PR #26, opened 2020-03-09
Hive - Topology driver

pradovic:
Points I tried to achieve:
1. `Discovery`, `Topology` interfaces not to know about multiaddr, the `addressbook` is used to fetch multiaddress in concrete implementations if needed
2. Extract business logic from `Hive` to rest of the system
3. Don't put specific logic into `libp2p`
4. Cover both known and connected peer added with/without connect 

I decided to do on a bit of a handler heavy approach, as I tried to do it with subscription and channels, and this seems much more simple. Wanted to cover full connectivity simple case, we can expand if needed.

Also, unrelated to the hive,  I did a few renaming along the way, @acud I renamed `GetterPutter` per Janoses comment on your PR that came a bit late, and I was guilty for the name  `GetterPutter`, I think `GetPutter` is more Go idiomatic.

Known issues:
1. Bootnodes - there is a conversation below
2. Connections between nodes at the same time - handshake error

### DIFF
--- a/pkg/debugapi/peer.go
+++ b/pkg/debugapi/peer.go
@@ -37,7 +37,11 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 
 	s.Addressbook.Put(address, addr)
 	if err := s.TopologyDriver.AddPeer(r.Context(), address); err != nil {
+		_ = s.P2P.Disconnect(address)
 		s.Logger.Debugf("debug api: topologyDriver.AddPeer %s: %v", addr, err)
+		s.Logger.Errorf("unable to connect to peer %s", addr)
+		jsonhttp.InternalServerError(w, err)
+		return
 	}
 
 	jsonhttp.OK(w, peerConnectResponse{
--- a/pkg/debugapi/peer.go
+++ b/pkg/debugapi/peer.go
@@ -6,6 +6,7 @@ package debugapi
 
 import (
 	"errors"
+	"fmt"
 	"net/http"
 
 	"github.com/ethersphere/bee/pkg/jsonhttp"
@@ -36,6 +37,7 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	s.Addressbook.Put(address, addr)
+	fmt.Printf("DEBUGLOG: debug api connect put address book peer %s, %s\n", address, addr)
 	if err := s.TopologyDriver.AddPeer(address); err != nil {
 		s.Logger.Debugf("debug api: topologyDriver.AddPeer %s: %v", addr, err)
 		s.Logger.Errorf("unable to connect to peer %s", addr)
--- a/pkg/debugapi/peer.go
+++ b/pkg/debugapi/peer.go
@@ -35,6 +35,14 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	s.Addressbook.Put(address, addr)
+	if err := s.TopologyDriver.AddPeer(address); err != nil {
+		s.Logger.Debugf("debug api: topologyDriver.AddPeer %s: %v", addr, err)
+		s.Logger.Errorf("unable to connect to peer %s", addr)
+		jsonhttp.InternalServerError(w, err)
+		return
+	}
+
 	jsonhttp.OK(w, peerConnectResponse{
 		Address: address.String(),
 	})
--- a/pkg/debugapi/peer.go
+++ b/pkg/debugapi/peer.go
@@ -38,9 +38,6 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 	s.Addressbook.Put(address, addr)
 	if err := s.TopologyDriver.AddPeer(r.Context(), address); err != nil {
 		s.Logger.Debugf("debug api: topologyDriver.AddPeer %s: %v", addr, err)
-		s.Logger.Errorf("unable to connect to peer %s", addr)
-		jsonhttp.InternalServerError(w, err)
-		return
 	}
 
 	jsonhttp.OK(w, peerConnectResponse{
--- a/pkg/debugapi/peer.go
+++ b/pkg/debugapi/peer.go
@@ -6,7 +6,6 @@ package debugapi
 
 import (
 	"errors"
-	"fmt"
 	"net/http"
 
 	"github.com/ethersphere/bee/pkg/jsonhttp"
@@ -37,7 +36,6 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	s.Addressbook.Put(address, addr)
-	fmt.Printf("DEBUGLOG: debug api connect put address book peer %s, %s\n", address, addr)
 	if err := s.TopologyDriver.AddPeer(address); err != nil {
 		s.Logger.Debugf("debug api: topologyDriver.AddPeer %s: %v", addr, err)
 		s.Logger.Errorf("unable to connect to peer %s", addr)
--- a/pkg/debugapi/peer.go
+++ b/pkg/debugapi/peer.go
@@ -36,7 +36,7 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	s.Addressbook.Put(address, addr)
-	if err := s.TopologyDriver.AddPeer(address); err != nil {
+	if err := s.TopologyDriver.AddPeer(r.Context(), address); err != nil {
 		s.Logger.Debugf("debug api: topologyDriver.AddPeer %s: %v", addr, err)
 		s.Logger.Errorf("unable to connect to peer %s", addr)
 		jsonhttp.InternalServerError(w, err)
--- a/pkg/discovery/mock/mock.go
+++ b/pkg/discovery/mock/mock.go
@@ -30,7 +30,9 @@ func (d *Discovery) BroadcastPeers(ctx context.Context, addressee swarm.Address,
 		d.mtx.Unlock()
 	}
 
+	d.mtx.Lock()
 	d.ctr++
+	d.mtx.Unlock()
 	return nil
 }
 
--- a/pkg/discovery/mock/mock.go
+++ b/pkg/discovery/mock/mock.go
@@ -14,23 +14,23 @@ import (
 type Discovery struct {
 	mtx     sync.Mutex
 	ctr     int //how many ops
-	records map[string]swarm.Address
+	records map[string][]swarm.Address
 }
 
 func NewDiscovery() *Discovery {
 	return &Discovery{
-		records: make(map[string]swarm.Address),
+		records: make(map[string][]swarm.Address),
 	}
 }
 
 func (d *Discovery) BroadcastPeers(ctx context.Context, addressee swarm.Address, peers ...swarm.Address) error {
 	for _, peer := range peers {
 		d.mtx.Lock()
-		d.ctr++
-		d.records[addressee.String()] = peer
+		d.records[addressee.String()] = append(d.records[addressee.String()], peer)
 		d.mtx.Unlock()
 	}
 
+	d.ctr++
 	return nil
 }
 
@@ -40,12 +40,12 @@ func (d *Discovery) Broadcasts() int {
 	return d.ctr
 }
 
-func (d *Discovery) AddresseeRecord(addressee swarm.Address) (overlay swarm.Address, exists bool) {
+func (d *Discovery) AddresseeRecords(addressee swarm.Address) (peers []swarm.Address, exists bool) {
 	d.mtx.Lock()
 	defer d.mtx.Unlock()
 	rec, exists := d.records[addressee.String()]
 	if !exists {
-		return swarm.Address{}, false
+		return []swarm.Address{}, false
 	}
 	return rec, true
 }
--- a/pkg/discovery/mock/mock.go
+++ b/pkg/discovery/mock/mock.go
@@ -43,9 +43,6 @@ func (d *Discovery) Broadcasts() int {
 func (d *Discovery) AddresseeRecords(addressee swarm.Address) (peers []swarm.Address, exists bool) {
 	d.mtx.Lock()
 	defer d.mtx.Unlock()
-	rec, exists := d.records[addressee.String()]
-	if !exists {
-		return []swarm.Address{}, false
-	}
-	return rec, true
+	peers, exists = d.records[addressee.String()]
+	return
 }
--- a/pkg/discovery/mock/mock.go
+++ b/pkg/discovery/mock/mock.go
@@ -8,29 +8,26 @@ import (
 	"context"
 	"sync"
 
-	"github.com/ethersphere/bee/pkg/discovery"
 	"github.com/ethersphere/bee/pkg/swarm"
-
-	ma "github.com/multiformats/go-multiaddr"
 )
 
 type Discovery struct {
 	mtx     sync.Mutex
 	ctr     int //how many ops
-	records map[string]discovery.BroadcastRecord
+	records map[string]swarm.Address
 }
 
 func NewDiscovery() *Discovery {
 	return &Discovery{
-		records: make(map[string]discovery.BroadcastRecord),
+		records: make(map[string]swarm.Address),
 	}
 }
 
-func (d *Discovery) BroadcastPeers(ctx context.Context, addressee swarm.Address, peers ...discovery.BroadcastRecord) error {
+func (d *Discovery) BroadcastPeers(ctx context.Context, addressee swarm.Address, peers ...swarm.Address) error {
 	for _, peer := range peers {
 		d.mtx.Lock()
 		d.ctr++
-		d.records[addressee.String()] = discovery.BroadcastRecord{Overlay: peer.Overlay, Addr: peer.Addr}
+		d.records[addressee.String()] = peer
 		d.mtx.Unlock()
 	}
 
@@ -43,12 +40,12 @@ func (d *Discovery) Broadcasts() int {
 	return d.ctr
 }
 
-func (d *Discovery) AddresseeRecord(addressee swarm.Address) (overlay swarm.Address, addr ma.Multiaddr) {
+func (d *Discovery) AddresseeRecord(addressee swarm.Address) (overlay swarm.Address, exists bool) {
 	d.mtx.Lock()
 	defer d.mtx.Unlock()
 	rec, exists := d.records[addressee.String()]
 	if !exists {
-		return swarm.Address{}, nil
+		return swarm.Address{}, false
 	}
-	return rec.Overlay, rec.Addr
+	return rec, true
 }
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -28,9 +28,10 @@ const (
 )
 
 type Service struct {
-	streamer    p2p.Streamer
-	addressBook addressbook.GetPutter
-	logger      logging.Logger
+	streamer     p2p.Streamer
+	addressBook  addressbook.GetPutter
+	peerHandlers []func(addr swarm.Address)
+	logger       logging.Logger
 }
 
 type Options struct {
@@ -74,6 +75,10 @@ func (s *Service) BroadcastPeers(ctx context.Context, addressee swarm.Address, p
 	}
 
 	return nil
+}
+
+func (s *Service) AddPeerHandler(h func(addr swarm.Address)) {
+	s.peerHandlers = append(s.peerHandlers, h)
 }
 
 func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swarm.Address) error {
@@ -122,6 +127,10 @@ func (s *Service) peersHandler(peer p2p.Peer, stream p2p.Stream) error {
 		}
 
 		s.addressBook.Put(swarm.NewAddress(newPeer.Overlay), addr)
+
+		for _, h := range s.peerHandlers {
+			h(swarm.NewAddress(newPeer.Overlay))
+		}
 	}
 
 	return nil
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -30,7 +30,7 @@ const (
 type Service struct {
 	streamer     p2p.Streamer
 	addressBook  addressbook.GetPutter
-	peerHandlers []func(addr swarm.Address)
+	peerHandlers []func(swarm.Address)
 	logger       logging.Logger
 }
 
@@ -104,6 +104,8 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
 		})
 	}
 
+	s.logger.Infof("sending peer request to peer %s,  req %s", peer, peersRequest)
+
 	if err := w.WriteMsg(&peersRequest); err != nil {
 		return fmt.Errorf("write Peers message: %w", err)
 	}
@@ -113,11 +115,14 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
 
 func (s *Service) peersHandler(peer p2p.Peer, stream p2p.Stream) error {
 	defer stream.Close()
+
 	_, r := protobuf.NewWriterAndReader(stream)
 	var peersReq pb.Peers
 	if err := r.ReadMsgWithTimeout(messageTimeout, &peersReq); err != nil {
 		return fmt.Errorf("read requestPeers message: %w", err)
 	}
+
+	s.logger.Infof("received peer request from peer %s,  req %s", peer, peersReq)
 
 	for _, newPeer := range peersReq.Peers {
 		addr, err := ma.NewMultiaddr(newPeer.Underlay)
@@ -127,7 +132,6 @@ func (s *Service) peersHandler(peer p2p.Peer, stream p2p.Stream) error {
 		}
 
 		s.addressBook.Put(swarm.NewAddress(newPeer.Overlay), addr)
-
 		for _, h := range s.peerHandlers {
 			h(swarm.NewAddress(newPeer.Overlay))
 		}
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -30,7 +30,7 @@ const (
 type Service struct {
 	streamer    p2p.Streamer
 	addressBook addressbook.GetPutter
-	peerHandler func(swarm.Address) error
+	peerHandler func(context.Context, swarm.Address) error
 	logger      logging.Logger
 }
 
@@ -77,7 +77,7 @@ func (s *Service) BroadcastPeers(ctx context.Context, addressee swarm.Address, p
 	return nil
 }
 
-func (s *Service) SetPeerAddedHandler(h func(addr swarm.Address) error) {
+func (s *Service) SetPeerAddedHandler(h func(ctx context.Context, addr swarm.Address) error) {
 	s.peerHandler = h
 }
 
@@ -128,7 +128,7 @@ func (s *Service) peersHandler(peer p2p.Peer, stream p2p.Stream) error {
 
 		s.addressBook.Put(swarm.NewAddress(newPeer.Overlay), addr)
 		if s.peerHandler != nil {
-			if err := s.peerHandler(swarm.NewAddress(newPeer.Overlay)); err != nil {
+			if err := s.peerHandler(context.Background(), swarm.NewAddress(newPeer.Overlay)); err != nil {
 				return err
 			}
 		}
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -88,7 +88,6 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
 	}
 
 	defer stream.Close()
-
 	w, _ := protobuf.NewWriterAndReader(stream)
 	var peersRequest pb.Peers
 	for _, p := range peers {
@@ -103,8 +102,6 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
 			Underlay: addr.String(),
 		})
 	}
-
-	s.logger.Infof("sending peer request to peer %s,  req %s", peer, peersRequest)
 
 	if err := w.WriteMsg(&peersRequest); err != nil {
 		return fmt.Errorf("write Peers message: %w", err)
@@ -121,8 +118,6 @@ func (s *Service) peersHandler(peer p2p.Peer, stream p2p.Stream) error {
 	if err := r.ReadMsgWithTimeout(messageTimeout, &peersReq); err != nil {
 		return fmt.Errorf("read requestPeers message: %w", err)
 	}
-
-	s.logger.Infof("received peer request from peer %s,  req %s", peer, peersReq)
 
 	for _, newPeer := range peersReq.Peers {
 		addr, err := ma.NewMultiaddr(newPeer.Underlay)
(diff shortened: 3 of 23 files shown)


janos:
> Bootnode spec - how should they behave in retrospect to topology?

Bootnodes should not contribute to the topology.

> How do we handle connection on a node that is already in the middle of a handshake (connecting). Example, node 2 is connecting to node1, but node 1 does a connect to node 2 in the middle. Rifght now one handshake will break, is this ok?

This should not be ok. We can treat this as the known issue if the handling is not trivial to add right now.

pradovic:
@janos , @acud One more questions about bootnodes. Can bootnodes connect to other nodes, or they are just being connected to. The reason for this question is that I would like to have an info about all connections if it's a bootnode or not, so I can decide weather to broadcast them or not. 
One example, you can get connection from another node, and if it can be a bootnode, you are actually not sure weather to put it in the topology or not. Also, we have a problem here when we get the list of known nodes. We don't know for sure if those are bootnodes or not.

pradovic:
@janos I will live the connection problem as known issue for now and maybe try to tackle this one in separate PR as it might be broader issue then hive related stuff. Sounds good?

janos:
> @janos , @acud One more questions about bootnodes. Can bootnodes connect to other nodes, or they are just being connected to. The reason for this question is that I would like to have an info about all connections if it's a bootnode or not, so I can decide weather to broadcast them or not.
> One example, you can get connection from another node, and if it can be a bootnode, you are actually not sure weather to put it in the topology or not. Also, we have a problem here when we get the list of known nodes. We don't know for sure if those are bootnodes or not.

That is a good question. Bootnodes do not have to broadcasted as they do not contribute to the topology. But it is important for them to have updated peers in address book beside the ones that they connect to. In that sense, it would be good to have them connected to other nodes to exchange peers over hive. In that way current bootnodes are kept up to date but they fill the spot in kademlia (or at least they used to do that). I have no concrete answer right now, except that they should not contribute to the topology, but they should somehow receive info about peers that are not only directly connecting to them.

janos:
> @janos I will live the connection problem as known issue for now and maybe try to tackle this one in separate PR as it might be broader issue then hive related stuff. Sounds good?

It sounds very good.

pradovic:
> > @janos , @acud One more questions about bootnodes. Can bootnodes connect to other nodes, or they are just being connected to. The reason for this question is that I would like to have an info about all connections if it's a bootnode or not, so I can decide weather to broadcast them or not.
> > One example, you can get connection from another node, and if it can be a bootnode, you are actually not sure weather to put it in the topology or not. Also, we have a problem here when we get the list of known nodes. We don't know for sure if those are bootnodes or not.
> 
> That is a good question. Bootnodes do not have to broadcasted as they do not contribute to the topology. But it is important for them to have updated peers in address book beside the ones that they connect to. In that sense, it would be good to have them connected to other nodes to exchange peers over hive. In that way current bootnodes are kept up to date but they fill the spot in kademlia (or at least they used to do that). I have no concrete answer right now, except that they should not contribute to the topology, but they should somehow receive info about peers that are not only directly connecting to them.

Hmmm, that's a very good input. I think I would then leave this for separate discussion and maybe even PR as well, cuz it might be an additionial thing to solve for bootnodes. I will think a bout it a bit more and let's discuss at some point. 

I think one solution could be not to initialize the same toplogy driver on bootnodes. We don't need a full connectivity driver, but just initialize the one that broadcastes peers. That way, they would receive new nodes when nodes connect to them (they would need to know about them), and they would broadcast new peers to those nodes.
On the other way, then it is easy to now that node you are connecting is bootnode (just because you now :) ), and we can have a flag if it's a bootnode or not, and just not broadcast bootnodes. Does this sounds reasonable as a first solution at least?

Also this goes well with another point that broadcast and connection strategy might be different partss of the topology driver, but that's a different and imo not urget topic, some amount of copy paste of code is just fine :)

janos:
> ... Does this sounds reasonable as a first solution at least?

Yes, it sounds as a very nice solution. Thanks.

pradovic:
> 
> 
> > ... Does this sounds reasonable as a first solution at least?
> 
> Yes, it sounds as a very nice solution. Thanks.

Thanks for the input! :) 🙏

pkg/debugapi/peer.go
```diff
@@ -35,6 +35,14 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	s.Addressbook.Put(address, addr)
```
acud: didn't way say that the topology driver will actually insert the entries to the address book?

pkg/debugapi/peer.go
```diff
@@ -35,6 +35,14 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
...
+	if err := s.TopologyDriver.AddPeer(address); err != nil {
```
acud: i wonder if we should add the flag forcing the connection to the `AddPeer` signature already. this is relevant to the debug api behavior. although it is true that right now this functionality will not really do anything, so, at your discretion

pkg/hive/hive.go
```diff
@@ -77,20 +77,29 @@ func (s *Service) BroadcastPeers(ctx context.Context, addressee swarm.Address, p
 	return nil
 }
 
-func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []discovery.BroadcastRecord) error {
+func (s *Service) AddPeerAddedHandler(h func(addr swarm.Address)) {
+	s.peerHandlers = append(s.peerHandlers, h)
```
acud: since there's only one handler right now, i think that `peerHandlers` can be just a `func(swarm.Address)` rather than a collection, then just set the value of the pluggable function after constructing the instance

pkg/hive/hive.go
```diff
@@ -116,8 +126,10 @@ func (s *Service) peersHandler(peer p2p.Peer, stream p2p.Stream) error {
 			continue
 		}
 
-		// todo: this might be changed depending on where do we decide to connect peers
 		s.addressBook.Put(swarm.NewAddress(newPeer.Overlay), addr)
+		for _, h := range s.peerHandlers {
```
acud: not needed if `peerHandlers` is just a pluggable function

pkg/node/node.go
```diff
@@ -92,17 +97,44 @@ func NewBee(o Options) (*Bee, error) {
...
+	// TODO: be more resilient on connection errors and connect in parallel
+	for _, a := range o.Bootnodes {
+		addr, err := ma.NewMultiaddr(a)
+		if err != nil {
+			return nil, fmt.Errorf("bootnode %s: %w", a, err)
+		}
+
+		overlay, err := p2ps.Connect(p2pCtx, addr)
+		if err != nil {
+			return nil, fmt.Errorf("connect to bootnode %s %s: %w", a, overlay, err)
```
acud: this is kind of tricky. if you can't connect to the bootnodes for some reason this does not necessarily infer that the node should be taken down.
i see a few options here, speaking purely from a usability perspective:
- if the bootnodes flag was supplied by the cli explicitly and the connection to all of them failed - then return an error (we must have this functionality for proper cluster bootstrapping with k8s i believe). you can enforce this with an `ErrorGroup`.
- if no cli flag was supplied for now we should never return an error here

pkg/p2p/p2p.go
```diff
@@ -14,15 +14,20 @@ import (
...
-	Disconnect(overlay swarm.Address) error
-	Peers() []Peer
+	Connecter
 }
 
 type Streamer interface {
 	NewStream(ctx context.Context, address swarm.Address, protocol, version, stream string) (Stream, error)
 }
 
+type Connecter interface {
```
acud: This doesn't really sit well with go interface naming conventions. `Connecter` would be a name for an interface that defines only a `Connect` method. Service might be better in this case

pkg/topology/full/full.go
```diff
@@ -20,81 +21,111 @@ func init() {
...
+		// update addr if it is wrong or it has been changed
+		if !addr.Equal(peerAddr) {
+			addr = peerAddr
+			d.addressBook.Put(peerAddr, ma)
+		}
+	}
+
+	connectedAddrs := []swarm.Address{}
+	for _, addressee := range connectedPeers {
+		// skip curenttly added peer
```
acud: `// skip newly added peer`

pkg/topology/full/full.go
```diff
@@ -20,81 +21,111 @@ func init() {
...
 
-	var connectedNodes []discovery.BroadcastRecord
-	for _, addressee := range d.connected {
-		cma, exists := d.addressBook.Get(addressee)
-		if !exists {
-			return topology.ErrNotFound
-		}
-
-		err := d.discovery.BroadcastPeers(context.Background(), addressee, discovery.BroadcastRecord{Overlay: overlay, Addr: ma})
+	if !isConnected(addr, connectedPeers) {
```
acud: `isConnected` is used only once. maybe inline the logic here?

pkg/topology/full/full.go
```diff
@@ -20,81 +21,111 @@ func init() {
...
-			return v, nil
+			return v.Address, nil
 		}
 		i++
 	}
 
 	return swarm.Address{}, topology.ErrNotFound
 }
+
+func (d *Driver) PeerAddedHandler(addr swarm.Address) {
```
acud: Why do we need this?

pkg/node/node.go
janos: This code is copied from libp2p package that I've written earlier. I think that it is outside of the scope of this PR and there already is a comment a few lines above.

> if no cli flag was supplied for now we should never return an error here

This is true as if Bootnodes slice is empty, this code will never be executed.

This package has nothing to do with cli flags. There are only passed options to the constructor. Just a note on the scope of this package, even it is used in cmd.

pkg/topology/full/full.go
janos: I think that isConnected is a nice function to have and makes the code in this function easier to read.

pkg/hive/hive.go
pradovic: Correct, I did it like this for libp2p as I don't expect much handlers, but here I just put it to illustrate the idiom. I can switch back to single one, and we can add more if needed later

pkg/debugapi/peer.go
pradovic: It's explained a bit in the desc of the PR. You are correct, we did say this, but when I was implementing I remembered that we also said on some meetings that we would like to avoid the mutliaddr in the interface of the topology and hive, so addressbook is here only as a collection of overlay->multiaddr pairs. Does this makes sense to you? @janos as well. If not, then we don't actually even need addressbook in the full connectivity implementation, but interface would know about mutliaddr, and also hive handler would have it as well. LTMWYT

pkg/debugapi/peer.go
pradovic: Again, related to the stuff above, depends if we want to expose multiaddr on topology interface or just implementation, it was related to the same decision. I took the liberty to do it a bit opinionated, but now when we see it, we can change if you guys think it's better the other way.

pkg/topology/full/full.go
pradovic: To avoid calling topology driver directly from hive. It's what we talked about a circular dependency, handelrs are one way to acomplish this. Right?

pradovic: Take a look at the node to see how all the parts (hive, topology driver etc..) are initialized and connected.

pkg/debugapi/peer.go
janos: Yes, I would like to say it again, topology driver should not know about underlay address in any of its form. Topology driver should be responsible only of overlay topology.

acud: cool :+1:

pkg/hive/hive.go
acud: let's have one handler for now :+1:

pkg/node/node.go
acud: > This package has nothing to do with cli flags. There are only passed options to the constructor. Just a note on the scope of this package, even it is used in cmd.


this is why I wrote `speaking purely from a usability perspective`.

acud: anyway - ok. let's keep as is

pkg/topology/full/full.go
acud: oops sorry is missed that.
do we have to have this in a separate goroutine though? is this not premature? or does this break tests? deadlocks? please elaborate

acud: i wonder if this is actually needed as an explicit method tbh.
you could just inject the handler in the when registering like so:
```go
	p2ps.SetPeerAddedHandler(func(addr swarm.Address){ topologyDriver.AddPeer(addr) })
```

acud: or even nicer, change the signatures that you can use `AddPeer` directly without having to wrap it

pradovic: Hmmm I have 2 points about this, let me know what you think:
1. I think it is a bad practice to block the handler of a different service
2. Doing it as a function in node makes node know about business logic, which I also don't like

What is a problem with having a special goroutine for a simple implementation, and a special function? Any downsides?

pradovic: The signature is not the same, but similar just because I don't actually want to handle error from handlers in hive, I think it adds no value, but that can be done as well. My goal was to separate this.

acud: Well, if you adjust the signature of the function that you're expecting to be `func(swarm.Address) error` and then just directly use `AddPeer` in `SetPeerAddedHandler` (but then the `SetHandler` naming should probably also change).

About goroutines, well, it introduces concurrency which is not necessarily trivial and could lead to flaky tests and certain data races, especially when the components work with the underlying data structures, so obviously we should make sure that the necessary changes are made to the address book first for example, but this is not so clear from the actual code (and someone else tampering with this code might not see this at first glance), and when we add more indirection and more concurrency this becomes less and less clear

pradovic: @acud , wrapping one line of code was done because our linter complains if function is just called with `go` and error is not handled. I personally don't like it either :)

pradovic: And in the more complex implementation (see todo comment) we could add various things here.

janos: I think that it is ok to have a todo comment about what it needs to be improved in the next iteration. And yes, the linter warning is handled making it obvious that the error and goroutine is yet to be handled. It is still needed to be tested where it should be improved.

acud: the questions was also if we need this invoked in a different goroutine @janos

pradovic: Just my 2 c, making handler blocking could also lead to additional problem in tests, because hive handler that receives messages is also async by nature. And I would not try to solve the problem for long blocking handlers, but for async handlers cuz it looks to me that this is what we would need.

pradovic: Since this is a first implementation, I don't think it makes sense to argue too much about it now, and both solution are good nenouigh, so I will switch to a blocking implementation and your suggestion @acud

janos: Yes, let's try the blocking implementation and have it async if we experience problems.

acud: thanks guys

pkg/discovery/mock/mock.go
```diff
@@ -43,12 +40,12 @@ func (d *Discovery) Broadcasts() int {
...
+func (d *Discovery) AddresseeRecords(addressee swarm.Address) (peers []swarm.Address, exists bool) {
 	d.mtx.Lock()
 	defer d.mtx.Unlock()
 	rec, exists := d.records[addressee.String()]
 	if !exists {
-		return swarm.Address{}, nil
+		return []swarm.Address{}, false
 	}
-	return rec.Overlay, rec.Addr
+	return rec, true
```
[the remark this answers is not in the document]
pradovic: Hmm, I think it's not possible because it's `comma ok` idiom and compiler is complaining because there is not two parameters. But I can do this: 
``` peers, exists = d.records[addressee.String()]
	return ```

File: pkg/topology/full/full.go [hunk 11]
```diff
@@ -20,81 +21,107 @@ func init() {
...
-type driver struct {
-	mtx         sync.Mutex
-	connected   map[string]swarm.Address
-	discovery   discovery.Driver
-	addressBook addressbook.Getter
+type Driver struct {
+	discovery     discovery.Driver
+	addressBook   addressbook.GetPutter
+	connecter     p2p.Service
+	receivedPeers map[string]bool
```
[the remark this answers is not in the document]
pradovic: Nice, I will switch to `struct{}` and add comment.  I was also thinking about the infinite grow, but I did not go into it because it is probably not going to be a production solution, but you are right, there should be a comment at least, and let's add expiration or cleanup if/when needed!

pkg/topology/full/full.go
```diff
@@ -20,81 +21,107 @@ func init() {
...
-func (d *driver) AddPeer(overlay swarm.Address) error {
-	d.mtx.Lock()
-	defer d.mtx.Unlock()
+func (d *Driver) AddPeer(addr swarm.Address) error {
+	if d.receivedPeers[addr.ByteString()] {
+		return nil
+	}
 
-	ma, exists := d.addressBook.Get(overlay)
+	d.receivedPeers[addr.ByteString()] = true
```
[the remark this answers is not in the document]
pradovic: I agree, nice catch!

pkg/discovery/mock/mock.go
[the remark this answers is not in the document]
janos: Ah, yes, you are right, thanks.

pradovic:
> Very nice. 👏 Comments are mostly for pedantic purposes.

All comments made sense to me. Fixed. Thanks!

pkg/hive/hive_test.go
```diff
@@ -153,12 +153,12 @@ func TestBroadcastPeers(t *testing.T) {
...
 
-		if !compareOverlays(exporter.Overlays(), tc.wantKeys) {
-			t.Errorf("Overlays got %v, want %v", exporter.Overlays(), tc.wantKeys)
+		if !compareOverlays(exporter.Overlays(), tc.wantOverlays) {
+			t.Errorf("Overlays got %v, want %v", exporter.Overlays(), tc.wantOverlays)
 		}
 
-		if !compareMultiaddrses(exporter.Multiaddresses(), tc.wantValues) {
-			t.Errorf("Multiaddresses got %v, want %v", exporter.Multiaddresses(), tc.wantValues)
+		if !compareMultiaddrses(exporter.Multiaddresses(), tc.wantMultiAddrses) {
```
acud: `wantMultiAddresses`

pkg/topology/full/full.go
```diff
@@ -20,81 +21,107 @@ func init() {
...
-func New(disc discovery.Driver, addressBook addressbook.Getter) topology.Driver {
-	return &driver{
-		connected:   make(map[string]swarm.Address),
-		discovery:   disc,
-		addressBook: addressBook,
+func New(disc discovery.Driver, addressBook addressbook.GetPutter, connecter p2p.Service, logger logging.Logger) *Driver {
+	return &Driver{
+		discovery:     disc,
+		addressBook:   addressBook,
+		connecter:     connecter,
```
acud: maybe some other name other than `connecter` because `Connecter` interface does no longer exist

pkg/topology/full/full.go
```diff
@@ -20,81 +21,107 @@ func init() {
...
 }
 
 // AddPeer adds a new peer to the topology driver.
 // The peer would be subsequently broadcasted to all connected peers.
 // All conneceted peers are also broadcasted to the new peer.
-func (d *driver) AddPeer(overlay swarm.Address) error {
-	d.mtx.Lock()
-	defer d.mtx.Unlock()
+func (d *Driver) AddPeer(ctx context.Context, addr swarm.Address) error {
+	if _, ok := d.receivedPeers[addr.ByteString()]; ok {
```
acud: is it not necessary to use a lock before accessing the map?

pkg/topology/full/full.go
```diff
@@ -20,81 +21,107 @@ func init() {
...
 
-	err := d.discovery.BroadcastPeers(context.Background(), overlay, connectedNodes...)
-	if err != nil {
+	if err := d.discovery.BroadcastPeers(context.Background(), addr, connectedAddrs...); err != nil {
 		return err
 	}
 
-	// add new node to connected nodes to avoid double broadcasts
-	d.connected[overlay.String()] = overlay
+	d.receivedPeers[addr.ByteString()] = struct{}{}
```
acud: lock?

pkg/topology/full/full_test.go
```diff
@@ -0,0 +1,185 @@
...
+		addressbook := inmem.New()
+		p2p := p2pmock.New(p2pmock.WithConnectFunc(func(ctx context.Context, addr ma.Multiaddr) (swarm.Address, error) {
+			t.Fatal("should not be called")
+			return swarm.Address{}, nil
+		}))
+
+		fullDriver := full.New(discovery, addressbook, p2p, logger)
+		err := fullDriver.AddPeer(context.Background(), overlay)
+		if !errors.Is(err, topology.ErrNotFound) {
+			t.Fatalf("full conn driver returned err %s", err.Error())
```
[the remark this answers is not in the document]
pradovic: Yup, correct, nice catch!

pkg/topology/topology.go
```diff
@@ -13,10 +14,14 @@ import (
 var ErrNotFound = errors.New("no peer found")
 
 type Driver interface {
-	AddPeer(overlay swarm.Address) error
+	Peerer
 	ChunkPeerer
 }
 
+type Peerer interface {
```
acud: I think `AddPeerer` is more in line with what the interface does

pkg/hive/hive.go
```diff
@@ -103,6 +112,7 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []dis
 
 func (s *Service) peersHandler(peer p2p.Peer, stream p2p.Stream) error {
 	defer stream.Close()
+
```
acud: not necessary to add whitespace

pkg/p2p/libp2p/libp2p.go
```diff
@@ -191,24 +187,18 @@ func New(ctx context.Context, o Options) (*Service, error) {
...
-		}
-
-		overlay, err := s.Connect(ctx, addr)
-		if err != nil {
-			return nil, fmt.Errorf("connect to bootnode %s %s: %w", a, overlay, err)
-		}
-	}
-
-	h.Network().SetConnHandler(func(_ network.Conn) {
+	h.Network().SetConnHandler(func(c network.Conn) {
```
acud: you don't need to declare `c`; `_` will do

pkg/p2p/libp2p/libp2p.go
```diff
@@ -326,6 +309,10 @@ func (s *Service) Peers() []p2p.Peer {
 	return s.peers.peers()
 }
 
+func (s *Service) SetPeerAddedHandler(h func(ctx context.Context, addr swarm.Address) error) {
```
acud: no need to declare `ctx` and `addr`, just the types would do. nicer would be to actually have the signature as a type, and just use that around since it is used in several places

pkg/topology/full/full_test.go
```diff
@@ -0,0 +1,185 @@
...
+		},
+		{
+			Address: swarm.MustParseHexAddress("ca1e9f3938cc1425c6061b96ad9eb93e134dfe8734ad490164ef20af9d1cf59d"),
+		},
+	}
+
+	t.Run("OK - no connected peers", func(t *testing.T) {
+		discovery := mock.NewDiscovery()
+		addressbook := inmem.New()
+		p2p := p2pmock.New(p2pmock.WithConnectFunc(func(ctx context.Context, addr ma.Multiaddr) (swarm.Address, error) {
```
acud: `_` for `ctx`

pkg/discovery/mock/mock.go
```diff
@@ -8,32 +8,29 @@ import (
...
+func (d *Discovery) BroadcastPeers(ctx context.Context, addressee swarm.Address, peers ...swarm.Address) error {
 	for _, peer := range peers {
 		d.mtx.Lock()
-		d.ctr++
-		d.records[addressee.String()] = discovery.BroadcastRecord{Overlay: peer.Overlay, Addr: peer.Addr}
+		d.records[addressee.String()] = append(d.records[addressee.String()], peer)
 		d.mtx.Unlock()
 	}
 
+	d.ctr++
```
acud: this should probably also be synchronized with a lock

pkg/p2p/mock/mock.go
```diff
@@ -44,6 +45,12 @@ func WithPeersFunc(f func() []p2p.Peer) Option {
 	})
 }
 
+func WithSetPeerAddedHandlerFunc(f func(func(ctx context.Context, overlay swarm.Address) error)) Option {
```
acud: nice to have: extract the function signature to a single definition that you can reuse

pkg/topology/mock/mock.go
```diff
@@ -0,0 +1,37 @@
...
+func (d *TopologyDriver) SetAddPeerErr(err error) {
+	d.addPeerErr = err
+}
+
+func (d *TopologyDriver) AddPeer(_ context.Context, addr swarm.Address) error {
+	if d.addPeerErr != nil {
+		return d.addPeerErr
+	}
+
+	d.peers = append(d.peers, addr)
```
acud: i wonder if we need a lock here

pkg/topology/topology.go
janos: Or `PeerAdder` ;)

pradovic: Hmm, I am not sure about this. Aren't we adding the peer to the toplogy? And `Peerer` for me is more like a word instead of something like `PeerAdder`. Could you explain, pls?

pradovic: Hahaha, @janos , same time :)

pkg/topology/mock/mock.go
pradovic: I did not expect to use it concurrently now, but I can add it if you think it might be useful for the future as well?

pkg/topology/topology.go
acud: Well, according to go conventions, an interface of name `Peerer` would infer a single method whose name would be `Peer()` or potentially `Peers()`. `PeerAdder` then, is better description of what we're passing around

pkg/debugapi/peer.go
```diff
@@ -38,9 +38,6 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 	s.Addressbook.Put(address, addr)
 	if err := s.TopologyDriver.AddPeer(r.Context(), address); err != nil {
 		s.Logger.Debugf("debug api: topologyDriver.AddPeer %s: %v", addr, err)
-		s.Logger.Errorf("unable to connect to peer %s", addr)
-		jsonhttp.InternalServerError(w, err)
```
acud: why is this now removed?

pkg/topology/mock/mock.go
acud: well, if we are mutating a slice then i think it might be a good idea. otherwise we would have to enforce that any test that uses the mock cannot introduce any concurrency at all, which would be difficult on the long run

pkg/topology/full/full_test.go
```diff
@@ -0,0 +1,185 @@
...
+		// check other nodes
+		for _, p := range connectedPeers {
+			if err := checkAddreseeRecords(discovery, p.Address, []p2p.Peer{{Address: overlay}}); err != nil {
+				t.Fatal(err)
+			}
+		}
+	})
+}
+
+func checkAddreseeRecords(discovery *mock.Discovery, addr swarm.Address, expected []p2p.Peer) error {
```
acud: you can change the signature to also accept `t *testing.T`, then declare `t.Helper()` at the head of the function, and then you can directly `t.Fatal` or `t.Errorf` inside the helper

pkg/topology/topology.go
pradovic: Ok, `PeerAdder` it is :)

pkg/debugapi/peer.go
```diff
@@ -35,6 +35,11 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	s.Addressbook.Put(address, addr)
+	if err := s.TopologyDriver.AddPeer(r.Context(), address); err != nil {
+		s.Logger.Debugf("debug api: topologyDriver.AddPeer %s: %v", addr, err)
```
acud: why doesn't the API return an error in this case?

pradovic: Oh, I thought you wanted it to be debug level? Error level message is the one visible to the user and I left it as it was. But then on the second thought I thought that it should not be an error, or we should cleanup and disconnect in this case. I don't know if we wanna go there at this point. LMNWYT? @janos as well as he was working on `debugapi`?

pkg/topology/full/full_test.go
pradovic: I feel like this makes more room to use this functions in different places. Do you think it has any downsides?

pkg/topology/mock/mock.go
pradovic: Done

pkg/debugapi/peer.go
pradovic: Basically, should we fail in this case, and how should we handle it. I am not sure what should be the best behaviour in the case of debug API.

acud: I meant the logging should be done on the debug level, but I think it would be nice to know if the requested debug api call failed for whatever reason. In this case it would just fail silently and indicate nothing both to the debug api caller nor to the user (unless the user has logging in debug level on).
We should be just careful about how we use the logger, with the understanding on who is reading which log levels, that is all. I did not mean that the api should not fail in this case

janos: This certainly need to be responded with the 500, and I am not sure why the error call is removed (conversation above). I think that it was just right initially.

acud: @janos, originally it was logging both on `error` and `debug` loglevels, this was changed here https://github.com/ethersphere/bee/pull/26/commits/a2034d29d5655cbe943636c73f8ca8c8e89fb470. i claim we should only log on debug (since this is a developer message, not message intended for the end user), and return a `500`, that is all.

pradovic: My point is, that if we return an error message in this case, like it was in the first place, the user would still be connected. Should we disconnect in this case? 
@acud I think that the pattern is:
Debug: message for the developer
Error: user facing message

That is why there was an detailed message in the debug, and more context based - connect failed in the error message

janos: Yes, I know, I think that both lines should be there, so that error message is detectable when this error happens if only error log level severity is set.

pradovic: So, no need to disconnect and stuff, just return an error to do user? I think that connect will still work in this case.

acud: I'm not sure what you mean by `user would still be connected` and what do you mean by disconnecting... if the `AddPeer` call failed I would not expect the peer to be connected and of the underlying data structures to mutate, so appropriate cleanups and rollbacks should be done. It might be possible to do a cleanup on the address book entry (which cannot be used), but that is optional

pradovic: If you take a look at the code above, we have a `p2p.Connect` call, and then we add stuff to the addressbook and add to the topology driver. So my question is, if adding to the topology driver fails, what should happen? Should we disconnect this peer? I am not sure how we should handle this 100% correct.

acud: aha, yes, should disconnect.

pradovic: Done :)

pradovic:
@acud No problem! Thanks you guys for the reviews!

PR #27, opened 2020-03-13
Stream headers and Tracing

janos:
This PR:

- adds stream headers concept by passing request headers to NewStream function, getting the response headers with Stream.Headers and processing them with protocols Headler function
- adds tracing by passing span context with headers
- adds an example tracing for pingpong protocol and api
- changes signature of NewStream which length of arguments should be addressed in the followup PR
- fixes missing address book and discovery driver instances in node package constructor

This PR is tested manually with jaeger server `docker run -p 6831:6831/udp -p 16686:16686 jaegertracing/all-in-one:latest` and with unit tests in code.

To test tracing manually:
- start the jaeger server with docker. access the jaeger web interface at http://localhost:16686/
- start locally two bee nodes (different data dirs and ports) and connect them as in https://github.com/ethersphere/bee#usage-experimental-api with `--tracing` flag provided for both nodes
- make a pingpong call over the exposed http api
- validate tracing in jaeger web interface


pkg/p2p/libp2p/headers_test.go
```diff
@@ -0,0 +1,132 @@
...
+	s1, overlay1, cleanup1 := newService(t, libp2p.Options{})
+	defer cleanup1()
+
+	s2, _, cleanup2 := newService(t, libp2p.Options{})
+	defer cleanup2()
+
+	var gotHeaders p2p.Headers
+	handled := make(chan struct{})
+	if err := s1.AddProtocol(newTestProtocol(func(_ context.Context, _ p2p.Peer, stream p2p.Stream) error {
+		gotHeaders = stream.Headers()
```
pradovic: Maybe it would be good to test if the input params are good in this kind of function? What do you think?

pkg/p2p/streamtest/streamtest.go
```diff
@@ -155,6 +161,10 @@ func (s *stream) Write(p []byte) (int, error) {
 	return s.in.Write(p)
 }
 
+func (s *stream) Headers() p2p.Headers {
+	return s.headers
```
pradovic: Should we add a method to set the headers for testing purposes maybe?

pkg/p2p/libp2p/headers.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	stream.headers = headersPBToP2P(headers)
+
+	var h p2p.Headers
+	if headler != nil {
+		h = headler(stream.headers)
+	}
+
+	if err := w.WriteMsgWithTimeout(time.Second, headersP2PToPB(h)); err != nil {
```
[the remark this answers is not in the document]
janos: Yes, this is deliberate as the receiving side expects to get a headers message first. If it reads the data from the stream it has to be predictable which message it should decode it to.

Nice catch that empty message test is missing. I will add it.

pkg/p2p/libp2p/headers_test.go
janos: You refer to the anonymous handler func here? That would be good to have, also. Will add checks on input parameters.

pradovic: Exactly, basically, just to check if the parameters received in function are good, if possible. So we know that it is called with right parameters.

pkg/p2p/streamtest/streamtest.go
janos: Currently, this `stream` is not exported and not used as the mock for the stream, but a recorder. They are set in NewStream if protocol has Headler. I would not add SetHeaders method until it is needed, as it would imply exporting the stream in this package or creating a new interface with that method. If that is ok?

pradovic: Yes, u r right, sorry, I did not see that 🙈  Sounds good!

janos:
Thanks for reviews, this one was not so trivial.

PR #32, opened 2020-03-31
api, storage: add http interface to accept chunks

acud:
This PR adds functionality to upload and download a chunk from `bee`.
Right now chunks are stored in-memory using a mock `storage.Storer` implementation.
Chunks are "verified" only through the test harness - i.e. it is ready to accept flows that would be generated from errors while putting the chunk, but does not expose which errors occurred (this will be done once the necessity will be shown).

Once localstore gets imported we can switch to that as storage engine. Same for BMT and chunk verification.
cc @jmozah @nolash

### DIFF
--- a/pkg/api/bzzchunk.go
+++ b/pkg/api/bzzchunk.go
@@ -5,10 +5,14 @@
 package api
 
 import (
+	"bytes"
+	"errors"
+	"io"
 	"io/ioutil"
 	"net/http"
 
 	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/gorilla/mux"
 )
@@ -40,4 +44,31 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	jsonhttp.OK(w, nil)
+}
+
+func (s *server) chunkGetHandler(w http.ResponseWriter, r *http.Request) {
+	addr := mux.Vars(r)["addr"]
+	ctx := r.Context()
+
+	address, err := swarm.ParseHexAddress(addr)
+	if err != nil {
+		s.Logger.Debugf("bzz-chunk: parse chunk address %s: %v", addr, err)
+		jsonhttp.BadRequest(w, "invalid chunk address")
+		return
+	}
+
+	data, err := s.Storer.Get(ctx, address)
+	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			jsonhttp.NotFound(w, "chunk not found")
+			return
+
+		}
+		s.Logger.Debugf("chunk read error: %v", err)
+		jsonhttp.InternalServerError(w, "chunk read error")
+		return
+	}
+	w.Header().Set("Content-Type", "binary/octet-stream")
+	w.WriteHeader(http.StatusOK)
+	io.Copy(w, bytes.NewReader(data))
 }
--- a/pkg/api/bzzchunk.go
+++ b/pkg/api/bzzchunk.go
@@ -70,7 +70,7 @@ func (s *server) chunkGetHandler(w http.ResponseWriter, r *http.Request) {
 			return
 
 		}
-		s.Logger.Debugf("bzz-chunk: chunk read error: %v", err)
+		s.Logger.Debugf("bzz-chunk: chunk read error: %v ,addr %s", err, address)
 		s.Logger.Error("bzz-chunk: chunk read error")
 		jsonhttp.InternalServerError(w, "chunk read error")
 		return
--- a/pkg/api/bzzchunk.go
+++ b/pkg/api/bzzchunk.go
@@ -70,5 +70,8 @@ func (s *server) chunkGetHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	w.Header().Set("Content-Type", "binary/octet-stream")
 	w.WriteHeader(http.StatusOK)
-	io.Copy(w, bytes.NewReader(data))
+	_, err = io.Copy(w, bytes.NewReader(data))
+	if err != nil {
+		s.Logger.Debugf("bzz-chunk: write to http writer: %v", err)
+	}
 }
--- a/pkg/api/bzzchunk.go
+++ b/pkg/api/bzzchunk.go
@@ -24,6 +24,7 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 	address, err := swarm.ParseHexAddress(addr)
 	if err != nil {
 		s.Logger.Debugf("bzz-chunk: parse chunk address %s: %v", addr, err)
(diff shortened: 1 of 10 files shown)


pkg/api/bzzchunk.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/gorilla/mux"
+)
+
+func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
+	addr := mux.Vars(r)["addr"]
+	ctx := r.Context()
+
+	address, err := swarm.ParseHexAddress(addr)
+	if err != nil {
+		s.Logger.Debugf("bzz-chunk: parse chunk address %s: %v", addr, err)
```
janos: Alongside these Debug log lines, I would add Error log lines that are for users who enable only Error log level, with messages that do not contain any internal implementation details. Debug log levels are for developers debugging internals.

pkg/api/bzzchunk.go
```diff
@@ -0,0 +1,77 @@
...
+		jsonhttp.BadRequest(w, "invalid chunk address")
+		return
+	}
+
+	data, err := s.Storer.Get(ctx, address)
+	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			jsonhttp.NotFound(w, "chunk not found")
```
janos: I think that trace log message that the chunk is not found would be useful here.

pkg/api/bzzchunk.go
```diff
@@ -0,0 +1,77 @@
...
+			return
+
+		}
+		s.Logger.Debugf("chunk read error: %v", err)
+		jsonhttp.InternalServerError(w, "chunk read error")
+		return
+	}
+	w.Header().Set("Content-Type", "binary/octet-stream")
+	w.WriteHeader(http.StatusOK)
```
janos: This line is not needed. Status OK is set implicitly.

pkg/api/bzzchunk_test.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		return true
+	}
+
+	mockValidatingStorer := mock.NewValidatingStorer(validatorF)
+	client, cleanup := newTestServer(t, testServerOptions{
+		Storer: mockValidatingStorer,
+	})
+
+	defer cleanup()
```
janos: Extremely pedantic comment: this defer should be moved one line up.

pkg/storage/mock/storer.go
```diff
@@ -23,6 +24,15 @@ func NewStorer() storage.Storer {
 	return s
 }
 
+func NewValidatingStorer(f storage.ChunkValidatorFunc) storage.Storer {
+	s := &mockStorer{
```
janos: It can be just `return &mockStorer{...`.

pkg/api/bzzchunk_test.go
acud: :D

pkg/api/bzzchunk.go
janos: Thanks for adding the Error log message. Could you add similar to the other places in these two new handlers? I am not sure if it is needed for io.Copy error, as that error is usually client or network related and without much significance to the server.

pkg/api/bzzchunk_test.go
```diff
@@ -0,0 +1,114 @@
...
+	})
+	defer cleanup()
+
+	t.Run("invalid hash", func(t *testing.T) {
+		jsonhttptest.ResponseDirect(t, client, http.MethodPost, resource(invalidHash), bytes.NewReader(validContent), http.StatusBadRequest, jsonhttp.StatusResponse{
+			Message: "chunk write error",
+			Code:    http.StatusBadRequest,
+		})
+
+		// make sure chunk not retrievable
```
pradovic, line 66:
```suggestion
		// make sure chunk is not retrievable
```

pkg/api/bzzchunk.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (s *server) chunkGetHandler(w http.ResponseWriter, r *http.Request) {
+	addr := mux.Vars(r)["addr"]
+	ctx := r.Context()
+
+	address, err := swarm.ParseHexAddress(addr)
+	if err != nil {
+		s.Logger.Debugf("bzz-chunk: parse chunk address %s: %v", addr, err)
+		s.Logger.Error("bzz-chunk: parse chunk address error")
```
pradovic, line 59: I would remove word error in error log maybe?

pkg/api/bzzchunk.go
```diff
@@ -0,0 +1,79 @@
...
+	data, err := s.Storer.Get(ctx, address)
+	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			s.Logger.Trace("bzz-chunk: chunk not found. addr %s", address)
+			jsonhttp.NotFound(w, "chunk not found")
+			return
+
+		}
+		s.Logger.Debugf("bzz-chunk: chunk read error: %v ,addr %s", err, address)
+		s.Logger.Error("bzz-chunk: chunk read error")
```
pradovic, line 73: Same here

pkg/api/bzzchunk.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
+	addr := mux.Vars(r)["addr"]
+	ctx := r.Context()
+
+	address, err := swarm.ParseHexAddress(addr)
+	if err != nil {
+		s.Logger.Debugf("bzz-chunk: parse chunk address %s: %v", addr, err)
+		s.Logger.Error("bzz-chunk: error uploading chunk")
```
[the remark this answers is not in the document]
acud: we agreed on not having too much info facing the end users that run with `error` loglevel, that's why there's duplicate logging on debug and error, providing the developer with the underlying reason for the failure

pkg/api/bzzchunk.go
```diff
@@ -0,0 +1,79 @@
...
+	"io/ioutil"
+	"net/http"
+
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/gorilla/mux"
+)
+
+func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
acud: cc @janos

pkg/api/bzzchunk_test.go
```diff
@@ -0,0 +1,114 @@
...
+			t.Fatal(err)
+		}
+
+		if !bytes.Equal(validContent, data) {
+			t.Fatal("data retrieved doesnt match uploaded content")
+		}
+	})
+}
+
+func request(t *testing.T, client *http.Client, method string, resource string, body io.Reader, responseCode int) *http.Response {
```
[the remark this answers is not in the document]
acud: i'd rather keep this here for now. past experience shows this kind of test helper doesn't get really reused all that well across packages. its not a big price to pay

PR #45, opened 2020-03-31
storage/disk: Added persistent storage

jmozah:
This PR adds the persistent storage (badger) to bee. 
This also add the shed implementation for future use.


pkg/swarm/swarm.go
```diff
@@ -83,3 +84,33 @@ func (a Address) MarshalJSON() ([]byte, error) {
 
 // ZeroAddress is the address that has no value.
 var ZeroAddress = NewAddress(nil)
+
+
+// Chunk defines a swarm chunk structure. It contains of the Address and the
+// chunk data.
+type Chunk struct {
```
Eknir: Maybe be explicit and call this an addressed chunk (don't want to review this now, but scanned over it and saw it). Good work so far! :)

Eknir:
Just a question, why did you decide to leave `storage/mem` in the code? Is this package still used somewhere? I see that a similar type of practice is done in `keystore`. There, we also have `keystore/mem` and `keystore/disk`. @janos, any good reason for this?

janos:
@Eknir I am not sure if this PR is ready for review. It looks to me that it is still in progress. Unused code should be removed, but not too early.

jmozah:
The tests are passing in my local setup, i don't know why they are failing here. 
Also the test cases that are failing are not part of this PR.

pkg/swarm/swarm.go
jmozah: Single Owner chunk will be called as SOC to avoid confusion. Normal chunk is chunk to maintain backward compatibiity.

jmozah:
> just a question, why did you decide to leave storage/mem in the code? Is this package still used somewhere? I see that a similar type of practice is done in keystore. There, we also have keystore/mem and keystore/disk. 

The reason for using a mem package in DB is to help other packages with DB dependency to test with memory impl. instead of the full fledged disk impl.  It is fast and simple to settup and teardown for each testcase.

PR #46, opened 2020-03-31
bootnode init improvement + libp2p incoming connection addr fix

pradovic:
* Bootnodes init in parallel, and does not fail if bootnode connect fails
* multiaddr creation fix in incoming connection in `libp2p`

@acud, @janos  
When testing I believe I found a problematic behavior which I believe is because the `AddPeer` handler is not async, or at least part of it. Imagine a scenario where you have nodes 0..10, and node 1 nodes about all other nodes. This can happen:
1. Node 2 connects to node 1
2. Node 1 will start to broadcast this node to all other nodes waiting for them to finish connect in order to finish the handle
3. Also, other nodes, will broadcast to other known nodes, in the handler, after connect

Every broadcast hive message is waiting for other side to close the stream, to finish communication, and this could result in lots of waits and timeouts. 
I would like to make `AddPeer` in topology driver async for 2 reasons:
1. It has a connect and broadcast calls which both wait for other nodes to confirm some information before exiting
2. I think that error handling of errors received from `AddPeer` is a bit hard to cover, because It might be different reasons for those errors and the code calling  `AddPeer` should probably not cancel the call if topology fails. 

For example, if one broadcast fails, or connect in the `AddPeer` fails, probably our incoming connection phase (after handshake) in `libp2p` should not actually fail and disconnect, etc... I think this one is very hard to cover. 
The other idea would be to identify parts the code that should async in handlers, but there are not much cuz the `AddPeer` handler is doing hive broadcasts and `libp2p` connect.

Other note, if we wanted some of this errors in `AddPeer` to disconnect the node, I would probably do it using Disconnect call in the handler itself other then wait for it in the calling function. Basically, I think that handlers should not be blocking, as they could block the whole chain and we can never know the implementations of those functions in the chain (like here we have `libp2p.Connect`  & `hive.Broadcast`)

LTKWYT

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -227,32 +227,31 @@ func NewBee(o Options) (*Bee, error) {
 		b.debugAPIServer = debugAPIServer
 	}
 
+	// Connect bootnodes
 	var wg sync.WaitGroup
-
-	// TODO: be more resilient on connection errors and connect in parallel
 	for _, a := range o.Bootnodes {
 		wg.Add(1)
 		go func(aa string) {
 			defer wg.Done()
 			addr, err := ma.NewMultiaddr(aa)
 			if err != nil {
-				logger.Debugf("multiaddress fail %s: %w", aa, err)
-				logger.Errorf("connect to bootnode %s: %w", aa, err)
+				logger.Debugf("multiaddress fail %s: %v", aa, err)
+				logger.Errorf("connect to bootnode %s", aa)
 				return
 			}
 
 			overlay, err := p2ps.Connect(p2pCtx, addr)
 			if err != nil {
-				logger.Debugf("connect fail %s: %w", aa, err)
-				logger.Errorf("connect to bootnode %s: %w", aa, err)
+				logger.Debugf("connect fail %s: %v", aa, err)
+				logger.Errorf("connect to bootnode %s", aa)
 				return
 			}
 
 			addressbook.Put(overlay, addr)
 			if err := topologyDriver.AddPeer(p2pCtx, overlay); err != nil {
 				_ = p2ps.Disconnect(overlay)
-				logger.Debugf("topology add peer fail %s %s: %w", aa, overlay, err)
-				logger.Errorf("connect to bootnode %s: %w", aa, err)
+				logger.Debugf("topology add peer fail %s %s: %v", aa, overlay, err)
+				logger.Errorf("connect to bootnode %s", aa)
 				return
 			}
 		}(a)
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -249,15 +249,12 @@ func NewBee(o Options) (*Bee, error) {
 			}
 
 			addressbook.Put(overlay, addr)
-			fmt.Printf("bootnode connected, added to addresbook %s, %s\n", addr, overlay)
 			if err := topologyDriver.AddPeer(p2pCtx, overlay); err != nil {
 				_ = p2ps.Disconnect(overlay)
 				logger.Debugf("topology add peer fail %s %s: %w", aa, overlay, err)
 				logger.Errorf("connect to bootnode %s: %w", aa, err)
 				return
 			}
-
-			fmt.Printf("bootnode connected, added to toplogy driver %s, %s\n", addr, overlay)
 		}(a)
 	}
 
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -12,6 +12,7 @@ import (
 	"net"
 	"net/http"
 	"path/filepath"
+	"sync"
 
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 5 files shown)


pkg/topology/full/full.go
```diff
@@ -56,6 +56,8 @@ func (d *Driver) AddPeer(ctx context.Context, addr swarm.Address) error {
 		d.mtx.Unlock()
 		return nil
 	}
+
+	d.receivedPeers[addr.ByteString()] = struct{}{}
```
pradovic, line 60: @janos I moved this back here, to avoid having multiple add peer calls being handled. I think it's either this or a global lock across the whole function? I think this is a better approach, and we can remove the map entry on errors if needed. LMKWYT?

pkg/node/node.go
```diff
@@ -239,6 +227,38 @@ func NewBee(o Options) (*Bee, error) {
...
+	var wg sync.WaitGroup
+
+	// TODO: be more resilient on connection errors and connect in parallel
+	for _, a := range o.Bootnodes {
+		wg.Add(1)
+		go func(aa string) {
+			defer wg.Done()
+			addr, err := ma.NewMultiaddr(aa)
+			if err != nil {
+				logger.Debugf("multiaddress fail %s: %w", aa, err)
```
janos, line 239: I think that verb `%w` is not well interpreted in logger. Should be `%v`.

This goes to other lines below.

pkg/node/node.go
```diff
@@ -239,6 +227,38 @@ func NewBee(o Options) (*Bee, error) {
...
+				logger.Errorf("connect to bootnode %s: %w", aa, err)
```
janos, line 240: I think that Error message should not expose the internal error, and that it may be better just to have "failed to connect to bootnode %s".

This goes to other lines below.

pkg/node/node.go
```diff
@@ -239,6 +227,38 @@ func NewBee(o Options) (*Bee, error) {
 		b.debugAPIServer = debugAPIServer
 	}
 
+	var wg sync.WaitGroup
+
+	// TODO: be more resilient on connection errors and connect in parallel
```
janos, line 232: This comment should be obsolete now.

pkg/topology/full/full.go
janos: I agree, it is ok.

janos:
Regarding the questions that you asked in the description. I do not think that libp2p handlers are blocking. They should all run in their own goroutine, for every stream.

pkg/node/node.go
acud: the error here is about multiaddress parse, why do we mention that we can't connect to the bootnode?

pkg/node/node.go
```diff
@@ -239,6 +227,38 @@ func NewBee(o Options) (*Bee, error) {
...
+			if err != nil {
+				logger.Debugf("multiaddress fail %s: %w", aa, err)
+				logger.Errorf("connect to bootnode %s: %w", aa, err)
+				return
+			}
+
+			overlay, err := p2ps.Connect(p2pCtx, addr)
+			if err != nil {
+				logger.Debugf("connect fail %s: %w", aa, err)
+				logger.Errorf("connect to bootnode %s: %w", aa, err)
```
acud, line 247: if you're logging the exact same things, why do we need to log it both in debug and in error?

pkg/p2p/libp2p/libp2p.go
```diff
@@ -193,13 +193,21 @@ func New(ctx context.Context, o Options) (*Service, error) {
 		}
 
 		s.peers.add(stream.Conn(), i.Address)
-		s.addrssbook.Put(i.Address, stream.Conn().RemoteMultiaddr())
+		remoteMultiaddr, err := ma.NewMultiaddr(fmt.Sprintf("%s/p2p/%s", stream.Conn().RemoteMultiaddr().String(), peerID.Pretty()))
+		if err != nil {
+			s.logger.Debugf("multiaddr error: handle %s: %v", peerID, err)
```
acud, line 198: this is quite different message than what is in error

pkg/node/node.go
pradovic: @janos I agree, it is a mistake :)
@acud My reasoning was that this is a users message, so user does not care why  the connect failed. Does it make sense?

pradovic:
> Regarding the questions that you asked in the description. I do not think that libp2p handlers are blocking. They should all run in their own goroutine, for every stream.

Agreed. My point was the combination of connect & receive peers in hive, which has a handler that blocks that whole stream, but I will move the handler after the stream close. I think that async handler would still be better :)

pkg/p2p/libp2p/libp2p.go
pradovic: The idea is that message in error is user scope, and debug is a developer scope which has a more detailed description. In this case, for the user, the result is that connection failed. For the developer is that it failed because because of the multiaddr generation error. 
This is the approach me & @janos took across the bee, you can see it across the project, and I believe that we all talked about it, but I might be wrong. If you don't like it, let's discuss it and switch to something else maybe?

pradovic:
@acud, @janos Thanks for the nice review guys!
I ended up leaving handlers as is, but moved them out of the stream scope in hive which should solve the described situation for now.
I see we have a discussions often  around the `debug` and `error` logging pair approach, and maybe we should discuss it on hangouts to see if we agree on it.

PR #54, opened 2020-04-03
add CODING.md

janos:
This PR adds a new file CODING.md with guidelines on writing code for this project.

I invite everyone in the team to review it, without creating a list of individual members which could end up incomplete.

### DIFF
--- a/CODING.md
+++ b/CODING.md
@@ -36,7 +36,7 @@ Explicit `error` values that are constructed with `errors.New` may have a messag
 
 Log messages are categorized in five different severities: error, warning, info, debug and trace, with semantic meaning associated with each of them.
 
-Two types of application users are identified: regular users and developers. Regular users should not be presented with confusing technical implementation details in log messages, but only with meaningful information related to application operability in a form of meaningful statements. Developers are users that are aware of implementation details and they will benefit from technical details to help them debug the application problematic state.
+Two types of application users are identified: regular users and developers. Regular users should not be presented with confusing technical implementation details in log messages, but only with meaningful information related to application operability in a form of meaningful statements. Developers are users that are aware of implementation details and they will benefit from technical details to help them debug the problematic application state.
 
 This means that the same problematic event may have two log lines but with different severities. This is the case with Error/Debug or Warning/Debug combo where Error or Warning is meant for the regular user and the Debug for developers to help them investigate the issue. Info and Trace log levels are informative about the expected state changes, but Info, with operable information, for regular user, and Trace, with technical details, for developer.
 
--- a/CODING.md
+++ b/CODING.md
@@ -1,6 +1,6 @@
 # Coding guide
 
-This project is written in Go programming language and follows a community accepted coding styles and practices defined in https://golang.org/doc/effective_go.html and https://github.com/golang/go/wiki/CodeReviewComments. This document defines additional guidelines specific to the Bee project.
+This project is written in Go programming language and follows a community accepted coding styles and practices defined in [Effective Go](https://golang.org/doc/effective_go.html) and [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments). This document defines additional guidelines specific to the Bee project.
 
 Developers should keep their Go tooling up to date with the latest stable versions.
 
@@ -10,7 +10,7 @@ Linting is done by using https://github.com/golangci/golangci-lint with configur
 
 ## Testing
 
-It is preferred to use separate test package to write tests. This allows explicit testing on exported behaviour of the package, not the internal implementation. In need to export only to testing package, export_test.go file should be used. Testing internal implementations discouraged, but not forbidden. Executing `make test` should pass without any warnings or errors.
+It is preferred to use a separate test package to write tests. This allows explicit testing on exported behaviour of the package, not the internal implementation. In need to export only to testing package, `export_test.go` file should be used. Testing internal implementations is discouraged, but not forbidden. Executing `make test` should pass without any warnings or errors.
 
 ## Packages
 
@@ -28,15 +28,19 @@ Every channel must have an owning goroutine. That goroutine is the only one that
 
 Errors must be propagated by the caller function. Error should not be logged and passed at the same time. It is up to the next caller function to decide what should be done in case of an error.
 
-If a function has multiple returns with, any returned error should be annotated either with fmt.Errorf using `%w` verb or with custom error type that has Unwrap method.
+If a function has multiple returns with, any returned error should be annotated either with `fmt.Errorf` using `%w` verb or with custom error type that has `Unwrap() error` method.
+
+Explicit `error` values that are constructed with `errors.New` may have a message prefixed with the package name and a colon character, if it would improve the message clarity.
 
 ## Logging
 
 Log messages are categorized in five different severities: error, warning, info, debug and trace, with semantic meaning associated with each of them.
 
-Two types of application users are identified: regular user and developer. Regular user should not be presented with confusing technical implementation details in log messages, but only with meaningful information related to application operability in a form of meaningful statements. Developers are users that are aware of implementation details and they will benefit from technical details to help them debug the application problematic state.
+Two types of application users are identified: regular users and developers. Regular users should not be presented with confusing technical implementation details in log messages, but only with meaningful information related to application operability in a form of meaningful statements. Developers are users that are aware of implementation details and they will benefit from technical details to help them debug the application problematic state.
 
 This means that the same problematic event may have two log lines but with different severities. This is the case with Error/Debug or Warning/Debug combo where Error or Warning is meant for the regular user and the Debug for developers to help them investigate the issue. Info and Trace log levels are informative about the expected state changes, but Info, with operable information, for regular user, and Trace, with technical details, for developer.
+
+Error and warning log messages should provide enough information to identify a problem in the codebase without the need to log the filename and line number alongside, even if that information is possible to be added by the logging library.
 
 ### Error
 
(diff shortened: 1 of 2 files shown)


CODING.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Logging
+
+Log messages are categorized in five different severities: error, warning, info, debug and trace, with semantic meaning associated with each of them.
+
+Two types of application users are identified: regular user and developer. Regular user should not be presented with confusing technical implementation details in log messages, but only with meaningful information related to application operability in a form of meaningful statements. Developers are users that are aware of implementation details and they will benefit from technical details to help them debug the application problematic state.
+
+This means that the same problematic event may have two log lines but with different severities. This is the case with Error/Debug or Warning/Debug combo where Error or Warning is meant for the regular user and the Debug for developers to help them investigate the issue. Info and Trace log levels are informative about the expected state changes, but Info, with operable information, for regular user, and Trace, with technical details, for developer.
```
acud, line 39: same here

CODING.md
```diff
@@ -0,0 +1,59 @@
...
+
+For every goroutine that is created it must be defined how and when it terminates.
+
+Every channel must have an owning goroutine. That goroutine is the only one that can send to the channel, close it or transfer ownership to another goroutine. Wherever is possible a readonly or writeonly channels should be used.
+
+## Error propagation, annotation and handling
+
+Errors must be propagated by the caller function. Error should not be logged and passed at the same time. It is up to the next caller function to decide what should be done in case of an error.
+
+If a function has multiple returns with, any returned error should be annotated either with fmt.Errorf using `%w` verb or with custom error type that has Unwrap method.
```
acud, line 31: maybe set up some baseline on how to format the errors, i.e. have everything in lowecase, maybe have the package name in the beginning of every error message?

CODING.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Logging
+
+Log messages are categorized in five different severities: error, warning, info, debug and trace, with semantic meaning associated with each of them.
+
+Two types of application users are identified: regular user and developer. Regular user should not be presented with confusing technical implementation details in log messages, but only with meaningful information related to application operability in a form of meaningful statements. Developers are users that are aware of implementation details and they will benefit from technical details to help them debug the application problematic state.
+
+This means that the same problematic event may have two log lines but with different severities. This is the case with Error/Debug or Warning/Debug combo where Error or Warning is meant for the regular user and the Debug for developers to help them investigate the issue. Info and Trace log levels are informative about the expected state changes, but Info, with operable information, for regular user, and Trace, with technical details, for developer.
+
```
acud, line 40: maybe also nice to have - do we have the annotation on which line the log has been logged from? we had this with the go-ethereum logger, not sure if this one supports this too. if not - the user of the logging library should make sure that the logged message can be traced back to a specific line in a file

CODING.md
janos: I did not put put these guidelines here as they are covered in https://github.com/golang/go/wiki/CodeReviewComments#error-strings. I am not sure that we have to make a package prefix in error a requirement. It is a practice in stdlib, but not in all cases. I would add that if it makes sense, like the error message is to general and can be ambiguous on its own, a package name can be added. If it is ok with you?

CODING.md
janos: I think that it is possible with logrus. I am more in favour that our log messages are written in a way that are obvious where they could be found by their meaning, but yes, sometimes it is helpful to have a file and line number. I am of the opinion that this is a sign of not so well written log line.

CODING.md
acud: yep sounds good :+1:

CODING.md
acud: true. it might still be good to mention if the feature exists in the logger, i think this is good for onboarded devs to know

janos: I agree. We may add the option to bee cli, and document here that is it encouraged only to be used in development.

CODING.md
```diff
@@ -0,0 +1,59 @@
+# Coding guide
+
+This project is written in Go programming language and follows a community accepted coding styles and practices defined in https://golang.org/doc/effective_go.html and https://github.com/golang/go/wiki/CodeReviewComments. This document defines additional guidelines specific to the Bee project.
```
mortelli, line 3: since this is markdown, why not add text to the hyperlinks? like so:

`[Effective Go](https://golang.org/doc/effective_go.html)`

CODING.md
```diff
@@ -0,0 +1,59 @@
...
+
+Developers should keep their Go tooling up to date with the latest stable versions.
+
+## Static code analysis
+
+Linting is done by using https://github.com/golangci/golangci-lint with configuration defined in the root of the project and also `go vet`. Executing `make lint vet` should pass without any warnings or errors.
```
mortelli, line 9: i am not up to speed why the CI on `bee`, but i run the linter in `swarm` through:

`golangci-lint run --tests --deadline=2m --disable-all --enable=goimports --enable=varcheck --enable=vet --enable=gofmt --enable=misspell --enable=goconst ./...`

is this essentially the same? if not: why was this changed?

CODING.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Testing
+
+It is preferred to use separate test package to write tests. This allows explicit testing on exported behaviour of the package, not the internal implementation. In need to export only to testing package, export_test.go file should be used. Testing internal implementations discouraged, but not forbidden. Executing `make test` should pass without any warnings or errors.
```
mortelli, line 13: minor corrections:
- `to use a separate test package` instead of `to use separate test package`
- `internal implementaitons is discouraged` rather than `internal implementations discouraged`
- `expot_test.go` rather than export_test.go

and a question: why are we discouraging internal testing? isn't it better to have both?

CODING.md
mortelli, line 31: here we can also format `fmt.Errorf` and `Unwrap` since they are references to code

CODING.md
mortelli, line 39: i would make `regular user` and `developer` plural here

CODING.md
```diff
@@ -0,0 +1,59 @@
...
+
+Error log messages are meant for regular users to be informed about the event that is happening which is not expected or may require intervention from the user for application to function properly. Log messages should be written as a statement, e.g. *unable to connect to peer 12345* and should not reveal technical details about internal implementation, such are package name, variable name or internal variable state. It should contain information only relevant to the user operating the application, concepts that can be operated on using exposed application interfaces.
+
+### Warning
+
+Warning log messages are also meant for regular users, as Error log messages, but are just informative about the state of the application that it can handle itself, by retry mechanism, for example. They should have the same form as the Error log messages.
+
+### Info
+
+Info log messages are informing users about the changes that are changing the application state in the expected manner, or presenting the user the information that can be used to operate the application, such as *node address: 12345*.
```
mortelli, line 51: maybe to not repeat "changes" here, say `about the changes that are affecting...`

CODING.md
janos: The makefile is there to make the command easy to run and to be able to update it if needed without communicating the change. In bee, we are using the default configuration, not a subset like in `swarm`. But with makefile, we can change it or add new commands.

CODING.md
janos: Thanks. Internal testing makes changes to the internal implementation harder, while behaviour testing on exposed api does not have to be changed when the internals change or improve.

mortelli: thanks for the clarification. i'm probably missing something though, because this sounds to me like you're saying "do not unit test"

janos: Maybe this would mean that it should be written a bi better. It should say that the unit tests should treat behaviours that package is exposing as units.

CODING.md
```diff
@@ -28,16 +28,20 @@ Every channel must have an owning goroutine. That goroutine is the only one that
...
 
 ## Logging
 
 Log messages are categorized in five different severities: error, warning, info, debug and trace, with semantic meaning associated with each of them.
 
-Two types of application users are identified: regular user and developer. Regular user should not be presented with confusing technical implementation details in log messages, but only with meaningful information related to application operability in a form of meaningful statements. Developers are users that are aware of implementation details and they will benefit from technical details to help them debug the application problematic state.
+Two types of application users are identified: regular users and developers. Regular users should not be presented with confusing technical implementation details in log messages, but only with meaningful information related to application operability in a form of meaningful statements. Developers are users that are aware of implementation details and they will benefit from technical details to help them debug the application problematic state.
```
acud, line 39: `... them debug the problematic application state.`

CODING.md
mortelli: requesting additional opinions here then (if it's just me i will resolve)

CODING.md
mortelli: just to clarify: i meant the very last sentence in the paragraph

CODING.md
acud: if you have an unexported function that provides some very specific business logic that should not be exposed outside the package then you can and should unit test it that it performs according to your expectations. 

for all other functionalities - you should test at the edges of the package, and determine that the package or the contract as a whole _behaves_ as expected. this is what is called behavior driven testing or behavior driven development. 

it is better that the tests we write assume a certain _behavior_ rather than implementation details, this way, it is easier to determine if a component is functioning correctly but also allows for a component to be replaced by another implementation without much hassle, since interface contracts are to be maintained (see `SOLID` and `Liskov substitution principle` in particular)

acud: I guess that the biggest misunderstanding is what we call a `unit`. In `bee` I know that @janos is pushing for an approach where we try ask much as we can to call the _exposed_ unit our _unit_, rather than some underlying implementation details

janos: Thanks @acud. That is exactly what I wanted to say.

@mortelli I have added one sentence that should provide a context of a unit, but not to get into reasons for that.

I am happy to hear suggestions how this can be explained more clearly, if needed.

CODING.md
janos: @mortelli thanks, I missed that, sorry.

PR #65, opened 2020-04-06
Connect from both sides at the same time

pradovic:
This is an attempt to solve situation where node1 connects to node 2 and node 2 connects to node 1 at the "same time".

The requirements:
1. Both connects should succeed or one should fail (depending on timing), but should result in one connection libp2p connection.
2. System state should be valid, such as number of peers in topology, number of connected nodes, etc..
3. Multiple handshakes from one side should still result in disconnect
4. Connect several times from one side should still fail


pkg/p2p/libp2p/connections_test.go
```diff
@@ -308,3 +278,41 @@ func TestConnectWithDisabledQUICAndWSTransports(t *testing.T) {
 	expectPeers(t, s2, overlay1)
 	expectPeersEventually(t, s1, overlay2)
 }
+
+func TestConnectRepeatHandshake(t *testing.T) {
```
acud, line 282: can you please document what the test is supposed to do? not just this one, it would be nice to have a godoc style comment on all tests

pkg/p2p/libp2p/connections_test.go
```diff
@@ -252,8 +215,15 @@ func TestDoubleConnectOnAllAddresses(t *testing.T) {
 		expectPeers(t, s2, overlay1)
 		expectPeersEventually(t, s1, overlay2)
 
-		if _, err := s2.Connect(ctx, addr); err == nil {
-			t.Fatal("second connect attempt should result with an error")
+		if _, err := s2.Connect(ctx, addr); !errors.Is(err, p2p.ErrAlreadyConnected) {
```
acud, line 218: so, the test says that it connects to all addresses which is not correct. it just connects to 1 address at a time. i would expect it to try to connect on all addresses (although i'm not sure which benefit it brings, except making sure the libp2p binds to the appropriate addresses, ports and protocols as expected from configuration, which should be tested as part of the libp2p abstraction, not as part of connection logic), or at least to try repeated connections and testing that the connect-disconnect logic that throws the peer out works as expected.

pkg/p2p/libp2p/connections_test.go
```diff
@@ -308,3 +278,41 @@ func TestConnectWithDisabledQUICAndWSTransports(t *testing.T) {
...
 }
+
+func TestConnectRepeatHandshake(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	s1, overlay1, cleanup1 := newService(t, libp2p.Options{NetworkID: 1})
+	defer cleanup1()
+
+	s2, overlay2, cleanup2 := newService(t, libp2p.Options{NetworkID: 1})
```
acud, line 289: do we have a test that checks if two nodes on different `NetworkID`s can (or cannot) connect to each other?

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -37,24 +41,27 @@ type PeerFinder interface {
...
 
 type Service struct {
-	peerFinder PeerFinder
-	overlay    swarm.Address
-	networkID  int32
-	logger     logging.Logger
+	overlay            swarm.Address
+	networkID          int32
+	receivedHandshakes map[libp2ppeer.ID]struct{}
+	mtx                sync.Mutex // guards received handshakes
```
acud, line 47: comment not necessary. you can also rename the mutex to be `muReceivedHandshakes`

pkg/p2p/libp2p/libp2p.go
```diff
@@ -173,9 +173,10 @@ func New(ctx context.Context, o Options) (*Service, error) {
 		return nil, fmt.Errorf("protocol version match %s: %w", id, err)
 	}
 
+	// handshake
```
acud, line 176: not sure if this is needed

pkg/p2p/libp2p/libp2p.go
```diff
@@ -187,12 +188,15 @@ func New(ctx context.Context, o Options) (*Service, error) {
 
 			s.logger.Debugf("handshake: handle %s: %v", peerID, err)
 			s.logger.Errorf("unable to handshake with peer %v", peerID)
-			// todo: test connection close and refactor
 			_ = s.disconnect(peerID)
 			return
 		}
 
-		s.peers.add(stream.Conn(), i.Address)
+		if s.peers.addIfNotExists(stream.Conn(), i.Address) {
```
acud, line 195: it would be nice if there is more explicit mentioning of the return value from the function, such as:
```go
if exists := s.peers.addIfNotExists(); exists {
  // peer already exists, don't notify again
  return
}
```

pkg/p2p/libp2p/libp2p.go
```diff
@@ -216,7 +220,8 @@ func New(ctx context.Context, o Options) (*Service, error) {
 		s.metrics.HandledConnectionCount.Inc()
 	})
 
-	h.Network().Notify(peerRegistry) // update peer registry on network events
+	h.Network().Notify(peerRegistry)       // update peer registry on network events
+	h.Network().Notify(s.handshakeService) // update handshaker service on network events
```
acud, line 224: `update handshake service` (unless there's a `handshaker` interface, which i might have missed)

pkg/p2p/libp2p/libp2p.go
```diff
@@ -317,7 +326,11 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (overlay swarm
 		return swarm.Address{}, err
 	}
 
-	s.peers.add(stream.Conn(), i.Address)
+	if s.peers.addIfNotExists(stream.Conn(), i.Address) {
```
acud, line 329: see the other comment about a similar line. also, i'm not sure that the `trace` line is needed at all

pkg/p2p/libp2p/peer.go
```diff
@@ -77,17 +77,23 @@ func (r *peerRegistry) peers() []p2p.Peer {
 	return peers
```
acud, line 77: in `peers()`, `r.mu.Lock()` should be changed to `r.mu.RLock()` (or switch to a normal mutex, not `RWMutex`)

pkg/topology/full/full.go
```diff
@@ -69,11 +70,13 @@ func (d *Driver) AddPeer(ctx context.Context, addr swarm.Address) error {
...
 		if err != nil {
-			return err
+			if !errors.Is(err, p2p.ErrAlreadyConnected) {
+				return err
+			}
 		}
 
 		// update addr if it is wrong or it has been changed
-		if !addr.Equal(peerAddr) {
+		if !addr.Equal(peerAddr) && !peerAddr.IsZero() {
```
acud, line 79: why do we need this `IsZero` check? why should zero addresses be transmitted in the first place?

pkg/hive/hive.go
```diff
@@ -23,7 +23,7 @@ const (
 	protocolName    = "hive"
 	protocolVersion = "1.0.0"
 	peersStreamName = "peers"
-	messageTimeout  = 5 * time.Second // maximum allowed time for a message to be read or written.
+	messageTimeout  = 1 * time.Minute // maximum allowed time for a message to be read or written.
```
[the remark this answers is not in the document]
pradovic: Well, I would argue that abysmal amount depends on the context :)  IMO, the point of timeout is just to guard against infinite wait, not to make the other side finish operation quickly. So the standard expiration times are usually very long, I would put it even longer tbh. LTMWYT, @janos  as well pls :)

pradovic:
@acud Some of the comments are related to this https://github.com/ethersphere/bee/pull/71. I guess you should refresh the github page before reviewing :)  🙈

go.sum
```diff
@@ -110,6 +110,7 @@ github.com/hashicorp/golang-lru v0.5.1 h1:0hERBMJE1eitiLkihrMvRVBYAkpHzc/J3QdDN+
 github.com/hashicorp/golang-lru v0.5.1/go.mod h1:/m3WP610KZHVQ1SGc6re/UDhFvYD7pJ4Ao+sR/qLZy8=
 github.com/hashicorp/hcl v1.0.0 h1:0Anlzjpi4vEasTeNFn2mLJgTSwt0+6sfsiTG8qcWGx4=
 github.com/hashicorp/hcl v1.0.0/go.mod h1:E5yfLk+7swimpb2L/Alb/PJmXilQ/rhwaUYs4T20WEQ=
+github.com/hpcloud/tail v1.0.0 h1:nfCOvKYfkgYP8hkirhJocXT2+zOD8yUNjXaWfTlyFKI=
```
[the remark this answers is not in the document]
pradovic: I am not sure to be honest, I will try ti hunt it down. Have been going in circles with go.sum last couple of days, I am not sure why this happens 🙈

pkg/p2p/libp2p/connections_test.go
pradovic: This is the old tests, I just changed the connection state after double connect.  @janos I believe you did the test, so could you comment pls?

pkg/p2p/libp2p/connections_test.go
pradovic: There is a test in handshake, I don't see the test in libp2p as well. Do you think we would need one?

acud:
i marked those comments as resolved :+1:

pkg/p2p/libp2p/connections_test.go
pradovic: Basically, this test is supposed to test if one node sends multiple handshake messages to another node. Because, double handshake can happen in the situation I explained in the PR description, but the try to perform another handshake after will be penalized by disconnect, as stated in the book of swarm. I will add some go doc to this test.

pkg/topology/full/full.go
pradovic: Few lines aboce I check `IsConnected` which is not thread-safe, and return an error only if it's not `ErrAlreadyConnected`. If it is, the code continues, and `peerAddr` will be zero in this case.

pkg/p2p/libp2p/peer.go
pradovic: Ah, I think my autosuggestion prefers `RWMutex` for some reason, I made a same mistake couple of times already 🙈

pkg/topology/full/full.go
acud: i'm sorry, i might have missed something, but why should the code continue execution when there is an error in the code block above. can you elaborate? why would you like to put the peer again in to the addressbook if its already there?

pkg/hive/hive.go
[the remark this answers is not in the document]
acud: in which context, except for [this one](https://en.wikipedia.org/wiki/Interplanetary_Internet), do you expect a handshake to take a minute?

pkg/p2p/libp2p/connections_test.go
janos: Yes it connects to all addresses, but one at the time. A parallel test would be a different one. This one ensures that everything is cleaned up on disconnects related to different peers. @acud If you think that this test is not good, feel free to address it in another pr. I do not think that it is a scope of this pr.

pkg/p2p/libp2p/connections_test.go
janos: I think that the handshake test with different network ids covers this situation, as it is related to the handshake.

pkg/p2p/libp2p/internal/handshake/handshake.go
janos: Yes that would make sense. I would prefer `receivedHandshakesMu` name.

pkg/hive/hive.go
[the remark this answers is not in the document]
acud: i think we should not be connected to a peer that takes 1m to handshake, but that's just me...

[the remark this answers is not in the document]
pradovic: @acud I agree in general, but I don't believe this is a place to handle this. For example, we will probably stress/load test this in simulations, or just plain test in in several network envs, and this should not be the reason for such tests to fail. If we want to disconnect slow peers, I believe we will probably end up doing it on another place that is more general. These timeouts are just sanity guards for protocols not to wait forever. Basically, as @janos said, it is not a restriction of a protocol, just a safety measure. Actually, when thinking about this, IRRC you had a similar point as well, warning us not to put timeouts on these places at all, and that is actually right. The big timeout is just as a safe guard, but not meant to intervene with the flow.

pkg/topology/full/full.go
pradovic: Well, basically, if you see, the code logic is that it connects if the node is not already connected. So, I don't see a reason to break it on already connected error, cuz it would be the same thing, right?

pradovic: Actually, you are right, it is the different error actually, so I will break in this case as well. It will make the code more simple, and it is still good enough. Thanks!

pkg/p2p/libp2p/peer.go
pradovic: Oh, actually, I thought this was something else. I don't think this is a part of this PR as well.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -326,7 +326,7 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (overlay swarm
 		return swarm.Address{}, err
 	}
 
-	if s.peers.addIfNotExists(stream.Conn(), i.Address) {
+	if exists := s.peers.addIfNotExists(stream.Conn(), i.Address); exists {
 		s.logger.Trace("skipped double connect")
```
acud, line 330: i still vote to get rid of this logline. it is not informative at all. it just says some general condition has happened. it neither mentions the base address of the reporting node, nor does it report with which node it skipped connecting to. in it's current form i would really not like to have such things in my logger output (as a developer)

pkg/p2p/libp2p/libp2p.go
```diff
@@ -192,7 +192,7 @@ func New(ctx context.Context, o Options) (*Service, error) {
 			return
 		}
 
-		if s.peers.addIfNotExists(stream.Conn(), i.Address) {
+		if exists := s.peers.addIfNotExists(stream.Conn(), i.Address); exists {
 			s.logger.Trace("skipped double connect")
```
acud, line 196: see other comment. this is not informative, please remove

pkg/p2p/libp2p/libp2p.go
pradovic: Deal!

pradovic: Is it ok to put it as debug? I would like to have it for testing?

acud: for testing you can add the line locally. in the old codebase we were at some points logging gigabytes per hour per cluster on `Trace`, these things must be taken into account. i prefer to log less but log meaningful things. either have this more informative as to what exactly happened or don't log at all

pradovic: ok

acud: also, i am not against logging. i am just arguing that it should be informative. so if you choose to log here you can do so as long as its informative and i'm totally cool with that. if you want to have a logline just add more info, that's all i'm asking

Issue #59, opened 2020-04-06
Addressbook persistence


pradovic:
# Task
implement addressbook persistence, pass errors in its methods and handle them

## Acceptance criterea

Test addressbook persistence when node is started, while node is up and running, and after the node is shutdown

PR #68, opened 2020-04-07
statestore: add initial implementation

acud:
uses leveldb for now. will be converted later to badger

dependency of https://github.com/ethersphere/bee/issues/59

### DIFF
--- a/pkg/statestore/store.go
+++ b/pkg/statestore/store.go
@@ -7,15 +7,11 @@ package statestore
 import (
 	"encoding"
 	"encoding/json"
-	"errors"
 
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/util"
 )
-
-// ErrNotFound is returned when no results are returned from the database
-var ErrNotFound = errors.New("ErrorNotFound")
 
 // Store uses LevelDB to store values.
 type Store struct {
--- a/pkg/statestore/store.go
+++ b/pkg/statestore/store.go
@@ -13,6 +13,8 @@ import (
 	"github.com/syndtr/goleveldb/leveldb/util"
 )
 
+var _ storage.StateStorer = (*Store)(nil)
+
 // Store uses LevelDB to store values.
 type Store struct {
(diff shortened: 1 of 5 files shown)


pkg/statestore/store.go
```diff
@@ -0,0 +1,91 @@
...
+	var bytes []byte
+	if marshaler, ok := i.(encoding.BinaryMarshaler); ok {
+		if bytes, err = marshaler.MarshalBinary(); err != nil {
+			return err
+		}
+	} else {
+		if bytes, err = json.Marshal(i); err != nil {
+			return err
+		}
+	}
```
pradovic, line 63:
```suggestion
	} else if bytes, err = json.Marshal(i); err != nil {
		return err
	}
```

acud:
thanks @pradovic.

> I have a feeling that named returned params are a bit overused. I don't think there is a reason to use them when the return params are just error, or error and an expected value like in constructor.

I'm not sure I understand what you mean. Can you please elaborate?

pkg/statestore/store.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/syndtr/goleveldb/leveldb/util"
+)
+
+// Store uses LevelDB to store values.
+type Store struct {
+	db *leveldb.DB
+}
+
+// New creates a new persistent state storage.
+func New(path string) (s storage.StateStorer, err error) {
```
pradovic, line 22:
```suggestion
func New(path string) (storage.StateStorer, error) {
```

pkg/statestore/store.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, err
+	}
+	return &Store{
+		db: db,
+	}, nil
+}
+
+// Get retrieves a value of the requested key. If not results are found,
+// storage.ErrNotFound will be returned.
+func (s *Store) Get(key string, i interface{}) (err error) {
```
pradovic, line 34:
```suggestion
func (s *Store) Get(key string, i interface{}) error {
```

pradovic:
@acud I have added a few code suggestions to illustrate the point.

acud:
Thanks @pradovic !

PR #71, opened 2020-04-07
Hive tests fix - expect values eventually

pradovic:
Fixes flacky tests for `BroadcastPeers` call in discovery.
caused by https://github.com/ethersphere/bee/pull/46

### DIFF
--- a/pkg/hive/hive_test.go
+++ b/pkg/hive/hive_test.go
@@ -153,47 +153,58 @@ func TestBroadcastPeers(t *testing.T) {
 			}
 		}
 
-		if !compareOverlays(exporter.Overlays(), tc.wantOverlays) {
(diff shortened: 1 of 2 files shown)


pkg/hive/hive_test.go
```diff
@@ -153,47 +153,58 @@ func TestBroadcastPeers(t *testing.T) {
...
+		sort.Strings(stringMultiaddresses)
+		sort.Strings(stringWantMultiAddresses)
+		if reflect.DeepEqual(stringMultiaddresses, stringWantMultiAddresses) {
+			return
+		}
 
-	var stringWantVal []string
-	for _, v := range wantValues {
-		stringWantVal = append(stringWantVal, v.String())
+		time.Sleep(50 * time.Millisecond)
```
janos, line 204: Can we have a retry mechanism here instead a fixed sleep? That would make this test more resilient on slower environments, like automated ci instances.

Issue #66, opened 2020-04-07
Migrate localstore from swarm to bee


janos:
# Task
Localstore is the chunk storage implementation in the swarm project that needs to me migrated to the bee project. It depends on the shed package which currently uses leveldb to store data. It is agreed that badgerdb should be used instead, but in order to have clear trace of the changes and for easier code reviews, a multiple pull requests are beneficial. It is fine to have swarm and ethereum as dependencies between PRs, even if it does not look very nice, but it contributes to the clarity in the process.

- [x] 1. A PR that would include swarm shed package as it is.
- [x] 2. A PR that changes StructField to use encoding/json instead ethereum's rlp for encoding, to remove a dependency on ethereum codebase
- [x] 3. A PR that replaces shed ethereum's metrics with bee metrics, to remove a dependency on ethereum codebase
- [x] 4. A PR that replaces shed log library with bee logging library, to remove a dependency on swarm codebase
- [x] 5. A PR that replaces leveldb with badgerdb
- [x] 6. A PR that contains the localstore from swarm
- [x] 7. A PR that replaces localstore ethereum's metrics with bee metrics, to remove a dependency on ethereum codebase
- [x] 8. A PR that replaces localstore log library with bee logging library, to remove a dependency on swarm codebase
- [x] 9. A PR that removes explicit localstore migrations, but leaves the functionality to add new migrations
- [x] 10. A PR that migrates only needed swarm ethersphere/swarm/chunk package types to bee and replacing them in localstore
- [x] 11. A PR that replaces swarm shed with bee shed in localstore, and any references to leveldb that are left in localstore
- [x] 12. move localstore out of `storage` package
- [ ] 13. fix SubscribePush tests that are set skipped in PR 11

## Acceptance criterea

The goal would be to have localstore migrated to bee with the same exported api without any dependencies to swarm or ethereum projects.

jmozah:
@janos can you shed some light on PR 9

PR #73, opened 2020-04-07
P3 - Replace ethereum's metrics with bee's metrics

jmozah:
This PR replaces shed ethereum's metrics with bee metrics, to remove a dependency on ethereum codebase
Refer P3 in #66


janos:
P2 is merged, targeted branch can be changed to master.

PR #74, opened 2020-04-07
P4 - Replace log from ethereum to bee's own log library

jmozah:
This PR replaces shed log library with bee logging library, to remove a dependency on swarm codebase.
Refer P3 in #66


pkg/shed/db.go
```diff
@@ -81,6 +84,7 @@ func NewDB(path string, metricsPrefix string) (db *DB, err error) {
 func (db *DB) Put(key []byte, value []byte) (err error) {
 	err = db.ldb.Put(key, value, nil)
 	if err != nil {
+		db.logger.Debugf("failed to insert in to DB", "err", err)
```
janos, line 87: This usage of Debugf is not correct.

janos: @jmozah could you just fix this, and we can merge?

pkg/shed/db.go
```diff
@@ -81,6 +84,7 @@ func NewDB(path string, metricsPrefix string) (db *DB, err error) {
 func (db *DB) Put(key []byte, value []byte) (err error) {
 	err = db.ldb.Put(key, value, nil)
 	if err != nil {
+		db.logger.Debugf("failed to insert in to DB. Error : ",  err.Error())
```
janos, line 87: This is still not functionally right. Missing `%s`.

pkg/shed/db.go
jmozah: @janos oops... forgot this one...Dne it now...

pkg/shed/db.go
jmozah: oops.. sorry..

jmozah:
@janos  Please merge this if you are satisfied.

PR #75, opened 2020-04-08
P5 - Replace levelDB with badgerDB in shed

jmozah:
This PR replaces leveldb with badgerdb.
Refer P5 in #66


janos:
This PR should be rebased with master when https://github.com/ethersphere/bee/pull/74 is merged. PR #74 has only one comment that needs to be resolved.

janos:
PR https://github.com/ethersphere/bee/pull/74 is merged, please, rebase.

pkg/shed/db.go
```diff
@@ -74,83 +88,361 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+				continue
+			}
+			count++
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error while counting keys in DB")
+		db.logger.Debugf("error while doing count. Error : %w", err.Error())
+		db.metrics.TotalFailCount.Inc()
```
janos, line 219: Should this block return and avoid the Trace log line below?

pkg/shed/db.go
```diff
@@ -74,83 +88,361 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+		// if prefix is nil, it is equivalent to counting from beginning
+		for i.Seek(prefix); i.Valid(); i.Next() {
+			count++
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error doing count from. prefix : %v", string(prefix))
+		db.logger.Debugf("error while doing count from. prefix : %v, Error : %w", string(prefix), err.Error())
+		db.metrics.CountPrefixFailCount.Inc()
```
janos, line 284: Should this block return and avoid the Trace log line below?

pkg/shed/db.go
```diff
@@ -74,83 +88,361 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+	db.metrics.LastCount.Inc()
+	err = db.bdb.View(func(txn *badger.Txn) (err error) {
+		o := badger.DefaultIteratorOptions
+		o.PrefetchValues = true
+		o.PrefetchSize = 1024
+
+		i := txn.NewIterator(o)
+		defer i.Close()
+
+		for i.Seek(prefix); i.ValidForPrefix(prefix); i.Next() {
```
janos, line 393: This will iterate over all keys with prefixes to get the last one. Looks expensive.

Shed in swarm uses more efficient approach to seek ahead and have only one Prev call afterwards https://github.com/ethersphere/swarm/blob/master/shed/index.go#L375. Why was that abandoned?

pkg/shed/db.go
```diff
@@ -74,83 +88,361 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
 	if err != nil {
-		db.logger.Debugf("could not writing batch. Error: %s ", err.Error())
-		db.metrics.WriteBatchFailCounter.Inc()
+		db.logger.Errorf("error while committing transaction. Error : %w", err.Error())
+		db.logger.Debugf("error while committing batch. Error : %w", err.Error())
+		db.metrics.WriteBatchFailCount.Inc()
 		return err
 	}
-	db.metrics.WriteBatchCounter.Inc()
+	txn.Discard()
```
janos, line 438: Why calling discard? Txn.Commit() also calls Txn.Discard() internally to cleanup the transaction.

pkg/shed/db.go
```diff
@@ -74,83 +88,361 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+	db.logger.Tracef("transaction committed successfully")
 	return nil
 }
 
-// Close closes LevelDB database.
-func (db *DB) Close() (err error) {
-	close(db.quit)
-	return db.ldb.Close()
+// Close shut's down the badger DB.
+func (db *DB) Close(ctx context.Context) (err error) {
```
janos, line 444: The context is not used. Why is it added?

pkg/shed/db.go
```diff
@@ -74,83 +88,361 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+	txn.Discard()
+	db.logger.Tracef("transaction committed successfully")
 	return nil
 }
 
-// Close closes LevelDB database.
-func (db *DB) Close() (err error) {
-	close(db.quit)
-	return db.ldb.Close()
+// Close shut's down the badger DB.
```
janos, line 443: Shouldn't be just `shuts`?

pkg/shed/field_string.go
```diff
@@ -65,6 +65,9 @@ func (f StringField) Put(val string) (err error) {
 
 // PutInBatch stores a string in a batch that can be
 // saved later in database.
-func (f StringField) PutInBatch(batch *leveldb.Batch, val string) {
-	batch.Put(f.key, []byte(val))
+func (f StringField) PutInBatch(batch *badger.Txn, val string) {
+	err := batch.Set(f.key, []byte(val))
+	if err != nil {
+		f.logger.Debugf("could not set values in PutInBatch")
```
janos, line 71: The error is not handled, not even put in log message.

This is very dangerous. We should return this error in order to stop any other actions on the batch. This can be cause of data inconstancies.

pkg/shed/field_struct.go
```diff
@@ -67,12 +67,15 @@ func (f StructField) Put(val interface{}) (err error) {
...
 	b, err := json.Marshal(val)
 	if err != nil {
-		f.logger.Debugf("could not PUT key %s in batch", string(f.key))
+		f.logger.Debugf("could not marshall while doing putiInBatch. Error: %s", err.Error())
 		return err
 	}
-	batch.Put(f.key, b)
+	err = batch.Set(f.key, b)
+	if err != nil {
+		f.logger.Debugf("could not put values in batch. Error: %s", err.Error())
```
janos, line 78: Why the error is not returned, just logged?

pkg/shed/field_uint64.go
```diff
@@ -67,16 +67,19 @@ func (f Uint64Field) Put(val uint64) (err error) {
 
 // PutInBatch stores a uint64 value in a batch
 // that can be saved later in the database.
-func (f Uint64Field) PutInBatch(batch *leveldb.Batch, val uint64) {
-	batch.Put(f.key, encodeUint64(val))
+func (f Uint64Field) PutInBatch(batch *badger.Txn, val uint64) {
+	err := batch.Set(f.key, encodeUint64(val))
+	if err != nil {
+		f.logger.Debugf("could not set uint64 value in PutInBatch. Error : %s", err.Error())
```
janos, line 73: The error is not handled!

This is very dangerous. We should return this error in order to stop any other actions on the batch. This can be cause of data inconstancies.

pkg/shed/index.go
```diff
@@ -162,32 +161,37 @@ func (f Index) Get(keyFields Item) (out Item, err error) {
...
 		key, err := f.encodeKeyFunc(item)
 		if err != nil {
 			f.logger.Debugf("keyfields encoding error in Fill. Error: %s", err.Error())
 			return err
 		}
-		value, err := snapshot.Get(key, nil)
+		badgerItem, err := txn.Get(key)
 		if err != nil {
+			if err == badger.ErrKeyNotFound {
+				f.logger.Debugf("error could not find key in fill. Error: %s", err.Error())
```
janos, line 176: Do not log error and return error.

pkg/shed/index.go
```diff
@@ -162,32 +161,37 @@ func (f Index) Get(keyFields Item) (out Item, err error) {
...
 			return err
 		}
-		v, err := f.decodeValueFunc(item, value)
+		var decodedItem Item
+		err = badgerItem.Value(func(val []byte) error {
+			decodedItem, err = f.decodeValueFunc(item, val)
+			if err != nil {
+				return err
+			}
+			return nil
```
janos, line 188: Can be `return err`.

pkg/shed/index.go
```diff
@@ -162,32 +161,37 @@ func (f Index) Get(keyFields Item) (out Item, err error) {
...
+		})
 		if err != nil {
 			f.logger.Debugf("error decofing keyfields in Fill . Error: %s", err.Error())
```
janos, line 191: Do not log error and return error.

pkg/shed/index.go
```diff
@@ -208,22 +212,30 @@ func (f Index) Has(keyFields Item) (bool, error) {
...
+
+	opts := badger.DefaultIteratorOptions
+	opts.PrefetchValues = false
+	it := txn.NewIterator(opts)
+	defer it.Close()
+
 	for i, keyFields := range items {
 		key, err := f.encodeKeyFunc(keyFields)
 		if err != nil {
 			f.logger.Debugf("keyfields encoding error in HasMulti. Error: %s", err.Error())
```
janos, line 227: Do not log error and return error.

pkg/shed/index.go
```diff
@@ -259,7 +271,11 @@ func (f Index) PutInBatch(batch *leveldb.Batch, i Item) (err error) {
 		f.logger.Debugf("keyfields encoding error in PutInBatch. Error: %s", err.Error())
 		return err
 	}
-	batch.Put(key, value)
+	err = batch.Set(key, value)
+	if err != nil {
+		f.logger.Debugf("could not set values in batch. Error : %s", err.Error())
```
janos, line 276: Why the error is not passed to the caller?

This may result in committing an incomplete batch.

pkg/shed/index.go
```diff
@@ -276,13 +292,16 @@ func (f Index) Delete(keyFields Item) (err error) {
...
+func (f Index) DeleteInBatch(batch *badger.Txn, keyFields Item) (err error) {
 	key, err := f.encodeKeyFunc(keyFields)
 	if err != nil {
 		f.logger.Debugf("keyfields encoding error in DeleteInBatch. Error: %s", err.Error())
 		return err
 	}
-	batch.Delete(key)
+	err = batch.Delete(key)
+	if err != nil {
+		f.logger.Debugf("could not delete items in batch. Error: %s", err.Error())
```
janos, line 303: Why the error is not passed to the caller?

This may result in committing an incomplete batch.

pkg/shed/index.go
```diff
@@ -322,102 +341,66 @@ func (f Index) Iterate(fn IndexIterFunc, options *IterateOptions) (err error) {
...
+		return i, err
+	}
+	return f.itemFromKeyValue(k, v, totalPrefix)
 }
 
-// itemFromIterator returns the Item from the current iterator position.
+// itemFromKeyValue returns the Item from the current iterator position.
 // If the complete encoded key does not start with totalPrefix,
-// leveldb.ErrNotFound is returned. Value for totalPrefix must start with
+// ErrNotFound is returned. Value for totalPrefix must start with
```
janos, line 373: This comment change is not correct. It will return `badger.ErrKeyNotFound` not `ErrNotFound`.

pkg/shed/index.go
```diff
@@ -322,102 +341,66 @@ func (f Index) Iterate(fn IndexIterFunc, options *IterateOptions) (err error) {
...
-	if l > 0 && nextPrefix != nil {
-		it.Seek(append(f.prefix, nextPrefix...))
-		it.Prev()
-	} else {
-		it.Last()
-	}
-
 	totalPrefix := append(f.prefix, prefix...)
-	return f.itemFromIterator(it, totalPrefix)
+	k, v, err := f.db.Last(totalPrefix)
```
janos, line 399: This change replaces an efficient implementation with one very inefficient. Related to comment to `db.Last`.

pkg/shed/metrics.go
```diff
@@ -13,95 +13,193 @@ type metrics struct {
...
 	subsystem := "shed"
 
 	return metrics{
-		PutCounter: prometheus.NewCounter(prometheus.CounterOpts{
+		DBOpenCount: prometheus.NewCounter(prometheus.CounterOpts{
 			Namespace: m.Namespace,
 			Subsystem: subsystem,
-			Name:      "put_count",
-			Help:      "Number of times the PUT operation is done.",
+			Name:      "db_db_open_count",
```
janos, line 51: Every metric name now has a prefix `db_`, this is not needed as there already is a subsystem "shed" set. It would be nicer to remove `db_` prefix for every metric in this PR.

pkg/shed/vector_uint64.go
```diff
@@ -66,16 +66,19 @@ func (f Uint64Vector) Put(i, val uint64) (err error) {
 
 // PutInBatch stores a uint64 value at index i in a batch
 // that can be saved later in the database.
-func (f Uint64Vector) PutInBatch(batch *leveldb.Batch, i, val uint64) {
-	batch.Put(f.indexKey(i), encodeUint64(val))
+func (f Uint64Vector) PutInBatch(batch *badger.Txn, i, val uint64) {
+	err := batch.Set(f.indexKey(i), encodeUint64(val))
+	if err != nil {
+		f.logger.Debugf("could not put uint64 value in batch. Error : ", err.Error())
```
janos, line 72: Again, error is not returned, which may result in data inconsistencies if the batch is committed, as the error is not propagated.

pkg/shed/db.go
jmozah: no. I want ed to print the count if trace is enabled.

pkg/shed/db.go
jmozah: no. I wanted to print the count if trace is enabled.

pkg/shed/db.go
jmozah: i also thought about this.. There is no prev() call in badger. i can raise a request for them. until then this is the only way.

janos:
There is also a panic from the badger codebase which is related to Go 1.14 unsafe.Pointer safety rules. This relates to badger 1.6.0 used in this PR. I didi not check if this issue is present in version 2.0.3.

```
=== RUN   TestNewDB
fatal error: checkptr: unsafe pointer arithmetic

goroutine 16 [running]:
runtime.throw(0xe12690, 0x23)
	/opt/hostedtoolcache/go/1.14.2/x64/src/runtime/panic.go:1116 +0x72 fp=0xc0000893d0 sp=0xc0000893a0 pc=0x446cd2
runtime.checkptrArithmetic(0xc00022a080, 0x0, 0x0, 0x0)
	/opt/hostedtoolcache/go/1.14.2/x64/src/runtime/checkptr.go:43 +0xb5 fp=0xc000089400 sp=0xc0000893d0 pc=0x418115
github.com/AndreasBriese/bbloom.Bloom.JSONMarshal(0x0, 0xe, 0xc00022a080, 0x8, 0x8, 0x9, 0x1ff, 0x7, 0x37, 0x0, ...)
	/home/runner/go/pkg/mod/github.com/!andreas!briese/bbloom@v0.0.0-20190306092124-e2d15f34fcf9/bbloom.go:238 +0xd1 fp=0xc0000894e8 sp=0xc000089400 pc=0xa086d1
github.com/dgraph-io/badger/table.(*Builder).Finish(0xc0002028a0, 0xc0002800cb, 0x14, 0x533322e)
	/home/runner/go/pkg/mod/github.com/dgraph-io/badger@v1.6.0/table/builder.go:229 +0x7c1 fp=0xc000089ac8 sp=0xc0000894e8 pc=0xa0b621
github.com/dgraph-io/badger.writeLevel0Table(0xc0001aa060, 0x2a00000000, 0x39, 0x0, 0x0, 0x0, 0xf096a0, 0xc000206018, 0x0, 0x0)
	/home/runner/go/pkg/mod/github.com/dgraph-io/badger@v1.6.0/db.go:845 +0x618 fp=0xc000089ca0 sp=0xc000089ac8 pc=0xa2a828
github.com/dgraph-io/badger.(*DB).handleFlushTask(0xc0000fd180, 0xc0001aa060, 0x2a00000000, 0x39, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/runner/go/pkg/mod/github.com/dgraph-io/badger@v1.6.0/db.go:884 +0x5af fp=0xc000089e00 sp=0xc000089ca0 pc=0xa2ae8f
github.com/dgraph-io/badger.(*DB).flushMemtable(0xc0000fd180, 0xc00012a0a0, 0x0, 0x0)
	/home/runner/go/pkg/mod/github.com/dgraph-io/badger@v1.6.0/db.go:918 +0x1cc fp=0xc000089f98 sp=0xc000089e00 pc=0xa2b62c
github.com/dgraph-io/badger.Open.func4(0xc000010318)
	/home/runner/go/pkg/mod/github.com/dgraph-io/badger@v1.6.0/db.go:298 +0x8b fp=0xc000089fd8 sp=0xc000089f98 pc=0xa70e2b
runtime.goexit()
	/opt/hostedtoolcache/go/1.14.2/x64/src/runtime/asm_amd64.s:1373 +0x1 fp=0xc000089fe0 sp=0xc000089fd8 pc=0x47a871
created by github.com/dgraph-io/badger.Open
	/home/runner/go/pkg/mod/github.com/dgraph-io/badger@v1.6.0/db.go:297 +0x2308
```

https://github.com/ethersphere/bee/pull/75/checks?check_run_id=573849921#step:10:311

pkg/shed/db.go
jmozah: good catch. But calling it twice does not harm it anyway. I will remove it for brevity.

pkg/shed/db.go
jmozah: if we want to cancel Close() if it takes too long or something in future...What do you say.. keep it or remove it...?

pkg/shed/field_string.go
jmozah: yeah. I will add error..

pkg/shed/db.go
janos: I do not see any reason to add an argument that is not used in the function body.

pkg/shed/db.go
janos: Should the count be irrelevant if there is an error?

pkg/shed/db.go
janos: Should the count be irrelevant if there is an error?

pkg/shed/index.go
jmozah: i think it is correct... it should be return nil

janos: The if is not needed, the comment is meant for all 4 lines, including if.

pkg/shed/index.go
jmozah: As i said in the older commit.. there is no way to do a Previous() while doing a iteration. Let me see how else can i fix this.

pkg/shed/db.go
janos: Couldn't the `IteratorOptions.Reverse` option be used to get the Prev functionality?

pkg/shed/index.go
janos: Couldn't the `IteratorOptions.Reverse` option be used to get the Prev functionality?

jmozah:
@janos fixed most of the comments by you. I am not sure about the dump though. Looks like it happened while opening the DB which i am doing so many times. May be this is related to CI machine... Lets see if it comes up again. I have raised a issue in badger repo also https://github.com/dgraph-io/badger/issues/1294

janos:
Thanks for addressing my comments, but now PutInBatch errors need to be handled https://github.com/ethersphere/bee/pull/75/checks?check_run_id=574344361#step:7:8.

jmozah:
@janos fixed the lints.. but the dump seems to be tricky for now. How do you want to proceed?

janos:
Thanks @jmozah. I think that we should not have panics related to pointer arithmetics, or actually, any type of panic. Would you mind to check if this problem is not present with badger 2.0.3?

jmozah:
2.0.3 also has the same issue... I dont think this is related to badger.. we are doing something wrong..

jmozah:
ok... looks like it works in 2.0.3...

janos:
> 2.0.3 also has the same issue... I dont think this is related to badger.. we are doing something wrong..

The panic is exactly pointing to badger using bbloom package code that handles pointers. It is clear what is the error.

jmozah:
> The panic is exactly pointing to badger using bbloom package code that handles pointers. It is clear what is the error.

They had some issues in the bloom filter i dont remember. @jarifibrahim  is looking in to it.

go.mod
```diff
@@ -5,7 +5,8 @@ go 1.14
 require (
 	github.com/btcsuite/btcd v0.20.1-beta
 	github.com/coreos/go-semver v0.3.0
-	github.com/ethereum/go-ethereum v1.9.12
+	github.com/dgraph-io/badger/v2 v2.0.3
+	github.com/ethereum/go-ethereum v1.9.12 // indirect
```
janos, line 9: We should not need go-ethereum anymore.

pkg/shed/db.go
janos: Please, remove the unused argument.

pkg/shed/field_struct.go
janos: Do not log error and return it.

pkg/shed/field_uint64.go
janos: Do not log error and return it.

pkg/shed/index.go
```diff
@@ -162,32 +161,35 @@ func (f Index) Get(keyFields Item) (out Item, err error) {
...
 			f.logger.Debugf("keyfields encoding error in Fill. Error: %s", err.Error())
 			return err
 		}
-		value, err := snapshot.Get(key, nil)
+		badgerItem, err := txn.Get(key)
 		if err != nil {
+			if err == badger.ErrKeyNotFound {
+				return ErrNotFound
+			}
 			f.logger.Debugf("error getting key %s in Fill. Error: %s", string(key), err.Error())
```
janos, line 178: Do not log error and return it.

pkg/shed/index.go
janos: Do not log error and return it.

pkg/shed/index.go
janos: Do not log error and return it.

pkg/shed/vector_uint64.go
janos: Do not log error and return it.

pkg/shed/db.go
```diff
@@ -74,83 +88,381 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
-		return err
-	}
-	db.metrics.PutCounter.Inc()
-	return nil
+	db.logger.Tracef("Put value len : %d for key %s", len(value), string(key))
+	return db.bdb.Update(func(txn *badger.Txn) (err error) {
+		db.metrics.PutCount.Inc()
+		err = txn.Set(key, value)
+		if err != nil {
+			db.logger.Errorf("Could not insert chunk with key %s in DB", string(key))
```
janos, line 104: Do not log error and return it.

pkg/shed/db.go
```diff
@@ -74,83 +88,381 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+		return item.Value(func(val []byte) error {
+			value = val
+			db.logger.Tracef("got valeu for key %s with length %d", string(key), len(val))
+			return nil
+		})
+	})
+	if err != nil {
 		db.logger.Errorf("key %s not found in DB", string(key))
-		db.metrics.GetFailCounter.Inc()
+		db.logger.Debug("key %s not found in DB. Error : %v", string(key), err.Error())
```
janos, line 137: Do not log error and return it.

pkg/shed/db.go
```diff
@@ -74,83 +88,381 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+			db.metrics.HasCount.Inc()
+			return nil
+		})
+	})
 	if err != nil {
-		db.logger.Debugf("encountered error during HAS of key %s. Error: %s ", string(key), err.Error())
-		db.metrics.HasFailCounter.Inc()
-		return false, err
+		db.logger.Errorf("key %s not found in DB", string(key))
+		db.logger.Debugf("key %s not found in DB. Error: %v", string(key), err.Error())
```
janos, line 167: Do not log error and never return it fi the function should return it (unless there is an explicit reason for this). Just return it.

jarifibrahim:
We fixed something similar in badger v2.0.3. I'll look at v1.6.1 and get back. :+1:

pkg/shed/db.go
```diff
@@ -74,83 +88,381 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
-	if err != nil {
-		db.logger.Debugf("could not DELETE key %s. Error: %s ", string(key), err.Error())
-		db.metrics.DeleteFailCounter.Inc()
+	db.logger.Tracef("deleting key %s along with value", string(key))
+	return db.bdb.Update(func(txn *badger.Txn) (err error) {
+		db.metrics.DeleteCount.Inc()
+		err = txn.Delete(key)
+		if err != nil {
+			db.logger.Errorf("could not delete key %s from DB", string(key))
+			db.logger.Debugf("could not delete key %s from DB. Error: %v", string(key), err.Error())
```
janos, line 188: Do not log error and return it.

pkg/shed/db.go
```diff
@@ -74,83 +88,381 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+			if k < 1 {
+				continue
+			}
+			count++
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error while counting keys in DB")
+		db.logger.Debugf("error while doing count. Error : %w", err.Error())
```
janos, line 218: Do not log error and return it.

pkg/shed/db.go
```diff
@@ -74,83 +88,381 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+					break
+				}
+			}
+			count++
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error while doing count. prefix : %v,", string(prefix))
+		db.logger.Debugf("error while doing count for prefix : %v, Error : %w", string(prefix), err.Error())
```
janos, line 254: Do not log error and return it.

pkg/shed/db.go
```diff
@@ -74,83 +88,381 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+			if stop {
+				return nil
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error while doing iteration. startKey : %v, skipStartKey: %t",
+			string(startKey), skipStartKey)
+		db.logger.Debugf("error while doing iteration. startKey : %v, skipStartKey: %t, Error : %w",
```
janos, line 336: Do not log error and return it.

pkg/shed/db.go
```diff
@@ -74,83 +88,381 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+		}
+		value, err = i.Item().ValueCopy(value)
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error while finding first key with prefix : %v, Error : %w", string(prefix), err.Error())
+		db.logger.Debugf("error while doing first. prefix : %v, Error : %w", string(prefix), err.Error())
```
janos, line 370: Do not log error and return it.

pkg/shed/db.go
```diff
@@ -74,83 +88,381 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+		}
+
+		if key == nil {
+			return ErrNotFound
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error while finding last value for prefix : %v", string(prefix))
+		db.logger.Debugf("error while doing last. prefix : %v, Error : %w", string(prefix), err.Error())
```
janos, line 433: Do not log error and return it.

pkg/shed/db.go
```diff
@@ -74,83 +88,381 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
-	err = db.ldb.Write(batch, nil)
+// WriteBatch commits the badger transaction after all the operations are over.
+func (db *DB) WriteBatch(txn *badger.Txn) (err error) {
+	db.metrics.WriteBatchCount.Inc()
+	err = txn.Commit()
 	if err != nil {
-		db.logger.Debugf("could not writing batch. Error: %s ", err.Error())
-		db.metrics.WriteBatchFailCounter.Inc()
+		db.logger.Errorf("error while committing transaction. Error : %w", err.Error())
+		db.logger.Debugf("error while committing batch. Error : %w", err.Error())
```
janos, line 455: Do not log error and return it.

janos, line 455: Do not log error and return it.

jmozah:
> We fixed something similar in badger v2.0.3. I'll look at v1.6.1 and get back. 👍

@jarifibrahim  i took this as anopportunityy and moved to 2.0.3 :-)

pkg/shed/db.go
jmozah: ok. Add this in an else

pkg/shed/db.go
jmozah: add this in an else.

pkg/shed/db.go
jmozah: did that exactly

pkg/shed/index.go
jmozah: fixed this.

jmozah:
@janos fixed all the comments. The build fails and it is not related to my changes.

janos:
@jmozah I ma sorry for no being clear enough. I think that no logging should be done related to errors if the error is returned. Could you please remove Errorf calls whenever you removed Debugf calls in the previous commit? They are not needed.

pkg/shed/field_string.go
```diff
@@ -65,6 +65,11 @@ func (f StringField) Put(val string) (err error) {
 
 // PutInBatch stores a string in a batch that can be
 // saved later in database.
-func (f StringField) PutInBatch(batch *leveldb.Batch, val string) {
-	batch.Put(f.key, []byte(val))
+func (f StringField) PutInBatch(batch *badger.Txn, val string) (err error) {
+	err = batch.Set(f.key, []byte(val))
+	if err != nil {
+		f.logger.Debugf("could not set values in PutInBatch. Error : %s", err.Error())
```
janos, line 71: Do not log error and return it.

pkg/shed/field_struct.go
```diff
@@ -67,12 +67,16 @@ func (f StructField) Put(val interface{}) (err error) {
 }
 
 // PutInBatch marshals provided val and puts it into the batch.
-func (f StructField) PutInBatch(batch *leveldb.Batch, val interface{}) (err error) {
+func (f StructField) PutInBatch(batch *badger.Txn, val interface{}) (err error) {
 	b, err := json.Marshal(val)
 	if err != nil {
-		f.logger.Debugf("could not PUT key %s in batch", string(f.key))
+		f.logger.Debugf("could not marshall while doing putiInBatch")
```
janos, line 73: Error is returned, this log line is not needed.

pkg/shed/field_struct.go
```diff
@@ -67,12 +67,16 @@ func (f StructField) Put(val interface{}) (err error) {
...
+		return err
+	}
+	err = batch.Set(f.key, b)
+	if err != nil {
+		f.logger.Debugf("could not put values in batch.")
```
janos, line 78: Error is returned, this log line is not needed.

pkg/shed/field_uint64.go
```diff
@@ -67,20 +67,25 @@ func (f Uint64Field) Put(val uint64) (err error) {
 
 // PutInBatch stores a uint64 value in a batch
 // that can be saved later in the database.
-func (f Uint64Field) PutInBatch(batch *leveldb.Batch, val uint64) {
-	batch.Put(f.key, encodeUint64(val))
+func (f Uint64Field) PutInBatch(batch *badger.Txn, val uint64) (err error){
+	err = batch.Set(f.key, encodeUint64(val))
+	if err != nil {
+		f.logger.Debugf("could not set uint64 value in PutInBatch",)
```
janos, line 73: Error is returned, this log line is not needed.

pkg/shed/field_uint64.go
```diff
@@ -67,20 +67,25 @@ func (f Uint64Field) Put(val uint64) (err error) {
...
 func (f Uint64Field) Inc() (val uint64, err error) {
 	val, err = f.Get()
 	if err != nil {
-		if err == leveldb.ErrNotFound {
+		if err == ErrNotFound {
 			f.logger.Debugf("key %s not found", string(f.key))
 			val = 0
 		} else {
-			f.logger.Errorf("key %s not found. Error: %s", string(f.key), err.Error())
+			f.logger.Errorf("key %s not found", string(f.key))
```
janos, line 88: Error is returned, this log line is not needed.

pkg/shed/index.go
```diff
@@ -141,17 +140,17 @@ func (db *DB) NewIndex(name string, funcs IndexFuncs, logger logging.Logger) (f
 func (f Index) Get(keyFields Item) (out Item, err error) {
 	key, err := f.encodeKeyFunc(keyFields)
 	if err != nil {
-		f.logger.Debugf("keyfields encoding error in Get. Error: %s", err.Error())
+		f.logger.Debugf("keyfields encoding error in Get")
```
janos, line 143: Error is returned, this log line is not needed.

pkg/shed/index.go
```diff
@@ -141,17 +140,17 @@ func (db *DB) NewIndex(name string, funcs IndexFuncs, logger logging.Logger) (f
...
 		return out, err
 	}
 	value, err := f.db.Get(key)
 	if err != nil {
-		f.logger.Debugf("error getting key %s in Get. Error: %s", string(key), err.Error())
+		f.logger.Debugf("error getting key %s in Get", string(key))
```
janos, line 148: Error is returned, this log line is not needed.

pkg/shed/index.go
```diff
@@ -141,17 +140,17 @@ func (db *DB) NewIndex(name string, funcs IndexFuncs, logger logging.Logger) (f
...
 		return out, err
 	}
 	out, err = f.decodeValueFunc(keyFields, value)
 	if err != nil {
-		f.logger.Debugf("error decofing keyfields in Get. Error: %s", err.Error())
+		f.logger.Debugf("error decofing keyfields in Get")
```
janos, line 153: Error is returned, this log line is not needed.

pkg/shed/index.go
```diff
@@ -198,7 +197,7 @@ func (f Index) Fill(items []Item) (err error) {
 func (f Index) Has(keyFields Item) (bool, error) {
 	key, err := f.encodeKeyFunc(keyFields)
 	if err != nil {
-		f.logger.Debugf("keyfields encoding error in Has. Error: %s", err.Error())
+		f.logger.Debugf("keyfields encoding error in Has.")
```
janos, line 200: Error is returned, this log line is not needed.

pkg/shed/index.go
```diff
@@ -234,12 +240,12 @@ func (f Index) HasMulti(items ...Item) ([]bool, error) {
 func (f Index) Put(i Item) (err error) {
 	key, err := f.encodeKeyFunc(i)
 	if err != nil {
-		f.logger.Debugf("keyfields encoding error in Put. Error: %s", err.Error())
+		f.logger.Debugf("keyfields encoding error in Put")
```
janos, line 243: Error is returned, this log line is not needed.

pkg/shed/index.go
```diff
@@ -234,12 +240,12 @@ func (f Index) HasMulti(items ...Item) ([]bool, error) {
...
 		return err
 	}
 	value, err := f.encodeValueFunc(i)
 	if err != nil {
-		f.logger.Debugf("keyfields encoding error in Put. Error: %s", err.Error())
+		f.logger.Debugf("keyfields encoding error in Put")
```
janos, line 248: Error is returned, this log line is not needed.

pkg/shed/index.go
```diff
@@ -248,42 +254,50 @@ func (f Index) Put(i Item) (err error) {
 // PutInBatch is the same as Put method, but it just
 // saves the key/value pair to the batch instead
 // directly to the database.
-func (f Index) PutInBatch(batch *leveldb.Batch, i Item) (err error) {
+func (f Index) PutInBatch(batch *badger.Txn, i Item) (err error) {
 	key, err := f.encodeKeyFunc(i)
 	if err != nil {
-		f.logger.Debugf("keyfields encoding error in PutInBatch. Error: %s", err.Error())
+		f.logger.Debugf("keyfields encoding error in PutInBatch")
```
janos, line 260: Error is returned, this log line is not needed.

pkg/shed/index.go
```diff
@@ -248,42 +254,50 @@ func (f Index) Put(i Item) (err error) {
...
 		return err
 	}
 	value, err := f.encodeValueFunc(i)
 	if err != nil {
-		f.logger.Debugf("keyfields encoding error in PutInBatch. Error: %s", err.Error())
+		f.logger.Debugf("keyfields encoding error in PutInBatch")
```
janos, line 265: Error is returned, this log line is not needed.

pkg/shed/index.go
```diff
@@ -248,42 +254,50 @@ func (f Index) Put(i Item) (err error) {
...
 		return err
 	}
-	batch.Put(key, value)
-	return nil
+	err = batch.Set(key, value)
+	if err != nil {
+		f.logger.Debugf("could not set values in batch")
```
janos, line 270: Error is returned, this log line is not needed.

pkg/shed/index.go
```diff
@@ -248,42 +254,50 @@ func (f Index) Put(i Item) (err error) {
...
+	return err
 }
 
 // Delete accepts Item to remove a key/value pair
 // from the database based on its fields.
 func (f Index) Delete(keyFields Item) (err error) {
 	key, err := f.encodeKeyFunc(keyFields)
 	if err != nil {
-		f.logger.Debugf("keyfields encoding error in Delete. Error: %s", err.Error())
+		f.logger.Debugf("keyfields encoding error in Delete",)
```
janos, line 281: Error is returned, this log line is not needed.

pkg/shed/index.go
```diff
@@ -248,42 +254,50 @@ func (f Index) Put(i Item) (err error) {
...
 }
 
 // DeleteInBatch is the same as Delete just the operation
 // is performed on the batch instead on the database.
-func (f Index) DeleteInBatch(batch *leveldb.Batch, keyFields Item) (err error) {
+func (f Index) DeleteInBatch(batch *badger.Txn, keyFields Item) (err error) {
 	key, err := f.encodeKeyFunc(keyFields)
 	if err != nil {
-		f.logger.Debugf("keyfields encoding error in DeleteInBatch. Error: %s", err.Error())
+		f.logger.Debugf("keyfields encoding error in DeleteInBatch")
```
janos, line 292: Error is returned, this log line is not needed.

pkg/shed/index.go
```diff
@@ -248,42 +254,50 @@ func (f Index) Put(i Item) (err error) {
...
 		return err
 	}
-	batch.Delete(key)
-	return nil
+	err = batch.Delete(key)
+	if err != nil {
+		f.logger.Debugf("could not delete items in batch")
```
janos, line 297: Error is returned, this log line is not needed.

pkg/shed/vector_uint64.go
```diff
@@ -66,16 +66,21 @@ func (f Uint64Vector) Put(i, val uint64) (err error) {
 
 // PutInBatch stores a uint64 value at index i in a batch
 // that can be saved later in the database.
-func (f Uint64Vector) PutInBatch(batch *leveldb.Batch, i, val uint64) {
-	batch.Put(f.indexKey(i), encodeUint64(val))
+func (f Uint64Vector) PutInBatch(batch *badger.Txn, i, val uint64) (err error){
+	err = batch.Set(f.indexKey(i), encodeUint64(val))
+	if err != nil {
+		f.logger.Debugf("could not put uint64 value in batch")
```
janos, line 72: Error is returned, this log line is not needed. Just return the error from batch.Set.

pkg/shed/db.go
janos: This log line is not needed, as the error is returned.

pkg/shed/db.go
```diff
@@ -74,83 +87,372 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+			if err == badger.ErrKeyNotFound {
+				db.metrics.GetNotFoundCount.Inc()
+				return ErrNotFound
+			}
+			db.metrics.GetFailCount.Inc()
+			return err
+		}
+		return item.Value(func(val []byte) error {
+			value = val
+			db.logger.Tracef("got valeu for key %s with length %d", string(key), len(val))
```
janos, line 130: valeu -> value

pkg/shed/db.go
```diff
@@ -74,83 +87,372 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+			return nil
+		})
+	})
+	if err != nil {
 		db.logger.Errorf("key %s not found in DB", string(key))
```
janos, line 135: This log line is not needed, as the error is returned.

pkg/shed/db.go
```diff
@@ -74,83 +87,372 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+			yes = true
+			db.metrics.HasCount.Inc()
+			return nil
+		})
+	})
 	if err != nil {
-		db.logger.Debugf("encountered error during HAS of key %s. Error: %s ", string(key), err.Error())
-		db.metrics.HasFailCounter.Inc()
-		return false, err
+		db.logger.Errorf("key %s not found in DB", string(key))
```
janos, line 163: This log line is not needed.

The error is not returned at all please return it, and remove the log line.

pkg/shed/db.go
```diff
@@ -74,83 +87,372 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
-	err = db.ldb.Delete(key, nil)
-	if err != nil {
-		db.logger.Debugf("could not DELETE key %s. Error: %s ", string(key), err.Error())
-		db.metrics.DeleteFailCounter.Inc()
+	db.logger.Tracef("deleting key %s along with value", string(key))
+	return db.bdb.Update(func(txn *badger.Txn) (err error) {
+		db.metrics.DeleteCount.Inc()
+		err = txn.Delete(key)
+		if err != nil {
+			db.logger.Errorf("could not delete key %s from DB", string(key))
```
janos, line 183: This log line is not needed, as the error is returned.

pkg/shed/db.go
```diff
@@ -74,83 +87,372 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+			k := item.KeySize()
+			if k < 1 {
+				continue
+			}
+			count++
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error while counting keys in DB")
```
janos, line 212: This log line is not needed, as the error is returned.

pkg/shed/db.go
```diff
@@ -74,83 +87,372 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+				if !bytes.HasPrefix(k, prefix) {
+					break
+				}
+			}
+			count++
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error while doing count. prefix : %v,", string(prefix))
```
janos, line 248: This log line is not needed, as the error is returned.

pkg/shed/db.go
```diff
@@ -74,83 +87,372 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+		defer i.Close()
+
+		// if prefix is nil, it is equivalent to counting from beginning
+		for i.Seek(prefix); i.Valid(); i.Next() {
+			count++
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error doing count from. prefix : %v", string(prefix))
```
janos, line 278: This log line is not needed, as the error is returned.

pkg/shed/db.go
```diff
@@ -74,83 +87,372 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+				return err
+			}
+			if stop {
+				return nil
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error while doing iteration. startKey : %v, skipStartKey: %t",
```
janos, line 329: This log line is not needed, as the error is returned.

pkg/shed/db.go
```diff
@@ -74,83 +87,372 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+			return ErrNotFound
+		}
+		value, err = i.Item().ValueCopy(value)
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error while finding first key with prefix : %v", string(prefix))
```
janos, line 362: This log line is not needed, as the error is returned.

pkg/shed/db.go
```diff
@@ -74,83 +87,372 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+			}
+		}
+
+		if key == nil {
+			return ErrNotFound
+		}
+		return nil
+	})
+	if err != nil {
+		db.logger.Errorf("error while finding last value for prefix : %v", string(prefix))
```
janos, line 424: This log line is not needed, as the error is returned.

pkg/shed/db.go
```diff
@@ -74,83 +87,372 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
-func (db *DB) WriteBatch(batch *leveldb.Batch) (err error) {
-	err = db.ldb.Write(batch, nil)
+// WriteBatch commits the badger transaction after all the operations are over.
+func (db *DB) WriteBatch(txn *badger.Txn) (err error) {
+	db.metrics.WriteBatchCount.Inc()
+	err = txn.Commit()
 	if err != nil {
-		db.logger.Debugf("could not writing batch. Error: %s ", err.Error())
-		db.metrics.WriteBatchFailCounter.Inc()
+		db.logger.Errorf("error while committing transaction",)
```
janos, line 445: This log line is not needed, as the error is returned.

jmozah:
@janos coding guidelines says "This means that the same problematic event may have two log lines but with different severities" thats why i had Errorf and Debugf messages. This is still not clear for me where to put Error and where to put debug and where to put both.

janos:
@jmozah It also states in https://github.com/ethersphere/bee/blob/master/CODING.md#error-propagation-annotation-and-handling section that errors should not be logged if they are returned. `Error should not be logged and passed at the same time.` That is the point of my comments. Sorry if that was not clear.

jmozah:
@janos no issues. it is clear now for me. I made the changes.

janos:
> @janos no issues. it is clear now for me. I made the changes.

Thanks @jmozah

pkg/shed/field_string.go
```diff
@@ -65,6 +65,10 @@ func (f StringField) Put(val string) (err error) {
...
 // PutInBatch stores a string in a batch that can be
 // saved later in database.
-func (f StringField) PutInBatch(batch *leveldb.Batch, val string) {
-	batch.Put(f.key, []byte(val))
+func (f StringField) PutInBatch(batch *badger.Txn, val string) (err error) {
+	err = batch.Set(f.key, []byte(val))
+	if err != nil {
+		return err
+	}
+	return nil
```
janos, line 73:
```suggestion
	return batch.Set(f.key, []byte(val))
```

pkg/shed/field_uint64.go
```diff
@@ -67,20 +67,23 @@ func (f Uint64Field) Put(val uint64) (err error) {
...
 // PutInBatch stores a uint64 value in a batch
 // that can be saved later in the database.
-func (f Uint64Field) PutInBatch(batch *leveldb.Batch, val uint64) {
-	batch.Put(f.key, encodeUint64(val))
+func (f Uint64Field) PutInBatch(batch *badger.Txn, val uint64) (err error){
+	err = batch.Set(f.key, encodeUint64(val))
+	if err != nil {
+		return err
+	}
+	return nil
```
janos, line 75:
```suggestion
	return batch.Set(f.key, encodeUint64(val))
```

pkg/shed/index.go
```diff
@@ -248,42 +248,44 @@ func (f Index) Put(i Item) (err error) {
...
-		f.logger.Debugf("keyfields encoding error in PutInBatch. Error: %s", err.Error())
 		return err
 	}
-	batch.Put(key, value)
-	return nil
+	err = batch.Set(key, value)
+	if err != nil {
+		return err
+	}
+	return err
```
janos, line 264:
```suggestion
	return batch.Set(key, value)
```

pkg/shed/index.go
```diff
@@ -248,42 +248,44 @@ func (f Index) Put(i Item) (err error) {
...
-		f.logger.Debugf("keyfields encoding error in DeleteInBatch. Error: %s", err.Error())
 		return err
 	}
-	batch.Delete(key)
-	return nil
+	err = batch.Delete(key)
+	if err != nil {
+		return err
+	}
+	return err
```
janos, line 288:
```suggestion
	return batch.Delete(key)
```

pkg/shed/vector_uint64.go
```diff
@@ -66,16 +66,20 @@ func (f Uint64Vector) Put(i, val uint64) (err error) {
...
 // PutInBatch stores a uint64 value at index i in a batch
 // that can be saved later in the database.
-func (f Uint64Vector) PutInBatch(batch *leveldb.Batch, i, val uint64) {
-	batch.Put(f.indexKey(i), encodeUint64(val))
+func (f Uint64Vector) PutInBatch(batch *badger.Txn, i, val uint64) (err error){
+	err = batch.Set(f.indexKey(i), encodeUint64(val))
+	if err != nil {
+		return err
+	}
+	return err
```
janos, line 74:
```suggestion
	return batch.Set(f.indexKey(i), encodeUint64(val))
```

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+			db.metrics.HasCount.Inc()
+			return nil
+		})
+	})
 	if err != nil {
-		db.logger.Debugf("encountered error during HAS of key %s. Error: %s ", string(key), err.Error())
-		db.metrics.HasFailCounter.Inc()
-		return false, err
+		db.metrics.HasFailCount.Inc()
+	}
```
janos, line 160: The error from line 144 is never returned. I think that it should be in this if block.

pkg/shed/db.go
janos: This comment left unaddressed.

go.sum
```diff
@@ -95,10 +99,17 @@ github.com/deckarep/golang-set v0.0.0-20180603214616-504e848d77ea h1:j4317fAZh7X
 github.com/deckarep/golang-set v0.0.0-20180603214616-504e848d77ea/go.mod h1:93vsz/8Wt4joVM7c2AVqh+YRMiUSc14yDtF28KmMOgQ=
 github.com/dgraph-io/badger v1.5.5-0.20190226225317-8115aed38f8f/go.mod h1:VZxzAIRPHRVNRKRo6AXrX9BJegn6il06VMTZVJYCIjQ=
 github.com/dgraph-io/badger v1.6.0-rc1/go.mod h1:zwt7syl517jmP8s94KqSxTlM6IMsdhYy6psNgSztDR4=
+github.com/dgraph-io/badger v1.6.0 h1:DshxFxZWXUcO0xX476VJC07Xsr6ZCBVRHKZ93Oh7Evo=
```
acud, line 102: i think this version is not needed anymore...

go.sum
```diff
@@ -109,14 +120,17 @@ github.com/docker/go-connections v0.4.0/go.mod h1:Gbd7IOopHjR8Iph03tsViu4nIes5Xh
...
 github.com/dustin/go-humanize v1.0.0/go.mod h1:HtrtbFcZ19U5GC7JDqmcUSB87Iq5E25KnS6fMYU6eOk=
 github.com/edsrzf/mmap-go v0.0.0-20160512033002-935e0e8a636c/go.mod h1:YO35OhQPt3KJa3ryjFM5Bs14WD66h8eGKpfaBNrHW5M=
+github.com/elastic/gosigar v0.0.0-20180330100440-37f05ff46ffa h1:o8OuEkracbk3qH6GvlI6XpEN1HTSxkzOG42xZpfDv/s=
 github.com/elastic/gosigar v0.0.0-20180330100440-37f05ff46ffa/go.mod h1:cdorVVzy1fhmEqmtgqkoE3bYtCfSCkVyjTyCIo22xvs=
 github.com/elastic/gosigar v0.8.1-0.20180330100440-37f05ff46ffa h1:XKAhUk/dtp+CV0VO6mhG2V7jA9vbcGcnYF/Ay9NjZrY=
 github.com/elastic/gosigar v0.8.1-0.20180330100440-37f05ff46ffa/go.mod h1:cdorVVzy1fhmEqmtgqkoE3bYtCfSCkVyjTyCIo22xvs=
 github.com/elazarl/goproxy v0.0.0-20170405201442-c4fc26588b6e/go.mod h1:/Zj4wYkgs4iZTTu3o/KG3Itv/qCCa8VVMlb3i9OVuzc=
 github.com/envoyproxy/go-control-plane v0.9.1-0.20191026205805-5f8ba28d4473/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
 github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
+github.com/ethereum/go-ethereum v1.9.2 h1:RMIHDO/diqXEgORSVzYx8xW9x2+S32PoAX5lQwya0Lw=
```
acud, line 133: this go ethereum stuff probably not too

pkg/shed/db.go
```diff
@@ -23,45 +23,58 @@
...
-		OpenFilesCacheCapacity: openFileLimit,
-	})
+	o := badger.DefaultOptions(path)
+	o.SyncWrites = DefaultSyncWrites
+	o.ValueThreshold = DefaultValueThreshold
+	o.ValueLogMaxEntries = DefaultValueLogMaxEntries
+	o.Logger = nil // Dont enable the badger logs
+	database, err := badger.Open(o)
 	if err != nil {
+		logger.Error("could not open database.")
```
acud, line 65: not needed to log _and_ return the error. enough just to return the error in this case

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
 		}
 	}
 
-	// Create a quit channel for the periodic metrics collector and run it
-	db.quit = make(chan struct{})
-
+	db.metrics.DBOpenCount.Inc()
+	logger.Debugf("Opened DB with path : %s, valueThreshold : %d, valueLogMaxEntries : %d",
```
acud, line 91: I don't think this is needed

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
 		}
 	}
 
-	// Create a quit channel for the periodic metrics collector and run it
-	db.quit = make(chan struct{})
-
+	db.metrics.DBOpenCount.Inc()
```
acud, line 90: not sure if this metric is all that useful. this should always be 1, at least for the time being...

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
 func (db *DB) Put(key []byte, value []byte) (err error) {
-	err = db.ldb.Put(key, value, nil)
-	if err != nil {
-		db.logger.Debugf("failed to insert in to DB. Error : %s",  err.Error())
-		db.metrics.PutFailCounter.Inc()
-		return err
-	}
-	db.metrics.PutCounter.Inc()
-	return nil
+	db.logger.Tracef("Put value len : %d for key %s", len(value), string(key))
```
acud, line 98: this will flood the logs on `trace` (imagine thousands of these lines when you upload and sync). i think this is not needed

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+	return db.bdb.Update(func(txn *badger.Txn) (err error) {
+		db.metrics.PutCount.Inc()
+		err = txn.Set(key, value)
+		if err != nil {
+			db.metrics.PutFailCount.Inc()
+			return err
+		}
+		db.logger.Tracef(" put success with key %s and value len %d", string(key), len(value))
```
acud, line 106: here too

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
 func (db *DB) Get(key []byte) (value []byte, err error) {
-	value, err = db.ldb.Get(key, nil)
-	if err == leveldb.ErrNotFound {
-		db.logger.Debugf("key %s not found during GET", string(key))
-		db.metrics.GetNotFoundCounter.Inc()
-		return nil, err
-	} else {
-		db.logger.Errorf("key %s not found in DB", string(key))
-		db.metrics.GetFailCounter.Inc()
+	db.logger.Tracef("Get value for key %s", string(key))
```
acud, line 115: again

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+			db.metrics.GetFailCount.Inc()
+			return err
+		}
+		return item.Value(func(val []byte) error {
+			value = val
+			return nil
+		})
+	})
+	if err == nil {
+		db.logger.Tracef("got value with len : %d for key : %s", len(value), string(key))
```
acud, line 133: i dont think this is needed. this will flood

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
-	return value, nil
+	return value, err
 }
 
-// Has wraps LevelDB Has method to increment metrics counter.
+// Has checks if the given key is present in the database.
+// it returns a bool indicating true or false OR error if it encounters one during the operation.
 func (db *DB) Has(key []byte) (yes bool, err error) {
-	yes, err = db.ldb.Has(key, nil)
+	db.logger.Tracef("Has value with key %s", string(key))
```
acud, line 142: ditto

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
 	if err != nil {
-		db.logger.Debugf("encountered error during HAS of key %s. Error: %s ", string(key), err.Error())
-		db.metrics.HasFailCounter.Inc()
+		db.metrics.HasFailCount.Inc()
 		return false, err
 	}
-	db.metrics.HasCounter.Inc()
+
+	if yes {
+		db.logger.Tracef("found key %s", string(key))
```
acud, line 164: ditto

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+	} else {
+		db.logger.Tracef("could not found key %s", string(key))
```
acud, line 166: ditto

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
 }
 
-// Delete wraps LevelDB Delete method to increment metrics counter.
+// Delete removed the key and value if a given key is present in the DB.
 func (db *DB) Delete(key []byte) (err error) {
-	err = db.ldb.Delete(key, nil)
-	if err != nil {
-		db.logger.Debugf("could not DELETE key %s. Error: %s ", string(key), err.Error())
-		db.metrics.DeleteFailCounter.Inc()
+	db.logger.Tracef("deleting key %s along with value", string(key))
```
acud, line 174: also

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+	return db.bdb.Update(func(txn *badger.Txn) (err error) {
+		db.metrics.DeleteCount.Inc()
+		err = txn.Delete(key)
+		if err != nil {
+			db.metrics.DeleteFailCount.Inc()
+		} else {
+			db.logger.Tracef("deleted key %s", string(key))
```
acud, line 181: you get the point

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+		}
 		return err
+	})
+}
+
+// Count gives a count of all the keys present in the DB.
+func (db *DB) Count(ctx context.Context) (count int, err error) {
+	db.logger.Tracef("counting all keys")
```
acud, line 189: not needed

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+				continue
+			}
+			count++
+		}
+		return nil
+	})
+	if err != nil {
+		db.metrics.TotalFailCount.Inc()
+	} else {
+		db.logger.Tracef("Total count is %d", count)
```
acud, line 209: i dont think this is needed when the count is returned to the calling context

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
-	db.metrics.DeleteCounter.Inc()
-	return nil
+	return count, err
+}
+
+// CountPrefix gives a count of all the keys that starts with a given key prefix.
+// a nil prefix acts like the total count of the DB
+func (db *DB) CountPrefix(prefix []byte) (count int, err error) {
+	if prefix != nil {
+		db.logger.Tracef("counting all keys with prefix %s", string(prefix))
```
acud, line 218: not sure if this is so helpful

pkg/shed/db.go
```diff
@@ -74,83 +87,359 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
+				}
+			}
+			count++
+		}
+		return nil
+	})
+	if err != nil {
+		db.metrics.CountPrefixFailCount.Inc()
+	} else {
+		db.logger.Tracef("Total count for prefix %s is %d", string(prefix), count)
```
acud, line 244: again

jmozah:
Bee is still dependent on Swarm and Ethereum as the example_store_test.go in shed uses it.
Once we move chun related stuff inside bee.. we will get rid of Swarm and ethereum.

go.sum
janos: It looks like that we still depend on swarm/storage in shed example, so tis sum is added here. We should remove completely swarm dependencies in the future PRs. And then this line could be removed then.

go.sum
janos: It looks like that we still depend on swarm/storage in shed example, so tis sum is added here. We should remove completely swarm dependencies in the future PRs. And then this line could be removed then. Now go tool keeps adding it.

go.mod
```diff
@@ -1,33 +1,33 @@
 module github.com/ethersphere/bee
 
-go 1.14
+go 1.13
```
acud, line 3: should be `1.14`

pkg/shed/index.go
```diff
@@ -162,32 +158,32 @@ func (f Index) Get(keyFields Item) (out Item, err error) {
...
 			return err
 		}
-		value, err := snapshot.Get(key, nil)
+		badgerItem, err := txn.Get(key)
 		if err != nil {
-			f.logger.Debugf("error getting key %s in Fill. Error: %s", string(key), err.Error())
+			if err == badger.ErrKeyNotFound {
+				return ErrNotFound
+			}
+			f.logger.Debugf("error getting key %s in Fill.", string(key))
```
acud, line 175: can also be removed, since error is returned

pkg/shed/vector_uint64.go
```diff
@@ -127,10 +130,10 @@ func (f Uint64Vector) Dec(i uint64) (val uint64, err error) {
...
 // The field is protected from overflow to a negative value.
-func (f Uint64Vector) DecInBatch(batch *leveldb.Batch, i uint64) (val uint64, err error) {
+func (f Uint64Vector) DecInBatch(batch *badger.Txn, i uint64) (val uint64, err error) {
 	val, err = f.Get(i)
 	if err != nil {
-		if err == leveldb.ErrNotFound {
+		if err == ErrNotFound {
 			val = 0
 		} else {
 			f.logger.Debugf("error getiing value while doing DecInBatch. Error: %s", err.Error())
```
acud, line 139: this too is not necessary since error is returned

pkg/shed/vector_uint64.go
```diff
@@ -89,18 +89,21 @@ func (f Uint64Vector) Inc(i uint64) (val uint64, err error) {
...
 // This operation is not goroutine safe.
-func (f Uint64Vector) IncInBatch(batch *leveldb.Batch, i uint64) (val uint64, err error) {
+func (f Uint64Vector) IncInBatch(batch *badger.Txn, i uint64) (val uint64, err error) {
 	val, err = f.Get(i)
 	if err != nil {
-		if err == leveldb.ErrNotFound {
+		if err == ErrNotFound {
 			val = 0
 		} else {
 			f.logger.Debugf("error getiing value while doing IncInBatch. Error: %s", err.Error())
```
acud, line 98: same

pkg/shed/vector_uint64.go
```diff
@@ -66,16 +66,16 @@ func (f Uint64Vector) Put(i, val uint64) (err error) {
...
 // Inc increments a uint64 value in the database.
 // This operation is not goroutine safe.
 func (f Uint64Vector) Inc(i uint64) (val uint64, err error) {
 	val, err = f.Get(i)
 	if err != nil {
-		if err == leveldb.ErrNotFound {
+		if err == ErrNotFound {
 			val = 0
 		} else {
 			f.logger.Debugf("error getiing value while doing Inc. Error: %s", err.Error())
```
acud, line 81: also

pkg/shed/index.go
```diff
@@ -322,102 +316,66 @@ func (f Index) Iterate(fn IndexIterFunc, options *IterateOptions) (err error) {
...
 	if err != nil {
 		f.logger.Debugf("error decoding key in itemFromIterator. Error: %s", err.Error())
 		return i, err
 	}
-	// create a copy of value byte slice not to share leveldb underlaying slice array
-	valueItem, err := f.decodeValueFunc(keyItem, append([]byte(nil), it.Value()...))
+	// create a copy of value byte slice not to share badger underlaying slice array
+	valueItem, err := f.decodeValueFunc(keyItem, append([]byte(nil), value...))
 	if err != nil {
 		f.logger.Debugf("error decoding value in itemFromIterator. Error: %s", err.Error())
```
acud, line 363: error returned, no need to log

pkg/shed/index.go
```diff
@@ -322,102 +316,66 @@ func (f Index) Iterate(fn IndexIterFunc, options *IterateOptions) (err error) {
...
+func (f Index) itemFromKeyValue(key []byte, value []byte, totalPrefix []byte) (i Item, err error) {
 	if !bytes.HasPrefix(key, totalPrefix) {
-		return i, leveldb.ErrNotFound
+		return i, badger.ErrKeyNotFound
 	}
-	// create a copy of key byte slice not to share leveldb underlaying slice array
+	// create a copy of key byte slice not to share badger underlaying slice array
 	keyItem, err := f.decodeKeyFunc(append([]byte(nil), key...))
 	if err != nil {
 		f.logger.Debugf("error decoding key in itemFromIterator. Error: %s", err.Error())
```
acud, line 357: error returned, no need to log

pkg/shed/field_uint64.go
```diff
@@ -131,10 +133,10 @@ func (f Uint64Field) Dec() (val uint64, err error) {
 // by retreiving a value from the database, not the same batch.
 // This operation is not goroutine save.
 // The field is protected from overflow to a negative value.
-func (f Uint64Field) DecInBatch(batch *leveldb.Batch) (val uint64, err error) {
+func (f Uint64Field) DecInBatch(batch *badger.Txn) (val uint64, err error) {
 	val, err = f.Get()
 	if err != nil {
-		if err == leveldb.ErrNotFound {
+		if err == ErrNotFound {
 			f.logger.Debugf("key %s not found", string(f.key))
```
acud, line 140: not necessary probably

pkg/shed/field_uint64.go
```diff
@@ -131,10 +133,10 @@ func (f Uint64Field) Dec() (val uint64, err error) {
...
 			val = 0
 		} else {
```
acud, line 142: one line below, error returned no need to log

pkg/shed/field_uint64.go
```diff
@@ -113,7 +115,7 @@ func (f Uint64Field) IncInBatch(batch *leveldb.Batch) (val uint64, err error) {
 func (f Uint64Field) Dec() (val uint64, err error) {
 	val, err = f.Get()
 	if err != nil {
-		if err == leveldb.ErrNotFound {
+		if err == ErrNotFound {
 			f.logger.Debugf("key %s not found", string(f.key))
```
acud, line 119: same, same below

pkg/shed/field_uint64.go
```diff
@@ -91,10 +90,10 @@ func (f Uint64Field) Inc() (val uint64, err error) {
 // IncInBatch increments a uint64 value in the batch
 // by retreiving a value from the database, not the same batch.
 // This operation is not goroutine save.
-func (f Uint64Field) IncInBatch(batch *leveldb.Batch) (val uint64, err error) {
+func (f Uint64Field) IncInBatch(batch *badger.Txn) (val uint64, err error) {
 	val, err = f.Get()
 	if err != nil {
-		if err == leveldb.ErrNotFound {
+		if err == ErrNotFound {
 			f.logger.Debugf("key %s not found", string(f.key))
```
acud, line 97: no need to log here i think

pkg/shed/field_string.go
```diff
@@ -49,7 +49,7 @@ func (db *DB) NewStringField(name string, logger logging.Logger) (f StringField,
 func (f StringField) Get() (val string, err error) {
 	b, err := f.db.Get(f.key)
 	if err != nil {
-		if err == leveldb.ErrNotFound {
+		if err == ErrNotFound {
 			f.logger.Errorf("key %s not found", string(f.key))
```
acud, line 53: if there's an error, you would would probably want to return it no? here you log it, but then return an empty string but nil error. i'm not sure this is right

pkg/shed/db.go
```diff
@@ -73,84 +85,307 @@ func NewDB(path string, logger logging.Logger) (db *DB, err error) {
...
 		return err
 	}
-	db.metrics.WriteBatchCounter.Inc()
 	return nil
 }
 
-// Close closes LevelDB database.
+// Close shuts down the badger DB.
 func (db *DB) Close() (err error) {
-	close(db.quit)
```
acud, line 154: i wonder, we dont need this anymore?

pkg/shed/field_string.go
jmozah: ok. The idea was to split the error in to KeyNotFound and rest. because the application might want to respond differently... but since you guys seems to like the idea of just returning the err as it is, i am fixing in all those places.

pkg/shed/db.go
jmozah: Where is this?   we have a Close() at the end of the file which doesn't look like this.

```
// Close shuts down the badger DB.
func (db *DB) Close() (err error) {
	db.logger.Tracef("database closed with path %s", db.path)
	return db.bdb.Close()
}
```

pkg/shed/field_string_test.go
```diff
@@ -37,7 +37,7 @@ func TestStringField(t *testing.T) {
 
 	t.Run("get empty", func(t *testing.T) {
 		got, err := simpleString.Get()
-		if err != nil {
+		if err == nil {
 			t.Fatal(err)
```
acud, line 41: you'd probably want to write a custom error message here for the test, since you're failing the test with a `nil` value.

acud: I'm merging the PR as is, you can piggy-back this change on a later PR

jmozah: Thanks @acud

PR #78, opened 2020-04-08
add debug api overlay address endpoint

janos:
This PR changes Debug API /addresses endpoint to expose node's overlay address. This information will be needed for integration tests. Most notably the connectivity test that we are preparing for.

It also changes `addresses` field to `underlay` in response.


pkg/debugapi/router.go
```diff
@@ -43,6 +43,9 @@ func (s *server) setupRouting() {
 	router.Handle("/addresses", jsonhttp.MethodHandler{
 		"GET": http.HandlerFunc(s.addressesHandler),
 	})
+	router.Handle("/overlay-address", jsonhttp.MethodHandler{
```
acud, line 46: why not `/addresses/overlay`?

janos: That is a good question. The main reason is that `/addresses` is returning underlay addresses that do not contain overlay address.

Since you asked, maybe I rushed with the implementation. Would it make sense just to add a field `overlay` in `/addresses` response and rename `addresses` field to `underlay` in it?

acud: yes, this sounds like a great idea :+1:

janos: Thanks, I will change it.

janos: Done, @acud ready for another review.

PR #81, opened 2020-04-08
addressbook: add support for persistence

### LINKED ISSUE #59: Addressbook persistence

acud:
This PR adds possibility to persist the addressbook in `bee`.

The changes in a nutshell:
- instead of having both an in-mem address book and a persisting addressbook implementation, I have instead implemented an in memory state-store that can be reused across other modules too, and prevents duplicate code for in memory and persisting implementations of components
- tests have been added and generalized between state store implementations
- in mem implementation of address book has been adapted to a generalized implementation that just needs an injected state store (whether in-memory or not - it does not care obviously)
- `exists` on `addressbook.Put` signature has been removed - it is not used anywhere
- `PeerEntry` struct has been moved to `addressbook` package, custom marshaler and unmarshaler from/to JSON have been implemented due to some quirk on deserialization of multiaddress from JSON. we can solve this later

Still TODO:
- [x] Decide where to persist the state-store in relation to the entire node configuration. @janos I would need your input on this

fixes #59


pkg/addressbook/addressbook.go
```diff
@@ -5,20 +5,142 @@
...
-	Get(overlay swarm.Address) (addr ma.Multiaddr, exists bool)
+	Get(overlay swarm.Address) (addr ma.Multiaddr, err error)
 }
 
 type Putter interface {
-	Put(overlay swarm.Address, addr ma.Multiaddr) (exists bool)
+	Put(overlay swarm.Address, addr ma.Multiaddr) (err error)
+}
+
+var _ GetPutter = (*store)(nil)
```
pradovic, line 33: 🌷 I would put this at the top of the page, next to const.

pkg/addressbook/addressbook_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+		book := addressbook.New(store)
+
+		return book, func() {
+			os.RemoveAll(dir)
+		}
+	})
+}
+
+func run(t *testing.T, f bookFunc) {
```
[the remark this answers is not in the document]
pradovic: I agree as well, seems a bit funky to me. IMO, in this kind of unit tests, it's enought to use mock in-mem statestore. And testing of different implementations of statestore should have their own tests.

pkg/node/node.go
```diff
@@ -253,7 +255,15 @@ func NewBee(o Options) (*Bee, error) {
 				return
 			}
 
-			addressbook.Put(overlay, addr)
+			err = addressbook.Put(overlay, addr)
+			if err != nil {
+				_ = p2ps.Disconnect(overlay)
+				logger.Debugf("addressboook error persisting %s %s: %v", aa, overlay, err)
+				logger.Errorf("error persisting node %s", aa)
```
pradovic, line 262: No need for error word.

pkg/statestore/persistent/store.go
```diff
@@ -2,41 +2,42 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.package storage
 
-package statestore
+package persistent
```
pradovic, line 5: I would call this package leveldb, as there could be multiple persistent implementations.

pkg/statestore/persistent/store_test.go
```diff
@@ -0,0 +1,41 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.package storage
+
+package persistent_test
```
pradovic, line 5: Same as above.

pkg/statestore/persistent/store.go
acud: it will change to badger anyway at some point. i wouldn't like to couple the name with the underlying engine

pradovic: I think it should be underlying engine, unless we plan to have multiple implementations of persistent store inside this package.  Because, you have statestore, an one statestore could be inmem, and the other one leveldb, the third badger, etc.. I don't see a need for another general package like `persistent`, unless there is a need for another separation.

pradovic: For example, if you use database, you would usually have `database` interface, and concrete implementations, unless you need to split it on even more granularity such as sql, no-sql, etc.. But you wuold probably not have a package named nosql and then have a mongo implementation there.

pradovic: Plus, it would probably be nice to be able to have both leveldb and badger code implementations there at once, at least at some point, for test and benchmark, for example. Just my 2 c.

pkg/addressbook/addressbook_test.go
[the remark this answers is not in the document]
pradovic: Why do we test both in mem and persisten in addressbook unit tests?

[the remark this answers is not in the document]
pradovic: But those are valid tests, I would them move them to `/addressbook/test/` and use them as integration tests for different implementations.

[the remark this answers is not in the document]
acud: true. i removed it. thanks for pointing this out

pkg/statestore/persistent/store.go
acud: done

pradovic:
@acud Sorry, I don't understand 🙈  Why did you completely remove `inmem` & `persistant`? Is there no value of having multiple implementations?

acud:
> /addressbook/inmem - inmem implementation
/addressbook/leveldb - leveldb implementation
/addressbook/badger - badger implementation

this is exactly what I wanted to avoid - having multiple implementations doing the same business logic over different storage engines. this is exactly what we should not have because then we end up with 5 different implementations for each component we implement that relies on in-memory and persistent storage.

therefore, i have removed the tests for persistence and for in mem, and left just the in-mem ones, because if the test pass on one storage engine but do not pass on the second one, then it means that the storage engine is broken and that the unit tests testing the storage should catch the fault, not the address book tests. at least, that is my renewed view of this. @janos wdyt

acud:
also @janos, please look at the PR description, there is one TODO that I need input from you on

pradovic:
Ok, I see your point. Fair enough. One more idea from me. Switch`inmem` to mock to be used inside the project, and put a implementation of `addressbook` in the `addressbook` package, and this implementation can use `Store` interface (as it used to be) . In the `addressbook` unit tests you can use `Store` mock then, and rest of the project can use an `addressbook` mocks in unit tests. We probably don't need an actual in memory implementation of `addressbook` anywayz, outside of mocking.

acud:
I don't understand

pradovic:
Ah, sorry if this is taking too long, maybe I am nitpicking 🙈 So my point is:
1. Have `addressbook.go` use `Store` interface
2. move inmem to `addressbook.mock` because we use it as mock
3. Different `Store` implementations can be technology dependent, such as `leveldb`, `badger`, `inmem` etc.. and have their own unit tests.
4. If different implementations are not needed, just keep `Store` implementation leveldb based for now, with no sub-packages based on technology

This way it is nicely resoloved, imo:
1. Bee components could use addressbook mock for unit  tests where addressbook is needed. For example, see hive_tests that use inmem. That should be mock, imo
2. Addressbook is uni tested on `Store` mock, and is not store implementation depended, just interface

Does this makes sense?

acud:
> Have addressbook.go use Store interface (as you already had I believe, at one point)

I'm not sure if you've read the code, but that's the interface which is currently used in the codebase, I'm not sure if I need to change this now or not. This is unclear.

> move inmem to addressbook.mock because we use it as mock

That's wrong. `inmem` is not a mock. Mock is a custom, dummy implementation that would ideally give certain properties for tests to assess whether a certain function has been called or not, or to a-priori inject a certain expected response for the api (let's say that a certain function call should result in this or that response no matter what the input is).
We are raping the term `mock` a bit too much in my opinion. The `inmem` implementation is an in memory implementation hence I'm not gonna rename it to be `mock`. we use it as a simple key-value store and not some fancy test mock. So this argument doesn't hold.


> Different Store implementations can be technology dependent if needed, such as leveldb, badger, inmem etc.. and have their own unit tests.

I strongly disagree and I think that different store implementation _ideally_ should not affect the contract of the package. For such a simple thing as the state store there is no need to introduce a notion of the underlying implementation in contracts.

pradovic:
No need to change, unit tests and package fix was enough, that was my main point. Rest was just general talk. Thanks for your patience. 👍

pkg/topology/full/full_test.go
```diff
@@ -128,7 +140,8 @@ func TestAddPeer(t *testing.T) {
 
 	t.Run("OK - connected peers - peer not already connected", func(t *testing.T) {
 		discovery := mock.NewDiscovery()
-		addressbook := inmem.New()
+		inmem := mockstate.NewStateStore()
```
pradovic, line 143:
```suggestion
		statestore:= mockstate.NewStateStore()
```

all the above where `inmem` name is used.

pkg/node/node.go
```diff
@@ -109,6 +112,19 @@ func NewBee(o Options) (*Bee, error) {
 		logger.Infof("new libp2p key created")
 	}
 
+	var stateStore storage.StateStorer
+	if o.DataDir == "" {
+		stateStore = mockinmem.NewStateStore()
+		logger.Warning("using in-mem state store. no node state will be persisted")
+	} else {
+		stateStore, err = leveldb.NewStateStore(filepath.Join(o.DataDir, "statestore.db"))
```
janos, line 120: Sorry for taking too long to respond to the question about state store data location. I see that you already set it here. Which is fine. I would just remove the `.db` suffix as this is a directory in leveldb case.

PR #83, opened 2020-04-09
P7 - All bee metrics to locatore

jmozah:
This PR replaces localstore ethereum's metrics with bee metrics, to remove a dependency on ethereum codebase
Refer P7 in #66


jmozah:
@janos  The test case fails bit it is not related to my changes

PR #84, opened 2020-04-09
P8 - Add Bee's logging to localstore

jmozah:
This PR  replaces localstore log library with bee logging library, to remove a dependency on swarm codebase.

Refer P8 in #66


jmozah:
@janos  The test case fails bit it is not related to my changes

PR #86, opened 2020-04-10
Flacky tests - NewStream

pradovic:
I believe this one is caused by https://github.com/ethersphere/bee/pull/65. I did not put peer registry synchronization in the right place. It was after the stream was already closed, resulting in possibly late addition to peer registry in retrospect to the other side of the stream.
To reproduce add  `time.Sleep` of 1s before ln 195 in `libp2p.go` on the code from master. What happens iis that stream handler returns not found from peer registry, and this error log is not visible in tests.  This should result in:
```
--- FAIL: TestHeaders (0.35s)
    headers_test.go:57: send headers: read message: stream reset
--- FAIL: TestHeaders_empty (0.31s)
    headers_test.go:106: send headers: read message: Application error 0x0
--- FAIL: TestHeadler (0.15s)
    headers_test.go:170: send headers: read message: Application error 0x0
--- FAIL: TestNewStream (0.44s)
    protocols_test.go:41: send headers: read message: stream reset
--- FAIL: TestNewStream_semanticVersioning (0.37s)
    protocols_test.go:138: create stream "/swarm/testing/2.0.0/messages" to "16Uiu2HAmNvXoptCFxi8TYbNZDSXz6FLois6CZfTAVtNMpRd5cnYw": stream reset
--- FAIL: TestTracing (0.58s)
    tracing_test.go:78: send headers: read message: Application error 0x0
```

I tested this locally, and by adding empty commits to github actions, you can see commit history. Tests seem to pass now every time. 
Do you guys think it would be useful/possible to enable error logs in tests in github actions for the future? Might be useful.

### DIFF
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -341,12 +341,16 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (overlay swarm
 		return swarm.Address{}, fmt.Errorf("handshake: %w", err)
 	}
 
-	if err := helpers.FullClose(stream); err != nil {
-		return swarm.Address{}, err
+	if exists := s.peers.addIfNotExists(stream.Conn(), i.Address); exists {
+		if err := helpers.FullClose(stream); err != nil {
+			return swarm.Address{}, err
+		}
+
+		return i.Address, nil
 	}
 
-	if exists := s.peers.addIfNotExists(stream.Conn(), i.Address); exists {
-		return i.Address, nil
+	if err := helpers.FullClose(stream); err != nil {
+		return swarm.Address{}, err
 	}
 
 	s.metrics.CreatedConnectionCount.Inc()
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -193,12 +193,16 @@ func New(ctx context.Context, o Options) (*Service, error) {
 			return
 		}
 
-		if err := stream.Close(); err != nil {
-			_ = stream.Reset()
+		if exists := s.peers.addIfNotExists(stream.Conn(), i.Address); exists {
+			if err := stream.Close(); err != nil {
+				_ = stream.Reset()
+			}
+
+			return
 		}
 
-		if exists := s.peers.addIfNotExists(stream.Conn(), i.Address); exists {
-			return
+		if err := stream.Close(); err != nil {
+			_ = stream.Reset()
 		}
 
 		remoteMultiaddr, err := ma.NewMultiaddr(fmt.Sprintf("%s/p2p/%s", stream.Conn().RemoteMultiaddr().String(), peerID.Pretty()))
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -178,7 +178,6 @@ func New(ctx context.Context, o Options) (*Service, error) {
 		peerID := stream.Conn().RemotePeer()
 		i, err := s.handshakeService.Handle(NewStream(stream), peerID)
 		if err != nil {
-			_ = stream.Close()
 			if err == handshake.ErrNetworkIDIncompatible {
 				s.logger.Warningf("peer %s has a different network id.", peerID)
 			}
@@ -257,7 +256,6 @@ func (s *Service) AddProtocol(p p2p.ProtocolSpec) (err error) {
 
 			// exchange headers
 			if err := handleHeaders(ss.Headler, stream); err != nil {
-				_ = stream.Close()
 				s.logger.Debugf("handle protocol %s/%s: stream %s: peer %s: handle headers: %v", p.Name, p.Version, ss.Name, overlay, err)
 				return
 			}
@@ -398,7 +396,6 @@ func (s *Service) NewStream(ctx context.Context, overlay swarm.Address, headers 
 
 	// exchange headers
 	if err := sendHeaders(ctx, headers, stream); err != nil {
-		_ = stream.Close()
 		return nil, fmt.Errorf("send headers: %w", err)
 	}
 
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -178,7 +178,7 @@ func New(ctx context.Context, o Options) (*Service, error) {
 		peerID := stream.Conn().RemotePeer()
 		i, err := s.handshakeService.Handle(NewStream(stream), peerID)
 		if err != nil {
-			_ = stream.Reset()
+			_ = stream.Close()
 			if err == handshake.ErrNetworkIDIncompatible {
 				s.logger.Warningf("peer %s has a different network id.", peerID)
 			}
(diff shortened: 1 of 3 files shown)


pkg/p2p/libp2p/libp2p.go
```diff
@@ -390,6 +404,7 @@ func (s *Service) NewStream(ctx context.Context, overlay swarm.Address, headers
 
 	// exchange headers
 	if err := sendHeaders(ctx, headers, stream); err != nil {
+		_ = stream.Reset()
```
acud, line 407: why do you need to reset it?

pkg/p2p/libp2p/libp2p.go
```diff
@@ -328,14 +338,18 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (overlay swarm
...
 	if exists := s.peers.addIfNotExists(stream.Conn(), i.Address); exists {
+		if err := helpers.FullClose(stream); err != nil {
+			return swarm.Address{}, err
+		}
+
 		return i.Address, nil
 	}
 
+	if err := helpers.FullClose(stream); err != nil {
+		return swarm.Address{}, err
```
acud, line 350: `return nil, err`, not a zero address and error

pkg/p2p/libp2p/libp2p.go
```diff
@@ -328,14 +338,18 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (overlay swarm
 		return swarm.Address{}, fmt.Errorf("handshake: %w", err)
 	}
 
-	if err := helpers.FullClose(stream); err != nil {
-		return swarm.Address{}, err
-	}
-
 	if exists := s.peers.addIfNotExists(stream.Conn(), i.Address); exists {
+		if err := helpers.FullClose(stream); err != nil {
+			return swarm.Address{}, err
```
acud, line 343: `return nil, err`

pkg/p2p/libp2p/libp2p.go
```diff
@@ -193,9 +194,17 @@ func New(ctx context.Context, o Options) (*Service, error) {
...
 		if exists := s.peers.addIfNotExists(stream.Conn(), i.Address); exists {
+			if err := stream.Close(); err != nil {
+				_ = stream.Reset()
+			}
+
 			return
 		}
 
+		if err := stream.Close(); err != nil {
+			_ = stream.Reset()
```
acud, line 205: if `Close` failed, won't `Reset` panic?

pkg/p2p/libp2p/libp2p.go
```diff
@@ -193,9 +194,17 @@ func New(ctx context.Context, o Options) (*Service, error) {
 		}
 
 		if exists := s.peers.addIfNotExists(stream.Conn(), i.Address); exists {
+			if err := stream.Close(); err != nil {
+				_ = stream.Reset()
```
acud, line 198: see comment below

pkg/p2p/libp2p/libp2p.go
```diff
@@ -178,6 +178,7 @@ func New(ctx context.Context, o Options) (*Service, error) {
 		peerID := stream.Conn().RemotePeer()
 		i, err := s.handshakeService.Handle(NewStream(stream), peerID)
 		if err != nil {
+			_ = stream.Reset()
```
acud, line 181: is this necessary?

pkg/p2p/libp2p/libp2p.go
pradovic: I don't think it is possible as `Connect` function returns `swarm.Address`?

pkg/p2p/libp2p/libp2p.go
pradovic: Same as above

pkg/p2p/libp2p/libp2p.go
pradovic: So, I have checked the libp2p examples, and it seem that they do `stream.Reset()` when they want to close the other side of the stream as well, if something fails for example. And they do `stream.Close()` if everything was ok and they want to close just this side of the stream.

pkg/p2p/libp2p/libp2p.go
pradovic: See comment about about `Reset`. It will not panic, it looks like they use this as safeguard, if close fails to at least try to close other side of the stream and indicate something is wrong. Multiple `Close` and `Reset` should not panic I think. You can check `helpers.FullClose` implementation of `libp2p`, the original one, not ours. Ours is calling that one, so it should be easy to find. I don't like this as well, but it looks like we should do this on each `Close()` maybe.

pkg/p2p/libp2p/libp2p.go
pradovic: Same as above, if error happened, I don't want to `Close` the stream, but actually `Reset`, which closes both side of the stream.

pradovic:
@acud Thanks. I saw a bit of libp2p examples and decided to try this new stuff. The solution to the tests would work with regular `Close()` call as well, the main thing was the ordering of closing the stream and adding peer to the peer registry.

pkg/p2p/libp2p/libp2p.go
acud: true. i didn't realise we switched to address which stores the byte slice in a struct (in the old codebase it was just a type alias to a byte slice). resolving :+1:

pkg/p2p/libp2p/libp2p.go
acud: Then why don't you just use `helpers.FullClose`?

acud:
Thanks @pradovic.
I think that the way we're handling libp2p streams within protocol implementation is starting to be very verbose, requiring developers to use more and more explicit calls to handle and use streams correctly and this is risky, because its error-prone and will require a lot of scrutiny in code-reviews and parallel protocol implementation in different languages. So, my questions would be:
- Could we make this easier to use, using for example, helper methods or abstraction (e.g. `helpers.FullClose()`
- Could we simply not care (or care _less_) about whether a stream gets reset or not on an error? Is this a must have?

pradovic:
@acud No problem. I agree, let's think about the helper methods and abstractions. I think we left it for later, until we are sure about the approach, as it is a bit hard to test, and it looks like streams have a bit funky behavior. Basically, my conclusions so far (might be not true):
1. `stream.Close()` - closes this side of the stream
2. `stream.Reset()` - closes both side of the stream, which can give the info to the other side that something was wrong
3. `helpers.FullClose()` does this:
``` 
if err := stream.Close(); err != nil {
				_ = stream.Reset()
AwaitEOF()
```

So, let's test this more, and see if we need it.
I  personally like the functionality of stream.Close() and stream.FullClose() because it let's us get the information from other side of the stream about the outcome of the operation. I also agree that we could abstract streams somehow, especially if we are using it in a similar way across protocols, maybe write some helpers functions that also work easy with protobuf or any other type of messages (buffers).

pkg/p2p/libp2p/libp2p.go
pradovic: See the explanation below, basically fullclose waits for other side to close the stream. I use it in Hive protocol and in handshake, as part of protocol to make sure that message is received, but here I don't think it is a right way to do it.

pradovic:
@janos As I said, they are not needed for this PR, so I will remove them, but we are already using them in `FullClose()` though, indirectly. Just the word of caution then, I think that instead of this then we should not close the stream on error, in some cases. So some of `defer stream.Close()` calls should  be addressed. But I agree it's not the scope of this PR.

pradovic:
@janos Done. I have replaced the `stream.Reset()` with `stream.Close()`. I will try to do some implement some tests about this in separate PR. For example, in headers, there was no closing the stream, in the case of an error, so I put the `stream.Close()` there. Is that ok?

pradovic:
Sure, make sense, I was going back and forth here so I ended up doing it along, because I was testing stuff locally. I completely agree with you guys. It should be tested as well, so it def need another PR 👍

PR #89, opened 2020-04-13
10/12: replace ethersphere/swarm/chunk and related types with bee ones

janos:
This PR is related to the 10th point in https://github.com/ethersphere/bee/issues/66.

It brings types from swarm needed for localstore to function without ethersphere/swarm/chunk.

This PR does not remove all dependencies from ethersphere/swarm, but it removes ethersphere/swarm/chunk, while preserving the functionality of localstore in bee.

Dependecies that has to be removed in the following PR:
- github.com/ethersphere/swarm/storage/mock (probably to remove mock from bee locastore)
- github.com/ethersphere/swarm/spancontext (replace with bee tracing)
- github.com/ethersphere/swarm/sctx
- github.com/ethersphere/swarm/shed (planed as PR 11 in #66)

### DIFF
--- a/pkg/storage/localstore/export_test.go
+++ b/pkg/storage/localstore/export_test.go
@@ -75,7 +75,7 @@ func TestExportImport(t *testing.T) {
 		}
 		got := ch.Data()
 		if !bytes.Equal(got, want) {
-			t.Fatalf("chunk %s: got data %s, want %x", addr, got, want)
+			t.Fatalf("chunk %s: got data %x, want %x", addr, got, want)
 		}
 	}
 }
--- a/pkg/storage/localstore/export_test.go
+++ b/pkg/storage/localstore/export_test.go
@@ -21,7 +21,8 @@ import (
 	"context"
 	"testing"
 
-	"github.com/ethersphere/swarm/chunk"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
 )
 
 // TestExportImport constructs two databases, one to put and export
@@ -37,11 +38,11 @@ func TestExportImport(t *testing.T) {
 	for i := 0; i < chunkCount; i++ {
 		ch := generateTestRandomChunk()
 
-		_, err := db1.Put(context.Background(), chunk.ModePutUpload, ch)
+		_, err := db1.Put(context.Background(), storage.ModePutUpload, ch)
 		if err != nil {
 			t.Fatal(err)
 		}
-		chunks[string(ch.Address())] = ch.Data()
+		chunks[ch.Address().String()] = ch.Data()
 	}
 
 	var buf bytes.Buffer
@@ -67,14 +68,14 @@ func TestExportImport(t *testing.T) {
 	}
 
 	for a, want := range chunks {
-		addr := chunk.Address([]byte(a))
-		ch, err := db2.Get(context.Background(), chunk.ModeGetRequest, addr)
+		addr := swarm.MustParseHexAddress(a)
+		ch, err := db2.Get(context.Background(), storage.ModeGetRequest, addr)
 		if err != nil {
 			t.Fatal(err)
 		}
 		got := ch.Data()
 		if !bytes.Equal(got, want) {
-			t.Fatalf("chunk %s: got data %x, want %x", addr.Hex(), got, want)
+			t.Fatalf("chunk %s: got data %s, want %x", addr, got, want)
 		}
 	}
 }
--- a/pkg/storage/store.go
+++ b/pkg/storage/store.go
@@ -7,6 +7,7 @@ package storage
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/ethersphere/bee/pkg/swarm"
 )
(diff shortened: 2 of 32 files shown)


pkg/storage/localstore/export_test.go
```diff
@@ -67,14 +68,14 @@ func TestExportImport(t *testing.T) {
...
-		ch, err := db2.Get(context.Background(), chunk.ModeGetRequest, addr)
+		addr := swarm.MustParseHexAddress(a)
+		ch, err := db2.Get(context.Background(), storage.ModeGetRequest, addr)
 		if err != nil {
 			t.Fatal(err)
 		}
 		got := ch.Data()
 		if !bytes.Equal(got, want) {
-			t.Fatalf("chunk %s: got data %x, want %x", addr.Hex(), got, want)
+			t.Fatalf("chunk %s: got data %s, want %x", addr, got, want)
```
acud, line 78: shouldn't the last `%x` be also `%s`?

pkg/storage/store.go
```diff
@@ -19,11 +20,131 @@ var (
...
+type Store interface {
+	Get(ctx context.Context, mode ModeGet, addr swarm.Address) (ch swarm.Chunk, err error)
+	GetMulti(ctx context.Context, mode ModeGet, addrs ...swarm.Address) (ch []swarm.Chunk, err error)
+	Put(ctx context.Context, mode ModePut, chs ...swarm.Chunk) (exist []bool, err error)
+	Has(ctx context.Context, addr swarm.Address) (yes bool, err error)
+	HasMulti(ctx context.Context, addrs ...swarm.Address) (yes []bool, err error)
+	Set(ctx context.Context, mode ModeSet, addrs ...swarm.Address) (err error)
+	LastPullSubscriptionBinID(bin uint8) (id uint64, err error)
+	SubscribePull(ctx context.Context, bin uint8, since, until uint64) (c <-chan Descriptor, stop func())
+	Close() (err error)
```
acud, line 145: can be `io.Closer`, but not critical for this pr

pkg/storage/testing/chunk.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// GenerateTestRandomChunk generates a Chunk that is not
+// valid, but it contains a random key and a random value.
+// This function is faster then storage.GenerateRandomChunk
+// which generates a valid chunk.
+// Some tests in do not need valid chunks, just
+// random data, and their execution time can be decreased
+// using this function.
+func GenerateTestRandomChunk() swarm.Chunk {
```
acud, line 38: there was some copy of this function somewhere above, maybe convert to use this?

pkg/swarm/swarm.go
```diff
@@ -83,3 +89,86 @@ func (a Address) MarshalJSON() ([]byte, error) {
...
+
+func (c *chunk) TagID() uint32 {
+	return c.tagID
+}
+
+func (self *chunk) String() string {
+	return fmt.Sprintf("Address: %v Chunksize: %v", self.addr.String(), len(self.sdata))
+}
+
+// Proximity returns the proximity order of the MSB distance between x and y
```
acud, line 146: if the proximity test is in a separate file then maybe this function can go into `proximity.go`? not a blocker

pkg/tags/tag.go
```diff
@@ -0,0 +1,276 @@
...
+		Uid:       uid,
+		Anonymous: anon,
+		Name:      s,
+		StartedAt: time.Now(),
+		Total:     total,
+	}
+
+	// context here is used only to store the root span `new.upload.tag` within Tag,
+	// we don't need any type of ctx Deadline or cancellation for this particular ctx
+	t.ctx, t.span = spancontext.StartSpan(context.Background(), "new.upload.tag")
```
acud, line 83: this probably needs to be broken down into a separate task to pull out the relevant wrappers from ethersphere/swarm into the tracing package no?

pkg/tags/tag.go
```diff
@@ -0,0 +1,276 @@
...
+	n, total, err := t.Status(s)
+	return err == nil && n == total
+}
+
+// DoneSplit sets total count to SPLIT count and sets the associated swarm hash for this tag
+// is meant to be called when splitter finishes for input streams of unknown size
+func (t *Tag) DoneSplit(address swarm.Address) int64 {
+	total := atomic.LoadInt64(&t.Split)
+	atomic.StoreInt64(&t.Total, total)
+	t.Address = address
```
acud, line 176: i wonder if we should copy the address here, it could result in quirks. since address is a struct that holds a slice, this infers a copy of the slice pointer. wdyt?

pkg/storage/localstore/export_test.go
janos: Actually, this change is not correct, the second word should stay the same.

pkg/tags/tag.go
janos: Yes, this function is copied as it is from the swarm. We should address this in a separate PR.

pkg/tags/tag.go
janos: I agree.

PR #92, opened 2020-04-13
11/12 integrate bee shed in localstore

janos:
This PR is integrating bee shed with badgerdb into the localstore replacing ethersphere/swarm/shed and removing swarm dependency.

This PR has two tests failing TestDB_SubscribePush and TestDB_SubscribePush_multiple (they are skipped), meaning that subscribe push is not working correctly as hashes in the response are not in the correct order. Since subscribe push will not be used immediately, I am setting this PR as ready for review and to fix subscribe push in the following PR.

This PR also changes shed so that no logger is needed to be passed in field constructors as fields can inherit shed logger. This was a complexity that should have been avoided in previous PRs, but missed in review.

There is also a fix to shed Has method, returning ErrNotFound even the retuned value should be false and no errors if the key is not found.


janos:
Thanks @acud. Let's fix subscribe push in some following PR.

PR #93, opened 2020-04-13
12/12: move localstore out of storage package

acud:
move `localstore` out of `storage` package


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.14
 
 require (
 	github.com/btcsuite/btcd v0.20.1-beta
+	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
```
janos, line 7: This one is strange. Do we need it, as this PR only moves some files?

janos:
This PR is fine, even if we have problems with allocating large memory since we integrated badger to localstore. I would expect that tests are failing on master as frequent as here on ci.

PR #95, opened 2020-04-14
Add in-memory support for testing using DB

jmozah:
This PR adds the in-memory support for badger DB which can be used while testing to reduce memory consumption.


jmozah:
@acud can you review this and merge

acud:
Please see that I've prepared a placeholder [here](https://github.com/ethersphere/bee/blob/master/pkg/node/node.go#L180) for in-mem node bootstrapping.
But so as @janos has mentioned, we need to allow `localstore` to also be configurable whether in-memory should be used. You might get away without adding another `bool` flag when constructing a new `localstore` - an empty path might be considered as implicit in-memory `localstore` (but that's just my suggestion). @janos WDYT?

janos:
> Please see that I've prepared a placeholder [here](https://github.com/ethersphere/bee/blob/master/pkg/node/node.go#L180) for in-mem node bootstrapping.
> But so as @janos has mentioned, we need to allow `localstore` to also be configurable whether in-memory should be used. You might get away without adding another `bool` flag when constructing a new `localstore` - an empty path might be considered as implicit in-memory `localstore` (but that's just my suggestion). @janos WDYT?

Yes, empty path may be just right not to use persistence, I agree. Thanks @acud.

PR #96, opened 2020-04-15
Init topology driver

pradovic:
The different way would be to try to connect to some `limit` amount of nodes, and add them to a topology driver already connected. This makes sense if we don't want to let topology driver choose which and amount of nodes connected when node is starting.

### DIFF
--- a/pkg/addressbook/addressbook.go
+++ b/pkg/addressbook/addressbook.go
@@ -22,6 +22,7 @@ var _ Interface = (*store)(nil)
 type Interface interface {
 	GetPutter
 	Overlays() ([]swarm.Address, error)
+	Multiaddresses() ([]ma.Multiaddr, error)
 }
 
 type GetPutter interface {
--- a/pkg/addressbook/addressbook.go
+++ b/pkg/addressbook/addressbook.go
@@ -17,7 +17,12 @@ import (
 
 const keyPrefix = "addressbook_entry_"
 
-var _ GetPutter = (*store)(nil)
+var _ Interface = (*store)(nil)
+
+type Interface interface {
+	GetPutter
+	Overlays() ([]swarm.Address, error)
+}
 
 type GetPutter interface {
 	Getter
@@ -36,7 +41,7 @@ type store struct {
 	store storage.StateStorer
 }
 
-func New(storer storage.StateStorer) GetPutter {
+func New(storer storage.StateStorer) Interface {
 	return &store{
 		store: storer,
 	}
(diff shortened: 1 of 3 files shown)


pkg/node/node.go
```diff
@@ -297,6 +298,26 @@ func NewBee(o Options) (*Bee, error) {
 	}
 
 	wg.Wait()
+
+	overlays, err := addressbook.Overlays()
+	if err != nil {
+		return nil, fmt.Errorf("addressbook: %w", err)
```
janos, line 304: I would suggest error message `addressbook overlays: %w`

pkg/node/node.go
```diff
@@ -297,6 +298,26 @@ func NewBee(o Options) (*Bee, error) {
...
+	}
+	for _, o := range overlays {
```
janos, line 306: Maybe to add a semaphore to limit the number of concurrent connection attempts?

pkg/addressbook/addressbook.go
```diff
@@ -17,7 +17,12 @@ import (
 
 const keyPrefix = "addressbook_entry_"
 
-var _ GetPutter = (*store)(nil)
+var _ Interface = (*store)(nil)
+
+type Interface interface {
+	GetPutter
+	Overlays() ([]swarm.Address, error)
```
acud, line 24: this is a duplicate interface of what is already defined in `hive_test.go`.
I would be happy if this method can be extracted to a separate interface which can be then reused in the test too.

pradovic: It's not the same, but it is similar. I have switched to using this interface in tests as well. I am not sure if we need any additional granularity, since `Store` seem to implement the whole interface anywayz.

acud: I meant that the method signature is the same as used in the hive test file.
It would be nice, if Overlay and Multiaddresses are broken down into separate interfaces which are referenced by this interface. Because then why bother breaking down `GetPutter` either with the same convention?
Also, it might be benefical to have `addressbook.New` just return `addressbook.Interface` type instead of `GetPutter`, and get rid of the interface cast in the test case

pradovic: Ups, my bad, forgot to delete the cast thanks! The constructor is already returning `Interface`. I am not sure if we need this granularity of `GetPutter` etc.. I would say it makes sense if it reduces are job mocking, and maybe saves some components of unnecessary functionality, but in be I don't see much benefit at the moment, tbh. Since you introduced `Getter`, `Putter` and `GetPutter`, feel free to refactor them any way you like, I like both approaches, but personally I would even just leave the `Interface` for now and delete the more granular ones.

PR #98, opened 2020-04-15
Add chunk validation interface and content addressed validator

nolash:


nolash:
part of https://github.com/ethersphere/bee/issues/40

go.mod
```diff
@@ -6,6 +6,8 @@ require (
 	github.com/btcsuite/btcd v0.20.1-beta
 	github.com/coreos/go-semver v0.3.0
 	github.com/dgraph-io/badger/v2 v2.0.3
+	github.com/ethereum/go-ethereum v1.9.12 // indirect
```
acud, line 9: this definitely should not be here. i guess it is used by `bmt` repo, so a cleanup might be needed

go.mod
```diff
@@ -23,7 +25,7 @@ require (
 	github.com/sirupsen/logrus v1.5.0
 	github.com/spf13/cobra v0.0.7
 	github.com/spf13/viper v1.6.3
-	github.com/syndtr/goleveldb v1.0.0
+	github.com/syndtr/goleveldb v1.0.1-0.20190923125748-758128399b1d
```
janos, line 28: I do not think that this change is relevant for this PR. I would keep dependency upgrades in separate PRs, unless this is something that go tool is insisting to change.

pkg/file/validator/validator.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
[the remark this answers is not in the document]
nolash: Hm. Then since `swarm` package declares chunk, and this is only about chunk, perhaps it should be there?

go.mod
nolash: I'll edit go.mod. No idea why they're there.

nolash: There seems to be some version-neutral includes here:

```
	golang.org/x/crypto v0.0.0-20200406173513-056763e48d71
	golang.org/x/sync v0.0.0-20200317015054-43a5402ce75a
```

should we nail down a version for them?

nolash: hmm these things don't even seem to _have_ versioning.

pkg/file/validator/validator.go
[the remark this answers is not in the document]
janos: I do no think that it would be necessary to be under `swarm`, it has a well defined purpose that it could be on the top level.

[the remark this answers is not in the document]
janos: Only this comment is left from my side.

pkg/validator/validator.go
```diff
@@ -0,0 +1,64 @@
...
+	// execute hash, compare and return result
+	v.hasher.Reset()
+	err := v.hasher.SetSpan(int64(span))
+	if err != nil {
+		v.logger.Debugf("SetSpan on bmt legacy hasher gave error: %v", err)
+		return false
+	}
+	_, err = v.hasher.Write(data[8:])
+	if err != nil {
+		v.logger.Debugf("Write on bmt legacy hasher gave error: %v", err)
```
acud, line 58: no need to capitalize on beginning of sentence

acud:
@nolash code is missing from the PR. Please add the necessary directory

pkg/validator/validator_test.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package validator_test
+
+import (
+	"encoding/binary"
+	"testing"
+
+	"github.com/ethersphere/bee/pkg/file/validator"
```
janos, line 10: This looks like an invalid path and tests are failing because of it.

acud:
@nolash please rebase master, `deepsource` file was added recently and the static analysis tool needs it

pkg/validator/validator.go
nolash: Well it's not really, it's the actual method call

Issue #49, opened 2020-04-01
Download fetches from closest node


Eknir:
# Epic

## Story

As a user of Swarm, I want to know to which node I should ask for delivery of a certain chunk, such that chunk that exists in the network is found by the node whom I ask.

### Global vs local closest
If chunk are stored by the node whose address is closest to the chunk (see #48), then the node should request chunk delivery from the node whose address is closest to the chunk. In a network with full connectivity, the node who is closest (local) is also the node who is closest (global), but when we implement partial connectivity, the decision to ask the closest (local) node should be a step in the direction towards the closest (global).

### Incentivized protocol
The retrieval protocol was incentivized in the old Swarm. The incentives team made a post-mortem based on their work. For the acceptance criteria for this PR, it is not needed to be incentivized, but the ground should be made ready for this protocol to become incentivized. 

## Acceptance criteria
- In a network where n random chunks are stored at the nodes whose addresses are closest to the respective chunks (see #48), a node is able to download all n chunks with 1 attempt per chunk

## Dependencies
- [x] Adjust full topology `ChunkPeer` method to return closest peer
- [x] Introduce notion of `NetStore`. Strong emphasis on how we are going to do caching
- [x] Wire HTTP JSON API to try to retrieve chunk from localstore; when that fails - initiate network request through retrieval protocol, get the chunk, optionally store it then serve it to the requester

Issue #48, opened 2020-04-01
Upload syncs to closest node


Eknir:
# Epic

## Story

As a user of Swarm, I want that my chunk is stored at the node whose address is closest to the address of the chunk, such that anybody who requests the chunk afterwards, knows where to find this chunk.

### Full connectivity
This epic can be fulfilled on a full-connectivity network. The node who uploads a chunk looks in his address book for the node whose address is closest to the address of the chunk. In a full connectivity network, this happens to be the node who is closest to the address of the chunk. In a partial connectivity network (i.e. Kademlia), this may or may not be the node who is closest to the address of the chunk. From the perspective of the node, however, this doesn't matter. He just forwards the push sync request to the connected node who is closest to the address of the chunk.

### Note on pricing
For this epic, it will be sufficient to upload the chunk to the closest connected node. However, in the future, there might be a possibility where the closest connected node is not the node to whom we should forward the push-sync, if there is a node who is further away, but cheaper. 

See (picture from BOS):
![image](https://user-images.githubusercontent.com/33000441/78660668-ca8ba680-78cd-11ea-894d-ee0c8431167f.png)

## Acceptance criteria
- In a network with full connectivity, 100 random chunks are uploaded, and every chunk is stored only at the node whose address is closest to the chunks address

## Dependencies
- [x] Modify topology driver interface to have a `SyncPeer` method
- [x] Modify full topology driver implementation to return closest peer for `SyncPeer`
- [x] Create protocol boilerplate
- [ ] Protocol sinks chunks from two sources - push sync localstore subscription and chunks coming from the network on the same protocol

Protocol test vectors:
- [x] make sure chunk gets picked up from localstore push index and gets sent to the closest peer
- [x] forwarding test - incoming message with chunk, check if there's a closer peer (there is, or maybe there is not) - forward to that peer
- [ ] no forwarding test - check that node does not forward when it is the closest (handle `toplogy.ErrWantSelf`)

PR #104, opened 2020-04-16
[bee #48, 49]: introduce netstore

acud:
Due to popular demand - the return of the immortal `NetStore`

needed for #48 #49, and also wires in #40 just for fun


pradovic:
@acud You can always export the service in `export_test.go`. But how would you use Service if it is not exported? It is needed to add protocol to `p2p`, right?

pkg/netstore/netstore_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/storage/mock"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+var chunkData = []byte("mockdata")
+
+type mockValidator struct{}
+
+func (m mockValidator) Validate(_ swarm.Chunk) bool { return true }
```
janos, line 23: An extremely minor suggestion, `m` receiver is not needed.

acud: but it has to implement the interface (the validator is an interface, not a function type alias)

janos: This is very very minor, but just as `swarm.Chunk` is `_`, `func (_ mockValidator) ... ` can have `_`.

PR #114, opened 2020-04-20
add debug api to expose localstore Has

acud:
this PR adds a debug api endpoint to query whether a node has a chunk. returns `404`, `200`, or `400` if the address was corrupt.

needed for syncing integration testing

### DIFF
--- a/pkg/debugapi/chunk.go
+++ b/pkg/debugapi/chunk.go
@@ -1,0 +1,36 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package debugapi
+
+import (
+	"net/http"
+
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/gorilla/mux"
+)
+
+func (s *server) hasChunkHandler(w http.ResponseWriter, r *http.Request) {
+	addr, err := swarm.ParseHexAddress(mux.Vars(r)["address"])
+	if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/debugapi/chunk.go
```diff
@@ -0,0 +1,36 @@
...
+		jsonhttp.BadRequest(w, err)
+		return
+	}
+
+	if has {
+		jsonhttp.OK(w, nil)
+		return
+	}
+
+	jsonhttp.NotFound(w, nil)
```
janos, line 35: I would make the if condition `!has` in order to have happy path not indented, and all other responses indented.

pkg/debugapi/router.go
```diff
@@ -52,6 +52,9 @@ func (s *server) setupRouting() {
 	router.Handle("/peers/{address}", jsonhttp.MethodHandler{
 		"DELETE": http.HandlerFunc(s.peerDisconnectHandler),
 	})
+	router.Handle("/chunk/{address}", jsonhttp.MethodHandler{
```
janos, line 55: I would suggest to have plurals in the api `"/chunks/{address}"`

pkg/debugapi/chunk_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"testing"
+
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/jsonhttp/jsonhttptest"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/storage/mock"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+func TestHasChunk(t *testing.T) {
```
pradovic, line 19: 🌷 `TestHasChunkHandler`

PR #115, opened 2020-04-21
[bee #48] pushsync: initial pushsync implementation

### LINKED ISSUE #48: Upload syncs to closest node

acud:
Naive implementation.

Defer to later stage:
- [ ] create a simple recorder that allows for an isolated protocol instance to be tested, then cleanup this code to use that recorder

closes #48


pkg/pushsync/pushsync.go
```diff
@@ -3,3 +3,218 @@
...
+
+	// timer, initially set to 0 to fall through select case on timer.C for initialisation
+	timer := time.NewTimer(0)
+	defer timer.Stop()
+	chunksInBatch := -1
+
+	for {
+		select {
+		// handle incoming chunks
+		case ch, more := <-chunks:
```
pradovic, line 134: Sorry if I am asking the stupid question, I am not very familiar with the algorithm here.  When is `chunks` channel being written to? 🙈

pkg/pushsync/pushsync_test.go
```diff
@@ -2,4 +2,336 @@
...
+
+	// instantiate a pushsync instance
+	ps := pushsync.New(pushsync.Options{
+		Streamer:      recorder,
+		Logger:        logger,
+		ClosestPeerer: fullDriver,
+		Storer:        storer,
+	})
+	defer ps.Close()
+	recorder.SetProtocols(ps.Protocol())
```
pradovic, line 93: Why is protocol being sent on the same recorder? Sorry, I might misunderstand something cuz I am not familiar with the protocol 🙈

pradovic:
@acud Nice work! I left a few comments, might be because I don't understand some of fine detailis of the protocol.

pkg/pushsync/pushsync_test.go
```diff
@@ -2,4 +2,336 @@
...
+	fullDriver := full.New(discovery, ab, p2ps, logger, pivotNode)
+
+	storer, err := localstore.New("", pivotNode.Bytes(), nil, logger)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// setup the stream recorder to record stream data
+	recorder := streamtest.New(
+		streamtest.WithMiddlewares(func(f p2p.HandlerFunc) p2p.HandlerFunc {
```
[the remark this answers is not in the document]
acud: please read the code. it is needed so that `handler` will not be called

pkg/pushsync/pushsync_test.go
acud: recorder in its current form is not suitable for this kind of testing that expects a protocol to initiate the creation of its own streams, that is, there is a mutual dependency between recorder and the protocol instance.

the pr description mentions that a simpler recorder implementation is needed, since recorder in its current form actually needs two protocol instances, but this is not needed for this test vector, since there is no initiating "other" node which creates the stream, only one unit is under test here. if you have any other suggestion i am open to hear it out.

pkg/pushsync/pushsync_test.go
```diff
@@ -2,4 +2,336 @@
...
+	var records []*streamtest.Record
+
+LOOP:
+	for i := 0; i < 10; i++ {
+		records, _ = recorder.Records(closestPeer, pushsync.ProtocolName, pushsync.ProtocolVersion, pushsync.StreamName)
+		switch l := len(records); l {
+		case 1:
+			break LOOP
+		case 0:
+			time.Sleep(time.Millisecond * 10)
```
nolash, line 110: ewww

pkg/pushsync/pushsync_test.go
```diff
@@ -2,4 +2,336 @@
...
+	}
+
+	messages, err := protobuf.ReadMessages(
+		bytes.NewReader(records[0].In()),
+		func() protobuf.Message { return new(pb.Delivery) },
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(messages) > 1 {
```
nolash, line 123: can this happen when you already checked `Records()` above?

pkg/pushsync/pushsync_test.go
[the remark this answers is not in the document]
nolash: shouldn't we have just `streamtest.WithNoHandler` that conceals exactly this?

Why wouldn't I call `streamtest.New().WithMiddleWares()`. I thought it was convention now to use chainables?

pkg/pushsync/pushsync_test.go
```diff
@@ -2,4 +2,336 @@
...
+	// this triggers the handler of the pivot with a delivery stream
+	err = w.WriteMsg(&pb.Delivery{
+		Address: chunkAddress.Bytes(),
+		Data:    chunkData,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for i := 0; i < 20; i++ {
```
nolash, line 230: does the divine inspiration for these numbers come from your levitation meditation?

pkg/pushsync/pushsync_test.go
```diff
@@ -2,4 +2,336 @@
...
+		if !called {
+			time.Sleep(10 * time.Millisecond)
+		}
+	}
+
+	records, err := recorder.Records(closestPeer, pushsync.ProtocolName, pushsync.ProtocolVersion, pushsync.StreamName)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if l := len(records); l != 1 {
```
nolash, line 244: why not use the same construction as in the previous test here for enforcing only one record?

pkg/pushsync/pushsync_test.go
```diff
@@ -2,4 +2,336 @@
...
+
+	// setup the stream recorder to record stream data
+	recorder := streamtest.New(
+		streamtest.WithMiddlewares(func(f p2p.HandlerFunc) p2p.HandlerFunc {
+			calls++
+			if runtime.GOOS == "windows" {
+				// windows has a bit lower time resolution
+				// so, slow down the handler with a middleware
+				// not to get 0s for rtt value
+				time.Sleep(100 * time.Millisecond)
```
nolash, line 296: `ewww^2`

pkg/pushsync/pushsync_test.go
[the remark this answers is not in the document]
pradovic: No need to get defensive, I was just curious as it was not obvious to me from the code or the comment ✌️

pkg/pushsync/pushsync_test.go
```diff
@@ -2,4 +2,336 @@
...
+
+	ps := pushsync.New(pushsync.Options{
+		Streamer:      recorder,
+		Logger:        logger,
+		ClosestPeerer: fullDriver,
+		Storer:        storer,
+	})
+	defer ps.Close()
+
+	// TODO: this needs to go away once we have a simpler recorder
```
nolash, line 310: repeating this comment doesn't make it go away more

nolash:
Oh, I forgot to mention there is a conspicuous lack of vogon commentary in the code

pkg/pushsync/pushsync.go
```diff
@@ -3,3 +3,218 @@
...
+
+type Options struct {
+	Streamer      p2p.Streamer
+	Storer        storage.Storer
+	ClosestPeerer topology.ClosestPeerer
+
+	Logger logging.Logger
+}
+
+var retryInterval = 10 * time.Second // time interval between retries
```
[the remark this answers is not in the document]
jmozah: like do you want to have a flag during startup to indicate timeout?

pkg/pushsync/pushsync.go
```diff
@@ -3,3 +3,218 @@
...
+	// create chunk and store it in the local store
+	addr := swarm.NewAddress(ch.Address)
+	chunk := swarm.NewChunk(addr, ch.Data)
+	_, err := ps.storer.Put(ctx, storage.ModePutSync, chunk)
+	if err != nil {
+		return err
+	}
+
+	// push this to the closest node too
+	peer, err := ps.peerSuggester.ClosestPeer(addr)
```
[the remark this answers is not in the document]
jmozah: forward nowhere or to all nodes.. whats your point.?

pkg/pushsync/pushsync.go
```diff
@@ -3,3 +3,218 @@
...
+	defer timer.Stop()
+	chunksInBatch := -1
+
+	for {
+		select {
+		// handle incoming chunks
+		case ch, more := <-chunks:
+			// if no more, set to nil, reset timer to 0 to finalise batch immediately
+			if !more {
+				chunks = nil
```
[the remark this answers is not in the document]
acud: L134

pkg/pushsync/pushsync.go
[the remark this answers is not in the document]
nolash: Hm never mind. The original push-sync implementation sought to send to neighborhood, but I guess it's not the case here, and wouldn't even have enough information to do so.

pkg/pushsync/pushsync.go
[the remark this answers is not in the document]
nolash: Ah, I see, sneaky.

pkg/pushsync/pushsync.go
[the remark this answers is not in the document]
nolash: Sorry, I should have been clearer, I was thinking of the numeric value.

pkg/pushsync/export_test.go
```diff
@@ -0,0 +1,7 @@
+package pushsync
```
janos, line 1: Add copyright header.

pkg/pushsync/pushsync.go
```diff
@@ -3,3 +3,218 @@
...
+// by opening a stream to the closest peer
+func (ps *PushSync) sendChunkMsg(ctx context.Context, peer swarm.Address, ch swarm.Chunk) error {
+	startTimer := time.Now()
+	streamer, err := ps.streamer.NewStream(ctx, peer, nil, protocolName, protocolVersion, streamName)
+	if err != nil {
+		return fmt.Errorf("new stream: %w", err)
+	}
+	defer streamer.Close()
+
+	w, _ := protobuf.NewWriterAndReader(streamer)
```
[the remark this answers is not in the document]
janos: This also come to my attention, but in the next PR we should have receipts, and the reader will be needed. It is a minor thingT he most correct thing would be to use NewWriter just as in handler NewReader should be used. But I do not think that the change is necessary.

pkg/pushsync/pushsync_test.go
janos: I agree, but, since we will have receipts, I think that the two sided recorder can be used fully in these tests and maybe shorten them a bit, compared to explicitly testing handler or client function.

pkg/pushsync/pushsync_test.go
janos: Is total of 100ms to wait good enough for the slow ci?

pkg/pushsync/pushsync_test.go
janos: I think that this is from the pingpong test and do not think that it is needed here.

pkg/pushsync/pushsync.go
```diff
@@ -3,3 +3,218 @@
...
+
+			// set chunk status to synced, insert to db GC index
+			if err := ps.storer.Set(ctx, storage.ModeSetSyncPush, ch.Address()); err != nil {
+				ps.logger.Error("pushsync: error setting chunks to synced", "err", err)
+			}
+
+			// retry interval timer triggers starting from new
+		case <-timer.C:
+			// initially timer is set to go off as well as every time we hit the end of push index
+			func() {
```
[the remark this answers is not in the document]
acud: i think it was originally done because probably some `defer` call to measure time... removing this

pkg/pushsync/pushsync_test.go
acud: i will raise it

pkg/pushsync/pushsync_test.go
acud: well, i guess anything can happen. do you want to leave it to fate? or to comparators?

pkg/pushsync/pushsync_test.go
acud: ...

acud:
@pradovic I have addressed your comments about mocking the topology driver, it has reduced code in the tests, thanks.
@nolash I also addressed your comments about reusing the functionality to wait for a certain amount of records in the recorder and have just extracted that method to be on `Recorder`, to wait there directly, or fail the test when the requirement is not satisfied.

pkg/pushsync/pushsync.go
```diff
@@ -3,3 +3,216 @@
...
+	ps := &PushSync{
+		streamer:      o.Streamer,
+		storer:        o.Storer,
+		peerSuggester: o.ClosestPeerer,
+		logger:        o.Logger,
+		metrics:       newMetrics(),
+		quit:          make(chan struct{}),
+	}
+
+	ctx := context.Background()
```
acud, line 59: @janos WDYT about getting rid of `quit` in favor of keeping a `quitFunc`, that will just call the cancel function of this context?

pkg/pushsync/pushsync.go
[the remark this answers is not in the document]
jmozah: it can be.. but for now this naive implementation i set it as constant

pkg/pushsync/pushsync.go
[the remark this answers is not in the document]
jmozah: This function just sends the message... so you are right.. i will change to Writer

pkg/pushsync/pushsync_test.go
jmozah: No. my third eye said so

pkg/pushsync/pushsync_test.go
pradovic: That's what I meant. Basically, we created streamtest very opinionated, to test both sides and not just one side. If you want to test one side you could basically just mock read, write and create stream calls, and no need to use streamtest at all. Do you think guys it would make sense to create mocks for new stream, read and write calls, and then use them if we want to test just one side, either handler or sender? It will be also easier when we abstract creating streams in general

pradovic: Sorry, maybe it's not the scope of this PR just a general thought :)

pkg/pushsync/pushsync.go
acud: but then, should the context not be cancelled when the quit is called? since this context is being passed to other components. cancelling it would make sure any pending operations are cancelled

janos: In this case, there is no cancel function there is only a background context.

But this is a good point. This context is not doing any cancelation in this code, it is just shared without much purpose. I do not think that it is needed at all, and chunksWorker can create its own contexts where a function with a context should be called.

pkg/pushsync/pushsync.go
```diff
@@ -3,3 +3,215 @@
...
+				ps.logger.Errorf("error sending chunk", "addr", ch.Address().String(), "err", err)
+			}
+
+			// set chunk status to synced, insert to db GC index
+			if err := ps.storer.Set(ctx, storage.ModeSetSyncPush, ch.Address()); err != nil {
+				ps.logger.Error("pushsync: error setting chunks to synced", "err", err)
+			}
+
+			// retry interval timer triggers starting from new
+		case <-timer.C:
```
acud, line 171: also, another thought is that this ugliness could be removed by making `SubscribePush` to just iterate until its empty. I think that this timer does not necessarily improve anything, but just slows things down. We have no proof that this actually makes things better. I think that infinitely iterating on push index will be nicer, rate limiting can be done implicitly by the length of the time it takes for the handler here to handle each chunk that is sent (i.e. if it is a blocking or an asynchronous operation)
cc @janos

janos: I agree that this is a bit strange. Even if there are chunks in from `chunks` channel it is possible to get to this case when the timer fires and resets the subscription. The reason to retry would be if there are chunks put in the push index at earlier store time then the chunks that we receive from the subscription. I do not think that this could happen. I think that this is a good suggestion. 👍

acud: > The reason to retry would be if there are chunks put in the push index at earlier store time then the chunks that we receive from the subscription

Then why shouldn't the subscription just iterate over all items in the index? this has to be done anyway to flush it empty no?

janos: It does the iterate over all items, but iteration is only in forward.

PR #116, opened 2020-04-21
Fixed badger iteration realted bug

jmozah:
This PR fixes the iteration related issue we had in the subscription_push test cases.
This also enables these test cases.
i would for this fix since badger is going to be our baseDB.... 
we should try to keep this to catch many bugs as possible..

### DIFF
--- a/pkg/localstore/subscription_push_test.go
+++ b/pkg/localstore/subscription_push_test.go
@@ -32,8 +32,6 @@ import (
 // push syncing subscription is created and validates if
 // all addresses are received in the right order.
 func TestDB_SubscribePush(t *testing.T) {
-	t.Skip("fails with badger shed")
-
 	db, cleanupFunc := newTestDB(t, nil)
 	defer cleanupFunc()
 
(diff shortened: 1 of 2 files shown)


jmozah:
If the fix is fine, i can resolve the conflicts and make it mergable.

jmozah:
Thanks @janos . I am eager to push badger and use it as mush as possible to avoid last minute surprises.  The more we use it the more we find bugs. 

To bring this to effort to a logical conclusion, i can add a test case in shed which captures this case.
WDYT?

janos:
It would be good to have badger switch in one PR that also includes this fix and the regression test, so that we can revert it easily (as @acud was able to do recently) if there are more problems that would delay some of the features. Would that be ok @acud, or should we wait with badger on master until we test it with real features that involve localstore usage?

janos:
I would suggest to include the revert of the commit made on master by merging #113 PR, as this fix does not make sense without it.

jmozah:
Since it is decided by @janos and @acud to not use badger  until they are satisfied with it, I am closing this PR.

janos:
@jmozah would you like for someone else to take over this PR? We did not made a final decision to close it, even so, my last comment suggests to bring back the badger. Even with all the problems that we had with it, it was still not ever once tested by using the application.

PR #120, opened 2020-04-23
add raw data upload api

acud:
add an api endpoint to upload raw data and get back a content address hash.

### DIFF
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -12,7 +12,6 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
-	"strings"
 
 	bmtlegacy "github.com/ethersphere/bmt/legacy"
 	"github.com/gorilla/mux"
@@ -22,6 +21,10 @@ import (
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
+
+type bzzPostResponse struct {
+	Hash string
+}
 
 func hashFunc() hash.Hash {
 	return sha3.NewLegacyKeccak256()
@@ -65,8 +68,7 @@ func (s *server) bzzUploadHandler(w http.ResponseWriter, r *http.Request) {
 		jsonhttp.InternalServerError(w, "write error")
 		return
 	}
-	w.Header().Set("Content-Type", "text/plain")
-	_, _ = io.Copy(w, strings.NewReader(addr.String()))
+	jsonhttp.OK(w, bzzPostResponse{Hash: addr.String()})
 }
 
 func (s *server) bzzGetHandler(w http.ResponseWriter, r *http.Request) {
@@ -94,6 +96,6 @@ func (s *server) bzzGetHandler(w http.ResponseWriter, r *http.Request) {
 		jsonhttp.InternalServerError(w, "read error")
 		return
 	}
-	w.Header().Set("Content-Type", "binary/octet-stream")
+	w.Header().Set("Content-Type", "application/octet-stream")
 	_, _ = io.Copy(w, bytes.NewReader(chunk.Data()))
 }
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -1,0 +1,99 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package api
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"hash"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	bmtlegacy "github.com/ethersphere/bmt/legacy"
+	"github.com/gorilla/mux"
+	"golang.org/x/crypto/sha3"
+
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+func hashFunc() hash.Hash {
+	return sha3.NewLegacyKeccak256()
+}
+
+func (s *server) bzzUploadHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	data, err := ioutil.ReadAll(r.Body)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/api/bzz.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/gorilla/mux"
+	"golang.org/x/crypto/sha3"
+
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+func hashFunc() hash.Hash {
+	return sha3.NewLegacyKeccak256()
```
janos, line 27: I wonder if we could use sha3.New256 not the legacy padding, since we have a new project. This is not in a scope of this PR, but I would like to see if it would be possible.

pkg/api/bzz.go
```diff
@@ -0,0 +1,99 @@
...
+	addr := swarm.NewAddress(hasher.Sum(nil))
+	_, err = s.Storer.Put(ctx, storage.ModePutUpload, swarm.NewChunk(addr, data[8:]))
+	if err != nil {
+		s.Logger.Debugf("bzz: write error: %v, addr %s", err, addr)
+		s.Logger.Error("bzz: write error")
+		jsonhttp.InternalServerError(w, "write error")
+		return
+	}
+	w.Header().Set("Content-Type", "text/plain")
+	_, _ = io.Copy(w, strings.NewReader(addr.String()))
```
janos, line 69: I think that we should have a json response here, since all other responses are json encoded. We could add check on the accept request header but I am of the opinion to keep to json as default, except where we need to return application/octet-stream.

pkg/api/bzz.go
```diff
@@ -0,0 +1,99 @@
...
+			jsonhttp.NotFound(w, "not found")
+			return
+
+		}
+		s.Logger.Debugf("bzz: read error: %v ,addr %s", err, address)
+		s.Logger.Error("bzz: read error")
+		jsonhttp.InternalServerError(w, "read error")
+		return
+	}
+	w.Header().Set("Content-Type", "binary/octet-stream")
```
janos, line 97: Should we use `application/octet-stream` instead? It looks to me that it is more commonly used and documented in various mime type lists.

pkg/api/bzz.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/gorilla/mux"
+	"golang.org/x/crypto/sha3"
+
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+type bzzPostResponse struct {
+	Hash string
```
janos, line 26: I would suggest to have this field as 
```
Address swarm.Address `json:"address"`
```
 It will automatically be marshaled and we will preserve the type information for clarity on the response. Also to add json tag and to mane it address as this is the name that we used so far for swarm addresses.

acud: I originally named this Address but later wondered if this should be so... due to the fact we still are introducing the notion of content addressed data, should reference to actual data be called `Hashes` or also `Addresses`? although they reuse the same underlying data structure to signify where the data should reside, there are clear distinctions in the semantics of content addresses and peer addresses, and so the question is if we should distinguish the two in the codebase. This could potentially have better clarity about whether some code handles content or peer addresses (although this should also be clear from the code itself... nevertheless...).
So this was just a thought experiment, albeit an unsuccessful one :) I'll change it to `Address`

janos: Yes, I see your point. It does make sense to name it hash and stick with that name in exposed APIs for chunk/file addressing. Another suggestion from me would be something like this:

```
Hash swarm.Address `json:"hash"`
```

WDYT? Maybe this is more clear that the address is related to chunk, not peer.

pkg/api/bzz.go
```diff
@@ -0,0 +1,101 @@
...
+	ctx := r.Context()
+	data, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		s.Logger.Debugf("bzz: read error: %v", err)
+		s.Logger.Error("bzz: read error")
+		jsonhttp.InternalServerError(w, "cannot read request")
+		return
+	}
+
+	p := bmtlegacy.NewTreePool(hashFunc, swarm.SectionSize, bmtlegacy.PoolSize)
```
nolash, line 43: Must be `swarm.Branches` not `swarm.sectionSize`

pkg/api/bzz_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/jsonhttp/jsonhttptest"
+	"github.com/ethersphere/bee/pkg/storage/mock"
+)
+
+func TestBzz(t *testing.T) {
+	var (
+		resource        = "/bzz/"
+		content         = []byte("foo")
+		expHash         = "b9d678ef39fa973b430795a1f04e0f2541b47c996fd300552a1e8bfb5824325f"
```
nolash, line 24: as you correctly spotted this is not correct hash when using correct segment count in bmt.

pkg/api/bzz_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/ethersphere/bee/pkg/api"
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/jsonhttp/jsonhttptest"
+	"github.com/ethersphere/bee/pkg/storage/mock"
+)
+
+func TestBzz(t *testing.T) {
```
nolash, line 20: Please always comment tests, however briefly.

pkg/api/bzz.go
acud: i addressed your change @janos. we can change later to `Address` if this is needed

pkg/api/bzz.go
acud: no such field exists. please elaborate

nolash: https://github.com/ethersphere/bee/blob/master/pkg/swarm/swarm.go#L17

PR #121, opened 2020-04-23
Connection Breaker

pradovic:
Please check linked issue for spec. 
This is a "basic" implementation of connection breaker with exponential back-off. Basically, it is tracking down consequent failed calls in a provided time interval. When the limit is reached, the breaker closes for some time in exponential back-off manner. 
The current implementation treats all errors in the same way. We can add specific error matching, if needed, in the future.
The implementation is very opinionated and I am aware that  it can be done in many different ways, so feel free to LMNWYT :)

TODO: 
1. I have added only dummy usage to `libp2p` as I would like to test it and add some `libp2p` tests in another PR 
2. Use `Breaker` in connectivity driver as well, also in another PR

### DIFF
--- a/pkg/p2p/libp2p/internal/breaker/breaker.go
+++ b/pkg/p2p/libp2p/internal/breaker/breaker.go
@@ -13,6 +13,7 @@ import (
 var _ Interface = (*breaker)(nil)
 
 var (
+	// ErrClosed is the special error type that indicates that breaker is closed and that is not executing functions at the moment.
 	ErrClosed = errors.New("breaker closed")
 	timeNow   = time.Now // used instead of time.Since() so it can be mocked
 )
@@ -66,9 +67,9 @@ func NewBreaker(o Options) Interface {
 	return breaker
 }
 
-// executes f() if the limit number of consecutive failed calls is not reached within fail interval.
-// f() call is not locked so it can still be executed concurently
-// returns `errClosed` if the limit is reached and f() result otherwise
+// Executes runs f() if the limit number of consecutive failed calls is not reached within fail interval.
+// f() call is not locked so it can still be executed concurently.
+// Returns `ErrClosed` if the limit is reached or f() result otherwise.
 func (b *breaker) Execute(f func() error) error {
 	if err := b.beforef(); err != nil {
 		return err
--- a/pkg/p2p/libp2p/internal/breaker/breaker.go
+++ b/pkg/p2p/libp2p/internal/breaker/breaker.go
@@ -10,10 +10,16 @@ import (
 	"time"
 )
 
+var _ Interface = (*breaker)(nil)
+
 var (
 	ErrClosed = errors.New("breaker closed")
 	timeNow   = time.Now // used instead of time.Since() so it can be mocked
 )
+
+type Interface interface {
+	Execute(f func() error) error
+}
 
 type breaker struct {
 	limit                int
@@ -33,7 +39,7 @@ type Options struct {
 	MaxBackoff   time.Duration
 }
 
-func NewBreaker(o Options) *breaker {
+func NewBreaker(o Options) Interface {
 	breaker := &breaker{
 		limit:        o.Limit,
 		backoff:      o.StartBackoff,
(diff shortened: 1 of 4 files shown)


pkg/p2p/libp2p/internal/breaker/breaker.go
```diff
@@ -0,0 +1,131 @@
...
+		breaker.backoff = 2 * time.Minute
+	}
+
+	return breaker
+}
+
+// Execute runs f() if the limit number of consecutive failed calls is not reached within fail interval.
+// f() call is not locked so it can still be executed concurently.
+// Returns `ErrClosed` if the limit is reached or f() result otherwise.
+func (b *breaker) Execute(f func() error) error {
```
janos, line 75: Since this type is not exported, this comment could be set on the interface method or as in the package doc, so that it will be visible in godoc.

pkg/p2p/libp2p/internal/breaker/breaker.go
```diff
@@ -0,0 +1,131 @@
...
+	// ErrClosed is the special error type that indicates that breaker is closed and that is not executing functions at the moment.
+	ErrClosed = errors.New("breaker closed")
+)
+
+type Interface interface {
+	Execute(f func() error) error
+}
+
+type breaker struct {
+	limit                int
```
acud, line 28: would be nice to have comments on the struct fields

pkg/p2p/libp2p/internal/breaker/breaker.go
```diff
@@ -0,0 +1,131 @@
...
+func NewBreaker(o Options) Interface {
+	breaker := &breaker{
+		limit:        o.Limit,
+		backoff:      o.StartBackoff,
+		maxBackoff:   o.MaxBackoff,
+		failInterval: o.FailInterval,
+	}
+
+	if o.Limit == 0 {
+		breaker.limit = 100
```
acud, line 54: would be nice to extract these value into `const`s at the head of the source file

pkg/p2p/libp2p/libp2p.go
```diff
@@ -46,6 +47,7 @@ type Service struct {
 	addrssbook       addressbook.Putter
 	peers            *peerRegistry
 	peerHandler      func(context.Context, swarm.Address) error
+	conectionBreaker breaker.Interface
```
acud, line 50: `-> connectionBreaker`

pradovic:
Hey @acud, no problem, i made a typo in the title, but the same naming was used in spec and in code. We can change the name, no problem ,but I don't actually like the name limiter. This is actually more a breaker, as it has a closed and open state, it just automatically changes state back to open after a back-off interval. Rate limiter, to me, would be something that limits the rate, but is not closing it for a some amount of time. For example, rate-limiter, limits traffic toward node, making it more rare, but still steady, just to not send to many packets, but it is sending them in a continuous fashion. Circuit breaker on the other hand closes the traffic towards some node, or half-closes it, to give it time to repair and not overflow with requests that would have no effect. It has a state usually, open, close, half-open, etc.. So, to me rate-limiter would mean something completely different.
In this concrete example it's a bit backward and global. 
Sooo, maybe` ConnectionBackoff`?

acud:
well you said it is a circuit breaker so maybe `CircuitBreaker`? :) I would be happy with a pattern which is also reusable across components and does not have some domain specific naming convention, so we would know that all (or some) components have a `.breaker` or `circuitBreaker` etc...

pradovic:
Well, it's not really a circuit breaker, nor rate-limiter. I am just saying it has a more similarity with circuit breaker, to me. That's why I would like to avoid maybe both, because they are kind of standard pattern. Here is one explanation of circuit breaker: https://martinfowler.com/bliki/CircuitBreaker.html. LTMWYT?

pradovic:
Hmmm, funny, I found one implementation of breaker in go: https://github.com/sony/gobreaker, It loos pretty similar, but it's not the same :)

acud:
cool merge as it is

pradovic:
Haha ok :) I don't mind changing it if you like, just tell me the name, or open another PR :)

acud:
it's ok as it is. you're free to merge

pradovic:
Ok, thanks for the comments!

Issue #118, opened 2020-04-22
Push sync receipts


jmozah:
Implement the receipts for push sync

- [ ] message format for the receipt in pb and generate the code
- [ ] while sending keep a note of chunks to receive receipts
- [ ] In handler if the chunk is stored successfully, send a receipt
- [ ] in the handler, if the message is a receipt, see if it is expected and remove from the list

- How to handle chunks which don't receive receipts?

PR #122, opened 2020-04-23
[bee #118] Receipts for pushsync

jmozah:
This PR adds the basic receipts support to the naive push sync protocol.

The actual strategy will come in another PR.


jmozah:
@janos @edinalovas @pradovic  I think i have completed the task and it is ready for review.
Please review this give your valuable comments.

pkg/pushsync/pushsync.go
```diff
@@ -77,54 +81,40 @@ func (ps *PushSync) Close() error {
...
-			return nil
-		}
-		return err
-	}
-
-	// create chunk and store it in the local store
-	addr := swarm.NewAddress(ch.Address)
-	chunk := swarm.NewChunk(addr, ch.Data)
-	_, err := ps.storer.Put(ctx, storage.ModePutSync, chunk)
+	chunk, err := ps.ReceiveChunkAndSendReceipt(ctx, stream)
```
janos, line 88: This is not correct forwarding. This handler will send receipt when it receives a delivery, but it should only if the peer that it forwards it to responds with its receipt, not before.

jmozah: 1) I get a delivery
2) i forward to closest peer
3) closest peer forwards to closest peer

final peer put in DB and send a receipt which travels back...
when you get a receipt from your closest peer.. then put in your DB and send receipt to your predecessor?

janos: This implementation does not do that. It sends the receipt before the peer that it is forwarded to sends back the receipt.

I have other comments on the current design of this package, but this is the most crucial point that is not correct.

pkg/pushsync/pushsync.go
```diff
@@ -77,54 +81,78 @@ func (ps *PushSync) Close() error {
...
-		if err == io.EOF {
+		ps.metrics.ReceivedChunkErrorCounter.Inc()
+		return err
+	}
+	ps.metrics.ChunksReceivedCounter.Inc()
+
+	// Select the closest peer to forward the received chunk
+	peer, err := ps.peerSuggester.ClosestPeer(swarm.NewAddress(ch.Address))
+	if err != nil {
+		if errors.Is(err, topology.ErrWantSelf) {
```
janos, line 102: Exiting here with nil means that the storer node is not sending back the receipt. This makes all intermediate nodes and the sender node never receive their own receipts.

pkg/pushsync/pushsync.go
jmozah: Thanks for the comment @janos. 

Where is this behavior you said documented? Swarm book section 2.3.2 says that once storer node (which is closest to the chunks address) receives this, it should store it and send back the receipt.  Infact the book doesn't talk about forwarding the chunk from there on. The forwarding given in the book is  about the route the chunk follows to reach the closest peer (to the chunk address) after kademlia is implemented. Is my understanding right @zelig

janos: I am saying that this code does not do what you described. Storer node does not send the receipt.

pkg/pushsync/pushsync.go
jmozah: if the swarm book logic is followed (as specified above), this comment becomes invalid.

pkg/pushsync/pushsync.go
jmozah: i am quoting again... Please read swarm book section 2.3.2.

pkg/pushsync/pushsync.go
```diff
@@ -195,25 +184,89 @@ func (ps *PushSync) chunksWorker() {
...
+	// create chunk and store it in the local store
+	addr := swarm.NewAddress(ch.Address)
+	chunk := swarm.NewChunk(addr, ch.Data)
+	_, err := ps.storer.Put(ctx, storage.ModePutSync, chunk)
+	if err != nil {
+		return nil, err
+	}
+	ps.metrics.TotalChunksStoredInDB.Inc()
+
+	// Send a receipt immediately once the storage of the chunk is successfull
```
zelig, line 209: this should happen  only if the node is closest to chunk, otherwise you need to wait for the receipt on the stream and  pass it back to the requestor

pkg/pushsync/pushsync.go
zelig: @jmozah strange. the second line of your comment above this function correctly describes what this funtion is doing:

>  // it sends a receipt for the chunk on the same stream.

this part is implemented as `ReceiveChunkAndSendReceipt` as the name suggests is what (only) the closest  peer should do.

> // ALso it forwards the received chunk to its closest peer.

jmozah: @zelig  So ReceiveChunkAndSendReceipt should wait to send receipt ..... only after sending the chunk to its closest peer and getting a receipt? 

Which of this 2 Algo is right?

Algo1:
  - Get a chunk delivery
  - add in DB
  - send a receipt in the same stream 
  - Find closest peer
  - Forward the received chunk  to the closest peer
   - receive receipt


Algo 2:
    - Get a chunk delivery
    - Find closest peer
    - Forward the received chunk  to the closest peer
     - receive receipt from closest peer
     - only then add chunk in localDB
     - send a receipt to original sender in the same stream

pkg/pushsync/pushsync.go
jmozah: it is...  As in the code now  we will et the delivery ONLY to the closest node.. the is no kademlia routing....
WHen we implement routing.. what you say is true.

jmozah: Yes... Now the chunk will reach only of the node is closest. Because we have not yet implemented kademlia. SO if you receive a chunk delivery, it means you are the closest node.. See function chunksWorker which delivers chunk only to closestpeers.

AFter kademlia is done... what you are saying is right.

pkg/pushsync/pushsync.go
jmozah: @zelig @janos @acud  I hae modified the code to follow Algo 2. Please review it and let me know your comments.

pkg/pushsync/metrics.go
```diff
@@ -14,48 +14,138 @@ type metrics struct {
...
+	ReceiptsSentCounter        prometheus.Counter
+	SendReceiptErrorCounter    prometheus.Counter
+	ReceiveReceiptErrorCounter prometheus.Counter
+
+	ErrorSettingChunkToSynced prometheus.Counter
+	RetriesExhaustedCounter   prometheus.Counter
+	InvalidReceiptReceived    prometheus.Counter
+
+	SendChunkTimer    prometheus.Counter
+	ReceiptRTT        prometheus.Counter
```
janos, line 36: This should be prometheus.Histogram, I do not think what counter is useful here. Example: api package ResponseDuration metric.

pkg/pushsync/metrics.go
```diff
@@ -14,48 +14,138 @@ type metrics struct {
 	// to be able to return them by Metrics()
 	// using reflection
 
-	SendChunkCounter      prometheus.Counter
-	SendChunkTimer        prometheus.Counter
-	SendChunkErrorCounter prometheus.Counter
-	MarkAndSweepTimer     prometheus.Counter
+	TotalChunksToBeSentCounter prometheus.Counter
+	TotalChunksSynced          prometheus.Counter
+	TotalChunksStoredInDB       prometheus.Counter
```
janos, line 19: This file is not gofmt-ed. An unexpected space character.

pkg/pushsync/pushsync.go
```diff
@@ -77,54 +80,139 @@ func (ps *PushSync) Close() error {
...
+	// Check if the receipt is valid
+	if !bytes.Equal(chunk.Address().Bytes(), receipt.Address) {
+		ps.metrics.InvalidReceiptReceived.Inc()
 		return err
 	}
 
-	// push this to the closest node too
-	peer, err := ps.peerSuggester.ClosestPeer(addr)
+	// forward the receipt to the received stream
+	err = ps.sendReceipt(w, chunk.Address().Bytes());
```
janos: This file is not gofmt-ed. Semicolon at the end.

pkg/pushsync/pushsync.go
```diff
@@ -54,8 +57,8 @@ func New(o Options) *PushSync {
 		quit:          make(chan struct{}),
 	}
 
-	go ps.chunksWorker()
-
+	ctx := context.Background()
+	go ps.chunksWorker(ctx)
```
janos, line 61: This context is deliberately moved in the previous pr to the chunkWorker function. Why is it here again?

pkg/pushsync/pushsync.go
janos: Thanks, there is one more in metrics.go.

pkg/pushsync/pushsync.go
jmozah: @janos Thanks for the input. My IDE was not configured for auto formatting and that was causing the issue. I fixed it now and even if it manually forget to format it before submission, autoformat should do the trick.

pkg/pushsync/metrics.go
jmozah: @janos  My IDE was not not configured all these days for auto formatting. That was causing the issue. I fixed it and now this formatting issues should not happen.

pkg/pushsync/pushsync.go
jmozah: We need contexts downstream anyway even if we dont use them.  I moved it to chunksworker according to your comment.

pkg/pushsync/metrics.go
```diff
@@ -14,48 +14,134 @@ type metrics struct {
...
+			Subsystem: subsystem,
+			Name:      "invalid_receipt_receipt",
+			Help:      "Invalid receipt received from peer.",
+		}),
+		SendChunkTimer: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "send_chunk_time_histogram",
+			Help:      "Histogram for Time taken to send a chunk.",
+			Buckets:   []float64{0.01, 0.1, 0.25, 0.5, 1, 2.5, 5, 10},
```
janos, line 130: Thanks for adding histograms @jmozah. I assume that you reused the same bucket distribution as in the ResponseDuration metric. I think that we can adjust that here as less than `0.01s` is not likely to happen, but more than `10s` is possible. Would it better to have a distribution from `0.1s` to `60s`, I think that this would be more useful.

pkg/pushsync/pushsync.go
janos: Thanks, yes, this is how it is on master. Context is created in chunkWorker.

pkg/pushsync/pushsync.go
```diff
@@ -77,54 +80,139 @@ func (ps *PushSync) Close() error {
...
+	// Check if the receipt is valid
+	if !bytes.Equal(chunk.Address().Bytes(), receipt.Address) {
+		ps.metrics.InvalidReceiptReceived.Inc()
 		return err
 	}
 
-	// push this to the closest node too
-	peer, err := ps.peerSuggester.ClosestPeer(addr)
+	// forward the receipt to the received stream
+	err = ps.sendReceipt(w, chunk.Address().Bytes())
```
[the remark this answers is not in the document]
acud: also, you are not really forwarding the receipt here, but creating a new one. you should _forward_ the exact receipt you have received from the downstream peer. the fact that they contain the same chunk address do not mean that they are the same receipt

pkg/pushsync/pushsync.go
```diff
@@ -195,25 +279,42 @@ func (ps *PushSync) chunksWorker() {
...
+		return nil
+	}
+	timeSpent := float64(time.Since(receiptRTTTimer))
+	ps.metrics.ReceiptRTT.Add(timeSpent)
+
+	// Check if the receipt is valid
+	if !bytes.Equal(ch.Address().Bytes(), receipt.Address) {
+		ps.metrics.InvalidReceiptReceived.Inc()
 		return err
 	}
```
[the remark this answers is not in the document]
jmozah: if the receipt address and the delivery address is not same in the same stream.. then it is an error right...?

pkg/pushsync/pushsync.go
jmozah: it is fixed to send receipt now since the above comment was accepted.

pkg/pushsync/pushsync.go
jmozah: Since the first comment is accepted, this comment was also taken care of in the fix.

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,139 @@ func (ps *PushSync) Close() error {
...
 func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) error {
-	_, r := protobuf.NewWriterAndReader(stream)
+	w, r := protobuf.NewWriterAndReader(stream)
 	defer stream.Close()
 
-	var ch pb.Delivery
+	// Get the delivery
+	ch, err := ps.getChunkDelivery(r)
+	if err != nil {
+		return fmt.Errorf("could not receive chunk delivery: %w ", err)
```
janos, line 90: The error annotation is better without `could not` in it as errors are chained.

Also, there is a blank space at the end of the message that should be removed. I believe that this is a copy/paste error from the chat, that was corrected in the chat.

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,139 @@ func (ps *PushSync) Close() error {
...
+	// Select the closest peer to forward the chunk
+	peer, err := ps.peerSuggester.ClosestPeer(chunk.Address())
+	if err != nil {
+		// If i am the closest peer then store the chunk and send receipt
+		if errors.Is(err, topology.ErrWantSelf) {
+
+			// Store the chunk in the local store
+			_, err := ps.storer.Put(ctx, storage.ModePutSync, chunk)
+			if err != nil {
+				return fmt.Errorf("could not store chunk in local DB: %w ", err)
```
janos, line 106: The error annotation is better without `could not` in it as errors are chained.

Also, there is a blank space at the end of the message that should be removed. I believe that this is a copy/paste error from the chat, that was corrected in the chat.

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,139 @@ func (ps *PushSync) Close() error {
...
+			}
+			ps.metrics.TotalChunksStoredInDB.Inc()
+
+			// Send a receipt immediately once the storage of the chunk is successfull
+			err = ps.sendReceipt(w, chunk.Address())
+			if err != nil {
+				return fmt.Errorf("could not send receipt: %w ", err)
```
janos, line 113: The error annotation is better without `could not` in it as errors are chained.

Also, there is a blank space at the end of the message that should be removed. I believe that this is a copy/paste error from the chat, that was corrected in the chat.

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,139 @@ func (ps *PushSync) Close() error {
...
-			return nil
-		}
+		return fmt.Errorf("new stream: %w", err)
+	}
+	defer streamer.Close()
+
+	wc, rc := protobuf.NewWriterAndReader(streamer)
+
+	if err := ps.sendChunkDelivery(wc, chunk); err != nil {
+		return fmt.Errorf("could not send chunk to closest peer: %w ", err)
```
janos, line 135: The error annotation is better without `could not` in it as errors are chained.

Also, there is a blank space at the end of the message that should be removed. I believe that this is a copy/paste error from the chat, that was corrected in the chat.

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,139 @@ func (ps *PushSync) Close() error {
...
+	}
+	receiptRTTTimer := time.Now()
+
+	receipt, err := ps.receiveReceipt(rc)
+	if err != nil {
+		return fmt.Errorf("could not receive receipt: %w ", err)
```
janos, line 141: The error annotation is better without `could not` in it as errors are chained.

Also, there is a blank space at the end of the message that should be removed. I believe that this is a copy/paste error from the chat, that was corrected in the chat.

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,139 @@ func (ps *PushSync) Close() error {
...
+	if !chunk.Address().Equal(swarm.NewAddress(receipt.Address)) {
+		ps.metrics.InvalidReceiptReceived.Inc()
 		return err
 	}
-	if err := ps.sendChunkMsg(ctx, peer, chunk); err != nil {
+
+	// forward the receipt to the received stream
+	err = ps.sendReceipt(w, chunk.Address())
+	if err != nil {
+		return fmt.Errorf("could not send receipt back: %w", err)
```
janos, line 154: The error annotation is better without `could not` in it as errors are chained.

pkg/pushsync/pushsync.go
```diff
@@ -195,25 +270,41 @@ func (ps *PushSync) chunksWorker() {
...
 		return err
 	}
 
-	timeSpent := float64(time.Since(startTimer))
-	ps.metrics.SendChunkTimer.Add(timeSpent)
-	return err
+	// set chunk status to synced, insert to db GC index
+	if err := ps.storer.Set(ctx, storage.ModeSetSyncPush, ch.Address()); err != nil {
+		ps.metrics.ErrorSettingChunkToSynced.Inc()
+		return fmt.Errorf("could not set chunk as synced in DB: %w", err)
```
janos, line 305: The error annotation is better without `could not` in it as errors are chained.

pkg/pushsync/pushsync.go
```diff
@@ -195,25 +270,41 @@ func (ps *PushSync) chunksWorker() {
...
+	}
+	receiptRTTTimer := time.Now()
 
-	if err := w.WriteMsg(&pb.Delivery{
-		Address: ch.Address().Bytes(),
-		Data:    ch.Data(),
-	}); err != nil {
+	receipt, err := ps.receiveReceipt(r)
+	if err != nil {
+		return fmt.Errorf("could not receive receipt: %w", err)
```
janos, line 292: The error annotation is better without `could not` in it as errors are chained.

pkg/pushsync/pushsync.go
```diff
@@ -195,25 +270,41 @@ func (ps *PushSync) chunksWorker() {
...
 	streamer, err := ps.streamer.NewStream(ctx, peer, nil, protocolName, protocolVersion, streamName)
 	if err != nil {
 		return fmt.Errorf("new stream: %w", err)
 	}
 	defer streamer.Close()
 
-	w, _ := protobuf.NewWriterAndReader(streamer)
+	w, r := protobuf.NewWriterAndReader(streamer)
+	if err := ps.sendChunkDelivery(w, ch); err != nil {
+		return fmt.Errorf("could not send chunk delivery : %w", err)
```
janos, line 286: The error annotation is better without `could not` in it as errors are chained.

There is a blank space before the colon that should be removed.

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,139 @@ func (ps *PushSync) Close() error {
...
+	if err != nil {
+		return fmt.Errorf("receive receipt: %w", err)
+	}
+	ps.metrics.ReceiptRTT.Observe(time.Since(receiptRTTTimer).Seconds())
+
+	// Check if the receipt is valid
+	if !chunk.Address().Equal(swarm.NewAddress(receipt.Address)) {
+		ps.metrics.InvalidReceiptReceived.Inc()
 		return err
 	}
```
janos, line 149: This code block is returning the error which is nil. It should return an appropriate error stating that the receipt address is not correct.

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,139 @@ func (ps *PushSync) Close() error {
...
 
-	// create chunk and store it in the local store
-	addr := swarm.NewAddress(ch.Address)
-	chunk := swarm.NewChunk(addr, ch.Data)
-	_, err := ps.storer.Put(ctx, storage.ModePutSync, chunk)
-	if err != nil {
+	//Dont forward this chunk to the same peer as the received peer
+	if p.Address.Equal(peer) {
 		return err
 	}
```
janos, line 123: This code block is returning the error which is nil.

This particular case is also an error which should be returned. If this happens, it means that a running node and a peer that is sending the chunk do not agree which one is closes to the node, as both are expecting that the another is closer.

This may need to be discussed, as it can happen and for now I think that it is ok to just return the concrete error, not nil error like currently.

Alternatively, this node can accept to store the chunk even if it wants to forward it back backward.

WDTY? @jmozah @acud

pkg/pushsync/pushsync.go
jmozah: good catch

pkg/pushsync/pushsync.go
jmozah: I think we should store it and send a receipt back...

janos:
Tests in this PR and pushsync can be simplified by testing the behaviour of message exchange and not separately handler and SendChunkAndReceiveReceipt method. It also only tests only happy path and error cases should be tested, also.

It is worrisome that with significant changes to the forwarding strategy tests were passing without changes.

I think that addressing these tests in this PR will prolong it even further. We should have an issue to refactor (simplify) and extend tests to accommodate special cases and more hop counts.

This will certainly be needed when the pusher package is created (pulling out chunksWorker from here) and localstore removed from the pushsync package.

jmozah:
@janos  It is true that only happy paths are covered.  But the tests were changed to accommodate the forwarding strategy change , so no worries.

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,153 @@ func (ps *PushSync) Close() error {
...
+	receipt, err := ps.receiveReceipt(rc)
+	if err != nil {
+		return fmt.Errorf("receive receipt: %w", err)
+	}
+	ps.metrics.ReceiptRTT.Observe(time.Since(receiptRTTTimer).Seconds())
+
+	// Check if the receipt is valid
+	if !chunk.Address().Equal(swarm.NewAddress(receipt.Address)) {
+		ps.metrics.InvalidReceiptReceived.Inc()
+		return fmt.Errorf("invalid receipt: %w", err)
```
janos, line 162: There is not need to log nil error. It should be just `errors.New("invalid receipt")`, or even better to have that an explicit error variable.

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,153 @@ func (ps *PushSync) Close() error {
...
 
-	var ch pb.Delivery
-
-	if err := r.ReadMsg(&ch); err != nil {
-		if err == io.EOF {
-			return nil
-		}
-		return err
+	// Get the delivery
+	ch, err := ps.getChunkDelivery(r)
```
acud, line 88: you could return the chunk directly from `getChunkDelivery`

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,153 @@ func (ps *PushSync) Close() error {
...
+
+			// Store the chunk in the local store
+			_, err := ps.storer.Put(ctx, storage.ModePutSync, chunk)
+			if err != nil {
+				return fmt.Errorf("chunk store: %w", err)
+			}
+			ps.metrics.TotalChunksStoredInDB.Inc()
+
+			// Send a receipt immediately once the storage of the chunk is successfull
+			err = ps.sendReceipt(w, chunk.Address())
```
acud, line 111: you can probably `return ps.sendReceipt(...)` and remove the error check and nil return

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,153 @@ func (ps *PushSync) Close() error {
...
+	ps.metrics.ReceiptRTT.Observe(time.Since(receiptRTTTimer).Seconds())
+
+	// Check if the receipt is valid
+	if !chunk.Address().Equal(swarm.NewAddress(receipt.Address)) {
+		ps.metrics.InvalidReceiptReceived.Inc()
+		return errors.New("invalid receipt")
+	}
+
+	// forward the receipt to the received stream
+	err = ps.sendReceipt(w, chunk.Address())
```
acud, line 166: this is still not _forwarding_ the receipt. you are just creating a new receipt here. if there was a signature in the receipt you are supposedly forwarding - it is now lost. you should forward the exact same receipt you got from the peer in `receiveReceipt`.
also, you can `return ps.sendReceipt()`

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,153 @@ func (ps *PushSync) Close() error {
...
+	if err != nil {
+		return fmt.Errorf("send receipt: %w", err)
+	}
+
+	return nil
+}
+
+func (ps *PushSync) getChunkDelivery(r protobuf.Reader) (ch pb.Delivery, err error) {
```
acud, line 174: it would be better if this chunk already returns a `swarm.Chunk` rather than the protobuf wrapper

pkg/pushsync/pushsync.go
```diff
@@ -77,59 +78,153 @@ func (ps *PushSync) Close() error {
...
 		ps.metrics.SendChunkErrorCounter.Inc()
 		return err
 	}
+	ps.metrics.SendChunkTimer.Observe(time.Since(startTimer).Seconds())
+	ps.metrics.ChunksSentCounter.Inc()
+	return nil
+}
+
+func (ps *PushSync) sendReceipt(w protobuf.Writer, addr swarm.Address) (err error) {
+	receipt := &pb.Receipt{
```
acud, line 198: i would pull the creation of a new receipt out of this method and call it with a `pb.Receipt` instead of a `swarm.Address`, this way you are not forced to generate a new canonical receipt any time you call the function, but rather you could reuse the same method to both send new receipts but also to forward existing receipts that you got from other peers and need to forward up the cascade

pkg/pushsync/pushsync_test.go
```diff
@@ -83,148 +73,191 @@ func TestSendToClosest(t *testing.T) {
...
-	// Create a mock connectivity driver
-	mockTopology := mock.NewTopologyDriver(mock.WithClosestPeer(closestPeer))
-	storer, err := localstore.New("", pivotNode.Bytes(), nil, logger)
-	if err != nil {
-		t.Fatal(err)
-	}
+	// create a pivot node and a mocked closest node
+	pivotPeer := swarm.MustParseHexAddress("0000000000000000000000000000000000000000000000000000000000000000")
+	triggerPeer := swarm.MustParseHexAddress("6000000000000000000000000000000000000000000000000000000000000000")
+	closestPeer := swarm.MustParseHexAddress("6000000000000000000000000000000000000000000000000000000000000000")
```
acud, line 121: triggerPeer and closestPeer have the same address and this is confusing, and could potentially eclipse other problems in the tests

pkg/pushsync/pushsync_test.go
```diff
@@ -83,148 +73,191 @@ func TestSendToClosest(t *testing.T) {
...
+		func() protobuf.Message { return new(pb.Delivery) },
+	)
 	if err != nil {
 		t.Fatal(err)
 	}
+	if len(messages) > 1 {
+		t.Fatal("too many messages")
+	}
+	delivery := messages[0].(*pb.Delivery)
+	rcvdChunk := swarm.NewChunk(swarm.NewAddress(delivery.Address), delivery.Data)
```
acud, line 172: this is actually not needed, you can below directly compare like:
```
if !bytes.Equal(delivery.Address, chunkAddress.Bytes()) {...}
if !bytes.Equal(delivery.Data, chunkData) {...}
```

pkg/pushsync/pushsync_test.go
```diff
@@ -83,148 +73,191 @@ func TestSendToClosest(t *testing.T) {
...
-	defer stream.Close()
-	w := protobuf.NewWriter(stream)
 
-	// this triggers the handler of the pivot with a delivery stream
-	err = w.WriteMsg(&pb.Delivery{
-		Address: chunkAddress.Bytes(),
-		Data:    chunkData,
-	})
+	// In pivot peer,  intercept the incoming delivery chunk from the trigger peer and check for correctness
+	records := pivotRecorder.WaitRecords(t, pivotPeer, pushsync.ProtocolName, pushsync.ProtocolVersion, pushsync.StreamName, 1, 5)
```
acud, line 160: the 10 lines of code from `L160` to `L171` repeat themselves constantly in the test, they could be refactored to make the code more readable

pkg/pushsync/pushsync.go
jmozah: What if we get error reading from the stream or decoding the protobuf?

pkg/pushsync/pushsync.go
jmozah: Same as above

pkg/pushsync/pushsync.go
acud: I meant that instead of returning the protobuf delivery message you can return the chunk. I did not say not to return an error

jmozah: aaha.. ok.. my bad..

pkg/pushsync/pushsync.go
jmozah: This is where i meant to say the above comment.
What if we get error sending to the stream or encoding in the protobuf?

jmozah: misunderstood the comment... got it done..

pkg/pushsync/pushsync_test.go
```diff
@@ -8,223 +8,205 @@ import (
...
-	mockTopology := mock.NewTopologyDriver(mock.WithClosestPeerErr(topology.ErrWantSelf))
+	// closing the ps so that the DB is closed after that to avoid data race
+	psClosestPeer.Close()
+	psPivot.Close()
+	psTriggerPeer.Close()
+
+	// closing all DBs
+	closestStorerPeerDB.Close()
+	storerPivotDB.Close()
+	triggerStorerDB.Close()
```
janos, line 140: I think that these storer Close calls can be deferred to make sure that they are called. They will certainly be called after pushsync close calls.

PR #124, opened 2020-04-27
Add splitter interface and simple splitter implementation stub

nolash:
This PR provides a simple implementation of the splitter interface, and a simple single-threaded bmt file hasher as backend.

While the separation of code into `internal` isn't strictly needed, I felt it more harmonious to mimic the structure of the `joiner` implementation as much as possible.

The initial version of this also tried to mimic the two-threaded channel structure of the joiner, but that complexity isn't really needed (thanks @acud)  and besides it was quirky to get to work for reporting error state properly. 

related to #141


pkg/file/io.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+// Read implements io.Reader.
+func (s *simpleReadCloser) Read(b []byte) (int, error) {
+	return s.buffer.Read(b)
+}
+
+// Close implements io.Closer.
+func (s *simpleReadCloser) Close() error {
+	return nil
```
acud, line 30: i am aware that this is supposed to be a `simple` read closer, but i am wondering if we should allow any reads after a `Close` has been called

pkg/file/joiner/joiner.go
```diff
@@ -69,9 +48,8 @@ func (s *simpleJoiner) Join(ctx context.Context, address swarm.Address) (dataOut
 	spanLength := binary.LittleEndian.Uint64(rootChunk.Data())
 	if spanLength <= swarm.ChunkSize {
 		s.logger.Tracef("simplejoiner root chunk %v is single chunk, skipping join and returning directly", rootChunk)
```
acud, line 50: i wonder if this trace is really needed

pkg/file/span.go
```diff
@@ -0,0 +1,42 @@
...
+	for i := 0; i < 9; i++ {
+		spans[i] = span
+		span *= branchesSixtyfour
+	}
+	return spans
+}
+
+// getLevelsFromLength calculates the last level index which a particular data section count will result in.
+// The returned level will be the level of the root hash.
+func GetLevelsFromLength(length int64, sectionSize, branches int) int {
```
acud, line 31: i think this function can simply be renamed to `Levels`. the documentation should explain what it gets levels for (length), and I don't think there's anything else we'd need to get levels for, except for length. for you consideration.

also, `length` should be of type `uint64` and not `int64`, since data length cannot be negative

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,201 @@
...
+// until the full data length has been written. The Sum should be called which will
+// return the SwarmHash of the data.
+//
+// Called Sum before the last Write, or Write after Sum has been called, may result in
+// error and will may result in undefined result.
+type SimpleSplitterJob struct {
+	ctx        context.Context
+	store      storage.Storer
+	spanLength int64    // target length of data
+	length     int64    // number of bytes written to the data level of the hasher
```
acud, line 41: i think these two should be `uint64`

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,201 @@
...
+	hasher     bmt.Hash // underlying hasher used for hashing the tree
+	buffer     []byte   // keeps data and hashes, indexed by cursors
+	logger     logging.Logger
+}
+
+// NewSimpleSplitterJob creates a new SimpleSplitterJob.
+//
+// The spanLength is the length of the data that will be written.
+func NewSimpleSplitterJob(ctx context.Context, store storage.Storer, spanLength int64) *SimpleSplitterJob {
+
```
acud, line 53: no need for whitespace

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,201 @@
...
+	p := bmtlegacy.NewTreePool(hashFunc, swarm.Branches, bmtlegacy.PoolSize)
+	j := &SimpleSplitterJob{
```
acud, line 55: `return &SimpleSplitterJob{...}`

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,201 @@
...
+		ctx:        ctx,
+		store:      store,
+		spanLength: spanLength,
+		sumCounts:  make([]int, 9),
+		cursors:    make([]int, 9),
+		hasher:     bmtlegacy.New(p),
+		buffer:     make([]byte, swarm.ChunkSize*9),
+		logger:     logging.New(os.Stderr, 6),
```
acud, line 63: please inject the logger through the constructor

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,201 @@
...
+//
+// The spanLength is the length of the data that will be written.
+func NewSimpleSplitterJob(ctx context.Context, store storage.Storer, spanLength int64) *SimpleSplitterJob {
+
+	p := bmtlegacy.NewTreePool(hashFunc, swarm.Branches, bmtlegacy.PoolSize)
+	j := &SimpleSplitterJob{
+		ctx:        ctx,
+		store:      store,
+		spanLength: spanLength,
+		sumCounts:  make([]int, 9),
```
acud, line 59: what does `9` stand for and could we extract it as a documented `const`?

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,201 @@
...
+		buffer:     make([]byte, swarm.ChunkSize*9),
+		logger:     logging.New(os.Stderr, 6),
+	}
+	return j
+}
+
+// Write adds data to the file splitter.
+func (j *SimpleSplitterJob) Write(b []byte) (int, error) {
+	if len(b) > swarm.ChunkSize {
+		return 0, fmt.Errorf("Write must be called with a maximum of %d bytes", swarm.ChunkSize)
```
acud, line 71: i'd consider even being more harsh here at this point and even `panic`, so we have an immediate idea that some grave mistake such as this has occured. @janos WDYT?

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	j.length += int64(len(b))
+	if j.length > j.spanLength {
+		return 0, errors.New("Write past span length")
+	}
+
+	j.writeToLevel(0, b)
+	if j.length == j.spanLength {
+		j.logger.Tracef("last write %d done for context %v, total length %d bytes", len(b), j.ctx, j.length)
```
acud, line 80: not sure if this is needed

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,201 @@
...
+	s.sumCounts[lvl]++
+	spanSize := file.Spans[lvl] * swarm.ChunkSize
+	span := (s.length-1)%spanSize + 1
+
+	sizeToSum := s.cursors[lvl] - s.cursors[lvl+1]
+
+	s.hasher.Reset()
+	err := s.hasher.SetSpan(span)
+	if err != nil {
+		return nil
```
acud, line 120: `return err`

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	_, err = s.hasher.Write(s.buffer[s.cursors[lvl+1] : s.cursors[lvl+1]+sizeToSum])
+	if err != nil {
+		return nil
```
acud, line 124: `return err`

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	ref := s.hasher.Sum(nil)
+	addr := swarm.NewAddress(ref)
+	ch := swarm.NewChunk(addr, s.buffer[s.cursors[lvl+1]:s.cursors[lvl]])
+	_, err = s.store.Put(s.ctx, storage.ModePutUpload, ch)
+	if err != nil {
+		return nil
```
acud, line 131: `return err`

pkg/file/splitter/internal/job.go
acud: or at least `panic`

pkg/file/splitter/internal/job.go
acud: or at least `panic`

pkg/file/splitter/internal/job.go
acud: ...

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,201 @@
...
+// The result will be:
+//
+//       SS
+//     F    F
+//   F   F   F
+// F   F   F   F
+//
+// After which the SS will be hashed to obtain the final root hash
+func (s *SimpleSplitterJob) moveDanglingChunk() {
+
```
acud, line 178: remove ws

pkg/file/splitter/internal/job_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/ethersphere/bee/pkg/file/splitter/internal"
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/storage/mock"
+	"github.com/ethersphere/bee/pkg/swarm"
+	mockbytes "gitlab.com/nolash/go-mockbytes"
+)
+
+var (
+	dataLengths = []int{31, // 0
```
acud, line 22: (extreme ocd nitpick) move 31 one line below? that way all is nice and aligned?

pkg/file/splitter/internal/job_test.go
```diff
@@ -0,0 +1,157 @@
...
+		"59de730bf6c67a941f3b2ffa2f920acfaa1713695ad5deea12b4a121e5f23fa1", // 18
+		"522194562123473dcfd7a457b18ee7dee8b7db70ed3cfa2b73f348a992fdfd3b", // 19
+		"ed0cc44c93b14fef2d91ab3a3674eeb6352a42ac2f0bbe524711824aae1e7bcc", // 20
+	}
+
+	start = 0
+	end   = len(dataLengths)
+)
+
+func TestSplitterJobPartialSingleChunk(t *testing.T) {
```
acud, line 72: add godoc

pkg/file/splitter/internal/job_test.go
```diff
@@ -0,0 +1,157 @@
...
+	store := mock.NewStorer()
+
+	//ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*250)
```
acud, line 75: remove commented code

pkg/file/splitter/internal/job_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	bmtHashOfFoo := "2387e8e7d8a48c2a9339c97c1dc3461a9a7aa07e994c5cb8b38fd7c1b3e6ea48"
+	address := swarm.MustParseHexAddress(bmtHashOfFoo)
+	if !addressResult.Equal(address) {
+		t.Fatalf("expected %v, got %v", address, addressResult)
+	}
+
+	_, err = j.Write([]byte("bar"))
+	if err == nil {
+		t.Fatal(err)
```
acud, line 101: `err` is `nil` here, so you're failing with `nil`. maybe fail with `t.Fatalf("expected error but got none")`? or are you expecting an error or not? can't really tell

pkg/file/splitter/internal/job_test.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+}
+
+func TestSplitterJobVector(t *testing.T) {
```
acud, line 105: doc

pkg/file/splitter/internal/job_test.go
```diff
@@ -0,0 +1,157 @@
...
+	for i := start; i < end; i++ {
+		dataLengthStr := strconv.Itoa(dataLengths[i])
+		runString := strings.Join([]string{dataLengthStr, expected[i]}, "/")
+		t.Run(runString, testSplitterJobVector)
+	}
+}
+
+func testSplitterJobVector(t *testing.T) {
+
```
acud, line 114: remove ws

pkg/file/splitter/internal/job_test.go
```diff
@@ -0,0 +1,157 @@
...
+	logger := logging.New(os.Stderr, 6)
+
+	paramstring := strings.Split(t.Name(), "/")
+	dataLength, _ := strconv.ParseInt(paramstring[1], 10, 0)
+	expect := paramstring[2]
+	logger.Debugf("job hash vector: %d", dataLength)
```
acud, line 120: needed?

pkg/file/splitter/internal/job_test.go
```diff
@@ -0,0 +1,157 @@
...
+		t.Run(runString, testSplitterJobVector)
+	}
+}
+
+func testSplitterJobVector(t *testing.T) {
+
+	logger := logging.New(os.Stderr, 6)
+
+	paramstring := strings.Split(t.Name(), "/")
+	dataLength, _ := strconv.ParseInt(paramstring[1], 10, 0)
```
acud, line 118: check err?

pkg/file/splitter/internal/job_test.go
```diff
@@ -0,0 +1,157 @@
...
+	expect := paramstring[2]
+	logger.Debugf("job hash vector: %d", dataLength)
+
+	store := mock.NewStorer()
+
+	ctx, cancel := context.WithCancel(context.Background())
```
acud, line 124: you could do all the variable initialization in a neat and tidy block in the beginning of the test:
```go
var (
  	logger = logging.New(ioutil.Discard, 0)
	paramstring = strings.Split(t.Name(), "/")
	dataLength, _ = strconv.ParseInt(paramstring[1], 10, 0)
	expect = paramstring[2]
	store = mock.NewStorer()
	g = mockbytes.New(0, mockbytes.MockTypeStandard).WithModulus(255)
....
)
defer cancel()

j := internal.NewSimpleSplitterJob(ctx, store, int64(len(data)))

```

pkg/file/splitter/internal/job_test.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+		if c < l {
+			t.Fatalf("short write %d", c)
+		}
+	}
+
+	hashResult := j.Sum(nil)
+	addressResult := swarm.NewAddress(hashResult)
+
+	addressHex := expect
```
acud, line 151: why not use just `expect` directly?

pkg/file/splitter/splitter.go
```diff
@@ -0,0 +1,66 @@
...
+type simpleSplitter struct {
+	store  storage.Storer
+	logger logging.Logger
+}
+
+// NewSimpleSplitter creates a new SimpleSplitter
+func NewSimpleSplitter(store storage.Storer) file.Splitter {
+	return &simpleSplitter{
+		store:  store,
+		logger: logging.New(os.Stderr, 6),
```
acud, line 31: inject the logger through the constructor

pkg/file/splitter/internal/job.go
nolash: It seems to me that anything `length` in `golang` is int.

pkg/file/span.go
nolash: >  I don't think there's anything else we'd need to get levels for, except for length. for you consideration.

there may be later on, but we can change it back when and if we get there.

> also, length should be of type uint64 and not int64, since data length cannot be negative

`len()` in `golang` returns `int`, not `uint`. `io.Reader.Read()` returns `int, err`. Why would we interpret length differently here?

pkg/file/splitter/internal/job_test.go
nolash: well fail is impossible here

pkg/file/splitter/internal/job_test.go
nolash: feels like a weird structure for the `err` and `cancel`. What if more than one decl that returns `err`. You have to move your eye to see where the `err` belongs. Same with `cancel`, easier to spot when it's on same indent.

If you insist, with a frowny emoticon or similar, I will change it.

pkg/file/span.go
acud: whoops you're right. in old swarm subtrie size is `int64`. apologies i stand corrected

pkg/file/splitter/internal/job_test.go
acud: <3

pkg/file/splitter/internal/job_test.go
acud: smiling emoticon it is then

pkg/file/splitter/internal/job.go
acud: i > linter

acud: maybe the linter is sick today

pkg/file/io.go
```diff
@@ -0,0 +1,41 @@
...
+
+func NewSimpleReadCloser(buffer []byte) io.ReadCloser {
+	return &simpleReadCloser{
+		buffer: bytes.NewBuffer(buffer),
+	}
+}
+
+// Read implements io.Reader.
+func (s *simpleReadCloser) Read(b []byte) (int, error) {
+	if s.closed {
```
acud, line 28: race detector might show this as a possible race, since there's no mutex to protect the field. fyi

pkg/file/io.go
```diff
@@ -0,0 +1,41 @@
...
+		return 0, errors.New("Read on closed reader")
```
acud, line 29: no need to capitalize errors. ditto for `Close`

pkg/file/error.go
```diff
@@ -26,3 +26,26 @@ func (e *AbortError) Unwrap() error {
...
 	return e.err.Error()
 }
+
+// HashError should be returned whenever a file operation is terminated
+// before it has completed.
+type HashError struct {
+	err error
+}
+
+// NewHashError creates a new ErrAborted instance.
```
janos, line 36: ErrAborted -> HashError

pkg/file/io.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	return s.buffer.Read(b)
+}
+
+// Close implements io.Closer.
+func (s *simpleReadCloser) Close() error {
+	if s.closed {
+		return errors.New("Close on already closed reader")
```
acud, line 37: could potentially panic here tbh

pkg/file/span.go
```diff
@@ -0,0 +1,42 @@
+package file
```
acud, line 1: add copyright

pkg/file/span.go
```diff
@@ -0,0 +1,42 @@
...
+
+var Spans []int64
+
+func init() {
+	Spans = GenerateSpanSizes(9, swarm.Branches)
+}
+
+// generates a dictionary of maximum span lengths per level represented by one SectionSize() of data
+//
+// TODO: level 0 should be SectionSize() not Branches()
```
acud, line 17: is this TODO still relevant? also, the godoc doesn't start with the function name

pkg/file/span.go
janos, line 1: Add a copyright header.

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// Write adds data to the file splitter.
+func (j *SimpleSplitterJob) Write(b []byte) (int, error) {
+	if len(b) > swarm.ChunkSize {
+		return 0, fmt.Errorf("Write must be called with a maximum of %d bytes", swarm.ChunkSize)
+	}
+	j.length += int64(len(b))
+	if j.length > j.spanLength {
+		return 0, errors.New("Write past span length")
```
acud, line 76: no need for caps

pkg/file/splitter/internal/job.go
janos: Since this is an exported Method, I would go with just an error.

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return nil, err
+	}
+	_, err = s.hasher.Write(s.buffer[s.cursors[lvl+1] : s.cursors[lvl+1]+sizeToSum])
+	if err != nil {
+		return nil, err
+	}
+	ref := s.hasher.Sum(nil)
+	addr := swarm.NewAddress(ref)
+	ch := swarm.NewChunk(addr, s.buffer[s.cursors[lvl+1]:s.cursors[lvl]])
```
acud, line 145: so you actually don't need to have the span in the chunk data here before the actual data?

pkg/file/splitter/internal/job_test.go
```diff
@@ -0,0 +1,159 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	g := mockbytes.New(0, mockbytes.MockTypeStandard).WithModulus(255)
+	data, err := g.SequentialBytes(int(dataLength))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	logger := logging.New(ioutil.Discard, 0)
```
acud, line 135: this can be moved into the `var` block above

pkg/file/splitter/internal/job_test.go
acud: you could, however, pull the context declaration to be adjacent to its usage (i.e. where the splitter job is created)

pkg/file/splitter/internal/job_test.go
```diff
@@ -0,0 +1,159 @@
...
+		t.Fatal("expected error writing after write/sum complete")
+	}
+}
+
+// TestSplitterJobVector verifies file hasher results of legacy test vectors
+func TestSplitterJobVector(t *testing.T) {
+	for i := start; i < end; i++ {
+		dataLengthStr := strconv.Itoa(dataLengths[i])
+		runString := strings.Join([]string{dataLengthStr, expected[i]}, "/")
+		t.Run(runString, testSplitterJobVector)
```
acud, line 114: i am wondering, there is not integration here to see that the joiner can actually unpack the split content. could we do that? WDYT?

pkg/file/io.go
acud: consider using [Value](https://golang.org/pkg/sync/atomic/#Value)

pkg/file/splitter/internal/job.go
```diff
@@ -0,0 +1,225 @@
...
+type SimpleSplitterJob struct {
+	ctx        context.Context
+	store      storage.Storer
+	spanLength int64    // target length of data
+	length     int64    // number of bytes written to the data level of the hasher
+	sumCounts  []int    // number of sums performed, indexed per level
+	cursors    []int    // section write position, indexed per level
+	hasher     bmt.Hash // underlying hasher used for hashing the tree
+	buffer     []byte   // keeps data and hashes, indexed by cursors
+	logger     logging.Logger
```
janos, line 49: It looks to me that this logger is not used. It can be removed and also removed from NewSimpleSplitterJob argument list. Then removed from `file/splitter.simpleSplitter` as that one is also not used.

pkg/file/span.go
nolash: actually I'm not entirely sure :D this is copied from the other code I wrote, and I since forgot exactly what I meant. Since it works, I'll remove it.

pkg/file/splitter/internal/job.go
nolash: `SetSpan` takes care of that, see above.

pkg/file/splitter/internal/job_test.go
nolash: I'll make one in the `file` package, but next PR?

nolash:
> Since this is an exported Method, I would go with just an error.

@janos I didn't understand this comment sorry

pkg/file/io.go
acud: k

pkg/file/splitter/internal/job.go
acud: yeah that's in regards to the hasher, that i have seen, but what i'm asking is, because the span is not encoded into the chunk data, once that chunk gets reloaded from the chunk store and passes through the validator will it still pass the content addressed validation?

acud: that would be a good test case

pkg/file/splitter/internal/job_test.go
acud: ok

janos:
> > Since this is an exported Method, I would go with just an error.
> 
> @janos I didn't understand this comment sorry

I am fine with returning an error instead of panicking.

nolash:
> I am fine with returning an error instead of panicking.

I changed it to panic. Somehow the context of the comment wasn't preserved in the chat history. SHouold I change it back?

nolash:
> OK to merge as it is given that this code is not wired in. There's still some open question in regards to correctness but I guess we can work them out on later iterations

? @acud

acud:
> > OK to merge as it is given that this code is not wired in. There's still some open question in regards to correctness but I guess we can work them out on later iterations
> 
> ? @acud

it's ok just merge it as it is... you can revert the panics here or on subsequent pr as far as i am concerned. no need to drag you with this one.

PR #146, opened 2020-04-28
Rebuild topology when back-off happens

pradovic:
1. Connect to bootnodes if there are not good nodes from the addressbook (spec in the linked epic + meeting from yesterday).
2. Rebuild topology when back-off happens (spec in the linked task)

The implementation in connectivity driver is not fully optimized, and does not cover all edge cases like some routines lingering for some time. since this is not a production level component, I think this is probably enough.  The functional edge cases should be covered, though, so it should be good enough for test envs. This should be implemented better with more advanced driver.

TODO: I fixed a little bug of handling `DisconnectError` in libp2p here. As it is unrelated I would add unit test for this in the separate PR.


pkg/p2p/libp2p/libp2p.go
```diff
@@ -35,7 +35,12 @@ import (
 	"github.com/multiformats/go-multistream"
 )
 
-var _ p2p.Service = (*Service)(nil)
+var (
+	_ p2p.Service = (*Service)(nil)
+
+	// ErrBadNetwork indicates that it is suspected that network is currently in bad condition
+	ErrBadNetwork = errors.New("bad network")
```
acud, line 42: is this used anywhere?

pkg/pushsync/pb/pushsync.proto
```diff
@@ -10,3 +10,7 @@ message Delivery {
   bytes Address = 1;
   bytes Data = 2;
 }
+
+message Receipt {
+  bytes Address = 1;
```
acud, line 15: please rebase, this is already on master, not sure why this is showing up on this PR

pkg/topology/full/full.go
```diff
@@ -31,17 +31,17 @@ var _ topology.Driver = (*driver)(nil)
...
-
+	base          swarm.Address // the base address for this node
 	discovery     discovery.Driver
-	addressBook   addressbook.GetPutter
+	addressBook   addressbook.Interface
 	p2pService    p2p.Service
 	receivedPeers map[string]struct{} // track already received peers. Note: implement cleanup or expiration if needed to stop infinite grow
-	mtx           sync.Mutex          // guards received peers
+	backoffActive bool
+	mtx           sync.Mutex // guards received peers
```
acud, line 40: comment can probably be removed, since it is used also for backoff

pkg/topology/full/full.go
```diff
@@ -77,6 +77,14 @@ func (d *driver) AddPeer(ctx context.Context, addr swarm.Address) error {
 	if !isConnected(addr, connectedPeers) {
 		peerAddr, err := d.p2pService.Connect(ctx, ma)
 		if err != nil {
+			d.mtx.Lock()
+			delete(d.receivedPeers, addr.ByteString())
+			d.mtx.Unlock()
+			var e *p2p.ConnectionBackoffError
+			if errors.Is(err, e) {
+				d.backoff(e.TryAfter())
```
acud, line 85: not sure if I missed anything, but, how is it that you're first declaring a nil pointer to `*p2p.ConnectionBackoffError`, then checking if `err` is of the same type (ok) but then calling `d.backoff(e.TryAfter())`. isn't this a null pointer exception?

pkg/topology/full/full.go
```diff
@@ -149,6 +157,36 @@ func (d *driver) ClosestPeer(addr swarm.Address) (swarm.Address, error) {
...
+		addresses, _ := d.addressBook.Overlays()
+		for _, addr := range addresses {
+			// todo: context should be in line with the shutdown of the driver itself
+			if err := d.AddPeer(context.Background(), addr); err != nil {
+				var e *p2p.ConnectionBackoffError
+				if errors.Is(err, e) {
+					d.mtx.Lock()
+					d.backoffActive = false
+					d.mtx.Unlock()
+					d.backoff(e.TryAfter())
```
acud, line 180: see my same comment above? how does this work?

pkg/p2p/libp2p/libp2p.go
pradovic: Huh I think not. it was there before so I forgot to erase it. Good catch!

pkg/pushsync/pb/pushsync.proto
pradovic: oh, I was building it for Zahoor and did not notice I commited to my branch :)

pkg/topology/full/full.go
pradovic: It should be type assertion, my bad 🙈

pradovic:
Fixed. Thanks @acud, really good catch!

pkg/p2p/error.go
```diff
@@ -7,14 +7,41 @@ package p2p
...
 )
 
-// ErrPeerNotFound should be returned by p2p service methods when the requested
-// peer is not found.
-var ErrPeerNotFound = errors.New("peer not found")
+var (
+	// ErrPeerNotFound should be returned by p2p service methods when the requested
+	// peer is not found.
+	ErrPeerNotFound = errors.New("peer not found")
+	// ErrAlreadyConnected is returned if connect was called for already connected node
```
acud, line 17: add period at end of sentence

pkg/p2p/libp2p/internal/breaker/breaker.go
```diff
@@ -33,6 +33,9 @@ type Interface interface {
 	// f() call is not locked so it can still be executed concurently.
 	// Returns `ErrClosed` if the limit is reached or f() result otherwise.
 	Execute(f func() error) error
+
+	// ClosedUntil retuns the timestamp when the breaker will become open again
```
acud, line 37: add period at end of sentence

pkg/pushsync/pb/pushsync.pb.go
```diff
@@ -74,22 +74,68 @@ func (m *Delivery) GetData() []byte {
 	return nil
 }
 
+type Receipt struct {
```
acud, line 77: this also doesn't belong here :)

pkg/topology/full/full.go
```diff
@@ -77,6 +77,14 @@ func (d *driver) AddPeer(ctx context.Context, addr swarm.Address) error {
 	if !isConnected(addr, connectedPeers) {
 		peerAddr, err := d.p2pService.Connect(ctx, ma)
 		if err != nil {
+			d.mtx.Lock()
+			delete(d.receivedPeers, addr.ByteString())
+			d.mtx.Unlock()
+			var e *p2p.ConnectionBackoffError
```
acud, line 83: since `e` is not really used to cast the error, wouldn't it be nicer to just remove this variable and check `errors.is(err, p2p.ConnectionBackoffError)`?

pkg/p2p/error.go
```diff
@@ -7,14 +7,41 @@ package p2p
...
+	ErrAlreadyConnected = errors.New("already connected")
+)
+
+// ConnectionBackoffError indicates that connection calls will not be executed until `tryAfter` timetamp.
+// The reason is provided in the wrappped error.
+type ConnectionBackoffError struct {
```
acud, line 23: definitely not a blocker, but as far as I can see we usually define errors  with naming like `ErrSomething...`, so maybe `ErrConnectionBackoff`? feel free to keep as is though

pkg/topology/full/full.go
pradovic: Ah sorry, I did not sleep for 2 days, so I am a bit tired today 🙈 It needs to be asserted with .As, so the first code was correct I just need to switch `errors.Is` to `errors.As`.

pkg/p2p/error.go
pradovic: I saw that Custom errors are usually not named this way, only exported error variables. Example for this and the idiom from above: https://blog.golang.org/go1.13-errors

pkg/topology/full/full.go
pradovic: https://blog.golang.org/go1.13-errors

pkg/p2p/libp2p/libp2p.go
```diff
@@ -278,7 +279,7 @@ func (s *Service) AddProtocol(p p2p.ProtocolSpec) (err error) {
 			s.metrics.HandledStreamCount.Inc()
 			if err := ss.Handler(ctx, p2p.Peer{Address: overlay}, stream); err != nil {
 				var e *p2p.DisconnectError
-				if errors.Is(err, e) {
+				if errors.As(err, &e) {
```
pradovic, line 282: This is a bug fix, not related to the scope of this PR really, but I saw it while debugging.

acud: Cool, but I think that if `e` is already a pointer, it is enough to call `errors.As(err,e)`, you're passing a pointer to a pointer, I'm not sure this is necessary. Another thing is that `e` is never used. Why not just check `Is` with the type directly?

pradovic: Resolved on mattermost 👍

pkg/topology/full/full.go
```diff
@@ -149,6 +159,60 @@ func (d *driver) ClosestPeer(addr swarm.Address) (swarm.Address, error) {
...
+
+	d.backoffActive = true
+
+	done := make(chan struct{})
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		select {
+		case <-done:
+		case <-d.quit:
+			cancel()
```
janos, line 182: Cancel should be called always to release context resources, so I am suggesting to move `cancel()` call at the very end of this anon function after the select is passed by either `done` or `d.quit`. Another approach would be maybe a bit more idiomatic, to have a `defer cancel()` just after the cancel function is declared, as defer is now not expensive and it is allowed to call cancel multiple times.

pradovic: Hmm, I agree yes, as `AddPeer` can be async. Nice catch

pkg/topology/full/full.go
```diff
@@ -149,6 +155,60 @@ func (d *driver) ClosestPeer(addr swarm.Address) (swarm.Address, error) {
 	return closest, nil
 }
 
+func (d *driver) Close() error {
```
janos, line 158: This function should be called by node.Bee.Shutdown and in pkg/topology/full/full_test.go tests.

pradovic: I did not call it on purpose, as I see we don't really have a consistent shuwdown strategy across the project. But I will add it anyway, and let's see how it fits in the future.

acud: actually the shutdown strategy is in `node.go`. we should obviously be mindful about inter-package dependencies when shutting down, but this is supposed to be handled by the shutdown sequence in the shutdown in `node.go`

janos: We have a shutdown strategy, and it should be consistent. Every service started by node.NewBee should be closed in node.Bee.Shutdown. If there is something that is missing we should address it as soon as possible.

Also, every test should cleanup resources that it acquires when it ends. If there are tests that do not do that, we should address them as soon as possible.

pradovic: Done!

pradovic:
@Eknir Why did you close this? :)

pradovic:
@janos No problem, thanks! That was a nice review!

PR #148, opened 2020-05-04
add beekeper (e2e) testing github actions workflow

vandot:
Added github actions that will be executed only after merge to `master`
It uses `k3d` to start k8s cluster inside docker (one node for now), set bees, downloads latest `beekeper` and execute tests.
For now docker image is still built on dockerhub


.github/bee.yaml
```diff
@@ -0,0 +1,36 @@
+image:
```
janos, line 1: At this file holding helm values, I would name it as such or put it in a directory to emphasise that. For example in `.github/helm/bee.yaml`. File `bee.yaml` can also represent other configurations, like the bee configuration file itself.

vandot: Moved it into `beekeper` subfolder

PR #160, opened 2020-05-06
add misspell, gofmt and unconvert linters

janos:
This PR adds more linters to golangci-lint suite to improve automatic static checking.

This PR also fixes some minor the problems found by new linters.

It updates the Go GitHub Actions workflow not to use autocrlf to avoid false positive gofmt errors on windows when gofmt is changing line endings.

It adds `linter` Makefile rule to install a specific version of golangci-lint as its master appears not to be stable.

It changes pre-push githook to use `linter` rule.


janos:
Linter gofmt fails only on windows now. I will have to check why.

PR #165, opened 2020-05-07
Git pre push hook

vandot:
Add git pre-push hook so that before pushing all tests are passed on local dev machine
Install it with `make hooks`, it will create symlink `.githooks/pre-push.bash -> .git/hooks/pre-push`

### DIFF
--- a/.githooks/pre-push.bash
+++ b/.githooks/pre-push.bash
@@ -8,6 +8,11 @@ if [ -z "$commits" ]; then
     exit 0
 fi
 
+if ! command -v golangci-lint &> /dev/null; then
+    echo "installing golangci-lint..."
+    go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
+fi
+
 make lint
 
 make vet
--- a/.githooks/pre-push.bash
+++ b/.githooks/pre-push.bash
@@ -13,10 +13,4 @@ if ! command -v golangci-lint &> /dev/null; then
     go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
 fi
 
-make lint
-
-make vet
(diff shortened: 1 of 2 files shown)


.githooks/pre-push.bash
```diff
@@ -0,0 +1,22 @@
...
+if ! command -v golangci-lint &> /dev/null; then
+    echo "installing golangci-lint..."
+    go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
+fi
+
+make lint
+
+make vet
+
+make test-race
```
janos, line 20: Is it ok to combine this into `make lint vet test-race`?

Makefile
```diff
@@ -39,6 +39,10 @@ build: export CGO_ENABLED=0
 build:
 	$(GO) build -trimpath -ldflags "$(LDFLAGS)" ./...
 
+.PHONY: hooks
```
janos, line 42: I would call this rule `githooks`.

.githooks/pre-push.bash
```diff
@@ -0,0 +1,22 @@
...
+
+exit 0
```
janos, line 22: `exit 0` should be implicit, just curious why it is needed?

.githooks/pre-push.bash
vandot: @janos didn't tried, I will test if make will exit when first commands exits with error

.githooks/pre-push.bash
vandot: probably not, I like it like that :)
will check without it

vandot:
@janos applied all suggested changes

PR #168, opened 2020-05-08
Add split join integration test

nolash:
An integration test making the necessary adjustments for the joiner to join what the splitter splits.


.githooks/pre-push.bash
```diff
@@ -1,15 +0,0 @@
-#!/usr/bin/env bash
```
acud, line 1: @nolash why is this removed? can you please rebase master?

nolash: @acud currently this hook fails on my system, and aborts the push.

pkg/file/file_test.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if l != int64(len(data)) {
+		t.Fatalf("data length return expected %d, got %d", len(data), l)
+	}
+
+	// read from joiner
+	var resultData []byte
+	for i := 0; i < len(data); i += swarm.ChunkSize {
```
acud, line 71: `ioutil.ReadAll()`?

nolash: Sorry, no; joiner needs chunk-wise reads.

pkg/file/joiner/internal/job.go
```diff
@@ -110,12 +107,24 @@ func (j *SimpleJoinerJob) nextReference(level int) error {
...
 	err := j.nextChunk(level-1, chunkAddress)
 	if err != nil {
-		return err
+		// if the last write is a "dangling chunk" the data chunk will have been moved
+		// to an intermediate level. In this edge case, the error must be suppressed,
+		// and the cursor manually to data length boundary to terminate the loop in
+		// the calling frame.
+		if j.readCount+int64(len(data)) == j.spanLength {
+			j.cursors[level] = len(j.data[level])
+			//j.dataC <- data
```
acud, line 116: rm

pkg/file/joiner/internal/job.go
```diff
@@ -110,12 +107,24 @@ func (j *SimpleJoinerJob) nextReference(level int) error {
...
+			err = j.sendChunkToReader(data)
```
acud, line 117: `return j.sendChunk...`?

pkg/file/joiner/internal/job.go
```diff
@@ -110,12 +107,24 @@ func (j *SimpleJoinerJob) nextReference(level int) error {
...
+			return err
+		}
+		if err != io.EOF {
```
acud, line 120: don't you want to fail-fast here and return the error directly?

pkg/file/joiner/internal/job.go
```diff
@@ -110,12 +107,24 @@ func (j *SimpleJoinerJob) nextReference(level int) error {
...
+			j.logger.Debugf("error in chunk join for %v: %v", chunkAddress, err)
+		}
 	}
 
 	// move the cursor to the next reference
 	j.cursors[level] += swarm.SectionSize
-	return nil
+	return err
```
acud, line 127: there should not be an error if the execution gets here. happy path no?

pkg/file/splitter/internal/job.go
```diff
@@ -139,7 +140,9 @@ func (s *SimpleSplitterJob) sumLevel(lvl int) ([]byte, error) {
 	}
 	ref := s.hasher.Sum(nil)
 	addr := swarm.NewAddress(ref)
-	ch := swarm.NewChunk(addr, s.buffer[s.cursors[lvl+1]:s.cursors[lvl]])
+	spanBytes := make([]byte, 8)
+	binary.LittleEndian.PutUint64(spanBytes, uint64(span))
+	ch := swarm.NewChunk(addr, append(spanBytes, s.buffer[s.cursors[lvl+1]:s.cursors[lvl]]...))
```
acud, line 145: i have vague recollection of a poor blind perl developer uttering something about heads, tails and appends... i'll know to ignore them next time

pkg/file/file_test.go
acud: hmmm. it would be nice if `Read` is then able to support this somehow. did you look at `ReadAll()` and why would it not work?

.githooks/pre-push.bash
vandot: @nolash you could only remove hook from `.git/hooks/pre-push`

pkg/file/splitter/internal/job.go
nolash: touché, mon frére.

pkg/file/joiner/internal/job.go
nolash: not quite;

We want the error to reach the end of the go-routine in `NewSimpleJoinerJob`, so that `Read` calls after EOF actually can return an `EOF` error.

pkg/file/joiner/internal/job.go
nolash: I would like to log the error if it's not EOF

pkg/file/joiner/internal/job.go
nolash: ... which I hereby did.

pkg/file/file_test.go
nolash: Also please consider; if this is a 100GB file, for example, would you want to ReadAll it at all?

nolash: Not sure if I understand this.

What would the end-user interface to the reads be, in a non-test scenario?

janos: Actually, with the amount of data in this test, it is not needed to have readers. But, once we have tests that are validating and measuring performance on large files, we may want to use readers approach. It is fine to keep these data slices in memory.

nolash:
spoke with @janos on phone, agreement that fixed length data in memory is ok for this iteration.

pkg/file/joiner/internal/job.go
acud: Not sure why this was marked as resolved. I see now that the logic of this function has changed extensively...

pkg/file/joiner/internal/job.go
```diff
@@ -110,6 +107,19 @@ func (j *SimpleJoinerJob) nextReference(level int) error {
...
+		// if the last write is a "dangling chunk" the data chunk will have been moved
+		// to an intermediate level. In this edge case, the error must be suppressed,
+		// and the cursor manually to data length boundary to terminate the loop in
+		// the calling frame.
+		if j.readCount+int64(len(data)) == j.spanLength {
+			j.cursors[level] = len(j.data[level])
+			err = j.sendChunkToReader(data)
+			return err
+		}
+		j.logger.Debugf("error in chunk join for %v: %v", chunkAddress, err)
```
janos, line 122: Error should not be returned and logged. I am suggesting to handle the error, and log it if necessary by the topmost caller in the execution stack. With both return and log, the same error may be logged multiple times.

nolash: Ok I will _wrap_ the error then. The problem was the store conceals the detail; it just says `not found` and doesn't output the address. Also warrants the question why shouldn't it include the address in the error from the store...?

janos: Yes, that may be useful. We will have to go through storage error messages and logging to clean and improve them.

nolash: so @janos should we instead just add  the address to the error from the store?

janos: I think that it is ok to make this small improvement in this PR and have another one which would address the whole storage.

acud:
@nolash on a sidenote can you please also have a look at the integration of splitter/joiner to the http api?

Issue #47, opened 2020-04-01
Kademlia connectivity


Eknir:
# Epic

Divided into 3 phases:
- [ ] phase 0 - basic kademlia implementation - depth calculation and connection management
- [ ] phase 1 - wire in kademlia to be used by the node by default (including all protocols)
- [ ] phase 2 - choose protocol and implement scoring function

## Story

As a user of Swarm, I don't want to connect to all online Swarm nodes, but still maintain the possibility to reach all online nodes, such that I can manage my peer connections more efficiently.

## Background

[Kademlia Connectivity](https://pdos.csail.mit.edu/~petar/papers/maymounkov-kademlia-lncs.pdf)

Tasks:
- [x] introduce notion of depth
- [x] create separate test vectors for `pslice` type
- [x] plug disconnection notification from peer registry to topology + tests
- [x] JSON API kademlia representation
- [x] Human consumable table format representation _(I've settled for `json.MarshalIndent` for now...)_
- [x] bootstrapping and connectivity maintenance:
- - [x] add peers to kademlia but not necessarily connect to them directly
- - [x] ~~sync.Cond for new (known but unconnected) peers? but also for when a peer disconnects~~
- - [x] worker goroutine to choose what to do with new peers added
- - [x] ~~mark and sweep when depth changes~~
- - [x] ~~short-circuit initial connections to build topology faster(not sure about this one)~~

PR #169, opened 2020-05-08
kademlia, topology: introduce pslice

acud:
this PR introduces a basic data structure (therein called `pslice`) which is essentially a slice which maintains indexes for POs, that is to index the different proximity order of peers that we populate in the main `peers` slice.

needed by #155 
inadvertently related to #47

### DIFF
--- a/pkg/kademlia/pslice/pslice.go
+++ b/pkg/kademlia/pslice/pslice.go
@@ -97,7 +97,6 @@ func (s *PSlice) EachBinRev(pf topology.EachPeerFunc) error {
 
 // ShallowestEmpty returns the shallowest empty bin if one exists.
 // If such bin does not exists, returns true as bool value.
-// Must be called under lock.
 func (s *PSlice) ShallowestEmpty() (bin uint8, none bool) {
 	s.Lock()
 	defer s.Unlock()
@@ -107,7 +106,7 @@ func (s *PSlice) ShallowestEmpty() (bin uint8, none bool) {
 	binCp[len(binCp)-1] = uint(len(s.peers))
 
 	for i := uint8(0); i < uint8(len(binCp)-1); i++ {
-		if binCp[i+1]-binCp[i] == 0 {
+		if binCp[i+1] == binCp[i] {
 			return i, false
 		}
 	}
@@ -128,7 +127,12 @@ func (s *PSlice) exists(addr swarm.Address) (bool, int) {
 	if len(s.peers) == 0 {
 		return false, 0
 	}
-	return peerExists(s.peers, addr)
+	for i, a := range s.peers {
+		if a.Equal(addr) {
+			return true, i
+		}
+	}
+	return false, 0
 }
 
 // Add a peer at a certain PO.
@@ -140,7 +144,9 @@ func (s *PSlice) Add(addr swarm.Address, po uint8) {
 		return
 	}
 
-	s.peers = append(s.peers[:s.bins[po]], append([]swarm.Address{addr}, s.peers[s.bins[po]:]...)...)
+	head := s.peers[:s.bins[po]]
+	tail := append([]swarm.Address{addr}, s.peers[s.bins[po]:]...)
+	s.peers = append(head, tail...)
 	s.incDeeper(po)
 }
 
@@ -185,14 +191,4 @@ func (s *PSlice) decDeeper(po uint8) {
 	for i := po + 1; i < uint8(len(s.bins)); i++ {
 		s.bins[i]--
 	}
-}
-
-// check if a certain address exists within the given slice of peers.
-func peerExists(peers []swarm.Address, peer swarm.Address) (exists bool, idx int) {
-	for i, addr := range peers {
-		if addr.Equal(peer) {
-			return true, i
-		}
-	}
-	return false, 0
 }
--- a/pkg/kademlia/pslice/pslice.go
+++ b/pkg/kademlia/pslice/pslice.go
@@ -1,0 +1,199 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package pslice
+
+import (
+	"sync"
+
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/ethersphere/bee/pkg/topology"
+)
+
+// PSlice maintaines a list of addresses, indexing them by their different proximity orders.
+// Currently, when peers are added or removed, their proximity order must be supplied, this is
+// in order to reduce duplicate PO calculation which is normally known and already needed in the
+// calling context.
+type PSlice struct {
+	peers []swarm.Address
+	bins  []uint
+
+	sync.Mutex
+}
+
+// New creates a new PSlice.
+func New(maxBins int) *PSlice {
+	return &PSlice{
+		peers: make([]swarm.Address, 0),
+		bins:  make([]uint, maxBins),
+	}
+}
+
+// iterates over all peers from deepest bin to shallowest.
+func (s *PSlice) EachBin(pf topology.EachPeerFunc) error {
+	s.Lock()
+	defer s.Unlock()
+
+	if len(s.peers) == 0 {
+		return nil
+	}
+
+	var binEnd = uint(len(s.peers))
+INNER:
+	for i := len(s.bins) - 1; i >= 0; i-- {
+		peers := s.peers[s.bins[i]:binEnd]
+		for _, v := range peers {
+			stop, next, err := pf(v, uint8(i))
+			if err != nil {
+				return err
+			}
+			if stop {
+				return nil
+			}
+			if next {
+				break INNER
+			}
+		}
+		binEnd = s.bins[i]
+	}
+
+	return nil
+}
+
+// EachBinRev iterates over all peers from shallowest to deepest.
+func (s *PSlice) EachBinRev(pf topology.EachPeerFunc) error {
+	s.Lock()
+	defer s.Unlock()
+
+	if len(s.peers) == 0 {
+		return nil
+	}
+
+	var binEnd int
+	for i := 0; i <= len(s.bins)-1; i++ {
+		if i == len(s.bins)-1 {
+			binEnd = len(s.peers)
+		} else {
+			binEnd = int(s.bins[i+1])
+		}
+
+		peers := s.peers[s.bins[i]:binEnd]
+	INNER:
+		for _, v := range peers {
+			stop, next, err := pf(v, uint8(i))
+			if err != nil {
+				return err
+			}
+			if stop {
+				return nil
+			}
+			if next {
+				break INNER
+			}
+		}
+	}
+	return nil
+}
+
+// ShallowestEmpty returns the shallowest empty bin if one exists.
+// If such bin does not exists, returns true as bool value.
+// Must be called under lock.
+func (s *PSlice) ShallowestEmpty() (bin uint8, none bool) {
+	s.Lock()
+	defer s.Unlock()
+
+	binCp := make([]uint, len(s.bins)+1)
+	copy(binCp, s.bins)
+	binCp[len(binCp)-1] = uint(len(s.peers))
+
+	for i := uint8(0); i < uint8(len(binCp)-1); i++ {
+		if binCp[i+1]-binCp[i] == 0 {
+			return i, false
+		}
+	}
+	return 0, true
+}
+
+// Exists checks if a peer exists.
+func (s *PSlice) Exists(addr swarm.Address) bool {
+	s.Lock()
+	defer s.Unlock()
+
+	b, _ := s.exists(addr)
+	return b
+}
+
+// checks if a peer exists. must be called under lock.
+func (s *PSlice) exists(addr swarm.Address) (bool, int) {
+	if len(s.peers) == 0 {
+		return false, 0
+	}
+	return peerExists(s.peers, addr)
+}
+
+// Add a peer at a certain PO.
+func (s *PSlice) Add(addr swarm.Address, po uint8) {
+	s.Lock()
+	defer s.Unlock()
+
+	if e, _ := s.exists(addr); e {
+		return
+	}
+
+	s.peers = append(s.peers[:s.bins[po]], append([]swarm.Address{addr}, s.peers[s.bins[po]:]...)...)
+	s.incDeeper(po)
+}
+
+// Remove a peer at a certain PO.
+func (s *PSlice) Remove(addr swarm.Address, po uint8) {
+	s.Lock()
+	defer s.Unlock()
+
+	e, i := s.exists(addr)
+	if !e {
+		return
+	}
+
+	s.peers = append(s.peers[:i], s.peers[i+1:]...)
+	s.decDeeper(po)
+}
+
+// incDeeper increments the peers slice bin index for proximity order > po.
+// Must be called under lock.
+func (s *PSlice) incDeeper(po uint8) {
+	if po > uint8(len(s.bins)) {
+		panic("po too high")
+	}
+
+	for i := po + 1; i < uint8(len(s.bins)); i++ {
+		// don't increment if the value in k.bins == len(k.peers)
+		// otherwise the calling context gets an out of bound error
+		// when accessing the slice
+		if s.bins[i] < uint(len(s.peers)) {
+			s.bins[i]++
+		}
+	}
+}
+
+// decDeeper decrements the peers slice bin indexes for proximity order > po.
+// Must be called under lock.
+func (s *PSlice) decDeeper(po uint8) {
+	if po > uint8(len(s.bins)) {
+		panic("po too high")
+	}
+
+	for i := po + 1; i < uint8(len(s.bins)); i++ {
+		s.bins[i]--
+	}
+}
+
+// check if a certain address exists within the given slice of peers.
+func peerExists(peers []swarm.Address, peer swarm.Address) (exists bool, idx int) {
+	for i, addr := range peers {
+		if addr.Equal(peer) {
+			return true, i
+		}
+	}
+	return false, 0
+}
--- a/pkg/kademlia/pslice/pslice.go
+++ b/pkg/kademlia/pslice/pslice.go
@@ -11,7 +11,7 @@ import (
 	"github.com/ethersphere/bee/pkg/topology"
 )
 
-// PSlice maintaines a list of addresses, indexing them by their different proximity orders.
+// PSlice maintains a list of addresses, indexing them by their different proximity orders.
 // Currently, when peers are added or removed, their proximity order must be supplied, this is
 // in order to reduce duplicate PO calculation which is normally known and already needed in the
 // calling context.
@@ -40,7 +40,7 @@ func (s *PSlice) EachBin(pf topology.EachPeerFunc) error {
 	}
 
 	var binEnd = uint(len(s.peers))
-INNER:
+
 	for i := len(s.bins) - 1; i >= 0; i-- {
 		peers := s.peers[s.bins[i]:binEnd]
 		for _, v := range peers {
@@ -52,7 +52,7 @@ INNER:
 				return nil
 			}
 			if next {
-				break INNER
+				break
 			}
 		}
 		binEnd = s.bins[i]
@@ -79,7 +79,6 @@ func (s *PSlice) EachBinRev(pf topology.EachPeerFunc) error {
 		}
 
 		peers := s.peers[s.bins[i]:binEnd]
-	INNER:
 		for _, v := range peers {
 			stop, next, err := pf(v, uint8(i))
 			if err != nil {
@@ -89,7 +88,7 @@ func (s *PSlice) EachBinRev(pf topology.EachPeerFunc) error {
 				return nil
 			}
 			if next {
-				break INNER
+				break
 			}
 		}
 	}
--- a/pkg/kademlia/pslice/pslice_test.go
+++ b/pkg/kademlia/pslice/pslice_test.go
@@ -18,16 +18,20 @@ func TestShallowestEmpty(t *testing.T) {
 	var (
 		ps    = pslice.New(16)
 		base  = test.RandomAddress()
-		peers = make([]swarm.Address, 16)
+		peers = make([][]swarm.Address, 16)
 	)
 
 	for i := 0; i < 16; i++ {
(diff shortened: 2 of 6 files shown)


pkg/topology/test/helper.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+	"math/rand"
+
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// RandomAddressAt generates a random address
+// at proximity order prox relative to address
+// if prox is negative a random address is generated.
+func RandomAddressAt(self swarm.Address, prox int) swarm.Address {
```
janos, line 17: It would be nice to have a test for this function. Just to check if expected proximity is correct.

pkg/topology/topology.go
```diff
@@ -28,3 +28,9 @@ type PeerAdder interface {
 type ClosestPeerer interface {
 	ClosestPeer(addr swarm.Address) (peerAddr swarm.Address, err error)
 }
+
+// ScoreFunc is implemented by components that need to score peers in a different way than XOR distance.
+type ScoreFunc func(peer swarm.Address) (score float32)
```
janos, line 33: ScoreFunc is not used in this PR. Maybe to add it later when needed?

pkg/topology/topology.go
```diff
@@ -28,3 +28,9 @@ type PeerAdder interface {
...
+
+// EachPeerFunc is a callback that is called with a peer and its PO
+type EachPeerFunc func(swarm.Address, uint8) (stop, jumpToNext bool, err error)
```
janos, line 36: EachPeerFunc is more related to kademlia as it has the concept of proximity order, while topology package is currently unaware of it. I think that it is ok to have it in kademlia or pslice package. WDYT?

pkg/kademlia/pslice/pslice.go
```diff
@@ -0,0 +1,199 @@
...
+		for _, v := range peers {
+			stop, next, err := pf(v, uint8(i))
+			if err != nil {
+				return err
+			}
+			if stop {
+				return nil
+			}
+			if next {
+				break INNER
```
janos, line 55: I am not sure if jumpToNext is tested in tests. Breaking here looks to me the same as like `stop` behaves, exiting the bins loop. It looks to me that it should be `continue INNER`.

pkg/kademlia/pslice/pslice.go
```diff
@@ -0,0 +1,199 @@
...
+	var binEnd int
+	for i := 0; i <= len(s.bins)-1; i++ {
+		if i == len(s.bins)-1 {
+			binEnd = len(s.peers)
+		} else {
+			binEnd = int(s.bins[i+1])
+		}
+
+		peers := s.peers[s.bins[i]:binEnd]
+	INNER:
```
janos, line 82: No need for this label. Break inside the loop will break it without the label.

pkg/kademlia/pslice/pslice_test.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	err := ps.EachBin(f)
+	if !errors.Is(err, e) {
+		t.Fatal("didnt get expected error")
+	}
+}
+
+// TestIterators tests that the EachBin and EachBinRev iterators work as expected.
+func TestIterators(t *testing.T) {
```
janos, line 196: It looks to me that stop and jumpToNext functionalities are not tested here.

pkg/topology/topology.go
acud: yeah, i put it here because if we still want to keep the full connectivity thing in the codebase it would have to have the same interface as kademlia with the iterators, then this function will belong here if the type reference is shared.

janos: That is ok, thanks.

pkg/topology/test/helper.go
acud: it was a fun one :+1: please double check

pkg/kademlia/pslice/pslice.go
acud: there was a bug in one of the iterators, the label wasn't correctly placed. in any case, removed both labels :+1:

pkg/kademlia/pslice/pslice_test.go
acud: added tests for that :+1: there was a bug there. nice catch :+1: :clap:

pkg/kademlia/pslice/pslice_test.go
```diff
@@ -0,0 +1,335 @@
...
+		}, {
+			rmPO:  0,
+			expSE: 0,
+		},
+	} {
+		po := uint8(swarm.Proximity(base.Bytes(), peers[tc.rmPO].Bytes()))
+		ps.Remove(peers[tc.rmPO], po)
+
+		sd, none := ps.ShallowestEmpty()
+		if sd != tc.expSE || none {
```
nolash, line 73: should this test perhaps be performed with _more_ than one peer in each bin, too?

pkg/kademlia/pslice/pslice_test.go
```diff
@@ -0,0 +1,335 @@
...
+	// 2 peers per bin
+	// indexes {0,1} {2,3} {4,5} {6,7}
+	for i := 0; i < 8; i += 2 {
+		a := test.RandomAddressAt(base, i)
+		peers[i] = a
+
+		b := test.RandomAddressAt(base, i)
+		peers[i+1] = b
+	}
+	chkNotExists(t, ps, peers...)
```
nolash, line 100: necessary?

pkg/kademlia/pslice/pslice_test.go
```diff
@@ -0,0 +1,335 @@
...
+	chkBins(t, ps, []uint{0, 2, 2, 2})
+	chkExists(t, ps, peers[:2]...)
+	chkNotExists(t, ps, peers[2:]...)
+
+	ps.Add(peers[2], 1)
+	ps.Add(peers[3], 1)
+	chkLen(t, ps, 4)
+	chkBins(t, ps, []uint{0, 2, 4, 4})
+	chkExists(t, ps, peers[:4]...)
+	chkNotExists(t, ps, peers[4:]...)
```
nolash, line 134: is it necessary to keep checking peers that cannot have been added?

pkg/kademlia/pslice/pslice_test.go
```diff
@@ -0,0 +1,335 @@
...
+		i++
+		return stop, skipNext, nil
+	}
+
+	err := ps.EachBinRev(f)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if i != iterations {
```
nolash, line 278: You don't have a matching check in `testIterator`. Is that deliberate?

pkg/kademlia/pslice/pslice_test.go
```diff
@@ -0,0 +1,335 @@
...
+	if lp := len(pp); lp != l {
+		t.Fatalf("length mismatch, want %d got %d", l, lp)
+	}
+}
+
+func chkBins(t *testing.T, ps *pslice.PSlice, seq []uint) {
+	pb := pslice.PSliceBins(ps)
+	for i, v := range seq {
+		if pb[i] != v {
+			t.Fatalf("bin seq wrong, get %d want %d", pb[i], v)
```
nolash, line 314: include index in output?

pkg/topology/test/helper.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"fmt"
+	"math/rand"
+
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// RandomAddressAt generates a random address
+// at proximity order prox relative to address
+// if prox is negative a random address is generated.
```
nolash, line 16: I would conceal this with a `randomAddressAt`, in order to not overload the functionality of `RandomAddress`

pkg/topology/test/helper_test.go
```diff
@@ -0,0 +1,40 @@
...
+	for i := 0; i < bitsInCommon; i++ {
+		bb0 |= (1 << (31 - i))
+	}
+	xorb := make([]byte, 2)
+	binary.BigEndian.PutUint16(xorb, bb0)
+
+	for i := 0; i < 2; i++ {
+		xb := hw0[i] ^ hw1[i]
+		if xorb[i] != xb {
+			t.Fatal("xor value mismatch. good luck debugging this")
```
nolash, line 37: :D

pkg/topology/test/helper_test.go
```diff
@@ -0,0 +1,40 @@
...
+	b0 := base.Bytes()
+	b1 := addr.Bytes()
+
+	hw0 := []byte{b0[0], b0[1]} // highest words of 0
+	hw1 := []byte{b1[0], b1[1]} // highest words of 1
+
+	// bb0 is the result of the XOR between b0 and b1 words
+	bb0 := uint16(0)
+	for i := 0; i < bitsInCommon; i++ {
+		bb0 |= (1 << (31 - i))
```
nolash, line 29: Why not just set the bits to `0xffff`?

pkg/topology/test/helper_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"testing"
+
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/ethersphere/bee/pkg/topology/test"
+)
+
+func TestRandomAddressAt(t *testing.T) {
+	base := swarm.MustParseHexAddress("ca1e9f3938cc1425c6061b96ad9eb93e134dfe8734ad490164ef20af9d1cf59c")
+
+	bitsInCommon := 16
```
nolash, line 18: Perhaps it would be cooler with something not stopping at byte boundary

pkg/topology/test/helper_test.go
acud: not sure if i understand. bits in common could vary. you have the same problem

pkg/topology/test/helper_test.go
acud: yeah i actually wanted to have this test do all the test cases from 0 to over a byte (or over 16 if to be exact). will do the change

pkg/topology/test/helper.go
acud: not sure if i follow. wdym?

pkg/kademlia/pslice/pslice.go
```diff
@@ -0,0 +1,198 @@
...
+func (s *PSlice) ShallowestEmpty() (bin uint8, none bool) {
+	s.Lock()
+	defer s.Unlock()
+
+	binCp := make([]uint, len(s.bins)+1)
+	copy(binCp, s.bins)
+	binCp[len(binCp)-1] = uint(len(s.peers))
+
+	for i := uint8(0); i < uint8(len(binCp)-1); i++ {
+		if binCp[i+1]-binCp[i] == 0 {
```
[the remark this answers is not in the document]
acud: you mean to compare if `binCp[i+1]>binCp[i]`?

pkg/kademlia/pslice/pslice.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// ShallowestEmpty returns the shallowest empty bin if one exists.
+// If such bin does not exists, returns true as bool value.
+// Must be called under lock.
+func (s *PSlice) ShallowestEmpty() (bin uint8, none bool) {
+	s.Lock()
+	defer s.Unlock()
+
+	binCp := make([]uint, len(s.bins)+1)
```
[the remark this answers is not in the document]
acud: no. the last place is added in order to add the length of the peers slice. its a bit ugly but i wanted to prevent the explicit check for which index we are in and if we are in the last index of the slice then compare agains `len(peers)`. note that there would always be another index at the end but that index is not indexed since `bins` indexes only where bins begin, not end!

pkg/kademlia/pslice/pslice.go
```diff
@@ -0,0 +1,198 @@
...
+func (s *PSlice) Remove(addr swarm.Address, po uint8) {
+	s.Lock()
+	defer s.Unlock()
+
+	e, i := s.exists(addr)
+	if !e {
+		return
+	}
+
+	s.peers = append(s.peers[:i], s.peers[i+1:]...)
```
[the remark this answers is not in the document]
acud: no.
https://github.com/golang/go/wiki/SliceTricks

pkg/kademlia/pslice/pslice.go
```diff
@@ -0,0 +1,198 @@
...
+func (s *PSlice) Exists(addr swarm.Address) bool {
+	s.Lock()
+	defer s.Unlock()
+
+	b, _ := s.exists(addr)
+	return b
+}
+
+// checks if a peer exists. must be called under lock.
+func (s *PSlice) exists(addr swarm.Address) (bool, int) {
```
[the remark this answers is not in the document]
acud: because you don't always want to lock when calling. it will be evident with the kademlia integration

pkg/kademlia/pslice/pslice_test.go
```diff
@@ -0,0 +1,335 @@
...
+			expSE: 1,
+		}, {
+			rmPO:  14,
+			expSE: 1,
+		}, {
+			rmPO:  0,
+			expSE: 0,
+		},
+	} {
+		po := uint8(swarm.Proximity(base.Bytes(), peers[tc.rmPO].Bytes()))
```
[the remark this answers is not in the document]
acud: i do, but i want it to be clear to whoever reviews

pkg/kademlia/pslice/pslice.go
```diff
@@ -0,0 +1,198 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+// ShallowestEmpty returns the shallowest empty bin if one exists.
+// If such bin does not exists, returns true as bool value.
+// Must be called under lock.
+func (s *PSlice) ShallowestEmpty() (bin uint8, none bool) {
```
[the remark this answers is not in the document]
acud: this is an optimization. i'd be happy to do it when it is demonstrated that it gives an improvement. increasing complexity and amount of state we maintain on objects without justification is something i'd like to avoid. this is the same reason why this data structure is a simple slice and not some fancy `b+-!#^tree`.

pkg/kademlia/pslice/pslice.go
```diff
@@ -0,0 +1,198 @@
...
+	if len(s.peers) == 0 {
+		return nil
+	}
+
+	var binEnd = uint(len(s.peers))
+
+	for i := len(s.bins) - 1; i >= 0; i-- {
+		peers := s.peers[s.bins[i]:binEnd]
+		for _, v := range peers {
+			stop, next, err := pf(v, uint8(i))
```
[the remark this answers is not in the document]
acud: the callback can have an error as a result of the actual callback. see `pkg/shed/index.go#Iterate` for a comparable example

pkg/topology/test/helper_test.go
nolash: Maybe I'm missing something but in this code `bitsInCommon` is set to 16. `bb0` is `0x0000` before loop. The loop sets all bits to 1? Why the loop not just set to `0xffff` directly

pkg/topology/test/helper.go
nolash: Remove the negative option on this method. If you want fully random only call the other.

```
randomAddress(address a, int x) : x € Z

RandomAddressAt(address a, int x) : x >= 0
* calls randomAddressAt(a, x)
RandomAddress()
* calls randomAddressAt(a, -1)
```

pkg/kademlia/pslice/pslice.go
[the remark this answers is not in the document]
nolash: it would be `==` no?

pkg/kademlia/pslice/pslice.go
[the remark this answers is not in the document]
nolash: > its a bit ugly 

I didn't mean it was ugly. Just wanted to lighten the mood :)

pkg/kademlia/pslice/pslice.go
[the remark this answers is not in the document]
nolash: I meant the `peerExists` call that follows...

pkg/kademlia/pslice/pslice.go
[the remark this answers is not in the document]
nolash: > i'd be happy to do it when it is demonstrated that it gives an improvement.

I don't think that needs a demonstration.

But putting it off till optimization stage is totally fine of course.

pkg/kademlia/pslice/pslice.go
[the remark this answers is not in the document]
nolash: > the callback can have an error as a result of the actual callback

that much is clear but in what cases with this particular use case - that is, what the callback actually does, can an error occur? I'm not sure I see it clearly.

pkg/kademlia/pslice/pslice.go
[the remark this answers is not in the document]
nolash: oh, way to go, `golang`, allowing out of bound index in your syntax.

pkg/kademlia/pslice/pslice.go
```diff
@@ -0,0 +1,198 @@
...
+// calling context.
+type PSlice struct {
+	peers []swarm.Address
+	bins  []uint
+
+	sync.Mutex
+}
+
+// New creates a new PSlice.
+func New(maxBins int) *PSlice {
```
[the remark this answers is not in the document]
acud: tbh this is a bit of a wider issue. the problem is that `swarm.Proximity` returns an `int`. and not a `uint8`. I think it would be nice to have a decision about how we treat proximity order across the whole codebase. either `Proximity` should return `uint8` or it should be `int` across the board. in any case lets have it like so for this iteration, we can do this wider change later

pkg/kademlia/pslice/pslice.go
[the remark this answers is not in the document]
acud: well if the callback needs to call some other component e.g. localstore or state store then yes some error can be resulted as part of the callback

pkg/kademlia/pslice/pslice.go
```diff
@@ -0,0 +1,198 @@
...
+			}
+		}
+		binEnd = s.bins[i]
+	}
+
+	return nil
+}
+
+// EachBinRev iterates over all peers from shallowest to deepest.
+func (s *PSlice) EachBinRev(pf topology.EachPeerFunc) error {
```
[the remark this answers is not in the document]
acud: i think `asc` `desc` in this context is confusing since po works the other way around... you insist on this change?

pkg/kademlia/pslice/pslice_test.go
```diff
@@ -0,0 +1,335 @@
...
+
+	for i, v := range peers {
+		ps.Add(v, uint8(i))
+		sd, none := ps.ShallowestEmpty()
+		if i == 15 {
+			if !none {
+				t.Fatal("expected last bin to be empty and return no empty bins")
+			}
+		} else {
+			if sd != uint8(i+1) {
```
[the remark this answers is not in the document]
acud: i am expecting the empty bins to precede the depth.
so adding a peer at bin 0 should make the shallowest empty bin 1, adding then a peer in bin 1 will make shallowest empty to be 2, etc

pkg/kademlia/pslice/pslice_test.go
acud: hurts?

pkg/kademlia/pslice/pslice_test.go
acud: this is actually not needed since the check is already done in the callback immediately when there are too many iterations

pkg/kademlia/pslice/pslice.go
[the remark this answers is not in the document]
nolash: ok. is there an issue for this @acud ?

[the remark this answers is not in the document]
acud: in my head yes. i'll handle it. no more issues

pkg/kademlia/pslice/pslice_test.go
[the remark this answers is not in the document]
nolash: ah yes you are right, my bad.

pkg/kademlia/pslice/pslice.go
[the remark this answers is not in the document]
nolash: well, no, it's up to you ... not sure if I personally like `Rev` so much though, I don't find that abbreviation that widely used in symbol naming...?

pkg/topology/test/helper.go
acud: so you end up with three functions that do exactly what these two functions do, and your `RandomAddressAt` still accepts an `int` meaning it can take a negative value, but the condition you mention here does not handle that case. so if it is negative is the function supposed to panic? i dont really understand...

pkg/kademlia/pslice/pslice_test.go
acud: i dont see much harm. yes it might be a trivial check but i don't see it as harmful

pkg/kademlia/pslice/pslice_test.go
acud: yes but then i have to rewrite this entire test from the beginning. which would suck badly. i'll create a different test for this then

pkg/kademlia/pslice/pslice_test.go
nolash: > a trivial check

I had to look twice when reading through the test to understand that it's not testing anything really. It does not help clarity.

pkg/topology/test/helper.go
nolash: It seems to me in general that in our code and in `golang` overloading a method argument's (or return value's) with special meaning for negative numbers is not standard.

Perhaps just remove the comment line for the negative value?

pkg/kademlia/pslice/pslice_test.go
nolash: `cat | /dev/debt`

pkg/topology/test/helper.go
acud: > It seems to me in general that in our code and in `golang` overloading a method argument's (or return value's) with special meaning for negative numbers is not standard.

That's not entirely correct. See [1](https://golang.org/pkg/strings/#Index) and [2](https://golang.org/pkg/strings/#Replace)

It is not common, but it is definitely something acceptable.

I'll remove the comment about the negative number. Thanks

Issue #159, opened 2020-05-05
Implement pusher


janos:
Current pushsync package contains logic in `chunksWorker` function that should be moved into its own package `pusher` so that responsibilities are clearly separated.

Package `pushsync` should be responsible to push chunk with forwarding and receiving receipts on demand only. This means only when SendChunkAndReceiveReceipt is called. It should use storage only to Put chunks and may use a new storage.Putter interface that only has the Put method from Storer. Tests would have to adjusted in `pushsync` so that accommodate removal of `chunksWorker`.

Package `pusher` should consume chunks by storage.Storer's SubscribePush and use `pushsync` protocol to push them. Package `pusher` should ideally only expose a Service constructor and a Service Close method to terminate the goroutine that pusher is waiting the chunks on. The pusher Service should depend on `pushsync.Interface` which would have only one method SendChunkAndReceiveReceipt, and to storage.Storer, or if a smaller interface can be used instead the whole Storer. A new package `pushsync/mock` should be made in order to test `pusher` functionalities (see `pingpong/mock` as an example).

Pusher Service should be constructed in node.NewBee as part of its integration in running node.

PR #171, opened 2020-05-08
Pusher: Move push sync logic in to pusher

jmozah:
- Moved chunksWorker in to pusher package
- Made pushsync protocol dummy and it handles only one put chunk using a restricted storer interface
- Added pushed to node start and shutdown

Refer https://github.com/ethersphere/bee/issues/159

### DIFF
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -1,0 +1,147 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package pusher
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/pushsync"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/ethersphere/bee/pkg/topology"
+)
+
+type Service struct {
+	storer            storage.Storer
+	peerSuggester     topology.ClosestPeerer
+	pushSyncer        pushsync.PushSyncer
+	logger            logging.Logger
+	metrics           metrics
+	quit              chan struct{}
+	chunksWorkerQuitC chan struct{}
+}
+
+type Options struct {
+	Storer        storage.Storer
+	PeerSuggester topology.ClosestPeerer
+	PushSyncer    pushsync.PushSyncer
+	Logger        logging.Logger
+}
+
+var retryInterval = 10 * time.Second // time interval between retries
+
+func New(o Options) *Service {
+	service := &Service{
+		storer:            o.Storer,
+		peerSuggester:     o.PeerSuggester,
+		pushSyncer:        o.PushSyncer,
+		logger:            o.Logger,
+		metrics:           newMetrics(),
+		quit:              make(chan struct{}),
+		chunksWorkerQuitC: make(chan struct{}),
+	}
+	go service.chunksWorker()
+	return service
+}
+
+// chunksWorker is a loop that keeps looking for chunks that are locally uploaded ( by monitoring pushIndex )
+// and pushes them to the closest peer and get a receipt.
+func (s *Service) chunksWorker() {
+	var chunks <-chan swarm.Chunk
+	var unsubscribe func()
+	// timer, initially set to 0 to fall through select case on timer.C for initialisation
+	timer := time.NewTimer(0)
+	defer timer.Stop()
+	defer close(s.chunksWorkerQuitC)
+	chunksInBatch := -1
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		<-s.quit
+		cancel()
+	}()
+	for {
+		select {
+		// handle incoming chunks
+		case ch, more := <-chunks:
+			// if no more, set to nil, reset timer to 0 to finalise batch immediately
+			if !more {
+				chunks = nil
+				var dur time.Duration
+				if chunksInBatch == 0 {
+					dur = 500 * time.Millisecond
+				}
+				timer.Reset(dur)
+				break
+			}
+
+			chunksInBatch++
+			s.metrics.TotalChunksToBeSentCounter.Inc()
+			peer, err := s.peerSuggester.ClosestPeer(ch.Address())
+			if err != nil {
+				if errors.Is(err, topology.ErrWantSelf) {
+					// set chunk status to synced
+					s.setChunkAsSynced(ctx, ch.Address())
+					continue
+				}
+			}
+
+			_, err = s.pushSyncer.SendChunkAndReceiveReceipt(ctx, peer, ch)
+			if err != nil {
+				s.logger.Errorf("pusher: error while sending chunk or receiving receipt: %v", err)
+				continue
+			}
+
+			// set chunk status to synced
+			s.setChunkAsSynced(ctx, ch.Address())
+			continue
+
+			// retry interval timer triggers starting from new
+		case <-timer.C:
+			// initially timer is set to go off as well as every time we hit the end of push index
+			startTime := time.Now()
+
+			// if subscribe was running, stop it
+			if unsubscribe != nil {
+				unsubscribe()
+			}
+
+			// and start iterating on Push index from the beginning
+			chunks, unsubscribe = s.storer.SubscribePush(ctx)
+
+			// reset timer to go off after retryInterval
+			timer.Reset(retryInterval)
+			s.metrics.MarkAndSweepTimer.Observe(time.Since(startTime).Seconds())
+
+		case <-s.quit:
+			if unsubscribe != nil {
+				unsubscribe()
+			}
+			return
+		}
+	}
+}
+
+func (s *Service) setChunkAsSynced(ctx context.Context, addr swarm.Address) {
+	if err := s.storer.Set(ctx, storage.ModeSetSyncPush, addr); err != nil {
+		s.logger.Errorf("pusher: error setting chunk as synced: %v", err)
+		s.metrics.ErrorSettingChunkToSynced.Inc()
+	} else {
+		s.metrics.TotalChunksSynced.Inc()
+	}
+}
+
+func (s *Service) Close() error {
+	close(s.quit)
+
+	// Wait for chunks worker to finish
+	select {
+	case <-s.chunksWorkerQuitC:
+	case <-time.After(3 * time.Second):
+	}
+	return nil
+}
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -6,7 +6,6 @@ package pusher
 
 import (
 	"context"
-	"errors"
 	"time"
 
 	"github.com/ethersphere/bee/pkg/logging"
@@ -18,7 +17,6 @@ import (
 
 type Service struct {
 	storer            storage.Storer
-	peerSuggester     topology.ClosestPeerer
 	pushSyncer        pushsync.PushSyncer
 	logger            logging.Logger
 	metrics           metrics
@@ -38,7 +36,6 @@ var retryInterval = 10 * time.Second // time interval between retries
 func New(o Options) *Service {
 	service := &Service{
 		storer:            o.Storer,
-		peerSuggester:     o.PeerSuggester,
 		pushSyncer:        o.PushSyncer,
 		logger:            o.Logger,
 		metrics:           newMetrics(),
@@ -81,18 +78,10 @@ func (s *Service) chunksWorker() {
 
 			chunksInBatch++
 			s.metrics.TotalChunksToBeSentCounter.Inc()
-			peer, err := s.peerSuggester.ClosestPeer(ch.Address())
-			if err != nil {
-				if errors.Is(err, topology.ErrWantSelf) {
-					// set chunk status to synced
-					s.setChunkAsSynced(ctx, ch.Address())
-					continue
-				}
-			}
 
 			// Later when we process receipt, get the receipt and process it
 			// for now ignoring the receipt and checking only for error
-			_, err = s.pushSyncer.ChunkPusher(ctx, peer, ch)
+			_, err := s.pushSyncer.PushChunkToClosest(ctx, ch)
 			if err != nil {
 				s.logger.Errorf("pusher: error while sending chunk or receiving receipt: %v", err)
 				continue
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -90,7 +90,9 @@ func (s *Service) chunksWorker() {
 				}
 			}
 
-			_, err = s.pushSyncer.SendChunkAndReceiveReceipt(ctx, peer, ch)
+			// Later when we process receipt, get the receipt and process it
+			// for now ignoring the receipt and checking only for error
+			_, err = s.pushSyncer.ChunkPusher(ctx, peer, ch)
 			if err != nil {
 				s.logger.Errorf("pusher: error while sending chunk or receiving receipt: %v", err)
 				continue
--- a/pkg/pusher/pusher_test.go
+++ b/pkg/pusher/pusher_test.go
@@ -1,0 +1,173 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package pusher_test
+
+import (
+	"context"
+	"errors"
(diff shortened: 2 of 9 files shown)


jmozah:
@janos @acud i am submitting this PR so that i can work on your comments over the weekend... to save time... Please review and let me know your valuable comments

pkg/pusher/pusher_test.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	return nil
+}
+
+func TestSendChunkToPushSync(t *testing.T) {
+	chunk := createChunk(t)
+
+	// create a trigger  and a closestpeer
+	triggerPeer := swarm.MustParseHexAddress("6000000000000000000000000000000000000000000000000000000000000000")
+	closestPeer := swarm.MustParseHexAddress("6000000000000000000000000000000000000000000000000000000000000000")
```
acud, line 46: both have same address

pkg/pusher/pusher_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	pushSyncService := pushsyncmock.New(func(ctx context.Context, address swarm.Address, chunk swarm.Chunk) (*pb.Receipt, error) {
+		receipt := &pb.Receipt{
```
acud, line 49: can `return &pb....`

pkg/pusher/pusher_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	_, err := storer.Put(context.Background(), storage.ModePutUpload, chunk)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Give some time for chunk to be pushed and receipt to be received
+	time.Sleep(10 * time.Millisecond)
+
+	err = checkIfModeSet(t, chunk.Address(), storage.ModeSetSyncPush, storer)
```
acud, line 65: you can fail directly inside the helper method, that's why passing `t` is helpful

pkg/pusher/pusher_test.go
```diff
@@ -0,0 +1,174 @@
...
+	if err == nil {
+		t.Fatalf("chunk not syned error expected")
+	}
+}
+
+func createChunk(t *testing.T) swarm.Chunk {
+	t.Helper()
```
acud, line 129: you don't need `t.Helper` and in fact `t` at all. can be safely removed

pkg/pusher/pusher_test.go
```diff
@@ -0,0 +1,174 @@
...
+		t.Fatal(err)
+	}
+}
+
+func TestSendChunkAndReceiveInvalidReceipt(t *testing.T) {
+	chunk := createChunk(t)
+
+	// create a trigger  and a closestpeer
+	triggerPeer := swarm.MustParseHexAddress("6000000000000000000000000000000000000000000000000000000000000000")
+	closestPeer := swarm.MustParseHexAddress("6000000000000000000000000000000000000000000000000000000000000000")
```
acud, line 76: same address

pkg/pusher/pusher_test.go
```diff
@@ -0,0 +1,174 @@
...
+		t.Fatalf("chunk not syned error expected")
+	}
+}
+
+func TestSendChunkAndTimeoutinReceivingReceipt(t *testing.T) {
+	chunk := createChunk(t)
+
+	// create a trigger  and a closestpeer
+	triggerPeer := swarm.MustParseHexAddress("6000000000000000000000000000000000000000000000000000000000000000")
+	closestPeer := swarm.MustParseHexAddress("6000000000000000000000000000000000000000000000000000000000000000")
```
acud, line 103: same

pkg/pusher/pusher_test.go
```diff
@@ -0,0 +1,174 @@
...
+	for k, v := range storer.ModeSet {
+		if addr.String() == k {
+			found = true
+			if v != mode {
+				return errors.New("chunk mode is not properly set as synced")
+			}
+		}
+	}
+	if !found {
+		return errors.New("Chunk not synced")
```
acud, line 171: you can fail directly here with `t.Fatal`, otherwise remove `t`

pkg/pushsync/pushsync.go
```diff
@@ -220,115 +204,32 @@ func (ps *PushSync) receiveReceipt(r protobuf.Reader) (receipt pb.Receipt, err e
...
-}
-
 // sendChunkAndReceiveReceipt sends chunk to a given peer
-// by opening a stream. It then waits for a receipt from that peer.
-// Once the receipt is received within a given time frame it marks that this chunk
-// as synced in the localstore.
-func (ps *PushSync) SendChunkAndReceiveReceipt(ctx context.Context, peer swarm.Address, ch swarm.Chunk) error {
+// by opening a stream. It then waits for a receipt from that peer and
+// returns error or nil based on the receiving and the validity of the receipt.
+func (ps *PushSync) SendChunkAndReceiveReceipt(ctx context.Context, peer swarm.Address, ch swarm.Chunk) (*pb.Receipt, error) {
```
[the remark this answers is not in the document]
acud: yeah I don't really like the protobuf types either. I wonder if we can have them somehow as `internal`? just a suggestion to prevent such cases

pkg/storage/store.go
```diff
@@ -139,6 +139,10 @@ type Storer interface {
 	io.Closer
 }
 
+type Putter interface {
+	Put(ctx context.Context, mode ModePut, chs ...swarm.Chunk) (exist []bool, err error)
```
acud, line 143: maybe replace line 132 with `Putter`?

pkg/pushsync/pushsync.go
```diff
@@ -25,40 +25,35 @@ const (
 	streamName      = "pushsync"
 )
 
+type PushSyncer interface {
+	SendChunkAndReceiveReceipt(ctx context.Context, peer swarm.Address, ch swarm.Chunk) (*pb.Receipt, error)
```
acud, line 29: I suggest to rename the method to `PushChunk` and interface to `Pusher` or `ChunkPusher`. Receipts should be implicit

pkg/pusher/pusher_test.go
```diff
@@ -0,0 +1,174 @@
...
+	pushsyncmock "github.com/ethersphere/bee/pkg/pushsync/mock"
+	"github.com/ethersphere/bee/pkg/pushsync/pb"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/ethersphere/bee/pkg/topology/mock"
+)
+
+type Store struct {
+	storage.Storer
+	ModeSet   map[string]storage.ModeSet
```
acud, line 28: why export?

acud:
@jmozah some other things that i remember that were discussed, is that push sync protocol will now be a really simple bare-bones protocol, for me this means also that all of the forwarding logic might have to be stripped out of there. this is evident by the fact that both `pusher` and `pushsync` service structs have `storer` and `peerSuggester` as fields they reference, and both components actively use these components to implement business logic that has to do with storing, sending and forwarding chunks and receipts. i think it is better that the protocol just handles sending and receiving chunks and receipts, and pusher can decide on whether to send a receipt back, keep forwarding or what not.
my 2 cents

janos:
Thanks for the review @acud. I missed some of the things. But, moving forwarding to pusher is not in the scope of this pr as per issue https://github.com/ethersphere/bee/issues/159. I think that forwarding is better suited for the protocol itself regarding waiting for the receipts and how streams help in this case. Forwarding is controlled by SuggestPeer, I think well enough, as logic where to forward is outsourced to a dependency, but forwarding as the protocol feature is implemented by most convenient way. I have an impression that moving forwarding to pusher would add more complexity then it is now in respect to receipts getting back to the pusher node.

pkg/pusher/pusher_test.go
jmozah: yeah but doesn't this looks readable and clean? if you  insist i will change

pkg/pusher/pusher_test.go
jmozah: we need to pass the error outside to make decisions as some testcase expect failure. I removed t as a argument which was redundant.,

pkg/storage/store.go
jmozah: cool..

pkg/pushsync/pushsync.go
jmozah: The interface name cannot be Pusher as it i will confuse with the pusher pkg we will already have.  May be i can rename the function to ChunkPusher.

jmozah:
@acud BTW I have a serious question? i really want to reduce the review comments on naming conventions. do we have any naming conventions for naming things.. like the change from SendChunkAndReceiveReceipt to ChunkPusher... how can i avoid review comments in future on adding functions names that reviewer don't seem to prefer?

pkg/pusher/pusher_test.go
jmozah: removed t  as said in the earlier comment.

jmozah:
@janos @acud Thanks for your first round of comment on a very short notice. I have addressed all of them. Please have a look at this PR once more and either add more comments or approve it.

acud:
> I have an impression that moving forwarding to pusher would add more complexity then it is now in respect to receipts getting back to the pusher node.

I agree with most of what you say @janos and I take your opinions, but I still have a feeling there is no clear separation of concerns here, and the reason is that pusher also has `peerSuggester` and uses it. I am fine if we agree that `pusher` component's responsibility is just to poll the localstore push index and queue the chunks for the protocol to send them out, then the protocol handles sending/fw/receipts (but then also selecting which peers to send to, meaning the whole peer selection part gets deleted out of `pusher` along with the `peerSuggester` field), so then we can really separate the syntactic glue between the localstore push index and the push sync protocol, that is, pusher, from the business logic of actually sending a chunk, forwarding it, then sending the receipt back.
WDYT?

jmozah:
@acud  @janos Thanks for the nice discussion above. These are my 2 cents regarding this. 

> I still have a feeling there is no clear separation of concerns here, and the reason is that pusher also has peerSuggester and uses it

The usage of peerSuggester in pusher will be used in the originating peer of the chunk and the peerSuggester inside the pushSync will be used in the forwarding peers. They solve different purposes so they should have their own instances is my view.

The reason is, If we remove peerSuggester from pusher and keep only in protocol, We loose forever the option of forwarding chunks to ANY peer that you want (not necessarily the closest) because you have glued that in protocol itself.  Now, only forwarding is handled by the protocol, not the destination. 

Also  in global pinning scenario the requesting node will not be the the closest node for sure. So if only protocol has the peerSuggester , it will not know who is the requesting peer for that chunk in the first place.

acud:
> The usage of `peerSuggester` in `pusher `will be used in the finding the destination peer of the chunk and the `peerSuggester `inside the `pushSync `will be used for finding closest forwarding peers. 

I'm not sure I understand. How are they different? Both use exactly the same method call.

> If we remove `peerSuggester `from pusher and keep only in protocol, We loose forever the option of forwarding chunks to ANY peer that you want (not necessarily the closest) because you have glued that in protocol itself. Now, only forwarding is handled by the protocol, not the destination.

Well, in the case this functionality is needed, push sync protocol can expose a method that can send a chunk to a specific peer, I see zero overhead in exposing such functionality, but in case this was not understood, I am arguing that `pusher` should have no notion of `peerSuggester`

jmozah:
@acud  I agree about the first point. you addressed above.  
Second point you seem to have accepted and provided an alternative solution which is okay for me,

What i seem to not get you is why should `peerSuggester ` not be part of pusher when other things like `storer `are shared between them. What is the logical design decision for this? Do you think anything to do with topology is part of the protocol? if that is so, then i agree what you say and lets please put it down somewhere please or shout it out in the core channel as a design desision?

acud:
> What is the logical design decision for this?

Ideally we should minimize the amount of duplicate business logic across components. If you get a chunk on pusher (from localstore), on pushsync protocol (due to handling of delivery) or on global pinning (due to chunk missing), and each one of those components need to forward the chunk (as an example, not as a binding fact of reality) according to the same heuristic, then it means that there is duplicate business logic between those components and that it could be refactored to be reused. 

Also, I think this is in line with the approach of trying to minimize the size of the interface surface that components should have with each other, i.e. we should not introduce interface points between components because we can, but because we must.

If this approach is taken, there is just less code to write, less code to test and less code to maintain.

jmozah:
> Ideally we should minimize the amount of duplicate business logic across components.

Agreed. But in this case, they are not duplicate. they are different in different scenarios. For ex: the chunk addressed used in the pushsync and the chunk address used in the global pinning should not be treated the same way. That's the logic of strategy layer like pusher. From a protocol layer perspective, pushsync sends a chunk to a given peer as dumb as that.

> I think this is in line with the approach of trying to minimize the size of the interface surface that components should have with each other

I agree with this too.  Few comments above you suggested to have two interface one with peer and one without which is contradictory.


The only argument which you said that looks right for me is... As a design decision, we consoder topology close to protocols (kad, discovery etc) and therefore we should use topology only in these modules and not in strategy modules. If that is agreed, i accept the argument and close it here.

jmozah:
@acud @janos For the record, if you think these questions i am asking are irrelevant and slowing down us please let me know... i have this issue only with subjective comments. As a team we have to try to be more tangible in giving comments so that people like me can understand. Thanks for your patience.  Let me know your final decision and i will do it. 

Thanks guys... had a great time working on this PR

pkg/pusher/pusher.go
```diff
@@ -0,0 +1,149 @@
...
+				if errors.Is(err, topology.ErrWantSelf) {
+					// set chunk status to synced
+					s.setChunkAsSynced(ctx, ch.Address())
+					continue
+				}
+			}
+
+			// Later when we process receipt, get the receipt and process it
+			// for now ignoring the receipt and checking only for error
+			_, err = s.pushSyncer.ChunkPusher(ctx, peer, ch)
```
janos, line 95: Thank you @acud for spotting this improvement, that I actually missed while writing the the task description.

I agree that peerSuggester is not needed in pusher and that `pushSyncer.ChunkPusher` method can accept only a chunk and deal with first peer, just as it deals with the forwarding. This would remove pusher dependency peerSuggester, which is used only here. I am also suggesting to change the name of the method `ChunkPusher` from a noun to something with a ver, like, `PushChunk`, or `PushChunkToClosestPeer`, maybe there could be something prettier.

jmozah: Thanks @janos. I am removing the topology from pusher per our discussion.

jmozah:
@janos @acud I think i fixed all your comments. Can you please go through this and either comment more or approve this PR?

pkg/pushsync/mock/mock.go
```diff
@@ -12,13 +12,13 @@ import (
 )
 
 type PushSync struct {
-	SendChunk func(ctx context.Context, peer swarm.Address, chunk swarm.Chunk) (*pushsync.Receipt, error)
+	SendChunk func(ctx context.Context, chunk swarm.Chunk) (*pushsync.Receipt, error)
```
acud, line 15: why export?

pkg/pushsync/pushsync.go
```diff
@@ -26,7 +26,7 @@ const (
 )
 
 type PushSyncer interface {
-	ChunkPusher(ctx context.Context, peer swarm.Address, ch swarm.Chunk) (*Receipt, error)
+	PushChunkToClosest(ctx context.Context, ch swarm.Chunk) (*Receipt, error)
```
acud, line 29: @jmozah golang convention on naming interfaces when it is a single-function interface, is that the name of the interface is the name of the function + `er` (or something in that manner, in a way that makes sense).
e.g.: 
- `fmt.Stringer` - is the interface for `String()` method
- `io.Closer` - interface for `Close() error` method
- `io.Reader` - interface for `Read()` method

The list goes on...
Can you please follow the same convention and rename the function to be consistent with the naming convention?
Also, I previously asked to rename the interface to `ChunkPusher`, I'm not sure why we ended up with `PushSyncer`

And another comment about the method name, is that you're leaking details about the internals of what the implementation is doing by the name. Personally I prefer shorter names that don't bind to some specific set of actions. We should treat implementation details as black-box on the package boundary

janos: In this case it is a bit more difficult to name this method, as we will have two methods in this interface, one for pushing to the closest peer (forwarding), and one to push directly to a specific peer (no forwarding). So, PushChunkToClosest does reveal the implementation detail, but it is also a requirement for that method.

jmozah:
@acud @janos I didn't close it for sure this time. I pushed the last commit and watching the builds.. suddenly i see that it closed in my name, janos updated and janos merged...  For sure this time i didn't close it  :-)

PR #173, opened 2020-05-11
set explicit protobuf package names

janos:
This PR is fixing a problem with namespacing protobuf message types with `proto.RegisterType` which registers protobuf messages globally under names that contain protobuf package name and message name. Currently we are using the same name for protobuf package to get the same go package name `pb`. This PR changes that to have different name for protobuf and its namespacing and an explicit one for go package name.

This issue was discovered on the same named (but different) message `Delivery` in `retrieval` and `pushsync`. The mistake is from the start how package names are referenced and this PR fixes that for all.


janos:
> Wow. This is pretty serious. How did you catch this and how did this not come up in the tests?

Tests are not registering all protocols, and the runtime error is just logged by protobuf library, not panicked. Also this did not come up as pushsync protocol was not added as protocol to p2p service (another pr).

PR #175, opened 2020-05-11
check if its only local branch

vandot:
fixing githook for case when there is no upstream branch

### DIFF
--- a/.githooks/pre-push.bash
+++ b/.githooks/pre-push.bash
@@ -4,7 +4,7 @@ set -euo pipefail
 
 # Get current branch name
 current=$(git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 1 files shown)


.githooks/pre-push.bash
```diff
@@ -2,10 +2,14 @@
 
 set -euo pipefail
 
-# Check if we actually have commits to push
-commits=$(git log @{u}..)
-if [ -z "$commits" ]; then
-    exit 0
+# Get current branch name
+current=$(git rev-parse --abbrev-ref HEAD)
+if git branch -r | grep "^  origin/${current}$" &> /dev/null; then
```
janos, line 7: If someone renames the remote differently than `origin` it probably would not work. Maybe to replace `origin` with `$1` which is a remote?

PR #186, opened 2020-05-14
Add CLI application for splitting data to chunks

nolash:
Adds a cli application that splits data from file or stdin and stores chunks to filesystem and/or `bee`'s http api.

It adds the `storage.Getter` interface and uses this in the `storage.Storer` interface. This enables mere `Putter` and `Getter` to be used symmetrically by `file/splitter` and `file/joiner`, whose implementations have been changed accordingly.

### DIFF
--- a/cmd/bee-split/main.go
+++ b/cmd/bee-split/main.go
@@ -76,7 +76,7 @@ func newFsStore(path string) storage.Putter {
 func (f *fsStore) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
 	for _, ch := range chs {
 		chunkPath := filepath.Join(f.path, ch.Address().String())
-		err := ioutil.WriteFile(chunkPath, ch.Data(), 0o777)
+		err := ioutil.WriteFile(chunkPath, ch.Data(), 0o666)
 		if err != nil {
 			return []bool{}, err
 		}
@@ -182,7 +182,7 @@ func Split(cmd *cobra.Command, args []string) (err error) {
 	// add the fsStore and/or apiStore, depending on flags
 	stores := newTeeStore()
 	if outdir != "" {
-		err := os.MkdirAll(outdir, 0750)
+		err := os.MkdirAll(outdir, 0o777) // skipcq: GSC-G301
 		if err != nil {
 			return err
 		}
@@ -228,10 +228,10 @@ Chunks are saved in individual files, and the file names will be the hex address
 		RunE: Split,
 	}
 
-	c.Flags().StringVar(&outdir, "output-dir", "", "saves chunks to given directory")
-	c.Flags().Int64Var(&inputLength, "count", 0, "read at most this many bytes")
+	c.Flags().StringVarP(&outdir, "output-dir", "d", "", "saves chunks to given directory")
+	c.Flags().Int64VarP(&inputLength, "count", "c", 0, "read at most this many bytes")
 	c.Flags().StringVar(&host, "host", "127.0.0.1", "api host")
-	c.Flags().IntVar(&port, "port", 8500, "api port")
+	c.Flags().IntVar(&port, "port", 8080, "api port")
 	c.Flags().BoolVar(&ssl, "ssl", false, "use ssl")
 	c.Flags().BoolVar(&noHttp, "no-http", false, "skip http put")
 	err := c.Execute()
--- a/cmd/bee-split/main.go
+++ b/cmd/bee-split/main.go
@@ -53,10 +53,10 @@ func (t *teeStore) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.C
 	for _, putter := range t.putters {
 		_, err := putter.Put(ctx, mode, chs...)
 		if err != nil {
-			return []bool{}, err
+			return nil, err
 		}
 	}
-	return []bool{}, nil
+	return nil, nil
 }
 
 // fsStore provides a storage.Putter that writes chunks directly to the filesystem.
@@ -78,10 +78,10 @@ func (f *fsStore) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Ch
 		chunkPath := filepath.Join(f.path, ch.Address().String())
 		err := ioutil.WriteFile(chunkPath, ch.Data(), 0o666)
 		if err != nil {
-			return []bool{}, err
+			return nil, err
 		}
 	}
-	return []bool{}, nil
+	return nil, nil
 }
 
 // apiStore provies a storage.Putter that adds chunks to swarm through the HTTP chunk API.
@@ -114,11 +114,11 @@ func (a *apiStore) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.C
 		c := &http.Client{}
 		res, err := c.Post(url, "application/octet-stream", buf)
 		if err != nil {
-			return []bool{}, err
+			return nil, err
 		}
 		_ = res
 	}
-	return []bool{}, nil
+	return nil, nil
 }
 
 // limitReadCloser wraps the input to the application to limit the input to the given count flag.
--- a/cmd/bee-split/main.go
+++ b/cmd/bee-split/main.go
@@ -170,7 +170,6 @@ func Split(cmd *cobra.Command, args []string) (err error) {
 		if err != nil {
 			return err
 		}
-		defer f.Close()
 		infile = newLimitReadCloser(f, f.Close, inputLength)
 	} else {
 		// this simple splitter is too stupid to handle open-ended input, sadly
--- a/cmd/bee-split/main.go
+++ b/cmd/bee-split/main.go
@@ -16,7 +16,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"time"
 
 	"github.com/ethersphere/bee/pkg/file/splitter"
 	"github.com/ethersphere/bee/pkg/storage"
@@ -107,16 +106,18 @@ func newApiStore(host string, port int, ssl bool) (storage.Putter, error) {
 
 // Put implements storage.Putter
 func (a *apiStore) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
+	c := http.DefaultClient
 	for _, ch := range chs {
 		addr := ch.Address().String()
 		buf := bytes.NewReader(ch.Data())
 		url := strings.Join([]string{a.baseUrl, addr}, "/")
-		c := &http.Client{}
 		res, err := c.Post(url, "application/octet-stream", buf)
 		if err != nil {
 			return nil, err
 		}
-		_ = res
+		if res.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("upload failed: %v", res.Status)
+		}
 	}
 	return nil, nil
 }
@@ -199,7 +200,7 @@ func Split(cmd *cobra.Command, args []string) (err error) {
 
 	// split and rule
 	s := splitter.NewSimpleSplitter(stores)
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*60)
+	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	addr, err := s.Split(ctx, infile, inputLength)
 	if err != nil {
--- a/cmd/file/split.go
+++ b/cmd/file/split.go
@@ -18,20 +18,20 @@ import (
 	"strings"
 	"time"
 
-	"github.com/ethersphere/bee/pkg/swarm"
-	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/file/splitter"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/spf13/cobra"
 )
 
 var (
 	notImplementedError = errors.New("method not implemented")
-	outdir string // flag variable, output dir for fsStore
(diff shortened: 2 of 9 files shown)


cmd/file/split.go
```diff
@@ -0,0 +1,241 @@
...
+	return &fsStore{
+		path: path,
+	}
+}
+
+// Put implements storage.Putter
+func (f *fsStore) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
+	for _, ch := range chs {
+		chunkPath := filepath.Join(f.path, ch.Address().String())
+		err := ioutil.WriteFile(chunkPath, ch.Data(), 0o777)
```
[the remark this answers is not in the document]
nolash: > Probably does not parse new octal notation

it complained about the dir octal, so yeah - peculiar. `666` :metal: it is

cmd/file/split.go
```diff
@@ -0,0 +1,241 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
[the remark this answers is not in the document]
nolash: Actually I am calling it `split` :) but you are saying that the  fact that the directory name `file` will have some effect here?

cmd/file/split.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	// output the resulting hash
+	fmt.Println(addr)
+	return nil
+}
+
+func main() {
+	c := &cobra.Command{
+		Use:   "split [datafile]",
```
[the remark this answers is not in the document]
nolash: Ok I see what you mean. So actually, I was imagining separate executables for `split` and `join` for now.

In general I don't like swiss-army-knife kind of binaries. I really prefer one root command doing one thing, and doing it well.

cmd/file/split.go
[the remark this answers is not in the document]
janos: Yes, `go build ./cmd/file` produces the binary named file. It is ok to rename the directory to split, or to call it bee-split or bzz-split.

cmd/file/split.go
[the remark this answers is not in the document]
janos: Of course, separate executables are fine.

cmd/bee-split/main.go
```diff
@@ -0,0 +1,241 @@
...
+func (t *teeStore) Add(putter storage.Putter) {
+	t.putters = append(t.putters, putter)
+}
+
+// Put implements storage.Putter
+func (t *teeStore) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
+	for _, putter := range t.putters {
+		_, err := putter.Put(ctx, mode, chs...)
+		if err != nil {
+			return []bool{}, err
```
acud, line 56: you can `return nil,err`, since `[]bool` is a slice, it is nullable, same goes to the other places in the PR where this is done

cmd/bee-split/main.go
```diff
@@ -0,0 +1,241 @@
...
+	}, nil
+}
+
+// Put implements storage.Putter
+func (a *apiStore) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
+	for _, ch := range chs {
+		addr := ch.Address().String()
+		buf := bytes.NewReader(ch.Data())
+		url := strings.Join([]string{a.baseUrl, addr}, "/")
+		c := &http.Client{}
```
acud, line 114: i think you can safely declare this before the `for` loop. also, if you don't customize the client (you're not), you can just declare `c := http.DefaultClient`

cmd/bee-split/main.go
```diff
@@ -0,0 +1,241 @@
...
+		res, err := c.Post(url, "application/octet-stream", buf)
+		if err != nil {
+			return []bool{}, err
+		}
+		_ = res
```
acud, line 119: `_ = ?`

cmd/bee-split/main.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	return []bool{}, nil
+}
+
+// limitReadCloser wraps the input to the application to limit the input to the given count flag.
+type limitReadCloser struct {
+	io.Reader
+	closeFunc func() error
```
acud, line 127: `closeFunc io.Closer`
or I'd even go as far as changing to:
```go
type limitReadCloser struct {
	io.ReadCloser
}
```

cmd/bee-split/main.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// newLimitReadCloser creates a new limitReadCloser.
+func newLimitReadCloser(r io.Reader, closeFunc func() error, c int64) io.ReadCloser {
+	return &limitReadCloser{
+		Reader:    io.LimitReader(r, c),
+		closeFunc: closeFunc,
```
acud, line 134: if you change the struct to the aforementioned suggestion above, you can set `Close` her with the provided `closeFunc`, then you can remove the `Close()` implementation below (if i am not mistaken). just a suggestion

cmd/bee-split/main.go
```diff
@@ -0,0 +1,241 @@
...
+		} else {
+			inputLength = fileLength
+		}
+
+		// open file and wrap in limiter
+		f, err := os.Open(args[0])
+		if err != nil {
+			return err
+		}
+		defer f.Close()
```
acud, line 172: this is strange to me and i think this is a potential panic. you're calling defer (this will be executed only when the entire `Split` functions runs to end, but you're providing `f.Close` to the `ReadCloser`, which means some other calling context might try to close the file twice. can you elaborate on why is this needed in this manner?

cmd/bee-split/main.go
```diff
@@ -0,0 +1,241 @@
...
+		infile = newLimitReadCloser(f, f.Close, inputLength)
+	} else {
+		// this simple splitter is too stupid to handle open-ended input, sadly
+		if inputLength == 0 {
```
acud, line 176: so do you must know the length of content a-priori? what happens if you have the length wrong? does it `Floor` the value to the actual length?

cmd/bee-split/main.go
```diff
@@ -0,0 +1,241 @@
...
+		store, err := newApiStore(host, port, ssl)
+		if err != nil {
+			return err
+		}
+		stores.Add(store)
+	}
+
+	// split and rule
+	s := splitter.NewSimpleSplitter(stores)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*60)
```
acud, line 202: since you're not using any cobra context and i can't see any code that listens to system signals, so the question asked is - can you even `kill` this process at all? if not - is this the intended behavior?

cmd/bee-split/main.go
```diff
@@ -0,0 +1,241 @@
...
+		baseUrl: u.String(),
+	}, nil
+}
+
+// Put implements storage.Putter
+func (a *apiStore) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
+	for _, ch := range chs {
+		addr := ch.Address().String()
+		buf := bytes.NewReader(ch.Data())
+		url := strings.Join([]string{a.baseUrl, addr}, "/")
```
acud, line 113: why not have a helper function that does this?

cmd/bee-split/main.go
nolash: > which means some other calling context might try to close the file twice

what do you mean? `os.Open` opens a unique file descriptor.

cmd/bee-split/main.go
nolash: For stdin you can't know, of course. So if you have less data than the `count` it will fail. If you have more, it will stop at the limit (that's what the limireader is for)

cmd/bee-split/main.go
nolash: `SIGINT` will terminate it

cmd/bee-split/main.go
nolash: Why would that be necessary?

cmd/bee-split/main.go
nolash: I tried that and it didn't work. I think the methods then have to be explicit. Or the object passed in the argument must itself be a readcloser.

cmd/bee-split/main.go
nolash: (See below)

cmd/bee-split/main.go
acud: wouldn't it be cleaner and more readable to have:
```go
res, err := c.Post(url(addr), "application/octet-stream", buf)
```
than this hardcoded string slice literal + `Join`?

cmd/bee-split/main.go
acud: how can it fail if you have less than the `count`, now that the timeout is gone?

cmd/bee-split/main.go
acud: since you're passing `f.Close` to the `ReadCloser` it means some other context can call the close, and it will cause the `defer` statement to panic since the file handle is already closed.

cmd/bee-split/main.go
nolash: If you insist I can do it, sure.

cmd/bee-split/main.go
nolash: The splitter is being told up front how much data to expect. If it doesn't get enough it will fail.

cmd/bee-split/main.go
acud: there you go:
```
diff --git a/cmd/bee-split/main.go b/cmd/bee-split/main.go
index 108be4f..9f52625 100644
--- a/cmd/bee-split/main.go
+++ b/cmd/bee-split/main.go
@@ -125,22 +125,17 @@ func (a *apiStore) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.C
 // limitReadCloser wraps the input to the application to limit the input to the given count flag.
 type limitReadCloser struct {
        io.Reader
-       closeFunc func() error
+       io.Closer
 }
 
 // newLimitReadCloser creates a new limitReadCloser.
-func newLimitReadCloser(r io.Reader, closeFunc func() error, c int64) io.ReadCloser {
+func newLimitReadCloser(r io.Reader, cl io.Closer, c int64) io.ReadCloser {
        return &limitReadCloser{
-               Reader:    io.LimitReader(r, c),
-               closeFunc: closeFunc,
+               Reader: io.LimitReader(r, c),
+               Closer: cl,
        }
 }
 
-// Close implements io.Closer
-func (l *limitReadCloser) Close() error {
-       return l.closeFunc()
-}
-
 // Split is the underlying procedure for the CLI command
 func Split(cmd *cobra.Command, args []string) (err error) {
        var infile io.ReadCloser
@@ -171,13 +166,13 @@ func Split(cmd *cobra.Command, args []string) (err error) {
                        return err
                }
                defer f.Close()
-               infile = newLimitReadCloser(f, f.Close, inputLength)
+               infile = newLimitReadCloser(f, f, inputLength)
        } else {
                // this simple splitter is too stupid to handle open-ended input, sadly
                if inputLength == 0 {
                        return errors.New("must specify length of input on stdin")
                }
-               infile = newLimitReadCloser(os.Stdin, func() error { return nil }, inputLength)
+               infile = newLimitReadCloser(os.Stdin, newNoopCloser(), inputLength)
        }
 
        // add the fsStore and/or apiStore, depending on flags
@@ -240,3 +235,14 @@ Chunks are saved in individual files, and the file names will be the hex address
                os.Exit(1)
        }
 }
+
+type noopCloser struct {
+}
+
+func newNoopCloser() io.Closer {
+       return noopCloser{}
+}
+
+func (_ noopCloser) Close() error {
+       return nil
+}
```

cmd/bee-split/main.go
nolash: you are right @acud. If you give it less data what you indicated, it will still hash with the data it has, but with the wrong span length. Hm. That needs to be enforced in the splitter, though. Thanks.

cmd/bee-split/main.go
nolash: After discussing this in person conclusion is it's ok to skip this.

nolash:
@janos thanks :) I noticed now though that I never actually committed the `pkg/file/splitter/splitter_test.go` file before. I just saw now when I was adding a test for the premature end of data problem pointed out by @acud. May I ask you please have a look at that too. Sorry.

pkg/file/splitter/splitter_test.go
```diff
@@ -0,0 +1,121 @@
...
+	g := mockbytes.New(0, mockbytes.MockTypeStandard).WithModulus(255)
+	testData, err := g.SequentialBytes(swarm.ChunkSize * swarm.Branches)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	store := mock.NewStorer()
+	s := splitter.NewSimpleSplitter(store)
+
+	//ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond * 500)
```
janos, line 86: Remove commented code.

pkg/file/splitter/splitter_test.go
```diff
@@ -0,0 +1,121 @@
...
+	testData, err := g.SequentialBytes(swarm.ChunkSize)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	store := mock.NewStorer()
+	s := splitter.NewSimpleSplitter(store)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
janos, line 51: There are no explicit parallel work being done in these tests, so cancel is not canceling any goroutine. I think that it is ok just to have context.Background() instead in all tests in this file.

Issue #182, opened 2020-05-13
Local pinning of chunks


Eknir:
## Story
As a Swarm developer, I want to have the possibility to pin chunks, such that I can develop more higher-level pinning solutions 

## Background
As we are exposing all APIs on the chunk-level, having the possibility to pin chunks makes sense. Doing this will allow other applications to pin anything on Swarm (files, databases, etc).

## Task Description
1) Create 3 debug APIs for already present chunks
       - POST       https://host:port/chunk-pin/<chunk hash>
       - DELETE   https://host:port/chunk-pin/<chunk hash>
       - GET          https://host:port/chunk-pin

2) You should be able to pin chunks while uploading if it has  a `x-swarm-pin` header
  
3) POST method should increment the pin counter in the `Lcalstore `with `ModeSetPin`

4) DELETE method should decrement the pin counter in the `Localstore ` with `ModeSetUnpin`. 
    If the counter becomes zero, make the chunk available for GC.

5) Until the `PinCounter `is more than 0, GC should not delete the chunk from the `localstore`
     Well.. GC is not done yet so just comment it there.

6) Get should show all the chunks that are pinned, in json format. if the no of chunks are large 
    this will cause issues.. may be we can use some of the guards
          - show only some MAX no of chunks
          - show chunks in pagination  (in future)
           - show chunks with some prefix in pagination (if badger comes in)
 
7) Use mocked localstore and extend it to check for `Modes `and `PinCounters `

8) All these 3 APIs should be in debug for now.

## Acceptance criteria
- Situation: User uploads two chunks, where one is pinned (a) and the other is not pinned (b).
- Situation: Enough data is synced to the the node, such that the non-pinned chunk is garbage-collected.
- Verify: The pinned chunk should still be in the database.

Eknir:
@jmozah

PR #187, opened 2020-05-15
Local pinning of chunk

jmozah:
Refer https://github.com/ethersphere/bee/issues/182 for task details

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /dist
 /.idea
 /.vscode
+/vendor
 
 # Compiled Object files, Static and Dynamic libs (Shared Objects)
 *.o
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 /dist
 /.idea
 /.vscode
-/vendor
 
 # Compiled Object files, Static and Dynamic libs (Shared Objects)
 *.o
--- a/pkg/api/bzzchunk.go
+++ b/pkg/api/bzzchunk.go
@@ -10,12 +10,16 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"strings"
 
 	"github.com/ethersphere/bee/pkg/jsonhttp"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/gorilla/mux"
 )
+
+// Presence of this header in the HTTP request indicates the chunk needs to be pinned
+const PinHeaderName = "x-swarm-pin"
 
 func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 	addr := mux.Vars(r)["addr"]
@@ -44,6 +48,18 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 		s.Logger.Error("bzz-chunk: chunk write error")
 		jsonhttp.BadRequest(w, "chunk write error")
 		return
+	}
+
+	// Check if this chunk needs to pinned and pin it
+	pinHeader := r.Header.Get(PinHeaderName)
+	if pinHeader != "" && strings.ToLower(pinHeader) == "true" {
+		err = s.Storer.Set(ctx, storage.ModeSetPin, address)
+		if err != nil {
+			s.Logger.Debugf("bzz-chunk: chunk pinning error: %v, addr %s", err, address)
+			s.Logger.Error("bzz-chunk: chunk pinning error")
+			jsonhttp.InternalServerError(w, "cannot pin chunk")
+			return
+		}
 	}
 
 	jsonhttp.OK(w, nil)
--- a/pkg/api/bzzchunk.go
+++ b/pkg/api/bzzchunk.go
@@ -18,7 +18,7 @@ import (
 	"github.com/gorilla/mux"
 )
 
-// Presence of this header in the HTTP request indicates the chunk needs to be pinned
+// Presence of this header in the HTTP request indicates the chunk needs to be pinned.
 const PinHeaderName = "x-swarm-pin"
 
 func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
@@ -51,8 +51,8 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Check if this chunk needs to pinned and pin it
-	pinHeader := r.Header.Get(PinHeaderName)
-	if pinHeader != "" && strings.ToLower(pinHeader) == "true" {
+	pinHeaderValues := r.Header.Get(PinHeaderName)
+	if pinHeaderValues != "" && strings.ToLower(pinHeaderValues) == "true" {
 		err = s.Storer.Set(ctx, storage.ModeSetPin, address)
 		if err != nil {
 			s.Logger.Debugf("bzz-chunk: chunk pinning error: %v, addr %s", err, address)
--- a/pkg/debugapi/pin.go
+++ b/pkg/debugapi/pin.go
@@ -21,14 +21,14 @@ import (
 func (s *server) pinChunk(w http.ResponseWriter, r *http.Request) {
 	addr, err := swarm.ParseHexAddress(mux.Vars(r)["address"])
 	if err != nil {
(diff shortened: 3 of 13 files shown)


jmozah:
TODO in future PRs
-  list pagination
- exclude pinned chunks in GC
- Add file support
- Add collection support

pkg/api/bzzchunk.go
```diff
@@ -10,13 +10,17 @@ import (
...
 	"net/http"
+	"strings"
 
 	"github.com/ethersphere/bee/pkg/jsonhttp"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/gorilla/mux"
 )
 
+// Presence of this header in the HTTP request indicates the chunk needs to be pinned
```
janos, line 21: End the sentence with a `.`.

.gitignore
```diff
@@ -1,6 +1,7 @@
 /dist
 /.idea
 /.vscode
+/vendor
```
janos, line 4: I would suggest to remove this change from this PR, and discuss if it is correct to be set in another place.

pkg/jsonhttp/jsonhttptest/jsonhttptest.go
```diff
@@ -11,12 +11,66 @@ import (
...
+	}
+	wantString := "[" + wantJson.Message + "]"
+
+	if wantString != string(got) {
+		t.Errorf("got response %s, want %s", string(got), wantString)
+	}
+}
+
+func ResponseDirectWithHeaders(t *testing.T, client *http.Client, method, url string, body io.Reader, responseCode int,
+	response interface{}, headers map[string]string) {
```
janos, line 70: There is a specific convenient type `http.Headers` that should be used to pass the headers instead of map.

pkg/debugapi/pin.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+// pinChunk pin's the already created chunk given its address.
+// it fails if the chunk is not present in the local store.
+// It also increments a pin counter to keep track of how many pin requests
+// are originating for this chunk.
+func (s *server) pinChunk(w http.ResponseWriter, r *http.Request) {
+	addr, err := swarm.ParseHexAddress(mux.Vars(r)["address"])
+	if err != nil {
+		s.Logger.Debugf("debug api: parse chunk address: %v", err)
```
janos, line 24: Log messages in this methods should mention the handler that they originate from, like `s.Logger.Debugf("debug api: pin chunk: parse chunk address: %v", err)`. This goes for all log messages in debug api.

pkg/debugapi/router.go
```diff
@@ -56,6 +56,14 @@ func (s *server) setupRouting() {
 	router.Handle("/chunks/{address}", jsonhttp.MethodHandler{
 		"GET": http.HandlerFunc(s.hasChunkHandler),
 	})
+	router.Handle("/chunks-pin/{address}", jsonhttp.MethodHandler{
+		"POST":   http.HandlerFunc(s.pinChunk),
+		"DELETE": http.HandlerFunc(s.unpinChunk),
+		"GET":    http.HandlerFunc(s.listPinnedChunks),
+	})
+	router.Handle("/chunks-pin/", jsonhttp.MethodHandler{
```
janos, line 64: There is no need for trailing slash. Currently it is required to be in the path, but should not be, should be removed.

pkg/storage/store.go
```diff
@@ -136,6 +142,8 @@ type Storer interface {
 	LastPullSubscriptionBinID(bin uint8) (id uint64, err error)
 	SubscribePull(ctx context.Context, bin uint8, since, until uint64) (c <-chan Descriptor, stop func())
 	SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, stop func())
+	GetPinnedChunks(ctx context.Context, cursor swarm.Address) (pinnedChunks []*PinInfo, err error)
+	GetPinInfo(address swarm.Address) (uint64, error)
```
janos, line 146: I would remove the `Get` prefix from these methods as not needed.

pkg/storage/mock/storer.go
```diff
@@ -6,28 +6,41 @@ package mock
 
 import (
 	"context"
+	"sync"
 
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
+type MockStorer interface {
```
janos, line 15: I am not sure if MockStorer interface is needed. Struct mockStore can be exported with all available methods that it provides. It just have to ensured that MockStore struct implements storage.Storer at compile time with `var _ storage.Storer = (*MockStorer)(nil)`. NewStorer would have to return a concrete type.

pkg/storage/mock/storer.go
```diff
@@ -6,28 +6,41 @@ package mock
...
+	storage.Storer
+	GetModeSet(addr swarm.Address) (mode storage.ModeSet)
+}
+
 type mockStorer struct {
 	store     map[string][]byte
+	modeSet   map[string]storage.ModeSet
+	modeSetMu *sync.Mutex
```
janos, line 23: mockStorer type is used with pointer semantics, so this mutex can be a value field.

pkg/debugapi/pin_test.go
```diff
@@ -0,0 +1,155 @@
...
+// TestPinChunkHandler checks for pinning, unpinning and listing of chunks.
+// It also check other edgw cases like chunk not present and checking for pinning,
+// invalid chunk address case etc. This test case has to be run in sequence and
+// it assumes some state of the DB before another case is run.
+func TestPinChunkHandler(t *testing.T) {
+	key := swarm.MustParseHexAddress("aabbcc")
+	value := []byte("data data data")
+	logger := logging.New(ioutil.Discard, 0)
+
+	storer, err := localstore.New("", key.Bytes(), nil, logger)
```
janos, line 30: Unit tests should test code in this package, which is the http layer and should preferably not relay on other components functionalities. A mock store should be used here instead, not the localstore, as that makes this test an integration test, which allowed new localstore pin methods not to be tested where they are implemented, in localstore package. Localstore should test for correctness of the code in it with its own tests and debugapi tests its own code.

.gitignore
jmozah: i can remove it.. but lets discuss and remove it. i don't understand why it is inside the code in the first place.

pkg/jsonhttp/jsonhttptest/jsonhttptest.go
jmozah: Thanks. The http.Header is of `type Header map[string][]string`. But when i do a `req.Header.Get(string`)  it returns only string...` not []string`.  strange na..

.gitignore
janos: We are not vendoring code in the bee project, just using modules. By having a vendor folder locally and ignored by git, it may present problems by using different code do run build or test, then the one defined in go.mod. If vendor directory is present in the project dir, go will use that code instead modules cache. This may result in different results when a developers runs code locally and on ci. If the vendor is not ignored it forces the developer to clean it if it is used for some experiment for example.

pkg/jsonhttp/jsonhttptest/jsonhttptest.go
janos: Yes, Header.Get returns the first value, but allows multiple values (properly encoded in the request), which can be accessed with Headers[headerName].

pkg/localstore/pin.go
```diff
@@ -0,0 +1,66 @@
...
+	it, err := db.pinIndex.First(prefix)
+	if err != nil {
+		return nil, fmt.Errorf("pin chunks: %w", err)
+	}
+	err = db.pinIndex.Iterate(func(item shed.Item) (stop bool, err error) {
+		pi := &storage.Pinner{
+			Address:    swarm.NewAddress(item.Address),
+			PinCounter: item.PinCounter,
+		}
+		pinnedChunks = append(pinnedChunks, pi)
```
pradovic, line 40:
```suggestion
		pinnedChunks = append(pinnedChunks, &storage.Pinner{
			Address:    swarm.NewAddress(item.Address),
			PinCounter: item.PinCounter,
		})
```

jmozah:
@pradovic Thanks for your comments.

Issue #56, opened 2020-04-03
Bootstrapping


Eknir:
# Epic

## Story

As a user of Swarm, I want to join the Swarm network and connect to a subset of the online nodes, chosen such that I can reach the whole network, such that I am fully connected efficiently and quick.

## Background

BOS:
```
As a node is being notified of new peer addresses, it stores them in a Kademlia table of known peers, also called \gloss{address book}. 

While it listens to incoming connections, it also proactively attempts to connect to nodes in order to achieve saturation: It 
(1) dials each known node that is within the \gloss{proximity order boundary} of $n$ \gloss{nearest neighbours} ($d$, \gloss{neighbourhood depth}) and 
(2) it tries to fill each bin up to $d$ with healthy peers (see figure \ref{fig:bootstrapping-kademlia}). 

To satisfy (1) most efficiently, it attempts to connect to the peer that is most needed at any point in time.
 Shallow (far) bins are more important to fill than deep (near) ones since they handle more volume. 
Filling an \gloss{empty bin} with one peer is more important than adding new peer to a non-empty bin, since it leads to a saturated Kademlia earlier. 
Therefore the protocol uses a \gloss{bottom-up  depth-first strategy} to choose a peer to connect to.   
Nodes that have been tried but failed to get connected are retried after a time interval that doubles after each attempt. 
After a certain number of attempts such nodes are no longer considered and eventually removed from the address book.
```
## Acceptance criteria
From the point of view of a new node, joining the network:
- Given a set of already running nodes, my node joins the network and establishes a connection with one node in every proximity order bin (see definition in BOS)

PR #188, opened 2020-05-15
[bee #47, #56] integrate kademlia instead of full driver

acud:
continued work on kademlia following #155.

this PR:
- implement `ClosestPeer` so that retrievals and push sync will work on kademlia
- improve debug api for topology, so that the actual kadmlia json string is a native JSON, not an escaped string
- hook up kademlia to be the default topology, instead of `full`
- clean up some of the connect logic in kademlia and pull it to a separate method, to clean up the `manage` loop logic
- add peer broadcasting functionality upon connecting to a peer
- add `Connected` hook on kademlia
- add peer announcement on dial-in

related to #47 #56 usw..

### DIFF
--- a/pkg/debugapi/topology.go
+++ b/pkg/debugapi/topology.go
@@ -5,15 +5,13 @@
 package debugapi
 
 import (
+	"bytes"
 	"encoding/json"
+	"io"
 	"net/http"
 
 	"github.com/ethersphere/bee/pkg/jsonhttp"
 )
-
-type topologyResponse struct {
-	Topology string `json:"topology"` //TODO: this is not so great since it makes the actual kademlia state to be an escaped string
-}
 
 func (s *server) topologyHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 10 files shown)


pkg/debugapi/topology.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 type topologyResponse struct {
-	Topology string `json:"topology"`
+	Topology string `json:"topology"` //TODO: this is not so great since it makes the actual kademlia state to be an escaped string
```
acud, line 15: @janos WDYT about this?

acud: see the change in https://github.com/ethersphere/bee/pull/188/commits/8f9c45951b96194f84a697ec39cf5e4624db37e0

janos: Yes, I actually did have a comment on this in my not submitted pr review to do something similar as in your changeset. I also think that it would work with just `jsonhttp.OK(w, ms)` and explicit call to MarshalJSON.

PR #189, opened 2020-05-15
Add join cli

nolash:
Adds a CLI command that retrieves a file from a Swarm Hash through Bee HTTP API.

### DIFF
--- a/cmd/bee-join/main.go
+++ b/cmd/bee-join/main.go
@@ -22,11 +22,11 @@ import (
 )
 
 var (
-	outFilePath string // flag variable, output file
-	outFileForce bool  // flag variable, overwrite output file if exists
-	host string // flag variable, http api host
-	port int    // flag variable, http api port
-	ssl  bool   // flag variable, uses https for api if set
+	outFilePath  string // flag variable, output file
+	outFileForce bool   // flag variable, overwrite output file if exists
+	host         string // flag variable, http api host
+	port         int    // flag variable, http api port
+	ssl          bool   // flag variable, uses https for api if set
 )
 
 // apiStore provies a storage.Getter that retrieves chunks from swarm through the HTTP chunk API.
@@ -90,7 +90,7 @@ func Join(cmd *cobra.Command, args []string) (err error) {
 			outFileFlags |= os.O_EXCL
 		}
 		// open the file
-		outFile, err= os.OpenFile(outFilePath, outFileFlags , 0o666)
+		outFile, err = os.OpenFile(outFilePath, outFileFlags, 0o666)
 		if err != nil {
 			return err
 		}
--- a/cmd/bee-join/main.go
+++ b/cmd/bee-join/main.go
@@ -90,7 +90,7 @@ func Join(cmd *cobra.Command, args []string) (err error) {
 			outFileFlags |= os.O_EXCL
 		}
 		// open the file
-		outFile, err = os.OpenFile(outFilePath, outFileFlags, 0o666)
+		outFile, err = os.OpenFile(outFilePath, outFileFlags, 0o666) // skipcq: GSC-G302
 		if err != nil {
 			return err
 		}
--- a/cmd/bee-join/main.go
+++ b/cmd/bee-join/main.go
@@ -7,7 +7,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


cmd/bee-join/main.go
```diff
@@ -0,0 +1,127 @@
+package main
```
acud, line 1: license

cmd/bee-join/main.go
```diff
@@ -0,0 +1,127 @@
...
+		baseUrl: u.String(),
+	}, nil
+}
+
+// Get implements storage.Getter
+func (a *apiStore) Get(ctx context.Context, mode storage.ModeGet, address swarm.Address) (ch swarm.Chunk, err error) {
+	addressHex := address.String()
+	url := strings.Join([]string{a.baseUrl, addressHex}, "/")
+	c := http.DefaultClient
+	res, err := c.Get(url)
```
acud, line 51: if you don't need to use `c` multiple times you can just `http.DefaultClient.Get`

cmd/bee-join/main.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, err
+	}
+	if res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("chunk %s not found", addressHex)
+	}
+	chunkData, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	err = res.Body.Close()
```
acud, line 62: do we care? i would just defer this after the error check you do after `get`

cmd/bee-join/main.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, err
+	}
+	ch = swarm.NewChunk(address, chunkData)
+	return ch, nil
+}
+
+// Join is the underlying procedure for the CLI command
+func Join(cmd *cobra.Command, args []string) (err error) {
+
+	outFile := os.Stdout
```
acud, line 73: you only use `stdout` here so why do you need the variable?
also, why not have the possibility of supplying an output filename? (then the variable will obviously be needed)

cmd/bee-join/main.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return nil, err
+	}
+	ch = swarm.NewChunk(address, chunkData)
+	return ch, nil
+}
+
+// Join is the underlying procedure for the CLI command
+func Join(cmd *cobra.Command, args []string) (err error) {
+
```
acud, line 72: i <3 whitespace

nolash:
> You're getting better at this

Monkey see, monkey do

cmd/bee-join/main.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	c.Flags().StringVarP(&outFilePath, "output-file", "o", "", "file to write output to")
+	c.Flags().BoolVarP(&outFileForce, "force", "f", false, "overwrite existing output file")
+	c.Flags().StringVar(&host, "host", "127.0.0.1", "api host")
+	c.Flags().IntVar(&port, "port", 8080, "api port")
+	c.Flags().BoolVar(&ssl, "ssl", false, "use ssl")
+
+	err := c.Execute()
+	if err != nil {
+		os.Exit(1)
```
acud, line 156: maybe print the error before exiting?

cmd/bee-join/main.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// create the join and get its data reader
+	j := joiner.NewSimpleJoiner(store)
+	r, l, err := j.Join(context.Background(), addr)
+	if err != nil {
+		return err
+	}
+
+	// join, rinse, repeat until done
+	teeReader := io.TeeReader(r, outFile)
```
acud, line 122: i kind of don't understand why do you need a `TeeReader` here. Isn't `tee` needed when you need to split the output into two different streams? but here it appears you read into two streams indeed, but one of them is never used (i.e. you never use `data`)

cmd/bee-join/main.go
nolash: It gets printed already when I run it with error. I think `cobra` takes care of it internally.

cmd/bee-join/main.go
nolash: > not sure if I see the benefit of having a tee here

it does exactly what you're describing, only without handling it directly.

cmd/bee-join/main.go
janos: I find too verbose for cobra to print usage when the error happens, so I would suggest to set `SilenceUsage: true` on the Command, if you agree, not required. Errors are still printed, just without the usage, while `-h` works correctly.

cmd/bee-join/main.go
acud: i think `Copy` won't work because the joiner needs to read at whole chunk-size intervals. but then in that case you could just `Read` as you do here and after the read explicitly `Write` to the `Writer`. `TeeReader` is still not needed. Please read the source code for what TeeReader does on Read

nolash: >> not sure if I see the benefit of having a tee here

> it does exactly what you're describing, only without handling it directly.

what's your argument exactly @acud? I don't get it.

acud: my argument is that you're reading in to the byte slice just in order to get the tee reader to `Write` into the writer, nothing else. then why don't you call `Write` explicitly from where you `Read` in the loop below and get rid of the tee reader?

nolash: Understood your point after a call.

PR #190, opened 2020-05-18
construct overlay address with network id

janos:
The core of this PR is to address the way how overlay address is constructed in respect to network id. It brings the implementation closer to the book `7.2.1 Overlay address` by including the network id in hashed data. We still do not have the ethereum address in bee, which is the difference from the book. We are constructing it from the local overlay private key.

- crypto.NewOverlayAddress updated to include network id
- changed network id to be uint64 type everywhere
- removed address from the keystore file encryptedKey as it changes with network id


pradovic:
LGTM!

PR #191, opened 2020-05-18
import interval store code from swarm

acud:
as a preparation step for commencing work on pull sync, we need to import the interval store code from old swarm.

test cases are not ideal since some of them inject some initial state into the interval by changing the field and not by constructor. @janos should we change this?


janos:
> test cases are not ideal since some of them inject some initial state into the interval by changing the field and not by constructor. @janos should we change this?

It can be improved later. I think that is better to only make functional changes when we move code from swarm. And in new commits (PRs) to add improvements.

PR #193, opened 2020-05-19
pull sync: initial version

acud:
### Review guide

_NOTE: This PR has been splitted into 2 commits for easier review process._

There are several components to be reviewed in this PR:
- `pullsync` protocol package - exposes APIs to synchronize given bins and intervals from a peer (implemented in first commit)
- `pullstorage` package - wraps a `localstore` instance to provide functionality to create batches from requested intervals in a bin (implemented in first commit)
- `puller` package - in an analogy to `pushsync` and `pusher`, I have created the `puller` component which is decoupled from `pullsync` package, and is tasked with syncing with peers. It receives notifications from the topology whenever a peer connects or disconnects, and similarly to the `manage` loop in kademlia that prioritizes peers to connect to, it prioritizes peers to start syncing with or to recalculate subscriptions with (when depth changes) (implemented in first commit)

My suggestion would be to start with reviewing the `pullsync` and `pullstorage` packages in the first commit as they are dependent upon each other (`pullstorage` is a `pullsync` dependency), then change to the second commit which introduces further changes to these packages (although pretty minor), and the `puller` component.

The PR tries to employ the principles that have been guiding us so far with `bee` development, thus mocks are heavily used across the packages. Inadvertently, since the real-world scenarios are quite complex (and ... deadlines) resulted in some of the mocks having quite some business logic in order to reproduce the desirable behavior needed for the test. This can be refactored later on, and I am aware of some of it being very verbose and with overlapping functionalities. I am also aware that it might not be entirely trivial for those who have not worked with this package in the old codebase to understand why a certain behavior is needed from the mock in a given test case. If this happens - please comment and ask in the PR or on mattermost.

This code has been thoroughly compared with all of the nuts and bolts in the stream package in `swarm` and it seems to cover the edge cases the we caught there (minus the localstore leakage fix that was reverted into the bee codebase @ #302).

One notable difference with the `swarm` stream implementation is that now peer initiation is made purely from changes in the connectivity. In `swarm`, each protocol is invoked with a `Run` function that bootstraps the protocol, and in the stream protocol the protocol would process each peer individually through the `Run` function. This causes all sorts of complexities in the protocol that must maintain a lot of state about peers. In the implementation presented in this PR peers are notified upon through the topology (i.e. whenever a change in peers occurs `puller` is notified, but does not get notified about what exactly happened but that a change has occurred, implying that peers should be recalculated). 

The depth notification from `kademlia` has been changed with a peer collection notification, which is more generalized but suits the needs of the protocol better. A simple example might be that a peer was added, but the depth did not change, and we would like to sync with that peer for whatever reason. Obviously this is needed since the abstraction does not contain the dreadful `Run` entry point.

---------

EDIT: Found bug where a peer that disconnects would not get removed correctly from `puller`. This is now fixed

---------
- [x] import bitvector from `swarm`
- [x] add db close channel to SubscribePull to signal db closed when iterator returns
- [x] implements only the dumb protocol that handles the message exchanges
- [x] implement `pullstorage` module that wraps `storage.Storer` and handles basic operations of getting, putting, setting chunks from the localstore in return for requests coming in from the protocol
- [x] implement puller
- [x] kademlia peer change subscriptions (done)
- [x] cursor exchange
- [x] handle peer slice change
- [x] handle depth changes
- [x] kademlia peer change subscription test disconnect trigger
- [x] integration testing with real localstore on pull sync
- [ ] falling edge kademlia channel wrapper -> defer to later pr
- [x] batch timeout test (for timer added code in pullstorage)
- [x] end to end context cancellation - currently only upstream peer quits but in fact there has to be some e2e closure of the stream so that downstream quits too, need to change to fullclose there
- [x] get rid of the localstore gc hack i introduced
- [x] integration testing
- [ ] metrics - later pr
- [ ] todo - check no goroutine/mem leaks
- [x] todo - peer that disconnects does not result in sync context removal from map. this can result in a mem leak and bugs when the peer gets reconnected


pradovic:
Looks good to me as well so far, no special comments. I will wait for ready for review, and go through it again once is ready for review. I agree about bitvector that it could be separate PR, but also I don't think this PR is huge, as @janos said, large part is auto-generated code, so it's seems fine!

acud:
thanks guys. I removed the deliberate panic in an earlier commit today. i now also removed the implicit methods of Has and Set. I left the Get that invokes GetMulti and the Put, but yeah we can also get rid of those too. I could maybe take care of it when the other PRs are merged into this one? just don't wanna deal too much with rebasing one on top of the other.

pkg/bitvector/bitvector.go
```diff
@@ -0,0 +1,124 @@
...
+var errInvalidLength = errors.New("invalid length")
+
+// BitVector is a convenience object for manipulating and representing bit vectors
+type BitVector struct {
+	len int
+	b   []byte
+}
+
+// New creates a new bit vector with the given length
+func New(l int) (bv *BitVector, err error) {
```
pradovic, line 32: 🌷 
```suggestion
func New(l int) (*BitVector, error) {
```

pkg/bitvector/bitvector.go
```diff
@@ -0,0 +1,124 @@
...
+	return NewFromBytes(make([]byte, l/8+1), l)
+}
+
+// NewFromBytes creates a bit vector from the passed byte slice.
+//
+// Leftmost bit in byte slice becomes leftmost bit in bit vector
+func NewFromBytes(b []byte, l int) (bv *BitVector, err error) {
```
pradovic, line 39: 🌷 
```suggestion
func NewFromBytes(b []byte, l int) (*BitVector, error) {
```

pkg/kademlia/kademlia.go
```diff
@@ -377,6 +399,44 @@ func (k *Kad) ClosestPeer(addr swarm.Address) (swarm.Address, error) {
...
+// SubscribePeersChange returns the channel that signals when the connected peers
+// set changes. Returned function is safe to be called multiple times.
+func (k *Kad) SubscribePeersChange() (c <-chan struct{}, unsubscribe func()) {
+	channel := make(chan struct{}, 1)
+	var closeOnce sync.Once
+
+	k.peerSigMtx.Lock()
+	defer k.peerSigMtx.Unlock()
+
+	k.peerSig = append(k.peerSig, channel)
```
pradovic, line 421: 🌷 Maybe it's a little bit nicer as the function below also locks, so someone might make a mistake while changing stuff.
```suggestion
	k.peerSigMtx.Lock()
	k.peerSig = append(k.peerSig, channel)
        k.peerSigMtx.Unlock()
```

pkg/pullsync/pullstorage/pullstorage.go
```diff
@@ -0,0 +1,147 @@
...
+		// is empty
+		topmost = to
+	}
+
+	return chs, topmost, nil
+}
+
+// Cursors gets the last BinID for every bin in the local storage
+func (s *ps) Cursors(ctx context.Context) (curs []uint64, err error) {
+	curs = make([]uint64, 16)
```
acud, line 127: this needs to be extracted (16)

acud, line 127: this needs to be extracted

pkg/kademlia/mock/kademlia.go
```diff
@@ -0,0 +1,175 @@
...
+	for _, c := range m.trigs {
+		select {
+		case c <- struct{}{}:
+		default:
+		}
+	}
+}
+
+func (m *Mock) Close() error {
+	panic("not implemented") // TODO: Implement
```
pradovic, line 167: I think that comment is not needed here as the string in panic is self-desribing

pkg/puller/puller.go
```diff
@@ -0,0 +1,447 @@
...
+						peerEntry := peer{addr: peerAddr, po: po}
+						peersToRecalc = append(peersToRecalc, peerEntry)
+					}
+				}
+
+				return false, false, nil
+			})
+
+			if err != nil {
+				panic(err)
```
pradovic, line 141: Why panic here?

pkg/pullsync/pullstorage/pullstorage_test.go
```diff
@@ -0,0 +1,376 @@
...
+var (
+	addrs = []swarm.Address{
+		swarm.MustParseHexAddress("0001"),
+		swarm.MustParseHexAddress("0002"),
+		swarm.MustParseHexAddress("0003"),
+		swarm.MustParseHexAddress("0004"),
+		swarm.MustParseHexAddress("0005"),
+		swarm.MustParseHexAddress("0006"),
+	}
+	limit = 5
```
acud, line 33: the limit here at the tests is injected through the function call. however, pull sync protocol enforces its own limit which is different (right now set at 50), some of the test cases around the pull sync protocol are also checking around the limit (of 50 - `pullsync.go#L37`) and so this is a bit of a debt that needs to be settled. up to you guys if this should be during this PR or not

pkg/pullsync/pullsync.go
```diff
@@ -0,0 +1,333 @@
...
+		if err != nil {
+			return 0, fmt.Errorf("storage has: %w", err)
+		}
+		if !have {
+			wantChunks[a.String()] = struct{}{}
+			ctr++
+			bv.Set(i / swarm.HashSize)
+		}
+	}
+	fmt.Println("sending want", len(bv.Bytes()))
```
pradovic, line 152: loggger or maybe debug leftover?

pkg/puller/puller.go
```diff
@@ -0,0 +1,449 @@
...
+		default:
+		}
+
+		s, _, _, err := p.nextPeerInterval(peer, bin)
+		if err != nil {
+			p.logger.Debugf("histSyncWorker nextPeerInterval: %v", err)
+			// wait and retry? this is a local error
+			// maybe just quit the peer entirely.
+			// not sure how to do this
+			<-time.After(30 * time.Second)
```
acud, line 313: probably log error and return here. not sure.

pkg/puller/puller.go
```diff
@@ -0,0 +1,449 @@
...
+		default:
+		}
+		p.logger.Tracef("liveSyncWorker peer %s syncing bin %d from %d", peer, bin, from)
+		top, err := p.syncer.SyncInterval(ctx, peer, bin, from, math.MaxUint64)
+		if err != nil {
+			p.logger.Debugf("liveSyncWorker exit on sync error. peer %s bin %d from %d err %v", peer, bin, from, err)
+			return
+		}
+		if top == 0 {
+			return //TODO need to deal with this somehow. not right
```
acud, line 357: this is not legal. should not happen. probably log and return.

pkg/kademlia/kademlia.go
acud: i don't think this is too confusing tbh and i prefer the `defer` statement

pkg/kademlia/mock/kademlia.go
acud: this is auto generated text from `:GoImpl` (`vim-go`)

pkg/pullsync/pullsync.go
```diff
@@ -0,0 +1,332 @@
...
+	if err != nil {
+		return err
+	}
+
+	time.Sleep(50 * time.Millisecond) //because of test, getting EOF w/o
+	return nil
+}
+
+func (s *Syncer) setChunks(ctx context.Context, addrs ...swarm.Address) error {
+	return s.storage.Set(ctx, storage.ModeSetSyncPull, addrs...)
```
zelig, line 244: I dont thin we need ModeSetSyncPull any more. it was legacy

pkg/localstore/subscription_pull.go
```diff
@@ -36,7 +36,7 @@ import (
 // function will terminate current and further iterations without errors, and also close the returned channel.
 // Make sure that you check the second returned parameter from the channel to stop iteration when its value
 // is false.
-func (db *DB) SubscribePull(ctx context.Context, bin uint8, since, until uint64) (c <-chan storage.Descriptor, stop func()) {
+func (db *DB) SubscribePull(ctx context.Context, bin uint8, since, until uint64) (c <-chan storage.Descriptor, closed <-chan struct{}, stop func()) {
```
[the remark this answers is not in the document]
acud: theoretically yes. in the old codebase for some reason though this wasn't so for some reason and you could get some empty batch on shutdown. the problem is also that `SubscribePull` does not provide an explicit signal about shutdown (before this change). I rather be on the safe side for now. But theoretically yes you are right and it could be that the entire section of code here which is handling shutdown could be removed :+1: Need to test this though to verify the assumption.

pkg/puller/puller.go
```diff
@@ -0,0 +1,501 @@
...
+func New(o Options) *Puller {
+	p := &Puller{
+		statestore: o.StateStore,
+		topology:   o.Topology,
+		syncer:     o.PullSync,
+		logger:     o.Logger,
+
+		cursors: make(map[string][]uint64),
+
+		syncPeers: make([]map[string]*syncPeer, bins),
```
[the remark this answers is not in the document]
acud: well not entirely. you need to maintain the contexts of which peers you sync with on which bins. I wouldn't like to overload this information in kademlia

pkg/puller/puller.go
```diff
@@ -0,0 +1,501 @@
...
+	"github.com/ethersphere/bee/pkg/intervalstore"
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/pullsync"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/ethersphere/bee/pkg/topology"
+)
+
+var (
+	bins = uint8(16)
```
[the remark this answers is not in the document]
acud: well actually not a const, since i'm altering this value in tests (to make the tests more readable), it has to be a `var` in order to do so. but it could be that it would just be nicer to inject this value through the constructor. there's a few other things that could follow

pkg/puller/puller.go
```diff
@@ -0,0 +1,501 @@
...
+			// way that it returns an error - the value must be checked.
+			_ = p.topology.EachPeerRev(func(peerAddr swarm.Address, po uint8) (stop, jumpToNext bool, err error) {
+				bp := p.syncPeers[po]
+				if _, ok := bp[peerAddr.String()]; ok {
+					delete(peersDisconnected, peerAddr.String())
+				}
+				syncing := len(bp)
+				switch {
+				case po < depth:
+					// outside of depth, sync peerPO bin only
```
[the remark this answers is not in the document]
acud: i remember about not syncing bin `0`. we can disable this of course.

pkg/puller/puller.go
```diff
@@ -0,0 +1,501 @@
...
+		syncer:     o.PullSync,
+		logger:     o.Logger,
+
+		cursors: make(map[string][]uint64),
+
+		syncPeers: make([]map[string]*syncPeer, bins),
+		quit:      make(chan struct{}),
+	}
+
+	for i := uint8(0); i < bins; i++ {
```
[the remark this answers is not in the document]
acud: not sure i understand what you mean

pkg/puller/puller.go
```diff
@@ -0,0 +1,501 @@
...
+				// we have sync running on this bin, cancel it
+				c()
+				delete(syncCtx.binCancelFuncs, bin)
+			}
+		}
+	} else {
+		// outside of depth
+		var (
+			want     = po
+			dontWant = []uint8{0} // never want bin 0
```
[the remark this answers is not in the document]
acud: because we said we never want to pull sync on bin `0`

pkg/pullsync/pullsync.go
acud: i dont remember this was defined anywhere. can we iterate on this?

pkg/puller/puller.go
```diff
@@ -0,0 +1,501 @@
...
+		cursors, err := p.syncer.GetCursors(ctx, peer)
+		if err != nil {
+			p.logger.Errorf("error getting cursors: %v", err)
+			delete(p.syncPeers[po], peer.String())
+			return
+			// remove from syncing peers list, trigger channel to find some other peer
+			// maybe blacklist for some time
+		}
+		p.cursorsMtx.Lock()
+		p.cursors[peer.String()] = cursors
```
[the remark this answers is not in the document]
acud: i dont understand. you need to keep a copy of the cursors in case you'd like to start syncing on some bins later on (if the depth moves back and forth). you don't want to ask the cursors if you already asked for them before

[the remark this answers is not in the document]
zelig: are they stored in db and here you just  cache them after query? then ok

pkg/puller/puller.go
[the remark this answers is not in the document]
zelig: this is not so,, i remember clearly  our call when we went through this. lets call

pkg/puller/puller.go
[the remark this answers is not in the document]
zelig: u  get clleaner code if you  hide the particulars of syncCtx and just call a function on it

pkg/puller/puller.go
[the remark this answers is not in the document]
zelig: yes please apply the minimal  design we so successfully kept so far. pullsync needed now for  only neighbourhood replication and adjustment (no need for filling in storage functionality now) or trickle chunks we got pushsync protocol

pkg/puller/puller.go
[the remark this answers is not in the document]
zelig: you prefer replicating the container.  ok

pkg/puller/puller.go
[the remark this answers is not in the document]
acud: no they are only stored in memory (in the map...)

pkg/puller/puller.go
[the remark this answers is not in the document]
acud: i'd like to be leave the flexibility to maintain many connections in kademlia, but not sync to all of them, so maintaining a subset of the state is mandatory in this case (or flagging them somehow in kademlia, which we agreed we shouldn't do)

Issue #178, opened 2020-05-11
Signature in handshake protocol


janos:
Add Signature in handshake protocol as:
```
message Syn {
    BzzAddress BzzAddress = 1;
    int32 NetworkID = 2;
    bool Light = 3;
}

message Ack {
    BzzAddress BzzAddress = 1;
}

message BzzAddress {
    bytes Underlay = 1;
    bytes Signature = 2;
    bytes Overlay = 3;
}
```

where 
`Signature=Sign(underlay|bzzNetworkID)`
and it is signed with pubkey of overlay.

PR #196, opened 2020-05-19
Signing in handshake

pradovic:
Implemented signing in `crypto` and handshake as per book and in https://github.com/ethersphere/bee/issues/178. 
Tested locally and it seems to work.

TODO:
1. Hive msgs


pkg/p2p/libp2p/internal/handshake/handshake_test.go
```diff
@@ -382,15 +488,46 @@ func TestHandle(t *testing.T) {
...
+			Underlay:  got.Syn.BzzAddress.Underlay,
 			NetworkID: got.Syn.NetworkID,
 			Light:     got.Syn.Light,
 		})
 
-		_, err = handshakeService.Handle(stream1, info.ID)
-		if err != ErrHandshakeDuplicate {
-			t.Fatalf("expected %s err, got %s err", ErrHandshakeDuplicate, err)
+		_, err = handshakeService.Handle(stream1, node2AddrInfo.ID)
+		if err == nil || err.Error() != "handshake duplicate" {
```
janos, line 499: Why is ErrHandshakeDuplicate removed? It was better to check for a concrete error type then for an error string.

pkg/p2p/libp2p/internal/handshake/handshake_test.go
```diff
@@ -164,37 +224,80 @@ func TestHandshake(t *testing.T) {
 			t.Fatal("res should be nil")
 		}
 
-		if err != ErrNetworkIDIncompatible {
-			t.Fatalf("expected %s, got %s", ErrNetworkIDIncompatible, err)
+		if err == nil || err.Error() != "invalid ack received" {
```
janos, line 227: Also here, it is better to check for the concrete error, then for the error message string.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -55,19 +56,16 @@ type Service struct {
...
 	DisableQUIC bool
-	NetworkID   uint64
 	Addressbook addressbook.Putter
 	Logger      logging.Logger
 	Tracer      *tracing.Tracer
 }
 
-func New(ctx context.Context, o Options) (*Service, error) {
-	host, port, err := net.SplitHostPort(o.Addr)
+func New(ctx context.Context, privateKey *ecdsa.PrivateKey, networkID uint64, overlay swarm.Address, addr string,
```
janos, line 66: Actually, private key is optional, it can be ephemeral in libp2p. I would put this back to Options, with the check if it is nil a few lines below.

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -50,20 +47,33 @@ type Service struct {
 	network.Notifiee // handhsake service can be the receiver for network.Notify
 }
 
-func New(overlay swarm.Address, networkID uint64, logger logging.Logger) *Service {
+func New(overlay swarm.Address, underlay []byte, signer crypto.SignRecoverer, networkID uint64, logger logging.Logger) (*Service, error) {
```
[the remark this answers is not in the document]
pradovic: Hmm I agree, I would actually even use underlay as a string in proto messages as well, I don't see the reason for it to be a byte array. I will change to what u suggested either way

pkg/p2p/libp2p/internal/handshake/handshake_test.go
pradovic: I removed those specific errors because they were used only in tests to compare, and in libp2p to log a warning which, now when I look at it, does not seem necessary, so I thought maybe removing some code is nicer. Although I agree that specific error is a better approach in general. Basically, in libp2p we are logging those anyways and r not doing anything specific. WDYT?

pkg/p2p/libp2p/libp2p.go
pradovic: Hmm yes, my bad, I had trouble getting this key from libp2p in that case, and I forgot to tackle it later 🙈 nice catch!

pkg/p2p/libp2p/internal/handshake/handshake_test.go
janos: Yes, we are suing them in tests only, but even that is justifiable, I think.

pradovic: Ok, I'll switch back to a specific errors them

pkg/p2p/libp2p/libp2p.go
pradovic: So,k I took another look at it. It seems that if the private key is not provided as one of the libp2p options, libp2p will use a default random identity then. This implementation seems to create a RSA key pair. What would you suggest in this case? I don't think those are interchangeable. Should we create a random private key in this case? Maybe I am missing something here, LTMWYT :)

pkg/p2p/libp2p/internal/handshake/handshake.go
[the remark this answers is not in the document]
pradovic: Done. I left underlay as bytearray as per spec and the book. We can talk about it with Viktor and I can change it to `string` if accepted. It can go well with the Hive msg changes PR.

pkg/p2p/libp2p/libp2p.go
janos: Yes, that is correct and I did not want to restrict this possibility, as libp2p is not required to have an identity option to work. Anyways, this is out of scope of this PR.

pradovic: So, you are OK with leaving it as required field for now?

janos: Well, no. I think that it should be in Options.

pradovic: Sorry, maybe I was not clear enough. What I am trying to say is, what key should I use for signing in this case, when private key is nil? If we don't specify the key to libp2p, it will create it's own random RSA key pair, which is of different type then what we are using in Signer, and might not be interchangeable.

pradovic: Ah ok, sorry, I confused swarm key with libp2p key 🙈 Should be fixed now!

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -74,21 +98,25 @@ func (s *Service) Handshake(stream p2p.Stream) (i *Info, err error) {
...
 
 	if err := w.WriteMsgWithTimeout(messageTimeout, &pb.Ack{
-		Address: resp.Syn.Address,
+		BzzAddress: resp.Syn.BzzAddress,
 	}); err != nil {
 		return nil, fmt.Errorf("write ack message: %w", err)
 	}
 
-	s.logger.Tracef("handshake finished for peer %s", address)
+	s.logger.Tracef("handshake finished for peer %s", resp.Syn.BzzAddress.Overlay)
```
janos, line 115: Overlay is []byte and I think that it will not be nicely printed with %s. Maybe to create an overlay variable and to use it here and in the returned Info?

pradovic: Sure, make sense!

pkg/crypto/signer.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// Recover verifies signature with the data base provided.
+// Is exported so it can be used in other `Recoverer` implementations as well.
+func Recover(signature, data []byte) (*ecdsa.PublicKey, error) {
+	p, _, err := btcec.RecoverCompact(btcec.S256(), signature, data)
+	return (*ecdsa.PublicKey)(p), err
+}
+
+type SignRecoverer interface {
```
ralph-pichler, line 29: Still don't like this merged interface. To me this makes the impression that the `Recover` operation depends on the signer you yourself use when really it depends on the signer the other peer uses (which we cannot know and which really should produce the signature in the same format as the `defaultSigner` anyway).

I think `Recoverer` and `Signer` should be passed to the service separately and there should only be a single implementation of `Recoverer` (at least as long as no information on the signature format is passed as part of the handshake).

pradovic: I think in this case this is only because of the code, not to use 2 components. So you have basically 3 options, to use `Signer`, `Recoverer`, and `SignRecoverer` if you need both, but there is nothing stopping you to implement them however you want. This is more like a helper abstraction for the code. There is no dependency between `Signer` and `Recoverer` imo, it's just extracted in the interface. LTMWYT

pradovic: Well, if we are not going to have multiple recover functions, I can completely delete `Recoverer` interface and use `crypto.Recover()` function directly in the code as well. I don't mind, no need to make abstraction if we don't need them and the function is small enough to directly use in unit tests as well. @janos what do you say as well? This way we will only have `Signer` interface

pkg/crypto/signer_test.go
```diff
@@ -0,0 +1,52 @@
...
+		if err != nil {
+			t.Fatal()
+		}
+
+		if pubKey.X.Cmp(privKey.PublicKey.X) != 0 || pubKey.Y.Cmp(privKey.PublicKey.Y) != 0 {
+			t.Fatalf("wanted %v but got %v", pubKey, &privKey.PublicKey)
+		}
+	})
+
+	t.Run("sign & recover with invalid signature", func(t *testing.T) {
```
ralph-pichler, line 37: Technically this tests that the public key returned from a valid signature is not equal to another random public key. There are also really invalid signatures (where there is no associated public key and `.Recover` would return an error) and the description makes it sound like this is being tested.

pkg/crypto/signer.go
janos: We can do that as we only use one elliptic curve. Does it make sense to make it explicit than like `RevoverSecp256k1`?

ralph-pichler: If we put it into the name it should probably be more explicit about the signature format. There are many different formats for `secp256k1` based signatures. This recover function would be specifically for compact signatures.

One case where `Recoverer` might make sense is if we allow different schemes / formats for the same curve. E.g. you could let your peer know that you sign data with the "Ethereum Signed Message"-Prefix and then it would use a different `Recoverer` for you. This would allow things like `clef` for external signing also for overlay but would require transmitting this information in the handshake.

pradovic: I would go for a single `crypto.Recover()` then, as this way it is not self-explanatory, so one who is using it will have to take a look at the code, but I don't see a way to overcome this. At least, it is obvious there is a single `Recover` function. But I think we are kind of over-engineering it for now, if we stumble upon different `Recover` functions, we can change the name, add the interface, do whatever we like :) 
I thought that we said we won't allow different formats nor curves. Actually, in the beginning of conversions, I used format and later we switched to curve :) WDYGT?

pkg/crypto/signer_test.go
pradovic: Yup, I thought about it but decided not to test those cases, I can actually add them as well, or change the signature, we'll do one of those :)

pradovic: Fixed

pkg/crypto/signer.go
```diff
@@ -0,0 +1,41 @@
...
+	key *ecdsa.PrivateKey
+}
+
+func NewDefaultSigner(key *ecdsa.PrivateKey) Signer {
+	return &defaultSigner{
+		key: key,
+	}
+}
+
+func (d *defaultSigner) PublicKey() (*ecdsa.PublicKey, error) {
```
acud, line 35: do you need the error in the return value? it is always nil

pkg/crypto/signer_test.go
```diff
@@ -0,0 +1,47 @@
...
+	signer := crypto.NewDefaultSigner(privKey)
+	signature, err := signer.Sign(testBytes)
+	if err != nil {
+		t.Fatal()
+	}
+
+	t.Run("OK - sign & recover", func(t *testing.T) {
+		pubKey, err := crypto.Recover(signature, testBytes)
+		if err != nil {
+			t.Fatal()
```
acud, line 29: `t.Fatal(err)`

pkg/crypto/signer_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"testing"
+
+	"github.com/ethersphere/bee/pkg/crypto"
+)
+
+func TestDefaultSigner(t *testing.T) {
+	testBytes := []byte("test string")
+	privKey, err := crypto.GenerateSecp256k1Key()
+	if err != nil {
+		t.Fatal()
```
acud, line 17: `t.Fatal(err)`

pkg/crypto/signer_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	signer := crypto.NewDefaultSigner(privKey)
+	signature, err := signer.Sign(testBytes)
+	if err != nil {
+		t.Fatal()
```
acud, line 23: `t.Fatal(err)`

pkg/crypto/signer_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+		if pubKey.X.Cmp(privKey.PublicKey.X) != 0 || pubKey.Y.Cmp(privKey.PublicKey.Y) != 0 {
+			t.Fatalf("wanted %v but got %v", pubKey, &privKey.PublicKey)
+		}
+	})
+
+	t.Run("OK - recover with invalid data", func(t *testing.T) {
+		pubKey, err := crypto.Recover(signature, []byte("invalid"))
+		if err != nil {
+			t.Fatal()
```
acud, line 40: `t.Fatal(err)`

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -24,16 +27,22 @@ const (
 	ProtocolName    = "handshake"
 	ProtocolVersion = "1.0.0"
 	StreamName      = "handshake"
-	messageTimeout  = 5 * time.Second // maximum allowed time for a message to be read or written.
+	messageTimeout  = 10 * time.Second // maximum allowed time for a message to be read or written.
```
acud, line 30: is this increase really needed?

pkg/p2p/libp2p/internal/handshake/handshake_test.go
```diff
@@ -382,15 +520,80 @@ func TestHandle(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		var buffer1 bytes.Buffer
+		var buffer2 bytes.Buffer
+		stream1 := mock.NewStream(&buffer1, &buffer2)
+		stream2 := mock.NewStream(&buffer2, &buffer1)
+
+		w, _ := protobuf.NewWriterAndReader(stream2)
```
acud, line 547: `w := protofbuf.NewWriter()`

pkg/p2p/libp2p/internal/handshake/handshake_test.go
```diff
@@ -382,15 +520,80 @@ func TestHandle(t *testing.T) {
...
```
acud, line 577: ditto

pkg/p2p/libp2p/internal/handshake/handshake_test.go
```diff
@@ -382,15 +520,80 @@ func TestHandle(t *testing.T) {
...
+		}); err != nil {
+			t.Fatal(err)
+		}
+
+		res, err := handshakeService.Handle(stream1, node2AddrInfo.ID)
+		if res != nil {
+			t.Fatal("res should be nil")
+		}
+
+		if err != ErrInvalidSignature {
```
acud, line 595: this made me noticed that this test is not namespaced under `handshake_test` but just `handshake`. I guess we'd like to change that @janos?

pkg/crypto/signer.go
pradovic: The interface was agreed with @ralph-pichler, as he will implement external signer. Default implementation does not need it, but I guess others might

pkg/p2p/libp2p/internal/handshake/handshake.go
pradovic: I was testing locally during debugging, so maybe not, I will switch it back and increase if needed.

pkg/p2p/libp2p/internal/handshake/handshake_test.go
pradovic: Fixed :)

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -144,8 +180,37 @@ func (s *Service) Disconnected(_ network.Network, c network.Conn) {
...
+	if syn.NetworkID != s.networkID {
+		return ErrNetworkIDIncompatible
+	}
+
+	recoveredPK, err := crypto.Recover(syn.BzzAddress.Signature, append(syn.BzzAddress.Underlay, strconv.FormatUint(syn.NetworkID, 10)...))
+	if err != nil {
+		return ErrInvalidSignature
+	}
+
+	recoveredOverlay := crypto.NewOverlayAddress(*recoveredPK, syn.NetworkID)
```
zelig, line 193: I checked this package and found that it is not conforming to spec, as the overlay is the hash of the eth address + network id not public key +. network iD

pkg/p2p/libp2p/libp2p.go
```diff
@@ -210,7 +206,7 @@ func New(ctx context.Context, o Options) (*Service, error) {
 			return
 		}
 
-		err = s.addrssbook.Put(i.Address, remoteMultiaddr)
+		err = s.addrssbook.Put(i.Overlay, remoteMultiaddr)
```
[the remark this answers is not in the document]
pradovic: Yup, I'm on it!

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -50,20 +62,32 @@ type Service struct {
 	network.Notifiee // handhsake service can be the receiver for network.Notify
 }
 
-func New(overlay swarm.Address, networkID uint64, logger logging.Logger) *Service {
+func New(overlay swarm.Address, underlay string, signer crypto.Signer, networkID uint64, logger logging.Logger) (*Service, error) {
+	signature, err := signer.Sign([]byte(underlay + strconv.FormatUint(networkID, 10)))
```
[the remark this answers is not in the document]
pradovic: I suppose it should actually be little endian as in the book?

PR #199, opened 2020-05-21
Tagging chunks

jmozah:
refer https://app.zenhub.com/workspaces/swarm-10-5e7c6d9966b97049231c3c0b/issues/ethersphere/bee/51

Everything under "Chunk upload" will be done in this PR


jmozah:
@janos The deepsource error is occuring again... looks like a false +ve

pkg/storage/mock/storer.go
```diff
@@ -60,7 +63,7 @@ func (m *MockStorer) Put(ctx context.Context, mode storage.ModePut, chs ...swarm
 		}
 		m.store[ch.Address().String()] = ch.Data()
 	}
-	return nil, nil
+	return []bool{true}, nil
```
acud, line 66: this is incorrect. the size of this slice should be the same length as `len(chs)`

pkg/api/bzzchunk.go
```diff
@@ -33,6 +39,18 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// create the Tag for the chunk
+	tagName := r.Header.Get(TagHeaderName)
+	if tagName == "" {
+		tagName = fmt.Sprintf("unnamed_tag_%d", time.Now().Unix())
+	}
+	tag, err := s.Tags.Create(tagName, 1, false)
```
acud, line 47: this is a bit tricky and we might need to change the API here. here's why:
- right now tags.Create creates a random number for a tag
- this means that if you need to upload several chunks that correlate to the same upload you cannot do so, since even if you provide the same tag name, you will get a new tag for every new chunk you upload, so you cannot keep track and associate multiple chunks with the same tag
- this might work for an API that accepts arbitrary length input, but not for a single chunk API

off the top of my head, i can think about 2 different possibilities of calling the single chunk api:
1. first chunk in a series of correlating chunks, we still don't have a tag at this point so: we call the api with the `x-swarm-tag` header, and create a new tag, this tag ID that was created should be returned in a header or the response somehow
2. chunk # 2 or more in the above sequence, we now have a tag ID and we would like to associate this chunk with the tag, so we could `POST` to the api with a separate `x-swarm-tag-id` header containing the actual tag, then in the handler we can do the appropriate increments to the tag. `x-swarm-tag` header in this case _should_ be ignored and is unnecessary since the tag already exists

WDYT about this? does this make sense?

jmozah: First of all ..this is meant for only single chunks, not a series of chunks that needs to be tagged ( i mean the  scope of the PR). If someone wants to send a series of chunks and tag them with same id or something... i would check if that hash a UUID already and not create tag, instead pull the tag from store.


- From the external application point of view, it should send first chunk, get the response, pick the UUID and use that in every subsequent chunks.

acud: well currently the `id` is just a string meant to identify the entire chain of chunks in a tag (in old swarm codebase, the functionality covers multiple chunks within one tag). so it was created in the sense that you could tag an upload like `"upload of directory ./documents"` (or some other general string identifier used to identify the tag from another L2 application that uses swarm as of such).
So but now you're saying that the `id` could also be a general string or a uuid? not sure if I'm following

acud: please note that right now the general `id` string is not enforcing any unique constraints in `tags`. I think mixing these things up using the same field will create confusion and possible bugs

jmozah: @acud I am no0t sure i follow you. I dont see any issues. now you have one chunk one tag because we can uppload one chunk unlike previous swarm code.

pkg/api/tag_test.go
```diff
@@ -0,0 +1,215 @@
...
+				_, err := strconv.ParseUint(v[0], 10, 32)
+				if err != nil {
+					t.Fatal(err)
+				}
+			}
+		}
+
+		if !found {
+			t.Fatalf("could not find tagid header in response")
+		}
```
janos, line 64: Can this be replaced with `if rcvdHeaders.Get(api.TagHeaderName) != uid { t.Fatalf(...) }`?

pkg/api/tag_test.go
```diff
@@ -0,0 +1,215 @@
...
+		mockPusher           = mp.NewMockPusher(tag)
+		client               = newTestServer(t, testServerOptions{
+			Storer: mockValidatingStorer,
+			Tags:   tag,
+		})
+	)
+
+	t.Run("tag-header-in-return", func(t *testing.T) {
+		headers := make(map[string][]string)
+		headers[api.TagHeaderName] = []string{"2341312131"} // the value doesn't matter, it is random so we cant test for that
```
janos, line 45: Simplify with http.Header.Set method?

pkg/pusher/pusher.go
```diff
@@ -79,9 +83,16 @@ func (s *Service) chunksWorker() {
 			chunksInBatch++
 			s.metrics.TotalChunksToBeSentCounter.Inc()
 
+			t, err := s.tag.GetByAddress(ch.Address())
+			if err != nil {
+				s.logger.Errorf("pusher: error while getting tag: %v", err)
```
janos, line 88: This log line should be of Debug severity as it exposes the error.
```suggestion
				s.logger.Debugf("pusher: get tag by address %s: %v", ch.Address(), err)
```

pkg/pusher/pusher.go
```diff
@@ -123,6 +134,11 @@ func (s *Service) setChunkAsSynced(ctx context.Context, addr swarm.Address) {
 		s.metrics.ErrorSettingChunkToSynced.Inc()
 	} else {
 		s.metrics.TotalChunksSynced.Inc()
+		ta, err := s.tag.GetByAddress(addr)
+		if err != nil {
+			s.logger.Errorf("pusher: error setting tag to synced: %v", err)
```
janos, line 139: This log line should be of Debug severity as it exposes the error.

This log line is not correct as the error is by GetByAddress, not setting the tag as synced (which is on the line below).

```suggestion
			s.logger.Debugf("pusher: get tag by address %s: %v", ch.Address(), err)
```

pkg/storage/mock/storer.go
janos: Yes, this should be corrected.

pkg/api/bzzchunk.go
```diff
@@ -7,17 +7,23 @@ package api
 import (
 	"bytes"
 	"errors"
+	"fmt"
+	"github.com/ethersphere/bee/pkg/tags"
```
janos, line 11: Somehow this import is among the stdlib's packages. Not so important, but it would be nice to have it a few lines below within the bee imports. Also noticed in some other places in this pr...

pkg/api/bzzchunk.go
janos: This question is on what functionality we want to have with chunk tags. As @acud, I also had the impression that we would like to be able have chunks grouped by the common tag for them so that their progress can be followed. We would need to increment Total in Tag when a new chunk is uploaded with the same tag name.

jmozah: @janos tag name or tag ID

janos: It should be able to be identified by header `x-swarm-tag`, so I suppose: tag name.

jmozah: i changed it to x-swarm-tag-uid to be specific.  It was very confusing as we have a name variable also in tag, but that is just for name sake :-)

janos: Thanks, that is fine.

pkg/api/bzzchunk.go
```diff
@@ -33,6 +40,37 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+			s.Logger.Debugf("bzz-chunk: parse taguid %s: %v", tagUidStr, err)
+			s.Logger.Error("bzz-chunk: error uploading chunk")
+			jsonhttp.BadRequest(w, "invalid taguid")
+			return
+		}
+
+		tag, err = s.Tags.Get(uint32(tagUid))
+		if err != nil {
+			s.Logger.Debugf("bzz-chunk: tag creation error: %v, addr %s", err, address)
+			s.Logger.Error("bzz-chunk: tag creation error")
```
janos, line 68: `tag creation error` -> `tag get error` in both lines

pkg/api/tag.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	tag, err := s.Tags.Create(tagName, 1, false)
+	if err != nil {
+		s.Logger.Debugf("bzz-chunk: tag creation error: %v", err)
+		s.Logger.Error("bzz-chunk: tag creation error")
+		jsonhttp.InternalServerError(w, "cannot create tag")
+		return
+	}
+	w.Header().Set("Content-Type", jsonhttp.DefaultContentTypeHeader)
```
janos, line 32: Content-Type is set by jsonhttp.OK, this line ken be removed.

pkg/api/tag.go
```diff
@@ -0,0 +1,82 @@
...
+func (s *server) getANewTag(w http.ResponseWriter, r *http.Request) {
+	tagName := mux.Vars(r)["name"]
+	if tagName == "" {
+		s.Logger.Debugf("bzz-tag: invalid tag name : %s", tagName)
+		s.Logger.Error("bzz-tag: invalid tag name")
+		jsonhttp.BadRequest(w, "invalid tag name")
+		return
+	}
+
+	tag, err := s.Tags.Create(tagName, 1, false)
```
janos, line 25: Should here be total set to 0, as no chunks are uploaded still? And when the chunk is uploaded to increment Total state in Tag.

pkg/api/tag.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		s.Logger.Debugf("bzz-chunk: tag creation error: %v", err)
+		s.Logger.Error("bzz-chunk: tag creation error")
+		jsonhttp.InternalServerError(w, "cannot create tag")
+		return
+	}
+	w.Header().Set("Content-Type", jsonhttp.DefaultContentTypeHeader)
+	w.Header().Set("Cache-Control", "no-cache, private, max-age=0")
+	jsonhttp.OK(w, tag)
```
janos, line 34: I think that the response here should be defined in this package and constructed from the tag, even if ti has the same fields. This would give clarity what is served by the api.

pkg/api/tag.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		s.Logger.Debugf("bzz-tag: tag not present %s : %v, ", address.String(), err)
+		s.Logger.Error("bzz-tag: tag not present")
+		jsonhttp.InternalServerError(w, "tag not present")
+		return
+	}
+
+	w.Header().Set("Content-Type", jsonhttp.DefaultContentTypeHeader)
+	w.Header().Set("Cache-Control", "no-cache, private, max-age=0")
+	jsonhttp.OK(w, tag)
```
janos, line 57: Here, also, an explicit response struct would be more correct.

pkg/api/tag.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		s.Logger.Debugf("bzz-tag: tag not present : %v, uuid %s", err, uidStr)
+		s.Logger.Error("bzz-tag: tag not present")
+		jsonhttp.InternalServerError(w, "tag not present")
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Cache-Control", "no-cache, private, max-age=0")
+	jsonhttp.OK(w, tag)
```
janos, line 81: Explicit response, here, also.

pkg/api/tag.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	tag, err := s.Tags.Get(uint32(uuid))
+	if err != nil {
+		s.Logger.Debugf("bzz-tag: tag not present : %v, uuid %s", err, uidStr)
+		s.Logger.Error("bzz-tag: tag not present")
+		jsonhttp.InternalServerError(w, "tag not present")
+	}
+
+	w.Header().Set("Content-Type", "application/json")
```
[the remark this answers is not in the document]
janos: Now, this line is not needed as Content-Type is set by jsonhttp.OK.

pkg/api/tag.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (s *server) getTagInfoUsingUUid(w http.ResponseWriter, r *http.Request) {
+	uidStr := mux.Vars(r)["uuid"]
+
+	uuid, err := strconv.ParseUint(uidStr, 10, 32)
+	if err != nil {
+		s.Logger.Debugf("bzz-tag: parse uuid  %s: %v", uidStr, err)
+		s.Logger.Error("bzz-tag: error uploading chunk")
+		jsonhttp.BadRequest(w, "invalid chunk address")
```
janos, line 67: These three log messages have inconstant messages.

pkg/api/tag.go
```diff
@@ -0,0 +1,82 @@
...
+	jsonhttp.OK(w, tag)
+}
+
+func (s *server) getTagInfoUsingAddress(w http.ResponseWriter, r *http.Request) {
+	addr := mux.Vars(r)["addr"]
+	address, err := swarm.ParseHexAddress(addr)
+	if err != nil {
+		s.Logger.Debugf("bzz-tag: parse chunk address %s: %v", addr, err)
+		s.Logger.Error("bzz-tag: error uploading chunk")
+		jsonhttp.BadRequest(w, "invalid chunk address")
```
janos, line 43: These three log messages have inconstant messages.

pkg/api/tag.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+func (s *server) getANewTag(w http.ResponseWriter, r *http.Request) {
+	tagName := mux.Vars(r)["name"]
+	if tagName == "" {
+		s.Logger.Debugf("bzz-tag: invalid tag name : %s", tagName)
+		s.Logger.Error("bzz-tag: invalid tag name")
+		jsonhttp.BadRequest(w, "invalid tag name")
+		return
+	}
```
janos, line 23: I do not think that we must check for mux vars if it is set.

pkg/api/router.go
```diff
@@ -45,6 +45,18 @@ func (s *server) setupRouting() {
 		"POST": http.HandlerFunc(s.chunkUploadHandler),
 	})
 
+	router.Handle("/bzz-tag/name/{name}", jsonhttp.MethodHandler{
+		"GET": http.HandlerFunc(s.getANewTag),
```
janos, line 49: As this method changes the state, it should be `POST`. Also, maybe to rename it to `createTag`?

pkg/api/tag.go
```diff
@@ -0,0 +1,82 @@
...
+
+	tag, err := s.Tags.GetByAddress(address)
+	if err != nil {
+		s.Logger.Debugf("bzz-tag: tag not present %s : %v, ", address.String(), err)
+		s.Logger.Error("bzz-tag: tag not present")
+		jsonhttp.InternalServerError(w, "tag not present")
+		return
+	}
+
+	w.Header().Set("Content-Type", jsonhttp.DefaultContentTypeHeader)
```
janos, line 55: No need to explicitly set Content-Type as jsonhttp.OK is used to serve response.

pkg/api/tag_test.go
```diff
@@ -0,0 +1,200 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
janos, line 1: It would be good to have a test where multiple chunks are uploaded with the same tag uid.

pkg/pusher/pusher.go
```diff
@@ -123,6 +134,11 @@ func (s *Service) setChunkAsSynced(ctx context.Context, addr swarm.Address) {
 		s.metrics.ErrorSettingChunkToSynced.Inc()
 	} else {
 		s.metrics.TotalChunksSynced.Inc()
+		ta, err := s.tag.GetByAddress(addr)
+		if err != nil {
+			s.logger.Debugf("pusher: get tag by address %s: %v", addr, err)
+		}
+		ta.Inc(tags.StateSynced)
```
janos, line 141: If the error is not nil, `ta` is probably nil or not usable. This code may panic or set invalid state if we do not return in case of `err != nil`.

pkg/storage/mock/storer.go
```diff
@@ -59,8 +62,19 @@ func (m *MockStorer) Put(ctx context.Context, mode storage.ModePut, chs ...swarm
 			}
 		}
 		m.store[ch.Address().String()] = ch.Data()
+		yes, err := m.Has(ctx, ch.Address())
+		if err != nil {
+			exist = append(exist, false)
+			continue
```
janos, line 68: Return error here?

pkg/api/tag_test.go
jmozah: There is a test "get-tag-indirectly-and-use-it-to-upload-chunk" .. but i will add a direct scenario also

pkg/storage/mock/storer.go
jmozah: If err != nil , then only that case should be affected... we should go through other cases anyway as they can be true.

janos: In that case, an error that may be relevant is hidden. Given that this is the mock, it does not have to be that strict, the only reason for it would be to have more information on test failure  and during its debug process. It can stay as it is now.

pkg/api/tag.go
```diff
@@ -0,0 +1,125 @@
...
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/gorilla/mux"
+)
+
+type TagResponse struct {
```
janos, line 17: This type can be unexported and export it only in export_test.go for tests. It also can have a constructor to avoid repeating the assignments in every handler.

pkg/api/tag.go
```diff
@@ -0,0 +1,125 @@
...
+	StartedAt time.Time     `json:"startedAt"`
+}
+
+func (s *server) getANewTag(w http.ResponseWriter, r *http.Request) {
+	tagName := mux.Vars(r)["name"]
+
+	tag, err := s.Tags.Create(tagName, 0, false)
+	if err != nil {
+		s.Logger.Debugf("bzz-chunk: tag get error: %v", err)
+		s.Logger.Error("bzz-chunk: tag get error")
```
janos, line 37: `tag get error` -> `tag create error` as the function is Tags.Create.

pkg/api/tag_test.go
janos: We should check for other tag counters, like Total for multiple chunks and that can be most clearly be done in a specific test for it.

jmozah:
> We should check for other tag counters, like Total for multiple chunks and that can be most clearly be done in a specific test for it.

In old code `total` was done in file splitter level in one go. Now i added it as sort of incrementing as we upload chunks.

pkg/api/tag.go
```diff
@@ -0,0 +1,106 @@
...
+	Sent      int64         `json:"sent"`
+	Synced    int64         `json:"synced"`
+	Uid       uint32        `json:"uid"`
+	Anonymous bool          `json:"anonymous"`
+	Name      string        `json:"name"`
+	Address   swarm.Address `json:"address"`
+	StartedAt time.Time     `json:"startedAt"`
+}
+
+func newTag(tag *tags.Tag) tagResponse {
```
janos, line 33: Should be newTagResponse.

pkg/api/tag_test.go
```diff
@@ -0,0 +1,237 @@
...
+			Message: http.StatusText(http.StatusOK),
+			Code:    http.StatusOK,
+		}, nil)
+
+		isTagFoundInResponse(t, rcvdHeaders, nil)
+	})
+
+	t.Run("get-tag-and-use-it-to-upload-chunk", func(t *testing.T) {
+		// Get a tag using API
+		ta := &tags.Tag{}
```
janos, line 61: To be completely correct, now, the response is api.TagResponse (can be exported in export_test.go). This type should be unmarshalled to, while leaving tags.Tag independent of API tests. This applies to other subtests in this package.

pkg/api/tag_test.go
jmozah: its done now

pkg/api/bzzchunk.go
janos: This comment is left.

pkg/api/router.go
janos: This comment is left.

pkg/api/tag.go
```diff
@@ -0,0 +1,106 @@
...
+		Seen:      tag.Seen,
+		Stored:    tag.Stored,
+		Sent:      tag.Sent,
+		Synced:    tag.Synced,
+		Uid:       tag.Uid,
+		Anonymous: tag.Anonymous,
+		Name:      tag.Name,
+		Address:   tag.Address,
+		StartedAt: tag.StartedAt,
+	}
```
janos, line 46: Could this be just `tagResponse(*tag)`?

pkg/api/tag.go
janos: This comment is left.

pkg/api/tag_test.go
janos: thanks @jmozah

pkg/api/bzzchunk.go
```diff
@@ -62,6 +105,9 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	tag.Address = address
```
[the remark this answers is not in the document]
jmozah: no. we don't need to open any discussion...  but you are right, we should not expose this API for chunks... we should expose them only for files that are split inside bee. @janos WDYS?

pkg/api/bzzchunk.go
```diff
@@ -28,11 +35,46 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+		s.Logger.Error("bzz-chunk: parse chunk address")
 		jsonhttp.BadRequest(w, "invalid chunk address")
 		return
 	}
 
+	// if tag header is not there create a new one
+	var tag *tags.Tag
+	tagUidStr := r.Header.Get(TagHeaderUid)
+	if tagUidStr == "" {
+		tagName := fmt.Sprintf("unnamed_tag_%d", time.Now().Unix())
```
acud, line 47: i suggest that if there is no longer a possibility to name tags (this was the original tag header purpose, which is now removed), then the tag name field should be removed from the codebase, since we're ending up with unnamed tags everywhere. either this or reinstate the header with possibility to give a name to a tag (a name which is not unique, as opposed to the uid that is generated there). it was not my intention that this functionality be removed. you are free to iterate on this in later pr if you want to.

jmozah: we now have two ways to use tags
- we have a seperate API to create `tags.` there you can supply `name` and get an `uid` back then use that  `uid` set in `TagHeaderUid` in subsequent chunks
- just send the first `chunk` without `TagHeaderUid` set, then get back a uid in return header, use that in subsequent chunks (this is where the unnamed_tags_ will be used

The first one is recommended way.
Hope this clarifies .

pkg/api/bzzchunk.go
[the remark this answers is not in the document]
janos: This line is indeed a questinable, but here we do not have the root address, we do not know which address is the root, or whether the root address exists at all for the chunks. But the tag uses the address only as a reference, whether it is a root hash or any other. Does it make sense to set the address of the first chunk for the same tag here and never to overwrite it?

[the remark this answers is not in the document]
jmozah: The most sensible thing i can think of is to add the last chunk's address to it. If we use a splitter that is similar to pyramid chunker, the the last chunk that will come out of it will be the root chunk. so this would work out in that case.

Since we dont know much about the final splitter implementation... i just added it like this.

PR #200, opened 2020-05-21
annotate errors in shed

janos:
This PR removes extensive logging next to returning the error in shed and properly annotates these errors so that their path is known. The consequence is that error checking has to be done with errors.Is function, which is also changed in this PR. This PR is based on the bad user experience when uploading and downloading chunks results with error messages that are not clear and not needed. There is no need to log errors in shed as they are all returned, so logger is removed, just as it does not exist in the swarm project code.

### DIFF
--- a/pkg/localstore/gc_test.go
+++ b/pkg/localstore/gc_test.go
@@ -19,6 +19,7 @@ package localstore
 import (
 	"bytes"
 	"context"
+	"errors"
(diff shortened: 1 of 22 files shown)


janos:
Thanks @acud. Yes, errors.Is is checking for all wrapped errors, even the ones in the middle of the chain.

PR #205, opened 2020-05-22
Binary encoding of networkID

pradovic:
@zelig 's comments from: https://github.com/ethersphere/bee/pull/196.
Moved handshake tests to single function to re-use data init.

### DIFF
--- a/pkg/p2p/libp2p/internal/handshake/handshake_test.go
+++ b/pkg/p2p/libp2p/internal/handshake/handshake_test.go
@@ -62,12 +62,26 @@ func TestHandshake(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	node1Overlay := crypto.NewOverlayAddress(privateKey1.PublicKey, 0)
-	node2Overlay := crypto.NewOverlayAddress(privateKey2.PublicKey, 0)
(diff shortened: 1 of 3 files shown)


pkg/p2p/libp2p/internal/handshake/handshake_test.go
```diff
@@ -42,13 +66,14 @@ func TestHandshake(t *testing.T) {
 	node2Overlay := crypto.NewOverlayAddress(privateKey2.PublicKey, 0)
 	signer1 := crypto.NewDefaultSigner(privateKey1)
 	signer2 := crypto.NewDefaultSigner(privateKey2)
-
-	signature1, err := signer1.Sign([]byte("underlay10"))
+	networkIDBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(networkIDBytes, 0)
```
acud, line 70: the code above uses little endian....

pradovic: Ah, 0 is so bad test network id 🙈

pradovic:
@acud Yup, I totally agree, that's why I went for little endian as it was already used for overlay, but now when I think about it it might be a mistake, as swarm also uses different endians. I will switch bach to big endian, as Viktor said then, and let's tlak some day to see if we should unify them or not.

PR #211, opened 2020-05-25
Squared gaps substitue / balanced bins

metacertain:
The aim of this PR is to introduce balanced bins in 6 steps

Create a list of ideal node addresses at node startup for each bin (nnlowwatermark number of bitwise diverse addresses)
	1) To do this, get prefix of Bin[k] by using Prefix(k) of base address
	2) Suffix this address with 0, 1 ... 2^n - 1 represented in n digit binary form to get nnlowwatermark number of bitwise diverse addresses where nnlowwatermark = 2^n
	3) Add the created ideal addresses to idealPeers PSlice at bin[k]

When checking saturation we are going to check for balancedness as well:
	1) for each ideal address check whether we have a peer connected with a matching k + n prefix in the same PO
	2) if yes for each, we can conclude we are saturated and balanced

If we are not saturated or balanced
prereq: Modify/clone ClosestPeer to be able to search either in connectedPeers or knownPeers for a closest peer
	1) for each ideal address, find ClosestPeer from knownPeers and connect if we haven't yet

Not included functionality:
When we are saturated and unbalanced, we will keep connecting to new peers if they help balancedness
This will not remove connections that were previously used for filling empty slots


Eknir:
What is the status @metacertain ?

PR #212, opened 2020-05-26
Add logging to cli commands

nolash:
This PR adds logging capability to cli commands for split and join, as well as a minimum set of logging info from the underlying packages.

Perhaps the single most annoying thing in the world is when it is impossible as an advanced user to get contextual information about what a tool does, at a minimum if it has understood the input parameters passed to it. Let's not be annoying.


nolash:
Obsoleted by https://github.com/ethersphere/bee/pull/214

PR #215, opened 2020-05-26
Add logging to cli

nolash:
This PR adds logging capability to cli commands for split and join, as well as a minimum set of logging info from the underlying packages.

Perhaps the single most annoying thing in the world is when it is impossible as an advanced user to get contextual information about what a tool does, at a minimum if it has understood the input parameters passed to it. Let's not be annoying.

### DIFF
--- a/cmd/bee-join/main.go
+++ b/cmd/bee-join/main.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 
 	"github.com/ethersphere/bee/pkg/file/joiner"
+	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/spf13/cobra"
@@ -26,6 +27,8 @@ var (
 	host         string // flag variable, http api host
 	port         int    // flag variable, http api port
 	ssl          bool   // flag variable, uses https for api if set
+	loglevel     int    // flag variable, sets loglevel for operation
+	logger       logging.Logger
 )
 
 // apiStore provies a storage.Getter that retrieves chunks from swarm through the HTTP chunk API.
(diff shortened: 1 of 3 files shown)


cmd/bee-join/main.go
```diff
@@ -153,6 +161,7 @@ Will output retrieved data to stdout.`,
 	c.Flags().StringVar(&host, "host", "127.0.0.1", "api host")
 	c.Flags().IntVar(&port, "port", 8080, "api port")
 	c.Flags().BoolVar(&ssl, "ssl", false, "use ssl")
+	c.Flags().IntVarP(&loglevel, "loglevel", "l", 0, "log verbosity")
```
acud, line 164: in `bee` this is called `verbosity`, let's keep the same convention across the codebase

cmd/bee-join/main.go
```diff
@@ -94,14 +106,10 @@ func Join(cmd *cobra.Command, args []string) (err error) {
 			return err
 		}
 		defer outFile.Close()
+		logger.Debugf("Writing data from Swarm Hash %s to file %s", addr, outFilePath)
```
acud, line 109: no need to capitalize the log messages. in the same manner as we do in `bee`

pkg/logging/logging.go
```diff
@@ -50,3 +50,12 @@ func New(w io.Writer, level logrus.Level) Logger {
 func (l *logger) NewEntry() *logrus.Entry {
 	return logrus.NewEntry(l.Logger)
 }
+
+// ToLogLevel bounds a numeric loglevel value to the amount of loglevels
+// provided by the underlying logging package
+func ToLogLevel(level int) logrus.Level {
```
janos, line 56: This function is aways called before `logging.New`. It can be implicit in `logging.New` to protect the log level without the need to be called explicitly and also should be unexported.

acud:
why `on hold`?

nolash:
@acud because this first: https://github.com/ethersphere/bee/pull/251

acud:
> @acud because this first: #251

what's the connection between the two? is this pr going to redo the logging on all of those binaries too then?

nolash:
Yeah I'd rather just do it on top.

nolash:
obsoleted by https://github.com/ethersphere/bee/pull/251

Issue #234, opened 2020-05-28
Peers are listed as disconnected in kademlia but they are not


janos:
Some of the peers are listed in kademlia topology response as disconnected even they are not according to debug api /peers response, log messages and pingpong functioning properly.

This issue is the a followup on https://github.com/ethersphere/bee/issues/219 with the difference that the consequence different but manifestation is similar.

vandot:
Trying to implement node discovery inside local network using multicast I get this issue every time. Implementation is that node is started and it sends discovery packet, every other running node that receives that message connects to that node.

Interesting thing is that in cluster of 5 nodes node `0` (started first so in the end it initiates connection to every node started after) has all 4 nodes listed as disconnected, node `1` has one connected node (node 0) and 3 disconnected, node `2` 2 connected (node 0 and 1) and two disconnected, up to last node that has all nodes listed as connected (all other nodes initiated connection to it)

TL;DR if the node initiate connection to other node it will set that node as disconnected in `/topology`, node that was connected to will keep initiator as connected.

fullconnectivity and pingpong tests passes and no log messages about disconnecting nodes

PR #236, opened 2020-05-29
kademlia, debugapi, node: fix connection issues

acud:
This PR fixes some bugs that were observed on the clusters and on local testing, namely inconsistencies of peers shown as disconnected while being in fact connected.

A short explanation of the changeset with its repercussions:
- `debugapi/peer.go` - the `AddPeer` method on kademlia was called _after connecting to the peer_. this is legacy from the full topology implementation and is incorrect in this case. The resulting behavior would be that the peer would be connected to, then would be added to the _known_ peers list, instead of both _known and connect_ lists. So the kademlia would try to iterate in `manage` but the peer would never show up in the known peers list, so would never show up in the iterator.
- it could be, that under certain circumstances that this peer would be gossiped to the node from another node, eventually triggering adding the peer through `AddPeer`, resulting in it being in the `knownPeers` slice. It is difficult to say what would be the behavior in such case, since `AddPeer` does not guarantee trying to connect to such a node, so we cannot guarantee in such case that the peer will ever shown in the `connectedPeers` list
- `kademlia.go` - moved the addition of a peer to the connect peers list after connect has finished. this is to prevent a data race when we try to connect to a peer, but it has already dialed in but was not yet added to the kademlia (the `Connected` callback was not yet called). In this case the peer should also be added to the connected peers list. This call is of no harm, since duplicates are not allowed as part of `pslice` guarantees (they are skipped and not added)
- `kademlia.go` - add peer to known peers slice on dial in - the problem was that a peer has dialed in, but was never known to us, resulting in an inconsistency in the kademlia
- `node.go` - peer should not be disconnected from when `AddPeer` fails in the case of kademlia (since `AddPeer` doesn't guarantee a connection)

fixes https://github.com/ethersphere/bee/issues/234

### DIFF
--- a/pkg/debugapi/peer.go
+++ b/pkg/debugapi/peer.go
@@ -44,7 +44,7 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	if err := s.TopologyDriver.Connected(r.Context(), address); err != nil {
 		_ = s.P2P.Disconnect(address)
-		s.Logger.Debugf("debug api: topologyDriver.AddPeer %s: %v", addr, err)
+		s.Logger.Debugf("debug api: topologyDriver.Connected %s: %v", addr, err)
 		s.Logger.Errorf("unable to connect to peer %s", addr)
 		jsonhttp.InternalServerError(w, err)
 		return
--- a/pkg/debugapi/peer.go
+++ b/pkg/debugapi/peer.go
@@ -42,7 +42,7 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 		jsonhttp.InternalServerError(w, err)
 		return
 	}
-	if err := s.TopologyDriver.AddPeer(r.Context(), address); err != nil {
+	if err := s.TopologyDriver.Connected(r.Context(), address); err != nil {
 		_ = s.P2P.Disconnect(address)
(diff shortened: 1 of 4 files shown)


pkg/debugapi/peer.go
```diff
@@ -42,7 +42,7 @@ func (s *server) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 		jsonhttp.InternalServerError(w, err)
 		return
 	}
-	if err := s.TopologyDriver.AddPeer(r.Context(), address); err != nil {
+	if err := s.TopologyDriver.Connected(r.Context(), address); err != nil {
 		_ = s.P2P.Disconnect(address)
 		s.Logger.Debugf("debug api: topologyDriver.AddPeer %s: %v", addr, err)
```
janos, line 47: Debug log message should be adjusted as well.

pradovic:
@acud one added rationale here, regarding the added peer and connected peer triggers. I noticed this also in the code while working on bzz address PR. Based on previous decisions ,this was somewhat separated and the only point of sync we have until the node is connected. 
On the other side `AddPeer` is separated cuz it can be called for not connected nodes as well. 
So anywayz, when `AddPeer` handler is called, the handler can check if node is connected or not, because connection should happen before `AddPeer` handler, but that does not hold vice versa. Node is not added to topology at the "same time" it is connected.

PR #242, opened 2020-06-01
Add file entry implementation

nolash:
This PR adds generic interfaces needed for representing nodes and leaves in a file- and directory-oriented view of Swarm, and a simple file metadata entry implementation.

The entry implementation encapsulates a data reference and a reference to metadata. The metadata in turn consists of file name and content type. this conforms with the buzz description of file entry, which separates file/info from reference (and other things like acl, erasure coding) as individual swarm objects.

It also moves address generation test helper code to swarm, as this is not specific to topology where it was last hiding out.


pkg/collection/entry/entry.go
```diff
@@ -0,0 +1,63 @@
...
+
+// New creates a new Entry.
+func New(reference swarm.Address) *Entry {
+	return &Entry{
+		reference: reference,
+	}
+}
+
+// SetMetadata sets the metadata for the data reference in Entry.
+func (e *Entry) SetMetadata(metadataAddress swarm.Address) {
```
acud, line 33: i wonder, why would you like to separate this? why doesn't the constructor take a reference _and_ a metadata address/entry

pkg/collection/entry/metadata.go
```diff
@@ -0,0 +1,50 @@
...
+package entry
+
+import (
+	"bytes"
+	"fmt"
+)
+
+// Metadata provides mime type and filename to file entry.
+type Metadata struct {
+	MimeType string
```
acud, line 14: why do you need to export these fields? you use getters in the entry type, why not here?

pkg/collection/entry/metadata.go
```diff
@@ -0,0 +1,50 @@
...
+
+// NewMetadata creates a new Metadata.
+func NewMetadata(fileName string) *Metadata {
+	return &Metadata{
+		Filename: fileName,
+	}
+}
+
+// WithMimeType adds mime type to Metadata.
+func (m *Metadata) WithMimeType(mimeType string) *Metadata {
```
acud, line 26: if `MimeType` is already exported, why do you need this function for?

pkg/collection/entry/metadata.go
```diff
@@ -0,0 +1,50 @@
...
+	m.MimeType = mimeType
+	return m
+}
+
+// MarshalBinary implements encoding.BinaryMarshaler
+func (m *Metadata) MarshalBinary() ([]byte, error) {
+	b := []byte(m.Filename)
+	if m.MimeType != "" {
+		bs := append(b, []byte{0x00}...)
```
acud, line 35: aren't strings terminated by a `nil` character? wouldn't this be a problem?

pkg/collection/collection.go
```diff
@@ -0,0 +1,30 @@
...
+package collection
+
+import (
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+type MetadataType uint8
+
+const (
+	FilenameMimeType MetadataType = iota
```
janos, line 15: Should this be `FilenameMetadataType`, or even `MetadataTypeFilename` just as for example http.Status* constants?

Also, why would this be needed now, as it is not used? I think that we should not add code in pr that is not directly usable within the scope of pr. We can add it later if needed, but for now there is only one metadata without any association to its MetadataType.

pkg/collection/entry/entry.go
janos: Right, this method is not needed. Metadata address can be set in the constructor.

pkg/collection/entry/metadata.go
janos: I do not think that we need this method, just set mime type in the constructor.

pkg/collection/entry/metadata.go
janos: This encoding is very arbitrary even if it is simple. It means that we need to escape 0 character in the filename and mime type. My opinion is that json encoding provides more benefits as it is well known and we will have easier time in the future when we extend metadata and want to ensure backward compatibility.

pkg/collection/entry/metadata_test.go
```diff
@@ -0,0 +1,36 @@
...
+package entry_test
+
+import (
+	"testing"
+
+	"github.com/ethersphere/bee/pkg/collection/entry"
+)
+
+// TestMetadataSerialize verifies integrity of serialization.
+func TestMetadataSerialize(t *testing.T) {
```
janos, line 14: This test needs to be extended for the known edge cases with the encoding which are 0 character in the filename and in the mime type. I would rather avoid custom encodings.

pkg/collection/entry/metadata.go
nolash: You may be right, I will double check if any text encoding would use zero byte. We would want `utf-8` filenames I guess. `0x00` as delimiter of string is a pretty normal thing, though.

pkg/collection/entry/metadata.go
nolash: The entry type has getters because they need to implement an interface. Happy to put getters here too, even if no interface is implemented, if you think it's better?

janos: Actually, there are no getters for these fields. I would keep it simple, exported fields are the simplest. :)

acud: cool with exporting, but then we dont get promise of immutability

pkg/collection/entry/metadata.go
acud: > 0x00 as delimiter of string is a pretty normal thing, though.

so then when you split, you get an extra entry no? since this byte constitutes an empty string then?

nolash: I'll just fastfood json it and end of discussion ;)

nolash: > It means that we need to escape 0 character in the filename and mime type.

that being said; this is not "0" character, this is 0 byte. There are no zero-bytes in string encoding, except to define end-of-string.

janos: Yes, of course, I did not mean "0" character, but character with all 0 bytes, just as you wrote 0x00, which is the same as 0.

pkg/collection/collection.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// MetadataType disambiguates different type of metadata structures between entry
+// implementations.
+type MetadataType uint8
+
+const (
+	CustomMetadataType MetadataType = iota
```
janos, line 17: This type is not used as Entry.Metadata() ignores its argument. I do not think that MetadataType is needed in this PR. If it is needed in the future, let's add it then.

nolash: but what about the interface?

janos: We can change the interface, when we have different metadata types. For now, I do not see that we need more.

pkg/collection/entry/entry_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"testing"
+
+	"github.com/ethersphere/bee/pkg/collection"
+	"github.com/ethersphere/bee/pkg/collection/entry"
+	"github.com/ethersphere/bee/pkg/swarm/test"
+)
+
+var (
+	_ = collection.Entry(&entry.Entry{})
```
janos, line 16: To ensure interface implementation during build, this line should not be in test file, but in entry.go file, as build is ignoring test files.

PR #245, opened 2020-06-02
Fix bzz addr in debugapi

pradovic:
While reverting changes in bzz address PR I forgot to fix `debugapi` part as well, which needed a slight change of `p2p.Connect` method as well.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -360,9 +360,9 @@ func NewBee(o Options) (*Bee, error) {
 					return
 				}
 
-				if err := topologyDriver.AddPeer(p2pCtx, bzzAddr.Overlay); err != nil {
+				if err := topologyDriver.Connected(p2pCtx, bzzAddr.Overlay); err != nil {
 					_ = p2ps.Disconnect(bzzAddr.Overlay)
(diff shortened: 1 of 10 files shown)


pkg/node/node.go
```diff
@@ -346,32 +345,24 @@ func NewBee(o Options) (*Bee, error) {
...
-					_ = p2ps.Disconnect(overlay)
-					logger.Debugf("addressbook error persisting %s %s: %v", a, overlay, err)
-					logger.Errorf("connect to bootnode %s", a)
-					return
-				}
-
-				if err := topologyDriver.AddPeer(p2pCtx, overlay); err != nil {
-					_ = p2ps.Disconnect(overlay)
-					logger.Debugf("topology add peer fail %s %s: %v", a, overlay, err)
+				if err := topologyDriver.AddPeer(p2pCtx, bzzAddr.Overlay); err != nil {
```
acud, line 363: this has slipped under my radar on the last round of fixes and is incorrect. the `Connected` notification should be called here, since we are already connected to the bootnode at this stage

pradovic: I am aware of this, as I wrote in `mattermost`, but I did no want to change the functionality now, but just revert, to avoid going through the whoel debugging process now. Can I change this in my second PR?

PR #250, opened 2020-06-03
Add db capacity flag

jmozah:
add a dbCapacity flag so that we can test local pinning

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -59,6 +59,7 @@ type Bee struct {
 
 type Options struct {
 	DataDir            string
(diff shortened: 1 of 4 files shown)


pkg/node/node.go
```diff
@@ -59,6 +59,7 @@ type Bee struct {
 
 type Options struct {
 	DataDir            string
+	DbCapacity         uint64
```
janos, line 62: `DbCapacity` -> `DBCapacity`

acud:
cool. i approved in any case, so no blockers

PR #251, opened 2020-06-03
Add CLI for file entry add and resolve

nolash:
This PR adds a `bee-file` binary that creates file entries and also writes files to fs based on the metadata in the file entry.

Furthermore, it makes the following changes:

* add a chunk equality method
* add loggers to all non-bee cli commands
* no http default for split, now just returns swarmhash by default
* fix possible oversight in bee binary where logging logs to stdout


cmd/bee-file/main.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	// open the file
+	outFile, err := os.OpenFile(outFilePath, outFileFlags, 0o666) // skipcq: GSC-G302
+	if err != nil {
+		return err
+	}
+	defer outFile.Close()
+
+	err = cmdfile.JoinReadAll(j, e.Reference(), outFile)
```
acud, line 119: can probably just return here `return cmdfile.Join....`

cmd/bee-file/main.go
```diff
@@ -0,0 +1,255 @@
...
+	// serialize metadata and send it to splitter
+	metadataBytes, err := json.Marshal(metadata)
+	if err != nil {
+		return err
+	}
+	logger.Debugf("metadata contents: %s", metadataBytes)
+
+	// set up splitter to process the metadata
+	s := splitter.NewSimpleSplitter(stores)
+	ctx, cancel := context.WithCancel(context.Background())
```
acud, line 168: i wonder if we need the cancel func at all. i cant see this being used except in the defer.

cmd/bee-file/main.go
```diff
@@ -0,0 +1,255 @@
...
+	fileEntryBuf := bytes.NewBuffer(fileEntryBytes)
+	fileEntryReader := io.LimitReader(fileEntryBuf, int64(len(fileEntryBytes)))
+	fileEntryReadCloser := ioutil.NopCloser(fileEntryReader)
+	fileEntryAddr, err := s.Split(ctx, fileEntryReadCloser, int64(len(fileEntryBytes)))
+	if err != nil {
+		return err
+	}
+
+	// output reference to file entry
+	fmt.Println(fileEntryAddr)
```
acud, line 196: i think the cobra `cmd` has some Println functionality that might be better

cmd/bee-join/main.go
```diff
@@ -26,50 +22,18 @@ var (
...
-	if res.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("chunk %s not found", addressHex)
-	}
-	chunkData, err := ioutil.ReadAll(res.Body)
+// Join is the underlying procedure for the CLI command
+func Join(cmd *cobra.Command, args []string) (err error) {
+	logger, err = cmdfile.SetLogger(cmd, verbosity)
 	if err != nil {
-		return nil, err
+		fmt.Fprintln(os.Stderr, err.Error())
```
acud, line 33: wouldn't returning the error from the function will result in this anyway?

cmd/bee-split/main.go
```diff
@@ -28,125 +23,23 @@ var (
...
-func (l *limitReadCloser) Close() error {
-	return l.closeFunc()
-}
-
 // Split is the underlying procedure for the CLI command
 func Split(cmd *cobra.Command, args []string) (err error) {
-	var infile io.ReadCloser
+	logger, err = cmdfile.SetLogger(cmd, verbosity)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err.Error())
```
acud, line 36: again, probably cleaner to return the error here no?

cmd/bee/cmd/start.go
```diff
@@ -56,15 +56,15 @@ func (c *command) initStartCmd() (err error) {
 			case "0", "silent":
 				logger = logging.New(ioutil.Discard, 0)
 			case "1", "error":
-				logger = logging.New(cmd.OutOrStdout(), logrus.ErrorLevel)
+				logger = logging.New(cmd.OutOrStderr(), logrus.ErrorLevel)
```
acud, line 59: why? i personally prefer to stdout, errors are anyway written to stderr afaik

cmd/file/io.go
```diff
@@ -0,0 +1,235 @@
...
+	putters []storage.Putter
+}
+
+// NewTeeStore creates a new TeeStore.
+func NewTeeStore() *TeeStore {
+	return &TeeStore{}
+}
+
+// Add adds a storage.Putter.
+func (t *TeeStore) Add(putter storage.Putter) {
```
acud, line 63: why not in the constructor?
```
NewTeeStore(putters ...storage.Putter) *TeeStore {

}
```

Also, `NewTeeStore` should probably return a `storage.Putter`. No need to return the concrete type here. then `TeeStore` type can be unexported

cmd/file/io.go
```diff
@@ -0,0 +1,235 @@
...
+
+// Put implements storage.Putter.
+func (t *TeeStore) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
+	for _, putter := range t.putters {
+		_, err := putter.Put(ctx, mode, chs...)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return nil, nil
```
acud, line 75: note you're silently breaking the contract of the function signature. this is not so great as the caller might expect a slice here, resulting in a panic. i would expect at least to return a slice in the same size as the length of the variadic function parameter, all with `false` values

cmd/file/io.go
```diff
@@ -0,0 +1,235 @@
...
+// Put implements storage.Putter.
+func (f *FsStore) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
+	for _, ch := range chs {
+		chunkPath := filepath.Join(f.path, ch.Address().String())
+		err := ioutil.WriteFile(chunkPath, ch.Data(), 0o666)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return nil, nil
```
acud, line 100: here too

cmd/file/io.go
```diff
@@ -0,0 +1,235 @@
...
+		url := strings.Join([]string{a.baseUrl, addr}, "/")
+		res, err := a.Client.Post(url, "application/octet-stream", buf)
+		if err != nil {
+			return nil, err
+		}
+		if res.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("upload failed: %v", res.Status)
+		}
+	}
+	return nil, nil
```
acud, line 140: ...

cmd/file/io.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+}
+
+// Write implements io.Writer.
+func (l *LimitWriteCloser) Write(b []byte) (int, error) {
+	if l.total+int64(len(b)) > l.limit {
+		return 0, errors.New("overflow")
+	}
+	c, err := l.WriteCloser.Write(b)
+	l.total += int64(c)
```
acud, line 183: if err is true, i'm not sure `total` should be incremented, or?

cmd/file/io.go
```diff
@@ -0,0 +1,235 @@
...
+	return c, err
+}
+
+// Join reads all output from the provided joiner.
```
acud, line 187: `JoinReadAll ....`

cmd/file/io.go
```diff
@@ -0,0 +1,235 @@
...
+			return err
+		}
+		if cw != cr {
+			return fmt.Errorf("short wrote %d of %d for chunk %d", cw, cr, i)
+		}
+	}
+	if total != l {
+		return fmt.Errorf("received only %d of %d total bytes", total, l)
+	}
+	return err
```
acud, line 213: `return nil`?

cmd/file/io.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func SetLogger(cmd *cobra.Command, verbosityString string) (logger logging.Logger, err error) {
+	v := strings.ToLower(verbosityString)
+	switch v {
+	case "0", "silent":
+		logger = logging.New(ioutil.Discard, 0)
+	case "1", "error":
+		logger = logging.New(cmd.OutOrStderr(), logrus.ErrorLevel)
```
acud, line 222: not sure about stderr as the default for everything in this block

nolash: It would be a very strange thing to log to stdout at all, from a cli usage perspective.

cmd/bee/cmd/start.go
nolash: that's perfectly normal operation, to want the result of cli cmd and the log chatter kept separate.

cmd/file/io.go
nolash: I find this easier to read

nolash: > Also, NewTeeStore should probably return a storage.Putter.

:+1:

cmd/file/io.go
nolash: from `go doc -all io | grep "type Writer interface -A10`:

```
Write writes len(p) bytes from p to the underlying data stream. It returns
    the number of bytes written from p (0 <= n <= len(p)) and any error
    encountered that caused the write to stop early.
```

in other words, `c` may be > 0 even if `err` is not `nil`.

cmd/bee-file/main.go
nolash: "better" :D

cmd/file/io.go
nolash: Actually, if you look at `cmd/bee-split/main.go` you can see why having `Add` would be the better interface here.

cmd/bee/cmd/start.go
janos: Yes, log messages should be sent to stderr, as usual practice, as to have the command output to stdout as the result of the execution. I have set stdout here as start does not have the resulted output, only logs, but I have nothing against using stderr for log messages.

I would also like to hear from @svetomir or @vandot if this is ok from the infrastructure point of view.

cmd/file/io.go
```diff
@@ -0,0 +1,238 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
janos, line 1: Package `cmd/file` is not the `main` package and from its path it suggests that. I think that it would be more correct to utilize here internal package feature and the move this package to `cmd/internal/file`.

nolash: Will it then still be accessible by `cmd/bee-split/main.go`

I was thinking of this structure (but left it like this now to make the diffs for split and join clearer)

```
cmd/file/io.go
cmd/file/bee-split/main.go
cmd/file/bee-join/main.go
cmd/file/bee-file/main.go
```

That still doesn't fit what you are describing as a problem though.

cmd/file/io.go
acud: well you could have alternatively:
```go
var stores []storage.Storer
if file { stores = append(stores, NewFileStore()) }
if http { stores = append(stores, NewHttpStore()) }
tee := NewTeeStore(stores...)
```
but this is a terribly tyrannic and selective-vision nitpick :)

cmd/bee-file/main.go
acud: :D

I think that its "better" because if you use the standard `fmt.Println` and the process quits, there's that strange `%` sign showing up when the process exits. I never understood why. The cobra print should mitigate that

cmd/bee/cmd/start.go
acud: all messages? error I understand (and i think the logger should do this regardless), but everything else? i am used that `WARN` and `ERROR` go to `stderr`, but the rest to `stdout` no?

nolash: From a cli user point of view I would absolutely _loathe_ that, and I dare ever make the assumption that that's not the way it usually works.

cmd/file/io.go
nolash: Aswell as ugly as f*****ck ;)

cmd/file/io.go
janos: I would preserve that all first subdirectories under cmd are only main packages.

cmd/bee/cmd/start.go
janos: This is true for applications that produce some output. In this case, `bee start` does not produce any output and it is a long running process, with logs that can be considered as the output. Maybe even this PR is not the place to change this. This change in `bee start` should be in a separate PR.

cmd/file/io.go
nolash: I will try with the `internal` but I was under the impression that `internal` will only work for the package in its immediate parent directory...? in that case sharing code between those binaries is not possible using that method.

janos: They just need to share the a common ancestor, which is here `github.com/ethersphere/bee/cmd`.

nolash:
@janos I've changed the dir to `internal`. And I have also reverted the `logging` change in `bee`. That was what we needed reply from @vandot @svetomir for right?

cmd/bee-join/main.go
```diff
@@ -26,50 +22,17 @@ var (
...
-	if res.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("chunk %s not found", addressHex)
-	}
-	chunkData, err := ioutil.ReadAll(res.Body)
+// Join is the underlying procedure for the CLI command
+func Join(cmd *cobra.Command, args []string) (err error) {
+	logger, err = cmdfile.SetLogger(cmd, verbosity)
 	if err != nil {
-		return nil, err
+		os.Exit(1)
```
janos, line 33: We should log the error here. Or just return it as it was before?

cmd/bee-split/main.go
```diff
@@ -28,125 +23,22 @@ var (
...
-func (l *limitReadCloser) Close() error {
-	return l.closeFunc()
-}
-
 // Split is the underlying procedure for the CLI command
 func Split(cmd *cobra.Command, args []string) (err error) {
-	var infile io.ReadCloser
+	logger, err = cmdfile.SetLogger(cmd, verbosity)
+	if err != nil {
+		os.Exit(1)
```
janos, line 36: Also here, we can return the error.

PR #255, opened 2020-06-04
Local pinning fix

jmozah:
**Problem**: Sometime a pinned chunk gets GCd and is not available in `localstore`. This was not simulatable. It once occurred in FairData setup and now in @santicomp2014 's setup in global pinning. 

**Fix**: Even though we don't know the exact scenario when this happens, a code walk through found out that we are not checking for `pinIndex` presence when a chunk is added in `gcIndex`. We found 4 such places that are missed out and fixing it in this PR.

### DIFF
--- a/pkg/localstore/gc.go
+++ b/pkg/localstore/gc.go
@@ -79,7 +79,6 @@ func (db *DB) collectGarbageWorker() {
 // the rest of the garbage as the batch size limit is reached.
 // This function is called in collectGarbageWorker.
 func (db *DB) collectGarbage() (collectedCount uint64, done bool, err error) {
-	fmt.Println("Starting GC")
 	db.metrics.GCCounter.Inc()
 	defer totalTimeMetric(db.metrics.TotalTimeCollectGarbage, time.Now())
 	defer func() {
@@ -116,7 +115,6 @@ func (db *DB) collectGarbage() (collectedCount uint64, done bool, err error) {
 
 		db.metrics.GCStoreTimeStamps.Set(float64(item.StoreTimestamp))
 		db.metrics.GCStoreAccessTimeStamps.Set(float64(item.AccessTimestamp))
-		fmt.Println("GCing item ", hex.EncodeToString(item.Address))
 
 		yes, err := db.pinIndex.Has(item)
 		if err == nil && yes {
@@ -196,7 +194,6 @@ func (db *DB) removeChunksInExcludeIndexFromGC() (err error) {
 			return false, nil
 		}
 		if ok {
-			fmt.Println("Excluding a item from GC ", hex.EncodeToString(item.Address))
 			err = db.gcIndex.DeleteInBatch(batch, item)
 			if err != nil {
 				return false, nil
--- a/pkg/localstore/gc.go
+++ b/pkg/localstore/gc.go
@@ -17,6 +17,7 @@
 package localstore
 
 import (
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"time"
@@ -78,6 +79,7 @@ func (db *DB) collectGarbageWorker() {
 // the rest of the garbage as the batch size limit is reached.
 // This function is called in collectGarbageWorker.
 func (db *DB) collectGarbage() (collectedCount uint64, done bool, err error) {
+	fmt.Println("Starting GC")
 	db.metrics.GCCounter.Inc()
 	defer totalTimeMetric(db.metrics.TotalTimeCollectGarbage, time.Now())
 	defer func() {
@@ -114,7 +116,12 @@ func (db *DB) collectGarbage() (collectedCount uint64, done bool, err error) {
 
 		db.metrics.GCStoreTimeStamps.Set(float64(item.StoreTimestamp))
 		db.metrics.GCStoreAccessTimeStamps.Set(float64(item.AccessTimestamp))
+		fmt.Println("GCing item ", hex.EncodeToString(item.Address))
 
+		yes, err := db.pinIndex.Has(item)
+		if err == nil && yes {
+			fmt.Println("GCing pinned item ", hex.EncodeToString(item.Address))
+		}
 		// delete from retrieve, pull, gc
 		err = db.retrievalDataIndex.DeleteInBatch(batch, item)
 		if err != nil {
@@ -189,6 +196,7 @@ func (db *DB) removeChunksInExcludeIndexFromGC() (err error) {
 			return false, nil
 		}
 		if ok {
+			fmt.Println("Excluding a item from GC ", hex.EncodeToString(item.Address))
 			err = db.gcIndex.DeleteInBatch(batch, item)
 			if err != nil {
 				return false, nil
--- a/pkg/localstore/gc.go
+++ b/pkg/localstore/gc.go
@@ -17,7 +17,6 @@
 package localstore
 
 import (
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"time"
@@ -116,10 +115,6 @@ func (db *DB) collectGarbage() (collectedCount uint64, done bool, err error) {
 		db.metrics.GCStoreTimeStamps.Set(float64(item.StoreTimestamp))
 		db.metrics.GCStoreAccessTimeStamps.Set(float64(item.AccessTimestamp))
 
-		yes, err := db.pinIndex.Has(item)
-		if err == nil && yes {
-			fmt.Println("GCing pinned item ", hex.EncodeToString(item.Address))
-		}
 		// delete from retrieve, pull, gc
 		err = db.retrievalDataIndex.DeleteInBatch(batch, item)
 		if err != nil {
--- a/pkg/localstore/gc_test.go
+++ b/pkg/localstore/gc_test.go
@@ -544,7 +544,7 @@ func TestPinAfterMultiGC(t *testing.T) {
 	pinnedChunks := make([]swarm.Address, 0)
 
 	// upload random chunks above db capacity to see if chunks are still pinned
-	for i := 0; i < 2000; i++ {
+	for i := 0; i < 200; i++ {
 		ch := generateTestRandomChunk()
 		_, err := db.Put(context.Background(), storage.ModePutUpload, ch)
 		if err != nil {
@@ -560,7 +560,7 @@ func TestPinAfterMultiGC(t *testing.T) {
 			pinnedChunks = append(pinnedChunks, rch.Address())
 		}
 	}
-	for i := 0; i < 2000; i++ {
+	for i := 0; i < 200; i++ {
 		ch := generateTestRandomChunk()
 		_, err := db.Put(context.Background(), storage.ModePutUpload, ch)
 		if err != nil {
@@ -571,7 +571,7 @@ func TestPinAfterMultiGC(t *testing.T) {
 			t.Fatal(err)
 		}
 	}
-	for i := 0; i < 2000; i++ {
+	for i := 0; i < 200; i++ {
 		ch := generateTestRandomChunk()
 		_, err := db.Put(context.Background(), storage.ModePutUpload, ch)
 		if err != nil {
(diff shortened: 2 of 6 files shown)


acud:
@jmozah If this PR is ready for review, please elaborate on what exactly is the bug and how it is fixed in the PR description.

pkg/localstore/gc.go
```diff
@@ -115,6 +116,10 @@ func (db *DB) collectGarbage() (collectedCount uint64, done bool, err error) {
 		db.metrics.GCStoreTimeStamps.Set(float64(item.StoreTimestamp))
 		db.metrics.GCStoreAccessTimeStamps.Set(float64(item.AccessTimestamp))
 
+		yes, err := db.pinIndex.Has(item)
+		if err == nil && yes {
+			fmt.Println("GCing pinned item ", hex.EncodeToString(item.Address))
+		}
```
janos, line 122: This can be removed as it is used only to print a line.

jmozah:
Update: Just now @santicomp2014 confirmed that this works in there setup.

jmozah:
> @jmozah If this PR is ready for review, please elaborate on what exactly is the bug and how it is fixed in the PR description.

@acud. Added the descriptions. Hope you are happy now.

pkg/localstore/gc_test.go
```diff
@@ -534,3 +535,194 @@ func TestSetTestHookCollectGarbage(t *testing.T) {
...
+		_, err := db.Put(context.Background(), storage.ModePutUpload, ch)
+		if err != nil {
+			t.Fatal(err)
+		}
+		err = db.Set(context.Background(), storage.ModeSetSyncPull, ch.Address())
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if len(pinnedChunks) < 10 {
```
acud, line 558: pinned chunks length here is always zero, why the check?

pkg/localstore/gc_test.go
```diff
@@ -534,3 +535,194 @@ func TestSetTestHookCollectGarbage(t *testing.T) {
...
+
+func TestPinAfterMultiGC(t *testing.T) {
+	db := newTestDB(t, &Options{
+		Capacity: 100,
+	})
+
+	pinnedChunks := make([]swarm.Address, 0)
+
+	// upload random chunks above db capacity to see if chunks are still pinned
+	for i := 0; i < 2000; i++ {
```
acud, line 547: if the capacity is 100, why do you need to insert 6000?

pkg/localstore/localstore_test.go
```diff
@@ -567,6 +567,6 @@ func TestDBDebugIndexes(t *testing.T) {
 	}
 
 	// assert that there's a pin and gc exclude entry now
-	testIndexCounts(t, 1, 1, 1, 1, 1, 1, 1, indexCounts)
+	testIndexCounts(t, 1, 1, 0, 1, 1, 1, 1, indexCounts)
```
acud, line 570: why has this been changed?

pkg/localstore/gc_test.go
```diff
@@ -534,3 +535,194 @@ func TestSetTestHookCollectGarbage(t *testing.T) {
...
+	t.Run("pin Index count", newItemsCountTest(db.pinIndex, len(pinnedChunks)))
+
+	// Check if all the pinned chunks are present in the data DB
+	for _, addr := range pinnedChunks {
+		outItem := shed.Item{
+			Address: addr.Bytes(),
+		}
+		gotChunk, err := db.Get(context.Background(), storage.ModeGetRequest, swarm.NewAddress(outItem.Address))
+		if err != nil {
+			fmt.Println("Pinned chunk missing ", addr)
```
acud, line 595: can you please remove all `fmt.Println` references? If there's a reason to annotate an error please use `t.Fatalf("something has occured %v" , err)`

pkg/localstore/gc_test.go
```diff
@@ -534,3 +535,194 @@ func TestSetTestHookCollectGarbage(t *testing.T) {
...
+	if err != nil {
+		t.Fatal("Pinned chunk missing ", err)
+	}
+	if !gotChunk.Address().Equal(rch.Address()) {
+		t.Fatal("Pinned chunk is not equal to got chunk")
+	}
+
+}
+
+func addRandomChunks(t *testing.T, count int, db *DB) {
```
acud, line 707: why don't you use this in all of the tests above?

acud:
> Hope you are happy now

You don't need to please me @jmozah. I want to know what I'm approving. Having a broad problem space that is difficult to define makes such PRs very problematic. Please note now that you're adding extra DB operations for every operation that we're doing. So instead of understanding what's causing this in the first place, we're taking the approach of blindly solving it by adding more redundant operations which will obviously slow down every db op.

pkg/localstore/gc_test.go
jmozah: Basically the idea is to add 200 chunks out of which 10 chunks needs to be pinned. The `pinnedChunks` is appended inside the loop, so the length increases. So it is not always zero.

pkg/localstore/gc_test.go
jmozah: That's the test case,...  The issue occurs when we bombard chunks greater that `dbCapacity` ..more than once...  2000 is a big number for CI, May be we can reduce it to 200.

pkg/localstore/localstore_test.go
jmozah: Previousl when you setAccesss, we add the chunk to gcIndex even of it is pinned. Now for a pinned chunk, even after a setAccess, the gcIndex should not be incremented as we have added our check their.

pkg/localstore/gc_test.go
jmozah: Because other tests dont need this.

jmozah:
> You don't need to please me @jmozah.

That was a joke. :-)

> So instead of understanding what's causing this in the first place, we're taking the approach of blindly solving it by adding more redundant operations which will obviously slow down every db op.

That not true, sorry. if you don't understand the fix, I can shepherd you through the problem and the fix.


>It is also unclear to me how is it, that chunks that are pinned are not cleared out of the gc index when the exclude trigger is hit within the garbage collection run. If I would put my money on something it would be that something there is broken (either some shed.Item does not have the right fields passed to the gc for the exclude index or some other peculiarity. If the item is in the exclude index too and gets GCd then we have some problem around that area).

Actually they will get cleared from `exchudeIndex` once they come there. Then after that the pinned chunk should not  come to `gcIndex` (we have the same condition in `ModeSetUpload`), that was the design. This is to avoid doin the `pinIndex.Has()` in every GC, For now this is the logical fix. I get that the performance will degrade, but in this release we are not not optimizing for the performance.. are we?

pkg/localstore/gc_test.go
```diff
@@ -534,3 +535,193 @@ func TestSetTestHookCollectGarbage(t *testing.T) {
...
+	if err != nil {
+		fmt.Println("Pinned chunk missing ", rch.Address())
+		t.Fatal(err)
+	}
+
+	addRandomChunks(t, 20, db)
+
+	_, err = db.Get(context.Background(), storage.ModeGetRequest, rch.Address())
+	if err != nil {
+		fmt.Println("Pinned chunk missing ", rch.Address())
```
acud, line 691: Again and again we ask to remove `fmt.Println` from the codebase. Can we please agree that you review your own code and get rid of such trivial mistakes that constant come up in every review?

jmozah:
>you're claiming that the chunks should not be in the garbage collection index if they are in the pin index, but that is not correct according to the old design. 

No. That is how we designed. I would be very happy if you could help me remember what the design was. You can look at the original [places](https://github.com/ethersphere/bee/blob/master/pkg/localstore/mode_set.go#L318) we did this. but left out in few other places, which is biting us now.

> That is why the gcExcludeIndex index was introduced, so that you could do a gcIndex ¬ gcExcludeIndex then GC.

`gcExcludeIndex` was introduced for another reason altogether. You first upload a file and then pin it in the code. If the chunk enters in to `gcIndex` before pinning actually happens, then `gcExcludeIndex` would come handy and will remove the item from `gcIndex` just before `GC` is done. You can also check in the code that once the chunk is removed from the `gcIndex`, it will also be [removed](https://github.com/ethersphere/bee/blob/master/pkg/localstore/gc.go#L200) from the `gcExclude` Index . So this is basically 1 time saving mechanism that we built. **After that, `gcIndex` is not supposed to have a pinned chunk by design.** So basically after after one round the `gcExcludeIndex` will be empty. `gcExcludeIndex` is only for the just uploaded chunks, `pinIndex` is for all chunks ever pinned. This [comment](https://github.com/ethersphere/bee/blob/master/pkg/localstore/gc.go#L96) might remind you. Also this is the @janos [comment](https://github.com/ethersphere/swarm/pull/1509#discussion_r308301381) that triggered the `gcExcludeIndex` design.

> The changes that you've done in this PR basically mean that gcExcludeIndex is no longer necessary. You can try to remove it completely and see if the problem persists, since in that case we would be doing a significant amount of redundant operations.

As said above, `gcExcludeIndex` is necessary. otherwise we might loose some chunks that enter `GC` between uploading and pinning the file. 


> Did you check that the chunks that are missing also exist in the exclude gc index? Their presence in the pin index before the fix is one thing, but their protection from being garbage collected is secured with their presence in the exclude index.

`gcExcludeIndex` becomes empty after one round. So there is no protection after the first round of protection from `gcExcludeIndex`.


> Another question about the tests that you've introduced here: do they fail persistently without the fix? do they flake? if yes by which ratio?

They always pass, as i said i could not reproduce this in unit tests cases here. The same scenario creates issue in @santicomp2014's  setup. So i added it as a precaution.

pkg/localstore/gc_test.go
jmozah: oops... my bad..  will remove it

acud:
So then the problem is clear no? If pinning the chunk is two-phased operation then there will always be a possibility for a data race where a GC runs between the upload phase and the pinning phase. Solving this thus must be an atomic operation where we can say that the chunk upload API can take a header saying to pin the chunk immediately on upload, and then create a `ModePutUploadPin` which takes care of adding the item into the pin index, gc exclude index and _not_ to the GC index.

And to emphasize, there is still a possibility for a data race even with this fix, where the chunk is added to the gc without it being in the pin index, this can happen when the chunk is accessed before it is pinned (you are relying on the fact that it is in the pin index when you `setAccess` on Get, so if it not yet pinned it will get garbage collected). This fix narrows the window of this possible data race, but does not eliminate it entirely.

acud:
I'm also willing to bet a few bucks that the problem is still there, and that it will show itself again when you try to upload and pin a large collection (let's say a large directory with a lot of files, that will take a bit longer to traverse for pinning)

jmozah:
> So then the problem is clear no? If pinning the chunk is two-phased operation then there will always be a possibility for a data race where a GC runs between the upload phase and the pinning phase.

Pinning is a two phased operation always. You can upload using a API  and later call a another API to pin.  That's the design at that time (although i wont do it if i do it in bee). That does not mean there will be a race condition. That means that there can be chunks that are going to be GC'd and we need to remove them from gcIndex.

> And to emphasize, there is still a possibility for a data race even with this fix, where the chunk is added to the gc without it being in the pin index, this can happen when the chunk is accessed before it is pinned

That's precisely the reason of gcExcludeIndex. You still don't seem to get the gcExcludeIndex sole existence.

> This fix narrows the window of this possible data race, but does not eliminate it entirely.

Can you explain the data race? where is the data race? without knowing here is the data race how are you claiming that it is narrowing the possibility.


>I'm also willing to bet a few bucks that the problem is still there, and that it will show itself again when you try to upload and pin a large collection (let's say a large directory with a lot of files, that will take a bit longer to traverse for pinning)

I am also of the same impression that the problem might still be there. until we know the issue exactly, we cannot say for sure that it is fixed. Nor i am claiming that this will fix it 100%. This is a guard fix that needs to be there logically as per the design and it is not here. This PR adds it. Got it.

Issue #258, opened 2020-06-05
Pushsync not working since tagging chunks merge


svetomir:
Pushsync not working since tagging chunks merge.

Build job failed: https://github.com/ethersphere/bee/runs/738397906?check_suite_focus=true

Logs:
`level=debug msg="pusher: get tag by address 0c2658e6324f889bdbffba7cbcbcc53a8ef28baef799d8ed762df5e0dfc310ec: tag not found"`
`level=debug msg="pusher: get tag by address a1f5f6feaa1f60052e2a9d406c6b9357511498fca993f743f4cf641fdea377e3: tag not found"`
`level=debug msg="pusher: get tag by address d7079e0eeb2f11dbd02ef3282dc8961fc961a47767467d71577bb576760eee75: tag not found"`
`level=debug msg="pusher: get tag by address 23b0f4d1f4e75d8f85aa5eba56163051fe93d9057388f1cd1c5d8d3ca9a136b4: tag not found"`
`level=debug msg="pusher: get tag by address 0c2658e6324f889bdbffba7cbcbcc53a8ef28baef799d8ed762df5e0dfc310ec: tag not found"`

Goroutine dump is attached.
[dump.txt](https://github.com/ethersphere/bee/files/4734904/dump.txt)

Command to replicate issue:
`beekeeper check pushsync -u 3 -p 2`

PR #260, opened 2020-06-05
Guard pusher against tags that are not present

jmozah:
bzz api does not implement splitter and tags yet, but pusher assumed that tags are present for all addresses and fails because of this. This PR put skips this condition until the bzz api implements splitter and then tags.

fixes https://github.com/ethersphere/bee/issues/258
Note: This should be temporary to make cluster tests pass... after bzz api implements splutter and tags, we should remove this guard fix.

### DIFF
--- a/pkg/pusher/pusher_test.go
+++ b/pkg/pusher/pusher_test.go
@@ -7,11 +7,12 @@ package pusher_test
 import (
 	"context"
 	"errors"
-	"github.com/ethersphere/bee/pkg/tags"
 	"io/ioutil"
 	"sync"
 	"testing"
 	"time"
+
+	"github.com/ethersphere/bee/pkg/tags"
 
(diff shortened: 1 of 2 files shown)


pkg/pusher/pusher_test.go
```diff
@@ -89,6 +89,50 @@ func TestSendChunkToPushSync(t *testing.T) {
...
+
+		err = checkIfModeSet(chunk.Address(), storage.ModeSetSyncPush, storer)
+		if err == nil {
+			break
+		}
+	}
+	if err != nil {
+		t.Fatal(err)
+	}
+	p.Close()
```
acud, line 133: please convert to `defer` statement

jmozah: This is done to make sure that we close the usher before the storer. Otherwise we Add on the Wait in the storer.Close(). remember, we talked in length about this.

PR #266, opened 2020-06-07
API for upload and download of multi-chunk data

nolash:


pkg/api/raw.go
```diff
@@ -0,0 +1,108 @@
...
+	ctx := r.Context()
+	responseObject, err := s.splitUpload(ctx, r.Body, r.ContentLength)
+	if err != nil {
+		jsonhttp.BadGateway(w, err.Error())
+		return
+	}
+	jsonhttp.OK(w, responseObject)
+}
+
+func (s *server) splitUpload(ctx context.Context, r io.ReadCloser, l int64) (interface{}, error) {
```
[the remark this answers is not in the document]
nolash: the idea was the @jmozah could use it for the bzz uploader

pkg/file/splitter/internal/job.go
```diff
@@ -60,7 +60,7 @@ func NewSimpleSplitterJob(ctx context.Context, putter storage.Putter, spanLength
 		sumCounts:  make([]int, levelBufferLimit),
 		cursors:    make([]int, levelBufferLimit),
 		hasher:     bmtlegacy.New(p),
-		buffer:     make([]byte, file.ChunkWithLengthSize*levelBufferLimit),
+		buffer:     make([]byte, file.ChunkWithLengthSize*levelBufferLimit*2),
```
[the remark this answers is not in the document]
nolash: It's a temporary workaround for a mistaken calculation that writes will never exceed single chunk per level. There's an issue for this here: https://github.com/ethersphere/bee/issues/267

pkg/api/raw.go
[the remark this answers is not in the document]
janos: Then, the log messages here would not be correct (they have `raw: ` prefix) and my suggestion is even more valid. This method can be written in a way to return annotated errors and a response that can be reused in multiple handers.

pkg/file/splitter/internal/job.go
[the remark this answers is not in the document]
janos: A comment on that in the code would be very helpful.

pkg/api/raw_test.go
```diff
@@ -8,30 +8,40 @@ import (
...
-		resource   = "/bzz"
-		content    = []byte("foo")
-		expHash    = "2387e8e7d8a48c2a9339c97c1dc3461a9a7aa07e994c5cb8b38fd7c1b3e6ea48"
+		resource   = "/bzz-raw"
+		expHash    = "29a5fb121ce96194ba8b7b823a1f9c6af87e1791f824940a53b5a7efe3f790d9"
 		mockStorer = mock.NewStorer()
 		client     = newTestServer(t, testServerOptions{
 			Storer: mockStorer,
+			Tags:   tags.NewTags(),
+			Logger: logging.New(os.Stderr, 5),
```
acud, line 34: change to `ioutil.Discard`

pkg/file/buffer_test.go
```diff
@@ -0,0 +1,70 @@
...
+		c, err = buf.Read(data)
+		if c != 2 {
+			errC <- fmt.Errorf("read expected 2, got %d", c)
+		}
+		if err != nil {
+			errC <- err
+		}
+		errC <- nil
+	}()
+	data := [swarm.ChunkSize - 2]byte{}
```
acud, line 40: 1. this is extremely ugly :) can you please switch to `data := make([]byte,ChunkSize-2)`/`var data [ChunkSize-2]byte`?
2. why the 2?

pkg/api/raw.go
```diff
@@ -0,0 +1,108 @@
...
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/octet-stream")
+	w.Header().Set("Content-Length", fmt.Sprintf("%d", dataSize))
+	err = file.JoinReadAll(j, address, w)
+	if err != nil {
+		s.Logger.Errorf("raw: data write %s: %v", address, err)
+		s.Logger.Error("raw: data input error")
+		jsonhttp.BadRequest(w, "failed to retrieve data")
```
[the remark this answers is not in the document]
nolash: I haven't looked down the stack on this and how things are buffered. I guess the client will be simply faced with a "server premature connection close" if things go wrong. Probably ok, but as an end user I remember this as a bit of a pain. We should probably have solid logging here down the road.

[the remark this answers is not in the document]
janos: This will not close the connection. The issue is only with served content and go net/http refusing to write another header since one is already written.

pkg/api/raw.go
```diff
@@ -0,0 +1,107 @@
...
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/octet-stream")
+	w.Header().Set("Content-Length", fmt.Sprintf("%d", dataSize))
+	err = file.JoinReadAll(j, address, w)
+	if err != nil {
+		s.Logger.Errorf("raw: data write %s: %v", address, err)
+		s.Logger.Error("raw: data input error")
+		return
```
janos, line 105: No need to return.

If no bytes are written to `w` (file.JoinReadAll returning an error before writing to writer), the client will receive an empty response. In that case, we should return a json response with internal server error. This can be left for later to be solved as I think that it will prolong this PR.

Additionally, Content-Type and Content-Length should be written only if file.JoinReadAll does not return an error before it writes to the writer.

nolash: both of your points describe something that means `JoinReadAll` has to change, which would hold up the PR...

janos: Actually, JoinReadAll does not have to change. There are multiple ways to implement changes from my comments.

nolash: so you mean wrap the writer _again_ and see if it's 0 or not?

nolash:
changed back to `in progress` because of `JoinReadAll` discussion.

pkg/api/raw.go
nolash: I'll just change readall to return read count

janos: > so you mean wrap the writer again and see if it's 0 or not?

This writer is not wrapped at all. Write method is called directly in JoinReadAll. Anyways, if the functionality is broken, why not to wrap.

PR #270, opened 2020-06-08
Empty overlay fix in peer registry

pradovic:


acud:
how is this possible? where have you seen this?

pkg/p2p/libp2p/peer.go
```diff
@@ -124,7 +124,7 @@ func (r *peerRegistry) remove(peerID libp2ppeer.ID) {
 	delete(r.connections, peerID)
 	r.mu.Unlock()
 
-	if r.disconnecter != nil {
+	if r.disconnecter != nil && !overlay.IsZero() {
```
acud, line 127: @pradovic can you please document in a comment one line above when is `overlay` zero?

janos: I know that I have suggested this change to @pradovic, but maybe it would be more correct to check `overlay, ok := r.overlays[peerID]; if !ok {return}` (with lock handling), instead for checking for a zero address. And yes, comment would be useful.

pradovic: Done already, or in about the same time :)

PR #281, opened 2020-06-10
Overlay from eth addr

janos:
This PR implements overlay address hashing explicit from ethereum address as described in the book of swarm. It also updates the file keystore to use version 3 so that it can be compatible with ethereum json key v3 and sets the ethereum address in that file. It is not possible to set the swarm overlay address as it depends on the network id. Ethereum address is constructed as it is in github.com/ethereum/go-ethereum/crypto.PubkeyToAddress, just with simpler code.


pkg/crypto/crypto.go
```diff
@@ -44,3 +44,17 @@ func DecodeSecp256k1PrivateKey(data []byte) (*ecdsa.PrivateKey, error) {
 	privk, _ := btcec.PrivKeyFromBytes(btcec.S256(), data)
 	return (*ecdsa.PrivateKey)(privk), nil
 }
+
+// NewEthereumAddress returns a bunary representation of ethereum blockchain address.
```
acud, line 48: bunary -> binary

PR #282, opened 2020-06-10
File upload with metadata

jmozah:
This PR should do the following
- Create  POST http://host:port/bzz-file api to upload a file
- Create GET http://host:port/bzz-file/{address} to download a file with the filename uploaded

See the user story for more details https://app.zenhub.com/workspaces/swarm-10-5e7c6d9966b97049231c3c0b/issues/ethersphere/bee/274

### DIFF
--- a/pkg/api/bytes.go
+++ b/pkg/api/bytes.go
@@ -11,14 +11,13 @@ import (
 	"io"
 	"net/http"
 
-	"github.com/gorilla/mux"
-
 	"github.com/ethersphere/bee/pkg/file"
 	"github.com/ethersphere/bee/pkg/file/joiner"
 	"github.com/ethersphere/bee/pkg/file/splitter"
 	"github.com/ethersphere/bee/pkg/jsonhttp"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/gorilla/mux"
 )
 
 type bytesPostResponse struct {
@@ -31,8 +30,10 @@ func (s *server) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 	responseObject, err := s.splitUpload(ctx, r.Body, r.ContentLength)
 	if err != nil {
 		s.Logger.Debugf("bytes upload: %v", err)
-		o := responseObject.(jsonhttp.StatusResponse)
-		jsonhttp.Respond(w, o.Code, o)
+		var response jsonhttp.StatusResponse
+		response.Message = "upload error"
+		response.Code = http.StatusInternalServerError
+		jsonhttp.Respond(w, response.Code, response)
 	} else {
 		jsonhttp.OK(w, responseObject)
 	}
@@ -61,11 +62,7 @@ func (s *server) splitUpload(ctx context.Context, r io.ReadCloser, l int64) (int
 	}()
 	sp := splitter.NewSimpleSplitter(s.Storer)
 	address, err := sp.Split(ctx, chunkPipe, l)
-	var response jsonhttp.StatusResponse
 	if err != nil {
-		response.Message = "upload error"
-		response.Code = http.StatusInternalServerError
-		err = fmt.Errorf("%s: %v", response.Message, err)
 		return swarm.ZeroAddress, err
 	}
 	return bytesPostResponse{Reference: address}, nil
--- a/pkg/api/bytes.go
+++ b/pkg/api/bytes.go
@@ -28,17 +28,17 @@ type bytesPostResponse struct {
 // bytesUploadHandler handles upload of raw binary data of arbitrary length.
 func (s *server) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
-	address, err := s.splitUpload(ctx, r.Body, r.ContentLength)
+	responseObject, err := s.splitUpload(ctx, r.Body, r.ContentLength)
 	if err != nil {
 		s.Logger.Debugf("bytes upload: %v", err)
 		o := responseObject.(jsonhttp.StatusResponse)
 		jsonhttp.Respond(w, o.Code, o)
 	} else {
-		jsonhttp.OK(w, &rawPostResponse{Hash: address})
+		jsonhttp.OK(w, responseObject)
 	}
 }
 
-func (s *server) splitUpload(ctx context.Context, r io.ReadCloser, l int64) (swarm.Address, error) {
+func (s *server) splitUpload(ctx context.Context, r io.ReadCloser, l int64) (interface{}, error) {
 	chunkPipe := file.NewChunkPipe()
 	go func() {
 		buf := make([]byte, swarm.ChunkSize)
--- a/pkg/api/bytes.go
+++ b/pkg/api/bytes.go
@@ -28,17 +28,17 @@ type bytesPostResponse struct {
 // bytesUploadHandler handles upload of raw binary data of arbitrary length.
 func (s *server) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
-	responseObject, err := s.splitUpload(ctx, r.Body, r.ContentLength)
+	address, err := s.splitUpload(ctx, r.Body, r.ContentLength)
 	if err != nil {
 		s.Logger.Debugf("bytes upload: %v", err)
 		o := responseObject.(jsonhttp.StatusResponse)
 		jsonhttp.Respond(w, o.Code, o)
 	} else {
-		jsonhttp.OK(w, responseObject)
+		jsonhttp.OK(w, &rawPostResponse{Hash: address})
 	}
 }
 
-func (s *server) splitUpload(ctx context.Context, r io.ReadCloser, l int64) (interface{}, error) {
+func (s *server) splitUpload(ctx context.Context, r io.ReadCloser, l int64) (swarm.Address, error) {
 	chunkPipe := file.NewChunkPipe()
 	go func() {
 		buf := make([]byte, swarm.ChunkSize)
@@ -66,7 +66,7 @@ func (s *server) splitUpload(ctx context.Context, r io.ReadCloser, l int64) (int
 		response.Message = "upload error"
 		response.Code = http.StatusInternalServerError
 		err = fmt.Errorf("%s: %v", response.Message, err)
-		return response, err
+		return swarm.ZeroAddress, err
 	}
 	return bytesPostResponse{Reference: address}, nil
 }
--- a/pkg/api/bzzfile.go
+++ b/pkg/api/bzzfile.go
@@ -226,18 +226,20 @@ func (s *server) storeMeta(ctx context.Context, dataBytes []byte) (swarm.Address
 	dataBuf := bytes.NewBuffer(dataBytes)
 	dataReader := io.LimitReader(dataBuf, int64(len(dataBytes)))
 	dataReadCloser := ioutil.NopCloser(dataReader)
-	dataAddr, err := s.splitUpload(ctx, dataReadCloser, int64(len(dataBytes)))
+	o, err := s.splitUpload(ctx, dataReadCloser, int64(len(dataBytes)))
 	if err != nil {
 		return swarm.ZeroAddress, err
 	}
-	return dataAddr, nil
+	bytesResp := o.(bytesPostResponse)
+	return bytesResp.Reference, nil
 }
 
 // storePartData stores file data belonging to one of the part of multipart.
 func (s *server) storePartData(ctx context.Context, part *multipart.Part, l uint64) (swarm.Address, error) {
-	dataAddr, err := s.splitUpload(ctx, part, int64(l))
+	o, err := s.splitUpload(ctx, part, int64(l))
 	if err != nil {
 		return swarm.ZeroAddress, err
 	}
-	return dataAddr, nil
+	bytesResp := o.(bytesPostResponse)
+	return bytesResp.Reference, nil
 }
--- a/pkg/api/bzzfile.go
+++ b/pkg/api/bzzfile.go
@@ -1,0 +1,243 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package api
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"mime"
+	"mime/multipart"
+	"net/http"
+	"strconv"
+
+	"github.com/ethersphere/bee/pkg/collection/entry"
+	"github.com/ethersphere/bee/pkg/file"
+	"github.com/ethersphere/bee/pkg/file/joiner"
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/gorilla/mux"
+)
+
+const (
+	MultiPartFormData = "multipart/form-data"
+)
+
+type FileUploadResponse struct {
+	Reference swarm.Address `json:"reference"`
+}
+
+// bzzFileUploadHandler uploads the file and its metadata supplied as a multipart http message.
+func (s *server) bzzFileUploadHandler(w http.ResponseWriter, r *http.Request) {
+	contentType, params, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
+	if contentType != MultiPartFormData {
+		s.Logger.Debugf("bzz-file: no mutlipart: %v", err)
+		s.Logger.Error("bzz-file: no mutlipart")
+		jsonhttp.BadRequest(w, "not a mutlipart/form-data")
+		return
+	}
+
+	mr := multipart.NewReader(r.Body, params["boundary"])
+	for {
+		part, err := mr.NextPart()
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			s.Logger.Debugf("bzz-file: read mutlipart: %v", err)
+			s.Logger.Error("bzz-file: read mutlipart")
+			jsonhttp.BadRequest(w, "error reading a mutlipart/form-data")
+			return
+		}
+
+		ctx := context.Background()
+
+		// find file size
+		fileSizeString := part.Header.Get("Content-Length")
+		if fileSizeString == "" {
+			s.Logger.Debugf("bzz-file: content length: %v", err)
+			s.Logger.Error("bzz-file: content length")
+			jsonhttp.BadRequest(w, "content length header missing")
+			return
+		}
+		fileSize, err := strconv.ParseUint(fileSizeString, 10, 64)
+		if err != nil {
+			s.Logger.Debugf("bzz-file: content length: %v", err)
+			s.Logger.Error("bzz-file: content length")
+			jsonhttp.BadRequest(w, "error parsing content length")
+			return
+		}
+
+		// first store the file and get its reference
+		fr, err := s.storePartData(ctx, part, fileSize)
+		if err != nil {
+			s.Logger.Debugf("bzz-file: file store: %v,", err)
+			s.Logger.Error("bzz-file: file store")
+			jsonhttp.InternalServerError(w, "could not store file data")
+			return
+		}
+
+		// try to find filename
+		// 1) in part header params
+		// 2) as formname
+		// 3) file reference hash
+		fileName := part.FileName()
+		if fileName == "" {
+			fileName = part.FormName()
+		}
+		if fileName == "" {
+			fileName = fr.String()
+		}
+		m := entry.NewMetadata(fileName)
+
+		// then find out content type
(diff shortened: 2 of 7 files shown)


jmozah:
@janos  @nolash  I have done most of the part except testing upload with multiple files. If you insist i will do that, but since that was not a requirement i am submitting for a review. Also the Coverage report failing looks like a false alarm.

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,214 @@
...
+	MultiPartFormData = "multipart/form-data"
+)
+
+type FileUploadResponse struct {
+	Address swarm.Address `json:"address"`
+}
+
+func (s *server) bzzFileUploadHandler(w http.ResponseWriter, r *http.Request) {
+	contentType, params, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
+	if contentType != MultiPartFormData {
```
nolash, line 37: Support for application/octet-stream upload will be added in later PR?

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,214 @@
...
+
+		cd := part.Header.Get("Content-Disposition")
+		_, params, err := mime.ParseMediaType(cd)
+		if err != nil {
+			s.Logger.Debugf("bzz-file: parse mediatype: %v", err)
+			s.Logger.Error("bzz-file: parse mediatype")
+			jsonhttp.BadRequest(w, "error parsing parse mediatype")
+			return
+		}
+		fileName := params["name"]
```
nolash, line 64: I see there is a method called `FileName` in `mime/multipart`. Would that be useful here?

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,214 @@
...
+		if fileName == "" {
+			fileName = part.FormName()
```
nolash, line 66:
```
func (p *Part) FormName() string
    FormName returns the name parameter if p has a Content-Disposition of type
    "form-data". Otherwise it returns the empty string.
```

It seems the filename can still be an empty string after this. The `bee-file` binary sets the filename to the hex string of the swarm address when no other filename has been set. Should we aim for the same behavior here?

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		// set up splitter to process the data
+		splitr := splitter.NewSimpleSplitter(s.Storer)
+		ctx := context.Background()
+
+		// first store the file and get the reference
+		buf, err := ioutil.ReadAll(part)
```
nolash, line 74: Isn't it better to pass this directly to the splitter instead of this extra buffering?

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,214 @@
...
+		if err != nil {
+			s.Logger.Debugf("bzz-file: read part: %v", err)
+			s.Logger.Error("bzz-file: read part ")
+			jsonhttp.BadRequest(w, "read a part data")
+			return
+		}
+		fr, err := s.storeData(ctx, splitr, buf)
```
nolash, line 81: I prepared the `pkg/api/raw.go:splitUpload` so it would be re-usable. Was it not good enough since you decided to write new code for this?

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,214 @@
...
+			s.Logger.Error("bzz-file: file store")
+			jsonhttp.InternalServerError(w, "could not store metadata")
+			return
+		}
+
+		// then create and store metadata.
+		m := entry.NewMetadata(fileName)
+		contentType := part.Header.Get("Content-Type")
+		if contentType == "" {
+			contentType = http.DetectContentType(buf)
```
nolash, line 93: This magic is the reason why you do the extra buffer? I am (still) of the opinion that this kind of magic is unnecessary for a lean bee node.

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,214 @@
...
+		if err != nil {
+			s.Logger.Debugf("bzz-file: metadata marshall: %v, file name %s", err, fileName)
+			s.Logger.Error("bzz-file: metadata marshall")
+			jsonhttp.InternalServerError(w, "metadata marshall error")
+			return
+		}
+		mr, err := s.storeData(ctx, splitr, metadataBytes)
+		if err != nil {
+			s.Logger.Debugf("bzz-file: metadata store: %v, file name %s", err, fileName)
+			s.Logger.Error("bzz-file: metadata store ")
```
nolash, line 106: space

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,214 @@
...
+
+const (
+	MultiPartFormData = "multipart/form-data"
+)
+
+type FileUploadResponse struct {
+	Address swarm.Address `json:"address"`
+}
+
+func (s *server) bzzFileUploadHandler(w http.ResponseWriter, r *http.Request) {
```
nolash, line 35: please describe the methods

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,214 @@
...
+	outBuffer := bytes.NewBuffer(nil)
+	_, err = file.JoinReadAll(j, e.Reference(), outBuffer)
+	if err != nil {
+		s.Logger.Debugf("bzz-file: data read %s: %v", addr, err)
+		s.Logger.Error("bzz-file: data read")
+		jsonhttp.InternalServerError(w, "error reading data")
+		return
+	}
+	w.Header().Set("ETag", fmt.Sprintf("%q", e.Reference()))
+	w.Header().Set("Content-Disposition", fmt.Sprintf("inline; filename=\"%s\"", metaData.Filename))
```
nolash, line 200: Would it be an idea to control through request query whether disposition should be inline or attachment?

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,214 @@
...
+	_, _ = io.Copy(w, outBuffer)
+}
+
+func (s *server) storeData(ctx context.Context, splitr file.Splitter, dataBytes []byte) (swarm.Address, error) {
+	dataBuf := bytes.NewBuffer(dataBytes)
+	dataReader := io.LimitReader(dataBuf, int64(len(dataBytes)))
+	dataReadCloser := ioutil.NopCloser(dataReader)
+	dataAddr, err := splitr.Split(ctx, dataReadCloser, int64(len(dataBytes)))
+	if err != nil {
+		return swarm.NewAddress([]byte{0}), err
```
nolash, line 211: Please use `swarm.ZeroAddress`

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/gorilla/mux"
+)
+
+const (
+	MultiPartFormData = "multipart/form-data"
+)
+
+type FileUploadResponse struct {
+	Address swarm.Address `json:"address"`
```
nolash, line 32: I think it should be named `Reference` rather than `Address`

pkg/api/bzzfile_test.go
```diff
@@ -0,0 +1,107 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		if params["filename"] != fileName {
+			t.Fatal("Invalid file name detected")
+		}
+	})
+
+	t.Run("upload-then-download-and-check-data", func(t *testing.T) {
+		fileName := "vacation.mp4"
```
nolash, line 70: mp4 but html content? I don't understand..

pkg/api/raw.go
```diff
@@ -66,9 +66,9 @@ func (s *server) splitUpload(ctx context.Context, r io.ReadCloser, l int64) (int
 		response.Message = "upload error"
 		response.Code = http.StatusInternalServerError
 		err = fmt.Errorf("%s: %v", response.Message, err)
-		return response, err
+		return swarm.NewAddress([]byte{0}), err
```
nolash, line 69: Please use `swarm.ZeroAddress`

pkg/collection/entry/metadata.go
```diff
@@ -21,6 +21,10 @@ func NewMetadata(fileName string) *Metadata {
 	}
 }
 
+func (m *Metadata) SetMimeType(mineType string) {
```
nolash, line 24: IIRC I was already reprimanded for adding this setter to a simple exported type.

pkg/jsonhttp/jsonhttptest/jsonhttptest.go
```diff
@@ -35,6 +39,80 @@ func ResponseDirect(t *testing.T, client *http.Client, method, url string, body
 	}
 }
 
+func ResponseDirectWithMultiPart(t *testing.T, client *http.Client, method, url, fileName string, data []byte,
+	responseCode int, response interface{}) http.Header {
+	body := &bytes.Buffer{}
```
nolash, line 44: `NewBuffer`?

pkg/jsonhttp/jsonhttptest/jsonhttptest.go
```diff
@@ -35,6 +39,80 @@ func ResponseDirect(t *testing.T, client *http.Client, method, url string, body
...
+	mw := multipart.NewWriter(body)
+	hdr := make(textproto.MIMEHeader)
+	hdr.Set("Content-Disposition", fmt.Sprintf("form-data; name=%q", fileName))
+	hdr.Set("Content-Type", http.DetectContentType(data))
+	hdr.Set("Content-Length", strconv.FormatInt(int64(len(data)), 10))
+	part, err := mw.CreatePart(hdr)
+	if err != nil {
+		t.Error(err)
+	}
+	_, err = io.Copy(part, ioutil.NopCloser(bytes.NewReader(data)))
```
nolash, line 54: Why do you need the `NopCloser` enhancement here?

pkg/jsonhttp/jsonhttptest/jsonhttptest.go
```diff
@@ -35,6 +39,80 @@ func ResponseDirect(t *testing.T, client *http.Client, method, url string, body
...
+	err = mw.Close()
+	if err != nil {
+		t.Error(err)
+	}
+
+	req, err := http.NewRequest(method, url, body)
+	if err != nil {
+		t.Error(err)
+	}
+	//req.Header.Set("Expect", "100-continue")
```
nolash, line 67: Please remove commented code

janos:
@jmozah could you update this branch with the recent changes from the master? We have update the API router.

pkg/api/router.go
```diff
@@ -28,6 +28,14 @@ func (s *server) setupRouting() {
...
 	})
 
+	router.Handle("/bzz-file", jsonhttp.MethodHandler{
+		"POST": http.HandlerFunc(s.bzzFileUploadHandler),
+	})
+
+	router.Handle("/bzz-file/{addr}", jsonhttp.MethodHandler{
+		"GET": http.HandlerFunc(s.bzzFileDownloadHandler),
+	})
+
```
janos, line 38: Could you update this branch to the latest master and use the handle function from there as:

```
	handle(router, "/files", jsonhttp.MethodHandler{
		"POST": http.HandlerFunc(s.fileUploadHandler),
	})
	handle(router, "/files/{address}", jsonhttp.MethodHandler{
		"GET": http.HandlerFunc(s.fileGetHandler),
	})
```

pkg/api/bzzfile.go
jmozah: I will add this in a new PR @nolash

pkg/api/bzzfile.go
jmozah: Good thought. For now i am hard coding it so that the browser can show it. later we can make it based on the query.

pkg/api/bzzfile_test.go
jmozah: well.. the detection is based on content.. filename makes no difference.. for easy reading i am making it sample.html :-)

jmozah:
> @jmozah could you update this branch with the recent changes from the master? We have updated the API router.

@janos Updated and fixed the review comments

pkg/api/bzzfile.go
nolash: still space here

pkg/api/bytes.go
```diff
@@ -66,7 +66,7 @@ func (s *server) splitUpload(ctx context.Context, r io.ReadCloser, l int64) (int
 		response.Message = "upload error"
 		response.Code = http.StatusInternalServerError
 		err = fmt.Errorf("%s: %v", response.Message, err)
-		return response, err
+		return swarm.ZeroAddress, err
```
janos, line 69: The `response` variable is not used, it can be removed, also the fmt.Errorf.

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,245 @@
...
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			s.Logger.Debugf("bzz-file: read mutlipart: %v", err)
+			s.Logger.Error("bzz-file: read mutlipart")
+			jsonhttp.BadRequest(w, "error reading a mutlipart/form-data")
+			return
+		}
+
+		ctx := context.Background()
```
janos, line 58: Use r.Context().

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		if fileName == "" {
+			fileName = fr.String()
+		}
+		m := entry.NewMetadata(fileName)
+
+		// then find out content type
+		contentType := part.Header.Get("Content-Type")
+		if contentType == "" {
+			reader := bufio.NewReader(part)
```
janos, line 101: `part` is already read and here cannot be read any further as it reached EOF. The whole content type detection has to be done before any Read call on `part`. Basically, before storePartData call, the whole code block from line 85 to line 111 has to be moved, with one important difference. If the bufio.Reader is constructed it has to be passed to storePartData, not directly the `part` variable as `part` reader already moved its cursor, while the bufio reader did not, it just peeked, so it can read from the start, as `part` reader has advanced 512 bytes. The best way to do that is to create `var reader io.Reader = part` before `if contentType == "" {` line and here to do `reader = bufio.NewReader(part)` (assign to the variable, not create a new one), and use that `reader` variable in storePartData `fr, err := s.storePartData(ctx, reader, fileSize)`.

It looks to me that `contentType == ""` block is not tested and it would be good to have that test.

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// send the file data back in the response
+	outBuffer := bytes.NewBuffer(nil)
+	_, err = file.JoinReadAll(j, e.Reference(), outBuffer)
+	if err != nil {
+		s.Logger.Debugf("bzz-file: data read %s: %v", addr, err)
+		s.Logger.Error("bzz-file: data read")
+		jsonhttp.InternalServerError(w, "error reading data")
+		return
+	}
```
janos, line 217: This whole block has the same problem as JoinReadAll described in https://github.com/ethersphere/bee/pull/266#discussion_r437311701. At least the jsonhtt call should be removed and we can handle the err != nil before any write to outBuffer in another PR. BYW, I think that outBuffer does not have any purpose in the current code.

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,245 @@
...
+	w.Header().Set("ETag", fmt.Sprintf("%q", e.Reference()))
+	w.Header().Set("Content-Disposition", fmt.Sprintf("inline; filename=\"%s\"", metaData.Filename))
+	w.Header().Set("Content-Type", metaData.MimeType)
+	_, _ = io.Copy(w, outBuffer)
+}
+
+// storeMeta is used to store metadata information as a whole.
+func (s *server) storeMeta(ctx context.Context, dataBytes []byte) (swarm.Address, error) {
+	dataBuf := bytes.NewBuffer(dataBytes)
+	dataReader := io.LimitReader(dataBuf, int64(len(dataBytes)))
```
janos, line 227: I do not think that LimitReader is needed. It limits the rad to exact how the data is long.

pkg/api/bzzfile.go
```diff
@@ -0,0 +1,245 @@
...
+	dataReadCloser := ioutil.NopCloser(dataReader)
+	o, err := s.splitUpload(ctx, dataReadCloser, int64(len(dataBytes)))
+	if err != nil {
+		return swarm.ZeroAddress, err
+	}
+	bytesResp := o.(bytesPostResponse)
```
janos, line 233: This is very unfortunate. But this is a consequence of a badly written splitUpload. We can address this later in the PR when we address technical dept created in the pr that added splitUpload.

pkg/api/router.go
```diff
@@ -38,6 +38,15 @@ func (s *server) setupRouting() {
 	handle(router, "/bytes", jsonhttp.MethodHandler{
 		"POST": http.HandlerFunc(s.bytesUploadHandler),
 	})
+
+	handle(router, "/bzz-file", jsonhttp.MethodHandler{
+		"POST": http.HandlerFunc(s.bzzFileUploadHandler),
+	})
+
+	handle(router, "/bzz-file/{addr}", jsonhttp.MethodHandler{
```
janos, line 46: We can call this endpoint `/files` right away as per https://hackmd.io/fGoeV1n2QficZwuHnYmb-g#File.

pkg/api/bzzfile.go
jmozah: I use it only in meta writing.. where we exactly know how much bytes to write.. is it not safe to have it?

jmozah:
> part is already read and here cannot be read any further as it reached EOF. The whole content type detection has to be done before any Read call on part. Basically, before storePartData call, the whole code block from line 85 to line 111 has to be moved, with one important difference. If the bufio.Reader is constructed it has to be passed to storePartData, not directly the part variable as part reader already moved its cursor, while the bufio reader did not, it just peeked, so it can read from the start, as part reader has advanced 512 bytes. The best way to do that is to create var reader io.Reader = part before if contentType == "" { line and here to do reader = bufio.NewReader(part) (assign to the variable, not create a new one), and use that reader variable in storePartData fr, err := s.storePartData(ctx, reader, fileSize).

Strange, i dont know how the unit testcase for this worked which also test and check the downloaded data with uploaded data. but i changed it as you said.

> It looks to me that contentType == "" block is not tested and it would be good to have that test.

Added arguments in test to cover the scenario where we dont send contentType, and check it on receiving.

pkg/api/bytes.go
jmozah: moved this up to the `bytesUploadHandler` for now

pkg/api/router.go
```diff
@@ -38,6 +38,15 @@ func (s *server) setupRouting() {
...
 		"POST": http.HandlerFunc(s.bytesUploadHandler),
 	})
+
+	handle(router, "/file", jsonhttp.MethodHandler{
+		"POST": http.HandlerFunc(s.bzzFileUploadHandler),
+	})
+
+	handle(router, "/file/{addr}", jsonhttp.MethodHandler{
+		"GET": http.HandlerFunc(s.bzzFileDownloadHandler),
+	})
```
janos, line 48: The endpoint should be in plural `/files`.

pkg/api/file.go
```diff
@@ -0,0 +1,265 @@
...
+	w.Header().Set("Content-Disposition", fmt.Sprintf("inline; filename=\"%s\"", metaData.Filename))
+	w.Header().Set("Content-Type", metaData.MimeType)
+	w.Header().Set("Content-Length", fmt.Sprintf("%d", dataSize))
+	_, _ = io.Copy(w, outBuffer)
+}
+
+// storeMeta is used to store metadata information as a whole.
+func (s *server) storeMeta(ctx context.Context, dataBytes []byte) (swarm.Address, error) {
+	dataBuf := bytes.NewBuffer(dataBytes)
+	dataReader := io.LimitReader(dataBuf, int64(len(dataBytes)))
```
janos, line 246: LimitReader is not needed.

nolash: Just make sure that splitter writes aren't called with more than one chunk of data at a time...

pkg/api/file.go
```diff
@@ -0,0 +1,265 @@
...
+				jsonhttp.BadRequest(w, "error reading content type")
+				return
+			}
+			contentType = http.DetectContentType(buf)
+		}
+
+		// find file size
+		fileSizeString := part.Header.Get("Content-Length")
+		if fileSizeString == "" {
+			s.Logger.Debugf("file: content length: %v", err)
```
janos, line 88: This error is nil here. The log message should be about missing content length.

Currently, I get no content length for a part when I submit with the form like this one:

```
<form action="http://localhost:8080/file" method="post" enctype="multipart/form-data">
 <div>
   <label for="file">Choose file to upload</label>
   <input type="file" id="file" name="file">
 </div>
 <div>
   <button>Submit</button>
 </div>
</form>
```

Do you maybe have an idea what is wrong? As I had this form working with the code from last week, which is not here anymore because of the force push.

janos: I have created a pr that handles this like we do in swarm already https://github.com/ethersphere/bee/pull/296. With it, the upload is working nicely.

jmozah: merged your PR

PR #283, opened 2020-06-10
Advertisable address

pradovic:
This is an attempt to solve/reduce a problem with the address that is broadcasted via hive in the case when the node is behind the NAT (UPNP). Problem is that the real public address is hard to find in this case, and it is changing through time. The quick fix is to use the mappings that NAT manager provides. the `host.addrs()` call is using these port mappings if they are present and returns the correct public reachable address. So we can send that address in the case when there is one available and is  same as received observed address but differs only in port.

Ideas for the future:
1.  Extend NAT manager with our of component and broadcast via hive whenever new address is found.
2. Other peers might also check peerstore and send updates when they get new addresses for peers
3. Inspect observed addresses more (maybe use addresses that are spotted by multiple peers and verified that they work), and maybe use this for identification if possible to deduce which address is relevant
4. Use libp2p identity protocol or implement our own (this is a bit tricky)


acud:
@pradovic nice work on this. Do we want this merged for this version? Also a question is, if UPnP is not trivial - why do we want to support it explicitly, when the hole punching method already works? i.e. not use the UPnP and then the default method of hole punching should work anyway no?

janos:
@acud unfortunately, only UPnP works for now automatically on home routers. I think that everything else requires either manual nat setup. I would prefer if this is merged before the alpha as it would improve the connectivity between nodes in home setups that discover themselves over the hive protocol.

pkg/node/node.go
```diff
@@ -150,6 +151,15 @@ func NewBee(o Options) (*Bee, error) {
 	}
 	b.p2pService = p2ps
 
+	// wait for nat manager to init
+	logger.Info("initializing NAT manager")
+	select {
+	case <-p2ps.NATManager().Ready():
+		logger.Info("NAT manager initialized")
+	case <-time.After(10 * time.Second):
+		logger.Error("Nat manager init timeout")
```
janos, line 160: This could be a warning only. `Nat` - `NAT`.

pkg/p2p/libp2p/upnp.go
```diff
@@ -0,0 +1,63 @@
...
+// In some NAT situations, for example in the case when nodes are behind upnp, observer might send the observed address with a wrong port.
+// In this case, observed address is compared to addresses provided by host, and if there is a same address but with different port, that one is used as advertisable address instead of provided observed one.
+// TODO: this is a quickfix and it will be improved in the future
+func (r *UpnpAddressResolver) Resolve(observedAddress ma.Multiaddr) (ma.Multiaddr, error) {
+	observableAddrInfo, err := libp2ppeer.AddrInfoFromP2pAddr(observedAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	observedAddrSplit := strings.Split(observableAddrInfo.Addrs[0].String(), "/")
```
janos, line 29: I would guard against index panic here.

pkg/p2p/libp2p/upnp.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// if address is not in a form of '/ipversion/ip/protocol/port/...` don't compare to addresses and return it
+	if len(observedAddrSplit) < 5 {
+		return observedAddress, nil
+	}
+
+	observedAddressPort := observedAddrSplit[4]
```
janos, line 36: I would guard against index panic here.

pkg/p2p/libp2p/upnp.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// observervedAddressShort is an obaserved address without port
+	observervedAddressShort := strings.Join(append(observedAddrSplit[:4], observedAddrSplit[5:]...), "/")
+
+	for _, a := range r.host.Addrs() {
+		asplit := strings.Split(a.String(), "/")
```
janos, line 42: It would be good to check for the length of asplit to avoid index panic below.

pkg/node/node.go
```diff
@@ -150,6 +151,15 @@ func NewBee(o Options) (*Bee, error) {
 	}
 	b.p2pService = p2ps
 
+	// wait for nat manager to init
+	logger.Info("initializing NAT manager")
+	select {
+	case <-p2ps.NATManager().Ready():
+		logger.Info("NAT manager initialized")
```
janos, line 158: Maybe to set this to debug level?

pradovic:
@acud this is actually not solving the problem with punching method, we are relying on it completely. This is trying to solve the problem of which address do is broadcasted with hive if the connection is established. Waiting on NAT to initialize should increase the chance of getting the right address, but it might not happen. Also, this address is changing over time, so I think we should def work on improving this if we want to cover home setups in the network, as they will be cases that address is not valid (either changed or not acquired in time) . I wrote some ideas in a description but I am still not sure on how to solve this completely, but I think we can improve it in the future. As @janos suggested, I think it is ok for now, at least we are aware of it and it is improving the correctness of broadcasted address a bit

pkg/p2p/libp2p/upnp.go
pradovic: Isn't the line above guarding it? There is a check for len?

pkg/p2p/libp2p/upnp.go
pradovic: There is a check right below that is comparing it to the len of `observedAddrSplit`, and that one is checked above. That should prevent it from panicking, right?

pkg/p2p/libp2p/upnp.go
janos: Ups, right, sorry.

pkg/p2p/libp2p/upnp.go
janos: Yes, right, indirectly, which is all fine.

pkg/node/node.go
```diff
@@ -150,6 +151,15 @@ func NewBee(o Options) (*Bee, error) {
 	}
 	b.p2pService = p2ps
 
+	// wait for nat manager to init
+	logger.Info("initializing NAT manager")
```
janos, line 155: This can be Debug also.

pkg/node/node.go
```diff
@@ -150,6 +151,19 @@ func NewBee(o Options) (*Bee, error) {
...
+	select {
+	case <-p2ps.NATManager().Ready():
+		logger.Debug("NAT manager initialized")
+	case <-time.After(10 * time.Second):
+		logger.Warning("NAT manager init timeout")
+	}
+
+	// this is magic sleep to give NAT time to sync the mappings
+	// this is a hack, kind of alchemy and should be improved
+	time.Sleep(3 * time.Second)
```
janos, line 165: I would put this sleep right after the Ready() channel reads and before the `NAT manager initialized` log line. If the timeout happens, there is no need to wait more and there is a delay after the `NAT manager initialized` log message.

pradovic: Correct! 🙈

pkg/p2p/libp2p/internal/handshake/handshake_test.go
```diff
@@ -241,8 +274,49 @@ func TestHandshake(t *testing.T) {
 		}
 	})
 
+	t.Run("Handshake - error advertisablea address", func(t *testing.T) {
```
acud, line 277: typo

pkg/p2p/libp2p/internal/handshake/handshake_test.go
```diff
@@ -241,8 +274,49 @@ func TestHandshake(t *testing.T) {
...
+		var buffer1 bytes.Buffer
+		var buffer2 bytes.Buffer
+		stream1 := mock.NewStream(&buffer1, &buffer2)
+		stream2 := mock.NewStream(&buffer2, &buffer1)
+
+		testError := errors.New("test error")
+		aaddresser.err = testError
```
acud, line 284: addresser?

pkg/p2p/libp2p/internal/handshake/handshake_test.go
```diff
@@ -241,8 +274,49 @@ func TestHandshake(t *testing.T) {
...
+				NetworkID: networkID,
+				Light:     false,
+			},
+		}); err != nil {
+			t.Fatal(err)
+		}
+
+		res, err := handshakeService.Handshake(stream1, node2AddrInfo.Addrs[0], node2AddrInfo.ID)
+		if err != testError {
+			t.Fatal("expected error")
```
acud, line 309: `Fatalf("expected error %v but got %v",testError, err)`

pkg/p2p/libp2p/internal/handshake/handshake_test.go
```diff
@@ -241,8 +274,49 @@ func TestHandshake(t *testing.T) {
...
+				},
+				NetworkID: networkID,
+				Light:     false,
+			},
+		}); err != nil {
+			t.Fatal(err)
+		}
+
+		res, err := handshakeService.Handshake(stream1, node2AddrInfo.Addrs[0], node2AddrInfo.ID)
+		if err != testError {
```
acud, line 308: use errors.Is?

pkg/p2p/libp2p/internal/handshake/handshake_test.go
```diff
@@ -496,6 +582,39 @@ func TestHandshake(t *testing.T) {
...
+
+		w := protobuf.NewWriter(stream2)
+		if err := w.WriteMsg(&pb.Syn{
+			ObservedUnderlay: node1maBinary,
+		}); err != nil {
+			t.Fatal(err)
+		}
+
+		res, err := handshakeService.Handle(stream1, node2AddrInfo.Addrs[0], node2AddrInfo.ID)
+		if err != testError {
```
acud, line 610: same here

pkg/p2p/libp2p/libp2p.go
```diff
@@ -166,17 +174,17 @@ func New(ctx context.Context, signer beecrypto.Signer, networkID uint64, overlay
...
 		libp2pPeerstore:  libp2pPeerstore,
 		metrics:          newMetrics(),
 		networkID:        networkID,
-		handshakeService: handshakeService,
 		peers:            peerRegistry,
 		addressbook:      o.Addressbook,
 		logger:           o.Logger,
 		tracer:           o.Tracer,
-		conectionBreaker: breaker.NewBreaker(breaker.Options{}), // todo: fill non-default options
+		conectionBreaker: breaker.NewBreaker(breaker.Options{}), // use default options
```
acud, line 186: `conection -> connection`

pkg/p2p/libp2p/upnp.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	observedAddrSplit := strings.Split(observableAddrInfo.Addrs[0].String(), "/")
+
+	// if address is not in a form of '/ipversion/ip/protocol/port/...` don't compare to addresses and return it
+	if len(observedAddrSplit) < 5 {
+		return observedAddress, nil
+	}
+
+	observedAddressPort := observedAddrSplit[4]
```
acud, line 41: how do you know that it is always in the 5th place?

pradovic:
@acud Thanks for the review! Yes, it's a bit of a hack, so I am afraid as well, let's test and see how it behaves. If you take a look at ln 36 in `upnp.go` there is a comment: "if address is not in a form of '/ipversion/ip/protocol/port/...` don't compare to addresses and return it". So basically, I tried to guard against  panics, and the compare algorithm is like that it should always return received observable address if it does not match the exact address. The 1st and the 5th place in the list are parts of the multiaddresses, so if node is started without `quic` it should just never match such an address. Only use-case I am worried about is if the node sending the observable address has an address with a right port, while the node receiving has a matching address but with a wrong port. We could not find a use-case when this happens.

pkg/p2p/libp2p/internal/handshake/handshake_test.go
pradovic: I added an a to be like `advertisable addresses`. I am not sure if that's problematic?

pkg/p2p/libp2p/internal/handshake/handshake_test.go
pradovic: I don't expect error to  be wrapped here as test error is really what should be returned, so while being more nice, `error.Is` would be just a bit inaccurate. Do you think ti makes more sense to use it than exact compare

PR #302, opened 2020-06-15
localstore: remove gc check hack

acud:
reverts the abomination of a fix we introduced here: https://github.com/ethersphere/swarm/pull/1940

we need to find a way to test of this properly with integration testing


acud:
@janos one test is failing, but is related to protobuf.
I am merging this anyway.

PR #306, opened 2020-06-15
support binary upload to api files endpoint

janos:
This PR adds a support to upload a file in its binary form, get its `content-type` from request header and name from `name` query parameter. This is done beside the regular multipart/form-data upload.

https://hackmd.io/fGoeV1n2QficZwuHnYmb-g?both#POST-v1files-typestringamp-namestring

- fileUploadHandler is reorganized to support binary data uploads, content type detection from header and filename detection from query parameter (Content-Disposition is only response header and is not used in requests)
- splitUpload returns only address to simplify its usage
- splitUpload removes requirement for io.ReadCloser as Close is never used, just io.Reader is required
- bytesUploadHandler structure is more consistent with other handlers
- minor adjustments to exported variables
- existing tests are extended to test for two types of uploading
- router decelerations are just reordered to group similar routes

### DIFF
--- a/pkg/api/file.go
+++ b/pkg/api/file.go
@@ -27,7 +27,7 @@ import (
 	"github.com/gorilla/mux"
 )
 
-const multipartFormDataContentType = "multipart/form-data"
+const multipartFormDataMediaType = "multipart/form-data"
 
 type fileUploadResponse struct {
 	Reference swarm.Address `json:"reference"`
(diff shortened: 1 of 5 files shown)


pkg/api/file.go
```diff
@@ -28,163 +27,171 @@ import (
...
-
 		// try to find filename
 		// 1) in part header params
 		// 2) as formname
 		// 3) file reference hash (after uploading the file)
-		fileName := part.FileName()
+		fileName = part.FileName()
 		if fileName == "" {
 			fileName = part.FormName()
 		}
```
pradovic, line 73:
```suggestion
		
		if fileName = part.FileName(); fileName == "" {
			fileName = part.FormName()
		}
```

acud: this is not correct, since fileName will only be scoped to the `if` condition

PR #308, opened 2020-06-15
Added translation from chunks to bytes in help and log

Eknir:
I believe that we should never use the metric "chunks" when we communicate about database size to our users, however I can image that doing such a change is debated (as @mortelli convincingly argued, in some tests it is good to have fine-grained control over the gc with chunks as metric) and it might be too big of a change for now. 

At a minimum, we should give the user a translation from db capacity (as measured in chunks) to db capacity (measured in bytes).

### DIFF
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -151,7 +151,7 @@ func (c *command) initStartCmd() (err error) {
 	}
 
 	cmd.Flags().String(optionNameDataDir, filepath.Join(c.homeDir, ".bee"), "data directory")
-	cmd.Flags().Uint64(optionNameDBCapacity, 5000000, "db capacity in chunks")
+	cmd.Flags().Uint64(optionNameDBCapacity, 5000000, "db capacity in chunks, multiply by 4096 to get bytes")
 	cmd.Flags().String(optionNamePassword, "", "password for decrypting keys")
 	cmd.Flags().String(optionNamePasswordFile, "", "path to a file that contains password for decrypting keys")
 	cmd.Flags().String(optionNameAPIAddr, ":8080", "HTTP API listen address")
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/node"
+	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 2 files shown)


pkg/localstore/localstore.go
```diff
@@ -166,7 +166,7 @@ func New(path string, baseKey []byte, o *Options, logger logging.Logger) (db *DB
 	if db.capacity == 0 {
 		db.capacity = defaultCapacity
 	}
-	db.logger.Infof("db capacity: %v", db.capacity)
+	db.logger.Infof("database capacity in chunks (and in Bytes): %v (%v)", db.capacity, db.capacity*swarm.ChunkSize)
```
acud, line 169: i think it might be nicer to have a message like:
`database capacity: %d chunks (%d bytes, %d megabytes)`.
counting in bytes is so 90s :)

cmd/bee/cmd/start.go
```diff
@@ -151,7 +151,7 @@ func (c *command) initStartCmd() (err error) {
 	}
 
 	cmd.Flags().String(optionNameDataDir, filepath.Join(c.homeDir, ".bee"), "data directory")
-	cmd.Flags().Uint64(optionNameDBCapacity, 5000000, "db capacity in chunks")
+	cmd.Flags().Uint64(optionNameDBCapacity, 5000000, "db capacity in chunks, multiply by 4096 to get bytes")
```
janos, line 154: It is more correct to say that a chunk is 4104 bytes long in its total size. But the storage also contains indexes and leveldb metadata that would add up. We could state that size in bytes is an approximation and have something less precise as it is currently in swarm cli option.

Eknir: @janos , thanks for comment. 

What about changing it here to `multiply by 4200 to get the approximate capacity in bytes` ?

Eknir: Or `multiply by swarm.ChunkSize to get the approximate capacity in bytes`

Eknir: > Or multiply by swarm.ChunkSize to get the approximate capacity in bytes

Otherwise it does not correspond with the log in `localstore.go:169`

pkg/localstore/localstore.go
Eknir: Updated with slight change from what you suggested

cmd/bee/cmd/start.go
Eknir: Updated

pkg/localstore/localstore.go
```diff
@@ -166,7 +166,7 @@ func New(path string, baseKey []byte, o *Options, logger logging.Logger) (db *DB
 	if db.capacity == 0 {
 		db.capacity = defaultCapacity
 	}
-	db.logger.Infof("db capacity: %v", db.capacity)
+	db.logger.Infof("database capacity: %d chunks, %d bytes, %.2f megabytes.", db.capacity, db.capacity*swarm.ChunkSize, float64(db.capacity*swarm.ChunkSize)*9.5367431640625e-7)
```
janos, line 169: I think that bytes is too precise. Can we just calculate the approximate value in appropriate units? Something like MB, GB depending on the value, `db capacity: 5000000 chunks (approximately 20GB)` .

PR #309, opened 2020-06-16
Clean addressbook after failed connection attempts

pradovic:
The basic idea is to stop infinitely retrying  and/or clean `addressbook` from the non-connectable nodes. This is not a full solution, as we would probably like to implement some kind of connection service/component to be used to make the code more reusable and less error prone, as we discussed when talking about handlers as well. This is just a quick fix, and if it works well, we can extract it to this kind of component and use it across  bee in a uniform way.
Please check  linked issue for additional details

### DIFF
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -249,24 +249,26 @@ func (k *Kad) connect(ctx context.Context, peer swarm.Address, ma ma.Multiaddr, 
 			return nil
 		}
 
+		k.logger.Debugf("error connecting to peer %s: %v", peer, err)
 		retryTime := time.Now().Add(timeToRetry)
 		var e *p2p.ConnectionBackoffError
-		if errors.As(err, &e) {
-			retryTime = e.TryAfter()
-		}
-
-		k.logger.Debugf("error connecting to peer %s: %v", peer, err)
 		k.waitNextMu.Lock()
 		failedAttempts := 0
-		info, ok := k.waitNext[peer.String()]
-		if ok {
-			failedAttempts = info.failedAttempts + 1
+		if errors.As(err, &e) {
+			retryTime = e.TryAfter()
+		} else {
+			info, ok := k.waitNext[peer.String()]
+			if ok {
+				failedAttempts = info.failedAttempts
+			}
+
+			failedAttempts++
 		}
 
 		if failedAttempts > maxConnAttempts {
 			delete(k.waitNext, peer.String())
 			if err := k.addressBook.Remove(peer); err != nil {
-				k.logger.Debug("could not remove peer from addressbook: %s", peer.String())
+				k.logger.Debugf("could not remove peer from addressbook: %s", peer.String())
 			}
 		} else {
 			k.waitNext[peer.String()] = retryInfo{tryAfter: retryTime, failedAttempts: failedAttempts}
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -356,13 +356,7 @@ func (k *Kad) Disconnected(addr swarm.Address) {
 	k.connectedPeers.Remove(addr, po)
 
 	k.waitNextMu.Lock()
-	failedAttempts := 0
-	info, ok := k.waitNext[addr.String()]
-	if ok {
-		failedAttempts = info.failedAttempts + 1
-	}
-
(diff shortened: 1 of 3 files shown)


pkg/kademlia/kademlia.go
```diff
@@ -242,12 +248,31 @@ func (k *Kad) connect(ctx context.Context, peer swarm.Address, ma ma.Multiaddr,
...
+		info, ok := k.waitNext[peer.String()]
+		if ok {
+			failedAttempts = info.failedAttempts + 1
+		}
 
-		// TODO: somehow keep track of attempts and at some point forget about the peer
+		if failedAttempts > maxConnAttempts {
+			delete(k.waitNext, peer.String())
+			if err := k.addressBook.Remove(peer); err != nil {
+				k.logger.Debug("could not remove peer from addressbook: %s", peer.String())
```
acud, line 269: debugf

pkg/kademlia/kademlia.go
```diff
@@ -23,8 +23,9 @@ import (
 )
 
 const (
-	maxBins        = 16
-	nnLowWatermark = 2 // the number of peers in consecutive deepest bins that constitute as nearest neighbours
+	maxBins         = 16
+	nnLowWatermark  = 2 // the number of peers in consecutive deepest bins that constitute as nearest neighbours
+	maxConnAttempts = 3 // when there is maxConnAttempts failed connect calls for a given peer it is considere non-connectable
```
acud, line 28: questions about this:
1. if i start my node and i am disconnected from the internet, does this mean that within 3 attempts to connect to peers will be address book get entirely truncated? is this the desirable behavior?
2. how long will it take to reach this limit in the above case?
3. if the connection breaker is hit how long does it take to reach the third attempt?

pradovic: So, first thing is, the PR is still in progress, sorry, so I might change stuff :)
But to answer anyway about the idea I am testing now:
1. So, the conn attempts and connection breaker are not related. The connection breaker should recognize the situation of multiple failed attempts across bee, and the connect calls will return the appropriate error and when to try again, so we don't try for now reason.
2. I left the same logic as before in kademlia, the only difference is that after 3 attemts to the same peer I would remove it from the wait next map and remove it from the addressbook so it is not retried again, unless it is added to the addressbook again somehow and then it will go all over again
3. If the connection breaker is hit, the all connection calls are postponed until returned `tryAfter`, and I am not taking into account those as failed connection attempts, only if the error is not from the connection breaker

Does this makes sense? Maybe I missed something, this was just to try quick fix, I am still testing how it behaves

pradovic: Regarding the release, I asked the same, you can check the linked issue :)

janos: This PR would be good to be merged before the release as it would reduce the error messages when connecting to staled nodes greatly. The issue was made with this in mind.

pkg/kademlia/kademlia.go
```diff
@@ -329,7 +356,13 @@ func (k *Kad) Disconnected(addr swarm.Address) {
 	k.connectedPeers.Remove(addr, po)
 
 	k.waitNextMu.Lock()
-	k.waitNext[addr.String()] = time.Now().Add(timeToRetry)
+	failedAttempts := 0
+	info, ok := k.waitNext[addr.String()]
+	if ok {
+		failedAttempts = info.failedAttempts + 1
```
janos, line 362: I would assume that fieldAttempts would be incremented only if the connection fails, not on every disconnect, as disconnect may be intentional by the peer. Also reseting failedAttempts to 0 should be on successful connection, I think.

pradovic: OMG, good catch, I was debugging and changing this and forgot to change back 🙈  thanks!

pradovic:
@janos Sure, I will add some tests as well and ask for a review again.

pradovic:
@janos I have added tests as well

pkg/kademlia/kademlia.go
```diff
@@ -242,12 +248,33 @@ func (k *Kad) connect(ctx context.Context, peer swarm.Address, ma ma.Multiaddr,
...
+		failedAttempts := 0
+		if errors.As(err, &e) {
+			retryTime = e.TryAfter()
+		} else {
+			info, ok := k.waitNext[peer.String()]
+			if ok {
+				failedAttempts = info.failedAttempts
+			}
 
-		// TODO: somehow keep track of attempts and at some point forget about the peer
```
janos, line 250: I have forgot about this todo. It is good that this PR addresses it.

PR #310, opened 2020-06-16
add File upload section to readme

janos:


README.md
```diff
@@ -75,6 +75,45 @@ api-addr: 127.0.0.1:8085
...
+
+To download a file, open an URL with that reference `http://localhost:8080/files/3b2791985f102fe645d1ebd7f51e522d277098fcd86526674755f762084b94ee`.
+
+If you need to specify manually content type or filename during upload:
+
+```sh
+curl -H "Content-Type: image/x-jpeg" --data-binary @kitten.jpg localhost:8081/files?name=cat.jpg
+```
+
+The same response with file reference is returned.
```
Eknir, line 102: What is the purpose of having this here? Is it 1) a feature of the Bee node, or 2) just a different way to instruct curl to do the same? 

If 2, I think it would be better to omit it. If 1, it is not clear to me :)

README.md
```diff
@@ -75,6 +75,45 @@ api-addr: 127.0.0.1:8085
...
+
+To avoid uploading with command line tools, this HTML file can be opened in the browser and used to select and submit a file:
```
Eknir, line 104: If you talk about doing things easily in the browser, you can also mention here that the user can locate the file in the browser, by using the address bar and navigating to `<API_address_node>/files/{hash}`

README.md
janos: Curl is not doing the same. These are two different requests. The first one is multipart, the second one is streaming file content. This is the illustration what is possible and users can choose what suits better for their needs, as these examples also suggest what is possible if someone wants to use different tools or libraries, without going too technical on http.

README.md
```diff
@@ -75,6 +75,47 @@ api-addr: 127.0.0.1:8085
 data-dir: /data/bees/bee5
 ```
 
+## File upload
+
+File can be uploaded by making an HTTP request like this one:
```
acud, line 80:
```suggestion
A file can be uploaded by making an HTTP request like this one:
```

README.md
```diff
@@ -75,6 +75,47 @@ api-addr: 127.0.0.1:8085
...
+
+```sh
+curl -F file=@kitten.jpg http://localhost:8080/files
+```
+
+Curl will make a properly encoded `multipart/form-data` request sending the filename, content type and file content to the bee API, and it will return a response with the reference to the uploaded file:
```
acud, line 86:
```suggestion
`curl` will form a `multipart/form-data` request with the filename, content type and file content to the `bee` API, returning a response with the reference to the uploaded file:
```

README.md
```diff
@@ -75,6 +75,47 @@ api-addr: 127.0.0.1:8085
...
+
+```json
+{"reference":"3b2791985f102fe645d1ebd7f51e522d277098fcd86526674755f762084b94ee"}
+```
+
+This reference is just an example, it will differ for every uploaded file.
+
+To download a file, open an URL with that reference `http://localhost:8080/files/3b2791985f102fe645d1ebd7f51e522d277098fcd86526674755f762084b94ee`.
```
acud, line 94:
```suggestion
To download or view the file, open an URL with that reference in your browser or through your favorite command line tool: `http://localhost:8080/files/3b2791985f102fe645d1ebd7f51e522d277098fcd86526674755f762084b94ee`.
```

README.md
```diff
@@ -75,6 +75,47 @@ api-addr: 127.0.0.1:8085
...
+
+If you need to specify manually content type or filename during upload:
```
acud, line 96:
```suggestion
In case you'd like to manually specify the content type during upload, you could do so using a query string parameter:
```

janos:
Thank you so much @acud, I have updated the PR with your suggestions.

PR #312, opened 2020-06-17
Add welcome message to handshake

paxthemax:
# Static welcome message in handshake

This PR adds a static welcome message exchanged during handshakes.

The message is set via node configuration.

TODO: Determine exact message format, determine maximum length


pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -214,6 +224,8 @@ func (s *Service) Handle(stream p2p.Stream, remoteMultiaddr ma.Multiaddr, remote
 	}
 
 	s.logger.Tracef("handshake finished for peer %s", remoteBzzAddress.Overlay.String())
+	s.logger.Tracef("received welcome message %s from peer", ack.WelcomeMessage)
```
janos, line 227: This message is in the Handle method, meaning that it will be printed on the node that is connected to. I think that the logging should happen on the node that is requesting a connection. WTDY?

This should be logged only if welcome message is not empty.

I think that Info log level would be better for visibility of this message, as Info is the default one.

Maybe to log also the peer overlay address at the end of the log message? @Eknir any suggestions on this?

cmd/bee/cmd/start.go
```diff
@@ -170,6 +172,7 @@ func (c *command) initStartCmd() (err error) {
 	cmd.Flags().String(optionNameTracingEndpoint, "127.0.0.1:6831", "endpoint to send tracing data")
 	cmd.Flags().String(optionNameTracingServiceName, "bee", "service name identifier for tracing")
 	cmd.Flags().String(optionNameVerbosity, "info", "log verbosity level 0=silent, 1=error, 2=warn, 3=info, 4=debug, 5=trace")
+	cmd.Flags().String(optionWelcomeMessage, "welcome-message", "send a welcome message string during handshakes")
```
janos, line 175: The default value should be blank, not "welcome-message".

pkg/p2p/libp2p/internal/handshake/handshake_test.go
```diff
@@ -23,6 +23,10 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 )
 
+const (
+	DefaultWelcomeMessage = "HelloWorld"
```
janos, line 27: I think that this can be in the scope of TestHandshake test.

pkg/p2p/libp2p/libp2p_test.go
```diff
@@ -22,6 +22,10 @@ import (
 	"github.com/multiformats/go-multiaddr"
 )
 
+const (
+	DefaultWelcomeMessage = "Hello World!"
```
janos, line 26: Let's leave a default message blank (not set at all), and explicitly set it in tests if needed.

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -128,13 +134,16 @@ func (s *Service) Handshake(stream p2p.Stream, peerMultiaddr ma.Multiaddr, peerI
...
-		Light:     s.lightNode,
+		NetworkID:      s.networkID,
+		Light:          s.lightNode,
+		WelcomeMessage: s.welcomeMessage,
 	}); err != nil {
 		return nil, fmt.Errorf("write ack message: %w", err)
 	}
 
 	s.logger.Tracef("handshake finished for peer %s", remoteBzzAddress.Overlay.String())
+	s.logger.Tracef("received welcome message %s from peer", resp.Ack.WelcomeMessage)
```
janos, line 145: This should be logged only if welcome message is not blank. And maybe to log the peer overlay address at the end. Also Info log level suggested for better visibility of the message as it is the default log level.

pkg/p2p/libp2p/internal/handshake/handshake.go
Eknir: I agree with what you write here. It should definitely be in the info. And maybe you can make it more active:

peer <overlay_address> greets you: <welcome_message>

pkg/p2p/libp2p/internal/handshake/pb/handshake.proto
```diff
@@ -16,6 +16,7 @@ message Ack {
     BzzAddress Address = 1;
     uint64 NetworkID = 2;
     bool Light = 3;
+    string WelcomeMessage  = 4;
```
pradovic, line 19: 🌷  Just one suggestion guys. I would put this on some high random index, not to intervene with the protocol, as we might change versions of the protocols and update the message, so we don't need to have a reserved field in the lower register.  So maybe something like: 
```suggestion
    string WelcomeMessage  = 99;
```

paxthemax: @pradovic Should we consider reserving some of the low field numbers?

pradovic: hmm, we can, but I don't see the need yet tbh. Reserving is useful when deleting and obsoleting fields, so they are not reused for some other fields. In this case I don't think we need to do it.

pradovic: So for example, if we end up removing some field, or `WelcomeMessage` field, then it would be good to reserve this one so it is not reused for something else

pkg/p2p/libp2p/internal/handshake/handshake.go
paxthemax: > This message is in the Handle method, meaning that it will be printed on the node that is connected to. I think that the logging should happen on the node that is requesting a connection. WTDY?

I agree, I will remove this when the PR is ready - just wanted to have a trace on both sides of the handshake when tracing.

pkg/p2p/libp2p/internal/handshake/pb/handshake.proto
paxthemax: Thanks @pradovic, change commited!

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -24,10 +24,16 @@ import (
...
+	// ProtocolName is the text of the name of the handshake protocol.
+	ProtocolName = "handshake"
+	// ProtocolVersion is the current handshake protocol version.
 	ProtocolVersion = "1.0.0"
-	StreamName      = "handshake"
-	messageTimeout  = 5 * time.Second // maximum allowed time for a message to be read or written.
+	// StreamName is the name of the stream used for handshake purposes.
+	StreamName = "handshake"
+	// MessageTimeout is the maximum allowed time for a message to be read or written.
+	MessageTimeout = 5 * time.Second
```
janos, line 34: Should MessageTimeout be exported? I am not sure that this change is needed.

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -122,25 +147,31 @@ func (s *Service) Handshake(stream p2p.Stream, peerMultiaddr ma.Multiaddr, peerI
...
+		NetworkID:      s.networkID,
+		Light:          s.lightNode,
+		WelcomeMessage: s.welcomeMessage,
 	}); err != nil {
 		return nil, fmt.Errorf("write ack message: %w", err)
 	}
 
 	s.logger.Tracef("handshake finished for peer %s", remoteBzzAddress.Overlay.String())
+	if len(resp.Ack.WelcomeMessage) > 0 {
+		s.logger.Infof("peer %s greets you: $s", resp.Ack.Address.String(), resp.Ack.WelcomeMessage)
```
janos, line 165: For some reason, I get a strangely formatted log line.

```
INFO[2020-06-19T11:04:26+02:00] peer Underlay:"\004\177\000\000\001\006\033\237\245\003'\000%\010\002\022!\003H\372\271V\344\365v\352\tm.\363H\316T}\016\213\260\255<\216\206\252_\007\344\232\324\271'+" Signature:"\037\3318Ug\344%`V\247/\020!\355\033\003m\002\265\336\251\355\013N!\030\"\341i\217[\242\330\0003ow\336\301'\017S\232\2069hOZ\376\314\007\302\312\315\003\300B\037\025\206R\344\030xO" Overlay:"0q\256(\250\003a7\214\235\324\233\016\\g^9\276\275\031\241\202\201?\273\007f){%\024\322"  greets you: $s%!(EXTRA string=Well, Hello there!)
```

I am not sure if the complete bzz address should be printed here (with signature), maybe just overlay would be enough, as we are logging the connection details already like this:

```
INFO[2020-06-19T11:04:26+02:00] successfully connected to peer [Overlay: 3071ae28a80361378c9dd49b0e5c675e39bebd19a182813fbb0766297b2514d2, Underlay: /ip4/127.0.0.1/tcp/7071/p2p/16Uiu2HAmHZsSr34G1LgdXEu7W6xa6QJgiS9MjXHEANNayqikJVri] 
```

A problem with the pattern is that `$s` should be `%s` at the end of it.

pradovic: I think it is because the address from the ack is used, so it is printing an proto generated address which probably does not have a String() method how we would like it. It's better to use remoteBzzAddress to print the peer, and resp to print the message, maybe?

paxthemax: @janos, @pradovic  - Should we just print the `remoteBzzAddress.Overlay.String()`?

pkg/p2p/libp2p/internal/handshake/handshake.go
paxthemax: Good point, unexporting.

pkg/p2p/libp2p/internal/handshake/handshake.go
janos: I've got the notification about the comment from @paxthemax 

> @janos, @pradovic - Should we just print the remoteBzzAddress.Overlay.String()?

But do not see it here.

My answer is yes, I think that it should be enough.

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -24,10 +24,15 @@ import (
...
+	ProtocolName = "handshake"
+	// ProtocolVersion is the current handshake protocol version.
 	ProtocolVersion = "1.0.0"
-	StreamName      = "handshake"
-	messageTimeout  = 5 * time.Second // maximum allowed time for a message to be read or written.
+	// StreamName is the name of the stream used for handshake purposes.
+	StreamName = "handshake"
+	// MaxWelcomeMessageLength is maximum number of characters allowed in the welcome message.
+	MaxWelcomeMessageLength = 140
+	messageTimeout          = 5 * time.Second
```
Eknir, line 35: I know it's not in your PR, but this `const` is not used anywhere. Can we remove it?

pkg/p2p/libp2p/libp2p.go
```diff
@@ -60,14 +60,15 @@ type Service struct {
...
-	DisableQUIC bool
-	LightNode   bool
-	Addressbook addressbook.Putter
-	Logger      logging.Logger
-	Tracer      *tracing.Tracer
+	PrivateKey     *ecdsa.PrivateKey
+	NATAddr        string
+	DisableWS      bool
+	DisableQUIC    bool
+	LightNode      bool
```
Eknir, line 67: @janos, we don't have light node capabilities atm. Why is this `bool` here?

janos: It is added because the handshake defines it. The feature is still not there but the protocol specifications has it and with this flag it is configurable.

Eknir:
@janos , is it an idea to 
1) either merge this PR to master before the release or if not:
2) already add the field in the handshake (just as lightnode), we can add it after the release

janos:
@Eknir @paxthemax  I think that this PR can be merged when conflicts are solved.

janos:
Thanks @Eknir, I will resolve the conflicts and we can merge to master.

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -128,19 +152,25 @@ func (s *Service) Handshake(stream p2p.Stream, peerMultiaddr ma.Multiaddr, peerI
...
+		NetworkID:      s.networkID,
+		Light:          s.lightNode,
+		WelcomeMessage: s.welcomeMessage,
 	}); err != nil {
 		return nil, fmt.Errorf("write ack message: %w", err)
 	}
 
 	s.logger.Tracef("handshake finished for peer %s", remoteBzzAddress.Overlay.String())
+	if len(resp.Ack.WelcomeMessage) > 0 {
+		s.logger.Infof("peer [%s] greets you: %s", remoteBzzAddress.Overlay.String(), resp.Ack.WelcomeMessage)
```
Eknir, line 164: Remove the []

PR #316, opened 2020-06-17
Printing bzzaddress

pradovic:
Print bzz addresses when the node is connected a bit nicer + plus improve bzzAddr debug logs a bit.

### DIFF
--- a/pkg/bzz/address.go
+++ b/pkg/bzz/address.go
@@ -126,25 +126,5 @@ func (a *Address) String() string {
 // ShortString returns shortened versions of overlay and underlay address in a format: [shortOver..., shortUnder...]
 // It can be used for logging
 func (a *Address) ShortString() string {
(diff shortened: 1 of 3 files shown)


pkg/bzz/address.go
```diff
@@ -86,34 +87,64 @@ func (a *Address) Equal(b *Address) bool {
...
+
+// ShortUnderlay returns shortened version of Underlay address
+// It can be used for logging
+func (a *Address) ShortUnderlay() string {
+	if uaddr := a.Underlay.String(); len(uaddr) >= 16 {
+		return uaddr[:16]
+	} else {
+		return uaddr
+	}
+}
```
janos, line 150: I think that we can keep complete addresses in all log messages, not to slice them to length of 16, to drop the Short* methods.

pradovic: Done

PR #319, opened 2020-06-17
Addressbook get  - not found

pradovic:
This PR addresses 2 minor issues:
1. In order to check the result of `addressbook.Get()` one must compare received error with `storage.NotFound` error, which is not in the same package and should not be used to check the result of the addressbook get function.
2. The returned `BzzAddress` is copied creating minor, but unnecessary overhead.
3. Makes the user of fetched `BzzAddress` pay attention to pass the address of BzzAddress on many places where it is convenient, such where  `Equal(), String()` methods are used and their appropriate interfaces

### DIFF
--- a/pkg/addressbook/addressbook.go
+++ b/pkg/addressbook/addressbook.go
@@ -5,6 +5,7 @@
 package addressbook
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
@@ -16,6 +17,8 @@ import (
 const keyPrefix = "addressbook_entry_"
 
 var _ Interface = (*store)(nil)
+
+var ErrNotFound = errors.New("addressbook: not found")
 
 type Interface interface {
 	GetPutter
@@ -57,7 +60,7 @@ func (s *store) Get(overlay swarm.Address) (*bzz.Address, error) {
 	err := s.store.Get(key, &v)
 	if err != nil {
 		if err == storage.ErrNotFound {
-			return nil, nil
+			return nil, ErrNotFound
 		}
 
 		return nil, err
--- a/pkg/addressbook/addressbook.go
+++ b/pkg/addressbook/addressbook.go
@@ -30,7 +30,7 @@ type GetPutter interface {
 }
 
 type Getter interface {
-	Get(overlay swarm.Address) (addr bzz.Address, err error)
+	Get(overlay swarm.Address) (addr *bzz.Address, err error)
 }
 
 type Putter interface {
@@ -51,12 +51,16 @@ func New(storer storage.StateStorer) Interface {
 	}
 }
 
-func (s *store) Get(overlay swarm.Address) (bzz.Address, error) {
+func (s *store) Get(overlay swarm.Address) (*bzz.Address, error) {
 	key := keyPrefix + overlay.String()
-	v := bzz.Address{}
+	v := &bzz.Address{}
 	err := s.store.Get(key, &v)
 	if err != nil {
-		return bzz.Address{}, err
+		if err == storage.ErrNotFound {
+			return nil, nil
+		}
+
+		return nil, err
 	}
 	return v, nil
 }
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -6,7 +6,6 @@ package hive
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"time"
 
@@ -16,7 +15,6 @@ import (
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/p2p"
 	"github.com/ethersphere/bee/pkg/p2p/protobuf"
-	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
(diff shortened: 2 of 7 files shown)


pkg/hive/hive.go
```diff
@@ -97,13 +95,14 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
...
 		if err != nil {
-			if errors.Is(err, storage.ErrNotFound) {
-				s.logger.Debugf("Peer not found %s", p)
-				continue
-			}
 			return err
 		}
 
+		if addr == nil {
+			s.logger.Debugf("Peer not found %s", p)
```
janos, line 102: Aaaand, this is the comment that was not submitted. I think that we can provide a bit more information in this message like "hive: peer not found in address book: %s".

pkg/addressbook/addressbook.go
```diff
@@ -51,12 +51,16 @@ func New(storer storage.StateStorer) Interface {
...
 
-func (s *store) Get(overlay swarm.Address) (bzz.Address, error) {
+func (s *store) Get(overlay swarm.Address) (*bzz.Address, error) {
 	key := keyPrefix + overlay.String()
-	v := bzz.Address{}
+	v := &bzz.Address{}
 	err := s.store.Get(key, &v)
 	if err != nil {
-		return bzz.Address{}, err
+		if err == storage.ErrNotFound {
```
acud, line 59: i find this change strange. if we dont find an entry this is an error, returning a `nil,nil` here is something i have never seen anywhere in the codebase. can you please explain the incentive of silencing this error? this is an anti pattern and breaks the convention of checking for error when accessing persistent storage. now all components using address book (and there's a few) will have to implement specific logic in order to avoid a nil pointer exception (panic!).

acud:
> In order to check the result of addressbook.Get() one must compare received error with storage.NotFound error, which is not in the same package and should not be used to check the result of the addressbook get function.

If this is a problem, you can introduce an `addressbook.ErrNotFound` if the dependency of `storage` really is bothering anyone (why). I feel this would a more appropriate change

pradovic:
@acud  So, I will try to tackle both your coments here.
So, from what I experienced in both Go and other programming languages so far, getting fetching the value from the map-like structure is usually done in one of 3 ways:
1. Returning nil if not found,  and error if there is a need for error at all (which I think we actually don't even need in the addressbook on the interface level, but I left it). Returning nil is considered an anti-pattern, but for generic maps because you don't know what return default values is. In this case, you actually do know and it's nil.
2. Returning the `res, ok` pair, which is common in go, used in golang maps
3. Returning special not found error, which is usually used when there is several different errors to cover, or when it is not possible to chck the return value, for example when we are returning a value type not a pointer

On the other hand, I would consider returning the underlying library error, in this case storage.Error a anti pattern because of the following:
1. It's hard to find documentation on how to check for not found error
2. What if you implement another addressbook, or Getter? You need to return storage error in order for users of addressbook not to break
3. I don't like if err!=nil { if err == specificError syntax}, but this is just me and is very opinionated, so this is not a reason not to do it. 

So I would propose to either:
1. Leave it like it is. Returning nil is a pretty common pattern actually for not found stuff when there is a pointer involved and nil is a legal value to check. I don't see a reason for panics here, it's pretty straighforward to check
2. Change to something like go maps, so returning `res, ok, err` tuple. I did not choose this cuz I don't like more then 2 return values + I don't see much benefit, the usage looks the same it's just checking weather ok is true or false instead of if res == nil, but I am ok with that as well
3. As you mentioned, specific `addressbook.ErrNotFound`

So, @janos, @acud , I leave you guys to choose, but I don't see any strong reason for it presented. It looks to me like more of you personal taste, same as returning nil was mine. I would just like to avoid returning storage error here, and value of`BzzAddress` as I don't see any benefit from it.

pkg/hive/hive.go
pradovic: I agree! Changed to a bit more info. Please LTMWYT

acud:
@pradovic address book is not a map but uses an underlying persistent storage, so I'm not sure where the claim that it is a map structure comes from.  There is a need to have an error return type because an error is possible, but we are silently ignoring it in the package which is contrary to what we do in other packages.

If you insist on treating address book as a map structure, then please remove the error type altogether and just return `*BzzAddress` or `*BzzAddress, bool`, and then use it like a map. I just find the fact that the return types of `*BzzAddress, error` are imposed to behave _like_ a map, but with an unconventional signature and this needs to be ironed out.

pradovic:
@acud There is a subtle difference between map-like structure and a map structure, and I used a map as an example because of the similar API. Basically, to me, I focus on the interface. Just a couple of arguments on how I see it:
1. I don't think I silenced an error, but just used another way to communicate not found situation. Other errors are still propagated, as you can see in ln63
2. I don't insist on using it like a map, I am just saying it has an map-like API
3. To me it's weird to have a not found error, like it's an error to not have an element. I feel like this is not an erroneous and find using an error to indicate this state is a bit weird. Especially since Go language is not the king of error comparisons :)

Maybe the middle ground is an options 3, using a special error, so i will switch back to it.

PR #320, opened 2020-06-17
Porting Swarm encryption in Bee

jmozah:
- Add encryption headers for /bytes and /files APIs
- If they are present , encrypt the chunks and store it
- while joining, decrypt and join the chunks


pkg/api/file.go
```diff
@@ -28,7 +30,10 @@ import (
 	"github.com/gorilla/mux"
 )
 
-const multipartFormDataMediaType = "multipart/form-data"
+const (
+	MultiPartFormData = "multipart/form-data"
```
janos, line 34: I do not see the need to export this variable.

pkg/file/splitter/internal/job.go
```diff
@@ -129,31 +139,45 @@ func (s *SimpleSplitterJob) sumLevel(lvl int) ([]byte, error) {
 
 	sizeToSum := s.cursors[lvl] - s.cursors[lvl+1]
 
-	// perform hashing
+	var ref encryption.Key
+	var chunkData []byte
+	data := s.buffer[s.cursors[lvl+1] : s.cursors[lvl+1]+sizeToSum]
```
janos, line 144: This can be done after SetSpan and avoid the allocation in case of an error. Also I am not sure why ref and chunkData declarations are moved here.

go.sum
```diff
@@ -1,8 +1,28 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+cloud.google.com/go v0.31.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
janos, line 2: go.mod is not changed, we should leave go.sum also without any changes.

pkg/api/bytes.go
```diff
@@ -37,6 +47,31 @@ func (s *server) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+func (s *server) splitUpload(ctx context.Context, r io.Reader, l int64, toEncrypt bool) (swarm.Address, error) {
```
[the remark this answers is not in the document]
jmozah: Because i think when i branched from master this was not there. Anyway i merged them now.

go.sum
jmozah: i took the so.sum from master.. i dont know how this creaps in.. let me do it again.

Eknir:
@jmozah , please make an issue on /ethersphere/bee-documentation to document this feature. Thanks!

PR #324, opened 2020-06-18
Announce order

pradovic:
The ordering of announce seems a bit different to me in the case of connect from manage and connected handler. @acud Could you pls take a look and verify if this is more correct? 
It seems to me that in manage it is first announced and then clean-ed up from maps, and in connected is vice versa + there is a manageC triggered in between, but I might be wrong 🙈

### DIFF
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -337,6 +337,10 @@ func (k *Kad) AddPeer(ctx context.Context, addr swarm.Address) error {
 
 // Connected is called when a peer has dialed in.
 func (k *Kad) Connected(ctx context.Context, addr swarm.Address) error {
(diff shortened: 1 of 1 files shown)


janos:
Let's make the base to `pull-0-update` so that we test with it on the cluster.

acud:
can we please have a problem statement?

pradovic:
@acud I saw that there is a race condition that is happening somewhere around announce, so I investigated and saw there is a different ordering. I am not sure this is the cause of the problem, just wanted to check with you if this makes sense. This is why I thought that connected handler should have a uniform code with connect call from managec, as it is basically, the same.  From what I sere here, it is similar, but the ordering is different.

pradovic:
@janos I have changed the base, sorry did not see before.

PR #327, opened 2020-06-18
Improve code in api/bytes.go

nolash:
There are not really big changes here in this PR, more or less just:

* removing the loglines in the goroutine, replacing them with errors (that we perhaps could just drop is what I asked, simplifying quite a bit)
* moving the splitUpload to files a method SplitWriteAll, which mirrors the already existing facilitator JoinReadAll.
* ~~typed error from splitter~~

### DIFF
--- a/pkg/file/error.go
+++ b/pkg/file/error.go
@@ -49,25 +49,3 @@ func (e *HashError) Unwrap() error {
 func (e *HashError) Error() string {
 	return e.err.Error()
 }
-
-// SplitError should wrap any error returned from the splitter.
-type SplitError struct {
-	err error
-}
-
-// NewSplitError returns a new SplitError instance.
-func NewSplitError(err error) error {
-	return &SplitError{
-		err: err,
-	}
-}
-
-// Unwrap returns an underlying error.
-func (e *SplitError) Unwrap() error {
-	return e.err
-}
-
-// Error implements standard go error interface.
-func (e *SplitError) Error() string {
-	return e.err.Error()
-}
--- a/pkg/file/error.go
+++ b/pkg/file/error.go
@@ -22,7 +22,7 @@ func (e *AbortError) Unwrap() error {
 	return e.err
 }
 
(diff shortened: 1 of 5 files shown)


nolash:
@janos I am unsure whether we really need to handle the errors and write mismatch in the go-routine. Would it be valid to leave the respondibility to the splitter to complain if something is wrong?

janos:
@nolash thanks, I am not sure if such a big change is needed. I think that @jmozah's approach with buffer handles the error case nicely https://github.com/ethersphere/bee/blob/master/pkg/api/file.go#L261.

It writes first to the buffer, checks if there is an error and if something is written and decides whether to serve the json error response or continue writing to the http.ResponseWriter trough the buffer. This approach requires much less changes to the code. Would that make sense to be done for bytes endpoint as well and avoid drastic changes? We have already improved splitUpload method by returning only the address.

nolash:
@janos I am not sure I understand, this is about the splitter not joiner? your link points to the join consumer.

Also there are not really big changes here in this PR, more or less just:

- removing the loglines in the goroutine
- moving the `splitUpload` to files a method `SplitWriteAll`, which mirrors the already existing facilitator `JoinReadAll`.
- typed error from splitter

nolash:
Aha, you are talking about another task. Yes I can make the improvement.

pkg/file/error.go
```diff
@@ -45,7 +45,29 @@ func (e *HashError) Unwrap() error {
 	return e.err
 }
 
-// Error implement standard go error interface.
+// Error implements standard go error interface.
 func (e *HashError) Error() string {
 	return e.err.Error()
 }
+
+// SplitError should wrap any error returned from the splitter.
```
janos, line 53: Split error type is used here to wrap other errors, but I do not see that it is checked anywhere, it is not used. I think that it is not needed until we explicitly handle it. I suggest to remove it and use fmt.Errorf to wrap errors, just as it it was.

PR #328, opened 2020-06-18
implement skip peers for ClosestPeerer

janos:
Currently, the retrieval asks only the closest peer (or fails if the closest one is self) to get the chunk. This PR makes retrieval more resilient by asking multiple peers by their proximity to the chunk and also makes Kademlia never return self.

This PR contains signature change hanse the number of lines changed, but the crucial parts are in the retrieval RetrieveChunk and kademlia ClosestPeer.

This change greatly improved the retrieval success rate when used by node connected to the public cluster where syncing is active and new nodes connect and disconnect.

I am basing this PR on `pull-0-update` to be able to test it with latest pull sync and connectivity fixes on the public cluster. Before it is merged to master.

This PR also changes some responses in api download file to NotFound instead InternalServerError to indicate that the file could not be found as result of errors in join from unsuccessful retrieval.

### DIFF
--- a/pkg/kademlia/kademlia_test.go
+++ b/pkg/kademlia/kademlia_test.go
@@ -377,7 +377,7 @@ func TestAddressBookPrune(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if !nonConnPeer.Equal(&p) {
+	if !nonConnPeer.Equal(p) {
 		t.Fatalf("expected %+v, got %+v", nonConnPeer, p)
 	}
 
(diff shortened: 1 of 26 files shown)


janos:
@acud I have improved the changes based on your suggestions but had to create a specific closestPeer in order to get peers ordered by proximity with skip peers, as peers are not ordered by proximity in EachPeerRev.

@pradovic this is probably a better implementation then that you have already reviewed, could you check it again?

pkg/kademlia/kademlia_test.go
```diff
@@ -480,10 +480,6 @@ func TestClosestPeer(t *testing.T) {
 			chunkAddress: swarm.MustParseHexAddress("5000000000000000000000000000000000000000000000000000000000000000"), // 0101, want peer 1
 			expectedPeer: 1,
 		},
-		{
```
acud, line 483: should this be removed? it seems like this functionality is still there and should be tested no?

janos: Yes, of course, I missed this to put back. Thanks.

PR #330, opened 2020-06-19
pusher: disable broken tags and make pusher push chunks in parallel

acud:
Tags are not working correctly and are polluting the terminal output.
Also, pusher pushes every chunk synchronously to NN, making the process painstakingly slow.
This PR adds some parallelism to the push process, therefore speeding it up.


Eknir:
I second this. I add it to the DISC release

PR #332, opened 2020-06-19
kademlia: add timeout to connect

acud:
this PR tries to solve a problem where libp2p connect might block for a significant amount of time, possibly slowing down kademlia iterations.
Therefore a timeout has been added to `connect` method.
The problem was noticed when shutting down nodes, a goroutine trying to connect to a peer would always write to the log at the edge of the log stream.


pkg/kademlia/kademlia.go
```diff
@@ -251,6 +251,7 @@ func (k *Kad) recalcDepth() uint8 {
 // connect connects to a peer and gossips its address to our connected peers,
 // as well as sends the peers we are connected to to the newly connected peer
 func (k *Kad) connect(ctx context.Context, peer swarm.Address, ma ma.Multiaddr, po uint8) error {
+	ctx = context.WithTimeout(ctx, 5*time.Second)
```
janos, line 254: This needs a cancel return parameter.

PR #334, opened 2020-06-19
Fix getting tags by address in pusher

jmozah:
tags are fetched by address in pusher which is wrong. They should be fetched by the uid in the chunk (if present, /bytes & /files api's don't have tags) and stored in a local map. When a receipt is received, the state in the local store should be synced and the tag counter for synced should be incremented.

### DIFF
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -7,34 +7,27 @@ package pusher
 import (
 	"context"
 	"errors"
-	"fmt"
-	"sync"
 	"time"
 
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/pushsync"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
-	"github.com/ethersphere/bee/pkg/tags"
 	"github.com/ethersphere/bee/pkg/topology"
 )
 
 type Service struct {
 	storer            storage.Storer
 	pushSyncer        pushsync.PushSyncer
-	tag               *tags.Tags
 	logger            logging.Logger
 	metrics           metrics
 	quit              chan struct{}
 	chunksWorkerQuitC chan struct{}
-	pushed            map[string]uint32 // address to tag Id
-	pushedMu          sync.Mutex
 }
 
 type Options struct {
 	Storer        storage.Storer
 	PeerSuggester topology.ClosestPeerer
-	Tags          *tags.Tags
 	PushSyncer    pushsync.PushSyncer
 	Logger        logging.Logger
 }
@@ -45,12 +38,10 @@ func New(o Options) *Service {
 	service := &Service{
 		storer:            o.Storer,
 		pushSyncer:        o.PushSyncer,
-		tag:               o.Tags,
 		logger:            o.Logger,
 		metrics:           newMetrics(),
 		quit:              make(chan struct{}),
 		chunksWorkerQuitC: make(chan struct{}),
-		pushed:            make(map[string]uint32),
 	}
 	go service.chunksWorker()
 	return service
@@ -89,26 +80,14 @@ func (s *Service) chunksWorker() {
 			chunksInBatch++
 			s.metrics.TotalChunksToBeSentCounter.Inc()
 
-			t, err := s.tag.Get(ch.TagID())
-			if err != nil {
-				s.logger.Tracef("pusher: get tag by uid %s: %v", ch.Address(), err)
-				//continue // // until bzz api implements tags, dont continue here
-			}
-
 			// Later when we process receipt, get the receipt and process it
 			// for now ignoring the receipt and checking only for error
-			_, err = s.pushSyncer.PushChunkToClosest(ctx, ch)
+			_, err := s.pushSyncer.PushChunkToClosest(ctx, ch)
 			if err != nil {
 				if !errors.Is(err, topology.ErrNotFound) {
 					s.logger.Errorf("pusher: error while sending chunk or receiving receipt: %v", err)
 				}
 				continue
-			}
-
-			if t != nil {
-				// update the tags only if we get it
-				t.Inc(tags.StateSent)
-				s.addUidToPushed(ch.Address().String(), t.Uid)
 			}
 
 			// set chunk status to synced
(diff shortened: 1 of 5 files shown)


pkg/pusher/pusher.go
```diff
@@ -84,13 +89,14 @@ func (s *Service) chunksWorker() {
 			chunksInBatch++
 			s.metrics.TotalChunksToBeSentCounter.Inc()
 
-			t, err := s.tag.GetByAddress(ch.Address())
+			t, err := s.tag.Get(ch.TagID())
 			if err != nil {
-				s.logger.Debugf("pusher: get tag by address %s: %v", ch.Address(), err)
-				//continue // // until bzz api implements tags dont continue here
+				s.logger.Debugf("pusher: get tag by uid %s: %v", ch.Address(), err)
+				//continue // // until bzz api implements tags, dont continue here
```
acud, line 95: can we please remove this? also, i dont see why we should ever continue here. continue means the chunk will never get pushed, will never get set as synced, and will result in index leakages.

pkg/pusher/pusher.go
```diff
@@ -103,6 +105,12 @@ func (s *Service) chunksWorker() {
...
 			}
 
+			if t != nil {
+				// update the tags only if we get it
+				t.Inc(tags.StateSent)
+				s.addUidToPushed(ch.Address().String(), t.Uid)
+			}
+
 			// set chunk status to synced
 			s.setChunkAsSynced(ctx, ch.Address())
```
acud, line 115: why don't you just add another parameter `*uint32` to `setChunkAsSynced` instead of adding a new field to the struct? and the mutex, and getters, delete, etc, wouldn't it be cleaner to just add one parameter here?

jmozah: > why don't you just add another parameter *uint32 to setChunkAsSynced instead of adding a new field to the struct? and the mutex, and getters, delete, etc, wouldn't it be cleaner to just add one parameter here?

@acud We were talking about making the pusher asyc right? we said it blocks until it gets receipt and does only 5 req/sec. That's the reason i moved to the old model of storing the chunk until we get a receipt. Anyway i will move to the model as you are suggesting now and come back to this when we do async.


>statesynced should be incremented after the receipt is received.

@Zelig True for intermediate nodes. But it is true for the sending node also.  Should we increment the tag after we set the ModeSetSyncPush in DB without errors? This is tricky to do. Fo now i would suggest i set the tag as soon as we receive receipt in all places. WDYS?

acud: > for intermediate nodes

intermediate nodes should not do any tag operations.

pkg/pusher/pusher.go
```diff
@@ -84,13 +89,10 @@ func (s *Service) chunksWorker() {
...
 
-			t, err := s.tag.GetByAddress(ch.Address())
+			t, err := s.tag.Get(ch.TagID())
 			if err != nil {
-				s.logger.Debugf("pusher: get tag by address %s: %v", ch.Address(), err)
-				//continue // // until bzz api implements tags dont continue here
-			} else {
-				// update the tags only if we get it
-				t.Inc(tags.StateSent)
+				s.logger.Tracef("pusher: get tag by uid %s: %v", ch.Address(), err)
```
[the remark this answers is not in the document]
jmozah: removed the line as this makes no sense before we implement  tags in /bytes and /files API.

pkg/pusher/pusher.go
jmozah: Then all the tags operation should be in pusher... not in protocol.

jmozah: > intermediate nodes should not do any tag operations.

@acud 
https://github.com/ethersphere/bee/pull/330/files PR removed what you you are referring. 
1)  If you are sending chunks async, we needed a a place to store them till we get receipt. 
2) if you should not store tags in intermediate nodes, then we should store them as i did before   in pusher.

What i fail to understand is why tags was removed in that PR altogether !!!!

pkg/pushsync/pushsync.go
```diff
@@ -134,13 +138,17 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
 	if err := ps.sendChunkDelivery(wc, chunk); err != nil {
 		return fmt.Errorf("forward chunk to peer %s: %w", peer.String(), err)
 	}
 	receiptRTTTimer := time.Now()
 
-	receipt, err := ps.receiveReceipt(rc)
+	// Most of the times now you wont get a tag because
+	// /bytes and /files API does not implement tags
+	// SO dont print any logs or returning error, if you dont find them
+	t, _ := ps.tagger.Get(chunk.TagID())
```
zelig, line 149: you only need  to call this once in handler.
be consistent. call it before `sendChunkDelivery`
and pass `t` to both `sendChunkDelivery` and `receiveReceipt`

pkg/pushsync/pushsync.go
```diff
@@ -134,13 +138,17 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
 
 	wc, rc := protobuf.NewWriterAndReader(streamer)
-
 	if err := ps.sendChunkDelivery(wc, chunk); err != nil {
 		return fmt.Errorf("forward chunk to peer %s: %w", peer.String(), err)
 	}
 	receiptRTTTimer := time.Now()
 
-	receipt, err := ps.receiveReceipt(rc)
+	// Most of the times now you wont get a tag because
```
zelig, line 146: this comment is not necessary, or turn into  a TODO

pkg/pushsync/pushsync.go
```diff
@@ -187,6 +195,16 @@ func (ps *PushSync) sendChunkDelivery(w protobuf.Writer, chunk swarm.Chunk) (err
 	}
 	ps.metrics.SendChunkTimer.Observe(time.Since(startTimer).Seconds())
 	ps.metrics.ChunksSentCounter.Inc()
+
+	//  if you manage to get a tag, just increment the respective counter
+	t, err := ps.tagger.Get(chunk.TagID())
+	if err == nil && t != nil {
+		// most of the times now you wont get a tag because
```
zelig, line 202: this comment is not necessary, or turn into  a TODO

pkg/pushsync/pushsync.go
jmozah: SendChunkDelivery is called from 2 places. one from this handler and the other from pusher. We have 2 choise

1) implemet this in handler and in pusher
2) implement it her and in SendChunkDelivery

if you chose the second method, then all the things related to tag stays in one package. Thats why i chose this method.

zelig: forwarded chunks do not have a tag, and therefore checking within the function or even passing  it  is incorrect.
Sorry all  I missed this, so no, this handler should have at no place do anything from this 
In fact this points to the solution being that not only tag calls should also be directly within the PushToClosest function body only (which is called from the pusher), the increments  of StateSent and StateSynced should come right after error checking `sendChunkDelivery` and the `setChunkSynced` calls, respectively.

jmozah:
@zelig since you were saying to review tags in today's daily , i  rebased to the latest master and resolved the conflicts.

pkg/pushsync/pushsync.go
jmozah: the above you described which was the original implementation.. I will fix this back.

jmozah:
@zelig Now the tags related stuff is inside PushToClosest and i added  a testcase also for that

jmozah:
> now LGTM

jmozah:
> now LGTM

To be more sure, we can increment the StateSynced counter after it is set in the localstore.

acud:
> To be more sure, we can increment the StateSynced counter after it is set in the localstore.

That indeed would be better

jmozah:
@acud addressed your minor comments. can i merge?

acud:
thanks @jmozah. i approved since these were not blockers

PR #336, opened 2020-06-19
Fullclose on incoming connection + stream close on send/handle headers

pradovic:
- The both incoming and sending side should wait for each other after adding the node to the peer registry, so we have a consistent situation when initiating protocols.
- handle/send headers stream close per https://github.com/ethersphere/bee/pull/354

### DIFF
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -286,9 +286,7 @@ func (s *Service) AddProtocol(p p2p.ProtocolSpec) (err error) {
 			if err := handleHeaders(ss.Headler, stream); err != nil {
 				s.logger.Debugf("handle protocol %s/%s: stream %s: peer %s: handle headers: %v", p.Name, p.Version, ss.Name, overlay, err)
 				s.logger.Errorf("handle protocol %s/%s: peer %s", p.Name, p.Version, overlay)
-				if err := stream.Close(); err != nil {
-					s.logger.Debugf("handle protocol %s/%s: stream %s: peer %s: handle headers close stream: %v", p.Name, p.Version, ss.Name, overlay, err)
-				}
+				_ = s.disconnect(peerID)
 				return
 			}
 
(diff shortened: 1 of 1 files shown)


acud:
@pradovic are we proceeding with this for the release?

pradovic:
@acud we wanted to wait a bit on Friday. This should be more correct and cut thr wait in the case of timeout. We canmerve it, but can also wait if we don't want to risk stability. @janos what do you think as well?

janos:
This commit https://github.com/ethersphere/bee/pull/336/commits/9884941c596f794ba66ce885a5c635e662325ec2 was reverted on https://github.com/ethersphere/bee/pull/354.

pradovic:
@janos I have reverted disconnect here as well then, but I left the fullclose sync it should be more correct

pkg/p2p/libp2p/libp2p.go
```diff
@@ -285,6 +285,10 @@ func (s *Service) AddProtocol(p p2p.ProtocolSpec) (err error) {
 			// exchange headers
 			if err := handleHeaders(ss.Headler, stream); err != nil {
 				s.logger.Debugf("handle protocol %s/%s: stream %s: peer %s: handle headers: %v", p.Name, p.Version, ss.Name, overlay, err)
+				s.logger.Errorf("handle protocol %s/%s: peer %s", p.Name, p.Version, overlay)
```
acud, line 288: sure we want this on `error`? maybe `debug`?

pradovic: I generally agree, but there will be no message for the user then, cuz we don't return an error here. Maybe warning?

acud: this is not really a user-facing error anyway... not sure what the mere mortal is supposed to make of this sort of error. it is pretty obscure. i'd even go as far as saying just have the debug line above and remove this line. anyway, as you wish

pradovic: Cool, good for me :)

PR #339, opened 2020-06-19
Use files join code in bytes

nolash:
Addresses cleanup request in https://github.com/ethersphere/bee/pull/327#issuecomment-646057120

### DIFF
--- a/pkg/api/bytes.go
+++ b/pkg/api/bytes.go
@@ -5,8 +5,10 @@
 package api
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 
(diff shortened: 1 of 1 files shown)


pkg/api/bytes.go
```diff
@@ -66,12 +68,19 @@ func (s *server) bytesGetHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	outBuffer := bytes.NewBuffer(nil)
+	c, err := file.JoinReadAll(j, address, outBuffer)
+	if err != nil && c == 0 {
```
acud, line 73: so if the error is not nil but c>0 then all good?

nolash:
@janos please see comment from @acud

PR #345, opened 2020-06-20
Ethereum address creation faulty

nolash:
Hashing of ethereum address by public key was not called correctly.

See https://github.com/ethersphere/bee/pull/345/files#diff-aa4c96b4d3b1b4d4fcd8f914144cd943L59

### DIFF
--- a/pkg/crypto/crypto_test.go
+++ b/pkg/crypto/crypto_test.go
@@ -38,7 +38,10 @@ func TestNewAddress(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	a := crypto.NewOverlayAddress(k.PublicKey, 1)
+	a, err := crypto.NewOverlayAddress(k.PublicKey, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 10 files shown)


pkg/crypto/crypto_test.go
```diff
@@ -37,7 +38,10 @@ func TestNewAddress(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	a := crypto.NewOverlayAddress(k.PublicKey, 1)
+	a, err := crypto.NewOverlayAddress(k.PublicKey, 1)
```
acud, line 41: how is it that here `NewOverlayAddress` is called with `1` and in the test above it is `3`?

nolash: no idea, I just left is as it was

acud: k

pradovic: @acud  That is a networkID argument. If you think the tests in bzz/address_test.go by test above, I think I just used number 3 randomly :)

PR #347, opened 2020-06-22
Introduce soc validator

nolash:
Introduces validation for single-owner chunks as per specifications in the book of swarm.

A new package `soc` is introduced which encapsulates a soc `Update` that works with or without signer, and theoretically can be used as a library to build and sign soc chunks.

It also dissolves the `validation` package, moving the `soc` validator to the `soc` package, and introducing the `content` package for both validator and convenience functions for _content addressed chunks_.


pkg/soc/soc.go
```diff
@@ -0,0 +1,243 @@
...
+// WithSpan is a chainable function that adds an arbitrary
+// span to the chunk data defined for the update.
+func (s *Update) WithSpan(span int64) *Update {
+	s.span = span
+	return s
+}
+
+// WithOwnerAddress provides the possibility of setting the ethereum
+// address for the owner of an update in the absence of a signer.
+func (s *Update) WithOwnerAddress(ownerAddress *Owner) *Update {
```
pradovic, line 77: The shorter way of setting multiple stuff could be an `Option` struct with span and owner address fields, but this is good for me as well. Just a suggestion ;)

pkg/soc/soc.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// Address returns the soc Chunk address of the update.
+func (s *Update) Address() (swarm.Address, error) {
+	var err error
+	hasher := swarm.NewHasher()
+	_, err = hasher.Write(s.id)
+	if err != nil {
+		return swarm.ZeroAddress, err
+	}
```
pradovic, line 117:
```suggestion
      if err := hasher.Write(s.id); err != nil {
		return swarm.ZeroAddress, err
	}
```

pkg/soc/soc.go
```diff
@@ -0,0 +1,243 @@
...
+	_, err = hasher.Write(s.owner.address)
+	if err != nil {
+		return swarm.ZeroAddress, err
+	}
```
pradovic, line 121:
```suggestion

	if err := hasher.Write(s.owner.address); err != nil {
		return swarm.ZeroAddress, err
	}
```

pkg/soc/soc.go
```diff
@@ -0,0 +1,243 @@
...
+	addressBytes := hasher.Sum(nil)
+	return swarm.NewAddress(addressBytes), nil
```
pradovic, line 123:
```suggestion
	return swarm.NewAddress(hasher.Sum(nil)), nil
```

pkg/soc/soc.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// add all the data fields
+	update := &Update{}
+	cursor := 0
+
+	update.id = chunkData[cursor : cursor+IdSize]
+	cursor += IdSize
+
+	update.signature = chunkData[cursor : cursor+SignatureSize]
+	cursor += SignatureSize
```
pradovic, line 141: Could you explain to me the cursors values in this function please? 🙏

pkg/soc/soc.go
nolash: The owner address is set in on of two ways:

* this call
* set the signer, which implicitly sets the address

I felt that having the chainable approach makes it seem more like an optional setting.

pkg/soc/soc.go
nolash: the cursor points to the offset position in the byte buffer from which we read the data to be deserialized for a particular type. For example, here signature follows the chunk "id" in the serialized byte vector, and you read a slice from there until the known length of a signature.

pkg/soc/soc.go
nolash: terribly sorry don't like this form

pkg/soc/soc.go
nolash: terribly sorry but I find it give better clarity to name each step of mutations as much as possible in any code.

pkg/soc/soc.go
```diff
@@ -0,0 +1,243 @@
...
+func NewOwner(address []byte) (*Owner, error) {
+	if len(address) != AddressSize {
+		return nil, fmt.Errorf("invalid address %x", address)
+	}
+	return &Owner{
+		address: address,
+	}, nil
+}
+
+// Update wraps a single soc update.
```
zelig, line 45: how the notion of update made its way here.
I dont think there. is update on the level of soc

pkg/soc/soc.go
```diff
@@ -0,0 +1,243 @@
...
+		return nil, err
+	}
+	_, err = bmtHasher.Write(update.payload)
+	if err != nil {
+		return nil, err
+	}
+	payloadSum := bmtHasher.Sum(nil)
+
+	// recover owner information
+	recoveredPublicKey, err := crypto.Recover(update.signature, payloadSum)
```
zelig, line 165: it is not the payloadSum that is signed. but the payloadSum against the id

pkg/soc/soc.go
```diff
@@ -0,0 +1,243 @@
...
+		return nil, err
+	}
+	_, err = bmtHasher.Write(s.payload)
+	if err != nil {
+		return nil, err
+	}
+	payloadSum := bmtHasher.Sum(nil)
+
+	// sign the update
+	signature, err := s.signer.Sign(payloadSum)
```
zelig, line 205: it is not the payloadSum that is signed. but the payloadSum against the id

pkg/soc/soc.go
nolash: in that case I would encourage revising the figure in the book.

pkg/soc/soc.go
nolash: ... as it looks like it could also be the address it's signed with. Also wonder why there is a dotted line.

nolash:
Agreed, a bit of DRYing wouldn't hurt.

nolash:
> also not sure why separate validator pkg is needed

This is how we chose to implement validation for content addressed chunks at the time. It seems to benefit clarity to have soc validation in the same place.

pkg/soc/soc.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil {
+		return nil, err
+	}
+	_, err = bmtHasher.Write(s.payload)
+	if err != nil {
+		return nil, err
+	}
+	payloadSum := bmtHasher.Sum(nil)
+	sha3Hasher := swarm.NewHasher()
+	_, err = sha3Hasher.Write(payloadSum)
```
jmozah, line 215: per BOS, first write should be `ID` and then `payloadSum`

pkg/soc/soc.go
```diff
@@ -0,0 +1,264 @@
...
+
+	sch.signature = chunkData[cursor : cursor+SignatureSize]
+	cursor += SignatureSize
+
+	spanBytes := chunkData[cursor : cursor+swarm.SpanSize]
+	span := binary.LittleEndian.Uint64(spanBytes)
+	sch.span = int64(span)
+	cursor += swarm.SpanSize
+
+	sch.payload = chunkData[cursor:]
```
jmozah, line 148: if len(chunkData) < (minChunkSize + sch.span) {
    error
}

nolash: I'm not sure I understand. Minimum length is checked in L128.

pkg/soc/soc.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil, err
+	}
+	_, err = sha3Hasher.Write(s.id)
+	if err != nil {
+		return nil, err
+	}
+	toSignBytes := sha3Hasher.Sum(nil)
+
+	// sign the update
+	signature, err := s.signer.Sign(toSignBytes)
```
zelig, line 226: i. dont think you need to hash them together, you can just use 

     toSignBytes. = bmtHash | ID

pkg/soc/soc.go
jmozah: after we get the spanSize, is it not good to check if we have data for that span?

nolash:
@zelig "rewrite the code"

ok. got you. thanks.

pkg/soc/soc.go
nolash: Reducing to a hash before signing is a fairly common concept, though?

nolash: I think it's a fairly common construct to sign digests rather than arbitrary length bytes.

In any case, the signer implementation seems to be acting up when trying to sign this concatenation. 

May I suggest that you adjust the spec to specify signature is digest instead?

@zelig

nolash:
@zelig I've added convenience accessors re your comments.

I find it pertinent to have a representation of soc which is independent of whether you have the signing key or not, for anyone wanting to manipulate this directly. I've DRY'ed the code, and also added the previously non-existent "content addressed" package which you solicited in your review by splitting it out from the package previously called validator.

pkg/soc/soc_test.go
```diff
@@ -0,0 +1,130 @@
...
+		t.Fatal(err)
+	}
+
+	toSignBytes, err := soc.ToSignDigest(id, ch.Address().Bytes())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// verify owner match
+	//recoveredPublicKey, err := crypto.Recover(signature, toSignBytes)
```
zelig, line 77: commented code

pkg/soc/soc.go
zelig: yes but the `signer.Sign` does that anyway no?

nolash: doesnt seem to.

pkg/content/content.go
```diff
@@ -0,0 +1,82 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/ethersphere/bee/pkg/swarm"
+	bmtlegacy "github.com/ethersphere/bmt/legacy"
+)
+
+// NewContentChunk creates a new content-addressed single-span chunk.
+// The length of the chunk data is set as the span.
+func NewContentChunk(data []byte) (swarm.Chunk, error) {
```
pradovic, line 19: Since the package is content, you could also just use a word Chunk instead of ContentChunk, because the usage of exported stuff will be `content.Chunk` or `content.ContentChunk`, choose whatever looks better for you :)

pkg/content/validator.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+var _ swarm.ChunkValidator = (*ContentAddressValidator)(nil)
+
+// ContentAddressValidator validates that the address of a given chunk
+// is the content address of its contents.
+type ContentAddressValidator struct {
```
pradovic, line 15: Same as the above, you can ommit `Content` in naming inside `content` package if you like.

pkg/soc/soc.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	_, err = h.Write(sum)
+	if err != nil {
+		return nil, err
+	}
+	return h.Sum(nil), nil
+}
+
+// CreateAddress creates a new soc address from the soc id and the ethereum address of the signer
+func CreateAddress(id Id, owner *Owner) (swarm.Address, error) {
```
pradovic, line 209: `NewAddress` might be more idiomatic.

nolash: I am not sure. `New` suggests - to me at least - instantiate an object, reserve memory etc. This is more like a helper function.

pradovic: Makes sense!

PR #356, opened 2020-06-24
kademlia: handle overlay mismatch

acud:


acud:
thanks for fixing this @pradovic <3

PR #365, opened 2020-06-24
Al rc1 disconnects

metacertain:
Upon experiencing a low number of connections being maintained after establishment, the changes included in this [commit ](https://github.com/ethersphere/bee/pull/365/commits/5149a5803006fd7fd4764f92297a0ade1d2a3225) helped to get a reasonable amount of peer connections

### DIFF
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -23,15 +23,15 @@ import (
 
 const (
 	maxBins         = 16
-	nnLowWatermark  = 2 // the number of peers in consecutive deepest bins that constitute as nearest neighbours
-	maxConnAttempts = 3 // when there is maxConnAttempts failed connect calls for a given peer it is considered non-connectable
+	nnLowWatermark  = 4  // the number of peers in consecutive deepest bins that constitute as nearest neighbours
+	maxConnAttempts = 10 // when there is maxConnAttempts failed connect calls for a given peer it is considered non-connectable
 )
 
 var (
 	errMissingAddressBookEntry = errors.New("addressbook underlay entry not found")
 	errOverlayMismatch         = errors.New("overlay mismatch")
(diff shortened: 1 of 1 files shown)


pkg/kademlia/kademlia.go
```diff
@@ -298,8 +305,8 @@ func (k *Kad) connect(ctx context.Context, peer swarm.Address, ma ma.Multiaddr,
 	}
 
 	if !i.Overlay.Equal(peer) {
-		_ = k.p2p.Disconnect(peer)
-		_ = k.p2p.Disconnect(i.Overlay)
+		//_ = k.p2p.Disconnect(peer)
```
acud, line 308: this should not be commented out. both lines

pkg/kademlia/kademlia.go
```diff
@@ -330,7 +337,7 @@ func (k *Kad) announce(ctx context.Context, peer swarm.Address) error {
 
 	err := k.discovery.BroadcastPeers(ctx, peer, addrs...)
 	if err != nil {
-		_ = k.p2p.Disconnect(peer)
+		// _ = k.p2p.Disconnect(peer)
```
acud, line 340: why?

acud:
this looks like base is wrong. `-> alpha-rc1?`

pkg/kademlia/kademlia.go
metacertain: A couple of things can go wrong here (namely in sendPeers)
stream, err := s.streamer.NewStream(ctx, peer, nil, protocolName, protocolVersion, peersStreamName)
addr, err := s.addressBook.Get(p)
if err := w.WriteMsg(&peersRequest); err != nil 

What I was experiencing was losing almost every connection within a couple of minutes of establishment, so I just commented this out so that I potentially don't lose so many connections

Eknir:
What is the status of this PR?

PR #370, opened 2020-06-25
puller, pull sync: clean shutdown

### LINKED ISSUE #360: Panic in local store
The local store panicked today twice in my setup (see prt screen below).

It happened on an empty data directory, maybe 10 seconds after I started the node, just after I shut down my node.

It can't be reproduced reliably.

acud:
fixes #360

### DIFF
--- a/pkg/pullsync/pullsync.go
+++ b/pkg/pullsync/pullsync.go
@@ -50,6 +50,8 @@ type Syncer struct {
 	streamer p2p.Streamer
 	logger   logging.Logger
 	storage  pullstorage.Storer
+	quit     chan struct{}
+	wg       sync.WaitGroup
 
 	ruidMtx sync.Mutex
 	ruidCtx map[uint32]func()
@@ -71,6 +73,8 @@ func New(o Options) *Syncer {
 		storage:  o.Storage,
 		logger:   o.Logger,
 		ruidCtx:  make(map[uint32]func()),
+		wg:       sync.WaitGroup{},
+		quit:     make(chan struct{}),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


acud:
@pradovic it is accustomed to add to waitgroup before the goroutine is scheduled, and these workers work in different goroutines, so i add to the waitgroup before calling them with `go ...`.

acud:
that is, at least from what i know. since goroutines might not be scheduled for a while, you are risking that it will add to the waitgroup after `wg.Wait()` has already returned

pradovic:
@acud I think this is done if the wait is right after in order to prevent fot wg.Wait() to happen before wg.Add(). Can this happen here actually? We should guard against wg.Add() happening after wg.Wait() is called in the Close().

acud:
Well, how do you know that a shutdown is not in progress? what is the difference between what you show above and the situation in the code in the PR?

pkg/pullsync/pullsync.go
```diff
@@ -398,5 +410,17 @@ func (s *Syncer) cancelHandler(ctx context.Context, p p2p.Peer, stream p2p.Strea
 }
 
 func (s *Syncer) Close() error {
+	s.logger.Info("pull syncer shutting down")
+	close(s.quit)
+	cc := make(chan struct{})
+	go func() {
+		defer close(cc)
+		s.wg.Wait()
```
pradovic, line 418: Maybe this needs to wait for for loop in manage to exit before calling `wg.Wait()` as more go routines might be added?

pradovic:
@acud Maybe I am wrong, if I am, sorry 🙈  But it looks to me that the code in manage loop would still be active, and can do wg.Add() calls so it looked to me that we need to guard for this anyway as these calls are in a manage() goroutine.

PR #371, opened 2020-06-25
pullsync: add explicit termination

acud:


pkg/pullsync/pullsync.go
```diff
@@ -320,6 +361,42 @@ func (s *Syncer) cursorHandler(ctx context.Context, p p2p.Peer, stream p2p.Strea
...
 }
 
+func (s *Syncer) CancelRuid(peer swarm.Address, ruid uint32) error {
+	stream, err := s.streamer.NewStream(context.Background(), peer, nil, protocolName, protocolVersion, cancelStreamName)
+	if err != nil {
+		return fmt.Errorf("new stream: %w", err)
+	}
+
+	w := protobuf.NewWriter(stream)
+	defer stream.Close()
```
pradovic, line 371: Maybe fullclose if we want to know that cancel succeeded?

PR #374, opened 2020-06-25
kademlia: saturate based on known peers

acud:
This PR changes the saturation function to use both known and connected peers for depth calculation, avoiding piling up peers in shallow bins when initially bootstrapping the topology. This results in much more stable topology. We also after some experimentation raised the number of peers needed to saturate a bin to `4`, referred to as `saturationPeers` thereafter in kademlia.

### DIFF
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -34,7 +34,7 @@ var (
 	shortRetry                 = 30 * time.Second
 )
 
-type binSaturationFunc func(bin, depth uint8, peers, connected *pslice.PSlice) bool
+type binSaturationFunc func(bin uint8, peers, connected *pslice.PSlice) bool
 
 // Options for injecting services to Kademlia.
 type Options struct {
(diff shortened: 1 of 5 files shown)


pkg/kademlia/kademlia.go
```diff
@@ -210,9 +210,11 @@ func (k *Kad) manage() {
 // binSaturated indicates whether a certain bin is saturated or not.
 // when a bin is not saturated it means we would like to proactively
 // initiate connections to other peers in the bin.
-func binSaturated(bin, depth uint8, peers *pslice.PSlice) bool {
+func binSaturated(bin, depth uint8, peers, connected *pslice.PSlice) bool {
```
janos, line 213: Looks like that depth argument is no longer used.

acud: true

janos:
This test is failing:

```
=== RUN   TestBinSaturation
##[error]    TestBinSaturation: kademlia_test.go:228: timed out waiting for counter to reach expected value. got 4 want 0
```

https://github.com/ethersphere/bee/pull/374/checks?check_run_id=810914686#step:11:423

I think that this branch can be merged after the test is adjusted.

PR #377, opened 2020-06-25
Added bee ascii

Eknir:
- Added a simple ascii Bee (source: https://www.asciiart.eu/animals/insects/bees), to be printed to the command line before anything else starts.
- Added a small welcome message

### DIFF
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -86,7 +86,7 @@ Welcome to the Swarm.... Bzzz Bzzzz Bzzzz
                  (%%%)                 
                    !                   `
 
-			fmt.Println(string(bee))
+			fmt.Println(bee)
 
 			debugAPIAddr := c.config.GetString(optionNameDebugAPIAddr)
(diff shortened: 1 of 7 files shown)


cmd/bee/cmd/start.go
```diff
@@ -71,6 +71,22 @@ func (c *command) initStartCmd() (err error) {
 			default:
 				return fmt.Errorf("unknown verbosity level %q", v)
 			}
+			bee := `
+Welcome to the Swarm.... Bzzz Bzzzz Bzzzz
```
vandot, line 75: @Eknir there is `B` with 3 z and then two times 4 z this is intentional?

vandot, line 75: @Eknir there is `B` with 3 z and then two times 4 z this is intentional?

vandot:
> @vandot @svetomir would this additional information that is printed to stdout influence how we parse log messages?

@janos we will have couple of lines that could not be parsed but no issues because of thar

janos:
@vandot cool, thanks.

PR #383, opened 2020-06-26
Close errorful but created streams

metacertain:


pkg/p2p/libp2p/libp2p.go
```diff
@@ -462,6 +462,9 @@ func (s *Service) newStreamForPeerID(ctx context.Context, peerID libp2ppeer.ID,
 	swarmStreamName := p2p.NewSwarmStreamName(protocolName, protocolVersion, streamName)
 	st, err := s.host.NewStream(ctx, peerID, protocol.ID(swarmStreamName))
 	if err != nil {
+		if st != nil {
+			_ = st.Close()
```
pradovic, line 466: I am not sure this could happen. st should be nil if the error happens. If it does happen, we could also do st.Reset(), since it is an erroureous state so it closes both sides of the stream.

PR #393, opened 2020-06-29
Add create tags in /files and /bytes API

jmozah:
- create tags for /files and /bytes API
- update the different tag counters like Address, Total, StateSeen, StateSplit, StateStored, StateSynced & Sate Sent etc at appropriate places


pkg/api/bytes.go
```diff
@@ -75,3 +83,38 @@ func (s *server) bytesGetHandler(w http.ResponseWriter, r *http.Request) {
...
+	// if tag header is not there create a new one
+	var tag *tags.Tag
+	tagUidStr := r.Header.Get(TagHeaderUid)
+	if tagUidStr == "" {
+		tagName := fmt.Sprintf("unnamed_tag_%d", time.Now().Unix())
+		var err error
+		tag, err = s.Tags.Create(tagName, 0, false)
+		if err != nil {
+			s.Logger.Debugf("bytes upload: tag creation error: %v", err)
+			s.Logger.Error("bytes upload: tag creation error")
```
pradovic, line 97:
```suggestion
			s.Logger.Error("bytes upload: tag creation")
```

pkg/api/bytes.go
```diff
@@ -75,3 +83,38 @@ func (s *server) bytesGetHandler(w http.ResponseWriter, r *http.Request) {
...
+			s.Logger.Debugf("bytes upload: parse taguid %s: %v", tagUidStr, err)
+			s.Logger.Error("bytes upload: parse taguid")
+			jsonhttp.BadRequest(w, "invalid taguid")
+			return nil
+		}
+
+		tag, err = s.Tags.Get(uint32(tagUid))
+		if err != nil {
+			s.Logger.Debugf("bytes upload: tag get error: %v", err)
+			s.Logger.Error("bytes upload: tag get error")
```
pradovic, line 114:
```suggestion
			s.Logger.Error("bytes upload: get tag")
```

pkg/debugapi/tag_test.go
```diff
@@ -202,6 +209,70 @@ func TestTags(t *testing.T) {
...
+		}
+		if finalTag.Seen != 3 {
+			t.Errorf("tag seen count mismatch. got %d want %d", finalTag.Seen, 3)
+		}
+		if finalTag.Stored != 3 {
+			t.Errorf("tag stored count mismatch. got %d want %d", finalTag.Stored, 3)
+		}
+
+		if !finalTag.Address.Equal(rootAddress) {
+			t.Errorf("Address mismatch")
```
pradovic, line 272: Would be useful to set the text to something like:
`expected %s, got %s`...

pkg/tags/tag.go
```diff
@@ -172,11 +172,15 @@ func (t *Tag) Done(s State) bool {
...
+// This is useful when you use the same tag for a file and its manifest upload
+// Also the Address is updated with the last called address, so the assumption
+// is that the manifest creation will be called last and this will have the root
+// hash of the manifest
 func (t *Tag) DoneSplit(address swarm.Address) int64 {
-	total := atomic.LoadInt64(&t.Split)
-	atomic.StoreInt64(&t.Total, total)
+	split := atomic.LoadInt64(&t.Split)
+	total := atomic.LoadInt64(&t.Total)
+	atomic.StoreInt64(&t.Total, total+split)
```
janos, line 183: These three atomic functions done separately are not atomic. There is a possibility of data race, as nothing is guarding Total in between lines 182 and 183. Total got on line 182 may be changed before new value is stored on line 183 by some other goroutine resulting an incorrect value. Mutex should be used.

janos: Your change is fixing the issue with minimal set of changes, but I find at least questionable whether both atomic and mutex should be used at the same time. If there would be any other place where Total would need to be set, both mutex lock should be held and atomic function used. There is even a strange code in UnmarshalBinary that sets Total with assignment, which should be set with an atomic function, but this is out of the scope of this pr.

jmozah: how else do you propose to fix this in this place? Other places a okay as they dont take it from splitCount.

The actual way to fix this would be to increment the total counter in a place somewhere between the API and the chunker. In the previous code it was the hasher...now we have to think it when we do pyramid chunker or something.

janos: This is a completely local issue of protecting the memory. I would not use atomic for Total field, just mutex as it is required here and makes it obvious how it is protected. I know that this involves a few more lines to be changed, but it makes clear how Total value should be handled. This is the danger of overusing the atomic package as you can create a false sense of data race free code.

jmozah: i am not sure if we should not use atomic for Total.... when other use cases comes in we might have to increase them atomically.

janos:
It appears that there are build problems https://github.com/ethersphere/bee/pull/393/checks?check_run_id=845702170#step:8:10.

jmozah:
@janos yeah there were conflicts with master and i am merging them. give few moments

nolash:
> You are adding tags to the splitter and joiner component outright. the bee-split and bee-join are both capable of dumping to and reading from filesystem, in which case tags are irrelevant. is there a way you can preserve the purity of the splitter and joiner components and coordinate tags in another manner?

for example, can it be put on the store level instead, like `TaggerStore`?

jmozah:
> for example, can it be put on the store level instead, like TaggerStore?

@nolash Tags are higher lever components which needs to know the file level or upload level structure. If you take it below to chunk level, then consolidating them will be a issue. One thing we can do is if you pass a `nil` in `NewSimpleSplitter` we can ignore incrementing tags altogether in `internal.NewSimpleSplitterJob`. This will make sure the clients which don't need tags can be somewhat pure. WDYT?

On the other hand, i will recommend that clients use tags and enhance UX.

nolash:
> > for example, can it be put on the store level instead, like TaggerStore?
> 
> @nolash Tags are higher lever components which needs to know the file level or upload level structure. If you take it below to chunk level, then consolidating them will be a issue. One thing we can do is if you pass a `nil` in `NewSimpleSplitter` we can ignore incrementing tags altogether in `internal.NewSimpleSplitterJob`. This will make sure the clients which don't need tags can be somewhat pure. WDYT?
> 
> On the other hand, i will recommend that clients use tags and enhance UX.

What does "know the file level or upload level structure" mean here?

To me it looks like the operations on tags here are only increments, who don't record anything about any splitter-specific data like level, section, batch etc. That suggests to me that probably these increments could be in a thin store wrapper instead, because the chunks are heading there anyway.

If it is possible to separate concerns here, I think we should do it.

(In the same spirit, I would say that the `toEncrypt` member also doesn't belong in this code.)

jmozah:
> What does "know the file level or upload level structure" mean here?

oh.. that means the root address and total chunks of the entire collection.
I am still not sure how to do it only in the chunk layer.. probably we could.. i will have another close look again.

> In the same spirit, I would say that the toEncrypt member also doesn't belong in this code. 


Yeah Encrypt can be pushed down.. What i would suggest we do this in another PR and not here.

nolash:
> What i would suggest we do this in another PR and not here.

yeah that's fine, but am I wrong in assuming that the encryption performs no action currently in the splitter part of the code? If so let's just remove all reference to it there now, as it's dead code anyway and if it doesn't belong no reason to keep it.

jmozah:
> but am I wrong in assuming that the encryption performs no action currently in the splitter part of the code? 

Logically, splitter and encryption are two seperate things.  But the toEncrypt flag is passed to encryption through splitter. We should find a way of passing things like pinFlag, encryptFlag to respective modules .

pkg/api/bytes.go
```diff
@@ -88,3 +97,38 @@ func (s *server) bytesGetHandler(w http.ResponseWriter, r *http.Request) {
...
+func (s *server) createTag(w http.ResponseWriter, r *http.Request) *tags.Tag {
+	// if tag header is not there create a new one
+	var tag *tags.Tag
+	tagUidStr := r.Header.Get(TagHeaderUid)
+	if tagUidStr == "" {
+		tagName := fmt.Sprintf("unnamed_tag_%d", time.Now().Unix())
+		var err error
+		tag, err = s.Tags.Create(tagName, 0, false)
+		if err != nil {
+			s.Logger.Debugf("bytes upload: tag creation error: %v", err)
```
zelig, line 110: is this common practice in bee to have debug and error lines? why not just error the first line

pkg/file/splitter/internal/job.go
```diff
@@ -47,14 +48,15 @@ type SimpleSplitterJob struct {
 	cursors    []int    // section write position, indexed per level
 	hasher     bmt.Hash // underlying hasher used for hashing the tree
 	buffer     []byte   // keeps data and hashes, indexed by cursors
-	toEncrypt  bool     // to encryrpt the chunks or not
+	tagg       *tags.Tag
```
zelig, line 51: i hope this is a typo,

pkg/tags/tag.go
```diff
@@ -172,15 +172,6 @@ func (t *Tag) Done(s State) bool {
 	return err == nil && n == total
 }
 
-// DoneSplit sets total count to SPLIT count and sets the associated swarm hash for this tag
```
zelig, line 175: why is this removed? it is important when you  upload from stream of unknown size

pkg/api/bytes.go
nolash: yes it is part of coding protocol the @janos presented starting out.

pkg/file/splitter/internal/job.go
jmozah: nope..
t - is testing
tag- is already there 
tags is already there 
tagger - should be a interface
so i chose tagg :-)

pkg/tags/tag.go
jmozah: Total is incremented inside splitter when a chunk is created.  So no need of calling a separate DoneSplit outside.

pkg/file/splitter/internal/job.go
zelig: oh. noo

pkg/tags/tag.go
zelig: ok fine but that means that progress bars will not be able to say 'we. dont know. yet'

jmozah: hmm.. yes. and i am not okay with it. I will add DoneSplit

nolash:
Had a long chat with @jmozah today about the general design of this. It seems we agree on:

- `toEncrypt` should be removed, and encryption, in the event that it should be added in bee (which of course I would say it should not), that should be a separate component either implementing an interface in common with `Splitter` and `Joiner`, or if possible, merely be a wrapper around the store.

- any other components at the same level of encryption should be pipelineable in the same fashion.

---

We also agree that it's worth looking at the possibility of keeping tags as a pluggable component aswell.

Instead of doing increments inside the splitter, we could wrap the underlying store with a "taggerstore." The taggerstore could increment when a chunk is submitted for storage, and when a chunk is stored. Pseudocode:

```
TaggerStore.Put(ch...) {
    tag.inc(havechunkstate)
    lowerstore.Put(ch...)
    tag.int(storedchunkstate)
}
```

However, contention arose on the issue of the states themselves. Specifically, it's not clear how `StateSplit` and `StateSeen` are distinctly useful. Also it was not clear what the `StateSeen` actually indicates.

My interpretation is that `StateSeen` means "i have one chunk of the total currently in memory," whereas `StateSplit` means "i have completed splitting out one more chunk from the data." But since the first thing that happens after split is put to the store, I'd argue these two are doing the same work in the current implementation.

Furthermore, I still consider the file splitter a pluggable module in bee. It will be part of the consumption build, yes, but bee can still work without it. And indeed, if I am doing splitting outside and adding chunks directly, I may still want to monitor the delivery status of the parts of the whole, in which case `StateSplit` will be a noop state.

That suggests to me that `StateSplit` should not be part of the base collection of tag states.

@jmozah raised the question which tag states are actually useful to the end user. I welcome this discussion, because after all then why shouldnt we have `StateEncrypted` `StateErasureCoded` `StateEntangled` aswell, and should we be actually hardcoding all of these in the tag structure as they are added as (pluggable) components in bee. I would say no.

Of course, if the decision is to keep all of these states, then they can still be defined as "extensions" of the tag structure, where each involved component could "getorcreate" a field in the tag and increment it as data units are processed by  it. However, then having the "tagstore" solution would make little sense, because the tag structure would have to pass through each element anyhow.

One last thing; we already have a mechanism in golang for passing along some state from component to component - that's the context. Haven't thought this through to conclusion, but would it rather make sense to embed such a tag that travels through all compoentns in the context instead?

pkg/file/splitter/internal/job.go
```diff
@@ -47,14 +48,15 @@ type SimpleSplitterJob struct {
...
+	toEncrypt  bool // to encryrpt the chunks or not
```
nolash, line 52: afaiu encryption is not currently being performed. Please let us follow our existing policy of not including something that "perhaps" will be implemented in the future.

in any case, I think embedding encryption into this component is an architecturally weak decision.

nolash:
After discussion today with @acud and @jmozah it seems the majority opinion is not spend time on cleanly separating tags' package knowledge of components. We spent a lot of time on this already, and didn't reach consensus. Since I'm in the minority, I withdraw my comment and approve, so feel free to wire it up as you please.

jmozah:
The issue raised by @nolash w,r,to this PR is
1) Dont pollute the splitter API
2) do we reuqire so many state in the tags. esp StateSplit
3) Along the discussion, we also found that StateSeen was not incremented in a proper place. 

The above fix did the following
- Splitter API does not have tag as an argument now, i pass it through the context.
- StateSeen is incremented in the proper place .i.e when a duplicate is found

I am pushing the state discussion (point 2) as out of scope of this PR. That is a larger discussion involving many more participants.

Issue #329, opened 2020-06-18
pull sync outstanding tasks


acud:
this issue enumerates the outstanding issues that arise out of the pull sync initial PR.

- [ ] falling edge kademlia channel wrapper
- [x] metrics
- [x] extract number of bins from pullstorage, pullsync protocol and puller to be injected from constructors
- [x] construct the tests in pullsync and pullstorage around the limit in a way which the limit can be adjusted through the const and that the test adjusts automatically
- [x] puller.go - handle error on next interval L313
- [x] puller.go L390 - eliminate return on top==0 (shouldnt happen, probably related to the mocks) (#810)
- [ ] subscribe pull in localstore - db shutdown might be removable altogether 
- [ ] sync only in NN
- [x] abstract syncCtx to a separate object (#810)
- [ ] get rid of mode set sync pull? just have mode set sync?

related to:
#193 #302 #144 #136 #254

Eknir:
@acud , is this in progress?

acud:
not yet. after the release.

Eknir:
@acud , is this in progress? If so, could you connect the PRs to this issue?

Eknir:
Status update?

PR #411, opened 2020-07-03
puller: extract maxpo and shallow bin peers

acud:
extract the max bins from puller to be easily injected from tests.
tried to also refactor `kademlia` to use `swarm.MaxBins` but I'm getting tests to hang for some reason.

as part of #329

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -280,13 +280,13 @@ func NewBee(o Options) (*Bee, error) {
 		return nil, fmt.Errorf("pullsync protocol: %w", err)
 	}
 
-	pullerOpts := puller.NewOptions()
-	pullerOpts.StateStore = stateStore
-	pullerOpts.Topology = topologyDriver
-	pullerOpts.PullSync = pullSync
-	pullerOpts.Logger = logger
+	puller := puller.New(puller.Options{
+		StateStore: stateStore,
+		Topology:   topologyDriver,
+		PullSync:   pullSync,
+		Logger:     logger,
+	})
 
-	puller := puller.New(pullerOpts)
 	b.pullerCloser = puller
 
 	var apiService api.Service
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -280,12 +280,13 @@ func NewBee(o Options) (*Bee, error) {
 		return nil, fmt.Errorf("pullsync protocol: %w", err)
 	}
 
-	puller := puller.New(puller.Options{
-		StateStore: stateStore,
-		Topology:   topologyDriver,
-		PullSync:   pullSync,
-		Logger:     logger,
-	})
+	pullerOpts := puller.NewOptions()
+	pullerOpts.StateStore = stateStore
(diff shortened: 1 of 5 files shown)


pkg/puller/puller.go
```diff
@@ -53,13 +66,15 @@ type Puller struct {
 }
 
 func New(o Options) *Puller {
+	bins = o.Bins
+	shallowBinPeers = o.ShallowBinPeers
```
janos, line 70: This constructor is setting global variables. How would that affect other Puller instances? There will be a data race when reading these global variables in manage loop and creating a new Puller instance.

pkg/puller/puller.go
```diff
@@ -19,19 +19,31 @@ import (
...
+	Logger          logging.Logger
+	Bins            uint8
+	ShallowBinPeers int
+}
+
+func NewOptions() Options {
+	return Options{
+		Bins:            swarm.MaxBins,
+		ShallowBinPeers: defaultShallowBinPeers,
+	}
```
janos, line 46: I am not sure if NewOptions is needed. Cold we just set Bins and ShallowBinPeers to their default values in New function if they are set to 0 or less? This would remove the need for constructor for options for another constructor.

pkg/node/node.go
```diff
@@ -280,12 +280,13 @@ func NewBee(o Options) (*Bee, error) {
...
-		StateStore: stateStore,
-		Topology:   topologyDriver,
-		PullSync:   pullSync,
-		Logger:     logger,
-	})
+	pullerOpts := puller.NewOptions()
+	pullerOpts.StateStore = stateStore
+	pullerOpts.Topology = topologyDriver
+	pullerOpts.PullSync = pullSync
+	pullerOpts.Logger = logger
```
mortelli, line 287: what is the criterion that determines that `Options.Bins` and `Options.ShallowBinPeers` should have default values in the options constructor, but not say, for example, `Options.Topology`?

mortelli: i now see @janos has made a somewhat similar comment

pkg/puller/puller.go
mortelli: why don't we put these global variables as `Puller` fields?

pkg/puller/puller_test.go
```diff
@@ -37,28 +37,24 @@ var (
 // then that adding another peer at the same po
 // does not start another syncing session
 func TestOneSync(t *testing.T) {
-	defer func(b uint8, p int) {
-		*puller.Bins = b
-		*puller.ShallowBinPeers = p
-
-	}(*puller.Bins, *puller.ShallowBinPeers)
-	*puller.Bins = 3
-	*puller.ShallowBinPeers = 1
```
mortelli, line 46: happy to see this gone

acud: ...

mortelli: just some positivity :)

PR #412, opened 2020-07-03
Kademlia: use swarm.MaxBins

acud:
minor cleanup, as a result of the previous PR that introduces `MaxBins` to `swarm` package


acud:
Sure, that's why I mentioned `Pending other merge` in the description :)

Issue #403, opened 2020-07-01
Stream closing, synchronisation & handling errors


pradovic:
Libp2p appears not to clean-up streams unless the stream is closed for both write and read: https://github.com/libp2p/go-libp2p/blob/master/NEWS.md.

We are currently relying on `stream.Close()` to close one side of the stream, but don't check if the other side is closed, always. This can result in a half-closed stream in a case of an error. The quick fix is to use `FullClose()` which will reset the stream on timeout, if other side did not close the stream.

This means that in order for stream to be cleaned up, one of the following must happen:
- Stream must be closed by both sending and receiving side. There is no way to know this unless we wait for other side to close the stream,
- If any error appears the stream should be reset, so it is closed for both writing and reading

In order to achieve this we should reset streams on errors, and this is already part of `helpers.FullClose()`. Taking the example of https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/identify/id.go, we can do the following:
- On the calling side, the one that creates the stream, we can call `FullClose()` in a separate goroutine, possibly hanging until the timeout, and resetting the stream in a case of timeout or error. This is done in a separate goroutine in order to not the block the caller, because maybe we don't actually care if the stream was reset or closed.
- On the handler side, there is not need for a separate gouroutine, as this is already a path that was initiated in the goroutine that initiates the handler

Handling errors:
This way we will reset the stream if the other side does not close the stream, but still we don't have a way to know that the error actually happened. The next step would be to introduce stream.Reset() on errors. 

So, addressing these issues should:
- Improve performance in some cases, where error cases was not handled and the timeout was induced
- Improve correctness, signal errors and synchronize between peers where needed
- Solve zombie thread problem described above

This will be addressed in multiple PRs, going protocol by protocol. The first batch of PRs should use `FullClose()` & `Close()` combinations more properly, and the second batch should introduce usage of `stream.Reset()` if needed.

PRs:
https://github.com/ethersphere/bee/pull/401 - merged
https://github.com/ethersphere/bee/pull/405 - merged
https://github.com/ethersphere/bee/pull/407 - merged
https://github.com/ethersphere/bee/pull/419 - merged
https://github.com/ethersphere/bee/pull/420 - merged
https://github.com/ethersphere/bee/pull/421 - merged
https://github.com/ethersphere/bee/pull/422 - merged
https://github.com/ethersphere/bee/pull/428 - merged
https://github.com/ethersphere/bee/pull/449 - merged

[...]

PR #420, opened 2020-07-07
Stream reset - hive

pradovic:

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1,4 +1,5 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+cloud.google.com/go v0.31.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
 cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
 cloud.google.com/go v0.37.0/go.mod h1:TS1dMSSfndXH133OKGwekG838Om/cQT0BUHV3HcBgoo=
 cloud.google.com/go v0.38.0/go.mod h1:990N+gfupTy94rShfmMCWGDn0LpTmnzTp2qbd1dvSRU=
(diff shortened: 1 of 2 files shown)


pkg/hive/hive.go
```diff
@@ -88,7 +88,7 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
 	if err != nil {
 		return fmt.Errorf("new stream: %w", err)
 	}
-	defer stream.Close()
+	defer stream.FullClose()
```
acud, line 91: this will block and cause other components to block until this is done. i am not sure if `defer stream.FullClose()` or `defer func(){ go stream.FullClose() }()`

go.sum
```diff
@@ -1,4 +1,5 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+cloud.google.com/go v0.31.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
acud, line 2: do we need to commit these changes?

pkg/hive/hive.go
pradovic: Yes, but it was already like this. The only difference that now on error the stream is reset, so it will not block as it will already be reset. If you look at the end of the function, there use to be FullClose call if there was no error, so the functionality should be the same, but more correct now cuz the stream will be reset on error.

go.sum
pradovic: I ma not sure, I noticed that this happened on @metacertain computer as well, so did anybody change go mod and did not commit the sum? should I remove it from PR?

pkg/hive/hive.go
pradovic: Actaully, I would change it to return stream.FullClose as it was

pradovic: Oooh, sorry, I remember now. since we are sending in batches, I wanted to get the response that the stream is closed before opening a new one so bathes have a bit more sense.

go.sum
acud: yep please remove it as it is not necessary for this changeset

pradovic: Ok, I have opened a new PR for this: https://github.com/ethersphere/bee/pull/423. Or how should we handle it?

PR #422, opened 2020-07-07
Pushsync - stream reset

pradovic:

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -80,14 +80,19 @@ func (s *PushSync) Protocol() p2p.ProtocolSpec {
 
 // handler handles chunk delivery from other node and forwards to its destination node.
 // If the current node is the destination, it stores in the local store and sends a receipt.
-func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) error {
+func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) (err error) {
 	w, r := protobuf.NewWriterAndReader(stream)
-	defer stream.FullClose()
+	defer func() {
+		if err != nil {
+			_ = stream.Reset()
+		} else {
+			_ = stream.FullClose()
+		}
+	}()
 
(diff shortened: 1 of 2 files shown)


pkg/pushsync/pushsync.go
```diff
@@ -82,11 +82,12 @@ func (s *PushSync) Protocol() p2p.ProtocolSpec {
 // If the current node is the destination, it stores in the local store and sends a receipt.
 func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) error {
```
acud, line 83: consider that if you have a named return here `err error`, and then in the code below return an error, then you don't need to scatter all these stream reset calls all over in every branch of the code but just solve it in one defer:
```go
defer func() {
if err != nil {
  _ = stream.Reset()
}
stream.FullClose()
}
```

I find this much more readable and fool proof for future changes

pradovic: Sounds good to me, I thought about this as well and then decided to skip it, but I will change it to this then

acud: thanks :pray:

pradovic: Hm, actually now I saw a reason against. There are 2 streams created here, both in the different parts of the function with the same error handling. So I would avoid this as it leads to a potential mistakes. Is that ok for you?

acud: it does not matter that there are two streams.
if you create the first defer statement, then the second defer statement, there cannot be a scenario that the error would be nil for the first one but not for the second one. you can in fact use exactly the same error check you do in the first defer statement. if the first one errors then the second one won't execute (since you don't even get to the second defer statement), and if the second errors then both should probably be in an error state. that's at least my opinion

pradovic: Makes sense. Done :)

pradovic: Hmm, I found something interesting here, but let's talk tomorrow. I need to check with you one use-case

PR #424, opened 2020-07-08
pullsync, puller: add metrics

acud:
This PR adds some prometheus metrics to the `puller` and `pullsync` packages.
These are very minimal metrics, and they should reflect some basic information as to the operation of these components.

One important note to the error counters is that due to the fact that context is being used to cancel the syncing jobs on depth change/peer disconnect, it in fact means that not all errors that are being returned from certain function calls should be accounted as actual protocol errors (or errors that we as developers would like to be notified about). 
This is due to the fact that once a sync job's underlying context is cancelled, the current function call will return an error `context.ErrContextCancelled`, which is not really an error in this case. This is why only certain branches of the code actually increment the error counters. 

part of #329

### DIFF
--- a/pkg/puller/metrics.go
+++ b/pkg/puller/metrics.go
@@ -1,0 +1,59 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package puller
+
+import (
+	m "github.com/ethersphere/bee/pkg/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type metrics struct {
+	HistWorkerIterCounter prometheus.Counter // counts the number of historical syncing iterations
+	HistWorkerDoneCounter prometheus.Counter // count number of finished historical syncing jobs
+	HistWorkerErrCounter  prometheus.Counter // count number of errors
+	LiveWorkerIterCounter prometheus.Counter // counts the number of live syncing iterations
+	LiveWorkerErrCounter  prometheus.Counter // count number of errors
+}
+
+func newMetrics() metrics {
+	subsystem := "puller"
+
+	return metrics{
+		HistWorkerIterCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "hist_worker_iterations",
+			Help:      "Total history worker iterations.",
+		}),
+		HistWorkerDoneCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "hist_worker_done",
+			Help:      "Total history worker jobs done.",
+		}),
+		HistWorkerErrCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "hist_worker_errors",
+			Help:      "Total history worker errors.",
+		}),
+		LiveWorkerIterCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "live_worker_iterations",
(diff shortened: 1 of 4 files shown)


pkg/pullsync/pullsync.go
```diff
@@ -196,7 +203,8 @@ func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8
 		}
 
 		delete(wantChunks, addr.String())
-
+		s.metrics.DbOpsCounter.Inc()
+		s.metrics.DeliveryCounter.Inc()
```
mortelli, line 207: shouldn't we increase this only if `s.storage.Put` is successfull?

pkg/puller/metrics.go
```diff
@@ -0,0 +1,59 @@
...
+package puller
+
+import (
+	m "github.com/ethersphere/bee/pkg/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type metrics struct {
+	HistWorkerIterCounter prometheus.Counter // counts the number of historical syncing iterations
+	HistWorkerDoneCounter prometheus.Counter // count number of finished historical syncing jobs
```
mortelli, line 14: afaik this one is never used. i'd suggest removing it or increasing it when the job is completed without errors. 

if we go for the 2nd option, i'd suggest adding a `LiveWorkerDoneCounter` as well.

pkg/pullsync/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+func newMetrics() metrics {
+	subsystem := "pullsync"
+
+	return metrics{
+		OfferCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "chunks_offered",
+			Help:      "Total chunks offered.",
+		}),
```
mortelli, line 28: if i understand this correctly (and please correct me if i don't) this is about the chunks that are offered to the running node, no?

if so, can we make this more explicit in the help text?

i initially thought this was about chunks offered to _other_ nodes

pkg/pullsync/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+			Subsystem: subsystem,
+			Name:      "chunks_wanted",
+			Help:      "Total chunks wanted.",
+		}),
+		DeliveryCounter: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "chunks_delivered",
+			Help:      "Total chunks delivered.",
+		}),
```
mortelli, line 40: same here, this means delivered _to us_, no?

pkg/puller/metrics.go
acud: @mortelli `live` syncing should never be done, hence the asymmetry. nice catch that it wasn't used though. added it now :+1:

pkg/pullsync/metrics.go
acud: it coincides with the terms used in the pull sync protocol

pkg/pullsync/metrics.go
acud: yeah, and that's where it is incremented. where's the question?

pkg/pullsync/pullsync.go
acud: we're counting deliveries (which is a protocol receiving a chunk), not db operation errors

pkg/pullsync/metrics.go
mortelli: no question, i would just make it more explicit in the `Help` text (_delivered to whom?_)

not a blocker though, and i'm happy to let it go if it's just me :)

pkg/pullsync/metrics.go
mortelli: fair enough

acud:
>   * `pullsync.go:311`

is the only one i did not add an increment for, since it does not necessarily imply an immediate Db operation right after the call. It might just wait for a chunk for a long time, hence to increment the counter will not be semantically correct.

PR #427, opened 2020-07-08
Check for validity before concluding retrieving chunk

metacertain:
In order to account only for valid chunks that are returned by others (i.e. not paying for junk) we need to check for chunk validity before we would change the account balance of the serving peer. 
As we want to do the accounting in for example retrieval, this required us to do the check before the end of retrieval, while the preexisting code only checked after retrieve returned. 

This also brought our attention to another possible issue, that if we were to recieve garbage chunk, we would not retrieve from another peer, as the verification takes place after the peer queries, so then we would probably fail, even if the valid chunk would be available from the next tried peer.

So we created the chunk in retrieve, and passed a validation function as an argument to retrieve that can do the validation and already return the chunk. This might cause performance issues because of the frequent passing of validation functions.

### DIFF
--- a/pkg/content/mock/validator.go
+++ b/pkg/content/mock/validator.go
@@ -8,14 +8,14 @@ import (
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
-var _ swarm.ChunkValidator = (*Validator)(nil)
+var _ swarm.Validator = (*Validator)(nil)
 
 type Validator struct {
 	rv bool
 }
 
 // NewValidator constructs a new Validator
-func NewValidator(rv bool) swarm.ChunkValidator {
+func NewValidator(rv bool) swarm.Validator {
 	return &Validator{rv: rv}
 }
 
--- a/pkg/content/mock/validator.go
+++ b/pkg/content/mock/validator.go
@@ -10,7 +10,6 @@ import (
 
 var _ swarm.ChunkValidator = (*Validator)(nil)
 
-//
 type Validator struct {
 	rv bool
 }
--- a/pkg/content/mock/validator.go
+++ b/pkg/content/mock/validator.go
@@ -10,17 +10,17 @@ import (
 
 var _ swarm.ChunkValidator = (*Validator)(nil)
 
-// ContentAddressValidator validates that the address of a given chunk
-// is the content address of its contents.
+//
 type Validator struct {
+	rv bool
 }
 
-// NewContentAddressValidator constructs a new ContentAddressValidator
-func NewValidator() swarm.ChunkValidator {
-	return &Validator{}
+// NewValidator constructs a new Validator
+func NewValidator(rv bool) swarm.ChunkValidator {
+	return &Validator{rv: rv}
 }
 
-// Validate performs the validation check.
+// Validate returns rv from mock struct
 func (v *Validator) Validate(ch swarm.Chunk) (valid bool) {
-	return true
+	return v.rv
 }
--- a/pkg/content/mock/validator.go
+++ b/pkg/content/mock/validator.go
@@ -1,0 +1,26 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mock
+
+import (
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+var _ swarm.ChunkValidator = (*Validator)(nil)
+
+// ContentAddressValidator validates that the address of a given chunk
+// is the content address of its contents.
+type Validator struct {
+}
+
+// NewContentAddressValidator constructs a new ContentAddressValidator
+func NewValidator() swarm.ChunkValidator {
+	return &Validator{}
+}
+
+// Validate performs the validation check.
+func (v *Validator) Validate(ch swarm.Chunk) (valid bool) {
+	return true
+}
--- a/pkg/netstore/netstore.go
+++ b/pkg/netstore/netstore.go
@@ -40,11 +40,6 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 				return nil, fmt.Errorf("netstore retrieve chunk: %w", err)
 			}
 
-			if !s.valid(ch) {
-				// invalid retrieve handling
-				return nil, storage.ErrInvalidChunk
-			}
-
 			_, err = s.Storer.Put(ctx, storage.ModePutRequest, ch)
 			if err != nil {
 				return nil, fmt.Errorf("netstore retrieve put: %w", err)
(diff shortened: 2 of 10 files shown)


Eknir:
@metacertain , what is the status of this PR? Could you be so kind to provide a description to this PR and/or link it to an existing issue? Right now, I cannot see which problem you intend to solve with this code. Thanks

metacertain:
@Eknir Updated description to elaborate

pkg/retrieval/retrieval.go
```diff
@@ -115,9 +120,11 @@ func (s *Service) retrieveChunk(ctx context.Context, addr swarm.Address, skipPee
 	defer cancel()
 
 	peer, err = s.closestPeer(addr, skipPeers)
+
```
zelig, line 123: why introduce newlines here?

pkg/retrieval/retrieval_test.go
```diff
@@ -132,3 +132,5 @@ func (s mockPeerSuggester) EachPeer(topology.EachPeerFunc) error {
 func (s mockPeerSuggester) EachPeerRev(f topology.EachPeerFunc) error {
 	return s.eachPeerRevFunc(f)
 }
+
+func FValidate(_ swarm.Chunk) bool { return true }
```
zelig, line 136: why not use the actual validators here if you do actual chunks?

metacertain: I think we used fake validation like this in netstore test as well,, which seems reasonable as we have a separate unit test for validation

pkg/retrieval/retrieval.go
```diff
@@ -93,14 +93,19 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address) (data [
 				continue
 			}
 			s.logger.Tracef("retrieval: got chunk %s from peer %s", addr, peer)
-			return data, nil
+			//
+			chunk := swarm.NewChunk(addr, data)
+			if valid(chunk) {
+				return chunk, nil
+			}
```
Eknir, line 100: Code style:
I usually find it better to return the error-conditions inside the if-block, and do the happy-path outside if blocks.

pkg/netstore/netstore.go
```diff
@@ -33,13 +33,13 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 	if err != nil {
 		if errors.Is(err, storage.ErrNotFound) {
 			// request from network
-			data, err := s.retrieval.RetrieveChunk(ctx, addr)
+			ch, err := s.retrieval.RetrieveChunk(ctx, addr, s.valid)
```
ralph-pichler, line 36: I don't think this is a good idea after all. I know I came up with this line but this could introduce problems (e.g. it's easy to get deadlocks if there are locks added to net store at some points). Also we are verifying the chunk twice which is unnecessary work.

I do think it would be more clean to return the peer we received the chunk from in `RetrieveChunk` and then find a way to disconnect from that peer after the `!s.valid(ch)` check further below. This does of course mean that we first `Credit` the peer for the invalid chunk which might trigger settlement, but if that happens its only because we already received a lot of valid chunks before, so the loss is not that big.

metacertain: I have to disagree partly. 
The problem is, if we don't verify chunk in retrieve, we do not query more peers if we get an invalid chunk.
So the current change makes retries possible when an invalid chunk is returned.

Removed redundant validity check, passing validator functions should probably be benchmarked.

acud: this all depends on how we'd like to do accounting. if accounting is implicit and is part of the protocol abstraction like in the old swarm codebase then this has to be done as part of the retrieval protocol validation. i'm not sure if passing the actual validator function is the best solution here. it could be that we would be better off with some reusable validator component that can be passed around both to the protocols.

that being said, if accounting for a chunk would be done declaratively outside the protocol then this change is not needed and bumping the accounting counters can be done after the validator check originally on `L41` on the left. not sure what you have in mind tho. my personal opinion is that it is nicer that we have the validator check only in netstore, and not have these code duplications across the protocols. although if this is very much necessary to do in this manner it is certainly doable, and we might find some way to make it nice and clean.

acud: having thought this through, i think that it is no longer needed to have this validator on netstore at all of we go with this approach of inspecting whether the chunk is invalid of not on the protocol level. passing it along as a function reference is clunky and adds unnecessary indirection

pkg/content/mock/validator.go
```diff
@@ -10,17 +10,17 @@ import (
 
 var _ swarm.ChunkValidator = (*Validator)(nil)
 
-// ContentAddressValidator validates that the address of a given chunk
-// is the content address of its contents.
+//
```
acud, line 13: empty comment :)

pkg/netstore/netstore_test.go
```diff
@@ -111,9 +113,10 @@ type retrievalMock struct {
...
 }
 
-func (r *retrievalMock) RetrieveChunk(ctx context.Context, addr swarm.Address) (data []byte, err error) {
+func (r *retrievalMock) RetrieveChunk(ctx context.Context, addr swarm.Address) (chunk swarm.Chunk, err error) {
 	r.called = true
 	atomic.AddInt32(&r.callCount, 1)
 	r.addr = addr
-	return chunkData, nil
+	chunk = swarm.NewChunk(addr, chunkData)
+	return chunk, nil
```
mortelli, line 121: you can replace `119:121` with:
```golang
return swarm.NewChunk(r.addr, chunkData), nil
```

pkg/swarm/swarm.go
```diff
@@ -185,3 +185,24 @@ func (c *chunk) WithType(t Type) Chunk {
 type ChunkValidator interface {
 	Validate(ch Chunk) (valid bool)
 }
+
+type chunkValidators struct {
+	set []ChunkValidator
+	ChunkValidator
+}
```
mortelli, line 192: why do we need both the `set` field plus the embedded `ChunkValidator` interface?

pkg/swarm/swarm.go
```diff
@@ -185,3 +185,24 @@ func (c *chunk) WithType(t Type) Chunk {
...
+
+func NewChunkValidators(v ...ChunkValidator) ChunkValidator {
```
mortelli, line 194: this returns a `ChunkValidator` but the function is called `NewChunkValidators` (plural).

we should be consistent here, take a look at the calls:
```
validator := swarm.NewChunkValidators(validatormock.NewValidator(true))
chunkvalidators := swarm.NewChunkValidators(soc.NewValidator(), content.NewValidator())
mockValidator := swarm.NewChunkValidators(mock.NewValidator(true))
```

i think it's OK to have an entity called _Validator_ even if it uses multiple private validators internally, but the vars which take the result of this func (plus the func name) should have consistent naming in this manner.

pkg/swarm/swarm.go
acud: because chunkValidators implements this interface. we might not need to include it explicitly in the struct definition. this was added just for sake of clarity. `set` has the actual validators

PR #428, opened 2020-07-09
Pullsync - reset stream on errors

metacertain:
https://github.com/ethersphere/bee/pull/403

### DIFF
--- a/pkg/pullsync/pullsync.go
+++ b/pkg/pullsync/pullsync.go
@@ -116,7 +116,7 @@ func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8
 		if err != nil {
 			_ = stream.Reset()
 		} else {
-			_ = stream.FullClose()
+			go stream.FullClose()
 		}
 	}()
 
@@ -362,7 +362,7 @@ func (s *Syncer) GetCursors(ctx context.Context, peer swarm.Address) (retr []uin
 		if err != nil {
 			_ = stream.Reset()
 		} else {
-			_ = stream.FullClose()
+			go stream.FullClose()
 		}
 	}()
 
@@ -401,7 +401,6 @@ func (s *Syncer) cursorHandler(ctx context.Context, p p2p.Peer, stream p2p.Strea
 	s.metrics.DbOpsCounter.Inc()
 	ints, err := s.storage.Cursors(ctx)
 	if err != nil {
-		_ = stream.Reset()
 		return err
 	}
 	ack.Cursors = ints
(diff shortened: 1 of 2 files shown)


pkg/pullsync/pullsync.go
```diff
@@ -112,6 +112,13 @@ func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8
 	if err != nil {
 		return 0, 0, fmt.Errorf("new stream: %w", err)
 	}
+	defer func() {
+		if err != nil {
+			_ = stream.Reset()
+		} else {
+			_ = stream.FullClose()
```
pradovic, line 119:
```suggestion
			go stream.FullClose()
```
It seems to me that we did not wait for conformation from the other side here, so i think we can do this in a separate go routine in order not to block the caller.

pkg/pullsync/pullsync.go
```diff
@@ -342,12 +353,18 @@ func (s *Syncer) processWant(ctx context.Context, o *pb.Offer, w *pb.Want) ([]sw
...
 	stream, err := s.streamer.NewStream(ctx, peer, nil, protocolName, protocolVersion, cursorStreamName)
 	if err != nil {
 		return nil, fmt.Errorf("new stream: %w", err)
 	}
-	defer stream.Close()
+	defer func() {
+		if err != nil {
+			_ = stream.Reset()
+		} else {
+			_ = stream.FullClose()
```
pradovic, line 365: Same as the above. The handlers don't need the separate go routines as they are not blocking anyone, but the actual side that opens the stream benefits from the separate go routine to avoid blocking the caller

pkg/pullsync/pullsync.go
```diff
@@ -376,7 +401,7 @@ func (s *Syncer) cursorHandler(ctx context.Context, p p2p.Peer, stream p2p.Strea
 	s.metrics.DbOpsCounter.Inc()
 	ints, err := s.storage.Cursors(ctx)
 	if err != nil {
-		_ = stream.FullClose()
+		_ = stream.Reset()
```
pradovic, line 404: No need for reset here as it will be done by deferred function.

pkg/pullsync/pullsync.go
```diff
@@ -387,14 +412,20 @@ func (s *Syncer) cursorHandler(ctx context.Context, p p2p.Peer, stream p2p.Strea
...
 		return fmt.Errorf("new stream: %w", err)
 	}
 
 	w := protobuf.NewWriter(stream)
-	defer stream.Close()
+	defer func() {
+		if err != nil {
+			_ = stream.Reset()
+		} else {
+			_ = stream.FullClose()
```
pradovic, line 426: Same as the above for the separate go routine.

pkg/pullsync/pullsync.go
pradovic: I think that FullClose was use inappropriate here before, no needf to wait for other side to reset here.

Eknir:
@metacertain , what is the status of this PR? @pradovic

pradovic:
@Eknir, @metacertain We should merge it :)

PR #429, opened 2020-07-09
Trojan chunk support

jmozah:
Implements the trojan structures as described in the Book of Swarm.

The purpose of deploying trojan messages is to message another node in obscurity; hence they are embedded into trojan chunks, which are indistinguishable from regular data chunks.

This is pulled from the old swarm code base and refitted for the bee

### Task-related for improvements
#445 Add unsealing and sealing
#444 optimize trojan chunk mining
#443 make trojan chunks work as a bit vector

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -56,14 +56,14 @@ func NewEthereumAddress(p ecdsa.PublicKey) ([]byte, error) {
 		return nil, errors.New("invalid public key")
 	}
 	pubBytes := elliptic.Marshal(btcec.S256(), p.X, p.Y)
-	pubHash, err := legacyKeccak256(pubBytes[1:])
+	pubHash, err := LegacyKeccak256(pubBytes[1:])
 	if err != nil {
 		return nil, err
 	}
 	return pubHash[12:], err
 }
 
-func legacyKeccak256(data []byte) ([]byte, error) {
+func LegacyKeccak256(data []byte) ([]byte, error) {
 	var err error
 	hasher := sha3.NewLegacyKeccak256()
 	_, err = hasher.Write(data)
--- a/pkg/file/file.go
+++ b/pkg/file/file.go
@@ -15,7 +15,7 @@ import (
 )
 
 var (
-	ChunkWithLengthSize = swarm.ChunkSize + 8
+	ChunkWithSpanSize = swarm.ChunkSize + 8
 )
 
 // Joiner returns file data referenced by the given Swarm Address to the given io.Reader.
--- a/pkg/file/splitter/internal/job.go
+++ b/pkg/file/splitter/internal/job.go
@@ -21,7 +21,7 @@ import (
 )
 
 // maximum amount of file tree levels this file hasher component can handle
-// (128 ^ (9 - 1)) * 4096 = 295147905179352825856 bytes
+// (128 ^ (9 - 1)) * 4096 = 295147905179352825856 bytesgit fetch origin
 const levelBufferLimit = 9
 
 // hashFunc is a hasher factory used by the bmt hasher
@@ -68,7 +68,7 @@ func NewSimpleSplitterJob(ctx context.Context, putter storage.Putter, spanLength
 		sumCounts:  make([]int, levelBufferLimit),
 		cursors:    make([]int, levelBufferLimit),
 		hasher:     bmtlegacy.New(p),
-		buffer:     make([]byte, file.ChunkWithLengthSize*levelBufferLimit*2), // double size as temp workaround for weak calculation of needed buffer space
+		buffer:     make([]byte, file.ChunkWithSpanSize*levelBufferLimit*2), // double size as temp workaround for weak calculation of needed buffer space
 		toEncrypt:  toEncrypt,
 		refSize:    refSize,
 	}
--- a/pkg/file/splitter/internal/job.go
+++ b/pkg/file/splitter/internal/job.go
@@ -60,7 +60,7 @@ func NewSimpleSplitterJob(ctx context.Context, putter storage.Putter, spanLength
 		sumCounts:  make([]int, levelBufferLimit),
 		cursors:    make([]int, levelBufferLimit),
 		hasher:     bmtlegacy.New(p),
-		buffer:     make([]byte, file.ChunkWithLengthSize*levelBufferLimit*2), // double size as temp workaround for weak calculation of needed buffer space
+		buffer:     make([]byte, file.ChunkWithSpanSize*levelBufferLimit*2), // double size as temp workaround for weak calculation of needed buffer space
 	}
 }
 
--- a/pkg/swarm/swarm.go
+++ b/pkg/swarm/swarm.go
@@ -15,16 +15,13 @@ import (
 )
 
 const (
-	SectionSize            = 32
-	Branches               = 128
-	ChunkSize              = SectionSize * Branches
-	HashSize               = 32
-	MaxPO            uint8 = 15
-	MaxBins                = MaxPO + 1
-	SpanSize               = 8
-	TrojanNonceSize        = 32
-	TrojanLengthSize       = 2
-	TrojanTopicSize        = 32
+	SectionSize       = 32
+	Branches          = 128
+	ChunkSize         = SectionSize * Branches
+	HashSize          = 32
+	MaxPO       uint8 = 15
+	MaxBins           = MaxPO + 1
+	SpanSize          = 8
 )
 
 var (
--- a/pkg/swarm/swarm.go
+++ b/pkg/swarm/swarm.go
@@ -10,15 +10,24 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"golang.org/x/crypto/sha3"
 )
 
 const (
-	SectionSize       = 32
-	Branches          = 128
-	ChunkSize         = SectionSize * Branches
-	HashSize          = 32
-	MaxPO       uint8 = 15
-	MaxBins           = MaxPO + 1
+	SectionSize            = 32
+	Branches               = 128
+	ChunkSize              = SectionSize * Branches
+	HashSize               = 32
+	MaxPO            uint8 = 15
+	MaxBins                = MaxPO + 1
+	SpanSize               = 8
+	TrojanNonceSize        = 32
+	TrojanLengthSize       = 2
+	TrojanTopicSize        = 32
+)
+
+var (
+	NewHasher = sha3.NewLegacyKeccak256
 )
 
 // Address represents an address in Swarm metric space of
@@ -119,7 +128,7 @@ type chunk struct {
 	sdata      []byte
 	pinCounter uint64
 	tagID      uint32
-	chunkType  Type
+	typ        Type
 }
 
 func NewChunk(addr Address, data []byte) Chunk {
@@ -164,11 +173,11 @@ func (c *chunk) Equal(cp Chunk) bool {
 }
 
 func (c *chunk) Type() Type {
-	return c.chunkType
+	return c.typ
 }
 
 func (c *chunk) WithType(t Type) Chunk {
-	c.chunkType = t
+	c.typ = t
 	return c
 }
 
--- a/pkg/swarm/swarm.go
+++ b/pkg/swarm/swarm.go
@@ -10,6 +10,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+
 	"golang.org/x/crypto/sha3"
 )
 
--- a/pkg/swarm/swarm.go
+++ b/pkg/swarm/swarm.go
@@ -94,6 +94,14 @@ func (a Address) MarshalJSON() ([]byte, error) {
 // ZeroAddress is the address that has no value.
 var ZeroAddress = NewAddress(nil)
 
+// Type describes a kind of chunk, whether it is content-addressed or other
+type Type int
+
+const (
+	Unknown Type = iota
+	ContentAddressed
+)
+
 type Chunk interface {
 	Address() Address
 	Data() []byte
@@ -102,6 +110,8 @@ type Chunk interface {
 	TagID() uint32
 	WithTagID(t uint32) Chunk
 	Equal(Chunk) bool
+	Type() Type
+	WithType(t Type) Chunk
 }
 
 type chunk struct {
@@ -109,6 +119,7 @@ type chunk struct {
 	sdata      []byte
 	pinCounter uint64
 	tagID      uint32
+	chunkType  Type
 }
 
 func NewChunk(addr Address, data []byte) Chunk {
@@ -150,6 +161,15 @@ func (c *chunk) String() string {
 
 func (c *chunk) Equal(cp Chunk) bool {
 	return c.Address().Equal(cp.Address()) && bytes.Equal(c.Data(), cp.Data())
+}
+
+func (c *chunk) Type() Type {
+	return c.chunkType
+}
+
+func (c *chunk) WithType(t Type) Chunk {
+	c.chunkType = t
+	return c
 }
 
 type ChunkValidator interface {
--- a/pkg/trojan/export_test.go
+++ b/pkg/trojan/export_test.go
@@ -1,8 +1,8 @@
 package trojan
 
 var (
-	Contains  = contains
-	HashBytes = hashBytes
-	PadBytes  = padBytesLeft
+	Contains    = contains
+	HashBytes   = hashBytes
+	PadBytes    = padBytesLeft
 	IsPotential = isPotential
 )
--- a/pkg/trojan/export_test.go
+++ b/pkg/trojan/export_test.go
@@ -4,4 +4,5 @@ var (
 	Contains  = contains
 	HashBytes = hashBytes
 	PadBytes  = padBytesLeft
+	IsPotential = isPotential
 )
--- a/pkg/trojan/export_test.go
+++ b/pkg/trojan/export_test.go
@@ -1,0 +1,7 @@
+package trojan
+
+var (
+	Contains  = contains
+	HashBytes = hashBytes
+	PadBytes  = padBytesLeft
+)
--- a/pkg/trojan/message.go
+++ b/pkg/trojan/message.go
@@ -8,8 +8,6 @@ import (
 	"bytes"
 	"crypto/rand"
 	"encoding/binary"
-	"errors"
-	"fmt"
 	"math/big"
 	"time"
 
@@ -20,7 +18,8 @@ import (
 // Topic is an alias for a 32 byte fixed-size array which contains an encoding of a message topic
 type Topic [32]byte
 
-// Target is an alias for an address which can be mined to construct a trojan message
+// Target is an alias for an address which can be mined to construct a trojan message.
+// Target is like partial address which helps to send message to a particular PO.
 type Target []byte
 
 // Targets is an alias for a collection of targets
@@ -43,23 +42,6 @@ const (
 	LengthSize     = 2
 	TopicSize      = 32
 	MinerTimeout   = 2 // seconds
-)
-
-var (
-	// ErrPayloadTooBig is returned when a given payload for a Message type is longer than the maximum amount allowed
-	ErrPayloadTooBig = fmt.Errorf("message payload size cannot be greater than %d bytes", MaxPayloadSize)
-
-	// ErrEmptyTargets is returned when the given target list for a trojan chunk is empty
-	ErrEmptyTargets = errors.New("target list cannot be empty")
-
-	// ErrVarLenTargets is returned when the given target list for a trojan chunk has addresses of different lengths
-	ErrVarLenTargets = errors.New("target list cannot have targets of different length")
-
-	// ErrUnMarshallingTrojanMessage is returned when a trojan message could not be de-serialized
-	ErrUnMarshallingTrojanMessage = errors.New("trojan message unmarshall error")
-
-	// ErrMinerTimeout is returned when mining a new nonce takes more time than swarm.TrojanMinerTimeout seconds
-	ErrMinerTimeout = errors.New("miner timeout error")
 )
 
 // NewTopic creates a new Topic variable with the given input string
--- a/pkg/trojan/message.go
+++ b/pkg/trojan/message.go
@@ -42,7 +42,7 @@ const (
 
 var (
 	// ErrPayloadTooBig is returned when a given payload for a Message type is longer than the maximum amount allowed
-	ErrPayloadTooBig = errors.New(fmt.Sprintf("message payload size cannot be greater than %d bytes", MaxPayloadSize))
+	ErrPayloadTooBig = fmt.Errorf("message payload size cannot be greater than %d bytes", MaxPayloadSize)
 
 	// ErrEmptyTargets is returned when the given target list for a trojan chunk is empty
 	ErrEmptyTargets = errors.New("target list cannot be empty")
--- a/pkg/trojan/message.go
+++ b/pkg/trojan/message.go
@@ -25,11 +25,10 @@ import (
 	"hash"
 	"math/big"
 
-	"github.com/ethersphere/swarm/bmt"
-	"golang.org/x/crypto/sha3"
-
 	"github.com/ethersphere/bee/pkg/crypto"
 	"github.com/ethersphere/bee/pkg/swarm"
+	bmtlegacy "github.com/ethersphere/bmt/legacy"
+	"golang.org/x/crypto/sha3"
 )
 
 // Topic is an alias for a 32 byte fixed-size array which contains an encoding of a message topic
@@ -75,7 +74,7 @@ func NewTopic(topic string) Topic {
 	if err != nil {
 		return tpc
 	}
-	copy(tpc[:], t[:])
+	copy(tpc[:], t)
 	return tpc
 }
 
@@ -83,7 +82,7 @@ func hashFunc() hash.Hash {
 	return sha3.NewLegacyKeccak256()
 }
 
-var hasher *bmt.Hasher
+var hasher *bmtlegacy.Hasher
 
 // NewMessage creates a new Message variable with the given topic and payload
 // it finds a length and nonce for the message according to the given input and maximum payload size
@@ -106,13 +105,13 @@ func NewMessage(topic Topic, payload []byte) (Message, error) {
 
 	// create new Message var and set fields
 	m := new(Message)
-	copy(m.length[:], lengthBuf[:])
+	copy(m.length[:], lengthBuf)
 	m.Topic = topic
 	m.Payload = payload
 	m.padding = padding
 
-	hashPool := bmt.NewTreePool(hashFunc, swarm.Branches, bmt.PoolSize)
-	hasher = bmt.New(hashPool)
+	hashPool := bmtlegacy.NewTreePool(hashFunc, swarm.Branches, bmtlegacy.PoolSize)
+	hasher = bmtlegacy.New(hashPool)
 
 	return *m, nil
 }
@@ -228,7 +227,11 @@ func hashBytes(s []byte) ([]byte, error) {
 		return nil, ErrInvalidHasher
 	}
 	hasher.Reset()
-	hasher.SetSpanBytes(s[:8])
+	span := binary.LittleEndian.Uint64(s[:8])
+	err := hasher.SetSpan(int64(span))
+	if err != nil {
+		return nil, err
+	}
 	if _, err := hasher.Write(s[8:]); err != nil {
 		return nil, err
 	}
--- a/pkg/trojan/message.go
+++ b/pkg/trojan/message.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"time"
 
 	"github.com/ethersphere/bee/pkg/swarm"
 	bmtlegacy "github.com/ethersphere/bmt/legacy"
@@ -38,6 +39,10 @@ const (
 	// MaxPayloadSize + Topic + Length + Nonce = Default ChunkSize
 	//    (4030)      +  (32) +   (2)  +  (32) = 4096 Bytes
 	MaxPayloadSize = 4030
+	NonceSize      = 32
+	LengthSize     = 2
+	TopicSize      = 32
+	MinerTimeout   = 2 // seconds
 )
 
 var (
@@ -50,7 +55,11 @@ var (
 	// ErrVarLenTargets is returned when the given target list for a trojan chunk has addresses of different lengths
 	ErrVarLenTargets = errors.New("target list cannot have targets of different length")
 
+	// ErrUnMarshallingTrojanMessage is returned when a trojan message could not be de-serialized
 	ErrUnMarshallingTrojanMessage = errors.New("trojan message unmarshall error")
+
+	// ErrMinerTimeout is returned when mining a new nonce takes more time than swarm.TrojanMinerTimeout seconds
+	ErrMinerTimeout = errors.New("miner timeout error")
 )
 
 // NewTopic creates a new Topic variable with the given input string
@@ -121,7 +130,7 @@ func Unwrap(c swarm.Chunk) (*Message, error) {
 }
 
 // IsPotential returns true if the given chunk is a potential trojan
-func IsPotential(c swarm.Chunk) bool {
+func isPotential(c swarm.Chunk) bool {
 	// chunk must be content-addressed to be trojan
 	if c.Type() != swarm.ContentAddressed {
 		return false
@@ -129,7 +138,7 @@ func IsPotential(c swarm.Chunk) bool {
 
 	data := c.Data()
 	// check for minimum chunk data length
-	trojanChunkMinDataLen := swarm.SpanSize + swarm.TrojanNonceSize + swarm.TrojanTopicSize + swarm.TrojanLengthSize
+	trojanChunkMinDataLen := swarm.SpanSize + NonceSize + TopicSize + LengthSize
 	if len(data) < trojanChunkMinDataLen {
 		return false
 	}
@@ -159,7 +168,7 @@ func checkTargets(targets Targets) error {
 // and its data set to the serialization of the trojan chunk fields which correctly hash into the matching address
 func (m *Message) toChunk(targets Targets, span []byte) (swarm.Chunk, error) {
 	// start out with random nonce
-	nonce := make([]byte, swarm.TrojanNonceSize)
+	nonce := make([]byte, NonceSize)
 	if _, err := rand.Read(nonce); err != nil {
 		return nil, err
 	}
@@ -173,8 +182,7 @@ func (m *Message) toChunk(targets Targets, span []byte) (swarm.Chunk, error) {
 	}
 
 	// hash chunk fields with different nonces until an acceptable one is found
-	// TODO: prevent infinite loop
-	for {
+	for start := time.Now(); ; {
 		s := append(append(span, nonce...), b...) // serialize chunk fields
 		hash1, err := hashBytes(s)
 		if err != nil {
@@ -189,11 +197,16 @@ func (m *Message) toChunk(targets Targets, span []byte) (swarm.Chunk, error) {
 		// else, add 1 to nonce and try again
 		nonceInt.Add(nonceInt, big.NewInt(1))
 		// loop around in case of overflow after 256 bits
-		if nonceInt.BitLen() > (swarm.TrojanNonceSize * 8) {
+		if nonceInt.BitLen() > (NonceSize * swarm.SpanSize) {
+			// Test if timeout after after every 256 iteration
+			if time.Since(start) > (MinerTimeout * time.Second) {
+				break
+			}
 			nonceInt = big.NewInt(0)
 		}
 		nonce = padBytesLeft(nonceInt.Bytes()) // pad in case Bytes call is not 32 bytes long
 	}
+	return nil, ErrMinerTimeout
 }
 
 // hashBytes hashes the given serialization of chunk fields with the hashing func
@@ -245,21 +258,21 @@ func (m *Message) MarshalBinary() (data []byte, err error) {
 
 // UnmarshalBinary deserializes a message struct
 func (m *Message) UnmarshalBinary(data []byte) (err error) {
-	if len(data) < swarm.TrojanLengthSize+swarm.TrojanTopicSize {
+	if len(data) < LengthSize+TopicSize {
 		return ErrUnMarshallingTrojanMessage
 	}
 
-	copy(m.length[:], data[:swarm.TrojanLengthSize])                                            // first 2 bytes are length
-	copy(m.Topic[:], data[swarm.TrojanLengthSize:swarm.TrojanLengthSize+swarm.TrojanTopicSize]) // following 32 bytes are topic
+	copy(m.length[:], data[:LengthSize])                    // first 2 bytes are length
+	copy(m.Topic[:], data[LengthSize:LengthSize+TopicSize]) // following 32 bytes are topic
 
 	length := binary.BigEndian.Uint16(m.length[:])
-	if len(data)-swarm.TrojanLengthSize-swarm.TrojanTopicSize < int(length) {
+	if (len(data) - LengthSize - TopicSize) < int(length) {
 		return ErrUnMarshallingTrojanMessage
 	}
 
 	// rest of the bytes are payload and padding
-	payloadEnd := swarm.TrojanLengthSize + swarm.TrojanTopicSize + length
-	m.Payload = data[swarm.TrojanLengthSize+swarm.TrojanTopicSize : payloadEnd]
+	payloadEnd := LengthSize + TopicSize + length
+	m.Payload = data[LengthSize+TopicSize : payloadEnd]
 	m.padding = data[payloadEnd:]
 	return nil
 }
--- a/pkg/trojan/message.go
+++ b/pkg/trojan/message.go
@@ -1,0 +1,280 @@
+// Copyright 2020 The Swarm Authors
+// This file is part of the Swarm library.
+//
+// The Swarm library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The Swarm library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the Swarm library. If not, see <http://www.gnu.org/licenses/>.
+
+package trojan
+
+import (
+	"bytes"
+	"crypto/rand"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"hash"
+	"math/big"
+
+	"github.com/ethersphere/swarm/bmt"
+	"golang.org/x/crypto/sha3"
+
+	"github.com/ethersphere/bee/pkg/crypto"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// Topic is an alias for a 32 byte fixed-size array which contains an encoding of a message topic
+type Topic [32]byte
+
+// Target is an alias for an address which can be mined to construct a trojan message
+type Target []byte
+
+// Targets is an alias for a collection of targets
+type Targets []Target
+
+// Message represents a trojan message, which is a message that will be hidden within a chunk payload as part of its data
+type Message struct {
+	length  [2]byte // big-endian encoding of Message payload length
+	Topic   Topic
+	Payload []byte // contains the chunk address to be repaired
+	padding []byte
+}
+
+const (
+	// MaxPayloadSize is the maximum allowed payload size for the Message type, in bytes
+	MaxPayloadSize = 4030
+)
+
+var (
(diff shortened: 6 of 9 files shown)


pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+	if err != nil {
+		return tpc
+	}
+	copy(tpc[:], t)
+	return tpc
+}
+
+func hashFunc() hash.Hash {
+	return sha3.NewLegacyKeccak256()
+}
```
mortelli, line 83: as in the `internal` package, you can move this declaration to the top of the file (after `var` and `const` declarations) and add the comment:
```
// hashFunc is a hasher factory used by the bmt hasher
```

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+
+var hasher *bmtlegacy.Hasher
```
mortelli, line 85: you can also move this `var` to be declared with the rest of them at the top of this file (in case we keep it as a global variable)

pkg/swarm/swarm.go
```diff
@@ -102,13 +110,16 @@ type Chunk interface {
...
+	Type() Type
+	WithType(t Type) Chunk
 }
 
 type chunk struct {
 	addr       Address
 	sdata      []byte
 	pinCounter uint64
 	tagID      uint32
+	chunkType  Type
```
mortelli, line 122: i would _really_ like to name this `type`, but it is a reserved word.

leaving this comment here in case anyone else has suggestions for something better than `chunk.chunkType`

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+	// ErrPayloadTooBig is returned when a given payload for a Message type is longer than the maximum amount allowed
+	ErrPayloadTooBig = fmt.Errorf("message payload size cannot be greater than %d bytes", MaxPayloadSize)
+
+	// ErrEmptyTargets is returned when the given target list for a trojan chunk is empty
+	ErrEmptyTargets = errors.New("target list cannot be empty")
+
+	// ErrVarLenTargets is returned when the given target list for a trojan chunk has addresses of different lengths
+	ErrVarLenTargets = errors.New("target list cannot have targets of different length")
+
+	ErrInvalidHasher = errors.New("hasher pool not initialized")
```
mortelli, line 66: as an exported `var`, we should add a comment here

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+// the input string is taken as a byte slice and hashed
+func NewTopic(topic string) Topic {
+	var tpc Topic
+	t, err := crypto.LegacyKeccak256([]byte(topic))
+	if err != nil {
+		return tpc
+	}
+	copy(tpc[:], t)
+	return tpc
+}
```
mortelli, line 79: could you please explain why the copying is necessary here in comparison to the old `swarm` `NewTopic` implementation?

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+
+	// create new Message var and set fields
+	m := new(Message)
+	copy(m.length[:], lengthBuf)
+	m.Topic = topic
+	m.Payload = payload
+	m.padding = padding
+
+	hashPool := bmtlegacy.NewTreePool(hashFunc, swarm.Branches, bmtlegacy.PoolSize)
+	hasher = bmtlegacy.New(hashPool)
```
mortelli, line 114: why are we creating a global hasher in the message constructor? 

this seems dangerous, what if 2 threads use the same hasher in parallel and call `Reset` on it?

or what if once a message is created, a 2nd one is created and the `hasher` var is re-set before the first message gets to use it?

pkg/trojan/message_test.go
```diff
@@ -0,0 +1,248 @@
...
+//
+// The Swarm library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the Swarm library. If not, see <http://www.gnu.org/licenses/>.
+
+package trojan
```
mortelli, line 17: following the `bee` code architecture, this should be a different package called `trojan_testing`

pkg/swarm/swarm.go
santicomp2014: how about chunk.cType?

pkg/trojan/message.go
santicomp2014: Maybe a test to verify it behaves correctly could be a good option.

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+	"github.com/ethersphere/bee/pkg/swarm"
+	bmtlegacy "github.com/ethersphere/bmt/legacy"
+	"golang.org/x/crypto/sha3"
+)
+
+// Topic is an alias for a 32 byte fixed-size array which contains an encoding of a message topic
+type Topic [32]byte
+
+// Target is an alias for an address which can be mined to construct a trojan message
+type Target []byte
```
nolash, line 38: What does it bring to type this separately, instead of just using `swarm.Address`?

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+// Message represents a trojan message, which is a message that will be hidden within a chunk payload as part of its data
+type Message struct {
+	length  [2]byte // big-endian encoding of Message payload length
+	Topic   Topic
+	Payload []byte // contains the chunk address to be repaired
+	padding []byte
+}
+
+const (
+	// MaxPayloadSize is the maximum allowed payload size for the Message type, in bytes
```
nolash, line 52: Please include the reason for the number, too. it's a peculiar size.

Also please note that in `soc` the chunk-and-span comes in addition to the metadata, which is explicitly so given in the book. In this code, however, it's deducted from it. I think we should choose the _same_ approach for both.

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+	MaxPayloadSize = 4030
+)
+
+var (
+	// ErrPayloadTooBig is returned when a given payload for a Message type is longer than the maximum amount allowed
+	ErrPayloadTooBig = fmt.Errorf("message payload size cannot be greater than %d bytes", MaxPayloadSize)
```
nolash, line 58: Please consider putting in `error.go`

pkg/trojan/message.go
nolash, line 66: Please explain this error.

pkg/trojan/message.go
nolash: Please see my previous comment.

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+		return Message{}, ErrPayloadTooBig
+	}
+
+	// get length as array of 2 bytes
+	payloadSize := uint16(len(payload))
+	lengthBuf := make([]byte, 2)
+	binary.BigEndian.PutUint16(lengthBuf, payloadSize)
+
+	// set random bytes as padding
+	paddingLen := MaxPayloadSize - payloadSize
```
nolash, line 100: shouldn't there _always_ be padding? In this case, any chunk at max capacity will always be the same and have weaker security properties those who have less.

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+	padding := make([]byte, paddingLen)
+	if _, err := rand.Read(padding); err != nil {
+		return Message{}, err
+	}
+
+	// create new Message var and set fields
+	m := new(Message)
+	copy(m.length[:], lengthBuf)
```
nolash, line 108: This is local assign to heap, no need to copy.

pkg/trojan/message.go
nolash: I don't understand, why would you re-use the message structure?

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+	if len(data) < 74 {
+		return false
+	}
+
+	// check for valid trojan message length in bytes #41 and #42
+	messageLen := int(binary.BigEndian.Uint16(data[40:42]))
+	return 74+messageLen <= len(data)
+}
+
+// checkTargets verifies that the list of given targets is non empty and with elements of matching size
```
nolash, line 167: Why must the sizes match?

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+// Wrap creates a new trojan chunk for the given targets and Message
+// a trojan chunk is a content-addressed chunk made up of span, a nonce, and a payload which contains the Message
+// the chunk address will have one of the targets as its prefix and thus will be forwarded to the neighbourhood of the recipient overlay address the target is derived from
+func (m *Message) Wrap(targets Targets) (swarm.Chunk, error) {
+	if err := checkTargets(targets); err != nil {
+		return nil, err
+	}
+
+	span := make([]byte, 8)
+	// 4064 bytes for trojan message + 32 bytes for nonce = 4096 bytes as payload for resulting chunk
```
nolash, line 128: This explanatiion in the `MaxPayloadSize` instead please. (cryptically, it doesn't even match that number)

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+// Unwrap creates a new trojan message from the given chunk payload
+// this function assumes the chunk has been validated as a content-addressed chunk
+// it will return the resulting message if the unwrapping is successful, and an error otherwise
+func Unwrap(c swarm.Chunk) (*Message, error) {
+	d := c.Data()
+
+	// unmarshal chunk payload into message
+	m := new(Message)
+	// first 40 bytes are span + nonce
+	err := m.UnmarshalBinary(d[40:])
```
nolash, line 143: Why is this not symmetrical to `Wrap` (eg. `MarshalBinary`)

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+
+	return m, err
+}
+
+// IsPotential returns true if the given chunk is a potential trojan
+func IsPotential(c swarm.Chunk) bool {
+	// chunk must be content-addressed to be trojan
```
nolash, line 150: We should defer to the validator for this check (in that case type is no longer needed).

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+	if c.Type() != swarm.ContentAddressed {
+		return false
+	}
+
+	data := c.Data()
+	// check for minimum chunk data length
+	// span (8) + nonce (32) + length (2) + topic (32) = 74
```
nolash, line 157: Please put this in const documentation instead.

I also find it nice to define each of the type lengths as consts, and add them together, like we had in swarm feeds earlier.

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+	if len(data) < 74 {
+		return false
+	}
+
+	// check for valid trojan message length in bytes #41 and #42
+	messageLen := int(binary.BigEndian.Uint16(data[40:42]))
```
nolash, line 163: umm, this isn't a very strong assertion. how about a checksum (2's complement, crc) instead?

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+	}
+	return nil
+}
+
+// toChunk finds a nonce so that when the given trojan chunk fields are hashed, the result will fall in the neighbourhood of one of the given targets
+// this is done by iteratively enumerating different nonces until the BMT hash of the serialization of the trojan chunk fields results in a chunk address that has one of the targets as its prefix
+// the function returns a new chunk, with the found matching hash to be used as its address,
+// and its data set to the serialization of the trojan chunk fields which correctly hash into the matching address
+func (m *Message) toChunk(targets Targets, span []byte) (swarm.Chunk, error) {
```
nolash, line 185: I am not so sure about this approach. FIrst of all I am not sure about the use-case where you'd want to send to "any-of-the-above" instead of "all-of-the-above" addresses. Perhaps it's in the book somewhere, of course.

But in any case, doesn't it make more sense to spawn parallell mine operations here? That would reduce the complexity of code in the core message component, pushing the responsibility to a(n optional) "miner driver" above that would orchestrate it, and thus probably ease debugging efforts.

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+	// start out with random nonce
+	nonce := make([]byte, 32)
+	if _, err := rand.Read(nonce); err != nil {
+		return nil, err
+	}
+	nonceInt := new(big.Int).SetBytes(nonce)
+	targetsLen := len(targets[0])
+
+	// serialize message
+	b, err := m.MarshalBinary() // TODO: this should be encrypted
```
nolash, line 195: why should it be encrypted? isn't that up to whoever is the client of this code?

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+// the function returns a new chunk, with the found matching hash to be used as its address,
+// and its data set to the serialization of the trojan chunk fields which correctly hash into the matching address
+func (m *Message) toChunk(targets Targets, span []byte) (swarm.Chunk, error) {
+	// start out with random nonce
+	nonce := make([]byte, 32)
+	if _, err := rand.Read(nonce); err != nil {
+		return nil, err
+	}
+	nonceInt := new(big.Int).SetBytes(nonce)
+	targetsLen := len(targets[0])
```
nolash, line 192: Shouldn't this be a same function `getTargetLength` as `checkTargets` would use?

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+
+	// serialize message
+	b, err := m.MarshalBinary() // TODO: this should be encrypted
+	if err != nil {
+		return nil, err
+	}
+
+	// hash chunk fields with different nonces until an acceptable one is found
+	// TODO: prevent infinite loop
```
nolash, line 201: Regarding comment above, if you have a driver on top, max iterations / max time could be part of the orchestration.

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+	if _, err := hasher.Write(s[8:]); err != nil {
+		return nil, err
+	}
+	return hasher.Sum(nil), nil
+}
+
+// contains returns whether the given collection contains the given elem
+func contains(col Targets, elem []byte) bool {
+	for i := range col {
+		if bytes.Equal(elem, col[i]) {
```
nolash, line 244: This is byte-level matching, not bit level. Is that correct?

Would looking at the "proximity" functions in `pkg/swarm` be relevant here?

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+
+		// take as much of the hash as the targets are long
+		if contains(targets, hash1[:targetsLen]) {
+			// if nonce found, stop loop and return chunk
+			return swarm.NewChunk(swarm.NewAddress(hash1), s), nil
+		}
+		// else, add 1 to nonce and try again
+		nonceInt.Add(nonceInt, big.NewInt(1))
+		// loop around in case of overflow
+		if nonceInt.BitLen() > 256 {
```
nolash, line 217: "insanity is doing the same thing over and over again and expecting different results"

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// padBytes adds 0s to the given byte slice as left padding,
+// returning this as a new byte slice with a length of exactly 32
+// given param is assumed to be at most 32 bytes long
+func padBytes(b []byte) []byte {
```
nolash, line 254: `padBytesLeft`?

pkg/trojan/message.go
```diff
@@ -0,0 +1,283 @@
...
+func (m *Message) UnmarshalBinary(data []byte) (err error) {
+	copy(m.length[:], data[:2])  // first 2 bytes are length
+	copy(m.Topic[:], data[2:34]) // following 32 bytes are topic
+
+	// rest of the bytes are payload and padding
+	length := binary.BigEndian.Uint16(m.length[:])
+	payloadEnd := 34 + length
+	m.Payload = data[34:payloadEnd]
+	m.padding = data[payloadEnd:]
+	return nil
```
nolash, line 282: It's a bit puzzling that unmarshaling cannot have an error condition. Please re-evaluate.

pkg/trojan/message_test.go
```diff
@@ -0,0 +1,248 @@
...
+	"math/big"
+	"reflect"
+	"testing"
+
+	chunktesting "github.com/ethersphere/bee/pkg/storage/testing"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// arbitrary targets for tests
+var t1 = Target([]byte{57, 120})
```
nolash, line 31: If you choose `Address` instead of  `Target` re: suggestion above, please consider extending the random address generation functions already available, so that we keep address-related things in the same place.

pkg/trojan/message_test.go
```diff
@@ -0,0 +1,248 @@
...
+// TestNewMessage tests the correct and incorrect creation of a Message struct
+func TestNewMessage(t *testing.T) {
+	smallPayload := make([]byte, 32)
+	m, err := NewMessage(testTopic, smallPayload)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// verify topic
+	if m.Topic != testTopic {
```
nolash, line 61: I would have expected a separate test for topic. Or is that implicitly covered by `padBytes` below?

pkg/trojan/message_test.go
```diff
@@ -0,0 +1,248 @@
...
+	s = []byte{255, 128, 64}
+	p = padBytes(s)
+	e := append(make([]byte, 29), s...) // 29 zeros plus the 3 original bytes
+	if !bytes.Equal(p, e) {
+		t.Fatalf("expected byte padding to result in %x, but is %x", e, p)
+	}
+
+	// simulate toChunk behavior
+	s = []byte{255, 255, 255}
+	i := new(big.Int).SetBytes(s) // byte slice to big.Int
```
nolash, line 160: Here it looks like you're testing whether `big.Int` interprets bytes correctly. That doesn't seem to belong here.

pkg/trojan/message_test.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	addrPrefix := addr.Bytes()[:len(testTargets[0])]
+	if !contains(testTargets, addrPrefix) {
+		t.Fatal("chunk address prefix does not match any of the targets")
+	}
+
+	data := c.Data()
+	dataSize := len(data)
+	expectedSize := 8 + swarm.ChunkSize // span + payload
```
nolash, line 101: Please consider using `file.ChunkWithLengthSize`

pkg/trojan/message_test.go
```diff
@@ -0,0 +1,248 @@
...
+	if dataSize != expectedSize {
+		t.Fatalf("chunk data has an unexpected size of %d rather than %d", dataSize, expectedSize)
+	}
+
+	span := binary.LittleEndian.Uint64(data[:8])
+	remDataLen := len(data[8:])
```
nolash, line 107: `rem`?

pkg/trojan/message.go
mortelli: targets are partial addresses. [eventually](https://github.com/ethersphere/swarm/issues/2156), they will be bit vectors

pkg/trojan/message.go
mortelli: afaik, this will always add padding as long as there is room for it

pkg/trojan/message_test.go
mortelli: `aining` 

(probably renaming this would be a good idea)

pkg/trojan/message_test.go
mortelli: indeed, this can probably be replaced by directly using `s = []byte{1, 0, 0, 0}`

pkg/trojan/message_test.go
mortelli: unit tests for topic could be added. right now it's covered by the tests that implicitly use topic functions, but we could have separate unit tests anyway

pkg/trojan/message.go
mortelli: it can `panic`. not sure we should add panic recovery code inside of an unmarshaling

pkg/trojan/message.go
mortelli: correct. i am too ignorant to know whether `Proximity` is faster than `bytes.Equal`, but this should be replaced by bit-level matching some point

pkg/trojan/message.go
nolash: well it returns error in the signature for a reason ;)

pkg/trojan/message.go
mortelli: i think [canonically](https://github.com/ethersphere/swarm/issues/2160), encryption is part of the whole trojan business.

having said that, this would be a good discussion to have. i guess clients could make this choice for themselves; whether to use encryption or not

pkg/trojan/message.go
mortelli: good question, requesting input from @zelig for this one

pkg/trojan/message.go
mortelli: i think it kind of is; there is a `MarshalBinary` invokation inside of the `toChunk` call (which is called by the `Wrap` function)

the main reason for the asymmetry here is that wrapping involves mining, whereas unwrapping does not

pkg/trojan/message.go
mortelli: agreed on the [parallelization](https://github.com/ethersphere/swarm/issues/2157) point.

regarding "any of the above" vs "all of the above": it will take longer to mine to more than 1 address. 

in terms of global pinning, we just need 1 target. if we are talking about a different use case, such as "message _all_ of these nodes", then yes, we'd need to change this.

maybe this distinction should be implemented. but we'd need to be careful in terms of code structure

pkg/trojan/message.go
mortelli: by this point, `checkTargets` has been called already, so all targets have matching length

pkg/trojan/message.go
mortelli: what about doing the same thing but changing 1 bit each time? :clown_face:

pkg/trojan/message.go
zelig: match a prefix of length d on any of n requires on averages 2^d/n cycles
the use-case is that typically one swarm node will operate multiple overlays each of which can serve as your resident address to receive messages on 'locally'

pkg/trojan/message.go
zelig: i believe i left the same comment in the orig code :)

pkg/trojan/message.go
zelig: at. what point? :0

pkg/trojan/message.go
zelig: ?

pkg/trojan/message_test.go
zelig: please consider doing topic sealing as per the spec instead of testing temporary stuff

pkg/trojan/message_test.go
zelig: with length size???

pkg/trojan/message.go
mortelli: i suppose at the point that the `Targets` become bit vectors

pkg/trojan/message_test.go
nolash: Yeah probably not best name. Perhaps rename this to `...WithSpanSize` while we're at it.

pkg/trojan/message.go
nolash: That is a `pss` construct then, not a `trojan` construct.

pkg/trojan/message_test.go
nolash: it seems to me that "sealing" belongs in the pss package? wouldnt putting it here be mixing areas of responsibility?

pkg/trojan/message.go
jmozah: t is []byte and Topic is fixed size byte. I dont know any other way to convert this.

pkg/swarm/swarm.go
```diff
@@ -94,6 +94,14 @@ func (a Address) MarshalJSON() ([]byte, error) {
 // ZeroAddress is the address that has no value.
 var ZeroAddress = NewAddress(nil)
 
+// Type describes a kind of chunk, whether it is content-addressed or other
+type Type int
```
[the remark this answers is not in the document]
jmozah: I accept that chunk struct in its purity  should only have Address and Data, and functions which applies on them.  Maybe its time to have a chunk package and define other surrounding things over that like type etc. OR
I would for now put type here and take care of moving tags/pin/type  to chunks package in another PR. @janos @nolash WDYT?

pkg/trojan/message.go
jmozah: @nolash what do you mean by "any chunk at max capacity will always be the same and have weaker security properties those who have less."

For now, the payload is the chunk address which is only 32 bytes and there will always be padding. In future if we support sending multiple chunks address that are missing .. then we might have chances that there is no padding

pkg/trojan/message.go
jmozah: Does this mean that all the targets should be in same PO? is this practically possible if we have multiple real world targets?

pkg/trojan/message.go
jmozah: nonceInt.Add(nonceInt, big.NewInt(1)) does it

pkg/trojan/message.go
jmozah: If i use errors.new(fmt.Sprintf()) ... deepsource complains as antipattern. Or did you mean something else ?

jmozah:
@santicomp2014 
> But the global hasher is my main concern also, maybe we can add a test for parallelization to verify it behaves correctly.

I removed the global hasher and now using the bmt pool on every method call (just like verifier)


>It's a good time to add the previous issues of trojan chunks, I will add them to BEE so we don't forget about these in the sprint.

Next Iteration :-)

jmozah:
@mortelli 

> this global hasher business worries me a little bit

I removed this and added like verifier

> the testing should be done in a different package, make sure to export only what is needed (it may be OK as it is right now)

Done
> i believe now is a good time to add (to the bee repo) the old unresolved swarm issues related to the code that is being ported here:

Iteration, iteration :-)

pkg/trojan/message.go
jmozah: Yeah, we will do trojan validator in a later PR, until then for testcases this can be used. Until then, this need not be exported and should use export_test instead.

pkg/trojan/message.go
jmozah: good idea for further improvement. 
Do we assert any incoming protocol message like this ?

pkg/trojan/message.go
jmozah: - as @mortelli said "all-of-the-above" does not fall under global pinning 

- parallel mining should be done using a miner as you said.. 
  We would take that as a improvement (see issue created #444 ).

pkg/crypto/crypto.go
```diff
@@ -56,14 +56,14 @@ func NewEthereumAddress(p ecdsa.PublicKey) ([]byte, error) {
...
-	pubHash, err := legacyKeccak256(pubBytes[1:])
+	pubHash, err := LegacyKeccak256(pubBytes[1:])
 	if err != nil {
 		return nil, err
 	}
 	return pubHash[12:], err
 }
 
-func legacyKeccak256(data []byte) ([]byte, error) {
+func LegacyKeccak256(data []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
Eknir: @nolash , why should we use one over the other? If one is a replacement of the other, should we delete the LegacyKeccak256? If you delete this in your branch, it'll be resolved through merge conflics with master if your or Zahoors branch gets first into master.

pkg/file/splitter/internal/job.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 // maximum amount of file tree levels this file hasher component can handle
-// (128 ^ (9 - 1)) * 4096 = 295147905179352825856 bytes
+// (128 ^ (9 - 1)) * 4096 = 295147905179352825856 bytesgit fetch origin
```
Eknir, line 24: typo

pkg/file/file.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 var (
-	ChunkWithLengthSize = swarm.ChunkSize + 8
+	ChunkWithSpanSize = swarm.ChunkSize + 8
```
Eknir, line 18: swam.Chunksize + swarm.SpanSize

zelig, line 18: there is `swarm.SpanSize` constant too and if you need this should be in `swarm` pkg too

pkg/swarm/swarm.go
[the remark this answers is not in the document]
Eknir: @nolash @jmozah , agreed on both of your points. @nolash , as you made the comment here, could you be so kind as to make an issue such that we document this cleanup?

pkg/trojan/message.go
```diff
@@ -50,7 +36,13 @@ type Message struct {
 
 const (
 	// MaxPayloadSize is the maximum allowed payload size for the Message type, in bytes
+	// MaxPayloadSize + Topic + Length + Nonce = Default ChunkSize
+	//    (4030)      +  (32) +   (2)  +  (32) = 4096 Bytes
 	MaxPayloadSize = 4030
+	NonceSize      = 32
```
zelig, line 42: nonce and topic are known to be hashes,  you can just use HashSize.
and better declare MaxPayloadSize transparently as:

     MaxPayloadSize = swarm.ChunkSize - NonceSize  - LengthSize - TopicSize

pkg/trojan/message.go
```diff
@@ -0,0 +1,278 @@
...
+	Topic   Topic
+	Payload []byte // contains the chunk address to be repaired
+	padding []byte
+}
+
+const (
+	// MaxPayloadSize is the maximum allowed payload size for the Message type, in bytes
+	// MaxPayloadSize + Topic + Length + Nonce = Default ChunkSize
+	//    (4030)      +  (32) +   (2)  +  (32) = 4096 Bytes
+	MaxPayloadSize = 4030
```
Eknir, line 41: If `MaxPayloadSize` completely dependent on other variables, then better define it as such:
`MaxPayloadSize = swarm.ChunkSize - TopicSize - LengthSize - NonceSize`

pkg/trojan/message.go
```diff
@@ -63,27 +55,27 @@ var (
 	// ErrVarLenTargets is returned when the given target list for a trojan chunk has addresses of different lengths
 	ErrVarLenTargets = errors.New("target list cannot have targets of different length")
 
-	ErrInvalidHasher = errors.New("hasher pool not initialized")
+	// ErrUnMarshallingTrojanMessage is returned when a trojan message could not be de-serialized
+	ErrUnMarshallingTrojanMessage = errors.New("trojan message unmarshall error")
```
zelig, line 59: just call this `ErrUnmarshal`  trojan message is implicit in the package, it is ungo to repeat it

pkg/trojan/export_test.go
```diff
@@ -0,0 +1,8 @@
+package trojan
```
zelig, line 1: is this a new pattern we use in bee? interesting

pkg/trojan/message.go
```diff
@@ -0,0 +1,278 @@
...
+	"fmt"
+	"math/big"
+	"time"
+
+	"github.com/ethersphere/bee/pkg/swarm"
+	bmtlegacy "github.com/ethersphere/bmt/legacy"
+)
+
+// Topic is an alias for a 32 byte fixed-size array which contains an encoding of a message topic
+type Topic [32]byte
```
zelig, line 21: i dont think this is needed

pkg/trojan/message.go
Eknir: I think it is important to document that the target length is dependent on the network size. I believe that right now, target is intended to be given as a parameter, right? So, it'll be the responsibility of the client to give the correct parameter size. 

Well, that's fine. But it should be documented somewhere how Target relates to the address and what the length of the Target should be. 

If this documentation is elsewhere, please ignore the comment!

pkg/trojan/message.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if _, err := hasher.Write(s[8:]); err != nil {
+		return nil, err
+	}
+	return hasher.Sum(nil), nil
+}
+
+// contains returns whether the given collection contains the given elem
```
Eknir, line 228: Please don't use abbreviations in comments

pkg/trojan/message.go
zelig: @jmozah it just means that the the message hash is required to match any target destination with the same amount of matching bits

pkg/trojan/message.go
zelig: @nolash @mortelli TBH trojan message itself  is a `pss` construct

pkg/file/file.go
Eknir: @zelig , samen comment, same time :D

pkg/trojan/message.go
Eknir: @zelig , same comment, same time (second time) :D

pkg/trojan/message.go
Eknir: @zelig , agreed. 
`/pkg/pss/trojan/trojan.go`

nolash:
@zelig

> TBH trojan message itself is a pss construct

Why isn't it just called `pss` then? And why isn't it in a package called `pss`? I would be curious to understand better where the one stops and the other takes over.

pkg/trojan/message.go
jmozah: may i know why?

pkg/crypto/crypto.go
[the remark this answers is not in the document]
jmozah: This is out of the scope of this PR.

jmozah:
> I see many unresolved comments still from 3 days ago. Please try to resolve them in the best way possible, so we can proceed.

@Eknir If you dont see"outdated" on top of the comment, it is either
1) responded and need awaiting closure or block from reviewer
2) fixed, but github is not showing "outdated" for some reason (https://github.com/ethersphere/bee/pull/429#discussion_r453450845)

pkg/trojan/export_test.go
jmozah: i think it is a general golang practice...But i saw this first in bee code like you

pkg/trojan/message.go
mortelli: fair enough

pkg/trojan/message.go
mortelli: i would say leave as is until we tackle pss

pkg/swarm/swarm.go
```diff
@@ -10,15 +10,23 @@ import (
...
+	ChunkSize               = SectionSize * Branches
+	HashSize                = 32
+	MaxPO             uint8 = 15
+	MaxBins                 = MaxPO + 1
+	SpanSize                = 8
+	ChunkWithSpanSize       = ChunkSize + SpanSize
+)
+
+var (
+	NewHasher = sha3.NewLegacyKeccak256
```
mortelli, line 29: is there any reason why we call it _new_ hasher in the variable? (and not just _hasher_)

pkg/trojan/message.go
mortelli: this can probably be made into a separate issue then

pkg/trojan/message.go
```diff
@@ -0,0 +1,278 @@
...
+			return swarm.NewChunk(swarm.NewAddress(hash1), s), nil
+		}
+		// else, add 1 to nonce and try again
+		nonceInt.Add(nonceInt, big.NewInt(1))
+		// loop around in case of overflow after 256 bits
+		if nonceInt.BitLen() > (NonceSize * swarm.SpanSize) {
+			// Test if timeout after after every 256 iteration
+			if time.Since(start) > (MinerTimeout * time.Second) {
+				break
+			}
```
mortelli, line 204: `break`ing after a time-out should not be inside this `if`.

it should be evaluated in every mining loop, not just in the case of having to loop around the 256 bits.

jmozah: well.. checking timeout on every operation is not optimized. instead checking it every X mining operation is good. and i chose 256. To be precise i will check how much time it takes for 256 operations in my laptop and see how it

pkg/crypto/crypto.go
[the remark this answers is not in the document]
nolash: The important thing is use same symbol everywhere for "hashing algorithm used in swarm". I believe that suggests it belongs in `swarm/swarm`.

pkg/trojan/message.go
nolash: I mean a separate file, same code.

pkg/trojan/message.go
nolash: > then we might have chances that there is no padding

I think we should be careful about thinking about this package only in the terms of "global pinning" (which your reply suggests). If I get my way, I will be able to craft trojan chunks independently and inject them as chunks in the system. That means that I may very well be filling data there to capacity.

usually padding is used to thwart replay-attacks in crypto. I am not entirely sure what use it has here, though?

pkg/trojan/message.go
nolash: > Do we assert any incoming protocol message like this ?

that depends on the layer further up. trojan is a construct built on content-addressed chunks, so in a way it's already a layer up from soc and content addressed validation. other applications on top may or may not warrant other checks, but that's up to the implementer, then.

pkg/trojan/message.go
nolash: > @zelig , agreed.
/pkg/pss/trojan/trojan.go

I was hoping for the other answer; that `trojan` and `pss` themselves could be kept apart as well-defined concepts.

pkg/trojan/message.go
nolash: I would say that should be done now... :)

pkg/crypto/crypto.go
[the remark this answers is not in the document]
jmozah: we should do this across the system..in another PR. as i said. i used this already in trojan PR.

pkg/trojan/message.go
mortelli: i don't entirely agree, but we can talk about this again when we tackle #444 

it's ok as it is for now :+1:

pkg/swarm/swarm.go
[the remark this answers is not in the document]
nolash: https://github.com/ethersphere/bee/issues/454

pkg/crypto/crypto.go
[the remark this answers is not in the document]
nolash: will you document this as an issue please thanks.

pkg/trojan/message.go
jmozah: > usually padding is used to thwart replay-attacks in crypto. I am not entirely sure what use it has here, though?

I pass this to @zelig

pkg/trojan/message.go
nolash: > I think it is important to document that the target length is dependent on the network size

proposal:

```
type Target swarm.Address
```

Then `Target` can add methods on top of that for anything opinionated. It does however represent an actual swarm address, as in there is nothing in the `pkg/swam/swarm.go` definition/implementation of `Address` that says anything about size.

pkg/trojan/message.go
jmozah: I would leave this to @zelig to decide weather to add this as an improvement to trojan. For now, this is per BOS.

pkg/trojan/message_test.go
nolash: So I read the sealing part in the book. As far as I understand now the sealing only has to do with topic obfuscation. I presume the effect of that is that for every message received you have to iterate over all registered topic and hash them together with the payload to find one that matches.

That begs the question; should it be possible to utilize the trojan functionality without being forced to spend cycles on this. And in extension of that, should topic obfuscation be an optional feature - if I don't care about being opaque, should I be forced to be.

pkg/trojan/message.go
jmozah: i disagree. I would put bitvectors implementation in another PR (#443)  as an improvement. otherwise we would be stuck in this PR for ever.

pkg/trojan/message.go
jmozah: Added validations, but github is not resolving this.

pkg/trojan/message_test.go
jmozah: What does "sealing" mean BTW?

pkg/trojan/message.go
nolash: do we want things to panic here?

pkg/trojan/message_test.go
nolash: `sealedtopic` on the wire is `hash(topic|payload)` so that the receiver would need for every registered `topic` try `hash(topic|payload)` to see if one of them matches `sealedtopic` ... if i understand correctly.

pkg/trojan/message.go
nolash: Ok then I would seriously encourage that this is the next thing that should happen.

pkg/trojan/message.go
mortelli: it's arguable. if none of the error `if` conditions are `true`, and we get an out of bounds error from line `274` onwards, then something has gone _very_ wrong.

`panic` might be preferable to silencing this problem. but i think a case could be made for both alternatives

pkg/trojan/message.go
nolash: @jmozah let's let this go for now, but please make an issue out of reevaluating padding re: these questions.

pkg/trojan/message.go
mortelli: it is a good proposal, but i don't know that we should invest time on it if we plan to implement this through bit vectors anyway.

thoughts?

pkg/swarm/swarm.go
nolash: well the method it aliases also has `New`... in the name

pkg/crypto/crypto.go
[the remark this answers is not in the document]
jmozah: Done https://github.com/ethersphere/bee/issues/455

pkg/trojan/message.go
mortelli: i agree with @nolash on this one

pkg/trojan/message.go
mortelli: > > then we might have chances that there is no padding
> 
> I think we should be careful about thinking about this package only in the terms of "global pinning" (which your reply suggests). If I get my way, I will be able to craft trojan chunks independently and inject them as chunks in the system. That means that I may very well be filling data there to capacity more often.

this would be ideal mo

pkg/swarm/swarm.go
mortelli: i don't think it is a good practice to have words like `new` in the code, since their meaning is questionable, relative, and does not age well in this case.

if it is done in this case to mimic or reference the `sha3` library, that's ok i guess (ftr: i still don't like it :upside_down_face:)

mortelli: thanks @nolash for reaching out and pointing out this was a function and not the actual hasher.

disregard my previous comment, fully support using `new` if we are talking about a constructor :+1:

pkg/trojan/message.go
Eknir: Agree. In my opinion, some documentation is minimally needed.

pkg/trojan/message.go
```diff
@@ -0,0 +1,278 @@
...
+type Message struct {
+	length  [2]byte // big-endian encoding of Message payload length
+	Topic   Topic
+	Payload []byte // contains the chunk address to be repaired
+	padding []byte
+}
+
+const (
+	// MaxPayloadSize is the maximum allowed payload size for the Message type, in bytes
+	// MaxPayloadSize + Topic + Length + Nonce = Default ChunkSize
```
Eknir, line 39: Could you please either remove this comment or make it to be aligned with the code? Thanks :)

pkg/trojan/message.go
Eknir: I think this is out of the scope for this PR anyways.

pkg/trojan/message.go
Eknir: Please, don't take my technical code comments as leading. Depending on you guys here :).

pkg/trojan/message.go
jmozah: I added unmarshall Error

pkg/trojan/message.go
nolash: @jmozah the todo comment is still there...

pkg/trojan/message.go
nolash: @mortelli panicking then means that theoretically if an exploit can be found you can crash the node by passing bogus data to it. I would prefer a rash of errors in the logs instead.

jmozah:
> I want to emphasize that I am not happy about the Type stuff in the swarm package here. It pollutes the namespace, 

@nolash If this PR starts polluting, then i agree that we should not break the first window. But it is already polluted with pinCounter and Tags related stuff. I would rather add this now and take this cleanup in another PR https://github.com/ethersphere/bee/issues/454

PR #430, opened 2020-07-09
api, file, netstore: add targets to query param on download path

santicomp2014:
This PR adds global pinning targets to the download path

Once the target is obtained from the query, then it's injected into the context for its use in netstore.get

Only the Get is being addressed for the time being.
The cmd join could also be added with a targets flag

Download with targets in curl/Browser
`localhost:8081/files/e49bcef241dc2e177ba05719330db67b39eafd060da426efdfb708bd45c24e22?targets=0x2222`


pkg/api/bytes.go
```diff
@@ -71,9 +72,15 @@ func (s *server) bytesGetHandler(w http.ResponseWriter, r *http.Request) {
 		jsonhttp.BadRequest(w, "invalid root chunk")
 		return
 	}
+	type targetsContextKey string
+	k := targetsContextKey("targets")
+
+	targets := r.URL.Query().Get("targets")
+	r = r.WithContext(context.WithValue(r.Context(), k, targets))
```
mortelli, line 79: it is good that you're not using the `string` type for the `WithValue` call, but a new type should not be defined inside of this function.

you can try something like:

```golang
type contextKey string

var (
    targetsContextKey = contextKey("targets")
)
```

alternatively, you can take a look at the `retrieval` and `tracing` packages which define context keys through the use of `struct` and maybe use this as inspiration.

pkg/api/file.go
```diff
@@ -206,11 +207,16 @@ func (s *server) fileDownloadHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	toDecrypt := len(address.Bytes()) == (swarm.HashSize + encryption.KeyLength)
+	type targetsContextKey string
+	k := targetsContextKey("targets")
```
mortelli, line 211: duplicated code

pkg/netstore/netstore.go
```diff
@@ -35,6 +35,12 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 			// request from network
 			data, err := s.retrieval.RetrieveChunk(ctx, addr)
 			if err != nil {
+				// START RECOVERY PROCESS
+				/*
+					if targets := ctx.Value("targets"); targets != "" {
+						//TODO: ADD RECOVERY PROCESS HERE
+					}
+				*/
```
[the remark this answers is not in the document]
santicomp2014: I will condense, no need for this template.
Thanks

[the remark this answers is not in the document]
santicomp2014: Fixed

pkg/api/bytes.go
santicomp2014: Good point, 
I extracted this out of the function.
It can be used as a struct, no need to define as a string.

pkg/api/file.go
santicomp2014: Removed this, as it's declared globally

santicomp2014:
> Thanks for the PR, Santi!
> 
> A couple of things which I noticed:
> 
> * There is no test which validate that we can get the targets via `chunk_test.go`
 added a test for this specific case
> * You explicitly assign the targets to the context in functions  `bzzDownloadHandler` and `fileDownloadHandler`, but you are not doing this in the `chunkGetHandler`. Is getting targets supported on the `/chunks` API endpoint?
I thought we were going to use the fileDownloadHandler, but this is a good point, added the endpoint for this as well.
> * If anything, I would rather push for validating the global pinning logic asap. To do so, only support for one endpoint is needed. I think the `/chunks` endpoint is most important. We can add support for the other endpoints easily after we have validated the logic on the `/chunks` endpoint.

pkg/netstore/netstore.go
```diff
@@ -35,6 +35,7 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 			// request from network
 			data, err := s.retrieval.RetrieveChunk(ctx, addr)
 			if err != nil {
+				//TODO: INVOKE RECOVERY PROCESS, obtain targets thru ctx
```
Eknir, line 38: Could you print something here to the console (perhaps just a log-line), such that I can test out that this works manually?

pkg/api/chunk_test.go
```diff
@@ -118,6 +120,29 @@ func TestChunkUploadDownload(t *testing.T) {
...
+		if !bytes.Equal(validContent, data) {
+			t.Fatal("data retrieved doesnt match uploaded content")
+		}
+
+		// Check if the target is obtained correctly
+		if resp.Header.Get("Targets") != targets {
+			t.Fatal("Invalid Targets")
+		}
+
+	})
```
mortelli, line 145: do we really need to:
1. upload a chunk
2. download it
3. verify its data

just to check that targets are obtained correctly?

pkg/api/file_test.go
```diff
@@ -190,4 +190,79 @@ func TestFiles(t *testing.T) {
 		})
 	})
 
+	t.Run("upload-then-download-with-targets", func(t *testing.T) {
```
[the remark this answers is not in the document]
santicomp2014: good point, reduced to minimal expression, so the test is light and dry.

pkg/api/bytes_test.go
```diff
@@ -57,6 +58,21 @@ func TestBytes(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		if !bytes.Equal(data, content) {
+			t.Fatalf("data mismatch. got %s, want %s", string(data), string(content))
+		}
+		if resp.Header.Get("Targets") != targets {
+			t.Fatalf("targets mismatch. got %s, want %s", resp.Header.Get("Targets"), targets)
+		}
+	})
```
mortelli, line 74: since the bytes download is already being tested separately, and this feature is about obtaining targets from a query parameter, i suggest the following:

```suggestion
	t.Run("download-with-targets", func(t *testing.T) {
		resp := request(t, client, http.MethodGet, resource+"/"+expHash+"?targets="+targets, nil, http.StatusOK)
		if resp.Header.Get("Targets") != targets {
			t.Fatalf("targets mismatch. got %s, want %s", resp.Header.Get("Targets"), targets)
		}
	})
```

i really don't see a need to download the chunk, let alone check that its content is correct (this is covered by `t.Run("download"...` anyways)

LMK what you think

pkg/api/chunk.go
```diff
@@ -140,5 +144,6 @@ func (s *server) chunkGetHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.Header().Set("Content-Type", "binary/octet-stream")
+	w.Header().Set("Targets", targets)
```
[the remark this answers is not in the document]
mortelli: nice! i stand corrected :clap:

pkg/api/file_test.go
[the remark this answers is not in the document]
mortelli: this looks great now. thanks!

pkg/api/file_test.go
```diff
@@ -190,4 +191,23 @@ func TestFiles(t *testing.T) {
...
+		headers := make(http.Header)
+		headers.Add("Content-Type", "text/html; charset=utf-8")
+
+		jsonhttptest.ResponseDirectSendHeadersAndReceiveHeaders(t, client, http.MethodPost, fileUploadResource+"?name="+fileName, bytes.NewReader(simpleData), http.StatusOK, api.FileUploadResponse{
+			Reference: swarm.MustParseHexAddress(rootHash),
+		}, headers)
+
+		fileDownloadResource = func(addr string) string { return "/files/" + addr + "?targets=" + targets }
+
+		rcvdHeader := jsonhttptest.ResponseDirectCheckBinaryResponse(t, client, http.MethodGet, fileDownloadResource(rootHash), nil, http.StatusOK, simpleData, nil)
```
mortelli, line 206: just a small suggestion here, so that we don't have to redefine this func and the test can be condensed further:

```suggestion
		rcvdHeader := jsonhttptest.ResponseDirectCheckBinaryResponse(t, client, http.MethodGet, fileDownloadResource(rootHash)+"?targets="+targets, nil, http.StatusOK, simpleData, nil)
```

not a big deal though :+1:

pkg/api/chunk_test.go
```diff
@@ -118,6 +120,21 @@ func TestChunkUploadDownload(t *testing.T) {
...
+
+		// try to fetch the same chunk
+		resp := request(t, client, http.MethodGet, resourceTargets(validHash), nil, http.StatusOK)
+
+		// Check if the target is obtained correctly
+		if resp.Header.Get("Targets") != targets {
+			t.Fatalf("targets mismatch. got %s, want %s", resp.Header.Get("Targets"), targets)
+		}
+
+	})
```
mortelli, line 137: what do you think if we were to do something like this here?

```suggestion
	t.Run("retrieve-targets", func(t *testing.T) {
		resp := request(t, client, http.MethodGet, resourceTargets(validHash), nil, http.StatusOK)

		// Check if the target is obtained correctly
		if resp.Header.Get("Targets") != targets {
			t.Fatalf("targets mismatch. got %s, want %s", resp.Header.Get("Targets"), targets)
		}
	})
```

pkg/api/chunk_test.go
mortelli: are you sure we need to? 

please see my suggestion in the review down below, just my 2¢

pkg/api/bytes_test.go
santicomp2014: No need for verification, I will change it to this.

pkg/api/file_test.go
santicomp2014: condensed

santicomp2014: fixed

pkg/api/chunk_test.go
santicomp2014: I left it with this implementation

santicomp2014:
> the signature in the first place.
> 
> please amend, thanks.

amended binaries thanks for spotting this.
The `ctx` is being passed into `JoinReadAll`

pkg/netstore/netstore.go
acud: i think that we should not have `TODO`s in the form of logger debug lines. it is pretty straightforward that some functionality needs to be implemented and where. please don't merge this sort of stuff into `master`

pkg/api/bytes_test.go
```diff
@@ -57,6 +58,14 @@ func TestBytes(t *testing.T) {
 		}
 	})
 
+	t.Run("download-with-targets", func(t *testing.T) {
+		resp := request(t, client, http.MethodGet, resource+"/"+expHash+"?targets="+targets, nil, http.StatusOK)
+
+		if resp.Header.Get("Targets") != targets {
```
acud, line 64: this should be extracted to some package level const that defines the `Targets` header. also, `Targets` in the sphere of `HTTP` is very general. it would be better to have some application level header, such as `swarm-recovery-targets` or `swarm-repair-targets` or `swarm-targets` etc you get the point

pkg/api/file_test.go
```diff
@@ -190,4 +191,21 @@ func TestFiles(t *testing.T) {
...
+		fileName := "simple_file.txt"
+		rootHash := "19d2e82c076031ec4e456978f839472d2f1b1b969a765420404d8d315a0c6123"
+		headers := make(http.Header)
+		headers.Add("Content-Type", "text/html; charset=utf-8")
+
+		jsonhttptest.ResponseDirectSendHeadersAndReceiveHeaders(t, client, http.MethodPost, fileUploadResource+"?name="+fileName, bytes.NewReader(simpleData), http.StatusOK, api.FileUploadResponse{
+			Reference: swarm.MustParseHexAddress(rootHash),
+		}, headers)
+
+		rcvdHeader := jsonhttptest.ResponseDirectCheckBinaryResponse(t, client, http.MethodGet, fileDownloadResource(rootHash)+"?targets="+targets, nil, http.StatusOK, simpleData, nil)
```
acud, line 204: i'm not sure if i totally understand the point of this test. what exactly is being tested here? that the server returns the exact same parameter it was provided with in the query-string?

pkg/api/file_test.go
```diff
@@ -190,4 +191,21 @@ func TestFiles(t *testing.T) {
...
+
+		if rcvdHeader.Get("Targets") != targets {
```
acud, line 206: again - hardcoded header name string. please extract

santicomp2014: I resolved it, I replaced it a const in Api.
thanks for the input.

pkg/api/file_test.go
santicomp2014: The idea is to return the Targets to the client, in this case, `swarm-recovery-targets`

pkg/api/bytes_test.go
santicomp2014: I went for `swarm-recovery-targets`
Very good point indeed.
Fixed

PR #439, opened 2020-07-14
Dynamic welcome message

paxthemax:
## Set dynamic welcome message

The purpose of this PR is twofold:

- One: to enable runtime changes of the welcome message for the handshake
- Two: to establish a pattern for runtime change of select configuration parameters through the debug API

### DIFF
--- a/pkg/debugapi/router.go
+++ b/pkg/debugapi/router.go
@@ -86,6 +86,10 @@ func (s *server) setupRouting() {
 	router.Handle("/topology", jsonhttp.MethodHandler{
 		"GET": http.HandlerFunc(s.topologyHandler),
 	})
+	router.Handle("/welcome-message", jsonhttp.MethodHandler{
+		"GET":  http.HandlerFunc(s.welcomeMessageSyncedHandler),
+		"POST": http.HandlerFunc(s.setWelcomeMessageHandler),
+	})
 
 	baseRouter.Handle("/", web.ChainHandlers(
 		logging.NewHTTPAccessLogHandler(s.Logger, logrus.InfoLevel, "debug api access"),
--- a/pkg/debugapi/router.go
+++ b/pkg/debugapi/router.go
@@ -87,7 +87,7 @@ func (s *server) setupRouting() {
 		"GET": http.HandlerFunc(s.topologyHandler),
 	})
 	router.Handle("/welcome-message", jsonhttp.MethodHandler{
-		"GET":  http.HandlerFunc(s.welcomeMessageSyncedHandler),
+		"GET":  http.HandlerFunc(s.getWelcomeMessageHandler),
 		"POST": http.HandlerFunc(s.setWelcomeMessageHandler),
 	})
 
--- a/pkg/debugapi/router.go
+++ b/pkg/debugapi/router.go
@@ -87,8 +87,11 @@ func (s *server) setupRouting() {
 		"GET": http.HandlerFunc(s.topologyHandler),
 	})
 	router.Handle("/welcome-message", jsonhttp.MethodHandler{
-		"GET":  http.HandlerFunc(s.getWelcomeMessageHandler),
-		"POST": http.HandlerFunc(s.setWelcomeMessageHandler),
+		"GET": http.HandlerFunc(s.getWelcomeMessageHandler),
+		"POST": web.ChainHandlers(
+			jsonhttp.NewMaxBodyBytesHandler(welcomeMessageMaxRequestSize),
+			web.FinalHandlerFunc(s.setWelcomeMessageHandler),
+		),
 	})
 
 	baseRouter.Handle("/", web.ChainHandlers(
--- a/pkg/debugapi/welcome_message.go
+++ b/pkg/debugapi/welcome_message.go
@@ -27,8 +27,6 @@ func (s *server) getWelcomeMessageHandler(w http.ResponseWriter, r *http.Request
 }
 
 func (s *server) setWelcomeMessageHandler(w http.ResponseWriter, r *http.Request) {
-	const maxBodySize = 256 // TODO: limit this on all requests?
-
 	var data welcomeMessageRequest
 	err := json.NewDecoder(r.Body).Decode(&data)
 	if err != nil {
--- a/pkg/debugapi/welcome_message.go
+++ b/pkg/debugapi/welcome_message.go
@@ -1,0 +1,38 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package debugapi
+
+import (
+	"io/ioutil"
+	"net/http"
+
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+)
+
+type welcomeMessageResponse struct {
+	WelcomeMesssage string `json:"welcome_message"`
+}
+
+func (s *server) welcomeMessageSyncedHandler(w http.ResponseWriter, r *http.Request) {
+	val := s.P2P.WelcomeMessageSynced()
+	jsonhttp.OK(w, welcomeMessageResponse{
+		WelcomeMesssage: val,
+	})
+}
+
+func (s *server) setWelcomeMessageHandler(w http.ResponseWriter, r *http.Request) {
+	data, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		s.Logger.Debugf("welcome message: request error: %v", err)
+		jsonhttp.InternalServerError(w, err)
+		return
+	}
+	if err := s.P2P.SetWelcomeMessage(string(data)); err != nil {
+		s.Logger.Debugf("welcome message: set error: %v", err)
+		jsonhttp.InternalServerError(w, err)
+		return
+	}
+	jsonhttp.OK(w, nil)
+}
--- a/pkg/debugapi/welcome_message.go
+++ b/pkg/debugapi/welcome_message.go
@@ -11,6 +11,8 @@ import (
 	"github.com/ethersphere/bee/pkg/jsonhttp"
 )
 
+const welcomeMessageMaxRequestSize = 512
+
 type welcomeMessageRequest struct {
 	WelcomeMesssage string `json:"welcome_message"`
 }
--- a/pkg/debugapi/welcome_message.go
+++ b/pkg/debugapi/welcome_message.go
@@ -5,12 +5,15 @@
 package debugapi
 
 import (
-	"io"
-	"io/ioutil"
+	"encoding/json"
 	"net/http"
 
 	"github.com/ethersphere/bee/pkg/jsonhttp"
 )
+
+type welcomeMessageRequest struct {
+	WelcomeMesssage string `json:"welcome_message"`
+}
 
 type welcomeMessageResponse struct {
 	WelcomeMesssage string `json:"welcome_message"`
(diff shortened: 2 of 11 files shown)


Eknir:
@paxthemax , I see you dragged this one to ready for review. Is the PR finished (to your standards)? If so, could you also assign the label "ready for review"?

pkg/debugapi/router.go
```diff
@@ -86,6 +86,10 @@ func (s *server) setupRouting() {
 	router.Handle("/topology", jsonhttp.MethodHandler{
 		"GET": http.HandlerFunc(s.topologyHandler),
 	})
+	router.Handle("/welcome-message", jsonhttp.MethodHandler{
+		"GET":  http.HandlerFunc(s.welcomeMessageSyncedHandler),
```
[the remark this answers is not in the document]
paxthemax: See comment below on RWMutex.

pkg/debugapi/welcome_message_test.go
```diff
@@ -0,0 +1,66 @@
+package debugapi_test
```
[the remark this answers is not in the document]
paxthemax: @mortelli - to reply to your question regarding the RWMutex (and the potentially unfortunate Synced naming):

My understanding is that the welcome message is read concurrently as there are concurrent connections to peers. A RWMutex would make sure that the value of the welcome message should stay synced - as in it should prevent a dirty read. The routine writing the welcome message would block until all routines doing handshakes complete the read.

I did this as an extra safety precaution, although I must admit I cannot estimate the performance implication on the handshake.

Lastly the name is kinda unfortunate - I wanted to attract attention to the fact that it could block.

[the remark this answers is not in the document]
mortelli: > @mortelli - to reply to your question regarding the RWMutex (and the potentially unfortunate Synced naming):
> 
> My understanding is that the welcome message is read concurrently as there are concurrent connections to peers. A RWMutex would make sure that the value of the welcome message should stay synced - as in it should prevent a dirty read. The routine writing the welcome message would block until all routines doing handshakes complete the read.
> 
> I did this as an extra safety precaution, although I must admit I cannot estimate the performance implication on the handshake.

i see. so we have a potential problem if through the API we are trying to set a new welcome message, while it's being read in other threads that are doing handshakes.

so this would be a fairly rare case, but it can definitely happen and it can certainly cause a `panic`. 
 
> Lastly the name is kinda unfortunate - I wanted to attract attention to the fact that it could block.

fair enough. i would switch it to `WelcomeMessage` or `GetWelcomeMessage` (although i think getter and setter wording is not so idiomatic in go) but i do not have a strong preference for either.

pkg/debugapi/welcome_message.go
```diff
@@ -0,0 +1,38 @@
...
+	val := s.P2P.WelcomeMessageSynced()
+	jsonhttp.OK(w, welcomeMessageResponse{
+		WelcomeMesssage: val,
+	})
+}
+
+func (s *server) setWelcomeMessageHandler(w http.ResponseWriter, r *http.Request) {
+	data, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		s.Logger.Debugf("welcome message: request error: %v", err)
```
pradovic, line 28: Maybe we should also log Errof here?

pkg/debugapi/welcome_message.go
```diff
@@ -0,0 +1,38 @@
...
+		jsonhttp.InternalServerError(w, err)
+		return
+	}
+	if err := s.P2P.SetWelcomeMessage(string(data)); err != nil {
+		s.Logger.Debugf("welcome message: set error: %v", err)
```
pradovic, line 33: Same as above

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -57,14 +57,20 @@ type AdvertisableAddressResolver interface {
 	Resolve(observedAdddress ma.Multiaddr) (ma.Multiaddr, error)
 }
 
+// GuardedMessage is a string message guarded by a RW mutex.
+type GuardedMessage struct {
```
pradovic, line 61: I think there is no need for Guarded in the name here.  Also, instead of a struct, you could have also just added `messageMu` field to service and use it to guard a message, so there would be no need for additional struct, especially since methods that actually use the mutex are on service level and not on `GuarderMessage` receiver.

pradovic: If you choose to leave the struct, I would ommit the `Guarded` in the name, and I would make get and set methood on a `GuardedMessage` receiver.

pradovic:
Nice job. However, could you please update the description with some explanation? I don't understand is it actually needed to expand the `p2p` interface for set and get of the welcome message? Do we plan to do these changes in runtime?

pkg/p2p/p2p.go
```diff
@@ -26,6 +26,8 @@ type Service interface {
 	Peers() []Peer
 	SetNotifier(topology.Notifier)
 	Addresses() ([]ma.Multiaddr, error)
+	SetWelcomeMessage(val string) error
+	WelcomeMessageSynced() string
```
pradovic, line 30: I would ommit `Synced` in the name of the method name here. Also, is it really needed to expand the p2p layer with these methods?

pkg/debugapi/welcome_message.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (s *server) welcomeMessageSyncedHandler(w http.ResponseWriter, r *http.Request) {
+	val := s.P2P.WelcomeMessageSynced()
+	jsonhttp.OK(w, welcomeMessageResponse{
+		WelcomeMesssage: val,
+	})
+}
+
+func (s *server) setWelcomeMessageHandler(w http.ResponseWriter, r *http.Request) {
+	data, err := ioutil.ReadAll(r.Body)
```
acud, line 26: if i generate a very large random string (gigabytes order) and upload it to this endpoint it will make the node crumble and die. this should be limited to a predefined size and used with `Read`

pkg/p2p/libp2p/internal/handshake/handshake.go
acud: you can use `atomic.Value` for [this](https://golang.org/pkg/sync/atomic/#Value).

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -259,6 +272,25 @@ func (s *Service) Disconnected(_ network.Network, c network.Conn) {
...
+		return ErrWelcomeMessageLength
+	}
+	s.welcomeMessage.mu.Lock()
+	defer s.welcomeMessage.mu.Unlock()
+	s.welcomeMessage.val = msg
+	return nil
+}
+
+// WelcomeMessageSynced returns the synced value of the current welcome message.
+func (s *Service) WelcomeMessageSynced() string {
```
acud, line 287: this is a classic getter. no need to use `Synced` in the future. `GetWelcomeMessage` would suffice in the future as a naming convention. please note my comment about `atomic.Value` for this purpose as it would suffice and make the type redundant

pkg/debugapi/welcome_message_test.go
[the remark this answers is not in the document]
paxthemax: I was never really sure about the idiomatic wording, but I agree.

pkg/debugapi/welcome_message.go
paxthemax: @acud - excellent point. I'll look to sanitizing all inputs in the PR.

paxthemax:
> Thanks @paxthemax. Please see my comments in the PR. I agree that some code could be removed from the scope of the PR. Also, extending the p2p interface methods with the welcome message stuff seems somehow misplaced and it would be nice if we could find some way around that. I think this can be somehow injected into the p2p layer without the need to extend the interface at all. we can pair on this if you want +1

@acud I'm not too happy with the current solution as well. It corrupts the purity of the P2P interface.

Spitballing solutions (in no order of preference):

- Don't expose it in P2P. Have the `debugapi` have a reference either to lib2p2 or directly to the handshake service. Broader question: does this belong in the P2P interface?
- Expose the handshake service through P2P (`p2p.HandshakeService()`). This removes the need for exposing it on libp2p.
- Define a generic method for setting runtime parameters on P2P (`Get/SetRuntimeOpts` ?). This reduces the triviality of the get/set method but can add unnecessary complexity 

I would be very interested in a more creative approach to this problem.

pradovic:
> > Thanks @paxthemax. Please see my comments in the PR. I agree that some code could be removed from the scope of the PR. Also, extending the p2p interface methods with the welcome message stuff seems somehow misplaced and it would be nice if we could find some way around that. I think this can be somehow injected into the p2p layer without the need to extend the interface at all. we can pair on this if you want +1
> 
> @acud I'm not too happy with the current solution as well. It corrupts the purity of the P2P interface.
> 
> Spitballing solutions (in no order of preference):
> 
>     * Don't expose it in P2P. Have the `debugapi` have a reference either to lib2p2 or directly to the handshake service. Broader question: does this belong in the P2P interface?
> 
>     * Expose the handshake service through P2P (`p2p.HandshakeService()`). This removes the need for exposing it on libp2p.
> 
>     * Define a generic method for setting runtime parameters on P2P (`Get/SetRuntimeOpts` ?). This reduces the triviality of the get/set method but can add unnecessary complexity
> 
> 
> I would be very interested in a more creative approach to this problem.

You could also define a new interface for the purpose of debugapi, which has a p2p interface inside plus these methods. This way you can still have an interface abstraction but would also be able to pass libp2p instance there.

pkg/debugapi/welcome_message.go
paxthemax: @pradovic Is there a general rule/pattern for logging errors in API handlers that I can follow?

acud:
Personally I would go for:
- expose a getter on the debug-api, such as `debugapi.GetWelcomeMessage`
- inject the getter into the p2p abstraction. right now you are doing this with a string, but if you pass a function instead of a string the p2p abstraction can just call the function every time it needs the message

this would require a bit of jugglery in `node.go` since debug api depends on p2p, so you cannot initialize the debug api before you initialize the p2p abstraction. so you need to initialize the p2p layer first with some function that would probably just return the default value as provided in the cli flag, then when the debug api is initialized you would set some other field that will be called within the getter after the debug api is initialized. so, something like:
```go
var customMsgGetter func() string
welcomeGetter := func() string {
 if customMsgGetter != nil { return customMsgGetter() }
 return o.WelcomeMessage
}
p2ps, err := libp2p.New( .... libp2p.Options{
 WelcomeMessageFunc: welcomeGetter,
 ...})

...
debugapi, err := debugapi.New()...
customMsgGetter = debugapi.GetWelcomeMessage()
```
it's a bit ugly and could probably be cleaned up even more if this is properly abstracted, but adding stuff like UI functionalities into a CLI application will always have this sort of overhead having to maintain UI business logic in the codebase (which is pretty much all of the time ugly)

pkg/debugapi/welcome_message.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (s *server) getWelcomeMessageHandler(w http.ResponseWriter, r *http.Request) {
+	val := s.P2P.GetWelcomeMessage()
+	jsonhttp.OK(w, welcomeMessageResponse{
+		WelcomeMesssage: val,
+	})
+}
+
+func (s *server) setWelcomeMessageHandler(w http.ResponseWriter, r *http.Request) {
```
janos, line 25: Given that this is a json-encoded api, using the the body as arbitrary string for the message is not consistent. I would suggest to have welcomeMessageRequest struct to parse and get the message from.

pkg/p2p/p2p.go
```diff
@@ -28,6 +28,13 @@ type Service interface {
 	Addresses() ([]ma.Multiaddr, error)
 }
 
+// DebugService extends the Service with method used for debugging.
+type DebugService interface {
```
paxthemax, line 32: @acud, @pradovic Is this new superset interface an acceptable solution for the scope of this PR?

pkg/debugapi/welcome_message.go
paxthemax: Agreed.

Also, I specifically do not check for the max body size, as @janos will create a new PR to add a new middleware specific for this task.

Issue #433, opened 2020-07-10
Support path in url and map to manifest traversal


pradovic:
Support resolving the relative path from the bzz url in order to be able to serve the website content. If the index.html is served and it ref paths that are present in the manifest, the bee should be able to recognize the relative path and traverse the root manifest from the url and serve appropriate file.

PR #446, opened 2020-07-14
Support manifest traversal

zbiljic:
Initial few commits for validating general idea.

https://github.com/ethersphere/bee/issues/433


pkg/api/router.go
```diff
@@ -54,6 +54,10 @@ func (s *server) setupRouting() {
 		"POST": http.HandlerFunc(s.chunkUploadHandler),
 	})
 
+	handle(router, "/bzz:/{address}/{path:.*}", jsonhttp.MethodHandler{
```
vandot, line 57: we agreed to use `/bzz:/` notation?
https://beehive.ethswarm.org/swarm/pl/fzgc13trxpgjz8k37ft3y8zzgo

pradovic: @zelig do you have an opinion on this maybe? @vandot asked if it would be better to use `/bzz/` instead: https://beehive.ethswarm.org/swarm/pl/fzgc13trxpgjz8k37ft3y8zzgo.

zbiljic: @vandot Sorry, I did not see that, I was following https://hackmd.io/YklWpLHPRDGWipoZavw50Q, and Book of Swarm: manifest API definition, page 227, section 10.2.4 (+).

pradovic:
Duplicate of  https://github.com/ethersphere/bee/pull/452.

PR #447, opened 2020-07-14
api: upload directory

### LINKED ISSUE #431: upload directory
Extend the upload functionality in `bee` to support uploading a directory.

The implementation can use mock (or a minimal implementations of) manifests if needed.

Related user story: #418

mortelli:
This PR implements the **upload directory** feature as described in #431. For this purpose, it uses the manifest code implemented in #452.

### Code overview
- adds the `/dirs` endpoint to the router, which can be used to upload directories.
  - directories should be uploaded in the form of a non-compressed tar file (i.e. `tar -cf`).
  - the directory upload function works in 2 steps: first, extract the dir upload data from the HTTP request, and then upload the directory content by iterating through all the files in the supplied tar.
  - directory content should be accessible through the `bzz` endpoint after being uploaded, using the manifest hash + the path to the content (as described in #452)
- uses the `jsonmanifest` pkg to create a manifest for the uploaded directory, with an entry for each individually uploaded file.
- extracts the `storeFile` func from `fileUploadHandler`.
  - the purpose of this is re-using this code for uploading files in the `storeDir` function.
  - eventually, this func should replace the code in `fileUploadHandler`, but this is going to be done in a separate (later) PR.
  - the `FileUploadInfo` struct was added to make the calling of this func less cumbersome.

### Follow-up issues
- #462: decouple api package
- #463: remove redundant entries in upload/download
- #476: extract jsonmanifest into manifest repo
- #480: add getter for jsonmanifest Entries
- #481: upload empty directory
- #482: upload encrypted directories

closes #431


pkg/upload/upload.go
```diff
@@ -0,0 +1,272 @@
...
+type FileUploadInfo struct {
+	FileName    string
+	FileSize    int64
+	ContentType string
+	ToEncrypt   bool
+	Reader      io.Reader
+}
+
+// GetFileHTTPInfo extracts file info for upload from HTTP request
+func GetFileHTTPInfo(r *http.Request) (*FileUploadInfo, error) {
```
mortelli, line 46: extracted from the current implementation of `pkg/api/file.go/fileUploadHandler`

pkg/upload/upload.go
```diff
@@ -0,0 +1,272 @@
...
+		FileName:    fileName,
+		FileSize:    int64(fileSize),
+		ContentType: contentType,
+		ToEncrypt:   toEncrypt,
+		Reader:      reader,
+	}, nil
+}
+
+// StoreFile stores the given file and returns its reference
+func StoreFile(ctx context.Context, fileInfo *FileUploadInfo, s storage.Storer) (swarm.Address, error) {
```
mortelli, line 129: extracted from the current implementation of `pkg/api/file.go/fileUploadHandler`

pkg/upload/upload.go
```diff
@@ -0,0 +1,280 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
pradovic, line 1: I would leave this in `/api` package for now, and just re-use the function across the api. End then we can separate it into the pkg. Just because I am not sure should we have an upload pkg, or maybe general transfer., or file  or something like this.

pkg/api/dirs.go
```diff
@@ -0,0 +1,39 @@
...
+type dirUploadResponse struct {
+	Reference swarm.Address `json:"reference"`
+}
+
+// dirUploadHandler uploads a directory
+// for now, adapted from old swarm tar upload code
+func (s *server) dirUploadHandler(w http.ResponseWriter, r *http.Request) {
+	dirInfo, err := upload.GetDirHTTPInfo(r)
+	if err != nil {
+		s.Logger.Errorf("dir upload get dir info err: %v", err)
```
pradovic, line 24: You can remove err from the text when Error log is used.

pkg/api/dirs.go
```diff
@@ -0,0 +1,39 @@
...
+		jsonhttp.BadRequest(w, "could not extract dir info from request")
+		return
+	}
+
+	reference, err := upload.StoreDir(r.Context(), dirInfo, s.Storer)
+	if err != nil {
+		s.Logger.Errorf("dir upload store dir err: %v", err)
```
pradovic, line 31: Same as above.

pkg/upload/upload.go
```diff
@@ -0,0 +1,280 @@
...
+			Reader:      bodyReader, // not sure about this one
+		}
+		fileReference, err := StoreFile(ctx, fileInfo, s)
+		logger.Info(fileReference)
+
+		if err != nil {
+			return swarm.ZeroAddress, fmt.Errorf("store dir file error: %v", err)
+		}
+
+		_ = fileReference // what do we do with each file ref?
```
pradovic, line 271: I think you will need a fileRef to construct a manifest, if it is not uploaded by the manifest itself

pkg/upload/upload.go
```diff
@@ -0,0 +1,280 @@
...
+			}*/
+
+			defaultPathFound = true
+		}
+
+		fileInfo := &FileUploadInfo{
+			FileName:    hdr.Name, // or hdr.FileInfo().Name() ??
+			FileSize:    hdr.Size,
+			ContentType: contentType,
+			ToEncrypt:   false,      // when could it be true?
```
pradovic, line 261: I guess it should be propagated from the API, if it should be encrypted or not.

pkg/upload/upload.go
mortelli: done! i will make sure to create another issue to detail these next steps.

pkg/api/dirs.go
mortelli: done

pkg/api/dirs.go
mortelli: done

pkg/upload/upload.go
mortelli: will look into this as soon as #452 is merged into this :+1:

pkg/upload/upload.go
mortelli: i ended up mimicking the `fileUploadHandler` behavior, and just requested a specific header on the dir upload HTTP request.

this is then passed down to all individual upload calls for every file contained in the tar file for the uploaded directory.

pkg/api/file.go
```diff
@@ -198,6 +200,143 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+// FileUploadInfo contains the data for a file to be uploaded
```
Eknir, line 203: Thanks for these changes. Good clean-up!

pkg/api/file.go
```diff
@@ -198,6 +200,143 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+type FileUploadInfo struct {
+	name        string // file name
+	size        int64  // file size
+	contentType string
+	toEncrypt   bool
+	reader      io.Reader
+}
+
+// GetFileHTTPInfo extracts data for a file to be uploaded from an HTTP request
+// this function was extracted from `fileUploadHandler` and should eventually replace its current code, along with `storeFile`
```
Eknir, line 213: Why don't you just do this right now? I understand that would make the changeset bigger, but right now there is a lot of duplication. Also, if the stuff that you wrote here is wrong, it won't be caught by unit tests. Let's be bold and just do the refactoring.

pkg/api/file.go
```diff
@@ -198,6 +200,143 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+func GetFileHTTPInfo(r *http.Request) (*FileUploadInfo, error) {
```
zbiljic, line 214: This function is not used. There also is no reason to export it from the package.

From the comment is apparent that the intention was to use it in some next PR, however I see little point in adding a function right now. This PR would be cleaner without this one, especially since there are no tests for this function.

Also, the code seems to differ from the one about from which it was yanked. Instead of having 2 log lines before returning error (which did introduce a lot of duplication), there are now just opaque errors being returned. How will that look like when the function is used. Are the log lines useful.
There does not seem to be system for handling different error types. The original code returns HTTP 'BadRequest' and 'InternalServerError' responses. How are errors from this function to be translated to those different HTTP errors; or is it to be assumed that only a single HTTP error is returned.

P.S. Sorry for maybe a rant, I would remove this one for now, and save it for some later PR.

pkg/api/file.go
```diff
@@ -198,6 +200,143 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+		size:        int64(fileSize),
+		contentType: contentType,
+		toEncrypt:   toEncrypt,
+		reader:      reader,
+	}, nil
+}
+
+// storeFile uploads the given file and returns its reference
+// this function was extracted from `fileUploadHandler` and should eventually replace its current code, along with `GetFileHTTPInfo`
+func storeFile(ctx context.Context, fileInfo *FileUploadInfo, s storage.Storer) (swarm.Address, error) {
```
zbiljic, line 298: Since this function was extracted from `fileUploadHandler`, why not use it there in this PR instead of leaving existing code?

Why not have this as part of `server` (no need for `storage.Storer` argument)? Was the intention was to make this into some business logic? If so maybe some other package would be better for this.

pkg/api/file.go
```diff
@@ -198,6 +200,143 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+	mr, err := file.SplitWriteAll(ctx, sp, bytes.NewReader(metadataBytes), int64(len(metadataBytes)), fileInfo.toEncrypt)
+	if err != nil {
+		return swarm.ZeroAddress, fmt.Errorf("split metadata error: %v", err)
+	}
+
+	// now join both references (mr, fr) to create an entry and store it
+	e := entry.New(fr, mr)
+	fileEntryBytes, err := e.MarshalBinary()
+	if err != nil {
+		return swarm.ZeroAddress, fmt.Errorf("entry marhsal error: %v", err)
```
zbiljic, line 329: Typo 'marhsal' -> 'marshal'.

pkg/api/file.go
```diff
@@ -198,6 +200,143 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+		return swarm.ZeroAddress, fmt.Errorf("metadata marshal error: %v", err)
+	}
+
+	sp = splitter.NewSimpleSplitter(s)
+	mr, err := file.SplitWriteAll(ctx, sp, bytes.NewReader(metadataBytes), int64(len(metadataBytes)), fileInfo.toEncrypt)
+	if err != nil {
+		return swarm.ZeroAddress, fmt.Errorf("split metadata error: %v", err)
+	}
+
+	// now join both references (mr, fr) to create an entry and store it
```
zbiljic, line 325: I also left this copy/paste (order in comment is different). Just mentioning it here, my plan was to create small PR for these typos later.

pkg/api/dirs.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/manifest/jsonmanifest"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// DirUploadResponse is returned when an HTTP request to upload a directory as a tar is successful
+type DirUploadResponse struct {
```
zbiljic, line 27: Why not using same system as for the `bytesPostResponse` and `fileUploadResponse` structs, which have non-exported main struct, and special ones for test-only, in [export_test.go](https://github.com/ethersphere/bee/blob/master/pkg/api/export_test.go)?

pkg/api/dirs.go
```diff
@@ -0,0 +1,155 @@
...
+	toEncrypt := strings.ToLower(r.Header.Get(encryptHeader)) == "true"
+	return &dirUploadInfo{
+		reader:    r.Body,
+		toEncrypt: toEncrypt,
+	}, nil
+}
+
+// storeDir stores all files recursively contained in the directory given as a tar
+// it returns the hash for the uploaded manifest corresponding to the uploaded dir
+func storeDir(ctx context.Context, dirInfo *dirUploadInfo, s storage.Storer, logger logging.Logger) (swarm.Address, error) {
```
zbiljic, line 72: Same comment as later on. This looks like the intention was to create some business logic function (not being part of `server`). Am I correct in the assumption ?

pkg/api/dirs.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return swarm.ZeroAddress, fmt.Errorf("manifest serialize error: %v", err)
+	}
+
+	// set up reader for manifest file upload
+	r := bytes.NewReader(b)
+
+	// then, upload manifest
+	manifestFileInfo := &FileUploadInfo{
+		name:        "manifest.json",
```
zbiljic, line 143: There is no special reason to have name for this file.
It basically has no difference. It just would not be correct for someone to assume that manifest file needs to be called like this. Especially since we are to move away from JSON structure.

pkg/api/bytes.go
```diff
@@ -23,7 +23,7 @@ type bytesPostResponse struct {
 func (s *server) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
-	toEncrypt := strings.ToLower(r.Header.Get(EncryptHeader)) == "true"
```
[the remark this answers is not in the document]
mortelli: you're right. this can be also done in the following technical debt clean up PR.

i've removed this change

pkg/api/dirs.go
```diff
@@ -0,0 +1,155 @@
...
+// getDirHTTPInfo extracts data for a directory to be uploaded from an HTTP request
+func getDirHTTPInfo(r *http.Request) (*dirUploadInfo, error) {
+	if r.Body == http.NoBody {
+		return &dirUploadInfo{}, errors.New("request has no body")
+	}
+	toEncrypt := strings.ToLower(r.Header.Get(encryptHeader)) == "true"
+	return &dirUploadInfo{
+		reader:    r.Body,
+		toEncrypt: toEncrypt,
+	}, nil
```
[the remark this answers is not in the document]
mortelli: i've gone ahead and created an issue for this since it is coming up a lot: #462

pkg/api/dirs.go
```diff
@@ -0,0 +1,155 @@
...
+		fileHeader, err := tarReader.Next()
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			return swarm.ZeroAddress, fmt.Errorf("read tar stream error: %v", err)
+		}
+
+		filePath := fileHeader.Name
+
+		// only store regular files
```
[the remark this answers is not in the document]
mortelli:
```
// IsRegular reports whether m describes a regular file.
// That is, it tests that no mode type bits are set.
```

this means that things like directories or symbolic links will be skipped

pkg/api/dirs.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	// set up reader for manifest file upload
+	r := bytes.NewReader(b)
+
+	// then, upload manifest
+	manifestFileInfo := &FileUploadInfo{
+		name:        "manifest.json",
+		size:        r.Size(),
+		contentType: ManifestContentType,
```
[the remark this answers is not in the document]
mortelli: regardless of this PR, could you please elaborate a bit on this?

pkg/api/file.go
```diff
@@ -198,6 +200,143 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+// FileUploadInfo contains the data for a file to be uploaded
+type FileUploadInfo struct {
```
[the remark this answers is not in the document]
mortelli: fixed, thanks

pkg/api/file.go
mortelli: per @zbiljic's suggestion, i've removed this function.

refactoring should be done through #462 to keep this PR as small as possible

pkg/api/file.go
mortelli: decoupling `api` code is a task in and of itself, which should be tackled through #462.

you are correct in that there is no `server` receiver as this would eventually exist in a separate component/package.

i had originally extracted this into a separate package; this was discussed [here](https://github.com/ethersphere/bee/pull/447#discussion_r455756952) in @pradovic's review

pkg/upload/upload.go
mortelli: #462

pkg/upload/upload.go
mortelli: this is now taken care of

pkg/api/file.go
mortelli: fixed, thanks

pkg/api/dirs.go
mortelli: entirely correct!

please see [my previous response](https://github.com/ethersphere/bee/pull/447#discussion_r458766684)

pkg/api/dirs.go
mortelli: fixed, thanks

pkg/api/dirs.go
mortelli: you're right, they are entirely the same.

i've removed this struct and changed the code so that it uses `fileUploadResponse`

pkg/api/file.go
mortelli: sure, i only made minor aesthetic changes when extracting this func, but thanks for the heads-up

pkg/api/file.go
mortelli: > This function is not used. There also is no reason to export it from the package.
> 
> From the comment is apparent that the intention was to use it in some next PR, however I see little point in adding a function right now. This PR would be cleaner without this one, especially since there are no tests for this function.
> ...
> P.S. Sorry for maybe a rant, I would remove this one for now, and save it for some later PR.

i've removed this func per yours and @Eknir's [comment](https://github.com/ethersphere/bee/pull/447#discussion_r458681392), good points made all around.

> Also, the code seems to differ from the one about from which it was yanked. Instead of having 2 log lines before returning error (which did introduce a lot of duplication), there are now just opaque errors being returned. How will that look like when the function is used. Are the log lines useful.
There does not seem to be system for handling different error types. The original code returns HTTP 'BadRequest' and 'InternalServerError' responses. How are errors from this function to be translated to those different HTTP errors; or is it to be assumed that only a single HTTP error is returned.

i've thought about this too. we can definitely revisit these points in #462, but from my perspective:
- i don't really see a reason for having the 2 log lines, so this is why i removed it in the extracted code. this wasn't really a change in logic, just removing what's (in my opinion) unnecessary code. i am open to hearing arguments from the team as to why we need to call the logger twice.
- it is true that we are losing granularity in terms of error handling/visibility by thinning the API and extracting these functionalities. i can see why the HTTP response would not include the error, and just log it instead. this also affects tests.

maybe it's a good idea to re-think how errors are handled from the internal packages out to the API, and how visible they are once we are only giving back an HTTP response—whether it is a bad request, internal server error, or something else.

pkg/api/dirs.go
zbiljic: I had no objections to `DirUploadResponse` struct, just noted that it could be non-exported, and in `export_test.go` have:
```
DirUploadResponse = dirUploadResponse
```
so it can be used if needed.

mortelli: but do you think we really need a different response in this sense?

after all, it really _is_ a file upload response, specifically the manifest upload.

pkg/api/file.go
pradovic: I am not sure what is the exact changset but we have established these kind of guidelines about logging and errors:
1. Never log and return an error, do either one of those
2. The pair log.Debugf and log.Errorf is used together. Debugf usually contains more information and it is used for debugging, and is useful on clusters as well. log.Errorf is usually a more sensible message that will be visible for the user.

pkg/api/dirs.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/manifest/jsonmanifest"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// dirUploadInfo contains the data for a directory to be uploaded
+type dirUploadInfo struct {
+	toEncrypt bool
```
acud, line 28: this is kind of meh. i find it better to inject these parameters onto the context object itself, then just pass the context to the underlying components that do the actual work. then you can safely avoid using all of those `toEncrypt` fields all over the place

pkg/api/dirs.go
```diff
@@ -0,0 +1,149 @@
...
+	return &dirUploadInfo{
+		reader:    r.Body,
+		toEncrypt: toEncrypt,
+	}, nil
+}
+
+// storeDir stores all files recursively contained in the directory given as a tar
+// it returns the hash for the uploaded manifest corresponding to the uploaded dir
+func storeDir(ctx context.Context, dirInfo *dirUploadInfo, s storage.Storer, logger logging.Logger) (swarm.Address, error) {
+	dirManifest := jsonmanifest.NewManifest()
```
acud, line 68: not sure, but why do we have a json manifest inside the bee repo? didn't we conclude this will be in a separate repo?

pkg/api/dirs.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// set up HTTP body reader
+	tarReader := tar.NewReader(dirInfo.reader)
```
acud, line 71: would be happy to see this `dirInfo` type gone

pkg/api/dirs.go
```diff
@@ -0,0 +1,149 @@
...
+			break
+		} else if err != nil {
+			return swarm.ZeroAddress, fmt.Errorf("read tar stream error: %v", err)
+		}
+
+		filePath := fileHeader.Name
+
+		// only store regular files
+		if !fileHeader.FileInfo().Mode().IsRegular() {
+			logger.Warningf("skipping file upload for %s as it is not a regular file", filePath)
```
acud, line 87: this can go into `Debug` or even `Trace`

pkg/api/file.go
```diff
@@ -198,6 +200,59 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	// if filename is still empty, use the file hash as the filename
+	if fileInfo.name == "" {
+		fileInfo.name = fr.String()
+	}
+
+	// then store the metadata and get its reference
+	m := entry.NewMetadata(fileInfo.name)
+	m.MimeType = fileInfo.contentType
```
acud, line 229: why not pass this in the constructor one line above?

pkg/api/file.go
```diff
@@ -198,6 +200,59 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	sp = splitter.NewSimpleSplitter(s)
+	mr, err := file.SplitWriteAll(ctx, sp, bytes.NewReader(metadataBytes), int64(len(metadataBytes)), fileInfo.toEncrypt)
+	if err != nil {
+		return swarm.ZeroAddress, fmt.Errorf("split metadata error: %v", err)
+	}
+
+	// now join both references (mr, fr) to create an entry and store it
+	e := entry.New(fr, mr)
```
acud, line 242: tons of indirection here and i am not sure if it is in the right place either....

pkg/api/file.go
```diff
@@ -198,6 +200,59 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+	toEncrypt   bool
+	reader      io.Reader
+}
+
+// storeFile uploads the given file and returns its reference
+// this function was extracted from `fileUploadHandler` and should eventually replace its current code
+func storeFile(ctx context.Context, fileInfo *fileUploadInfo, s storage.Storer) (swarm.Address, error) {
+	// first store the file and get its reference
+	sp := splitter.NewSimpleSplitter(s)
+	fr, err := file.SplitWriteAll(ctx, sp, fileInfo.reader, fileInfo.size, fileInfo.toEncrypt)
```
acud, line 217: the `toEncrypt` here can be removed if it is set on the context

pkg/api/file.go
```diff
@@ -198,6 +200,59 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	// now join both references (mr, fr) to create an entry and store it
+	e := entry.New(fr, mr)
+	fileEntryBytes, err := e.MarshalBinary()
+	if err != nil {
+		return swarm.ZeroAddress, fmt.Errorf("entry marshal error: %v", err)
+	}
+	sp = splitter.NewSimpleSplitter(s)
+	reference, err := file.SplitWriteAll(ctx, sp, bytes.NewReader(fileEntryBytes), int64(len(fileEntryBytes)), fileInfo.toEncrypt)
```
acud, line 248: ditto ctx encrypt

pkg/api/dirs_test.go
```diff
@@ -0,0 +1,247 @@
...
+		if err := tw.WriteHeader(hdr); err != nil {
+			t.Fatal(err)
+		}
+
+		// open and read the file data
+		r, err := os.Open(file.Name())
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer r.Close()
```
acud, line 229: not good. this will be called only when the entire function returns. please close this explicitly after `tw.Write` line

pkg/api/dirs.go
```diff
@@ -0,0 +1,149 @@
...
+			size:        fileHeader.FileInfo().Size(),
+			contentType: contentType,
+			toEncrypt:   dirInfo.toEncrypt,
+			reader:      tarReader,
+		}
+		fileReference, err := storeFile(ctx, fileInfo, s)
+		if err != nil {
+			return swarm.ZeroAddress, fmt.Errorf("store dir file error: %v", err)
+		}
+		logger.Tracef("uploaded dir file %v with reference %v", filePath, fileReference)
```
acud, line 106: do we really need this?

pkg/api/dirs.go
```diff
@@ -0,0 +1,149 @@
...
+			Name:      fileName,
+			Headers:   headers,
+		}
+
+		// add entry to dir manifest
+		dirManifest.Add(filePath, fileEntry)
+	}
+
+	// check if files were uploaded through the manifest
+	if len(dirManifest.Entries) == 0 {
```
acud, line 122: oops. you're accessing an exposed map field which is not protected. so this implementation is not thread safe. would be better if jsonmanifest does not expose this field and has designated getters with mutexes protecting field access. obviously this is not an issue right now since the implementation is single threaded. still better to have this as thread safe from the get-go

pkg/api/dirs.go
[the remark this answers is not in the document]
acud: so this won't support empty directories. would be nice to have the support so we don't lose information when uploading. this was happening in old swarm and i think this is counter-intuitive and an undesired behavior

pkg/api/dirs.go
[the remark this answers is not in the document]
acud: i think what rinke means is what happens if we change the manifest json data structure in future versions. the reader/parser needs to know which schema to expect. i support this idea in general

pkg/api/dirs_test.go
```diff
@@ -0,0 +1,247 @@
...
+		defer os.Remove(f.Name())
+		defer f.Close()
+
+		jsonhttptest.ResponseDirectSendHeadersAndReceiveHeaders(t, client, http.MethodPost, dirUploadResource, f, http.StatusInternalServerError, jsonhttp.StatusResponse{
+			Message: "could not store dir",
+			Code:    http.StatusInternalServerError,
+		}, nil)
+	})
+
+	for _, tc := range []dirTestCase{
```
acud, line 79: inlining the test case data struct into the for loop makes things more readable, but also writeable, since you dont need to scroll all the time to look for the struct definition:
```go
for _,tc := range []struct{
 name string
 field int...
} {
 { name: "test1", field: 123},
 { name: "test2", field: 456}, ...
} {
 //do stuff
}
```

pkg/api/dirs_test.go
```diff
@@ -0,0 +1,247 @@
...
+					if _, err := os.Stat(file.path); os.IsNotExist(err) {
+						if err := os.Mkdir(file.path, 0700); err != nil {
+							t.Fatal(err)
+						}
+						defer os.RemoveAll(file.path)
+					}
+				}
+				f := writeAndOpenFile(t, path.Join(file.path, file.name), file.data)
+				defer os.Remove(f.Name())
+				defer f.Close()
```
acud, line 151: not all that critical but the `defer` will only execute after the entire test case is finished. in general i would avoid usage of `defer` statements inside `for` loops as it is misleading to think the execute after the for loop is done, which is not the case

pkg/api/dirs_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	t.Run("empty request body", func(t *testing.T) {
+		jsonhttptest.ResponseDirectSendHeadersAndReceiveHeaders(t, client, http.MethodPost, dirUploadResource, bytes.NewReader(nil), http.StatusBadRequest, jsonhttp.StatusResponse{
+			Message: "could not extract dir info from request",
+			Code:    http.StatusBadRequest,
+		}, nil)
+	})
+
+	t.Run("empty file", func(t *testing.T) {
+		f := writeAndOpenFile(t, "empty-file", nil)
```
acud, line 58: it is not entirely transparent where these files are created. it would be good to explicitly write to a [temp directory](https://golangcode.com/creating-temp-files/)

pkg/api/dirs.go
```diff
@@ -0,0 +1,149 @@
...
+
+// dirUploadInfo contains the data for a directory to be uploaded
+type dirUploadInfo struct {
+	toEncrypt bool
+	reader    io.ReadCloser
+}
+
+// dirUploadHandler uploads a directory supplied as a tar in an HTTP request
+func (s *server) dirUploadHandler(w http.ResponseWriter, r *http.Request) {
+	dirInfo, err := getDirHTTPInfo(r)
```
acud, line 34: what about checking if the `Content-Type` header indicates that the request is of type tar? [link](https://stackoverflow.com/questions/39058006/proper-mime-media-type-for-tar-files)

pkg/api/dirs.go
mortelli: created #476 for this.

thanks

pkg/api/dirs_test.go
mortelli: this was my initial approach; the problem with this is that temp dirs and files use names with randomly generated appendices. this is good for concurrency.

the reason that this is a problem is that we cannot predict the hash of these temp assets. 

so the alternatives i see right now are:

- leave as is and have the test verify the resulting hashes
- change to explicitly temp files, and do not verify hashes
- somehow verify the hashes for each generated name (no idea how this could be done or even _if_ it could be done)

i don't know that there is a way to generate _explicitely_ temp files and dirs without a randomly generated name other than what i've done here. although if we keep it this way, we lose the benefit of better concurrency handling in case of parallel testing. not sure how much weight to put into this though.

some direction on which alternative to take here would be much appreciated :pray:

pkg/api/dirs.go
mortelli: yeah... you're right. i initially started out with this struct since it had more fields, and was mainly inspired by the `fileUploadInfo` struct. but i don't see much of a reason to keep it around at this point.

i've gone with the context route as you suggested. please have another look and let me know your thoughts on this current implementation.

thanks

pkg/api/dirs.go
mortelli: *poof*

pkg/api/dirs.go
mortelli: i disagree, for me this is an arguably unexpected piece of information about business logic which is worth warning about.

pkg/api/file.go
mortelli: this is extracted code from what was already in the codebase.

we will probably iterate _this_ particular section through #463

pkg/api/file.go
mortelli: also taken from (the already existing) `fileUploadHandler`... i suggest we work on this one through #462 or #463

pkg/api/file.go
mortelli: if i remove `toEncrypt` from `fileInfo`, why not also `size` and `reader`?

i think this discussion is healthy to have, but outside of the scope for this PR for now. we have follow-up issues where we can work on this stuff.

pkg/api/file.go
mortelli: same answer as above

pkg/api/dirs_test.go
mortelli: what if `tw.Write` outputs and error and this `defer` is not in place?

pkg/api/dirs.go
mortelli: i think it's useful when debugging a dir upload. it came in handy while coding the feature, too.

pkg/api/dirs.go
mortelli: nice catch! added this to #480

pkg/api/dirs.go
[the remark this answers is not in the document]
mortelli: sure, but can we tackle this through a subsequent PR? 

it seems like something good to iterate upon, specially since i don't really know how to implement this as of now, and don't want to drag on with this PR much further.

pkg/api/dirs.go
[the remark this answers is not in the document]
mortelli: @Eknir could you please create an issue for this?

pkg/api/file.go
acud: so the reasoning ass follows:
context passing around of values is meant to pass around _request specific_ state and can pass key/value pairs (it is not a map tho); a value _might_ be found on a context, there is not a way to assure that a certain key/value pair will be on the context object except for verifying and tightly guarding the points from which the context propagates that the values are indeed there, propagating correctly to the entire cascade of calls. this is quite difficult to maintain as an approach. theoretically, you can create a context that will carry around even pointers to certain data structs (or readers in this case). but as i said there is no way of verifying that they will be there, thus making the code in the cascade of calls clunky will all sorts of possible `nil` checks and typecasts that slow down execution, etc. the `toEncrypt` value though, is a classic example of what _can_ be passed on context, since it is an optional parameter and you can assume it is false as long as it is not declaratively specified otherwise. `size` and `reader` do not have the same status IMO and trying to pass them around is not really idiomatic go (i've never seen anything like that anywhere), however certainly doable.

acud: also:
https://medium.com/@matryer/context-keys-in-go-5312346a868d
https://stackoverflow.com/questions/40379960/golang-context-withvalue-how-to-add-several-key-value-pairs

pkg/api/dirs_test.go
mortelli: thanks. i've changed this, but it looks a little weird by having to define—what previously was—the `dirTestCaseFile` type twice, since it is inline now.

is there a way around this? am i missing something?

i've looked around and did not find anything explicit for this.

pkg/api/dirs_test.go
mortelli: where would you put these statements instead?

it cannot be before line `170`, and we'd have to keep track of every file and dir created in the test

pkg/api/dirs.go
mortelli: done.

also added a test case for this.

pkg/api/dirs_test.go
acud: i did not mean `dirTestCaseFile`, i meant `dirTestCase`. for `dirTestCaseFile` you can try to employ a similar approach to what i did in for example `pullsync_test.go` with `someChunks` or in `puller_test.go` with `expLiveCalls`. can't think of anything else but it surely increases readability

pkg/api/dirs_test.go
acud: you could create the tar and delete the files in the same function, the indirection you're using here is a bit crooked since you're opening file handles, keeping them open and collecting them for some other function to use. you could create the files one by one and just add them sequentially to the tarball and delete them. also, it might be way simpler to just use a test fixture here and have a prefabricated small tarball in the codebase in a testdata subdirectory. just load that up, unpack, verify, tschuess, instead of all of the constant regeneration of the same data. increases predictability too

acud: see old swarm codebase, `migration_test.go#TestMigrateSanctuaryFixture`

pkg/api/dirs.go
acud: ...

pkg/api/dirs_test.go
acud: then call `r.Close()` inside the `err!=nil` block before `Fatal()`?

pkg/api/dirs.go
acud: stuff that is useful for coding a feature is not necessarily valuable when looking at production code logs, and think about users complaining about what they see in the terminal window ("i see all these errors/warnings, what's wrong with my bee node?")

pkg/api/dirs_test.go
acud: whoops i missed something, why do you need to open the file again? aren't you already passing an open file handle when calling `tarFiles`??

pkg/api/dirs.go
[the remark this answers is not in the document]
mortelli: i've created issue #481 for this.

pkg/api/file.go
mortelli: i really don't think this logging practice is worth it.

having said that, i have added back the double log call to be consistent with the rest of the codebase.

pkg/api/dirs.go
mortelli: it's not _just_ that, though. please have a look at other `Trace` logs we have in the codebase right now, we're logging far more minor stuff than this.

i am open to hearing more arguments about how this `logger` call negatively impacts the codebase, though.

pkg/api/file.go
mortelli: i'm convinced. i've removed this field from the file info struct and am extracting it from the context now. 

thanks for the links too

pkg/api/dirs_test.go
mortelli: thanks. went with something like what was done in `puller_test.go`.

it's definitely more readable now

pkg/api/dirs_test.go
mortelli: > then call r.Close() inside the err!=nil block before Fatal()?

**in general**, i can't see how having 2 `Close()` calls for the same file is better than leaving 1 `defer` inside a `for`.

can you please explain why it is a problem that deferred calls are invoked when the entire function returns? isn't that the whole point of `defer`?

i see this as a good thing; this way we know these statements will not slip by the cracks.

> whoops i missed something, why do you need to open the file again? aren't you already passing an open file handle when calling `tarFiles`??

you're right. i removed the redundant `Open()` in `tarFiles`.

--- 

leaving this aside, i am happy to integrate `dir_test.go` lines `154` to `169` into the `tarFiles` aux func if you think all the file & dir opening/closing logic is better contained this way.

the more i think about it, the more i am leaning towards this alternative.

please lmk your thoughts on this

mortelli: ok, looking at [this comment](https://github.com/ethersphere/bee/pull/447#discussion_r461166611) you made down below.

will update soon

pkg/api/dirs_test.go
mortelli: thanks for all this thoughtful input :heart: 

so i've opted to continue to generate the tar buffer, but without the interaction to disk. as a result, the code is much simpler now.

please see [this reply above](https://github.com/ethersphere/bee/pull/447#discussion_r461297210) and lmk what you think

pkg/api/dirs_test.go
mortelli: this is now somewhat of a moot point due to [the most recent changes](https://github.com/ethersphere/bee/pull/447#discussion_r461297210)

pkg/api/file.go
```diff
@@ -201,6 +202,61 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+	name        string // file name
+	size        int64  // file size
+	contentType string
+	reader      io.Reader
+}
+
+// storeFile uploads the given file and returns its reference
+// this function was extracted from `fileUploadHandler` and should eventually replace its current code
+func storeFile(ctx context.Context, fileInfo *fileUploadInfo, s storage.Storer) (swarm.Address, error) {
+	v := ctx.Value(toEncryptContextKey{})
```
acud, line 216: ideally this would not even be here, and the `toEncryptContextKey` would be exposed in the `sctx` library, and then you would check for the value on the context inside the `SplitWriteAll` function in `file` package. I guess we can live without it for now though since this is a bit of a wider change.

pkg/api/file.go
```diff
@@ -201,6 +202,61 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+	mr, err := file.SplitWriteAll(ctx, sp, bytes.NewReader(metadataBytes), int64(len(metadataBytes)), toEncrypt)
+	if err != nil {
+		return swarm.ZeroAddress, fmt.Errorf("split metadata error: %v", err)
+	}
+
+	// now join both references (mr, fr) to create an entry and store it
+	e := entry.New(fr, mr)
+	fileEntryBytes, err := e.MarshalBinary()
+	if err != nil {
+		return swarm.ZeroAddress, fmt.Errorf("entry marshal error: %v", err)
```
acud, line 249: we usually wrap errors in the codebase when we propagate them up. this helps with type assertions on errors. so please use `%w` instead of `%v`. if you can just skim through the PR and find all of the other occurrences of this type and correct them :pray:

pkg/api/dirs.go
zbiljic: As it was mentioned, this `jsonmanifest` implementation was initially created for tests around this feature, and after some iteration decided to be left as first version. This was all while trying to be in accordance with [feature notes](https://hackmd.io/YklWpLHPRDGWipoZavw50Q).
If we are to have full-blown JSON implementation for this, there are multiple things that need to be changed. Not just for manifest structure, but around entry one also.

Regarding protecting map access with mutex, that is easy, however not sure if there are some cases where we would have the case where this is needed (at least for now). If the policy is to always have a mutex when there is map in external struct, that is okay, and I will follow this going forward.

Eknir:
Connected to issue #431

pkg/api/file.go
mortelli: is this happening in the rest of the codebase then? do we have an issue to tackle this?

acud: no... let's leave it for now

pkg/api/file.go
mortelli: done, thanks!

pkg/api/file.go
janos: @mortelli @zbiljic reasons behind such logging is explained in https://github.com/ethersphere/bee/blob/master/CODING.md#logging.

PR #449, opened 2020-07-15
Reset in streamtest

pradovic:
Update the `streamtest` to support reset functionality as well.

### DIFF
--- a/pkg/p2p/streamtest/streamtest_test.go
+++ b/pkg/p2p/streamtest/streamtest_test.go
@@ -374,7 +374,7 @@ func TestRecorder_resetAfterPartialWrite(t *testing.T) {
 			return fmt.Errorf("flush: %w", err)
 		}
 
-		// deliberately close the stream before the new line character is
+		// deliberately reset the stream before the new line character is
 		// written to the stream
 		if err := stream.Reset(); err != nil {
 			return err
--- a/pkg/p2p/streamtest/streamtest_test.go
+++ b/pkg/p2p/streamtest/streamtest_test.go
@@ -314,6 +314,72 @@ func TestRecorder_closeAfterPartialWrite(t *testing.T) {
 			return err
 		}
 
+		// stream should be closed and write should return err
+		if _, err := rw.WriteString("expect err message"); err != nil {
+			return fmt.Errorf("write: %w", err)
+		}
+
+		if err := rw.Flush(); err == nil {
+			return fmt.Errorf("expected err")
+		}
+
+		return nil
+	}
+
+	err := request(context.Background(), recorder, swarm.ZeroAddress)
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/p2p/streamtest/streamtest_test.go
```diff
@@ -314,6 +314,72 @@ func TestRecorder_closeAfterPartialWrite(t *testing.T) {
 			return err
 		}
 
+		// stream should be closed and write should return err
```
acud, line 317: the comment says `stream should be closed and write should return error`, but below we check if the error is not nil, then we wrap it, so i expect it to fail here and to return the error on line 319. but it doesn't, otherwise the test would fail in line 331. afaik `Close` should close the stream for _writing_ right? 
```go
// Stream is a bidirectional io pipe within a connection.
type MuxedStream interface {
	io.Reader
	io.Writer

	// Close closes the stream for writing. Reading will still work (that
	// is, the remote side can still write).
	io.Closer
        ...
```
so i would assume we should actually check in `318`:
`if _, err := rw.WriteString("expect err message"); err == nil {`
WDYT?

pkg/p2p/streamtest/streamtest_test.go
```diff
@@ -314,6 +314,72 @@ func TestRecorder_closeAfterPartialWrite(t *testing.T) {
...
+		if _, err := rw.WriteString("expect err message"); err != nil {
+			return fmt.Errorf("write: %w", err)
+		}
+
+		if err := rw.Flush(); err == nil {
+			return fmt.Errorf("expected err")
+		}
```
acud, line 324: also, maybe we can test that `Read` still works (since it should after a `Close`)

pkg/p2p/streamtest/streamtest_test.go
pradovic: This is an old test, so I did not want to change the approach. I think that result is the similar, but the error is returned on flush, since it is buffered. I think we would need to use a different reader writer implementation. Do you think we should change it?

pradovic: I can actually do this easily if I wrap existing write with a method that will do flush as well.

pkg/p2p/streamtest/streamtest_test.go
pradovic: I agree, I will also add both read and write checks after reset as well, in the `TestRecorder_resetAfterPartialWrite`

pkg/p2p/streamtest/streamtest_test.go
acud: yeah true, the original test checked only for read (which i believe is in fact incorrect). as you wish, we can keep this section as is and iterate :+1:

pradovic:
@acud Can we merge this one then, I will open a separate request to rewrite the tests?

PR #452, opened 2020-07-16
Support manifest traversal

zbiljic:
See #433

**NOTE**: This is PoC, and uses simple, JSON-based manifest structure, which will be changed to solidity friendly implementation later on. It also lacks various protections against possible attacks.

Includes initial support for resolving the relative path from the `bzz` URL.
If the referenced address (hash) is of type manifest, it will attempt to lookup noted path from URL in that manifest, and if matched return the content of that file.

Route used is `/bzz:/{address}/{path:.*}`.

This implementation borrowed heavily from [`file.go`](https://github.com/ethersphere/bee/blob/master/pkg/api/file.go) and [`bytes.go`](https://github.com/ethersphere/bee/blob/master/pkg/api/bytes.go). Initial PoC uses JSON-based manifest (for ease of testing), which is identified by `application/bzz-manifest+json` MIME type in file metadata. Manifest entry structure contains headers for the referenced file, which will be populated on response. All manifest-based operations are hidden behind interfaces, which should allow for easier switch to any other structure for manifest later on.

#### step 1 - upload 'index.html'

```bash
http POST http://127.0.0.1:8081/files "Content-Type:text/html" < index.html
```

Used HTML content:

```html
<!doctype html>
<html>
<head>
    <title>Swarm website</title>
</head>
<body>
    <img src="images/ethersphere.png" alt="ethersphere">
    <p>Swarm: Censorship resistant storage and communication infrastructure for a truly sovereign digital society</p>
</body>
</html>
```

#### step 2 - upload some image

For example, 'ethersphere.png' was taken.

```bash
http POST http://127.0.0.1:8081/files "Content-Type:image/png" < ethersphere.png
```

#### step 3 - manually create manifest JSON

References used are the ones received for each file upload.
Note that 'index.html' uses 'ethersphere.png' on specific path.

```json
{
    "entries": {
        "index.html": {
            "reference": "<INDEX_HTML_REFERENCE>",
            "name": "index.html",
            "headers": {
                "Content-Type": ["text/html"]
            }
        },
        "images/ethersphere.png": {
            "reference": "<PNG_REFERENCE>",
            "name": "ethersphere.png",
            "headers": {
                "Content-Type": ["image/png"]
            }
        }
    }
}
```

#### step 4 - upload created manifest

We use `/files` API, and set special content type for manifest file (for this PoC).

```bash
http POST http://127.0.0.1:8081/files "Content-Type:application/bzz-manifest+json" < manifest.json
```

#### step 5 - check "website"

It should now be possible to open the "website" in browser by going to:

```
http://127.0.0.1:8081/bzz:/<MANIFEST_FILE_REFERENCE>/index.html
```


pkg/api/bzz.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/ethersphere/bee/pkg/file"
+	"github.com/ethersphere/bee/pkg/file/joiner"
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/gorilla/mux"
+)
+
+const (
+	ManifestType = "application/bzz-manifest+json"
```
pradovic, line 27:
```suggestion
	ManifestContentType = "application/bzz-manifest+json"
```

pkg/api/bzz.go
```diff
@@ -0,0 +1,189 @@
...
+func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
+	addressHex := mux.Vars(r)["address"]
+	path := mux.Vars(r)["path"]
+
+	ctx := r.Context()
+
+	address, err := swarm.ParseHexAddress(addressHex)
+	if err != nil {
+		s.Logger.Debugf("bzz download: parse address %s: %v", addressHex, err)
+		s.Logger.Error("bzz download: parse address error")
```
pradovic, line 39: No need for error word in error logs.
```suggestion
		s.Logger.Error("bzz download: parse address")
```

pkg/api/bzz_test.go
```diff
@@ -0,0 +1,145 @@
...
+		<h1>My First Heading</h1>
+
+		<p>My first paragraph.</p>
+
+		</body>
+		</html>`
+
+		var err error
+		var fileReference swarm.Address
+		var manifestFileReference swarm.Address
```
pradovic, line 61: I would maybe put this in the `TestBzz` scope and just have these individual tests below. Don't see a need for sub-tests.

pkg/api/router.go
```diff
@@ -54,6 +54,10 @@ func (s *server) setupRouting() {
 		"POST": http.HandlerFunc(s.chunkUploadHandler),
 	})
 
+	handle(router, "/bzz:/{address}/{path:.*}", jsonhttp.MethodHandler{
```
vandot, line 57: we agreed to use /bzz:/ notation?
https://beehive.ethswarm.org/swarm/pl/fzgc13trxpgjz8k37ft3y8zzgo

pradovic: https://beehive.ethswarm.org/swarm/pl/sfyot87eqin68cpsum3qspwp7w, I guess we just agreed to try `/bzz/`?

pkg/api/bzz.go
```diff
@@ -0,0 +1,191 @@
...
+
+	entry, err := manifest.FindEntry(path)
+	if err != nil {
+		jsonhttp.BadRequest(w, "invalid path address")
+		return
+	}
+
+	entryAddress := entry.GetReference()
+
+	toDecryptEntry := len(entryAddress.Bytes()) == (swarm.HashSize + encryption.KeyLength)
```
zelig, line 116: maybe we should extract this as a function on addresses

pkg/api/bzz_test.go
```diff
@@ -0,0 +1,143 @@
...
+			t.Fatal(err)
+		}
+
+		mr, err := file.SplitWriteAll(context.Background(), sp, bytes.NewReader(metadataBytes), int64(len(metadataBytes)), false)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// now join both references (mr,fr) to create an entry and store it.
+		entrie := entry.New(fr, mr)
```
zelig, line 105: entrie? nah

pkg/api/bzz.go
```diff
@@ -0,0 +1,191 @@
...
+	outBuffer := bytes.NewBuffer(nil)
+	c, err := file.JoinReadAll(j, entryAddress, outBuffer, toDecryptEntry)
+	if err != nil && c == 0 {
+		s.Logger.Debugf("bzz download: data join %s: %v", entryAddress, err)
+		s.Logger.Errorf("bzz download: data join %s", entryAddress)
+		jsonhttp.NotFound(w, nil)
+		return
+	}
+
+	pr, pw := io.Pipe()
```
zelig, line 141: why all this complexity here? this should be the same as downloading a file from here? why no shared code?

pradovic: I agree, we mentioned this already: https://beehive.ethswarm.org/swarm/pl/83bayftmzfgqbqgocpyxwkn3xc. But we decided to c/p for dev speed :) @zbiljic maybe we could extract it as a separate function in `api` if not in the special component right away?

pkg/api/bzz_test.go
zbiljic: copy/paste from existing code
it looked strange to me as well, however I opted to leave it as-is

pkg/api/bzz_test.go
```diff
@@ -0,0 +1,143 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		if params["filename"] != fileName {
+			t.Fatal("Invalid file name detected")
+		}
+		if rcvdHeader.Get("ETag") != fmt.Sprintf("%q", fileReference) {
+			t.Fatal("Invalid ETags header received")
+		}
+		if rcvdHeader.Get("Content-Type") != "text/html; charset=utf-8" {
```
mortelli, line 130: can we extract `text/hml; charset=utf-8` as a `const`?

zbiljic: This was left as is, because it was used as here on 3 places in `file_test.go` file, and I figured there was a reason for it.
It I were to extract it, it would be changed for that other file.

Since this is for test, and there is no other point for this string, I would leave it as is. It was meant to illustrate point from [pkg/api/bzz_test.go#78](https://github.com/ethersphere/bee/pull/452/files/5f94d6c9d4306315ca4ce316d5efd9f14596bcc0#diff-804afbcbcdabfa91b45c34d0af02358eR78).
If we ever need some greater dealings for this to be `const`, it would probably be better to have another package that contains commonly used mime types.

mortelli: fair enough

pkg/api/bzz.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/gorilla/mux"
+)
+
+const (
+	// ManifestContentType represents content type used for noting that specific
+	// file should be processed as manifest
+	ManifestContentType = "application/bzz-manifest+json"
+)
+
+func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
```
nolash, line 28: This is a _very_ long function

mortelli:
@zbiljic thank you very much for addressing all my comments :+1:

pkg/api/bzz.go
zbiljic: That it is. It was intentionally left like that, to have linear steps that are being performed. There is also a lot of boilerplate regarding HTTP API.
BTW, not the longest one, the [file.go](https://github.com/ethersphere/bee/blob/master/pkg/api/file.go) wins in that regard.

nolash: well mistakes don't need to be repeated

pradovic: I would say this function is not specially long, and having a "long" function is not bad in any general way.  Also, we are generally trying to isolate the PRs to be small, and limit the changset, so I would not spend to much time here to refactor the `files.go` code. Especially, since we asked about this on mattermost and got the info that the code inside was not considered business logic and that it should be used in this way.
In order to refactor this function we should probably:
1. Extract specific error types
2. Remove logging and request rendering form the function and move it to calling functions, based on extracted error types
3. Maybe move this to a special upload/download/transfer component

So, anyways, my 2 c are that we should keep this PR minimal and focuse on the user-story, and open the separate PR to refactor the code, since this code will move stuff from `files.go` and possibily other api calls as well.

pkg/api/common.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	w.Header().Set("ETag", fmt.Sprintf("%q", reference))
+	w.Header().Set("Content-Length", fmt.Sprintf("%d", dataSize))
+	w.Header().Set("Decompressed-Content-Length", fmt.Sprintf("%d", dataSize))
+	if _, err = io.Copy(w, bpr); err != nil {
+		s.Logger.Debugf("api download: data read %s: %v", reference, err)
+		s.Logger.Errorf("api download: data read %s", reference)
+	}
+}
```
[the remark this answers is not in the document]
mortelli: #462

mortelli:
related follow-up issue: #463

PR #461, opened 2020-07-22
Accounting debug api 

metacertain:
The aim of this pull request is to create the

```
/balances
```
```
/balances/{peer}
```

endpoints in the debug api that return the swap balances associated with peers in json format

### DIFF
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -231,6 +231,7 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
 	return peerBalance, nil
 }
 
+// Balances gets balances for all peers, first from memory, than completing from store
 func (a *Accounting) Balances() (map[string]int64, error) {
 	peersBalances := make(map[string]int64)
 
@@ -240,8 +241,7 @@ func (a *Accounting) Balances() (map[string]int64, error) {
 	}
 	a.balancesMu.Unlock()
 
-	err := a.CompleteFromStore(peersBalances)
-
+	err := a.completeFromStore(peersBalances)
 	if err != nil {
 		return nil, err
 	}
@@ -249,12 +249,14 @@ func (a *Accounting) Balances() (map[string]int64, error) {
 	return peersBalances, nil
 }
 
-func (a *Accounting) CompleteFromStore(s map[string]int64) error {
-
+// Get balances from store for keys (peers) that do not already exist in argument map
+// Used to get all balances not loaded in memory at the time the Balances() function is called
+func (a *Accounting) completeFromStore(s map[string]int64) error {
 	err := a.store.Iterate(balancesPrefix, func(key, val []byte) (stop bool, err error) {
-
-		addr := balanceKeyPeer(key)
-
+		addr, err := balanceKeyPeer(key)
+		if err != nil {
+			return false, nil
+		}
 		if _, ok := s[addr.String()]; !ok {
 			var storevalue int64
 			err = a.store.Get(peerBalanceKey(addr), &storevalue)
@@ -262,32 +264,30 @@ func (a *Accounting) CompleteFromStore(s map[string]int64) error {
 				a.logger.Debugf("store operation error for peer %v: %v", addr.String(), err)
 				return false, nil
 			}
+
 			s[addr.String()] = storevalue
 		}
-
 		return false, nil
 	})
 
 	return err
-
 }
 
-func balanceKeyPeer(key []byte) swarm.Address {
+// get the embedded peer from the balance storage key
+func balanceKeyPeer(key []byte) (swarm.Address, error) {
 	k := string(key)
-	var ret swarm.Address
 
 	split := strings.SplitAfter(k, balancesPrefix)
 	if len(split) != 2 {
-		return ret
+		return swarm.ZeroAddress, errors.New("No peer in key")
 	}
 
 	addr, err := swarm.ParseHexAddress(split[1])
-
 	if err != nil {
-		return ret
+		return swarm.ZeroAddress, err
 	}
 
-	return addr
+	return addr, nil
 }
 
 func (pb *PeerBalance) freeBalance() int64 {
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -7,15 +7,18 @@ package accounting
 import (
 	"errors"
 	"fmt"
-	"sync"
-
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/p2p"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
+	"strings"
+	"sync"
 )
 
-var _ Interface = (*Accounting)(nil)
+var (
+	_              Interface = (*Accounting)(nil)
+	balancesPrefix string    = "balance_"
+)
 
 // Interface is the main interface for Accounting
 type Interface interface {
@@ -31,6 +34,8 @@ type Interface interface {
 	Debit(peer swarm.Address, price uint64) error
 	// Balance returns the current balance for the given peer
 	Balance(peer swarm.Address) (int64, error)
+	// Balances returns balances for all known peers
+	Balances() (map[string]int64, error)
 }
 
 // PeerBalance holds all relevant accounting information for one peer
@@ -129,7 +134,7 @@ func (a *Accounting) Credit(peer swarm.Address, price uint64) error {
 
 	a.logger.Tracef("crediting peer %v with price %d, new balance is %d", peer, price, nextBalance)
 
-	err = a.store.Put(balanceKey(peer), nextBalance)
+	err = a.store.Put(peerBalanceKey(peer), nextBalance)
 	if err != nil {
 		return err
 	}
@@ -159,7 +164,7 @@ func (a *Accounting) Debit(peer swarm.Address, price uint64) error {
 
 	a.logger.Tracef("debiting peer %v with price %d, new balance is %d", peer, price, nextBalance)
 
-	err = a.store.Put(balanceKey(peer), nextBalance)
+	err = a.store.Put(peerBalanceKey(peer), nextBalance)
 	if err != nil {
 		return err
 	}
@@ -188,8 +193,8 @@ func (a *Accounting) Balance(peer swarm.Address) (int64, error) {
 }
 
 // get the balance storage key for the given peer
-func balanceKey(peer swarm.Address) string {
-	return fmt.Sprintf("balance_%s", peer.String())
+func peerBalanceKey(peer swarm.Address) string {
+	return fmt.Sprintf("%s%s", balancesPrefix, peer.String())
 }
 
 // getPeerBalance gets the PeerBalance for a given peer
@@ -203,7 +208,7 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
 	if !ok {
 		// balance not yet in memory, load from state store
 		var balance int64
-		err := a.store.Get(balanceKey(peer), &balance)
+		err := a.store.Get(peerBalanceKey(peer), &balance)
 		if err == nil {
 			peerBalance = &PeerBalance{
 				balance:  balance,
@@ -224,6 +229,65 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
 	}
 
 	return peerBalance, nil
+}
+
+func (a *Accounting) Balances() (map[string]int64, error) {
+	peersBalances := make(map[string]int64)
+
+	a.balancesMu.Lock()
+	for peer, balance := range a.balances {
+		peersBalances[peer] = balance.balance
+	}
+	a.balancesMu.Unlock()
+
+	err := a.CompleteFromStore(peersBalances)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return peersBalances, nil
+}
+
+func (a *Accounting) CompleteFromStore(s map[string]int64) error {
+
+	err := a.store.Iterate(balancesPrefix, func(key, val []byte) (stop bool, err error) {
+
+		addr := balanceKeyPeer(key)
+
+		if _, ok := s[addr.String()]; !ok {
+			var storevalue int64
+			err = a.store.Get(peerBalanceKey(addr), &storevalue)
+			if err != nil {
+				a.logger.Debugf("store operation error for peer %v: %v", addr.String(), err)
+				return false, nil
+			}
+			s[addr.String()] = storevalue
+		}
+
+		return false, nil
+	})
+
+	return err
+
+}
+
+func balanceKeyPeer(key []byte) swarm.Address {
+	k := string(key)
+	var ret swarm.Address
+
+	split := strings.SplitAfter(k, balancesPrefix)
+	if len(split) != 2 {
+		return ret
+	}
+
+	addr, err := swarm.ParseHexAddress(split[1])
+
+	if err != nil {
+		return ret
+	}
+
+	return addr
 }
 
 func (pb *PeerBalance) freeBalance() int64 {
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -7,12 +7,13 @@ package accounting
 import (
 	"errors"
 	"fmt"
+	"strings"
+	"sync"
+
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/p2p"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
-	"strings"
-	"sync"
 )
 
 var (
@@ -235,43 +236,40 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
 func (a *Accounting) Balances() (map[string]int64, error) {
 	peersBalances := make(map[string]int64)
 
+	// get peer balances from store first as it may be outdated
+	// compared to the in memory map
+	if err := a.balancesFromStore(peersBalances); err != nil {
+		return nil, err
+	}
+
 	a.balancesMu.Lock()
 	for peer, balance := range a.balances {
 		peersBalances[peer] = balance.balance
 	}
 	a.balancesMu.Unlock()
 
-	err := a.completeFromStore(peersBalances)
-	if err != nil {
-		return nil, err
-	}
-
 	return peersBalances, nil
 }
 
-// Get balances from store for keys (peers) that do not already exist in argument map
-// Used to get all balances not loaded in memory at the time the Balances() function is called
-func (a *Accounting) completeFromStore(s map[string]int64) error {
-	err := a.store.Iterate(balancesPrefix, func(key, val []byte) (stop bool, err error) {
+// Get balances from store for keys (peers) that do not already exist in argument map.
+// Used to get all balances not loaded in memory at the time the Balances() function is called.
+func (a *Accounting) balancesFromStore(s map[string]int64) error {
+	return a.store.Iterate(balancesPrefix, func(key, val []byte) (stop bool, err error) {
 		addr, err := balanceKeyPeer(key)
 		if err != nil {
-			a.logger.Debugf("could not parse address from key: %v: %v", string(key), err)
-			return false, nil
+			return false, fmt.Errorf("parse address from key: %s: %v", string(key), err)
 		}
 		if _, ok := s[addr.String()]; !ok {
 			var storevalue int64
 			err = a.store.Get(peerBalanceKey(addr), &storevalue)
 			if err != nil {
-				a.logger.Debugf("store operation error for peer %v: %v", addr.String(), err)
-				return false, nil
+				return false, fmt.Errorf("get peer %s balance: %v", addr.String(), err)
 			}
 
 			s[addr.String()] = storevalue
 		}
 		return false, nil
 	})
-
-	return err
 }
 
 // get the embedded peer from the balance storage key
@@ -280,7 +278,7 @@ func balanceKeyPeer(key []byte) (swarm.Address, error) {
 
 	split := strings.SplitAfter(k, balancesPrefix)
 	if len(split) != 2 {
-		return swarm.ZeroAddress, errors.New("No peer in key")
+		return swarm.ZeroAddress, errors.New("no peer in key")
 	}
 
 	addr, err := swarm.ParseHexAddress(split[1])
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -255,6 +255,7 @@ func (a *Accounting) completeFromStore(s map[string]int64) error {
 	err := a.store.Iterate(balancesPrefix, func(key, val []byte) (stop bool, err error) {
 		addr, err := balanceKeyPeer(key)
 		if err != nil {
+			a.logger.Debugf("could not parse address from key: %v: %v", string(key), err)
 			return false, nil
 		}
 		if _, ok := s[addr.String()]; !ok {
--- a/pkg/accounting/mock/accounting.go
+++ b/pkg/accounting/mock/accounting.go
@@ -5,45 +5,136 @@
 package mock
 
 import (
-	"github.com/ethersphere/bee/pkg/swarm"
 	"sync"
+
+	"github.com/ethersphere/bee/pkg/accounting"
+	"github.com/ethersphere/bee/pkg/swarm"
 )
 
-type MockAccounting struct {
-	lock     sync.Mutex
-	balances map[string]int64
+// Service is the mock Accounting service.
+type Service struct {
+	lock         sync.Mutex
+	balances     map[string]int64
+	reserveFunc  func(peer swarm.Address, price uint64) error
+	releaseFunc  func(peer swarm.Address, price uint64)
+	creditFunc   func(peer swarm.Address, price uint64) error
+	debitFunc    func(peer swarm.Address, price uint64) error
+	balanceFunc  func(swarm.Address) (int64, error)
+	balancesFunc func() (map[string]int64, error)
 }
 
-func (ma *MockAccounting) Reserve(peer swarm.Address, price uint64) error {
+// WithReserveFunc sets the mock Reserve function
+func WithReserveFunc(f func(peer swarm.Address, price uint64) error) Option {
+	return optionFunc(func(s *Service) {
+		s.reserveFunc = f
+	})
+}
+
+// WithReleaseFunc sets the mock Release function
+func WithReleaseFunc(f func(peer swarm.Address, price uint64)) Option {
+	return optionFunc(func(s *Service) {
+		s.releaseFunc = f
+	})
+}
+
+// WithCreditFunc sets the mock Credit function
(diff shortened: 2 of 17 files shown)


pkg/debugapi/balances.go
```diff
@@ -0,0 +1,72 @@
...
+	Balances []balanceResponse `json:"balances"`
+}
+
+func (s *server) balancesHandler(w http.ResponseWriter, r *http.Request) {
+
+	balances, err := s.Accounting.Balances()
+
+	if err != nil {
+		// TODO: Do we need to return an explicit error?
+		s.Logger.Debugf("debug api: balances: %v", err)
```
mortelli, line 28: i would call internal server error on `w` here and `return`

pkg/debugapi/balances_test.go
```diff
@@ -0,0 +1,103 @@
...
+func TestBalancesError(t *testing.T) {
+	wantErr := errors.New("ASDF")
+	balancesFunc := func() (ret map[string]int64, err error) {
+		return nil, wantErr
+	}
+	testServer := newTestServer(t, testServerOptions{
+		AccountingOpts: []mock.Option{mock.WithBalancesFunc(balancesFunc)},
+	})
+
+	jsonhttptest.ResponseDirect(t, testServer.Client, http.MethodGet, "/balances", nil, http.StatusOK, debugapi.BalancesResponse{})
```
mortelli, line 59: if you change the balances handler so that it returns an error if one occurs, this should be updated so that it does not use `http.StatusOK`

pkg/accounting/mock/accounting.go
```diff
@@ -1,49 +1,136 @@
...
 
-func NewAccounting() *MockAccounting {
-	return &MockAccounting{
-		balances: make(map[string]int64),
+// Balances is the mock function wrapper that calls the set implementation
+func (s *Service) Balances() (map[string]int64, error) {
+	if s.balancesFunc != nil {
+		return s.balancesFunc()
 	}
+	return nil, nil
```
mortelli, line 126: could you please explain why this does not return `s.balances` if `s.balancesFunc == nil`?

pkg/accounting/accounting.go
```diff
@@ -219,3 +226,34 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
 }
+
+func (a *Accounting) getPeersBalances() (map[string]int64, error) {
+	peersBalances := make(map[string]int64)
+
+	for peer, balance := range a.balances {
+		peersBalances[peer] = balance.balance
+	}
+
+	err := a.store.Iterate("balance_", func(key, val []byte) (stop bool, err error) {
```
ralph-pichler, line 237: Should we perhaps extract `balance_` to a constant? it appears now in at least 3 places.

pkg/accounting/mock/accounting.go
```diff
@@ -1,49 +1,136 @@
-// Copyright 2020 The Swarm Authors. All rights reserved.
```
ralph-pichler, line 1: copyright header needs to be restored

pkg/debugapi/balances.go
```diff
@@ -0,0 +1,72 @@
+package debugapi
```
ralph-pichler, line 1: add copyright header

pkg/debugapi/balances.go
ralph-pichler: Something like
```go
jsonhttp.InternalServerError(w, "cannot get balances")
```
should do the trick. That's how the other packages seem to do it.

pkg/debugapi/balances.go
```diff
@@ -0,0 +1,72 @@
...
+		s.Logger.Debugf("debug api: balances peer: parse peer address: %v", err)
+		jsonhttp.BadRequest(w, "malformed peer address")
+		return
+	}
+
+	balance, err := s.Accounting.Balance(peer)
+
+	if err != nil {
+		s.Logger.Debugf("debug-api: balances peer: get peer balance: %v", err)
+		jsonhttp.InternalServerError(w, err)
```
ralph-pichler, line 63: I've looked at other api calls and it seems usually specific error message are never returned through http, only logged (twice).

Errors in the other apis are always handled like this:
```go
s.Logger.Debugf("file upload: file store, file %q: %v", fileName, err)
s.Logger.Errorf("file upload: file store, file %q", fileName)
jsonhttp.InternalServerError(w, "could not store file data")
```

mortelli: yes, that seems to be the case. 

i'm not a huge fan of this logging twice business, so i want to bring this up with @janos or @acud when they come back. i also [discussed it](https://github.com/ethersphere/bee/pull/447#discussion_r458803331) with @zbiljic in another PR.

not returning `err` info in the HTTP response is OK though.

pkg/accounting/accounting.go
```diff
@@ -219,3 +226,34 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+
+		k := string(key)
+		split := strings.SplitAfter(k, "balance_")
+		if len(split) != 2 {
+			return false, fmt.Errorf("invalid overlay key: %s", k)
+		}
```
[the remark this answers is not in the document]
metacertain: if len(split) != 2, that means the split operation couldn't get some string from after "balances_" in the iteration. the only thing the iterator could select like this is the index "balances_" so if we would ever saved a record with empty string as peer address, that could cause this condition to fail

pkg/accounting/accounting.go
```diff
@@ -219,3 +226,34 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+
+		if val, ok := peersBalances[addr.String()]; !ok {
+			peersBalances[addr.String()] = val
+		}
+
+		return false, nil
+	})
+
+	return peersBalances, err
```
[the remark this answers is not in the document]
metacertain: probably yes, as the error could be from the statestore operation, however, before that, we already read items from memory that are probably worth returning.

pkg/debugapi/balances.go
metacertain: but, even if we have an error from Balances(), we probably got most of it right from memory (if statestore has some problem)
I'd like to keep this as is

pkg/accounting/accounting.go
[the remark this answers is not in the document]
mortelli: got it, thanks

pkg/debugapi/balances.go
mortelli: the problem with this approach is that if you ignore `err` on `balancesHandler` (`balances.go:28`) and return a partial resort, you're still doing a `jsonhttp.OK` at the end of the call.

how could the caller tell apart a response with an error from one without one?

e.g.:
- empty `Balances` response: did something go wrong, or are there just no balances?
- non-empty `Balances` response: is this the full list of balances, or am i missing some because an `err` occurred somewhere along the way?

if you keep as is, you need to supply some sort of error info along with the result through `jsonhttp`

paxthemax: @mortelli We were considering this approach, however, does the client care if the internal store access is compromised in this case?

Also, if the internal store is not accessible/responding - what should we return as the result code?

pkg/accounting/accounting.go
```diff
@@ -177,9 +182,13 @@ func (a *Accounting) Balance(peer swarm.Address) (int64, error) {
 	return peerBalance.balance, nil
 }
 
+func (a *Accounting) Balances() (map[string]int64, error) {
```
acud, line 185: why do we need a function that wraps another function? you can just rename `getPeerBalances` to `Balances`

pkg/accounting/accounting.go
```diff
@@ -177,9 +182,13 @@ func (a *Accounting) Balance(peer swarm.Address) (int64, error) {
...
+	return a.getPeersBalances()
+}
+
 // get the balance storage key for the given peer
 func balanceKey(peer swarm.Address) string {
```
acud, line 190: would be nicer to have `peerBalanceKey`

pkg/accounting/accounting.go
```diff
@@ -216,6 +225,37 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+	peersBalances := make(map[string]int64)
+
+	a.balancesMu.Lock()
+	for peer, balance := range a.balances {
+		peersBalances[peer] = balance.balance
+	}
+	a.balancesMu.Unlock()
+
+	err := a.store.Iterate(balancesPrefix, func(key, val []byte) (stop bool, err error) {
+		k := string(key)
```
acud, line 238: would be nice to extract everything from here to `L247` into a helper function in the same manner like in `balanceKey()`, just returning a `swarm.Address, error`

pkg/accounting/accounting.go
```diff
@@ -216,6 +225,37 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+		if len(split) != 2 {
+			return false, fmt.Errorf("invalid overlay key: %s", k)
+		}
+		addr, err := swarm.ParseHexAddress(split[1])
+
+		if err != nil {
+			return false, err
+		}
+
+		if val, ok := peersBalances[addr.String()]; !ok {
```
acud, line 249: this is kind of sketchy. it would work in most cases but it questions the status of the persisted balances and whether they have precedence at all. i find it not so great that there could be some value mismatch between the in-memory and persisted value, and it will be silently ignored here. this is definitely something we need to review, but i guess we can live with this as-is for now.

pkg/accounting/accounting.go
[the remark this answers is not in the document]
acud: > probably yes, as the error could be from the statestore operation, however, before that, we already read items from memory that are probably worth returning.

so basically what you're saying is that the calling context should ignore the error in case there was one (since you are assuming values should still be inspected although an error has occurred). i second what marcello is writing here and would vote to have an explicit error check returning a `nil, err` if there was an error. otherwise `peerBalances, nil`

pkg/accounting/mock/accounting.go
```diff
@@ -1,49 +1,140 @@
...
-	defer ma.lock.Unlock()
-	ma.balances[peer.String()] += int64(price)
+// Debit is the mock function wrapper that calls the set implementation
+func (s *Service) Debit(peer swarm.Address, price uint64) error {
+	if s.debitFunc != nil {
+		return s.debitFunc(peer, price)
+	}
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	s.balances[peer.String()] += int64(price)
```
acud, line 111: would you still not like to increment the changes to balances even though the pluggable function was called? it would be nice to have both working no? maybe this can be in some defer closure which is called at the top of the method (this goes for all of the functions that were modified to accommodate for this change)

pkg/debugapi/balances.go
```diff
@@ -0,0 +1,85 @@
...
+type balanceResponse struct {
+	Peer    string `json:"peer"`
+	Balance int64  `json:"balance"`
+}
+
+type balancesResponse struct {
+	Balances []balanceResponse `json:"balances"`
+}
+
+type balancesResponseWErr struct {
```
acud, line 26: it would be cool if we can just return an error if there was one, and not have to maintain separate contracts. if there's an error return an error, if it was ok return the result

pkg/debugapi/balances.go
acud: guys, if there's an error from statestore make the api return an explicit error message without a balance response. otherwise we end up with a whole lot of messy code on the clients that have to implement these apis. also, statestore error is very severe, and suggests a whole lot of other inconsistency errors which are happening. we should assume the node is already misbehaving/malfunctioning and treat it as such, and not consider the response as `200 OK` in these kind of scenarios

pkg/debugapi/balances.go
```diff
@@ -0,0 +1,85 @@
...
+		s.Logger.Debugf("debug api: balances: %v", err)
+	}
+
+	var balResponses []balanceResponse
+
+	keys := make([]string, 0, len(balances))
+	for k := range balances {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
acud, line 46: ... why?

pkg/debugapi/balances.go
```diff
@@ -0,0 +1,85 @@
...
+
+	for _, k := range keys {
+		balResponses = append(balResponses, balanceResponse{
+			Peer:    k,
+			Balance: balances[k],
+		})
+	}
+
+	if err != nil {
+		jsonhttp.OK(w, balancesResponseWErr{Balances: balResponses, Error: fmt.Sprintf("%v", err)})
```
acud, line 56: `InternalServerError`

pkg/debugapi/balances.go
acud: let's stick with the log twice convention for now. in any case, it will be better that we _don't_ leak developer/internal errors outside the application e.g. on api returned errors

pkg/accounting/accounting.go
```diff
@@ -216,6 +225,37 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+			peersBalances[addr.String()] = val
```
ralph-pichler, line 250: Wrong `val` used here.

pkg/accounting/accounting.go
metacertain: our idea was (with debouncing in mind) that in-memory values are the source of truth, and we just append stuff from store that we don't have in memory

pkg/debugapi/balances.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (s *server) balancesHandler(w http.ResponseWriter, r *http.Request) {
+
+	balances, err := s.Accounting.Balances()
+
+	if err != nil {
+		jsonhttp.InternalServerError(w, "Can not get balances")
+		s.Logger.Debugf("debug api: balances: %v", err)
+		s.Logger.Errorf("debug api: balances: %v", err)
```
janos, line 32: Error log messages should not leak internals, link in this case the message from err. It can just state something like "debug api: could not get balances"

pkg/debugapi/balances.go
```diff
@@ -0,0 +1,77 @@
...
+		return
+	}
+
+	var balResponses []balanceResponse
```
janos, line 36: `balResponses` can also be preallocated, just as `keys` are.

pkg/debugapi/balances.go
```diff
@@ -0,0 +1,77 @@
...
+
+	keys := make([]string, 0, len(balances))
```
janos, line 38: If the `keys` exist only to be sorted for the response, `keys` slice can be avoided (less allocations). `balResponses` can be appended with elements in one loop only and sorted with `sort.Slice` (or to implement `sort.Interface` and use `sort.Sort`) afterwards.

pkg/debugapi/balances.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		s.Logger.Debugf("debug api: balances peer: parse peer address: %v", err)
+		jsonhttp.BadRequest(w, "malformed peer address")
+		return
+	}
+
+	balance, err := s.Accounting.Balance(peer)
+
+	if err != nil {
+		s.Logger.Debugf("debug-api: balances peer: get peer balance: %v", err)
```
janos, line 67: `debug-api:...` -> `debug api:...`

pkg/debugapi/balances.go
janos: @mortelli @ralph-pichler reasons behind such logging is explained in https://github.com/ethersphere/bee/blob/master/CODING.md#logging.

Yes, http response should not expose arbitrary internal information.

pkg/debugapi/balances.go
```diff
@@ -0,0 +1,77 @@
...
+type balancesResponse struct {
+	Balances []balanceResponse `json:"balances"`
+}
+
+func (s *server) balancesHandler(w http.ResponseWriter, r *http.Request) {
+
+	balances, err := s.Accounting.Balances()
+
+	if err != nil {
+		jsonhttp.InternalServerError(w, "Can not get balances")
```
acud, line 30: i missed this, but this error could be extracted to a separate `var` declaration in the beginning of the package, then you can use the error when returning the error here. you can then expose this error in `export_test.go` and verify the error inside the test using the actual error, and not against the hardcoded string which asserts the error now

pkg/accounting/accounting.go
```diff
@@ -216,6 +221,65 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+	err := a.CompleteFromStore(peersBalances)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return peersBalances, nil
+}
+
+func (a *Accounting) CompleteFromStore(s map[string]int64) error {
```
acud, line 242: why is this method exported? also, it is not clear to me from the method name (and the fact that it is not documented) what this function is meant to do

pkg/accounting/accounting.go
```diff
@@ -216,6 +221,65 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+	peersBalances := make(map[string]int64)
+
+	a.balancesMu.Lock()
+	for peer, balance := range a.balances {
+		peersBalances[peer] = balance.balance
+	}
+	a.balancesMu.Unlock()
+
+	err := a.CompleteFromStore(peersBalances)
+
```
acud, line 234: dangling whitespace

pkg/accounting/accounting.go
```diff
@@ -216,6 +221,65 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+		return nil, err
+	}
+
+	return peersBalances, nil
+}
+
+func (a *Accounting) CompleteFromStore(s map[string]int64) error {
+
+	err := a.store.Iterate(balancesPrefix, func(key, val []byte) (stop bool, err error) {
+
```
acud, line 245: whitespace

pkg/accounting/accounting.go
```diff
@@ -216,6 +221,65 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+		addr := balanceKeyPeer(key)
+
```
acud, line 247: whitespace... etc... try to minimize the amount of whitespaces you introduce

pkg/accounting/accounting.go
```diff
@@ -216,6 +221,65 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+				return false, nil
+			}
+			s[addr.String()] = storevalue
+		}
+
+		return false, nil
+	})
+
+	return err
+
```
acud, line 262: another one

pkg/accounting/accounting.go
```diff
@@ -216,6 +221,65 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+}
+
+func balanceKeyPeer(key []byte) swarm.Address {
+	k := string(key)
+	var ret swarm.Address
```
acud, line 267: this is not necessary to allocate.
you can remove this line and in the cases below where you return `ret` just `return swarm.ZeroAddress` instead. this is not contradiction even if you change the signature to `Address,error`, since you would still need to return some value for the address when returning an error

pkg/accounting/accounting.go
```diff
@@ -216,6 +221,65 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+		}
+
+		return false, nil
+	})
+
+	return err
+
+}
+
+func balanceKeyPeer(key []byte) swarm.Address {
```
acud, line 265: it would be clearer if this function returns `(swarm.Address, error)`

pkg/debugapi/balances.go
```diff
@@ -0,0 +1,70 @@
...
+			Balance: balances[k],
+		})
+	}
+
+	jsonhttp.OK(w, balancesResponse{Balances: balResponses})
+
+}
+
+func (s *server) peerBalanceHandler(w http.ResponseWriter, r *http.Request) {
+	// TODO: Currently, we do not check the length of the hex address (such as prepending zeroes), should we?
```
acud, line 49: this `TODO` can be removed. if such check is made, it can be made inside `ParseHexAddress`, returning an error to the calling context if need be

pkg/jsonhttp/jsonhttptest/jsonhttptest.go
```diff
@@ -148,6 +148,24 @@ func ResponseUnmarshal(t *testing.T, client *http.Client, method, url string, bo
...
+	resp := request(t, client, method, url, body, responseCode, nil)
+	defer resp.Body.Close()
+
+	got, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		t.Fatal(err)
+	}
+	got = bytes.TrimSpace(got)
+	tmp := response
+	_ = json.Unmarshal(got, &tmp)
```
acud, line 163: don't ignore the error here, also, `tmp` is not needed. i believe you can unmarshal directly into `response`

pkg/soc/validator_test.go
```diff
@@ -54,7 +54,7 @@ func TestSocValidator(t *testing.T) {
 	// check invalid address
 	sch.Data()[0] = 0x00
 	wrongAddressBytes := sch.Address().Bytes()
-	wrongAddressBytes[0] = 255 - wrongAddressBytes[0]
+	wrongAddressBytes[0] ^= wrongAddressBytes[0]
```
ralph-pichler, line 57: You're reintroducing a recently fixed issue here. Revert this.

pkg/soc/validator_test.go
```diff
@@ -40,7 +40,7 @@ func TestSocValidator(t *testing.T) {
 	}
 
 	// check valid chunk
-	v := soc.NewValidator()
+	v := soc.NewSocValidator()
```
ralph-pichler, line 43: Why is this in this PR? And all the other validator naming changes?

metacertain: https://github.com/ethersphere/bee/pull/461/commits/c00182cd1114c466dd7d9421191827257a1f64f3

these come from the rebase conflict fixing commit, will revert this

metacertain:
Some fixes in the latest commit addresses nearly-missed issues from https://github.com/ethersphere/bee/pull/427#pullrequestreview-459122006

pkg/accounting/accounting.go
```diff
@@ -203,7 +208,7 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
 	if !ok {
 		// balance not yet in memory, load from state store
 		var balance int64
-		err := a.store.Get(balanceKey(peer), &balance)
+		err := a.store.Get(peerBalanceKey(peer), &balance)
```
mortelli, line 211: comment unrelated to this particular diff, but since `reserved` is an `int`, its default value will be `0`. 

and you are anyway ignoring it outside of the `getPeerBalance` func, so you can just skip the setting of the `reserved` value in the structs returned in this function.

pkg/accounting/accounting.go
```diff
@@ -226,6 +231,66 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+			err = a.store.Get(peerBalanceKey(addr), &storevalue)
+			if err != nil {
+				a.logger.Debugf("store operation error for peer %v: %v", addr.String(), err)
+				return false, nil
+			}
+
+			s[addr.String()] = storevalue
+		}
+		return false, nil
+	})
```
mortelli, line 272: it is a bit unclear to me how we are making use of this `Iterate` call. it seems like the in-line func you are passing to it as a parameter never returns an error, so unless i'm seeing it wrong, `completeFromStore` will _always_ return `nil`. 

please correct me if i'm mistaken, and if not, could you please explain why we are silencing errors from the store iteration?

pkg/debugapi/balances_test.go
```diff
@@ -0,0 +1,151 @@
...
+	got := jsonhttptest.ResponseReturnDirect(t, testServer.Client, http.MethodGet, "/balances", nil, http.StatusOK, &debugapi.BalancesResponse{
+		[]debugapi.BalanceResponse{
+			{
+				Peer:    "J",
+				Balance: 5,
+			},
+		},
+	}).(*debugapi.BalancesResponse)
+
+	if !comparisonOfBalances(got, expected) {
```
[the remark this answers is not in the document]
metacertain: yes, it flaked because of maps unorderedness. an alternative would be 

 if fmt.Sprint(expected) != fmt.Sprint(got) {
        t.Errorf("Got: %v, not as expected: %v", got, expected)
    }

as uncovered by @paxthemax

janos:
I have updated this PR by addressing some comments that are left and also made some improvements. It is ready for another review. I am not sure why Validator naming changes are here, but I did not revert them.

pkg/debugapi/balances_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// We expect a list of items unordered by peer:
+	got := jsonhttptest.ResponseReturnDirect(t, testServer.Client, http.MethodGet, "/balances", nil, http.StatusOK, &debugapi.BalancesResponse{
+		[]debugapi.BalanceResponse{
+			{
+				Peer:    "J",
+				Balance: 5,
+			},
+		},
+	}).(*debugapi.BalancesResponse)
```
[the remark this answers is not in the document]
mortelli: very nice change @janos

pkg/debugapi/balances_test.go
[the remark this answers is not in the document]
mortelli: @janos could you please clear this up for me?

i no longer see any references to balances being sorted, although i might be missing this.

if i am correct, do we really need the `equalBalances` func instead of just doing a `reflect.DeepEqual`?

mortelli:
> I have updated this PR by addressing some comments that are left and also made some improvements. It is ready for another review. I am not sure why Validator naming changes are here, but I did not revert them.

thank you! 

regarding the naming changes, please see [this comment](https://github.com/ethersphere/bee/pull/461#issuecomment-668566226) from @metacertain.

also, it seems @ralph-pichler made an additional comment [previously](https://github.com/ethersphere/bee/pull/461#pullrequestreview-460852504), although i am not exactly sure of what he is referring to.

pkg/debugapi/balances_test.go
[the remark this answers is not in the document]
acud: reflect deep equal checks for order, not just content, and here the sequence of balances is not guaranteed, so you must check for content but without giving importance to order

acud:
> I have updated this PR by addressing some comments that are left and also made some improvements. It is ready for another review. I am not sure why Validator naming changes are here, but I did not revert them.

the whole validator interface naming is annoying and tbh I am a bit oblivious to why on earth does a single validator need to implement an _interface_ in its own. then this whole ordeal of `Validator` vs `ChunkValidator` vs `ChunksValidator` vs `ChunkValidators` is prevented. or we can find some completely orthogonal name for the validator that checks all validators (validator, validator)

pkg/debugapi/balances_test.go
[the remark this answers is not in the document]
mortelli: thanks for the clarification :+1:

janos:
Thanks guys for very thoughtful reviews and the merge.

PR #465, opened 2020-07-22
settlement protocol and payment threshold

ralph-pichler:
This PR introduces a simple pretend settlement protocol where peers send payment messages (which are just integers of the fake payment). It introduces the notion of payment threshold upon which payments are triggered and drops the disconnect threshold configuration in favour of a tolerance above the payment threshold (the tolerance must be less than half the payment threshold).

To make the various comparisons in accounting a bit clearer the following terminology is introduced:
* `expectedBalance` is the balance we expect with a peer if all reserved actions are successful (=`balance-reserved`)
* `expectedDebt` is the debt we expect to have with a peer if all reserved actions are successful (=`max(-expectedBalance, 0)`)

The accounting logic is changed in the following way:
* `Reserve` only blocks calls once the `expectedDebt` has already exceeded the payment threshold.
* `Credit` triggers settlement of the actual debt when `expectedDebt` exceeds the payment threshold.
* `Debit` returns a disconnect error when a peers debt exceeds `paymentThreshold` + `paymentTolerance`
* `Accounting` exposes a function to be notified of payments which adjust the balance. Payments which would put the node into more debt than the `paymentTolerance` are rejected.

The rationale behind this is as follows:

* If a node is retrieving a lot of chunks from one of its peers concurrently the reserved balance for this peer might make up a significant part of the payment threshold. If it continues making requests and they all return successfully the node might get disconnected from the peer as it cannot send a payment message in time. Blocking further retrieves until the payment threshold is actually hit is not good either, so instead it sends a payment as soon as it might potentially cross the threshold but only for the amount of debt that is confirmed so far. This implies that during high load with a peer smaller payments will be made (but more frequently).
* There is still blocking inbetween the first retrieval where the expected debt exceeds the threshold and the next successful retrieval request with the peer, but this should already be much less than in the previous design (also retrieval will still try `maxPeers-1` other peers, so the retrieve operation might still work). If this is still too much blocking we can initiate settlement during `Credit` already once expected debt goes near the payment threshold.
* If there are no concurrent requests nothing changes compared to the old design. Expected debt and actual balance will cross the payment threshold during the same `Credit` call.


pkg/settlement/pseudosettle/pseudosettle.go
```diff
@@ -0,0 +1,101 @@
...
+
+	s.logger.Tracef("sending payment message from peer %v of %d", peer, amount)
+	w := protobuf.NewWriter(stream)
+	err = w.WriteMsgWithContext(ctx, &pb.Payment{
+		Amount: amount,
+	})
+	return err
+}
+
+func (s *Service) SetPaymentObserver(observer settlement.PaymentObserver) {
```
Eknir, line 99: Can you explain to me what this is?

pkg/settlement/pseudosettle/pseudosettle.go
```diff
@@ -0,0 +1,101 @@
...
+	var req pb.Payment
+	if err := r.ReadMsg(&req); err != nil {
+		return fmt.Errorf("read request: %w peer %v", err, p.Address)
+	}
+
+	s.logger.Tracef("received payment message from peer %v of %d", p.Address, req.Amount)
+	return s.observer.NotifyPayment(p.Address, req.Amount)
+}
+
+func (s *Service) Pay(ctx context.Context, peer swarm.Address, amount uint64) error {
```
Eknir, line 75: Missing comment

pkg/settlement/pseudosettle/pseudosettle.go
Eknir: Never mind, I get it. But more documentation around this would be useful :), also around line 258 in `node.go`

pkg/accounting/accounting.go
```diff
@@ -216,6 +261,44 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+// expectedDebt returns the debt we expect to have with a peer if all reserved funds are actually credited
+func (pb *PeerBalance) expectedDebt() uint64 {
+	expectedBalance := pb.expectedBalance()
+	if expectedBalance >= 0 {
+		return 0
+	}
+	return uint64(-expectedBalance)
+}
+
+// NotifyPayment is called by Settlement when we received payment
```
Eknir, line 278: Maybe a comment that, by implementing this function, Accounting is a PaymentObserver

pkg/accounting/accounting.go
```diff
@@ -216,6 +261,44 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+		return err
+	}
+
+	balance.lock.Lock()
+	defer balance.lock.Unlock()
+
+	nextBalance := balance.balance - int64(amount)
+
+	if nextBalance < 0 {
+		return fmt.Errorf("refusing to accept payment which would put us in debt, new balance would have been %d", nextBalance)
```
Eknir, line 291: I think this check is too harsh. 
First concern: 
- A node might actually send a higher amount that what we think he should send us, due to differences in balances
Second concern:
- Why shouldn't I just accept a higher payment and go on?

My suggestion would be to: always accept a payment, where we do proper accounting if the payment received makes the balance above zero (but doesn't trigger a payment from our side) and where we don't do "proper" accounting when the payment would trigger us to send a cheque back immediately.

pkg/accounting/accounting.go
```diff
@@ -126,19 +131,59 @@ func (a *Accounting) Credit(peer swarm.Address, price uint64) error {
...
+	if expectedDebt >= a.paymentThreshold {
+		a.settle(peer, balance)
+	}
 
 	return nil
 }
 
+// settle all debt with a peer
+// the lock on balance must be held when called
+func (a *Accounting) settle(peer swarm.Address, balance *PeerBalance) {
```
janos, line 154: If settle function would return an error, error messages would not have to contain function arguments, just reasons for failure, and the logging could be done where the function is called with any additional information required in its context, or to act accordingly if there is an error. One case is if we want to continue on calling settle if error rate is very high, which we cannot detect currently.

pkg/accounting/accounting.go
```diff
@@ -216,6 +261,44 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
 	return peerBalance, nil
 }
 
-func (pb *PeerBalance) freeBalance() int64 {
+// expectedBalance returns the balance we expect to have with a peer if all reserved funds are actually credited
+func (pb *PeerBalance) expectedBalance() int64 {
```
janos, line 265: This function has a potential panic of int64 overflow.

pkg/settlement/pseudosettle/pseudosettle.go
```diff
@@ -0,0 +1,101 @@
...
+	defer func() {
+		if err != nil {
+			_ = stream.Reset()
+		} else {
+			_ = stream.FullClose()
+		}
+	}()
+	var req pb.Payment
+	if err := r.ReadMsg(&req); err != nil {
+		return fmt.Errorf("read request: %w peer %v", err, p.Address)
```
janos, line 68: Chained error in error message is usually at the end.

ralph-pichler: fixed. I took this part of the code from `retrieval`. It should probably be fixed there as well for consistent log messages.

cmd/bee/cmd/start.go
```diff
@@ -194,7 +196,8 @@ Welcome to the Swarm.... Bzzz Bzzzz Bzzzz
 	cmd.Flags().String(optionNameTracingServiceName, "bee", "service name identifier for tracing")
 	cmd.Flags().String(optionNameVerbosity, "info", "log verbosity level 0=silent, 1=error, 2=warn, 3=info, 4=debug, 5=trace")
 	cmd.Flags().String(optionWelcomeMessage, "", "send a welcome message string during handshakes")
-	cmd.Flags().Uint64(optionNameDisconnectThreshold, 100000000000, "threshold in BZZ until which you allow peers to be indebted before disconnecting")
+	cmd.Flags().Uint64(optionNamePaymentThreshold, 100000000000, "threshold in BZZ where you expect to get paid from your peers")
+	cmd.Flags().Uint64(optionNamePaymentTolerance, 100000000000, "excess debt above payment threshold in BZZ where you disconnect from your peer")
```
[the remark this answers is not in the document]
ralph-pichler: Let's go with absolutes for now. I also reduced both of those values significantly so that regular payments actually occur.

pkg/accounting/accounting.go
```diff
@@ -42,18 +44,22 @@ type PeerBalance struct {
...
-	balancesMu          sync.Mutex // mutex for accessing the balances map
-	balances            map[string]*PeerBalance
-	logger              logging.Logger
-	store               storage.StateStorer
-	disconnectThreshold uint64 // the debt threshold at which we will disconnect from a peer
+	balancesMu       sync.Mutex // mutex for accessing the balances map
+	balances         map[string]*PeerBalance
+	logger           logging.Logger
+	store            storage.StateStorer
+	paymentThreshold uint64 // the debt threshold we communicate to our peers
```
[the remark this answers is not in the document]
ralph-pichler: fixed

pkg/accounting/accounting.go
```diff
@@ -42,18 +44,22 @@ type PeerBalance struct {
...
+	paymentTolerance uint64 // the amount we let peers exceed the payment threshold before disconnected
```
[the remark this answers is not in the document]
ralph-pichler: fixed

pkg/settlement/pseudosettle/pseudosettle.go
ralph-pichler: fixed

pkg/accounting/accounting.go
ralph-pichler: It now allows putting us into debt up to the payment tolerance

pkg/accounting/accounting.go
```diff
@@ -126,19 +131,59 @@ func (a *Accounting) Credit(peer swarm.Address, price uint64) error {
...
 
+// settle all debt with a peer
+// the lock on balance must be held when called
+func (a *Accounting) settle(peer swarm.Address, balance *PeerBalance) {
+	// don't do anything if there is no actual debt
+	// this might be the case if the peer owes us and the total reserve for a peer exceeds the payment treshhold
+	if balance.balance >= 0 {
+		return
+	}
+	paymentAmount := uint64(-balance.balance)
```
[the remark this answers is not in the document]
Eknir: I think this is not needed.

pkg/accounting/accounting.go
Eknir: Cool, thanks! This works if the payment tolerance is strictly less than they payment threshold. Do you think there is a need for a hard check that this is the case when we start up the node? Or perhaps only documenting it is enough?
Don't forget to update the PR description to include this too!

pkg/settlement/pseudosettle/pseudosettle.go
```diff
@@ -0,0 +1,102 @@
...
+		} else {
+			_ = stream.FullClose()
+		}
+	}()
+	var req pb.Payment
+	if err := r.ReadMsg(&req); err != nil {
+		return fmt.Errorf("read request from peer %v: %w", p.Address, err)
+	}
+
+	s.logger.Tracef("received payment message from peer %v of %d", p.Address, req.Amount)
```
Eknir, line 71: Why is this `trace`?

I think this should be on the `info` or at least `debug` level for sure!

The reason is that receiving a payment is something relevant to the user :)

pkg/settlement/pseudosettle/pseudosettle.go
Eknir: Missing comment as well

cmd/bee/cmd/start.go
```diff
@@ -194,7 +196,8 @@ Welcome to the Swarm.... Bzzz Bzzzz Bzzzz
 	cmd.Flags().String(optionNameTracingServiceName, "bee", "service name identifier for tracing")
 	cmd.Flags().String(optionNameVerbosity, "info", "log verbosity level 0=silent, 1=error, 2=warn, 3=info, 4=debug, 5=trace")
 	cmd.Flags().String(optionWelcomeMessage, "", "send a welcome message string during handshakes")
-	cmd.Flags().Uint64(optionNameDisconnectThreshold, 100000000000, "threshold in BZZ until which you allow peers to be indebted before disconnecting")
+	cmd.Flags().Uint64(optionNamePaymentThreshold, 100000, "threshold in BZZ where you expect to get paid from your peers")
+	cmd.Flags().Uint64(optionNamePaymentTolerance, 10000, "excess debt above payment threshold in BZZ where you disconnect from your peer")
```
Eknir, line 200: Suggest to make comment that this should be maximally 1/2 * PaymentThreshold. Or add a hard check elsewhere

pkg/accounting/accounting.go
```diff
@@ -216,6 +261,45 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+	if err != nil {
+		return err
+	}
+
+	balance.lock.Lock()
+	defer balance.lock.Unlock()
+
+	nextBalance := balance.balance - int64(amount)
+
+	// don't allow a payment to put use more into debt than the tolerance
```
Eknir, line 290: Perhaps it might make sense to describe why we do this. I.e. the attack scenario

pkg/accounting/accounting.go
```diff
@@ -216,6 +261,45 @@ func (a *Accounting) getPeerBalance(peer swarm.Address) (*PeerBalance, error) {
...
+	if nextBalance < -int64(a.paymentTolerance) {
+		return fmt.Errorf("refusing to accept payment which would put us too much in debt, new balance would have been %d", nextBalance)
```
Eknir, line 292: How is this error handled higher-up in the stack? The desired behavior is a disconnect from this peer--and never reconnect again.

pkg/settlement/pseudosettle/pseudosettle.go
ralph-pichler: fixed

pkg/accounting/accounting.go
ralph-pichler: fixed

pkg/accounting/accounting.go
ralph-pichler: do we really need to disconnect here?

cmd/bee/cmd/start.go
ralph-pichler: Added a check during startup

pkg/accounting/accounting.go
```diff
@@ -126,19 +131,59 @@ func (a *Accounting) Credit(peer swarm.Address, price uint64) error {
...
+	paymentAmount := uint64(-balance.balance)
+
+	oldBalance := balance.balance
+	nextBalance := oldBalance + int64(paymentAmount)
+
+	// try to save the next balance first
+	// otherwise we might pay and then not be able to save, thus paying again after restart
+	err := a.store.Put(balanceKey(peer), nextBalance)
+	if err != nil {
+		a.logger.Errorf("failed to persist balance for peer %v: %w", peer, err)
```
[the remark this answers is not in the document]
ralph-pichler: I extracted the saving into a function where it now generates an error.

pkg/accounting/accounting.go
```diff
@@ -126,19 +131,59 @@ func (a *Accounting) Credit(peer swarm.Address, price uint64) error {
...
+		return
+	}
+
+	err = a.settlement.Pay(context.Background(), peer, paymentAmount)
+	if err != nil {
+		a.logger.Errorf("payment for peer %v with amount %d failed: %v", peer, paymentAmount, err)
+		// if the payment didn't work we should restore the old balance in the state store
+		err = a.store.Put(balanceKey(peer), oldBalance)
+		if err != nil {
+			a.logger.Errorf("failed to persist balance for peer %v: %w", peer, err)
```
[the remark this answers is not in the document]
ralph-pichler: same as above

pkg/accounting/accounting.go
ralph-pichler: I changed the error handling a bit. This now returns an error. I also removed the peer from all the other error messages in accounting.

pkg/accounting/accounting.go
ralph-pichler: comment added

pkg/settlement/pseudosettle/pseudosettle.go
ralph-pichler: `info` seems to be way too spammy. during high load with one peer payment might be quite frequent.

cmd/bee/cmd/start.go
ralph-pichler: @janos not sure if this is the right place in the code for this kind of check: https://github.com/ethersphere/bee/pull/465/files#diff-dabbd93cd3d162371b93260a9f99c90cR103. What do you think? Should this be here, in `NewBee` or somewhere else?

pkg/accounting/accounting.go
Eknir: Don't forget this comment @ralph-pichler

pkg/settlement/pseudosettle/pseudosettle.go
Eknir: If it is a small fix, I would suggest just to do it right away. If it is a bigger task, it would be good to create an issue (`cleanup`).

pkg/accounting/accounting.go
Eknir: I don't see the comment. On which line?

pkg/accounting/accounting.go
Eknir: Not needed to disconnect, indeed. 
But still the question: how is this error handled higher up in the stack? Does the peer get notified somehow that you refused the payment?

pkg/accounting/accounting.go
ralph-pichler: the line directly below. 311.

pkg/accounting/accounting.go
ralph-pichler: AFAIK as long as its not a disconnect error it just gets logged inside the p2p code (settlement is returning this in its handler).

cmd/bee/cmd/start.go
janos: Sorry for a late response. I think that this is a logic that belongs to the accounting. Since both values are passed to the accounting constructor it should be validated there. It would contain the accounting logic into one package, not to leak to cmd.

pkg/accounting/accounting.go
```diff
@@ -114,7 +120,7 @@ func (a *Accounting) Release(peer swarm.Address, price uint64) {
 
 	if price > balance.reserved {
 		// If Reserve and Release calls are always paired this should never happen
-		a.logger.Error("attempting to release more balance than was reserved for peer")
+		a.logger.Errorf("attempting to release more balance than was reserved for peer")
```
janos, line 123: Errorf is not needed, just Error.

pkg/accounting/accounting.go
```diff
@@ -135,22 +141,70 @@ func (a *Accounting) Credit(peer swarm.Address, price uint64) error {
...
 
 	return nil
 }
 
+func (a *Accounting) saveBalance(peer swarm.Address, nextBalance int64) error {
+	err := a.store.Put(peerBalanceKey(peer), nextBalance)
+	if err != nil {
+		return fmt.Errorf("failed to persist balance: %w", err)
+	}
+	return nil
```
janos, line 172: I do not see the need to annotate this error. It would be just fine to return it without the need for the if statement. I would explicitly annotate errors in places where the function is called instead implicitly here.

pkg/accounting/accounting.go
ralph-pichler: fixed

pkg/accounting/accounting.go
ralph-pichler: I dropped this function entirely and moved the errorf to where put is called in the individual functions.

cmd/bee/cmd/start.go
ralph-pichler: I moved this check into the accounting constructor which can now return an error. I also added a test for this as this is now testable.

pkg/accounting/accounting.go
ralph-pichler: As discussed let's address this in a separate PR. I made an issue #542 for this.

Issue #441, opened 2020-07-14
Global Pinning Phase 1


santicomp2014:
Tasks for global pinning, port from Swarm to BEE
This phase consists

## Tasks
- [ ]  Mine and send to pinners
- [ ]  Identification of chunk as trojan and sending to chunk repair module
- [ ]  decode trojan and repair the chunk by sending the missing chunk to n/w
- [ ]  parallelize minning for trojan chunk
- [x]  integrate PSS into BEE

## PRs
| # | title  | status  |
|---|---|---|
| #466 | pss port from Swarm | **merged** |

Eknir:
Is this done @santicomp2014 ? If so: could you update Zenhub? If not, could you mention what needs to be done? Generally, would appreciate updated issues, given that you guys have achieved a good milestone!

PR #466, opened 2020-07-22
trojan, pss, pss_test: pss port from Swarm

santicomp2014:
This PR is the first step to import PSS for global pinning
The initial phase is to integrate this package into BEE, the next step is to define it's boundaries appropriately for the BEE architecture.

This is related to global pinning tasks #441


pkg/pss/pss.go
```diff
@@ -0,0 +1,134 @@
...
+			time.Sleep(100 * time.Millisecond)
+		}
+	}
+}
+
+// Register allows the definition of a Handler func for a specific topic on the pss struct
+func (p *Pss) Register(topic trojan.Topic, hndlr Handler) {
+	p.handlersMu.Lock()
+	defer p.handlersMu.Unlock()
+	p.handlers[topic] = hndlr
```
nolash, line 103: I would put this handler map in `trojan` package.

pkg/pss/pss.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// Monitor is used for tracking status changes in sent trojan chunks
+type Monitor struct {
+	// returns the state of the trojan chunk that is being monitored
+	State chan tags.State
+}
+
+// NewPss inits the Pss struct with the localstore
+func NewPss(localStore storage.Storer, tags *tags.Tags) *Pss {
```
nolash, line 35: the name suggests `localStore` but type is store interface. If it can be only localstore, then the type should be specific.

pkg/pss/pss.go
```diff
@@ -0,0 +1,134 @@
...
+	go monitor.updateState(tag)
+
+	return monitor, nil
+}
+
+// updateState sends the change of state thru the State channel
+// this is what enables monitoring the trojan chunk after it's sent
+func (m *Monitor) updateState(tag *tags.Tag) {
+	for _, state := range []tags.State{tags.StateStored, tags.StateSent, tags.StateSynced} {
+		for {
```
nolash, line 88: this seems a bit un-go to me; wouldn't more go be count sends on a channel?

pkg/pss/pss_test.go
```diff
@@ -0,0 +1,223 @@
...
+	var err error
+	ctx := context.TODO()
+	testTags := tags.NewTags()
+
+	localStore := mock.NewTagsStorer(testTags)
+	pss := pss.NewPss(localStore, testTags)
+
+	target := trojan.Target([]byte{1}) // arbitrary test target
+	targets := trojan.Targets([]trojan.Target{target})
+	payload := []byte("RECOVERY CHUNK")
```
nolash, line 34: perhaps choose a more neutral string as value. pss is not a recovery-specific package

pkg/pss/pss_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+	// call Send to store trojan chunk in localstore
+	if _, err = pss.Send(ctx, targets, topic, payload); err != nil {
+		t.Fatal(err)
+	}
+
+	var chunkAddress swarm.Address
+	// this code iterates over the localstore
+	// this will get the chunk that was stored in pss.Send
+	for po := uint8(0); po <= swarm.MaxPO; po++ {
```
nolash, line 45: this looks rather clumsy, tbh. doesn't it make more sense to have an interface to get the addess from the trojan object instead?

pkg/pss/pss_test.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+	}
+
+	// verify store, that trojan chunk has been stored correctly
+	var storedChunk swarm.Chunk
+	if storedChunk, err = localStore.Get(ctx, storage.ModeGetRequest, chunkAddress); err != nil {
+		t.Fatal(err)
+	}
+
+	// create a stored chunk artificially
```
nolash, line 67: ... if you had that address check interface probably you wouldn't need this code.

pkg/pss/pss_test.go
```diff
@@ -0,0 +1,223 @@
...
+	if monitor, err = pss.Send(ctx, targets, topic, payload); err != nil {
+		t.Fatal(err)
+	}
+
+	storeTags := testTags.All()
+	if len(storeTags) != 1 {
+		t.Fatalf("expected %d tags got %d", 1, len(storeTags))
+	}
+
+	timeout := 1 * time.Second
```
nolash, line 120: may I suggest `context.WithDeadline` for this?

pkg/pss/pss_test.go
```diff
@@ -0,0 +1,223 @@
...
+	for _, expectedState := range []tags.State{tags.StateStored, tags.StateSent, tags.StateSynced} {
+		storeTags[0].Inc(expectedState)
+	loop:
+		for {
+			// waits until the monitor state has changed or timeouts
+			select {
+			case state := <-monitor.State:
+				if state == expectedState {
+					break loop
```
nolash, line 129: return?

pkg/pss/pss_test.go
```diff
@@ -0,0 +1,223 @@
...
+				}
+			case <-time.After(timeout):
+				t.Fatalf("no message received")
+			}
+		}
+	}
+}
+
+// TestRegister verifies that handler funcs are able to be registered correctly in pss
+func TestRegister(t *testing.T) {
```
nolash, line 139: Ill read this later depending on reply about handler above

pkg/pss/pss.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// GetHandler returns the Handler func registered in pss for the given topic
+func (p *Pss) GetHandler(topic trojan.Topic) Handler {
+	p.handlersMu.RLock()
+	defer p.handlersMu.RUnlock()
+	return p.handlers[topic]
+}
+
+// GetAllHandlers returns all the Handler funcs registered in pss
```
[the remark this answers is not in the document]
santicomp2014: Before in Swarm, the testing was made in the same package and handlers were visible from the test.
Since it's now in another package, there is no visibility.
I added this for testing capabilities.

pkg/pss/pss_test.go
```diff
@@ -0,0 +1,223 @@
...
+	// call Send to store trojan chunk in localstore
+	if _, err = pss.Send(ctx, targets, topic, payload); err != nil {
+		t.Fatal(err)
+	}
+
+	var chunkAddress swarm.Address
+	// this code iterates over the localstore
+	// this will get the chunk that was stored in pss.Send
+	for po := uint8(0); po <= swarm.MaxPO; po++ {
+		last, err := localStore.LastPullSubscriptionBinID(po)
```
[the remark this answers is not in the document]
santicomp2014: This was used for testing that the pss.Send was storing correctly the chunk for sending.
Is PullSubscription not needed in this implementation in BEE.

What should be the correct way to implement this?.

pkg/pss/pss.go
santicomp2014: But this would mean the trojan packages has the map of handlers, I think it's pss responsibility to have the list of handlers.
That being said maybe it's correct to move it, @zelig what do you think?.

pkg/pss/pss.go
santicomp2014: this is wrong, I agree I will change it. good point.

santicomp2014: fixed

pkg/pss/pss.go
santicomp2014: This is a bit contrived, This should be changed, do any examples?, so I can make it more idiomatic. thanks

pkg/pss/pss_test.go
santicomp2014: We were using this pss only for Global Pinning that's why it got hardcoded.
I will change it to a generic topic

santicomp2014: fixed

pkg/pss/pss_test.go
santicomp2014: This is related to @zelig [comment](https://github.com/ethersphere/bee/pull/466#discussion_r460360511)

pkg/pss/pss_test.go
santicomp2014: replaced

pkg/pss/pss_test.go
santicomp2014: This has to wait for all 3 states to be processed.
It loops back to the `loop` tag or until it times out

pkg/pss/pss.go
nolash: AFAIR we agreed that trojan would resolve the topics, pss would register handler code to that on the same level as eg. an external subscription service, that would know how to unwrap itself.

I remember now I promised to produce a diagram to outline the workflow. Sorrry I haven't had time for this yet. I will try to get it done tomorrow friday.

pkg/pss/pss.go
nolash: At first glance it would require changes to the tags package, so it would emit on a channel when a count is increased. I don't know the package very well, though. Perhaps @acud @jmozah have better perspective.

pkg/pss/pss_test.go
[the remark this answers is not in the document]
santicomp2014: This was removed since it now uses pushsync directly with it's mock

pkg/pss/pss_test.go
```diff
@@ -0,0 +1,256 @@
...
+	storage.Storer
+	modeSet   map[string]storage.ModeSet
+	modeSetMu *sync.Mutex
+}
+
+// TestTrojanChunkRetrieval creates a trojan chunk
+// mocks the localstore
+// calls pss.Send method and verifies it's properly stored
+func TestTrojanChunkRetrieval(t *testing.T) {
+	var err error
```
acud, line 39: you can make the test more readable by either creating a helper method that will bootstrap the dependencies:
`ctx, pss, target, targets(???), payload := newTestPss()` 
or
```go
var (
   ctx = context.Background()
   testTags = tags.NewTags()
   pushSyncService = pushsyncmock.New(...)
   etc...
)
```

pkg/pss/pss_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+	mtags := tags.NewTags()
+	pusherStorer := &Store{
+		Storer:    storer,
+		modeSet:   make(map[string]storage.ModeSet),
+		modeSetMu: &sync.Mutex{},
+	}
+	peerSuggester := mocktopology.NewTopologyDriver(mockOpts...)
+
+	pusherService := pusher.New(pusher.Options{Storer: pusherStorer, PushSyncer: pushSyncService, Tagger: mtags, PeerSuggester: peerSuggester, Logger: logger})
```
acud, line 254: why do you need a fully blown `pusher` here? wouldn't it be easier to mock it? that's the whole point of having interfaces

pkg/pss/pss.go
Eknir: Is this diagram already produced @nolash ? Would be of a great help.

pkg/pss/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+
+package pss
+
+import (
+	m "github.com/ethersphere/bee/pkg/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type metrics struct {
+	// all metrics fields must be exported
```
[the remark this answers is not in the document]
janos: I believe that this is a common comment, copied from other packages. It can be removed here, but also in other places, if not needed.

pkg/pss/pss.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Register allows the definition of a Handler func for a specific topic on the pss struct
+func (p *Pss) Register(topic trojan.Topic, hndlr Handler) {
+	p.handlersMu.Lock()
+	defer p.handlersMu.Unlock()
+	p.handlers[topic] = hndlr
+}
+
+// Deliver allows unwrapping a chunk as a trojan message and calling its handler func based on its topic
+func (p *Pss) Deliver(c swarm.Chunk) {
```
janos, line 88: I would suggest to add context to Deliver and Handler, and probably also to return error. Something noticed from the chunk repair pr.

pkg/pss/pss.go
santicomp2014: We should definitely do this, but for the time being we should merge and reiterated on the PSS and Trojan chunk boundaries.

pkg/pss/pss.go
santicomp2014: This was replaced with tags

pkg/pss/pss.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+}
+
+// Handler defines code to be executed upon reception of a trojan message
+type Handler func(trojan.Message)
+
+// Send constructs a padded message with topic and payload,
+// wraps it in a trojan chunk such that one of the targets is a prefix of the chunk address
+// uses push-sync to deliver message
+func (p *Pss) Send(ctx context.Context, targets trojan.Targets, topic trojan.Topic, payload []byte) (*tags.Tag, error) {
```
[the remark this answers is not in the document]
santicomp2014: Maybe we should refactor this in a new PR, so the boundaries are well defined.
This was also mentioned by @nolash [here](https://github.com/ethersphere/bee/pull/466#discussion_r460746977)

pkg/pss/pss.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var tc swarm.Chunk
+	tc, err = m.Wrap(targets)
+	if err != nil {
+		return nil, err
+	}
+
+	tag, err := p.tags.Create("pss-chunks-tag", 1, false)
```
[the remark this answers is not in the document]
santicomp2014: We should review this in the PSS refactoring, which I think is not going to be critical for global pinning as of now.

pkg/pss/pss_test.go
santicomp2014: The mock is not created for pushSync and also it was not necessary for my test.
I removed it.
Thanks

pkg/pss/pss.go
```diff
@@ -0,0 +1,134 @@
...
+func (p *Pss) Register(topic trojan.Topic, hndlr Handler) {
+	p.handlersMu.Lock()
+	defer p.handlersMu.Unlock()
+	p.handlers[topic] = hndlr
+}
+
+// Deliver allows unwrapping a chunk as a trojan message and calling its handler func based on its topic
+func (p *Pss) Deliver(c swarm.Chunk) {
+	if trojan.IsPotential(c) {
+		m, _ := trojan.Unwrap(c) // if err occurs unwrapping, there will be no handler
```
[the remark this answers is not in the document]
santicomp2014: Fixed

janos:
Linter is reporting some warnings https://github.com/ethersphere/bee/pull/466/checks?check_run_id=971858621#step:8:10.

pkg/pss/pss_test.go
```diff
@@ -85,70 +69,70 @@ func TestTrojanChunkRetrieval(t *testing.T) {
...
-			case state := <-monitor.State:
-				if state == expectedState {
-					break loop
-				}
-			case <-time.After(timeout):
-				t.Fatalf("no message received")
-			}
-		}
+	if tag.Get(tags.StateStored) != 1 && tag.Get(tags.StateSent) != 1 && tag.Get(tags.StateSynced) != 1 {
+		t.Fatalf("Trojan Chunk expected to be Stored == %d, Sent == %d and Synced == %d", tag.Stored, tag.Sent, tag.Synced)
```
nolash, line 127: why would trojan chunk be _stored_?

pkg/pss/pss.go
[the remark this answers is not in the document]
acud: @santicomp2014 if `PushChunkToClosest` returns without an error then the operation succeeded. all of the state you need and know about the chunk is already in this function. pushsync will not retry pushing the chunk on error so the tag does not bring any added value, i.e. even if you continue polling the tag in another context with the hopes of the chunk being eventually synced, the state of it wont change. this can be removed

pkg/pss/pss_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+	if !reflect.DeepEqual(tc, storedChunk) {
+		t.Fatalf("trojan chunk created does not match sent chunk. got %s, want %s", storedChunk.Address().ByteString(), tc.Address().ByteString())
+	}
+}
+
+// TestPssTags creates a trojan chunk
+// mocks the localstore
+// calls pss.Send method
+// updates the tag state (Stored/Sent/Synced)
```
zelig, line 85: stored will not be incremented if  we  use the pushsync protocol directly

pkg/pss/pss.go
```diff
@@ -0,0 +1,115 @@
...
+func (p *pss) TryUnwrap(ctx context.Context, c swarm.Chunk) error {
+	if trojan.IsPotential(c) {
+		m, err := trojan.Unwrap(c) // if err occurs unwrapping, there will be no handler
+		if err != nil {
+			return err
+		}
+		h := p.GetHandler(m.Topic)
+		if h != nil {
+			p.logger.Trace("executing handler for trojan process global-pinning chunk", c.Address().ByteString())
+			h(*m)
```
acud, line 103: not sure why a pointer dereferencing is needed here, this is just causing more allocations. why can't you pass a pointer in the signature?

pkg/pss/pss.go
```diff
@@ -0,0 +1,115 @@
...
+			return nil
+		}
+	}
+	return fmt.Errorf("chunk not trojan or no handler found process global-pinning chunk %v", c.Address().ByteString())
```
acud, line 107: this message is very long, plus, it mentions global pinning. seems misplaced. also, it seems that this function will return an error for everything which is not a content addressed chunk (whatever `!IsPotential()`). i suggest to remove the `ispotential` check and just try to unwrap everything

pkg/pss/pss.go
```diff
@@ -0,0 +1,115 @@
...
+
+// TryUnwrap allows unwrapping a chunk as a trojan message and calling its handler func based on its topic
+func (p *pss) TryUnwrap(ctx context.Context, c swarm.Chunk) error {
+	if trojan.IsPotential(c) {
+		m, err := trojan.Unwrap(c) // if err occurs unwrapping, there will be no handler
+		if err != nil {
+			return err
+		}
+		h := p.GetHandler(m.Topic)
+		if h != nil {
```
acud, line 101: this should be inverted, please check that `h == nil` and return the error from L107 inside the block. Also, try to keep happy-path code with minimum nesting. Here the happy path ends in the code branch in this block

pkg/storage/mock/storer.go
```diff
@@ -41,6 +43,18 @@ func WithSubscribePullChunks(chs ...storage.Descriptor) Option {
 	})
 }
 
+func WithBaseAddress(a swarm.Address) Option {
```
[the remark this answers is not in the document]
santicomp2014: This was removed, This is not necessary anymore since we use the push sync mock, this was code to emulate what we had in Swarm.
But It was changed accordingly,
Thanks for the correction.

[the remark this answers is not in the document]
santicomp2014: If somebody needs this, we can change it in another PR.

pkg/pss/pss.go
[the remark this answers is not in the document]
santicomp2014: I see your point, so `PushChunkToClosest `  would block until message is sent, and if error arrives then it can be considered a failure and somebody up the stack can try a retry logic or not.

About tags, I thought the idea of the tag was to view the status of the chunk in its journey.
Is the tag not Sent and Synced currently?

This is an important question, I could remove tags altogether, but I would like @zelig and @jmozah opinion on this, before removing this from PSS.

pkg/pss/pss_test.go
santicomp2014: This was a wrong assumption from when we ported.
Corrected, good point

pkg/pss/pss.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// push the chunk using push sync so that it reaches it destination in network
+	if _, err = p.pusher.PushChunkToClosest(ctx, tc.WithTagID(tag.Uid)); err != nil {
+		return nil, err
+	}
+
+	tag.Total = 1
```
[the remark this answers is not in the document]
santicomp2014: huh, that's odd, we missed this in Swarm also.
No issue removed.

pkg/pss/pss_test.go
santicomp2014: fixed

pkg/pss/pss.go
santicomp2014: The message is long and Global Pinning was here because it was the only use case.
This was inherited from Swarm, I will change this message now.

`IsPotential` does validation that the chunk is content addresses and is valid.
If we try to unwrap everything it will be done multiple times on chunk that are not Trojan Chunks.

We had this issue in Swarm and we fixed it with @mortelli with this implementation.
Maybe there is another way in BEE and we should remove this, but for the time being, I would insist on keeping `IsPotential`

pkg/pss/pss.go
santicomp2014: `why can't you pass a pointer in the signature`

```
// GetHandler returns the Handler func registered in pss for the given topic
func (p *pss) GetHandler(topic trojan.Topic) Handler {
	p.handlersMu.RLock()
	defer p.handlersMu.RUnlock()
	return p.handlers[topic]
}
```

This uses type Handler func(trojan.Message) could I change the method to be a pointer func?, Maybe I did not understand,
Thanks

PR #467, opened 2020-07-23
Accounting metrics

metacertain:
The aim of this pull request is to add accounting related metrics to the metrics endpoint
- Total debited amount 
- Total credited amount
- Total credit events
- Total debit events
- Total accounting related disconnects
- Total accounting related reserve blocks ( when temporarily skipping a peer for the safety of not going through thresholds before we could settle)


pkg/accounting/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+
+func newMetrics() metrics {
+	subsystem := "accounting"
+
+	return metrics{
+		DebitCount: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "total_debited_amount",
+			Help:      "Amount of bzz debited to peers (potential income of the node)",
```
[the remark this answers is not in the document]
ralph-pichler: Not sure if this has been settled but in the command line description in merged accounting PR I used `BZZ` (capital letters).

pkg/accounting/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+	CreditEvents      prometheus.Counter
+	CreditDisconnects prometheus.Counter
+	CreditBlocks      prometheus.Counter
+}
+
+func newMetrics() metrics {
+	subsystem := "accounting"
+
+	return metrics{
+		DebitCount: prometheus.NewCounter(prometheus.CounterOpts{
```
ralph-pichler, line 28: The naming seems inconsistent. The field is called is called Debit**Count**  but the metric is total_debited_**amount**.  Further down there is actually a total_debit_events_**count** metric.

`DebitEvent` should probably be `DebitCount` and `DebitCount` `DebitAmount`

pkg/accounting/metrics.go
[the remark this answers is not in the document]
paxthemax: @mortelli AFAIK - this is supposed to be the Buzz token, and **BZZ** is the ticker name.

pkg/accounting/accounting.go
```diff
@@ -162,9 +168,13 @@ func (a *Accounting) Debit(peer swarm.Address, price uint64) error {
 
 	if nextBalance >= int64(a.disconnectThreshold) {
 		// peer to much in debt
+		a.metrics.CreditDisconnects.Inc()
 		return p2p.NewDisconnectError(fmt.Errorf("disconnect threshold exceeded for peer %s", peer.String()))
 	}
 
+	a.metrics.DebitCount.Add(float64(price))
+	a.metrics.DebitEvents.Inc()
```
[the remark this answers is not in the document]
paxthemax: @mortelli I agree, and I've updated to reflect the change.

pkg/accounting/accounting.go
```diff
@@ -162,9 +168,13 @@ func (a *Accounting) Debit(peer swarm.Address, price uint64) error {
 
 	if nextBalance >= int64(a.disconnectThreshold) {
 		// peer to much in debt
+		a.metrics.CreditDisconnects.Inc()
```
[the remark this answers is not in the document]
paxthemax: @mortelli So the most important fact is that the relevant counter names are `credit_related_disconnects_count` and `credit_related_blocks_count` so let's make sure that the names of the counters match the actual data written to the Prometheus stream.

I think the credit term is related no to a credit balance, but to the concept of credit. But same could be said to accounting so it's okay either way.

pkg/accounting/metrics.go
paxthemax: @ralph-pichler Agreed, we need a better naming scheme.

pkg/accounting/accounting.go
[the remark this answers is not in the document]
mortelli: fully agree on your point that the names should match, i'm just not so sure about this name in particular.

anyway for me this is not a blocker, so i am open to hearing other opinions on this from the rest of the team.

pkg/accounting/metrics.go
paxthemax: I would propose to rename the counters/metrics in the following way:

- `TotalDebitedAmount` to track `total_debited_amount`
- `TotalCreditedAmount` to track `total_credited_amount`
- `DebitEventsCount` to track `debit_events_count`
- `CreditEventsCount` to track `credit_events_count`
- `AccountingDisconnectsCount` to track `accounting_disconnects_count`
- `AccointingBlocksCount` to track `accounting_blocks_count`

paxthemax:
Unrelated to the PR: we do not have **unit** tests for metrics. I assume they are not a priority for the PR - are they required? If so, is there a good way to write unit tests for Prometheus Counters, Gauges and other Metrics collectors?

pkg/accounting/metrics.go
ralph-pichler: Sounds good to me

pkg/accounting/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type metrics struct {
+	// all metrics fields must be exported
+	// to be able to return them by Metrics()
+	// using reflection
+	DebitCount        prometheus.Counter
+	CreditCount       prometheus.Counter
+	DebitEvents       prometheus.Counter
```
acud, line 18: i'm pretty sure that we don't need these explicitly. this can probably be set up in grafana to display the counter of how many times the debitCount was written to. cc @vandot

vandot: @acud probably not, Counter holds current value and prometheus scrapes that value every 10s
if you need how many times something happened you have to expose separate metric or for everything you will have 6 occurrences per minute (number of times prometheus scrapes metric)

acud: cool. thanks

pkg/accounting/accounting.go
```diff
@@ -84,6 +86,7 @@ func (a *Accounting) Reserve(peer swarm.Address, price uint64) error {
 	// since we pay this we have to reduce this (positive quantity) from the balance
 	// the disconnectThreshold is stored as a positive value which is why it must be negated prior to comparison
 	if balance.freeBalance()-int64(price) < -int64(a.disconnectThreshold) {
+		a.metrics.AccointingBlocksCount.Inc()
```
ralph-pichler, line 89: Accointing

Issue #471, opened 2020-07-24
rectify swarm-encrypt returns an error


significance:
# Task

there is a bug when using the swarm-encrypt header, the command as follows:

`sig :: ~/Desktop » curl -H "swarm-encrypt: true" -F file=@Spinach-and-Chick-Pea-e1434488129314.jpg http://localhost:8081/files`

has response

`{"message":"could not store file data","code":500}`

## Acceptance criterea

a reference should be returned

jmozah:
https://github.com/ethersphere/bee/pull/472

j1n6:
I was testing if the header would encrypt the file, both local and public gateway URL allow seeing the file without decryption. 

@jmozah how would you expect encrypted file to behave when replicated to the public gateway?

Eknir:
@j1n6 , if you use the `swarm-encrypt` header, the data will be encrypted before sent to swarm. You get back a reference which contains the reference to the data + the decryption key. If you pass this reference, indeed, our public gateways will decrypt automatically with the decryption key, as provided. 

Let me know if this answers your question.

j1n6:
@Eknir is there anyway to protect the data only allowing to be decrypted at local node to protect sensitive information? or would you recommend a different/better way to achieve this objective?

Eknir:
@j1n6 , it would be best to continue this conversation at https://beehive.ethswarm.org/ . You can find me there as @Eknir

PR #472, opened 2020-07-24
Fix file upload/download during encryption

jmozah:
If a file is uploaded with encryption flag set as true, bee does not upload the file, instead it returns a http 500 error.
This is because in upload path `validation` and in download path the entry `un-marshalling ` does not take care of hash size properly

This was not caught earlier because there was an error in the test case. Instead of adding api.EncryptHeader in the header,  "Header" string was added... so in essence encryption was not at all triggered and so the test case was passing

Fix for https://github.com/ethersphere/bee/issues/471


pkg/api/file_test.go
```diff
@@ -51,16 +52,18 @@ func TestFiles(t *testing.T) {
...
 
-		_ = jsonhttptest.ResponseDirectSendHeadersAndReceiveHeaders(t, client, http.MethodPost, fileUploadResource+"?name="+fileName, bytes.NewReader(simpleData), http.StatusOK, api.FileUploadResponse{
-			Reference: swarm.MustParseHexAddress(rootHash),
-		}, headers)
+		_, respBytes := jsonhttptest.ResponseDirectSendHeadersAndDontCheckResponse(t, client, http.MethodPost, fileUploadResource+"?name="+fileName, bytes.NewReader(simpleData), http.StatusOK, headers)
+		read := bytes.NewReader(respBytes)
+
+		// get the reference as everytime it will change because of random encryption key
+		var resp api.FileUploadResponse
+		json.NewDecoder(read).Decode(&resp)
```
mortelli, line 64: please do not ignore the `err` that `Decode` returns

pkg/api/file_test.go
```diff
@@ -51,16 +52,18 @@ func TestFiles(t *testing.T) {
...
 
-		rcvdHeader := jsonhttptest.ResponseDirectCheckBinaryResponse(t, client, http.MethodGet, fileDownloadResource(rootHash), nil, http.StatusOK, simpleData, nil)
+		rcvdHeader := jsonhttptest.ResponseDirectCheckBinaryResponse(t, client, http.MethodGet, fileDownloadResource(resp.Reference.String()), nil, http.StatusOK, simpleData, nil)
```
mortelli, line 66: i think you should extract `resp.Reference.String()` into `rootHash` and leave this as it was, it's more readable that way imo

pkg/file/joiner/joiner.go
```diff
@@ -42,7 +54,17 @@ func (s *simpleJoiner) Size(ctx context.Context, address swarm.Address) (dataSiz
...
+	var chunkData []byte
+	if toDecrypt {
+		originalData, err := internal.DecryptChunkData(rootChunk.Data(), key)
+		if err != nil {
+			return 0, err
+		}
+		chunkData = originalData
+	} else {
+		chunkData = rootChunk.Data()
+	}
```
mortelli, line 66:
```suggestion
	chunkData := rootChunk.Data()
	if toDecrypt {
		decryptedData, err := internal.DecryptChunkData(chunkData, key)
		if err != nil {
			return 0, err
		}
		chunkData = decryptedData
	}
```

ftr: i would also prefer to have it this way in `Join`, but it is not a blocker.

pkg/file/joiner/joiner.go
```diff
@@ -31,6 +31,18 @@ func NewSimpleJoiner(getter storage.Getter) file.Joiner {
 }
 
 func (s *simpleJoiner) Size(ctx context.Context, address swarm.Address) (dataSize int64, err error) {
+	toDecrypt := len(address.Bytes()) == (swarm.HashSize + encryption.KeyLength)
```
mortelli, line 34: should `toDecrypt` be a parameter?

also, isn't the right side of the `==` actually `EncryptedHashSize`?

in any case, a comment here would be nice

pkg/file/joiner/joiner.go
```diff
@@ -31,6 +31,18 @@ func NewSimpleJoiner(getter storage.Getter) file.Joiner {
...
+
+	var addr []byte
```
mortelli, line 36: we need a different name here. we already have the `address` param, having `addr` on top of this makes things confusing.

i think line `45` also supports this change

pkg/file/joiner/joiner.go
```diff
@@ -31,6 +31,18 @@ func NewSimpleJoiner(getter storage.Getter) file.Joiner {
...
+	var key encryption.Key
+	if toDecrypt {
+		addr = address.Bytes()[:swarm.HashSize]
+		key = address.Bytes()[swarm.HashSize : swarm.HashSize+encryption.KeyLength]
+	} else {
+		addr = address.Bytes()
+	}
```
mortelli, line 43: as in my previous comment, here i would define default assignments (no encryption) and then add a following `if` without `else`.

i think this makes things more readable, but it is not a blocker

pkg/collection/entry/entry.go
```diff
@@ -50,11 +51,17 @@ func (e *Entry) MarshalBinary() ([]byte, error) {
 }
 
 // UnmarshalBinary implements encoding.BinaryUnmarshaler
-func (e *Entry) UnmarshalBinary(b []byte) error {
-	if len(b) != serializedDataSize {
+func (e *Entry) UnmarshalBinary(b []byte, toDecrypt bool) error {
```
zelig, line 54: `UnmarshalBinary implements encoding.BinaryUnmarshaler`

not if u introduce an argument.
i think it is ok to pass the `toDecrypt` flag within the entry  (as an unserialised or even serialised field).

pkg/jsonhttp/jsonhttptest/jsonhttptest.go
```diff
@@ -137,6 +137,21 @@ func ResponseDirectSendHeadersAndReceiveHeaders(t *testing.T, client *http.Clien
 	return resp.Header
 }
 
+func ResponseDirectSendHeadersAndDontCheckResponse(t *testing.T, client *http.Client, method, url string, body io.Reader, responseCode int, headers http.Header) (http.Header, []byte) {
```
zelig, line 140: maybe a comment would be nice, although this mouthful of a name is almost self commenting

pkg/swarm/swarm.go
```diff
@@ -20,6 +21,7 @@ const (
 	Branches                = 128
 	ChunkSize               = SectionSize * Branches
 	HashSize                = 32
+	EncryptedHashSize       = HashSize + encryption.KeyLength
```
zelig, line 24: no, please call it 'encrypted reference' size

pkg/collection/entry/entry.go
```diff
@@ -12,8 +12,9 @@ import (
 )
 
 var (
-	_                  = collection.Entry(&Entry{})
-	serializedDataSize = swarm.SectionSize * 2
+	_                           = collection.Entry(&Entry{})
+	serializedDataSize          = swarm.SectionSize * 2
```
zelig, line 16: what is this meant to be?

pkg/collection/entry/entry.go
```diff
@@ -12,8 +12,9 @@ import (
...
+	encryptedSerializedDataSize = swarm.EncryptedHashSize * 2
```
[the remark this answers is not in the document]
jmozah: There are two references stored in an entry now. One for metadata and the other for the data. Thats why it is multiplied by 2. 

There is swarm.EncryptedHashSize which is the size of hash for a encrypted data.

pkg/jsonhttp/jsonhttptest/jsonhttptest.go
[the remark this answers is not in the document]
jmozah: yeah. I agree.

jmozah:
> one important question, though: t.Run("encrypt-decrypt") was definitely not failing until now.

@mortelli There was an error in the test case. Instead of adding api.EncryptHeader in the header, i added "Header" string... so in essence encryption was not at all triggered and so the test case was passing.

pkg/collection/entry/entry.go
jmozah: Entry stores two references, one for metadata and other for the actual data. so it is multiplied by 2.  But for Encrypted file, the references will be 64 ... that was added as a new constant.

jmozah:
> i haven't left a specific comment in job.go, but i don't really understand the reason for those changes in that part of the diff. some clarification here would be much appreciated 🙏

@mortelli That is the place where the chunk is actually encrypted. If toEncrypt is set, then the actual data is encrypted and then that is hashed to find the corresponding content address.

pkg/file/joiner/joiner.go
jmozah: toDecrypt need not be a parameter since we can find that out by seeing the length of the address itself. 

Also i added some comments

pkg/collection/entry/entry.go
jmozah: Removed the flag so that it remains like the old code.

pkg/jsonhttp/jsonhttptest/jsonhttptest.go
jmozah: :-)  Added some comments

acud:
it would be nice to have a problem statement because i can't seem to find one, not in the issue reporting the bug nor in the PR which supposedly fixes it

jmozah:
> it would be nice to have a problem statement because i can't seem to find one, not in the issue reporting the bug nor in the PR which supposedly fixes it

@acud I think #471 is super clear and explains that it is not able to upload a file when encryption is enabled. It also states the received error and expected result. And this PR fixes the issue. Also the [reply](https://github.com/ethersphere/bee/pull/472#issuecomment-663754805) to @mortelli comment indicates the reason why it was not caught earlier.

acud:
@jmozah the issue you're pointing at describes how to reproduce the error. it does not mention on what is wrong in the codebase that actually causes this problem, therefore, it is not clear to the naive reviewer on what is exactly the underlying problem _is_ here. also, the reply you're referring me to has to do with a test case, not with production code which is affected by the original issue, or am I getting something wrong here?

pkg/jsonhttp/jsonhttptest/jsonhttptest.go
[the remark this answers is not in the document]
mortelli: #477

pkg/file/joiner/joiner.go
```diff
@@ -31,6 +31,17 @@ func NewSimpleJoiner(getter storage.Getter) file.Joiner {
 }
 
 func (s *simpleJoiner) Size(ctx context.Context, address swarm.Address) (dataSize int64, err error) {
+	// Handle size based on weather the root chunk is encrypted or not
```
mortelli, line 34: typo in `weather` (should be `whether`)

jmozah:
> it does not mention on what is wrong in the codebase that actually causes this problem

@acud @mortelli  ok ok... I added the reason in PR starting. I hope everyone adds comment like this when issuing a PR.

Eknir:
Hi @jmozah , I connected the PR to the issue (Zenhub feature). Please do so in the future. Thanks for taking care of this!

j1n6:
is this released?

Eknir:
Our [last release](https://github.com/ethersphere/bee/releases) was on June 29. This feature got merged after that. You can try out the feature on the latest master. We will make an official release (include an update of the documentation) upcoming Monday.

j1n6:
Thank you for confirming the release status - saved me time to debug why “swarm-encrypt” header didn’t work in the 0.1.0 release.

Eknir:
@j1n6 , you could join our public testing session today at 1300 CET on meet.jit.si/SwarmDev

Issue #440, opened 2020-07-14
Global Pinning Phase 0


santicomp2014:
Tasks for global pinning, port from Swarm to BEE
Initial porting process, barebones for this phase 0

## Tasks
- [x]  Trojan chunk implementation porting as a standalone package in bee
- [x]  take targets as arguments in download path
- [x]  logic to detect chunk as missing
- [x]  send a trojan chunk 
- [x]  inspect Bee project to find possible integration places
- [x]  create integration test issue for @svetomir
- [x]  plan with @zelig and @nolash PSS porting

## PRs
| # | title  | status  |
|---|---|---|
| #429 | Trojan chunk support | **closed** |
| #430 | Add targets to query param on download path | **closed** |
| #466 | PSS | **closed** |
| #479 | Chunk repair | **closed** |
| #556 | remove chunk debug API | **closed** |
| #453 | Integration test | **issue created** |

Eknir:
Hi guys, what is the status of this one?

jmozah:
It is done. We just have to merge. I need to check if i need to wait and merge in master or merge it right no in global-pinin-pss branch itself.

acud:
looks like done

PR #479, opened 2020-07-27
Chunk repair

jmozah:
This PR is part of global pinning. This does the following
- Detect if a chunk is not present in the network
- Send a chunk recovery message to the targets (get targets while download and store in context)
- Targets receive the message from PSS ( register while booting up) 
- Repair the chunk by uploading it to the network

builds on #466


Eknir:
Hi @jmozah . Great to see this PR!

I am doing the following things now. Please do it next time (it would have saved you and me some time to figure out what is the status of Global Pinning and allows us to properly estimate our velocity as a team).

- Connect the PR with an existing issue (I assume #440, but maybe also #441?), using Zenhubs feature `Connect this pull request with an existing issue`
- Set the issue under the global pinning epic
- Set the issue under the persistent websites release

pkg/chunk/repair.go
```diff
@@ -0,0 +1,95 @@
...
+			logger.Debugf("chunk repair: error while sending chunk or receiving receipt: %v", err)
+			return
+		}
+	}
+}
+
+// getPinners returns the specific target pinners for a corresponding chunk by
+// reading the prefix targets sent in the download API
+func getPinners(ctx context.Context) (trojan.Targets, error) {
+	targetString := ctx.Value(api.TargetsContextKey{}).(string)
```
mortelli, line 81: we're extracting the targets twice by this point, not sure there's a need for this

pkg/chunk/repair.go
```diff
@@ -0,0 +1,95 @@
...
+	prefixes := strings.Split(targetString, ",")
+
+	var targets trojan.Targets
+	for _, prefix := range prefixes {
+		var target trojan.Target
+		prefix = strings.TrimPrefix(prefix, "0x")
```
mortelli, line 87: would this also work if the prefix is `0X`?

pkg/chunk/repair.go
```diff
@@ -0,0 +1,95 @@
...
+		target, err := hex.DecodeString(prefix)
+		if err != nil {
+			return nil, ErrTargetPrefix
```
mortelli, line 90: an alternative here is: skip the targets which cannot be decoded, and return the err if none can be decoded.

pkg/chunk/repair.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		// push the chunk using push sync so that it reaches it destination in network
+		_, err = pushSyncer.PushChunkToClosest(ctx, ch)
+		if err != nil {
+			logger.Debugf("chunk repair: error while sending chunk or receiving receipt: %v", err)
+			return
+		}
+	}
+}
```
mortelli, line 76: what would happen if after calling `s.Get` the chunk is garbage collected due to not being locally pinned?

pkg/chunk/repair_test.go
```diff
@@ -0,0 +1,296 @@
...
+		t.Fatal(err)
+	}
+
+	if hookWasCalled != true {
+		t.Fatalf("recovery hook was not called")
+	}
+}
+
+// RecoveryHookTestCase is a struct used as test cases for the TestRecoveryHookCalls func
+type RecoveryHookTestCase struct {
```
mortelli, line 59: this one does not need to be exported

pkg/chunk/repair_test.go
```diff
@@ -0,0 +1,296 @@
...
+type mockPeerSuggester struct {
+	eachPeerRevFunc func(f topology.EachPeerFunc) error
+}
+
+func (s mockPeerSuggester) EachPeer(topology.EachPeerFunc) error {
+	return errors.New("not implemented")
+}
+func (s mockPeerSuggester) EachPeerRev(f topology.EachPeerFunc) error {
+	return s.eachPeerRevFunc(f)
+}
```
mortelli, line 296: can't we import these from `retrieval`?

pkg/chunk/repair_test.go
```diff
@@ -0,0 +1,296 @@
...
+func newTestNetStore(t *testing.T, recoveryFunc chunk.RecoveryHook) storage.Storer {
+	t.Helper()
+	storer := mock.NewStorer()
+	logger := logging.New(ioutil.Discard, 5)
+
+	mockStorer := storemock.NewStorer()
+	serverMockAccounting := accountingmock.NewAccounting()
+	price := uint64(10)
+	pricerMock := accountingmock.NewPricer(price, price)
+	peerID := swarm.MustParseHexAddress("9ee7add7")
```
mortelli, line 260: could you please explain the reason behind the `price` and `peerID` values?

pkg/chunk/repair_test.go
```diff
@@ -0,0 +1,296 @@
...
+	ctx            context.Context
+	expectsFailure bool
+}
+
+// TestRecoveryHookCalls verifies that recovery hooks are being called as expected when net store attempts to get a chunk
+func TestRecoveryHookCalls(t *testing.T) {
+	// generate test chunk, store and publisher
+	c := chunktesting.GenerateTestRandomChunk()
+	ref := c.Address()
+	p := "0xbE165fe06c03e4387F79615b7A0b79d535e8D325"
```
mortelli, line 70: i know this might not impact the test, but i would call this `target` and shorten it to 1 byte for clarity's sake

pkg/chunk/repair_test.go
```diff
@@ -0,0 +1,296 @@
...
+		{
+			name:           "no targets in context",
+			ctx:            dummyContext,
+			expectsFailure: true,
+		},
+		{
+			name:           "targets set in context",
+			ctx:            targetContext,
+			expectsFailure: false,
+		},
```
mortelli, line 86: we could also add (maybe for a next iteration) test cases for targets set/unset, but with chunks that are actually retrievable.

that way we could verify that the recovery hooks are _not_ being called, even though they have enough data for that to be the case.

pkg/chunk/repair_test.go
```diff
@@ -0,0 +1,296 @@
...
+				if tc.expectsFailure {
+					return
+				}
+				t.Fatal("recovery hook was not called when expected")
+			}
+		})
+	}
+}
+
+func TestNewRepairHandler(t *testing.T) {
```
mortelli, line 122: it is kind of hard to read/follow this test.

do you think it's possible to do this through a `for _, tc in range tcs...`-type of statement and re-use some of the code here?

mortelli, line 122: please add a brief comment on what this test covers

pkg/chunk/repair.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/pss"
+	"github.com/ethersphere/bee/pkg/pushsync"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/ethersphere/bee/pkg/trojan"
+)
+
+const (
+	// RecoveryTopicText is the string used to construct the recovery topic
```
mortelli, line 23: please go through all of the definition comments in the diff and add a period (`.`) at the end of each one, so that we follow the coding guidelines

pkg/api/file.go
```diff
@@ -41,7 +41,7 @@ const (
 	EncryptHeader     = "swarm-encrypt"
 )
 
-type targetsContextKey struct{}
+type TargetsContextKey struct{}
```
janos, line 44: This type should not be exported, or not in the `api` package. This is a good example of a problematic design. By exporting it in `api` package, other packages that need this information must import `api` package creating a possibility for circular imports and breaking the layers of abstraction. Move this type to an appropriate package. Package `api` is a "top level" package that imports and depends other packages. This change makes `chunk` and `netstore` packages unimportable in this package.

I would even suggest not to export it as the purpose of the custom types for context keys is to avoid conflicts. Functions that would set and get required values with using unexported type for context key would be ideal, but not strictly required. Of course, not to define them in the `api` package.

acud: Yes, this stands out. I would suggest to move this to the `sctx` package, and use setters and getters on the context in the same manner it is used on the other values there

pkg/chunk/repair.go
```diff
@@ -0,0 +1,95 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package chunk
```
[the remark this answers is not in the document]
acud: also - I have a bit of a problem with having a `chunk` package per-se. The `Chunk` type is in `swarm` package, so having a `chunk` package that does not even relate to the actual chunk data type seems counter intuitive for me. `recovery` would be a possible alternative

pkg/chunk/repair.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// RecoveryHook defines code to be executed upon failing to retrieve pinned chunks
+type RecoveryHook func(ctx context.Context, chunkAddress swarm.Address) error
+
+// sender is the function call for sending trojan chunks
+type sender func(ctx context.Context, targets trojan.Targets, topic trojan.Topic, payload []byte) (*pss.Monitor, error)
+
+// NewRecoveryHook returns a new RecoveryHook with the sender function defined
+func NewRecoveryHook(send sender) RecoveryHook {
```
acud, line 42: I'm not sure there is justification to have `sender` type alias here. Why not inject the PSS protocol interface as a dependency?

pkg/chunk/repair.go
acud: well the repair will print the error and return. I think this is OK. I would even go as far as putting the debug line in the error handling block to trace

pkg/chunk/repair.go
acud: this code with the normalization of values should be moved to `sctx` package

acud: this code with the normalization of values should be moved to `sctx` package

pkg/chunk/repair.go
acud: It is actually not necessary to `TrimPrefix` here, since `hex.DecodeString` already handles this

pkg/netstore/netstore.go
```diff
@@ -9,6 +9,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/ethersphere/bee/pkg/api"
```
acud, line 12: the code additions in this package have no unit tests in `netstore` package

pkg/netstore/netstore.go
```diff
@@ -65,7 +82,19 @@ func (s *store) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chun
...
-	return s.Storer.Put(ctx, mode, chs...)
+	exist, err = s.Storer.Put(ctx, mode, chs...)
+	if err != nil {
+		return nil, err
+	}
+	// if callback is defined, call it for every new, valid chunk
+	if s.deliveryCallback != nil {
+		for i, exists := range exist {
+			if !exists {
+				go s.deliveryCallback(chs[i])
```
acud, line 93: i am not 100% sure that a closure is not needed here. cc @janos

pkg/chunk/repair.go
mortelli: that's fair, but i don't think we should be repairing chunks which are not locally pinned.

to me, it seems kind of counter intuitive to mark chunks as globally pinned which we are not pinning locally, but we can certainly discuss this further.

pkg/chunk/repair.go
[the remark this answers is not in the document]
jmozah: my actual intention of adding a chunk package is to move the chunks related things from swarm to here. This was discussed here as part of trojan chunks  https://github.com/ethersphere/bee/pull/429/files#r453443685

pkg/chunk/repair.go
jmozah: > but i don't think we should be repairing chunks which are not locally pinned.

Why? Is this a hard requirement? as long as the pinner is able to access the chunk.. it should repair it.

Eknir: I think this is an edge-condition. Normally, you would expect a chunk to be locally pinned, but if a chunk is not locally pinned and the "pinner" can still repair, then why not.

Eknir: The pinner might use ways outside of Swarm to make sure a chunk is persisted. Hence, a chunk might not be locally pinned from Swarm-perspective, but it is still persisted by the node. I think it's fine for a chunk not to be locally pinned, but still repair-able.

pkg/netstore/netstore.go
```diff
@@ -17,15 +19,21 @@ import (
...
+var (
+	ErrRecoveryAttempt = errors.New("chunk recovery initiated")
+)
+
 // New returns a new NetStore that wraps a given Storer.
-func New(s storage.Storer, r retrieval.Interface, logger logging.Logger, validators ...swarm.ChunkValidator) storage.Storer {
-	return &store{Storer: s, retrieval: r, logger: logger, validators: validators}
+func New(s storage.Storer, rcb chunk.RecoveryHook, dcb func(swarm.Chunk), r retrieval.Interface, logger logging.Logger,
+	validators ...swarm.ChunkValidator) storage.Storer {
+	return &store{Storer: s, recoveryCallback: rcb, deliveryCallback: dcb, retrieval: r, logger: logger, validators: validators}
```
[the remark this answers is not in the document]
jmozah: Well... I didn't want to pollute the storage.Storer interface with the call back functions. 
Dont know if this is the right thing to do

pkg/netstore/netstore.go
```diff
@@ -37,9 +45,18 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 			// request from network
 			data, err := s.retrieval.RetrieveChunk(ctx, addr)
 			if err != nil {
+				targets := ctx.Value(api.TargetsContextKey{})
```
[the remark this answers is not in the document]
jmozah: The return value if the `ctx.Value` is nil if it is not set. SO we have to check for nil first.
I am not what will happen when we cast a nil as string. so checking for nil first.
If this is passed, we pass the context inside the `getPinners` where it is casted to string and extracted.

pkg/netstore/netstore.go
```diff
@@ -37,9 +45,18 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
+				if s.recoveryCallback != nil && targets != "" && targets != nil {
```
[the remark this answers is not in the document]
jmozah: see above comment for detail

pkg/netstore/netstore.go
```diff
@@ -65,7 +82,19 @@ func (s *store) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chun
...
+	}
+	// if callback is defined, call it for every new, valid chunk
+	if s.deliveryCallback != nil {
+		for i, exists := range exist {
+			if !exists {
+				go s.deliveryCallback(chs[i])
+			}
+		}
+	}
+	return exist, nil
```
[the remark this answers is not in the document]
jmozah: yes. if we get `exists == true`, it means that the particular `trojan` message has already been received. and we don't want to  deliver the message twice.

pkg/chunk/repair.go
```diff
@@ -0,0 +1,95 @@
...
+	return func(ctx context.Context, chunkAddress swarm.Address) error {
+		targets, err := getPinners(ctx)
+		if err != nil {
+			return err
+		}
+		payload := chunkAddress
+
+		if _, err := send(ctx, targets, RecoveryTopic, payload.Bytes()); err != nil {
```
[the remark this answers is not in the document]
jmozah: pss.Send takes bytes as the last argument

pkg/chunk/repair_test.go
jmozah: not from retrieval_test

pkg/chunk/repair_test.go
jmozah: no reason. They just need to be there for the retreival to work without panic. I changed the values so they they clearly look dummy.

pkg/chunk/repair_test.go
jmozah: This test is just to check if the hook is called or not.. The sending of data is covered in pss and similarly other packages. What you are saying is a integration test that will not have mock packages and not unit test. We should do that in cluster test.

pkg/chunk/repair_test.go
jmozah: Its kinda difficult to do that since every test case needs different initialization.

pkg/chunk/repair.go
jmozah: PSS doesnt have a interface yet. @santicomp2014 is working on it. i will add it once it is done.

acud: > for mocking in the test this could be easier, plus we really need just a senderfunc

PSS interface should be lean and short, so it should also be easy to mock. Even if PSS does not yet have an interface, the interface definition can be defined in this package, and the PSS instance can be passed as a dependency.

pkg/chunk/repair.go
```diff
@@ -0,0 +1,95 @@
...
+			logger.Debugf("chunk repair: error while sending chunk or receiving receipt: %v", err)
+			return
+		}
+	}
+}
+
+// getPinners returns the specific target pinners for a corresponding chunk by
+// reading the prefix targets sent in the download API.
+func getPinners(ctx context.Context, logger logging.Logger) (trojan.Targets, error) {
+	targetString := sctx.GetTargets(ctx)
```
acud, line 81: `sctx.GetTargets` could return `(trojan.Targets, error)`, why just settle for a string when you can move all of this business logic there?

pkg/chunk/repair.go
```diff
@@ -0,0 +1,95 @@
...
+		if err != nil {
+			logger.Debugf("chunk repair: error while getting chunk for repairing: %v", err)
+			return
+		}
+
+		// push the chunk using push sync so that it reaches it destination in network
+		_, err = pushSyncer.PushChunkToClosest(ctx, ch)
+		if err != nil {
+			logger.Debugf("chunk repair: error while sending chunk or receiving receipt: %v", err)
+			return
```
acud, line 73: this can be removed

pkg/netstore/netstore.go
```diff
@@ -37,9 +45,18 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 			// request from network
 			data, err := s.retrieval.RetrieveChunk(ctx, addr)
 			if err != nil {
+				targets := sctx.GetTargets(ctx)
+				if s.recoveryCallback != nil && targets != "" {
```
acud, line 49: `sctx.GetTargets` should not return a string

pkg/netstore/netstore.go
acud: also, having given this further thought, i am not sure that we want to have this here in general. 

why: this means that every incoming chunk we check if its a trojan chunk. for example, this means that every chunk we retrieve with the `retrieval` protocol will be checked to be a trojan chunk. this is unnecessary computation that need not be done.

the way i see it, trojan chunks have two ingress points - one on pushsync protocol, the second on pullsync (for mailboxing). not sure if this needs to be on both (it might be that pushsync is only necessary right now), but it would be more efficient to have this hook only called in push and pull sync protocols.

pkg/chunk/repair.go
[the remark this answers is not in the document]
acud: i don't think that the chunk repair business logic should be in the same package as the chunk type.

pkg/netstore/netstore.go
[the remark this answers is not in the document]
mortelli: i'm not saying we should change `storage.Storer`. we can (and i think we should) keep these new callback functions in the `store` struct.

my only suggestion is that you remove the delivery callback from the _constructor_, and set it instead with a separate function.

i guess a case could be made for both, but i'd rather we have the signature reflect the fact that recovery callbacks should always be there, but that this is not the case for the delivery callbacks.

pkg/chunk/repair.go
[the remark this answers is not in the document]
mortelli: > afair, thought u had decided to use `prod/client prod/server`.

same here.

asking for content repair and globally pinning it are 2 related but separate things.

if you we are not going to separate them now, please create an issue to so that this is documented.

(if this issue already exists, i apologize, but i have not been able to find it)

pkg/node/node.go
```diff
@@ -254,8 +257,16 @@ func NewBee(o Options) (*Bee, error) {
...
 
-	ns := netstore.New(storer, retrieve, logger, content.NewValidator(), soc.NewValidator())
+	// instantiate the pss object
+	psss := pss.NewPss(storer, tagg)
 
+	// create recovery callback for content repair
+	recoverFunc := chunk.NewRecoveryHook(psss.Send)
+
+	// delivery call back for delivery of the registered messages
+	deliverFunc := psss.Deliver
```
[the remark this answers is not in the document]
mortelli: @jmozah

pkg/chunk/repair.go
mortelli: why would we advertise content as globally pinned which is not being pinned locally?

if a node goes through the trouble of enabling itself as a global pinner and having its address advertised, i don't think it's too much to ask to also pin the content it's supposed to be able to restore.

let's also look at the flip side: if we don't force local pinning here, we are potentially having a ton of wasted repair requests, traffic, etc.

finally, with regards to outside mechanisms of chunk persistence, i suggest we cross that bridge when we get to it.

pkg/chunk/repair_test.go
mortelli: (heads-up: comment still is in uppercase)

pkg/chunk/repair_test.go
mortelli: maybe i am missing something here, is there a problem with exporting these from the test package rather than duplicating the code?

pkg/chunk/repair_test.go
mortelli: thanks

pkg/chunk/repair_test.go
mortelli: > This test is just to check if the hook is called or not.

right, and the hook being called depends on:
- the chunk being missing
- targets being present
- a recovery hook being defined

no?

pkg/chunk/repair.go
mortelli: at least the following new functions are missing comments:
- `WithPushSyncer` 
- `deliverToPSS`
- `handleDeliveryResponse`

pkg/netstore/netstore.go
```diff
@@ -37,9 +45,18 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
 			if err != nil {
+				targets := ctx.Value(api.TargetsContextKey{})
+				if s.recoveryCallback != nil && targets != "" && targets != nil {
+					go func() {
+						err := s.recoveryCallback(ctx, addr)
+						if err != nil {
+							s.logger.Debugf("netstore: error while recovering chunk: %v", err)
+						}
+					}()
+					return nil, ErrRecoveryAttempt
```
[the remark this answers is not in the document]
mortelli: maybe we are talking about different things, but i'd thought we'd settled on something like: if a chunk get fails and recovery is requested, enhance your calm, then try your luck again later.

pkg/chunk/repair.go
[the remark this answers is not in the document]
jmozah: > asking for content repair and globally pinning it are 2 related but separate things.

@mortelli I dont understand. They are same, one side is asking (client) other side is doing (server). Both are what is called chunk repair in toto. I never understood why it was called prod in the first place. so changed it to retrieve per @acud's comment.

pkg/chunk/repair.go
jmozah: As i said in the `mattermost`, we should recommend the pinner's to pin it for providing proper service. But we don't have to check if it is pinned. We just have to see if it is accessible (which is the same). If we can get the chunk, we repair it, otherwise not.

pkg/netstore/netstore.go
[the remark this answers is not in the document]
jmozah: yes. @mortelli  is right. For now we start a async chunk repair process and return a http 202 reponse in the API asking him to try again later. How much kater and how many time he has to retry, is up to the client for now.

pkg/recovery/repair.go
```diff
@@ -0,0 +1,67 @@
...
+
+// NewRepairHandler creates a repair function to re-upload globally pinned chunks to the network with the given store.
+func NewRepairHandler(s storage.Storer, logger logging.Logger, pushSyncer pushsync.PushSyncer) pss.Handler {
+	return func(m trojan.Message) {
+		chAddr := m.Payload
+		ctx := context.Background()
+
+		ch, err := s.Get(ctx, storage.ModeGetRequest, swarm.NewAddress(chAddr))
+		if err != nil {
+			logger.Tracef("chunk repair: error while getting chunk for repairing: %v", err)
```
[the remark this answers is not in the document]
jmozah: @acud wnated it to be Trace

pkg/recovery/repair.go
```diff
@@ -0,0 +1,67 @@
...
+			return
+		}
+
+		// push the chunk using push sync so that it reaches it destination in network
+		_, err = pushSyncer.PushChunkToClosest(ctx, ch)
+		if err != nil {
+			logger.Tracef("chunk repair: error while sending chunk or receiving receipt: %v", err)
```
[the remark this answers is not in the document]
jmozah: @acud wanted it to be Trace

pkg/recovery/repair.go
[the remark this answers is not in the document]
janos: Ideally, this error would be passed, by my opinion. My reason for Debug is that we associate Trace with additional information on successful execution https://github.com/ethersphere/bee/blob/master/CODING.md#trace, while Debug for specifying internal information about errors.

pkg/netstore/netstore.go
```diff
@@ -36,9 +44,18 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 			// request from network
 			ch, err := s.retrieval.RetrieveChunk(ctx, addr)
 			if err != nil {
+				targets, err := sctx.GetTargets(ctx)
+				if s.recoveryCallback != nil && err == nil {
```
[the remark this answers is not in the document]
jmozah: If recoveryCallback is nil, this would not enter the block and so the ErrRecoveryAttempt would not be sent. Added the test case.

pkg/netstore/netstore.go
```diff
@@ -36,9 +44,18 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
+					go func() {
+						err := s.recoveryCallback(ctx, addr, targets)
```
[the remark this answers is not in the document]
mortelli: this actually happened in the global pinning implementation back in `swarm`. you'll need a new context.

pkg/recovery/repair.go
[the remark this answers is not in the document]
jmozah: This error is passed now. SO i am leaving it as Trace.  Please let me know if this is okay.

pkg/chunk/repair.go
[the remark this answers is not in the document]
mortelli: ultimately i don't mind one name or the other, but you've already differentiated them by calling them client and server:

- all nodes should be able to request a chunk repair (this is default behaviour)
- only nodes that are configured as such will be able to do the repair routine (global pinners)

so `prod/client` and `prod/server` (replace `prod` with `recovery` if you wish) would keep these things related but separate.

pkg/chunk/repair.go
mortelli: i understand yours and @acud's points, but i still think this is conceptually unsound.

it seems i'm the only one who is insisting on this though, so i will gladly resolve this thread and not block on this any longer.

pkg/pushsync/pushsync.go
```diff
@@ -276,3 +255,29 @@ func (ps *PushSync) PushChunkToClosest(ctx context.Context, ch swarm.Chunk) (*Re
 
 	return rec, nil
 }
+
+func (ps *PushSync) deliverToPSS(ch swarm.Chunk) {
+	// if callback is defined, call it for every new, valid chunk
+	if ps.deliveryCallback != nil {
+		go ps.deliveryCallback(ch)
```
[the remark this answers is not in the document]
jmozah: Good question. Delivery callback uses `PSS.send` which internally uses `pushSync`.
`Pushsync` has a timeout (`timeToWaitForReceipt`) when sending a chunk  which should make this goroutine quit in 3 seconds max.

pkg/netstore/netstore.go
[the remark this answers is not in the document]
mortelli: resolving, just saw the updated signature

pkg/pss/pss_test.go
```diff
@@ -174,15 +175,16 @@ func TestRegister(t *testing.T) {
 	}
 
 	registeredHandler := pss.GetHandler(testTopic)
-	registeredHandler(trojan.Message{}) // call handler to verify the retrieved func is correct
+	registeredHandler(context.Background(), trojan.Message{}) // call handler to verify the retrieved func is correct
```
mortelli, line 178: missing err check here

pkg/pss/pss_test.go
```diff
@@ -191,7 +193,7 @@ func TestRegister(t *testing.T) {
 	}
 
 	registeredHandler = pss.GetHandler(testTopic)
-	registeredHandler(trojan.Message{}) // call handler to verify the retrieved func is correct
+	registeredHandler(context.Background(), trojan.Message{}) // call handler to verify the retrieved func is correct
```
mortelli, line 196: here as well

pkg/pss/pss_test.go
```diff
@@ -223,8 +225,9 @@ func TestDeliver(t *testing.T) {
 
 	// create and register handler
 	var tt trojan.Topic // test variable to check handler func was correctly called
-	hndlr := func(m trojan.Message) {
+	hndlr := func(ctx context.Context, m trojan.Message) error {
 		tt = m.Topic // copy the message topic to the test variable
+		return nil
 	}
 	pss.Register(topic, hndlr)
```
mortelli, line 232: also, it's not showing up in the diff here, but on line `235` the `Deliver` call is also not error checked

pkg/recovery/repair_test.go
```diff
@@ -0,0 +1,314 @@
...
+func (s mockPeerSuggester) EachPeer(topology.EachPeerFunc) error {
+	return errors.New("not implemented")
+}
+func (s mockPeerSuggester) EachPeerRev(f topology.EachPeerFunc) error {
+	return s.eachPeerRevFunc(f)
+}
+
+type mockPssSender struct {
+	hookC     chan bool
+	pssSender recovery.PssSender
```
mortelli, line 307: are we using this `pssSender` attribute at all?

pkg/netstore/netstore_test.go
```diff
@@ -92,26 +97,105 @@ func TestNetstoreNoRetrieval(t *testing.T) {
 	}
 }
 
+func TestRecovery(t *testing.T) {
+	hookWasCalled := make(chan bool, 1)
+	rec := &mockRecovery{
+		hookC: hookWasCalled,
+	}
+
+	retreive, _, nstore := newRetrievingNetstore(rec)
```
mortelli, line 106: typo, should be `retrieve`

pkg/netstore/netstore_test.go
```diff
@@ -92,26 +97,105 @@ func TestNetstoreNoRetrieval(t *testing.T) {
...
+	select {
+	case <-hookWasCalled:
+		break
+	case <-time.After(100 * time.Millisecond):
+		t.Fatal("recovery hook was not called")
+	}
+}
+
+func TestInvalidRecoveryFunction(t *testing.T) {
+	retreive, _, nstore := newRetrievingNetstore(nil)
```
mortelli, line 126: same typo here

pkg/netstore/netstore_test.go
```diff
@@ -92,26 +97,105 @@ func TestNetstoreNoRetrieval(t *testing.T) {
...
+	}
+}
+
+func TestInvalidTargets(t *testing.T) {
+	hookWasCalled := make(chan bool, 1)
+	rec := &mockRecovery{
+		hookC: hookWasCalled,
+	}
+
+	retreive, _, nstore := newRetrievingNetstore(rec)
```
mortelli, line 144: and here

pkg/node/node.go
```diff
@@ -258,22 +261,45 @@ func NewBee(o Options) (*Bee, error) {
...
 
-	ns := netstore.New(storer, retrieve, logger, chunkvalidators)
+	// instantiate the pss object
+	psss := pss.NewPss(pss.Options{
+		Logger: logger,
+		Tags:   tagg,
+	})
+
+	// create recovery callback for content repair
+	recoverFunc := recovery.NewRecoveryHook(psss)
```
janos, line 271: This variable can be put inside `if o.GlobalPinningEnabled {`.

pkg/pushsync/pushsync.go
```diff
@@ -35,32 +35,35 @@ type Receipt struct {
...
-	storer        storage.Putter
-	peerSuggester topology.ClosestPeerer
-	tagg          *tags.Tags
-	logger        logging.Logger
-	metrics       metrics
+	streamer         p2p.Streamer
+	storer           storage.Putter
+	peerSuggester    topology.ClosestPeerer
+	tagg             *tags.Tags
+	deliveryCallback func(swarm.Chunk) error// callback func to be invoked to deliver chunks to PSS
```
mortelli, line 42: gofmt would add a whitespace char before this comment

pkg/pushsync/pushsync_test.go
```diff
@@ -145,20 +145,27 @@ func TestHandler(t *testing.T) {
 	triggerPeer := swarm.MustParseHexAddress("6000000000000000000000000000000000000000000000000000000000000000")
 	closestPeer := swarm.MustParseHexAddress("f000000000000000000000000000000000000000000000000000000000000000")
 
+	// mock call back function to see if pss message is delivered when it is received in the destination (closestPeer in this testcase)
+	hookWasCalled := make(chan bool, 1) // channel to check if hook is called
+	pssDeliver := func(ch swarm.Chunk) error{
```
mortelli, line 150: this line also needs gofmt (space before `{`)

pkg/pss/pss.go
```diff
@@ -85,17 +89,18 @@ func (p *Pss) Register(topic trojan.Topic, hndlr Handler) {
 }
 
 // Deliver allows unwrapping a chunk as a trojan message and calling its handler func based on its topic
-func (p *Pss) Deliver(c swarm.Chunk) {
+func (p *Pss) Deliver(c swarm.Chunk) error{
 	if trojan.IsPotential(c) {
+		ctx := context.Background()
```
janos, line 94: Would it be better to accept the context in Deliver function? This context here is not so useful for cancelation.

pkg/pushsync/pushsync.go
```diff
@@ -276,3 +255,36 @@ func (ps *PushSync) PushChunkToClosest(ctx context.Context, ch swarm.Chunk) (*Re
 
 	return rec, nil
 }
+
+func (ps *PushSync) deliverToPSS(ch swarm.Chunk) error{
+	// if callback is defined, call it for every new, valid chunk
+	errC := make(chan error)
+	defer close(errC)
+	if ps.deliveryCallback != nil {
+		go func () {
```
janos, line 264: What is the purpose of this goroutine and errC? With current code, it is just blocking or channel read and the equivalent code would be `return ps.deliveryCallback(ch)`.

pkg/pushsync/pushsync.go
```diff
@@ -276,3 +255,31 @@ func (ps *PushSync) PushChunkToClosest(ctx context.Context, ch swarm.Chunk) (*Re
 
 	return rec, nil
 }
+
+func (ps *PushSync) deliverToPSS(ctx context.Context, ch swarm.Chunk) error {
+	// if callback is defined, call it for every new, valid chunk
+	errC := make(chan error)
```
janos, line 261: This channel has no purpose.

pkg/pss/pss_test.go
jmozah: it is checked....

pkg/recovery/repair.go
```diff
@@ -0,0 +1,66 @@
...
+// sender is the function call for sending trojan chunks.
+type PssSender interface {
+	Send(ctx context.Context, targets trojan.Targets, topic trojan.Topic, payload []byte) (*tags.Tag, error)
+}
+
+// NewRecoveryHook returns a new RecoveryHook with the sender function defined.
+func NewRecoveryHook(pss PssSender) RecoveryHook {
+	return func(chunkAddress swarm.Address, targets trojan.Targets) error {
+		payload := chunkAddress
+		ctx := context.Background()
```
janos, line 41: Would it be more useful for RecoveryHook to accept context instead to create a new one here?

pkg/pushsync/pushsync.go
jmozah: i am not sure which version are you looking.. i have a space here...

pkg/pushsync/pushsync_test.go
jmozah: here too i see it in my code...

pkg/recovery/repair.go
jmozah: i changed like this because of this review comment https://github.com/ethersphere/bee/pull/479#discussion_r466899136.

pkg/pushsync/pushsync.go
jmozah: oopsss...

pkg/recovery/repair.go
janos: I am aware. I am sorry, I assumed that the reason for context propagation is cleat.

pkg/chunk/repair_test.go
jmozah: It is problem exporting with test packages

pkg/chunk/repair_test.go
jmozah: the targets set testcase is there in netstore.. because it is there where we check the targets.

pkg/chunk/repair.go
jmozah: Do we have to add comments to un-exported functions too? please point me to the coding guideline for the same.

pkg/pss/pss_test.go
mortelli: now it is. thanks

pkg/pushsync/pushsync.go
mortelli: i was looking at the version which was up at the time of submitting my review. this is why the `outdated` tag is present.

pkg/pushsync/pushsync_test.go
mortelli: same answer as before.

pkg/chunk/repair.go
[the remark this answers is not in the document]
jmozah: so you want to block on this?

pkg/chunk/repair_test.go
mortelli: > the targets set testcase is there in netstore.. because it is there where we check the targets.

? this is `repair_test.go`

pkg/chunk/repair.go
mortelli: no. we don't have to. but it's better if we do.

not a blocker.

pkg/netstore/netstore.go
```diff
@@ -34,11 +42,23 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
 			if err != nil {
-				return nil, fmt.Errorf("netstore retrieve chunk: %w", err)
+				if s.recoveryCallback == nil {
+					return nil, err
+				}
+				targets, err := sctx.GetTargets(ctx)
+				if err != nil {
+					return nil, err
+				}
+				go func() {
```
janos, line 54: Related to my comments on the context, if recovery.RecoveryHook would accept a context argument, it would be possible to terminate this goroutine when netstore is closed, or maybe to set it a timeout. I think that it is important to control the lifetime of our goroutines.

jmozah: I am not taking this for now in this PR as the plan is to have a "instant repair" feature, which means we have to block in netstore until the response comes or timeouts.

pkg/netstore/netstore.go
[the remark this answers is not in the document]
jmozah: @zelig For the next iteration, we are going to add instant chunk repair, where we wait for the "repaired chunk" from the pinner itself. See this https://github.com/ethersphere/bee/issues/548.
So we will not require retry logic at all anywhere.

pkg/chunk/repair.go
```diff
@@ -0,0 +1,95 @@
...
+		ctx := context.Background()
+
+		ch, err := s.Get(ctx, storage.ModeGetRequest, swarm.NewAddress(chAddr))
+		if err != nil {
+			logger.Debugf("chunk repair: error while getting chunk for repairing: %v", err)
+			return
+		}
+
+		// push the chunk using push sync so that it reaches it destination in network
+		_, err = pushSyncer.PushChunkToClosest(ctx, ch)
```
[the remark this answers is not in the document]
jmozah: See this https://github.com/ethersphere/bee/issues/548

pkg/pushsync/pushsync.go
```diff
@@ -276,3 +255,29 @@ func (ps *PushSync) PushChunkToClosest(ctx context.Context, ch swarm.Chunk) (*Re
 
 	return rec, nil
 }
+
+func (ps *PushSync) deliverToPSS(ctx context.Context, ch swarm.Chunk) error {
```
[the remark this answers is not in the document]
jmozah: because PSS uses push sync to send messages. when it receives a message it is the closest point from which we can deliver to PSS. Otherwise the chunk will land in localstore and we have to intercept there. which means that for any normal upload too the chunks will be delivered to PSS which is not good.

See this comment by @acud https://github.com/ethersphere/bee/pull/479#discussion_r466326141

PR #488, opened 2020-07-28
Debounce statestore writes for Accounting

metacertain:
As part of this PR we create the accounting/debouncer package,
The debouncer package is meant to be a disk-write-bufferzone, that limits the number of disk writes for a store key within the time scope of the debouncer interval to 1 / interval.


Eknir:
What is the status of this PR?

PR #490, opened 2020-07-29
add test for tags

Eknir:
Shows failing test for adding an upload with tags in the header. The bug was uncovered by running:

`curl -H "swarm-tag-uid: abcdef" -F file=@README.md http://localhost:8080/file` which got
` curl: (52) Empty reply from server`

Log-lines show:

image


Eknir:
I can report that it works if you don't pass tags explicitly:
image

jmozah:
There is a test available already which passes. see debugapi/tag-test.go

PR #493, opened 2020-07-29
Fix panic in tags debug api

jmozah:
When the debug api is enabled,  /tags and /tags/{uid} api panic.
This is because the tag is not initialized in the debug server.
This was not caught in the unit tests as the test cases uses mock server and in real world we use the debug server which is initialized in "node" package

### DIFF
--- a/pkg/api/bytes.go
+++ b/pkg/api/bytes.go
@@ -27,6 +27,9 @@ type bytesPostResponse struct {
 // bytesUploadHandler handles upload of raw binary data of arbitrary length.
 func (s *server) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 	ta := s.createTag(w, r)
+	if ta == nil {
+		return
+	}
 
 	// Add the tag to the context
 	r = r.WithContext(context.WithValue(r.Context(), tags.TagsContextKey{}, ta))
--- a/pkg/api/file.go
+++ b/pkg/api/file.go
@@ -67,6 +67,9 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
 	var fileSize uint64
 
 	ta := s.createTag(w, r)
+	if ta == nil {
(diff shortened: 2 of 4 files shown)


pkg/api/bytes.go
```diff
@@ -27,6 +27,9 @@ type bytesPostResponse struct {
 // bytesUploadHandler handles upload of raw binary data of arbitrary length.
 func (s *server) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 	ta := s.createTag(w, r)
+	if ta == nil {
+		return
```
janos, line 31: I would skip this check or add an internal server error response.

pkg/api/file.go
```diff
@@ -67,6 +67,9 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
 	var fileSize uint64
 
 	ta := s.createTag(w, r)
+	if ta == nil {
```
janos, line 70: I would skip this check or add an internal server error response.

pkg/api/bytes.go
jmozah: CreateTag already creates a internal response Error. Here we just dont proceed so that the error is sent back as response.

pkg/api/file.go
jmozah: same as above

pkg/file/splitter/internal/job.go
```diff
@@ -176,7 +176,13 @@ func (s *SimpleSplitterJob) sumLevel(lvl int) ([]byte, error) {
 	ref := s.hasher.Sum(nil)
 	addr = swarm.NewAddress(ref)
 
-	ch := swarm.NewChunk(addr, c)
+	var ch swarm.Chunk
```
Eknir, line 179: A comment, explaining what you are doing here would be good :)

Issue #476, opened 2020-07-27
extract jsonmanifest into manifest repo


mortelli:
The `manifest` interfaces defined on `bee` should remain as they are, and they can be iterated on if we come to a situation in which we need additional functionalities, different signatures, etc.

However, the `jsonmanifest` implementation should ideally be migrated to the `ethersphere/manifest` repo.

This will allow us to keep a contract from `bee` to whatever manifest implementation(s) we choose to make use of.

`jsonmanifest` might be temporary or it could be iterated should this be needed down the road. In this way, this task could be executed in this separate repo while keeping `bee` intact.

cc: @zbiljic @pradovic @zelig @acud

Issue #480, opened 2020-07-27
add getter for jsonmanifest Entries


mortelli:
Protect the `Entries` field of the `jsonmanifest.JSONManifest` type with a mutex, since it is a `map` and directly accessing it is not thread safe.

Original comment:
> oops. you're accessing an exposed map field which is not protected. so this implementation is not thread safe. would be better if jsonmanifest does not expose this field and has designated getters with mutexes protecting field access. obviously this is not an issue right now since the implementation is single threaded. still better to have this as thread safe from the get-go

_Originally posted by @acud in https://github.com/ethersphere/bee/pull/447#discussion_r460697867_

mortelli:
cc @zbiljic

PR #495, opened 2020-07-29
manifest, jsonmanifest, api, node: add encoding marshaling to manifest interface

### LINKED ISSUE #494: add encoding marshaling to manifest interface
Instead of using parser structs/funcs, `manifest.Interface` should embed the binary marshaler and unmarshaler interfaces.

As a consequence, the `jsonmanifest` implementation should also be updated accordingly.

Related user story: #418

mortelli:
This PR removes the `manifest` parser-related structures, interfaces and functions and embeds the `encoding` binary marshaller and unmarshaller interfaces in its stead.

It also:
- updates the `jsonmanifest` implementation to comply with the new interface.
- piggybacks the splitting of `jsonmanifest/manifest.go` into 2 files (one for `jsonmanifest.Entry`) for easier maintainability.
- removes the word `Get` from `manifest.Entry` func names
  - as a result, `jsonmanifest.JSONEntry` fields had to be unexported, a constructor added, and custom marshalling/unmarshalling funcs were added as well. this struct now uses pointer semantics
- updates some `manifest` and `jsonmanifest` comments.

We are still eventually migrating all of this out of the codebase, through #476.

closes #494

### DIFF
--- a/pkg/manifest/jsonmanifest/entry.go
+++ b/pkg/manifest/jsonmanifest/entry.go
@@ -1,0 +1,37 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package jsonmanifest
+
+import (
+	"net/http"
+
+	"github.com/ethersphere/bee/pkg/manifest"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// JSONEntry is a JSON representation of a file entry for a JSONManifest
+type JSONEntry struct {
+	Reference swarm.Address `json:"reference"`
+	Name      string        `json:"name"`
+	Headers   http.Header   `json:"headers"`
+}
+
+// verify JSONEntry implements manifest.Entry
+var _ manifest.Entry = (*JSONEntry)(nil)
+
+// GetReference returns the address of the entry for a file
+func (me JSONEntry) GetReference() swarm.Address {
+	return me.Reference
+}
+
+// GetName returns the name of the file for the entry
+func (me JSONEntry) GetName() string {
+	return me.Name
+}
+
+// GetHeaders returns the headers for manifest entry for a file
+func (me JSONEntry) GetHeaders() http.Header {
+	return me.Headers
+}
--- a/pkg/manifest/jsonmanifest/entry.go
+++ b/pkg/manifest/jsonmanifest/entry.go
@@ -62,7 +62,7 @@ func (me *JSONEntry) MarshalJSON() ([]byte, error) {
 	})
 }
 
-// UnmarshalJSON implements the json.Marshaler interface.
+// UnmarshalJSON implements the json.Unmarshaler interface.
 func (me *JSONEntry) UnmarshalJSON(b []byte) error {
 	e := exportEntry{}
 	if err := json.Unmarshal(b, &e); err != nil {
--- a/pkg/manifest/jsonmanifest/entry.go
+++ b/pkg/manifest/jsonmanifest/entry.go
@@ -11,27 +11,27 @@ import (
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
-// verify JSONEntry implements manifest.Entry
+// verify JSONEntry implements manifest.Entry.
 var _ manifest.Entry = (*JSONEntry)(nil)
 
-// JSONEntry is a JSON representation of a single manifest entry for a JSONManifest
+// JSONEntry is a JSON representation of a single manifest entry for a JSONManifest.
 type JSONEntry struct {
 	Reference swarm.Address `json:"reference"`
 	Name      string        `json:"name"`
 	Headers   http.Header   `json:"headers"`
 }
 
-// GetReference returns the address of the file in the entry
+// GetReference returns the address of the file in the entry.
 func (me JSONEntry) GetReference() swarm.Address {
 	return me.Reference
 }
 
-// GetName returns the name of the file in the entry
+// GetName returns the name of the file in the entry.
 func (me JSONEntry) GetName() string {
 	return me.Name
 }
 
-// GetHeaders returns the headers for the file in the manifest entry
+// GetHeaders returns the headers for the file in the manifest entry.
 func (me JSONEntry) GetHeaders() http.Header {
 	return me.Headers
 }
--- a/pkg/manifest/jsonmanifest/entry.go
+++ b/pkg/manifest/jsonmanifest/entry.go
@@ -11,27 +11,27 @@ import (
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
-// JSONEntry is a JSON representation of a file entry for a JSONManifest
+// verify JSONEntry implements manifest.Entry
+var _ manifest.Entry = (*JSONEntry)(nil)
+
+// JSONEntry is a JSON representation of a single manifest entry for a JSONManifest
 type JSONEntry struct {
 	Reference swarm.Address `json:"reference"`
 	Name      string        `json:"name"`
 	Headers   http.Header   `json:"headers"`
 }
 
-// verify JSONEntry implements manifest.Entry
-var _ manifest.Entry = (*JSONEntry)(nil)
-
-// GetReference returns the address of the entry for a file
+// GetReference returns the address of the file in the entry
 func (me JSONEntry) GetReference() swarm.Address {
 	return me.Reference
 }
 
-// GetName returns the name of the file for the entry
+// GetName returns the name of the file in the entry
 func (me JSONEntry) GetName() string {
 	return me.Name
 }
 
-// GetHeaders returns the headers for manifest entry for a file
+// GetHeaders returns the headers for the file in the manifest entry
 func (me JSONEntry) GetHeaders() http.Header {
 	return me.Headers
 }
--- a/pkg/manifest/jsonmanifest/entry.go
+++ b/pkg/manifest/jsonmanifest/entry.go
@@ -16,22 +16,31 @@ var _ manifest.Entry = (*JSONEntry)(nil)
 
 // JSONEntry is a JSON representation of a single manifest entry for a JSONManifest.
 type JSONEntry struct {
-	Reference swarm.Address `json:"reference"`
-	Name      string        `json:"name"`
-	Headers   http.Header   `json:"headers"`
+	reference swarm.Address
+	name      string
+	headers   http.Header
 }
 
-// GetReference returns the address of the file in the entry.
-func (me JSONEntry) GetReference() swarm.Address {
-	return me.Reference
+// NewEntry creates a new JSONEntry struct and returns it.
+func NewEntry(reference swarm.Address, name string, headers http.Header) JSONEntry {
+	return JSONEntry{
+		reference: reference,
+		name:      name,
+		headers:   headers,
+	}
 }
 
-// GetName returns the name of the file in the entry.
-func (me JSONEntry) GetName() string {
-	return me.Name
+// Reference returns the address of the file in the entry.
+func (me JSONEntry) Reference() swarm.Address {
+	return me.reference
 }
 
-// GetHeaders returns the headers for the file in the manifest entry.
-func (me JSONEntry) GetHeaders() http.Header {
-	return me.Headers
+// Name returns the name of the file in the entry.
+func (me JSONEntry) Name() string {
+	return me.name
+}
+
+// Headers returns the headers for the file in the manifest entry.
+func (me JSONEntry) Headers() http.Header {
+	return me.headers
 }
--- a/pkg/manifest/jsonmanifest/entry.go
+++ b/pkg/manifest/jsonmanifest/entry.go
@@ -5,6 +5,7 @@
 package jsonmanifest
 
 import (
+	"encoding/json"
 	"net/http"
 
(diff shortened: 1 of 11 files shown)


pkg/manifest/jsonmanifest/entry.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"net/http"
+
+	"github.com/ethersphere/bee/pkg/manifest"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// verify JSONEntry implements manifest.Entry
+var _ manifest.Entry = (*JSONEntry)(nil)
```
janos, line 15: I know that this is out of scope for this PR, but i have a naming suggestion to drop `JSON` prefixes from types in this package. I think that json is clear from the package name. I would go even further and rename package to just `json`, but that requires package naming in imports, which I do not mind, but it may be too much. :)

pkg/manifest/manifest.go
```diff
@@ -5,38 +5,36 @@
...
 
 import (
+	"encoding"
 	"errors"
 	"net/http"
 
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
+// ErrNotFound is returned when an Entry is not found in the manifest
```
janos, line 15: I see that this package does not follow the full sentence comment guideline (missing dots at the end). I would suggest to have that in order to have better documentation in godoc. https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences referenced in https://github.com/ethersphere/bee/blob/master/CODING.md.

This is also more related to the current state of the code in master, than to this PR.

pkg/manifest/jsonmanifest/manifest.go
```diff
@@ -6,39 +6,27 @@ package jsonmanifest
...
 var _ manifest.Interface = (*JSONManifest)(nil)
 
+// JSONManifest is a JSON representation of a manifest
+// it stores manifest entries in a map based on string keys
 type JSONManifest struct {
 	Entries map[string]JSONEntry `json:"entries,omitempty"`
 }
 
+// NewManifest creates a new JSONManifest struct and returns a pointer to it
 func NewManifest() *JSONManifest {
```
acud, line 23: this should ideally return `manifest.Interface`, although, i think that due to the fact you're accessing the exposed `Entries` from outside the package its a no-go (at least until that's sorted). Also, once the constructor returns the interface type, `JSONManifest` can be safely unexported.

pkg/manifest/manifest.go
```diff
@@ -5,38 +5,36 @@
...
 	Add(string, Entry)
-	// Remove reference from file on specified path
+	// Remove a manifest entry on the specified path
 	Remove(string)
-	// FindEntry returns manifest entry if one is found on specified path
+	// FindEntry returns a manifest entry if one is found in the specified path
 	FindEntry(string) (Entry, error)
-	// Serialize return encoded manifest
-	Serialize() ([]byte, error)
+	// BinaryMarshaler is the interface implemented by an object that can marshal itself into a binary form
```
acud, line 26: there's no need to document these interface methods (both marshal and unmarshal)

pkg/manifest/manifest.go
```diff
@@ -5,38 +5,36 @@
...
+	encoding.BinaryMarshaler
+	// BinaryUnmarshaler is the interface implemented by an object that can unmarshal a binary representation of itself
+	encoding.BinaryUnmarshaler
 }
 
-// Entry represents single manifest entry
+// Entry represents a single manifest entry
 type Entry interface {
-	// GetReference returns address of the entry file
+	// GetReference returns the address of the file in the entry
```
acud, line 34: I think we can remove the `Get` prefixes here too. Just as `swarm.Address` has `String()` and not `GetString()`, or `Bytes()` and not `GetBytes()`

pkg/manifest/manifest.go
```diff
@@ -5,38 +5,36 @@
...
 	GetReference() swarm.Address
-	// GetName returns the name of the file for the entry, if added
+	// GetName returns the name of the file in the entry
 	GetName() string
-	// GetHeaders returns the headers for manifest entry, if configured
+	// GetHeaders returns the headers for the file in the manifest entry
```
acud, line 38: I'm not really sure why we need to tie the manifest interface definition to the `http` package. This rapes manifests into an opinionated implementation which deviates from the generalisation it is supposed to provide. This is not a blocker, but the business logic on which headers to return to the requester need to live in the http api package, not inside manifests.

pkg/manifest/manifest.go
```diff
@@ -5,38 +5,36 @@
...
 type Interface interface {
-	// Add a manifest entry to specified path
+	// Add a manifest entry to the specified path
 	Add(string, Entry)
-	// Remove reference from file on specified path
+	// Remove a manifest entry on the specified path
 	Remove(string)
-	// FindEntry returns manifest entry if one is found on specified path
+	// FindEntry returns a manifest entry if one is found in the specified path
 	FindEntry(string) (Entry, error)
```
acud, line 25: as with my comment below on `Get` prefixes, I think that we could live with just `Entry(path string)` here

pkg/manifest/jsonmanifest/entry.go
mortelli: > I know that this is out of scope for this PR, but i have a naming suggestion to drop `JSON` prefixes from types in this package. I think that json is clear from the package name. I would go even further and rename package to just `json`, but that requires package naming in imports, which I do not mind, but it may be too much. :)

we've previously spoken about this with @zbiljic and we 100% agree.

we decided to keep this as-is while it is inside of `bee`, and then actually renaming it to just `json` when ported out, and using a named import when adding back the dependency.

pkg/manifest/manifest.go
mortelli: shoot, i missed this.

thanks, fixed

pkg/manifest/manifest.go
mortelli: fixed

pkg/manifest/jsonmanifest/entry.go
janos: Cool, thanks @mortelli. 👍

pkg/manifest/manifest.go
mortelli: fixed, thanks

pkg/manifest/manifest.go
mortelli: done

pkg/manifest/jsonmanifest/manifest.go
mortelli: right, i'm currently working on this through #480

mortelli:
re-requesting reviews from @janos and @acud since:
- based on @acud's comments, `Get` was removed from the `manifest.Entry` interface func names
- as a result, `jsonmanifest.JSONEntry` fields were made private
- because of this, custom `MarshalJSON` and `UnmarshalJSON` funcs had to be added
  - an auxiliary `struct` was created for this

pkg/manifest/jsonmanifest/entry.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/ethersphere/bee/pkg/manifest"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// verify JSONEntry implements manifest.Entry.
+var _ manifest.Entry = (*JSONEntry)(nil)
+
+// JSONEntry is a JSON representation of a single manifest entry for a JSONManifest.
+type JSONEntry struct {
```
janos, line 19: Given that JSONEntry only implements manifest.Entry interface, it could be unexported and wherever it is returned or accepted as argument manifest.Entry can be used.

A different problem is with serialization and different manifest implementations. Different implementations would require compatible serialization format in order to swap them. With this approach it is expected for all implementations to have the same json encoding. In that case I would expect that both manifest.Interface and mainfest.Entry require json marshaler and unmarshaller interfaces, not arbitrary binary serialization or to define what is the serialization that is required in order to have compatibility between implementations.

pkg/manifest/jsonmanifest/manifest.go
janos: Since Entries are required to be accessed outside, I see that as a miss in the manifest.Interface interface. If manifest package defines required interfaces, they should be complete. If there is a need to go from interface to a concrete type to perform some functionality, then the interface is not complete.

I am suggesting to unexport JSONManifest and JSONEntry, use appropriate manifest interfaces instead and see where the abstraction needs to be improved.

Also have in mind that a compatible encoding is also important for different implementations in this case.

pkg/manifest/jsonmanifest/entry.go
mortelli: > Given that JSONEntry only implements manifest.Entry interface, it could be unexported and wherever it is returned or accepted as argument manifest.Entry can be used.

i entirely agree. in fact, i've already implemented this change in PR #503; i didn't want to add more to this PR.

> A different problem is with serialization and different manifest implementations. 

considering [the more robust version of the manifest](https://github.com/ethersphere/manifest/pull/2), we definitely need binary marshalling—but regardless of what particular implementation we use at any given moment, i think it's wise to have `manifest.Interface` embed these 2 interfaces (rather than JSON).

> Different implementations would require compatible serialization format in order to swap them. 

in a discussion with @acud and @zbiljic we thought of adding integration tests through issue #476. that is: once we pull this implementation out of `bee` and into `manifest`, we would add tests on `bee` which can verify at least _some_ things about marshaling (or other behaviors).

> With this approach it is expected for all implementations to have the same json encoding. In that case I would expect that both manifest.Interface and mainfest.Entry require json marshaler and unmarshaller interfaces, not arbitrary binary serialization or to define what is the serialization that is required in order to have compatibility between implementations.

for the `jsonmanifest.JSONEntry` type, i'm not sure whether it matters or not what type of serialization we use as long as `jsonmanifest.JSONManifest` uses the binary marshaller, but i feel like you're telling me otherwise here.

could you please explain this a bit further? specifically: why do the manifest and the entry need to use the same serialization if only the manifest interface (and not the entry) defines an encoding?

should we maybe also embed the binary marshaller signature in the `manifest.Entry` interface?

pkg/manifest/jsonmanifest/manifest.go
mortelli: yes, i apologize for this :pray: 

i always try to do 1 PR per issue, but stuff kept coming up through reviews with this one. i tried to separate as much as possible but clearly i could have done better :sweat_smile:  

i thought maybe having this entry business in a 2nd PR (which builds on this one) would be good to tackle maybe the serialization first, and then the entries stuff later; to try and keep things as lean as possible.

in any case, i apologize for the confusion.

pkg/manifest/jsonmanifest/entry.go
acud: > In that case I would expect that both manifest.Interface and mainfest.Entry require json marshaler and unmarshaller interfaces

@janos I believe that there was some confusion with serialization here. The idea was that we should not be bound to json as a serialization format. Manifests, to the API, should only define a manifest -> bytes and bytes -> manifest interface. If we coerce manifests to implement JSON marshalling then we cannot implement other encoding, whereas binary marshalling allows us to be encoding agnostic.
It is a bit confusing that we are calling `json.Marshal/Unmarshal` on the types but this is only in order to generate and parse the actual encoded data, so that we don't need to manually generate and parse the manifest json.

pkg/manifest/jsonmanifest/entry.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// exportEntry is a struct used for marshaling and unmarshaling JSONEntry structs.
+type exportEntry struct {
+	Reference swarm.Address `json:"reference"`
+	Name      string        `json:"name"`
+	Headers   http.Header   `json:"headers"`
+}
+
+// MarshalJSON implements the json.Marshaler interface.
```
acud, line 56: @janos this was needed because of:
```go
// MarshalBinary implements encoding.BinaryMarshaler.
func (m *JSONManifest) MarshalBinary() (data []byte, err error) {
	return json.Marshal(m)
}
```

acud:
btw it would be cool if we can somehow along the way change the name of the implementation to something like `simpleManifest`. The name should not leak anything in regards to the format or encoding, this way we can also prevent confusion when reading the code.

pkg/manifest/jsonmanifest/entry.go
janos: Thanks for the responses @mortelli @acud. My assumption that different manifest implementation can be interchangeable with compatible serialization was wrong. Sorry. I know that binary serialization can be anything including json. Basically not all manifest implementations can be used in all places where mainfest.Interface is specified, as the serialization can differ.

The purpose of different serialization interfaces like JSON or YAML or any other specific ones is to clearly define what is the serialization possible, so that it is possible to relay on their functionalities. With having a general binary serialization interface with arbitrary serializations between implementations, it is not clear at compile time or by type checking which implementation implements which serializations. I am not sure if this was relevant for you when you designed the manifest abstraction, if not, please disregard this comment.

mortelli:
> btw it would be cool if we can somehow along the way change the name of the implementation to something like `simpleManifest`. The name should not leak anything in regards to the format or encoding, this way we can also prevent confusion when reading the code.

agreed, will make sure to make this change through #476

Issue #478, opened 2020-07-27
Debug API endpoint /chunks-pin returns error if no chunks are pinned


svetomir:
If no chunks are pinned `/chunks-pin` endpoint returns:
`{
    "message": "get first pin: leveldb: not found",
    "code": 500
}`

It should return empty list instead of an error.

jmozah:
Thanks.. will have a look in to it

jmozah:
@svetomir this is merged

PR #500, opened 2020-07-30
return empty response when nothing is pinned

jmozah:
When /chunks-pin (list pinned addresses) API is called and if nothing is pinned, the response is 
`{ "message": "get first pin: leveldb: not found", "code": 500 }`

Instead it would be good if an empty response is returned, like
`{"chunks":[]}`

This fixes the issue https://github.com/ethersphere/bee/issues/478

### DIFF
--- a/pkg/storage/mock/storer.go
+++ b/pkg/storage/mock/storer.go
@@ -263,6 +263,9 @@ func (m *MockStorer) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, s
 func (m *MockStorer) PinnedChunks(ctx context.Context, cursor swarm.Address) (pinnedChunks []*storage.Pinner, err error) {
 	m.pinSetMu.Lock()
 	defer m.pinSetMu.Unlock()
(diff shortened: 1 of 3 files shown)


pkg/storage/mock/storer.go
```diff
@@ -263,6 +263,9 @@ func (m *MockStorer) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, s
 func (m *MockStorer) PinnedChunks(ctx context.Context, cursor swarm.Address) (pinnedChunks []*storage.Pinner, err error) {
 	m.pinSetMu.Lock()
 	defer m.pinSetMu.Unlock()
+	if m.pinnedAddress == nil || len(m.pinnedAddress) == 0 {
```
acud, line 266: it is actually not necessary to check whether `m.pinnedAddress` is `nil`, since `len()` of uninitialized slice [will return zero anyway](https://play.golang.org/p/wTsFgdfr_qx)

Issue #481, opened 2020-07-28
upload empty directory


mortelli:
Extending the work in #447, implement the upload of empty directories.

Original [comment](https://github.com/ethersphere/bee/pull/447#discussion_r460699102) by @acud:
> so this won't support empty directories. would be nice to have the support so we don't lose information when uploading. this was happening in old swarm and i think this is counter-intuitive and an undesired behavior

zbiljic:
This came up during discussion I had with Viktor, and he mentioned that we should not created entries for empty directories, unless there is at least one file in them (@zelig please correct me if I misunderstood). Which means if there is information about directory from TAR archive, entry for that directory should be created, unless directory is empty.

Personally, I do not have any opinions about this (at this time at least), just mentioning for a reference.

acud:
@zbiljic @zelig I personally don't like the idea of omitting information from a given upload. Directory structure is as important as physical files. I object to omitting empty directories, they should be given consideration just as any other file type. This is counter intuitive for users too

Eknir:
I second @acud s opinion. Being able to upload empty directories is important for a user.

zelig:
> This came up during discussion I had with Viktor, and he mentioned that we should not created entries for empty directories, unless there is at least one file in them (@zelig please correct me if I misunderstood). Which means if there is information about directory from TAR archive, entry for that directory should be created, unless directory is empty.
> 
this must have been a misunderstanding. i specifically  want empty dirs to be represented and see it as a shortcoming of the old manifest format not to allow that.

the solution currently in mantaray is not as i designed it and want to change so 

bump this @Eknir @zbiljic

Eknir:
Hi @zelig , I would like to request a meeting on this. As can be seen in this discussion, @acud opposes to omitting empty directories and I also see the value of uploading them. Nevertheless, I assume there are solid reasons for you bumping this issue.
What about upcoming roundtable? If yes, could you put it on the agenda?

PR #503, opened 2020-07-30
api, jsonmanifest, manifest: getter for jsonmanifest entries

### LINKED ISSUE #480: add getter for jsonmanifest Entries

mortelli:
This PR aims to solve issue #480, which is about thread safety and exposure of the `entries` `map` field in the `JSONManifest` type. It adds a `sync.RWMutex` field to the struct for this purpose.

As a result, `JSONManifest` functions were modified to make use of this field.

This PR also:
- makes the `JSONManifest` field `entries` private. 
- removes the `Entries` function—we don't actually need to get all of the manifest entries at once.
  - in its stead, a `Length` function was added, which returns the amount of entries in a manifest variable. 
  - the `Length` function was also added to `manifest.Interface`, since it is used outside of the package in a way that other implementations should be able to comply with.
- unexports the `JSONManifest` and `JSONEntry` types.
  - custom marshalling/unmarshalling code for these types could be safely removed due to this, which simplifies the code.
- adds unit tests for the `jsonmanifest` package.

We are still eventually migrating all of this out of the codebase, through #476.

closes #480


pkg/api/dirs.go
```diff
@@ -117,24 +117,20 @@ func storeDir(ctx context.Context, reader io.ReadCloser, s storage.Storer, logge
...
+		fileEntry := jsonmanifest.NewEntry(fileReference, fileName, headers)
 
 		// add entry to dir manifest
 		dirManifest.Add(filePath, fileEntry)
 	}
 
 	// check if files were uploaded through the manifest
-	if len(dirManifest.Entries) == 0 {
-		return swarm.ZeroAddress, fmt.Errorf("no files in tar")
+	if dirManifest.Size() == 0 {
```
acud, line 127: I think `Length` is more legible here as `Size` might infer actual size in bytes

mortelli: agreed, done

pkg/manifest/jsonmanifest/manifest.go
```diff
@@ -6,51 +6,94 @@ package jsonmanifest
...
 }
 
 // UnmarshalBinary implements encoding.BinaryUnmarshaler.
-func (m *JSONManifest) UnmarshalBinary(data []byte) error {
-	return json.Unmarshal(data, m)
+func (m *jsonManifest) UnmarshalBinary(b []byte) error {
+	m.entriesMu.Lock()
+	defer m.entriesMu.Unlock()
+
+	e := exportManifest{}
```
mortelli, line 93: do we need a `make(map[string]*jsonEntry)` statement here?

why/why not?

pkg/manifest/jsonmanifest/manifest.go
```diff
@@ -6,51 +6,94 @@ package jsonmanifest
...
+	defer m.entriesMu.RUnlock()
+
+	entry, ok := m.entries[path]
+	if !ok {
+		return nil, manifest.ErrNotFound
 	}
 
-	return nil, manifest.ErrNotFound
+	// return a copy to prevent external modification
+	return NewEntry(entry.Reference(), entry.Name(), entry.Headers()), nil
```
mortelli, line 62: not 100% on this, but since we are using pointer semantics, we should be careful not to have the manifest entries modified from outside, no?

acud: you can easily check if this is necessary or not with a unit test (create an entry, get it, change it, and get it again from the source to see if the instance was changed). this test should also probably be shipped with the code. I am not entirely sure though, we actually _want_ to allow editing of entries in this manner, but I would assume no for now.

pkg/manifest/jsonmanifest/manifest.go
acud: i think we do, because of the json serializer attribute: ``json:"entries,omitempty"`` - it will omit the entry in case it is empty. i'm not sure about this 100% but again - classic for a unit test

pkg/manifest/jsonmanifest/manifest.go
```diff
@@ -6,51 +6,94 @@ package jsonmanifest
...
 }
 
 // Add adds a manifest entry to the specified path.
-func (m *JSONManifest) Add(path string, entry manifest.Entry) {
-	m.Entries[path] = NewEntry(entry.Reference(), entry.Name(), entry.Headers())
+func (m *jsonManifest) Add(path string, entry manifest.Entry) {
+	m.entriesMu.Lock()
+	defer m.entriesMu.Unlock()
+
+	m.entries[path] = &jsonEntry{
```
janos, line 36: Use `NewEntry`?

mortelli: since `NewEntry` returns a `manifest.Entry` type, and `entries` is of the `map[string]*jsonEntry` type, using `NewEntry` here would require a type assertion.

i don't think it's worth it in this case since this is internal to the package, and—for now—we are not handling errors in the `Add` function anyways (i.e. what if the assertion fails?)

pkg/manifest/jsonmanifest/entry.go
```diff
@@ -12,49 +12,49 @@ import (
...
+// exportEntry is a struct used for marshaling and unmarshaling jsonEntry structs.
 type exportEntry struct {
 	Reference swarm.Address `json:"reference"`
 	Name      string        `json:"name"`
 	Headers   http.Header   `json:"headers"`
 }
 
 // MarshalJSON implements the json.Marshaler interface.
-func (me *JSONEntry) MarshalJSON() ([]byte, error) {
+func (me *jsonEntry) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
mortelli: i like this idea, but in that case these fields would conflict with the `Reference`, `Name` and `Headers` funcs, which are defined by the `manifest.Entry` interface.

we could export them but call them something different, wdyt?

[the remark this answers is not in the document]
janos: Ah yes, the conflict, maybe to prefix them with a single letter, like X, or to name them only R, N and H.

pkg/manifest/jsonmanifest/manifest.go
janos: Yes, I see, it is fine like this.

pkg/manifest/jsonmanifest/entry.go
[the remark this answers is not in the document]
mortelli: > name them only R, N and H.

went with this one, thanks!

pkg/manifest/jsonmanifest/manifest.go
mortelli: thanks, i added the `TestEntryModification` unit test. turns out we did need the copy, after all.

afaik, manifest entries should only be modifiable (as part of the manifest) through `manifest.Interface` or `manifest.Entry` functions, whether it is `Add` or some other eventual method.

pkg/manifest/jsonmanifest/manifest.go
mortelli: after some discussion with @pradovic, we ended up simplifying the marshaling of the `jsonEntry` struct.

since it is now private, we can have the fields as exported with the json serializer attribute set, which allows us to just go ahead with the `json.Marshal` and `json.Unmarshal` calls.

pkg/manifest/jsonmanifest/entry.go
```diff
@@ -5,71 +5,42 @@
...
 }
 
 // Name returns the name of the file in the entry.
-func (me *JSONEntry) Name() string {
-	return me.name
+func (me *jsonEntry) Name() string {
+	return me.N
 }
 
 // Headers returns the headers for the file in the manifest entry.
```
acud, line 43: comment says `headers` but this returns just one `header` no?

pkg/manifest/jsonmanifest/jsonmanifest_test.go
```diff
@@ -0,0 +1,243 @@
...
+			{
+				reference: test.RandomAddress(),
+				name:      "entry-2.png",
+				headers:   http.Header{"Content-Type": {"image/png"}},
+				path:      "",
+			},
+		},
+	},
+	{
+		name: "nested-entries",
```
acud, line 53: what is nested here?

pkg/manifest/jsonmanifest/jsonmanifest_test.go
```diff
@@ -0,0 +1,243 @@
...
+		},
+	},
+}
+
+// TestEntries verifies that manifests behave as expected when adding and removing entries.
+// It also tests the Length and Entry functions.
+func TestEntries(t *testing.T) {
+	// get non-trivial test case
+	tc := testCases[len(testCases)-1] // last test case
+	if len(tc.entries) < 2 {
```
acud, line 88: wtf... why do we need `L87-L90`?

pkg/manifest/jsonmanifest/jsonmanifest_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+	m := jsonmanifest.NewManifest()
+	if m.Length() != 0 {
+		t.Fatalf("expected length to be %d, but is %d instead", 0, m.Length())
+	}
+
+	// add and check all entries
+	for i, e := range tc.entries {
+		entry := jsonmanifest.NewEntry(
+			e.reference,
```
acud, line 100: no need for carriage returns here

pkg/manifest/jsonmanifest/jsonmanifest_test.go
```diff
@@ -0,0 +1,243 @@
...
+		lastEntry.name,
+		lastEntry.headers,
+	)
+
+	// replace manifest entry by adding to the same path
+	path := filepath.Join(lastEntry.path, lastEntry.name)
+	m.Add(path, entry)
+
+	// length should not have changed
+	if m.Length() != manifestLen {
```
acud, line 130: these repetitions of `m.Length() ....` could be a helper function:
```go
func checkManifestLength(t *testing.T, m *manifest, count int){
if m.length != count { t.Fatal(yada yada) } ...
}

File: pkg/manifest/jsonmanifest/jsonmanifest_test.go [hunk 11]
```diff
@@ -0,0 +1,243 @@
...
+		t.Fatalf("expected length to be %d, but is %d instead", manifestLen, m.Length())
+	}
+	// check retrieved entry
+	verifyEntry(t, m, entry, path)
+
+	// try removing inexistent entry
```
acud, line 136: lots of comments here. i think that the code is very much self explanatory, it is not needed to comment before every function call. it would be easier to read if there's a description of what the test case is supposed to do at the comment on the top of the test (godoc), and avoid verbose commenting in the body of the test.

pkg/manifest/jsonmanifest/jsonmanifest_test.go
```diff
@@ -0,0 +1,243 @@
...
+		t.Fatal(err)
+	}
+	if reflect.DeepEqual(rre, re) {
+		t.Fatalf("manifest entry %v was unexpectedly modified externally", rre)
+	}
+}
+
+// TestMarshal verifies that created manifests are successfully marshalled and unmarshalled.
+func TestMarshal(t *testing.T) {
+	for _, tc := range testCases {
```
acud, line 200: you don't need multiple test cases for marshaling. one test will do

pkg/manifest/jsonmanifest/manifest.go
```diff
@@ -6,51 +6,82 @@ package jsonmanifest
...
+	if !ok {
+		return nil, manifest.ErrNotFound
 	}
 
-	return nil, manifest.ErrNotFound
+	// return a copy to prevent external modification
+	return NewEntry(entry.Reference(), entry.Name(), entry.Headers().Clone()), nil
+}
+
+// Length returns the amount of entries in the manifest.
```
acud, line 65: `amount` -> `number`, also, this comment might be misleading as one might think this is recursive and shows all the deep entries too, which might be the case for json (mention this explicitly), but might not be trivial for other implementations

pkg/manifest/manifest.go
```diff
@@ -23,6 +23,8 @@ type Interface interface {
 	Remove(string)
 	// Entry returns a manifest entry if one is found in the specified path.
 	Entry(string) (Entry, error)
+	// Length returns the amount of entries in the manifest.
```
acud, line 26: This is misleading. In implementations which are not the flat simple json this would be not so trivial. I would go with a general comment like `Length returns an implementation specific count of elements in the manifest.`

pkg/manifest/jsonmanifest/entry.go
mortelli: correct. i've changed _all_ references to singular (that was fun...!)

pkg/manifest/jsonmanifest/jsonmanifest_test.go
mortelli: the entries here represent the following structure:
```
nested-entries
├── readme.md
├── text/
│   └──robots.txt
└── img/
     ├── 1.png
     └── 2.png
```

pkg/manifest/jsonmanifest/jsonmanifest_test.go
mortelli: sorry :sweat_smile: i was paranoid that an eventual change to the `testCases` var would negatively impact `TestEntries`, since it always takes the last test case of the array. 

and i wondered: _what if this _new_ last test case was not big enough for this test to make sense any longer?_

...on second thought, if that were to be the case, i would rather this test failed.

removed

pkg/manifest/jsonmanifest/jsonmanifest_test.go
mortelli: fixed, thanks

pkg/manifest/jsonmanifest/jsonmanifest_test.go
mortelli: extracted `checkLength` helper func. thanks

pkg/manifest/jsonmanifest/jsonmanifest_test.go
mortelli: i'm not so convinced. a few commits ago, a tc with `nil` entries for the manifest was failing the marshalling, while _all_ the rest were succeeding.

we have the following cases:
- nil entries
- 1 plain entry
- 2 plain entries
- 4 nested entries 

i tried to think of representative test cases here. maybe i went a bit overboard, but i don't think it's doing more bad than good at this point.

also the last of these is actually used for another test, which verifies the `Add`, `Remove`, `Entry` and `Length` functions, so i think it should have at least _some_ complexity.

pkg/manifest/manifest.go
mortelli: done, thanks

pkg/manifest/jsonmanifest/manifest.go
mortelli: fixed, thanks

pkg/manifest/jsonmanifest/jsonmanifest_test.go
mortelli: thanks for this recommendation. i've iterated the godoc comments as well as the ones in the body.

pkg/manifest/jsonmanifest/jsonmanifest_test.go
acud: the (unmarshal) just calls json marshal. not sure why we need 4 tests to see that this builtin go functionality works. it is true that it comes at a low price in this case, but you could probably have 1 test case that covers everything. also, for `nil` entries it is not clear to me how and why such manifest would even be created. this should potentially not do anything or return an error

pkg/manifest/jsonmanifest/jsonmanifest_test.go
acud: oh wow this wasn't so clear at first glance. i wonder what `path` even gives as an added value here. why not have just one `path` field that stores the full path string and just provide `Name()` on the fly by returning whatever is after the last slash?

pkg/api/bzz.go
```diff
@@ -118,18 +118,18 @@ func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
 
 	manifestEntryAddress := me.Reference()
 
-	var additionalHeaders http.Header
```
zbiljic, line 121: @mortelli Please note that `http.Header` is actually alias for type `map[string][]string`, which is why plural was used.

pkg/manifest/jsonmanifest/jsonmanifest_test.go
mortelli: > the (un)marshal just calls json marshal. not sure why we need 4 tests to see that this builtin go functionality works. it is true that it comes at a low price in this case, but you could probably have 1 test case that covers everything. 

ultimately, you're right. the thing is that i coded these 4 test cases when the (un)marshaling functions were not just built-in go functionalities, but had some logic instead (unexported structs with json annotations and yada yada). 

after this, you, @janos and @pradovic helped me out in simplifying the structs (and their fields) so that we ended up not really needing anything else but the `json.Marshal` calls.

since the nil entries test-case is an interesting edge case (i will go into this shortly) and the last test case is used for testing the `Add`, `Remove`, `Entry` and `Length` funcs, we really only have 2 more test cases than what we really _need_; i.e. `one-entry` and `two-entries`.

> also, for `nil` entries it is not clear to me how and why such manifest would even be created. this should potentially not do anything or return an error

a newly created manifest starts out empty (meaning, no entries), so this can for sure exist in the code. whether it _should_ though, is a different story.

as far as i know, we don't (currently) have a place in the code which outputs such result, but i—maybe prematurely—anticipated this for issue #481. it is my understanding that an empty directory should have no entries; but right now, if you upload a tar with no files in it, this will result in an error. so yes, there is some inconsistency here that should be dealt with.

i hope to settle this when we tackle #481 or continue to iterate on manifests. i will keep it in mind—if it turns out that there is no good reason to keep an empty manifest around, then for sure let's have this as an error case.

pkg/api/bzz.go
mortelli: yes, thanks. i think both singular and plural are arguably acceptable here.

in any case, i reverted this change _only_ for `bzz.go` since i want this part of the code to be consistent with the `downloadHandler` signature.

pkg/manifest/jsonmanifest/jsonmanifest_test.go
mortelli: nice catch.

i've simplified the test case entry structure so that only `path` is used (which is a full path) and then the entry name is extracted from there.

PR #505, opened 2020-07-31
limit max http body size and detect it

janos:
This PR adds a simple http middleware constructor NewMaxBodyBytesHandler that can be chained on various router paths to limit the maximal allowed body size. Since the error is returned by reading from the body it has to be handled in the handler where it happens so a helper function is added HandleBodyReadError, which can be used to serve an appropriate Request Entity Too Large response.

This handler is integrated into the chunks api endpoint and should be used in other handlers where the protection from large response bodies is required, like in https://github.com/ethersphere/bee/pull/439.

### DIFF
--- a/pkg/jsonhttp/handlers.go
+++ b/pkg/jsonhttp/handlers.go
@@ -19,3 +19,37 @@ func (h MethodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 func NotFoundHandler(w http.ResponseWriter, _ *http.Request) {
 	NotFound(w, nil)
 }
+
+// NewMaxBodyBytesHandler is an http middleware constructor that limits the
+// maximal number of bytes that can be read from the request body. When a body
+// is read, the error can be handled with a helper function HandleBodyReadError
+// in order to respond with Request Entity Too Large response.
+// See TestNewMaxBodyBytesHandler as an example.
+func NewMaxBodyBytesHandler(limit int64) func(http.Handler) http.Handler {
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.ContentLength > limit {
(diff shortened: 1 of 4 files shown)


pkg/jsonhttp/handlers.go
```diff
@@ -19,3 +19,37 @@ func (h MethodHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+// HandleBodyReadError checks for particular errors and writes appropriate
+// response accordingly. If no known error is found, no response is written and
+// the function returns false.
+func HandleBodyReadError(err error, w http.ResponseWriter) (responded bool) {
+	if err == nil {
+		return false
+	}
+	// http.MaxBytesReader returns an unexported error,
+	// this is the only way to detect it
+	if err.Error() == "http: request body too large" {
```
paxthemax, line 50: Maybe extract this into an error type?

paxthemax:
@janos This looks very good and ready to merge. Excellent job!

paxthemax:
To the side: would it be a valid approach to add this middleware to every endpoint (as a default on the start of each handler chain), then explicitly override it on endpoints that tolerate lower content lengths or explicitly have no limit on request content length?

pkg/jsonhttp/handlers.go
janos: That would be another approach to handling. This function does not return an error, but handles the unfortunately unexported error from stdlib's http package. Another approach would be to create our own error and return it here, but I did this in order to minimize the code repeated in http handlers by having the response here and allow for other scenarios to be handled here if needed.

janos:
@paxthemax thanks for the review and merge. We can do something like that to ensure that every now handler is protected by default. Even we have only one or two where it is needed currently. This is a good improvement as we have only two handlers where should allow limitless upload size. The only downside is a small performance overhead to check at runtime the overridden value for every handler. Currently everything is set at compile time.

PR #511, opened 2020-08-03
fix forever loop timeout issue

jmozah:
When a wrap takes time (race test) or is supplied a longer target, it gets stuck in a forever loop mining the trojan chunk.
This had a timeout which was inside the longest target loop (i.e. timeout will be checked after trojan reaches 255 bytes long ).
Now it is fixed to check the timeout every few(256) iterations.

The solution is to increase the timeout and bring down the target length from 2 bytes to 1 bytes in tests. 

Fixes https://github.com/ethersphere/bee/issues/509

### DIFF
--- a/pkg/trojan/message.go
+++ b/pkg/trojan/message.go
@@ -11,8 +11,9 @@ import (
 	"math/big"
 	"time"
 
-	"github.com/ethersphere/bee/pkg/swarm"
 	bmtlegacy "github.com/ethersphere/bmt/legacy"
+
+	"github.com/ethersphere/bee/pkg/swarm"
 )
 
 // Topic is an alias for a 32 byte fixed-size array which contains an encoding of a message topic
@@ -37,11 +38,11 @@ const (
 	// MaxPayloadSize is the maximum allowed payload size for the Message type, in bytes
 	// MaxPayloadSize + Topic + Length + Nonce = Default ChunkSize
 	//    (4030)      +  (32) +   (2)  +  (32) = 4096 Bytes
-	MaxPayloadSize          = swarm.ChunkSize - NonceSize - LengthSize - TopicSize
-	NonceSize               = 32
-	LengthSize              = 2
-	TopicSize               = 32
-	MinerTimeout            = 5   // seconds after which the mining will fail
+	MaxPayloadSize = swarm.ChunkSize - NonceSize - LengthSize - TopicSize
+	NonceSize      = 32
+	LengthSize     = 2
+	TopicSize      = 32
+	MinerTimeout   = 5 // seconds after which the mining will fail
 )
 
 // NewTopic creates a new Topic variable with the given input string
@@ -163,21 +164,26 @@ func (m *Message) toChunk(targets Targets, span []byte) (swarm.Chunk, error) {
 		return nil, err
 	}
 
-	done := make(chan bool, 1)
+	doneC := make(chan bool, 1)
+	errC := make(chan error, 1)
 	var hash, s []byte
 	go func() {
+		defer close(doneC)
+		defer close(errC)
+
 		// mining operation: hash chunk fields with different nonces until an acceptable one is found
 		for {
 			s = append(append(span, nonce...), b...) // serialize chunk fields
 			hash, err = hashBytes(s)
 			if err != nil {
+				errC <- err
 				return
 			}
 
 			// take as much of the hash as the targets are long
 			if contains(targets, hash[:targetsLen]) {
 				// if nonce found, stop loop and return chunk
-				done <- true
+				doneC <- true
 				return
 			}
 			// else, add 1 to nonce and try again
@@ -192,7 +198,9 @@ func (m *Message) toChunk(targets Targets, span []byte) (swarm.Chunk, error) {
 
 	// checks whether the mining is completed or times out
 	select {
-	case <-done:
+	case err := <-errC:
+		return nil, err
+	case <-doneC:
 		return swarm.NewChunk(swarm.NewAddress(hash), s), nil
 	case <-time.After(MinerTimeout * time.Second):
 		return nil, ErrMinerTimeout
--- a/pkg/trojan/message.go
+++ b/pkg/trojan/message.go
@@ -37,11 +37,12 @@ const (
 	// MaxPayloadSize is the maximum allowed payload size for the Message type, in bytes
 	// MaxPayloadSize + Topic + Length + Nonce = Default ChunkSize
 	//    (4030)      +  (32) +   (2)  +  (32) = 4096 Bytes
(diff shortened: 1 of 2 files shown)


pkg/trojan/message_test.go
```diff
@@ -124,6 +125,23 @@ func TestWrapFail(t *testing.T) {
...
 
+func TestWrapTimeout(t *testing.T) {
+	m := newTestMessage(t)
+
+	// a large target will take more than MinerTimeout seconds, so timeout error will be triggered
+	buf := make([]byte, swarm.SectionSize)
+	_, err := rand.Read(buf)
+	if err != nil {
+		t.Fatal(err)
+	}
```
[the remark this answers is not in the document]
jmozah: sorry. i will like to stick to my original style, since i thin that is more readable.
if this is part of the go-coding guideline, then please quote the place so that i know.

jmozah:
> my suggestion here is to make the mining timeout as part of at least some configuration. this would allow it to be mockable, and would get around the 2 points aforementioned.

@mortelli  I would rather stick to the fixed timeout. If we parallelize in future we can still need this so that inc case if we could not get the nonce, we rather timeout. But there is a good news, The reduced the timeout because we reduced the target size from 2 bytes to 1 byte in the test cases and even in race test they get over soon.

pkg/trojan/message.go
```diff
@@ -163,32 +163,40 @@ func (m *Message) toChunk(targets Targets, span []byte) (swarm.Chunk, error) {
...
+	done := make(chan bool, 1)
+	var hash, s []byte
+	go func() {
+		// mining operation: hash chunk fields with different nonces until an acceptable one is found
+		for {
+			s = append(append(span, nonce...), b...) // serialize chunk fields
+			hash, err = hashBytes(s)
+			if err != nil {
+				return
+			}
```
janos, line 175: If this function would return an error, not that it will not be visible, but also this function will wait on MinerTimeout in order to return, which is not needed.

jmozah: good catch. Fixed

pkg/trojan/message.go
```diff
@@ -163,32 +164,47 @@ func (m *Message) toChunk(targets Targets, span []byte) (swarm.Chunk, error) {
...
 	}
 
-	// hash chunk fields with different nonces until an acceptable one is found
-	for start := time.Now(); ; {
-		s := append(append(span, nonce...), b...) // serialize chunk fields
-		hash1, err := hashBytes(s)
-		if err != nil {
-			return nil, err
-		}
+	doneC := make(chan bool, 1)
```
janos, line 167: I think that there is no need to have two channels now. `doneC` can be removed as sending a nil to `errC` would act the same, just to check it on the receiving end if it is nil. This would also, it would not require a buffered channel.

pkg/trojan/message_test.go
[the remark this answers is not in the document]
mortelli: no problem :+1:

mortelli:
> > my suggestion here is to make the mining timeout as part of at least some configuration. this would allow it to be mockable, and would get around the 2 points aforementioned.
> 
> @mortelli I would rather stick to the fixed timeout because of two reasons
> 
>     * if we expose these through config, we may confuse the user

fair point.
 
>     * If we parallelize in future we can still need this so that inc case if we could not get the nonce, we rather timeout.

sorry, i don't understand what you mean here. time-outs will be needed whether we parallelize or not, no?

> But there is a good news, i reduced the timeout because we reduced the target size from 2 bytes to 1 byte in the test cases (as that was the real world usecase) and even in race test they get over soon. It is taking only 0.4 sec in CI with race test, which should be the worst condition for this testcase.

this is indeed good news, but please be aware that this is something that might change in the future and therefore affect the test in some unpredictable manner.

in any case, if i'm the only one with this pov here, i am happy to consider this a non-blocker and approve this PR as soon as the rest of the changes are dealt with :+1:

pkg/trojan/message.go
acud: Alternatively, the mining code could have been written like this:
```go
t := time.After(MinerTimeout * time.Second)
for {
   select {
      case <-t:
          return nil, t.Err()
      default:
   }
   ...mine stuff here...
}
```
Personally I find this simpler and easier to read. But up to you

jmozah:
> sorry, i don't understand what you mean here. time-outs will be needed whether we parallelize or not, no?

Yeah that's what i am saying too.. the sentence has too many typo.. sorry

> this is indeed good news, but please be aware that this is something that might change in the future and therefore affect the test in some unpredictable manner.

Even if we parallelize in future, i don't think this will have any effect. But for now, i think this is fair enough.

acud:
> my suggestion here is to make the mining timeout as part of at least some configuration

This PR is above everything to clean up our CI pipeline from flaky tests. Let's not make this an overhaul PR

mortelli:
> > my suggestion here is to make the mining timeout as part of at least some configuration
> 
> This PR is above everything to clean up our CI pipeline from flaky tests. Let's not make this an overhaul PR

sure. an overhaul wasn't my intention. my intention was not to have _one_ test which took an entire minute to be successful every time it's executed.

acud:
> > > my suggestion here is to make the mining timeout as part of at least some configuration
> > 
> > 
> > This PR is above everything to clean up our CI pipeline from flaky tests. Let's not make this an overhaul PR
> 
> sure. an overhaul wasn't my intention. my intention was not to have _one_ test which took an entire minute to be successful every time it's executed.

yeah but I quoted the note about configuration. In general I agree we can decrease the timeout (even 10 seconds is close to eternity in this context), but pulling it out to configurable value seems maybe out of scope

jmozah:
> even 10 seconds is close to eternity in this context

its 5 seconds  :-)

mortelli:
> > > > my suggestion here is to make the mining timeout as part of at least some configuration
> > > 
> > > 
> > > This PR is above everything to clean up our CI pipeline from flaky tests. Let's not make this an overhaul PR
> > 
> > 
> > sure. an overhaul wasn't my intention. my intention was not to have _one_ test which took an entire minute to be successful every time it's executed.
> 
> yeah but I quoted the note about configuration. In general I agree we can decrease the timeout (even 10 seconds is close to eternity in this context), but pulling it out to configurable value seems maybe out of scope

so if not a configurable value, how would you go about having this test take less time to be completed?

just have the code be fast enough so that we use a small timeout? i would like to know if there are alternatives to this as for me this is not such a sure thing.

jmozah:
> so if not a configurable value, how would you go about having this test take less time to be completed?

@mortelli  The test taking time is becauese of CI environment and race detector switched on. Previously it was due to multi byte targets. Now it is fixed, so it should take less time.. I saw it taking less than 1/2 sec on Ci with race enabled.

Issue #515, opened 2020-08-04
panic in proximity function


janos:
After running a bee node built with code state on git commit `6efcbfdfc1f8d0f71b889d5a95a44aa2a773ec8a` as part of the testing session within the small network size, the node eventually got terminated with the panic below. It suggests that the Proximity function received a zero length address as the second argument `other` by the kademlia Disconnect function passed by the peerRegistry. 

```
panic: runtime error: index out of range [0] with length 0

goroutine 3194398 [running]:
github.com/ethersphere/bee/pkg/swarm.Proximity(0xc0000ac6a0, 0x20, 0x20, 0x0, 0x0, 0x0, 0x927749a22a34de01)
	github.com/ethersphere/bee/pkg/swarm/proximity.go:28 +0xc4
github.com/ethersphere/bee/pkg/kademlia.(*Kad).Disconnected(0xc0024c8000, 0x0, 0x0, 0x0)
	github.com/ethersphere/bee/pkg/kademlia/kademlia.go:414 +0x6e
github.com/ethersphere/bee/pkg/p2p/libp2p.(*peerRegistry).Disconnected(0xc000304fc0, 0x2061b60, 0xc000083380, 0x205d420, 0xc001136780)
	github.com/ethersphere/bee/pkg/p2p/libp2p/peer.go:77 +0x3fc
github.com/libp2p/go-libp2p-swarm.(*Conn).doClose.func1.1(0x2055b40, 0xc000304fc0)
	github.com/libp2p/go-libp2p-swarm@v0.2.8/swarm_conn.go:84 +0x5b
github.com/libp2p/go-libp2p-swarm.(*Swarm).notifyAll.func1(0xc0018075e4, 0xc0027491c0, 0x2055b40, 0xc000304fc0)
	github.com/libp2p/go-libp2p-swarm@v0.2.8/swarm.go:493 +0x62
created by github.com/libp2p/go-libp2p-swarm.(*Swarm).notifyAll
	github.com/libp2p/go-libp2p-swarm@v0.2.8/swarm.go:491 +0x118
```

janos:
https://github.com/ethersphere/bee/pull/517 is merged, but it only hardens the Proximity function, does not solve the root cause for this issue, so this issue should stay open to investigate why the nil address was passed to the Proximity function as the second argument.

pradovic:
[...]

PR #517, opened 2020-08-04
prevent panic in proximity function with shorter second argument

janos:
This PR is a proposal how to handle the symptom of the problem described in https://github.com/ethersphere/bee/issues/515. This changes the behaviour for proximity function to handle shorter second argument than the first one instead to panic. This behaviour would hide the root cause of the issue https://github.com/ethersphere/bee/issues/515, but the current implementation is not commutative binary function as it would handle shorter first argument, but not the second one.

The root cause of https://github.com/ethersphere/bee/issues/515 should be found, regardless of this PR being merged or not.

Additionally, mockValidator is removed in order to make the linter pass.

### DIFF
--- a/pkg/netstore/netstore_test.go
+++ b/pkg/netstore/netstore_test.go
@@ -21,10 +21,6 @@ import (
 
 var chunkData = []byte("mockdata")
 
-type mockValidator struct{}
-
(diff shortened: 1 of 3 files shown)


Eknir:
Suggesting to open an issue to find the root-cause, or not close #515 before the rootcause is found.

PR #549, opened 2020-08-09
change http header names per BOS / Standard

jmozah:
This PR alighn's the http header names used in the code as per the BOS.
The BOS says that the header should start with capitalised SWARM and hyphen and the the feature name.
like SWARM-XXXX where XXXX is the feature name.

This affects the following features
- tags
- local pinning
- encryption
- global pinning targets

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -35,7 +35,7 @@ type Options struct {
 
 const (
 	// TargetsRecoveryHeader defines the Header for Recovery targets in Global Pinning
-	TargetsRecoveryHeader = "swarm-recovery-targets"
+	TargetsRecoveryHeader = "SWARM-RECOVERY-TARGETS"
 )
 
 func New(o Options) Service {
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -35,7 +35,7 @@ type Options struct {
 
 const (
 	// TargetsRecoveryHeader defines the Header for Recovery targets in Global Pinning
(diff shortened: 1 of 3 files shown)


jmozah:
There are different option for us

1) x-swarm-zzzz
2) swarm-zzzz
3)Swarm-Zzzz
4) SWARM-ZZZZ
5) X-SWARM-ZZZZ

jmozah:
The official recommendation for non standard http headers seems to name the header sensibly without the X- prefix.
Refer https://tools.ietf.org/html/rfc6648

Per RFC 2616, field names are case insensitive. So we can have 2,3 or 4 options. But option 3 seems to be the unwritten standard.

janos:
I think that the most correct representation of header names is how https://golang.org/pkg/net/http/#CanonicalHeaderKey returns them.

jmozah:
> I think that the most correct representation of header names is how https://golang.org/pkg/net/http/#CanonicalHeaderKey returns them.

@janos Yes. We can add that in the book. But in the code, this should be case insensitive per rfc 6648.

janos:
> But in the code, this should be case insensitive per rfc 6648.

Yes, this PR is not needed, because of that.

jmozah:
> Yes, this PR is not needed, because of that.

@janos this PR is still needed to make sure we send the header names in proper format.

janos:
> @janos this PR is still needed to make sure we send the header names in proper format.

According to go stdlib Header implementation, it is not needed as keys will be canonicalized internally https://golang.org/pkg/net/http/#Header.Add.

PR #557, opened 2020-08-11
api, debugapi, internal, sctx, tags: tags cleanup

### LINKED ISSUE #563: restore tags api
Undoing the work done in #291, the `tags` API endpoint must be migrated back from `debugapi`.

Additionally:
- we should make sure that the API matches the Book of Swarm as close as possible.
- the `tags` `DoneSplit` endpoint (although not formally specified) should also be added.
- the code has to be cleaned up in terms of tidiness/maintainability.

mortelli:
This PR aims to solve issue #563, which is about moving back the `tags` API from `debugapi` and iterating on it.

Follow-up issue: #576.

closes #563 
fixes https://github.com/ethersphere/bee/issues/572


pkg/api/tag.go
```diff
@@ -0,0 +1,129 @@
...
+		return
+	}
+	w.Header().Set("Cache-Control", "no-cache, private, max-age=0")
+	jsonhttp.Created(w, newTagResponse(tag))
+}
+
+func (s *server) getTag(w http.ResponseWriter, r *http.Request) {
+	idStr := mux.Vars(r)["id"]
+
+	id, err := strconv.ParseUint(idStr, 10, 32)
```
acud, line 91: Atoi

pkg/tags/tags.go
```diff
@@ -26,17 +26,14 @@ import (
...
 	"time"
 
-	"github.com/ethersphere/bee/pkg/sctx"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
 var (
 	TagUidFunc  = rand.Uint32
-	ErrNotFound = errors.New("tag not found")
+	NotFoundErr = errors.New("tag not found")
```
acud, line 34: keep as `ErrNotFound`, it is the same convention as in other packages

mortelli: done

pkg/api/tag.go
mortelli: done, also in `deleteTag`.

thanks

pkg/api/tag.go
```diff
@@ -145,3 +146,51 @@ func (s *server) deleteTag(w http.ResponseWriter, r *http.Request) {
...
+		jsonhttp.InternalServerError(w, "error unmarshaling metadata")
+		return
+	}
+
+	tag, err := s.Tags.Get(uint32(id))
+	if err != nil {
+		if errors.Is(err, tags.ErrNotFound) {
+			s.Logger.Debugf("done split: tag not present: %v, id %s", err, idStr)
+			s.Logger.Error("done split: tag not present")
+			jsonhttp.NotFound(w, "tag not present")
```
acud, line 186: return after this line

mortelli: done, thanks

pkg/api/bytes.go
```diff
@@ -26,27 +26,28 @@ type bytesPostResponse struct {
...
+		s.Logger.Debugf("bytes upload: get or create tag: %v", err)
+		s.Logger.Error("bytes upload: get or create tag")
+		jsonhttp.InternalServerError(w, "cannot get or create tag")
 		return
 	}
 
 	// Add the tag to the context
-	r = r.WithContext(context.WithValue(r.Context(), tags.TagsContextKey{}, ta))
-	ctx := r.Context()
+	ctx := sctx.SetTag(r.Context(), tag)
```
mortelli, line 38: no need to also modify `r` since we are not using it further

pkg/api/router.go
```diff
@@ -66,6 +66,22 @@ func (s *server) setupRouting() {
...
+
+	router.Handle("/tags/{id}", jsonhttp.MethodHandler{
+		"GET": http.HandlerFunc(s.getTag),
+	})
+
+	router.Handle("/tags/{id}", jsonhttp.MethodHandler{
+		"DELETE": http.HandlerFunc(s.deleteTag),
+	})
+
+	router.Handle("/tags/{id}", jsonhttp.MethodHandler{
```
mortelli, line 81: not formally defined in the book

pkg/api/tag_test.go
```diff
@@ -29,37 +29,32 @@ type fileUploadResponse struct {
 func TestTags(t *testing.T) {
 	var (
 		bytesResource        = "/bytes"
-		resource             = func(addr swarm.Address) string { return "/chunks/" + addr.String() }
-		tagResourceUidCreate = func(name string) string { return "/tags?name=" + name }
-		tagResourceUUid      = func(uuid uint64) string { return "/tags/" + strconv.FormatUint(uuid, 10) }
+		chunksResource       = func(addr swarm.Address) string { return "/chunks/" + addr.String() }
+		createTagResource    = "/tags"
+		getTagResource       = func(id uint64) string { return "/tags/" + strconv.FormatUint(id, 10) }
```
mortelli, line 34: `Atoi` also here? @acud

pkg/sctx/sctx.go
```diff
@@ -39,18 +42,18 @@ func GetHost(ctx context.Context) string {
...
 
-// GetTag gets the tag unique identifier from the context
-func GetTag(ctx context.Context) uint32 {
-	v, ok := ctx.Value(tagKey{}).(uint32)
-	if ok {
-		return v
+// GetTag gets the tag instance from the context
+func GetTag(ctx context.Context) *tags.Tag {
+	v, ok := ctx.Value(tagKey{}).(*tags.Tag)
+	if !ok {
```
mortelli, line 53: changed for to have happy path outside of `if`

mortelli:
requesting preliminary reviews from @janos, @acud and @zelig while i add new tests, the rest of the code should be ready

pkg/api/tag_test.go
acud: `return fmt.Sprintf("/tags/%d",id)`?

pkg/api/tag.go
```diff
@@ -0,0 +1,199 @@
...
+			return
+		}
+		s.Logger.Debugf("delete tag: tag %v: %v", idStr, err)
+		s.Logger.Errorf("delete tag: %v", idStr)
+		jsonhttp.InternalServerError(w, "cannot get tag")
+		return
+	}
+
+	s.Tags.Delete(tag.Uid)
+	jsonhttp.Respond(w, http.StatusNoContent, nil)
```
acud, line 149: consider returning a json response here, just to indicate that operation was successful, e.g. https://github.com/ethersphere/bee/blob/master/pkg/debugapi/status.go

pkg/api/tag.go
```diff
@@ -0,0 +1,199 @@
...
+
+	id, err := strconv.Atoi(idStr)
+	if err != nil {
+		s.Logger.Debugf("done split tag: parse id  %s: %v", idStr, err)
+		s.Logger.Error("done split tag: parse id")
+		jsonhttp.BadRequest(w, "invalid id")
+		return
+	}
+
+	body, err := ioutil.ReadAll(r.Body)
```
acud, line 163: we should limit how many bytes we read here

pkg/api/tag_test.go
```diff
@@ -69,128 +64,140 @@ func TestTags(t *testing.T) {
...
 	})
 
 	t.Run("tag-counters", func(t *testing.T) {
-		rcvdHeaders := jsonhttptest.ResponseDirectSendHeadersAndReceiveHeaders(t, apiClient, http.MethodPost, resource(validHash), bytes.NewReader(validContent), http.StatusOK, jsonhttp.StatusResponse{
+		rcvdHeaders := jsonhttptest.ResponseDirectSendHeadersAndReceiveHeaders(t, client, http.MethodPost, chunksResource(validHash), bytes.NewReader(validContent), http.StatusOK, jsonhttp.StatusResponse{
 			Message: http.StatusText(http.StatusOK),
 			Code:    http.StatusOK,
 		}, nil)
-		uuid1 := isTagFoundInResponse(t, rcvdHeaders, nil)
+		id := isTagFoundInResponse(t, rcvdHeaders, nil)
```
acud, line 184: `id := uint32(...)` then you can remove the casts below, or just return `uint32` from `isTagFoundIn...`

pkg/tags/tags.go
```diff
@@ -101,16 +98,6 @@ func (ts *Tags) GetByAddress(address swarm.Address) (*Tag, error) {
 	return t, nil
 }
 
-// GetFromContext gets a tag from the tag uid stored in the context
-func (ts *Tags) GetFromContext(ctx context.Context) (*Tag, error) {
```
mortelli, line 105: not used anywhere, apparently

pkg/api/tag_test.go
```diff
@@ -0,0 +1,309 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
mortelli, line 1: this file wasn't just moved, but also iterated on.

if you want to see what was changed here, please inspect a diff between this file and `debugapi/tag_test.go` (which is removed through this PR)

acud: it would have been nice if we would have used `git mv`. i'm not sure why it was not shown as moved in the diff. it can still be done though

pkg/api/tag_test.go
mortelli: done, thanks

pkg/api/tag.go
mortelli: fixed by you already, thanks  :+1:

pkg/api/tag.go
mortelli: sorry, not sure about this, limit to how many?

pkg/api/api.go
```diff
@@ -50,10 +58,25 @@ func New(tags *tags.Tags, storer storage.Storer, corsAllowedOrigins []string, lo
 	return s
 }
 
-const (
-	SwarmPinHeader = "Swarm-Pin"
-	TagHeaderUid   = "swarm-tag-uid"
-)
+func (s *server) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
+	// if tag header is not there create a new one
```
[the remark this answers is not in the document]
mortelli: fixed, thanks

pkg/api/api.go
```diff
@@ -50,10 +58,25 @@ func New(tags *tags.Tags, storer storage.Storer, corsAllowedOrigins []string, lo
...
+			return nil, false, fmt.Errorf("cannot create tag: %w", err)
+		}
+		return tag, true, nil
+	}
+	// if the tag uid header is present, then use the tag sent
+	uid, err := strconv.Atoi(tagUid)
+	if err != nil {
+		return nil, false, fmt.Errorf("cannot parse taguid: %w", err)
+	}
+	t, err := s.Tags.Get(uint32(uid))
```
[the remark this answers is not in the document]
mortelli: you're right; i have added a docstring to this function.

regarding the behavior: a change could be made so that if an id is supplied but the tag cannot be fetched, then create a new tag.

however, the id supplied would not match the id for the newly created tag. i think this would be somewhat confusing, so i'd vote to keep this as-is.

mortelli:
i've created a follow-up issue here: #576.

pkg/pushsync/pushsync.go
```diff
@@ -209,6 +209,12 @@ func (ps *PushSync) PushChunkToClosest(ctx context.Context, ch swarm.Chunk) (*Re
 	peer, err := ps.peerSuggester.ClosestPeer(ch.Address())
 	if err != nil {
 		if errors.Is(err, topology.ErrWantSelf) {
+			// this is to make sure that the sent number does not diverge from the synced counter
+			t, err := ps.tagg.Get(ch.TagID())
```
zelig, line 213: sigh....not this tagg again...

pkg/api/tag.go
```diff
@@ -0,0 +1,203 @@
...
+			jsonhttp.NotFound(w, "tag not present")
+			return
+		}
+		s.Logger.Debugf("done split: tag %v: %v", idStr, err)
+		s.Logger.Errorf("done split: %v", idStr)
+		jsonhttp.InternalServerError(w, "cannot get tag")
+		return
+	}
+
+	tag.DoneSplit(tagr.Address)
```
zelig, line 201: what address are we using here??

pkg/pushsync/pushsync.go
acud: whatss wrongg?

pkg/api/api.go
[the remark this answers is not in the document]
acud: https://github.com/ethersphere/bee/issues/579 :+1:

pkg/api/tag.go
```diff
@@ -0,0 +1,203 @@
...
+		s.Logger.Error("get tag: parse id")
+		jsonhttp.BadRequest(w, "invalid id")
+		return
+	}
+
+	tag, err := s.Tags.Get(uint32(id))
+	if err != nil {
+		if errors.Is(err, tags.ErrNotFound) {
+			s.Logger.Debugf("get tag: tag not present: %v, id %s", err, idStr)
+			s.Logger.Error("get tag: tag not present")
```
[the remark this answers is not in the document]
acud: we don't want to leak details when logging with `Error`, that's why we add the ID only in the `Debug` line. This is in line with the logging in the rest of the package

pkg/api/tag.go
acud: the address that the api client has called the endpoint with.
two scenarios:
- i am an API client and I have multiple uploads that have different root hashes but there is not one canonical root hash that represents all of these uploads, then I call this api with an empty body. this calls the `DoneSplit` and sets `Split = Total` but the root hash is empty (because there is not a root hash)
- i am an API client and I have one single upload (or multiple) that can be represented by one root hash. I call this api endpoint with the JSON request and set the root hash of the tag to the one which represents my uploads.

pkg/pushsync/pushsync.go
mortelli: we fixed a bunch of these, but it seems this one managed to slip by.

i'll make sure to include this change through #576

pkg/api/tag.go
mortelli: > the address that the api client has called the endpoint with.
> two scenarios:
> 
>     * i am an API client and I have multiple uploads that have different root hashes but there is not one canonical root hash that represents all of these uploads, then I call this api with an empty body. this calls the `DoneSplit` and sets `Split = Total` but the root hash is empty (because there is not a root hash)

i'm a bit confused about this... should we be allowing `DoneSplit` calls with an empty address? 

because currently we are not

Issue #521, opened 2020-08-05
Fallback to bootnodes at runtime


janos:
Upon node start, connecting to bootnodes is tried. If non of the connections are succesfull and there are no peers in the addressbook, or at leas no peers that can be connected to, the node will not have try to connect to any other peers. If the node will periodically try to connect to the bootnodes if there are no peers in the addressbook, this problem would be solved. The node will be able to recover from no peers only by connecting to bootnodes again during runtime. This is maybe the best to do in kademlia manage goroutine by passing bootnodes as trusted peers that can fallback to connect again. This change my involve some generalizations with the bootnode connect code in node package and it should use the Discover function for dns bootnode discovery.

This issue was observed on the private clusters when a node had network problems connecting the bootnode at start, but the expectation was that it would be more resilient.

PR #558, opened 2020-08-11
Fallback to bootnodes

pradovic:
https://github.com/ethersphere/bee/issues/521

I have left `Start()` method only on kademlia for now, so I did not extract any extra interfaces. This is probably ok, as there is no other implementation of topology driver at the moment, and full connectivity does not need it.

### DIFF
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -27,11 +27,11 @@ const (
 )
 
 type Service struct {
-	streamer         p2p.Streamer
-	addressBook      addressbook.GetPutter
-	peerAddedHandler func(context.Context, ...swarm.Address) error
-	networkID        uint64
-	logger           logging.Logger
+	streamer        p2p.Streamer
+	addressBook     addressbook.GetPutter
+	addPeersHandler func(context.Context, ...swarm.Address) error
+	networkID       uint64
+	logger          logging.Logger
 }
 
 type Options struct {
@@ -79,8 +79,8 @@ func (s *Service) BroadcastPeers(ctx context.Context, addressee swarm.Address, p
 	return nil
 }
 
-func (s *Service) SetPeersAddedHandler(h func(ctx context.Context, addr ...swarm.Address) error) {
-	s.peerAddedHandler = h
+func (s *Service) SetAddPeersHandler(h func(ctx context.Context, addr ...swarm.Address) error) {
+	s.addPeersHandler = h
 }
 
 func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swarm.Address) (err error) {
@@ -151,8 +151,8 @@ func (s *Service) peersHandler(ctx context.Context, peer p2p.Peer, stream p2p.St
 		peers = append(peers, bzzAddress.Overlay)
 	}
 
-	if s.peerAddedHandler != nil {
-		if err := s.peerAddedHandler(ctx, peers...); err != nil {
+	if s.addPeersHandler != nil {
+		if err := s.addPeersHandler(ctx, peers...); err != nil {
 			return err
 		}
 	}
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -151,7 +151,7 @@ func (s *Service) peersHandler(ctx context.Context, peer p2p.Peer, stream p2p.St
 		peers = append(peers, bzzAddress.Overlay)
 	}
 
-	if s.peerAddedHandler != nil && len(peers) > 0 {
+	if s.peerAddedHandler != nil {
 		if err := s.peerAddedHandler(ctx, peers...); err != nil {
 			return err
 		}
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -134,6 +134,7 @@ func (s *Service) peersHandler(ctx context.Context, peer p2p.Peer, stream p2p.St
 	// but we still want to handle not closed stream from the other side to avoid zombie stream
 	go stream.FullClose()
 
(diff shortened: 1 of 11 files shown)


Eknir:
Hi Petar, could you write a description and/or connect this one with an issue?

pradovic:
@Eknir Sure, done!

pkg/kademlia/kademlia.go
```diff
@@ -211,10 +216,55 @@ func (k *Kad) manage() {
...
+
+			if k.connectedPeers.Length() == 0 {
+				k.connectBootnodes(ctx)
+			}
+
 		}
 	}
 }
 
+func (k *Kad) Start(ctx context.Context) error {
```
acud, line 228: why not execute this code as part of the kademlia constructor?

pkg/kademlia/mock/kademlia.go
```diff
@@ -58,9 +58,9 @@ func NewMockKademlia(o ...Option) *Mock {
 	return m
 }
 
-// AddPeer is called when a peer is added to the topology backlog
+// AddPeers  is called when a peers are added to the topology backlog
```
acud, line 61: change from AddPeer to AddPeers added a space in some occurrences

pkg/node/node.go
```diff
@@ -387,59 +396,8 @@ func NewBee(addr string, logger logging.Logger, o Options) (*Bee, error) {
...
-					return count > 3, nil
-				}); err != nil {
-					logger.Debugf("discover fail %s: %v", a, err)
-					logger.Warningf("discover to bootnode %s", a)
-					return
-				}
-			}(a)
-		}
-		wg.Wait()
+	if err := kad.Start(p2pCtx); err != nil {
```
acud, line 399: not sure if i understand why this is needed as a discrete step in the kademlia start up sequence? is there anything preventing the address book to populate kademlia as part of the kademlia constructor?

pradovic: I guess not, but I had a weird feeling when using it that way. Because, we need kademlia as a dependency on some other components so it is created pretty "early", while this kind of logic where there is an ongoing loop looks more like a "service" to me. To put it simple, I liked the idea of being able to instantiate other components with kademlia, but not start the logic that is driving rest of the system as well, such as connecting, notifying, etc... LMT if this makes sense?

pkg/hive/hive.go
```diff
@@ -27,11 +27,11 @@ const (
...
 
 type Service struct {
-	streamer    p2p.Streamer
-	addressBook addressbook.GetPutter
-	peerHandler func(context.Context, swarm.Address) error
-	networkID   uint64
-	logger      logging.Logger
+	streamer         p2p.Streamer
+	addressBook      addressbook.GetPutter
+	peerAddedHandler func(context.Context, ...swarm.Address) error
```
janos, line 32: A very minor naming suggestion `addPeersHandler`?

pkg/kademlia/doc.go
```diff
@@ -15,7 +15,7 @@ was persisted in the address book and the node has been restarted).
 
 So the information has been changed, and potentially upon disconnection,
 the depth can travel to a shallower depth in result.
-If a peer gets added through AddPeer, this does not necessarily infer
+If a peer gets added through AddPeers , this does not necessarily infer
```
janos, line 18: Remove additional whitespace.

pkg/node/node.go
```diff
@@ -202,10 +199,22 @@ func NewBee(addr string, logger logging.Logger, o Options) (*Bee, error) {
...
-	topologyDriver := kademlia.New(kademlia.Options{Base: address, Discovery: hive, AddressBook: addressbook, P2P: p2ps, Logger: logger})
-	b.topologyCloser = topologyDriver
-	hive.SetPeerAddedHandler(topologyDriver.AddPeer)
-	p2ps.AddNotifier(topologyDriver)
+	var bootnodes []ma.Multiaddr
+	for _, a := range o.Bootnodes {
+		addr, err := ma.NewMultiaddr(a)
+		if err != nil {
+			logger.Debugf("multiaddress fail %s: %v", a, err)
+			logger.Warningf("connect to bootnode %s", a)
```
janos, line 207: These log messages should be adjusted.

pkg/node/node.go
janos: This does give more control to the process, indeed. Even it may seem an overhead, I think that it is a safe approach.

pkg/topology/mock/mock.go
```diff
@@ -16,14 +16,14 @@ type mock struct {
 	peers           []swarm.Address
 	closestPeer     swarm.Address
 	closestPeerErr  error
-	addPeerErr      error
+	AddPeersErr     error
```
janos, line 19: It looks to me that this does not have to be exported.

pradovic: Ah, that was a rename. This is the last time I used rename all functionality, I am apparently very bad at using it :)

pkg/kademlia/doc.go
pradovic: Ah, did not save all files after auto rename 🙈

pkg/node/node.go
acud: that's fine, but the `manage` loop is started in the constructor. it falls under this same case, so either pull the manage goroutine start to this function, or have both in the ctor. WDYT?

pradovic: Sure, done!

pradovic:
handle  bootnodes in kademlia

PR #565, opened 2020-08-13
api: add pin on upload

acud:
This PR adds a `ModePutUploadPin` which allows to upload and pin content simultaneously in one atomic operation. `Swarm-Pin` header is also added to the API to enable pinning on all the relevant endpoints.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -6,6 +6,7 @@ package api
 
 import (
 	"net/http"
+	"strings"
 
 	"github.com/ethersphere/bee/pkg/logging"
 	m "github.com/ethersphere/bee/pkg/metrics"
@@ -47,4 +48,17 @@ func New(tags *tags.Tags, storer storage.Storer, corsAllowedOrigins []string, lo
 	s.setupRouting()
 
 	return s
+}
+
+const (
+	SwarmPinHeader = "Swarm-Pin"
+	TagHeaderUid   = "swarm-tag-uid"
+)
+
+// requestModePut returns the desired storage.ModePut for this request based on the request headers.
+func requestModePut(r *http.Request) storage.ModePut {
+	if h := strings.ToLower(r.Header.Get(SwarmPinHeader)); h == "true" {
+		return storage.ModePutUploadPin
+	}
+	return storage.ModePutUpload
 }
--- a/pkg/api/dirs.go
+++ b/pkg/api/dirs.go
@@ -91,7 +91,7 @@ func storeDir(ctx context.Context, reader io.ReadCloser, s storage.Storer, mode 
 		if err == io.EOF {
 			break
 		} else if err != nil {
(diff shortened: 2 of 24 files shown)


pkg/debugapi/tag_test.go
```diff
@@ -260,8 +247,8 @@ func TestTags(t *testing.T) {
 		if finalTag.Total != 3 {
 			t.Errorf("tag total count mismatch. got %d want %d", finalTag.Total, 3)
 		}
-		if finalTag.Seen != 3 {
-			t.Errorf("tag seen count mismatch. got %d want %d", finalTag.Seen, 3)
+		if finalTag.Seen != 1 {
```
acud, line 250: this has changed due to a bug in the mock storage that resulted in incorrect count of `seen` chunks on insertion

pkg/api/api.go
```diff
@@ -48,3 +49,16 @@ func New(tags *tags.Tags, storer storage.Storer, corsAllowedOrigins []string, lo
 
 	return s
 }
+
+const (
+	SwarmPinHeader = "Swarm-Pin"
+	TagHeaderUid   = "swarm-tag-uid"
```
janos, line 55: As we changed the format of headers names to Pascal kebeb case, this value can follow that.

pkg/api/dirs.go
```diff
@@ -144,10 +148,57 @@ func storeDir(ctx context.Context, reader io.ReadCloser, s storage.Storer, logge
...
+// this function was extracted from `fileUploadHandler` and should eventually replace its current code
+func storeFile(ctx context.Context, fileInfo *fileUploadInfo, s storage.Storer, mode storage.ModePut) (swarm.Address, error) {
+	v := ctx.Value(toEncryptContextKey{})
+	toEncrypt, _ := v.(bool) // default is false
+
+	// first store the file and get its reference
+	sp := splitter.NewSimpleSplitter(s, mode)
+	fr, err := file.SplitWriteAll(ctx, sp, fileInfo.reader, fileInfo.size, toEncrypt)
+	if err != nil {
+		return swarm.ZeroAddress, fmt.Errorf("split file error: %w", err)
```
janos, line 169: Word error is redundant it the error message. This goes for other errors in this function.

pkg/api/api.go
acud: yeah, i didn't change that because of the open PR from @mortelli that has extensive changes to the tag api. it will introduce more conflicts. That's also why I did not change the variable name (should be `SwarmTagUidHeader`)

pkg/localstore/pin.go
```diff
@@ -65,12 +65,12 @@ func (db *DB) PinnedChunks(ctx context.Context, cursor swarm.Address) (pinnedChu
 }
 
 // Pinner returns the pin counter given a swarm address, provided that the
-// address has to be pinned already.
+// address has been pinned.
```
zelig, line 68: Pinner? in the comment

Issue #534, opened 2020-08-05
Tags persistence


Eknir:
# User story

## Story

As a user of Swarm, I want to refer to my collection of chunks with a `tag-uid`, even after restart of the node, such that I can follow the progress of a big-file upload after the node restarts.

## Acceptance criteria

- I can query the status of an unfinished upload after I restarted my node

## Tasks

| Task | Assignee |  Done (1), or 0 |
| ------ |------------| -----|

Eknir:
@jmozah , fyi

PR #573, opened 2020-08-14
node: add tag persistence across sessions

acud:
pulled from swarm codebase. no tests since no tests in `node` package.

fixes https://github.com/ethersphere/bee/issues/534


janos:
This approach would not preserve tags on panic or forced node termination. Is this intensional? More correct way is to use state store in the tags package for Tags storage backend, not just sync.Map.

acud:
Yes correct, this won't survive the aforementioned cases. I think we somehow converged on this since it would just do a write and a read on every read or change of the state. also, right now the entire sync map is persisted in one go on the shutdown in a single json object serialized into the localstore. otherwise we'd have to use a single state store entry for every tag, it would be better, but means we need to rewrite the `Tags` type. at your discretion. we can do it, just a bit more effort.

janos:
@acud correct, this would need a bit more time to have a reliable persistence. This PR is the improvement. It looks to me that tags.Tags.Delete method is not used anywhere, and over time, loading and saving tags will get longer. This is in essence a memory leak, but now it will be persisted memory leak. :)

acud:
> @acud correct, this would need a bit more time to have a reliable persistence. This PR is the improvement. It looks to me that tags.Tags.Delete method is not used anywhere, and over time, loading and saving tags will get longer. This is in essence a memory leak, but now it will be persisted memory leak. :)

I think the delete method should be used in @mortelli's PR which adds a tag delete endpoint. It indeed is a leak, and I also have concerns that it keeps your "history" of what you uploaded (privacy concern). But, we can improve and add some pruning mechanism (or a `--senile` flag that does not keep track with tags)

pkg/node/node.go
```diff
@@ -471,7 +484,12 @@ func (b *Bee) Shutdown(ctx context.Context) error {
 		errs.add(fmt.Errorf("tracer: %w", err))
 	}
 
-	if err := b.stateStoreCloser.Close(); err != nil {
+	err := b.stateStore.Put("tags", b.tags)
+	if err != nil {
+		errs.add(fmt.Errorf("tag persistence:%w", "err", err))
```
zelig, line 489: remove the "err" argument

zelig:
> @acud correct, this would need a bit more time to have a reliable persistence. This PR is the improvement. It looks to me that tags.Tags.Delete method is not used anywhere, and over time, loading and saving tags will get longer. This is in essence a memory leak, but now it will be persisted memory leak. :)

indeed,  tag persistance is meant to be for resumable uploads, not to serve as user history feature although it can serve as one.

zelig:
on  an unrelated note,  the `anonymous` field should  be removed

jmozah:
Rework on this PR includes
 - Store every tag separately
 - Update every tag 2 times (first when local store is done, second when entire sync is done)
 - on system shutdown, store all the tags that are in memory in statestore ( each tag separately)
 - After bootup don't load any tag, load them if they are accessed
 - Remove tag from memory after the tag states are done and store it
 - If API asks for a tag, check the memory first and then state store to retrieve the tag

cmd/internal/file/io_test.go
```diff
@@ -153,7 +154,9 @@ func TestLimitWriter(t *testing.T) {
 // newTestServer creates an http server to serve the bee http api endpoints.
 func newTestServer(t *testing.T, storer storage.Storer) *url.URL {
 	t.Helper()
-	s := api.New(tags.NewTags(), storer, nil, logging.New(ioutil.Discard, 0), nil)
+	logger := logging.New(ioutil.Discard, 0)
+	store := statestore.NewStateStore()
+	s := api.New(tags.NewTags(store, logger), storer, nil, logging.New(ioutil.Discard, 0), nil)
```
acud, line 159: you don't need another `logging.New` here, reuse `logger`

pkg/file/pipeline/store_writer.go
```diff
@@ -28,10 +28,14 @@ func newStoreWriter(ctx context.Context, l storage.Putter, mode storage.ModePut,
 
 func (w *storeWriter) chainWrite(p *pipeWriteArgs) error {
 	tag := sctx.GetTag(w.ctx)
+	var c swarm.Chunk
 	if tag != nil {
 		tag.Inc(tags.StateSplit)
+		c = swarm.NewChunk(swarm.NewAddress(p.ref), p.data).WithTagID(tag.Uid)
```
acud, line 34: nice catch

pkg/tags/tag.go
```diff
@@ -280,3 +300,21 @@ func decodeInt64Splice(buffer *[]byte) int64 {
 	*buffer = (*buffer)[n:]
 	return val
 }
+
+// UpdateTag update the tag in the state store
+func (tag *Tag) UpdateTag() {
+	key := "tags_" + strconv.Itoa(int(tag.Uid))
+	value, err := tag.MarshalBinary()
+	if err != nil {
+		tag.logger.Warning("tag: ", err)
```
acud, line 309: why swallow the error here? it's better to return the error here so we know if there's a problem

pkg/tags/tag.go
```diff
@@ -280,3 +300,21 @@ func decodeInt64Splice(buffer *[]byte) int64 {
...
+		return
+	}
+
+	if tag.stateStore != nil {
+		err = tag.stateStore.Put(key, value)
+		if err != nil {
+			tag.logger.Warning("tag: ", err)
```
acud, line 316: same

pkg/tags/tags.go
```diff
@@ -143,3 +156,31 @@ func (ts *Tags) UnmarshalJSON(value []byte) error {
 
 	return err
 }
+
+// GetTagFromStore get a given tag from the state store.
+func (ts *Tags) GetTagFromStore(uid uint32) (*Tag, error) {
+	key := "tags_" + strconv.Itoa(int(uid))
```
acud, line 162: it would be good to extract the method that creates the key for the tag into a helper method:
```func getTagKey(uid uint32) string {
   return fmt.Sprintf("tags_%d",uid)
}
```
It is also used in `UpdateTag`

pkg/tags/tag.go
```diff
@@ -280,3 +300,21 @@ func decodeInt64Splice(buffer *[]byte) int64 {
 	*buffer = (*buffer)[n:]
 	return val
 }
+
+// UpdateTag update the tag in the state store
+func (tag *Tag) UpdateTag() {
```
acud, line 305: UpdateTag probably does not need to be exported as it is not used outside the package

pkg/tags/tags.go
```diff
@@ -143,3 +156,31 @@ func (ts *Tags) UnmarshalJSON(value []byte) error {
...
+
+// Close is called when the node goes down. This is when all the tags in memory is persisted.
+func (ts *Tags) Close() (err error) {
+	// store all the tags in memory
+	tags := ts.All()
+	for _, t := range tags {
+		ts.logger.Trace("updaing tag: ", t.Uid)
+		t.UpdateTag()
+	}
+	fmt.Println("closed all tags")
```
acud, line 184: remove Println

pkg/node/node.go
```diff
@@ -399,6 +401,10 @@ func (b *Bee) Shutdown(ctx context.Context) error {
 		errs.add(fmt.Errorf("tracer: %w", err))
 	}
 
+	if err := b.tagsCloser.Close(); err != nil {
+		errs.add(fmt.Errorf("tag persistence:%w", err))
```
janos, line 405: Missing a space between `:` and `%w`.

pkg/tags/tag.go
janos: I agree, and a name like `persistTag` or `saveTag` would be more appropriate.

pkg/tags/tag.go
janos: I agree, error here should be just returned, not logged. This function should return error and should be handler/propagated by the caller.

pkg/tags/tags.go
```diff
@@ -143,3 +156,31 @@ func (ts *Tags) UnmarshalJSON(value []byte) error {
 
 	return err
 }
+
+// GetTagFromStore get a given tag from the state store.
+func (ts *Tags) GetTagFromStore(uid uint32) (*Tag, error) {
```
janos, line 161: This method does not have to be exported. The correct way to get the tag from outside of the package is by the Get method.

pkg/tags/tags.go
```diff
@@ -143,3 +156,31 @@ func (ts *Tags) UnmarshalJSON(value []byte) error {
...
+	key := "tags_" + strconv.Itoa(int(uid))
+	var data []byte
+	err := ts.stateStore.Get(key, &data)
+	if err != nil {
+		return nil, err
+	}
+	var ta Tag
+	err = ta.UnmarshalBinary(data)
+	if err != nil {
+		return nil, nil
```
janos, line 171: Why is error not returned, but only nil?

pkg/tags/tags_test.go
```diff
@@ -52,3 +59,57 @@ func TestAll(t *testing.T) {
...
+	if ta.Total != rcvd1.Total {
+		t.Fatalf("invalid total: expected %d got %d", ta.Total, rcvd1.Total)
+	}
+
+	// See if tag update hapens after syncing is over
+	for i := 0; i < 8; i++ {
+		ta.Inc(StateSent)
+		ta.Inc(StateSynced)
+	}
+	ts.Delete(ta.Uid) // simulate node going down
```
janos, line 95: Instead to use Delete, `ts` should be closed and a new one should be created with the same storage. This is test is not testing the persistence without the actual closing of the tags.

pkg/tags/tag_test.go
```diff
@@ -189,30 +189,40 @@ func TestTagsMultipleConcurrentIncrementsSyncMap(t *testing.T) {
 	n := 100
 	wg := sync.WaitGroup{}
 	wg.Add(10 * 5 * n)
+	errC := make(chan error)
```
acud, line 192: both of these channels are not necessary. you can just call `t.Errorf(..)` on line 206, then just wait on the waitgroup outside the for loop. the extra channels are not needed

jmozah: t.Fatal was complaining... t.Errorf is not... Fixed. Thanks

pkg/tags/tags_test.go
```diff
@@ -52,3 +59,81 @@ func TestAll(t *testing.T) {
...
+
+	// check if the values ae intact after the bootup
+	if ta.Uid != rcvd1.Uid {
+		t.Fatalf("invalid uid: expected %d got %d", ta.Uid, rcvd1.Uid)
+	}
+	if ta.Total != rcvd1.Total {
+		t.Fatalf("invalid total: expected %d got %d", ta.Total, rcvd1.Total)
+	}
+
+	// See if tag is saved after syncing is over
```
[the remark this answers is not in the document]
jmozah: Since updating the tag on every inc() is expensive, @janos and me had a discussion before implementing this and decided that we will store in the following cases
1) When the splitting is over 
2) When the syncing is over
3) When the system does a shutdown

This is so that we can reload the tags when sync is resumed after the bootup.

pkg/tags/tag.go
```diff
@@ -172,15 +191,20 @@ func (t *Tag) Done(s State) bool {
...
+func (t *Tag) DoneSplit(address swarm.Address) (int64, error) {
 	total := atomic.LoadInt64(&t.Split)
 	atomic.StoreInt64(&t.Total, total)
 
 	if !address.Equal(swarm.ZeroAddress) {
 		t.Address = address
 	}
 
-	return total
+	// persist the tag
```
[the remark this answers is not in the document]
jmozah: same as above

pkg/tags/tag.go
```diff
@@ -112,11 +119,23 @@ func (t *Tag) IncN(state State, n int) {
 		v = &t.Synced
 	}
 	atomic.AddInt64(v, int64(n))
+
+	// check if syncing is over and persist the tag
```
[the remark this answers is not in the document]
jmozah: same as above

jmozah:
@zelig The user story says "wants to follow the progress of a big-file upload after the node restarts". The issue we faced during the tests was, the upload is fast.. but the syncing is too slow. So if someone restarts the node before the sync is done, he wants to follow the sync progress after bootup.

@janos and me had a discussion before i started implementation and we decided that we cannot store on every Inc(), instead we can store the tag data when the splitting is over (i.ie when DoneSPlit() is called), when syncing is Over and also during system shutdown. The bullet points of the discussion is captured [here](https://github.com/ethersphere/bee/pull/573#issuecomment-682424833). This make sure that the user can follow the sync progress even after a restart. 


> we should periodically save tags representing partial uploads
    for collection uploads, save manifest too
    within current file roothash, offset and equal counts

How often should we save tags? Why should we save manifest and other information?

> forget/delete synced tags (yes, the opposite of what this PR is doing)

We can delete it as soon as the file is synced... but there is high chance that it will be accessed again soon. in this PR is Get it from the statestore and also store it in memory.. are you saying that i dont store it in memory when it is accessed?

> report partial tags on bootup in interactive mode

Interactive mode?

>  support resumed upload if tag shows partial upload

It is supported now... The partial tags are stored on shutdown and when sync is started a Get is issued and the partial tag is loaded again and syncing continues.

> catch os signals for graceful shutdown

CTRL-C is caught... tags are stored in graceful shutdown. Few other signals like OOM is not caught...


@janos WDYT of the above points?

janos:
It is correct that @jmozah and me concluded that persisting on every Inc call is expensive, especially for a fast changing counters, like split or store. Syncing counter is much slower and persisting on every sync increment would probably not be so noticeable, but that would require some drastic changes in how different counters are treated.

Deleting (synced) tags right away may introduce problem for the ux as when a user would receive that the tag does not exist, should it be considered as synced or unknown.

Tags should be lazy loaded when accessed or incremented, not to load all tags from the store in memory on start, or at least, it is too drastic, as most of them will never be changed or accessed.

There is no need to catch os signals, as we are already doing that at one place when node.Bee.Shutdown is called. This way, it is only required for a component to call its Close in Shutdown to release its resources, handle teardown in the right order in respect to dependants and in this case to do the final store. In short, only Shutdown should be handled correctly and relay that whoever created the node will call it appropriately. In this case, cmd/bee is creating the node and handles os signals, and the lifetime of the node is its responsibility, and the initialization, dependency injection and termination of components is the responsibility of the node.

I have already talked with @jmozah, @acud and @pradovic about my view of Tags implementation architecture which would by my opinion make more maintainable and clear code. Basically, a service that would operate on tags using tag IDs, instead passing around the Tag object. That was considered as too drastic change for the current code state and left for potential improvements later, if possible. Even thought I think that it goes along with the architecture that bee started with.

zelig (approved):

pkg/api/tag.go
```diff
@@ -197,6 +197,12 @@ func (s *server) doneSplit(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	tag.DoneSplit(tagr.Address)
+	_, err = tag.DoneSplit(tagr.Address)
+	if err != nil {
+		s.Logger.Debugf("done split: %v", err)
+		s.Logger.Error("done split: %v", err)
```
janos, line 203: The Error log message should not expose internals. The message should be something like this `"done split failed for address %v", tagr.Address`. Also, the Debugf would be more informative with the address in the message.

pkg/tags/tag.go
```diff
@@ -62,18 +65,22 @@ type Tag struct {
 	StartedAt time.Time     // tag started to calculate ETA
 
 	// end-to-end tag tracing
-	ctx      context.Context  // tracing context
-	span     opentracing.Span // tracing root span
-	spanOnce sync.Once        // make sure we close root span only once
+	ctx        context.Context     // tracing context
+	span       opentracing.Span    // tracing root span
+	spanOnce   sync.Once           // make sure we close root span only once
+	stateStore storage.StateStorer // to persist the tag
```
janos, line 71: We are now passing the database object through the context, which brings its own problems and is discouraged to do so. I think that tags would benefit from a different architecture where only the tag id would be passed in the context, and a tag service would be responsible for getting, incrementing and storing counters.

janos (approved):
The most significant comment is how tag is now passed as context value with the state store in it, which is not recommended. As I already wrote, I find that the current implementation can be improved, but I am approving this PR, nevertheless, as for that change an architecture proposal is needed.

pkg/tags/tag.go
jmozah: I agree. But this is out of the scope of this PR.

Issue #477, opened 2020-07-27
clean up jsonhttptest signatures


mortelli:
Based on the work done in #472, it is apparent we need to work on the signatures for the functions in `jsonhttptest.go` for readability's sake.

Here's what we have right now:
```
- ResponseDirect
- ResponseDirectWithMultiPart
- ResponseDirectCheckBinaryResponse
- ResponseDirectSendHeadersAndReceiveHeaders
- ResponseDirectSendHeadersAndDontCheckResponse
- ResponseUnmarshal
- request
```

There are no name suggestions as of now, but let's aim at having something more readable and self-explanatory.

We could also refactor these into different functions.

PR #577, opened 2020-08-14
cleanup jsonhttptest package

janos:
This PR consolidates all functions in jsonhttptest package that piled up over time bloating its api. Unfortunately, only the initial two functions were tested, non of the followup were, so new tests are added and existing one improved. Also the package is better documented.

Tests in api and debugapi packages are adjusted.

fixes: #477

### DIFF
--- a/pkg/api/chunk_test.go
+++ b/pkg/api/chunk_test.go
@@ -97,7 +97,7 @@ func TestChunkUploadDownload(t *testing.T) {
 				Message: http.StatusText(http.StatusOK),
 				Code:    http.StatusOK,
 			}),
-			jsonhttptest.WithRequestHeader(api.SwarmPinHeader, "hdgdh"),
+			jsonhttptest.WithRequestHeader(api.SwarmPinHeader, "invalid-pin"),
 		)
 
 		// Also check if the chunk is NOT pinned
--- a/pkg/api/chunk_test.go
+++ b/pkg/api/chunk_test.go
@@ -44,30 +44,39 @@ func TestChunkUploadDownload(t *testing.T) {
 	)
 
 	t.Run("invalid hash", func(t *testing.T) {
-		jsonhttptest.ResponseDirect(t, client, http.MethodPost, resource(invalidHash), bytes.NewReader(validContent), http.StatusBadRequest, jsonhttp.StatusResponse{
-			Message: "chunk write error",
-			Code:    http.StatusBadRequest,
-		})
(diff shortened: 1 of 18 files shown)


pkg/jsonhttp/jsonhttptest/testing_mock_test.go
```diff
@@ -0,0 +1,71 @@
...
+package jsonhttptest_test
+
+import (
+	"errors"
+	"fmt"
+	"testing"
+)
+
+// assert is a test helper that validates a functionality of another helper
+// function by mocking Errorf, Fatal and Helper methods on testing.TB.
```
mortelli, line 14: are you sure this should read `testing.TB`?

pkg/api/chunk_test.go
```diff
@@ -82,37 +91,43 @@ func TestChunkUploadDownload(t *testing.T) {
...
-			Message: http.StatusText(http.StatusOK),
-			Code:    http.StatusOK,
-		}, headers)
+		jsonhttptest.Request(t, client, http.MethodPost, resource(validHash), http.StatusOK,
+			jsonhttptest.WithRequestBody(bytes.NewReader(validContent)),
+			jsonhttptest.WithExpectedJSONResponse(jsonhttp.StatusResponse{
+				Message: http.StatusText(http.StatusOK),
+				Code:    http.StatusOK,
+			}),
+		)
```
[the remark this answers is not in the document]
janos: Yes headers are just empty map, so it works. The explicit set of an empty map had no use in the previous version of this test.

pkg/jsonhttp/jsonhttptest/testing_mock_test.go
janos: I am using testing.TB here in assert and mock for consistency with the Request function which uses testing.TB interface in order to inject the mock. Type mock can use testing.T struct, as well. I can change it if you think that it would be better.

janos:
Thank you @mortelli.

Yes, there is a redundancy in providing the http status code and expecting StatusResponse in the body. I went with explicit set on all values http status code, response message and response code for clarity even it introduces more code. I could not find a nice way to make the test obvious that expected json-encoded body is expected to be StatusResponse. Another reason for not assuming the StatusResponse code and http status code to be the same is the possibility to have our own codes which are >= 600. Sometimes APIs want to communicate different messages and provide specific codes for easier use by the clients. I am not sure if we will have such codes, but there is nothing that stands in a way to have them.

PR #578, opened 2020-08-15
api: communicate tag uid right away

### LINKED ISSUE #492: Communicate tagUID right away
# User story

## Story

As a user of Swarm who is uploading a big file, I want to know the status of my file throughout the whole update process, such that I know that my node is working and I wait patiently.

## Acceptance criteria

- User can figure out the state of the upload throughout the whole process. For this to work, he must know the tag immediately after sending the request to upload data.

## Background

This issue came along after trying out to use Swarm for uploading a bigger file. It took some time, so we figured out it might be handy to verify the state of the upload by using tags. The first issue encountered was https://github.com/ethersphere/bee/pull/490, when using explicit tags. When not using explicit tags, Swarm is auto-generating a tag, but this tag is only communicated to the user when chunking is done. Hence, I don't know how the process of chunking is proceeding.

More information in a chat with @acud :

## Tasks

acud:
works with `curl`. not sure if we can/want to unit test this as a separate feature (it will also be a bit tricky to correctly unit test)

closes #492

### DIFF
--- a/pkg/api/bytes.go
+++ b/pkg/api/bytes.go
@@ -30,6 +30,11 @@ func (s *server) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 		jsonhttp.InternalServerError(w, "cannot get or create tag")
 		return
 	}
+	w.Header().Set(SwarmTagUidHeader, fmt.Sprint(tag.Uid))
+	w.WriteHeader(http.StatusContinue)
+	if f, ok := w.(http.Flusher); ok {
+		f.Flush()
+	}
 
 	// Add the tag to the context
 	ctx := sctx.SetTag(r.Context(), tag)
@@ -46,7 +51,6 @@ func (s *server) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 	if created {
 		tag.DoneSplit(address)
 	}
-	w.Header().Set(SwarmTagUidHeader, fmt.Sprint(tag.Uid))
 	w.Header().Set("Access-Control-Expose-Headers", SwarmTagUidHeader)
 	jsonhttp.OK(w, bytesPostResponse{
 		Reference: address,
--- a/pkg/api/chunk.go
+++ b/pkg/api/chunk.go
@@ -40,6 +40,12 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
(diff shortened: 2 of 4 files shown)


zelig:
so does this mean you no longer need to pregenerate the tag to monitor split/store instead you can use the auto generated tag uid from response header?

acud:
not necessarily @zelig, but it _does_ mean that you would be able to monitor the chunking and storing before the root chunk is returned. but anyway i'm not sure if it was too early to pop the champagnes. the tests are failing and i'm not sure if i can really get this to work without implementing my own response writer

pkg/api/chunk.go
```diff
@@ -40,6 +40,12 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	w.Header().Set(SwarmTagUidHeader, fmt.Sprint(tag.Uid))
```
zelig, line 43: DRY these 5 lines into a function call, no?

pkg/api/bytes.go
```diff
@@ -30,6 +30,11 @@ func (s *server) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 		jsonhttp.InternalServerError(w, "cannot get or create tag")
 		return
 	}
+	w.Header().Set(SwarmTagUidHeader, fmt.Sprint(tag.Uid))
+	w.WriteHeader(http.StatusContinue)
```
janos, line 34: Why setting the status to Continue as at the end it is written OK?

Status should be written only once.

acud: AFAIK `1xx` status codes can be sent before the final response code has been established. Even the golang built in http package actually is supposed to send a 100 continue while the body of the request is being read

acud: Hm. Ok. So, sending a 1xx status code before the end of the response is permitted by the HTTP protocol spec, but the golang http server doesn't support it. Except for the fact that it automatically sends a 100-continue automatically when reading the request body. But even if I write this extra header before the 100 continue is sent, the header doesn't get flushed with the 100 continue :/ I will continue testing this. Might be worth to open an issue on the golang repo for this

acud: closing this for now as with the built-in http server this is not possible.

https://github.com/golang/go/issues/40838

acud:
closing this for now as with the built-in http server this is not possible.

https://github.com/golang/go/issues/40838

PR #580, opened 2020-08-17
Always load balance from store

ralph-pichler:
This PR removes the in-memory copy of accounting balances (as suggested by @janos). Balances are always loaded from disk when needed. The `balances` struct has been renamed to `accountingPeer` as it no longer holds the balance. Instead it is only for in-memory data (such as the reserved balance and the payment threshold). The accounting logic itself remains unchanged.

### DIFF
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -262,10 +262,10 @@ func (a *Accounting) Debit(peer swarm.Address, price uint64) error {
 // Balance returns the current balance for the given peer
 func (a *Accounting) Balance(peer swarm.Address) (balance int64, err error) {
 	err = a.store.Get(peerBalanceKey(peer), &balance)
-	if err == storage.ErrNotFound {
-		return 0, nil
-	}
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/accounting/accounting.go
```diff
@@ -239,76 +260,43 @@ func (a *Accounting) Debit(peer swarm.Address, price uint64) error {
 }
 
 // Balance returns the current balance for the given peer
-func (a *Accounting) Balance(peer swarm.Address) (int64, error) {
-	peerBalance, err := a.getPeerBalance(peer)
+func (a *Accounting) Balance(peer swarm.Address) (balance int64, err error) {
+	err = a.store.Get(peerBalanceKey(peer), &balance)
+	if err == storage.ErrNotFound {
+		return 0, nil
+	}
```
janos, line 267: This can go inside `err != nil` (very minor optimization). Also, errors.Is would be preferred on the imported package error check.

ralph-pichler: changed

Issue #581, opened 2020-08-17
upload does not terminate when ctrl-c curl


acud:
when uploading from curl, if the curl process is exited with `ctrl-c`, the upload on the bee node does not terminate correctly and the chunker hangs, which in turns does not allow the bee client to close correctly if it is then `ctrl-c`d too

PR #582, opened 2020-08-17
seekjoiner: add pluggable decryption

acud:
This PR adds pluggable decryption to the joiner. Actually, there are no changes needed for the joiner apart from intercepting the initial address size and iterating the hash trie according to the initial reference length.
The PR base has been changed to the new pipeline encryption branch which it is based on.

fixes: https://github.com/ethersphere/bee/issues/581


pkg/api/bytes_test.go
```diff
@@ -67,9 +67,9 @@ func TestBytes(t *testing.T) {
 	})
 
 	t.Run("not found", func(t *testing.T) {
-		jsonhttptest.ResponseDirect(t, client, http.MethodGet, resource+"/abcd", nil, http.StatusNotFound, jsonhttp.StatusResponse{
-			Message: "not found",
-			Code:    http.StatusNotFound,
+		jsonhttptest.ResponseDirect(t, client, http.MethodGet, resource+"/abcd", nil, http.StatusBadRequest, jsonhttp.StatusResponse{
+			Message: "invalid root chunk",
+			Code:    http.StatusBadRequest,
```
janos, line 72: This should be a valid request, just the chunk is not found. I think that 404 is correct to return.

pkg/encryption/decrypt_store.go
```diff
@@ -0,0 +1,98 @@
...
+	storage.Getter
+}
+
+func NewDecryptingStore(s storage.Getter) storage.Getter {
+	return &decryptingStore{s}
+}
+
+func (s *decryptingStore) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Address) (ch swarm.Chunk, err error) {
+	switch l := len(addr.Bytes()); l {
+	case 32:
```
janos, line 27: Should here be a constant instead literal?

pkg/encryption/decrypt_store.go
```diff
@@ -0,0 +1,98 @@
...
+		// normal, unencrypted content
+		return s.Getter.Get(ctx, mode, addr)
+
+	case 64:
```
janos, line 31: Here also.

pkg/encryption/decrypt_store.go
```diff
@@ -0,0 +1,98 @@
...
+		// encrypted reference
+		ref := addr.Bytes()
+		address := swarm.NewAddress(ref[:32])
+		ch, err := s.Getter.Get(ctx, mode, address)
+		if err != nil {
+			return nil, err
+		}
+
+		decryptionKey := make([]byte, KeyLength)
+		copy(decryptionKey, ref[32:])
```
janos, line 41: Is the copy needed? Should be safe to pass ref[32:] to decryptChunkData as it does not mutate the slice.

pkg/file/seekjoiner/internal/job.go
```diff
@@ -21,18 +21,20 @@ type SimpleJoinerJob struct {
...
 	levels     int
+	refLength  int
 
 	ctx    context.Context
 	getter storage.Getter
 }
 
 // NewSimpleJoinerJob creates a new simpleJoinerJob.
-func NewSimpleJoinerJob(ctx context.Context, getter storage.Getter, rootChunk swarm.Chunk) *SimpleJoinerJob {
+func NewSimpleJoinerJob(ctx context.Context, getter storage.Getter, refLength int, rootChunk swarm.Chunk) *SimpleJoinerJob {
```
janos, line 31: refLength can be measured from rootChunk here. No need to pass both rootChunk and refLength.

acud:
encryption seems to behave strangely above a certain content span. i will revert the encryption in the splitter and also use the same pattern to do the encryption in the splitter and see if the problem persists. it is very difficult to debug

pkg/encryption/decrypt_store.go
```diff
@@ -0,0 +1,144 @@
...
+
+package encryption
+
+import (
+	"context"
+	"encoding/binary"
+	"fmt"
+	"hash"
+
+	"github.com/ethersphere/bee/pkg/storage"
```
zelig, line 13: how come encryption depends on storage?

pkg/encryption/decrypt_store.go
```diff
@@ -0,0 +1,144 @@
...
+
+func hashFunc() hash.Hash {
+	return sha3.NewLegacyKeccak256()
+}
+
+type decryptingGetter struct {
+	storage.Getter
+}
+
+func NewDecryptingGetter(s storage.Getter) storage.Getter {
```
zelig, line 26: i think thereshould be an encrypted store in storage rather. this pkg should not depend on storage

pkg/encryption/decrypt_store.go
```diff
@@ -0,0 +1,144 @@
...
+			return nil, err
+		}
+		return swarm.NewChunk(address, d), nil
+
+	default:
+		return nil, storage.ErrReferenceLength
+	}
+}
+
+type Encrypting struct {
```
zelig, line 63: why this complexity?

nolash (commented):
I won't review this PR, but one thing caught my eye at a glance; the term "reference size" is used in the filehasher, I presume to perform the calculations of spans. In the previous filehasher work (which is still unreviewed and still in the swarm repo) any chainable hasher would define what _sectionsize_ it operates with, and setting/getting this was an interface. `sectionsize` seems to me to be a more correct term, since the filehasher really only constructs proofs of the data, not caring really whether the sections it's hashing are symkeys, references or whatever. I believe using the term reference here suggests an unnecessary and even faulty coupling to higher level concepts.

acud:
@nolash @zelig this PR is in progress, due to bugs I found with encryption I had to stop working on this (hence the `in progress` label). Once the problems with encryption are solved I will reiterate on this one again and invite your reviews. The PR scope is only to change the joiner, not splitter. Thanks

pkg/file/seekjoiner/internal/job.go
acud: well not entirely, because the root chunk we have here is the root chunk that came from a retrieve request. when we retrieve the root chunk of encrypted content we retrieve the content addressed chunk, we don't concatenate the decryption key to it. i'd leave it as is for now

pkg/api/bytes_test.go
acud: the inner error is `invalid reference length` which is due to some more strict checks that i am now doing further down the call stack. the resource is `/abcd` so i actually think that `bad request` is legible here. if you insist i'll change it. the message should not be 'invalid root chunk' tho. probably just `bad request`

zelig (approved):
LGTM  
see https://github.com/ethersphere/bee/pull/618#issuecomment-681036905

zelig (changes requested):
actually my previously  left comment  still valid.
decryption 'store' should be parallel to encryption pipeline and not polluting  the  encryption pkg

pkg/api/file.go
```diff
@@ -309,7 +304,7 @@ func (s *server) downloadHandler(w http.ResponseWriter, r *http.Request, referen
 		}
 		s.Logger.Debugf("api download: invalid root chunk %s: %v", reference, err)
 		s.Logger.Error("api download: invalid root chunk")
-		jsonhttp.BadRequest(w, "invalid root chunk")
+		jsonhttp.NotFound(w, "not found")
```
janos, line 307: Instead of `"not found"`, nil can be passed.

pkg/encryption/decrypt_store.go
```diff
@@ -0,0 +1,96 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
janos, line 1: I agree that decryptingStore should be in its own package like `pkg/encrytpion/store`. And alos to depend on encrytion.Interface in order to be tested without testing the encryption, but just the store.

pkg/file/seekjoiner/internal/job.go
janos: You are saying that the refLength there does not represent the length of the rootChunk.Address()? I see that this function is used in https://github.com/ethersphere/bee/pull/582/files#diff-f069298b920c3d61a6ec16bace277a31R63 and it would be very strange that address and rootChunk.Address() are different in simpleJoiner.Join function where getter.Get is used.

I think that it need to be clarified what is refLength is expected to be if it is different then rootChunk.Address().

acud: so... the problem is the following:
you get a request to join the whole thing, which is address+key, but you need the root chunk, so you ask it from the retrieval protocol, which brings it back, but it does not bring it back with the key, since it is already decrypted. so you get back from the retrieve protocol the chunk with the content addressed hash of the encrypted content (no key) and the content. i am a bit afraid to augment the key into the chunk decryption key along with the decrypted content since it kind of takes away the purity of the chunk type and adds more information to it, since in this case Address is not just Address but Address+Key, which is kind of not so great. I can intercept the chunk before it is sent to the joiner and create a new chunk, and augment the decryption key there, but i find this a bit ugly

janos (approved):
Thanks @acud for addressing my comments.

pkg/encryption/store/decrypt_store.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+
+		d, err := decryptChunkData(ch.Data(), ref[swarm.HashSize:])
+		if err != nil {
+			return nil, err
+		}
+		return swarm.NewChunk(address, d), nil
+
+	default:
+		return nil, storage.ErrReferenceLength
```
zelig, line 48: you  see?this whole thing belongs to under `storage/encryption`  and encryption as used in pipeline should be a  putter, maybe lets sort this all out in a later PR?

pkg/encryption/store/decrypt_store.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+func decryptChunkData(chunkData []byte, encryptionKey encryption.Key) ([]byte, error) {
+	if len(chunkData) < swarm.SpanSize {
```
zelig, line 53: this is  impossible right?

pkg/encryption/store/decrypt_store.go
```diff
@@ -0,0 +1,97 @@
...
+
+	c := make([]byte, length+8)
+	copy(c[:8], decryptedSpan)
+	copy(c[8:], decryptedData[:length])
+
+	return c, nil
+}
+
+func decrypt(chunkData []byte, key encryption.Key) ([]byte, []byte, error) {
+	encryptedSpan, err := newSpanEncryption(key).Encrypt(chunkData[:swarm.SpanSize])
```
zelig, line 79: should this not be Decrypt instead of Encrypt? lol because of symmetry it works but super confusing

pkg/encryption/store/decrypt_store.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	encryptedData, err := newDataEncryption(key).Encrypt(chunkData[swarm.SpanSize:])
```
zelig, line 83: should this not be `Decrypt` instead of Encrypt? lol  because of symmetry  it works but super confusing

pkg/encryption/store/decrypt_store.go
zelig: var should be `decryptedSpan` or well... just `span`

pkg/encryption/store/decrypt_store.go
zelig: var should be just `data`?

pkg/file/seekjoiner/joiner_test.go
```diff
@@ -114,3 +166,62 @@ func TestJoinerWithReference(t *testing.T) {
...
+				if err != nil && err != io.EOF {
+					t.Fatal(err)
+				}
+				copy(totalGot[index:], resultBuffer[:n])
+				index += n
+			}
+
+			if !bytes.Equal(testData, totalGot) {
+				t.Fatal("input data and output data does not match")
+			}
```
zelig, line 224: you do not test the equality of data.

pkg/file/seekjoiner/joiner_test.go
```diff
@@ -7,32 +7,84 @@ package seekjoiner_test
...
+	var err error
+	_, _, err = joiner.Join(ctx, swarm.ZeroAddress)
+	if !errors.Is(err, storage.ErrReferenceLength) {
+		t.Fatalf("expected ErrReferenceLength %x but got %v", swarm.ZeroAddress, err)
+	}
+}
+
 // TestJoiner verifies that a newly created joiner returns the data stored
 // in the store when the reference is one single chunk.
 func TestJoinerSingleChunk(t *testing.T) {
```
zelig, line 43: why not have this test with encryption too?

pkg/file/seekjoiner/joiner_test.go
```diff
@@ -114,3 +166,62 @@ func TestJoinerWithReference(t *testing.T) {
...
+			store := mock.NewStorer()
+			joinner := seekjoiner.NewSimpleJoiner(store)
+
+			g := mockbytes.New(0, mockbytes.MockTypeStandard).WithModulus(255)
+			testData, err := g.SequentialBytes(tt.chunkLength)
+			if err != nil {
+				t.Fatal(err)
+			}
+			ctx := context.Background()
+			pipe := pipeline.NewPipelineBuilder(ctx, store, storage.ModePutUpload, true)
```
zelig, line 194: this true means you encrypt the data right? maybe worth a comment

zelig (changes requested):

pkg/encryption/store/decrypt_store.go
```diff
@@ -0,0 +1,97 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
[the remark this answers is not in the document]
acud: its in a separate package `store` under `encryption` package. the dependency you're describing does no exist

pkg/file/seekjoiner/joiner_test.go
acud: which comment? the test name describes that it tests for encryption, and the bool parameter in NewPipelineBuilder is called `encrypt` do we need more explicit comments here?

pkg/encryption/store/decrypt_store.go
acud: encryption cannot be a putter, since before you put the encrypted content you must hash it through bmt

pkg/api/bytes_test.go
```diff
@@ -72,7 +72,7 @@ func TestBytes(t *testing.T) {
 	t.Run("not found", func(t *testing.T) {
 		jsonhttptest.Request(t, client, http.MethodGet, resource+"/abcd", http.StatusNotFound,
 			jsonhttptest.WithExpectedJSONResponse(jsonhttp.StatusResponse{
-				Message: "not found",
+				Message: "Not Found",
```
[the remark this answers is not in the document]
acud: this is the built in jsonhttp message....

pkg/file/seekjoiner/joiner_test.go
acud: this is covered in the encryption test

pkg/api/bytes_test.go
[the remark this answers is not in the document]
janos: Becasue it is http.StatusNotFound message.

janos:
I can confirm that this PR fixes https://github.com/ethersphere/bee/issues/581.

zelig (approved):

PR #585, opened 2020-08-17
announce payment threshold through pricing protocol

### LINKED ISSUE #417: Payment threshold and disconnects
# Task

Instruct settlement to do a payment if threshold is met. Disconnect if disconnect threshold is exceeded. A mock settlement layer can be used here.

- [x] `Settlement` interface with a mock implementation where settlement always works
  - [x] Create a simple protocol with a pay message containing the value
  - [x] Adjust accounting balance on received pay message
- [x] Introduce accounting/pricing protocol
  - [x] Communicate payment threshold to connected peers
  - [x] Introduce per-peer payment threshold (their threshold)
  - [x] Adjust peers payment threshold on received message
- [x] Changes to `Accounting`
  - [x] Introduce a fixed payment threshold (for now)
  - [x] Call `Settlement` if payment threshold exceeded or if we are close (depends on details of `Reserve`)
  - [x] Adjust `Reserve` to work with peers payment threshold
  - [x] Return error in `Add` if disconnect threshold is met
  - [x] Instead of configurable disconnect threshold, use payment treshold plus tolerance

## Acceptance criterea

Peers exchange payment threshold and pay each other with fake payment to restore balance.

ralph-pichler:
closes #417


Eknir:
@ralph-pichler , given that this one is still in draft, I assume you are not aiming to have it in for the next release, right?

ralph-pichler:
> @ralph-pichler , given that this one is still in draft, I assume you are not aiming to have it in for the next release, right?

yes, since this depends on some yet to do changes in the protocol layer (init function) I would not include it in the release.

janos:
> > @ralph-pichler , given that this one is still in draft, I assume you are not aiming to have it in for the next release, right?
> 
> yes, since this depends on some yet to do changes in the protocol layer (init function) I would not include it in the release.

Yes, sorry for that. There was some short discussion on the core channel https://beehive.ethswarm.org/swarm/pl/isjam1ro9fnpjfssi7d61p6scc about the Init function in ProtocolSpec, but not a final conclusion if it is a good approach. I know that @pradovic liked the idea, but also would like to her if @acud would approve that. Somehow this was left unresolved. @acud would it be ok to proceed with the architectural proposal on Init function in ProtocolSpec? I am asking you mainly as you implemented topology.Notifier interface which is similar but does notifications only when a peer connects to node, not when the node makes a connection to a peer, as Connect is used and not ConnectNotify.

pkg/accounting/accounting.go
```diff
@@ -60,6 +60,7 @@ type Options struct {
 	Logger           logging.Logger
 	Store            storage.StateStorer
 	Settlement       settlement.Interface
+	Pricing          pricing.Interface
```
pradovic, line 63: I would move all fields out of the Options struct that are not really optional.

pkg/accounting/accounting.go
```diff
@@ -495,3 +496,26 @@ func addI64pU64(a int64, b uint64) (result int64, err error) {
...
+
+	accountingPeer.lock.Lock()
+	defer accountingPeer.lock.Unlock()
+
+	accountingPeer.paymentThreshold = paymentThreshold
+	return nil
+}
+
+// Connected is called when a peer connects to us
+func (a *Accounting) Connected(ctx context.Context, peer swarm.Address) error {
```
pradovic, line 515: How is this going to be triggered?

pkg/accounting/accounting.go
```diff
@@ -55,13 +55,6 @@ type accountingPeer struct {
 
 // Options are options provided to Accounting.
 type Options struct {
```
ralph-pichler, line 57: if it's empty drop this entirely

pkg/accounting/accounting_test.go
```diff
@@ -662,3 +599,78 @@ func TestAccountingInvalidPaymentTolerance(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = acc.NotifyPaymentThreshold(peer1Addr, testPaymentThreshold)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// TODO: check the success somehow
```
pradovic, line 675: todo?

pkg/node/node.go
```diff
@@ -282,19 +283,15 @@ func NewBee(addr string, swarmAddress swarm.Address, keystore keystore.Service,
...
 
-	acc, err := accounting.NewAccounting(accounting.Options{
-		Logger:           logger,
-		Store:            stateStore,
-		PaymentThreshold: o.PaymentThreshold,
-		PaymentTolerance: o.PaymentTolerance,
-		EarlyPayment:     o.PaymentEarly,
-		Settlement:       settlement,
-	})
+	pricing := pricing.New(p2ps, logger, o.PaymentThreshold)
```
pradovic, line 286: Should `p2ps.AddProtocol()` be called for pricing as well?

pradovic (commented):
Nice work, code looks very nice. Left some questions about some stuff that might be missing, if I am not wrong.

paxthemax (approved):
Code looks good to go as far as I'm concerned.

pkg/node/node.go
ralph-pichler: added

pkg/accounting/accounting_test.go
ralph-pichler: test completed

pkg/accounting/accounting.go
ralph-pichler: removed

pkg/accounting/accounting.go
ralph-pichler: removed

pradovic (approved):
👍

PR #588, opened 2020-08-18
accounting in push sync

ralph-pichler:
This PR integrates push sync with accounting. Nodes pay their peers after receiving the receipt.

The accounting flow is as follows:

During handling a chunk delivery:
* if we are the closest we debit the pushing peer with our price after sending the receipt (in `handleDeliveryResponse`)
* if we need to forward we credit (with the corresponding reserve/release calls) the closest peer with their chunk price after receiving the receipt . after sending back the receipt to the pusher we debit it with our price (at the end of the `handler`) 

During initiating a push:
* if we are not the closest node, the closest node is credited with their price for the chunk after receiving the receipt (with the corresponding reserve/release calls before sending any messages).

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -276,7 +276,7 @@ func NewBee(addr string, logger logging.Logger, o Options) (*Bee, error) {
 	}
 	retrieve.SetStorer(ns)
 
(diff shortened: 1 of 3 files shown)


zelig (approved):

Eknir (approved):
I can give an approval for the logic. Did not look specifically into the code implementation.

janos (approved):
LGTM

Issue #463, opened 2020-07-22
remove redundant entries in upload/download


mortelli:
because the file upload and download functionalities were implemented before manifests, the current code uses variables from `collection/entry` to reference metadata about a file (before referencing its actual content).

now that we have json manifests, we have a redundant entry in the resulting references:

> dir manifest → `manifest` entry →  `collection` entry → file content

consequently, `entry.Entry` variables should be removed from the code and just `manifest` entries should be use in their stead.

this issue is a technical debt follow-up issue stemming from #452 and #447.

mortelli:
we've discussed this further with with @zbiljic and we're not entirely sure about how to proceed with this issue.

we'll try to set up a call with @zelig to sync about this.

Eknir:
@zbiljic , could you tell me what is the status here? Is this something we want to resolve before the next release?

zbiljic:
@Eknir This is related to the old discussion we had (little over a month ago), that I explained more detailed here: https://github.com/ethersphere/bee/pull/591#discussion_r473879567.

My initial implementation was one way ('bytes'), after which we decided on another ('file'), and this issue was for asking the question whether it should be returned to previous way ('bytes'), for storing content of manifest-related files.

Current way is storing files as through the 'files' API, which is by itself fine.

More detailed discussion is needed on whether we want to leave it like this, or change it to storing file bytes (as is using 'bytes' API).

Eknir:
@zbiljic : to me you are the technical owner of this part. I recall a discussion with @mortelli, but not with @zelig about that. At that time, nobody was really sure what to do. You had some arguments for `bytes`, and @mortelli had some arguments for using `files`. If, in hindsight, the decision is wrong, please feel free to add it back.

zbiljic:
@Eknir There was not wrong decision. Each of us had arguments for one or the other side. I had no special idea whether one or other is better. In the end on the meeting with Viktor we decided to go with `files` API.
We can discuss about changes.

PR #591, opened 2020-08-19
Add support for multiple manifest types

zbiljic:
Builds upon ethersphere/manifest#6

(will be updated after that PR is merged)

Local 'manifest' package (with interfaces) is used, and there external implementations have instance, which can be used in rest of the project. In this case only API package.

Here initial "simple" implementation is used, which mostly replaces existing code (without the need for extra headers and name found in previous code).

Different manifest implementations are differentiated by content type metadata of the manifest file. Here the simple implementation uses `application/bzz-manifest-simple+json` content type (open for comments).


janos (commented):
Very nice!

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/coreos/go-semver v0.3.0
 	github.com/davidlazar/go-crypto v0.0.0-20200604182044-b73af7476f6c // indirect
 	github.com/ethersphere/bmt v0.1.2
+	github.com/ethersphere/manifest v0.0.0-20200819112455-88c44943e28f
```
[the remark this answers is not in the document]
zbiljic: This must be changed once that other PR is merged.
In regards to versioning, I am for it, however we have hash anyway.

pkg/api/dirs.go
```diff
@@ -133,42 +141,54 @@ func storeDir(ctx context.Context, reader io.ReadCloser, s storage.Storer, mode
...
 		logger.Tracef("uploaded dir file %v with reference %v", filePath, fileReference)
 
-		// create manifest entry for uploaded file
-		headers := http.Header{}
-		headers.Set("Content-Type", contentType)
-		fileEntry := jsonmanifest.NewEntry(fileReference, fileName, headers)
+		// add file entry to dir manifest
+		err = dirManifest.Add(filePath, manifest.NewEntry(fileReference))
+		if err != nil {
+			return swarm.ZeroAddress, fmt.Errorf("add to manifest error: %w", err)
```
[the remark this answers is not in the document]
zbiljic: Copied from some previous code, will change.

janos (approved):
LGTM 👍

pkg/api/bzz.go
```diff
@@ -84,55 +79,35 @@ func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// we are expecting manifest Mime type here
-	if ManifestContentType != metadata.MimeType {
+	m, err := manifest.NewManifestReference(
+		ctx,
```
acud, line 83: nitpicking but we usually don't have line feeds between parameters on a method call

pkg/manifest/manifest.go
```diff
@@ -5,36 +5,95 @@
...
+	// Lookup returns a manifest entry if one is found in the specified path.
+	Lookup(string) (Entry, error)
+	// Store stores the manifest, returning the resulting address.
+	Store(storage.ModePut) (swarm.Address, error)
 }
 
 // Entry represents a single manifest entry.
 type Entry interface {
-	// Reference returns the address of the file in the entry.
+	// Reference returns the address of the file.
```
acud, line 42: so now a manifest entry only knows to represent an address and that's it?

pkg/manifest/manifest.go
```diff
@@ -5,36 +5,95 @@
...
+	ctx context.Context,
+	encrypted bool,
+	storer storage.Storer,
+) (Interface, error) {
+	return NewManifest(ctx, DefaultManifestType, encrypted, storer)
+}
+
+// NewManifest creates a new manifest.
+func NewManifest(
+	ctx context.Context,
```
acud, line 57: see comment above (this goes for the entire PR)

pkg/manifest/simple.go
```diff
@@ -0,0 +1,142 @@
...
+const (
+	// ManifestSimpleContentType represents content type used for noting that
+	// specific file should be processed as 'simple' manifest
+	ManifestSimpleContentType = "application/bzz-manifest-simple+json"
+)
+
+type simpleManifest struct {
+	manifest simple.Manifest
+
+	ctx       context.Context
```
acud, line 30: i am not entirely sure this is desirable to have on the manifest level. you might create a manifest and pass it on to some other calling context to handle and that way you lose control over cancellations in deeper calls in the cascade

acud (approved):

pkg/api/bzz.go
zbiljic: I opted for this (on several places) because I figured it would be better (easier to read and compare) longer lines.
There is no official requirement, however my thoughts were that once the line goes over 100 characters it might be smart to consider other options (personally I try to go to 80). I also find limited line easier for terminal (but this might be due to my configuration).
If we are to decide to push single-line only, that is fine by me, we can add it to `CODING.md`, and I will change these.

pkg/manifest/manifest.go
zbiljic: Yes, that is the change (that I mentioned in the description of this PR).

Name, as you previously noted should be removed as it is a bit pointless in this structure, can just be extracted from the path using `filepath.Base` at any point. As for the headers go, only header that is being added is the `Content-Type` header during upload (again, based on file name using `mime.TypeByExtension`).

Both of these things were also being saved in the actual file on the Swarm in the special `entry.Entry` object, with reference to `entry.Metadata` which contains filename and mime type.

At the moment there is no way (at least I count not find one) to modify the name of already uploaded file after the fact. Also there is no way to add additional headers to the file, or modify existing manifest (through some API) to add additional headers.

Regarding uploading TAR collection using `/dirs` API: Marcelo and myself have been working on separate parts for "website support", and while I added initial manifest implementation (that had 'name' and 'headers') and serving content via `/bzz` API, actual expectation I had was that in that case content would be saved as is using 'bytes' API, while Marcelo used 'files' API (saving files with metadata) on the `/dirs` API side.
During the merge time we had a meeting with Rinke and Viktor regarding this, and decision was to use 'files' for storing files, meaning manifest will contain references to files, which themselves already have metadata which was initially added in manifest. Since it is already located there, I see no point in having it here also.

Now, we could change this back and say that we want to save only bytes for manifest uploaded files, in which case it is easy to return this to previous state.

P.S.
Important note for this is that the initial `mantaray` implementation does not have support for complex content inside. This can easily be changed though.
Also I would like to note that this fact was not the reason for removing fields.

pkg/manifest/simple.go
zbiljic: Personally I never saw manifest as something that was suppose to be shared in any special way around the code, apart from few simple operations with it. Until now there were not usages of the context in regards with manifest.

As far as the context goes, it is only ever used to save or load manifest.
Load will happen during actual constructor, while the save might be separated, however it could not devise a way to initialize manifest at single point, add stuff from multiple context-es, and save in some completely other one.

It will be easy to change this anyway.

Eknir:
@zbiljic , as far as I understood, this will be for the next release. Is this still the case?

zbiljic:
@Eknir I cannot give answer to that question.

As mentioned in the plan there are several steps that were intended to be completed. This is just one (of several) PR's.
Multiple questions/concerns were raised, and I do not know what are design decisions regarding some parts.
I plan to write architectural proposal explaining our state, and possible directions, so that we have somewhere to discuss, and finalize decisions.

As far as the amount of time needed for each PR, it depends on multiple factors, and that is also one of things that might decide whether this is completed in time.

Eknir:
@zbiljic , thanks for the quick answer. I see this PR already has 2 approvals, so that's already good :). Good suggestion to make architecture proposals, but also let's make sure that the upcoming release contains what we want it to contain. I'm happy to jump on a call on this topic, perhaps with @acud and @janos , to make sure we are putting the effort where it is most effective.
Thanks again.

zbiljic:
@Eknir Related PR from manifest project (noted in the description here) needs to be merged first, after which this to be changed, before this one can be merged.

pkg/manifest/manifest.go
Eknir: Related: https://github.com/ethersphere/bee/issues/463#issuecomment-677590074

acud (approved):

Issue #522, opened 2020-08-05
Print config CLI command


janos:
Based on the feedback https://github.com/ethersphere/bee/issues/508, it would be helpful to have a new cli command that would print a bee configuration in yaml format to show available parameters.

This can be achieved with viper AllSettings() method on command.config and marshalling it with yaml encoder just as it is done in viper internally, pretty straightforward.

Eknir:
@metacertain , if you already have a PR for this, could you please link it?

PR #593, opened 2020-08-20
Config command

metacertain:
#522 

To achieve a greater degree of configuration overviewability, we are introducing the bee config command.

Behavior of the command is to show all options and default values of configuration used without a provided configuration file, 
Or - in presence of a configuration file provided to bee - to show all of the options and the actual values used with that configuration file loaded.

The aim is to display the output formatted as valid yaml

### DIFF
--- a/cmd/bee/cmd/cmd.go
+++ b/cmd/bee/cmd/cmd.go
@@ -61,7 +61,9 @@ func newCommand(opts ...option) (c *command, err error) {
 
 	c.initVersionCmd()
 
-	c.initConfigurateOptionsCmd()
+	if err := c.initConfigurateOptionsCmd(); err != nil {
+		return nil, err
+	}
 
 	return c, nil
 }
--- a/cmd/bee/cmd/cmd.go
+++ b/cmd/bee/cmd/cmd.go
@@ -60,6 +60,9 @@ func newCommand(opts ...option) (c *command, err error) {
 	}
 
 	c.initVersionCmd()
+
+	c.initConfigurateOptionsCmd()
+
 	return c, nil
 }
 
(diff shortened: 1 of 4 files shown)


janos (approved):
LGTM. Thanks @metacertain!

Eknir:
@ldeffenb , this one based on your feedback. Would be amazing if you could test and report your findings

Eknir:
Trying out this feature as a user.

1. I can find directly what I need, by just typing `bee`, showing me that `bee config` is a command.
2. When I do `bee config -h`, I get a suggestion to do `bee config [flags], showing all available flags of the `bee start` command. This is not indended.
3. Ideally, I would like to have the option `bee config --print [path]`, to easily print the default config file (and modify myself afterwards).
4. As a user of Swarm, I would like to print a config file of my current configuration, such that I can easily re-use this and don't have to type the flags all the time
5. Edit on 2 (and follow up on 4), I see now that if I do `bee config [flag with custom value]`, the printed-out config shows that value. Good! It solves 4 and clears up 2. This is just not clear from the `bee config -h`. I suggest to make 
`Print configuration options` somewhat more verbose, explaining that this command can be used as `bee start` (with all ways to pass config values), but that it prints out the yaml config file, which afterwards can be used to easily start up your Swarm node. Something like that. The idea is that a user just starts op Bee, and can work with it, without consulting our user documentation.
6. --config (flag) and config (command) look really alike, but are completely different. I suggest to change the command to some other name
7. Alternatively to 6, you could also just make a flag (--printConfig [path]), which you would pass with `bee start`, to print the current configuration to your specified file.

I don't approve this one yet, as this feature is really user-focused, but the UX is not great yet. If you can think about my experience (as a user) and think about my suggestions (or come up with other suggestions), apply those and give this feature a bit more love, then I will approve!

Great work nevertheless Abel!

PR #594, opened 2020-08-20
Add basic metrics counters for total sent/recieved pseudosettle amounts

metacertain:
The aim of this PR is to introduce some monitoring opportunity to sent / recieved pseudosettlements through the /metrics endpoint of the debug API

### DIFF
--- a/pkg/settlement/pseudosettle/metrics.go
+++ b/pkg/settlement/pseudosettle/metrics.go
@@ -25,13 +25,13 @@ func newMetrics() metrics {
 			Namespace: m.Namespace,
 			Subsystem: subsystem,
 			Name:      "total_received_pseudosettlements",
-			Help:      "Amount of BZZ received from peers (actual income of the node)",
+			Help:      "Amount of pseudotokens received from peers (income of the node)",
 		}),
 		TotalSentPseudoSettlements: prometheus.NewCounter(prometheus.CounterOpts{
 			Namespace: m.Namespace,
 			Subsystem: subsystem,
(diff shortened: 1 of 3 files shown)


janos (approved):

pkg/settlement/pseudosettle/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+
+func newMetrics() metrics {
+	subsystem := "pseudosettle"
+
+	return metrics{
+		TotalReceivedPseudoSettlements: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "total_received_pseudosettlements",
+			Help:      "Amount of BZZ received from peers (actual income of the node)",
```
ralph-pichler, line 28: Should the help maybe mention that there are no real `BZZ` being sent here? or is that already obvious from the `pseudosettle` namespace?

pkg/settlement/pseudosettle/pseudosettle.go
```diff
@@ -90,6 +94,9 @@ func (s *Service) Pay(ctx context.Context, peer swarm.Address, amount uint64) er
 	}()
 
 	s.logger.Tracef("sending payment message to peer %v of %d", peer, amount)
+
+	s.metrics.TotalSentPseudoSettlements.Add(float64(amount))
```
ralph-pichler, line 98: this should be after the `WriteMsgWithContext`, so it does not increase if sending fails

ralph-pichler (approved):

PR #602, opened 2020-08-21
global integration test related fixes

jmozah:
- Removed chunk type test in `IsPotential` as this is unnecessary
- Removed chunk type from swarm.Chunk as it is not needed now 
- Added local existence for chunk when a request for repair is received, otherwise there will be unnecessary network Get (which will fail since someone triggered repair because the chunk is not there in network)

### DIFF
--- a/pkg/recovery/repair.go
+++ b/pkg/recovery/repair.go
@@ -48,6 +48,19 @@ func NewRepairHandler(s storage.Storer, logger logging.Logger, pushSyncer pushsy
 	return func(ctx context.Context, m *trojan.Message) error {
 		chAddr := m.Payload
 
+		// check if the chunk exists in the local store and proceed.
+		// otherwise the Get will trigger a unnecessary network retrieve
+		exists, err := s.Has(ctx, swarm.NewAddress(chAddr))
+		if err != nil {
+			logger.Info("chunk repair: error while getting chunk for repairing: ", err)
+			return err
+		}
+		if !exists {
+			logger.Info("chunk repair: chunk not present in local store for repairing")
+			return err
+		}
+
+		// retrieve the chunk from the local store
 		ch, err := s.Get(ctx, storage.ModeGetRequest, swarm.NewAddress(chAddr))
 		if err != nil {
 			logger.Tracef("chunk repair: error while getting chunk for repairing: %v", err)
--- a/pkg/recovery/repair.go
+++ b/pkg/recovery/repair.go
@@ -6,7 +6,7 @@ package recovery
 
 import (
 	"context"
-	"fmt"
+	"errors"
 
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/pss"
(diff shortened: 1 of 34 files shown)


pkg/recovery/repair.go
```diff
@@ -48,6 +48,19 @@ func NewRepairHandler(s storage.Storer, logger logging.Logger, pushSyncer pushsy
 	return func(ctx context.Context, m *trojan.Message) error {
 		chAddr := m.Payload
 
+		// check if the chunk exists in the local store and proceed.
+		// otherwise the Get will trigger a unnecessary network retrieve
+		exists, err := s.Has(ctx, swarm.NewAddress(chAddr))
+		if err != nil {
+			logger.Info("chunk repair: error while getting chunk for repairing: ", err)
+			return err
```
janos, line 56: Why logging an error and returning it?

pkg/recovery/repair.go
```diff
@@ -48,6 +48,19 @@ func NewRepairHandler(s storage.Storer, logger logging.Logger, pushSyncer pushsy
...
+		}
+		if !exists {
+			logger.Info("chunk repair: chunk not present in local store for repairing")
```
janos, line 59: Why logging an error and returning it?

pkg/recovery/repair.go
jmozah: Creaped in from debugging session.. removing it

pkg/recovery/repair.go
jmozah: Creaped in from debugging session.. removing it

pkg/recovery/repair_test.go
```diff
@@ -201,7 +201,7 @@ func TestNewRepairHandler(t *testing.T) {
 
 		// invoke the chunk repair handler
 		err = repairHandler(context.Background(), &msg)
-		if err != nil && err.Error() != "storage: not found" {
+		if err != nil && err.Error() != "chunk repair: chunk not present in local store for repairing" {
```
acud, line 204: you can use `errors.Is()` if the error is extracted and exposed through `export_test`

pkg/recovery/repair.go
```diff
@@ -48,6 +49,17 @@ func NewRepairHandler(s storage.Storer, logger logging.Logger, pushSyncer pushsy
...
 		chAddr := m.Payload
 
+		// check if the chunk exists in the local store and proceed.
+		// otherwise the Get will trigger a unnecessary network retrieve
+		exists, err := s.Has(ctx, swarm.NewAddress(chAddr))
+		if err != nil {
+			return err
+		}
+		if !exists {
+			return fmt.Errorf("chunk repair: chunk not present in local store for repairing")
```
acud, line 59: maybe would be nice to extract the error to be a `const` or `var` at the beginning of the file?

acud (approved):
minor comments. otherwise LGTM

pkg/recovery/repair_test.go
janos: Yes, correct, lets have a proper error checking, not string validation.

jmozah: oh yeah.. sure

janos (approved):
Nice, thanks for addressing my comments!

PR #603, opened 2020-08-23
pipeline splitter

acud:
This PR changes the splitter to be a pipeline oriented one.
The `Pipeline` interface exposes an `io.Writer` and `Sum` method, for components to use as a black box. Within a pipeline, writers are chainable. It is up for the implementer to decide whether a writer calls the next writer. Implementers should always implement the `Sum` method and call the next writer's `Sum` method (in case there is one), returning its result to the calling context.
The package has a breakdown of the pipeline into several components:
- `chunkFeederWriter` - exposed as `Interface` - the head of the pipeline into which data is written verbatim. The writes are buffered up to one chunk, and partial writes are supported. `Sum()` flushes existing data in the buffer and calls the next writer's `Sum`.
- `bmtWriter` - a writer that hashes provided data using the `bmt` hashing algorithm. Partial writes are not supported
- `storeWriter` - writes the provided data into a `storage.Putter` of some sort
- `hashTrieWriter` - expects full references with their respective spans. Partial writes are not supported. `Sum` returns the swarm merkle root hash of the represented trie. Has a dependency on a constructor function (`type pipelineFunc`) to create a shorter, hashing pipeline for internal purposes when wrapping levels and creating the final hash.
- `resultWriter` - a connector that allows a given caller to receive the result of a shorter pipeline. Needed for using the short pipeline in the `hashTrieWriter`.

NOTE: all of the aforementioned components except for `chunkFeederWriter` assume that the span of the data is prepended to the data's byte slice. `chunkFeederWriter` prepends the span to the data.

Right two pipelines exist through the following constructors:
- `NewPipeline` - a normal pipeline; does not encrypt, chunk branching factor (!=BMT BRANCHING FACTOR!) 128, stores into localstore and provides a merkle root hash of the trie when calling `Sum`
- `NewShortPipeline` - dependency for the `hashTrieWriter` used by the normal pipeline. Also does not encrypt, only hashes data through BMT and stores it in the localstore. Gives back the result of the write chain through the `resultWriter`

The ticket reuses the test vectors created for the original splitter which is in the `file` package.

Still TODO:
- Context propagation on writes. is this necessary with these interfaces?
- Change enclosing `Interface` to be of `hash.Hash`
- Possible minor cleanup on the `pipeWriteArgs` struct. However, for building the merkle trie it would be nicer to have `span` and `ref` separate
- `swarm.Chunk` spans are `uint64`, but in some places I have a feeling we're still doing `int64` due to `io.Writer`
- unit tests for individual writers
- Benchmarks


pkg/file/pipeline/writer.go
```diff
@@ -0,0 +1,27 @@
...
+
+import "io"
+
+// ChainWriter is a writer in a pipeline.
+// It is up to the implementer to decide whether a writer
+// calls the next writer or not. Implementers should
+// call the Sum method of the subsequent writer in case there
+// exists one.
+type ChainWriter interface {
+	ChainWrite(*pipeWriteArgs) (int, error)
```
janos, line 15: This method makes ChainWriter implementable only in this package as pipeWriteArgs is unexported. As I see, the complete implementation is in one package, so this interface does not need to be exported.

pkg/file/pipeline/pipeline.go
```diff
@@ -0,0 +1,50 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package pipeline
+
+import (
+	"github.com/ethersphere/bee/pkg/storage"
+)
+
+type pipeWriteArgs struct {
```
janos, line 11: This structure looks very similar to what the Chunk is. Could it be removed?

pkg/file/pipeline/pipeline.go
```diff
@@ -0,0 +1,50 @@
...
+	head Interface
+}
+
+func (p *Pipeline) Write(b []byte) (int, error) {
+	return p.head.Write(b)
+}
+
+func (p *Pipeline) Sum() ([]byte, error) {
+	return p.head.Sum()
+}
```
janos, line 27: Pipeline type just wraps two methods of the Interface in a concrete type. I am not sure why it is needed, and it can be completely removed.

pkg/file/pipeline/pipeline.go
```diff
@@ -0,0 +1,50 @@
...
+	b := NewBmtWriter(128, lsw)
+	feeder := NewChunkFeederWriter(4096, b)
+
+	return &Pipeline{head: feeder}
+}
+
+type pipelineFunc func(p *pipeWriteArgs) ChainWriter
+
+// this is just a hashing pipeline. needed for level wrapping inside the hash trie writer
+func NewShortPipeline(s storage.Storer) func(*pipeWriteArgs) ChainWriter {
```
janos, line 42: This should be called NewShortChainWriterFunc or something like that, as it does not return a pipeline.

pkg/file/pipeline/writer.go
janos, line 15: All ChainWriter implementations return 0 int and ChainWrite calls do not check for this return value. Is it really needed?

pkg/file/pipeline/pipeline.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func NewPipeline(s storage.Storer) Interface {
+	// DATA -> FEEDER -> BMT -> STORAGE -> TRIE
+	tw := NewHashTrieWriter(4096, 128, 32, NewShortPipeline(s))
+	lsw := NewStoreWriter(s, tw)
+	b := NewBmtWriter(128, lsw)
+	feeder := NewChunkFeederWriter(4096, b)
+
+	return &Pipeline{head: feeder}
```
janos, line 36: `feeder` can be returned without wrapping it with the Pipeline type.

pkg/file/pipeline/chunkfeeder_writer.go
```diff
@@ -0,0 +1,94 @@
...
+const span = 8
+
+type chunkFeeder struct {
+	size      int
+	next      ChainWriter
+	buffer    []byte
+	bufferIdx int
+}
+
+func NewChunkFeederWriter(size int, next ChainWriter) Interface {
```
janos, line 20: It would be very good to comment what is feeder writer and why it exists, as it is a very important component.

pkg/file/pipeline/result_writer.go
```diff
@@ -0,0 +1,24 @@
...
+type resultWriter struct {
+	target *pipeWriteArgs
+}
+
+func NewResultWriter(b *pipeWriteArgs) ChainWriter {
+	return &resultWriter{target: b}
+}
+
+func (w *resultWriter) ChainWrite(p *pipeWriteArgs) (int, error) {
+	*w.target = *p
```
janos, line 18: Is dereference needed?

janos (commented):
I would add, that having all writers, their integrations among each other and their abstractions is too much responsibility for a single package. It would be better that every writer is in its own package with dedicated unit tests. It is currently not possible to implement a ChainWriter outside of this package, so many of exported identifiers are not needed to be exported.

pkg/file/pipeline/chunkfeeder_writer.go
```diff
@@ -0,0 +1,94 @@
...
+			return w + n, nil
+		}
+
+		// fill stuff up from the incoming write
+		n = copy(d[span+f.bufferIdx:], b[i:])
+		i += n
+		sp += n
+
+		binary.LittleEndian.PutUint64(d[:span], uint64(sp))
+		args := &pipeWriteArgs{data: d[:span+sp]}
```
zbiljic, line 68: Why not fill `pipeWriteArgs.span` field here also ?

pkg/file/pipeline/chunkfeeder_writer.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if l+f.bufferIdx < f.size {
+		// write the data into the buffer and return
+		n := copy(f.buffer[f.bufferIdx:], b)
+		f.bufferIdx += n
+		return n, nil
+	}
+
+	// if we are here it means we have to do at least one write
+	d := make([]byte, f.size+span)
```
zbiljic, line 41: Could it be possible to only use `f.buffer`, instead of creating new array each time ?

pkg/file/pipeline/pipeline.go
acud: well, no. since i need a way to pass references and unencrypted spans to the trie writer when doing encrypted content. https://github.com/ethersphere/bee/pull/613/files#diff-2fa7bcdc8b2ddbe0c7c7d1b4fb852b23R18

janos: That is not part of this PR, not sure what problems would it make with encryption, but that this type defined in abstraction does create a coupling with a specific (encrypt) writer in that sense.

pkg/file/pipeline/result_writer.go
acud: i believe so. otherwise you just set the address of the local pointer to be this of `p`, but i need the content to be visible in the calling context in hashTrieWriter

pkg/file/pipeline/pipeline.go
acud: we will have more complexities and dependencies created between different writers in the pipeline when erasure coding and entanglement codes are introduced. having this struct makes _some_ things easier to express declaratively in the code. If we remove this type then I'd just have to offload all of these members into the writer method signature. There will be coupling here either way

pkg/file/pipeline/result_writer.go
janos: I am not sure what the result writer purpose is. It just sets the target and I do not see that it it used anywhere. I am not sure how hashTrieWriter uses this information set here.

pkg/file/pipeline/pipeline.go
janos: Yes, I see, the pipeline needs to be coupled because of complexities of different writers. This is what makes my most significant concern, which is having everything in one package, making abstraction leaks very easy.

acud: Fair enough. I see and respect your concerns. It is only a question to which degree we need abstraction within this component of pipeline. Food for thought in the meanwhile.

pkg/file/pipeline/hashtrie_writer.go
```diff
@@ -0,0 +1,172 @@
...
+	refSize    int
+	fullChunk  int    // full chunk size in terms of the data represented in the buffer (span+refsize)
+	cursors    []int  // level cursors, key is level. level 0 is data level
+	buffer     []byte // keeps all level data
+	pipelineFn pipelineFunc
+}
+
+func NewHashTrieWriter(chunkSize, branching, refLen int, pipelineFn pipelineFunc) ChainWriter {
+	return &hashTrieWriter{
+		cursors:    make([]int, 9),
```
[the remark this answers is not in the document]
acud: 9 is the amount of levels we'd like to support in the hash trie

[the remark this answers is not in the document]
zbiljic: Got it.

pkg/file/pipeline/result_writer.go
acud: hmm... it is true, i can already just mutate the fields in `pipeWriteArgs`. in hashTrieWriter it is used since the target pointer here is set when `pipelineFn` is called, then once we reach this line it dereferences the content onto the pointer provided in the previous call to create the short pipeline. is it ok that i do this change as part of the integration PR?

zbiljic (approved):

janos (approved):
In order to have the encryption for the next release, this PR needs to be merged, but I have some concerns that would be good to be addressed later:

- splitting every writer in separate packages
- propagate context where needed, not to create now ones
- I do not understand the purpose of resultWriter.target, but I may be missing something

pkg/file/pipeline/result_writer.go
janos: Yes, it is, thanks.

pkg/file/pipeline/pipeline.go
```diff
@@ -0,0 +1,55 @@
+package pipeline
+
+import (
+	"github.com/ethersphere/bee/pkg/storage"
+)
+
+type pipeWriteArgs struct {
```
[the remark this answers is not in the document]
acud: its not a chunk. please have a look ad the discussion above: https://github.com/ethersphere/bee/pull/603#discussion_r476433268

acud:
will address more of the concerns in the integration PR

pkg/file/pipeline/chunkfeeder_writer.go
acud: `f.buffer` stores the actual data. we could theoretically do that, but then it means that `f.buffer` must be of size `f.size+span`, and contains the span at the beginning of the data, which means we should also probably constantly marshal the span into the first eight bytes constantly with the data, even on partial writes. this can save this one allocation and probably can also get rid of `bufferIdx` but I feel it is a bit less clear to the reader. i can try to do this when i finish working on encryption and see how it looks

PR #605, opened 2020-08-24
Add multiresolver

paxthemax:
# Mutliresolver and resolver chains

Multiresolver allows a registered resolver chain to be selected based on a TLD in a name (eg. "name.eth" -> ENS resolver chain).
A resolver chain attempts to resolve a name using all resolvers in a chain, returning an error if all resolvers fail.

### DIFF
--- a/pkg/resolver/mock/resolver.go
+++ b/pkg/resolver/mock/resolver.go
@@ -8,18 +8,14 @@ import (
 	"fmt"
 
 	"github.com/ethersphere/bee/pkg/resolver"
-	"github.com/ethersphere/bee/pkg/swarm"
 )
-
-// Address is the swarm bzz address.
-type Address = swarm.Address
 
 // Assure mock Resolver implements the Resolver interface.
 var _ resolver.Interface = (*Resolver)(nil)
 
 // Resolver is the mock Resolver implementation.
 type Resolver struct {
-	resolveFunc func(string) (Address, error)
+	resolveFunc func(string) (resolver.Address, error)
 }
 
 // Option function sets the option on the mock Resolver.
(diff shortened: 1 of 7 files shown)


pkg/resolver/mock/resolver.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+
+	"github.com/ethersphere/bee/pkg/resolver"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// Address is the swarm bzz address.
+type Address = swarm.Address
```
janos, line 15: No need for this type alias.

pkg/resolver/multi.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	chain := mr.resolvers[tld]
+
+	if len(chain) == 0 {
+		return Address{}, ErrResolverChainEmpty
+	}
+
+	for _, res := range chain {
+		adr, err := res.Resolve(name)
+		if err == nil {
```
janos, line 119: This error will not be propagated and could be useful after all resolvers fail.

pkg/resolver/multi.go
```diff
@@ -0,0 +1,133 @@
...
+		mr.ForceDefault = true
+	}
+}
+
+// PushResolver will push a new Resolver to the name resolution chain for the
+// given TLD.
+// TLD names should be prepended with a dot (eg ".tld"). An empty TLD will push
+// to the default resolver chain.
+func (mr *MultiResolver) PushResolver(tld string, r Interface) error {
+	if tld != "" && !isTLD(tld) {
```
janos, line 60: I would allow empty tls value as the value for default resolvers. If there is no tld specified, to try to resolve with them.

pkg/resolver/multi.go
paxthemax: Good point. Which error should be propagated (as there is a whole chain of errors) - the last one?

pkg/resolver/mock/resolver.go
paxthemax: Done

pkg/resolver/multi.go
janos: The last one would be good enough.

zelig (approved):
neat

pkg/resolver/multi.go
paxthemax: Empty value is specifically allowed. I've added a test with that specific case to confirm.

janos (approved):
Cool, thanks @paxthemax.

PR #606, opened 2020-08-24
push already built and tested docker image to docker hub

vandot:
We need to move latest docker image build to github actions for easier CI/CD integration, tag/release will still be built on dockerhub


svetomir (approved):

.github/workflows/beekeeper.yaml
```diff
@@ -56,3 +56,9 @@ jobs:
         run: ./beekeeper check retrieval --api-scheme http --debug-api-scheme http --disable-namespace --debug-api-domain localhost --api-domain localhost --node-count "${REPLICA}" --upload-node-count "${REPLICA}" --chunks-per-node 3
       - name: Test pullsync
         run: ./beekeeper check pullsync --api-scheme http --debug-api-scheme http --disable-namespace --debug-api-domain localhost --api-domain localhost --node-count "${REPLICA}" --upload-node-count "${REPLICA}" --chunks-per-node 3
+      - name: Retag docker image and push
+        run: |
+          docker tag registry.localhost:5000/ethersphere/bee:latest ethersphere/bee:latest
+          printf ${{ secrets.DOCKERHUB_PASSWORD }} > password
+          cat password | docker login --username ${{ secrets.DOCKERHUB_USERNAME }} --password-stdin
```
janos, line 63: Not sure how relevant it is in ci, but could write to a file be avoided with echo to docker login pipe?

janos (approved):

PR #610, opened 2020-08-25
pipeline splitter integration

acud:
This PR integrates the pipeline splitter into `bee`.


pradovic:
Some tests fail.

acud:
yeah i'm on it @pradovic

acud:
should be good to go now :+1:

pkg/api/dirs.go
```diff
@@ -169,8 +168,8 @@ func storeDir(ctx context.Context, reader io.ReadCloser, s storage.Storer, mode
 		return swarm.ZeroAddress, fmt.Errorf("metadata marshal: %w", err)
 	}
 
-	sp := splitter.NewSimpleSplitter(s, mode)
-	mr, err := file.SplitWriteAll(ctx, sp, bytes.NewReader(metadataBytes), int64(len(metadataBytes)), toEncrypt)
+	pipe := pipeline.NewPipeline(ctx, s, mode)
+	mr, err := pipeline.FeedPipeline(ctx, pipe, bytes.NewReader(metadataBytes), int64(len(metadataBytes)))
```
pradovic, line 172: Maybe you can inline pipe here?

pkg/api/dirs.go
```diff
@@ -182,8 +181,8 @@ func storeDir(ctx context.Context, reader io.ReadCloser, s storage.Storer, mode
 		return swarm.ZeroAddress, fmt.Errorf("entry marshal: %w", err)
 	}
 
-	sp = splitter.NewSimpleSplitter(s, mode)
-	manifestFileReference, err := file.SplitWriteAll(ctx, sp, bytes.NewReader(fileEntryBytes), int64(len(fileEntryBytes)), toEncrypt)
+	pipe = pipeline.NewPipeline(ctx, s, mode)
+	manifestFileReference, err := pipeline.FeedPipeline(ctx, pipe, bytes.NewReader(fileEntryBytes), int64(len(fileEntryBytes)))
```
pradovic, line 185: Same as the above as it appears it is not used anywhwre else.

pkg/api/file.go
```diff
@@ -179,8 +177,8 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
 		jsonhttp.InternalServerError(w, "metadata marshal error")
 		return
 	}
-	sp = splitter.NewSimpleSplitter(s.Storer, mode)
-	mr, err := file.SplitWriteAll(ctx, sp, bytes.NewReader(metadataBytes), int64(len(metadataBytes)), toEncrypt)
+	pipe = pipeline.NewPipeline(ctx, s.Storer, mode) // TODO just implement `Reset` on the pipeline
```
pradovic, line 180: Did we agree to not leave TODO comments in the codebase before? I remember the comment about keeping personal todos somewhere else.

pkg/api/dirs.go
acud: let's leave as it is as things might get more complicated when encryption is introduced

pkg/api/dirs.go
acud: but it reuses the same variable so hopefully less allocations

pradovic: Probably the same, as it will do the copy in the constructor which will happen in both cases. Not sure, though :)

pkg/manifest/simple.go
```diff
@@ -106,9 +106,9 @@ func (m *simpleManifest) Store(ctx context.Context, mode storage.ModePut) (swarm
 		return swarm.ZeroAddress, fmt.Errorf("manifest marshal error: %w", err)
 	}
 
-	sp := splitter.NewSimpleSplitter(m.storer, mode)
-
-	address, err := file.SplitWriteAll(ctx, sp, bytes.NewReader(data), int64(len(data)), m.encrypted)
+	pipe := pipeline.NewPipeline(ctx, m.storer, mode)
```
acud, line 109: on a sidenote this is not so great that we have this dependency buried here. the ability to marshal a manifest should be possible from the calling context

janos (approved):

pradovic (approved):

pkg/manifest/simple.go
zbiljic: That will be problem for new 'mantaray' implemetation, which uses some internal `LoadSaver` interface, which lazily saves and loads required Trie nodes, to optimize space/time.

PR #615, opened 2020-08-25
ENS Resolver client

paxthemax:
# Add ENS resolver client

The client should use an Ethereum node at a given endpoint (with ENS deployed) for the name resolution service.

ENS resolver chain can be set up via the `--ens-endpoints` flag.

## Example

```
bee start --ens-endpoints "nope,https://sokol.poa.network,https://cloudflare-eth.com"
```

Should return a "could not dial error" for the first endpoint, "no contract deployed" for the second, and should establish connection with the last one.

Current implementation does not provide a default ENS endpoint.


pkg/node/node.go
```diff
@@ -307,6 +311,21 @@ func NewBee(addr string, logger logging.Logger, o Options) (*Bee, error) {
...
+	// TODO: consider refactoring into resolver.Service that handles connection
+	// logic and logging.
+	multiResolver := resolver.NewMultiResolver()
+	for idx, ep := range o.ENSResolverEndpoints {
+		ensR := ens.NewClient()
+		if err := ensR.Connect(ep); err != nil {
+			logger.Warningf("ens resolver [%d]: failed to connect: %v", idx, err)
+		} else {
+			logger.Infof("ens resolver [%d]: connected to endpoint %q", idx, ep)
+			_ = multiResolver.PushResolver(".eth", ensR)
```
janos, line 324: Check for error.

janos, line 324: This will able to resolve only .eth domains, not .test, for example. Swarm `--ens-api` flag provides possibilities to specify default, or any tld resolvers with their specific contract addresses.

janos (commented):
I think that complete functionality for multiresolver should be utilized with cli flags, providing a way to specify resolvers for any tld and default resolvers, similar to swarm's `--ens-api` flag.

pkg/resolver/client/ens/ens_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+func TestConnect(t *testing.T) {
+	ep := "test"
+
+	t.Run("no dial func error", func(t *testing.T) {
+		c := ens.NewClient(
+			ens.WithDialFunc(nil),
+		)
+		err := c.Connect(ep)
+		if err == nil && err.Error() != "no dial function implementation" {
```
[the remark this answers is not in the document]
zelig: `!=` or `||`?

pkg/node/node.go
```diff
@@ -57,6 +59,7 @@ type Bee struct {
 	p2pCancel        context.CancelFunc
 	apiServer        *http.Server
 	debugAPIServer   *http.Server
+	multiResolver    resolver.Interface
```
janos, line 62: The Bee struct only keeps the multi resolver in order to close it on Shutdown. As with other closers this should be `resolverCloser io.Closer`. Also, it is just a resolver, multi is specific to the implementation, which is here not relevant.

paxthemax:
Pushed updates to the implementation. Now we use a connection string (as in the old swarm):
```
[tld:][address@]name
--------
eth:0xada....adad@cloudflare-eth.com
```

Note that the address string is NOT used as of now, only recorded for compatibility purposes.

janos:
@paxthemax could you resolve merge conflicts?

pkg/resolver/client/ens/ens.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Connect implements the resolver.Client interface.
+func (c *Client) Connect(ep string) error {
+	if c.dialFn == nil {
+		return errors.New("no dial function implementation")
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
janos, line 63: This clock can be moved bellow the dialFn call and its error check. Mutex is only protecting the Client fields, and dialFn can take arbitrary long time and it should not hold a lock.

pkg/resolver/client/ens/ens.go
```diff
@@ -0,0 +1,121 @@
...
+// Function obtains a read lock while interacting with the Ethereum client.
+func (c *Client) Resolve(name string) (Address, error) {
+	if c.resolveFn == nil {
+		return swarm.ZeroAddress, errors.New("no resolve function implementation")
+	}
+
+	// Retrieve the content hash from ENS. Obtain a read lock.
+	c.mu.RLock()
+	hash, err := c.resolveFn(c.ethCl, name)
+	c.mu.RUnlock()
```
janos, line 95: There is no need to protect the resolve function call here.

It could be done something like this, to reduce the locking time:

```go
	c.mu.RLock()
	cl := c.ethCl
	c.mu.RUnlock()
	hash, err := c.resolveFn(cl, name)
```

pkg/resolver/client/ens/ens_test.go
```diff
@@ -0,0 +1,140 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package ens_test
```
janos, line 5: These tests need to close clients that are creating.

pkg/resolver/multi.go
```diff
@@ -5,17 +5,18 @@
...
-	"errors"
 	"path"
 	"strings"
 )
 
 // MultiResolver errors.
 var (
-	ErrInvalidTLD         = errors.New("invalid TLD")
-	ErrResolverChainEmpty = errors.New("resolver chain empty")
+// ErrResolverChainEmpty = errors.New("resolver chain empty")
```
janos, line 14: Remove commented code.

pkg/resolver/resolver.go
```diff
@@ -14,4 +14,5 @@ type Address = swarm.Address
 // Interface can resolve an URL into an associated Ethereum address.
 type Interface interface {
 	Resolve(url string) (Address, error)
+	Close()
```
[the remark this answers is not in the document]
janos: This is still unresolved.

pkg/resolver/service/service.go
```diff
@@ -0,0 +1,145 @@
...
+	"unicode"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/resolver"
+	"github.com/ethersphere/bee/pkg/resolver/client/ens"
+)
+
+// Service is the name resolution service ready for integration with bee.
+type Service struct {
```
janos, line 19: It looks to me like that Service is redundant to Multiresolver. This package provides the service only to have a Connect method doing something. Could the Service type be removed and only have a Connect and Parse* functions in multiresolver as that all is in function for it. After that the whole service package is no longer needed.

pkg/resolver/service/service.go
```diff
@@ -0,0 +1,145 @@
...
+// Close implements the Closer interface.
+func (s *Service) Close() error {
+	s.multi.Close()
+	return nil
+}
+
+// ParseConnectionString will try to parse a connection string used to connect
+// the Resolver to a name resolution service. The resulting config can be
+// used to initialize a resovler Service.
+func ParseConnectionString(cs string) (*ConnectionConfig, error) {
```
janos, line 87: This function is used only internally, can be unexported.

janos (commented):
There are some merge conflicts that need to be resolved.

cmd/bee/cmd/start.go
```diff
@@ -11,18 +11,47 @@ import (
...
 	"os"
 	"os/signal"
+	"path/filepath"
 	"strings"
 	"syscall"
 	"time"
 
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/node"
+	"github.com/ethersphere/bee/pkg/resolver/service"
```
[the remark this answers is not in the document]
paxthemax: Good point, will do.

pkg/node/node.go
```diff
@@ -38,6 +38,7 @@ import (
 	"github.com/ethersphere/bee/pkg/pusher"
 	"github.com/ethersphere/bee/pkg/pushsync"
 	"github.com/ethersphere/bee/pkg/recovery"
+	"github.com/ethersphere/bee/pkg/resolver/service"
```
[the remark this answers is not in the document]
paxthemax: As above, will do.

cmd/bee/cmd/start.go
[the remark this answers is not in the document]
janos: Have look at the comment about the service package, also. I think that it may not be needed at all.

pkg/resolver/resolver.go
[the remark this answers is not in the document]
paxthemax: It should be fixed now, I apologize.

pkg/resolver/service/service.go
paxthemax: I agree, I will remove the `Service` struct. Probably the `MultiResolver` is sufficient on it's own.

janos (commented):
@paxthemax there are changes that revert some of the previous merge commits. Could you rebase this branch with master?

pkg/resolver/service/service.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/resolver"
+	"github.com/ethersphere/bee/pkg/resolver/client/ens"
+)
+
+// InitMultiResolver will create a new MultiResolver, create the appropriate
+// resolvers, push them to the resolver chains and attempt to connect.
+func InitMultiResolver(logger logging.Logger, cfgs []*resolver.ConnectionConfig) *resolver.MultiResolver {
```
janos, line 15: This looks to me like a constructor for MultiResolver. I think that it can be in the resolver package, where MultiResolver and ConnectionConfig are. And that service package is not needed.

pkg/resolver/service/service.go
```diff
@@ -0,0 +1,58 @@
...
+	if len(cfgs) > 0 {
+		logger.Info("no name resolution service provided")
+	}
+
+	logger.Debug("connecting to name resolution services")
```
janos, line 20: This log line does not provide much information. Maybe to add a log line for every connection that is made with some context about that connection (address, etc...).

pkg/resolver/multi_test.go
```diff
@@ -226,4 +227,15 @@ func TestResolve(t *testing.T) {
 			}
 		})
 	}
+
+	t.Run("close all", func(t *testing.T) {
+		mr.Close()
```
janos, line 232: Error is not checked.

pkg/resolver/multi.go
```diff
@@ -57,7 +52,7 @@ func WithForceDefault() Option {
 // to the default resolver chain.
 func (mr *MultiResolver) PushResolver(tld string, r Interface) error {
 	if tld != "" && !isTLD(tld) {
-		return ErrInvalidTLD
+		return ErrInvalidTLD("tld")
```
janos, line 55: Should the argument be the `tld` variable, not `"tld"` string?

pkg/resolver/mock/resolver.go
```diff
@@ -5,16 +5,23 @@
...
 // Assure mock Resolver implements the Resolver interface.
 var _ resolver.Interface = (*Resolver)(nil)
 
+// ErrNotImplemented denotes a function has not been implemented.
+type ErrNotImplemented struct{}
+
+// Error returns the formatted not implemented error.
+func (e ErrNotImplemented) Error() string {
+	return "function not implemented"
+}
```
janos, line 20: This error type is pretty static. Could it be `var ErrNotImplemented = errors.New("function not implemented")`?

pkg/resolver/client/ens/errors.go
```diff
@@ -0,0 +1,24 @@
...
+
+import "fmt"
+
+// ErrNotImplemented denotes that the function has not been implemented.
+type ErrNotImplemented struct{}
+
+// Error returns the fomatted not implemented error message.
+func (e ErrNotImplemented) Error() string {
+	return "function not implemented"
+}
```
janos, line 15: This error type is pretty static. Could it be var ErrNotImplemented = errors.New("function not implemented")?

pkg/resolver/client/ens/ens_test.go
```diff
@@ -0,0 +1,149 @@
...
+		)
+
+		if err := c.Connect("test"); err == nil {
+			t.Fatal("expected error")
+		}
+		c.Close()
+	})
+
+	t.Run("ok", func(t *testing.T) {
+		c := ens.NewClient(
```
janos, line 49: Defer a function that will close the client if it is still connected, as it will consume resources in case of an error in between the explicit Close call.

pkg/resolver/client/ens/ens_integration_test.go
```diff
@@ -0,0 +1,95 @@
...
+			wantAdr: "00cb23598c2e520b6a6aae3ddc94fed4435a2909690bdd709bf9d9e7c2aadfad",
+		},
+	}
+	for _, tC := range testCases {
+		t.Run(tC.desc, func(t *testing.T) {
+			if tC.endpoint == "" {
+				tC.endpoint = defaultEndpoint
+			}
+
+			eC := ens.NewClient()
```
janos, line 65: Defer a function that will close the client if it is still connected, as it will consume resources in case of an error in between the explicit Close call.

janos (approved):

janos (commented):
I have just a few more comments.

pkg/resolver/client/ens/ens_test.go
paxthemax: Specifically, in this test, the Dial function is a noop. Calling Close() will cause a panic, as the ethereumClient underneath has not been properly initialized.

I will refactor the test in one of the future PRs to avoid this behavior.

pkg/resolver/service/service.go
paxthemax: Personally, I like that there is a separate factory-like component as it decouples the MultiResolver logic from eg. mapping of domains to resolver clients, parsing connection strings etc.

I believe it breaches the scope of this PR. I will create a separate refactor issue for the resolver.

pkg/resolver/client/ens/impl.go
```diff
@@ -0,0 +1,50 @@
...
+		return nil, err
+	}
+
+	return cl, nil
+}
+
+func wrapResolve(backend bind.ContractBackend, name string) (string, error) {
+
+	// Connect to the ENS resolver for the provided name.
+	ensR, err := goens.NewResolver(backend, name)
```
zelig, line 33: why create this on  each resolve call? is  it not  better just  to reuse  it?

pkg/resolver/errors.go
```diff
@@ -0,0 +1,70 @@
...
+
+// ErrResolverChainEmpty denotes trying to pop an empty resolver chain.
+type ErrResolverChainEmpty string
+
+// Error returns the formatted resolver chain empty error.
+func (e ErrResolverChainEmpty) Error() string {
+	return fmt.Sprintf("Resolver chain for %q empty", string(e))
+}
+
+// TODO: implement MultiError as separate package.
```
zelig, line 36: i  feel this may be unnecessary to have this as an error type 
just collect the errors in situ in an array
and return error if non-empty

pkg/resolver/multi.go
```diff
@@ -121,10 +117,24 @@ func (mr *MultiResolver) Resolve(name string) (Address, error) {
...
 	return Address{}, err
 }
 
+// Close all will call Close on all resolvers in all resolver chains.
+func (mr *MultiResolver) Close() error {
+	errs := new(CloseError)
+
+	for _, chain := range mr.resolvers {
+		for _, r := range chain {
+			if err := r.Close(); err != nil {
```
zelig, line 129: put the  nil check inside `add`

pkg/resolver/errors.go
```diff
@@ -0,0 +1,70 @@
...
+// had an error when Close was called.
+type CloseError struct {
+	errs []error
+}
+
+func (me CloseError) add(err error) {
+	me.errs = append(me.errs, err)
+}
+
+func (me CloseError) resolve() error {
```
zelig, line 48: it  is unkind to call this `resolve` inside a resolver pkg IMHO  :) maybe `eval` or explicitly `errIfNotEmpty`

pkg/resolver/errors.go
```diff
@@ -0,0 +1,70 @@
...
+
+// TODO: implement MultiError as separate package.
+
+// CloseError denotes that at least one resolver in the MultiResolver has
+// had an error when Close was called.
+type CloseError struct {
+	errs []error
+}
+
+func (me CloseError) add(err error) {
```
zelig, line 44: put the nil check here?

zelig (changes requested):
quite thorough.
complexity could be reduced, agree with @janos comments

pkg/resolver/client/ens/ens_test.go
```diff
@@ -0,0 +1,149 @@
...
+	bzzAddress := swarm.MustParseHexAddress(
+		"6f4eeb99d0a144d78ac33cf97091a59a6291aa78929938defcf967e74326e08b",
+	)
+
+	t.Run("no resolve func error", func(t *testing.T) {
+		c := ens.NewClient(
+			ens.WithResolveFunc(nil),
+		)
+		_, err := c.Resolve("test")
+		if err == nil && errors.Is(err, ens.ErrNotImplemented{}) {
```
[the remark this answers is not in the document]
paxthemax: Thank you, that was a nasty mishap!

pkg/resolver/client/ens/impl.go
paxthemax: 'NewResolver' will return an ENS resolver which is specific for a given domain name. Internally, it perform checks and fetches the contract address so that it can be queried for ENS records.

Caching is another matter that IMHO should be discussed separately.

pkg/resolver/errors.go
paxthemax: I would consider importing https://github.com/hashicorp/go-multierror, as `node.go` also contains multierror code. However, `node.go` is kind of a mess presently, and it is outside of the scope of this PR in any case.

paxthemax:
@zelig, @janos I apologize for the large PR and the somewhat rushed implementation. I you are satisfied, I would delegate further enhancements to new issues and focus attention to integration with the api

pkg/resolver/client/ens/ens_test.go
janos: Wouldn't something like this work?

```go
defer func() {
    if c.IsConnected() {
        if err := c.Close(); err != nil {
            t.Fatal(err)
        }
    }
}()
```

Or maybe to create a resilient Close method, as it should not cause panic.

As in these tests, Close is called after the Connect call, but Close is usually associated with releasing resources acquired with the constructor, in this case NewClient. I would assume to always call Close when the instance is successfully created with the constructor.

pkg/resolver/mock/resolver.go
```diff
@@ -45,5 +50,12 @@ func (r *Resolver) Resolve(name string) (resolver.Address, error) {
 	if r.resolveFunc != nil {
 		return r.resolveFunc(name)
 	}
-	return resolver.Address{}, fmt.Errorf("not implemented")
+	return resolver.Address{}, fmt.Errorf("%w: resolveFunc", ErrNotImplemented)
```
janos, line 53: `%w` should be at the end of the format string.

pkg/resolver/multi.go
```diff
@@ -103,28 +99,41 @@ func (mr *MultiResolver) GetChain(tld string) []Interface {
...
+
 	tld := ""
 	if !mr.ForceDefault {
 		tld = getTLD(name)
 	}
 	chain := mr.resolvers[tld]
 
 	if len(chain) == 0 {
-		return Address{}, ErrResolverChainEmpty
+		return Address{}, fmt.Errorf("%w: %s", ErrResolverChainEmpty, tld)
```
janos, line 110: `%w` should be at the end of the format string.

pkg/resolver/service/service.go
janos: I do not see that what of decoupling is done, as mapping of resolver clients is the responsibility of multiresolver, and parsing connection strings is already in the resolver package ParseConnectionStrings. This function just uses implementations from there. At least do not call it `service` as it does not contain any services, just one constructor function.

pkg/resolver/errors.go
janos: I think that we do not need go-multierror. We can pull put multierror from `node` packge, we just did not need it elsewhere.

Please, write issues for what is required to change in the `node` package, so that we can address them.

zelig (approved):

janos (approved):

pkg/resolver/client/ens/ens.go
```diff
@@ -0,0 +1,132 @@
...
+	ethCl     *ethclient.Client
+	dialFn    dialType
+	resolveFn resolveType
+}
+
+// Option is a function that applies an option to a Client.
+type Option func(*Client)
+
+// NewClient will return a new Client.
+func NewClient(opts ...Option) *Client {
```
acud, line 42: it would be better to return `client.Interface` here, unless there's some specific reason not to

pkg/resolver/client/ens/ens_integration_test.go
```diff
@@ -0,0 +1,96 @@
...
+			defer eC.Close()
```
acud, line 66: this will be called twice (L90)

pkg/resolver/client/ens/ens_test.go
acud: > Specifically, in this test, the Dial function is a noop. Calling Close() will cause a panic, as the ethereumClient underneath has not been properly initialized.

Then the `Close` function should have a nil check to prevent a panic

pkg/resolver/client/ens/ens_test.go
```diff
@@ -0,0 +1,166 @@
...
+func TestConnect(t *testing.T) {
+	ep := "test"
+
+	t.Run("no dial func error", func(t *testing.T) {
+		c := ens.NewClient(
+			ens.WithDialFunc(nil),
+		)
+		err := c.Connect(ep)
+		defer c.Close()
+		if !errors.Is(err, ens.ErrNotImplemented) {
```
acud, line 32: i don't understand the semantics of `not implemented` as an error for no dial function.... this is confusing

pkg/resolver/client/ens/errors.go
```diff
@@ -0,0 +1,22 @@
...
+
+package ens
+
+import (
+	"errors"
+)
+
+var (
+	// ErrNotImplemented denotes that the function has not been implemented.
+	ErrNotImplemented = errors.New("function not implemented")
```
acud, line 13: is this really something we need to export on a package level?

pkg/resolver/config_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+func TestParseConnectionStrings(t *testing.T) {
+	testCases := []struct {
+		desc       string
+		conStrings []string
+		wantCfg    []resolver.ConnectionConfig
+		wantErr    error
+	}{
+		{
+			desc: "tld too long",
```
acud, line 22: according to whom? is this an IETF definition? if so, please quote it. also `ffffff.....` is not a TLD, TLD is `com`, `ffff... ` is a subdomain

pkg/resolver/errors.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+var (
+	// ErrTLDTooLong denotes when a TLD in a name exceeds maximum length.
+	ErrTLDTooLong = fmt.Errorf("TLD exceeds maximum length of %d characters", maxLabelLength)
```
acud, line 15: should `maxLabelLength` mean `maxTldLength`?

Issue #222, opened 2020-05-26
Improve UX around creating an account and giving password


Eknir:
Right now, the first thing which a user sees when booting up Bee is:

`password`

This is not super user-friendly. 

It would be better to give the user slightly more context, telling him that, as it is the first time that he boots up Bee, he should make a new password.

Then, the user should be re-prompted to enter this password again to unlock the Bee.

Eknir:
@janos

Eknir:
@janos , do we want this in for next release? If so: let' move the card in our Kanban board. If not, also fine.

Eknir:
Maybe you could tell @pradovic what your idea was?

pradovic:
@Eknir, @janos Sorry, I was a bit late on this. I have implemented some logic, but now that I take a look at it I don't really like it. We are currently not remembering the password and the user. If we start to do it, then what should we do if the user enters a wrong password? We would need to add the way to change the password etc.. Only thing that we are checking at the moment is that if the keys exists or not, and if not we create the new ones, but password is typed every time (unless provided by config or file). I think this is kind of a good approach. Maybe we should just change the text? Or should we remember the password for the user and refuse the user if new/wrong password is used?

janos:
There is no need to remember (store) the password.

First case: there are no keys and user must provide password to protect newly generated keys. Two password inputs on creation is just a repercussion that the user entered the password correctly. They are just in memory, checked if they are equal before new key is created.

Second case: keys already exist and user needs to provide a password only once over terminal to decrypt it.

These two cases are only when the password is required to be entered over terminal. When the password is provided with cli configuration options (flag, env var or config file), there is no interaction over terminal.

There is no need for a password to be changed, at least not at this time. If that would be needed, a dedicated command or even binary can be provided.

PR #617, opened 2020-08-25
Improve password creation UX

pradovic:
https://github.com/ethersphere/bee/issues/222

This is the text that appears when the node is started for the first time:

```
Bee node is booting up for the first time. Please provide a new password.
Password:
Confirm password:
```

Let me know if you have better idea for the text :) cc @Eknir


Eknir (approved):
Much better like this @pradovic , thanks for doing this. I tested out the feature and can confirm that it works. I did not look at the specific implementation.

janos (approved):
Thanks @pradovic.

PR #621, opened 2020-08-27
Add support for 'mantaray' manifest, and make it default

zbiljic:
This change adds `mantaray` manifest implementation for `bee`.

Externally from `pkg/manifest` package there are no changes, expect to some test, where some edits were required.
Since the `mantaray` implementation contains `obfuscationKey`, which is generated randomly for new manifest, each time collection is uploaded, new hash will be returned by the service.

### DIFF
--- a/pkg/api/dirs_test.go
+++ b/pkg/api/dirs_test.go
@@ -163,8 +163,7 @@ func TestDirs(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			// NOTE: cannot have expected hash for manifest because each time
-			//       it is created it is obfuscated using random key
+			// NOTE: reference will be different each time, due to manifest randomness
 
 			if resp.Reference.String() == "" {
 				t.Fatalf("expected file reference, did not got any")
--- a/pkg/api/dirs_test.go
+++ b/pkg/api/dirs_test.go
@@ -82,13 +82,11 @@ func TestDirs(t *testing.T) {
 
 	// valid tars
 	for _, tc := range []struct {
-		name         string
-		expectedHash string
-		files        []f // files in dir for test case
+		name  string
+		files []f // files in dir for test case
 	}{
 		{
-			name:         "non-nested files without extension",
-			expectedHash: "685f591d0482a57e172aecb7f58babd7eb50fcb8411f875cae5c7b96fa44ff82",
+			name: "non-nested files without extension",
 			files: []f{
 				{
 					data:      []byte("first file data"),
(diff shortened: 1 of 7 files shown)


janos (approved):
👍

pkg/api/dirs_test.go
```diff
@@ -166,8 +163,11 @@ func TestDirs(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			if tc.expectedHash != resp.Reference.String() {
-				t.Fatalf("expected file reference to match %s, got %x", tc.expectedHash, resp.Reference)
+			// NOTE: cannot have expected hash for manifest because each time
+			//       it is created it is obfuscated using random key
```
pradovic, line 167: Rephrase comment a bit maybe?

pkg/api/file_test.go
```diff
@@ -339,20 +340,42 @@ func TestRangeRequests(t *testing.T) {
...
-				jsonhttptest.WithExpectedJSONResponse(api.FileUploadResponse{
-					Reference: swarm.MustParseHexAddress(upload.reference),
-				}),
 				jsonhttptest.WithRequestHeader("Content-Type", upload.contentType),
+				jsonhttptest.WithPutResponseBody(&respBytes),
 			)
 
+			if uploadReference == "" {
+				// NOTE: cannot have expected hash for manifest because each time
+				//       it is created it is obfuscated using random key
```
pradovic, line 355: Same here

pradovic (approved):
LGTM 👍

pkg/manifest/mantaray.go
```diff
@@ -0,0 +1,176 @@
...
+) *mantarayLoadSaver {
+	return &mantarayLoadSaver{
+		ctx:       ctx,
+		encrypted: encrypted,
+		storer:    storer,
+		modePut:   modePut,
+	}
+}
+
+func (ls *mantarayLoadSaver) Load(ref []byte) ([]byte, error) {
```
acud, line 151: i don't really understand why this stuff is necessary. while refactoring the splitter i noticed that the dependencies to the joiner and splitter have crept into this package and i don't really understand why. this seems like an unnecessary abstraction over simple generic interfaces like `BinaryMarshaler`/`Unmarshaler`.  also i don't really understand the semantics here, from `Load` of a `mantarayLoader` I would expect to return a `mantaray` type, not a byte slice. this is just an abstraction over splitter and joiner. why?

PR #622, opened 2020-08-27
tags: remove anonymous field

acud:
removes the unused anonymous field from tag struct


Eknir:
From what I got in #core, the tag name is also not used. Is there a possibility to remove this one as well?

janos (approved):

zbiljic (approved):

PR #634, opened 2020-08-30
Resolver integration

paxthemax:
# Integrate resolver with the API service

This PR integrates the name resolution to the API service. All endpoints that receive an `{addr}` parameter will expect either a bzz address or a name. During the execution of the request, if a name has been passed, the api service will try to resolve it into a hash. If the resolution is successful, execution proceeds as normal.

## Acceptance criteria

Start the bee node with the following resolver:
```
bee start -resolver-options "eth:https://cloudflare-eth.com" --verbosity debug
``` 

Log should output:
```
...
DEBU[...] name resolver: resolver for ".eth": connecting to endpoint https://cloudflare-eth.com
INFO[...] name resolver: resolver for ".eth" domain: connected to https://cloudflare-eth.com
...
```

Resolving 
```
localhost:8080\files\latest.bookofswarm.eth
```

Should yield the following messages in the log:
```
INFO[...] name resolve: resolved name latest.bookofswarm.eth to a983eae4225c19068f6b795ffaa3fe7b23170842027d908a7b11c6fc36247f01
```

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -5,6 +5,7 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"strconv"
@@ -15,16 +16,19 @@ import (
 	m "github.com/ethersphere/bee/pkg/metrics"
 	"github.com/ethersphere/bee/pkg/resolver"
 	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/ethersphere/bee/pkg/tags"
 	"github.com/ethersphere/bee/pkg/tracing"
 )
 
+// Swarm headers.
 const (
 	SwarmPinHeader     = "Swarm-Pin"
-	SwarmTagUidHeader  = "Swarm-Tag-Uid"
+	SwarmTagUIDHeader  = "Swarm-Tag-Uid"
 	SwarmEncryptHeader = "Swarm-Encrypt"
 )
 
+// Service is the API service interface.
 type Service interface {
 	http.Handler
 	m.Collector
@@ -46,6 +50,7 @@ const (
 	TargetsRecoveryHeader = "swarm-recovery-targets"
 )
 
+// New will create a and initialize a new API service.
 func New(tags *tags.Tags, storer storage.Storer, resolver resolver.Interface, corsAllowedOrigins []string, logger logging.Logger, tracer *tracing.Tracer) Service {
 	s := &server{
 		Tags:               tags,
@@ -64,9 +69,9 @@ func New(tags *tags.Tags, storer storage.Storer, resolver resolver.Interface, co
 
 // getOrCreateTag attempts to get the tag if an id is supplied, and returns an error if it does not exist.
 // If no id is supplied, it will attempt to create a new tag with a generated name and return it.
-func (s *server) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
+func (s *server) getOrCreateTag(tagUID string) (*tags.Tag, bool, error) {
 	// if tag ID is not supplied, create a new tag
-	if tagUid == "" {
+	if tagUID == "" {
 		tagName := fmt.Sprintf("unnamed_tag_%d", time.Now().Unix())
 		var err error
 		tag, err := s.Tags.Create(tagName, 0)
@@ -75,12 +80,35 @@ func (s *server) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
 		}
 		return tag, true, nil
 	}
-	uid, err := strconv.Atoi(tagUid)
+	uid, err := strconv.Atoi(tagUID)
 	if err != nil {
 		return nil, false, fmt.Errorf("cannot parse taguid: %w", err)
 	}
 	t, err := s.Tags.Get(uint32(uid))
 	return t, false, err
+}
+
+var errInvalidChunkAddress = errors.New("invalid chunk address")
+
+func (s *server) resolveNameOrAddress(str string) (swarm.Address, error) {
+	log := s.Logger
+
+	// Try and parse the name as a bzz address.
+	adr, err := swarm.ParseHexAddress(str)
+	if err == nil {
+		log.Debugf("name parse: valid bzz address %q", str)
+		return adr, nil
+	}
+	// Try and resolve the name using the provided resolver.
+	log.Debugf("name parse: attempting to resolve %s to bzz address", str)
+	adr, err = s.Resolver.Resolve(str)
+	if err == nil {
+		log.Debugf("name parse: resolved name %s to %s", str, adr)
+		return adr, nil
+	}
+	log.Errorf("name parse: failed to resolve name %s", str)
+
+	return swarm.ZeroAddress, fmt.Errorf("%w: %v", errInvalidChunkAddress, err)
 }
 
 // requestModePut returns the desired storage.ModePut for this request based on the request headers.
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -111,7 +111,7 @@ func (s *server) resolveNameOrAddress(str string) (swarm.Address, error) {
 	log.Debugf("name resolve: attempting to resolve %s to bzz address", str)
 	adr, err = s.Resolver.Resolve(str)
 	if err == nil && !adr.IsZero() {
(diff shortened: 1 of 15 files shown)


pkg/api/api.go
```diff
@@ -72,14 +80,45 @@ func (s *server) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
...
+
+	// If no resolver is not available, return an error.
+	if s.Resolver == nil {
+		log.Errorf("name resolve: no name resolver available", str)
+		return swarm.ZeroAddress, errNoResolver
+	}
+
+	// Try and resolve the name using the provided resolver.
+	log.Debugf("name resolve: attempting to resolve %s to bzz address", str)
+	adr, err = s.Resolver.Resolve(str)
```
zelig, line 112: addr please

zelig (approved):
lgtm

pkg/api/api.go
paxthemax: Whoops...

pkg/api/api.go
```diff
@@ -80,6 +88,37 @@ func (s *server) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
...
+var errInvalidChunkAddress = errors.New("invalid chunk address")
+var errNoResolver = errors.New("no resolver connected")
+
+func (s *server) resolveNameOrAddress(str string) (swarm.Address, error) {
+	log := s.Logger
+
+	// Try and parse the name as a bzz address.
+	addr, err := swarm.ParseHexAddress(str)
+	if err == nil {
+		log.Debugf("name resolve: valid bzz address %q", str)
```
janos, line 100: This should be Trace.

pkg/api/api.go
```diff
@@ -80,6 +88,37 @@ func (s *server) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
...
+		return addr, nil
+	}
+
+	// If no resolver is not available, return an error.
+	if s.Resolver == nil {
+		log.Errorf("name resolve: no name resolver available", str)
```
janos, line 106: The error is returned, no need to have this log line.

pkg/api/api.go
```diff
@@ -80,6 +88,37 @@ func (s *server) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
...
+		return swarm.ZeroAddress, errNoResolver
+	}
+
+	// Try and resolve the name using the provided resolver.
+	log.Debugf("name resolve: attempting to resolve %s to bzz address", str)
+	addr, err = s.Resolver.Resolve(str)
+	if err == nil && !addr.IsZero() {
+		log.Infof("name resolve: resolved name %s to %s", str, addr)
```
janos, line 114: I think that this also can be Trace level.

pkg/api/api.go
```diff
@@ -80,6 +88,37 @@ func (s *server) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
...
+		return addr, nil
+	}
+	log.Errorf("name resolve: failed to resolve name %s", str)
```
janos, line 117: Error is returned bellow, no need for this log line.

pkg/api/api.go
```diff
@@ -13,17 +14,21 @@ import (
...
 	"github.com/ethersphere/bee/pkg/logging"
 	m "github.com/ethersphere/bee/pkg/metrics"
+	"github.com/ethersphere/bee/pkg/resolver"
 	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/ethersphere/bee/pkg/tags"
 	"github.com/ethersphere/bee/pkg/tracing"
 )
 
+// Swarm headers.
```
acud, line 24: this is not necessary. if you want this to be useful please document each header separately in godoc format

pkg/api/api.go
```diff
@@ -80,6 +88,35 @@ func (s *server) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
 	return t, false, err
 }
 
+var errInvalidChunkAddress = errors.New("invalid chunk address")
```
acud, line 91: pull to the top of the file in a 
```go
var (
... ...
)
```
format

pkg/api/api.go
```diff
@@ -80,6 +88,35 @@ func (s *server) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
...
+
+	// If no resolver is not available, return an error.
+	if s.Resolver == nil {
+		return swarm.ZeroAddress, errNoResolver
+	}
+
+	// Try and resolve the name using the provided resolver.
+	log.Debugf("name resolve: attempting to resolve %s to bzz address", str)
+	addr, err = s.Resolver.Resolve(str)
+	if err == nil && !addr.IsZero() {
```
acud, line 112: try to keep happy path non-branching. if the error is not nil, then return the error, otherwise return the address. also, if the error is nil, the address should not be the zero address. if it is then we have a bug. we should try to mitigate overarching conditions

janos:
It would be good to have a default chain of resolvers as a fallback, similar as we have in swarm. So that for example `--resolver-options "https://cloudflare-eth.com"` would work. Currently it is not possible to have such resolver configuration.

pkg/api/api.go
```diff
@@ -80,6 +88,35 @@ func (s *server) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
...
+		log.Tracef("name resolve: resolved name %s to %s", str, addr)
+		return addr, nil
+	}
+
+	return swarm.ZeroAddress, fmt.Errorf("%w: %v", errInvalidChunkAddress, err)
```
acud, line 117: i don't understand this annotation, and i'm not sure why you are using one error to annotate another. also the assumption that the chunk address is invalid seems somewhat naive. `Resolve` cannot return other errors?

pkg/api/api_test.go
```diff
@@ -45,3 +53,79 @@ func newTestServer(t *testing.T, o testServerOptions) *http.Client {
...
+			tC.log = logging.New(ioutil.Discard, 0)
+		}
+		if tC.res == nil && !tC.noResolver {
+			tC.res = resolverMock.NewResolver(
+				resolverMock.WithResolveFunc(func(string) (swarm.Address, error) {
+					return tC.wantAdr, nil
+				}))
+		}
+
+		s := api.New(nil, nil, tC.res, nil, tC.log, nil).(*api.Server)
```
acud, line 118: why the cast?

pkg/api/api_test.go
```diff
@@ -45,3 +53,79 @@ func newTestServer(t *testing.T, o testServerOptions) *http.Client {
 		}),
 	}
 }
+
+func TestParseName(t *testing.T) {
```
acud, line 57: it would be nice to have a test where you explicitly try to parse a hash and assert that the resolver _was not_ called. a problem we had in old swarm is that we used to try to query and resolve normal hashes too. this was later discovered by infura, not us.

janos (approved):
I think that it is good enough for the release. I agree with @acud comments, but I would suggest that, what is left of the comments, address in a followup pr.

pkg/api/api_test.go
paxthemax: Currently, the `server` struct is internal to the api packagem and the `ResolveNameOrAddress` is it's member. This is admittedly a patchwork solution, but a PR for refactoring is already open #637, which will solve this issue.

pkg/api/api.go
paxthemax: I think this is a bug in the ENS resolver, but I haven't been able to track it down before the release. I've opened a separate PR to handle it #638

pkg/api/api_test.go
paxthemax: Good point, current fallback (default) validator may still allow this behavior. The test for this, however, should be in the `resolver` package.

paxthemax:
@janos, @zelig, @acud - thank you for your comments.

I've opened separate issues to resolve some of your comments outside the scope of the release: #637 #638

paxthemax:
This should fix #638

pkg/api/api_test.go
acud: i don't see what you mean. `api.New` returns the interface, not the unexported type:
`func New(tags *tags.Tags, storer storage.Storer, corsAllowedOrigins []string, logger logging.Logger, tracer *tracing.Tracer) Service`

PR #640, opened 2020-08-31
Fix loop break bug in discover

pradovic:
1. Change kademlia code to connect to bootnodes without notification 
2. Fix loop bug in `Discover`

### DIFF
--- a/pkg/p2p/discover.go
+++ b/pkg/p2p/discover.go
@@ -14,7 +14,7 @@ import (
 	madns "github.com/multiformats/go-multiaddr-dns"
 )
 
-func Discover(ctx context.Context, addr ma.Multiaddr, f func(ma.Multiaddr) (stop bool, err error)) (stopped bool, err error) {
+func Discover(ctx context.Context, addr ma.Multiaddr, f func(ma.Multiaddr) (bool, error)) (bool, error) {
 	if comp, _ := ma.SplitFirst(addr); comp.Protocol().Name != "dnsaddr" {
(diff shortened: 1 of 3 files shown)


janos (approved):
Thank you @pradovic.

acud (approved):

pkg/p2p/discover.go
```diff
@@ -14,7 +14,7 @@ import (
 	madns "github.com/multiformats/go-multiaddr-dns"
 )
 
-func Discover(ctx context.Context, addr ma.Multiaddr, f func(ma.Multiaddr) (stop bool, err error)) (stopped bool, err error) {
+func Discover(ctx context.Context, addr ma.Multiaddr, f func(ma.Multiaddr) (bool, error)) (bool, error) {
```
acud, line 17: would be nice at some point to clean this signature up... not relevant for this PR

janos: @acud would you create the issue with that to clean up, so that we do not forget?

pradovic: Oups, sorry, did not see this.. what should we clean-up there? Do you think I should bring back the named parameters? I did not like cuz they are used inside in some checks and I did not like them initializxed before

janos: In general I am fine with removing the named parameters. In this case they were providing some form of documentation, as the bools were described with them. But I have nothing agains that you removed them and only line in the comment to describe them would be nice.

PR #641, opened 2020-08-31
Remove connect notify

pradovic:
Potentially remove connect notify from toplogy driver as a cleanup, as it is used only in debugapi.

TODO:
Polish up the p2pMock a little bit and check toplogy notifier in `peer_tests` in debugapi.

### DIFF
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -239,26 +239,30 @@ func (k *Kad) Start(ctx context.Context) error {
 
 func (k *Kad) connectBootnodes(ctx context.Context) {
 	var count int
-
 	for _, addr := range k.bootnodes {
-		if count == 3 {
+		if count >= 3 {
 			return
 		}
 
 		if _, err := p2p.Discover(ctx, addr, func(addr ma.Multiaddr) (stop bool, err error) {
 			k.logger.Tracef("connecting to bootnode %s", addr)
-			_, err = k.p2p.ConnectNotify(ctx, addr)
+			bzzAddress, err := k.p2p.Connect(ctx, addr)
 			if err != nil {
 				if !errors.Is(err, p2p.ErrAlreadyConnected) {
 					k.logger.Debugf("connect fail %s: %v", addr, err)
 					k.logger.Warningf("connect to bootnode %s", addr)
+					return false, err
 				}
 				return false, nil
+			}
+
+			if err := k.connected(ctx, bzzAddress.Overlay); err != nil {
+				return false, err
 			}
 			k.logger.Tracef("connected to bootnode %s", addr)
 			count++
 			// connect to max 3 bootnodes
-			return count == 3, nil
+			return count >= 3, nil
 		}); err != nil {
 			k.logger.Debugf("discover fail %s: %v", addr, err)
 			k.logger.Warningf("discover to bootnode %s", addr)
@@ -427,6 +431,8 @@ func (k *Kad) AddPeers(ctx context.Context, addrs ...swarm.Address) error {
 		k.knownPeers.Add(addr, po)
 	}
 
+	fmt.Println("triggering manage C from addPeers")
+
 	select {
 	case k.manageC <- struct{}{}:
 	default:
@@ -437,6 +443,20 @@ func (k *Kad) AddPeers(ctx context.Context, addrs ...swarm.Address) error {
 
 // Connected is called when a peer has dialed in.
 func (k *Kad) Connected(ctx context.Context, addr swarm.Address) error {
+	if err := k.connected(ctx, addr); err != nil {
+		return err
+	}
+
+	fmt.Println("triggering manage C from connected")
+	select {
+	case k.manageC <- struct{}{}:
+	default:
+	}
+
+	return nil
+}
+
+func (k *Kad) connected(ctx context.Context, addr swarm.Address) error {
 	if err := k.announce(ctx, addr); err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


pkg/p2p/p2p.go
```diff
@@ -17,9 +17,6 @@ import (
 // Service provides methods to handle p2p Peers and Protocols.
 type Service interface {
 	AddProtocol(ProtocolSpec) error
-	// ConnectNotify connects to the given multiaddress and notifies the topology once the
-	// peer has been successfully connected.
-	ConnectNotify(ctx context.Context, addr ma.Multiaddr) (address *bzz.Address, err error)
```
janos, line 22: Thank you @pradovic for taking time to clean this method. This change makes this interface more clear. 👍

janos (approved):
Great! LGTM

pkg/p2p/p2p.go
pradovic: No problem. I think there is still some problems with this kind of architecture in which there are notifiers on the incoming connections but not on the outgoing connections:
1. Now the caller, in this case `debugapi` needs to notify the appropriate components
2. If there are both calls present, as it was, then this is only partially solved. It is solved from the debug_api buyt not in the case of callers of `p2p.Connect`.
3. In both  cases of the second argument, the caller of connect actually needs to know all the notifiers set on the `p2p` in order to mimic the same behavior of the incoming connection. This leads to a potentially problems in the case of multiple notifiers
4. Some duplicated code, as it can be seen in `kademlia` service which is a bit tricky to follow and maintain.

To solve this problem I would propose to have a more granular notifiers, incoming, ougoing on which the services can register + the init method in protocolSpec.

@acud, @janos LMNWYT :)

janos: I support the Init method for ProtoclSpec, and also the Close/OnDisconnect/Teardown (not sure about the name). A feature like this is already required for other features, like https://github.com/ethersphere/bee/pull/585#issuecomment-682599786.

acud: > 2. If there are both calls present, as it was, then this is only partially solved. It is solved from the debug_api buyt not in the case of callers of p2p.Connect.
3. In both cases of the second argument, the caller of connect actually needs to know all the notifiers set on the p2p in order to mimic the same behavior of the incoming connection. This leads to a potentially problems in the case of multiple notifiers

I'm not sure if I'm following you here. what does these two points mean?

> To solve this problem I would propose to have a more granular notifiers, incoming, outgoing on which the services can register + the init method in protocolSpec.

Can you please explain what you mean by `more granular notifiers`?  an example with the interface you intend to expose might be useful

pkg/kademlia/kademlia.go
```diff
@@ -239,26 +239,30 @@ func (k *Kad) Start(ctx context.Context) error {
...
 			if err != nil {
 				if !errors.Is(err, p2p.ErrAlreadyConnected) {
 					k.logger.Debugf("connect fail %s: %v", addr, err)
 					k.logger.Warningf("connect to bootnode %s", addr)
+					return false, err
 				}
 				return false, nil
 			}
+
+			if err := k.connected(ctx, bzzAddress.Overlay); err != nil {
```
acud, line 259: this has the unwanted side effect that you would try to announce a bootnode. presumably unnecessary

acud (approved):
LGTM. See my comments about further changes to this

pkg/p2p/p2p.go
pradovic: 1 & 2
```
// node.go
p2p.SetNotifier(notifier1)
p2p.SetNotifier(notifier2)
p2p.SetNotifier(notifier3)
// some component
p2p.Connect()
notifier1.Connected()
notifier2.Connected()
// oops forgot notifier3
// oops all notifier need to have exposed connected method
```

3.
```
type Interface struct {
OnConnectedIn
OnConnectedOut
OnDisconnect
}

type ProtocolSpec struct {
// what we already have in p2p.go
Init() // should either be both OnConnectedIn & OnConnectedOut  or a bit more sophisticated
}
```

4. Async idempotent wonderland
```
type Interface struct {
OnConnected
OnDisconnected
} 

// protocols are written in the way that they threat both inbound & outbound connections the same
// when registering protocols, we can pay attention to the order in which they are registered and call handlers in the same order
// different types of errors could indicate if the loop should be broken (and disconnected for example) or just continue // disregarding  that error happen in one particular protocol

```

pkg/kademlia/kademlia.go
pradovic: I noticed this but did not want to change the existing logic. The bootnodes used to be connected with connectNotify which triggers connected handler, which I believe also announces. `k.connect` & `k.Connected` both announces. I really think it is a bad approach to have to think about these different cases all the time. It is very easy to make a mistake and really hard to see the implications of adding some code to kademlia service, for example.
So, if you are sure that announce is not required, I would erase it gladly. But then I need to copy some code again, wihich is different only in announce.

acud: as we discussed before, different connection flows have different requirements, so as much as I'm a big fan of refactoring and reusing of repeated code, sometimes it is inevitable.

pkg/p2p/p2p.go
acud: about `notifyPeerSig` - this is totally subjective, also `k.connectPeers.Add` is called from different places so maybe we should introduce an abstraction for that too? I'm not sure what are your arguments here. You are free to refactor the package as you wish (although it is questionable whether we should pour engineering hours over something that works).

pradovic: In `idempotent wonderland` example, this would be only on one place probably + I already mentioned before that it's hard to tackle a problem with having abstractions in sync (peer registry & kademlia), so that I would actually rely on peer registry, but let's not get back there :) I will merge the PR :)

PR #666, opened 2020-09-02
bump bee version

acud:


janos (approved):

svetomir (commented):
What do you think about setting it to 0.2.1 so we can have new tag (0.2.1) with it?

acud:
@svetomir i was actually thinking to change it to 0.3.0 since we are technically now already on `0.3.0-unstable`

svetomir:
> @svetomir i was actually thinking to change it to 0.3.0 since we are technically now already on `0.3.0-unstable`

That works, too :) @janos @vandot What do you think about it?

janos:
@svetomir @acud makes sense. I had the same thought.

acud:
pushed to `0.3.0`. ok to merge?

svetomir (approved):

janos (approved):

PR #667, opened 2020-09-02
seekjoiner: fill read buffer, integrate langos

acud:
This PR improves the seekjoiner to fill the buffer it has been provided with, and does so in parallel, optimizing the retrievals from the network.


Eknir:
This PR is open for 20 days. What are we doing with it?

Eknir:
This PR improves my download speed from 10kb/s to 200kb/s

janos (approved):
I can confirm that parallel retrievals are happening as it is visible on tracing charts, and that the speed improvement when downloading a file is around 25x, from 10KB/s to 250KB/s. It is an amazing improvement. 👏

acud:
> I can confirm that parallel retrievals are happening as it is visible on tracing charts, and that the speed improvement when downloading a file is around 25x, from 10KB/s to 250KB/s. It is an amazing improvement.

we are slowly entering the DSL age :D

pradovic (approved):
Nice!

janos:
Measuring download time on staging cluster. In all download attempts, the same node key was used and localstore and statestore were deleted.

The measurement was done with wget downloading a file `f624c1cfe80e9dfd9f2eaa35abf5d9a48b3d750407ff748ac25f580587b075de` on the staging cluster.

## getFileBufferSize = 4 * 32 * 1024

Speed: 346KB/s
Time: 4m 24s


## getFileBufferSize = 8 * 32 * 1024

Speed: 451KB/s
Time: 2m 39s


## getFileBufferSize = 16 * 32 * 1024

Speed: 665KB/s
Time: 2m 7s


## getFileBufferSize = 32 * 32 * 1024

Speed: 691KB/s
Time: 2m 6s



Give that the buffer size influences the user experience when downloading, the download happens in steps of buffer size, and that at start joiner does have much less parallelization than when it goes deeper, I would suggest to have an adaptable buffer size in downloadHandler depending on the file size that is downloaded. Basically, to choose the buffer size from 4, 8 or 16 times 32kB in relation to the file size. I would not suggest buffer of 32*32kB as it is very large and it looks like that 16*32kB is the saturation point that I managed to get. 16*32kb is 512kB which is quite large and would use that for files larger than 10MB, for example.

zelig (approved):

PR #670, opened 2020-09-03
Resolver refactor service

### LINKED ISSUE #637: Refactor resolver service package
# Task
Resolver package currently exposes a `service` package, which is redundant.
Refactor the resolver package so that it does not expose superfluous packages or interfaces.

## Acceptance criteria
Superfluous package and interfaces are removed. All tests pass, functionality is not modified.

paxthemax:
# Refactor resolver

Simplify and refactor resolver implementation:

- Remove redundant `service` package
- Remove 'Connect' from interface
- Remove locks
- Simplify `ens` package
- Extract `multierror` package

fixes #637.


pkg/resolver/multiresolver/multiresolver.go
```diff
@@ -0,0 +1,212 @@
...
+
+	for _, chain := range mr.resolvers {
+		for _, r := range chain {
+			if err := r.Close(); err != nil {
+				errs.Append(err)
+			}
+		}
+	}
+
+	return errs.WrapErrorOrNil(ErrCloseFailed)
```
janos, line 194: I am not sure if ErrCloseFailed is needed, as the caller already knows that this is a Close method call that it returned error. It is a redundant information.

pkg/resolver/multiresolver/multiresolver.go
```diff
@@ -0,0 +1,212 @@
...
+	for _, res := range chain {
+		addr, err = res.Resolve(name)
+		if err == nil {
+			return addr, nil
+		}
+		errs.Append(err)
+	}
+
+	return addr, errs.WrapErrorOrNil(
+		fmt.Errorf("resolve chain for tld %q: %w", tld, ErrResolverChainFailed))
```
janos, line 179: Also ErrResolverChainFailed looks redundant, as if the error is nil, it is know to the caller that Resolve function failed.

pkg/resolver/multiresolver/multiresolver.go
```diff
@@ -0,0 +1,212 @@
...
+		switch c.TLD {
+		case "eth":
+			mr.connectENSClient(c.TLD, c.Endpoint)
+		case "rsk":
+			mr.connectENSClient(c.TLD, c.Endpoint)
+		case "":
+			mr.connectENSClient("", c.Endpoint)
+		default:
+			log.Errorf("default domain resolution not supported")
+		}
```
janos, line 92: Do we need this specific tld handling here? Should we just relay what is passed in the config?

paxthemax: The case can be useful if different resolver clients are selected based on the TLD. The intent was to use a `RSKClient` for RNS domains etc.

paxthemax: I will remove this for now, until it is required.

pkg/resolver/multiresolver/multiresolver.go
paxthemax: Does the caller need to filter using `errors.Is`, specifically for `Close`-type errors? Clients may return different errors, this should wrap it at the top level.

If, however, this is not idiomatic (or applicable to the project) - I'm happy to remove it.

pkg/resolver/multiresolver/multiresolver.go
paxthemax: See above, same issue.

pkg/resolver/multiresolver/multiresolver.go
janos: Right, but RNS resolver should be the same as ENS resolver, just needs to be tested. This whole switch statement can be replaced with just `mr.connectENSClient(c.TLD, c.Endpoint)`. We can add more complex handling when is actually needed.

pkg/resolver/multiresolver/multiresolver.go
janos: Since this error is only returned by the Close method, `errors.Is` is not needed, `err != nil` gives the same information that there has been an error. Clients can return different errors, but I see not point to have a special error like ErrCloseFailed if that error is returned by only one method. I do not see that it provides any new information.

zelig (approved):
@paxthemax @janos I really like this refactor and the much improved/simplified code

pkg/resolver/multiresolver/multiresolver.go
paxthemax: Removed.

pkg/resolver/multiresolver/multiresolver.go
paxthemax: Removed.

janos (approved):
Thanks @paxthemax 👍

acud (approved):

PR #672, opened 2020-09-03
add opentracing instrumentation to http package

acud:
Adds OpenTracing instrumentation to the http package.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -129,3 +129,15 @@ func requestModePut(r *http.Request) storage.ModePut {
 func requestEncrypt(r *http.Request) bool {
 	return strings.ToLower(r.Header.Get(SwarmEncryptHeader)) == "true"
 }
+
+func (s *server) instrumentOpenTracing() func(h http.Handler) http.Handler {
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			spanName := fmt.Sprintf("http.%s", r.Method)
+			span, _, ctx := s.Tracer.StartSpanFromContext(r.Context(), spanName, s.Logger)
+			defer span.Finish()
+
+			h.ServeHTTP(w, r.WithContext(ctx))
+		})
+	}
+}
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -130,10 +130,10 @@ func requestEncrypt(r *http.Request) bool {
 	return strings.ToLower(r.Header.Get(SwarmEncryptHeader)) == "true"
 }
 
-func (s *server) instrumentOpenTracing() func(h http.Handler) http.Handler {
+func (s *server) newTracingHandler(spanTag string) func(h http.Handler) http.Handler {
 	return func(h http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			spanName := fmt.Sprintf("http.%s", r.Method)
+			spanName := fmt.Sprintf("http.%s.%s", r.Method, spanTag)
 			span, _, ctx := s.Tracer.StartSpanFromContext(r.Context(), spanName, s.Logger)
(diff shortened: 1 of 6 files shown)


pkg/api/api.go
```diff
@@ -129,3 +129,15 @@ func requestModePut(r *http.Request) storage.ModePut {
 func requestEncrypt(r *http.Request) bool {
 	return strings.ToLower(r.Header.Get(SwarmEncryptHeader)) == "true"
 }
+
+func (s *server) instrumentOpenTracing() func(h http.Handler) http.Handler {
```
janos, line 133: This is actually a constructor. I am suggesting to call it `newTracingHandler`

pkg/api/router.go
```diff
@@ -37,21 +37,35 @@ func (s *server) setupRouting() {
 	})
 
 	handle(router, "/files", jsonhttp.MethodHandler{
-		"POST": http.HandlerFunc(s.fileUploadHandler),
+		"POST": web.ChainHandlers(
+			s.instrumentOpenTracing(),
+			http.HandlerFunc(s.fileUploadHandler),
```
janos, line 42: This should be `web.FinalHandlerFunc(s.fileUploadHandler)`, like in other places below where compiler complains.

pkg/api/api.go
```diff
@@ -129,3 +129,15 @@ func requestModePut(r *http.Request) storage.ModePut {
...
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			spanName := fmt.Sprintf("http.%s", r.Method)
```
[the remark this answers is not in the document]
acud: great suggestion. was not sure how to get around this. thanks :+1:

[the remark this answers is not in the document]
acud: i just add the name of the endpoint after the method. i think it should be clear by `http.get.files` what the api was requested with, not sure if we need more stuff like `http.get.files-download`, seems to be duplicate information there. wdyt?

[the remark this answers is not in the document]
janos: We have two methods for uploading and downloading. I think that have different tracing for them so they should have a different name. I am not sure if we need `http.{method}` as the method is the formal concept, having just `files-download` is more clear to me instead `http.get.files`, where `get` may relate to list or be associated with some other semantics. I am just thinking how would someone understand the span name the first time in jeager for example, without the context of how download is implemented.

pkg/api/api.go
```diff
@@ -129,3 +129,15 @@ func requestModePut(r *http.Request) storage.ModePut {
 func requestEncrypt(r *http.Request) bool {
 	return strings.ToLower(r.Header.Get(SwarmEncryptHeader)) == "true"
 }
+
+func (s *server) newTracingHandler(spanTag string) func(h http.Handler) http.Handler {
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			spanName := fmt.Sprintf("http.%s.%s", r.Method, spanTag)
```
janos, line 136: I think there is no need for http.$method prefix, basically `spanName := spanTag` is just enough.

pkg/api/router.go
```diff
@@ -63,7 +78,10 @@ func (s *server) setupRouting() {
 	})
 
 	handle(router, "/bzz/{address}/{path:.*}", jsonhttp.MethodHandler{
-		"GET": http.HandlerFunc(s.bzzDownloadHandler),
+		"GET": web.ChainHandlers(
+			s.newTracingHandler("bzz-download"),
+			web.FinalHandlerFunc(http.HandlerFunc(s.bzzDownloadHandler)),
```
janos, line 83: `http.HandlerFunc` can be dropped in all changes in router.

```suggestion
			web.FinalHandlerFunc(s.bzzDownloadHandler),
```

janos (approved):
Thanks @acud. LGTM.

PR #673, opened 2020-09-04
move chunks-pin endpoints from debugapi to api

### LINKED ISSUE #669: Move chunks-pin endpoints from debugapi to api
All chunks-pin endpoints should be moved from debugapi to the main api and a [gateway mode flag](https://beehive.ethswarm.org/swarm/pl/hzdc83wza7niijxmrap6cgn11r) should be implemented to disable them on our public gateways. More information https://beehive.ethswarm.org/swarm/pl/hzdc83wza7niijxmrap6cgn11r.

janos:
This PR moves chunks-pin endpoints from debugapi to api as per #669.

fixes #669


pkg/api/router.go
```diff
@@ -81,6 +81,15 @@ func (s *server) setupRouting() {
 		),
 	})
 
+	handle(router, "/chunks-pin/{address}", jsonhttp.MethodHandler{
```
acud, line 84: maybe just `/pin`?

acud (approved):
LGTM. Not sure about the endpoint name though. Is this how it is specified in the book?

pkg/api/router.go
janos: We should have also endpoints to pin bytes, files and collections. I also do not like so much chunks-pin. Maybe to have:

```
/pinning/chunks
/pinning/bytes
/pinning/files
/pinning/dirs
```

For now just `/pinning/chunks`?

acud: SGTM :+1:

janos: Cool, thanks. Changed.

metacertain (approved):

ldeffenb:
Is there a master list of changes like this that break existing (debug) API users?  My app was using /chunks-pin/{address} and all of a sudden it's getting back the following:

{
  "message": "Not Found",
  "code": 404
}

A search through the closed pull requests found this, but most mere mortals might not be so lucky...

janos:
@ldeffenb I am sorry for this change to cause trouble and your suggestion is very good. We should have this type of documentation and also it would be even more useful in this case to have a deprecation error instead of just not found. I think that this will be even more import when we pass the beta stage of development. Maybe we should version the debugapi as well as the main api.

Issue #668, opened 2020-09-02
Gateway mode flag


janos:
Bee should have a gateway mode flag to disable certain features for nodes that run on our public gateway. More information https://beehive.ethswarm.org/swarm/pl/hzdc83wza7niijxmrap6cgn11r.

PR #675, opened 2020-09-07
api gateway mode

janos:
This PR ads a single api and cli option to disable some features, most useful when exposing the api in public.

Gateway mode disables in the API level:

- tags endpoints
- pinning chunks endpoints
- encryption on upload endpoints
- pinning on upload endpoints

Additionally, this PR adds api.Options with the new Gateway option and already existing cors headers as both being optional features.

fixes: #668


acud (approved):
Great @janos. Thanks

pradovic (approved):

PR #676, opened 2020-09-07
Instant chunk repair

jmozah:
As explained in the "instant chunk repair" section of Architecture proposal https://hackmd.io/NSYAwNabQjakaAYDficraw. 
Also see Fig:2. for strategy


pkg/netstore/netstore.go
```diff
@@ -51,13 +73,34 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
 					return nil, err
 				}
-				go func() {
-					err := s.recoveryCallback(addr, targets)
-					if err != nil {
-						s.logger.Debugf("netstore: error while recovering chunk: %v", err)
+				chunkC := make(chan swarm.Chunk, 1)
+				socAddress, err := s.recoveryCallback(addr, targets, chunkC)
+				if err != nil {
+					s.logger.Debugf("netstore: error while recovering chunk: %v", err)
```
janos, line 79: The error is just logged and execution of the function continues, but it should not. Error should be returned.

pkg/recovery/repair.go
```diff
@@ -64,7 +89,7 @@ func NewRepairHandler(s storage.Storer, logger logging.Logger, pushSyncer pushsy
 		}
 
 		// retrieve the chunk from the local store
-		ch, err := s.Get(ctx, storage.ModeGetRequest, swarm.NewAddress(chAddr))
+		ch, err := s.Get(ctx, storage.ModeGetRequest, addrToRetrieve)
 		if err != nil {
 			logger.Tracef("chunk repair: error while getting chunk for repairing: %v", err)
 			return err
```
janos, line 95: This is missed in some previous pr. Errors should not be logged and returned.

pkg/recovery/repair.go
```diff
@@ -76,6 +101,107 @@ func NewRepairHandler(s storage.Storer, logger logging.Logger, pushSyncer pushsy
 			logger.Tracef("chunk repair: error while sending chunk or receiving receipt: %v", err)
 			return err
```
janos, line 102: This is missed in some previous pr. Errors should not be logged and returned.

janos (changes requested):
There is a reoccurring pattern in PRs for a long time that errors are logged and returned. Please, correct these trivial and common comments first, then I will continue with my review.

pkg/netstore/netstore.go
```diff
@@ -23,16 +26,35 @@ type store struct {
...
+		recoveryCallback: rcb,
+		retrieval:        r,
+		logger:           logger,
+		validator:        validator,
+		chunkChanMap:     make(map[string]chan swarm.Chunk),
+		chunkChanMu:      &sync.RWMutex{},
+	}
+}
+
+func SetTimeout(timeout time.Duration) {
```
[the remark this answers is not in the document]
jmozah: This is called in netstore also, that why it is exported.

> Or maybe, to have repairTimeout as a configurable field in store.

do you mean have it as a configurable value from command line?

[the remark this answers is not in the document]
janos: > This is called in netstore also, that why it is exported.

It is in the `netstore`, that is the point of my comment.

> 
> > Or maybe, to have repairTimeout as a configurable field in store.
> 
> do you mean have it as a configurable value from command line?

No, as I said, the field to the store, so that it can be configured in netstore constructor for tests. That is just another approach instead the package global variable.

pkg/netstore/netstore_test.go
```diff
@@ -187,15 +245,36 @@ func (r *retrievalMock) RetrieveChunk(ctx context.Context, addr swarm.Address) (
...
+		return dummyChunk
+	}
+	signer := crypto.NewDefaultSigner(privKey)
+	id := make([]byte, 32)
+	sch, err := soc.NewChunk(id, c, signer)
+	if err != nil {
+		return dummyChunk
+	}
+	return sch
+}
```
janos, line 280: It would be nice that the errors are propagated, or at least to have this function as test helper with errors handled.

pkg/recovery/repair.go
```diff
@@ -5,57 +5,81 @@
...
+			return errInvalidEnvelope
+		}
+
+		// get the envelopeAddress and the payloadId from the soc envelope
+		cursor := 0
+		envelopeAddress := make([]byte, swarm.SectionSize)
+		envelopePayloadId := make([]byte, swarm.SectionSize)
+		copy(envelopeAddress, m.Payload[cursor:swarm.SectionSize])
+		cursor += swarm.SectionSize + soc.IdSize + soc.SignatureSize + swarm.SpanSize
+		copy(envelopePayloadId, m.Payload[cursor:cursor+swarm.SectionSize])
```
janos, line 77: It would be nice to have this encoding logic encapsulated in a function.

pkg/recovery/repair.go
```diff
@@ -64,18 +88,113 @@ func NewRepairHandler(s storage.Storer, logger logging.Logger, pushSyncer pushsy
...
+		data := make([]byte, soc.IdSize+soc.SignatureSize+swarm.SpanSize+len(ch.Data()))
+		cursor = 0
+		envelopeCursor := swarm.SectionSize
+		copy(data[cursor:cursor+soc.IdSize], m.Payload[envelopeCursor:envelopeCursor+soc.IdSize])
+		cursor += soc.IdSize
+		envelopeCursor += soc.IdSize
+		copy(data[cursor:cursor+soc.SignatureSize], m.Payload[envelopeCursor:envelopeCursor+soc.SignatureSize])
+		cursor += soc.SignatureSize
+		copy(data[cursor:cursor+len(ch.Data())], ch.Data())
+		socChunk := swarm.NewChunk(swarm.NewAddress(envelopeAddress), data)
```
janos, line 112: It would be nice to have this encoding logic encapsulated in a function.

pkg/recovery/repair.go
```diff
@@ -64,18 +88,113 @@ func NewRepairHandler(s storage.Storer, logger logging.Logger, pushSyncer pushsy
...
+	s := soc.New(id, chunk)
+	err := s.AddSigner(signer)
+	if err != nil {
+		return nil, err
+	}
+	socEnvelope, err := s.ToChunk()
+	if err != nil {
+		return nil, err
+	}
+	return socEnvelope, nil
```
janos, line 186:
```suggestion
	return s.ToChunk()
```

pkg/soc/soc_test.go
```diff
@@ -122,4 +128,7 @@ func TestFromChunk(t *testing.T) {
 	if !bytes.Equal(ownerEthereumAddress, u2.OwnerAddress()) {
 		t.Fatalf("owner address mismatch %x %x", ownerEthereumAddress, u2.OwnerAddress())
 	}
+
+	fmt.Println(ch.Address(), ch.Data())
+	fmt.Println(u2.Chunk.Address(), u2.Chunk.Data())
```
janos, line 133: Remove Println calls and properly test values.

pkg/recovery/repair_test.go
```diff
@@ -119,41 +233,66 @@ func TestRecoveryHookCalls(t *testing.T) {
...
 		defer mockStorer.Close()
-		_, err := mockStorer.Put(context.Background(), storage.ModePutRequest, c1)
+		_, err = mockStorer.Put(context.Background(), storage.ModePutRequest, c1)
 		if err != nil {
 			t.Fatal(err)
 		}
 
 		// create a mock pushsync service to push the chunk to its destination
 		var receipt *pushsync.Receipt
+		var socChunk swarm.Chunk
```
zelig, line 258: call it  recoveryResponse or something

pkg/recovery/repair_test.go
```diff
@@ -119,41 +233,66 @@ func TestRecoveryHookCalls(t *testing.T) {
...
+		id := make([]byte, 32)
+		data := make([]byte, swarm.SpanSize+swarm.SectionSize)
+		binary.BigEndian.PutUint64(data[:swarm.SpanSize], swarm.SectionSize)
+		copy(data[swarm.SpanSize:swarm.SpanSize+swarm.SectionSize], c1.Address().Bytes())
+		c := swarm.NewChunk(c1.Address(), data)
+		sch, err := soc.NewChunk(id, c, signer)
+		if err != nil {
+			t.Fatal(err)
+		}
+		maxPayload := append(sch.Address().Bytes(), sch.Data()...)
```
zelig, line 288: maxPayload?

pkg/recovery/repair_test.go
zelig: call     `pushSyncService` `mockPushSyncService`

pkg/recovery/repair_test.go
```diff
@@ -169,11 +308,26 @@ func TestNewRepairHandler(t *testing.T) {
 			t.Fatalf("invalid address in receipt: expected %s received %s", c1.Address(), receipt.Address)
 		}
 
+		data = socChunk.Data()
```
zelig, line 311: receipt checks are not needed since you conscruc t them

zelig (commented):
this pr needs a lot of work still

pkg/netstore/netstore.go
```diff
@@ -51,13 +67,36 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
-					if err != nil {
-						s.logger.Debugf("netstore: error while recovering chunk: %v", err)
+				chunkC := make(chan swarm.Chunk, 1)
+				fmt.Println("----- Calling recovery")
+				socAddress, err := s.recoveryCallback(addr, targets, chunkC)
+				if err != nil {
+					return nil, err
+				}
+
+				// add the expected soc address and the channel in which the chunk is awaited
```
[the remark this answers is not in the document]
jmozah: Well you don't know the envelope address before you call this function. This is as soon as we can add it.

pkg/netstore/netstore.go
```diff
@@ -51,13 +67,36 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
+
+				// wait for chunk or timeout
+				var socChunk swarm.Chunk
+				select {
+				case socChunk = <-chunkC:
+					fmt.Println("---- Got chunk repaired")
+				case <-time.After(s.repairTimeout):
+					if chunkC, ok := s.chunkChanMap[socAddress.String()]; ok {
+						close(chunkC)
+						delete(s.chunkChanMap, socAddress.String())
```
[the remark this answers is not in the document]
jmozah: oops... good catch on delete

pkg/netstore/netstore.go
```diff
@@ -78,6 +117,16 @@ func (s *store) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chun
 		if !s.validator.Validate(ch) {
 			return nil, storage.ErrInvalidChunk
 		}
+
+		// if the chunk is expected as part of the repair process, then deliver it
+		s.chunkChanMu.Lock()
```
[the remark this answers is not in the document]
jmozah: Yeah.. it is validated as SOC in the previous step...

pkg/netstore/netstore.go
```diff
@@ -78,6 +117,16 @@ func (s *store) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chun
...
+		key := ch.Address().String()
+		if chunkC, ok := s.chunkChanMap[key]; ok {
+			chunkC <- ch
+			close(chunkC)
```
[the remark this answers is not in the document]
jmozah: is it not good to close a channel after its lifetime is done, so that it will be GCd?

pkg/netstore/netstore_test.go
```diff
@@ -97,20 +101,74 @@ func TestNetstoreNoRetrieval(t *testing.T) {
...
+	if !bytes.Equal(c.Data(), rcvdChunk.Data()) {
+		t.Fatal()
+	}
+}
+
+// TestRecoveryTimeout verifies if the timout function of the netstore works properly
+func TestRecoveryTimeout(t *testing.T) {
+	hookWasCalled := make(chan bool, 1)
+	payload := make([]byte, swarm.ChunkSize)
+	_, err := rand.Read(payload)
```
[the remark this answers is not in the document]
jmozah: There is a GenerateTestRandomChunk function but it generate an invalid chunk i.e. it is not content addressed. It fails in most of the recovery testcases..   

But in this particular test.. we can use it.. i just avoided it.

pkg/netstore/netstore_test.go
```diff
@@ -97,20 +101,74 @@ func TestNetstoreNoRetrieval(t *testing.T) {
...
+	}
+
+	select {
+	case <-hookWasCalled:
+		break
+	case <-time.After(100 * time.Millisecond):
+		t.Fatal("recovery hook was not called")
+	}
+
+	if !c.Address().Equal(rcvdChunk.Address()) {
```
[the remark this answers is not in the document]
jmozah: It make sense because, the actual chunk is embedded in to a SOC in the recovery function and unwrapped in the Get function. so this test the unwrapping if nothing else.

pkg/netstore/netstore_test.go
```diff
@@ -187,15 +245,39 @@ func (r *retrievalMock) RetrieveChunk(ctx context.Context, addr swarm.Address) (
 }
 
 type mockRecovery struct {
-	hookC chan bool
+	hookC     chan bool
```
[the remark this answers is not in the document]
jmozah: Thats what TestRecovery does.

pkg/netstore/netstore_test.go
```diff
@@ -187,15 +245,39 @@ func (r *retrievalMock) RetrieveChunk(ctx context.Context, addr swarm.Address) (
...
+		return sch.Address(), nil
+	}
+	return swarm.NewAddress([]byte("deadbeef")), nil
 }
 
 func (r *mockRecovery) RetrieveChunk(ctx context.Context, addr swarm.Address) (chunk swarm.Chunk, err error) {
 	return nil, fmt.Errorf("chunk not found")
 }
+
+func getSocChunk(c swarm.Chunk) (swarm.Chunk, error) {
```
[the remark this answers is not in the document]
jmozah: well this is the envelope in which the chunk is embedded.. This should be done because the recovery response will come like this only.

pkg/recovery/repair.go
```diff
@@ -5,57 +5,80 @@
...
+		}
+
+		// get the envelopeAddress and the payloadId from the soc envelope
+		envelopeAddress, envelopePayloadId := getAddressAndPayloadId(m.Payload)
 
 		// check if the chunk exists in the local store and proceed.
 		// otherwise the Get will trigger a unnecessary network retrieve
-		exists, err := s.Has(ctx, swarm.NewAddress(chAddr))
+		addrToRetrieve := swarm.NewAddress(envelopePayloadId)
+		exists, err := s.Has(ctx, addrToRetrieve)
```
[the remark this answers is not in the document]
jmozah: good idea... but i would like to push that as part of another PR.

pkg/netstore/netstore.go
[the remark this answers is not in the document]
jmozah: i delete in normal scenario in PUT :-)

pkg/node/node.go
```diff
@@ -270,7 +272,7 @@ func NewBee(addr string, swarmAddress swarm.Address, keystore keystore.Service,
 
 	if o.GlobalPinningEnabled {
 		// register function for chunk repair upon receiving a trojan message
-		chunkRepairHandler := recovery.NewRepairHandler(ns, logger, pushSyncProtocol)
+		chunkRepairHandler := recovery.NewRepairHandler(ns, pushSyncProtocol)
```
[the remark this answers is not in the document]
jmozah: i think one doesn't make send without the other. So keeping them under one flag is the best in my opinion.

pkg/recovery/repair_test.go
```diff
@@ -42,8 +56,24 @@ func TestRecoveryHook(t *testing.T) {
...
+	signer := crypto.NewDefaultSigner(privKey)
+	publicKey, err := signer.PublicKey()
+	if err != nil {
+		t.Fatal(err)
+	}
+	overlayBytes, err := crypto.NewEthereumAddress(*publicKey)
+	if err != nil {
+		t.Fatal(err)
+	}
+	overlayAddres := swarm.NewAddress(overlayBytes)
```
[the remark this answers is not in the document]
jmozah: ethereum address is to create the owner and also to have the overlay derived from signer.

pkg/recovery/repair_test.go
```diff
@@ -63,9 +156,16 @@ type recoveryHookTestCase struct {
 
 // TestRecoveryHookCalls verifies that recovery hooks are being called as expected when net store attempts to get a chunk.
 func TestRecoveryHookCalls(t *testing.T) {
-	// generate test chunk, store and publisher
-	c := chunktesting.GenerateTestRandomChunk()
-	ref := c.Address()
+	//// generate test chunk, store and publisher
```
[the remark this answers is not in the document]
jmozah: Yeah.. there recovery is mocked.. here the actual recovery function is tested.

pkg/recovery/repair_test.go
```diff
@@ -119,41 +233,66 @@ func TestRecoveryHookCalls(t *testing.T) {
...
 			t.Fatal(err)
 		}
 
 		// create a mock pushsync service to push the chunk to its destination
 		var receipt *pushsync.Receipt
+		var socChunk swarm.Chunk
 		pushSyncService := pushsyncmock.New(func(ctx context.Context, chunk swarm.Chunk) (*pushsync.Receipt, error) {
-			receipt = &pushsync.Receipt{
-				Address: swarm.NewAddress(chunk.Address().Bytes()),
+			if receipt == nil {
```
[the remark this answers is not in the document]
jmozah: This tests the RepairHandler, and if that functions properly the message is sent to push sync. So push sync and receipt is mocked here

pkg/recovery/repair_test.go
jmozah: it is needed as the receipt is constructed from the address that is send to PSS. If the Repairhandler sends a wrong chunk address.. then receipt will fail...

pkg/netstore/netstore.go
```diff
@@ -23,16 +26,25 @@ type store struct {
...
 var (
-	ErrRecoveryAttempt = errors.New("failed to retrieve chunk, recovery initiated")
+	errDecodingRecoveryResponse = errors.New("error decoding recovery response")
 )
 
 // New returns a new NetStore that wraps a given Storer.
 func New(s storage.Storer, rcb recovery.RecoveryHook, r retrieval.Interface, logger logging.Logger,
-	validator swarm.Validator) storage.Storer {
-	return &store{Storer: s, recoveryCallback: rcb, retrieval: r, logger: logger, validator: validator}
+	validator swarm.Validator, timeout time.Duration) storage.Storer {
```
janos, line 39: `timeout` argument is not used.

pkg/netstore/netstore.go
```diff
@@ -47,17 +59,34 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
+				socAddress, err1 := s.recoveryCallback(ctx, addr, targets, chunkC)
+				if err1 != nil {
+					return nil, err1
+				}
+
+				// add the expected soc address and the channel in which the chunk is awaited
+				s.chunkChanMu.Lock()
+				s.chunkChanMap[socAddress.String()] = chunkC
+				s.chunkChanMu.Unlock()
+				defer delete(s.chunkChanMap, socAddress.String())
```
janos, line 76: This deferred delete on chunkChanMap is not protected by the lock.

pkg/netstore/netstore.go
```diff
@@ -47,17 +59,34 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
+				var recoveryResponse swarm.Chunk
+				select {
+				case recoveryResponse = <-chunkC:
+				case <-ctx.Done():
+					return nil, err // return the original retrieval error
+				}
+				s, err := soc.FromChunk(recoveryResponse)
 				if err != nil {
-					return nil, err
+					return nil, errDecodingRecoveryResponse
```
janos, line 87: The actual error is not propagated, it is shadowed with errDecodingRecoveryResponse, which is actually not checked anywhere.

pkg/node/node.go
```diff
@@ -252,14 +252,16 @@ func NewBee(addr string, swarmAddress swarm.Address, keystore keystore.Service,
 	var ns storage.Storer
 	if o.GlobalPinningEnabled {
 		// create recovery callback for content repair
-		recoverFunc := recovery.NewRecoveryHook(psss)
-		ns = netstore.New(storer, recoverFunc, retrieve, logger, chunkvalidator)
+		recoverFunc := recovery.NewRecoveryHook(psss, swarmAddress, signer)
+		// because the miner timeout is 20 seconds, we give enough time for the miner to be over
+		repairTimeout := 20 * time.Second
```
janos, line 257: repairTimeout is not used in netstore and can be removed.

pkg/recovery/repair.go
```diff
@@ -64,18 +91,40 @@ func NewRepairHandler(s storage.Storer, logger logging.Logger, pushSyncer pushsy
...
+
+		// construct a new soc chunk from the soc envelope by adding the chunk data to the envelope
+		recoveryResponse := newRecoveryResponseChunk(envelopeAddress, m.Payload, ch.Data())
+
+		// send the soc envelope which has the chunk data, to the requester
+		_, err = pushSyncer.PushChunkToClosest(ctx, recoveryResponse)
+		if err != nil {
 			return err
 		}
 		return nil
```
janos, line 113:
```suggestion
		return err
```

pkg/soc/soc.go
```diff
@@ -232,3 +235,84 @@ func recoverAddress(signature, digest []byte) ([]byte, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		return s.ToChunk()
+	}
+	return mine(ctx, f)
+}
+
+// mine finds out the nonce that is valid to create a envelope address
+func mine(ctx context.Context, f func(nonce []byte) (swarm.Chunk, error)) (swarm.Chunk, error) {
```
janos, line 267: This function looks the same as https://github.com/ethersphere/bee/pull/695/files#diff-3fc846f43f3c15fe3d5e0d8f0ad7b4bcR217 in a different PR. Consider reusing the same function when both PRs get merged, instead copying it.

janos (commented):
There are some comments by @zelig left unaddressed.

pkg/soc/soc.go
jmozah: whichever comes first :-)

pkg/recovery/repair_test.go
```diff
@@ -229,19 +269,31 @@ func TestNewRepairHandler(t *testing.T) {
...
-		// create the chunk repair handler
-		repairHandler := recovery.NewRepairHandler(mockStorer, logger, pushSyncService)
-
 		//create a trojan message to trigger the repair of the chunk
 		testTopic := trojan.NewTopic("foo")
-		maxPayload := make([]byte, swarm.SectionSize)
-		var msg trojan.Message
-		copy(maxPayload, c3.Address().Bytes())
-		msg, err = trojan.NewMessage(testTopic, maxPayload)
+		privKey, err := crypto.GenerateSecp256k1Key()
```
zelig, line 274: but  why have  this test, to check if pushsync error is p
you do not  need this to test `recieve error`

pkg/soc/soc_test.go
```diff
@@ -122,4 +129,111 @@ func TestFromChunk(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	// with target len 1
+	ctx := context.Background()
+	envelope1, err := soc.CreateSelfAddressedEnvelope(ctx, owner, overlayAddres, 1, c1.Address(), signer)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if envelope1.Address().Bytes()[0] != overlayAddres.Bytes()[0] {
```
zelig, line 174: check a matching prefix of length `targetLen`

pkg/soc/soc_test.go
```diff
@@ -122,4 +129,111 @@ func TestFromChunk(t *testing.T) {
...
+		t.Fatal(" invalid prefix mined")
+	}
+
+	// with target len 2
+	envelope2, err := soc.CreateSelfAddressedEnvelope(ctx, owner, overlayAddres, 2, c1.Address(), signer)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if envelope2.Address().Bytes()[0] != overlayAddres.Bytes()[0] {
```
zelig, line 183: check a matching prefix of length targetLen

zelig (changes requested):

pkg/recovery/repair.go
jmozah: why is lint not catching this... is code simplification not enabled in our lint?

pkg/netstore/netstore.go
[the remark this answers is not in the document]
jmozah: i understand what you say. But it is hard to execute this only in the case of SOC chunk without changing the validator interface? i am not sure about changing that interface. @janos WDYT?

The only other way to find it as a SOC chunks is to check the payload length. But it is not guaranteed to pass for all SOC chunks.

pkg/recovery/repair_test.go
jmozah: This test was written during normal chunk repair and i just modified it to adjust to instant chunk repair. I wanted to cover all error conditions during the test. i will remove it if you feel it is not required.

pkg/netstore/netstore.go
jmozah: good catch. Thanks

pkg/netstore/netstore.go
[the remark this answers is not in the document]
janos: We already can use soc.Validator.Validate function. There is no need to extend he interface, just use the specific validator that already exists.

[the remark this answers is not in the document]
jmozah: @janos That validator is called in the above few lines and the control comes down only if the SOC chunk is valid.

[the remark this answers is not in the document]
janos: I know, the validator is returning true if a chunk is valid of any type. I am saying that we already have a way to check if a chunk is valid for a specific type without changing the interface in this pr.

If you like to introduce the validator change in this PR I think that it will be out of scope. I would propose to create a PR where a validator would work in a way to return which types a chunk is identified (validated) as.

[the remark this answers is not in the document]
jmozah: Even though i think that adding a SOC validate will be more costlier than checking the `chunkChanMap` for every chunk, i accept this comment and added a SOC validate. I also added a comment there saying that the validate should spew out the chunk type in a future PR.

pkg/netstore/netstore.go
```diff
@@ -74,10 +102,28 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 // returns a storage.ErrInvalidChunk error when
 // encountering an invalid chunk.
 func (s *store) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
+	var chunks []swarm.Chunk
 	for _, ch := range chs {
 		if !s.validator.Validate(ch) {
 			return nil, storage.ErrInvalidChunk
 		}
+
+		// TODO: his double validate in case of a SOC chunk should be avoided in future by
```
zelig, line 111: no  future sorry, i really cannot afford technical debt like this going  in the code for a not so important feature,, especially when it is not needed.
put the callback within the validator or make the validator return type and condition on that etc

pkg/recovery/repair_test.go
```diff
@@ -42,8 +56,24 @@ func TestRecoveryHook(t *testing.T) {
 	}
 
 	// create recovery hook and call it
-	recoveryHook := recovery.NewRecoveryHook(pssSender)
-	if err := recoveryHook(chunkAddr, targets); err != nil {
+	privKey, err := crypto.GenerateSecp256k1Key()
```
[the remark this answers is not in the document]
zelig: ^^^

pkg/recovery/repair_test.go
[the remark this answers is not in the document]
zelig: you  should not test stuff that you explicitly extend the test  with

pkg/recovery/repair_test.go
```diff
@@ -183,111 +203,52 @@ func TestNewRepairHandler(t *testing.T) {
...
-
-		// create the chunk repair handler
-		repairHandler := recovery.NewRepairHandler(mockStorer, logger, pushSyncService)
-
-		//create a trojan message to trigger the repair of the chunk
-		testTopic := trojan.NewTopic("foo")
-		maxPayload := make([]byte, swarm.SectionSize)
-		var msg trojan.Message
-		copy(maxPayload, c3.Address().Bytes())
+		maxPayload := append(sch.Address().Bytes(), sch.Data()...)
```
zelig, line 223: maxPayload?

pkg/soc/soc.go
zelig: janos left a comment here about  not copying..........

zelig (commented):
starting to doubt this pr should make it. too much technical debt

pkg/netstore/netstore.go
janos: I have suggested that validator is not changed in this pr. It is already getting large and validator change should add even more changes across the codebase. I think that it should be done in a different pr, before or after this pr is merged.

jmozah:
> starting to doubt this pr should make it. too much technical debt

Sorry @zelig. Except [this](https://github.com/ethersphere/bee/pull/676#discussion_r490787609) what are the other technical debts.

pkg/recovery/repair_test.go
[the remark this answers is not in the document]
jmozah: it is not tested.. since the pushsync is called twice... this mockpushsyncserice is handling both the case..
in the first case it should send the receipt, in the second case it (if the receipt is sent), it should send the receovery response.

pkg/netstore/netstore.go
jmozah: Changed the Validator interface to expose chunk type also.

pkg/soc/soc.go
jmozah: pointed this to tojan.Mine

jmozah:
@janos i suggest to close this PR and open it when required. Infact i would say we open a bug to fix the existing Global pinning and fix that first;

janos:
Closing this PR but the branch should be kept as a reference. Thanks @jmozah for the work done here.

PR #677, opened 2020-09-07
integrate clef signer

### LINKED ISSUE #139: Key management
## User-story
As a node operator, I don't want the Swarm node to directly manage my private keys, such that it is easier to validate what is done with my private keys and when.

## Background
- [whiteboarding](https://hackmd.io/aOn6nJxvTJ2t50rRH3iqfw?view)

ralph-pichler:
This PR introduces clef as a signer backend for the overlay address. It also changes the signing format of the in-memory signer for compatibility by adding the Ethereum message prefix prior to hashing and moving the recovery value to the end of the encoded signature.

Clef must be enabled using the `--clef-signer-enable` flag. The path to the ipc socket can be specified using `--clef-signer-endpoint` flag, if this is not specified it will use the default location for the os.

A complete bash script to initialise and startup clef with the correct rules can be found here: https://gist.github.com/ralph-pichler/d3cd67b2325ca5a057d67fdb4027ec80. Alternatively you can start clef interactively like described here: https://github.com/ralph-pichler/swap-clef-test.

fixes #139


acud:
@ralph-pichler i will review this in more detail tomorrow, but just on sidenote - maybe it would be good to maintain (in this repo or another) an example/baseline rule file?

pkg/crypto/clef.go
```diff
@@ -0,0 +1,91 @@
+package crypto
```
janos, line 1: Add copyright header.

pkg/crypto/clef.go
```diff
@@ -0,0 +1,91 @@
...
+	"os"
+	"path/filepath"
+	"runtime"
+
+	"github.com/ethereum/go-ethereum/accounts"
+)
+
+var (
+	ErrNoAccounts       = errors.New("no accounts found in clef")
+	ClefRecoveryMessage = []byte("public key recovery message")
```
janos, line 15: This variable does not have to be exported, it is used only in tests, so it can be exported with export_test.go only.

pkg/crypto/clef_test.go
```diff
@@ -0,0 +1,101 @@
+package crypto_test
```
janos, line 1: Add copyright header.

pkg/crypto/clef.go
janos, line 1: It looks to me that clef signer implementation can be in its own package and to drop Clef in all identifiers here. Also clef imports go-ethereum, and if somewhere crypto package is imported for anything other than clef, go-ethereum would be an unused dependency.

janos (commented):
Very nice, @ralph-pichler!

zelig (approved):
great stuff

cmd/bee/cmd/cmd.go
```diff
@@ -42,6 +42,8 @@ const (
 	optionNamePaymentTolerance     = "payment-tolerance"
 	optionNameResolverEndpoints    = "resolver-options"
 	optionNameGatewayMode          = "gateway-mode"
+	optionNameClefSignerEnable     = "clef-signer-enable"
```
acud, line 45: do we need both flags? maybe infer that `clef-signer-enabled` whenever `clef-signer-endpoint` is defined?

acud (approved):
Brilliant

pkg/crypto/clef.go
ralph-pichler: fixed

pkg/crypto/clef.go
ralph-pichler: fixed

pkg/crypto/clef_test.go
ralph-pichler: fixed

pkg/crypto/clef.go
ralph-pichler: created a clef package and moved everything there

janos (approved):
Cool, thanks @ralph-pichler.

cmd/bee/cmd/cmd.go
ralph-pichler: To not infer things seems more consistent with how the other "enable"-flags in bee work. E.g. `debug-api-addr`does not infer `debug-api-enable` and `tracing-endpoint` does not infer `tracing-enable` (I think).

pkg/crypto/clef/clef.go
```diff
@@ -0,0 +1,96 @@
...
+	Accounts() []accounts.Account
+}
+
+type clefSigner struct {
+	clef    ExternalSignerInterface
+	account accounts.Account // the account this signer will use
+	pubKey  *ecdsa.PublicKey // the public key for the account
+}
+
+// DefaultIpcPath returns the os-dependent default ipc path for clef
```
paxthemax, line 35:
```

paxthemax (commented):
Looks good to go.

Annoying pedantics: comments should be complete sentences (ending with period).

Issue #683, opened 2020-09-08
Print git commit version on `bee version`


acud:
Add the possibility to print the git commit on `bee version` ([ref](https://github.com/ethersphere/bee/commit/dc1951843baeebbe7dc417216e86cc5ab6cd9fc4#diff-b67911656ef5d18c4ae36cb6741b7965L4)).

We should also add to api the possibility to get a version of a running bee node (possibly under health check or under heartbeat result).

acud:
(discussion at https://github.com/ethersphere/bee/issues/664)

PR #684, opened 2020-09-08
add git commit to version and expose version in debugapi

janos:
This is a simple change to include the git commit when building with Makefile and .git directory is available very similar to what we had before https://github.com/ethersphere/bee/commit/dc1951843baeebbe7dc417216e86cc5ab6cd9fc4#diff-b67911656ef5d18c4ae36cb6741b7965L4, but if the COMMIT environment variable is provided, the git command will not be executed. When building a docker image COMMIT environment must be provided as .git directory is in .dockerignore, this is not changed.

fixes: #683


acud (approved):
very cool. thanks janos

svetomir (approved):

PR #686, opened 2020-09-09
api: add pss websockets api

### LINKED ISSUE #650: PSS public facing API
support pss notifications through websockets

acud:
Test code in https://github.com/acud/bee-pss-websockets

fixes #650

manual testing:
- [x] open websocket handler, check that connection keepalive is ok for several minutes
- [x] open websocket handler, get a message, do the same (check keepalive), receive another message, check keepalive
- [x] open ws, close the bee node, check that the close message arrives
- [x] open ws, kill the client (no close message)
- [x] open ws, client gone gracefully (close message arrives at bee)

### DIFF
--- a/pkg/api/pss.go
+++ b/pkg/api/pss.go
@@ -6,7 +6,10 @@ package api
 
 import (
 	"context"
+	"encoding/hex"
+	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/ethersphere/bee/pkg/jsonhttp"
@@ -29,6 +32,8 @@ var (
 
 	// Send pings to peer with this period. Must be less than pongWait.
 	pingPeriod = (pongWait * 9) / 10
+
+	targetMaxLength = 2 // max target length in bytes, in order to prevent grieving by excess computation
 )
 
 type PssMessage struct {
@@ -37,21 +42,64 @@ type PssMessage struct {
 }
 
 func (s *server) pssPostHandler(w http.ResponseWriter, r *http.Request) {
+	t, ok := mux.Vars(r)["topic"]
+	if !ok {
+		s.Logger.Error("pss send: no topic")
+		jsonhttp.BadRequest(w, nil)
+		return
+	}
+	topic := trojan.NewTopic(t)
 
+	tg, ok := mux.Vars(r)["targets"]
+	if !ok {
+		s.Logger.Error("pss send: no targets")
+		jsonhttp.BadRequest(w, nil)
+		return
+	}
+	var targets trojan.Targets
+	tgts := strings.Split(tg, ",")
+	for _, v := range tgts {
+		target, err := hex.DecodeString(v)
+		if err != nil || len(target) > targetMaxLength {
+			s.Logger.Debugf("pss send: bad targets: %v", err)
+			s.Logger.Error("pss send: bad targets")
+			jsonhttp.BadRequest(w, nil)
+			return
+		}
+		targets = append(targets, target)
+	}
+
+	payload, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		s.Logger.Debugf("pss read payload: %v", err)
+		s.Logger.Error("pss read payload")
+		jsonhttp.InternalServerError(w, nil)
+		return
+	}
+
+	err = s.Pss.Send(r.Context(), targets, topic, payload)
+	if err != nil {
+		s.Logger.Debugf("pss send payload: %v", err)
+		s.Logger.Error("pss send payload")
+		jsonhttp.InternalServerError(w, nil)
+		return
+	}
+
+	jsonhttp.OK(w, nil)
 }
 
 func (s *server) pssWsHandler(w http.ResponseWriter, r *http.Request) {
 	conn, err := upgrader.Upgrade(w, r, nil)
 	if err != nil {
-		logger.Debugf("pss ws: upgrade: %v", err)
-		logger.Error("pss ws: cannot upgrade")
+		s.Logger.Debugf("pss ws: upgrade: %v", err)
+		s.Logger.Error("pss ws: cannot upgrade")
 		jsonhttp.InternalServerError(w, nil)
 		return
 	}
 
 	t, ok := mux.Vars(r)["topic"]
 	if !ok {
-		logger.Error("pss ws: no topic")
+		s.Logger.Error("pss ws: no topic")
 		jsonhttp.BadRequest(w, nil)
 		return
 	}
@@ -59,7 +107,7 @@ func (s *server) pssWsHandler(w http.ResponseWriter, r *http.Request) {
 	go s.pumpWs(conn, t)
 }
 
-func (s *server) pumpWs(conn *websocket.Conn, topic string) {
+func (s *server) pumpWs(conn *websocket.Conn, t string) {
 	defer s.wsWg.Done()
 
 	var (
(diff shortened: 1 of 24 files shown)


pkg/api/pss_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/pss"
+	resolverMock "github.com/ethersphere/bee/pkg/resolver/mock"
+	"github.com/ethersphere/bee/pkg/storage/mock"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/ethersphere/bee/pkg/trojan"
+	"github.com/gorilla/websocket"
+)
+
+func newTestWsServer(t *testing.T, o testServerOptions) *httptest.Server {
```
janos, line 22: This looks the same as the test server constructor, just returns the server. Nothing specific to ws, even more, there should not be a need for a specific ws test server. Existing newTestServer can just return also the server or ws client if needed.

pkg/api/router.go
```diff
@@ -85,6 +85,11 @@ func (s *server) setupRouting() {
 		),
 	})
 
+	handle(router, "/pss/{topic}/ws", jsonhttp.MethodHandler{
```
janos, line 88: This should be (POST,DELETE) `/pss/subscribe/{topic}` for subscriptions and there should be `/pss/send/{topic}` for sending.

acud: cool. i think `DELETE` is not necessary since the client can just hang-up the websocket, or the server can also close the websocket on shutdown. cc @zelig

janos: Right, good point @acud.

pkg/api/router.go
```diff
@@ -85,6 +85,17 @@ func (s *server) setupRouting() {
...
 
+	handle(router, "/pss/send/{topic}/{targets}", jsonhttp.MethodHandler{
+		"POST": web.ChainHandlers(
+			jsonhttp.NewMaxBodyBytesHandler(swarm.ChunkSize),
+			web.FinalHandlerFunc(s.pssPostHandler),
+		),
+	})
+
+	handle(router, "/pss/subscribe/{topic}", jsonhttp.MethodHandler{
+		"GET": http.HandlerFunc(s.pssWsHandler),
```
zelig, line 96: this should also be `POST` no? 
you  convinced me that there  is no  need for `DELETE` to cancel a subscription

pkg/pss/pss_test.go
```diff
@@ -146,23 +104,99 @@ func TestDeliver(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	checkCalls(t, h1Calls, 2)
+	checkCalls(t, h2Calls, 0)
+	checkCalls(t, h3Calls, 1)
 
-	// set handler for test topic
-	pss.Register(testTopic, testHandler)
+	cleanup() // remove the last handler
```
zelig, line 169: should this not be called with defer before fatal  calls after ?

zelig (approved):
a top quality PR. as long as it is  integration tested i approve

pkg/pss/pss_test.go
acud: well not explicitly since `pss.Register` does not take any resources per-se, i.e. it does not start any goroutine for the handlers on registration, only on notification. That's why I omitted the cleanups on the other handlers

pkg/api/router.go
acud: > this should also be POST no?

Tried it. Didn't work. Also the websocket library does not offer any notion of HTTP verbs. Since WS is an actual protocol, I would not assume that it behaves as HTTP does.

zelig: ?? I mean creating   thewwebsocket should be 

     POST /pss/subscribe/{topic}

acud: because when you create the websocket you don't `POST`. As a client you don't even use any HTTP notions at all, you dial to `ws://...`, you don't `HTTP POST /path/to/endpoint` and magically get a websocket back

acud: as of such, you dont even need to mention the HTTP verb in the book spec. You just specify `ws://localhost:8080/pss/subscribe/<topic>` and that's that

janos: `jsonhttp.MethodHandler` can be even removed here.

zelig: i understood that this is an http API endpoint. otherwise what are the http-s all  around it?  i  must be missing  something, but  i should  do my homework first sorry guys for interfering

pkg/api/pss.go
```diff
@@ -0,0 +1,169 @@
...
+		return
+	}
+	topic := trojan.NewTopic(t)
+
+	tg, ok := mux.Vars(r)["targets"]
+	if !ok {
+		s.Logger.Error("pss send: no targets")
+		jsonhttp.BadRequest(w, nil)
+		return
+	}
```
janos, line 50: Checks for mux vars are not needed. They are always provided by the mux router.

pkg/api/pss.go
```diff
@@ -0,0 +1,169 @@
...
+			jsonhttp.BadRequest(w, nil)
+			return
+		}
+		targets = append(targets, target)
+	}
+
+	payload, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		s.Logger.Debugf("pss read payload: %v", err)
+		s.Logger.Error("pss read payload")
```
janos, line 67: A topic can be a useful information in logs in this handler.

pkg/api/pss.go
```diff
@@ -0,0 +1,169 @@
...
+		jsonhttp.InternalServerError(w, nil)
+		return
+	}
+
+	t, ok := mux.Vars(r)["topic"]
+	if !ok {
+		s.Logger.Error("pss ws: no topic")
+		jsonhttp.BadRequest(w, nil)
+		return
+	}
```
janos, line 97: No need to check for mux var.

pkg/api/pss.go
```diff
@@ -0,0 +1,169 @@
...
+		case <-gone:
+			// client gone
+			return
+		case <-ticker.C:
+			err = conn.SetWriteDeadline(time.Now().Add(writeDeadline))
+			if err != nil {
+				s.Logger.Debugf("pss set write deadline: %v", err)
+				return
+			}
+			if err = conn.WriteMessage(websocket.PingMessage, nil); err != nil {
```
janos, line 164: A comment that this error is not useful to log here would be good.

pkg/api/pss_test.go
```diff
@@ -0,0 +1,393 @@
...
+	m, err := trojan.NewMessage(topic, payload)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	tc, err = m.Wrap(targets)
+	if err != nil {
+		t.Fatal(err)
+	}
+	<-time.After(500 * time.Millisecond) // wait to see that the websocket is kept alive
```
janos, line 266: This is the same as calling time.Sleep.

pkg/api/pss_test.go
```diff
@@ -0,0 +1,393 @@
...
+			// ignore pings
+			continue
+		}
+
+		if message != nil {
+			mtx.Lock()
+			copy(targetContent, message)
+			mtx.Unlock()
+		}
+		<-time.After(50 * time.Millisecond)
```
janos, line 303: time.Sleep here also.

pkg/api/pss_test.go
acud: personal preference to using channels :) anyway i changed to Sleeps

janos (approved):

PR #693, opened 2020-09-11
log pss error instead of stream reset

ralph-pichler:
Log errors from the pss delivery handler instead of returning them to the push sync handler. This is to avoid them causing a stream reset which sometimes prevents the receiving of the receipt by the other peer.

The loglevel was chosen as `DEBUG` because this is how it would be logged before this PR (as an `error handle protocol`).


janos (approved):

pradovic (approved):
Is it ok to do this for any error or do we need to catch a specific case?

PR #700, opened 2020-09-17
pipeline hashing: refactor to different packages

### LINKED ISSUE #619: pipeline splitter tasks
Still todo in the context of pipeline splitter:
- [x] split writers to separate packages, add unit tests
- [x] remove dependency on encryption package, have composable encryption and add unit tests for encryption with mock encryption
- [ ] Change enclosing Interface to be of hash.Hash (optional)
- [x] Possible minor cleanup on the pipeWriteArgs struct. However, for building the merkle trie it would be nicer to have span and ref separate
- [ ]  swarm.Chunk spans are uint64, but in some places I have a feeling we're still doing int64 due to io.Writer
- [ ] Benchmarks
- [x] https://github.com/ethersphere/bee/pull/603#discussion_r476301769

ref #610 #603 #618  #613 #582

acud:
Following the architectural proposal documented [here](https://hackmd.io/2Q3NjLESTi2drTVnOqLAVw?view), this PR pulls out the different components to their respective packages within the `pipeline` package. Unit tests have been added to the respective packages.

fixes #619 (other future tasks have been extracted to separate issues)

The followup tasks that have been created are #708 #709 #710.

As documented in #710, there is still a missing test to the hashtrie writer, which can be used to get someone else familiar with this part of the codebase.


pkg/encryption/chunk_encryption.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/ethersphere/bee/pkg/swarm"
+	"golang.org/x/crypto/sha3"
+)
+
+// ChunkEncrypter encrypts chunk data.
+type ChunkEncrypter interface {
+	EncryptChunk([]byte) (key Key, encryptedSpan, encryptedData []byte, err error)
+}
+
+type ce struct{}
```
janos, line 13: It can be called chunkEncryptor, for clarity, even the code is short, but requires more reading, than to just read the name.

pkg/encryption/mock/chunk_encryption.go
```diff
@@ -0,0 +1,24 @@
+package mock
```
janos, line 1: Add copyright header.

pkg/encryption/mock/chunk_encryption.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"github.com/ethersphere/bee/pkg/encryption"
+)
+
+type ce struct {
```
janos, line 7: Also chunkEncryptor name would be nicer.

pkg/encryption/chunk_encryption.go
```diff
@@ -0,0 +1,37 @@
...
+	encryptedData, err := newDataEncryption(key).Encrypt(chunkData[8:])
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	return key, encryptedSpan, encryptedData, nil
+}
+
+func newSpanEncryption(key Key) Interface {
+	refSize := int64(swarm.HashSize + KeyLength)
+	return New(key, 0, uint32(swarm.ChunkSize/refSize), sha3.NewLegacyKeccak256)
```
zbiljic, line 32: We seem to be increasing number of uses of `sha3.NewLegacyKeccak256` (2 here).

It might be prudent at some point to switch to specific other one. Since other `swarm` constant is used, maybe we could use `swarm.NewHasher` intead.

pkg/file/pipeline/builder/builder.go
```diff
@@ -0,0 +1,128 @@
...
+	var total int64
+	data := make([]byte, swarm.ChunkSize)
+	var eof bool
+	for !eof {
+		c, err := r.Read(data)
+		total += int64(c)
+		if err != nil {
+			if err == io.EOF {
+				if total < dataLength {
+					return swarm.ZeroAddress, fmt.Errorf("pipline short write: read %d out of %d bytes", total+int64(c), dataLength)
```
zbiljic, line 85: Why doing again `total+int64(c)` in error message ?
Isn't this already done in previous line, right before checking for error.

pkg/file/pipeline/builder/builder.go
```diff
@@ -0,0 +1,128 @@
...
+				if c > 0 {
+					cc, err := pipeline.Write(data[:c])
+					if err != nil {
+						return swarm.ZeroAddress, err
+					}
+					if cc < c {
+						return swarm.ZeroAddress, fmt.Errorf("pipeline short write: %d mismatches %d", cc, c)
+					}
+				}
+				continue
```
zbiljic, line 97: I think I understand this, however would it not be same to just have `break` here (instead if `continue`), and completely remove `eof` variable ?

pkg/encryption/chunk_encryption.go
acud: This was already thoroughly discussed. As you may notice, it is the official golang `sha3` library that has this name. This is the keccak256 standard that is defined as the sha3 flavor we're using and that's how they call it. I'm not sure if we should wrap or try to alter this. It came up several times but I'm not sure if and what we should do with this.

zbiljic (approved):

janos (approved):
Thanks @acud!

acud:
@zelig can you submit your pending review?

PR #703, opened 2020-09-17
chequebook deployment

### LINKED ISSUE #659: Deploy and fund chequebook on first startup
# Task

A swap chequebook needs to be deployed on startup if one already does not already exist.

- [x] Integrate `go-sw3` library with bee.
- [x] Add a cli options to specify eth endpoint and factory address
- [x] On startup check if chequebook address already in store
  - [x] if not, deploy one with the factory and store that address
  - [x] flag to not do this (zero eth entry), or pass existing chequebook address
  - [x] prefund chequebook with tokens
  - [x] if there is a chequebook verify that instead

ralph-pichler:
Adds chequebook deployment and initial top-up. 

Added features:
* On startup the node checks if a chequebook has been previously deployed otherwise it will deploy one. If `--swap-initial-deposit` is specified with a non-0 value it will also transfer bzz token from the overlay to the new chequebook.
* Adds a temporary `--swap-enable` flag so integration tests don't break instantly due to lack of blockchain backend. With the flag a working blockchain connection and funds to deploy the chequebook are necessary for starting the node.
* The bytecode of the supplied factory is verified and your own chequebook is verified against it.
* Signer interface has been extended with a `SignTx` function.
* As there are no hard-coded factories at the moment, factory addresses need to be passed with `--swap-factory-address` for all networks. For goerli this is `0xc5a26Fc99B40C7fe3174D9311Aa803e88be78c73`. A factory can be setup on a local ganache with the script from [here](https://gist.github.com/ralph-pichler/dc0386f81ffa09c220271d4389b4a139), this will also output the environment variables you can use to configure bee.

closes #659


Eknir:
Thanks for this PR!

My findings:

Good:
- The node doesn't boot up without Gorli Eth
- The node boots up after funding my address with Gorli Eth and creates chequebook
- The node recognizes that a chequebook was already deployed when restarting

Still some problems:
- I can pass `-swap-initial-deposit`, even without having BZZ tokens on my Gorli address. No log, no error
- If I pass `swap-initial-deposit` after having BZZ tokens (I believe @rpicc funded [my account](https://goerli.etherscan.io/address/0xe62cc17e2ade461d6688a97ceaa933415cb85fbe), I don't get a feedback from the logs that my chequebook was funded.

To me, it seems as if my account doesn't have BZZ (basing myself on Etherscan). In this case, an error log is missing that deposit is failing. I could not try out the case where deposit is succeeding, as I don't have BZZ yet

[My chequebook](https://goerli.etherscan.io/address/0x3b865358e2919d31642bf2a4030ca13bcac4d04c)

ralph-pichler:
`swap-initial-deposit` only does something when initially deploying the chequebook. On further runs it is ignored. To deploy a new chequebook (with initial deposit) you have to delete the state store first.

pkg/settlement/swap/chequebook/init.go
```diff
@@ -0,0 +1,99 @@
...
+			return nil, err
+		}
+
+		chequebookService, err = New(swapBackend, transactionService, chequebookAddress, erc20Address, overlayEthAddress, simpleSwapBindingFunc, erc20BindingFunc)
+		if err != nil {
+			return nil, err
+		}
+
+		if swapInitialDeposit != 0 {
+			logger.Info("depositing into new chequebook")
```
Eknir, line 69: Depositing <amount> into new chequebook

cmd/bee/cmd/cmd.go
```diff
@@ -44,6 +44,10 @@ const (
 	optionNameGatewayMode          = "gateway-mode"
 	optionNameClefSignerEnable     = "clef-signer-enable"
 	optionNameClefSignerEndpoint   = "clef-signer-endpoint"
+	optionNameSwapEndpoint         = "swap-endpoint"
+	optionNameSwapFactoryAddress   = "swap-factory-address"
+	optionNameSwapInitialDeposit   = "swap-initial-deposit"
+	optionNameSwapEnable           = "swap-enable"
```
Eknir, line 50: Let's work together with @vandot to remove this flag asap.

pkg/node/node.go
```diff
@@ -138,6 +146,53 @@ func NewBee(addr string, swarmAddress swarm.Address, keystore keystore.Service,
...
+		}
+		overlayEthAddress, err := signer.EthereumAddress()
+		if err != nil {
+			return nil, err
+		}
+
+		// print ethereum address so users know which address we need to fund
+		logger.Infof("using ethereum address %x", overlayEthAddress)
+
+		// TODO: factory address discovery for well-known networks (goerli for beta)
```
Eknir, line 166: This is important

Eknir (commented):
I can verify that this works (both positive and negative scenarios) on Goerli.

I did not have a thorough look at the implementation (only skimmed through).

I won't approve, as I did not have a look at the implementation.

pkg/node/node.go
```diff
@@ -138,6 +146,53 @@ func NewBee(addr string, swarmAddress swarm.Address, keystore keystore.Service,
...
+
+		if o.SwapFactoryAddress == "" {
+			return nil, errors.New("no known factory address")
+		} else if !common.IsHexAddress(o.SwapFactoryAddress) {
+			return nil, errors.New("invalid factory address")
```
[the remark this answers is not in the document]
ralph-pichler: this already happens in `chequebook.Init`

pkg/node/node.go
ralph-pichler: yes, will do this in a separate PR when we deployed the real token and factory.

cmd/bee/cmd/cmd.go
ralph-pichler: we might want to keep this around for now and just default to true. otherwise every dev will have to spawn local blockchains with factories and token prefundings just to test completely unrelated stuff.

pkg/settlement/swap/chequebook/init.go
Eknir: Aha. Github removed part of my comment. I wanted to write `depositing <amount> into new chequebook`. The capitalization was not the point (and should indeed not be capitalized).

cmd/bee/cmd/cmd.go
Eknir: Agree.

pkg/settlement/swap/chequebook/chequebook.go
```diff
@@ -0,0 +1,137 @@
...
+		Context: ctx,
+	}, s.ownerAddress)
+	if err != nil {
+		return common.Hash{}, err
+	}
+
+	// check we can afford this so we don't waste gas
+	if balance.Cmp(amount) < 0 {
+		return common.Hash{}, errors.New("insufficient token balance")
+	}
```
Eknir, line 97: Would be nice to do this check also before we deploy the cheqeubook (especially given that deposit is only possible initially at the moment).

cmd/bee/cmd/cmd.go
```diff
@@ -190,4 +194,8 @@ func (c *command) setAllFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool(optionNameGatewayMode, false, "disable a set of sensitive features in the api")
 	cmd.Flags().Bool(optionNameClefSignerEnable, false, "enable clef signer")
 	cmd.Flags().String(optionNameClefSignerEndpoint, "", "clef signer endpoint")
+	cmd.Flags().String(optionNameSwapEndpoint, "http://localhost:8545", "swap ethereum blockchain endpoint")
+	cmd.Flags().String(optionNameSwapFactoryAddress, "", "swap factory address")
+	cmd.Flags().Uint64(optionNameSwapInitialDeposit, 0, "initial deposit if deploying a new chequebook")
+	cmd.Flags().Bool(optionNameSwapEnable, false, "enable swap")
```
Eknir, line 200: As we spoke about. Don't remove this flag, but set default to `true` whenever possible from devops (@vandot) side.

https://github.com/ethersphere/bee/pull/703/files#r491862821

pkg/crypto/signer.go
```diff
@@ -82,3 +85,35 @@ func (d *defaultSigner) Sign(data []byte) (signature []byte, err error) {
 	signature[64] = v
 	return signature, nil
 }
+
+func (d *defaultSigner) SignTx(transaction *types.Transaction) (*types.Transaction, error) {
```
Eknir, line 89: Comment missing

pkg/crypto/signer.go
```diff
@@ -82,3 +85,35 @@ func (d *defaultSigner) Sign(data []byte) (signature []byte, err error) {
...
+	v := signature[0]
+	copy(signature, signature[1:])
+	// v value needs to be adjusted by 27 as transaction.WithSignature expects it to be 0 or 1
+	signature[64] = v - 27
+	return transaction.WithSignature(&types.HomesteadSigner{}, signature)
+}
+
+type EthAddress = [20]byte
+
+func (d *defaultSigner) EthereumAddress() (EthAddress, error) {
```
Eknir, line 107: Comment missing

pkg/settlement/swap/chequebook/chequebook.go
```diff
@@ -0,0 +1,137 @@
...
+	})
+}
+
+// WaitForDeposit waits for the deposit transaction to confirm and verifies the result
+func (s *service) WaitForDeposit(ctx context.Context, txHash common.Hash) error {
+	receipt, err := s.transactionService.WaitForReceipt(ctx, txHash)
+	if err != nil {
+		return err
+	}
+	if receipt.Status != 1 {
```
Eknir, line 133: Is it possible to propagate the revert reason?

Eknir (approved):
I went over the implementation, now taking more time.

I am confident to approve it, as I didn't see anything very odd. I verified the logic (manual testing) and ran the tests.

Looking forward to see the other components of settlement hooked-on!

pkg/crypto/signer.go
ralph-pichler: added

pkg/crypto/signer.go
ralph-pichler: added

pkg/settlement/swap/chequebook/chequebook.go
ralph-pichler: no, not returned by most clients (other than ganache)

pkg/crypto/signer.go
```diff
@@ -82,3 +91,37 @@ func (d *defaultSigner) Sign(data []byte) (signature []byte, err error) {
 	signature[64] = v
 	return signature, nil
 }
+
+// SignTx signs an ethereum transaction
+func (d *defaultSigner) SignTx(transaction *types.Transaction) (*types.Transaction, error) {
```
janos, line 96: It would be good to have tests for defultSigner's SignTx and EthereumAddress, maybe just to have the points what should be done in next iterations?

pkg/crypto/signer.go
```diff
@@ -82,3 +91,37 @@ func (d *defaultSigner) Sign(data []byte) (signature []byte, err error) {
...
+
+	// Convert to Ethereum signature format with 'recovery id' v at the end.
+	v := signature[0]
+	copy(signature, signature[1:])
+	// v value needs to be adjusted by 27 as transaction.WithSignature expects it to be 0 or 1
+	signature[64] = v - 27
+	return transaction.WithSignature(&types.HomesteadSigner{}, signature)
+}
+
+type EthAddress = [20]byte
```
janos, line 112: EthAddress is defined here, but in checkbook and other places common.Address is used. Since go-ethereum types is imported here, it should be ok to use common.Address from ethereum also?

pkg/settlement/swap/chequebook/common_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethersphere/bee/pkg/settlement/swap/chequebook"
+	"github.com/ethersphere/sw3-bindings/v2/simpleswapfactory"
+)
+
+type backendMock struct {
```
janos, line 19: It would be good to panic on all not implemented methods for this mock, just to make sure that nobody relies on them in the future, until implemented.

janos (approved):
This is very nice, but rather large PR. LGTM.

pkg/crypto/signer.go
ralph-pichler: I changed it back to `common.Address`.

pkg/settlement/swap/chequebook/common_test.go
ralph-pichler: Done.

pkg/crypto/signer.go
ralph-pichler: I added tests for both

PR #707, opened 2020-09-18
kademlia: TestClosestPeer wait for peers to be added to the backing data structure

### LINKED ISSUE #512: TestClosestPeer flaking
```
=== RUN   TestClosestPeer
##[error]    kademlia_test.go:509: peers not equal. got 4000000000000000000000000000000000000000000000000000000000000000 expected 6000000000000000000000000000000000000000000000000000000000000000
--- FAIL: TestClosestPeer (0.26s)
```

acud:
In `TestClosestPeer`, there was a possibility for a race condition where the `conns` counter that was used has been incremented, but the peer was not yet added to the backing `pslice` data structure.
This is possible since the `conns` counter was incremented in the mock p2p connect function, meaning that kademlia initiated a dial, but the actual addition to `connectedPeers` is done some lines of code after the connection is done on the p2p level, resulting in a data race where indeed 3 connections were initiated, but not all 3 peers are in `connectedPeers`.

This is also evident, since the test was flaking on the first test case:
```
=== RUN   TestClosestPeer
##[error]    kademlia_test.go:509: peers not equal. got 4000000000000000000000000000000000000000000000000000000000000000 expected 6000000000000000000000000000000000000000000000000000000000000000
--- FAIL: TestClosestPeer (0.26s)
```
So it is reasonable to assume that this is happening due to the aforementioned race condition.

fixes #512

### DIFF
--- a/pkg/kademlia/kademlia_test.go
+++ b/pkg/kademlia/kademlia_test.go
@@ -477,9 +477,8 @@ func TestClosestPeer(t *testing.T) {
 
 	disc := mock.NewDiscovery()
 	ab := addressbook.New(mockstate.NewStateStore())
-	var conns int32
 
-	kad := kademlia.New(base, ab, disc, p2pMock(ab, nil, &conns, nil), logger, kademlia.Options{})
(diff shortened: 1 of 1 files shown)


janos (approved):
👍

pradovic (approved):

PR #713, opened 2020-09-18
Add support for manifest metadata

zbiljic:
Includes support for metadata that can be stored in manifest.

Initially it i used to store `index-document` and `error-document` in the manifest.
These two are stored in root path (`/`) entry metadata.
During serving files from paths, on the `bzz` endpoint, if some of these documents are configured, attempt will be made to redirect to some of them (based on the actual state of the request).

Some naming changes are done by this PR (headers), which should be open for discussion.

One additional change was made to `api_test.go`, where additional option was added, that allows caller to disable client redirect. This was done to support some additional tests related to redirection to these mentioned (optional) documents.

### DIFF
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -6,6 +6,7 @@ package api
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -21,6 +22,7 @@ import (
 	"github.com/ethersphere/bee/pkg/jsonhttp"
 	"github.com/ethersphere/bee/pkg/manifest"
 	"github.com/ethersphere/bee/pkg/sctx"
+	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/ethersphere/bee/pkg/tracing"
 )
 
@@ -102,15 +104,15 @@ func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
 
 		indexDocumentSuffixKey := bzzDownloadHandlerManifestRedirect(m, manifestWebsiteIndexDocumentSuffixKey)
 		if indexDocumentSuffixKey != "" {
-			u := r.URL
-			u.Path += "/"
-			u.Path += indexDocumentSuffixKey
-			redirectURL := u.String()
+			pathWithIndex := path.Join(pathVar, indexDocumentSuffixKey)
+			indexDocumentManifestEntry, err := m.Lookup(pathWithIndex)
+			if err == nil {
+				// index document exists
+				logger.Debugf("bzz download: serving path: %s", pathWithIndex)
 
-			logger.Debugf("bzz download: redirecting to %s: %v", redirectURL, err)
-
-			http.Redirect(w, r, redirectURL, http.StatusTemporaryRedirect)
-			return
+				s.bzzDownloadHandlerServeManifestEntry(w, r, ctx, j, address, indexDocumentManifestEntry.Reference())
+				return
+			}
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


acud:
is this ready for review?

pkg/api/bzz.go
```diff
@@ -96,12 +97,69 @@ func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
...
+		indexDocumentSuffixKey := bzzDownloadHandlerManifestRedirect(m, manifestWebsiteIndexDocumentSuffixKey)
+		if indexDocumentSuffixKey != "" {
+			u := r.URL
+			u.Path += "/"
+			u.Path += indexDocumentSuffixKey
+			redirectURL := u.String()
+
+			logger.Debugf("bzz download: redirecting to %s: %v", redirectURL, err)
+
+			http.Redirect(w, r, redirectURL, http.StatusTemporaryRedirect)
```
zelig, line 112: temporary?

zelig (approved):

pkg/api/bzz.go
zbiljic: Yes, that is great question.
I started with temporary redirect, but question should it be replaced by permanent, or just serve the content of `index-document` content ?

janos: As this will redirect `/` to `/index.html`, I think that we should serve the index document, not redirect.

janos (changes requested):
This PR adds redirections for index and error documents. But, instead it should serve the content of that documents at requested path. This is even more important for the error document in order to have js routers working. There is one need for redirection, and that id when accessing the path with index document without trailing slash, then, the redirection should happen to the path with trailing slash in order to support relative html assets paths.

In a few examples:

- request to `/some-path-with-index-document` should redirect to `/some-path-with-index-document/` (support relative asset paths)
- request to `/some-path-with-index-document/` should serve the content of `/some-path-with-index-document/index-document.ext` (serve directly on the path, not redirecting to index.html)
- request to `/some-path-with-error-document/non-existing-manifest-entry` should serve the content of `/some-path-with-error-document/error-document.ext` (support external links for apps with js routers)

zbiljic:
Updated per @janos requested changes. Few PR were created on 'manifest' project side, that were deemed as needed to support relative asset paths.

pkg/api/bzz.go
```diff
@@ -96,24 +103,103 @@ func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
...
+						s.bzzDownloadHandlerServeManifestEntry(w, r, ctx, j, address, indexDocumentManifestEntry.Reference())
+						return
+					}
+				}
+			}
+
+			// check if error document is to be shown
+			errorDocumentPath := bzzDownloadHandlerManifestRedirect(m, manifestWebsiteErrorDocumentPathKey)
+			if errorDocumentPath != "" {
+				if pathVar != errorDocumentPath {
```
acud, line 167: maybe reduce nesting with `if errorDocumentPath != "" && pathVar != errorDocumentPath`?

pkg/api/bzz.go
```diff
@@ -96,24 +103,103 @@ func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
...
 		}
 		return
 	}
 
-	manifestEntryAddress := me.Reference()
+	// serve requested path
+	s.bzzDownloadHandlerServeManifestEntry(w, r, ctx, j, address, me.Reference())
+}
+
+func (s *server) bzzDownloadHandlerServeManifestEntry(
```
acud, line 190: maybe just `func serveManifestEntry` is enough? not sure if we need too much nesting in function naming too

pkg/api/bzz.go
```diff
@@ -156,3 +242,19 @@ func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
 
 	s.downloadHandler(w, r, fileEntryAddress, additionalHeaders)
 }
+
+func bzzDownloadHandlerManifestRedirect(manifest manifest.Interface, metadataKey string) string {
```
acud, line 246: the function name does not describe its purpose very well

acud (commented):
Nice @zbiljic. I have mainly aesthetic comments

janos (approved):
Thanks @zbiljic!

janos:
@zbiljic just have a look at https://deepsource.io/gh/ethersphere/bee/run/9b8ba3ee-59c9-40df-9411-47a4005ef7b6/go/ reports.

zbiljic:
@janos I did not know DeepSource will report it, but I was fully aware of my changes, and I have intentionally left is as-is.

First one (context) was because I figured it would be better to pass on already extracted context even though `http.ResponseWriter` and `http.Request` are forwarded also.
The second reported issue was because after the line became a little too long I have split it into multiple, and having single line `swarm.Address` would join last two (or leave only in one line which was not nice to me).

I do overthink some things maybe. I will fix this.

janos (approved):
I see, makes sense. Deep source is sometimes very aggressive. Thanks for making the change.

PR #715, opened 2020-09-18
Charge for MaxPO chunks

metacertain:
This PR only increases the prices in pricer by 1* poPrice so that maxPO chunks are not unaccounted for


ralph-pichler (approved):
A small change, but an important one. Thanks.

janos (approved):

PR #716, opened 2020-09-18
pushsync: remove some indirection and redundant error logging

acud:


pkg/pushsync/pushsync.go
```diff
@@ -319,10 +300,11 @@ func (ps *PushSync) handleDeliveryResponse(ctx context.Context, w protobuf.Write
...
 	}
 
-	// since all PSS messages comes through push sync, deliver them here if this node is the destination
-	err = ps.deliverToPSS(ctx, chunk)
-	if err != nil {
-		ps.logger.Debugf("error pss delivery for chunk %v: %v", chunk.Address(), err)
+	if ps.deliveryCallback != nil {
+		// since every chunk received with pushsync will be
+		// tried to be unwrapped, we swallow the error here
+		_ = ps.deliveryCallback(ctx, chunk)
```
janos, line 306: Even this log line can be printed for a contend addressed chunk identified as a trojan chunk, I would keep the error log and remove if it appears too frequent later.

acud: but then it would constantly appear for all content addressed chunks that are unwrapped as garbage and will say that the topic is not found, which is annoying. so maybe we can have the logline, but remove the pss error which is returned when no handlers are found? i find this easier and better. if there's no handlers - too bad

janos: Ok, let's ignore ErrNoHandler but log other errors.

janos: It appears to create an import circle `github.com/ethersphere/bee/pkg/pss: import cycle not allowed`. :(

acud: hm. lovely :)

acud: hm. great. i don't know how to get around this. not returning this error in the first place seems like the cleanest thing to do.

pkg/node/node.go
```diff
@@ -263,7 +263,15 @@ func NewBee(addr string, swarmAddress swarm.Address, keystore keystore.Service,
 	}
 	retrieve.SetStorer(ns)
 
-	pushSyncProtocol := pushsync.New(p2ps, storer, kad, tagg, psss.TryUnwrap, logger, acc, accounting.NewFixedPricer(swarmAddress, 10))
+	silenceNoHandlerFunc := func(ctx context.Context, ch swarm.Chunk) error {
+		err := psss.TryUnwrap(ctx, ch)
+		if err != nil && err == pss.ErrNoHandler {
```
janos, line 268: `err == pss.ErrNoHandler` implies that `err != nil`, which can be removed.

janos (approved):
LGTM.

PR #720, opened 2020-09-21
add cheque signing

### LINKED ISSUE #657: Integrate signer for Ethereum transactions and messages
# Task

`crypto.Signer` only exposes `Sign`. To be able to use external signers as they are intended we will need to differentiate between signing Ethereum messages and Ethereum transactions. Also this signer will have to return signatures in the format expected by the contract (as opposed to the compact signatures).

- [x] signed message
- [x] signed transaction
- [x] signed type data

ralph-pichler:
adds cheque signing functionality

a few remarks:
* unfortunately the clef wrapper from go-ethereum does not expose `account_signTypedData`. therefore a lower-level `rpc.Client` is also passed into the signer. in the future we could either use the rpc directly for everything (as it is just a very thin wrapper), or try to get a `SignedTypedData` function in the go-ethereum abstraction.
* the package which defines the `TypedData` structures (with the correct json encoding) is unfortunately called `core` (from `signer/core`) on the go-ethereum side. To avoid having to have `core` everywhere or resort to import aliasing I created type aliases for the relevant stuff in a `eip712` package which also has some additional helpers.
* `IssueCheque` might change in the next PR depending on the needs of the swap protocol code (which is why it is not properly tested here).

closes #657

### DIFF
--- a/pkg/crypto/clef/clef.go
+++ b/pkg/crypto/clef/clef.go
@@ -25,14 +25,14 @@ var (
 	clefRecoveryMessage = []byte("public key recovery message")
 )
 
-// ExternalSignerInterface is the interface for the clef client from go-ethereum
+// ExternalSignerInterface is the interface for the clef client from go-ethereum.
 type ExternalSignerInterface interface {
 	SignData(account accounts.Account, mimeType string, data []byte) ([]byte, error)
 	SignTx(account accounts.Account, tx *types.Transaction, chainID *big.Int) (*types.Transaction, error)
 	Accounts() []accounts.Account
 }
 
-// Client is the interface for rpc.RpcClient
+// Client is the interface for rpc.RpcClient.
 type Client interface {
 	Call(result interface{}, method string, args ...interface{}) error
 }
@@ -44,7 +44,7 @@ type clefSigner struct {
 	pubKey  *ecdsa.PublicKey // the public key for the account
 }
 
-// DefaultIpcPath returns the os-dependent default ipc path for clef
+// DefaultIpcPath returns the os-dependent default ipc path for clef.
 func DefaultIpcPath() (string, error) {
 	socket := "clef.ipc"
 	// on windows clef uses top level pipes
@@ -66,8 +66,8 @@ func DefaultIpcPath() (string, error) {
 	return filepath.Join(home, ".clef", socket), nil
 }
 
-// NewSigner creates a new connection to the signer at endpoint
-// As clef does not expose public keys it signs a test message to recover the public key
+// NewSigner creates a new connection to the signer at endpoint.
+// As clef does not expose public keys it signs a test message to recover the public key.
 func NewSigner(clef ExternalSignerInterface, client Client, recoverFunc crypto.RecoverFunc) (signer crypto.Signer, err error) {
 	// get the list of available ethereum accounts
 	clefAccounts := clef.Accounts()
@@ -98,28 +98,28 @@ func NewSigner(clef ExternalSignerInterface, client Client, recoverFunc crypto.R
 	}, nil
 }
 
-// PublicKey returns the public key recovered during creation
+// PublicKey returns the public key recovered during creation.
 func (c *clefSigner) PublicKey() (*ecdsa.PublicKey, error) {
 	return c.pubKey, nil
 }
 
-// SignData signs with the text/plain type which is the standard Ethereum prefix method
+// SignData signs with the text/plain type which is the standard Ethereum prefix method.
 func (c *clefSigner) Sign(data []byte) ([]byte, error) {
 	return c.clef.SignData(c.account, accounts.MimetypeTextPlain, data)
 }
 
-// SignTx signs an ethereum transaction
+// SignTx signs an ethereum transaction.
 func (c *clefSigner) SignTx(transaction *types.Transaction) (*types.Transaction, error) {
 	// chainId is nil here because it is set on the clef side
 	return c.clef.SignTx(c.account, transaction, nil)
 }
 
-// EthereumAddress returns the ethereum address this signer uses
+// EthereumAddress returns the ethereum address this signer uses.
 func (c *clefSigner) EthereumAddress() (common.Address, error) {
 	return c.account.Address, nil
 }
 
-// SignTypedData signs data according to eip712
+// SignTypedData signs data according to eip712.
 func (c *clefSigner) SignTypedData(typedData *eip712.TypedData) ([]byte, error) {
 	var sig hexutil.Bytes
 	err := c.client.Call(&sig, "account_signTypedData", c.account.Address, typedData)
--- a/pkg/crypto/clef/clef.go
+++ b/pkg/crypto/clef/clef.go
@@ -14,8 +14,10 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethersphere/bee/pkg/crypto"
+	"github.com/ethersphere/bee/pkg/crypto/eip712"
 )
 
 var (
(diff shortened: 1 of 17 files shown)


pkg/crypto/clef/clef.go
```diff
@@ -30,7 +32,13 @@ type ExternalSignerInterface interface {
 	Accounts() []accounts.Account
 }
 
+// ExternalSignerInterface is the interface for rpc.RpcClient
```
acud, line 35: wrong comment

pkg/crypto/clef/clef.go
```diff
@@ -30,7 +32,13 @@ type ExternalSignerInterface interface {
...
+type RpcClient interface {
+	Call(result interface{}, method string, args ...interface{}) error
+}
+
 type clefSigner struct {
+	client  RpcClient // low-level rpc client to clef as ExternalSigner does not implement account_signTypedData
```
acud, line 41: is the comment needed?

pkg/crypto/clef/clef.go
```diff
@@ -30,7 +32,13 @@ type ExternalSignerInterface interface {
 	Accounts() []accounts.Account
 }
 
+// ExternalSignerInterface is the interface for rpc.RpcClient
+type RpcClient interface {
```
acud, line 36: consider interface name to be `Client`, or `Caller` (more "go")

pkg/crypto/clef/clef.go
```diff
@@ -109,3 +118,14 @@ func (c *clefSigner) SignTx(transaction *types.Transaction) (*types.Transaction,
 func (c *clefSigner) EthereumAddress() (common.Address, error) {
 	return c.account.Address, nil
 }
+
+// SignTypedData signs data according to eip712
+func (c *clefSigner) SignTypedData(typedData *eip712.TypedData) ([]byte, error) {
```
acud, line 123: maybe `SignTyped` is enough?

pkg/crypto/eip712/typeddata.go
```diff
@@ -0,0 +1,46 @@
+package eip712
```
acud, line 1: add copy

pkg/crypto/signer.go
```diff
@@ -124,3 +115,56 @@ func (d *defaultSigner) EthereumAddress() (common.Address, error) {
...
+
+	sighash, err := LegacyKeccak256(rawData)
+	if err != nil {
+		return nil, err
+	}
+
+	return d.sign(sighash, false)
+}
+
+// sign signs the provided hash and converts it to the ethereum (r,s,v) format
```
acud, line 134: `// sign the provided...`

pkg/settlement/swap/chequebook/cheque.go
```diff
@@ -0,0 +1,94 @@
+package chequebook
```
acud, line 1: add copy

pkg/settlement/swap/chequebook/cheque.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Cheque represents a cheque for a SimpleSwap chequebook
+type Cheque struct {
+	Chequebook       common.Address
+	Beneficiary      common.Address
+	CumulativePayout *big.Int
+}
+
+// SignedCheque represents a cheque together with its signature
+type SignedCheque struct {
```
acud, line 21: i wonder if we need both types. a `Cheque` could just have the signature field, and have a helper method `.Signed()` that returns a bool to indicate whether it is signer or not

pkg/settlement/swap/chequebook/cheque_test.go
```diff
@@ -0,0 +1,99 @@
+package chequebook_test
```
acud, line 1: add copy

pkg/settlement/swap/chequebook/chequebook.go
```diff
@@ -135,3 +144,44 @@ func (s *service) WaitForDeposit(ctx context.Context, txHash common.Hash) error
 	}
 	return nil
 }
+
+// Issue a new cheque
```
acud, line 148: `// IssueCheque issues a new cheque` (godoc)

acud (changes requested):
great work @ralph-pichler. i have a few minor comments

pkg/crypto/clef/clef.go
ralph-pichler: fixed

pkg/crypto/clef/clef.go
ralph-pichler: I would leave it because it explains why there are two different clients to the same clef.

pkg/crypto/clef/clef.go
ralph-pichler: went with `Client`

pkg/crypto/eip712/typeddata.go
ralph-pichler: fixed

pkg/crypto/signer.go
ralph-pichler: fixed

pkg/settlement/swap/chequebook/cheque.go
ralph-pichler: fixed

pkg/settlement/swap/chequebook/cheque_test.go
ralph-pichler: fixed

pkg/settlement/swap/chequebook/chequebook.go
ralph-pichler: fixed

pkg/crypto/clef/clef.go
ralph-pichler: Possibly. But this signing method is called `signTypedData` both on an rpc level and in all other libraries I saw.

pkg/settlement/swap/chequebook/chequebook.go
```diff
@@ -135,3 +144,44 @@ func (s *service) WaitForDeposit(ctx context.Context, txHash common.Hash) error
 	}
 	return nil
 }
+
+// IssueCheque issues a new cheque
```
acud, line 148: add period end of sentence (godoc)

pkg/settlement/swap/chequebook/cheque.go
ralph-pichler: Not sure I see the advantage this would bring. Having these types separate makes it explicit at a type level where we expect a signature to be present and where not.

pkg/settlement/swap/chequebook/chequebook.go
```diff
@@ -135,3 +144,44 @@ func (s *service) WaitForDeposit(ctx context.Context, txHash common.Hash) error
...
+func (s *service) IssueCheque(beneficiary common.Address, amount *big.Int) (*SignedCheque, error) {
```
acud, line 149: i wonder if we need the endpoint to be `IssueCheque`, since it is in the chequebook already, so you would usually call some `chequebook.IssueCheque()`, maybe `chequebook.Issue()` is enough in that sense.

pkg/settlement/swap/chequebook/chequebook.go
ralph-pichler: fixed. added it to all the other sentences in chequebook package as well.

pkg/settlement/swap/chequebook/chequebook.go
ralph-pichler: makes sense. changed.

acud (approved):
Thanks @ralph-pichler

pkg/crypto/clef/clef.go
```diff
@@ -109,3 +118,14 @@ func (c *clefSigner) SignTx(transaction *types.Transaction) (*types.Transaction,
 func (c *clefSigner) EthereumAddress() (common.Address, error) {
 	return c.account.Address, nil
 }
+
+// SignTypedData signs data according to eip712
```
acud, line 122: period at end of sentence (goes for the rest of this file)

ralph-pichler: fixed

janos (approved):
LGTM.

Issue #653, opened 2020-09-01
Protocol init


acud:
cc https://github.com/ethersphere/bee/pull/585#issuecomment-682599786

janos:
Yes, @pradovic that was one of the inspiration points for this description. Initially, I thought that only Connect and Disconnect events are enough, but it is better to have more control with their direction.

pradovic:
I believe that current protocols work with single Disconnect call, so that one might be possible. The single Connect call is a bit more tricky as Kademlia has a different use-case and needs to treat them differently.
One more thing, maybe I would rename those to `OnConnectedIn`, `OnDisconnctedIn`,  etc.. Not a requirement, but it might be more obvious that those are event handlers.

acud:
LGTM @janos @pradovic. The only question from my side is that if I understand correctly, this will be called directly from the p2p abstraction, however doesn't this also make the peer notification channel redundant?

Also, one reason why I did not, for example, use such notification when implementing pull syncing (where peer connect/disconnect notifications are needed), is that because IMO topology should reflect the source of truth to consumer APIs (for example, APIs that use `ClosestPeer`, kademlia iterators etc), this potentially introduces data races between p2p abstraction and kademlia, since some disconnections can happen between notifications. Therefore, I chose to implement "dumb" notifications and iterate over the peers, rather than providing an exact overlay (which might not be found when looked for in kademlia due to possible data race).

This is at least my 2 cents from what I've stumbled across. Godspeed

janos:
Pull syncing is a bit different, as it relays on the changes to the topology, not just to call handlers per peer connect. For this purpose topology SubscribePeersChange and NeighborhoodDepth is needed, but here (like for payment threshold protocol) a declarative way to specify events per peer regardless of the topology is needed.

acud:
Cool. Looks good from my side then :+1:

pradovic:
[...]

PR #722, opened 2020-09-21
Granular notifiers in protocol spec

pradovic:
- Added connect/disconnect notifiers per: https://github.com/ethersphere/bee/issues/653
- Removed `topology.Notifier` to remove p2p and toplogy dependency
- Reduced to single notifier as there is only one notifier needed for now, this is easy to expand to array if needed 
- Some code in libp2p  uses closePeer() instead of disconnect, as this is part of connection process and there is no need to signal disconnect in this case as the peer for not yet even added to peer registry.

Todo: more unit tests


janos (commented):
As already stated in the description, unit tests for new notifiers would be good. Otherwise it looks good.

acud (commented):
Looks good @pradovic. I'm not sure why the signature change in the notifiers is needed to `p2p.Peer`, where `p2p.Peer` in fact stores only overlay address. `swarm.Address` usage maintains clarity that only overlay is conveyed, whereas `p2p.Peer` might infer that other information that should maybe not be leaked to higher level components, such as underlay or other p2p info.

pradovic:
> Looks good @pradovic. I'm not sure why the signature change in the notifiers is needed to `p2p.Peer`, where `p2p.Peer` in fact stores only overlay address. `swarm.Address` usage maintains clarity that only overlay is conveyed, whereas `p2p.Peer` might infer that other information that should maybe not be leaked to higher level components, such as underlay or other p2p info.

I saw that @janos proposal uses Peer, so I wanted to make it uniform. No other reason, but I guess it could be useful if we expand Peer with more info. On the other hand, the major change in `Notifier` interfaces was just reducing it to one + decouple it from topology.

pradovic:
@janos I talked with @metacertain and he wanted to work on unit tests to get more familiar with this. So, I will wait for him to implement the tests and hlep him in the process, and we will add them as well.

pradovic:
@janos Tests are done. I used a bit different approach than what was proposed in the task, just because of the simplicity and I think this is good enouth. LMNWYT :)

janos (approved):

janos:
@pradovic yes, this is just enough for testing the functionality. 👍 Thanks.

ralph-pichler (approved):
LGTM

PR #731, opened 2020-09-22
throw error when target string is empty

jmozah:
When target is set in context and is empty, do not trigger recovery, instead throw target decode error and return.
this was reported in [mattermost](https://beehive.ethswarm.org/swarm/pl/ht1phqu9ffn7bdcbmk6w36b75a)

### DIFF
--- a/pkg/netstore/netstore.go
+++ b/pkg/netstore/netstore.go
@@ -42,14 +42,17 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 	if err != nil {
 		if errors.Is(err, storage.ErrNotFound) {
 			// request from network
-			ch, err = s.retrieval.RetrieveChunk(ctx, addr)
-			if err != nil {
+			ch, err1 := s.retrieval.RetrieveChunk(ctx, addr)
+			if err1 != nil {
 				if s.recoveryCallback == nil {
-					return nil, err
+					return nil, err1
 				}
-				targets, err := sctx.GetTargets(ctx)
-				if err != nil {
-					return nil, err
+				targets, err2 := sctx.GetTargets(ctx)
+				if err2 != nil {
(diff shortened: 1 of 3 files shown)


janos (approved):

acud:
The problem is that the targets are always set in context, even if they are empty:
```go
func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
    logger := tracing.NewLoggerWithTraceID(r.Context(), s.Logger)
    targets := r.URL.Query().Get("targets")
    r = r.WithContext(sctx.SetTargets(r.Context(), targets))
    ctx := r.Context()
```
```go
// fileDownloadHandler downloads the file given the entry's reference.
func (s *server) fileDownloadHandler(w http.ResponseWriter, r *http.Request) {
    logger := tracing.NewLoggerWithTraceID(r.Context(), s.Logger)
    nameOrHex := mux.Vars(r)["addr"]
    address, err := s.resolveNameOrAddress(nameOrHex)
    if err != nil {
        logger.Debugf("file download: parse file address %s: %v", nameOrHex, err)
        logger.Errorf("file download: parse file address %s", nameOrHex)
        jsonhttp.BadRequest(w, "invalid file address")
        return
    }
    targets := r.URL.Query().Get("targets")
    r = r.WithContext(sctx.SetTargets(r.Context(), targets))
```

Also, the error that should be returned from netstore is supposed to be `ErrNotFound`, now it will always return a recovery started error, which is wrong. returning a targets empty error does not improve the situation.

jmozah:
> The problem is that the targets are always set in context, even if they are empty:

Even though i think that "setting targets to empty strings" needs to be fixed , to fix the issue. 
If fix this your way by catching empty string and sending ErrNotFound from netstore.

acud:
fixed in a different PR

PR #728, opened 2020-09-22
simple El-Gamal encryption for trojan chunks

zelig:
a prerequisite to  pss encryption #663 
builds on #727 
This PR implements el-Gamal encryption using  symmetric encryption of the `encryption` package and `ECDH` secret key generation in  the `crypto` package

PR #727, opened 2020-09-22
crypto: add ECDH shared secret support, in memory variant

zelig:
a prerequisite to pss encryption #663 submitted seperately for easier review

PR #737, opened 2020-09-23
Pss encryption  and topic matching, refactor, cleaner test

### LINKED ISSUE #663: pss outstanding tasks
# Epic

- [ ] encryption
- [ ] #443 bit vector targets 
- [x] #444 optimised mining, PR #695
- [ ] #445 sealing/unsealing

zelig:
This PR builds on #727 and #728 
It is a refactor of the trojan pkg into pss pkg
-simplifies the structure
- adds zero leak topic matching
- adds el-Gamal encryption with topic as salt
- implements plaintext length encoding and msg integrity checking

closes #663


pkg/pss/pss.go
```diff
@@ -116,18 +112,24 @@ func (p *pss) Register(topic trojan.Topic, handler Handler) (cleanup func()) {
 	}
 }
 
+func (p *pss) topics() (ts []Topic) {
+	p.handlersMu.Lock()
+	defer p.handlersMu.Unlock()
+	for t, _ := range p.handlers {
+		ts = append(ts, t)
```
janos, line 119: This is just something that can be easily done, but probably without so significant performance gain, `ts` slice can be preallocated with `make([]Topic, 0, len(p.handlers))` before the  for loop.

pkg/pss/pss_test.go
```diff
@@ -27,202 +25,218 @@ func TestSend(t *testing.T) {
...
-	for i := 0; i < 10; i++ {
-		mtx.Lock()
 
-		eq := bytes.Equal(tt[:], msg.Topic[:])
-		mtx.Unlock()
-		if eq {
-			return
-		}
-		<-time.After(50 * time.Millisecond)
+	message := <-msgChan
```
janos, line 116: Just not to wait for test to reach the global timeout, a select case with some reasonable large timeout would be nice.

janos (commented):
Package trojan is also imported in sctx package, resulting a failed build.

acud:
@zelig needs a rebase

acud:
build still red

zbiljic:
@zelig One outstanding (possible) issue for this PR is how should the `recipient` for the `"/pss/send/{topic}/{recipient}/{targets}"` API endpoint be represented.

My understanding is that that should be some public key, and going in that direction for the test I am serializing some generated public key, and then hex-encoding bytes to string. On the other side reversed is done, and that public key is used for sending data on the topic to the targets.

In total, there are some number of changes, and I would like if someone who tested this feature before to check it again.

pkg/pss/pss.go
```diff
@@ -116,18 +116,27 @@ func (p *pss) Register(topic trojan.Topic, handler Handler) (cleanup func()) {
 	}
 }
 
+func (p *pss) topics() []Topic {
+	p.handlersMu.Lock()
+	defer p.handlersMu.Unlock()
+
+	ts := make([]Topic, 0, len(p.handlers))
+	for t := range p.handlers {
```
zelig, line 124: this loop shouldnt use append then

zelig:
thatnks nemanja,  awesome job. 
yes  indeed , recipient is expected in all forms that btcec can  parse,  see https://github.com/btcsuite/btcd/blob/master/btcec/pubkey.go#L79

zelig:
> Another note is that for some reason the `pkg/pss/NewTopic` function is using `sha3.Sum256` function to create bytes for `pss.Topic` from topic textual name. I don't know it this was intentional, or just an omission, and whether this should be changed to usual `sha3.NewLegacyKeccak256`.

yes better to  have uniformity

zelig:
> @zelig One outstanding (possible) issue for this PR is how should the `recipient` for the `"/pss/send/{topic}/{recipient}/{targets}"` API endpoint be represented.

it should be last var and if left blank, topic-based encryption is used, which shouuld be generalised out  of recovery pkg

pkg/pss/pss.go
zbiljic: I think comment @janos made was on the point. The pre-allocated array creates slice with length 0 (but enough capacity). Append operations should add into existing slice until capacity is reached, after which each append will reallocate and create new slice. Since we only ever add up to `len(p.handlers)` we should be fine.

zbiljic:
> > @zelig One outstanding (possible) issue for this PR is how should the `recipient` for the `"/pss/send/{topic}/{recipient}/{targets}"` API endpoint be represented.
> 
> it should be last var and if left blank, topic-based encryption is used, which shouuld be generalised out of recovery pkg

One thing about suggested change. We cannot have route match where one variable is empty, that will not match this route.
I will change this one, and create another route without `recipient` part.

pkg/api/pss.go
```diff
@@ -52,6 +49,23 @@ func (s *server) pssPostHandler(w http.ResponseWriter, r *http.Request) {
...
+		// use topic-based encryption
+		privkey := crypto.Secp256k1PrivateKeyFromBytes(topic[:])
+		recipient = &privkey.PublicKey
+	} else {
+		var err error
+		recipient, err = pss.ParseRecipient(recipientVar)
+		if err != nil {
+			s.Logger.Debugf("pss recipient: %v", err)
+			s.Logger.Error("pss recipient")
+			jsonhttp.InternalServerError(w, nil)
```
acud, line 64: BadRequest

pkg/api/pss.go
```diff
@@ -28,18 +30,13 @@ var (
...
 
-type PssMessage struct {
-	Topic   string
-	Message string
-}
-
 func (s *server) pssPostHandler(w http.ResponseWriter, r *http.Request) {
-	t := mux.Vars(r)["topic"]
-	topic := trojan.NewTopic(t)
+	topicVar := mux.Vars(r)["topic"]
```
acud, line 34: since most things are `var`s, maybe `topic :=` and `targets := ` would suffice

pkg/api/pss_test.go
```diff
@@ -15,29 +17,31 @@ import (
...
 	"github.com/ethersphere/bee/pkg/swarm"
-	"github.com/ethersphere/bee/pkg/trojan"
 	"github.com/gorilla/websocket"
 )
 
 var (
-	target  = trojan.Target([]byte{1})
-	targets = trojan.Targets([]trojan.Target{target})
+	target  = pss.Target([]byte{1})
+	targets = pss.Targets([]pss.Target{target})
```
acud, line 34: if you are already refactoring these things between packages, i would suggest to completely remove `Targets` type. It does not serve any purpose, we don't need type alias to represent a slice. it also makes the code clunky in many places

pkg/api/pss_test.go
```diff
@@ -115,9 +113,10 @@ func TestPssWebsocketSingleHandlerDeregister(t *testing.T) {
 
 func TestPssWebsocketMultiHandler(t *testing.T) {
 	var (
-		pss, cl, listener = newPssTest(t, opts{})
-		u                 = url.URL{Scheme: "ws", Host: listener, Path: "/pss/subscribe/testtopic"}
-		cl2, _, err       = websocket.DefaultDialer.Dial(u.String(), nil)
+		p, publicKey, cl, listener = newPssTest(t, opts{})
+
```
acud, line 117: newline addition not needed

pkg/api/router.go
```diff
@@ -90,6 +90,12 @@ func (s *server) setupRouting() {
 		),
 	})
 
+	handle(router, "/pss/send/{topic}/{targets}/{recipient}", jsonhttp.MethodHandler{
```
acud: why not reuse the same handler and just define the recipient mux variable as optional?

zbiljic:
**UPDATE**: In the last commit I have made change to the `/pss/send/` API endpoint, to be as is noted in the book, i.e. `/pss/send/{topic}(?targets={targets}&recipient={recipient})`.

This looks much better to me than previous, where the `target` and `recipient` variable is in the path. If someone has objection to this, we can remove this commit.

pkg/encryption/elgamal/encryption.go
```diff
@@ -23,10 +23,10 @@ func New(key *ecdsa.PrivateKey, pub *ecdsa.PublicKey, salt []byte, padding int,
 	return encryption.New(sk, padding, 0, hashfunc), nil
 }
 
-// NewEncrypter constructs an El-Gamal encryptor
+// NewEncryptor constructs an El-Gamal encryptor
 // this involves generating an ephemeral key pair the public part of which is returned
 // as it is needed for the counterparty to decrypt
-func NewEncrypter(pub *ecdsa.PublicKey, salt []byte, padding int, hashfunc func() hash.Hash) (encryption.Encrypter, *ecdsa.PublicKey, error) {
+func NewEncryptor(pub *ecdsa.PublicKey, salt []byte, padding int, hashfunc func() hash.Hash) (encryption.Encrypter, *ecdsa.PublicKey, error) {
```
acud, line 29: why renaming to `Encryptor`? can we please have `Encrypter`?

pkg/pss/pss.go
acud: i agree with viktor about this. as a convention using `make` then append is possible source of trouble and we've had some pretty serious bugs due to this. it would also prefer this not to use append

pkg/pss/pss_test.go
```diff
@@ -27,202 +25,225 @@ func TestSend(t *testing.T) {
...
+
 	// create and register handler
-	var tt trojan.Topic // test variable to check handler func was correctly called
-	hndlr := func(ctx context.Context, m *trojan.Message) {
-		mtx.Lock()
-		copy(tt[:], m.Topic[:]) // copy the message topic to the test variable
-		mtx.Unlock()
+	handler := func(ctx context.Context, m []byte) {
+		msgChan <- topicMessage{
+			topic: topic,
```
acud, line 104: not sure why we need `topicMessage` type. you register a handler on a topic then you always know which topic you are subscribed to and which data come on that topic

pkg/pss/pss_test.go
```diff
@@ -27,202 +25,225 @@ func TestSend(t *testing.T) {
...
-		<-time.After(100 * time.Millisecond)
 	}
-	t.Fatal("timed out waiting for value")
+}
+
+func ensureCalls(t *testing.T, name string, calls *int, exp int) {
+	t.Helper()
+
+	if exp != *calls {
+		t.Fatalf("%s expected %d calls, found %d", name, exp, *calls)
```
acud, line 247: adding the name of the handler that failed in not necessary, since `ensureCalls` calls `t.Helper()`, when failing it will show the line in which the failure occured on the calling context, so you would know exactly with which handler it was called and as of such you know which handler failed. so the string is not necessary. you will see that this pattern does not exist elsewhere in the codebase for this reason

acud (changes requested):
requesting some changes

acud:
> **UPDATE**: In the last commit I have made change to the `/pss/send/` API endpoint, to be as is noted in the book, i.e. `/pss/send/{topic}(?targets={targets}&recipient={recipient})`.
> 
> This looks much better to me than previous, where the `target` and `recipient` variable is in the path. If someone has objection to this, we can remove this commit.

targets should be in path since you cannot send a message without targets. query params are by convention optional parameters so targets should be mandatory. recipient can be a query param or an optional mux variable. that's at least my take on this

pkg/api/pss.go
zbiljic: That is pre-existing code, I will change it.

pkg/api/pss_test.go
zbiljic: @zelig do you have suggestion for this, since you modified packages ?

pkg/encryption/elgamal/encryption.go
zbiljic: @zelig had some comment about this.

pkg/api/pss_test.go
zbiljic: Yes, but this way fewer lines are changed.

acud: Not sure if i understand what you mean. There's three lines of code on the left and there's three lines of code on the right

pkg/encryption/elgamal/encryption.go
acud: maybe a better name here would be `NewElGamal`. I guess the rename came due to some ambiguity with the interface name but it would be still more informative as `ElGamal` and not a general `Encrypter/or`

pkg/api/pss_test.go
zbiljic: This is not only place in this file with new line added.

pkg/pss/pss.go
janos: Sorry for the confusion, I thought that it is an obvious thing. Feel free to remove it.

zbiljic: I personally do not have special preference to specific way (although I prefer more optimized code).

The options for this are now following (omiting mutex):

- named return value

```
func (p *pss) topics() (ts []Topic) {
	for t, _ := range p.handlers {
		ts = append(ts, t)
	}
	return ts
}
```

- exactly allocated slice with append

```
func (p *pss) topics() []Topic {
	ts := make([]Topic, 0, len(p.handlers))
	for t := range p.handlers {
		ts = append(ts, t)
	}
	return ts
}
```

- exactly allocated slice with helper variable for index

```
func (p *pss) topics() []Topic {
	ts := make([]Topic, len(p.handlers))
    i := 0
	for t := range p.handlers {
		ts[i] = t
        i++
	}
	return ts
}
```

If someone has some other suggestion please add so it can be decided.

There is no special reason to select the code with additional variable, but if that looks easier, we can change to that one.

acud: no need to explicitly declare the helper variable as it is provided by `range`:
```
func (p *pss) topics() []Topic {
	ts := make([]Topic, len(p.handlers))
	for i, t := range p.handlers {
		ts[i] = t
	}
	return ts
}
```

janos: It is needed in this case, as the range is over the map, it does not have the index.

pradovic: @acud - "as a convention using make then append is possible source of trouble and we've had some pretty serious bugs due to this" - could you elaborate more pls? I am interested in this convention.

acud: > It is needed in this case, as the range is over the map, it does not have the index.

true sorry.

acud: > could you elaborate more pls

as an example we had a problem in old swarm codebase where we had errors when setting chunks as synced in localstore.
this was caused by the fact that we used:
```go
someSlice := make([]chunkType, expectedSize)
<then later...>
someSlice = append(someSlice, chunk)
```
So obviously the first entries were empty, and the context down the line did not try to fulfill all elements in the slice, but returned an error on the first item that errored (the first one, since it was nil), as you would normally do.

It took a lot of time to find this problem and it is a very fine difference to spot out when you're skimming down a codebase, i.e. `make([]type,len,cap); then append` vs `make([]type,len); then append`, at least my opinion.

pkg/api/pss_test.go
zelig: yes, but ultimately we want to support bit-level variable length targets but i dont mind\

pkg/encryption/elgamal/encryption.go
zelig: We merged tjhe rename in #728 why revert it?  objections

pkg/pss/pss.go
pradovic: I see, thanks for the explanation :) I still think this is more the case of "once bitten twice shy" then convention ;)

acud: yeah true. i meant it in the way that if we get in the habit of using this as convention we might be possibly burying some bugs under the hood. so i mean that i find the convention of `make then append` risky to use in a sense. anyway, won't block on this. total nitpick :)

acud (approved):

janos (approved):

janos:
There is a trivial to solve merge conflict, only.

zbiljic:
@janos already prepared rebase

pkg/pss/trojan.go
```diff
@@ -0,0 +1,324 @@
...
+		if !contains(targets, hash[:targetsLen]) {
+			return nil, nil
+		}
+		chunk := swarm.NewChunk(swarm.NewAddress(hash), append(hint, append(nonce, payload...)...))
+		return chunk, nil
+	}
+	return mine(ctx, odd, f)
+}
+
+// Unwrap takes a chunk, a topic and a private key, and tries to decrypt the payload
```
zelig, line 131: slightly incorrect  comment: no topic , but topics  are given

PR #740, opened 2020-09-24
cashout cheques

ralph-pichler:
adds endpoints:
* `POST /chequebook/cashout/<peer>` to initiate a cashout for the latest cheque for this peer.
* `GET /chequebook/cashout/<peer>` to view the status or result for the latest cashout for this peer.

Info about the latest cashout per peer is stored in statestore. Builds on top of `lastcheque-endpoint` as it uses some of the interfaces / mocks from there.

This is a very minimal implementation. It does not actively monitor the transaction for success, only when queried by the user.

### DIFF
--- a/pkg/debugapi/chequebook.go
+++ b/pkg/debugapi/chequebook.go
@@ -208,15 +208,19 @@ func (s *server) swapCashoutHandler(w http.ResponseWriter, r *http.Request) {
 	jsonhttp.OK(w, swapCashoutResponse{TransactionHash: txHash.String()})
 }
 
+type swapCashoutStatusResult struct {
+	Recipient  common.Address `json:"recipient"`
+	LastPayout *big.Int       `json:"lastPayout"`
+	Bounced    bool           `json:"bounced"`
+}
+
 type swapCashoutStatusResponse struct {
-	Peer             swarm.Address  `json:"peer"`
-	Chequebook       common.Address `json:"chequebook"`
-	CumulativePayout *big.Int       `json:"cumulativePayout"`
-	Beneficiary      common.Address `json:"beneficiary"`
-	TransactionHash  common.Hash    `json:"transactionHash"`
-	Recipient        common.Address `json:"recipient"`
-	LastPayout       *big.Int       `json:"lastPayout"`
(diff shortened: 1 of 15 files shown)


pkg/settlement/swap/chequebook/cashout.go
```diff
@@ -0,0 +1,234 @@
+package chequebook
```
janos, line 1: Add copyright header.

pkg/settlement/swap/chequebook/cashout_test.go
```diff
@@ -0,0 +1,443 @@
+package chequebook_test
```
janos, line 1: Add copyright header.

pkg/debugapi/chequebook.go
```diff
@@ -184,3 +190,88 @@ func (s *server) chequebookAllLastHandler(w http.ResponseWriter, r *http.Request
...
+	status, err := s.Swap.CashoutStatus(r.Context(), peer)
+	if err != nil {
+		if errors.Is(err, chequebook.ErrNoCheque) {
+			jsonhttp.NotFound(w, errNoCheque)
+			return
+		}
+		if errors.Is(err, chequebook.ErrNoCashout) {
+			jsonhttp.NotFound(w, errNoCashout)
+			return
+		}
```
janos, line 253: Add debug log lines.

janos (commented):
I gave only a formal review as I would need more time to properly review it. The code looks pretty solid, but I did not validate that it is correct from the usability perspective. It would be good to have openapi specs also updated.

pkg/settlement/swap/chequebook/cashout.go
ralph-pichler: added

pkg/settlement/swap/chequebook/cashout_test.go
ralph-pichler: added

pkg/debugapi/chequebook.go
ralph-pichler: added

janos (approved):

zelig (approved):
brilliant

PR #742, opened 2020-09-24
allow first retrieval request to be upstream

janos:
This PR changes slightly the logic of retrieval. It allows the upstream chunk request by the requester node. All forwarded requests are still only downstream. This change improves the situation where the requester node is the closest to he chunk of all its peers. By allowing the first request to be upstream, retrieval is more efficient and forwarding loops are still not possible.

### DIFF
--- a/pkg/retrieval/retrieval.go
+++ b/pkg/retrieval/retrieval.go
@@ -115,18 +115,25 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address) (swarm.
 
 func (s *Service) retrieveChunk(ctx context.Context, addr swarm.Address, skipPeers []swarm.Address) (chunk swarm.Chunk, peer swarm.Address, err error) {
 	v := ctx.Value(requestSourceContextKey{})
+	// allow upstream requests if this node is the source of the request
+	// i.e. the request was not forwarded, to improve retrieval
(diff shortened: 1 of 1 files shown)


Eknir:
This PR fixes an error which I got when trying out `prefetch-buffer` PR from @acud.

Error below:

image

pkg/retrieval/retrieval.go
```diff
@@ -115,18 +115,25 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address) (swarm.
...
+		allowUpstream = false
 	}
 	ctx, cancel := context.WithTimeout(ctx, retrieveChunkTimeout)
 	defer cancel()
 
-	peer, err = s.closestPeer(addr, skipPeers)
+	peer, err = s.closestPeer(addr, skipPeers, allowUpstream)
 	if err != nil {
-		return nil, peer, fmt.Errorf("get closest: %w", err)
+		return nil, peer, fmt.Errorf("get closest for address %s: %w", addr.String(), err)
```
acud, line 136: maybe we can add the value of `allowUpstream` in the error?

acud (approved):
Beautiful. Thanks @janos

zbiljic (approved):

PR #745, opened 2020-09-24
trigger settlement before expected debt hits payment threshold

ralph-pichler:
This changes accounting to trigger payment a bit earlier than the peers payment threshold. This is to avoid needlessly blocking reserve request if already near the payment threshold when several concurrent request occur.

### DIFF
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -57,6 +57,7 @@ type accountingPeer struct {
 type Options struct {
 	PaymentThreshold uint64
 	PaymentTolerance uint64
+	EarlyPayment     uint64
 	Logger           logging.Logger
 	Store            storage.StateStorer
 	Settlement       settlement.Interface
@@ -74,6 +75,7 @@ type Accounting struct {
 	// The amount in BZZ we let peers exceed the payment threshold before we
 	// disconnect them.
 	paymentTolerance uint64
+	earlyPayment     uint64
 	settlement       settlement.Interface
 	metrics          metrics
 }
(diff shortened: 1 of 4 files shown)


pkg/accounting/accounting.go
```diff
@@ -229,7 +232,7 @@ func (a *Accounting) Credit(peer swarm.Address, price uint64) error {
 
 	// If our expected debt exceeds our payment threshold (which we assume is
 	// also the peers payment threshold), trigger settlement.
-	if uint64(expectedDebt) >= a.paymentThreshold {
+	if uint64(expectedDebt) >= a.paymentThreshold-a.earlyPayment {
```
Eknir, line 235: Update comment above. Also explain why we do it

Eknir (approved):
Yes, this makes sense.

pkg/accounting/accounting.go
```diff
@@ -227,9 +230,17 @@ func (a *Accounting) Credit(peer swarm.Address, price uint64) error {
...
-	// If our expected debt exceeds our payment threshold (which we assume is
+	// If our expected debt is less than earlyPayment away from our payment threshold (which we assume is
 	// also the peers payment threshold), trigger settlement.
-	if uint64(expectedDebt) >= a.paymentThreshold {
+	// we pay early to avoid needlessly blocking request later when concurrent requests occur and we are already close to the payment threshold
+	threshold := a.paymentThreshold
+	if threshold > a.earlyPayment {
+		threshold -= a.earlyPayment
+	} else {
+		threshold = 0
```
ralph-pichler, line 240: we should probably enforce some kind of minimum payment threshold we accept, so we don't have peers where we constantly send cheques for every chunk.

janos (approved):

PR #746, opened 2020-09-24
pusher: add tracing

acud:
adds a main trace to the pusher under which all the chunks in the snapshot will be visible on jaeger.

### DIFF
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -170,6 +170,10 @@ LOOP:
 			if unsubscribe != nil {
 				unsubscribe()
 			}
+			if span != nil {
+				span.Finish()
+			}
+
 			break LOOP
 		}
 	}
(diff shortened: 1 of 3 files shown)


janos (approved):
Nice @acud. LGTM.

pkg/pusher/pusher.go
```diff
@@ -79,6 +85,10 @@ LOOP:
 				break
 			}
 
+			if span == nil {
+				span, _, ctx = s.tracer.StartSpanFromContext(cctx, "pusher-sync-batch", s.logger)
```
pradovic, line 89: where will this new ctx be canceled?

pradovic (approved):
LGTM mod one question 👍

pkg/pusher/pusher.go
acud: `ctx` will be on the first run the actual root context (`cctx`) which is being cancelled on `L64`, and once we reach this line then `ctx` becomes a child context of `cctx` and does not have its own cancel function, therefore it is not cancelled. if `cctx` is cancelled, then this context will be cancelled too. at least that is to the best of my understanding

pradovic: SGTM

PR #747, opened 2020-09-25
api, netstore: dont set empty targets, return correct errors from netstore

acud:
We are constantly rewriting errors with chunk repair and trojan errors and this must be changes so that netstore returns the correct errors when failing to retrieve a chunk. Also, since targets are not a mandatory argument when doing requests, it should not be considered an error when one does not have any targets set.

### DIFF
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -28,7 +28,9 @@ import (
 func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
 	logger := tracing.NewLoggerWithTraceID(r.Context(), s.Logger)
 	targets := r.URL.Query().Get("targets")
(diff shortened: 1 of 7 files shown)


janos:
It looks like that TestInvalidTargets test should be adjusted.

```
=== RUN   TestInvalidTargets
Error:     netstore_test.go:152: netstore retrieve chunk: chunk not found
```

acud:
> It looks like that TestInvalidTargets test should be adjusted.
> 
> ```
> === RUN   TestInvalidTargets
> Error:     netstore_test.go:152: netstore retrieve chunk: chunk not found
> ```

i removed the test. i don't see the point in checking the validity of the targets while getting it from context. if anything, we should be checking the target format on input, not on processing it later.

janos (approved):

pradovic (approved):

PR #749, opened 2020-09-25
fix goerli factory address

ralph-pichler:
changes goerli factory address to the correct one. the other address is a valid factory but it does not use the token we want to use. (it was the test factory used before deploying the final gbbz token)

also checks we can afford initial deposit before deploying.


Eknir (approved):
Thanks for this PR!

janos (approved):

Issue #753, opened 2020-09-25
panic in pss.extractPublicKey


ralph-pichler:
```
panic: runtime error: slice bounds out of range [:72] with capacity 32

goroutine 1195 [running]:
github.com/ethersphere/bee/pkg/pss.extractPublicKey(0xc003926060, 0x16, 0x20, 0x20, 0x1, 0xb848d6c7d44639f3)
    github.com/ethersphere/bee/pkg/pss/trojan.go:262 +0x13b
github.com/ethersphere/bee/pkg/pss.Unwrap(0x17e6100, 0xc00390a750, 0xc000b54090, 0x17efca0, 0xc0039187c0, 0xc003926920, 0x1, 0x1, 0x0, 0x0, ...)
    github.com/ethersphere/bee/pkg/pss/trojan.go:135 +0x9a
github.com/ethersphere/bee/pkg/pss.(*pss).TryUnwrap(0xc000a5c050, 0x17e6100, 0xc00390a750, 0x17efca0, 0xc0039187c0, 0xc003926040, 0x20)
    github.com/ethersphere/bee/pkg/pss/pss.go:133 +0xf8
github.com/ethersphere/bee/pkg/node.NewBee.func1(0x17e6100, 0xc00390a750, 0x17efca0, 0xc0039187c0, 0xa0, 0x0)
    github.com/ethersphere/bee/pkg/node/node.go:363 +0x5b
github.com/ethersphere/bee/pkg/pushsync.(*PushSync).handleDeliveryResponse(0xc000a8a2c0, 0x17e6100, 0xc00390a750, 0x7f7dcb6fcb78, 0xc003918400, 0xc000afe8e0, 0x20, 0x20, 0x17efca0, 0xc0039187c0, ...)
    github.com/ethersphere/bee/pkg/pushsync/pushsync.go:313 +0x41b
github.com/ethersphere/bee/pkg/pushsync.(*PushSync).handler(0xc000a8a2c0, 0x17e6100, 0xc00390a750, 0xc000afe8e0, 0x20, 0x20, 0x17ed120, 0xc0038a1100, 0x17c3220, 0xc000048770)
    github.com/ethersphere/bee/pkg/pushsync/pushsync.go:112 +0x140b
github.com/ethersphere/bee/pkg/p2p/libp2p.(*Service).AddProtocol.func1(0x17f6e00, 0xc0038a10e0)
[...]

PR #754, opened 2020-09-26
Pss trojan  length check and no  error return in  delivery callback

zelig:
add length check  to `TryUnwrap` 
fixes https://github.com/ethersphere/bee/issues/753
remove error return value `TryUnwrap/deliveryCallback` to make sure no protocol  error  can result from pss dispatching

### DIFF
--- a/pkg/pss/pss.go
+++ b/pkg/pss/pss.go
@@ -129,12 +129,12 @@ func (p *pss) topics() []Topic {
 
 // TryUnwrap allows unwrapping a chunk as a trojan message and calling its handlers based on the topic.
 func (p *pss) TryUnwrap(ctx context.Context, c swarm.Chunk) {
-	if len(c.Data()) < 4104 {
+	if len(c.Data()) < swarm.ChunkWithSpanSize {
 		return // chunk not full
 	}
 	topic, msg, err := Unwrap(ctx, p.key, c, p.topics())
 	if err != nil {
-		return // cannor unwrap
+		return // cannot unwrap
 	}
 	h := p.getHandlers(topic)
 	if h == nil {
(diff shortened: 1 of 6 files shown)


pkg/pss/pss.go
```diff
@@ -129,14 +128,17 @@ func (p *pss) topics() []Topic {
 }
 
 // TryUnwrap allows unwrapping a chunk as a trojan message and calling its handlers based on the topic.
-func (p *pss) TryUnwrap(ctx context.Context, c swarm.Chunk) error {
+func (p *pss) TryUnwrap(ctx context.Context, c swarm.Chunk) {
+	if len(c.Data()) < 4104 {
```
zbiljic, line 132: Could we create constant for `4104` ?

zbiljic (approved):

pkg/pss/pss.go
```diff
@@ -129,14 +128,17 @@ func (p *pss) topics() []Topic {
...
+		return // chunk not full
+	}
 	topic, msg, err := Unwrap(ctx, p.key, c, p.topics())
 	if err != nil {
-		return err
+		return // cannor unwrap
```
janos, line 137: cannor -> cannot

janos (approved):

pkg/pss/pss.go
acud: it would be better to extract this to a `var` which is calculated once, since it is not trivial to assume the reader knows why this specific number was selected (such as in the previous check: `trojanChunkMinDataLen = swarm.SpanSize + NonceSize + TopicSize + LengthSize`)

acud (approved):

PR #778, opened 2020-09-28
default initial deposit

ralph-pichler:
changes the default initial deposit to the amount dispensed by the faucet. this ensure users don't end up with an unfunded chequebook unless they explicitly want to.


Eknir (approved):
This is important from a UX perspective.

janos (approved):

PR #784, opened 2020-09-29
first commit for slash command beekeeper run

vandot:


vandot:
we will take different approach

PR #785, opened 2020-09-30
Validator callbacks

zelig:
This PR is primarily motivated by allowing mailboxing on pss messages.
It simplifies and refactors Validators, and introduces the  possibility to define asyncronous chunk-specific hooks  on the multiple validator in swarm, see pss.Handle in node.go

The PR is incorrect in its current form since it wrongly removes validation from pushsyncer, which will now forward invalid chunks. 
Also passing the context to the asyncronous  chunk hooks and  in general termination of the hook goroutines need rethinking


pkg/swarm/validator.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package swarm contains most basic and general Swarm concepts.
```
janos, line 5: Invalid comment.

pkg/swarm/validator.go
```diff
@@ -0,0 +1,29 @@
...
+import "context"
+
+type ValidatorFunc func(context.Context, Chunk) bool
+
+type Validator struct {
+	Valid    func(ch Chunk) (valid bool)
+	Callback func(context.Context, Chunk)
+}
+
+func NewValidator(vs ...Validator) ValidatorFunc {
```
janos, line 17: To be precise and clear, this function should be called `NewValidatorFunc` as it conflicts with the `Validator` interface here. I would even go and suggest these name changes:

`ValidatorFunc` -> `Func` (validator is the package name)
`NewValidator` -> `New`

or even 

`ValidatorFunc` -> interface type `Service`
`Validator` -> `NewService`

but that may be too much.

janos (commented):
I like the approach but naming could be approved a bit, just to improve the readability of the code.

zelig:
but do  you like the callback assignment ?
pushsync also needs validator for forwarding
but for stufff saved it should go to netstore validator which contains the pss.Handle callback
this smells, since validators are all over the place and netstore only called when need to save.

janos:
Yes, I think that the callback provides good enough flexibility that can be specified for different requirements for different components, there is no need to have a universal validator instance, like here, netstore and pushsync have different requirements, and that is ok for me.

Eknir:
TODO:
- @eknir, make user story about this (PSS mailboxing)
- @janos and @zelig , to discuss the best approach

Eknir:
closed this one, in favor of #913

PR #786, opened 2020-09-30
actually disconnect on no swap beneficiary

ralph-pichler:
Unfortunately the last PR did not actually work the way intended. It didn't disconnect at all as the error was ignored further upstream and even if it was handled it would have disconnected the wrong peer.

This PR passes `p2p.Service` into `Swap` which can then disconnect the right peer.


janos (approved):

janos (commented):
Actually, this test should be adjusted, mostly on the mock side.

```
=== RUN   TestPayUnknownBeneficiary
Error:     swap_test.go:421: wrong error. wanted unknown beneficiary for peer, got function Disconnect not configured
```

zelig (approved):

PR #791, opened 2020-10-02
accounting: settle in reserve

ralph-pichler:
This PR changes the accounting flow so that settlement of actual debt is already triggered in the `Reserve` stage.

The new accounting flow is as follows:
* `Credit` like `Debit` now only updates balances and no longer attempts to settle.
* Settlement now uses the right context
* If in `Reserve` we detect we are close enough to payment threshold we trigger settlement right there and fail if settlement fails.
* We still block if the expected debt would exceed the threshold. this is possible even if settlement occurred (e.g. if all outstanding requests actually exceed the payment threshold and we had positive balance with the peer before).

Motivation:
* Currently it is possible to get in a state where we are above the payment threshold but we still will never attempt to settle again. The only way to always have a chance retry without intentionally going arbitrarily high over the payment threshold is to do this in reserve stage already.
* This PR should also reduce the frequency of overdraft errors in case of many concurrent requests.

### DIFF
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -28,12 +28,12 @@ var (
 
 // Interface is the Accounting interface.
 type Interface interface {
-	// Reserve reserves a portion of the balance for peer. Returns an error if
-	// the operation risks exceeding the disconnect threshold.
+	// Reserve reserves a portion of the balance for peer and attempts settlements if necessary.
+	// Returns an error if the operation risks exceeding the disconnect threshold.
 	//
-	// This should be called (always in combination with Release) before a
+	// This has to be called (always in combination with Release) before a
 	// Credit action to prevent overspending in case of concurrent requests.
-	Reserve(peer swarm.Address, price uint64) error
+	Reserve(ctx context.Context, peer swarm.Address, price uint64) error
 	// Release releases the reserved funds.
 	Release(peer swarm.Address, price uint64)
 	// Credit increases the balance the peer has with us (we "pay" the peer).
@@ -116,8 +116,8 @@ func NewAccounting(
 	}, nil
 }
 
-// Reserve reserves a portion of the balance for peer.
-func (a *Accounting) Reserve(peer swarm.Address, price uint64) error {
+// Reserve reserves a portion of the balance for peer and attempts settlements if necessary.
+func (a *Accounting) Reserve(ctx context.Context, peer swarm.Address, price uint64) error {
 	accountingPeer, err := a.getAccountingPeer(peer)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


pkg/settlement/pseudosettle/mock/pseudosettle.go
```diff
@@ -62,7 +70,10 @@ func NewSettlement(opts ...Option) settlement.Interface {
 	return mock
 }
 
-func (s *Service) Pay(_ context.Context, peer swarm.Address, amount uint64) error {
+func (s *Service) Pay(c context.Context, peer swarm.Address, amount uint64) error {
+	if s.pay != nil {
+		return s.Pay(c, peer, amount)
```
zelig, line 75: meant to be  `s.pay`  no?

ralph-pichler: yes, tests worked anyway because I used the old mock for the tests which actually paid something. I switched all tests over to this one and removed the old mock.

metacertain (approved):
LGTM

Eknir (commented):
I looked at the PR description, as well as the change set to the implementation of `Reserve` as well as `Credit`.

I think this is an elegant solution to make sure we are not in an never-ending overdraft blockage.

I am approving the logic, but won't approve the PR because there are quite a lot of lines of code that I didn't have a look at.

Thanks for the PR and I hope it gets in soon.

pkg/settlement/pseudosettle/mock/pseudosettle.go
```diff
@@ -23,6 +23,8 @@ type Service struct {
 
 	settlementsSentFunc func() (map[string]uint64, error)
 	settlementsRecvFunc func() (map[string]uint64, error)
+
+	pay func(context.Context, swarm.Address, uint64) error
```
janos, line 27: For consistency, this `pay` field should be called `payFunc`.

janos (approved):

pkg/settlement/pseudosettle/mock/pseudosettle.go
ralph-pichler: fixed

pkg/accounting/accounting.go
```diff
@@ -28,12 +28,12 @@ var (
 
 // Interface is the Accounting interface.
 type Interface interface {
-	// Reserve reserves a portion of the balance for peer. Returns an error if
-	// the operation risks exceeding the disconnect threshold.
+	// Reserve reserves a portion of the balance for peer and attempts settlements if necessary.
+	// Returns an error if the operation risks exceeding the disconnect threshold.
```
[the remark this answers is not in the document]
ralph-pichler: updated

ldeffenb:
It looks like this is working as intended.  The balance with my peers approaches 100K and then a settlement is issued.  The most observable effect is that checks used to be 100K+ and now they tend to be shy of 100K (like in the 80s) when I'm pushing lots of chunks into the network.

I haven't seen a peer get to 100K and stall as was happening previously when connected to (apparently) non-swapping peers.

PR #805, opened 2020-10-06
decouple retrieval forwarding from netstore

janos:
This PR decouples forwardign logic from netstore into retreival, leaving retrieval not dependent on netstore behaviour.


acud (approved):
very cool. thanks janos

zelig (approved):

PR #809, opened 2020-10-07
Export additional metrics

zbiljic:
While debugging "pushsync" package, I noticed that metrics are not exported. After including those to debug API, I also noticed some other seem to be missing from there. 

This change includes those missing metrics from other packages, and fixes names of metrics from `localstore` (discovered after metrics were added).

One remaining metric group that was not included here is from `shed` package. Since that is only internally used in `localstore`, different change would be needed to include it.
Since I am not totally sure about that package anyway, I have left it out. If someone thinks that it should be added, I can update this PR.

### DIFF
--- a/pkg/localstore/metrics.go
+++ b/pkg/localstore/metrics.go
@@ -216,7 +216,7 @@ func newMetrics() metrics {
 		ModeGetMultiFailure: prometheus.NewCounter(prometheus.CounterOpts{
 			Namespace: m.Namespace,
 			Subsystem: subsystem,
-			Name:      "mode_get_failure_count",
+			Name:      "mode_get_multi_failure_count",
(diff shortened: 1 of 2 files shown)


janos (approved):

ldeffenb:
Still missing PushSyncPusher metrics unless I missed something?

zbiljic:
> Still missing PushSyncPusher metrics unless I missed something?

You are correct.

acud (approved):
nice. shows to which extent we use metrics :D

ldeffenb:
When we get to the backlog task in the roadmap, whoever gets assigned the following tasks should appreciate this:

create Bee specific dashboards in Grafana
create dashboards with Beekeeper metrics in Grafana

At least, I assume Grafana will be consuming the Prometheus metrics.

PR #812, opened 2020-10-07
move transaction service to its own package

ralph-pichler:
* transaction service to its own package
* proper mock for `crypto.Signer` and `transaction.Service`

### DIFF
--- a/pkg/settlement/swap/transaction/mock/transaction.go
+++ b/pkg/settlement/swap/transaction/mock/transaction.go
@@ -1,0 +1,62 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mock
+
+import (
+	"context"
+	"errors"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethersphere/bee/pkg/settlement/swap/transaction"
+)
+
+type transactionServiceMock struct {
+	send           func(ctx context.Context, request *transaction.TxRequest) (txHash common.Hash, err error)
+	waitForReceipt func(ctx context.Context, txHash common.Hash) (receipt *types.Receipt, err error)
+}
+
+func (m *transactionServiceMock) Send(ctx context.Context, request *transaction.TxRequest) (txHash common.Hash, err error) {
+	if m.send != nil {
+		return m.send(ctx, request)
+	}
+	return common.Hash{}, errors.New("error")
+}
+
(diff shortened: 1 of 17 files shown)


janos (approved):

pkg/settlement/swap/transaction/mock/transaction.go
```diff
@@ -0,0 +1,62 @@
...
+type transactionServiceMock struct {
+	send           func(ctx context.Context, request *transaction.TxRequest) (txHash common.Hash, err error)
+	waitForReceipt func(ctx context.Context, txHash common.Hash) (receipt *types.Receipt, err error)
+}
+
+func (m *transactionServiceMock) Send(ctx context.Context, request *transaction.TxRequest) (txHash common.Hash, err error) {
+	if m.send != nil {
+		return m.send(ctx, request)
+	}
+	return common.Hash{}, errors.New("error")
```
zelig, line 25: these `errors.New("error")` errors are a bit strange. throughout, why are we sabotaging. extra information?

pkg/settlement/swap/transaction/transaction.go
```diff
@@ -54,7 +46,7 @@ type transactionService struct {
 }
 
 // NewTransactionService creates a new transaction service.
-func NewTransactionService(logger logging.Logger, backend Backend, signer crypto.Signer) (TransactionService, error) {
+func NewTransactionService(logger logging.Logger, backend Backend, signer crypto.Signer) (Service, error) {
```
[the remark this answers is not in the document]
acud: :+1:  for `New`

pkg/settlement/swap/transaction/mock/transaction.go
acud: this is a mock so its not that bad. the errors that the service below can return are coming from the go-ethereum package if i understand correctly. not sure if the mock should use these errors or not, as the tests don't do assertions on the error type. @ralph-pichler if a mock returns an error it would be nice to be able to set the error from the test. see [topology mock](https://github.com/ethersphere/bee/blob/master/pkg/topology/mock/mock.go#L36) as an example

janos: Actually, as `send` function can be set as any and can return arbitrary error for checking in tests, this error is of not so significance. It can state `not implemented` or `this method is not specified`, just to be more informative.

ralph-pichler: I went with `not implemented`. As @janos said if you want to return a specific error in the mock you can just override that function (this is also how its done in swap tests).

pkg/settlement/swap/transaction/transaction.go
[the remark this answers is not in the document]
ralph-pichler: went with `NewService`, as `transaction.New` might be misinterpreted as creating a transaction, rather than a transaction service.

PR #817, opened 2020-10-07
bump version to 0.3.2

janos:
Set the next version in code.

### DIFF
--- a/version.go
+++ b/version.go
@@ -5,7 +5,7 @@
 package bee
 
 var (
-	version = "0.3.0" // manually set semantic version number
+	version = "0.3.2" // manually set semantic version number
 	commit  string    // automatically set git commit hash
 
 	Version = func() string {


zbiljic:
@janos Should this PR maybe also update OpenAPI documentation files ?

janos:
> @janos Should this PR maybe also update OpenAPI documentation files ?

I am not sure if we need to update that version if api is not changed. Maybe it is good to hear other opinions also.

vandot (approved):

svetomir (approved):

acud:
> > @janos Should this PR maybe also update OpenAPI documentation files ?
> 
> I am not sure if we need to update that version if api is not changed. Maybe it is good to hear other opinions also.

i think we never really said anything about maintaining a separate version for the api. in which case we should just leave it alone for now, unless some changes arise. not sure if i really have an opinion here. would make sense to bump it just when it changes

acud:
also, given that we make the same error of not bumping the version numbers - could we please prepare together a release checklist?

acud (approved):
Yassu

janos:
Release checklist is needed. We should make it.

Let's update api version in specs when we agree on the checklist.

PR #821, opened 2020-10-08
localstore/pin: refactor and add paging

### LINKED ISSUE #464: Delete all pins (20 at a time?)
# User story

## Story

As a file host I want to be able to delete ALL current pins as I load a new consistent set of files into the Swarm.

I'm currently doing "Get list of pinned chunks" followed by "Unpin chunk with given address" for each returned address (times the returned pinCounter).  But imagine my surprise when this doesn't delete all the pins?  Another "Get list of pinned chunks" returns still more (and different) pins.

It seems that "Get list of pinned chunks" only returns 20 pins per invocation, but this is not mentioned in the documentation for the Debug API.  There also does not seem to be a way to specify how many pins are desired.

Is there a way to query ALL the pins?  Or is there a context so that different sets of pins can be queried?

## Acceptance criteria

At least document "Get list of pinned chunks" to have a limit of 20 pins.  But hopefully also describe how to query all existing pins.

Or provide an (Debug?) API that deletes all current pins with full understanding of the ramifications of this request.

## Background

I have a script that adds lots (1,000s) of files to the swarm and automatically pins each address while also recording said address externally.  As this file set is refreshed, I want to delete all current pins without consulting the external source and before re-adding the new file set.

## Tasks

acud:
This PR adds pagination to the pin api, as well as some minor cleanups and refactorings.
A default 100 entry response is returned, and a hard limit of 1000 entries per response is enforced.

fixes #464

### DIFF
--- a/pkg/api/pin.go
+++ b/pkg/api/pin.go
@@ -23,6 +23,7 @@ func (s *server) pinChunk(w http.ResponseWriter, r *http.Request) {
 	addr, err := swarm.ParseHexAddress(mux.Vars(r)["address"])
 	if err != nil {
 		s.Logger.Debugf("pin chunk: parse chunk address: %v", err)
+		s.Logger.Error("pin chunk: parse address")
 		jsonhttp.BadRequest(w, "bad address")
 		return
 	}
@@ -30,6 +31,7 @@ func (s *server) pinChunk(w http.ResponseWriter, r *http.Request) {
 	has, err := s.Storer.Has(r.Context(), addr)
 	if err != nil {
 		s.Logger.Debugf("pin chunk: localstore has: %v", err)
+		s.Logger.Error("pin chunk: store")
 		jsonhttp.InternalServerError(w, err)
 		return
 	}
@@ -42,6 +44,7 @@ func (s *server) pinChunk(w http.ResponseWriter, r *http.Request) {
 	err = s.Storer.Set(r.Context(), storage.ModeSetPin, addr)
 	if err != nil {
 		s.Logger.Debugf("pin chunk: pinning error: %v, addr %s", err, addr)
+		s.Logger.Error("pin chunk: cannot pin chunk")
 		jsonhttp.InternalServerError(w, "cannot pin chunk")
 		return
 	}
@@ -54,6 +57,7 @@ func (s *server) unpinChunk(w http.ResponseWriter, r *http.Request) {
 	addr, err := swarm.ParseHexAddress(mux.Vars(r)["address"])
 	if err != nil {
 		s.Logger.Debugf("pin chunk: parse chunk address: %v", err)
+		s.Logger.Error("pin chunk: parse address")
 		jsonhttp.BadRequest(w, "bad address")
 		return
 	}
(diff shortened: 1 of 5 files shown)


pkg/localstore/pin_test.go
```diff
@@ -12,114 +12,132 @@ import (
...
+	err = db.Set(context.Background(), storage.ModeSetPin, swarm.NewAddress(chunk.Address().Bytes()))
+	if err != nil {
+		t.Fatal(err)
+	}
+	pinCounter, err = db.PinCounter(swarm.NewAddress(chunk.Address().Bytes()))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if pinCounter != 2 {
+		t.Fatalf("want pin counter 1 but got %d", pinCounter)
```
zelig, line 78: 1. -> 2

ldeffenb:
Are there defaults for limit and offset?  My (possibly incorrect) reading looks like they'll default to zero which is fine (and one could argue expected) for offset, but defaulting the limit to 0 results in (I think) just a single pin being returned.

For backward compatibility it would seem that a default limit, if not specified, of the old 20 would be good?

ldeffenb:
I just re-read and saw "A default 100 entry response", but I didn't see that in the code?  I confess, I might have missed it though.

ldeffenb:
Ignore me.  I now see the 100 limit in the "add error" commit.  Didn't expect that comment on that commit.

acud:
@ldeffenb there's also a short description in the PR ^^ :)

pkg/api/pin.go
```diff
@@ -96,16 +104,41 @@ type listPinnedChunksResponse struct {
...
+			s.Logger.Debugf("list pins: parse limit: %v", err)
+			s.Logger.Errorf("list pins: bad limit")
+			jsonhttp.BadRequest(w, "bad limit")
+		}
+	}
+
+	pinnedChunks, err := s.Storer.PinnedChunks(r.Context(), offset, limit)
 	if err != nil {
-		s.Logger.Debugf("debug-api: pin chunk: listing pinned chunks: %v", err)
+		s.Logger.Debugf("pin chunk: list pinned: %v", err)
```
janos, line 131: `pin chunk` -> `list pins`

pkg/api/pin.go
```diff
@@ -131,13 +166,14 @@ func (s *server) getPinnedChunk(w http.ResponseWriter, r *http.Request) {
...
 
-	pinCounter, err := s.Storer.PinInfo(addr)
+	pinCounter, err := s.Storer.PinCounter(addr)
 	if err != nil {
 		if errors.Is(err, storage.ErrNotFound) {
 			jsonhttp.NotFound(w, nil)
 			return
 		}
-		s.Logger.Debugf("debug-api: pin chunk: listing pinned chunks: %v", err)
+		s.Logger.Debugf("pin chunk: get pin counter: %v", err)
```
janos, line 175: `pin chunk` -> `pin counter`

janos (approved):

ldeffenb:
> @ldeffenb there's also a short description in the PR ^^ :)

I read that, but was checking the code just to see if I could see how it worked.

pkg/localstore/pin.go
```diff
@@ -37,36 +32,34 @@ func (db *DB) PinnedChunks(ctx context.Context, cursor swarm.Address) (pinnedChu
...
+		if limit == 0 {
 			return true, nil
-		} else {
-			return false, nil
 		}
-
+		return false, nil
 	}, &shed.IterateOptions{
-		StartFrom:         &it,
 		SkipStartFromItem: false,
```
zbiljic, line 55: No need for `SkipStartFromItem` option if the `StartFrom` is removed.
Options here can be replaced by `nil`.

zbiljic (approved):

PR #823, opened 2020-10-09
all: use bmt hasher pool

acud:
The PR integrates the new bmt hasher pool from the bmt package in order to reduce allocations.

### DIFF
--- a/pkg/bmtpool/bmtpool.go
+++ b/pkg/bmtpool/bmtpool.go
@@ -1,0 +1,28 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package bmtpool
+
+import (
+	"github.com/ethersphere/bee/pkg/swarm"
+	bmtlegacy "github.com/ethersphere/bmt/legacy"
+	"github.com/ethersphere/bmt/pool"
+)
+
+var instance pool.Pooler
+
+func init() {
+	instance = pool.New(8, swarm.BmtBranches)
+}
+
+// Get a bmt Hasher instance.
+// Instances are reset before being returned to the caller.
+func Get() *bmtlegacy.Hasher {
+	return instance.Get()
+}
+
+// Put a bmt Hasher back into the pool
+func Put(h *bmtlegacy.Hasher) {
+	instance.Put(h)
+}
--- a/pkg/bmtpool/bmtpool.go
+++ b/pkg/bmtpool/bmtpool.go
@@ -24,5 +24,5 @@ func Get() *bmtlegacy.Hasher {
 
 // Put a bmt Hasher back into the pool
 func Put(h *bmtlegacy.Hasher) {
-	instance.Put(h)
+	go instance.Put(h)
 }
--- a/pkg/bmtpool/bmtpool.go
+++ b/pkg/bmtpool/bmtpool.go
@@ -24,5 +24,5 @@ func Get() *bmtlegacy.Hasher {
 
 // Put a bmt Hasher back into the pool
 func Put(h *bmtlegacy.Hasher) {
-	go instance.Put(h)
+	instance.Put(h)
 }
--- a/pkg/file/pipeline/bmt/bmt.go
+++ b/pkg/file/pipeline/bmt/bmt.go
@@ -37,10 +37,12 @@ func (w *bmtWriter) ChainWrite(p *pipeline.PipeWriteArgs) error {
 	hasher := bmtpool.Get()
 	err := hasher.SetSpanBytes(p.Data[:swarm.SpanSize])
 	if err != nil {
+		bmtpool.Put(hasher)
 		return err
 	}
 	_, err = hasher.Write(p.Data[swarm.SpanSize:])
 	if err != nil {
+		bmtpool.Put(hasher)
 		return err
 	}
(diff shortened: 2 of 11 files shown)


pkg/bmtpool/bmtpool.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/ethersphere/bmt/pool"
+)
+
+var instance pool.Pooler
+
+func init() {
+	instance = pool.New(8, swarm.BmtBranches)
+}
+
+func Pool() pool.Pooler {
```
janos, line 18: How about to add Get at Put functions in this package and instead calling `bmtpool.Pool().Get()`, just to have `bmtpool.Get()`? And to remove the Pool() function?

janos (commented):
I am curious if this change made some performance improvements to be shown.

janos (approved):

pkg/bmtpool/bmtpool.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+// Get a bmt Hasher instance.
+// Instances are reset before being returned to the caller.
+func Get() *bmtlegacy.Hasher {
+	return instance.Get()
+}
+
+// Put a bmt Hasher back into the pool
+func Put(h *bmtlegacy.Hasher) {
```
zbiljic, line 26: Side comment (not important): "Release" might be more precise name for 'pool' resource.

pkg/file/pipeline/bmt/bmt.go
```diff
@@ -39,24 +34,22 @@ func (w *bmtWriter) ChainWrite(p *pipeline.PipeWriteArgs) error {
...
 		return err
 	}
-	_, err = w.b.Write(p.Data[swarm.SpanSize:])
+	_, err = hasher.Write(p.Data[swarm.SpanSize:])
 	if err != nil {
 		return err
 	}
-	p.Ref = w.b.Sum(nil)
+	p.Ref = hasher.Sum(nil)
+	bmtpool.Put(hasher)
```
zbiljic, line 47: This "hasher" will never be returned to the pool if some error occurs (however unlikely these two operation will fail).

pkg/file/splitter/internal/job.go
```diff
@@ -167,16 +157,19 @@ func (s *SimpleSplitterJob) sumLevel(lvl int) ([]byte, error) {
...
 		return nil, err
 	}
-	_, err = s.hasher.Write(c[8:])
+	_, err = hasher.Write(c[8:])
 	if err != nil {
 		return nil, err
 	}
-	ref := s.hasher.Sum(nil)
+	ref := hasher.Sum(nil)
+	bmtpool.Put(hasher)
```
zbiljic, line 171: Same issue with leaking resource.

pkg/node/node.go
```diff
@@ -357,22 +357,22 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey,
 		return nil, fmt.Errorf("swarm key: %w", err)
 	}
 
-	psss := pss.New(swarmPrivateKey, logger)
-	b.pssCloser = psss
+	pss := pss.New(swarmPrivateKey, logger)
```
zbiljic, line 360: I do not think it is good idea to rename this variable to `pss`.

Even though it might be nicer to have it like that, package with the same name is used here, and prevent any further use of that package. While in this instance there might not be any further need for something from `pss` package, we should not override name. I vote to maybe have this as guideline for Go code.

pkg/node/node.go
```diff
@@ -344,8 +344,8 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey,
 	chunkvalidator := swarm.NewChunkValidator(soc.NewValidator(), content.NewValidator())
 
 	retrieve := retrieval.New(swarmAddress, storer, p2ps, kad, logger, acc, accounting.NewFixedPricer(swarmAddress, 10), chunkvalidator, tracer)
-	tagg := tags.NewTags(stateStore, logger)
-	b.tagsCloser = tagg
+	tags := tags.NewTags(stateStore, logger)
```
zbiljic, line 347: Same comment as for `pss` package.

zbiljic (changes requested):

pkg/node/node.go
acud: @zelig wanted this renamed and I second that. I don't see a problem with having a variable with the same name like the package. If the compiler allows it then its probably legal no? @janos @zelig would be happy to hear your opinion.

janos: The problem with shadowing package name with a variable is in maintainability, as if the package is needed later in the code and time, a larger changeset is needed in order to rename the variable.

In this particular case, the last `s` in `psss` may refer to the Service, or to name the pss instance to `pssService` for clarity.

Even the compiler allows this type of shadowing, maybe there is a linter that warns about this. I would go with @zbiljic's suggestion to have a guideline.

pkg/file/splitter/internal/job.go
zbiljic: @acud this comment is from previous review, same as the other (you fixed), where the `hasher` is not returned to the 'bmtpool' if one of these two error occur.

acud: sorry missed this. fixed :+1:

zbiljic (approved):

PR #826, opened 2020-10-09
blocklist peer on retrieval timeout

janos:
This PR implements blocklisting i retrieval when the retrieval times out by a peer. 

Also, it passes a context in message exchange where it was missing currently for correctness.

### DIFF
--- a/.github/workflows/slash-beekeeper.yml
+++ b/.github/workflows/slash-beekeeper.yml
@@ -20,7 +20,7 @@ jobs:
           command: beekeeper
           reaction-type: "rocket"
       - name: Get PR Branch
-        if: steps.command.outputs.command-arguments == 'beekeeper'
+        if: steps.command.outputs.command-name == 'beekeeper'
         id: comment-branch
         uses: xt0rted/pull-request-comment-branch@v1
         with:
--- a/.github/workflows/slash-beekeeper.yml
+++ b/.github/workflows/slash-beekeeper.yml
@@ -1,18 +1,36 @@
-name: Beekeeper
+name: Slash Beekeeper
 
 on:
-  repository_dispatch:
-    types: [trigger-beekeeper]
+  issue_comment:
+    types: [created]
 
 jobs:
   beekeeper:
+    name: Execute "/beekeeper" slash command
+    if: github.event.issue.pull_request != '' && contains(github.event.comment.body, '/beekeeper')
     env:
       REPLICA: 3
     runs-on: ubuntu-latest
     steps:
+      - name: Check for Command
+        id: command
+        uses: xt0rted/slash-command-action@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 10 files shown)


vandot:
testing if this comment will trigger pipeline

ldeffenb:
Yes, please!  This will hopefully stop (or at least slow down) the endless requesting of chunks that I sometimes see in my clients.

I'm not great at reading diff listings on the commit, but how long does the peer remain blocklisted?

ldeffenb:
Can/should we do this also in the pusher when it gets a stream-reset?

time="2020-10-09T10:17:45-04:00" level=debug msg="pusher: error while sending chunk 49795cb965cbacd7a11bdf9678bff60f86fbf1ca84ec40e75d2b063a04d8e3f4 or receiving receipt: receive receipt from peer 7b8dcf3cc74795801ca6f6ef8c90c49ac82af0348eae1e2c18338cdccca55476: stream reset"

This is the other loop I see my clients going into.  This one is actually worse from a bandwidth perspective because it pushes the entire chunk out instead of just a CID on a retrieval.

ldeffenb:
But then again, now that I've written it down, disconnecting and blocklisting the peer might be too extreme as this loop seems to be CID-related and not generally something that the specific peer is "doing bad".  But right now, the pusher cannot tell what the issue was with the pushed-to-target because it doesn't return errors, but simply resets the stream.

ldeffenb:
Here's a graphical image of why I'm commenting on the pusher.  This is my network bandwidth (outbound is positive) with a local peer looping on a bunch of chunks that all are resulting in a stream reset from 1 or 2 upstream peers.

https://gateway.ethswarm.org/files/1e6f979c6402b889a99ddf0f2fa7d2a7d9a9fe1c9e343637bbcb435e15d32afc

janos:
@ldeffenb 

> I'm not great at reading diff listings on the commit, but how long does the peer remain blocklisted?

Currently only 1 minute. This is completely my subjective estimation.

> Can/should we do this also in the pusher when it gets a stream-reset?

As you already stated the problems, there should be a way to identify conditions when blocklisting should happen in pushsync as well. This would improve the protocol effectiveness for sure.

zbiljic (approved):

.github/workflows/slash-beekeeper.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           command: beekeeper
           reaction-type: "rocket"
       - name: Get PR Branch
-        if: steps.command.outputs.command-arguments == 'beekeeper'
+        if: steps.command.outputs.command-name == 'beekeeper'
```
acud, line 23: is this needed?

acud (approved):

.github/workflows/slash-beekeeper.yml
janos: Not actually, as it was merged to master recently.

Issue #532, opened 2020-08-05
Upload while uploading


Eknir:
# User story

## Story

As a user of Swarm who has just uploaded a big file, I want to be able to do another upload while the big file is still uploading, such that I don't get the feeling that my node is crashed.

## Acceptance criteria

- User uploads big file
- User can upload another file while big file is not yet finished

## Background

This issue came up during the [testing session](https://hackmd.io/hVr1MklhSRGcbuM7DgTypQ), where people had the idea that their node didn't work, whereas in fact their pusher was just pushing other stuff.

A solution could be to implement a FIFI queue for the pusher (now it is LIFO), or somehow being able to stop/pause/continue an existing upload.

## Tasks

| Task | Assignee |  Done (1), or 0 |
| ------ |------------| -----|

mortelli:
we can test this through 2 scenarios:

- upload a big file through the api, and while it is being chunked/pushed, repeat the same operation for a smaller file, and then verify which is finished first
- upload a big file through the api, and while iti s being chunked/pushed, stop the node. then, start up the node again and upload the smaller file and verify

acud:
i dont understand this ticket. what is meant by `such that I don't get the feeling that my node is crashed`

Eknir:
This ticket was based on real user feedback. It happened that the user uploaded a big file, didn't have verbosity on 5. Now, it took ages for this file to be push-synced, so the user just had the feeling that the upload crashed. He tried to upload another (smaller file), and uploading this file also didn't work, because the pusher was still pushing the big file.

Eknir:
@zbiljic , are you working on this?

Eknir:
@significance ,

Could you be so kind to do some QA on this issue? Verify that it works for some uploads and comment on the UX?

significance:
@Eknir i would love to!

PR #827, opened 2020-10-09
Reverse the order of the queue for push sync index

zbiljic:
This change adds support for reverse iteration over index items, and uses that for processing push index.

Since the order is reversed, last successful processed item is no longer tracked (to be used to start on next iteration). First item from iteration is saved, and each item from iteration is compared to that one, and if reached iteration stopped.
That backward processing results in FIFO processing of added chunks into push index.

relates to: #532

### DIFF
--- a/pkg/localstore/subscription_push.go
+++ b/pkg/localstore/subscription_push.go
@@ -17,6 +17,7 @@
 package localstore
 
 import (
+	"bytes"
 	"context"
 	"sync"
 	"time"
@@ -52,14 +53,15 @@ func (db *DB) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, stop fun
 		// close the returned chunkInfo channel at the end to
 		// signal that the subscription is done
 		defer close(chunks)
-		// sinceItem is the Item from which the next iteration
-		// should start. The first iteration starts from the first Item.
-		var sinceItem *shed.Item
+		// lastItem is the first Item received in the last iteration.
+		var lastItem *shed.Item
+		// toItemKey is the key for the Item that was oldest in the last iteration.
+		var toItemKey []byte
 		for {
 			select {
 			case <-trigger:
 				// iterate until:
-				// - last index Item is reached
+				// - last non-processed Item is reached
 				// - subscription stop is called
 				// - context is done.met
 				db.metrics.SubscribePushIteration.Inc()
@@ -73,12 +75,28 @@ func (db *DB) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, stop fun
 						return true, err
 					}
 
+					// check if we reached item that was already processed
+					// and stop
+					if toItemKey != nil {
+						dataItemKey, err := db.pushIndex.ItemKey(dataItem)
+						if err != nil {
+							return true, err
+						}
+
+						if bytes.Equal(dataItemKey, toItemKey) {
+							toItemKey = nil
+							return true, nil
+						}
+					}
+
 					select {
 					case chunks <- swarm.NewChunk(swarm.NewAddress(dataItem.Address), dataItem.Data).WithTagID(item.Tag):
 						count++
-						// set next iteration start item
-						// when its chunk is successfully sent to channel
-						sinceItem = &item
+						// when the chunk is successfully sent to channel
+						// we set first one sent, which is "oldest" at that point
+						if lastItem == nil {
+							lastItem = &item
+						}
 						return false, nil
 					case <-stopChan:
 						// gracefully stop the iteration
@@ -92,10 +110,7 @@ func (db *DB) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, stop fun
 						return true, ctx.Err()
 					}
 				}, &shed.IterateOptions{
-					StartFrom: sinceItem,
-					// sinceItem was sent as the last Address in the previous
-					// iterator call, skip it in this one
-					SkipStartFromItem: true,
+					Reverse: true,
 				})
 
 				totalTimeMetric(db.metrics.TotalTimeSubscribePushIteration, iterStart)
@@ -105,6 +120,23 @@ func (db *DB) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, stop fun
 					db.logger.Debugf("localstore push subscription iteration: %v", err)
 					return
 				}
+
+				// save last Item from this iteration in order to know were
+				// to stop on next iteration
+				if lastItem != nil && toItemKey == nil {
+					var lastItemKey []byte
+					// move 'toItemKey' to point to last item in previous iteration
+					lastItemKey, err = db.pushIndex.ItemKey(*lastItem)
+					if err != nil {
+						return
+					}
+
+					toItemKey = append(lastItemKey[:0:0], lastItemKey...)
+				}
+
+				// 'lastItem' should be populated on next iteration again
+				lastItem = nil
+
 			case <-stopChan:
 				// terminate the subscription
 				// on stop
--- a/pkg/localstore/subscription_push.go
+++ b/pkg/localstore/subscription_push.go
@@ -92,7 +92,6 @@ func (db *DB) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, stop fun
 					select {
 					case chunks <- swarm.NewChunk(swarm.NewAddress(dataItem.Address), dataItem.Data).WithTagID(item.Tag):
 						count++
-						// when the chunk is successfully sent to channel
 						// we set first one sent, which is "oldest" at that point
 						if lastItem == nil {
 							lastItem = &item
@@ -121,17 +120,14 @@ func (db *DB) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, stop fun
 					return
 				}
 
-				// save last Item from this iteration in order to know were
+				// save last Item from this iteration in order to know where
 				// to stop on next iteration
 				if lastItem != nil && toItemKey == nil {
-					var lastItemKey []byte
 					// move 'toItemKey' to point to last item in previous iteration
-					lastItemKey, err = db.pushIndex.ItemKey(*lastItem)
+					toItemKey, err = db.pushIndex.ItemKey(*lastItem)
 					if err != nil {
 						return
 					}
-
-					toItemKey = append(lastItemKey[:0:0], lastItemKey...)
 				}
 
 				// 'lastItem' should be populated on next iteration again
--- a/pkg/localstore/subscription_push_test.go
+++ b/pkg/localstore/subscription_push_test.go
@@ -57,7 +57,7 @@ func TestDB_SubscribePush(t *testing.T) {
 		}
 	}
 
-	// REQUIRED_LOCK - chunksMu
+	// caller expected to hold lock on chunksMu
 	findChunkIndex := func(chunk swarm.Chunk) int {
 		for i, c := range chunks {
 			if chunk.Address().Equal(c.Address()) {
@@ -178,7 +178,7 @@ func TestDB_SubscribePush_multiple(t *testing.T) {
 		}
 	}
 
-	// REQUIRED_LOCK - addrsMu
+	// caller expected to hold lock on addrsMu
 	findAddressIndex := func(address swarm.Address) int {
 		for i, a := range addrs {
 			if a.Equal(address) {
--- a/pkg/localstore/subscription_push_test.go
+++ b/pkg/localstore/subscription_push_test.go
@@ -37,6 +37,8 @@ func TestDB_SubscribePush(t *testing.T) {
 	chunks := make([]swarm.Chunk, 0)
 	var chunksMu sync.Mutex
 
+	chunkProcessedTimes := make([]int, 0)
+
 	uploadRandomChunks := func(count int) {
 		chunksMu.Lock()
 		defer chunksMu.Unlock()
@@ -50,7 +52,20 @@ func TestDB_SubscribePush(t *testing.T) {
 			}
 
 			chunks = append(chunks, ch)
+
+			chunkProcessedTimes = append(chunkProcessedTimes, 0)
 		}
+	}
+
+	// REQUIRED_LOCK - chunksMu
+	findChunkIndex := func(chunk swarm.Chunk) int {
+		for i, c := range chunks {
+			if chunk.Address().Equal(c.Address()) {
+				return i
+			}
+		}
+
+		return -1
 	}
 
 	// prepopulate database with some chunks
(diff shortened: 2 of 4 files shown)


pkg/localstore/subscription_push.go
```diff
@@ -52,14 +53,15 @@ func (db *DB) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, stop fun
 		// close the returned chunkInfo channel at the end to
 		// signal that the subscription is done
 		defer close(chunks)
-		// sinceItem is the Item from which the next iteration
-		// should start. The first iteration starts from the first Item.
-		var sinceItem *shed.Item
+		// lastItem is the first Item received in the last iteration.
+		var lastItem *shed.Item
+		// toItemKey is the key for the Item that was oldest in the last iteration.
```
acud, line 58: what does this mean? if `lastItem` is the first item that was received in the last iteration, then it is also the oldest item in the last iteration, and then why the differentiation calling one `last` and one `to`

pkg/localstore/subscription_push.go
```diff
@@ -73,12 +75,28 @@ func (db *DB) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, stop fun
...
+						}
+					}
+
 					select {
 					case chunks <- swarm.NewChunk(swarm.NewAddress(dataItem.Address), dataItem.Data).WithTagID(item.Tag):
 						count++
-						// set next iteration start item
-						// when its chunk is successfully sent to channel
-						sinceItem = &item
+						// when the chunk is successfully sent to channel
```
acud, line 95: this comment is not really necessary, the code is self explanatory

pkg/localstore/subscription_push.go
```diff
@@ -105,6 +120,23 @@ func (db *DB) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, stop fun
 					db.logger.Debugf("localstore push subscription iteration: %v", err)
 					return
 				}
+
+				// save last Item from this iteration in order to know were
```
acud, line 124: were -> where

pkg/localstore/subscription_push.go
```diff
@@ -105,6 +120,23 @@ func (db *DB) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, stop fun
...
+				// to stop on next iteration
```
acud, line 125: i dont really understand the comment. the last item from this iteration is actually where we _start_ the next iteration no?

pkg/localstore/subscription_push.go
```diff
@@ -105,6 +120,23 @@ func (db *DB) SubscribePush(ctx context.Context) (c <-chan swarm.Chunk, stop fun
...
+				if lastItem != nil && toItemKey == nil {
+					var lastItemKey []byte
+					// move 'toItemKey' to point to last item in previous iteration
+					lastItemKey, err = db.pushIndex.ItemKey(*lastItem)
+					if err != nil {
+						return
+					}
+
+					toItemKey = append(lastItemKey[:0:0], lastItemKey...)
```
acud, line 134: the three index slice is rather exotic, why no just `toItemKey = lastItemKey`? `lastItemKey` is scoped in this block. I don't see why you need this overhead?
you could also just `toItemKey, err = db.pushIndex.ItemKey(*lastItem)` directly

pkg/localstore/subscription_push_test.go
```diff
@@ -50,9 +52,22 @@ func TestDB_SubscribePush(t *testing.T) {
 			}
 
 			chunks = append(chunks, ch)
+
+			chunkProcessedTimes = append(chunkProcessedTimes, 0)
 		}
 	}
 
+	// REQUIRED_LOCK - chunksMu
```
acud, line 60: what does the comment mean? that you're supposed to call this under lock? if so - you can just write in the comment `// caller expected to hold lock on chunksMu`

pkg/localstore/subscription_push_test.go
```diff
@@ -138,6 +178,17 @@ func TestDB_SubscribePush_multiple(t *testing.T) {
 		}
 	}
 
+	// REQUIRED_LOCK - addrsMu
```
acud, line 181: same

pkg/shed/index.go
```diff
@@ -303,21 +310,68 @@ func (f Index) Iterate(fn IndexIterFunc, options *IterateOptions) (err error) {
...
+	ok = it.Seek(startKey)
+
+	if !options.Reverse {
+		if !ok {
+			// stop iterator if seek has failed
+			return it.Error()
+		}
+	} else {
+		// reverse seeker
+		if options.StartFrom != nil {
```
acud, line 329: lots of nesting here. can't you just check that `!ok` just after the seek? if the original seek failed the whole function should return anyway no?

pkg/shed/index.go
```diff
@@ -303,21 +310,68 @@ func (f Index) Iterate(fn IndexIterFunc, options *IterateOptions) (err error) {
...
+			if !ok {
+				return it.Error()
+			}
+		} else {
+			// find last key for this index (and prefix)
+
+			// move cursor to last key
```
acud, line 336: duplicate or somehow related to the comment 2 lines above. code is self explanatory, not sure if they are needed

pkg/shed/index.go
```diff
@@ -303,21 +310,68 @@ func (f Index) Iterate(fn IndexIterFunc, options *IterateOptions) (err error) {
...
+					return it.Error()
+				}
+
+				// previous key should have proper prefix
+				ok = it.Prev()
+				if !ok {
+					return it.Error()
+				}
+			}
+
```
acud, line 361: whitespace not necessary

pkg/shed/index.go
```diff
@@ -336,6 +390,26 @@ func (f Index) Iterate(fn IndexIterFunc, options *IterateOptions) (err error) {
 	return it.Error()
 }
 
+// Increment the last byte that is not 0xFF, and returned a new byte array
```
acud, line 393: `and return`

acud: also, no need to capitalize the first word. if godoc spec is used, you should start the sentence with the name of the function

pkg/shed/index.go
```diff
@@ -303,21 +310,68 @@ func (f Index) Iterate(fn IndexIterFunc, options *IterateOptions) (err error) {
...
+			// find last key for this index (and prefix)
+
+			// move cursor to last key
+			ok = it.Last()
+			if !ok {
+				return it.Error()
+			}
+
+			if lastKeyHasPrefix := bytes.HasPrefix(it.Key(), prefix); !lastKeyHasPrefix {
+				// increment last prefix byte (that is not 0xFF) to try to find last key
```
acud, line 343: `(that is not 0xff)` should not be in brackets. `// increment the last prefix byte which is not 0xff to try to find the last key`.

also, i don't really understand why incrementing the last byte which is not `ff` should find the last key. this alters the value from which you would like to seek to (since when you provide the key you would like to start with adds a prefix to the schema prefix, so you might be starting from a wrong item or an item that does not exist). it is not really clear why this brute forcing is needed (or why it should yield the correct item to start with).

acud (changes requested):
nice work @zbiljic. it would make it easier to review if the changes to `shed` would be isolated to a different PR than the changes to the subscribe push logic

zbiljic:
> nice work @zbiljic. it would make it easier to review if the changes to `shed` would be isolated to a different PR than the changes to the subscribe push logic

Since the change was pretty trivial, and only in single commit (which can be seen separately), I figured to add actual wanted change in this PR.

pkg/localstore/subscription_push.go
zbiljic: comment left from previous code, will remove it

Eknir:
@zbiljic, I linked this with #532. Good for you?

zbiljic:
@Eknir I already mentioned that issue in this PR, however I did not want to link it, because there may be few other smaller PRs after this one, all of which should expand on this one, and further improve problem from the issue.

pkg/shed/index.go
zbiljic: Excuse this commit, I figured that I have left enough comments for any reader to follow logic.

This change revolves around adding support for reverse iteration over items in index, which are stored (in this case) in LevelDB (but changes should be same for many other similar style KV databases). Change being made is not to just support reverse iteration, but to add support for it, while also supporting "standard" iteration (from first key in index onward). Before first seek no check is performed to know exactly what is contained in the `startKey`.

Because there are multiple possible conditions, case in which error should checked must exists on multiple places. It may be possible to have more complex single condition first, however I think due to the nature of conditions second must be present also. I might have not considered some condition, but I see no way to make it while also leaving code in "readable" condition. I am open for suggestions.

First seek to `startKey` is left from old code, and it is fine to exists (even though currently not used for new 'reverse' use-case, it might be used on some other occasion if the `options.StartFrom` is configured with `options.Reverse`). That `startKey` could either be just index prefix, index prefix with some prefix, or some of those prefix combinations with exact key that should be found.

First `if` condition is to to check if we are iterating from topmost item from LevelDB (at least with prefix), and failing if there was error in that case. Conditions `!options.Reverse` and `!ok` here cannot be joined, as we need `else` condition for the case where 'reverse' iteration is performed. Of course we could have separate condition with both `if !options.Reverse && !ok {...`, and then after that another `if options.Reverse {...`, but those are functionally same (or have some couple of other complex conditions).

At this point we are in the case for reverse order. Separate check if `options.StartFrom` is configured is needed in this case because exact key should be found in any case.

In the case where we only have prefix, and are doing reverse iteration, a bit more complex operation for finding end is needed. This is especially true because this index structure can contain multiple indexes in the same LevelDB database.

Initial step is to just try to find last item in the database, using `it.Last()`. In the case where there is only single index, we have last item, and are done there. However, when there are multiple indexes we have additional work to find last item for needed index (and prefix).

This is done in the `if lastKeyHasPrefix := bytes.HasPrefix(it.Key(), prefix); !lastKeyHasPrefix {...` condition. This is only performed in the case when `options.StartFrom` is not configured.
Way to find last item in index (for prefix) is to find first next item, and then go to previous item, at which point cursor should be on valid item. Note here that I also partly rely on later logic which checks if item has proper prefix before returning item.

In the case where multiple indexes exists, or we also have some prefix, finding key after some other is (sort-of) known. Since keys are basically sorted byte arrays, incrementing last byte in array for key and seeking to it should move to subsequent key, even if exact key that was put in the `Seek` function does not exists.

For this example (for index `\x02` - first index ID, and prefix `want`):

```txt
"\x00" "{\"fields\":{},\"indexes\":{\"2\":{\"name\":\"retrieval-index\"},\"3\":{\"name\":\"second-index\"}}}
"\x02skip-hash-01" "value"
"\x02skip-hash-02" "value"
"\x02want-hash-00" "value"
"\x02want-hash-04" "value"
"\x02want-hash-05" "value"
"\x03iterate-hash-10" "value"
```

we want to iterate over items with prefix `\x02want`, incremented prefix will be `\x02wanu`, which will on `Seek` operation move cursor to `\x03iterate-hash-10`. Moving to previous database item should move cursor to correct index.
For this case also we could use new prefix of `\x02wao` (if the actual prefix was `\x02wan\xff`), and which would result in the same end result. In the case where only index prefix is used, i.e. prefix is `\x02`, the new prefix would be `\x03`, and same end goal would be reached.

In the case where last byte of the key is `0xFF` (which is, of course, highly unlikely or impossible in our case) we remove that last byte, and increment previous one.

I do not consider this in any way brute-force way, there are only 2 operations here: one to find next item, and one to go backward.
This logic cannot be compared to something like in PR #821, where there actual numerical offset is used to retrieve page. Even though actual logic is correct, which is why I approved it, it is a bit too much. Not sure why actual pages are used, and why have we not returned some cursor, which could be used for further iterations. Or even last pin could be used to start from that one.

The tests I added (copied and changed) should cover most conditions pretty well.
I am open to precise suggestion on how current change can be improved.

pkg/localstore/subscription_push.go
zbiljic: Actually no, last item is temporary variable, used to save first received item from iterator, which should be currently last one in the database.
As far as I could find through the code, once started, all elements are being processed from `pushIndex` once started.
If all are processed, first one that was received (which is last in the database), is copied to `toItemKey` variable, which is used to compare on next loops for iteration to know on which item was already processed before (depending on the reason why the iteration stopped).

In the example where we start with this initial elements in the database:

```txt
01
02
03
```

first item processed will be `03`, and will be stored in `lastItem` variable. Next item will be `02` and then `01`.
If some new items are added in the meantime, we might come to situation like this (assuming no removal of previous items happened):

```txt
01
02
03
04
05
```

On the next iteration first item will be `05`, and we again go backward, to `04`. However once we come to `03` we do not want to process that item again, as we already did. That is why each item key is compared to `toItemKey`, and why we stop iteration in that case.

This should basically result in same process that was here previously, where iteration was being started from last processed chunk.

We cannot configure option to start from some other item, as that would mean starting from some middle of items. We always want to start from last added item.
I have considered having that option in addition to these changes, however I could not find condition where this would actually be used, which is why it was removed.

janos (approved):
I could not find any missed edge cases, it looks very thoroughly written. 👍

pkg/localstore/subscription_push.go
zbiljic: In theory it is possible. However, as I noted in the comment, I have gone through all the cases that I could, and could not find scenario where that happens. If any error occurs, it should complete enclosing goroutine.

As far as I understood previous code, it basically iterates through all items, until it reaches end (processes all elements), and then when triggered starts from last processed item (previously used `sinceItem` variable). If there was some way that iteration could stop prematurely and without an error I would like to know, to make sure I do no miss it. As for now, I could not find it.

New code should to the same, only in reversed order. Once iteration starts, it should go through all elements until it reaches end (sorted first key in LevelDB). On the next trigger, it should start from last element in the index, and at most to the previous element that was already processed (skipping it).

Only possible scenario that I could imagine is that somehow during iteration of LevelDB, some element in the middle of the queue is removed (before it is processed), and `leveldb.ErrNotFound` error is reached, at which point internal iteration loop is stopped, at which point we have a gap in the middle, which should require re-subscription to start whole process again (which does happen in some cases).
However I could not find case when random items are removed from push index before being processed.

pkg/shed/index.go
acud: Cool. Thanks for the explanation. If you have suggestions to improve other pull request you should make them :+1:

acud (approved):

pkg/localstore/subscription_push.go
acud: cool. very thorough :+1: thanks

Eknir:
Hi @zbiljic , I see your comment now, nevertheless, I still connected them (before your comment). Feel free to de-connect them and make issues for things which you deem are needed to complete the main issue (let's call it epic).

PR #840, opened 2020-10-15
propagate context to message read and write in p2p protocols

janos:
This PR adjusts protobuf message read and write calls to use context for cancelation, by either adding it or replacing with timeout variant. More cancelation control is provided with context, so functions with timeout are removed. These changes may improve overall communication in not perfect network conditions and with overloaded peers.

Since pseudosettle is changed to use context in handler while reading a message, TestPayment was very flaky, as the context cancled in Pay method would cancel the context passed to the handler through streamtest Recorder. Two changes were made to fix this. First not to couple client call context with handler context in streamtest Recorder, and the second is are improvements to TestPayment to check for record error and also to synchronize observer call with a channel.

### DIFF
--- a/pkg/p2p/libp2p/internal/handshake/handshake.go
+++ b/pkg/p2p/libp2p/internal/handshake/handshake.go
@@ -5,6 +5,7 @@
 package handshake
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"sync"
@@ -33,7 +34,7 @@ const (
 	StreamName = "handshake"
 	// MaxWelcomeMessageLength is maximum number of characters allowed in the welcome message.
 	MaxWelcomeMessageLength = 140
-	messageTimeout          = 5 * time.Second
+	handshakeTimeout        = 15 * time.Second
 )
 
 var (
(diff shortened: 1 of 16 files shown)


pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -33,7 +34,7 @@ const (
 	StreamName = "handshake"
 	// MaxWelcomeMessageLength is maximum number of characters allowed in the welcome message.
 	MaxWelcomeMessageLength = 140
-	messageTimeout          = 5 * time.Second
+	handshakeTimeout        = 15 * time.Second
```
acud, line 37: why the increase in timeout? this is quite a lot of time no?

acud (approved):
very cool. thanks janos

pkg/p2p/libp2p/internal/handshake/handshake.go
janos: This is now a different timeout. `messageTimeout` covered only one message in the handshake, while `handshakeTimeout` covers the whole handshake process. I think that this timeout should be a protective one. But it is completely set by a guess by me. Would 5s for the whole handshake be more reasonable?

acud: I see. Let's leave it as is for now :+1:

zbiljic (approved):

PR #844, opened 2020-10-16
Wait until pusher active jobs are done on Close

zbiljic:
Change increases timeout for waiting for pending push operations to terminate. When the `Close()` function is invoked on push service, it closes the `quit` channel which triggers closing of the worker, which has logic that waits for some time to terminate all active goroutines through semaphore.

### DIFF
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -214,6 +214,7 @@ func (s *Service) setChunkAsSynced(ctx context.Context, ch swarm.Chunk) error {
 }
 
 func (s *Service) Close() error {
+	s.logger.Info("pusher shutting down")
 	close(s.quit)
 
 	// Wait for chunks worker to finish
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -193,7 +193,7 @@ LOOP:
 
 	select {
 	case <-closeC:
-	case <-time.After(2 * time.Second):
+	case <-time.After(5 * time.Second):
 		s.logger.Warning("pusher shutting down with pending operations")
 	}
 }
@@ -219,7 +219,7 @@ func (s *Service) Close() error {
 	// Wait for chunks worker to finish
 	select {
 	case <-s.chunksWorkerQuitC:
-	case <-time.After(3 * time.Second):
+	case <-time.After(6 * time.Second):
 	}
 	return nil
 }
--- a/pkg/pusher/pusher_test.go
+++ b/pkg/pusher/pusher_test.go
@@ -31,18 +31,38 @@ var noOfRetries = 20
 // Wrap the actual storer to intercept the modeSet that the pusher will call when a valid receipt is received
 type Store struct {
 	storage.Storer
-	modeSet   map[string]storage.ModeSet
-	modeSetMu *sync.Mutex
+	internalStorer storage.Storer
+	modeSet        map[string]storage.ModeSet
+	modeSetMu      *sync.Mutex
+
+	closed              bool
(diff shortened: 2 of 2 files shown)


pkg/pusher/pusher_test.go
```diff
@@ -214,6 +234,107 @@ func TestSendChunkAndTimeoutinReceivingReceipt(t *testing.T) {
...
+	storer.modeSetMu.Unlock()
+
+	select {
+	case goFuncAfterCloseC <- struct{}{}:
+	case <-time.After(5 * time.Second):
+		t.Fatal("timed out waiting for chunk")
+	}
+
+	// we need this to allow some goroutines to complete
+	time.Sleep(10 * time.Millisecond)
```
janos, line 316: This sleep can be longer.

pkg/pusher/pusher_test.go
```diff
@@ -214,6 +234,107 @@ func TestSendChunkAndTimeoutinReceivingReceipt(t *testing.T) {
...
+	}
+	if storer.setAfterCloseCount != 0 {
+		t.Fatalf("store 'Set' called %d times after close, expected 0", storer.setAfterCloseCount)
+	}
+	storer.modeSetMu.Unlock()
+
+	// should be closed by now
+	select {
+	case <-pusherClosedC:
+	case <-time.After(time.Millisecond):
```
janos, line 333: This timeout can be longer.

janos (approved):
👍

pkg/pusher/pusher_test.go
zbiljic: My bad, forgot to push latest local change.

acud (approved):
LGTM

pkg/pusher/pusher.go
```diff
@@ -219,7 +219,7 @@ func (s *Service) Close() error {
 	// Wait for chunks worker to finish
 	select {
 	case <-s.chunksWorkerQuitC:
-	case <-time.After(3 * time.Second):
+	case <-time.After(6 * time.Second):
```
acud, line 222: you could theoretically return an error here, and then it would be printed in the shutdown sequence. this should be done on other components too (but as a separate issue)

zbiljic: This point should not be reached due to other change, with lower timeout.

However, it might be useful to have that other change in another small loop, and print some message every second (or so) while waiting to stop worker ?

acud: it should not be reached, but if it does - we should get some sort of signalling. you can have a look at other services that shut down and have this sort of shutdown sequence, they usually print a warning when the timeout is reached without clean shutdown. this should ideally be replaced by using the errors returned in the shutdown sequence in `node.go` (at least i am at this opinion)

acud: about printing the message at intervals :+1: but janos was at an opposition to this

zbiljic: There is one such log line next to other changed timeout, and I have added `s.logger.Info("pusher shutting down")` line in the beginning of `Close()` function.
This makes it same as for `pullsync` package.

zbiljic: On another point, I have checked how it looks when printing log every second or so yesterday, and did not like how it looks (that is personal opinion). Unless we can change existing line with some progress (while possible, we should not bother here), we probably should not print on interval until closed.
I think that line where notifying user that operation has started is good enough.

acud: sounds good :+1:

acud (approved):

PR #866, opened 2020-10-23
Update package documentation

zbiljic:
This change adds some additional documentation on several packages.

### DIFF
--- a/pkg/addressbook/addressbook.go
+++ b/pkg/addressbook/addressbook.go
@@ -20,10 +20,13 @@ var _ Interface = (*store)(nil)
 
 var ErrNotFound = errors.New("addressbook: not found")
 
+// Interface is the AddressBook interface.
 type Interface interface {
 	GetPutter
 	Remover
+	// Overlays returns a list of all overlay addresses saved in addressbook.
 	Overlays() ([]swarm.Address, error)
+	// Addresses returns a list of all bzz.Address-es saved in addressbook.
 	Addresses() ([]bzz.Address, error)
 }
 
@@ -33,14 +36,17 @@ type GetPutter interface {
 }
 
 type Getter interface {
+	// Get returns pointer to saved bzz.Address for requested overlay address.
 	Get(overlay swarm.Address) (addr *bzz.Address, err error)
 }
 
 type Putter interface {
+	// Put saves relation between peer overlay address and bzz.Address address.
 	Put(overlay swarm.Address, addr bzz.Address) (err error)
 }
 
 type Remover interface {
+	// Remove removes overlay address.
 	Remove(overlay swarm.Address) error
 }
 
@@ -48,6 +54,7 @@ type store struct {
 	store storage.StateStorer
 }
 
+// New creates new addressbook for state storer.
 func New(storer storage.StateStorer) Interface {
 	return &store{
 		store: storer,
--- a/pkg/addressbook/doc.go
+++ b/pkg/addressbook/doc.go
@@ -5,6 +5,7 @@
 /*
 Package addressbook provides persisted mapping between overlay (topology) address
 and bzz.Address address, which contains underlay (physical) address.
+The underlay address contains both physical and p2p addresses.
 
 It is single point of truth about known peers and relations of their
 overlay and underlay addresses.
--- a/pkg/addressbook/doc.go
+++ b/pkg/addressbook/doc.go
@@ -1,0 +1,12 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+/*
+Package addressbook provides persisted mapping between overlay (topology) address
+and bzz.Address address, which contains underlay (physical) address.
+
+It is single point of truth about known peers and relations of their
+overlay and underlay addresses.
+*/
+package addressbook
--- a/pkg/keystore/keystore.go
+++ b/pkg/keystore/keystore.go
@@ -9,9 +9,16 @@ import (
 	"errors"
 )
 
+// ErrInvalidPassword is returned when the password for decrypting content where
+// private key is stored is not valid.
 var ErrInvalidPassword = errors.New("invalid password")
 
+// Service for managing keystore private keys.
 type Service interface {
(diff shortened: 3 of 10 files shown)


pkg/addressbook/addressbook.go
```diff
@@ -33,21 +36,25 @@ type GetPutter interface {
 }
 
 type Getter interface {
+	// Get returns pointer to saved bzz.Address for requested overlay address.
 	Get(overlay swarm.Address) (addr *bzz.Address, err error)
 }
 
 type Putter interface {
+	// Put saves relation between peer overlay address, and bzz.Address address.
```
janos, line 44: I think that `,` is not needed.

pkg/addressbook/doc.go
```diff
@@ -0,0 +1,12 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+/*
+Package addressbook provides persisted mapping between overlay (topology) address
+and bzz.Address address, which contains underlay (physical) address.
```
janos, line 7: I would say that underlay contains both physical and p2p addresses.

janos (approved):
Great! My comments are very minor.

pkg/addressbook/doc.go
zbiljic: My thinking for this instance was that that reader would follow `bzz.Address` implementation and figure out rest there.
Anyway, I update this comment according to comment.

janos (approved):
Thanks. Very nice.

pkg/keystore/keystore.go
```diff
@@ -9,9 +9,16 @@ import (
...
 
+// ErrInvalidPassword is returned when the password for decrypting content where
+// private key is stored is not valid.
 var ErrInvalidPassword = errors.New("invalid password")
 
+// Service for managing keystore private keys.
 type Service interface {
+	// Key returns private key for specified name that was encrypted with
+	// provided password. If the private key does not exists it creates new one
+	// with name and password, and returns with created set to true.
```
zelig, line 20: // Key returns the private key for a  specified name that was encrypted with the
	// provided password. If the private key does not exists it creates a  new one
	// with a name and the password, and returns with `created` set to true.

pkg/keystore/mem/service.go
```diff
@@ -15,11 +15,17 @@ import (
 
 var _ keystore.Service = (*Service)(nil)
 
+// Service is the memory-based keystore.Service implementation.
+//
+// Keys are stored in the in-memory map, where the key is the name of the
+// private key, and value is structure where actual private key and password are
+// stored.
```
zelig, line 22: // Keys are stored in an in-memory map, where the key is the name of the
// private key, and the value is the structure where the actual private key and the password are
// stored.

pkg/keystore/test/test.go
```diff
@@ -12,6 +12,8 @@ import (
 	"github.com/ethersphere/bee/pkg/keystore"
 )
 
+// Service is utility testing function that can be used to test implementations
+// of keystore.Service interface.
```
zelig, line 16: // Service is a utility testing function that can be used to test implementations
// of the keystore.Service interface.

pkg/tracing/doc.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+/*
+Package tracing helps with propagation of tracing span through context in the
+system. It does this for operations contained to single node, as well as across
+nodes, by injecting special headers.
+
```
zelig, line 9: Package tracing helps with the propagation of the tracing span through context in the
system. It does this for operations contained to single node, as well as across
nodes, by injecting special headers.

pkg/tracing/doc.go
```diff
@@ -0,0 +1,51 @@
...
+To use tracing package Tracer instance must be created, which contains functions
+for starting new span contexes, injecting them in other data, and extracting
+active span them from context.
```
zelig, line 12: To use the tracing package, a Tracer instance must be created, which contains functions
for starting new span contexts, injecting them in other data, and extracting the
active span them from the context.

pkg/tracing/doc.go
```diff
@@ -0,0 +1,51 @@
...
+
+To use tracing package Tracer instance must be created:
```
zelig, line 14: To use the tracing package a Tracer instance must be created:

pkg/tracing/doc.go
```diff
@@ -0,0 +1,51 @@
...
+		ServiceName: "bee",
+	})
+	if err != nil {
+		// handle error
+	}
+	defer tracerCloser.Close()
+	// ...
+
+The tracer instance contains functions for starting new span contexes, injecting
+them in other data, and extracting active span them from context:
```
zelig, line 28: The tracer instance contains functions for starting new span contexts, injecting
them in other data, and extracting the active span them from the context:

pkg/tracing/doc.go
```diff
@@ -0,0 +1,51 @@
...
+
+	span, _, ctx := tracer.StartSpanFromContext(ctx, "operation-name", nil)
+
+Once the operation is finished, the open span should be finished:
+
+	span.Finish()
+
+The tracing package also provides function for creating logger which will inject
+"traceid" field entry to log line, which helps in finding out which log lines
+belong to specific trace.
```
zelig, line 38: The tracing package also provides a function for creating  a  logger which will inject
a "traceid" field entry to the log line, which helps in finding out which log lines
belong to a specific trace.

pkg/tracing/doc.go
```diff
@@ -0,0 +1,51 @@
...
+
+To create logger with trace just wrap existing logger:
```
zelig, line 40: To create a logger with trace just wrap an existing logger

zelig (changes requested):
the definite and indefinite articles are a bitch for slavic speakers but without them the text is very difficult to read.

zelig (approved):

acud (approved):
LGTM

PR #868, opened 2020-10-26
disable port reuse in libp2p

janos:
LibP2P TCP transport default behaviour is to use port reuse which allows to open multiple listeners on the same local port. This is confusing as to or more bee nodes can run on the same system with the same p2p port but makes non of them usable as packets are distributed among them.

With this change, bee node will exit with an error if another process is already listening on the same port, just as it is the case with api and debugapi listeners.


acud (approved):
sweet, thanks!

zbiljic (approved):

PR #870, opened 2020-10-27
Surplus balances for peers

metacertain:
The aim of this PR is to add safety for settlements that have an amount over current balance.
If a peer sends us an oversettlement, we want the extra amount to be stored as a reserve for future forwarding services towards the peer, but we do not want it to be part of regular balance, so that it can not form the base for triggering a settlement back.

### DIFF
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -306,7 +306,7 @@ func (a *Accounting) Debit(peer swarm.Address, price uint64) error {
 	cost := price
 
 	// see if peer has surplus balance to deduct this transaction of
-	if accountingPeer.surplus == true {
+	if accountingPeer.surplus {
 		surplusBalance, err := a.SurplusBalance(peer)
 		if err != nil {
 			if !errors.Is(err, ErrPeerNoBalance) {
@@ -343,6 +343,9 @@ func (a *Accounting) Debit(peer swarm.Address, price uint64) error {
 
 		// if surplus balance didn't cover full transaction, let's continue with leftover part as cost
 		debitIncrease, err := subtractU64mI64(price, surplusBalance)
+		if err != nil {
+			return err
+		}
 		// conversion to uint64 is safe because we know the relationship between the values by now, but let's make a sanity check
 		if debitIncrease <= 0 {
 			return fmt.Errorf("sanity check failed for partial debit after surplus balance drawn")
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -547,10 +547,6 @@ func (a *Accounting) NotifyPayment(peer swarm.Address, amount uint64) error {
 			return ErrOverflow
 		}
 
-		if increasedSurplus > math.MaxInt64 {
-			return ErrOverflow
-		}
-
 		err = a.store.Put(peerSurplusBalanceKey(peer), increasedSurplus)
 		if err != nil {
 			return fmt.Errorf("failed to persist surplus balance: %w", err)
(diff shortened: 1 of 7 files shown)


pkg/accounting/accounting.go
```diff
@@ -299,6 +303,56 @@ func (a *Accounting) Debit(peer swarm.Address, price uint64) error {
 	accountingPeer.lock.Lock()
 	defer accountingPeer.lock.Unlock()
 
+	cost := price
+
+	// see if peer has surplus balance to deduct this transaction of
+	if accountingPeer.surplus {
```
ralph-pichler, line 309: how does surplus become true after a restart?

pkg/accounting/accounting.go
```diff
@@ -51,6 +54,7 @@ type accountingPeer struct {
 	lock             sync.Mutex // lock to be held during any accounting action for this peer
 	reservedBalance  uint64     // amount currently reserved for active peer interaction
 	paymentThreshold uint64     // the threshold at which the peer expects us to pay
+	surplus          bool       // signals wether peer has surplus to be used for debiting
```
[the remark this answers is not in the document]
metacertain: to know at debit whether we have any surplus balance to deduct from.
otherwise to check whether surplus balance is available, we'd do a store get for the surplus balance stored (in each debit). so this way we just check this bool, lighter

pkg/accounting/accounting.go
metacertain: it does not, currently
however, we could initialize it in getAccountingPeer? either set it to true by default, or make a store get to see if it has > 0 surplus balance

pkg/accounting/accounting.go
```diff
@@ -345,11 +392,30 @@ func (a *Accounting) Balance(peer swarm.Address) (balance int64, err error) {
 	return balance, nil
 }
 
+// SurplusBalance returns the current balance for the given peer.
+func (a *Accounting) SurplusBalance(peer swarm.Address) (balance int64, err error) {
+	err = a.store.Get(peerSurplusBalanceKey(peer), &balance)
+
+	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			return 0, ErrPeerNoBalance
```
ralph-pichler, line 401: why not just return `0, nil`

pkg/accounting/accounting.go
```diff
@@ -299,6 +302,50 @@ func (a *Accounting) Debit(peer swarm.Address, price uint64) error {
...
+			return err
+		}
+
+		// if nothing left for debiting, store new surplus balance and return from debit
+		if newSurplusBalance >= 0 {
+			err = a.store.Put(peerSurplusBalanceKey(peer), newSurplusBalance)
+			if err != nil {
+				return fmt.Errorf("failed to persist surplus balance: %w", err)
+			}
+			return nil
```
ralph-pichler, line 327: what about metrics in that case? `TotalDebitedAmount` and `DebitEventsCount`. Technically this is still a debit operation.

ralph-pichler (approved):

janos (approved):

pkg/accounting/accounting.go
```diff
@@ -83,6 +90,8 @@ var (
 	ErrPeerNoBalance = errors.New("no balance for peer")
 	// ErrOverflow denotes an arithmetic operation overflowed.
 	ErrOverflow = errors.New("overflow error")
+	// ErrInValue denotes an invalid value read from store
+	ErrInValue = errors.New("invalid value")
```
acud, line 94: consider renaming to `ErrInvalidValue` since `InValue` might be a bit misleading

acud (approved):

Issue #872, opened 2020-10-27
upload with postage stamp: HTTP API 


acud:
As per the [architectural proposal](https://hackmd.io/rYze0NZ5Ryi2HNmc62ogWA?view) that deals with uploading with postage stamps, the following changes are needed in the HTTP API package:
- The API should accept a `Swarm-Postage-Batch-Id` header
- The API should pass the `BatchID` to the splitting pipeline

related to #654

PR #886, opened 2020-10-30
api: add postage batch id header

acud:
Adds a postage batch ID header to the API

related to #872

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -5,6 +5,7 @@
 package api
 
 import (
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
@@ -25,11 +26,12 @@ import (
 )
 
 const (
-	SwarmPinHeader           = "Swarm-Pin"
-	SwarmTagUidHeader        = "Swarm-Tag-Uid"
-	SwarmEncryptHeader       = "Swarm-Encrypt"
-	SwarmIndexDocumentHeader = "Swarm-Index-Document"
-	SwarmErrorDocumentHeader = "Swarm-Error-Document"
+	SwarmPinHeader            = "Swarm-Pin"
+	SwarmTagUidHeader         = "Swarm-Tag-Uid"
+	SwarmEncryptHeader        = "Swarm-Encrypt"
+	SwarmIndexDocumentHeader  = "Swarm-Index-Document"
+	SwarmErrorDocumentHeader  = "Swarm-Error-Document"
+	SwarmPostageBatchIdHeader = "Swarm-Postage-Batch-Id"
 )
 
 // The size of buffer used for prefetching content with Langos.
@@ -47,6 +49,7 @@ const (
 var (
 	errInvalidNameOrAddress = errors.New("invalid name or bzz address")
 	errNoResolver           = errors.New("no resolver connected")
+	errInvalidPostageBatch  = errors.New("invalid postage batch id")
 )
 
 // Service is the API service interface.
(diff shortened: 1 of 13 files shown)


pkg/post/stamper.go
```diff
@@ -0,0 +1,12 @@
+package post
```
zelig, line 1: this file is not needed and should be removed

zelig (commented):
- i do not see any tests
- the post package should be removed, see pr #890 
- batch id should be inherited through the context, not added to every method

acud:
> batch id should be inherited through the context, not added to every method

context attached values are optional by definition. if i understand correctly, batch id is not optional. correct me if i'm wrong

acud:
> i do not see any tests

I don't see what to test here (I kid you not). Let's add tests to this package once the stamper is wired into the pipeline, this way we can mock it and assert it was called with the correct value

pkg/api/api.go
```diff
@@ -47,8 +49,15 @@ const (
 var (
 	errInvalidNameOrAddress = errors.New("invalid name or bzz address")
 	errNoResolver           = errors.New("no resolver connected")
+	errInvalidPostageBatch  = errors.New("invalid postage batch id")
+
+	fallbackPostageBatch []byte
```
janos, line 54: This is a mutable shared variable between multiple goroutines and we cannot be sure that it will not be mutated, as it is passed to other components. I assume that it is an optimization to avoid allocations, but, I think that this optimization is not needed, and it is better to be safe.

janos (commented):
This PR needs tests for at least that the http header is handled as expected, like we have for other headers.

pkg/api/api.go
```diff
@@ -180,6 +189,18 @@ func requestEncrypt(r *http.Request) bool {
...
 
+func requestPostageBatchId(r *http.Request) ([]byte, error) {
+	if h := strings.ToLower(r.Header.Get(SwarmPostageBatchIdHeader)); h != "" {
+		if len(h) != 64 {
+			return nil, errInvalidPostageBatch
+		}
+		return hex.DecodeString(h)
+	}
+	b := make([]byte, len(fallbackPostageBatch))
+	_ = copy(b, fallbackPostageBatch)
```
janos, line 200: I suppose that this approach is to have a fallback postage back clearly defined on the package level, instead here just to return `make([]byte, 32)`. Could you add a comment?

janos (approved):
Thanks for addressing my comments. 👍

zelig (approved):
i do not understand fallback to all zeros batch id otherwise ok

pkg/api/api.go
```diff
@@ -180,6 +183,18 @@ func requestEncrypt(r *http.Request) bool {
 	return strings.ToLower(r.Header.Get(SwarmEncryptHeader)) == "true"
 }
 
+func requestPostageBatchId(r *http.Request) ([]byte, error) {
+	if h := strings.ToLower(r.Header.Get(SwarmPostageBatchIdHeader)); h != "" {
+		if len(h) != 64 {
+			return nil, errInvalidPostageBatch
+		}
+		return hex.DecodeString(h)
```
zbiljic, line 191: If value in the header is not hexadecimal it will return some internal error. This might have been handled, and returned `errInvalidPostageBatch` again.

zbiljic (approved):

pkg/api/api.go
acud: makes sense :+1:  thanks

PR #906, opened 2020-11-03
api: increase pss test read timeout before mining

acud:
Tests in `api/pss_test.go` are flaking. This is apparently since the read timeout for the websocket is set before mining the chunk using the real pss type. We should improve this by mocking pss in this test in a way that we don't need to mine the chunk every time and run through the actual pss code. This PR is meant to serve as a quick fix so everyone can continue working with the CI without needing the restart the builds constantly.

### DIFF
--- a/pkg/api/pss_test.go
+++ b/pkg/api/pss_test.go
@@ -30,11 +30,12 @@ import (
 )
 
 var (
-	target  = pss.Target([]byte{1})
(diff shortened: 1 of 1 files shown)


zbiljic (approved):
I have not observed this, however change seem fine.

janos (approved):

Issue #920, opened 2020-11-07
Get value of chunks


zelig:
# Story
As a storer node, I want to have an up-to-date representation of the value of all chunks that I store in my _reserve_, such that I can use this information to decide on which chunks to discard and what chunks to keep.

# Background
There exists an architecture proposal https://hackmd.io/o8RGsNwZSN6IE0lkBAZCPw, which most likely will be changed based on changes in the postage stamp smart contract ([see](https://hackmd.io/tvacCgfBSWmeConc3rFOtw)). 

The reserve is the name of storage space dedicated to chunks in the node’s area of responsibity.

# Acceptance criteria
- For all chunks that a node is storing as part of his _reserve_, the node knows the value

Eknir:
@zelig , I am adjusting this epic to be more user-focused, such that it is generic of the particular implementation. Changes in https://hackmd.io/tvacCgfBSWmeConc3rFOtw most likely require also changes in implementation. We would like to discuss this with you on Monday.

acud:
@Eknir can we close this epic? all issues associated with it are closed and I don't see much value in keeping this open

Issue #896, opened 2020-11-02
Stamping chunks during upload


Eknir:
Architecture proposal: https://hackmd.io/k42tvB44RWGXtpbpmdcv2g

PR #921, opened 2020-11-07
generic (smart contract) event listener 

zelig:
The PR implements an event listener as per the appendix of https://hackmd.io/o8RGsNwZSN6IE0lkBAZCPw?both (#920)
this is a prerequisite of #922 and #934 (#896).

### DIFF
--- a/pkg/listener/listener.go
+++ b/pkg/listener/listener.go
@@ -1,0 +1,78 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package listener
+
+import (
+	"context"
+	"math/big"
+
+	ethereum "github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+// Listener provides a blockchain event iterator
+type Listener interface {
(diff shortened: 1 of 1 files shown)


pkg/listener/listener.go
```diff
@@ -0,0 +1,78 @@
...
+// - it terminates  with no error when quit channel is closed
+// - if the update function returns an error, the call returns with that error
+func (lis *EthListener) Listen(from uint64, quit chan struct{}, f func(types.Log) error) error {
+	// need to cancel context even if terminate without error with quit channel
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	// using an event channel, we subscribe to filter logs emmitted by the postage contract(s)
+	// since the last block recorded that had a relevant event
+	events := make(chan types.Log)
+	sub, err := lis.client.SubscribeFilterLogs(ctx, query(from, lis.addrs...), events)
```
ralph-pichler, line 50: The fromBlock used to be ignored in the past (see https://github.com/ethereum/go-ethereum/issues/15063). Not sure if this was fixed, will have to check this.

zelig: if that is the case we may need to read the historical logs to catch up
https://goethereumbook.org/event-read/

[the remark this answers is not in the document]
zelig: otherwise we cannot do subscription since there is no push notiffications, so what to do?

acud:
closed in favor of https://github.com/ethersphere/bee/pull/1099

PR #922, opened 2020-11-07
postage batch store 

zelig:
deprecated by #1046. - based on AP #928 https://hackmd.io/o8RGsNwZSN6IE0lkBAZCPw

### DIFF
--- a/pkg/postage/batchstore/store.go
+++ b/pkg/postage/batchstore/store.go
@@ -1,0 +1,159 @@
+package batchstore
+
+import (
+	"math/big"
+	"sync"
+
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/postage"
+	"github.com/ethersphere/bee/pkg/storage"
+)
+
+var (
+	batchKeyPrefix = "batchKeyPrefix"
+	valueKeyPrefix = "valueKeyPrefix"
+)
+
+var _ postage.EventUpdater = (*Store)(nil)
+
+// Store is a local store  for postage batches
+type Store struct {
+	store  storage.StateStorer // state store backend to persist batches
+	mu     sync.Mutex          // mutex to lock statestore during atomic changes
+	cancel func()              // cancel sync and wait till done
+	state  *state              // the current state
+	logger logging.Logger
+}
+
+// New constructs a new postage batch store
+func New(store storage.StateStorer, events postage.Events, logger logging.Logger) (*Store, error) {
+	// initialise state from statestore or start with 0-s
+	st := &state{}
+	if err := st.load(store); err != nil {
+		return nil, err
+	}
+	s := &Store{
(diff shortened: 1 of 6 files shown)


pkg/postage/batchstore/store.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// settle retrieves the current state
+// - sets the cumulative outpayment normalised, cno+=price*period
+// - sets the new block number
+// caller holds the store mutex
+func (s *Store) settle(block uint64) {
+	period := int64(block - s.state.block)
+	s.state.block = block
+	s.state.total.Add(s.state.total, new(big.Int).Mul(s.state.price, big.NewInt(period)))
```
ralph-pichler, line 52: this reintroduces a lot of the reorg issues we wanted to get rid of. instead of doing this price update thing here we should refetch this value using the `totalOutPayment` getter when there is a `PriceUpdate` event on the oracle contract.

pkg/postage/batchstore/store.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	return s.put(b)
+}
+
+func (s *Store) UpdateDepth(id []byte, depth uint8) error {
+	b, err := s.get(id)
+	if err != nil {
+		return err
+	}
+	b.Depth = depth
```
ralph-pichler, line 106: what is the unit of `balance`? is this total amount paid into it, or per chunk (like the normalised balance). If the latter, changing the depth also changes the balance.

pkg/postage/batchstore/store.go
```diff
@@ -110,34 +70,28 @@ func (s *Store) replace(id []byte, oldValue, newValue *big.Int) error {
...
 	b, err := s.get(id)
 	if err != nil {
 		return err
 	}
-	value, err := s.balance(b, amount)
-	if err != nil {
-		return err
-	}
-	err = s.replace(id, b.Value, value)
+	value, err := s.balance(b, value)
```
zelig, line 94: value can cime durectly from the event]

pkg/postage/batchstore/store.go
```diff
@@ -110,34 +70,28 @@ func (s *Store) replace(id []byte, oldValue, newValue *big.Int) error {
...
 		Depth: depth,
 	}
-	value, err := s.balance(b, amount)
-	if err != nil {
-		return err
-	}
-	err = s.replace(id, b.Value, value)
+
+	panic("@zelig - should we have b.Add(value)?")
+	value, err := s.balance(b, value)
```
zelig, line 82: value can cime durectly from the event]

zelig: we need the relace call bothg in topup and increasedepth

zelig (commented):
- mutex should be used
- entrypoint for latest state
- entrypoint for Get by batch id

PR #923, opened 2020-11-08
Add file joiner function for iterating through chunk addresses

zbiljic:
Adds additional function to `file.Joiner` interface for iterating through chunk addresses for the file.

### DIFF
--- a/pkg/file/joiner/joiner.go
+++ b/pkg/file/joiner/joiner.go
@@ -210,7 +210,7 @@ func (j *joiner) Seek(offset int64, whence int) (int64, error) {
 
 }
 
-func (j *joiner) IterateChunkAddresses(fn file.AddressIterFunc) error {
+func (j *joiner) IterateChunkAddresses(fn swarm.AddressIterFunc) error {
 	// report root address
 	stop := fn(j.addr)
 	if stop {
@@ -228,7 +228,7 @@ func (j *joiner) IterateChunkAddresses(fn file.AddressIterFunc) error {
 	return nil
 }
 
-func (j *joiner) processChunkAddresses(fn file.AddressIterFunc, data []byte, subTrieSize int64, eg *errgroup.Group) {
+func (j *joiner) processChunkAddresses(fn swarm.AddressIterFunc, data []byte, subTrieSize int64, eg *errgroup.Group) {
 	// we are at a leaf data chunk
 	if subTrieSize <= int64(len(data)) {
 		return
--- a/pkg/file/joiner/joiner.go
+++ b/pkg/file/joiner/joiner.go
@@ -210,6 +210,60 @@ func (j *joiner) Seek(offset int64, whence int) (int64, error) {
 
 }
 
+func (j *joiner) IterateChunkAddresses(fn file.AddressIterFunc) error {
+	// report root address
+	stop := fn(j.addr)
+	if stop {
+		return nil
+	}
+
+	var eg errgroup.Group
+	j.processChunkAddresses(fn, j.rootData, j.span, &eg)
+
+	err := eg.Wait()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


pkg/file/joiner/joiner.go
```diff
@@ -210,6 +210,60 @@ func (j *joiner) Seek(offset int64, whence int) (int64, error) {
...
+
+	var eg errgroup.Group
+	j.processChunkAddresses(fn, j.rootData, j.span, &eg)
+
+	err := eg.Wait()
+	if err != nil {
+		return err
+	}
+
+	return nil
```
janos, line 228:
```suggestion
	return eg.Wait()
```

janos (approved):

acud (commented):
@zbiljic I am not sure that changes are needed for the actual joiner tbh. How about an approach which is the symmetric of the [following](https://github.com/ethersphere/beekeeper/blob/master/pkg/check/localpinning/localpinning.go#L19)? You could just inject a simple `storage.Storer` wrapper that logs the addresses (or calls a callback) as they get requested from the underlying store.
You could then just use `file.JoinReadAll` which just writes to `io.Discard`

zbiljic:
@acud The solution you presented sounds viable, however it does appear a bit wasteful. Not sure if it is the best idea to actually read all the content of the file, just to write into the air. Even if this new store does not write anywhere, the chunk content will be read from store, right?

I will see to create something like this and see how it looks with other changes.

acud:
Yes, the chunk will be read from the store, just as you are doing on `pkg/file/joiner/joiner.go#L248` [here](https://github.com/ethersphere/bee/pull/923/files#diff-0f05e0a55213856cc3c2846487aa0dcef1d956b541552814149254f2ca895149R248)

zbiljic:
Yes, you are correct, that was omission from my side. It should just have read the first chunks, that contain addresses to other chunks, and not read the chunks with end-data.

acud:
I see what you mean. It is true that it would traverse the entire trie. A possible benefit of this is that if you traverse all content locally, then this could potentially be used to signal that chunks are missing (I'm referring to the actual use case of re-uploading pinned content - this way you would know a chunk is missing before trying to re-upload it).

acud (approved):

pkg/file/joiner/joiner.go
```diff
@@ -210,6 +210,55 @@ func (j *joiner) Seek(offset int64, whence int) (int64, error) {
...
+		address := swarm.NewAddress(data[cursor : cursor+j.refLength])
+
+		stop := fn(address)
+		if stop {
+			break
+		}
+
+		func(address swarm.Address, eg *errgroup.Group) {
+			eg.Go(func() error {
+				ch, err := j.getter.Get(j.ctx, storage.ModeGetRequest, address)
```
acud, line 248: at a second look, since you first `Get` but only decide if to quit the recursive call when the subtrie size is `<= len(data)` then it means you have already retrieved the data chunk itself

zbiljic:
Created PR with suggested change here #932.
That one may replace this one.

acud:
Thanks @zbiljic. If you'd like to use this one under the assumption that data chunk should not be retrieved, then you would need to estimate the subtrie size before retrieving it, similarly to what is done in the joiner when retrieving (there's a helper function for that - `subtrieSection`)

zbiljic:
Closing in favor of: #932.

zbiljic:
Reopened because alternative solution is an issue, as `joiner.New` requests root chunk initially, and then multiple times.
Because there is external loop for each `swarm.ChunkSize` byte slice size, and then it goes down to the leaf node, always processing each parent chunk, and in doing so invoking `storer.Get` method, which will invoke function each time.

This means the larger the file, more duplicate calls for the same chunk will be invoked.

That is why this solution should be better, at least for the use-case it is indented for.

Anyway, I have added code that uses `subtrieSection` to prevent processing leaf chunks.

acud (approved):

PR #924, opened 2020-11-08
Chunk traversal

zbiljic:
Introduces package for traversing chunk addresses. It does so for bytes, files, and manifests.

It does make some (less-than-ideal) assumptions for detecting if a content is actually file or manifest.

Contains some additional optimization for checking various for types.

### DIFF
--- a/pkg/manifest/mantaray.go
+++ b/pkg/manifest/mantaray.go
@@ -113,6 +113,7 @@ func (m *mantarayManifest) HasPrefix(prefix string) (bool, error) {
 func (m *mantarayManifest) Store(ctx context.Context, mode storage.ModePut) (swarm.Address, error) {
 
 	saver := newMantaraySaver(ctx, m.encrypted, m.storer, mode)
+	m.loader = saver
 
 	err := m.trie.Save(saver)
 	if err != nil {
@@ -122,6 +123,53 @@ func (m *mantarayManifest) Store(ctx context.Context, mode storage.ModePut) (swa
 	address := swarm.NewAddress(m.trie.Reference())
 
 	return address, nil
+}
+
+func (m *mantarayManifest) IterateAddresses(ctx context.Context, fn swarm.AddressIterFunc) error {
+	reference := swarm.NewAddress(m.trie.Reference())
+
+	if swarm.ZeroAddress.Equal(reference) {
+		return ErrMissingReference
+	}
+
+	walker := func(path []byte, node *mantaray.Node, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if node != nil {
+			var stop bool
+
+			if node.Reference() != nil {
+				ref := swarm.NewAddress(node.Reference())
+
+				stop = fn(ref)
+				if stop {
+					return errStopIterator
+				}
+			}
+
+			if node.IsValueType() && node.Entry() != nil {
+				entry := swarm.NewAddress(node.Entry())
+				stop = fn(entry)
+				if stop {
+					return errStopIterator
+				}
+			}
+		}
+
+		return nil
+	}
+
+	err := m.trie.WalkNode([]byte{}, m.loader, walker)
+	if err != nil {
+		if !errors.Is(err, errStopIterator) {
+			return fmt.Errorf("manifest iterate addresses: %w", err)
+		}
+		// ignore error if interation stopped by caller
+	}
+
+	return nil
 }
 
 // mantarayLoadSaver implements required interface 'mantaray.LoadSaver'
--- a/pkg/manifest/mantaray.go
+++ b/pkg/manifest/mantaray.go
@@ -45,6 +45,24 @@ func NewMantarayManifest(
 	}, nil
 }
 
+// NewMantarayManifestWithObfuscationKeyFn creates a new mantaray-based manifest
+// with configured obfuscation key
+//
+// NOTE: This should only be used in tests.
+func NewMantarayManifestWithObfuscationKeyFn(
+	encrypted bool,
+	storer storage.Storer,
+	obfuscationKeyFn func([]byte) (int, error),
+) (Interface, error) {
+	mm := &mantarayManifest{
+		trie:      mantaray.New(),
+		encrypted: encrypted,
+		storer:    storer,
+	}
+	mantaray.SetObfuscationKeyFn(obfuscationKeyFn)
+	return mm, nil
+}
+
 // NewMantarayManifestReference loads existing mantaray-based manifest.
 func NewMantarayManifestReference(
 	ctx context.Context,
--- a/pkg/manifest/simple.go
+++ b/pkg/manifest/simple.go
@@ -27,6 +27,7 @@ const (
 type simpleManifest struct {
 	manifest simple.Manifest
 
+	reference swarm.Address
 	encrypted bool
 	storer    storage.Storer
 }
@@ -52,6 +53,7 @@ func NewSimpleManifestReference(
 ) (Interface, error) {
 	m := &simpleManifest{
 		manifest:  simple.NewManifest(),
+		reference: reference,
 		encrypted: encrypted,
 		storer:    storer,
 	}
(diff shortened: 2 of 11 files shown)


janos (approved):
👍 and also, good test cases.

janos (approved):

pkg/manifest/mantaray.go
```diff
@@ -124,6 +143,53 @@ func (m *mantarayManifest) Store(ctx context.Context, mode storage.ModePut) (swa
...
+				if stop {
+					return errStopIterator
+				}
+			}
+
+			if node.IsValueType() && node.Entry() != nil {
+				entry := swarm.NewAddress(node.Entry())
+				stop = fn(entry)
+				if stop {
+					return errStopIterator
```
zelig, line 174: just  return nil here and then just

     return  m.trie.WalkNode([]byte{}, m.loader, walker)

no?

pkg/manifest/simple.go
```diff
@@ -116,9 +118,51 @@ func (m *simpleManifest) Store(ctx context.Context, mode storage.ModePut) (swarm
...
+		}
+
+		ref, err := swarm.ParseHexAddress(entry.Reference())
+		if err != nil {
+			return err
+		}
+
+		stop := fn(ref)
+		if stop {
+			return errStopIterator
```
zelig, line 149: same here  no?

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,293 @@
...
+
+	bytesReference := e.Reference()
+
+	err = s.processBytes(ctx, bytesReference, chunkAddressFunc)
+	if err != nil {
+		// possible it was custom JSON bytes, which matches entry JSON
+		// but in fact is not file, and does not contain reference to
+		// existing address, which is why it was not found in storage
+		if !errors.Is(err, storage.ErrNotFound) {
+			return nil
```
zelig, line 156: return err no??

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,293 @@
...
+	err = j.IterateChunkAddresses(func(addr swarm.Address) (stop bool) {
+		stop = chunkAddressFunc(addr)
+		return
+	})
+	if err != nil {
+		return fmt.Errorf("traversal: iterate chunks: %s: %w", reference, err)
+	}
+
+	return nil
+}
```
zelig, line 293:
```suggestion
	return j.IterateChunkAddresses(chunkAddressFunc)
}
```

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,293 @@
...
+	if err != nil {
+		return
+	}
+
+	if stop := chunkAddressFunc(reference); stop {
+		return nil
+	}
+
+	return nil
+}
```
zelig, line 173:
```suggestion
	     _  = chunkAddressFunc(reference) 
    return nil
}
```

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,293 @@
...
+		}
+	} else {
+		err = s.processBytes(ctx, reference, chunkAddressFunc)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
zelig, line 139:
```suggestion
   if isFile {
      return s.traverseChunkAddressesAsFile(ctx, reference, chunkAddressFunc, e)
  }
 return  s.processBytes(ctx, reference, chunkAddressFunc)
}
```

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,293 @@
...
+			})
+			if err != nil {
+				return fmt.Errorf("traversal: iterate chunks: %s: %w", reference, err)
+			}
+
+			metadataReference := e.Metadata()
+
+			err = s.processBytes(ctx, metadataReference, chunkAddressFunc)
+			if err != nil {
+				return nil
```
zelig, line 90: not return err?

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,293 @@
...
+			}
+
+			if stop := chunkAddressFunc(reference); stop {
```
zelig, line 93: it seems not to matter if stop is true you return nil anyways

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,293 @@
...
+
+	} else {
+		err = s.processBytes(ctx, reference, chunkAddressFunc)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
zelig, line 112:
```suggestion
		} 
		return s.traverseChunkAddressesAsFile(ctx, reference, chunkAddressFunc, e)
	} 
	return s.processBytes(ctx, reference, chunkAddressFunc)
}
```

zelig (changes requested):
wow how come this complexity ? It ought to be much  simpler

test look very good

pkg/manifest/mantaray.go
zbiljic: The reason for this flow was because this code is using some other library for iteration (i.e. `manifest` project).
Implementation(s) in the `manifest` project for "walker" function were modeled according to Go `filepaht.WalkFunc`, which does not have some easy way to "stop" iteration (as far as I remember). Only way is to return error from (this) implemented "walker" function, which will be propagated to code here. Because of this, internal error is retuned to stop iteration, which is ignored later on.

This is in contrast to the `swarm.AddressIterFunc`, which is more modeled in the line with some DB iteration functions.

pkg/traversal/traversal.go
zbiljic: This was done in order to cover as much as possible edge-cases for requirements here.

As noted in use case issue (and discussed), one would have some address, whose chunks should be discovered, and unpinned. Since we have multiple types in Swarm, there is no exact way to know what type is actually (bytes, files and collections). For this reason, the code basically tries to "detect" what type of address it has, and then find all chunks for it.
This concrete part tries to cover case where someone used `/bytes` (or `/chunks`) API to upload JSON which is compatible with `pkg/collection/entry.Metadata`, and then used that address with another address to create new chunk, which unmarshals to `pkg/collection/entry.Entry`, but there reference used for file does not actually exists.
In that case, since the content in that reference address does not exist, for that side there are no chunks, so this is covers that.

I figure all of that might be far-fetched case, but if I am not mistaken with that, this should be fine.

Of course, another possibility is that somehow the bytes content for the file was removed, but not the metadata or root entry chunk (which could be possible), in which case maybe error should be returned?
Or maybe errors on other places should be checked for this same condition and ignored?

pkg/traversal/traversal.go
zbiljic: Yeah, I had several iterations of various stuff, and changed orders on some places, which is why I had some leftovers. Thanks for comments.

zbiljic:
Added new functions, that are for specific internal type. If specific function is called with wrong reference, error will be thrown.

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,364 @@
...
+var (
+	// ErrInvalidType is returned when the reference was not expected type.
+	ErrInvalidType = errors.New("traversal: invalid type")
+)
+
+// Service is the service to find dependent chunks for an address.
+type Service interface {
+	// TraverseChunkAddresses iterates through each address related to
+	// the supplied one, if possible.
+	TraverseChunkAddresses(context.Context, swarm.Address, swarm.AddressIterFunc) error
```
janos, line 31: TraverseChunkAddresses name may relate to a single chunk traversal, even if it does not make sense to have that. I would consider naming it without a Chunk word, but have no nice example. Maybe just TraverseAddresses?

janos (approved):
Thanks @zbiljic.

zelig (approved):
ok lets merge it, bbut. affter. release lets address technical debt:

- all simplified, some parts move to mantaray
- adapt to context passed in loadsaver
- no stop condition, but errgroup with context
-

PR #932, opened 2020-11-09
Add chunk addresses getter iterator

zbiljic:
Adds new `storage.Getter` implementation that can be used to iterate through chunk addresses for the file.


acud (approved):
Very cool. Thanks

janos (approved):

PR #934, opened 2020-11-10
postage/events: pkg for postage events and syncing with events

zelig:
this incomplete PR implements the go side of postage events and syncing with these events as per #920
see AP https://hackmd.io/o8RGsNwZSN6IE0lkBAZCPw?both

known to be missing: 
- tests (unclear to me how to mock `types.Log` events without ethclient calls on actual contracts.
- handling of reorgs `types.Log.Removed` field and/or 'n-block confirmation delay'

### DIFF
--- a/pkg/postage/events/events.go
+++ b/pkg/postage/events/events.go
@@ -1,0 +1,161 @@
+package events
+
+import (
+	"encoding/hex"
+	"fmt"
+	"math/big"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethersphere/bee/pkg/crypto"
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/postage"
+)
+
+var (
+	contractAddress      = common.HexToAddress("0x147B8eb97fD247D06C4006D269c90C1908Fb5D54")
+	contractABI          = parseABI("{}")
+	contractEventDigests = digestSig(eventSignatures)
+)
+
+var (
+	eventSignatures = []string{
+		"BatchCreated(bytes32,uint256,address,uint256)",
+		"BatchTopUp(bytes32,uint256)",
+		"BatchDepthIncrease(bytes32,uint256",
+		"PriceUpdate(uint256)",
+	}
+	eventNames = []string{
+		"BatchCreated",
+		"BatchTopUp",
+		"BatchDepthIncrease",
+		"PriceUpdate",
+	}
+)
+
+func eventTypes(i int) postage.Event {
+	switch i {
+	case 0:
+		return &batchCreatedEvent{}
+	case 1:
+		return &batchTopUpEvent{}
+	case 2:
+		return &batchDepthIncreaseEvent{}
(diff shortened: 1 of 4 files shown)


pkg/postage/events/events.go
```diff
@@ -0,0 +1,161 @@
...
+
+var (
+	contractAddress      = common.HexToAddress("0x147B8eb97fD247D06C4006D269c90C1908Fb5D54")
+	contractABI          = parseABI("{}")
+	contractEventDigests = digestSig(eventSignatures)
+)
+
+var (
+	eventSignatures = []string{
+		"BatchCreated(bytes32,uint256,address,uint256)",
```
ralph-pichler, line 25: depth is now `uint8`

ralph-pichler, line 25: `depth` is now `uint8`

pkg/postage/events/events.go
```diff
@@ -0,0 +1,161 @@
...
+		"BatchTopUp(bytes32,uint256)",
+		"BatchDepthIncrease(bytes32,uint256",
```
ralph-pichler, line 27: same here about `depth`. also missing `)`

pkg/postage/events/events.go
```diff
@@ -0,0 +1,161 @@
...
+	price *big.Int
+}
+
+func (e *priceUpdateEvent) Update(s postage.EventUpdater) error {
+	return s.UpdatePrice(e.price)
+}
+
+// parse reifies the event log type as a struct
+//  weakly unsafe in that if there is another event, nil is returned
+func parse(log types.Log) postage.Event {
```
ralph-pichler, line 97: this approach does not work as some event parameters are indexed (only appears as topics, not events). see this for reference: https://github.com/ethereum/go-ethereum/blob/master/accounts/abi/bind/base.go#L350.

ralph-pichler, line 97: at the very least we must check for `Removed` here. Otherwise not only are we not going to catch the reorganisation, we'll apply the removed event twice.

ralph-pichler (commented):
Error

ralph-pichler (commented):
Regarding the tests you can just create instances of `types.Log` with the right topics (like here https://github.com/ethersphere/bee/pull/934, except with the real topics if you want to test the parse function).

Reorg handling will be tricky. Functions here which take the block number should at least also take the block hash as the same block number later might refer to a different block but block hashes never change. 
Another important point to consider is that if the socket connection to the client breaks and we open a new subscription from the last block we processed we will not get `Removed` events even in case of reorgs of events still received on the old subscription. This will lead to an inconsistent batch database. The same applies on bee restarts for blocks which were already processed but reorganised during the down time.

ralph-pichler (commented):
The most solid approach to the event tracking would be to store the block hash for every block number whenever we receive a (non-removed) event through the subscription. Then on startup and whenever the subscription is remade we can check if we still agree with the node on the last processed event (by checking the block hash which also includes the receipt hash in its computation). Otherwise we can walk back through the events and find the first event that is no longer valid and revert all of the ones following it. Then we can subscribe again from the block after the last agreed event.

zelig:
@ralph-pichler    could you take over? ;)

acud:
closed in favor of https://github.com/ethersphere/bee/pull/1099

Issue #873, opened 2020-10-27
upload with postage stamps: hashing pipeline changes


acud:
As per the [architectural proposal](https://hackmd.io/rYze0NZ5Ryi2HNmc62ogWA?view) that deals with uploading with postage stamps, the following changes are needed in the `pipeline` package:
- a postage stamp writer needs to be created in accordance with the spec

related to #654

PR #939, opened 2020-11-11
api, manifest: abstract saver/loader functionality

acud:
this is needed so we don't need to pass `crypto.Signer` and batch IDs to the manifest package as part of the storage incentives effort

needed for #873

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -248,11 +248,11 @@ func equalASCIIFold(s, t string) bool {
 	return s == t
 }
 
-type pipelineFunc func(io.Reader, int64) (swarm.Address, error)
+type pipelineFunc func(context.Context, io.Reader, int64) (swarm.Address, error)
 
-func requestPipelineFn(ctx context.Context, s storage.Storer, r *http.Request) pipelineFunc {
+func requestPipelineFn(s storage.Storer, r *http.Request) pipelineFunc {
 	mode, encrypt := requestModePut(r), requestEncrypt(r)
-	return func(r io.Reader, l int64) (swarm.Address, error) {
+	return func(ctx context.Context, r io.Reader, l int64) (swarm.Address, error) {
 		pipe := builder.NewPipelineBuilder(ctx, s, mode, encrypt)
 		return builder.FeedPipeline(ctx, pipe, r, l)
 	}
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -5,6 +5,7 @@
 package api
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -15,6 +16,7 @@ import (
 	"time"
 	"unicode/utf8"
 
+	"github.com/ethersphere/bee/pkg/file/pipeline/builder"
 	"github.com/ethersphere/bee/pkg/logging"
 	m "github.com/ethersphere/bee/pkg/metrics"
 	"github.com/ethersphere/bee/pkg/pss"
@@ -195,6 +197,13 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
 	}
 }
 
+func lookaheadBufferSize(size int64) int {
+	if size <= largeBufferFilesizeThreshold {
+		return smallFileBufferSize
+	}
+	return largeFileBufferSize
+}
+
 // checkOrigin returns true if the origin is not set or is equal to the request host.
 func (s *server) checkOrigin(r *http.Request) bool {
 	origin := r.Header["Origin"]
@@ -213,13 +222,6 @@ func (s *server) checkOrigin(r *http.Request) bool {
 	}
 
 	return false
-}
-
-func lookaheadBufferSize(size int64) int {
-	if size <= largeBufferFilesizeThreshold {
-		return smallFileBufferSize
-	}
-	return largeFileBufferSize
 }
 
 // equalASCIIFold returns true if s is equal to t with ASCII case folding as
@@ -244,4 +246,14 @@ func equalASCIIFold(s, t string) bool {
 		}
 	}
 	return s == t
+}
+
+type pipelineFunc func(io.Reader, int64) (swarm.Address, error)
+
+func requestPipelineFn(ctx context.Context, s storage.Storer, r *http.Request) pipelineFunc {
+	mode, encrypt := requestModePut(r), requestEncrypt(r)
+	return func(r io.Reader, l int64) (swarm.Address, error) {
+		pipe := builder.NewPipelineBuilder(ctx, s, mode, encrypt)
+		return builder.FeedPipeline(ctx, pipe, r, l)
+	}
 }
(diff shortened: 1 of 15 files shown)


pkg/file/io.go
```diff
@@ -107,3 +107,18 @@ func SplitWriteAll(ctx context.Context, s Splitter, r io.Reader, l int64, toEncr
 	}
 	return addr, nil
 }
+
+type Loader interface {
+	// Load a reference in byte slice representation and return all content associated with the reference.
+	Load([]byte) ([]byte, error)
```
zbiljic, line 113: Since this is in Bee project, I think that `Load`/`Save` functions should return `swarm.Address` directly.

And maybe allow passing context to both operations.

zbiljic (changes requested):

zelig (changes requested):
lets revisit this when this is merged https://github.com/ethersphere/manifest/pull/16

pkg/file/io.go
acud: this interface was made to be compatible with `manifest.Loader` in the manifest package. have a closer look that i am injecting this interface directly into the mantaray dependency on load and save. if you want it as an address then we need to change the interface in mantaray first

janos:
Compilation errors should be fixed https://github.com/ethersphere/bee/pull/939/checks?check_run_id=1390345361#step:7:9.

acud:
@janos errors are intentional. waiting for viktor's PR to be merged to the manifest project

pkg/file/io.go
zbiljic: I am fully aware this was the reason. My comment is that we should not force external interface into this project.

The interface in "manatary" project was created to be independent from some external project. That interface is how it is, and should not be changed.
I would rather we have another level of indirection instead of having something like this. Especially since some other parts of Bee project might want to use this package.

pkg/file/io.go
```diff
@@ -107,3 +107,18 @@ func SplitWriteAll(ctx context.Context, s Splitter, r io.Reader, l int64, toEncr
 	}
 	return addr, nil
 }
+
```
zelig, line 110: these interfaces are defined twice. we should use only one and have it in `storage` package maybe

pkg/traversal/traversal_test.go
```diff
@@ -464,12 +465,12 @@ func TestTraversalManifest(t *testing.T) {
 
 			switch tc.manifestType {
 			case manifest.ManifestSimpleContentType:
-				dirManifest, err = manifest.NewSimpleManifest(false, mockStorer)
+				dirManifest, err = manifest.NewSimpleManifest(loadsave.New(mockStorer, storage.ModePutRequest, false))
```
zelig, line 468: define once before switch

pkg/api/api.go
```diff
@@ -245,3 +247,13 @@ func equalASCIIFold(s, t string) bool {
 	}
 	return s == t
 }
+
+type pipelineFunc func(io.Reader, int64) (swarm.Address, error)
+
+func requestPipelineFn(ctx context.Context, s storage.Storer, r *http.Request) pipelineFunc {
+	mode, encrypt := requestModePut(r), requestEncrypt(r)
+	return func(r io.Reader, l int64) (swarm.Address, error) {
```
zelig, line 255: i. think we should redo the pipelinefunc to take the context, not the constructors

pkg/api/api.go
```diff
@@ -245,3 +247,13 @@ func equalASCIIFold(s, t string) bool {
 	}
 	return s == t
 }
+
+type pipelineFunc func(io.Reader, int64) (swarm.Address, error)
```
zelig, line 251: i. think we should redo the pipelinefunc to take the context, not the constructors

pkg/api/dirs.go
```diff
@@ -146,7 +147,7 @@ func storeDir(ctx context.Context, reader io.ReadCloser, s storage.Storer, mode
 			contentType: contentType,
 			reader:      tarReader,
 		}
-		fileReference, err := storeFile(ctx, fileInfo, s, mode, encrypt)
+		fileReference, err := storeFile(fileInfo, p)
```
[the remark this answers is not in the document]
zelig: this must be parallel to storeDir so that it takes context, no?

pkg/api/dirs.go
```diff
@@ -220,10 +219,9 @@ func storeDir(ctx context.Context, reader io.ReadCloser, s storage.Storer, mode
 
 // storeFile uploads the given file and returns its reference
 // this function was extracted from `fileUploadHandler` and should eventually replace its current code
-func storeFile(ctx context.Context, fileInfo *fileUploadInfo, s storage.Storer, mode storage.ModePut, encrypt bool) (swarm.Address, error) {
+func storeFile(fileInfo *fileUploadInfo, p pipelineFunc) (swarm.Address, error) {
```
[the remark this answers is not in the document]
zelig: why would the context be linked to manifests only?

pkg/api/file.go
```diff
@@ -152,9 +150,10 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request) {
 		reader = tmp
 	}
 
+	p := requestPipelineFn(ctx, s.Storer, r)
```
zelig, line 153: i think the  context should be passed directly to `p`

pkg/file/io.go
acud: these interfaces provide functionalities which are wrapping components in the `file` package. that's why I put them here. it belongs here more than in `storage`, which relates more to `localstore` and of such

zelig (approved):
LGTM

pkg/file/io.go
```diff
@@ -107,3 +107,18 @@ func SplitWriteAll(ctx context.Context, s Splitter, r io.Reader, l int64, toEncr
...
+type Loader interface {
+	// Load a reference in byte slice representation and return all content associated with the reference.
+	Load(context.Context, []byte) ([]byte, error)
```
zbiljic, line 113: I think since we are in the `bee` project, the `Load` and `Save` function should work directly with `swarm.Address`, instead of using `[]byte`.

This may require having some proxy implementation for manifest, where we have implementation with byte slice, which translates to `swarm.Address`.

pkg/manifest/simple.go
```diff
@@ -28,51 +24,41 @@ type simpleManifest struct {
...
-	encrypted bool
-	storer    storage.Storer
+	ls        file.LoadSaver
 }
 
 // NewSimpleManifest creates a new simple manifest.
 func NewSimpleManifest(
-	encrypted bool,
-	storer storage.Storer,
+	ls file.LoadSaver,
```
zbiljic, line 32: Function signature may be collapsed now (it should be enough shorter).

zbiljic (changes requested):
Only things I think should be changed is `LoadSaver` function signatures, to use `swarm.Address`. That way package may be usable for other things in the future.

pkg/file/io.go
acud: but then it means I need an extra abstraction that would use this one. I'm not sure why this is necessary. I only created this implementation to be compatible with the manta ray interfaces. I think that once other components need to use similar functionality we can generalize, but I'm not sure why to do so prematurely

zbiljic: Yeah, I just do not think application should cater to library completely.

zbiljic (approved):

PR #941, opened 2020-11-12
Bee compatibility as windows service

metacertain:
Packaging .msi raised a technical limit that our executable can't run as service under windows 10.
These iterations are experimental, attempting to alleviate the issue


vandot:
@metacertain problem with implementation is that now binary is always executed as a service
try `bee version`
it should be started as a service only with `start` command

PR #942, opened 2020-11-12
validators/callbacks reorg to enable pss mailboxing

zelig:
- api, pss: pss.TryUnwrap  takes no context
- netstore: has no validator, (FIXME: chunk API needs validation)
   recovery simplified
- recovery: renames, callback is async and results in no error
- pullsync: add validator with callback (FIXME: always called)
- pushsync: add validator with callbakck ,
   called async, called only if peer is closest node, tested
- swarm: validator.go, multiValidator implements ValidatorWithCallback

### DIFF
--- a/pkg/pss/pss.go
+++ b/pkg/pss/pss.go
@@ -31,7 +31,7 @@ type Interface interface {
 	// Register a Handler for a given Topic.
 	Register(Topic, Handler) func()
 	// TryUnwrap tries to unwrap a wrapped trojan message.
-	TryUnwrap(context.Context, swarm.Chunk)
+	TryUnwrap(swarm.Chunk)
 
 	SetPushSyncer(pushSyncer pushsync.PushSyncer)
 	io.Closer
@@ -128,10 +128,11 @@ func (p *pss) topics() []Topic {
 }
 
 // TryUnwrap allows unwrapping a chunk as a trojan message and calling its handlers based on the topic.
-func (p *pss) TryUnwrap(ctx context.Context, c swarm.Chunk) {
+func (p *pss) TryUnwrap(c swarm.Chunk) {
 	if len(c.Data()) < swarm.ChunkWithSpanSize {
 		return // chunk not full
 	}
+	ctx := context.Background()
 	topic, msg, err := Unwrap(ctx, p.key, c, p.topics())
 	if err != nil {
 		return // cannot unwrap
--- a/pkg/pullsync/pullsync.go
+++ b/pkg/pullsync/pullsync.go
@@ -18,7 +18,6 @@ import (
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/p2p"
 	"github.com/ethersphere/bee/pkg/p2p/protobuf"
-
 	"github.com/ethersphere/bee/pkg/pullsync/pb"
 	"github.com/ethersphere/bee/pkg/pullsync/pullstorage"
 	"github.com/ethersphere/bee/pkg/storage"
@@ -57,12 +56,13 @@ type Interface interface {
 }
 
 type Syncer struct {
-	streamer p2p.Streamer
-	metrics  metrics
-	logger   logging.Logger
-	storage  pullstorage.Storer
-	quit     chan struct{}
-	wg       sync.WaitGroup
+	streamer  p2p.Streamer
+	metrics   metrics
+	logger    logging.Logger
+	storage   pullstorage.Storer
+	quit      chan struct{}
+	wg        sync.WaitGroup
+	validator swarm.ValidatorWithCallback
 
 	ruidMtx sync.Mutex
 	ruidCtx map[uint32]func()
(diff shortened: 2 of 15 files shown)


pkg/pss/pss.go
```diff
@@ -128,10 +128,11 @@ func (p *pss) topics() []Topic {
 }
 
 // TryUnwrap allows unwrapping a chunk as a trojan message and calling its handlers based on the topic.
-func (p *pss) TryUnwrap(ctx context.Context, c swarm.Chunk) {
+func (p *pss) TryUnwrap(c swarm.Chunk) {
 	if len(c.Data()) < swarm.ChunkWithSpanSize {
 		return // chunk not full
 	}
+	ctx := context.Background()
 	topic, msg, err := Unwrap(ctx, p.key, c, p.topics())
```
acud, line 136: i dont understand why we need a context here. consider removing it from `Unwrap` altogether.

pkg/recovery/repair.go
```diff
@@ -16,27 +16,26 @@ import (
 )
 
 const (
-	// RecoveryTopicText is the string used to construct the recovery topic.
-	RecoveryTopicText = "RECOVERY"
+	// TopicText is the string used to construct the recovery topic.
+	TopicText = "RECOVERY"
```
acud, line 20: does this need to be exported?

pkg/recovery/repair.go
```diff
@@ -16,27 +16,26 @@ import (
...
 )
 
 var (
-	// RecoveryTopic is the topic used for repairing globally pinned chunks.
-	RecoveryTopic = pss.NewTopic(RecoveryTopicText)
+	// Topic is the topic used for repairing globally pinned chunks.
+	Topic = pss.NewTopic(TopicText)
```
acud, line 25: ditto

pkg/swarm/validator.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+	}
+
+	return false
+}
+func (mv *multiValidator) ValidWithCallback(ch Chunk) (bool, func()) {
+	for i, v := range mv.validators {
+		if v.Validate(ch) {
+			if i < len(mv.callbacks) {
+				return true, func() { mv.callbacks[i](ch) }
```
acud, line 55: not sure why to return the callback instead of directly invoking it in a separate goroutine.
also, i thought we would use the same validator defined with callbacks or without callbacks to fulfill both validation with and without side effects. not sure why we need `ValidatorWithCallback`.

pkg/pullsync/pullsync.go
```diff
@@ -212,9 +213,18 @@ func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8
...
+
+		chunk := swarm.NewChunk(addr, delivery.Data)
+		valid, callback := s.validator.ValidWithCallback(chunk)
+		if !valid {
+			return 0, ru.Ruid, swarm.ErrInvalidChunk
+		}
+		if err = s.storage.Put(ctx, storage.ModePutSync, chunk); err != nil {
 			return 0, ru.Ruid, fmt.Errorf("delivery put: %w", err)
 		}
+		if callback != nil {
```
acud, line 225: i would expect the validator to do this

pkg/pullsync/pullsync_test.go
```diff
@@ -217,6 +217,20 @@ func haveChunks(t *testing.T, s *mock.PullStorage, addrs ...swarm.Address) {
...
+type mockValidator struct {
+	c chan swarm.Chunk
+}
+
+func (*mockValidator) Validate(swarm.Chunk) bool {
+	return true
+}
+
+func (mv *mockValidator) ValidWithCallback(c swarm.Chunk) (bool, func()) {
+	return true, func() { mv.c <- c }
```
acud, line 235: is this channel used somehow? not sure why not to return an empty function here

acud (changes requested):

acud:
@zelig seeing the changes in the pushsync (saving the callback to the point where you need to process and call it only later in a different code branch) makes me feel this is a bit sinister concept that diverges from a concept of a _validator_, which I would assume has no side effects in general. If we choose to have it with side effects then I would expect it to execute the callback exactly when it is assumed that the data is valid and the callback for it exists. but somehow returning the callback for later processing seems a bit like this concept is a bit skewed.

That being said, nitpicking on this is a waste of time and I won't insist to change this. I see this merely as moving coins from one pocket to another.

pkg/pss/pss.go
janos: As context is removed from TryUnwrap, it should be removed from Unwrap also, as it serves no purpose now. No cancelation is possible. I would argue that cancelation has its value here, but not so significant. I think that context can be removed from Unwrap.

pkg/recovery/repair.go
```diff
@@ -16,27 +16,26 @@ import (
...
+func NewCallback(pssSender pss.Sender) Callback {
+	privk := crypto.Secp256k1PrivateKeyFromBytes([]byte(TopicText))
 	recipient := privk.PublicKey
-	return func(chunkAddress swarm.Address, targets pss.Targets) error {
+	return func(chunkAddress swarm.Address, targets pss.Targets) {
 		payload := chunkAddress
 		ctx := context.Background()
-		err := pssSender.Send(ctx, RecoveryTopic, payload.Bytes(), &recipient, targets)
-		return err
+		_ = pssSender.Send(ctx, Topic, payload.Bytes(), &recipient, targets)
```
janos, line 38: This error was logged by the callback caller. Should we log it here?

pkg/swarm/validator.go
zelig: in pushsync it is clear why the callback needs to be returned and not called within the validator:
because there are additional conditions for calling it, namely that the node is the closest one to the chunk.

pkg/pullsync/pullsync_test.go
zelig: so that we test that it was called just like in your code https://github.com/ethersphere/bee/blob/master/pkg/pushsync/pushsync_test.go#L234

acud: i see you are passing this mock in the tests, but i can't see where do you actually assert that the call was made. this infers changes to the actual tests (which was not done). also, that is not my code :)

acud (approved):
approving this with regards to https://github.com/ethersphere/bee/issues/974

janos (approved):

Eknir:
@zelig , @acud . This one is done reviewing. Please, next time drag the issue to the good pipeline.

PR #944, opened 2020-11-16
add init command

ralph-pichler:
add a `bee init` command which only does key generation.

### DIFF
--- a/cmd/bee/cmd/init.go
+++ b/cmd/bee/cmd/init.go
@@ -1,0 +1,58 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func (c *command) initInitCmd() (err error) {
+	cmd := &cobra.Command{
+		Use:   "init",
+		Short: "Initialise a Swarm node",
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			if len(args) > 0 {
+				return cmd.Help()
+			}
+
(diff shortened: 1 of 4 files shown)


cmd/bee/cmd/init.go
```diff
@@ -0,0 +1,58 @@
...
+				logger = logging.New(cmd.OutOrStdout(), logrus.TraceLevel)
+			default:
+				return fmt.Errorf("unknown verbosity level %q", v)
+			}
+
+			_, err = c.configureSigner(cmd, logger)
+			if err != nil {
+				return err
+			}
+			return nil
```
janos, line 48:
```suggestion
			return err
```

cmd/bee/cmd/start.go
```diff
@@ -33,6 +33,131 @@ import (
...
+type signerConfig struct {
+	keystore         keystore.Service
+	signer           crypto.Signer
+	address          swarm.Address
+	publicKey        *ecdsa.PublicKey
+	libp2pPrivateKey *ecdsa.PrivateKey
+	password         string
+}
+
+func (c *command) configureSigner(cmd *cobra.Command, logger logging.Logger) (config *signerConfig, err error) {
```
janos, line 45: I would put this to the end of this file or even in init.go file, but it is a very minor thing, nothing crucial, just a suggestion.

janos (approved):

vandot:
When you execute `bee init` it prints log lines (with timestamps and log lvl to stdout)
@Eknirwhat do you think?

Eknir:
> When you execute `bee init` it prints log lines (with timestamps and log lvl to stdout)
> @Eknir what do you think?

Just went over it together with you. From a product perspective, it seems to behave as it should.

acud (approved):

cmd/bee/cmd/init.go
ralph-pichler: fixed

cmd/bee/cmd/start.go
ralph-pichler: I just put it at the end of start.go

Issue #892, opened 2020-11-01
avoid disconnects if peer does not verify cheque fast enough


ralph-pichler:
# Task

Currently it is possible to get blocklisted by a peer after sending a cheque because we already used more bandwidth before the peers has finished processing the cheque (which involves a liquidity check against the chain). This is because we don't wait for a full close anymore when sending a cheque. This was removed recently as it lead to a deadlock situation if two nodes were sending cheques to each other at the same time.

possible solutions:
* reintroduce fullclose. avoid deadlock by updating balance in a go routine
* apply balance change immediately. revert later if cheque was not ok
* remove chain-based liquidity check entirely (does not solve this fully, as we cannot avoid the factory check)
* introduce a tolerance of 50 refused requests before disconnecting someone for overdrawing their payment threshold, reseted upon settlement 

## Acceptance criterea

We don't get disconnected if we send a cheque to a peer and consume bandwidth afterwards while avoiding the deadlock issue we had previously.

PR #953, opened 2020-11-17
call notifypayment in go routine in swap and restore full close

ralph-pichler:
restores waiting for full close on sending cheques which solves problems on the real network when nodes use external blockchain nodes. to avoid reintroducing the deadlock fixed in #860, swap now calls notifypayment in a go routine. Furthermore to simplify the code this PR removes the payment observer and just passes the notify func to settlement directly.

solution 1 from #892

### DIFF
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -455,8 +455,8 @@ func (a *Accounting) NotifyPayment(peer swarm.Address, amount uint64) error {
 	return nil
 }
 
-// AsyncNotifyPayment calls notify payment in a go routine
-// This is needed when accounting needs to be notified but the accounting lock is already held
+// AsyncNotifyPayment calls notify payment in a go routine.
+// This is needed when accounting needs to be notified but the accounting lock is already held.
 func (a *Accounting) AsyncNotifyPayment(peer swarm.Address, amount uint64) error {
 	go func() {
 		err := a.NotifyPayment(peer, amount)
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -416,8 +416,7 @@ func balanceKeyPeer(key []byte) (swarm.Address, error) {
 	return addr, nil
 }
 
-// NotifyPayment implements the PaymentObserver interface. It is called by
-// Settlement when we receive a payment.
+// NotifyPayment is called by Settlement when we receive a payment.
 func (a *Accounting) NotifyPayment(peer swarm.Address, amount uint64) error {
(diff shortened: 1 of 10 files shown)


pkg/accounting/accounting.go
```diff
@@ -456,6 +455,18 @@ func (a *Accounting) NotifyPayment(peer swarm.Address, amount uint64) error {
 	return nil
 }
 
+// AsyncNotifyPayment calls notify payment in a go routine
```
acud, line 458: add period end of line

pkg/accounting/accounting.go
```diff
@@ -456,6 +455,18 @@ func (a *Accounting) NotifyPayment(peer swarm.Address, amount uint64) error {
...
+// This is needed when accounting needs to be notified but the accounting lock is already held
```
acud, line 459: here too

pkg/settlement/swap/swapprotocol/swapprotocol.go
```diff
@@ -186,8 +186,7 @@ func (s *Service) EmitCheque(ctx context.Context, peer swarm.Address, cheque *ch
 		if err != nil {
 			_ = stream.Reset()
 		} else {
-			// don't wait for full close to avoid deadlocks if cheques are sent simultaneously in both directions
-			go stream.FullClose()
```
acud, line 190: why remove the goroutine

pkg/accounting/accounting.go
ralph-pichler: fixed

pkg/accounting/accounting.go
ralph-pichler: fixed

pkg/settlement/swap/swapprotocol/swapprotocol.go
ralph-pichler: because we want to wait for the stream to be closed. this gives us at least some basic assurance that the counterparty has had time to process the cheque. in the meantime settlement is still holding that peers accounting lock, therefore preventing overspending.

Eknir:
Hi @ralph-pichler , are we planning to merge this, or do we do this after the release?

ralph-pichler:
@Eknir at the very least we need to merge the sync full close again, otherwise this will cause trouble in production (where most ppl will use external eth nodes).

Eknir:
I would suggest we proceed to merge this before todays testing session.
An interactive review session on this PR would be good.

acud (approved):

janos (approved):

PR #957, opened 2020-11-17
Add API for changing pin counter

zbiljic:
This change adds additional HTTP route, which allows caller to change pin counter of some chunk.

This can be done through `PUT` call on `/pin/chunks/{address}` endpoint, with the following JSON request:

```json
{
  "pinCounter": 0
}
```

If the request is performed with the example request body, it should completely unpin specified chunk.

### DIFF
--- a/pkg/api/pin_chunks.go
+++ b/pkg/api/pin_chunks.go
@@ -197,7 +197,7 @@ func (s *server) updatePinnedChunkPinCounter(w http.ResponseWriter, r *http.Requ
 	if err != nil {
 		s.Logger.Debugf("update pin counter: parse chunk ddress: %v", err)
 		s.Logger.Errorf("update pin counter: parse address")
-		jsonhttp.NotFound(w, nil)
+		jsonhttp.BadRequest(w, "bad address")
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/api/pin_chunks.go
```diff
@@ -184,6 +187,129 @@ func (s *server) getPinnedChunk(w http.ResponseWriter, r *http.Request) {
...
+		s.Logger.Errorf("update pin counter: parse address")
+		jsonhttp.NotFound(w, nil)
+		return
+	}
+
+	has, err := s.Storer.Has(r.Context(), addr)
+	if err != nil {
+		s.Logger.Debugf("update pin counter: localstore has: %v", err)
+		s.Logger.Errorf("update pin counter: store")
+		jsonhttp.NotFound(w, nil)
```
janos, line 208: This should be internal server error, as `!has` is not found.

janos (approved):

pkg/api/pin_chunks.go
zbiljic: I copied existing `getPinnedChunk` function. Will change it there also.

zelig (approved):

Issue #680, opened 2020-09-08
Standardize error codes on download API endpoints


acud:
We should return `http.StatusNotFound` for cases where the chunk cannot be found, but should also be inclusive of cases where hashes are invalid, since this information is not relevant for the users.

related to #474

PR #959, opened 2020-11-17
Normalize API error responses

zbiljic:
Updates responses on some API functions, to make it more same across all.


janos (approved):

acud:
this is reverting work that was done before pretty much because of opposite reasons. not that i care about zig-zagging, but we should at least be consistent as to which kind of internals we'd like to expose on error codes. i believe the #680 was exactly about that

acud (approved):

zbiljic:
@acud You are right, I did not consider previous changes. This PR was in response to comments from pin/unpin PR changes, where it was the comment was different. I will close this one.

PR #963, opened 2020-11-18
Allow cross origin connections using websockets.

significance:
This is a small amendment allowing users to specify origins Bee is permitted to allow websockets connections from.

I have re-used to the http CORS configuration for simplicity and to minimise code disruption. 

At present I can not think of a common use case which would preclude this assumption.

### DIFF
--- a/pkg/api/pss.go
+++ b/pkg/api/pss.go
@@ -12,6 +12,7 @@ import (
 	"net/http"
 	"strings"
 	"time"
+	"fmt"
 
 	"github.com/ethersphere/bee/pkg/crypto"
 	"github.com/ethersphere/bee/pkg/jsonhttp"
@@ -22,10 +23,6 @@ import (
 )
 
 var (
-	upgrader = websocket.Upgrader{
-		ReadBufferSize:  swarm.ChunkSize,
-		WriteBufferSize: swarm.ChunkSize,
-	}
 	writeDeadline   = 4 * time.Second // write deadline. should be smaller than the shutdown timeout on api close
 	targetMaxLength = 2               // max target length in bytes, in order to prevent grieving by excess computation
 )
@@ -86,6 +83,18 @@ func (s *server) pssPostHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *server) pssWsHandler(w http.ResponseWriter, r *http.Request) {
+	upgrader := websocket.Upgrader{
+		ReadBufferSize:  swarm.ChunkSize,
+		WriteBufferSize: swarm.ChunkSize,
+		CheckOrigin: func(r *http.Request) bool {
+			fmt.Println(r.Header)
+			o := r.Header.Get("Origin")
+			if o == "" || o != "" && (s.CORSAllowedOrigins == nil || containsOrigin(o, s.CORSAllowedOrigins)) {
+				return true
+			}
+			return false
+		},
+	}
 	conn, err := upgrader.Upgrade(w, r, nil)
 	if err != nil {
 		s.Logger.Debugf("pss ws: upgrade: %v", err)


janos (commented):
Thanks @significance. ~~Just run gofmt or goimports on pss.go file to format it exactly as expected :) https://github.com/ethersphere/bee/pull/963/checks?check_run_id=1417165761#step:7:11.~~ Just saw that you already did.

significance:
@janos yes, thankyou!

significance:
ah just noticed all my working commits are there too - sorry thought i had excluded those 🙄 

i guess gh will squash on the merge, otherwise let me know how best to get rid so that i don't cause problems 😅

pkg/api/pss.go
```diff
@@ -86,6 +82,16 @@ func (s *server) pssPostHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *server) pssWsHandler(w http.ResponseWriter, r *http.Request) {
+	upgrader := websocket.Upgrader{
+		ReadBufferSize:  swarm.ChunkSize,
+		WriteBufferSize: swarm.ChunkSize,
+		CheckOrigin: func(r *http.Request) bool {
+			if o := r.Header.Get("Origin"); o != "" && (s.CORSAllowedOrigins == nil || containsOrigin(o, s.CORSAllowedOrigins)) {
```
janos, line 89: Currently the tests are failing as this function returns false if `r.Header.Get("Origin") == ""`, which stops the handshake. If there are no "Origin" header, this function should return true `a per https://github.com/gorilla/websocket/blob/v1.4.2/server.go#L90.

significance: great, thanks!

yes, must run tests before pushing code 😂 

baby steps Dan 🤦

pkg/api/pss.go
```diff
@@ -86,6 +82,20 @@ func (s *server) pssPostHandler(w http.ResponseWriter, r *http.Request) {
...
+	upgrader := websocket.Upgrader{
+		ReadBufferSize:  swarm.ChunkSize,
+		WriteBufferSize: swarm.ChunkSize,
+		CheckOrigin: func(r *http.Request) bool {
+			o := r.Header.Get("Origin")
+			if o != "" {
+				return true
+			}
+			if o != "" && (s.CORSAllowedOrigins == nil || containsOrigin(o, s.CORSAllowedOrigins)) {
+				return true
```
acud, line 94: this condition will never get hit, because you will always return anyway on the previous `if` statement, as it comes before.

pkg/api/pss.go
```diff
@@ -86,6 +82,20 @@ func (s *server) pssPostHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *server) pssWsHandler(w http.ResponseWriter, r *http.Request) {
+	upgrader := websocket.Upgrader{
+		ReadBufferSize:  swarm.ChunkSize,
+		WriteBufferSize: swarm.ChunkSize,
+		CheckOrigin: func(r *http.Request) bool {
+			o := r.Header.Get("Origin")
+			if o != "" {
```
acud, line 90: this will always return true when origin is set. you'd probably want to get rid of this statement and just have the following one. I also think that `containsOrigin` might need to compare whether the origin header value has the compared item's value as a substring, since origin header can [contain port also](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin), it means that even if you define `localhost` as an allowed origin, it will not compare since the request came in with `localhost:1234` value on the actual header.

pkg/api/pss.go
significance: omg, yep, fixed, i am having one of those mornings, fml 🙈

pkg/api/pss.go
significance: if i understand right first point is as above i think?

second point - i think you are right - i have taken the condition directly from [here](https://github.com/ethersphere/bee/blob/2b491c9f6e7279d1c8928ad03f68ce1f6ee05b41/pkg/api/router.go#L169), the CORS is port specific so just `localhost` would imply `localhost:80`. 

so i could add `80` if port ommitted, or could add a check to include the port number in the config, as we do with api-addr et al?

acud: i'm not sure at all that the port incoming on the `Origin` header is `80` at all. If anything I think it should reflect the incoming request port and my guess is that this would be a high port. It would be cool if you could test it with the websocket on the browser side, and just print the header when you're in the upgrader (add `fmt.Println(r.Header))`) to see which port we get there. If I remember correctly when we were both testing it the port was a high port

Eknir:
@significance , are you still working on this PR or is it in review? If in review, could you please drag it to the review pipeline in the Active Sprint board and also to the correct pipeline in the review board? Thanks.

significance:
@Eknir afaik this is good to go

the ci keeps failing but not with any consistency as far as i can see, but i maybe lacking a little insight here at this point 😁

acud:
it is failing because you're PRing from your fork. Deepsource env var will only be set when PRing from `ethersphere` branches

pkg/api/pss.go
acud: i wonder if `o == "" || o != ""` makes sense here. maybe `if s.CORSAllowedOrigins == nil || containsOrigin(o,s.CORS...)` is enough. can you confirm that this works? I guess it would be nice to have a unit test on this, but manual testing due to lack of time will do at this point

Eknir:
@significance , would it be an idea for you to pair on this with for example @acud ? I would like to have this in before today's testing session.
Thanks

acud:
closed in favor of #988

PR #972, opened 2020-11-18
Surplus balances for peers, debug API addons

metacertain:
This PR involves all changes from:
https://github.com/ethersphere/bee/pull/870

And...

introduces new endpoints:
/consumed
/consumed/{peer}

The /balances/{peer} and /balances endpoints now show balances deducted by surplus balances for easy comparison of balances in integration tests.

The consumed endpoint shows current balance disregarding the surplus balance, as we consider the surplus balance sort of a pre-paid service solution.

The other part of the PR concerns Reserve checking for the balance adjusted by surplus balance so that we avoid an overdraft when a situation occurs, in which our node gets an oversettlement, and keeps reserving balance while the surplus balance does not naturally drain by requests coming from the peer.
In this situation, the other side might have processed transactions in a different order, and might consider our node's surplus balance as debt, which we don't want to increase further if this is the case, rather settle early.

### DIFF
--- a/pkg/debugapi/balances_test.go
+++ b/pkg/debugapi/balances_test.go
@@ -171,8 +171,6 @@ func equalBalances(a, b *debugapi.BalancesResponse) bool {
 	return true
 }
 
-//#####
-
 func TestConsumedBalances(t *testing.T) {
 	balancesFunc := func() (ret map[string]int64, err error) {
 		ret = make(map[string]int64)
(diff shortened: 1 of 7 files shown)


janos (approved):

pkg/debugapi/balances_test.go
```diff
@@ -170,3 +170,128 @@ func equalBalances(a, b *debugapi.BalancesResponse) bool {
 
 	return true
 }
+
+//#####
```
acud, line 174: can be removed

acud (approved):

ralph-pichler (approved):

PR #980, opened 2020-11-19
add PSS dedicated private key

janos:
This PR introduces a dedicated PSS private key, log it to the user and exposes it in debug api /addresses endpoint.

This PR also makes a change to check for libp2p key if it exists when testing if keys are created for the first time. Previously, the swarm key was checked, but it may not be there if clef is used as an external signer, resulting in asking for password twice.

### DIFF
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -99,7 +99,7 @@ Welcome to the Swarm.... Bzzz Bzzzz Bzzzz
 				return err
 			}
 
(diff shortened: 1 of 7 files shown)


acud (approved):

ralph-pichler (approved):

ralph-pichler:
one remark: with this change no longer need to inject keystore and password into `NewBee` as it's no longer used there.

janos:
> one remark: with this change we no longer need to inject keystore and password into `NewBee` as it's no longer used there.

Thanks, we will remove them in the next PR.

PR #981, opened 2020-11-19
remove unused arguments and options from node.NewBee

janos:
As per comment https://github.com/ethersphere/bee/pull/980#issuecomment-730351322.


ralph-pichler (commented):
keystore and password could be removed from the signerConfig struct too

acud (approved):

ralph-pichler (approved):

PR #987, opened 2020-11-20
remove payment tolerance check

ralph-pichler:
as payment tolerance is no longer used as a limit for incoming debt we can remove the `PaymentTolerance > PaymentThreshold/2` check.


acud (approved):

zelig (approved):

pkg/accounting/accounting.go
```diff
@@ -108,10 +108,6 @@ func NewAccounting(
 		return nil, fmt.Errorf("tolerance plus threshold too big: %w", ErrOverflow)
 	}
 
-	if PaymentTolerance > PaymentThreshold/2 {
-		return nil, ErrInvalidPaymentTolerance
```
zelig, line 112: is this error still used somewhere else? If not should probably remove it on top

ralph-pichler: removed

PR #988, opened 2020-11-20
api: add pss CORS check

acud:
supersedes https://github.com/ethersphere/bee/pull/963

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"net/url"
 	"strconv"
 	"strings"
 	"sync"
@@ -202,13 +201,9 @@ func (s *server) checkOrigin(r *http.Request) bool {
 	if len(origin) == 0 {
 		return true
 	}
-	u, err := url.Parse(origin[0])
-	if err != nil {
-		return false
-	}
-	hosts := append(s.CORSAllowedOrigins, r.Host)
+	hosts := append(s.CORSAllowedOrigins, "http://"+r.Host)
 	for _, v := range hosts {
-		if equalASCIIFold(u.Host, v) || v == "*" {
+		if equalASCIIFold(origin[0], v) || v == "*" {
 			return true
 		}
 	}
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -196,7 +196,7 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
 	}
 }
 
-// checkSameOrigin returns true if the origin is not set or is equal to the request host.
+// checkOrigin returns true if the origin is not set or is equal to the request host.
 func (s *server) checkOrigin(r *http.Request) bool {
 	origin := r.Header["Origin"]
 	if len(origin) == 0 {
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -201,7 +201,11 @@ func (s *server) checkOrigin(r *http.Request) bool {
 	if len(origin) == 0 {
 		return true
 	}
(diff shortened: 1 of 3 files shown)


pkg/api/api.go
```diff
@@ -194,9 +196,53 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
...
+	if len(origin) == 0 {
+		return true
+	}
+	u, err := url.Parse(origin[0])
+	if err != nil {
+		return false
+	}
+	hosts := append(s.CORSAllowedOrigins, r.Host)
+	for _, v := range hosts {
+		if equalASCIIFold(u.Host, v) || v == "*" {
```
janos, line 211: This check is validating only the Host part of the URL, ignoring the Schema part. This is a backward incompatible change as s.CORSAllowedOrigins were required to be a full URL and now just a Host part of it on order to pass this condition.

acud: i see. but then we can't append `r.Host`, can we? since it only hostname+port, but does not include the schema

janos: Correct, we don't even need to urlparse origin[0], just to append origin[0] as it is.

acud: i added some solution, let me know what you think. it could be that having a really correct solution which is totally by-the-book is not realistic right now, as there are some details here and we need to document how different client set the header

pkg/api/api.go
```diff
@@ -194,9 +195,49 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
 	}
 }
 
+// checkOrigin returns true if the origin is not set or is equal to the request host.
+func (s *server) checkOrigin(r *http.Request) bool {
+	origin := r.Header["Origin"]
+	if len(origin) == 0 {
+		return true
+	}
+	hosts := append(s.CORSAllowedOrigins, "http://"+r.Host)
```
janos, line 204: Scheme could be https, not just http.

```suggestion
	scheme := "http"
	if r.TLS != nil {
		scheme = "https"
	}
	hosts := append(s.CORSAllowedOrigins, scheme+"://"+r.Host)
```

pkg/api/api.go
janos: The only thing is having the correct scheme. (comment below)

pkg/api/api.go
```diff
@@ -194,9 +195,49 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
...
+}
+
 func lookaheadBufferSize(size int64) int {
 	if size <= largeBufferFilesizeThreshold {
 		return smallFileBufferSize
 	}
 	return largeFileBufferSize
 }
+
+// equalASCIIFold returns true if s is equal to t with ASCII case folding as
```
zelig, line 221: so does this mean one can abuse the cors check by creating a url that asciifold-equal to the ligit one?

acud: the ascii fold is to convert to lowercase (that's the IETF standard as per the RFC), and AFAIK hostname is case insensitive

pkg/api/api.go
acud: i merged it, although i'm not sure how this behave behind a load balancer that rewrites an HTTPS request into a regular HTTP request that goes into the actual bee node (since bee does not offer HTTPS)

janos: Correct, we do not have tls connections, but if someone would add them, then this code will not be a surprise to be handled, it will work. For proxies, the configuration would need to include the exact origin in CORSAllowedOrigins with https scheme. We are already setting `*` on gateways so it should work there.

janos (approved):

zelig (approved):

PR #989, opened 2020-11-20
use correct payment threshold in final reserve check

ralph-pichler:
fixes a check where the wrong payment threshold was used (ours instead of the peers). (this bug was already in the 0.3 releases but probably did not surface because everybody is using the same threshold)


acud (approved):

metacertain (approved):
Thanks!

PR #993, opened 2020-11-21
Replace log with stdout print on init cmd

paxthemax:
# Issue

Print bee initialization results to stdout instead of logging them. These include:
1. Swarm public key
2. PSS public key
3. Ethereum overlay address of the node

## Enhancments

Use different outputs for the print (to be consumed the user or machine clients):

Pretty output:

```
Swarm Public Key:         0x0399952b2d1aacd9dec71f42e08419402a51bf238f4793ff1b8a1c39c2c8b8f6fc
PSS Public Key:           0x0302053a6edbabeb1516b491e4efaad783e5fe8702bc0f2ece2332003961f443bc
Overlay Ethereum Address: 0xbfe535b409aa4c26426170043e683fb3c2bd6ef3
```

Bare output (for easy consumption by shell):
```
0x0399952b2d1aacd9dec71f42e08419402a51bf238f4793ff1b8a1c39c2c8b8f6fc
0x0302053a6edbabeb1516b491e4efaad783e5fe8702bc0f2ece2332003961f443bc
0xbfe535b409aa4c26426170043e683fb3c2bd6ef3
```

JSON Output for ease of consumption by non shell clients:
```
{
    "swarm_pubkey": "0x0399952b2d1aacd9dec71f42e08419402a51bf238f4793ff1b8a1c39c2c8b8f6fc",
    "pss_pubkey": "0x0302053a6edbabeb1516b491e4efaad783e5fe8702bc0f2ece2332003961f443bc",
    "eth_address": "0xbfe535b409aa4c26426170043e683fb3c2bd6ef3"
}
```

## Notes
IMO the `cmd` code is starting to smell and could use a refactor.

### DIFF
--- a/cmd/bee/cmd/init.go
+++ b/cmd/bee/cmd/init.go
@@ -5,12 +5,11 @@
 package cmd
 
 import (
-	"fmt"
 	"io/ioutil"
-	"strings"
+	"os"
 
+	"github.com/ethersphere/bee/pkg/crypto"
 	"github.com/ethersphere/bee/pkg/logging"
-	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


Eknir:
@paxthemax @acud , what is the status of this PR?

Eknir:
@paxthemax , could you please make an issue and then connect this PR with that issue?

cmd/bee/cmd/init.go
```diff
@@ -23,25 +22,13 @@ func (c *command) initInitCmd() (err error) {
...
-			case "3", "info":
-				logger = logging.New(cmd.OutOrStdout(), logrus.InfoLevel)
-			case "4", "debug":
-				logger = logging.New(cmd.OutOrStdout(), logrus.DebugLevel)
-			case "5", "trace":
-				logger = logging.New(cmd.OutOrStdout(), logrus.TraceLevel)
-			default:
-				return fmt.Errorf("unknown verbosity level %q", v)
-			}
+			res, err := c.configureSigner(cmd, logging.New(ioutil.Discard, 0))
```
acud, line 25: @paxthemax `configureSigner` uses the passed logger to print information about the process. 

Using `ioutil.Discard` here essentially means we're silencing all possible warnings and other information that is printed there. I think that some of the information is also good for debugging purposes so i wouldn't be in a rush to remove it.

Why not remove the log lines that print the key info in `configureSigner` and just print the keys nicely here once the configuration part is over. Also, you mentioned different types of printouts and I think that for the terminal output we should print the keys with the descriptions of them (like in the first example in the PR description), since this one only prints the raw keys.

Eknir:
@paxthemax , could you please help me here. I recall that you were having a PR, but in the end did not merge that. What happened with that code? 
It would be an option to close this issue.

Eknir:
Whoops, I see that this is the PR. Would be good to either merge the PR (in the current state or with minimal effort, if there is an improvement of the code) or close the PR otherwise. My personal preference would be to merge it (without looking at the state of the code).

Issue #1026, opened 2020-11-30
Perform clean for TAR archive paths


zbiljic:
# Task

We should use `filepath.Clean` for paths fro TAR archive.

related issue: #999

Eknir:
This needs documentation. @zbiljic , could you please provide an a list or some text, mentioning what kind of directory structures cannot be uploaded to Bee?

zbiljic:
I think more in the way that we should support any "normal" directory structure (that is archived using TAR). Current state is that there might be some cases that are not covered (because we have not encountered them yet).

As noted in the #1024 (where further discussion should be continued), if someone (when creating TAR archive) adds some files/directories, that are in some parent directory, that is relative from the current working directory from which archive is being created, even though paths are valid strings, they will not be resolvable through Bee.
Issue can be read for some examples, as well as suggestions for possible changes.

PR #1027, opened 2020-11-30
Clean file path from tar archive

zbiljic:
related issue: #1026

### DIFF
--- a/pkg/api/dirs.go
+++ b/pkg/api/dirs.go
@@ -128,7 +128,12 @@ func storeDir(ctx context.Context, reader io.ReadCloser, s storage.Storer, mode 
 			return swarm.ZeroAddress, fmt.Errorf("read tar stream: %w", err)
 		}
 
-		filePath := fileHeader.Name
+		filePath := filepath.Clean(fileHeader.Name)
+
+		if filePath == "." {
+			logger.Warningf("skipping file upload empty path")
+			continue
+		}
 
 		// only store regular files
 		if !fileHeader.FileInfo().Mode().IsRegular() {
--- a/pkg/api/dirs.go
+++ b/pkg/api/dirs.go
@@ -131,7 +131,7 @@ func storeDir(ctx context.Context, reader io.ReadCloser, s storage.Storer, mode 
 		filePath := filepath.Clean(fileHeader.Name)
 
 		if filePath == "." {
-			logger.Warningf("skipping file upload empty path")
+			logger.Warning("skipping file upload empty path")
 			continue
 		}
 
--- a/pkg/api/dirs.go
+++ b/pkg/api/dirs.go
@@ -15,6 +15,7 @@ import (
 	"mime"
 	"net/http"
 	"path/filepath"
+	"runtime"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/api/dirs.go
```diff
@@ -128,7 +128,15 @@ func storeDir(ctx context.Context, reader io.ReadCloser, s storage.Storer, mode
...
-		filePath := fileHeader.Name
+		filePath := filepath.Clean(fileHeader.Name)
+
+		if filePath == "." {
+			logger.Warningf("skipping file upload empty path")
+			continue
+		}
+
+		// always use Unix path separator
+		filePath = strings.ReplaceAll(filePath, "\\", "/")
```
acud, line 139: minor comment, that `\\` is a valid POSIX filename. there's no other solution here though, since we cannot know on which operating system the tarball was created, and I don't know if a tarball contained on a windows machine writes the metadata header with POSIX compatible filenames

acud (approved):

zbiljic:
Updated to replace to slash only if on Windows (and path was actually changed by the `filepath.Clean` function).

Unless we go into full custom `Clean` function, I do not see the way to cover all cases. Even with this custom work, we would still have some possible inconsistencies.

acud:
Thanks @zbiljic. Obviously it won't catch the edge case where a tar created on a windows machine is uploaded to a bee which runs on *nix, so it might screw things up on the gateway but unless we go for our own path encoding I don't see a way around it.

zbiljic:
@acud Note that even if the TAR is created on Windows, we would most probably prefer to have Unix-style paths, especially because in the end we will use it via HTTP.
I think that some browsers even do same replacement.

zbiljic:
I will perform some additional tests on Windows.

pkg/api/dirs.go
```diff
@@ -128,7 +128,15 @@ func storeDir(ctx context.Context, reader io.ReadCloser, s storage.Storer, mode
 			return swarm.ZeroAddress, fmt.Errorf("read tar stream: %w", err)
 		}
 
-		filePath := fileHeader.Name
+		filePath := filepath.Clean(fileHeader.Name)
+
+		if filePath == "." {
+			logger.Warningf("skipping file upload empty path")
```
janos, line 134: `Warningf` -> `Warning`

janos (approved):
👍

zbiljic:
I have attempted several ways of creating TAR archive on Windows, using command line tool, and each time the archive used slashes (`/`) for paths. Another thing that I attempted was to create TAR archive using Go, with the built-in library, and that one also used slashes in the resulting archive.

Glazing over TAR specification there is only mentioning of `/` as separator. Apart from going into implementations of different libraries, I would surmise that slash is the standard, and replacement after clean is required (as it is for Windows). This may end up missing some files, with `\` in the path (when uploading to Windows), as commented, however unless we manually code the cleanup function (to maybe properly account for `\` on Windows), we will miss things.

Of course, it might be possible for someone to manually create archive which uses `\` instead of slash, or maybe some random library does that, however I would not consider that as standard, unless otherwise proven.

PR #1031, opened 2020-12-02
github: disable pullsync integration tests

acud:
pullsync integration tests are not needed at this point, since we cannot ensure that a chunk has designated replicators.


vandot (approved):

vandot:
@acud should I disable pullsync test for nightly test as well https://github.com/ethersphere/bee-argo/blob/master/infra-test/beekeeper.yaml#L49 ?

acud:
@vandot yes please :+1:

PR #1032, opened 2020-12-02
Resolve peer topology inconsistencies

zbiljic:
relates to: #1034 

Change the way how we check if some peer is connected.

In addition to checking `libp2p` peer ID, it also checks if that peer is connected through expected address (see related issue for more in-depth explanation of the problem). Only in the case when both parts are same can we say for sure that we have multiple connections to same peer.

One additional change was to one test, `TestDoubleConnectOnAllAddresses`, which was failing most of the times (few times it passed). Change involves actually changing when the other "peer" is created, and we create new peer for each address that we test for. This was done because there does not seem to be a way to make sure connection to other peer is completely terminated.
What was happening was that once a connection is created to address to a peer, even if `Disconnect` is called, on the next attempt to connect to the same peer, the `libp2p` library may end up using previous address for connection, which will then cause this new function not to recognize peer as already connected.

### DIFF
--- a/pkg/p2p/libp2p/peer.go
+++ b/pkg/p2p/libp2p/peer.go
@@ -165,26 +165,32 @@ func (r *peerRegistry) overlay(peerID libp2ppeer.ID) (swarm.Address, bool) {
 }
 
 func (r *peerRegistry) isConnected(peerID libp2ppeer.ID, remoteAddr ma.Multiaddr) (swarm.Address, bool) {
+	if remoteAddr == nil {
+		return swarm.ZeroAddress, false
+	}
+
 	r.mu.RLock()
 	defer r.mu.RUnlock()
 
 	overlay, found := r.overlays[peerID]
+	if !found {
+		return swarm.ZeroAddress, false
+	}
 
-	if found && remoteAddr != nil {
-		// check connection remote address
-		if conns, ok := r.connections[peerID]; ok {
-			found = false
-			for c := range conns {
-				if c.RemoteMultiaddr().Equal(remoteAddr) {
-					// we ARE connected to the peer on expected address
-					found = true
-					break
-				}
-			}
+	// check connection remote address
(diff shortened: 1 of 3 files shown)


acud (approved):

pkg/p2p/libp2p/peer.go
```diff
@@ -163,6 +164,26 @@ func (r *peerRegistry) overlay(peerID libp2ppeer.ID) (swarm.Address, bool) {
 	return overlay, found
 }
 
+func (r *peerRegistry) isConnected(peerID libp2ppeer.ID, remoteAddr ma.Multiaddr) (swarm.Address, bool) {
+	r.mu.RLock()
+	overlay, found := r.overlays[peerID]
```
janos, line 169: I would suggest to have a little less nesting. These suggestions are just opinionated cosmetics, the code as it is is fine also.

- check if `remoteAddr == nil` and return if it is before acquiring the lock
- acquire lock and defer unlock
- return if `!found`
- return if connections map lookup `!ok`
- return instead of break in the for loop

janos (approved):
Nice find, LGTM.

pkg/p2p/libp2p/peer.go
zbiljic: The function that was used for checking if peer is connected was the one before, named `overlay`. From that one, I just added additional check inside.

I do not think we should remote just if `remoteAddr == nil` because the `peerID` is the first and main thing that is checked. Admittedly current solution does additionally check for remote address.

The lookup in connections map should never return `false` in the `found == true`, there just was no reason not to check it (just to prevent possible panic).

However, I do understand that changes may make things more clear.

janos: Thanks for the explanation and for the change @zbiljic.

janos (approved):

PR #1033, opened 2020-12-02
add support for docker-compose

vandot:
@Eknir should I include README inside `packaging/docker` folder or I should just add docker-compose section to docs?


Dockerfile
```diff
@@ -19,6 +19,9 @@ RUN apt-get update && apt-get install -y \
     groupadd -r bee --gid 999; \
     useradd -r -g bee --uid 999 --no-log-init -m bee;
 
+# make sure mounted volumes have correct permissions
+RUN mkdir -p /home/bee/.bee && chown 999:999 /home/bee/.bee
```
vandot, line 23: If folder is not precreated and permissions set when you mount external volume it will be owned by root

packaging/docker/env.example
```diff
@@ -0,0 +1,75 @@
...
+# CLEF_CHAINID=12345
+
+### BEE
+
+## HTTP API listen address (default ":1633")
+# BEE_API_ADDR=:1633
+## initial nodes to connect to (default [/dnsaddr/bootnode.ethswarm.org])
+# BEE_BOOTNODE=[/dnsaddr/bootnode.ethswarm.org]
+## enable clef signer
+# BEE_CLEF_SIGNER_ENABLE=false
```
vandot, line 15: @Eknir should I enable clef and set clef endpoint or should I leave it to the user?

Eknir:
> @Eknir should I include README inside `packaging/docker` folder or I should just add docker-compose section to docs?

Hey,

Please add it to  `packaging/docker` . @significance will decide to put it in the docs if this is relevant.

packaging/docker/env.example
Eknir: Please make it in such a way as to subtly nudge the user to use Clef wherever possible and feasible.

packaging/docker/docker-compose.yml
```diff
@@ -0,0 +1,61 @@
...
+    image: ethersphere/clef:0.4.2
+    restart: unless-stopped
+    environment:
+      - CLEF_CHAINID
+    volumes:
+      - clef:/app/data
+    command: full
+
+  bee:
+    image: ethersphere/bee:0.4.0
```
zbiljic, line 14: We will forget to update version at some point.

Maybe we should add make step to update on all places?

zbiljic (approved):

packaging/docker/env.example
vandot: @Eknir check README

packaging/docker/docker-compose.yml
vandot: I agree with you but not sure how to set that up

vandot:
> > @Eknir should I include README inside `packaging/docker` folder or I should just add docker-compose section to docs?
> 
> Hey,
> 
> Please add it to `packaging/docker` . @significance will decide to put it in the docs if this is relevant.

Added README

Eknir (approved):

Issue #1037, opened 2020-12-03
Refactor password prompt from cmd


paxthemax:
# Task

Currently, the password prompt is buried in the bee cmd and tightly coupled with the command. I propose we refactor the password prompt into a separate interface as an internal package available in cmd. This will also make the implementation testable.

## Acceptance criteria

Refactor with no functionality lost, add tests.

PR #1038, opened 2020-12-03
Refactor password prompt to internal package

paxthemax:
# Description

Currently, the password prompt is buried in the `bee\cmd\bee` package, is not testable, and is not available to other commands.

This PR reimplements the password prompt functionality in an internal package to `bee\cmd`, with unit tests. It exposes the following interfaces:
- `PasswordReader` - for reading passwords securely from stdin or reading from a (plaintext) file
- `PasswordPrompter` - for generating a prompt and reading from a `PasswordReaded`
- `PasswordConfirmer` - for confirming a password with two prompts (which must match)

Note that this PR does replace terminal functionality (I will do it separately).

Related to #1037

### DIFF
--- a/cmd/internal/terminal/export_test.go
+++ b/cmd/internal/terminal/export_test.go
@@ -1,0 +1,33 @@
+package terminal
+
+import (
+	"io"
+)
+
+// WithPromptOut sets the out file to write the prompt to.
+func WithPromptOut(out io.Writer) PasswordPrompterOption {
+	return func(fpp *WriterPasswordPrompter) {
+		fpp.out = out
+	}
+}
+
+// WithPromptPasswordReader sets the password reader.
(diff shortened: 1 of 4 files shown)


janos (commented):
It is good to have a new package with tests, even I think that it may not be strictly necessary, as I see no significant need for a special terminal package. But if it is prioritised already, that is fine.

In general, the abstraction looks pretty heavy given that what is actually tested is whether two strings are the same or that the file contains an expected data. I do not see that `StdinPasswordReader` is tested, which contains the crucial ssh/teminal.ReadPassword call.

Eknir:
@paxthemax , what is the status of this PR?

paxthemax:
Closing this PR until I come up with a better solution.

Issue #571, opened 2020-08-14
Pin in retrospect


Eknir:
# User story

## Story
As a user of Swarm, I want to pin a file which is already in the network, such that I can persist this content

## Background

Right now, pinning is architected for upload + pin, but it is expected that users want to pin content which is already in the network (and not necessarily already in their local store).

## Tasks

| Task | Assignee |  Done (1), or 0 |
| ------ |------------| -----|

ldeffenb:
I'll second this request as it will support pinning services to simply request the pin instead of needing to fetch the block and then pin, hoping that the fetched block is actually still local in the cache.

zbiljic:
First PR for this (#1045), main change was that now chunks are actually being fetched through `netstore`, which means we are actually checking network for chunks that are being pinned.

The thing that is not changed, and is still "open issue" is what happens if we cannot fetch some chunk that was found for some "root" hash. This relates to traversed chunks for complex structures we support. Current behavior is that as soon as we reach some chunk that is un-accessible, process is stopped. In that case user is left with not complete data, however trying to fetch other chunks really has little point, if something is missing.

There are multiple possible solutions for solving this problem, some of this @Eknir and myself talked about, but I would like to open this issue for further discussion.

Eknir:
Hi Nemanja,

Thanks for your RFC. Indeed, we spoke about it. Let me outline what I can still remember:

For a user, who wants content to be pinned from the network, it is not great if the upload stops somewhere midway. Ideally, the node should continue pinning the best it can, and only then return, with the information that it only partly succeeded.
Obviously, half-pinned content is not valuable for the user. As a user, you only care about stuff that is fully pinned. That is why we must somehow allow for retrying. Retrying makes sense as the content might be re-uploaded or becomes retrievable again.

We spoke about launching an in-node process that would do this retrying, but as you pointed out, this is opening a box of pandora and will look to obscure to the user. Instead, it would be much better, if we facilitate the primitives inside the bee-node for the user to retry. 
The biggest problem is that pinning something for the second time (retrying) will increase the pin-counter for all chunks. The user does not want this; it merely wants to increment the pincounter once for all the content of the root-hash across different retry sessions.
This, you called "idempotent pinning". As a user-story:
`As a pinner of content content that is already in the network, I want to be able to retry pinning content if it was not available or not reachable, such that I will be able to pin all content under a root hash and afterwards easily unpin it again`

[...]

PR #1045, opened 2020-12-10
Fetch chunks from network when pinning

zbiljic:
relates to: #571 

Pinning operations are changed to fetch all (locally) missing chunks.
New storage mode was added, the `storage.ModePutRequestPin`, which performs pin operation in the same time when the chunk is retrieved. This new mode might not be required, but it is safer than having fetch-pin-check operations in succession.

Modified `storage.ModeSetPin` operation, to check if the chunk exists locally before performing pin operation, under same lock (changed Mock storage also). This caused some changes on how the chunk is being pinned.

### DIFF
--- a/pkg/api/pin_bytes.go
+++ b/pkg/api/pin_bytes.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 
 	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/ethersphere/bee/pkg/traversal"
 	"github.com/gorilla/mux"
@@ -33,8 +34,14 @@ func (s *server) pinBytes(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if !has {
-		jsonhttp.NotFound(w, nil)
-		return
+		_, err := s.Storer.Get(r.Context(), storage.ModeGetRequest, addr)
+		if err != nil {
+			s.Logger.Debugf("pin chunk: netstore get: %v", err)
+			s.Logger.Error("pin chunk: netstore")
+
+			jsonhttp.NotFound(w, nil)
+			return
+		}
 	}
 
 	ctx := r.Context()
--- a/pkg/api/pin_chunks.go
+++ b/pkg/api/pin_chunks.go
@@ -328,11 +328,19 @@ func (s *server) pinChunkAddressFn(ctx context.Context, reference swarm.Address)
 
 		if !has {
 			// chunk not found locally, try to get from netstore
-			_, err := s.Storer.Get(ctx, storage.ModeGetRequest, address)
+			ch, err := s.Storer.Get(ctx, storage.ModeGetRequest, address)
 			if err != nil {
 				s.Logger.Debugf("pin traversal: storer get: for reference %s, address %s: %w", reference, address, err)
 				return true
 			}
+
+			_, err = s.Storer.Put(ctx, storage.ModePutRequestPin, ch)
+			if err != nil {
+				s.Logger.Debugf("pin traversal: storer put pin: for reference %s, address %s: %w", reference, address, err)
+				return true
+			}
+
+			return false
 		}
 
 		err = s.Storer.Set(ctx, storage.ModeSetPin, address)
--- a/pkg/api/pin_chunks.go
+++ b/pkg/api/pin_chunks.go
@@ -41,8 +41,14 @@ func (s *server) pinChunk(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if !has {
-		jsonhttp.NotFound(w, nil)
-		return
+		_, err := s.Storer.Get(r.Context(), storage.ModeGetRequest, addr)
+		if err != nil {
+			s.Logger.Debugf("pin chunk: netstore get: %v", err)
+			s.Logger.Error("pin chunk: netstore")
+
+			jsonhttp.NotFound(w, nil)
+			return
+		}
 	}
 
 	err = s.Storer.Set(r.Context(), storage.ModeSetPin, addr)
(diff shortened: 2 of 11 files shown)


pkg/api/pin_chunks.go
```diff
@@ -312,10 +318,34 @@ func (s *server) updatePinCount(ctx context.Context, reference swarm.Address, de
...
+		if err != nil {
+			s.Logger.Debugf("pin traversal: localstore has: for reference %s, address %s: %w", reference, address, err)
+			return true
+		}
+
+		if !has {
+			// chunk not found locally, try to get from netstore
+			ch, err := s.Storer.Get(ctx, storage.ModeGetRequest, address)
+			if err != nil {
+				s.Logger.Debugf("pin traversal: storer get: for reference %s, address %s: %w", reference, address, err)
```
acud, line 333: this too

pkg/api/pin_chunks.go
```diff
@@ -312,10 +318,34 @@ func (s *server) updatePinCount(ctx context.Context, reference swarm.Address, de
...
+				return true
+			}
+
+			_, err = s.Storer.Put(ctx, storage.ModePutRequestPin, ch)
+			if err != nil {
+				s.Logger.Debugf("pin traversal: storer put pin: for reference %s, address %s: %w", reference, address, err)
```
acud, line 339: this too

pkg/api/pin_chunks.go
```diff
@@ -312,10 +318,34 @@ func (s *server) updatePinCount(ctx context.Context, reference swarm.Address, de
...
+				return true
+			}
+
+			return false
+		}
+
+		err = s.Storer.Set(ctx, storage.ModeSetPin, address)
```
acud, line 346: just an observation - you might have a data race between `L324` and here, and the chunk might be already garbage collected before it can be pinned. I'm trying to think if we need a separate `ModeGet` that also pins a chunk atomically.

pkg/api/pin_files.go
```diff
@@ -33,8 +34,14 @@ func (s *server) pinFile(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if !has {
-		jsonhttp.NotFound(w, nil)
-		return
+		_, err := s.Storer.Get(r.Context(), storage.ModeGetRequest, addr)
```
acud, line 37: can be reduced to one block as in the other api endpoints

acud (commented):
looks good, i have a few questions

pkg/api/pin_bytes.go
```diff
@@ -33,8 +34,14 @@ func (s *server) pinBytes(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if !has {
-		jsonhttp.NotFound(w, nil)
-		return
+		_, err := s.Storer.Get(r.Context(), storage.ModeGetRequest, addr)
```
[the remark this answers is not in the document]
zbiljic: That would be possible, however the reason I left existing code, i.e. `store.Has` function call (and why I would write it with it if I was doing it from scratch), was because I wanted to avoid unnecessary disk read operations.

As the operation flow goes:
- in the "best" case we use `store.Has`, which just does `leveldb.Has` operation, which should only check if key exists
  - after that we would write/update other key, for actual pinning operation
- if we were to use `store.Get` we have 2 internal cases
  - `netstore` case, where we do not have chunk locally
    - this would be worst case, and has to be done as is, no going around this
    - chunk will be retrieved (if possible) and added to localstore
  - localstore case, in which case actual content for the chunk would be read into newly allocated byte slice
    - we have no use for the chunk content at all, and will produce heap usage which is discarded

In regards to "more disk read", I know that that is a bit of a stretch comment, especially knowing how LevelDB actually works (not to mention also file system internal caching and prefetching), however I do think it is valid one. At least having less allocation should be enough of a reason for having current code as-is.

pkg/api/pin_chunks.go
zbiljic: Yes, that is correct. This was existing way of doing things, and I figured to leave it as it was.

Reason for adding `storage.ModePutRequestPin` mode was to have in in single operation for fetch and pin.
This relates to comment I made here already. We would have to completely read the chunk in order to have it pinned "atomically" (under same lock).

Even though recognize we should have something, I do not completely agree with suggested solution. I do not think that `ModeGet` operation should perform writes in the database. Read operations should only be used to read.

I would rather change the `ModeSetPin` operation to check if the chunk actually exists before adding the pin or return error otherwise, however in that case we also access different indexes.

pkg/api/pin_chunks.go
```diff
@@ -312,10 +318,34 @@ func (s *server) updatePinCount(ctx context.Context, reference swarm.Address, de
 
 func (s *server) pinChunkAddressFn(ctx context.Context, reference swarm.Address) func(address swarm.Address) (stop bool) {
 	return func(address swarm.Address) (stop bool) {
-		err := s.Storer.Set(ctx, storage.ModeSetPin, address)
+		// NOTE: stop pinning on first error
+
+		has, err := s.Storer.Has(ctx, address)
+		if err != nil {
+			s.Logger.Debugf("pin traversal: localstore has: for reference %s, address %s: %w", reference, address, err)
```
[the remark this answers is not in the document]
zbiljic: Note about this comment.

At first I did make the suggested change, I have since moved it back to `Debug` level. Reason for this is because on any error the iteration will be stopped abruptly, and I think that `Debug` level is useful, as only single line will be written. If the case would be that for each missing chunk we log something I would move it to `Trace` level.

If you still feel we should move it to `Trace` level, please comment, and I will add the change.

pkg/api/pin_bytes.go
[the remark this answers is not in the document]
acud: cool. you've convinced me :+1:

pkg/api/pin_chunks.go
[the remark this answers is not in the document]
acud: OK fair enough :+1:
My reasoning was because: when logging actions on individual chunks I tend to stick to Trace but if we're talking about 1 log line that will be written and then the operation terminates then all good :+1:

pkg/api/pin_chunks.go
acud: so the only suggestion that comes into mind would to (unfortunately) use `Put` here with a `ModePutUploadPin`, but have `Put` implemented in a way that if the chunk already exists in the database, then it would just increment the pin counter (which is already the case).

zbiljic: Yeah, every time I look at `Mode` enums, I think maybe I can bitmask them, and then remember otherwise.

Not sure if you checked updated code (I updated description about the change), it read as:

> Modified storage.ModeSetPin operation, to check if the chunk exists locally before performing pin operation, under same lock (changed Mock storage also). This caused some changes on how the chunk is being pinned.

So basically solution I did here is modify existing operation. My thinking was that it is acceptable to change `storage.ModeSetPin` to expect the chunk to exists locally. If this is not acceptable we could discuss other solutions.

acud (approved):

janos (approved):
Very nice. LGTM.

PR #1050, opened 2020-12-15
Fix version docker compose

vandot:
With new release fix for Dockerfile will be applied in released container, hardcoding version in docker-compose.yaml

### DIFF
--- a/packaging/docker/docker-compose.yml
+++ b/packaging/docker/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3"
 
 services:
   signer:
-    image: ethersphere/clef:0.4.2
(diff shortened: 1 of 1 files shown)


svetomir (approved):

acud (approved):

Eknir:
@vandot , could you please ensure that, if we need to update the version in docker-compose, that this is in the roamResearch release workflow?

vandot:
> @vandot , could you please ensure that, if we need to update the version in docker-compose, that this is in the roamResearch release workflow?

@Eknir what do you mean by that? To include change inside release notes?

PR #1056, opened 2020-12-15
set only two fresh clusters in ci and slash tests

vandot:


vandot:
Its almost twice as fast ~14min compared to ~24min before
@acud @ralph-pichler all tests are executed twice, once without clef, second time with clef enabled, maybe there is something we can trim more check tests run here https://github.com/ethersphere/bee/runs/1557979282?check_suite_focus=true

acud:
great improvement @vandot.
from my perspective I don't know if we really need to run the whole suite twice.
My suggestion would be to have the test clef cover what could be more relevant to actually having clef plugged in as a signer.
Therefore, when testing _with_ clef I would not test: 
- gc
- pin chunk 
- pin bytes
- manifest

I am almost tempted to say that we even should not test pushsync and retrieval, since both are covered as part of the balances and settlements tests (balances and settlements use upload and download functionalities, therefore using pushsync and retrieval under the hood). But I am aware this might be a bit extreme. On the very least, I think that the 4 tests above could potentially be removed. But would be happy to hear others. cc @janos @svetomir @ralph-pichler

ralph-pichler:
@acud the 4 tests you mentioned seem very independent of clef, so they can surely be removed. Maybe leave one of pushsync / retrieval tests in, so we at least have a test that shows that the basic stuff work even if there are accounting issues.
We could also get rid of balances here also as settlement already includes checking the balance (actually we could probably remove the balances test altogether as the same argument also applies to the non-clef case).

Eknir:
No specific opinion on this. Thanks

janos:
If removing some tests reduce time significantly, it is good to have only targeted clef enabled tests. Anyways, this is great to have in CI pipeline, even as it is.

vandot:
We would gain the most by removing balances test completely and if we remove these 4 from clef run we would strip ~68s in total which is not bad at all
- gc 8s
- pin chunk 7s
- pin bytes 10s
- manifest 5s
- balances 19s
after this PR is merged I will try to further speed up ci tests by optimizing cluster init (this step takes the most time) and using caching.
@acud removing this 4 makes sense if they are not hitting signer :+1: 
@ralph-pichler please confirm that we can remove balances test completely

acud:
> @acud removing this 4 makes sense if they are not hitting signer +1

:+1:

Eknir:
@vandot , if you got your comments, could you please remove rfc label?

acud (approved):
LGTM. Very nice improvement

ralph-pichler:
@vandot balances can be removed

ralph-pichler (approved):

PR #1059, opened 2020-12-16
If P2P connection exists compare overlay address

zbiljic:
If (for some reason) peer changes its Swarm network address (`overlay` address), and then joins again, because it will have same P2P peer ID, some peer may be marked as already connected even though it is not.

Handling of `p2p.ErrAlreadyConnected` error is changed to actually check if the already connected `overlay` address matches to the one we are trying to connect. If that is in fact the case, we correct that is the same peer, otherwise we return an error or try to connect again.

### DIFF
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -336,13 +336,11 @@ func (k *Kad) connect(ctx context.Context, peer swarm.Address, ma ma.Multiaddr, 
 	i, err := k.p2p.Connect(ctx, ma)
 	if err != nil {
 		if errors.Is(err, p2p.ErrAlreadyConnected) {
-			if i == nil || swarm.ZeroAddress.Equal(i.Overlay) || i.Overlay.Equal(peer) {
-				return nil
-			}
-
 			if !i.Overlay.Equal(peer) {
 				return errOverlayMismatch
 			}
+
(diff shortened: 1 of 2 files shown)


pkg/kademlia/kademlia.go
```diff
@@ -336,7 +336,13 @@ func (k *Kad) connect(ctx context.Context, peer swarm.Address, ma ma.Multiaddr,
 	i, err := k.p2p.Connect(ctx, ma)
 	if err != nil {
 		if errors.Is(err, p2p.ErrAlreadyConnected) {
-			return nil
+			if i == nil || swarm.ZeroAddress.Equal(i.Overlay) || i.Overlay.Equal(peer) {
+				return nil
```
acud, line 340: if the peer is already connected, how could it be that `i` is `nil` or `ZeroAddress`?

zbiljic: Originally the `i` was not being returned when error (i.e. `p2p.ErrAlreadyConnected`) was response from `Connect` call. Since one of the change here now returns `bzz.Address` on that error, it should be fine to remove additional checks. I had them while testing possible solutions; will remove it.

acud (approved):
LGTM

janos (approved):

PR #943, opened 2020-11-13
support windows services for the start command

janos:
This PR is based on the work done by @metacertain in https://github.com/ethersphere/bee/pull/941. It makes use of github.com/kardianos/service to add support for windows service.

PR #1060, opened 2020-12-16
Windows service support

zbiljic:
Adds support for running Bee as a service.

This change similar to previous one, main thing here is that we add support for writing logs to Windows Event Log, which is only enabled if we are running as Windows service. The #943 was making usage of `github.com/kardianos/service` library fully, while these changes only use it if we detect that Bee is running as Windows service.

The name for the service is `BeeSvc`. This same name must be fixed, and same one must be used when installing service on Windows machine.

To create the service on Windows run:
```
sc.exe create BeeSvc binPath="[INSTALLDIR]bee.exe start --config=[INSTALLDIR]bee.yaml" type=share start=auto displayName="Bee"
```

Note that configuration file must be properly modified, in order for Bee to be able to start properly. I suggest trying to start it from the command line first, before starting the created service.

Does not contain additional required changes, for different packaging methods so that Bee can be installed on Windows. That is planned after this one.

### DIFF
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -44,44 +44,27 @@ func (c *command) initStartCmd() (err error) {
 				return cmd.Help()
 			}
 
-			isWindowsService, err := isWindowsService()
-			if err != nil {
-				return fmt.Errorf("failed to determine if we are running in service: %w", err)
-			}
-
-			var level logrus.Level
+			var logger logging.Logger
 			switch v := strings.ToLower(c.config.GetString(optionNameVerbosity)); v {
 			case "0", "silent":
-				break
+				logger = logging.New(ioutil.Discard, 0)
 			case "1", "error":
-				level = logrus.ErrorLevel
+				logger = logging.New(cmd.OutOrStdout(), logrus.ErrorLevel)
 			case "2", "warn":
-				level = logrus.WarnLevel
+				logger = logging.New(cmd.OutOrStdout(), logrus.WarnLevel)
 			case "3", "info":
-				level = logrus.InfoLevel
+				logger = logging.New(cmd.OutOrStdout(), logrus.InfoLevel)
 			case "4", "debug":
-				level = logrus.DebugLevel
+				logger = logging.New(cmd.OutOrStdout(), logrus.DebugLevel)
 			case "5", "trace":
-				level = logrus.TraceLevel
+				logger = logging.New(cmd.OutOrStdout(), logrus.TraceLevel)
 			default:
 				return fmt.Errorf("unknown verbosity level %q", v)
 			}
 
-			if isWindowsService {
-				// max 'warning' level for Windows service
-				switch level {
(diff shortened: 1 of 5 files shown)


metacertain (approved):
Worked great, code looks great, thank you!

cmd/bee/cmd/start.go
```diff
@@ -29,10 +29,15 @@ import (
 	"github.com/ethersphere/bee/pkg/node"
 	"github.com/ethersphere/bee/pkg/resolver/multiresolver"
 	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/kardianos/service"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
+const (
+	serviceName = "BeeSvc"
```
acud, line 38: I'm just thinking if we'd like to choose a longer name here, just to avoid possible conflicts with other services. Maybe `SwarmBeeSvc`

cmd/bee/cmd/start_windows.go
```diff
@@ -0,0 +1,98 @@
...
+func (l *windowsEventLogger) Tracef(format string, args ...interface{}) {
+	// ignore
+}
+
+func (l *windowsEventLogger) Trace(args ...interface{}) {
+	// ignore
+}
+
+func (l *windowsEventLogger) Debugf(format string, args ...interface{}) {
+	// ignore
```
acud, line 53: does this mean we will never be able to log anything > INFO?

acud (approved):
Brilliant, thanks!

cmd/bee/cmd/start.go
vandot: @acud or `BeeSwarmSvc` when you open Services to be at top :)
I agree to set more unique name for service itself @zbiljic

zbiljic: I am totally fine with any name. The one here was copied from some other branch.

cmd/bee/cmd/start_windows.go
zbiljic: Yes, as we talked about, 'Debug' and 'Trace' level logs will not be pushed to Windows Event Log.

While it could be possible to setup (which I already did and tested) to log those lower levels, I am not completely confident in Windows Event Log to push possibly huge amount of logs into. If we were actually developing for Windows, I assume we would have greater knowledge about this and could made assumptions what can and cannot be done without problems. As for now, I think better option is to just not log anything under `Info` level.

acud: :+1:

cmd/bee/cmd/start.go
acud: I don't mind either. @zbiljic I leave it up to you

vandot (approved):

vandot:
@zbiljic yes that is the correct name

janos (approved):
Very nice, thanks.

Issue #1006, opened 2020-11-23
Deterministic manifest address


agazso:
#### Summary
When I upload a tar file using the `/dirs` endpoint twice I get different manifests. It would be nice if there was a way to upload the same file twice and getting back the same reference.

zbiljic:
This is intentional behavior, random obfuscation key is generated for each manifest (to prevent some possible attacks, for example [Known-plaintext attack](https://en.wikipedia.org/wiki/Known-plaintext_attack)).

@zelig can provide much better explanation about this.

zelig:
this is still annoying. we should be able to set the obfuscation key deterministically

agazso:
I can imagine having a query parameter on the endpoint where you can pass in the obfuscation key.

zbiljic:
@zelig Of course, I am not opposed to this. Support for that is in the manifest library. Only question is how do we want to implement this change.

Eknir:
@zbiljic , what about the suggestion by @agazso ?

zbiljic:
@Eknir I do not have any opposing arguments. There are numerous possibilities for solving this, some could be done without user involvement.

However, as I mentioned, I would want comment from @zelig about this, as discussion contradicts his previous comments.
One comment here is that we should create obfuscation key deterministically, which should be possible, and that can be done, I would just like some confirmation about this.

Eknir:
Hi @agazso , I was trying to make a user story out of this, and it actually strikes me that I couldn't find a real reason why a user would want this, apart from "it would be nice to have this". I also believe this myself. Actually, I realized this is due to not understanding why similar content would have a different hash.
Would you agree that this "not understanding" is the real reason? Do you think that this issue can also be solved with a proper reasoning as to why the address of similar content uploaded via the `dirs` api is different?

Eknir:
Note: I also don't know why this is the case myself. Maybe @zbiljic can enlighten us (I ready up about the known-plaintext attack, but still don't understand)?

agazso:
> Hi @agazso , I was trying to make a user story out of this, and it actually strikes me that I couldn't find a real reason why a user would want this, apart from "it would be nice to have this". I also believe this myself. Actually, I realized this is due to not understanding why similar content would have a different hash.
> Would you agree that this "not understanding" is the real reason? Do you think that this issue can also be solved with a proper reasoning as to why the address of similar content uploaded via the `dirs` api is different?

[...]

PR #1065, opened 2020-12-18
Random manifest obfuscation key only when encrypting

zbiljic:
relates to: #1006 

Change behavior for upload to `/dirs` endpoint so that non-encrypted uploads will force `mantaray`-based manifest not to use random obfuscation key for each upload. This means that for multiple uploads of the same content to have same root hash will be generated.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/ethereum/go-ethereum v1.9.20
 	github.com/ethersphere/bmt v0.1.4
 	github.com/ethersphere/langos v1.0.0
-	github.com/ethersphere/manifest v0.3.6-0.20201219094757-ca4bdf6230ce
+	github.com/ethersphere/manifest v0.3.6-0.20201219111500-ac38cbddff4d
 	github.com/ethersphere/sw3-bindings/v2 v2.1.0
 	github.com/gogo/protobuf v1.3.1
 	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e // indirect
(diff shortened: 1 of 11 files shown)


zelig:
@zbiljic  @Eknir 
no need to do this. 
the obfuscation should only be used whe the manifest is encrypted.

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/ethereum/go-ethereum v1.9.20
 	github.com/ethersphere/bmt v0.1.4
 	github.com/ethersphere/langos v1.0.0
-	github.com/ethersphere/manifest v0.3.5
+	github.com/ethersphere/manifest v0.3.6-0.20201219111500-ac38cbddff4d
```
acud, line 13: tidy? should probably be just `v0.3.6` no?

acud (approved):
Looks good :beer:

zelig (approved):

go.mod
zbiljic: @acud Yes, before this PR can be merged, I am waiting on [ethersphere/manifest#18](https://github.com/ethersphere/manifest/pull/18) to me merged before, after which I planed to tag to `0.3.6`.

zbiljic: @acud I have updated this PR now to use `v0.3.6` version of the `github.com/ethersphere/manifest` library. Could you check it again?

zelig:
this should now be merged i  guess

zbiljic:
> this should now be merged i guess

Yes, after I merged manifest library, I updated here, and asked @acud to confirm (as he made comment about version).

Issue #945, opened 2020-11-16
Tackle technical debt related to traversal services


zbiljic:
# Task

Some changes introduced with #924 were noted as technical debt, and should be fixed in the future.
Few of the issues expressed were:
* all simplified, some parts move to mantaray
* adapt to context passed in loadsaver
* no stop condition, but errgroup with context
* `processChunkAddresses` should return an error
* use context, check. ctx.done at beginning
* no need to recursively pass errgroup (this was copied from previously existing code)
* no need for `sync.WaitGroup` (was added to make processing synchronous)
* use `errgroup.Group` with context for the for loop
* rename function to `Walk???Async` or `Each??Async`

For more detailed comments, refer the the PR comment [here](https://github.com/ethersphere/bee/pull/924#pullrequestreview-530576490).

PR #1066, opened 2020-12-19
Return error from address iteration function

zbiljic:
related to: #945 

Tackle some technical depth (details in the issue).

### DIFF
--- a/pkg/api/pin_bytes_test.go
+++ b/pkg/api/pin_bytes_test.go
@@ -123,6 +123,7 @@ func TestPinBytesHandler(t *testing.T) {
 		)
 
 		hashes := []string{rootHash, data1Hash, data2Hash}
+		sort.Strings(hashes)
 
 		expectedResponse := api.ListPinnedChunksResponse{
 			Chunks: []api.PinnedChunk{},
--- a/pkg/api/pin_bytes_test.go
+++ b/pkg/api/pin_bytes_test.go
@@ -6,6 +6,7 @@ package api_test
 
 import (
 	"bytes"
+	"encoding/json"
 	"io/ioutil"
 	"net/http"
 	"sort"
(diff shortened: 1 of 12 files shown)


pkg/api/pin_bytes_test.go
```diff
@@ -123,21 +124,41 @@ func TestPinBytesHandler(t *testing.T) {
...
-			expectedResponse.Chunks = append(expectedResponse.Chunks, api.PinnedChunk{
-				Address:    swarm.MustParseHexAddress(h),
-				PinCounter: 1,
-			})
-		}
+		var respBytes []byte
 
 		jsonhttptest.Request(t, client, http.MethodGet, pinChunksResource, http.StatusOK,
-			jsonhttptest.WithExpectedJSONResponse(expectedResponse),
+			jsonhttptest.WithPutResponseBody(&respBytes),
```
acud, line 134: unless you need the actual bytes for something, you can just use `jsonhttptest.WithUnmarshalJSONResponse` and unmarshal directly to `api.ListPinnedChunksResponse`

acud (approved):
just a minor comment about the usage of `jsonhttptest`, otherwise LGTM

pkg/manifest/mantaray.go
```diff
@@ -138,22 +138,20 @@ func (m *mantarayManifest) IterateAddresses(ctx context.Context, fn swarm.Addres
...
 		if node != nil {
-			var stop bool
-
 			if node.Reference() != nil {
 				ref := swarm.NewAddress(node.Reference())
 
-				stop = fn(ref)
-				if stop {
-					return errStopIterator
+				err = fn(ref)
```
zelig, line 144: are these two conditions not mutually exclusive?
if they are you can simplify the code

zelig (approved):

pkg/manifest/mantaray.go
zbiljic: They are not.
First one (`node.Reference()`) contains address to mantaray trie node, and second (`node.Entry()`) for the reference to the actual thing that was stored on that node (if path contains some value, i.e. file).

pkg/file/joiner/joiner.go
```diff
@@ -213,37 +213,36 @@ func (j *joiner) Seek(offset int64, whence int) (int64, error) {
...
+	}
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
 	}
 
+	eg, ectx := errgroup.WithContext(ctx)
+
```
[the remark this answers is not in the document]
zbiljic: No really, no. However I did have discussion with @acud, and decision was to leave the WaitGroup for now because it currently makes code more linear, and easier to follow (and I agree with decision). Maybe once we have some additional tests we can make the switch.

acud (approved):

PR #1068, opened 2020-12-22
kademlia: limit bootnode connection attempts

acud:
our staging cluster is constantly trying to connect to itself, and because attempt counts are not limited the kademlia iterator never returns from the `connectBootnodes` call.

we probably have other connectivity problems, but we don't have enough visibility on what's going on due to insufficient logging which is mended in this PR.

Another problem is that nodes may get their own address from the DNS discovery and try to connect to themselves. I therefore added a check to `parseCheckAck` to check that a node does not try to connect to itself. This verification is done in the Ack, when a node sends his overlay address.

### DIFF
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -250,12 +250,11 @@ func (k *Kad) connectBootnodes(ctx context.Context) {
 
 		if _, err := p2p.Discover(ctx, addr, func(addr ma.Multiaddr) (stop bool, err error) {
 			k.logger.Tracef("connecting to bootnode %s", addr)
-			attempts++
 			if attempts >= maxBootnodeAttempts {
 				return true, nil
 			}
-
 			bzzAddress, err := k.p2p.Connect(ctx, addr)
+			attempts++
 			if err != nil {
 				if !errors.Is(err, p2p.ErrAlreadyConnected) {
 					k.logger.Debugf("connect fail %s: %v", addr, err)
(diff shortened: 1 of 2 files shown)


acud:
2020-12-22-103934_1920x1080_scrot

pkg/kademlia/kademlia.go
```diff
@@ -238,31 +240,38 @@ func (k *Kad) Start(ctx context.Context) error {
...
+
 	for _, addr := range k.bootnodes {
-		if count >= 3 {
+		if attempts >= totalAttempts || connected >= 3 {
 			return
 		}
 
 		if _, err := p2p.Discover(ctx, addr, func(addr ma.Multiaddr) (stop bool, err error) {
 			k.logger.Tracef("connecting to bootnode %s", addr)
+			if attempts >= maxBootnodeAttempts {
```
zbiljic, line 253: In the case where we have a problem with `bootnodes`, and some "fresh" node cannot connect to any, it will still end up in a loop, trying to connect to some boot node. This is done by the `kademlia.manage` loop, which is fine, because if there the node is not connected to any peer, it should try to connect to some.

acud: yes, that is correct, and the connection will be tried periodically. i don't see a reason why we should keep on trying over and over again (without any delay either). also, the other problem is that even if some other node dialed in to the node which is trying to connect to the bootnodes (and for some reason cannot), will never go out of the bootnodes loop, although some other peers may already be connected.

zbiljic (approved):

janos (approved):

PR #1046, opened 2020-12-11
batch service

acud:
2 in 1 for better flexibility

PR #1069, opened 2020-12-22
storage incentives: batch service and backing store

acud:
#1046 #922 and parts of #934


acud:
closed in favor of #1070

Issue #1054, opened 2020-12-15
api request logline should have traceid


acud:
we are currently logging HTTP requests on bee, but the loglines do not include the trace information. this could be helpful to filter logs and correlate traces to errors that we see coming from a related request.

cc @janos WDYT?

janos:
In order for log lines to have trace id, is must exist in the net/http handler. Currently tracing is on the lower p2p level, but adding traces to the http level can be added with opentracing HTTPHeadersCarrier extraction into the request context in one http middleware that could be set for every request. If the tracing span is not in request headers, a new one can be created there. The basic logic for the middleware function is in this example https://github.com/opentracing/opentracing-go#deserializing-from-the-wire.

Logging calls in http handlers have to be changed use a logger with the trace id field set, similarly as it is done on the p2p layer.

Eknir:
Added label RFC. While I don't get the specifics of this issue, I'm all in for improving our instrumentation and debug possibilities. These are IMO always a priority, especially if you feel the real need to have them during a debug session.

Eknir:
@acud , is this issue ready for implementation?

acud:
yeah so I think that we can do this. The task is to add a middleware that will try to get a `Swarm-TraceId` header from the http request, if it is not found, the middleware should inject a new opentracing span into the context of the request.

I'm tagging https://github.com/ethersphere/bee/issues/724, since we could have this done in a way which could allow other tracing capabilities which are more granular than what we have today. So, perhaps have the tracer to inject onto different carriers, rather than just to a `p2p.Header` object. This is discussed in #724

PR #1071, opened 2020-12-24
Support tracing in HTTP headers

zbiljic:
relates to: #1054

### DIFF
--- a/pkg/tracing/tracing.go
+++ b/pkg/tracing/tracing.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"errors"
 	"io"
+	"net/http"
 	"time"
 
 	"github.com/ethersphere/bee/pkg/logging"
@@ -34,6 +35,14 @@ type contextKey struct{}
 
 // LogField is the key in log message field that holds tracing id value.
 const LogField = "traceid"
+
+const (
+	// TraceContextHeaderName is the http header name used to propagate tracing context.
+	TraceContextHeaderName = "trace-id"
+
+	// TraceBaggageHeaderPrefix is the prefix for http headers used to propagate baggage.
+	TraceBaggageHeaderPrefix = "ctx-"
+)
 
 // Tracer connect to a tracing server and handles tracing spans and contexts
 // by using opentracing Tracer.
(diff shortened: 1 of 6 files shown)


acud:
@agazso @nugaon @vojtechsimetka once this is merged you could use the tracing header to continue traces initialized in external contexts (i.e. javascript) into bee and subsequently see the whole cascade across the network using jaeger. fyi :)

acud (approved):
awesome work @zbiljic

pkg/tracing/tracing.go
```diff
@@ -35,6 +36,14 @@ type contextKey struct{}
 // LogField is the key in log message field that holds tracing id value.
 const LogField = "traceid"
 
+const (
+	// TraceContextHeaderName is the http header name used to propagate tracing context.
+	TraceContextHeaderName = "trace-id"
```
acud, line 41: @janos @agazso @vojtechsimetka @nugaon do you have any preference for the header name? should we go with `Trace-Id`? `Swarm-Trace-Id`? any other ideas?

agazso: I like `swarm-trace-id` for the consistency with other custom headers.

vojtechsimetka: 👍

janos: `swarm-trace-id` is good as it is analogue to default value `uber-trace-id` https://github.com/jaegertracing/jaeger-client-go/blob/master/constants.go#L58 and it has to be in lower case as per comment https://github.com/jaegertracing/jaeger-client-go/blob/master/header.go#L33.

agazso: @janos I didn't know it has to be lowercase, but in the HTTP standard the header name is case-insensitive so I just prefer to refer to header names in lowercase :)

https://tools.ietf.org/html/rfc7230#section-3.2

janos: @agazso I did not know either, I've just fount this int the comments in the library that is used and found this interesting fact. I thought that it would be good to share it here mainly because of the case-insensitivity you mentioned. It looks like just a specific requirement for this particular library.

acud: very strange :)

zbiljic: Yes, the library requires lowercase values.

zbiljic:
Pushed change with updated header name.

janos (approved):
Thanks @zbiljic. Looks very nice.

PR #1095, opened 2021-01-08
update libp2p version

acud:


acud:
```
=== RUN   TestHeaders_empty
    headers_test.go:100: send headers: read message: context deadline exceeded
--- FAIL: TestHeaders_empty (10.48s)
```

acud:
```
== RUN   TestBlocklisting
time="2021-01-09T10:48:15Z" level=info msg="successfully connected to peer (outbound) [Overlay: 62c5ccd15c43131270ce44678074c7852662e6daa9a84c7093e7b04fcc5f1aeb, Underlay: /ip4/10.1.0.4/tcp/46335/p2p/QmVGTF26x84HSjhyV9HCg3GUvLeYJyttUHz6K49JkWXgV2]"
time="2021-01-09T10:48:15Z" level=info msg="successfully connected to peer (inbound) [Overlay: a50db6bd6ff4b94630e82edabd3ac9a41d8a8597af7ba9ee17af80d5e1e7dee6, Underlay: /ip4/10.1.0.4/tcp/36107/p2p/Qmempp5cT2xk68yuSeqNGqYSwWReAj6ujQMTETJhvY4nf8]"
time="2021-01-09T10:48:16Z" level=error msg="blocked connection from blocklisted peer QmVGTF26x84HSjhyV9HCg3GUvLeYJyttUHz6K49JkWXgV2"
time="2021-01-09T10:48:16Z" level=debug msg="handshake: could not close stream Qmempp5cT2xk68yuSeqNGqYSwWReAj6ujQMTETJhvY4nf8: stream reset"
    connections_test.go:349: got peers 1, want 0
--- FAIL: TestBlocklisting (2.57s)
```

pkg/p2p/libp2p/headers.go
```diff
@@ -23,12 +23,11 @@ func sendHeaders(ctx context.Context, headers p2p.Headers, stream *stream) error
 	defer cancel()
 
 	if err := w.WriteMsgWithContext(ctx, headersP2PToPB(headers)); err != nil {
-		return fmt.Errorf("write message: %w", err)
+		return fmt.Errorf("send write message: %w", err)
```
janos, line 26: I am not sure why these log messages are changed.

pkg/p2p/libp2p/stream.go
```diff
@@ -29,5 +32,28 @@ func (s *stream) Headers() p2p.Headers {
 }
 
 func (s *stream) FullClose() error {
-	return helpers.FullClose(s)
+	if err := s.CloseWrite(); err != nil {
+		_ = s.Reset()
+		return err
+	}
+	// So we don't wait forever
+	_ = s.SetDeadline(time.Now().Add(5 * time.Second))
```
janos, line 40: This is too short timeout. I would suggest to have it 60s as it was in earlier versions of helper package, maybe to lower it to 30s, as this deadline is the worst case scenario, and 5s is pretty short.

janos (commented):
AutoNAT functionality works as expected. 👍

pkg/p2p/libp2p/headers.go
acud: when debugging the tests the logged messages were exactly the same (both messages are the same for `sendHeader` and `handleHeaders`).

janos: This should have been annotated by the caller function. The consequence is that the log message in production will contain redundant message part `handle`, for example for handleHeaders here https://github.com/ethersphere/bee/blob/9c5bf9c2ea2908c767b7dde9a1c052b14b09e268/pkg/p2p/libp2p/libp2p.go#L342. If the problem is with the tests, it should be improved there, this way it has the impact on the clarity of log messages in production. Returned error annotations should not reference the function that is returned them, that is the responsibility of the caller function.

acud: cool. removed :+1:

pkg/p2p/libp2p/libp2p.go
```diff
@@ -176,11 +176,7 @@ func New(ctx context.Context, signer beecrypto.Signer, networkID uint64, overlay
 	// If you want to help other peers to figure out if they are behind
 	// NATs, you can launch the server-side of AutoNAT too (AutoRelay
 	// already runs the client)
-	if _, err = autonat.NewAutoNATService(ctx, h,
-		// Support same non default security and transport options as
-		// original host.
-		append(transports, security)...,
-	); err != nil {
+	if _, err = autonat.New(ctx, h); err != nil {
```
zbiljic, line 179: This change seem to completely remove AutoNAT service.

If we were to update library version, while keeping "existing" behavior, we would maybe have to have change like this:

```go
// Support same non default security and transport options as
// original host.
dialer, err := libp2p.New(ctx, append(transports, security)...)
if err != nil {
	return nil, err
}

// If you want to help other peers to figure out if they are behind
// NATs, you can launch the server-side of AutoNAT too (AutoRelay
// already runs the client)
if _, err = autonat.New(ctx, h, autonat.EnableService(dialer.Network())); err != nil {
	return nil, fmt.Errorf("autonat: %w", err)
}
```

acud: Thanks @zbiljic for pointing this out today. This indeed seems like the right solution, however I am wondering if we need a clean way to close the autonat instance and whether we should perform cleanups on shutdown on that `dialer` libp2p instance. I wonder, if this has to be a libp2p instance at all...

acud: I pushed these changes @zbiljic

pkg/p2p/libp2p/stream.go
```diff
@@ -29,5 +36,29 @@ func (s *stream) Headers() p2p.Headers {
 }
 
 func (s *stream) FullClose() error {
-	return helpers.FullClose(s)
+	if err := s.CloseWrite(); err != nil {
```
zbiljic, line 39: While this change is extracted from previous library version, and same as before, I would posit that we actually no longer need this "hack".

As far as I could see from `github.com/libp2p/go-libp2p-core` release notes for `v0.7.0` version, previously the `Close()` function closed streams for writing, but left them open for reading. Because of this, they had "helper" function that handled fully closing the stream (to prevent leaking streams).

Since we actually do not handle anything special, and we just want to close the stream, instead of all this code, it might me enough here simply to call `Close`. In that case `Close` will not wait for any form of acknowledgment, and should free the stream object (I have not gone too deep into this).

Even with this said, I think it might be better to leave it as it was before in this PR, and have some next PR where this change will be done.

zbiljic (approved):

pkg/p2p/libp2p/headers.go
```diff
@@ -25,7 +25,6 @@ func sendHeaders(ctx context.Context, headers p2p.Headers, stream *stream) error
 	if err := w.WriteMsgWithContext(ctx, headersP2PToPB(headers)); err != nil {
 		return fmt.Errorf("write message: %w", err)
 	}
-
```
zbiljic, line 28: Maybe remove this changed line, as it is nothing.

pkg/p2p/libp2p/stream.go
acud: The way we use `FullClose` actually needs the acknowledgement, since it synchronizes the flow control in certain situations, like in the handshake protocol.

janos (approved):

Issue #1087, opened 2021-01-05
add uncashed amount to some cheque or cashout endpoint


ralph-pichler:
# Task

Currently computing which cheques are uncashed (and to which amount) requires using two different endpoints and some client side computation. Exposing the uncashed amount as part of one of the endpoints can significantly simplify cashout scripts.

Eknir:
Thanks Ralph,
I think this makes sense.

Eknir:
For reference: current cashout script that could be simplified: https://gist.githubusercontent.com/ralph-pichler/3b5ccd7a5c5cd0500e6428752b37e975/raw/7ba05095e0836735f4a648aefe52c584e18e065f/cashout.sh

Eknir:
@paxthemax , this could be something

PR #1103, opened 2021-01-12
crypto: add deterministic signer test

acud:
This PR adds a deterministically generated signature, so that other language implementations can use it as a reference for implementation.

cc @agazso


janos (approved):

ralph-pichler (approved):

porn35:
#1087

PR #1109, opened 2021-01-13
event listener: sync forever

acud:
naive implementation to keep on syncing forever


ralph-pichler (approved):
looks good as a first step

paxthemax (approved):
As @ralph-pichler mentioned, good as a first step, ok to merge in that context.

PR #1111, opened 2021-01-14
make value the normalised balance

ralph-pichler:


paxthemax:
As far as I can tell, this is what it says on the tin - you removed the `NormalisedBalance` field and added new semantics to `Value`?

paxthemax (approved):
Small change, looks OK, good to merge :+1:

acud (approved):

PR #1114, opened 2021-01-15
wire up postage stamp syncing

ralph-pichler:
* add flags for contract addresses
* this reuses swap endpoint which means swap endpoint is required even if swap is off
* adds updateBlockNumber to event updater
* node.go wiring

### DIFF
--- a/pkg/postage/listener/listener.go
+++ b/pkg/postage/listener/listener.go
@@ -178,6 +178,16 @@ func (l *listener) sync(from uint64, updater postage.EventUpdater) error {
 			return err
 		}
 
+		if to < from {
+			// if the blockNumber is actually less than what we already, it might mean the backend is not synced or some reorg scenario
+			continue
+		}
+
+		if to < tailSize {
+			// in a test blockchain there might be not be enough blocks yet
+			continue
+		}
+
 		// consider to-tailSize as the "latest" block we need to sync to
 		to = to - tailSize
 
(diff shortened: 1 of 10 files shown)


pkg/postage/listener/listener.go
```diff
@@ -198,6 +208,11 @@ func (l *listener) sync(from uint64, updater postage.EventUpdater) error {
 			}
 		}
 
+		err = updater.UpdateBlockNumber(to)
```
acud, line 211: I think that moving the block height to the actual individual event signatures on `EventUpdater` and using the `Log.BlockNumber` (and thus updating the chain state on every write in batch service) would be more consistent. Then we can avoid this method altogether. WDYT?

acud (approved):
LGTM

pkg/postage/listener/listener.go
ralph-pichler: the advantage of this approach is that the block number is only updated after the entire block has been processed. whereas if its on individual events it's already updated after the first event. if the bee node stops/crashes mid-processing a block it will reprocess some events but that should be fine as all of them should be idempotent (at least better than skipping anything).

pkg/postage/listener/listener_test.go
```diff
@@ -155,9 +157,10 @@ func TestListener(t *testing.T) {
 				depthIncrease.toLog(),
 				priceUpdate.toLog(),
 			),
+			WithBlockNumber(blockNumber),
 		)
-		listener := listener.New(logging.New(ioutil.Discard, 0), mf, postageStampAddress, priceOracleAddress)
-		listener.Listen(0, ev)
+		l := listener.New(logging.New(ioutil.Discard, 0), mf, postageStampAddress, priceOracleAddress)
```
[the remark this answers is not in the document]
ralph-pichler: fixed

zelig (approved):

paxthemax (approved):
LGTM

PR #1142, opened 2021-01-21
postage: add create endpoint

ralph-pichler:
adds the `/stamps/{amount}/{depth}` endpoint where amount is the initial balance per chunk.

related changes:
* chainId is now always queried when not running in standalone. even with swap disabled swap-endpoint is required.
* swap is always disabled in standalone, even if swap is explicitly enabled

open issues:
* currently postage stamps don't work at all with swap disabled as the token address is queried through the swap factory. this is only an issue if not standalone and `--swap-enable=false`.
* package naming could be improved
* bucketDepth is set incorrectly


pkg/api/postage.go
```diff
@@ -0,0 +1,39 @@
+package api
```
acud, line 1: add code copyright

pkg/api/postage.go
```diff
@@ -0,0 +1,39 @@
...
+	amount, ok := big.NewInt(0).SetString(mux.Vars(r)["amount"], 10)
+	if !ok {
+		jsonhttp.BadRequest(w, "invalid postage amount")
+		s.Logger.Error("api: invalid postage amount")
+		return
+	}
+
+	depth, err := strconv.ParseUint(depthStr, 10, 8)
+	if err != nil {
+		jsonhttp.InternalServerError(w, "invalid depth")
```
acud, line 25: bad request

pkg/postage/postagecontract/contract.go
```diff
@@ -0,0 +1,187 @@
+package postagecontract
```
acud, line 1: code copy missing

pkg/postage/postagecontract/contract.go
```diff
@@ -0,0 +1,187 @@
...
+			var createdEvent batchCreatedEvent
+			err = parseEvent(&postageStampABI, "BatchCreated", &createdEvent, *ev)
+			if err != nil {
+				return nil, err
+			}
+
+			batchID := createdEvent.BatchId[:]
+
+			c.postageService.Add(postage.NewStampIssuer(
+				"label",
```
acud, line 144: @zelig please elaborate what these are for and if we should even set them now

pkg/postage/postagecontract/contract.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/ethersphere/bee/pkg/settlement/swap/transaction"
+	"github.com/ethersphere/sw3-bindings/v2/simpleswapfactory"
+)
+
+var (
+	postageStampABI   = parseABI(listener.PostageStampABIJSON)
+	erc20ABI          = parseABI(simpleswapfactory.ERC20ABI)
+	batchCreatedTopic = postageStampABI.Events["BatchCreated"].ID
+
+	ErrBatchCreationFailed = errors.New("batch creation failed")
```
acud, line 25: ErrBatchCreate` probably enough

pkg/postage/postagecontract/contract.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+type Interface interface {
```
acud, line 28: single method interfaces in go are usually called after the name of the function + `er`.
`io.Reader`, `Closer`, `LoadSaver`, etc... so `BatchCreater`  (i know, its a typo)

pkg/postage/postagecontract/contract.go
acud: also, i think @zelig would probably want to call this package `subscription` or something like that

acud (commented):
minor comments. otherwise looks good

acud (commented):
one more thing - we should have an api test that covers this new endpoint - injecting a mock Interface and asserting it was called (or not) given different parameters.

pkg/postage/postagecontract/contract.go
ralph-pichler: my idea of this package was more to put all of the postage stamp (write) code there, as the code for topup and alike will also look similar and require the same helper code. then the interface would not be a single method anymore.

acud: ok

pkg/postage/postagecontract/contract.go
ralph-pichler: changed

pkg/api/postage.go
ralph-pichler: added to all new files

pkg/api/postage.go
ralph-pichler: fixed

ralph-pichler:
@acud basic api test added. which cases should we test here?

pkg/api/postage.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/gorilla/mux"
+)
+
+func (s *server) postageCreateHandler(w http.ResponseWriter, r *http.Request) {
+	depthStr := mux.Vars(r)["depth"]
+
+	amount, ok := big.NewInt(0).SetString(mux.Vars(r)["amount"], 10)
+	if !ok {
+		jsonhttp.BadRequest(w, "invalid postage amount")
```
acud, line 22: it should be:
```go
s.Logger.Debugf("create batch: parse amount: %v",err)
s.Logger.Error("create batch: invalid amount")
jsonhttp.BadRequest(w, "invalid postage amount")
return
```

pkg/api/postage.go
```diff
@@ -0,0 +1,43 @@
...
+		s.Logger.Error("api: invalid postage amount")
+		return
+	}
+
+	depth, err := strconv.ParseUint(depthStr, 10, 8)
+	if err != nil {
+		jsonhttp.InternalServerError(w, "invalid depth")
+		s.Logger.Error("api: invalid depth")
```
acud, line 30: see logging pattern above ^^

pkg/api/postage.go
```diff
@@ -0,0 +1,43 @@
...
+		return
+	}
+
+	batchID, err := s.postageContract.CreateBatch(r.Context(), amount, uint8(depth))
+	if err != nil {
+		jsonhttp.InternalServerError(w, "cannot create batch")
+		s.Logger.Error("api: cannot create batch")
```
acud, line 37: see logging pattern above ^^. first logging (on the two levels, debug first), then return status code and return

pkg/api/postage_test.go
```diff
@@ -0,0 +1,46 @@
...
+			}
+			return batchId, nil
+		}),
+	)
+	createBatch := func(amount int64, depth uint8) string { return fmt.Sprintf("/stamps/%d/%d", amount, depth) }
+	client, _, _ := newTestServer(t, testServerOptions{
+		PostageContract: contract,
+	})
+
+	jsonhttptest.Request(t, client, http.MethodPost, createBatch(initialBalance, depth), http.StatusOK,
```
acud, line 41: consider other cases:
- send depth as some wrong input (`abcabc`) and make sure you get a 400
- send balance as such and make sure you get a 400
- mock an error that is returned from the contract mock - what happens for example when you have insufficient funds to do the txn? what happens if the txn fails because not enough gas?

pkg/api/postage.go
ralph-pichler: what if there is no error like in this case. should there still be two log lines?

pkg/api/postage.go
ralph-pichler: fixed

pkg/api/postage.go
ralph-pichler: fixed

pkg/api/postage.go
```diff
@@ -0,0 +1,43 @@
...
+		s.Logger.Debugf("api: cannot create batch: %v", err)
+		return
+	}
+
+	jsonhttp.OK(w, hex.EncodeToString(batchID))
```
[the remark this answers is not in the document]
ralph-pichler: fixed

pkg/api/postage_test.go
ralph-pichler: more tests added

pkg/api/postage.go
acud: sorry, yeah in this case just an error would probably do :+1:

pkg/api/postage.go
```diff
@@ -0,0 +1,50 @@
...
+		s.Logger.Error("create batch: invalid amount")
+		jsonhttp.BadRequest(w, "invalid postage amount")
+		return
+
+	}
+
+	depth, err := strconv.ParseUint(depthStr, 10, 8)
+	if err != nil {
+		s.Logger.Debugf("create batch: invalid depth: %v", err)
+		s.Logger.Error("create batch: invalid amount")
```
acud, line 34: this should be `invalid depth` (says `amount`)

pkg/node/node.go
```diff
@@ -171,6 +172,14 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey,
 			return nil, err
 		}
 
+		chainID, err = swapBackend.ChainID(p2pCtx)
+		if err != nil {
+			logger.Infof("could not connect to backend at %v. A working blockchain node (for goerli network in production) is required. Check your node or specify another node using --swap-endpoint.", o.SwapEndpoint)
```
acud, line 177: not sure about using `Info` here. maybe `Warn` or even `Error`? also, the message says `"could not connect to backend"` but could it not be something else?

pkg/node/node.go
```diff
@@ -216,22 +225,24 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey,
...
+	batchStore := batchstore.New(stateStore)
+
+	post := postage.NewService(stateStore, chainID.Int64())
+
+	// this is needed until postage API gets wired in (since our actual API
+	// falls back to a 32 byte slice of zeros as batch id
+	fallbackBatch := make([]byte, 32)
+	post.Add(postage.NewStampIssuer("empty batch", "", fallbackBatch, 32, 8))
 
+	var postageContractService postagecontract.Interface
```
zelig, line 244: some TLC to naming here please.. i got no better idea though

pkg/postage/postagecontract/contract.go
zelig: not this one, no

pkg/postage/postagecontract/contract.go
```diff
@@ -0,0 +1,191 @@
...
+func New(
+	owner common.Address,
+	postageStampAddress common.Address,
+	bzzTokenAddress common.Address,
+	transactionService transaction.Service,
+	postageService postage.Service,
+) Interface {
+	return &postageContract{
+		owner:               owner,
+		postageStampAddress: postageStampAddress,
```
zelig, line 53: this is the contract address no? please do not call it stamp address

pkg/node/node.go
```diff
@@ -216,22 +225,24 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey,
...
+	if !o.Standalone {
```
zelig, line 245: postageStampAddress  is a misnomer below

pkg/postage/postagecontract/contract.go
```diff
@@ -0,0 +1,191 @@
...
+			return createdEvent.BatchId[:], nil
+		}
+	}
+
+	return nil, ErrBatchCreate
+}
+
+type batchCreatedEvent struct {
+	BatchId           [32]byte
+	TotalAmount       *big.Int
```
zelig, line 164: lets have one value called `Value` ok?

pkg/postage/postagecontract/contract_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"fmt"
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	postageMock "github.com/ethersphere/bee/pkg/postage/mock"
+	"github.com/ethersphere/bee/pkg/postage/postagecontract"
+	"github.com/ethersphere/bee/pkg/settlement/swap/transaction"
+	transactionMock "github.com/ethersphere/bee/pkg/settlement/swap/transaction/mock"
```
zelig, line 20: time to move transaction pkg out of swap

zelig (approved):

pkg/node/node.go
```diff
@@ -216,22 +225,24 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey,
...
-		if err != nil {
-			logger.Infof("could not connect to backend at %v. A working blockchain node (for goerli network in production) is required. Check your node or specify another node using --swap-endpoint.", o.SwapEndpoint)
-			return nil, fmt.Errorf("could not get chain id from ethereum backend: %w", err)
-		}
+	batchStore := batchstore.New(stateStore)
+
+	post := postage.NewService(stateStore, chainID.Int64())
+
+	// this is needed until postage API gets wired in (since our actual API
+	// falls back to a 32 byte slice of zeros as batch id
```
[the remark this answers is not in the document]
ralph-pichler: not sure about this. this was already here previously, I just moved it around. @acud is this still needed?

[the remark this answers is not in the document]
acud: no :+1:

pkg/node/node.go
ralph-pichler: not sure why this was not `Error` before, fixed now. There should not be any other error (if there were it would be printed anyway due to the returned error in the next line). The reason why it explicitly mentions the connection here is because bee will otherwise just terminate with `Error: could not get chain id from ethereum backend: Post "http://localhost:8545": dial tcp [::1]:8545: connect: connection refused` which could be confusing to users.

pkg/api/postage.go
ralph-pichler: fixed

pkg/node/node.go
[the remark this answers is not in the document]
ralph-pichler: removed this. but what does that mean for integration tests? is there a need to call the create endpoint on a node before doing any uploads?

pkg/postage/postagecontract/contract.go
ralph-pichler: these names need to exactly match the contract names as the parser uses reflection underneath.

pkg/postage/postagecontract/contract_test.go
ralph-pichler: agreed, but let's please do code moves in a separate PR after this

pkg/postage/postagecontract/contract.go
zelig: yes, please set them via query param and/or form-data .
They are meeant to support 2 dimensional classification of stamps. 
The UX presented in the book can then be simply supported from js side, 
https://gateway.ethswarm.org/bzz/latest.bookofswarm.eth/#subsection.6.1.3 
with `label` being generated from:

```
8\minimal (a few hours)– Useful for immediate delivery of pss messages or singleowner chunks part of ephemeral chat or other temporary files.
•temporary (week)– Files or mailboxed messages meant not to be stored long butto be picked up by third parties asynchronously.
•long term (year)– Default long term storage.
•forever (10years)– Important content not to be lost/forgotten;  to survive thegrowth  of  the  network  and  subsequent  increase  of  batch  depth  even  if  theuploader remains completely offline.
```

and batch reuse strategy maybe

"keyid" can be recorded as the batch owner etheraddress that would be used to call the appropriate signer with the right batch owner key.

ralph-pichler: `label` query param added. `keyID` is hex representation of owner address.

pkg/node/node.go
ralph-pichler: the contract is actually called `PostageStamp`. For now I went with `postageContractAddress`. We should also change the flag which currently is `--postage-stamp-address` but we should reach a decision on naming first as changing that also means changing the helm chart.

pkg/postage/postagecontract/contract.go
ralph-pichler: see https://github.com/ethersphere/bee/pull/1142/files#r564473904

acud (approved):
LGTM

PR #1143, opened 2021-01-22
api: split chunk and soc api

acud:


acud:
@zelig would need your input on this since I need to start adjusting the integration tests for this change to be merged

pkg/api/chunk.go
```diff
@@ -68,22 +62,38 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
...
 
-	chunk := swarm.NewChunk(address, data)
-	if !content.Valid(chunk) {
-		if !soc.Valid(chunk) {
-			s.Logger.Debugf("chunk upload: invalid chunk: %s", address)
-			s.Logger.Error("chunk upload: invalid chunk")
-			jsonhttp.BadRequest(w, nil)
-			return
-		}
+	if len(data) < swarm.SpanSize {
```
agazso, line 71: This would allow zero payload length chunks when the span is all zero. Before this change this was not allowed, minimum payload size was 1 (therefore the minimum data size was `SpanSize + 1`). I don't have a strong opinion about this just wanted to mention.

acud: yeah this is indeed a problem. It is not clearly defined whether a chunk with span `0` and `nil` data is a valid chunk or not

pkg/api/api_test.go
```diff
@@ -88,6 +89,23 @@ func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.
 	return httpClient, conn, ts.Listener.Addr().String()
 }
 
+func request(t *testing.T, client *http.Client, method, resource string, body io.Reader, responseCode int) *http.Response {
```
zelig, line 92: `checkResponseCode` or  something better

pkg/api/chunk.go
zelig: no the span can be arbitrary metadata really, it only matters for swarm file chunks.
if the query  paramater is not given, it falls back to datalength by  default

pkg/api/router.go
```diff
@@ -71,11 +71,21 @@ func (s *server) setupRouting() {
...
+			jsonhttp.NewMaxBodyBytesHandler(swarm.ChunkWithSpanSize),
+			web.FinalHandlerFunc(s.chunkUploadHandler),
+		),
+	})
+
 	handle(router, "/chunks/{addr}", jsonhttp.MethodHandler{
 		"GET": http.HandlerFunc(s.chunkGetHandler),
+	})
+
+	handle(router, "/soc/{owner}/{id}/{sig}", jsonhttp.MethodHandler{
```
zelig, line 85: sig should be query parameter. if missing we fall back on calling  the signer

pkg/api/soc.go
```diff
@@ -0,0 +1,112 @@
...
+		return
+	}
+	id, err := hex.DecodeString(mux.Vars(r)["id"])
+	if err != nil {
+		s.Logger.Debugf("soc upload: bad id: %v", err)
+		s.Logger.Error("soc upload: bad id")
+		jsonhttp.BadRequest(w, "bad id")
+		return
+	}
+	sig, err := hex.DecodeString(mux.Vars(r)["sig"])
```
zelig, line 38: query param, see above

pkg/api/soc.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		s.Logger.Debugf("soc upload: read chunk data error: %v", err)
+		s.Logger.Error("soc upload: read chunk data error")
+		jsonhttp.InternalServerError(w, "cannot read chunk data")
+		return
+	}
+	if len(data) < swarm.SpanSize {
+		s.Logger.Debugf("soc upload: chunk data too short")
+		s.Logger.Error("soc upload: chunk data")
+		jsonhttp.BadRequest(w, "short chunk data")
```
zelig, line 59: https://gateway.ethswarm.org/bzz/latest.bookofswarm.eth/#subsection.10.2.1

201   CreatedReference in response body
400  Bad request     owner, id or span is not well formed
401   Unauthorized signing fails
404  Not found  no owner keypair is found
413  Payload Too Large Payload size exceeds span value or 4096
422  Unprocessable entity    owner/id pair already exists

pkg/api/soc.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		if jsonhttp.HandleBodyReadError(err, w) {
+			return
+		}
+		s.Logger.Debugf("soc upload: read chunk data error: %v", err)
+		s.Logger.Error("soc upload: read chunk data error")
+		jsonhttp.InternalServerError(w, "cannot read chunk data")
+		return
+	}
+	if len(data) < swarm.SpanSize {
```
zelig, line 56: span should be query param

pkg/api/soc.go
```diff
@@ -0,0 +1,112 @@
...
+		s.Logger.Debugf("soc upload: chunk data too short")
+		s.Logger.Error("soc upload: chunk data")
+		jsonhttp.BadRequest(w, "short chunk data")
+		return
+	}
+
+	ch, err := chunk(data)
```
zelig, line 63: the length of data is not checked. it must be <=4096 
we dont necessarily require span length <4096 to match data - we could allow arbitrary metadata

pkg/api/soc.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		s.Logger.Debugf("soc upload: create content addressed chunk: %v", err)
+		s.Logger.Error("soc upload: chunk data error")
+		jsonhttp.BadRequest(w, "chunk data error")
+		return
+	}
+
+	chunk, err := soc.NewSignedChunk(id, ch, owner, sig)
```
zelig, line 71: if no sig given we should fall back to calling the signer

pkg/api/soc.go
```diff
@@ -0,0 +1,112 @@
...
+	if !soc.Valid(chunk) {
+		s.Logger.Debugf("soc upload: invalid chunk: %v", err)
+		s.Logger.Error("soc upload: invalid chunk")
+		jsonhttp.BadRequest(w, "invalid chunk")
+		return
+
+	}
+	ctx := r.Context()
+
+	_, err = s.Storer.Put(ctx, requestModePut(r), chunk)
```
zelig, line 88: why request mode? it should be `uploadModePut`

pkg/api/soc_test.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+
+		if !bytes.Equal(ch.Data(), data) {
+			t.Fatal("data retrieved doesnt match uploaded content")
+		}
+	})
+}
+
+// returns a valid, mocked SOC
+func mockSoc(t *testing.T) (*soc.Soc, []byte, []byte) {
```
zelig, line 135: why do we need this, use the soc pkg constructors, no need to reimplement it here

zelig (changes requested):

pkg/api/soc.go
acud: gets the `ModePut` for this request. therefore `requestModePut`

pkg/api/chunk.go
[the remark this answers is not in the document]
acud: if span can be arbitrary metadata why do we want to complicate things by coercing the caller to encode it in a different format than the actual chunk data? wouldn't it be just easier and clearer to prepend it to the data, just in the same manner that we do not make a distinction between chunk span and chunk data in the actual chunk data type?

pkg/api/api_test.go
acud: not just, since it also returns a `Response`

pkg/api/soc.go
acud: it is not optional. why QP?

pkg/api/chunk.go
zelig: > yeah this is indeed a problem. It is not clearly defined whether a chunk with span `0` and `nil` data is a valid chunk or not

true, I think it is valid according to definition but i would not allow 0  length content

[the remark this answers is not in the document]
zelig: so be it

pkg/api/api_test.go
zelig: `testedRequest`?

pkg/api/soc.go
zelig: whatevs have it in the data then

pkg/api/soc.go
zelig: no  longer checked then

pkg/api/soc.go
zelig: @acud ^

pkg/api/soc.go
zelig: no no, `requestModePut` is for chunks that arrive as response to retrieve requests.  this must be `uploadModePut`

pkg/api/soc_test.go
acud: they dont provide the necessary functionality

pkg/api/soc.go
```diff
@@ -16,26 +17,39 @@ import (
...
 		s.Logger.Debugf("soc upload: bad id: %v", err)
-		s.Logger.Error("soc upload: bad id")
+		s.Logger.Error("soc upload: %v", errBadRequestParams)
 		jsonhttp.BadRequest(w, "bad id")
 		return
 	}
-	sig, err := hex.DecodeString(mux.Vars(r)["sig"])
+
+	sigStr := r.URL.Query().Get("sig")
+	if sigStr == "" {
```
zelig, line 45: TODO: fallback to bee signer

pkg/api/soc.go
```diff
@@ -16,26 +17,39 @@ import (
...
+		s.Logger.Debugf("soc upload: empty signature")
+		s.Logger.Error("soc upload: bad signature")
```
zelig, line 47: bad  -> empty?

pkg/api/soc.go
```diff
@@ -53,13 +67,21 @@ func (s *server) socSigUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+
 	if len(data) < swarm.SpanSize {
 		s.Logger.Debugf("soc upload: chunk data too short")
-		s.Logger.Error("soc upload: chunk data")
+		s.Logger.Error("soc upload: %v", errBadRequestParams)
 		jsonhttp.BadRequest(w, "short chunk data")
 		return
 	}
 
+	if len(data) > swarm.ChunkSize {
```
zelig, line 78: wait, now we said Spanbytes is part  of the  data, then you need to use a different constant and amend comment 4096 --> 4104

zelig (approved):
LGTM needing one correction when checking the size

PR #1145, opened 2021-01-22
make logs more user friendly

Eknir:
- Changed warning message `cannot connect to peer` to make users not freak about about this
- Removed redundant information from connected to peer info message and moved that to debug
- some other improvements to logging


acud:
some unrelated files are changed

Eknir:
> some unrelated files are changed

Should be better now. Thx

pkg/p2p/libp2p/libp2p.go
```diff
@@ -304,7 +304,9 @@ func New(ctx context.Context, signer beecrypto.Signer, networkID uint64, overlay
 		}
 
 		s.metrics.HandledStreamCount.Inc()
-		s.logger.Infof("successfully connected to peer (inbound) %s", i.BzzAddress.ShortString())
+		s.logger.Debugf("successfully connected to peer (inbound) %s", i.BzzAddress.ShortString())
```
acud, line 307: i don't understand why we need this change. what information is this providing when running in less than Debug loglevel? I prefer to revert this

pkg/p2p/libp2p/libp2p.go
```diff
@@ -531,7 +533,8 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (address *bzz.
 	s.protocolsmu.RUnlock()
 
 	s.metrics.CreatedConnectionCount.Inc()
-	s.logger.Infof("successfully connected to peer (outbound) %s", i.BzzAddress.ShortString())
+	s.logger.Debugf("successfully connected to peer (outbound) %s", i.BzzAddress.ShortString())
+	s.logger.Infof("successfully connected to peer (outbound)")
```
acud, line 537: also

pkg/kademlia/kademlia.go
```diff
@@ -340,6 +340,7 @@ func recalcDepth(peers *pslice.PSlice) uint8 {
 // connect connects to a peer and gossips its address to our connected peers,
 // as well as sends the peers we are connected to to the newly connected peer
 func (k *Kad) connect(ctx context.Context, peer swarm.Address, ma ma.Multiaddr, po uint8) error {
+	k.logger.Infof("attempting to connect to peer %s", peer)
```
acud, line 343: there is already a logline relating to this in `kademlia.go#L170`

pkg/p2p/libp2p/libp2p.go
Eknir: I believe that this log line, together with the inbound equivalent is telling a lot. It tells the user that the node is being connected through outbound and inbound connection requests. 

These particular changes were motivated by me trying to let the log lines tell a story to the average early-adopter user that is understandable, but without any redundant detail (which should be in debug) . The previous log line contained, in my opinion, a lot of information that is of no use for telling a story to the user:
` successfully connected to peer (outbound) [Overlay: 7f7ea3501344f7312e1fa80dc57279b483138a6667b9d1af3748056176d9bbfc, Underlay: /ip4/3.93.49.53/tcp/31024/p2p/16Uiu2HAmM7LrsvEnaQHTKkf9wKc2bEPE2TF59JKPJKBqPJxLxoSt]`

This is why I removed the redundant information and moved it to debug. I can, however, see value in logging the overlay only.

`successfully connected to peer 7f7ea3501344f7312e1fa80dc57279b483138a6667b9d1af3748056176d9bbfc (outbound)`

pkg/kademlia/kademlia.go
Eknir: Perfect. So you suggest to remove this log and use the log on line 170? I would, however, prefer to use my wording, as this contains less jargon and is therefore easier to understand for the average early-adopter user. Besides, I would use the info log level, as this is an important action that tells the user a story on what the node is doing.

Eknir: I had a look at it, and I would suggest to leave my log line and remove the debug log line on `170`. The reason for this is that `func (k *Kad) connect(ctx context.Context, peer swarm.Address, ma ma.Multiaddr, po uint8)` may also be called from another place and you want to have the info in the function itself to always have the correct info.

pkg/p2p/libp2p/libp2p.go
acud: saying that a peer was dialed to without the overlay does not give much information, as you might disconnect from that peer and retry and connect again, so two lines without much information will be logged for the same peer.

pkg/kademlia/kademlia.go
acud: Please change the logline in `L170` to display the information you log here. I don't mind changing the loglevel. Why do we need to log the same information twice?

> may also be called from another place

but it is not called from anywhere else.

either remove `L170` or change it please

pkg/p2p/libp2p/libp2p.go
Eknir: I already changed this per your suggestion it and added the overlay address (see most recent change: https://github.com/ethersphere/bee/pull/1145/commits/df416faeca92026191ccffdd7d18863789fd353f )

pkg/kademlia/kademlia.go
Eknir: > either remove L170 or change it please

￼This was done here: https://github.com/ethersphere/bee/pull/1145/commits/865f64f733c32a8f268a537d770cffc664c0301d

pkg/kademlia/kademlia.go
```diff
@@ -177,8 +175,8 @@ func (k *Kad) manage() {
 							k.logger.Debugf("could not remove peer from addressbook: %s", peer.String())
 						}
 					}
-					k.logger.Debugf("could not connect to peer from kademlia %s: %v", bzzAddr.String(), err)
-					k.logger.Warningf("could not connect to peer")
+					k.logger.Debugf("peer not reachable from kademlia not reachable %s: %v", bzzAddr.String(), err)
```
acud, line 178: this message says: `peer not reachable from kademlia not reachable: ....`. maybe `peer not reachable from kademlia %s: %v"` is enough

acud (approved):
minor cleanup needed on one of the messages. otherwise good

Issue #1147, opened 2021-01-22
Simplify tags API


zbiljic:
# Task

Current `/tags/{id}` API returns response object which contains internal tag state, and was shown that it constantly confuses users. We should simplify this response to return only data that are relevant to the user. If possible we should make response usable to track actual "upload" process for the chunks (that can be used to track progress) to the network.

related: #918

## Acceptance criterea

Have update tags API which is more understandable to the users.

Eknir:
Working on documenting this now in our bee docs

Issue #1084, opened 2021-01-05
API Stability


Eknir:
# User story

## Story

As a consumer of the Swarm APIs, I want to have a reasonable certainty that they won't change anymore or be prepared when they change, such that I can build products on top of Swarm and/or don't need to relearn again when the API endpoints changed

## Acceptance criteria

- All envisioned changes to the API endpoints which can be done without touching the architecture of Bee are done
- For all API endpoints, we explicitly mention the stability in the API docs (https://nodejs.org/dist/latest-v6.x/docs/api/all.html#documentation_stability_index)
- A blog is written, detailing all our API endpoints 

## Tasks

| Index | Task | reasoning |  Done (1), or 0 |
|-------- | ------ |------------| -----|
| 0 |  POST bzz: return 201 instead of 200 | fits better with [http response codes definitions](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) | 1 | 
| 1 |  POST bzz: swarm-encrypt to swarm-encryption | more proper | 1 | 
| 2 |  POST bzz: swarm-tag-uid to swarm-tag | shorter | 1 | 
| 3 |  POST bzz: Content-Type (by `multipart/form-data` request) should be defined as `multipart/form-data; boundary=...` . Other file attributes can be placed in Content-Disposition header | 0 |
| 4 | GET  bzz/{reference}: use 408 request timeout error when request times out | more in line with semantic error code definitions | 0 | 
| 6 |  GET  bzz/{reference}: use error code 416 when applicable | when possible, use semantic error codes | 0 | 
| 7 |  GET tags?offset={offset}&length={length}: this is not implemented yet, but should be implemented | mentioned in BOS. There might be usecases and easy to implement. Requested by #985 | 0 | 
| 8 |  issue #1078 | improve consistency | 1 | 
| 9 | add 404 to API response code for GET tags/{id} | verify code and API doc congruence | 1 |
| 10 | document the PATCH tags functionality in API docs and in the book | | 0 |
| 11 | add chunk estimation to tar reader when uploading directories and do final chunk accounting after each file in the tar has been split + manifests | | 0 |

Eknir:
Proposed API changes: https://docs.google.com/spreadsheets/d/1JE4wY0i9ASt4oRiO7Q0D7qfb70ZF4Ds5MjWbbQ-8UtU/edit#gid=1710849048

Eknir:
Related: #1113

Eknir:
[...]

Issue #1051, opened 2020-12-15
Reupload to network


Eknir:
# Story
- As a content uploader, I want to be able to instruct my node to re-upload the same content, such that I have another chance of syncing the content if this was not successful before
- Also a user story to prepare for #852
- Re uploading the same chunk twice is also a way how postage stamps can propagate to the nearest-neighborhood

# Acceptance criteria
- Uploading one chunk for two times results in the nearest neighborhood receiving the chunk two times 

# Background
This issue was created when planning #852. It was mentioned by @acud that merely re uploading to the network is not even possible. It was also mentioned that this issue has the potential of simplifying the design of the local store.

Eknir:
@acud , I copied this issue to our backlog, as it is blocking #852. The issue doesn't contain enough information, however. Could you please provide this?

ldeffenb:
Bullet one has "change" instead of "chance"?
Bullet three is incomplete:  "how postage stamps can..."  what?

Eknir:
> Bullet one has "change" instead of "chance"?
> Bullet three is incomplete: "how postage stamps can..." what?

Thanks. Adjusted

significance:
+1 please 😀

Eknir:
Completed per #1705 .

@acud , I would appreciate if we can go over the change you made, so I can see if it fulfills the acceptance criteria

Issue #985, opened 2020-11-19
Query existing tags?


ldeffenb:
#### Summary
I've been creating tags and uploading files with those tag UIDs, but is there a way to query what tags actually exist?  If my upload process doesn't complete cleanly, I'd need to be able to query the tags to delete them as no longer needed.

And can tag creation specify a name so that my client can recognize a queried tag as one that it created?

#### Steps to reproduce
1) Create a tag
2) Upload a file using that tag
3) loop 1 and 2 for a while resulting in lots of existing tags

How to know what tags were created in the event that the 1/2 loop fails?

#### Expected behavior
Ability to query existing tags so that they can be checked and/or cleaned up (deleted).

#### Actual behavior
Created tag UIDs seem to be invisible once created.

Eknir:
@ldeffenb , you must have been creating a lot of tags. 
Just so that I understand your issue clearly;

First issue: the probability of collision in the createTag function is big. Solution: increase the length of the tag.
Second issue: No possibility to query existing tags. You would like to call to the bee node and request all existing tag-uids, potentially filtering on name

Eknir:
This one is completed: https://docs.ethswarm.org/api/#tag/Tag/paths/~1tags/get

PR #1148, opened 2021-01-22
Simplify public tags API response

zbiljic:
relates to: #1147

Changes to public `/tags/{id}` API.

The current tags API response returns following object:
```json
{
  "uid": 1066766601,
  "name": "unnamed_tag_1611322188",
  "startedAt": "2021-01-22T14:29:48+01:00",
  "address": "<hash>",
  "total": 26,
  "split": 26,
  "seen": 4,
  "stored": 26,
  "sent": 33,
  "synced": 22
}
```

Suggested change to this response would produce following object:
```json
{
  "uid": 1066766601,
  "startedAt": "2021-01-22T14:29:48+01:00",
  "total": 26,    // the total number of chunks for upload(s) related with this tag (estimated until tag is done)
  "processed": 26,    // the total number of chunks stored and queued for sending
  "synced": 26    // the total number of chunks that are synced with the network (now of before)
}
```

Idea behind this is to have 2 names that would represent number of chunks that are directly uploaded through request and the number of chunks that have completed syncing to the network (are were already present, i.e. already pushed previously).
In this change the `synced` is no longer actual value from the internal tag value, but represents `seen` + `synced` count.

I have considered several different combinations for these 2 properties. Any suggestion for changes (that may be more user-friendly) are welcome.

In addition to change to public API, original tag response is added to Debug API.

For the public API, the "address" is no longer returned. The address will be populated in the case where tag is automatically created by the system, during upload to `/bytes`, `/files`, or `/dirs` endpoint. (In this same case, after upload is completed (but may not yet be synced), the `split` counter will be copied into `total` and address added.)

The "name" has been completely removed from the tag.

Update (2021-01-29):
- changed "Swarm-Tag-Uid" header name to "Swarm-Tag"
- added endpoint to public API, `GET /tags`, which can be used to list tags from the node
  - this endpoint accepts `offset` and `limit`
  - since the tags exists in-memory until they are "done", and then in the statestore after, the listing first returns entries from memory, and then from statestore (ignoring ones already returned from memory)
    - addition of new tags while using `offset` will mess up ordering, and some may be skipped or returned more than once, but this limitation is because of the ID type that was chosen for the tags

Update (2021-02-02):
- renamed `stored` response field to `processed`
- returned `total` field in response
- the `total` field (for manually created tags) will always be populated with calculated number of chunks for upload
  - this only works in the case when tag is manually created and used for upload
  - for bytes content just from `Content-Length` header
  - for files based on calculation from `Content-Length` header with `2` additional chunks, for file metadata content, and for collection entry for those two addresses
  - for dirs upload for each file content from TAR archive calculated based on file size with 2 additional (for metadata and collection entry, as well with additional chunks produced in the process of saving manifest content with 2 additional (for manifest file metadata and collection entry)

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -132,9 +132,7 @@ func (s *server) Close() error {
 func (s *server) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
 	// if tag ID is not supplied, create a new tag
 	if tagUid == "" {
-		tagName := fmt.Sprintf("unnamed_tag_%d", time.Now().Unix())
-		var err error
-		tag, err := s.Tags.Create(tagName, 0)
+		tag, err := s.Tags.Create(0)
 		if err != nil {
 			return nil, false, fmt.Errorf("cannot create tag: %w", err)
 		}
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -30,7 +30,7 @@ import (
 
 const (
 	SwarmPinHeader           = "Swarm-Pin"
-	SwarmTagUidHeader        = "Swarm-Tag-Uid"
+	SwarmTagHeader           = "Swarm-Tag"
 	SwarmEncryptHeader       = "Swarm-Encrypt"
 	SwarmIndexDocumentHeader = "Swarm-Index-Document"
 	SwarmErrorDocumentHeader = "Swarm-Error-Document"
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -295,3 +295,10 @@ func calculateNumberOfChunks(contentLength int64, isEncrypted bool) int64 {
 
 	return int64(totalChunks) + 1
 }
+
+func requestCalculateNumberOfChunks(r *http.Request) int64 {
+	if !strings.Contains(r.Header.Get(contentTypeHeader), "multipart") && r.ContentLength > 0 {
+		return calculateNumberOfChunks(r.ContentLength, requestEncrypt(r))
+	}
+	return 0
+}
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -17,6 +17,7 @@ import (
 	"time"
 	"unicode/utf8"
 
+	"github.com/ethersphere/bee/pkg/encryption"
 	"github.com/ethersphere/bee/pkg/file/pipeline/builder"
 	"github.com/ethersphere/bee/pkg/logging"
 	m "github.com/ethersphere/bee/pkg/metrics"
@@ -276,15 +277,15 @@ func requestPipelineFn(s storage.Storer, r *http.Request) pipelineFunc {
 // calculateNumberOfChunks calculates the number of chunks in an arbitrary
 // content length.
 func calculateNumberOfChunks(contentLength int64, isEncrypted bool) int64 {
-	if contentLength < 4096 {
+	if contentLength < swarm.ChunkSize {
 		return 1
 	}
-	branchingFactor := 128
+	branchingFactor := swarm.Branches
 	if isEncrypted {
-		branchingFactor = 64
+		branchingFactor = encryption.EncryptionBranches
 	}
 
-	dataChunks := math.Ceil(float64(contentLength) / float64(4096))
+	dataChunks := math.Ceil(float64(contentLength) / float64(swarm.ChunkSize))
 	totalChunks := dataChunks
 	intermediate := dataChunks / float64(branchingFactor)
 
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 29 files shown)


zelig:
why are we eroding functionality instead of fixing  whats needed and making use of it.
most of  all how do we want   progress bar without total count?

@Eknir afair what we agreed on with  @janos was
- we should stop persisting tags
- weshould providemore  correct result for tar stream 
- provide `retries` on top of `sent`

as for this pr, i  dont mind simplifying output but at least  keep `split` and `sent` as well as `total`. 
the js team should  really build

Eknir (approved):
I just had a conversation with Viktor about this issue and we agreed that the simplification, as proposed by here, is a good one. Removing the `total` counter means that we won't be able to estimate the ETA of splitting, but given that we needed to make changes on how we treat total (we should estimate the total before we start splitting in order for this number to be useful), it is logical to remove it. 

> Removing the `total` counter means that we won't be able to estimate the ETA of splitting

On a second thought, I think that even this user story is satisfied, as the process that created the tag and ordered the upload can just do the estimation on how many chunks are expected and then compare `stored` and `synced` with this information. No need to keep track of `total` inside the `bee` process.

Based on my testing experience, the user story is satisfied, while significantly simplifying the interface of this API. Great work! 
The possibility to debug with the tags API is still possible, because you replaced the "old" functionality to the debugAPI. 

I have two requests:
1. Could you please update the API endpoint `swarm-tag-uid` to `-swarm-tag` (see https://github.com/ethersphere/bee/issues/1084 point 2)
2. Could you please have a look if it is trivial to add the functionality to query all existing tags (https://github.com/ethersphere/bee/issues/985) and BOS (prt screen below). If trivial, it would be great if you can add it. If not trivial, Please leave a comment under https://github.com/ethersphere/bee/issues/1084 with your findings.

image

acud (approved):
LGTM. Thanks @zbiljic!

zelig:
Lets make a compromise:
- not to change the semantics of what stored means (number of chunks locally stored)
- report count synced to include seen 
- report count stored to include seen also 
- keep total, split, stored (the  original) and synced - this would also reduce the amount of work otherwise generated for     
      - documentation, api spec, and the book.
      - earlier implementations of progress bars etc.
- the rest of the PR is fine.
- ASAP in  another  PR we fix the total count set by http server based on Content-Length header as per the old swarm: https://github.com/ethersphere/swarm/blob/master/api/http/middleware.go#L123

zbiljic:
@Eknir, @acud, @zelig Please note that I have updated this PR with changes requested by @Eknir.
This includes changing `swarm-tag` header name, as well as adding API endpoint for listing tags (this one might be strange as we have tags in-memory map until tag is saved).

zbiljic:
This came up after I wrote my comment. To comment on it:

> * not to change the semantics of what stored means (number of chunks locally stored)

semantics of `stored` is not changed

> * report count synced to include seen

with these changes `synced` == `synced + seen`

> * report count stored to include seen also

not sure what this means

> * keep total, split, stored (the  original) and synced - this would also reduce the amount of work otherwise generated for
>   - documentation, api spec, and the book.
>   - earlier implementations of progress bars etc.

so basically comment is that we should just remove `seen` (i.e. move in into `synced`), and leave everything else as-is ?

> * ASAP in  another  PR we fix the total count set by http server based on Content-Length header as per the old swarm: https://github.com/ethersphere/swarm/blob/master/api/http/middleware.go#L123

we can create separate issue for this

acud:
we had a chat about this on friday.
viktor suggested to go ahead with this but even reduce further the amount of internals we expose to the user and have:
- `total` - same as today, and will be adjusted by the issue viktor mentioned above, to pre-estimate the total according to content length
- `processed` - the amount of chunks we've processed locally (whatever that may be. but the count should be stored+seen)
- `synced` - the amount of chunks synced into the network (synced+seen)
I think that this gives the right amount of wiggle of exposing progress while not confusing the user with terms she does not understand. WDYT @zbiljic ? @zelig did I get it right?

zbiljic:
> * `processed` - the amount of chunks we've processed locally (whatever that may be. but the count should be stored+seen)

One note regarding this. Each chunk for an upload will increment `stored` variable. For chunks that were already uploaded at some previous date `seen` will be incremented. If we have variable that adds "stored+seen" we might end up with some number that is larger than total possible number of chunks.

acud:
I meant it in the sense that when we get the value to display to the user we just sum these two up. I did not mean to introduce a new field or to store a different value in the actual tag, but rather just calculate it on the fly from the value of the other counters.

zbiljic:
> I meant it in the sense that when we get the value to display to the user we just sum these two up. I did not mean to introduce a new field or to store a different value in the actual tag, but rather just calculate it on the fly from the value of the other counters.

My comment is that we CANNOT sum these two values because any `seen` is already in the `stored`.

acud:
gotcha. so return `stored`.

zelig (approved):
I approve this but 
-  I would really like `Total` added to the public API before the JS side implements  a progress bar with the suboptimal `Synced/Stored` ratio of completeness instead of `Synced/Total` if `Total>0` needed once we implement presetting `Total` based on `Content-Length` header value https://github.com/ethersphere/bee/issues/918. 
- Renaming public `Stored` to `Processed` would make the tag  response ready for the reupload functionality as well.  https://github.com/ethersphere/bee/issues/1051

Eknir:
This needs documentation. Whole `tags` page should be rewritten and openAPI spec should be updated if not done yet.

zbiljic:
> This needs documentation. Whole `tags` page should be rewritten and openAPI spec should be updated if not done yet.

Not sure what documentation is mentioned for this PR. The OpenAP spec with changes related to this PR are present.

zelig (approved):
amazing job! thank you

pkg/api/api.go
```diff
@@ -276,15 +277,15 @@ func requestPipelineFn(s storage.Storer, r *http.Request) pipelineFunc {
 // calculateNumberOfChunks calculates the number of chunks in an arbitrary
 // content length.
 func calculateNumberOfChunks(contentLength int64, isEncrypted bool) int64 {
-	if contentLength < 4096 {
+	if contentLength < swarm.ChunkSize {
```
zelig, line 280: should  this  not  be <=?

pkg/encryption/encryption.go
```diff
@@ -21,13 +21,19 @@ import (
...
+	"github.com/ethersphere/bee/pkg/swarm"
 )
 
 const (
 	KeyLength     = 32
 	ReferenceSize = 64
 )
 
+const (
+	EncryptionBranches = swarm.Branches / 2
```
zelig, line 34: this constant better in `swarm` package no?

zelig (approved):

pkg/api/api.go
zbiljic: This whole function was just copied from `swarm` project (with constants replaced).
Will change this.

pkg/encryption/encryption.go
zbiljic: That was my initial idea, but then I moved it here. Will move it back.

acud (approved):
Thank you for the very thorough work, as usual :pray:

PR #1150, opened 2021-01-23
more explicit wording that bee requires ETH and BZZ to start up

attila-lendvai:


acud (approved):

pkg/settlement/swap/chequebook/init.go
```diff
@@ -48,9 +48,9 @@ func checkBalance(
 		}
 
 		if balance.Cmp(big.NewInt(int64(swapInitialDeposit))) < 0 {
-			logger.Warningf("please make sure there is sufficient eth and bzz available on %x", overlayEthAddress)
+			logger.Warningf("with swap enabled, we cannot begin operation until there is sufficient ETH and BZZ available on %x", overlayEthAddress)
```
Eknir, line 51: I would go with my suggestion below, because we want to avoid user to think "Oh! I need to disable swap and then all works"
```suggestion
			logger.Warningf("we cannot begin operation until there is sufficient ETH and BZZ available on %x", overlayEthAddress)
```

attila-lendvai: @Eknir i updated the PR to reflect that.

acud (approved):

Eknir (approved):
LGTM! Thanks

PR #1151, opened 2021-01-24
add a warning message when the user is not using clef

Eknir:
to discourage the usage of bee without bee-clef

### DIFF
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -321,6 +321,7 @@ func (c *command) configureSigner(cmd *cobra.Command, logger logging.Logger) (co
 
 		logger.Infof("using swarm network address through clef: %s", address)
 	} else {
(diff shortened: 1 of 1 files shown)


cmd/bee/cmd/start.go
```diff
@@ -321,6 +321,7 @@ func (c *command) configureSigner(cmd *cobra.Command, logger logging.Logger) (co
 
 		logger.Infof("using swarm network address through clef: %s", address)
 	} else {
+		logger.Warning("clef is not enabled; portability and security of your keys is sub optimal")
```
acud, line 324: WDYM with `portability`? consider `clef is not enabled; we encourage using clef to ensure the safety of your keys`

Eknir: Thank you for the suggestion! With portability, I mean that using your swarm keys in any wallet that is not Swarm is difficult to achieve. This is solved when the user uses clef.

acud (approved):

PR #1154, opened 2021-01-25
wait for eth backend to be synced on startup

ralph-pichler:
waits for eth backend to be synced (latest block less than a minute in the past) on startup. this is necessary to ensure we don't reject valid chequebooks or the chequebook factory if running against an unsynced eth node (e.g. if goerli-geth was installed via dappnode at the same time as bee).


acud (approved):

acud:
@svetomir @ralph-pichler @vandot if we merge this to `master`, how will this affect our integration tests?

ralph-pichler:
@acud should not affect it at all. there is only a single node there and it is always considered synced as it regularly creates new blocks every few seconds.

metacertain (approved):

Issue #1074, opened 2020-12-25
Bee listens on localhost only for p2p on ipv6


d10r:
After starting bee with default options for the p2p address, I get this:
```
# netstat -tulpen | grep 1634
tcp        0      0 0.0.0.0:1634            0.0.0.0:*               LISTEN      1012       1013819126  4198/bee        
tcp6       0      0 ::1:1634                :::*                    LISTEN      1012       1013819127  4198/bee
```
While listening on the global interface for ipv4, it listens on ::1 (localhost) only for ipv6.
Shouldn't this instead be :::1634 for tcp6?

zbiljic:
This is correct, as can be seen in [pkg/p2p/libp2p/libp2p.go](https://github.com/ethersphere/bee/blob/207c03bcf542014b6c52a493b081e715e4749d7c/pkg/p2p/libp2p/libp2p.go#L87-L88):
```go
ip4Addr := "0.0.0.0"
ip6Addr := "::1"
```

We should change IPv6 address to `::` in other to have it same as for IPv4.
Although it would seem that even after that change the `libp2p` library again reports address in the format `/ip6/::1/tcp/7071/p2p/...`. It more looks like `libp2p` attempts to expand all addresses from "base" ones (at least that is what it shows).

PR #1167, opened 2021-01-27
IPv6 default host address be same as for IPv4

zbiljic:
relates to: #1074

In IPv6, the all-zeros address is represented by "`::`" (in short notation).


janos (approved):
Thanks!

acud (approved):

Issue #1053, opened 2020-12-15
debugAPI for blocklisted peers


Eknir:

Eknir:
@acud , did you update the openAPI spec?

PR #1171, opened 2021-01-28
libp2p, debugapi: add API to view blocklisted peers

acud:
adds a debugapi to view currently blocklisted peers. if a peer is in the blocklist but the duration has already expired, it will not show up in the endpoint. however, unlike in the `blocklist.Exists` method, the entry will not get deleted while querying this endpoint, in order to reduce side-effects.

ref https://github.com/ethersphere/bee/issues/1053

### DIFF
--- a/pkg/debugapi/router.go
+++ b/pkg/debugapi/router.go
@@ -68,6 +68,10 @@ func (s *server) setupRouting() {
 	router.Handle("/peers", jsonhttp.MethodHandler{
 		"GET": http.HandlerFunc(s.peersHandler),
 	})
+	router.Handle("/peers/blocklisted", jsonhttp.MethodHandler{
+		"GET": http.HandlerFunc(s.blocklistedPeersHandler),
+	})
+
 	router.Handle("/peers/{address}", jsonhttp.MethodHandler{
 		"DELETE": http.HandlerFunc(s.peerDisconnectHandler),
 	})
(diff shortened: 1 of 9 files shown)


ralph-pichler (approved):

pkg/debugapi/router.go
```diff
@@ -68,6 +68,10 @@ func (s *server) setupRouting() {
 	router.Handle("/peers", jsonhttp.MethodHandler{
 		"GET": http.HandlerFunc(s.peersHandler),
 	})
+	router.Handle("/peers/blocklisted", jsonhttp.MethodHandler{
```
zbiljic, line 71: This endpoint clashes with `"/peers/{address}"` (just bellow it).
While they do use different method, and this is not public API, I still think some other than `"/peers/blocklisted"` should be used (maybe just reverse words). At least to keep the API "clean" and discourage similar practice for the future.

zbiljic (changes requested):

pkg/debugapi/router.go
acud: Thanks @zbiljic. I renamed the endpoint to `/blocklist`

pkg/p2p/p2p.go
```diff
@@ -21,6 +21,7 @@ type Service interface {
 	Connect(ctx context.Context, addr ma.Multiaddr) (address *bzz.Address, err error)
 	Disconnecter
 	Peers() []Peer
+	BlocklistedPeers() ([]Peer, error)
```
zbiljic, line 24: It looks a bit strange that function for listing blocklisted peers returns an error, but list of peers does not.

zbiljic (approved):

pkg/p2p/p2p.go
acud: true. but `Peers` does not involve any storage operations, therefore does not return a possible error :+1:

Issue #1168, opened 2021-01-27
AvailableBalance metric


Eknir:
# User story

## Story

As a big node operator, I want to have metrics available that inform me about the availableBalance, such that I can generate a warning when my chequebooks are about to run out of funds

## Acceptance criteria

- There exists a metric, which is updated whenever the bee process is handing out a cheque (negative), cashing out (positive) or processing a withdrawal (negative). 

Question for above acceptance criteria: are there any more places where the bee process updates the available balance?

## Background

As a node operator, you don't want your nodes to run out of balance. There is currently no easy way to generate such warnings. Having a metric would help. @vandot and @ralph-pichler are aware of this task.
warn: Whenever there is a deposit (excluding initial deposit) or withdrawal from the Bee node that is not initiated by the bee process, this metric can be off. Please use the deposit/withdraw functions from the Bee node until this is resolved.

## Tasks

| Task | Assignee |  Done (1), or 0 |
| ------ |------------| -----|

Eknir:
Would it be an option to do a call to the chequebook whenever we are about to update the metric?

Eknir:
Documentation: in advanced section we might want to explain how node operators create warnings with this metric

Eknir:
@vandot , is this now set-up on our cluster?

PR #1172, opened 2021-01-28
settlement: add available balance metric

acud:
adds a metric to show available balance. it is updated only when we pay, but I think it might be necessary in other places too. @ralph-pichler @metacertain can you please advise?

#1168

### DIFF
--- a/pkg/settlement/swap/swap.go
+++ b/pkg/settlement/swap/swap.go
@@ -126,7 +126,7 @@ func (s *Service) Pay(ctx context.Context, peer swarm.Address, amount uint64) er
 		return err
 	}
 
-	s.metrics.AvailableBalance.Set(float64(balance.Int64()))
+	s.metrics.AvailableBalance.Set(big.NewFloat(0).SetInt(balance).Float64())
 	s.metrics.TotalSent.Add(float64(amount))
 	return nil
 }
--- a/pkg/settlement/swap/swap.go
+++ b/pkg/settlement/swap/swap.go
@@ -119,17 +119,13 @@ func (s *Service) Pay(ctx context.Context, peer swarm.Address, amount uint64) er
 		}
 		return ErrUnknownBeneficary
 	}
-	err = s.chequebook.Issue(ctx, beneficiary, big.NewInt(int64(amount)), func(signedCheque *chequebook.SignedCheque) error {
+	balance, err := s.chequebook.Issue(ctx, beneficiary, big.NewInt(int64(amount)), func(signedCheque *chequebook.SignedCheque) error {
 		return s.proto.EmitCheque(ctx, peer, signedCheque)
 	})
 	if err != nil {
 		return err
 	}
 
-	balance, err := s.chequebook.AvailableBalance(ctx)
-	if err != nil {
-		return err
-	}
 	s.metrics.AvailableBalance.Set(float64(balance.Int64()))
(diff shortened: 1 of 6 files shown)


pkg/settlement/swap/swap.go
```diff
@@ -125,6 +125,12 @@ func (s *Service) Pay(ctx context.Context, peer swarm.Address, amount uint64) er
 	if err != nil {
 		return err
 	}
+
+	balance, err := s.chequebook.AvailableBalance(ctx)
```
ralph-pichler, line 129: a peer's accounting lock is held during `Pay`. we should avoid adding additional blockchain calls here if possible.

ralph-pichler:
@acud to catch all relevant changes it would have to regularly fetch this in the background. if we only want to react to changes that are the result of our own nodes actions we should check after deposit / withdrawal / cashing as well. however the current code does not actually wait for any of these to terminate, so there currently is no place where you could check. maybe upon sending the transactions wait in a go routine, then check and update this afterwards.

pkg/settlement/swap/swap.go
acud: I'd like to ideally avoid the complexities of managing more goroutines. Would it be an idea to add the current available balance as a return value on the `unreserveTotalIssued` and then add it to the return value of `Issue`?

ralph-pichler: the call is in `reserve` not `unreserve`. it's also not the correct value but you should get the right one if you reduce it by `amount`.

acud: whoops sorry my mistake. so adding the return value to `reserve`?

ralph-pichler: yeah. actually it might also be fine (for now at least) to have this the only point where it is updated as this is the only place where low available balance actual matters (other than a withdraw call) and this will be called quite frequently so the metric should not get too old.

ralph-pichler (approved):
Not too pretty, but it seems to do the right thing. Maybe add some comments as to why `Issue` returns this (and that this is the availableBalance immediately after the cheque has been sent as this might not be obvious).

pkg/settlement/swap/swap.go
```diff
@@ -119,12 +119,14 @@ func (s *Service) Pay(ctx context.Context, peer swarm.Address, amount uint64) er
...
 	}
-	err = s.chequebook.Issue(ctx, beneficiary, big.NewInt(int64(amount)), func(signedCheque *chequebook.SignedCheque) error {
+	balance, err := s.chequebook.Issue(ctx, beneficiary, big.NewInt(int64(amount)), func(signedCheque *chequebook.SignedCheque) error {
 		return s.proto.EmitCheque(ctx, peer, signedCheque)
 	})
 	if err != nil {
 		return err
 	}
+
+	s.metrics.AvailableBalance.Set(float64(balance.Int64()))
```
ralph-pichler, line 129: note that `balance` does not necessarily fit into `int64`, especially with the new bzz token. so you might want to avoid the `int64` conversion step using `big.NewFloat(0).SetInt(balance).Float64()`.

metacertain (approved):

pkg/settlement/swap/swap.go
acud: nice catch. thanks

PR #1177, opened 2021-01-29
postage: check balance before attempting stamp creation

ralph-pichler:


pkg/api/postage.go
```diff
@@ -39,6 +41,12 @@ func (s *server) postageCreateHandler(w http.ResponseWriter, r *http.Request) {
 	label := r.URL.Query().Get("label")
 
 	batchID, err := s.postageContract.CreateBatch(r.Context(), amount, uint8(depth), label)
+	if errors.Is(err, postagecontract.ErrInsufficientFunds) {
```
acud, line 44: this should probably go into the `if err!=nil` block below

acud (approved):

PR #1179, opened 2021-01-29
api: use hex encoding in postage api

ralph-pichler:
api returned base64 encoded id instead of hex.

### DIFF
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -23,11 +23,6 @@ func (b batchID) MarshalJSON() ([]byte, error) {
 	return json.Marshal(hex.EncodeToString(b))
 }
 
-func (b batchID) UnmarshalJSON(d []byte) error {
-	_, err := hex.DecodeString(string(d))
-	return err
-}
-
 type postageCreateResponse struct {
 	BatchID batchID `json:"batchID"`
 }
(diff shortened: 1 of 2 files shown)


pkg/api/postage.go
```diff
@@ -10,11 +10,12 @@ import (
 	"strconv"
 
 	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/gorilla/mux"
 )
 
 type postageCreateResponse struct {
-	BatchID []byte `json:"batchID"`
+	BatchID swarm.Address `json:"batchID"`
```
acud, line 18: not sure we'd actually want to use this type here, strictly speaking, this isn't an `address`

ralph-pichler: i only used this because it has the right encoding. `[]byte` uses base64 unfortunately.

acud: `[]byte` and implement `json.Marshaler`?

acud (approved):

pkg/api/postage.go
ralph-pichler: you cannot implement anything on `[]byte`. I made a custom type for this field.

acud: i meant the define the JSON marshaler on `postageCreateResponse`

acud (approved):

Issue #756, opened 2020-09-26
Fix big.Int (incorrectly) marshals to a number instead of a string


paxthemax:
# Task
Currently, big.Int incorrectly maps to integer instead of string. This is a known "feature" of golang (see [here](https://github.com/golang/go/issues/28154)).

A possible fix is to wrap the big.Int into a type and marshal said type to a string.

## Acceptance criteria

Big int (or wrapped type) correctly marshals to string.

paxthemax:
This issue only affects `debugapi` for now.

PR #1187, opened 2021-02-02
swap, accounting: move to bigint

ralph-pichler:
* moves api function of settlement to `big.Int` from `uint64` as this might not work with high token amounts
* moves balances and alike in accounting to `big.Int`
* the `pseudosettle` and `swap` mock have been merged as they were basically the same thing
* options for thresholds and alike are now of type string

balances have been stored as `uint64` in the past. reading those past values as `big.Int` seems to work.


metacertain (approved):
massive change, LGTM !

pkg/accounting/accounting.go
```diff
@@ -168,35 +155,29 @@ func (a *Accounting) Reserve(ctx context.Context, peer swarm.Address, price uint
...
-	if err != nil {
-		return err
-	}
+	increasedExpectedDebt := new(big.Int).Add(expectedDebt, additionalDebt)
 
 	// If our expected debt is less than earlyPayment away from our payment threshold
 	// and we are actually in debt, trigger settlement.
 	// we pay early to avoid needlessly blocking request later when concurrent requests occur and we are already close to the payment threshold.
-	if increasedExpectedDebt >= int64(threshold) && currentBalance < 0 {
+	if increasedExpectedDebt.Cmp(threshold) >= 0 && currentBalance.Cmp(big.NewInt(0)) < 0 {
```
acud, line 167: i think that you could extract a package level `big.NewInt(0)` and reuse it in all of the comparators that check whether a value is smaller/bigger than zero

pkg/debugapi/balances.go
```diff
@@ -22,8 +23,8 @@ var (
 )
 
 type balanceResponse struct {
-	Peer    string `json:"peer"`
-	Balance int64  `json:"balance"`
+	Peer    string   `json:"peer"`
+	Balance *big.Int `json:"balance"`
```
acud, line 27: openapi spec should probably be modified, since this is a special data type and should be treated as a string natively in js i believe

pkg/settlement/pseudosettle/pseudosettle.go
```diff
@@ -128,7 +129,7 @@ func (s *Service) Pay(ctx context.Context, peer swarm.Address, amount uint64) er
 	s.logger.Tracef("sending payment message to peer %v of %d", peer, amount)
 	w := protobuf.NewWriter(stream)
 	err = w.WriteMsgWithContext(ctx, &pb.Payment{
-		Amount: amount,
+		Amount: amount.Uint64(),
```
acud, line 132: consider changing this protocol field for type safety.

acud (approved):
minor comments. otherwise lgtm

pkg/settlement/pseudosettle/pseudosettle.go
ralph-pichler: well, pseudosettle is only used for testing (and only if you run a cluster without swap). in such a setting this should never exceed uint64. anyway we can still do this at a later point.

pkg/debugapi/balances.go
ralph-pichler: in the json itself it is an integer (apparently integers in json don't have an upper bound). but I guess this might make problem with `JSON.parse` @paxthemax opened this issue back when we introduced `big.Int`s in debug api: https://github.com/ethersphere/bee/issues/756.

pkg/accounting/accounting.go
ralph-pichler: Will look into this in a follow up. This variable will be accessed from many threads but since all of them only read its probably fine.

zelig (approved):

PR #1189, opened 2021-02-02
OAS description updates

metacertain:
POST @ /soc
POST @ /chunk

debugapi / blocklist

### DIFF
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -168,8 +168,6 @@ paths:
                 $ref: 'SwarmCommon.yaml#/components/schemas/Status'
         '400':
           $ref: 'SwarmCommon.yaml#/components/responses/400'
-        '403':
-          $ref: 'SwarmCommon.yaml#/components/responses/403'
         '500':
           $ref: 'SwarmCommon.yaml#/components/responses/500'
         default:
@@ -845,6 +843,10 @@ paths:
             application/json:
               schema:
                 $ref: 'SwarmCommon.yaml#/components/schemas/ReferenceResponse'
+        '400':
+           $ref: 'SwarmCommon.yaml#/components/responses/400'
+        '401':
+           $ref: 'SwarmCommon.yaml#/components/responses/401'
         '500':
            $ref: 'SwarmCommon.yaml#/components/responses/500'
         default:
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -1,7 +1,7 @@
-openapi: 3.0.0
+openapi: 3.0.3
 
 info:
-  version: 0.4.0
+  version: 0.5.0
   title: Swarm API
   description: 'A list of the currently provided Interfaces to interact with the swarm, implementing file operations and sending messages'
 
@@ -134,6 +134,8 @@ paths:
           $ref: 'SwarmCommon.yaml#/components/responses/500'
         default:
           description: Default response
+
+  '/chunks':
     post:
       summary: 'Upload Chunk'
       tags:
@@ -151,12 +153,6 @@ paths:
             type: boolean
           required: false
           description: Represents the pinning state of the chunk
-        - in: path
-          name: reference
-          schema:
-            $ref: 'SwarmCommon.yaml#/components/schemas/SwarmReference'
-          required: true
-          description: Swarm address of chunk
       requestBody:
         content:
           application/octet-stream:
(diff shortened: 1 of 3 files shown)


openapi/SwarmCommon.yaml
```diff
@@ -138,7 +138,7 @@ components:
     DateTime:
       type: string
       format: date-time
-      pattern: '^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{7}\+\d{2}:\d{2})$'
+      #pattern: '^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{7}\+\d{2}:\d{2})$'
```
acud, line 141: not sure if this was intentional. uncomment?

openapi/SwarmCommon.yaml
```diff
@@ -1,6 +1,6 @@
-openapi: 3.0.0
+openapi: 3.0.3
```
acud, line 1: not sure if the patch update is needed

openapi/SwarmDebug.yaml
```diff
@@ -1,6 +1,6 @@
-openapi: 3.0.0
+openapi: 3.0.3
```
acud, line 1: do they all have to have the same numbers? @janos

acud (commented):
minor comments

openapi/SwarmDebug.yaml
metacertain: not necessary to update, just checked and this is the current version of openapi

openapi/SwarmCommon.yaml
metacertain: yes, intentional, the format and pattern used together caused a problem with latest openapi generator

openapi/Swarm.yaml
```diff
@@ -1,7 +1,7 @@
-openapi: 3.0.0
+openapi: 3.0.3
```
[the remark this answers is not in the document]
metacertain: this is the current version of the openapi specification standard.
our description version is below in info, increased to 0.5.0 with this update

openapi/SwarmCommon.yaml
acud: so maybe we can just keep the old version?

acud (approved):

openapi/Swarm.yaml
```diff
@@ -818,3 +812,43 @@ paths:
...
+  '/soc/{owner}/{id}':
+    post:
+      summary: Upload single owner chunk
+      tags:
+        - Single owner chunk
+      parameters:
+        - in: path
+          name: owner
+          schema:
+            $ref: 'SwarmCommon.yaml#/components/schemas/SwarmAddress'
```
acud, line 824: this is actually an ethereum address

openapi/SwarmCommon.yaml
metacertain: No we can't, our spec becomes useless with latest openapi generator if we do

openapi/Swarm.yaml
metacertain: thanks!

openapi/SwarmCommon.yaml
acud: i meant not to increment the open api version to `3.0.3`. the generator should then just use the old version. and then we don't need to comment this out. are we using any specific `3.0.3` spec features?

metacertain: ah the specification version increase was unrelated, this broke with the new version of the openapi generator (5.0.0) which must have been recently released, and this ecosystem is full of bugs and inconsistencies, but it's worth to keep checking against latest version of this tool as earlier versions become obsolete anyway

acud (approved):

openapi/Swarm.yaml
```diff
@@ -818,3 +812,43 @@ paths:
...
+          name: id
+          schema:
+            $ref: 'SwarmCommon.yaml#/components/schemas/HexString'
+          required: true
+          description: Id
+        - in: query
+          name: sig
+          schema:
+            $ref: 'SwarmCommon.yaml#/components/schemas/HexString'
+          required: true
```
zelig, line 837: TODO: when implementing falling back to bee signer, set required   to false

zelig (approved):

PR #1192, opened 2021-02-03
localstore: attach stamp to outgoing chunk

acud:
Fixes a bug where outgoing chunk on pushsync did not have stamp attached to it


ralph-pichler (approved):
LGTM

PR #1193, opened 2021-02-03
api, feeds: wire-in feeds resolution

acud:
This PR integrates the feeds package, providing API endpoints to lookup (and update) feeds.
Currently calling the bee  signer is not supported, so the update  endpoint needs  to  supply the signature.
However, for the signature, both the next update index and the wrapped chunk('s hash) are  needed.
so it can only be used after a lookup. 
The lookup endpoint returns a  response with header values for `swarm-feed-index` and `swarm-feed-index-next` specifying the 32-byte current and next index in hex encoding.

NOTE: this feature  is not implemented for epochs only sequence-indexing.

The PR also provides support for automatic feed resolution on the `bzz` endpoint.
This requires the feed attributes to  be encoded as entry metadata in a mantaray entry  on the '/' path.
This is a temporary workaround until mantaray is fixedd to support  entry on the empty path.
such a manifest is created by calling  the `feeds` POST endpoint

The API is  detailed in the opanapi spec


openapi/Swarm.yaml
```diff
@@ -818,3 +818,162 @@ paths:
...
+            $ref: 'SwarmCommon.yaml#/components/schemas/HexString'
+          required: false
+          description: Feed update index
+        - in: query
+          name: type
+          schema:
+            $ref: 'SwarmCommon.yaml#/components/schemas/FeedType'
+          required: false
+          description: Feed indexing scheme (default: sequence)
+        - in: query
```
agazso, line 889: Why do we need the signature for finding a feed update?

openapi/Swarm.yaml
```diff
@@ -818,3 +818,162 @@ paths:
...
+                $ref: 'SwarmCommon.yaml#/components/schemas/ReferenceResponse'
+        '400':
+           $ref: 'SwarmCommon.yaml#/components/responses/400'
+        '401':
+           $ref: 'SwarmCommon.yaml#/components/responses/401'
+        '500':
+           $ref: 'SwarmCommon.yaml#/components/responses/500'
+        default:
+          description: Default response
+    get:
```
agazso, line 860: The time (`at`) seems to be missing here if I'm understand correctly.

openapi/Swarm.yaml
zelig: this is messed up  yes, post params are under get.

`at` should be param in all 3 methods

openapi/Swarm.yaml
```diff
@@ -818,3 +818,162 @@ paths:
...
+          name: type
+          schema:
+            $ref: 'SwarmCommon.yaml#/components/schemas/FeedType'
+          required: false
+          description: Feed indexing scheme (default: sequence)
+        - in: query
+          name: sig
+          schema:
+            $ref: 'SwarmCommon.yaml#/components/schemas/HexString'
+          required: false
```
agazso, line 962: I think the signature should be required

acud: :+1: setting it to `true`. once we implement fallback to internal signer this can be switched off

agazso: Isn't it a security issue if the users of this endpoint can pass in any data and Bee signs it? Or is this functionality is going to be private somehow to trusted users?

pkg/api/feed.go
```diff
@@ -0,0 +1,117 @@
...
+			return
+		}
+	}
+
+	var at int64
+	atStr := r.URL.Query().Get("at")
+	if atStr != "" {
+		at, err = strconv.ParseInt(atStr, 10, 64)
+		if err != nil {
+			panic(err)
```
zelig, line 57: panics will be cleaned up yeah?

pkg/api/feed.go
```diff
@@ -0,0 +1,117 @@
...
+	ch, err := lookup.At(r.Context(), int64(at), 0)
+	if err != nil {
+		panic(err)
+	}
+
+	update := ch.Data()
+
+	// split the timestamp and reference
+	if len(update) < swarm.SpanSize*2 {
+		panic("malformed data")
```
zelig, line 77: why not explicitly require 8+8+32=48 or 8+8+64=80  bytes

pkg/api/feed_test.go
```diff
@@ -0,0 +1,383 @@
...
+
+		if h := respHeaders[api.SwarmFeedIndexHeader]; len(h) > 0 {
+			b, err := hex.DecodeString(h[0])
+			if err != nil {
+				t.Fatal(err)
+			}
+			if !bytes.Equal(b, timestampBytes) {
+				t.Fatalf("feed index header mismatch. got %v want %v", b, timestampBytes)
+			}
+		} else {
```
zelig, line 128: check if len(header) == 0  then no need for else

pkg/api/feed_test.go
```diff
@@ -0,0 +1,383 @@
...
+		return l.chunk, nil
+	}
+	fmt.Println(at, after, l.at, l.after)
+	if at == l.at && after == l.after {
+		return l.chunk, nil
+	}
+	return nil, errors.New("no feed update found")
+}
+
+func mockUpdate(t int64, a swarm.Address) swarm.Chunk {
```
zelig, line 339: no need to have a mock when the mock does what the real thing does. just use the feed toChunk method etc

pkg/api/feed_test.go
```diff
@@ -0,0 +1,383 @@
...
+	}
+	_, err = hasher.Write(b)
+	if err != nil {
+		panic(err)
+	}
+	s := swarm.NewAddress(hasher.Sum(nil))
+	return swarm.NewChunk(s, append(b, tb...))
+}
+
+func socChunk(t *testing.T, ch swarm.Chunk) (*soc.Soc, []byte, []byte) {
```
zelig, line 362: here too. lets not duplicate code

pkg/api/feed.go
acud: of course

pkg/feeds/epochs/finder.go
```diff
@@ -164,10 +165,14 @@ func (f *asyncFinder) at(ctx context.Context, at int64, p *path, e *epoch, c cha
 		}
 	}
 }
+func (f *asyncFinder) At(ctx context.Context, at, after int64) (swarm.Chunk, feeds.Index, feeds.Index, error) {
```
zelig, line 168: comment FIXME/TODO that this needs to be implemented

pkg/feeds/factory/factory.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
```
zelig, line 1: TODO: This needs to be implemented directly on the feed struct which should have an index type field.

pkg/feeds/getter.go
```diff
@@ -17,7 +17,7 @@ import (
 
 // Lookup is the interface for time based feed lookup
 type Lookup interface {
-	At(ctx context.Context, at, after int64) (swarm.Chunk, error)
+	At(ctx context.Context, at, after int64) (chunk swarm.Chunk, currentIndex, nextIndex Index, err error)
```
zelig, line 20: TODO: in fact At when used with noncurrent time should not return a next index.
only  latest  should..

zelig (approved):

pkg/api/bzz.go
```diff
@@ -69,6 +78,34 @@ func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	// there's a possible ambiguity here, right now the data which was
+	// read can be an entry.Entry or a mantaray feed manifest. Try to
+	// unmarshal as mantaray first and possibly resolve the feed, otherwise
+	// go on normally.
+	if !feedDereferenced {
+		if l, err := s.manifestFeed(ctx, ls, buf.Bytes()); err == nil {
+			//we have a feed manifest here
+			ch, cur, next, err := l.At(ctx, time.Now().Unix(), 0)
+			fmt.Println(cur, next)
```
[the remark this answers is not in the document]
acud: :+1: forgot about this. added the current index in the return headers. this was missing :+1:

pkg/feeds/factory/factory.go
acud: but then how do you inject the mock factory with the mock update?

openapi/Swarm.yaml
```diff
@@ -861,11 +861,107 @@ paths:
...
+          $ref: 'SwarmCommon.yaml#/components/responses/400'
         '401':
-           $ref: 'SwarmCommon.yaml#/components/responses/401'
+          $ref: 'SwarmCommon.yaml#/components/responses/401'
         '500':
            $ref: 'SwarmCommon.yaml#/components/responses/500'
         default:
           description: Default response
 
+  '/feed/{owner}/{topic}':
```
agazso, line 872: Signature seems to be missing.

acud: not needed for `POST` since no SOC is being created

agazso, line 872: In `router.go` the endpoint is called `/feeds`

https://github.com/ethersphere/bee/blob/aa61c6532d8b7c57c5776250c62264e333feef35/pkg/api/router.go#L92

zelig (approved):

PR #1203, opened 2021-02-06
check eth balance on initial startup and show required funds

ralph-pichler:
due to obtaining gETH and gBZZ no longer being the same operation, users might run into the scenario of having BZZ on their overlay but no ETH.

This PR
* checks there is sufficient ETH available on initial startup (previously only BZZ was checked).
* shows how much BZZ is needed (this might not be obvious to the user if the deposit amount is not specified manually)
* rounds up BZZ for the bazaar link if not an exact number of BZZ
* increases default deposit to 10 BZZ

### DIFF
--- a/pkg/settlement/swap/chequebook/init.go
+++ b/pkg/settlement/swap/chequebook/init.go
@@ -82,7 +82,11 @@ func checkBalance(
 			select {
 			case <-time.After(backoffDuration):
 			case <-timeoutCtx.Done():
-				return fmt.Errorf("insufficient token for initial deposit")
+				if insufficientERC20 {
+					return fmt.Errorf("insufficient token for initial deposit")
+				} else {
+					return fmt.Errorf("insufficient eth for initial deposit")
+				}
 			}
 			continue
 		}
--- a/pkg/settlement/swap/chequebook/init.go
+++ b/pkg/settlement/swap/chequebook/init.go
@@ -60,7 +60,7 @@ func checkBalance(
 		minimumEth := gasPrice.Mul(gasPrice, big.NewInt(2000000))
 
 		insufficientERC20 := erc20Balance.Cmp(swapInitialDeposit) < 0
-		insufficientETH := ethBalance.Cmp(minimumEth) == 0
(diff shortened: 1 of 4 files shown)


acud (approved):

Eknir:
Hi Ralph,
I entered into two problems:

1. I did not have BZZ nor Ether. After getting BZZ from the faucet, the node errored with message:
`Error: insufficient token for initial deposit`
I would expect the process not to end at this point, but to continue looping and telling the user that there is enough BZZ, but not enough eth. Besides, "insufficient token" does not tell the average user that there is enough token (BZZ), but not enough ETH. 
2. Starting the node after that error gives me the message:
`Error: Insufficient funds. The account you tried to send transaction from does not have enough funds. Required 28259140000000000 and got: 0.` before stopping the process. Again, I don't understand this message as "funds" can mean BZZ or Eth. Besides, it would be more user friendly to go into a loop, instructing the user to deposit and wait until that is done.

I hope you can address my comments and that they are in scope for this PR!

pkg/settlement/swap/chequebook/init.go
```diff
@@ -40,22 +40,53 @@ func checkBalance(
...
+			neededERC20, mod := new(big.Int).DivMod(swapInitialDeposit, big.NewInt(10000000000000000), new(big.Int))
+			if mod.Cmp(big.NewInt(0)) > 0 {
+				// always round up the division as the bzzaar cannot handle decimals
+				neededERC20.Add(neededERC20, big.NewInt(1))
+			}
+
+			if insufficientETH && insufficientERC20 {
+				logger.Warningf("cannot continue until there is sufficient ETH and at least %d BZZ available on %x", neededERC20, overlayEthAddress)
+			} else if insufficientETH {
+				logger.Warningf("cannot continue until there is sufficient ETH available on %x", overlayEthAddress)
```
Eknir, line 75:
```suggestion
				logger.Warningf("cannot continue until there is sufficient ETH (for Gas) available on %x", overlayEthAddress)
```

Eknir (approved):
Minor suggestion for improvement left. Otherwise: good work!

Issue #57, opened 2020-04-03
Balanced Kademlia


Eknir:
# User story

## Story

As a node operator, I want to evenly spread the connections that I make each bin, such that I minimize the expected hop count for any random request (given a fixed amount of connections)

## Acceptance criteria

- The average hopcount is reduced (compared to the current implementation)

### Old Swarm
Work on this was done in old Swarm ([this PR](https://github.com/ethersphere/swarm/pull/1869) as a reference).
### Metabins
@metacertain has done research on this and he proposes to implement this concept with `metabins`. Each normal Kademlia bin contains a metabin, that on it's turn can contain 2 additional metabins or an array of peers. A metabin contains a value `R` which stands for the required number of peers in that bin and it is defined as the `binSize` (for the top metabin) or 1/2*R of the parent. When a metabin splits in two children metabins, the address space of the parent metabin is equally divided to the two children.
With this structure in place in both the connected peers and known peers, it becomes possible to implement the balanced kademlia logic, needed to fulfill the acceptance criteria of this issue.

## Tasks

| Task | Assignee |  Done (1), or 0 |
| ------ |------------| -----|

Eknir:
@metacertain, are you going to work on this?

metacertain:
@Eknir yes!

Eknir:
This issue is still in progress @metacertain . could you please update the status, connect it to a PR (if it exists) and put it back into the backlog if you are not currently working on it?
Thanks!

metacertain:
Status - this is in progress still, experiencing performance problems for now, so it's under development

PR #1207, opened 2021-02-08
Balanced peers in kademlia

zbiljic:
relates to: #57 

Using base address for starting node of `00000000` rest of description is made here.
For that base address we have following (random) addresses for different proximities (first 8 here):

```
0 -> "11011011"
1 -> "01011110"
2 -> "00100010"
3 -> "00010111"
4 -> "00001000"
5 -> "00000100"
6 -> "00000010"
7 -> "00000001"
```

Bits in common are what defines bin of some other address in relation to base address.

To have some sort of balancing, we try to have some dispersion of addresses for each bin (proximity).
We define this as bit suffix for bin, and if should be number > 0. If we set it to `1` we would have 2 possibilities, with `2` we have 4, with `3` we have 8, and so on.

For 2 following suffixes are produced for address:

```
00
01
10
11
```

Using this config, for each bin we generate table of "balanced" addresses toward which we aspire to have connections to (this is generated on Kademlia instantiation and is based on node 'base' address). We use prefix for each bin as a staring point, and to it we append each of these suffixes, to create pseudo address prefixes:

```
bin(0)
prefix = ""
balanced prefixes:
'100'
'101'
'110'
'111'

bin(1)
prefix = "0"
balanced prefixes:
'0100'
'0101'
'0110'
'0111'

bin(2)
prefix = "00"
balanced prefixes:
'00100'
'00101'
'00110'
'00111'

...

bin(4)
prefix = "0000"
balanced prefixes:
'0000100'
'0000101'
'0000110'
'0000111'

...
```

The additional bit in these examples comes after the prefix, and is opposite of the bit of the one in the base address. Index of that bit is at `PO + 1`.

With these prepared "balanced" address prefixes, for each bin we can check if we have balanced connections. We also use this information to decide if we want to connect to some other known peer in order to have balanced connected state.

The algorithm to determine if we need to connect to some other peer is done by:

- going over each generated pseudo address (prefix) from each bin
- then finding closest already connected peer to that pseudo address
- if the proximity of that closest connected peer (to pseudo address) is less than 'bin + suffix length'
- we connect to the closest known peer
  - closest known peer is again found in relation to pseudo address
  - and if proximity of it in relation to pseudo address is what we expect for balanced connection we attempt to connect to it

This process is added as separate function just before existing process for connecting to peers in Kademlia, which looks at neighborhood depth to decide if more peer connections need to be made.

### DIFF
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -147,9 +147,7 @@ func (k *Kad) generateCommonBinPrefixes() {
 		}
 	}
 
-	//fmt.Printf("\n %08b", addrBytes)
 	for i := range binPrefixes {
-		//fmt.Printf("\n bin %v", i)
 		for j := range binPrefixes[i] {
 			pseudoAddrBytes := binPrefixes[i][j].Bytes()
 
@@ -180,7 +178,6 @@ func (k *Kad) generateCommonBinPrefixes() {
 				index, pos := l/8, l%8
 				pseudoAddrBytes[index] = bits.Reverse8(clearBit(bits.Reverse8(pseudoAddrBytes[index]), uint8(pos)))
 			}
-			//fmt.Printf("\n %08b", pseudoAddrBytes)
 		}
 	}
 
@@ -245,14 +242,12 @@ func (k *Kad) manage() {
 			err := func() error {
 				// for each bin
 				for i := range k.commonBinPrefixes {
-					skipConnectedPeers := make([]swarm.Address, 0)
-					skipKnownPeers := make([]swarm.Address, 0)
 
 					// and each pseudo address
 					for j := range k.commonBinPrefixes[i] {
 						pseudoAddr := k.commonBinPrefixes[i][j]
 
-						closestConnectedPeer, err := closestPeer(k.connectedPeers, pseudoAddr, skipConnectedPeers...)
+						closestConnectedPeer, err := closestPeer(k.connectedPeers, pseudoAddr, swarm.ZeroAddress)
 						if err != nil {
 							if errors.Is(err, topology.ErrNotFound) {
 								break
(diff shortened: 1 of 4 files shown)


zelig (approved):

acud (approved):

acud:
Looks good. Let's merge this after the release tomorrow

pkg/kademlia/kademlia.go
```diff
@@ -137,7 +233,118 @@ func (k *Kad) manage() {
...
+
+							skipKnownPeers = append(skipKnownPeers, closestKnownPeer)
+
+							if k.connectedPeers.Exists(closestKnownPeer) {
+								continue
+							}
+
+							closestKnownPeerPO := swarm.Proximity(closestKnownPeer.Bytes(), pseudoAddr.Bytes())
+
+							if int(closestKnownPeerPO) != i+k.bitSuffixLength {
```
metacertain, line 284: I think this should be <  i+k.bitSuffixLength+1

pkg/kademlia/kademlia.go
```diff
@@ -637,6 +884,53 @@ func (k *Kad) neighborhoodDepth() uint8 {
...
+
+	binCheckBitLen := int(bin) + k.bitSuffixLength
+
+	connectedBalancedPeers := 0
+
+	// for each pseudo address
+	for i := range k.commonBinPrefixes[bin] {
+		pseudoAddr := k.commonBinPrefixes[bin][i]
+
+		err := k.connectedPeers.EachBinRev(func(peer swarm.Address, po uint8) (bool, bool, error) {
```
metacertain, line 904: This could be substituted with checking proximity of closest connected peer

pkg/kademlia/kademlia.go
```diff
@@ -637,6 +884,53 @@ func (k *Kad) neighborhoodDepth() uint8 {
...
+// IsBalanced returns if Kademlia is balanced to bin.
+func (k *Kad) IsBalanced(bin uint8) bool {
+	k.depthMu.RLock()
+	defer k.depthMu.RUnlock()
+
+	if int(bin) > len(k.commonBinPrefixes) {
+		return false
+	}
+
+	binCheckBitLen := int(bin) + k.bitSuffixLength
```
metacertain, line 896: I think this should be int(bin) + k.bitSuffixLength + 1 because of flipped bit in common addresses

metacertain (approved):
some occurences of a problem with check lengths, probably we should add some simple test as well that could show this

acud:
@metacertain let's do a revision of the branch before merging it

metacertain:
So one other thing that was missing, is that actually we need an extended Proximity function that has a higher limit than MaxPO
Because, if we want balance connections in bin [MaxPO-1], and suffix length is 3, than we need to be able to see proximities beyond MaxPO ( actually MaxPO-1 + suffix length + 1 ). So I duplicated this Proximity function as ExtendedProximity, after being riddled by anomalies beginning at PO 13. It was not obvious at all.

I think we can merge this now!

acud (approved):

PR #1208, opened 2021-02-08
postage:  update contract addresses for new token

ralph-pichler:
update the goerli postage stamp defaults to contracts which use the new token. should be merged after `storage-incentives` was rebased on `master`. nodes which synced the batch store before should reset their state store.


acud (approved):

acud:
do we need this when we rebase over master?

ralph-pichler:
we can merge now and rebase or rebase first, does not matter. but if you're planning for people to use this during this week it's important both things happens, otherwise you will need both tokens (and eth) for the node to start successfully.

PR #1228, opened 2021-02-10
fix funding addr for both linux and mac bee-get-addr

vandot:
Changed funding URL to bzzaar for `bee-get-addr` script


Eknir (approved):
lgtm! Thanks

acud (approved):

PR #1230, opened 2021-02-10
set swap-endpoint to slock.it goerli for deb, rpm and brew package

vandot:
Set `https://rpc.slock.it/goerli` as a `swap-endpoint` for deb, rpm and brew packages
@Eknir should I set that as a default for docker-compose?


Eknir (approved):
I support this change, but I did not look into the implementation. @acud , could you please let us know if you are supportive of this?

acud (approved):

PR #1231, opened 2021-02-10
check if password file already exists

vandot:
upgrades using brew overwrites existing password, with this PR installation will generate password only if password doesnt exist


Eknir (commented):
I can't validate this PR, as I am not the best reader of `yml` scripts and don't have macOS. However, I have seen the user with this problem.
I can approve this PR if needed, but think it is better that somebody can approve who can actually validate this. 

Could you please add the bug report (prt screen or message) to the PR description?

acud (approved):
beautiful

acud (approved):

Issue #1235, opened 2021-02-10
Possible infinite loop waiting for ethereum backend to be synced


r0qs:
#### Summary
Using the goerli endpoint and depending on your timezone the function `IsSynced` may never ends: https://github.com/ethersphere/bee/blob/master/pkg/settlement/swap/transaction/backend.go#L27

I believe that the problem happens because the time delta comparison is not using the same timezone, both should be UTC.
So the `blockTime.After` may never get in the range of the `maxDelay`. The `maxDelay` is hardcoded for one minute in the `node.go`.

Here are some results of my debug:
```
BLOCKTIME:  2021-02-10 23:54:26 +0100 CET
MY TIME: 2021-02-10 22:57:05.911896254 +0000 UTC
DELAY TIME:  2021-02-10 22:56:05.911901424 +0000 UTC
TIMEOUT? false
BLOCKNUMBER: 4261271
BLOCKTIME:  2021-02-10 23:54:26 +0100 CET
MY TIME: 2021-02-10 22:57:06.020386141 +0000 UTC
DELAY TIME:  2021-02-10 22:56:06.02040609 +0000 UTC
TIMEOUT? false
BLOCKNUMBER: 4261271
```
The blocknumber were correct and matched with the ones shown here: https://goerli.etherscan.io/
The time of my machine:
```
$ timedatectl 
               Local time: Wed 2021-02-10 23:58:13 CET
           Universal time: Wed 2021-02-10 22:58:13 UTC
                 RTC time: Wed 2021-02-10 22:58:13    
                Time zone: Europe/Oslo (CET, +0100)   
System clock synchronized: no                         
              NTP service: inactive                   
          RTC in local TZ: no
```

#### Steps to reproduce
Change your timezone. For example, mine is in Norway (UTC+1).
Run a bee node using goerli endpoint: `bee start --swap-endpoint=https://rpc.slock.it/goerli`

#### Expected behavior
The node eventually sync.

#### Actual behavior
It gets stuck waiting for sync.

#### Fix

To fix it, I just converted the blockTime to UTC before compare:
```go
blockTime := time.Unix(int64(header.Time), 0).UTC()
```

haha665:
Hello, are there any specific steps to modify the block time? I didn't understand it very well when I first arrived. thank you very much.

PR #1236, opened 2021-02-10
Convert to UTC before compare blocktime

r0qs:
Proposal to fix https://github.com/ethersphere/bee/issues/1235.


r0qs:
Checking the [tests](https://github.com/ethersphere/bee/blob/master/pkg/settlement/swap/transaction/backend_test.go#L33) I realized that `time.Unix` already return a UTC time. So convert it to UTC again does not make sense and wasn't what solves the problem.
Was only a wrong time setup of the time in my system.

Eknir:
Ok! Thanks anyways for the effort!

Issue #1232, opened 2021-02-10
Error: statestore  - after updating bee node to version 0.5.0


eth-swarm:
#### Summary

After updating bee node to version 0.5.0 (on Daapnode) I got  Error bellow:
Error: statestore: leveldb: manifest corrupted (field 'comparer'): missing [file=MANIFEST-000622]

Bee node keeps restarting

![image](https://user-images.githubusercontent.com/63300164/107497832-54ae3680-6b93-11eb-82b0-bb2cd2f5b540.png)

#### Steps to reproduce

1. install bee node v0.4.2 over dappnode
2. update bee node v0.4.2 to v0.5.0

Eknir:
Thanks for making this bug report. We'll have a look into it!

acud:
I have made a PR to possibly address this, but I also have a hunch that the dirty shutdown might be related to how DappNode shuts bee down. @Eknir can you get me in touch with them so we could have some pointers?

petenickless:
@acud I've got the same issue on the Raspberry Pi 4 running without DappNode - just thought i'd flag.  I've subscribed and will follow updates.  

`Error: statestore: leveldb: manifest corrupted (field 'comparer'): missing [file=MANIFEST-000041]`

PR #1253, opened 2021-02-15
statestore: try to recover leveldb on corrupted db

acud:
It appears that the statestore might shut down in an unclean state, resulting in a corrupted db.
This PR adds an attempt to recover the statestore whenever a corrupt state is detected.

refers to a bug reported in https://github.com/ethersphere/bee/issues/1232

This is obviously controversial as it can yield some other strange states, I am open for suggestions or scratching this entirely.

### DIFF
--- a/pkg/intervalstore/store_test.go
+++ b/pkg/intervalstore/store_test.go
@@ -39,7 +39,7 @@ func TestDBStore(t *testing.T) {
 	}
 	defer os.RemoveAll(dir)
 
-	store, err := leveldb.NewStateStore(dir)
+	store, err := leveldb.NewStateStore(dir, nil)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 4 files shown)


janos (approved):
Very nice.

Should the same recovery attempt be added to the shed as it uses leveldb also?

acud:
Thanks @janos :)
I am a bit more cautions with having localstore use this. I was even in doubts about doing this automatically here. Another suggestion was maybe to have a `bee db repair` and `bee statestore repair` commands that would do this explicitly, and have `node.go` fail as it did before, pointing the user to use the explicit command to try and repair the db/statestore. WDYT?

janos:
Recovery could take a much longer with larger database size, but I am not sure how long. I think that adding new commands does not add more value or control. The user either wants or does not want to do a repair. This is just a preemptive repair approach instead an explicit call for a user interaction. If a user does not want to repair, the process can be stopped and data removed. It is ok to wait until leveldb that stores chunks needs such functionality (that the issue is reported).

zelig (approved):

Issue #1252, opened 2021-02-13
Persisted tags are never deleted


ldeffenb:
#### Summary
Tags are stored in two locations, a memory-resident sync.Map (tags) and the persistent storage.StateStorer (stateStore).  If a tag has been persisted, either through having been created and not deleted before a node shutdown, or by having been completed with PATCH, then that tag will continue to be queriable even after having been DELETEd.

#### Steps to reproduce

Note: All of this is on the main, not debug, API

1) Create a tag with POST /tags
2) Query the tag with GET /tags/{uid}
3) Restart the node to flush to the stateStore
4) Query the tag again with GET /tags/{uid}
5) Delete the tag with DELETE /tags/{uid}
6) Query the tag again with GET /tags/{uid} - It still exists!
7) Restart the node again, to be sure
8) Query the tag again with GET /tags/{uid} - It still exists!

Alternatively, with a PATCHed tag:

1) Create a tag with POST /tags
2) Query the tag with GET /tags/{uid}
3) Persist the tag with PATCH /tags/{uid}
4) Query the tag again with GET /tags/{uid}
5) Delete the tag with DELETE /tags/{uid}
6) Query the tag again with GET /tags/{uid} - It still exists!
7) Restart the node just to be sure
8) Query the tag again with GET /tags/{uid} - It still exists!

However, the following sequence actually works, because the tag was never persisted to the stateStore.

1) Create a tag with POST /tags
2) Query the tag with GET /tags/{uid}
3) Delete the tag with DELETE /tags/{uid}
4) Query the tag again with GET /tags/{uid} - It is gone
5) Restart the node just to be sure
6) Query the tag again with GET /tags/{uid} - It really is gone

#### Expected behavior
DELETE /tags/{uid} should make any tag disappear, including from the stateStore

#### Actual behavior
Persisted tags are never deleted.

This is really bad because there is a limited address space (Uint32) for tag {uid}s and if a random collision occurs, the tag create API will return errExists, something that would not really make sense to the user.  I suspect my own tag space is getting quite full which is why I was querying and deleting obsolete tags, only to discover that they don't really go away!

I suspect the problem is in the Delete method of https://github.com/ethersphere/bee/blob/master/pkg/tags/tags.go.  It only deletes the tag from the tags sync.Map and nothing that I can find ever deletes the key from the stateStore.

acud:
Great catch @ldeffenb! A fix is underway!

ldeffenb:
[...]

PR #1255, opened 2021-02-15
tags: delete persisted tag

acud:
Persisted tags were never removed from statestore on `Delete`, which has resulted in deleted tags to be reloaded on node restart.

addresses https://github.com/ethersphere/bee/issues/1252

### DIFF
--- a/pkg/tags/tags_test.go
+++ b/pkg/tags/tags_test.go
@@ -210,4 +210,21 @@ func TestPersistence(t *testing.T) {
 	if ta.Synced != rcvd2.Synced {
 		t.Fatalf("invalid synced: expected %d got %d", ta.Synced, rcvd2.Synced)
 	}
+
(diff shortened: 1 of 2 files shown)


pkg/tags/tags_test.go
```diff
@@ -210,4 +210,21 @@ func TestPersistence(t *testing.T) {
...
+	// simulate node closing down and booting up
+	err = ts.Close()
+	if err != nil {
+		t.Fatal(err)
+	}
+	ts = NewTags(mockStatestore, logger)
+
+	// get the tag after the node boot up
+	_, err = ts.Get(ta.Uid)
+	if err == nil {
```
janos, line 227: It would be better to have a check on a specific error, as any error may result in a wrong state treated as expected.

janos (approved):
I have only one minor comment.

zelig (approved):

Issue #1247, opened 2021-02-12
Uploading SOC with the same owner and identifier should be rejected


agazso:
#### Summary
I uploaded two single owner chunks on the same node to the `/soc` endpoint with the same owner and identifier and it was successful.  As per Book of Swarm (pg. 115 footnote 3): 

> If the chunks are uploaded using the same route, the chunk that comes later will be rejected as already known.

#### Steps to reproduce
```sh
% echo '010000000000000000' | xxd -r -p | curl --data-binary @- --no-buffer 'http://localhost:61633/soc/8d3766440f0d7b949a5e32995d09619a7f86e632/0000000000000000000000000000000000000000000000000000000000000000?sig=6c7ea367cce127578597309f90f5c8996ae01e0e288a25ca5fdd82bcfb999e8f31ac14a64cfc28c4984160ed160f5fa3156ec3c38afc0b413e60cb62daf362171c' 
{"reference":"9d453ebb73b2fedaaf44ceddcf7a0aa37f3e3d6453fea5841c31f0ea6d61dc85"}

% echo '010000000000000001' | xxd -r -p | curl --data-binary @- --no-buffer 'http://localhost:61633/soc/8d3766440f0d7b949a5e32995d09619a7f86e632/0000000000000000000000000000000000000000000000000000000000000000?sig=94b5a602b9621bc2ab4e703d7c4e219d2754a7e06f6dbea4b1de528a8c43a1ed443bea80abc54506a7b06408714c92e1f8d8eecde556bfd414d34cc49fa032501c'  
{"reference":"9d453ebb73b2fedaaf44ceddcf7a0aa37f3e3d6453fea5841c31f0ea6d61dc85"}
```

#### Expected behavior
The second command should not succeed but instead should be rejected.

Eknir:
@agazso, could you please do some basic QA on this task and report your outcome here? Question: does bee behave now as expected? Is the returned error useful?

agazso:
> @agazso, could you please do some basic QA on this task and report your outcome here? Question: does bee behave now as expected? Is the returned error useful?

Yes, this is fixed in 0.5.1 and 409 (Conflict) is returned.

PR #1256, opened 2021-02-15
api: dont accept two soc's with the same address

acud:
SOCs are not content addressed, therefore SOCs with different content addressed chunks in their payload can result in the same SOC address.
This PR adds a check in the API, to verify that a chunk with the same address does not exist in the localstore before inserting it.

addresses #1247

### DIFF
--- a/pkg/api/soc.go
+++ b/pkg/api/soc.go
@@ -105,7 +105,21 @@ func (s *server) socUploadHandler(w http.ResponseWriter, r *http.Request) {
 		return
 
 	}
+
 	ctx := r.Context()
+
+	has, err := s.storer.Has(ctx, chunk.Address())
+	if err != nil {
+		s.logger.Debugf("soc upload: store has: %v", err)
+		s.logger.Error("soc upload: store has")
+		jsonhttp.InternalServerError(w, "storage error")
+		return
+	}
+	if has {
+		s.logger.Error("soc upload: chunk already exists")
+		jsonhttp.Conflict(w, "chunk already exists")
+		return
+	}
 
 	_, err = s.storer.Put(ctx, requestModePut(r), chunk)
 	if err != nil {
--- a/pkg/api/soc_test.go
+++ b/pkg/api/soc_test.go
@@ -124,6 +124,32 @@ func TestSoc(t *testing.T) {
 			t.Fatal("data retrieved doesnt match uploaded content")
 		}
 	})
+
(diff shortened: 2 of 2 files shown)


janos (commented):
This check has a potential data race as Has and Put calls are not locked together. I would suggest to call a lock on them in the upload hander.

In general, this PR looks to me like a not correct patch to do. I would say that the addressing soc and cac would require more research given the problem described in this PR, but for now this change will technically work.

acud:
@janos having a lock inside the API does not necessarily mitigate the data race completely no? As it might potentially be put from another source. So if I understand correctly we might need a new ModePut here to do the check in localstore during the Put, and avoid the Has call altogether.

janos:
Yes, the lock would protect this case in this handler. This goes along with this PR changeset only addressing the problem in the handler. Basically, this PR adds a check in the handler, and my comment is about the completeness of that approach only within the scope of solution here. As you suggested, a Put that would prevent overwrite would be a more general solution, and probably more correct and safer.

acud:
@janos I started updating the PR to have a new mode for putting an `soc`, but then realised we lose the ability to support upload and atomically pin a chunk. `soc`s will require two calls to upload and pin and I feel that this is a deviation from the current API. I am leaning towards keeping this solution, as it is good enough for now. If a client chooses to put two same `soc` chunks then this is a bit of an edge case and maybe for now with our current design supporting more might be a bit more difficult. Would be happy for your opinion

janos (approved):
@acud That is fine as long as the problem is known and documented. I think that it is important to reiterate on the pinning soc chunks and have a decision on how to handle that problem and have reliable api.

pkg/api/soc.go
```diff
@@ -105,8 +105,22 @@ func (s *server) socUploadHandler(w http.ResponseWriter, r *http.Request) {
 		return
 
 	}
+
 	ctx := r.Context()
 
+	has, err := s.storer.Has(ctx, chunk.Address())
```
zelig, line 111: why not check in swarm why only locally?

Eknir:

pkg/api/soc.go
acud: so uploading a chunk to the API might take up to a minute for all requests to time-out and conclude that the chunk doesn't exist. I don't find this so user-friendly. Can make the change but meh

Issue #1248, opened 2021-02-12
Feed update lookup always return the same update


agazso:
#### Summary
Using the [Find feed update endpoint](https://docs.ethswarm.org/api/#tag/Feed/paths/~1feeds~1{owner}~1{topic}/get) always return the first or second update.

#### Steps to reproduce
1. Start from empty, standalone node.
2. Using the Find feed update endpoint with owner and topic returns 404 (this is good because it's empty).
3. Upload SOC containing value `0` with owner and topic using index `0`.
4. Using the Find feed update endpoint with owner and topic returns value `0`, index `0`, next index `1` (so far so good).
5. Upload SOC containing value `1` with owner and topic using index `1`.
4. Using the Find feed update endpoint with owner and topic returns value `0`, index `0`, next index `1` (same as step 4, this is an error).

#### Expected behavior
Find feed update returns the latest update.

It doesn't matter if you wait between the uploads or not. If you change `sequence.NewAsyncFinder` to `sequence.NewFinder` in `pkg/feeds/factory/factory.go` then it works correctly.

zelig:
@agazso i dont see how this can happen. Knowing the exact test setup would be needed.
Note, that if step 5, the constructed chunk contains a timestamp in the future, then looking up latest returns the  latest at the current time.

agazso:
I uploaded an app to help debugging this issue.

https://gateway.ethswarm.org/files/3ed0350080fd2fab4a9420e0b91571b777389bc87acfda3f21f7cd1712da1ddf

What it does is that it uses the Find feed update endpoint to get the next index and then it creates an SOC with that index and writes an update there. So in the Bee log you are supposed to see something like this:

```
INFO[2021-02-14T19:56:18+01:00] api access                                    duration=0.0170698 ip="::1" method=GET proto=HTTP/1.1 size=45 status=200 uri="/feeds/8d3766440f0d7b949a5e32995d09619a7f86e632/0000000000000000000000000000000000000000000000000000000000000000?type=sequence" user-agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.150 Safari/537.36"
[...]

PR #1259, opened 2021-02-15
feeds: sequence feed resolution fixed and timestamp based heuristic dropped

zelig:
#1248 

- eliminate timestamp based heuristic
- fix sequence async finder bug
- simplify and harden tests
- add test for returned index

### DIFF
--- a/pkg/feeds/sequence/sequence.go
+++ b/pkg/feeds/sequence/sequence.go
@@ -170,7 +170,6 @@ func (f *asyncFinder) At(ctx context.Context, at, after int64) (ch swarm.Chunk, 
 			// if there is a chunk for this path, then the  latest chunk has surely beed sent
 			// since `at` starts from log interval
 			if p.level == r.level && r.level < DefaultLevels {
-
 				return r.chunk, &index{r.index}, &index{r.index + 1}, nil
 			}
 			p.min = r.level
@@ -184,7 +183,6 @@ func (f *asyncFinder) At(ctx context.Context, at, after int64) (ch swarm.Chunk, 
 				return p.chunk, &index{p.index}, &index{p.index + 1}, nil
 			}
 			np := p.next()
-
 			go f.at(ctx, at, np, c, quit)
 		}
 		if p.max < p.min {
--- a/pkg/feeds/sequence/sequence.go
+++ b/pkg/feeds/sequence/sequence.go
@@ -21,6 +21,8 @@ import (
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
+const DefaultLevels = 8
+
 var _ feeds.Index = (*index)(nil)
 var _ feeds.Lookup = (*finder)(nil)
 var _ feeds.Lookup = (*asyncFinder)(nil)
@@ -34,6 +36,10 @@ func (i *index) MarshalBinary() ([]byte, error) {
 	indexBytes := make([]byte, 8)
 	binary.BigEndian.PutUint64(indexBytes, i.index)
 	return indexBytes, nil
+}
+
+func (i *index) Next(last int64, at uint64) feeds.Index {
+	return &index{i.index + 1}
 }
 
 // finder encapsulates a chunk store getter and a feed and provides
@@ -56,14 +62,17 @@ func (f *finder) At(ctx context.Context, at, after int64) (ch swarm.Chunk, curre
 			if !errors.Is(err, storage.ErrNotFound) {
 				return nil, nil, nil, err
 			}
-			return ch, &index{i - 1}, &index{i}, nil
+			if i > 0 {
+				current = &index{i - 1}
+			}
+			return ch, current, &index{i}, nil
 		}
 		ts, err := feeds.UpdatedAt(u)
 		if err != nil {
 			return nil, nil, nil, err
 		}
 		if ts > uint64(at) {
-			return ch, &index{i}, nil, nil
+			return ch, &index{i - 1}, nil, nil
 		}
 		ch = u
 	}
@@ -80,6 +89,9 @@ func NewAsyncFinder(getter storage.Getter, feed *feeds.Feed) feeds.Lookup {
 	return &asyncFinder{feeds.NewGetter(getter, feed)}
 }
 
+// path represents a series of exponential lookahead lookups
+// - starting from base + 1 as level 1
+// - upto base + 2^DefaultLevel - 1
 type path struct {
 	latest    result
 	base      uint64
@@ -88,6 +100,7 @@ type path struct {
 	cancelled bool
 }
 
+// path can be closed if there is a returned chunk
 func (p *path) close() {
 	if !p.cancelled {
 		close(p.cancel)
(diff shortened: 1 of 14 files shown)


agazso:
It's better now, because it doesn't get stuck at index `1` anymore so the original bug is gone.

I see two new issues when filling up a feed with updates:
- Sometimes the Find feed update does not return. The request is sent, Bee doesn't close the connection but does not return any response.
- Sometimes the next index is out of order. For example after `0f` comes `20` (instead of `10`). Then the subsequent index is `11` but then later `20` is skipped. I also saw out-of-order indexes `40`, `80` and `100`.

This suggests to me that more testing is required.

agazso:
Here is the the tool I use for testing:

https://gist.githubusercontent.com/agazso/f504cfd20c13d723e80e101577b0a6ac/raw/9650dc772d64a20c4bc3990ad71c8ef7673e2556/feed-debug-auto.html

It gets the latest update from the `Find feed update` endpoint, then writes an update using the `swarm-feed-index-next` for calculating the index. Then it waits 1.5 seconds and starts from the beginning. It also keeps track of the index, so if the index is not continuous it will write an error.

Usage: curl the link and save the html file locally. 
Open the file from the browser (or alternatively upload it to swarm and run from there).

For example file:///Users/attila/Projects/bee-js/feed-test-auto.html?bee=http://localhost:1633

You can specify different options in query parameters:
`bee`: the Bee node to connect to
`topic`: change the topic
`privateKey`: change the privateKey
`index`: expect this index to be updated, otherwise error

Running this demonstrates the errors I mentioned above in about 10-20 updates.

pkg/feeds/sequence/sequence.go
```diff
@@ -80,89 +94,110 @@ func NewAsyncFinder(getter storage.Getter, feed *feeds.Feed) feeds.Lookup {
...
-			np := newPath(p.latest.seq)
-			np.level = p.level
-			np.latest.chunk = p.latest.chunk
+			np := p.next()
+			go f.at(ctx, at, np, c, quit)
+		}
+		if p.max < p.min {
+			np := newPath(p.base)
+			np.level = p.max
+			np.max = p.level
```
zelig, line 191: this should be np.max=p.max or np.level

pkg/feeds/sequence/sequence.go
```diff
@@ -80,89 +94,110 @@ func NewAsyncFinder(getter storage.Getter, feed *feeds.Feed) feeds.Lookup {
...
 		p.cancelled = true
 	}
 }
-
+func (p *path) next() *path {
+	return &path{
+		base:   p.index,
+		index:  p.index,
+		max:    p.level,
+		level:  p.level,
```
zelig, line 123: this should be `= p.max`

pkg/feeds/sequence/sequence.go
```diff
@@ -80,89 +94,110 @@ func NewAsyncFinder(getter storage.Getter, feed *feeds.Feed) feeds.Lookup {
...
 		close(p.cancel)
 		p.cancelled = true
 	}
 }
-
+func (p *path) next() *path {
+	return &path{
+		base:   p.index,
+		index:  p.index,
+		max:    p.level,
```
zelig, line 122: this should be `= p.max`

pkg/feeds/sequence/sequence.go
```diff
@@ -80,89 +94,110 @@ func NewAsyncFinder(getter storage.Getter, feed *feeds.Feed) feeds.Lookup {
...
 			}
-			p.close()
-			np := newPath(p.latest.seq)
-			np.level = p.level
-			np.latest.chunk = p.latest.chunk
+			np := p.next()
+			go f.at(ctx, at, np, c, quit)
+		}
+		if p.max < p.min {
+			np := newPath(p.base)
```
zelig, line 189: this is incorrect. we should just use p.Next() and reset max and level to p.level

pkg/feeds/testing/lookup.go
```diff
@@ -73,77 +95,98 @@ func TestFinderBasic(t *testing.T, finderf func(storage.Getter, *feeds.Feed) fee
...
-		})
+		}
 	}
 }
 
 func TestFinderRandomIntervals(t *testing.T, finderf func(storage.Getter, *feeds.Feed) feeds.Lookup, updaterf func(putter storage.Putter, signer crypto.Signer, topic []byte) (feeds.Updater, error)) {
 	for i := 0; i < 5; i++ {
 		t.Run(fmt.Sprintf("random intervals %d", i), func(t *testing.T) {
 			storer := mock.NewStorer()
+			// storer := &Timeout{mock.NewStorer()}
```
zelig, line 189: remove

pkg/feeds/sequence/sequence.go
zelig: ^

pkg/feeds/sequence/sequence.go
zelig: ^

zelig: ^

zelig (commented):
LGTM

Eknir:

Eknir:
@zelig , is this ready for review?

zelig:
@Eknir now it is. you can  trust my use of labels :)

agazso:
`index` query parameter is not implemented, it would be good to remove from the documentation (or implement it):

https://docs.ethswarm.org/api/#tag/Feed/paths/~1feeds~1{owner}~1{topic}/get

pkg/feeds/testing/lookup.go
```diff
@@ -17,10 +19,30 @@ import (
...
+// Get overrides the mock storer and introduces latency
+func (t *Timeout) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Address) (swarm.Chunk, error) {
+	ch, err := t.Storer.Get(ctx, mode, addr)
+	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			time.Sleep(searchTimeout)
+		}
+		return ch, err
+	}
+	time.Sleep(time.Duration(rand.Intn(10)) * time.Millisecond)
```
acud, line 40: I would increase this significantly, perhaps to 100 or more milliseconds delay. Also, this assures that chunks which are always returned before the chunks which are not found. Theoretically OK, but does not give us any test coverage that can simulate jitter

acud (changes requested):

pkg/feeds/testing/lookup.go
zelig: I think it is ok, i tested and it does what it is meant to  do. 
increasing the time 10 fold would just slow down tests

pkg/feeds/sequence/sequence.go
```diff
@@ -80,132 +99,157 @@ func NewAsyncFinder(getter storage.Getter, feed *feeds.Feed) feeds.Lookup {
...
-			np.latest.chunk = p.latest.chunk
-			go f.at(ctx, at, np, c, quit)
+			go f.at(ctx, at, 0, i.next(), c, quit)
+		}
+		// inconsistent feed, retry
+		if i.notFound < i.found.level {
+			go f.at(ctx, at, i.found.level, i.retry(), c, quit)
 		}
 	}
 	return nil, nil, nil, nil
```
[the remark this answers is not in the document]
zelig: ok I will do this

acud (approved):

acud:
Let's address further changes in subsequent PRs

PR #1265, opened 2021-02-16
transaction: add support for deployments

ralph-pichler:
add support for deployment transaction in the transaction service. this is not currently used in bee itself, but is useful for some tools which are used to deploy the swap core contracts which currently have to include a slightly modified version of bees `transaction.Service`.


acud (approved):

paxthemax (approved):
All looks well, as far as I can tell. Approved.

Issue #1175, opened 2021-01-28
Bug in hashing process crashes retrieve process


Eknir:
#### Summary

When we upload files of with 16519168 < size < 16773119,  (with encryption) or 67100000 without encryption, a hash is generated but there are unretrievable  
These size generate a Data length of 0 in the metadata, crashing the retrieve process

#### Steps to reproduce

1. 
```bash
truncate -s 16154k bee.jpg
curl -F file=@bee.jpg -H "Swarm-Encrypt: true" http://localhost:1633/files
```
2. Download using the hash generated

The size of problematic files can be calculated on how levels are filled

#### Extra
- Related: https://github.com/ethersphere/swarm/issues/2211
- suspected cause: According to our research, the problem comes from the function "hoistLevels" in hastrie.go, when the target level in the merkle tree is empty after a recursive wrap.
- fix: A proposed quick fix is to launch this function again with the upper level
adding this at line 138 :
```golang
if tlen == 0 {
 return h.hoistLevels(level + 1) 
}
```

PR #1267, opened 2021-02-16
pipeline: hasher bug

acud:
There was a bug in our hashing pipeline that was discovered in #1175.
The problem was that we assumed that the trie depth was known in advanced, however, when wrapping deeper levels under certain circumstances the upper levels would also wrap, resulting in a trie that is deeper than originally assumed. The result was that the trie was being traversed only until the preassumed depth, leaving us with an "empty" level once it has wrapped into the next.

This has been demonstrated with an explicit regression test that is now in the `hashtrie` package. Furthermore, unit tests have now been added to the `hashtrie` package. It covers various scenarios and edge cases.

The `Sum` method logic has now been inlined into the method instead of the helper method, and the logic has been significantly simplified, with extensive comments and explanations on how things work, with the hope that this improves developer onboarding and open-source contributions, which have revealed this bug.

The logic of generating the root hash has changed and indeed simplified. In the past, we would wrongly assume which level we're aiming at and try to read the resulting data out of that level. However as mentioned, this did not take into account the case where several subsequent levels were wrapped as a result of one level wrap (causing the trie to balance, basically).
After the simplification we just iterate on all levels and constantly carry over hashes or wrap levels as needed, carrying over the result to higher level. In the end of the operation, we just take the resulting hash from the top level of the trie.

I have used `TestRegression` to verify that this problem exists on current `master` and is solved with this change, as well as manual tests. In the unit tests added to the `hashtrie` package, I have chosen an approach where one predefined hash is constantly being written to the trie, with constant spans of value `1`. This allows the eventual root chunk span to be compared against the linear number of writes that were done in the test. Also, custom branching and chunk size values were used in tests in order to decrease the number of writes needed in order to trigger a level wrap of the trie.

### DIFF
--- a/pkg/file/pipeline/hashtrie/hashtrie.go
+++ b/pkg/file/pipeline/hashtrie/hashtrie.go
@@ -7,7 +7,6 @@ package hashtrie
 import (
 	"encoding/binary"
 	"errors"
-	"fmt"
 
 	"github.com/ethersphere/bee/pkg/file/pipeline"
 	"github.com/ethersphere/bee/pkg/swarm"
@@ -17,6 +16,8 @@ var (
 	errInconsistentRefs = errors.New("inconsistent reference lengths in level")
 	errTrieFull         = errors.New("trie full")
 )
+
+const maxLevel = 8
 
 type hashTrieWriter struct {
 	branching  int
@@ -138,11 +139,8 @@ func (h *hashTrieWriter) levelSize(level int) int {
 //		the hash to the next level, potentially resulting in a level wrap
 //	- more than one hash, in which case we _do_ perform a hashing operation, appending the hash to
 //		the next level.
-
 func (h *hashTrieWriter) Sum() ([]byte, error) {
-	fmt.Println("sum", h.cursors, h.levelSize(1))
 	oneRef := h.refSize + swarm.SpanSize
-	maxLevel := 8
 	for i := 1; i < maxLevel; i++ {
 		l := h.levelSize(i)
 		if l%oneRef != 0 {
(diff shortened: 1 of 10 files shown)


pkg/storage/mock/storer.go
```diff
@@ -93,7 +93,9 @@ func (m *MockStorer) Put(ctx context.Context, mode storage.ModePut, chs ...swarm
 			po := swarm.Proximity(ch.Address().Bytes(), m.baseAddress)
 			m.bins[po]++
 		}
-		m.store[ch.Address().String()] = ch.Data()
+		b := make([]byte, len(ch.Data()))
```
acud, line 96: this is needed since the chunk feeder shares memory across calls to the pipeline. this is in order to avoid multiple allocations. this change mimics the behavior of shed and localstore, and copies the data from the call into the in-memory store

janos (approved):
👏

pkg/file/pipeline/hashtrie/hashtrie.go
```diff
@@ -12,15 +12,21 @@ import (
...
 
-var errInconsistentRefs = errors.New("inconsistent reference lengths in level")
+var (
+	errInconsistentRefs = errors.New("inconsistent references")
+	errTrieFull         = errors.New("trie full")
+)
+
+const maxLevel = 8
 
 type hashTrieWriter struct {
```
zelig, line 22: call this just writer

pkg/file/pipeline/hashtrie/hashtrie.go
```diff
@@ -12,15 +12,21 @@ import (
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
-var errInconsistentRefs = errors.New("inconsistent reference lengths in level")
+var (
+	errInconsistentRefs = errors.New("inconsistent references")
```
zelig, line 16: not  exported?

pkg/file/pipeline/hashtrie/hashtrie.go
```diff
@@ -12,15 +12,21 @@ import (
...
+	errTrieFull         = errors.New("trie full")
```
zelig, line 17: i thought we agreed this should be handled but  its ok

zelig (approved):
LGTM. please change trie to tree everywhere. this is not  a trie

Eknir:

pkg/file/pipeline/hashtrie/hashtrie.go
acud: what for?

acud:
> change trie to tree everywhere

Let's throttle renaming. I'll do this in a separate PR

PR #1277, opened 2021-02-17
packaging/docker: remove quotes from compose .env file

skylenet:
Related to https://github.com/ethersphere/docs.github.io/pull/133

The values inside the docker-compose .env file should not contain quotes. Otherwise these will be included in the values and we start seeing errors i ike :

```
Error: parse "\"https://rpc.slock.it/goerli\"": first path segment in URL cannot contain colon
```

### DIFF
--- a/packaging/docker/env
+++ b/packaging/docker/env
@@ -7,23 +7,23 @@
 
 ### BEE
 
-## HTTP API listen address (default ":1633")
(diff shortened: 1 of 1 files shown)


vandot (approved):
LGTM
Thanks

acud (approved):

PR #1278, opened 2021-02-17
enable dns entry for nat-addr

vandot:
when forming multiaddr there is a check if dns entry is valid so we need to check if dns entry has only one type of A record or has both A and AAAA. we set `dns4`, `dns6` or `dns` as proto for multiaddr


janos (approved):
Cool!

acud (approved):

PR #1281, opened 2021-02-17
Cac refactor

r0qs:
This PR refactor the chunk construction methods and remove some duplicated code.

### DIFF
--- a/pkg/cac/cac.go
+++ b/pkg/cac/cac.go
@@ -15,26 +15,28 @@ var (
 
 // New creates a new content address chunk by initializing a span and appending the data to it.
 func New(data []byte) (swarm.Chunk, error) {
-	if len(data) > swarm.ChunkSize+swarm.SpanSize {
+	dataLength := len(data)
+	if dataLength > swarm.ChunkSize {
 		return nil, errTooLargeChunkData
 	}
 
-	if len(data) < swarm.SpanSize {
+	if dataLength < swarm.SpanSize {
 		return nil, errTooShortChunkData
 	}
 
 	span := make([]byte, swarm.SpanSize)
-	binary.LittleEndian.PutUint64(span, uint64(len(data)))
+	binary.LittleEndian.PutUint64(span, uint64(dataLength))
 	return NewWithSpan(data, span)
 }
 
 // NewWithDataSpan creates a new chunk assuming that the span precedes the actual data.
 func NewWithDataSpan(data []byte) (swarm.Chunk, error) {
-	if len(data) > swarm.ChunkSize+swarm.SpanSize {
+	dataLength := len(data)
+	if dataLength > swarm.ChunkSize+swarm.SpanSize {
 		return nil, errTooLargeChunkData
 	}
 
-	if len(data) < swarm.SpanSize {
+	if dataLength < swarm.SpanSize {
 		return nil, errTooShortChunkData
 	}
 	return NewWithSpan(data[swarm.SpanSize:], data[:swarm.SpanSize])
--- a/pkg/cac/cac.go
+++ b/pkg/cac/cac.go
@@ -17,7 +17,7 @@ var (
 func New(data []byte) (swarm.Chunk, error) {
 	span := make([]byte, swarm.SpanSize)
 	binary.LittleEndian.PutUint64(span, uint64(len(data)))
-	return newWithSpan(data, span)
+	return NewWithSpan(data, span)
 }
 
 // NewWithDataSpan creates a new chunk assuming that the span precedes the actual data.
@@ -30,11 +30,11 @@ func NewWithDataSpan(data []byte) (swarm.Chunk, error) {
 		return nil, errTooShortChunkData
 	}
 
-	return newWithSpan(data[swarm.SpanSize:], data[:swarm.SpanSize])
+	return NewWithSpan(data[swarm.SpanSize:], data[:swarm.SpanSize])
 }
 
-// newWithSpan creates a new chunk prepending the given span to the data.
-func newWithSpan(data []byte, span []byte) (swarm.Chunk, error) {
+// NewWithSpan creates a new chunk prepending the given span to the data.
+func NewWithSpan(data []byte, span []byte) (swarm.Chunk, error) {
 	h := hasher(data)
 	hash, err := h(span)
 	if err != nil {
--- a/pkg/cac/cac.go
+++ b/pkg/cac/cac.go
@@ -43,7 +43,7 @@ func NewWithDataSpan(data []byte) (swarm.Chunk, error) {
 }
 
 // newWithSpan creates a new chunk prepending the given span to the data.
-func newWithSpan(data []byte, span []byte) (swarm.Chunk, error) {
+func newWithSpan(data, span []byte) (swarm.Chunk, error) {
 	h := hasher(data)
 	hash, err := h(span)
 	if err != nil {
--- a/pkg/cac/cac.go
+++ b/pkg/cac/cac.go
@@ -2,25 +2,59 @@ package cac
 
 import (
 	"encoding/binary"
+	"errors"
 
 	"github.com/ethersphere/bee/pkg/bmtpool"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
+var (
+	errTooShortChunkData = errors.New("short chunk data")
+	errTooLargeChunkData = errors.New("data too large")
+)
+
+// New creates a new content address chunk by initializing a span and appending the data to it.
 func New(data []byte) (swarm.Chunk, error) {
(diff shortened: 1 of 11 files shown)


pkg/cac/cac.go
```diff
@@ -2,25 +2,59 @@ package cac
...
+var (
+	errTooShortChunkData = errors.New("short chunk data")
+	errTooLargeChunkData = errors.New("data too large")
+)
+
+// New creates a new content address chunk by initializing a span and appending the data to it.
 func New(data []byte) (swarm.Chunk, error) {
-	hasher := bmtpool.Get()
-	defer bmtpool.Put(hasher)
+	span := make([]byte, swarm.SpanSize)
```
acud, line 18: you should do the input validation the same way you do in `NewWithDataSpan`, since `data` can be larger than a single chunk, which should result in an error

pkg/cac/cac.go
```diff
@@ -2,25 +2,59 @@ package cac
...
+}
+
+// NewWithSpan creates a new chunk prepending the given span to the data.
+func NewWithSpan(data []byte, span []byte) (swarm.Chunk, error) {
+	h := hasher(data)
+	hash, err := h(span)
 	if err != nil {
 		return nil, err
 	}
+	return swarm.NewChunk(swarm.NewAddress(hash), append(span, data...)), nil
```
acud, line 43: This is a bit dangerous. Since `span` can actually be a slice which references storage that is not guaranteed to be immutable, you can potentially end up with a data race here, this is why the race detector complains. The problem is that you're reslicing data when calling this function from line 33. The slice you're shuffling around can actually cause all sorts of quirks. I would copy at least the span in `NewWithDataSpan` to a new slice, so that when you append `data` you actually append it to a new slice, preventing all sorts of peculiar bugs in the future.

acud (changes requested):

pkg/cac/cac.go
```diff
@@ -2,25 +2,70 @@ package cac
...
+var (
+	errTooShortChunkData = errors.New("short chunk data")
+	errTooLargeChunkData = errors.New("data too large")
+)
+
+// New creates a new content address chunk by initializing a span and appending the data to it.
 func New(data []byte) (swarm.Chunk, error) {
-	hasher := bmtpool.Get()
-	defer bmtpool.Put(hasher)
+	if len(data) > swarm.ChunkSize+swarm.SpanSize {
```
acud, line 18: actually, here `data` can only be `0<len<=swarm.ChunkSize`, since the span is created separately below. A unit test that checks these edge cases would be good.

pkg/cac/cac_test.go
```diff
@@ -21,3 +23,68 @@ func TestCac(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !c.Address().Equal(address) {
+		t.Fatalf("address mismatch. got %s want %s", c.Address().String(), address.String())
+	}
+}
+
+var testChunkData = []struct {
```
acud, line 60: You can inline this into the actual test below. See [here](https://github.com/ethersphere/bee/blob/master/pkg/encryption/mock/mock_test.go#L18)

pkg/cac/cac_test.go
```diff
@@ -21,3 +23,68 @@ func TestCac(t *testing.T) {
 		t.Fatalf("address mismatch. got %s want %s", c.Address().String(), address.String())
 	}
 }
+
+func TestNewWithDataSpan(t *testing.T) {
+	foo := "greaterthanspan"
```
acud, line 28: I think that having table-drive tests for all constructors would be good, so that we can validate all input and outputs.
Note that the original sanity test that I wrote here checks only for the address of the generated chunk, however it should also check the actual `Data()` itself is correct (check that it equals `span|data`). There's no need to do this on all tests, I think that on the sanity check would suffice.

pkg/cac/export_test.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
acud, line 1: 2021 :)

pkg/feeds/putter.go
```diff
@@ -57,21 +57,7 @@ func (u *Putter) Put(ctx context.Context, i Index, at int64, payload []byte) err
 }
 
 func toChunk(at uint64, payload []byte) (swarm.Chunk, error) {
-	hasher := bmtpool.Get()
-	defer bmtpool.Put(hasher)
-
-	ts := make([]byte, 8)
+	ts := make([]byte, swarm.SpanSize)
```
acud, line 60: note that `ts` is a timestamp here, so it is indeed `8` bytes because of the fact it is a Uint64, but I would prefer to initialize the variable with a hardcoded `8` bytes since using the `const` in `swarm` package might mislead the reader to think this is somehow related to the chunk span.

pkg/soc/soc.go
```diff
@@ -154,7 +154,7 @@ func FromChunk(sch swarm.Chunk) (*Soc, error) {
 	spanBytes := chunkData[cursor : cursor+swarm.SpanSize]
 	cursor += swarm.SpanSize
 
-	ch, err := contentAddressedChunk(chunkData[cursor:], spanBytes)
+	ch, err := cac.NewWithSpan(chunkData[cursor:], spanBytes)
```
acud, line 157: You could potentially simplify this with removing `spanBytes` and the cursor increment on `L154-155`, then calling `cac.NewWithDataSpan(chunkData[cursor:])`. This way you can unexport `NewWithSpan` since it is only used here (and simplify the tests as a result).

acud (changes requested):
Few more small changes. Thanks for helping us maintaining this important part of the codebase :)

pkg/cac/cac.go
r0qs: I fixed it now and it is already tested in the new `TestChunkInvariants`. But currently I'm only aware of two cases:
1. Data is smaller than the spanSize
2. Data is bigger than the maxChunkSize

Do you see more cases?

pkg/soc/soc.go
r0qs: True! Thanks. I fixed it now.

pkg/cac/cac_test.go
r0qs: I could also change the sanity tests to a table-driven test and reduce some repetition in those tests, but I'm not sure if this will be good in terms of legibility.

pkg/cac/cac_test.go
```diff
@@ -1,23 +1,101 @@
...
+			{
+				name:    "too large data chunk",
+				data:    []byte(strings.Repeat("a", swarm.ChunkSize+swarm.SpanSize+1)),
+				wantErr: cac.ErrTooLargeChunkData,
+			},
+		} {
+			testName := fmt.Sprintf("%s-%s", f.name, cc.name)
+			t.Run(testName, func(t *testing.T) {
+				_, err := f.chunker(cc.data)
+				if err != nil {
```
acud, line 91: Since you always expect errors here, I would:
```go
_,err := f.chunker()
if !errors.Is(err,cc.wantErr) { .... } 
```
Then you don't need to handle the `else` block, since `errors.Is` will return false, since `err` would be `nil`

pkg/cac/cac_test.go
```diff
@@ -1,23 +1,101 @@
...
+		},
+	}
+
+	for _, f := range chunkerFunc {
+		for _, cc := range []struct {
+			name    string
+			data    []byte
+			wantErr error
+		}{
+			{
```
acud, line 77: I would add test cases for:
- data is a `0` length slice
- data is `nil`

pkg/soc/soc_test.go
```diff
@@ -25,8 +26,8 @@ func TestToChunk(t *testing.T) {
 
 	id := make([]byte, 32)
 
-	payload := []byte("foo")
-	ch, err := chunk(payload)
+	payload := []byte("greaterthanspan")
```
acud, line 29: This actually does not need to be longer than span, since you're calling `New`

pkg/soc/soc_test.go
```diff
@@ -92,8 +93,8 @@ func TestFromChunk(t *testing.T) {
...
```
acud, line 96: here too

pkg/soc/soc_test.go
r0qs: Oh, I just noticed that I wrongly assumed that in the `New` the data also should be smaller than the span. Reading carefully your comments again I noticed that this is only the case for the `NewWithDataSpan`. I adjusted the code and the tests for both cases now.

acud (approved):
Brilliant! :clap: Thank you very much for this contribution :pray:

janos (approved):
Very nice. 👏

PR #1282, opened 2021-02-18
libp2p: always close streams

acud:
Using our `FullClose` helper method might not always result in streams being garbage collected. This PR adds a defer call to close the stream in any case when the method returns


janos (approved):

acud:
2021-02-18-155711_1920x1080_scrot
I have deployed this change to the node which was displaying the memory leak, without deploying to the rest of the cluster. This seems to improve the situation ^^

Issue #1280, opened 2021-02-17
Allow CORS for debug API [enhancement]


alsakhaev:
"cors-allowed-origins" config doesn't affect a debug :1635 API

Use-case: web interface (gui) for administration of bee node

alsakhaev:
looks like this piece of code is responsible for CORS in the Main API
https://github.com/ethersphere/bee/blob/5e1a341dc125e6a7d6594cedae7ad494e664c8be/pkg/api/router.go#L199-L205

and something similar is missing here:
https://github.com/ethersphere/bee/blob/5e1a341dc125e6a7d6594cedae7ad494e664c8be/pkg/debugapi/router.go#L153-L157

vojtechsimetka:
This looks resolved by https://github.com/ethersphere/bee/pull/1288

Eknir:
Documentation: document this and document sane usage of cors and document potential security concerns

Eknir:
QA task : @sig or @vojtechsimetka , can you please let me know if this works as expected?

PR #1288, opened 2021-02-19
Apply CORS Options for Debug Api

significance:
This is a tentative PR to attempt to address #1280 

I think it makes sense that the same CORS settings are valid for the API are also valid for the Debug API on the basis that these settings are only supposed to apply for entities with privileged network access and without Gateway Mode enabled, pending implementing some sort of remote access.

Note I have duplicated code for `checkOrigin` and `equalASCIIFold` in `api.go` and `debugapi.go` and would appreciate guidance on how best to DRY this.

Otherwise it's pretty much a brutalistic copy paste job so please do feel free to close if it's not appropriate. 😀

### DIFF
--- a/pkg/debugapi/debugapi_test.go
+++ b/pkg/debugapi/debugapi_test.go
@@ -57,7 +57,7 @@ func newTestServer(t *testing.T, o testServerOptions) *testServer {
 	settlement := swapmock.New(o.SettlementOpts...)
 	chequebook := chequebookmock.NewChequebook(o.ChequebookOpts...)
 	swapserv := swapmock.NewApiInterface(o.SwapOpts...)
-	s := debugapi.New(o.Overlay, o.PublicKey, o.PSSPublicKey, o.EthereumAddress, o.P2P, o.Pingpong, topologyDriver, o.Storer, logging.New(ioutil.Discard, 0), nil, o.Tags, acc, settlement, true, swapserv, chequebook)
+	s := debugapi.New(o.Overlay, o.PublicKey, o.PSSPublicKey, o.EthereumAddress, o.P2P, o.Pingpong, topologyDriver, o.Storer, logging.New(ioutil.Discard, 0), nil, o.Tags, acc, settlement, true, swapserv, chequebook, api.Options{})
 	ts := httptest.NewServer(s)
 	t.Cleanup(ts.Close)
 
--- a/pkg/debugapi/debugapi.go
+++ b/pkg/debugapi/debugapi.go
@@ -10,6 +10,7 @@ package debugapi
 import (
 	"crypto/ecdsa"
 	"net/http"
+	"unicode/utf8"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethersphere/bee/pkg/accounting"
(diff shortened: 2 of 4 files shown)


pkg/debugapi/debugapi_test.go
```diff
@@ -57,7 +57,7 @@ func newTestServer(t *testing.T, o testServerOptions) *testServer {
 	settlement := swapmock.New(o.SettlementOpts...)
 	chequebook := chequebookmock.NewChequebook(o.ChequebookOpts...)
 	swapserv := swapmock.NewApiInterface(o.SwapOpts...)
-	s := debugapi.New(o.Overlay, o.PublicKey, o.PSSPublicKey, o.EthereumAddress, o.P2P, o.Pingpong, topologyDriver, o.Storer, logging.New(ioutil.Discard, 0), nil, o.Tags, acc, settlement, true, swapserv, chequebook)
+	s := debugapi.New(o.Overlay, o.PublicKey, o.PSSPublicKey, o.EthereumAddress, o.P2P, o.Pingpong, topologyDriver, o.Storer, logging.New(ioutil.Discard, 0), nil, o.Tags, acc, settlement, true, swapserv, chequebook, api.Options{})
```
acud, line 60: referencing the wrong package. should be `debugapi.Options`

acud (commented):
I am a bit wary of exposing debugapi with CORS. This smells a bit. If this is a huge blocker for someone to build something really exciting over swarm right now then I think we can make a compromise, but that being said we will have to, as a result, prioritize institutionalizing this properly in the short-run, otherwise this will potentially come up in the security audit and increases our attack surface. If it's a nice-to-have then I'd not rush doing this, especially since people will start treating this API as a customer facing API, and then later on we will have to break this. Lots of compromises in various directions. Would be happy to hear what others think

pkg/debugapi/debugapi_test.go
significance: thanks have fixed 👍

significance:
i think it would be good to get this in as a quick fix, the debug API should already be firewalled off and only used via local network, there are warnings in the docs to this effect, i can also augment this. 

enabling with it would facilitate the development of GUI node management browser apps which arguably should run on the local network now in any case, and only later remotely through an authenticated api endpoint if we choose to implement this.

i would be in favour of adding checks to ensure the debug api is not being accessed over the public internet, i.e. if headers do not show origin ip's that are reserved for use on internal networking.

Eknir:
This feature has been requested to me via several channels. I do think it is a blocker to build amazing applications on the short-term over Swarm. That being said, I also don't see this as a great long-term option.
I would be OK to have this in now, but we need to proceed the discussion at #1289

janos (commented):
I agree with @acud  on the direction that cors would put the information from debug api. As this looks to me like a frequently requested feature, it would be good to put a nice warning sign in docs, or even a dedicated option for api cors headers and debug api cors headers, to be more explicit.

janos (approved):
Looks good, just to fix this issue with deepsource integration https://github.com/ethersphere/bee/runs/1952940065#step:11:157.

acud:
> Looks good, just to fix this issue with deepsource integration

Not possible. This is due to the fact @significance is PRing from a private fork. @significance can you please use branches inside the bee repo in the future?

acud (approved):

PR #1290, opened 2021-02-19
libp2p: add more stream cleanups

acud:
This PR adds more stream cleanups in some places where a dangling stream may still be left half-open.


janos (approved):
This should be fine. It has to be tested in integration, of course.

acud:
Already deployed on staging and memory consumption is back to normal :+1:

PR #1294, opened 2021-02-19
Feature/keys command

syspulse:
Add **keys** command to bee client to dump all keys (including Private Keys!)
This could be useful when not using clef (e.g. import Eth account into Wallet by private key, like Metamask)

### DIFF
--- a/cmd/bee/cmd/cmd.go
+++ b/cmd/bee/cmd/cmd.go
@@ -100,6 +100,10 @@ func newCommand(opts ...option) (c *command, err error) {
 		return nil, err
 	}
 
+	if err := c.initKeysCmd(); err != nil {
+		return nil, err
+	}
+
 	c.initVersionCmd()
 
 	if err := c.initConfigurateOptionsCmd(); err != nil {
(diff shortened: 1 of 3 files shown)


Eknir:
Thank you for the PR @syspulse. The PR is against our policy of recommending the usage of bee-clef, but I can imagine this might be useful for users. We will consider your PR.

zelig (approved):

acud:
Hi @syspulse, thanks for you contribution with this PR.

We unfortunately cannot merge this into the codebase, since this goes against our security policy of not exposing private keys altogether in the `bee` process. The changes that you're introducing would leak the private key into the process even under normal operation, since the bee process reuses the same `configureSigner` section under normal node bootstrapping.

In case you're using `bee-clef`, you can use the `bee-clef keys` command that exports the private keys for importing in metamask, for example. There is also the `exportSwarmKey` util which was made by @jmozah, you can find a fork on our organization [here](https://github.com/ethersphere/exportSwarmKey).

We welcome such contributions of external open-source tools, but we have to be prudent with what we introduce to the `bee` codebase. Thank you for the understanding :)

PR #1310, opened 2021-02-22
chequebook: continue waiting for previous deploy transaction

ralph-pichler:
wait for previous deployment attempt to succeed if bee is stopped while waiting for deploy tx

### DIFF
--- a/pkg/settlement/swap/chequebook/init.go
+++ b/pkg/settlement/swap/chequebook/init.go
@@ -17,7 +17,10 @@ import (
 	"github.com/ethersphere/bee/pkg/storage"
 )
 
-const chequebookKey = "swap_chequebook"
+const (
+	chequebookKey           = "swap_chequebook"
+	chequebookDeploymentKey = "swap_chequebook_transaction_deployment"
+)
 
 func checkBalance(
 	ctx context.Context,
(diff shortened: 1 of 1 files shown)


pkg/settlement/swap/chequebook/init.go
```diff
@@ -127,21 +130,35 @@ func Init(
...
+			logger.Info("no chequebook found, deploying new one.")
+			if swapInitialDeposit.Cmp(big.NewInt(0)) != 0 {
+				err = checkBalance(ctx, logger, swapInitialDeposit, swapBackend, chainId, overlayEthAddress, erc20BindingFunc, erc20Address, 20*time.Second, 10)
+				if err != nil {
+					return nil, err
+				}
+			}
+
+			// if we don't yet have a chequebook, deploy a new one
+			txHash, err = chequebookFactory.Deploy(ctx, overlayEthAddress, big.NewInt(0))
```
acud, line 148: consider:
```
store.Put()
err := Deploy()
if err != nil {
   store.Delete()
   ...
}
...
```
This can eliminate the possibility of sending the transaction but not logging the actual key since the user did a `Ctrl-c` in between.

acud (approved):
Minor suggestion. Otherwise LGTM

pkg/settlement/swap/chequebook/init.go
ralph-pichler: there is nothing useful we can store before `Deploy` returns. The best way would be to already store the hash before the actual sending but that would require further refactor which I would not do in this PR.

paxthemax (approved):
This all looks good, approved and ready to merge.

PR #1314, opened 2021-02-22
Add deploy command to bee

paxthemax:
This PR adds the deploy command to bee. The `deploy` command will deploy the chequebook using the configured factory address. This is the same functionality from `start`, extracted to a separate command.

I've performed a basic test:

1. Calling deploy
2. Funding BZZ and ETH
3. Waiting for deploy to finish
4. Calling start

Expected outcome was reached: upon running `start`, bee detects and uses the same chequebook from deploy.

### DIFF
--- a/cmd/bee/cmd/deploy.go
+++ b/cmd/bee/cmd/deploy.go
@@ -52,6 +52,8 @@ func (c *command) initDeployCmd() error {
 				return
 			}
 
+			defer stateStore.Close()
+
 			signerConfig, err := c.configureSigner(cmd, logger)
 			if err != nil {
 				return
@@ -70,6 +72,7 @@ func (c *command) initDeployCmd() error {
 			if err != nil {
 				return
 			}
+			defer swapBackend.Close()
 
 			chequebookFactory, err := node.InitChequebookFactory(
 				logger,
--- a/cmd/bee/cmd/deploy.go
+++ b/cmd/bee/cmd/deploy.go
@@ -1,0 +1,109 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/node"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func (c *command) initDeployCmd() error {
+	cmd := &cobra.Command{
+		Use:   "deploy",
+		Short: "Deploy and fund the chequebook contract",
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			if (len(args)) > 0 {
+				return cmd.Help()
+			}
+
+			var logger logging.Logger
+			switch v := strings.ToLower(c.config.GetString(optionNameVerbosity)); v {
+			case "0", "silent":
+				logger = logging.New(ioutil.Discard, 0)
+			case "1", "error":
+				logger = logging.New(cmd.OutOrStdout(), logrus.ErrorLevel)
+			case "2", "warn":
+				logger = logging.New(cmd.OutOrStdout(), logrus.WarnLevel)
+			case "3", "info":
+				logger = logging.New(cmd.OutOrStdout(), logrus.InfoLevel)
+			case "4", "debug":
+				logger = logging.New(cmd.OutOrStdout(), logrus.DebugLevel)
+			case "5", "trace":
+				logger = logging.New(cmd.OutOrStdout(), logrus.TraceLevel)
+			default:
+				return fmt.Errorf("unknown verbosity level %q", v)
+			}
+
+			dataDir := c.config.GetString(optionNameDataDir)
+			factoryAddress := c.config.GetString(optionNameSwapFactoryAddress)
+			swapInitialDeposit := c.config.GetString(optionNameSwapInitialDeposit)
+			swapEndpoint := c.config.GetString(optionNameSwapEndpoint)
+
+			stateStore, err := node.InitStateStore(logger, dataDir)
+			if err != nil {
+				return
+			}
+
+			signerConfig, err := c.configureSigner(cmd, logger)
+			if err != nil {
+				return
+			}
+			signer := signerConfig.signer
+
+			ctx := cmd.Context()
+
+			swapBackend, overlayEthAddress, chainID, transactionService, err := node.InitChain(
+				ctx,
+				logger,
+				stateStore,
+				swapEndpoint,
+				signer,
+			)
+			if err != nil {
+				return
+			}
+
+			chequebookFactory, err := node.InitChequebookFactory(
+				logger,
+				swapBackend,
+				chainID,
+				transactionService,
+				factoryAddress,
+			)
+			if err != nil {
+				return
+			}
+
+			_, err = node.InitChequebookService(
+				ctx,
+				logger,
+				stateStore,
+				signer,
+				chainID,
+				swapBackend,
+				overlayEthAddress,
+				transactionService,
+				chequebookFactory,
+				swapInitialDeposit,
+			)
+
+			return
+		},
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return c.config.BindPFlags(cmd.Flags())
+		},
+	}
+
+	c.setAllFlags(cmd)
+	c.root.AddCommand(cmd)
+
+	return nil
+}
--- a/pkg/node/chain.go
+++ b/pkg/node/chain.go
@@ -1,0 +1,220 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package node
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math/big"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethersphere/bee/pkg/crypto"
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/p2p/libp2p"
+	"github.com/ethersphere/bee/pkg/settlement/swap"
+	"github.com/ethersphere/bee/pkg/settlement/swap/chequebook"
+	"github.com/ethersphere/bee/pkg/settlement/swap/swapprotocol"
+	"github.com/ethersphere/bee/pkg/settlement/swap/transaction"
+	"github.com/ethersphere/bee/pkg/storage"
+	"gopkg.in/src-d/go-log.v1"
+)
+
+const (
+	maxDelay = 1 * time.Minute
+)
+
+// InitChain will initialize the Ethereum backend at the given endpoint and
+// set up the Transacton Service to interact with it using the provided signer.
+func InitChain(
+	ctx context.Context,
+	logger logging.Logger,
+	stateStore storage.StateStorer,
+	endpoint string,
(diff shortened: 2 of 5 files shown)


pkg/node/chain.go
```diff
@@ -0,0 +1,249 @@
...
+	)
+	if err != nil {
+		err = fmt.Errorf("failed to init chequebook service: %w", err)
+		return
+	}
+
+	return
+}
+
+func initChequeStoreCahsout(
```
ralph-pichler, line 178: cashout

pkg/node/chain.go
```diff
@@ -0,0 +1,249 @@
...
+	stateStore storage.StateStorer,
+	swapBackend *ethclient.Client,
```
ralph-pichler, line 180: use `transaction.Backend` here

cmd/bee/cmd/deploy.go
```diff
@@ -0,0 +1,111 @@
...
+				swapBackend,
+				overlayEthAddress,
+				transactionService,
+				chequebookFactory,
+				swapInitialDeposit,
+			); err != nil {
+				return err
+			}
+
+			return nil
```
ralph-pichler, line 100: you can drop the last if and just return the err from `InitChequebookService `

cmd/bee/cmd/deploy.go
```diff
@@ -0,0 +1,111 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
ralph-pichler, line 1: wrong year

pkg/node/chain.go
```diff
@@ -0,0 +1,249 @@
+package node
```
ralph-pichler, line 1: missing header

pkg/node/chain.go
```diff
@@ -0,0 +1,249 @@
...
+	transactionService transaction.Service,
+	chequebookFactory chequebook.Factory,
+	initialDeposit string,
+) (
+	chequebookService chequebook.Service,
+	err error,
+) {
+	chequeSigner := chequebook.NewChequeSigner(signer, chainID)
+
+	isSynced, err := transaction.IsSynced(ctx, backend, maxDelay)
```
ralph-pichler, line 135: does this not belong better to `InitChain`?

ralph-pichler (changes requested):

pkg/node/statestore.go
```diff
@@ -0,0 +1,25 @@
+package node
```
ralph-pichler, line 1: header

ralph-pichler (changes requested):

acud (commented):
my 2 cents:
The `node` package is used to construct nodes, so I find it a bit out of place to abstract some of the node creation logic used in node bootstrapping outside of the node package. I might be off here by more than a bit though, so I would be happy to hear @janos opinion on this

pkg/node/statestore.go
```diff
@@ -0,0 +1,25 @@
...
+	if dataDir == "" {
+		ret = mock.NewStateStore()
+		log.Warning("using in-mem state store, no node state will be persisted")
+		return ret, nil
+	}
+	ret, err = leveldb.NewStateStore(filepath.Join(dataDir, "statestore"), log)
+	if err != nil {
+		return nil, fmt.Errorf("statestore: %w", err)
+	}
+	return ret, nil
```
janos, line 24: I think that this would be good enough.

```suggestion
	return leveldb.NewStateStore(filepath.Join(dataDir, "statestore"), log)
```

pkg/node/statestore.go
```diff
@@ -0,0 +1,25 @@
...
+	"fmt"
+	"path/filepath"
+
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/statestore/leveldb"
+	"github.com/ethersphere/bee/pkg/statestore/mock"
+	"github.com/ethersphere/bee/pkg/storage"
+)
+
+// InitStateStore will initialze the stateStore with the given data directory.
```
janos, line 13: The comment is not correct completely. The usage of empty dir should be documented as ti is the main purpose of this function.

janos (commented):
Cool. I agree with @ralph-pichler comments.

janos:
> my 2 cents:
> The `node` package is used to construct nodes, so I find it a bit out of place to abstract some of the node creation logic used in node bootstrapping outside of the node package. I might be off here by more than a bit though, so I would be happy to hear @janos opinion on this

Ideally, the node package should construct a node out of only configuration parameters and as less as possible already created services or instances of more complex types. Logger may be a good example as an exception, or keys, but core components should not leave the abstraction of node. My view is that this is not so hard requirement if there are very practical reasons to pull out some of the node dependencies and provide them for the node constructor. But, to keep that at less as possible.

cmd/bee/cmd/deploy.go
paxthemax: Time flies, huh 😁 fixing asap

pkg/node/chain.go
paxthemax: Fixed typo.

pkg/node/chain.go
paxthemax: Updated to use interface. Should the same apply elsewhere in the codebase?

cmd/bee/cmd/deploy.go
paxthemax: Fixed redundant if.

pkg/node/chain.go
paxthemax: Fixed.

pkg/node/statestore.go
paxthemax: Fixed.

pkg/node/statestore.go
paxthemax: Removed redundant error wrap.

pkg/node/statestore.go
paxthemax: Good point, updated the comment to clarify the function in more detail.

pkg/node/chain.go
paxthemax: I would agree, I just wasn't sure if there is a case where we want to initialize the chain backend without explicitly syncing it. I will move it there.

paxthemax:
@janos @acud - just to make sure if I understand correctly: would it be better if we put all of the deploy command functionality into a separate function in `node.go` and just forward the initialized parameters from the `cmd` package - namely the context, state store, signer, and command line parameters?

acud:
Its OK as it is. @janos made his point

pkg/node/chain.go
```diff
@@ -0,0 +1,253 @@
...
+	if !isSynced {
+		log.Infof("waiting to sync with the Ethereum backend")
+		err = transaction.WaitSynced(ctx, backend, maxDelay)
+		if err != nil {
+			err = fmt.Errorf("could not wait to sync with the Ethereum backend: %w", err)
+			return
+		}
+	}
+
+	return
```
ralph-pichler, line 92: is this necessary? (same for the other functions?)

pkg/node/chain.go
```diff
@@ -0,0 +1,253 @@
...
+
+// InitChequebookService will initialize the chequebook service with the given
+// chequebook factory and chain backend.
+func InitChequebookService(
+	ctx context.Context,
+	log logging.Logger,
+	stateStore storage.StateStorer,
+	signer crypto.Signer,
+	chainID int64,
+	backend *ethclient.Client,
```
ralph-pichler, line 143: backend interface

pkg/node/chain.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	return
+}
+
+// InitChequebookFactory will initialize the chequebook factory with the given
+// chain backend.
+func InitChequebookFactory(
+	log logging.Logger,
+	backend *ethclient.Client,
```
ralph-pichler, line 99: backed interface

pkg/node/chain.go
```diff
@@ -0,0 +1,253 @@
...
+	chainID int64,
+	transactionService transaction.Service,
+	factoryAddress string,
+) (
+	chequebookFactory chequebook.Factory,
```
acud, line 104: please no named returns

pkg/node/chain.go
```diff
@@ -0,0 +1,253 @@
...
+	signer crypto.Signer,
+	chainID int64,
+	backend *ethclient.Client,
+	overlayEthAddress common.Address,
+	transactionService transaction.Service,
+	chequebookFactory chequebook.Factory,
+	initialDeposit string,
+) (
+	chequebookService chequebook.Service,
+	err error,
```
acud, line 150: same here

pkg/node/chain.go
```diff
@@ -0,0 +1,253 @@
...
+
+func initChequeStoreCashout(
+	stateStore storage.StateStorer,
+	swapBackend transaction.Backend,
+	chequebookFactory chequebook.Factory,
+	chainID int64,
+	overlayEthAddress common.Address,
+	transactionService transaction.Service,
+) (
+	chequeStore chequebook.ChequeStore,
```
acud, line 190: same

pkg/node/chain.go
```diff
@@ -0,0 +1,253 @@
...
+func InitSwap(
+	p2ps *libp2p.Service,
+	log logging.Logger,
+	stateStore storage.StateStorer,
+	networkID uint64,
+	overlayEthAddress common.Address,
+	chequebookService chequebook.Service,
+	chequeStore chequebook.ChequeStore,
+	cashoutService chequebook.CashoutService,
+) (swapService *swap.Service, err error) {
```
acud, line 229: same

pkg/node/statestore.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/statestore/leveldb"
+	"github.com/ethersphere/bee/pkg/statestore/mock"
+	"github.com/ethersphere/bee/pkg/storage"
+)
+
+// InitStateStore will initialze the stateStore with the given path to the
+// data directory. When given an empty directory path, the function will instead
+// initialize an in-memory state store that will not be persisted.
+func InitStateStore(log logging.Logger, dataDir string) (ret storage.StateStorer, err error) {
```
acud, line 19: same

acud (changes requested):
minor comments

pkg/node/chain.go
```diff
@@ -0,0 +1,253 @@
...
+// InitChain will initialize the Ethereum backend at the given endpoint and
+// set up the Transacton Service to interact with it using the provided signer.
+func InitChain(
+	ctx context.Context,
+	log logging.Logger,
+	stateStore storage.StateStorer,
+	endpoint string,
+	signer crypto.Signer,
+) (
+	backend *ethclient.Client,
```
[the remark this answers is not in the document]
janos: In this case, some of the named returns help to understand the semantics behind types like chainID. But, it is ok with only a comment and no named returns.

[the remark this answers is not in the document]
acud: I agree that using named returns helps with code readability under some circumstances, but I find the setting of the variable and a blank return just not very nice, and makes code branching to become non-linear, plus leaves a lot of room for error (forgot to return?)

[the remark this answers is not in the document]
paxthemax: Is there general style consensus in the bee codebase? I believe I've found examples of both cases. I can make the change (quite easily) but it would be prudent to have a style guide for future work.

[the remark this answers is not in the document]
acud: We use named returns very rarely:
- In the case that there are quite a lot of return values (more than a few, in cases where you cannot infer which kind of object you're getting. e.g. `func someFunc() (int, int, int, int)` would probably be more readable as `func someFunc() (height int, width int, topleft int, bottomright int)`
- In cases where the compiler optimization that it provides actually gives benefit

acud (approved):

ralph-pichler (approved):

PR #1322, opened 2021-02-24
localstore reserve logic

zelig:
the pr also fixes an index leak which happened (ModeGetRequest) when an as yet unsynced chunk was requested. It was put to GC incorrectly but moreover this entry was not removed when the chunk was synced (setSync was called) 

#1346 

AP https://hackmd.io/9ejVZEvSQg69zXG91YsFEQ


acud:
@zelig adding a ton of renaming to fields and methods unnecessarily increases the mental taxation of reviewing an already large changeset. can you please revert all unrelated changes from the scope of this PR?

pkg/localstore/localstore.go
```diff
@@ -276,9 +277,9 @@ func New(path string, baseKey []byte, o *Options, logger logging.Logger) (db *DB
 		return nil, err
 	}
 	// pull index allows history and live syncing per po bin
-	db.pullIndex, err = db.shed.NewIndex("PO|BinID->Hash|Tag", shed.IndexFuncs{
+	db.pullIndex, err = db.shed.NewIndex("PO|BinID->Hash|BatchID", shed.IndexFuncs{
```
acud, line 280: not sure, why is this needed if the information is already on the retrievalDataIndex?

pkg/localstore/mode_put.go
```diff
@@ -85,13 +85,6 @@ func (db *DB) put(mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err e
 			}
 			exist[i] = exists
 			gcSizeChange += c
-
-			if mode == storage.ModePutRequestPin {
```
acud, line 89: I'm reinstating this part since its needed, but I'm not sure how this will behave with the `gcSizeChange` line at 87

pkg/localstore/postage.go
```diff
@@ -0,0 +1,202 @@
...
+		return err
+	}
+	item.Radius = radius
+	if err = db.postage.counts.PutInBatch(batch, item); err != nil {
+		return err
+	}
+	return db.shed.WriteBatch(batch)
+}
+
+func (p *postageBatches) withinRadius(item shed.Item) bool {
```
zelig, line 200: oops this needs to be:

    return p.po(item.Address()) >= item.Radius

pkg/node/node.go
```diff
@@ -258,7 +258,6 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey,
 		return nil, fmt.Errorf("localstore: %w", err)
 	}
 	b.localstoreCloser = storer
```
zelig, line 260: there needs to be a change here. setting up the batchstore with localstore.Unreserve hood

pkg/localstore/mode_put.go
```diff
@@ -259,21 +262,31 @@ func (db *DB) putSync(batch *leveldb.Batch, binIDs map[uint8]uint64, item shed.I
...
+// (the retrieval access index and the gc index)
+func (db *DB) preserveOrCache(batch *leveldb.Batch, item shed.Item, forcePin bool) (gcSizeChange int64, err error) {
+	// item needs to be populated with Radius
+	item2, err := db.postageRadiusIndex.Get(item)
+	if err != nil {
+		return 0, fmt.Errorf("postage radius index: %w", err)
+	}
+	item.Radius = item2.Radius
+
+	if withinRadiusFn(db, item) || forcePin {
```
zelig, line 287: when this is from a simple setsync call we also should not set the access timestamp to now.
only if it is putRequest.

but surely not ike this, instead:
*make access update consistent irrespective of reserve/cache status*. (as reserve can be evicted after a second and then will be deleted even before their earlier cached (and earlier accessed) sister chunks.

pkg/localstore/reserve.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+		batch = new(leveldb.Batch)
+	)
+
+	i, err := db.postageRadiusIndex.Get(item)
+	if err != nil {
+		if !errors.Is(err, leveldb.ErrNotFound) {
+			return err
+		}
+		item.Radius = radius
```
zelig, line 34: the batch needs to save the new radius even if the batch IS found in the postageRadiusIndex.

pkg/localstore/reserve.go
```diff
@@ -0,0 +1,95 @@
...
+			return false, fmt.Errorf("unpin: %w", err)
+		}
+
+		// if the batch is unreserved we should remove the chunk
+		// from the pull index
+		item2, err := db.retrievalDataIndex.Get(item)
+		if err != nil {
+			return false, err
+		}
+		err = db.pullIndex.DeleteInBatch(batch, item2)
```
zelig, line 54: this is not consistent. The pullindex must include chunk batches otherwise nothing will be in the pullindex from any bin outside the neighbourhood radius

pkg/localstore/mode_put.go
```diff
@@ -307,11 +310,17 @@ func (db *DB) setGC(batch *leveldb.Batch, item shed.Item) (gcSizeChange int64, e
 		return 0, err
 	}
 	if !ok {
-		err = db.gcIndex.PutInBatch(batch, item)
-		if err != nil {
+		exists, err := db.gcIndex.Has(item)
```
zelig, line 313:
```go
if ok {
    return gcSizeChange, nil
}
exists, err := db.gcIndex.Has(item)
```

pkg/localstore/mode_put.go
```diff
@@ -307,11 +310,17 @@ func (db *DB) setGC(batch *leveldb.Batch, item shed.Item) (gcSizeChange int64, e
...
+		if err != nil && !errors.Is(err, leveldb.ErrNotFound) {
 			return 0, err
 		}
-		gcSizeChange++
+		if !exists {
```
zelig, line 317:
```
if exists {
    return 0, nil
}
```

pkg/localstore/mode_set.go
```diff
@@ -179,65 +175,51 @@ func (db *DB) setSync(batch *leveldb.Batch, addr swarm.Address, mode storage.Mod
...
-	}
-
-	// Add in gcIndex only if this chunk is not pinned
-	ok, err := db.pinIndex.Has(item)
-	if err != nil {
-		return 0, err
-	}
-	if !ok {
-		err = db.gcIndex.PutInBatch(batch, item)
+		item.AccessTimestamp = now()
```
zelig, line 183: TODO: here please take note of this. This should leave the access timestamp alone.

zelig: and the following line also dropped

pkg/localstore/reserve_test.go
```diff
@@ -0,0 +1,278 @@
...
+		err = db.Set(context.Background(), storage.ModeSetSync, ch.Address())
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	var gcChs []swarm.Chunk
+	for i := 0; i < 100; i++ {
+		gcch := generateTestRandomChunkAt(swarm.NewAddress(db.baseKey), 2).WithBatch(2, 3)
+		err := db.UnreserveBatch(gcch.Stamp().BatchID(), 2)
```
zelig, line 212: maybe a better test would be to add a few chunks of at PO 4 and show they are intact.
otherwise this is a very good test`

pkg/localstore/mode_get.go
```diff
@@ -151,26 +151,28 @@ func (db *DB) updateGC(item shed.Item) (err error) {
...
-	ok, err := db.pinIndex.Has(item)
-	if err != nil {
-		return err
-	}
-	if !ok {
+	updatedTimestamp := false
+	_, err = db.gcIndex.Get(item)
+	if err == nil {
+		updatedTimestamp = true
+		item.AccessTimestamp = now()
```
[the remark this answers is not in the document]
acud: look 9 lines above?

pkg/localstore/mode_put.go
```diff
@@ -259,46 +271,37 @@ func (db *DB) putSync(batch *leveldb.Batch, binIDs map[uint8]uint64, item shed.I
 	if err != nil {
 		return false, 0, err
 	}
-	gcSizeChange, err = db.setGC(batch, item)
+	err = db.postageChunksIndex.PutInBatch(batch, item)
+	if err != nil {
+		return false, 0, err
+	}
+
+	item.AccessTimestamp = now()
```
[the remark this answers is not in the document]
acud: ?

pkg/localstore/mode_set.go
acud: ....don't understand.... i thought you said setSync should update access timestamp

acud: let's please have a writeup or at least a truth table about what happens to which index under which circumstances. i find documenting this in the code not the best thing to do and it just pollutes the code with hypothetical reminders. let's start with defining, then make changes to the codebase and in the meanwhile try to keep it clean

zelig: here it is quite simple, you update accesstimestamp if and only if the chunk iis requested

zelig (commented):
lgtm

pkg/localstore/mode_set.go
acud: truth table?

acud (approved):

acud:
famous last words?

PR #1334, opened 2021-02-24
fix brew post_install

vandot:
There is an issue when you brew upgrade the package.
Clef is already released with same fix and it works


acud (approved):

Eknir (approved):
lgtm!

PR #1341, opened 2021-02-25
kademlia: return ErrNotFound on skip peers

acud:
As the change suggests, we might be returning `topology.ErrWantSelf` when we've already skipped some peers. This might suggest that we don't have very good connectivity, and the result is that we as a fallback are designated as the chunk storer, resulting in the chunk exiting the push sync index. This can specifically backfire if the node has very degraded connection, resulting in all chunks to be stored on the node, instead of being sent to others.

### DIFF
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -581,7 +581,16 @@ func (k *Kad) ClosestPeer(addr swarm.Address, skipPeers ...swarm.Address) (swarm
 
 	// check if self
 	if closest.Equal(k.base) {
-		return swarm.Address{}, topology.ErrWantSelf
+		switch l := len(skipPeers); l {
+		case 0:
+			return swarm.Address{}, topology.ErrWantSelf
+		default:
+			// we skipped some peers and found that we are
+			// the closest one - return an ErrNotFound,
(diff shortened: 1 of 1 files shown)


zelig (approved):

pkg/kademlia/kademlia.go
```diff
@@ -581,7 +581,16 @@ func (k *Kad) ClosestPeer(addr swarm.Address, skipPeers ...swarm.Address) (swarm
 
 	// check if self
 	if closest.Equal(k.base) {
-		return swarm.Address{}, topology.ErrWantSelf
+		switch l := len(skipPeers); l {
```
zelig, line 584: on second  thought this is not  a good idea. 
also disconnected peers etc clearly shows this should be an iterator function.
either way, checking  the length of skipPeers is a hack and not the right thing.
For instance if we had to prepopulate skipPeers with some nodes, which seems legit, suddenly, 
the closest node will not be found. 

Instead you need to check closest before you  check skippeers and  simply never return errwantself iff there is a closer peer even if it is to be skipped.

zelig (changes requested):

Eknir:
Is this issue merged?

PR #1352, opened 2021-03-01
Limit incoming connections

metacertain:
Introducing the oversaturation limit and disconnections based on

### DIFF
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -32,11 +32,10 @@ var (
 	errMissingAddressBookEntry    = errors.New("addressbook underlay entry not found")
 	errOverlayMismatch            = errors.New("overlay mismatch")
 	errDisconnectByOverSaturation = errors.New("connection refused for oversaturated bin")
-
-	timeToRetry         = 60 * time.Second
-	shortRetry          = 30 * time.Second
-	saturationPeers     = 4
-	overSaturationPeers = 32
+	timeToRetry                   = 60 * time.Second
+	shortRetry                    = 30 * time.Second
+	saturationPeers               = 4
+	overSaturationPeers           = 32
 )
 
 type binSaturationFunc func(bin uint8, peers, connected *pslice.PSlice) bool
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -29,12 +29,14 @@ const (
 )
 
 var (
-	errMissingAddressBookEntry = errors.New("addressbook underlay entry not found")
-	errOverlayMismatch         = errors.New("overlay mismatch")
-	timeToRetry                = 60 * time.Second
-	shortRetry                 = 30 * time.Second
-	saturationPeers            = 4
-	overSaturationPeers        = 32
+	errMissingAddressBookEntry    = errors.New("addressbook underlay entry not found")
+	errOverlayMismatch            = errors.New("overlay mismatch")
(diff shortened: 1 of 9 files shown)


ldeffenb (commented):
Ignore this, I guess I was reading the order of the commits backwards and the "uncommitted changes" certainly didn't invite me to inspect that one!

Shouldn't the following be binOverSaturated?

	if o.OverSaturationFunc == nil {
		o.OverSaturationFunc = binSaturated
	}

ldeffenb (commented):
Shouldn't the following be binOverSaturated?

	if o.OverSaturationFunc == nil {
		o.OverSaturationFunc = binSaturated

And I'm not a go programmer, but is there a difference between these two when the connect() function is returning an error?  What does an empty return actually return if not nil?  And I didn't notice any changes that will cause the over saturation incoming (supposedly) connection to actually be disconnected?  It's just not added to our structures leaving it an orphan or something?

		k.notifyPeerSig()
		return nil
	}
	return
	}

pkg/kademlia/kademlia.go
```diff
@@ -34,9 +34,10 @@ var (
 	timeToRetry                = 60 * time.Second
 	shortRetry                 = 30 * time.Second
 	saturationPeers            = 4
+	overSaturationPeers        = 16
 )
 
-type binSaturationFunc func(bin uint8, peers, connected *pslice.PSlice) bool
+type binSaturationFunc func(bin uint8, peers, connected *pslice.PSlice) (bool, bool)
```
janos, line 40: It would be good to comment on what return values represent.

pkg/kademlia/kademlia.go
```diff
@@ -451,8 +452,19 @@ func (k *Kad) AddPeers(ctx context.Context, addrs ...swarm.Address) error {
 	return nil
 }
 
+func (k *Kad) Pick(peer p2p.Peer) bool {
+	po := swarm.Proximity(k.base.Bytes(), peer.Address.Bytes())
+	_, oversaturated := k.saturationFunc(po, k.knownPeers, k.connectedPeers)
+	return oversaturated
```
janos, line 458: Kademlia pick function would return true here if the bin is oversaturated. Is this expected behaviour for pick? Especially in the context of usage in libp2p notifier in handshake handler.

acud: nice catch. added this part to the unit test that checks the relevant functionality

janos (approved):
LGTM. Thanks @metacertain and @acud!

jmozah:
@metacertain  @Eknir will this limit apply to qBee's too? 
If yes, will the possibility of getting qBZZ come down because of this limit?

ldeffenb:
I've been testing this fix with the branch merged into my build and have uncovered an unfortunate side effect.

I run multiple nodes and use the Debug API POST /connect/<MultiAddr> to keep all of my peers connected to each other.  Unfortunately with this change, /connect no longer works if the target node has already oversaturated the bin of the connecting node.

For bee node farms to be able to explicitly manage their internal cross-connections, I think that the /connect API should be able to exceed the incoming limit on the target node.

acud:
> For bee node farms to be able to explicitly manage their internal cross-connections

I don't see why this should be necessary at all. Different nodes have different addresses which are in different neighborhoods, therefore those nodes should connect to nodes in their address space neighbors, not data-center neighbors, and this property should not be affected by this PR.

zelig (approved):
LGTM

Issue #1295, opened 2021-02-19
Find out status of Ethereum connection and chequebook deployment


Eknir:
# User story

## Story

As a user who is setting up his/her bee node, I want to view in a browser application whether I am connected to Ethereum; whether my chequebook is deployed; whether my node has sufficient eth/BZZ, such that I don't need to look in the logs on clues how to setup my bee node and I can get instructions on this browser page when something is not yet working

## Acceptance criteria

There exists an API endpoint for:
   - connection with Ethereum
   - gETH on overlay
   - gBZZ on overlay
And these endpoints are exposed before the chequebook is deployed

## Background

We want to put less emphasis on our logs and enable bee dashboards to give rich experience on the status of the Bee node. Above-mentioned statuses are important to help the user in starting the Bee node.
Related project: https://hackmd.io/SZ-YbMRDSUytRjhcxg6Xtg?both 
Currently, it is not possible to know all this, because the API router is only instantiated after the chequebook is deployed. That order should be the other way around in order to support his

## Tasks

| Task | Assignee |  Done (1), or 0 |
| ------ |------------| -----|

vojtechsimetka:
Created repo where the development will take place https://github.com/ethersphere/bee-status

Eknir:
To which Ethereum network am I connected?

acud:
I'm not sure whether this is needed and what is the purpose here. It would be difficult to assert whether the node has a living eth backend and display the progress of the chequebook deployment. Some information is already being displayed in the terminal.

acud:
@agazso @vojtechsimetka

agazso:
I think this has been fixed and the debug API is initialized before the node is connected to the blockchain endpoint or other nodes. This may be problematic again in the future when the Bebug and Bee APIs are going to be merged.

This is going to be more important in Milestone 2 but I think we can close it for now.

acud:
Thanks @agazso

PR #1358, opened 2021-03-01
debug api basic and full routing

janos:
This PR introduces a two step Debug API router configuration.

Currently, debug api http listener is exposed at the end of the node starting process, when all of the services are constructed. This is a problem if there is a long running process, like localstore migration, and the /health endpoint is not exposed for a long time for kubernetes health check. This PR creates a debug api listener at the start of node initialization, exposing /health, /addresses and some other available debug endpoints, like metrics, pprof and vars. After all of the debug api dependencies are constructed, debug api service is configured with them and a full router is set.

This PR is with the minimal set of changes for this to be accomplished. A followup PR can be done to remove debugapi.Service interface and expose server in that package to avoid Configure method to be in the interface also. The interface is not needed. This change would require to unexport fields in the server struct which would raise the number of changed lines significantly here. I am avoiding that.

### DIFF
--- a/pkg/debugapi/debugapi.go
+++ b/pkg/debugapi/debugapi.go
@@ -10,6 +10,7 @@ package debugapi
 import (
 	"crypto/ecdsa"
 	"net/http"
+	"sync"
 	"unicode/utf8"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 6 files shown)


pkg/debugapi/debugapi.go
```diff
@@ -30,33 +31,32 @@ import (
 
 type Service interface {
 	http.Handler
+	Configure(overlay swarm.Address, publicKey, pssPublicKey ecdsa.PublicKey, ethereumAddress common.Address, p2p p2p.DebugService, pingpong pingpong.Interface, topologyDriver topology.Driver, storer storage.Storer, tags *tags.Tags, accounting accounting.Interface, settlement settlement.Interface, chequebookEnabled bool, swap swap.ApiInterface, chequebook chequebook.Service)
```
janos, line 34: As explained in PR description, this interface should be removed for simplification and to avoid to have this method here.

Eknir:
Hey Janos: This is related: https://github.com/ethersphere/bee/issues/1295 Can that be completed after your PR?

pkg/debugapi/debugapi.go
```diff
@@ -103,28 +103,49 @@ func equalASCIIFold(s, t string) bool {
...
-		metricsRegistry:   newMetricsRegistry(),
-		ChequebookEnabled: chequebookEnabled,
-		Chequebook:        chequebook,
-		Swap:              swap,
-	}
+// New creates a new Debug API Service with only basic routers enabled in order
+// to expose /health endpoint, Go metrics and pprof. It is useful to expose
+// these endpoints before all dependencies are configured and injected to have
+// access to basic debugging tools and /health endpoint.
+func New(logger logging.Logger, tracer *tracing.Tracer, corsAllowedOrigins []string) Service {
```
acud, line 110: Minor suggestion: I think that `overlay swarm.Address, publicKey, pssPublicKey ecdsa.PublicKey, ethereumAddress common.Address` can already be set in this constructor as they are static and not expected to change during runtime.

acud (approved):
Very cool :clap: Thanks @janos

svetomir (approved):

pkg/debugapi/debugapi.go
janos: That is a very good suggestion. The only problem is `ethereumAddress common.Address` which needs to wait for `InitChain` function to complete. I am suggesting to have explicit `SetEthereumAddress` method and move other address arguments to constructor as you suggested. The `EthereumAddress` would require lock, but that is not so drastic.

This way there would be a debug api available as soon as possible for debugging and health and also with /addresses endpoint exposed, but ethereum address would be set only if and after `InitChain` is done. This would address some of the points in https://github.com/ethersphere/bee/issues/1295 that @Eknir mentioned. WTYT @acud @Eknir?

acud: Actually, I don't think that you need `InitChain` for this. You can have the ethereum address from the signer which is passed to `NewBee`, exactly as it is used in `InitChain`

janos: Yes, correct. Great. :)

janos:
PR is updated to expose `/addresses` on node start before all other services which is a nice consequence of the change suggested by @acud.

The documentation should be updated with the new behaviour after this PR is merged. Underlay addresses array in `/addresses` response is empty until p2p service is constructed, including the nat relay.

acud (approved):
:clap:

svetomir (approved):

PR #1359, opened 2021-03-01
set correct logger in node initchain

janos:
This PR removes the dependency on gopkg.in/src-d/go-log.v1 by using the logger variable in node package. It is my assumption that the logger was just set to a different on by mistake. I may be wrong.


ralph-pichler:
@janos this change is already include in pr #1357 as well

janos:
Thanks @ralph-pichler. Closing in favour of https://github.com/ethersphere/bee/pull/1357.

PR #1361, opened 2021-03-01
chequebook: remove use of bindings for factory

ralph-pichler:
this is the second in a series of PRs which remove the use of abigen-generated bindings in favour of just using the few things we need directly building on top of #1360.

This PR
* removes the use of bindings in `factory.go` instead doing the call via the transaction service (`Call` was copied from `storage-incentives`)
* adds abi-aware helpers to the `transaction` mock for more ergonomic checking of wether the `TxRequest`s are correct.
* adds the `transactionmock.WithABICallSequence` helper which is not used directly in this PR but in follow-ups where functions with more than one call need to be tested.

### DIFF
--- a/pkg/settlement/swap/chequebook/factory.go
+++ b/pkg/settlement/swap/chequebook/factory.go
@@ -20,6 +20,7 @@ import (
 var (
 	ErrInvalidFactory       = errors.New("not a valid factory contract")
 	ErrNotDeployedByFactory = errors.New("chequebook not deployed by factory")
+	errDecodeABI            = errors.New("could not decode abi data")
 
 	factoryABI                  = transaction.ParseABIUnchecked(simpleswapfactory.SimpleSwapFactoryABI)
 	simpleSwapDeployedEventType = factoryABI.Events["SimpleSwapDeployed"]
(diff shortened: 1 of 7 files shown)


pkg/settlement/swap/chequebook/factory.go
```diff
@@ -121,12 +113,25 @@ func (c *factory) VerifyBytecode(ctx context.Context) (err error) {
...
+	if err != nil {
+		return err
+	}
+
+	results, err := factoryABI.Unpack("deployedContracts", output)
 	if err != nil {
 		return err
 	}
+
+	deployed := *abi.ConvertType(results[0], new(bool)).(*bool)
```
acud, line 134: this can potentially panic

pkg/settlement/swap/chequebook/factory.go
```diff
@@ -135,13 +140,25 @@ func (c *factory) VerifyChequebook(ctx context.Context, chequebook common.Addres
...
 		return common.Address{}, err
 	}
-	return erc20Address, nil
+
+	results, err := factoryABI.Unpack("ERC20Address", output)
+	if err != nil {
+		return common.Address{}, err
+	}
+
+	return *abi.ConvertType(results[0], new(common.Address)).(*common.Address), nil
```
acud, line 161: here too

acud (approved):
Looks OK. Minor comments

zelig (approved):

pkg/settlement/swap/chequebook/factory.go
ralph-pichler: yes, but that is how those abi functions are supposed to be used, it is exactly the same code as you would find in the bindings. the fact the unpack did not error should guarantee that this conversion will work. Also `abi.ConvertType` already panics internally if the cast is not possible, so even if we checked something here it would already be too late.

ralph-pichler: I added some basic sanity checks for now.

pkg/settlement/swap/chequebook/factory.go
ralph-pichler: same as above

PR #1362, opened 2021-03-01
chequebook: remove erc20 bindings

ralph-pichler:
this is the third in a series of PRs which remove the use of abigen-generated bindings in favour of just using the few things we need directly building on top of #1361.

This PR
* removes use of bindings for erc20
* introduces an erc20 service instead which uses go-ethereum abi directly (this has been extracted as some of its functions are also needed in `storage-incentives` where some code is currently duplicated)

### DIFF
--- a/pkg/settlement/swap/chequebook/init.go
+++ b/pkg/settlement/swap/chequebook/init.go
@@ -10,9 +10,9 @@ import (
 	"math/big"
 	"time"
 
-	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/settlement/swap/erc20"
 	"github.com/ethersphere/bee/pkg/settlement/swap/transaction"
 	"github.com/ethersphere/bee/pkg/storage"
 )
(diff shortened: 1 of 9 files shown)


pkg/settlement/swap/erc20/erc20.go
```diff
@@ -0,0 +1,92 @@
...
+func (c *erc20Service) Transfer(ctx context.Context, address common.Address, value *big.Int) (common.Hash, error) {
+	callData, err := erc20ABI.Pack("transfer", address, value)
+	if err != nil {
+		return common.Hash{}, err
+	}
+
+	request := &transaction.TxRequest{
+		To:       &c.address,
+		Data:     callData,
+		GasPrice: nil,
```
acud, line 81: if the `GasPrice` is `nil` and `GasLimit` is `0` then what happens to the transaction once it gets sent? Not really clear what is the expected behavior here?

acud (commented):
Few minor comments

pkg/settlement/swap/erc20/erc20.go
ralph-pichler: those are the values for the transaction request, not the transaction itself. gas price nil means the transaction service will query the ethereum node for a gas price, gas limit 0 means it will do a gas estimation.

pkg/settlement/swap/chequebook/init.go
```diff
@@ -138,7 +131,7 @@ func Init(
 		if err == storage.ErrNotFound {
 			logger.Info("no chequebook found, deploying new one.")
 			if swapInitialDeposit.Cmp(big.NewInt(0)) != 0 {
-				err = checkBalance(ctx, logger, swapInitialDeposit, swapBackend, chainId, overlayEthAddress, erc20BindingFunc, erc20Address, 20*time.Second, 10)
+				err = checkBalance(ctx, logger, swapInitialDeposit, swapBackend, chainId, overlayEthAddress, erc20Service, 20*time.Second, 10)
```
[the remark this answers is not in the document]
ralph-pichler: done. as they are constants I also removed them as arguments.

pkg/settlement/swap/erc20/erc20.go
acud: cool. maybe good to document this inline?

acud (approved):

zelig (approved):
neat

PR #1373, opened 2021-03-03
hive: add metrics

acud:


pkg/hive/hive.go
```diff
@@ -153,9 +162,7 @@ func (s *Service) peersHandler(ctx context.Context, peer p2p.Peer, stream p2p.St
 	}
 
 	if s.addPeersHandler != nil {
-		if err := s.addPeersHandler(ctx, peers...); err != nil {
-			return err
-		}
+		return s.addPeersHandler(ctx, peers...)
```
janos, line 165: This is changing how branching is done. It is intended to continue with function execution if `addPeersHandler` returns nil, even though there is not much to be done currently, but this code may be overlooked in the future if something else is added after this if block.

janos (approved):
👍 Only one small comment.

ralph-pichler (approved):

PR #1313, opened 2021-02-22
clef: allow specifying account

ralph-pichler:
add a flag to enable specifying which clef account to use. It's important that if this is specified it also needs to be specified on any further run of that bee node. in a future pr the used overlay should be stored and checked.

this helps against the following problems:
* new keys are added to clef and the original is no longer at index 0
* multiple bees using the same clef

PR #1377, opened 2021-03-03
prevent nodes from starting with a different overlay

ralph-pichler:
currently a bee node will happily start up with a different overlay key which breaks accounting, settlement and probably a ton of other stuff. this pr saves the expected overlay address on first boot and then checks it on every further startup. the check happens on `start`, `init` and `deploy` command.

this prevents user errors such as
* deleting swarm.key but not statestore
* changing key in clef but not deleting statestore
* adding key to clef which takes index 0 in the account list but not deleting statestore or specifying the account explicitly
* specifying another account with `clef-signer-ethereum-address` (from PR #1313) on a second run.

### DIFF
--- a/cmd/bee/cmd/init.go
+++ b/cmd/bee/cmd/init.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/node"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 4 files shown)


cmd/bee/cmd/init.go
```diff
@@ -41,7 +42,24 @@ func (c *command) initInitCmd() (err error) {
...
+				return err
+			}
+
+			defer stateStore.Close()
+
+			err = node.CheckOverlayWithStore(signerConfig.address, stateStore)
+			if err != nil {
+				return err
+			}
+
```
janos, line 62:
```suggestion
			return node.CheckOverlayWithStore(signerConfig.address, stateStore)

```

pkg/node/statestore.go
```diff
@@ -24,3 +26,22 @@ func InitStateStore(log logging.Logger, dataDir string) (ret storage.StateStorer
...
+	err := storer.Get(overlayKey, &storedOverlay)
+	if err != nil {
+		if !errors.Is(err, storage.ErrNotFound) {
+			return err
+		}
+		return storer.Put(overlayKey, overlay)
+	}
+
+	if !storedOverlay.Equal(overlay) {
+		return errors.New("overlay changed")
```
janos, line 44: Just a suggestion. It would be good to include expected and found overlay addresses in the error message.

janos (approved):
Only minor comments. Looks good.

cmd/bee/cmd/init.go
ralph-pichler: fixed

pkg/node/statestore.go
ralph-pichler: fixed

acud (approved):
LGTM

Issue #1286, opened 2021-02-18
systemd service file should specify nice level


jpritikin:
I suggest,

```
[Service]
Nice=10
```

I like nice bees.

Eknir:
@jpritikin , could you please provide more context?

jpritikin:
If I'm going to run bee on my work laptop then I don't want bee to hurt performance of other stuff that I'm doing.

jpritikin:
Actually ionice may be equally important, https://web.archive.org/web/20150423145758/http://b.zekjur.net/2012/09/23/systemd-ionice.html

jpritikin:
I encourage adding [memory limits](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#LimitCPU=) too. However, I'm not sure what to recommend as defaults.

jpritikin:
I added,

```
ProtectHome=yes
```

It would be great if more sandboxing protections could be enabled.

vandot:
Hi @jpritikin thank for submitting this bug report.
Bee's systemd service file is placed inside `/lib/systemd/system/` folder, you are free to customize it as much as you want and just place custom systemd service file inside `/etc/systemd/system/` and systemctl will use that one for managing your service.
In this way your changes will not be overwritten with bee updates.
I will close this ticket, you are free to comment on it or to create new one if needed.

jpritikin:
I encourage you to at least add my suggested customizations as **comments** in the distributed service file. I'm not a systemd expert; it took me hours of research just to come up with the tweaks that I recorded here.

PR #1378, opened 2021-03-03
Suggest how to minimize the impact of cpu & i/o demands (#1286)

jpritikin:
This took me a few hours to figure out. It's definitely worth a comment.


Eknir:
@vandot , can you have a look at this?

acud:
This can indeed work but it will degrade service you're giving out to other peers. We've made (and are still making) significant changes to the bee codebase so that such changes should not be necessary. All in all, slow peer behavior should be discouraged for the sake of network reliability and performance, and should be sanctioned in the medium term by just disconnecting or de-prioritizing peers for retrievals or syncing.

PR #1379, opened 2021-03-03
some housekeeping

attila-lendvai:
mostly just rewordings, but

- fixes the Bzzaar URL's status=%d
- turns an error log into a warning

### DIFF
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -254,7 +254,7 @@ func waitForClef(logger logging.Logger, maxRetries uint64, endpoint string) (ext
 			return nil, err
 		}
 		maxRetries--
-		logger.Errorf("cannot connect to clef signer: %v", err)
+		logger.Warningf("failing to connect to clef signer: %v", err)
 
 		time.Sleep(5 * time.Second)
 	}
--- a/packaging/bee-get-addr
+++ b/packaging/bee-get-addr
@@ -2,28 +2,37 @@
 
 if [ ! `id -u` -eq 0 ] ; then
     echo "
-use sudo!
+This script requires root priviledges, use sudo.
 "
     exit 1
 fi
 
 RESP=$(/usr/bin/bee init --config /etc/bee/bee.yaml 2>&1)
 case "$RESP" in
-    Error*)
+    Error*|*'No such file'*)
         echo "
-Enabled clef-signer but clef is not running.
+The bee node has returned an error, probably because the bee-clef external signer is enabled, but the node couldn't connect to it.
+
 Check https://docs.ethswarm.org/ for more info.
 
-After you fix configuration run 'sudo bee-get-addr' again.
+Fix the configuration and try again.
+
+The node's output was:
         "
+        echo "$RESP"
         ;;
     *)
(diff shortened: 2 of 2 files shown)


cmd/bee/cmd/start.go
```diff
@@ -254,7 +254,7 @@ func waitForClef(logger logging.Logger, maxRetries uint64, endpoint string) (ext
 			return nil, err
 		}
 		maxRetries--
-		logger.Errorf("cannot connect to clef signer: %v", err)
+		logger.Warningf("failing to connect to clef signer: %v", err)
```
acud, line 257: I think the wording on the left is better

acud (approved):
Minor comments. Otherwise good to go

cmd/bee/cmd/start.go
attila-lendvai: @acud "on the left"? you mean the original? my problem with the original is that it doesn't communicate that it's an ongoing process, and that the failure is temporary. hence my issue with the use of `Errorf`, and the original wording that suggests that it's something final.

note that this change was motivated by an actual confusion of mine, that lasted until i looked at the code.

attila-lendvai: an alternative: `trying to connect to clef signer, but failing due to error %v`

acud: This is not really in the English standard-library :) trying and failing at the same time? failed since the failure is already known therefore in past tense. Sorry for the grammar harassment.  

> "on the left"? you mean the original

yes

zelig (approved):

cmd/bee/cmd/start.go
attila-lendvai: the grammar of spoken languages is not set in stone... ;) i don't care how, but what i pointed out should be expressed in a way that is not confusing. the original wording suggests a final, failed situation.

i still think it's clear english: the trying is still happening, continuously, but it also keeps failing, for now. but it's rather funny when two non-native speakers get into such discussions... :)

attila-lendvai:
a question: who is supposed to press the merge button?

1. the PR opener, once it has been enabled by the two approvals?
2. or shall i just leave it to the regular devs, who go through the PRs every once in a while?

cmd/bee/cmd/start.go
acud: right. let's ask a native then. @significance?

Eknir: No strong opinion here. I would like to ask @significance to cut the ties here.

Eknir (approved):
lgtm! Thanks @attila-lendvai . For the grammar issue; no strong opinion here. I think it was good as it was, but I am also not against your change.

cmd/bee/cmd/start.go
significance: i guess maybe @attila-lendvai's rephrased alternative is the most proper? `trying to connect to clef signer, but failing due to error %v` but i think any would work for the use case presuming they are appearing multiple times in the logs?

personal vote goes to @acud's lolzcatz version 😹

attila-lendvai: FTR, it usually appears only once in the log, because at the first retry clef is already ready and accepting connections.

PR #1383, opened 2021-03-04
streamtest: add option to add a correct base address

acud:
This rectifies the mistake we had in the recorder, where the `p2p.Peer` that we called the handler call with was the wrong address, carrying over the destination address, instead of the origin address of the message.

There is still one todo here, and that is to rewrite the preemptive retry test to use mocked server responses instead of the current behavior which is injected through a middleware if statement. Since this change is a blocker for @esadakar's PR, I'd rather have it in so he can rebase, then we can mend the skipped test.

### DIFF
--- a/pkg/pushsync/pushsync_test.go
+++ b/pkg/pushsync/pushsync_test.go
@@ -357,8 +357,6 @@ func TestHandler(t *testing.T) {
 		t.Fatalf("unexpected balance on trigger. want %d got %d", -int64(fixedPrice), balance)
 	}
 
-	// we need to check here for pivotPeer instead of triggerPeer because during
-	// streamtest the peer in the handler is actually the receiver
 	balance, err = pivotAccounting.Balance(triggerPeer)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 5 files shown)


pkg/accounting/mock/accounting.go
```diff
@@ -153,6 +154,7 @@ func (s *Service) Debit(peer swarm.Address, price uint64) error {
 
 // Balance is the mock function wrapper that calls the set implementation
 func (s *Service) Balance(peer swarm.Address) (*big.Int, error) {
+	fmt.Println(s.balances)
```
esadakar, line 157: I can the print statement in my PR

esadakar (approved):

pkg/accounting/mock/accounting.go
acud: reverted :+1:

pkg/pushsync/pushsync_test.go
```diff
@@ -358,8 +357,9 @@ func TestHandler(t *testing.T) {
 		t.Fatalf("unexpected balance on trigger. want %d got %d", -int64(fixedPrice), balance)
 	}
 
-	// we need to check here for pivotPeer instead of triggerPeer because during streamtest the peer in the handler is actually the receiver
-	balance, err = pivotAccounting.Balance(pivotPeer)
+	// we need to check here for pivotPeer instead of triggerPeer because during
```
ralph-pichler, line 360: you may want to remove the comment as the hack is no longer necessary.

janos (approved):
LGTM

PR #1384, opened 2021-03-04
localstore: improve CPU usage due to leveldb iterator

acud:
Bee is demonstrating high CPU usage due to open (push and pull) subscriptions and these are being triggered whenever something is written to the db, usually incoming from either a pushsync incoming message or a pullsync incoming message.

The notification to the subscriptions all launch simultaneously, bringing up potentially hundreds of iterators online at once. So far so good, but the problem is that usually whenever there's a write, there would also be another write very soon. Examples are with pull sync that writes each delivery separately, and pushsync does the same. So ideally we'd be waking these subscriptions up after a batch was written in pullsync for example, but the current design does not allow this. With pushsync the situation is a bit different, since deliveries _should_ be more sporadic.

My approach would be therefore, to apply some sort of buffering to the way we notify the subscriptions about new data to query, with the hope that the actual notification comes in after the batch has finished writing, allowing the iterators to iterate without the risk of creating those merged iterators that are causing the high CPU usage.

The downside would be some slowdown to pull-syncing, but with the incoming puhsync spread PR this should not be a big issue, and I would argue that the performance hit should not be so big, since larger batches could potentially be written on live syncing.

I borrow the `falling edge detection` approach from how DSP is done on digital signal detection, to detect when a signal goes from `on` to `off` state. This is a common technique and feature used on oscilloscopes and DSP analyzers. So in this case I applied this same technique to basically allow multiple writes into the channel (which is buffered), but allowing for some sort of a time window to pass before actually writing to the output of that channel. I have written a simple library specifically to do this [here](https://github.com/acud/flipflop). It also defines a max time window, to write to the output channel anyway, just in case signals just keep on coming and coming endlessly, resulting in the buffer to never dry up and actually write to the output channel.

This is still work in progress, but so far tests have shown improvement. Still need to home in on the actual correct settings in terms of timeouts, but this should serve as a baseline and an entry point to a discussion of whether this is a viable solution or not.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/ethersphere/bee
 go 1.14
 
 require (
-	github.com/acud/flipflop v0.0.2
 	github.com/btcsuite/btcd v0.21.0-beta
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
 	github.com/coreos/go-semver v0.3.0
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/ethersphere/bee
 go 1.14
 
 require (
+	github.com/acud/flipflop v0.0.2
 	github.com/btcsuite/btcd v0.21.0-beta
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
 	github.com/coreos/go-semver v0.3.0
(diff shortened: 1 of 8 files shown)


ldeffenb:
Could this be an explanation for what I just observed in the pushsync handler?  A single peer is pushing a single chunk into my node multiple times in relatively quick succession.  This is not only a load on my incoming bandwidth to redundantly receive the chunk, but the original pusher is also pushing it out multiple times to me!

I looked at the client-side of pushsync, but haven't nailed down any particular cause for this behavior yet.

You can see a log of this by browsing the following CID at https://gateway.ethswarm.org/

3b2ca13a7637d278b1b6e58eb076653e688f903c8893409b03ed68c9d8d4fe23

That log file shows a single chunk being received from a few peers, many times from a single peer.   I understand that I might receive a chunk from multiple peers, but multiple times from a single peer doesn't seem right.

acud:
@ldeffenb regretfully there's no connection between the two. Once we improve the CPU usage I'll be looking into the issue you've mentioned.

janos (approved):
Thanks, nice @acud!

Issue #1104, opened 2021-01-12
Prevent accidental non-syncing


Eknir:
# User story

## Story

As a storer node, I only want to send out a storage receipt after I send the chunk to my neighborhood peers, such that the receipt which I give becomes more reliable estimator for the uploader and the content does not get lost when I shut down my node and I did not sync the content to my peers

## Acceptance criteria

- If the node is shut down immediately after sending the receipt, the content is still retrievable

## Background

With this issue in place, we prevent accidental deletion of chunks and we make the network more robust

## Tasks

| Task | Assignee |  Done (1), or 0 |
| ------ |------------| -----|
|  Make architecture proposal | unassigned | 0 |

PR #1176, opened 2021-01-28
pushsync: clean up

acud:
There's quite a bit of overlap in pushsync package logic. `handler` makes use of the same logic as `PushToClosest`. This PR changes the package to reuse that logic. It will also make use of the retry logic which is present in the `PushToClosest` method, resulting in retries when forwarding pushed chunks.

Also, some tests needed some clean up since they resulted in flow control events which are unnatural to the package in production.

PR #1385, opened 2021-03-04
storer nodes send chunks to closest peers as the receipt is sent back  #1104

esadakar:
- [x]  some tests are broken, waiting on acud's PR before this can be merged

See issue #1104
This change makes sure that as the receipt is being back from the storing node during pushsync, the chunk is sent off to N neighbor peers in parallel so that chunks can be replicated immediately. 

Balance is also updated amongst the neighbor nodes.

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -50,9 +50,10 @@ type Receipt struct {
 }
 
 type PushSync struct {
+	address       swarm.Address
 	streamer      p2p.StreamerDisconnecter
 	storer        storage.Putter
-	peerSuggester topology.ClosestPeerer
+	peerSuggester topology.EachPeerer
 	tagger        *tags.Tags
 	unwrap        func(swarm.Chunk)
 	logger        logging.Logger
@@ -63,10 +64,13 @@ type PushSync struct {
 	signer        crypto.Signer
 }
 
-var timeToLive = 5 * time.Second // request time to live
+var timeToLive = 5 * time.Second                      // request time to live
+var timeToWaitForPushsyncToNeighbor = 3 * time.Second // time to wait to get a receipt for a chunk
+var nPeersToPushsync = 3                              // number of peers to replicate to as receipt is sent upstream
 
-func New(streamer p2p.StreamerDisconnecter, storer storage.Putter, closestPeerer topology.ClosestPeerer, tagger *tags.Tags, unwrap func(swarm.Chunk), logger logging.Logger, accounting accounting.Interface, pricer pricer.Interface, signer crypto.Signer, tracer *tracing.Tracer) *PushSync {
+func New(address swarm.Address, streamer p2p.StreamerDisconnecter, storer storage.Putter, closestPeerer topology.EachPeerer, tagger *tags.Tags, unwrap func(swarm.Chunk), logger logging.Logger, accounting accounting.Interface, pricer pricer.Interface, signer crypto.Signer, tracer *tracing.Tracer) *PushSync {
 	ps := &PushSync{
+		address:       address,
 		streamer:      streamer,
 		storer:        storer,
 		peerSuggester: closestPeerer,
@@ -126,6 +130,15 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 		return swarm.ErrInvalidChunk
 	}
 
+	//if p's address is closer to the chunk than my address, then simply store it and return
+	if dcmp, _ := swarm.DistanceCmp(chunk.Address().Bytes(), p.Address.Bytes(), ps.address.Bytes()); dcmp == 1 {
+		_, err = ps.storer.Put(ctx, storage.ModePutSync, chunk)
+		if err != nil {
+			return fmt.Errorf("chunk store: %w", err)
+		}
+		return ps.accounting.Debit(p.Address, ps.pricer.PeerPrice(p.Address, chunk.Address()))
+	}
+
 	span, _, ctx := ps.tracer.StartSpanFromContext(ctx, "pushsync-handler", ps.logger, opentracing.Tag{Key: "address", Value: chunk.Address().String()})
 	defer span.Finish()
 
@@ -151,6 +164,61 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 			if err != nil {
 				return fmt.Errorf("receipt signature: %w", err)
 			}
+
+			// push the chunk to the closest peers in parallel for replication
+			// any errors here should NOT impact the rest of the handler
+			var skipPeers []swarm.Address
+			for i := 0; i < nPeersToPushsync; i++ {
+				peer, err := ps.closestPeer(chunk.Address(), skipPeers, true)
+				if err != nil {
+					ps.logger.Tracef("pushsync replication closest peer: %w", err)
+					break
+				}
+				skipPeers = append(skipPeers, peer)
+				go func(peer swarm.Address) {
+
+					var err error
+					defer func() {
+						if err != nil {
+							ps.logger.Tracef("pushsync replication: %v", err)
+						}
+					}()
+
+					receiptPrice := ps.pricer.PeerPrice(peer, chunk.Address())
+
+					err = ps.accounting.Reserve(ctx, peer, receiptPrice)
+					if err != nil {
+						return
+					}
+					defer ps.accounting.Release(peer, receiptPrice)
+
+					streamer, err := ps.streamer.NewStream(ctx, peer, nil, protocolName, protocolVersion, streamName)
+					if err != nil {
+						return
+					}
+
+					w := protobuf.NewWriter(streamer)
+					ctx, cancel := context.WithTimeout(ctx, timeToWaitForPushsyncToNeighbor)
+					defer cancel()
+
+					err = w.WriteMsgWithContext(ctx, &pb.Delivery{
+						Address: chunk.Address().Bytes(),
+						Data:    chunk.Data(),
+					})
+					if err != nil {
+						_ = streamer.Reset()
+						return
+					}
+
+					err = ps.accounting.Credit(peer, receiptPrice)
+					if err != nil {
+						return
+					}
+
+				}(peer)
+			}
+
+			// return back receipt
 			receipt := pb.Receipt{Address: chunk.Address().Bytes(), Signature: signature}
 			if err := w.WriteMsgWithContext(ctx, &receipt); err != nil {
 				return fmt.Errorf("send receipt to peer %s: %w", p.Address.String(), err)
@@ -211,8 +279,15 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
 
 		defersFn()
 
-		// find next closest peer
-		peer, err := ps.peerSuggester.ClosestPeer(ch.Address(), skipPeers...)
+		deferFuncs = append(deferFuncs, func() {
+			if lastErr != nil {
+				ps.metrics.TotalErrors.Inc()
+				logger.Errorf("pushsync: %v", lastErr)
+			}
+		})
+
+		// find the next closest peer
+		peer, err := ps.closestPeer(ch.Address(), skipPeers, false)
 		if err != nil {
 			// ClosestPeer can return ErrNotFound in case we are not connected to any peers
 			// in which case we should return immediately.
@@ -323,4 +398,53 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
 	}
 
 	return nil, topology.ErrNotFound
+}
+
+// closestPeer returns address of the peer that is closest to the chunk with
+// provided address addr. This function will ignore peers with addresses
+// provided in skipPeers
+func (ps *PushSync) closestPeer(addr swarm.Address, skipPeers []swarm.Address, allowUpstream bool) (swarm.Address, error) {
+	closest := swarm.Address{}
+	err := ps.peerSuggester.EachPeer(func(peer swarm.Address, po uint8) (bool, bool, error) {
+		for _, peerToSkip := range skipPeers {
+			if peerToSkip.Equal(peer) {
+				return false, false, nil
+			}
+		}
+		if closest.IsZero() {
+			closest = peer
+			return false, false, nil
+		}
+		dcmp, err := swarm.DistanceCmp(addr.Bytes(), peer.Bytes(), closest.Bytes())
+		if err != nil {
+			return false, false, fmt.Errorf("distance compare error. addr %s closest %s peer %s: %w", addr.String(), closest.String(), peer.String(), err)
+		}
+		if dcmp == 1 {
+			closest = peer
+		}
+		return false, false, nil
+	})
+
+	if err != nil {
+		return swarm.Address{}, err
+	}
+
+	if closest.IsZero() {
+		return swarm.Address{}, topology.ErrNotFound
+	}
+
+	if allowUpstream {
+		return closest, nil
+	}
+
+	dcmp, err := swarm.DistanceCmp(addr.Bytes(), ps.address.Bytes(), closest.Bytes())
+	if err != nil {
+		return swarm.Address{}, fmt.Errorf("distance compare addr %s closest %s base address %s: %w", addr.String(), closest.String(), ps.address.String(), err)
+	}
+
+	if dcmp == 1 {
+		return closest, topology.ErrWantSelf
+	}
+
+	return closest, nil
 }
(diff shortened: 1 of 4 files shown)


pkg/pushsync/pushsync.go
```diff
@@ -130,6 +144,9 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 				return fmt.Errorf("chunk store: %w", err)
 			}
 
+			// push the chunk to the second closest peers
+			go ps.pushToPeers(ctx, chunk)
```
acud, line 148: this means that only one push will happen at any given point.
I would inline `pushToPeers` here instead and launch every individual secondry send in a separate goroutine. This way all sends happen simultaneously and one send does not depend on the success of the other in terms of timing (some sends might fail, slowing down other sends)

pkg/pushsync/pushsync.go
```diff
@@ -279,3 +296,81 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
...
+	if dcmp == 0 {
+		return closest, topology.ErrWantSelf
+	}
+
+	return closest, nil
+}
+
+func (ps *PushSync) pushToPeers(ctx context.Context, ch swarm.Chunk) (err error) {
+
+	skip := []swarm.Address{ps.address}
```
acud, line 347: this is incorrect - this should be the origin address, not the base address. also, for uniformity, I would add a `origin swarm.Address` and `allowUpstream bool` flag to `pushToClosest` method, and include the origin address when doing the call to `pushToClosest` when forwarding (`allowUpstream=false` in this case). When calling it from `PushToClosest` then `allowUpstream=true`.

pkg/pushsync/pushsync_test.go
```diff
@@ -80,7 +80,85 @@ func TestSendChunkAndReceiveReceipt(t *testing.T) {
...
+	waitOnRecordAndTest(t, closestPeer, recorder, chunk.Address(), chunk.Data())
+
+	// this intercepts the outgoing delivery message from storer node to second storer node
+	waitOnRecordAndTest(t, secondPeer, secondRecorder, chunk.Address(), chunk.Data())
+
+	// this intercepts the incoming receipt message
+	waitOnRecordAndTest(t, closestPeer, recorder, chunk.Address(), nil)
+
+	_, err = storerEmpty.Get(context.Background(), storage.ModeGetSync, chunk.Address())
+	if err != storage.ErrNotFound {
```
acud, line 148: we try to use `errors.Is(err, expectedErr)` whenever possible. it leverages the go error wrapping nicely

acud (commented):
very cool start :) some comments above

pkg/pushsync/pushsync.go
esadakar: right, I agree

pkg/pushsync/pushsync_test.go
esadakar: ok

pkg/pushsync/pushsync.go
esadakar: I believe base address is correct here because we do not want the final storing node to send the chunk to itself again. Origin address should be able to receive the chunk if the storing node decides that it is the closest after itself.

 I need a bit more clarification on the other comments.

acud: A node cannot send to itself. You will never get the base address from the kademlia iterator, if you do - we have a serious issue. This should never happen.

> Origin address should be able to receive the chunk if the storing node decides that it is the closest after itself

That is indeed true

esadakar: ok

acud (commented):
Getting there :)

A few notes:
- About `pushToPeer` - this package was just recently rewritten to avoid code branching and indirection. See [here](https://github.com/ethersphere/bee/pull/1176/files). Also the name is misleading, since it specifically handles the logic of replicating, and `pushToPeer` might suggest a variety of things.
- Some pending TODOs - mainly about logging. Since we don't fail the operation I suggest you log them on `Trace` log level, as they are unhandled errors. 
- Some code is commented out

go.mod
```diff
@@ -70,6 +70,7 @@ require (
 	golang.org/x/tools v0.0.0-20200626171337-aa94e735be7f // indirect
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
 	gopkg.in/ini.v1 v1.57.0 // indirect
+	gopkg.in/src-d/go-log.v1 v1.0.1
```
ralph-pichler, line 73: please rebase on master and make sure this import is not here.

acud: rebasing won't help since the CS is in the branch and will be reapplied. `git checkout master -- go.mod`

pkg/pushsync/pushsync.go
```diff
@@ -130,6 +143,54 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 				return fmt.Errorf("chunk store: %w", err)
 			}
 
+			// push the chunk to the closest peers in parallel for replication
+			// any errors here should NOT impact the rest of the handler
+			var skipPeers []swarm.Address
+			for i := 0; i < nPeersToPushsync; i++ {
+				peer, err := ps.closestPeer(chunk.Address(), skipPeers, true)
+				if err != nil {
+					ps.logger.Tracef("closest peer: %w", err)
```
acud, line 152: consider adding a bit more context here so that we know where it came from. See `pushsync.go#L256`. The same holds for all the other loglines that were added

pkg/pushsync/pushsync.go
```diff
@@ -130,6 +143,54 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+					break
+				}
+				skipPeers = append(skipPeers, peer)
+				go func(peer swarm.Address) {
+					receiptPrice := ps.pricer.PeerPrice(peer, chunk.Address())
+					if err := ps.accounting.Reserve(ctx, peer, receiptPrice); err != nil {
+						ps.logger.Tracef("reserve balance for peer %s: %w", peer.String(), err)
```
acud, line 159: You don't need to log explicitly in every error case. Consider having a variable defined at the beginning of the closure, then a defer call that checks the value of it, logging if necessary:
```go
var err error
defer func() {
if err != nil {
 ps.logger.Tracef("pushsync replication: %v",err)
}
....
err = ps.someCall()
if err != nil { return }
... 

```

pkg/pushsync/pushsync.go
```diff
@@ -130,6 +143,54 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+						return
+					}
+					defer ps.accounting.Release(peer, receiptPrice)
+
+					streamer, err := ps.streamer.NewStream(ctx, peer, nil, protocolName, protocolVersion, streamName)
+					if err != nil {
+						ps.logger.Tracef("new stream for peer %s: %w", peer.String(), err)
+						return
+					}
```
acud, line 168: if you use the `defer` example above, you could just add here:
```go
defer func() {if err != nil { _ = streamer.Reset() } }

```

pkg/pushsync/pushsync_test.go
```diff
@@ -179,23 +223,27 @@ func TestPushChunkToNextClosest(t *testing.T) {
...
-		peer1,
-		peer2,
-	}
+	peer1 := swarm.MustParseHexAddress("6000000000000000000000000000000000000000000000000000000000000000")
+	peer2 := swarm.MustParseHexAddress("5000000000000000000000000000000000000000000000000000000000000000")
 
-	// peer is the node responding to the chunk receipt message
-	// mock should return ErrWantSelf since there's no one to forward to
-	psPeer1, storerPeer1, _, peerAccounting1 := createPushSyncNode(t, peer1, nil, nil, mock.WithClosestPeerErr(topology.ErrWantSelf))
+	// emptyPeer := swarm.MustParseHexAddress("5000000000000000000000000000000000000000000000000000000000000000") // node that satisfies replication
```
acud, line 231: commented code

pkg/pushsync/pushsync_test.go
```diff
@@ -22,37 +23,51 @@ import (
...
 	fixedPrice = uint64(10)
 )
 
-// TestSendChunkAndGetReceipt inserts a chunk as uploaded chunk in db. This triggers sending a chunk to the closest node
-// and expects a receipt. The message are intercepted in the outgoing stream to check for correctness.
-func TestSendChunkAndReceiveReceipt(t *testing.T) {
+// TestReplicateBeforeReceipt tests that a chunk is pushed and a receipt is received.
+// Also the storer node initiates a pushsync to N closest nodes of the chunk as it's sending back the receipt.
+// The second storer should only store it and not forward it. The balance of all nodes is tested.
+func TestReplicateBeforeReceipt(t *testing.T) {
```
acud, line 40: if the original test case still passes after the changes, I would add this specific test as an extra test case

acud (changes requested):
requesting a few more minor changes

pkg/pushsync/pushsync.go
```diff
@@ -58,10 +59,13 @@ type PushSync struct {
 	tracer        *tracing.Tracer
 }
 
-var timeToWaitForReceipt = 3 * time.Second // time to wait to get a receipt for a chunk
+var timeToWaitForReceipt = 3 * time.Second            // time to wait to get a receipt for a chunk
+var timeToWaitForPushsyncToNeighbor = 3 * time.Second // time to wait to get a receipt for a chunk
+var nPeersToPushsync = 3                              // number of peers to pushsync a chunk as the receipt is being sent upstream from storage node
```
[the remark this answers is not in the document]
esadakar: ok

pkg/pushsync/pushsync.go
```diff
@@ -275,3 +349,52 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
...
+	if allowUpstream {
+		return closest, nil
+	}
+
+	dcmp, err := swarm.DistanceCmp(addr.Bytes(), ps.address.Bytes(), closest.Bytes())
+	if err != nil {
+		return swarm.Address{}, fmt.Errorf("distance compare addr %s closest %s base address %s: %w", addr.String(), closest.String(), ps.address.String(), err)
+	}
+
+	if dcmp == 1 {
```
zelig, line 395: ?

pkg/pushsync/pushsync.go
```diff
@@ -275,3 +349,52 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
 
 	return nil, topology.ErrNotFound
 }
+
+// closestPeer returns address of the peer that is closest to the chunk with
```
zelig, line 353: ok now. but can  we please rewrite  these as iterators on kademlia TODO

zelig (changes requested):

pkg/pushsync/pushsync.go
```diff
@@ -119,6 +123,15 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 		return swarm.ErrInvalidChunk
 	}
 
+	//if p's address is closer to the chunk than my address, then simply store it and return
+	if dcmp, _ := swarm.DistanceCmp(chunk.Address().Bytes(), p.Address.Bytes(), ps.address.Bytes()); dcmp == 1 {
+		_, err = ps.storer.Put(ctx, storage.ModePutSync, chunk)
+		if err != nil {
+			return fmt.Errorf("chunk store: %w", err)
```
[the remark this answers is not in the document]
esadakar: this is similar to how the storing node behaves. An error in the Put call, returns from the handler with error.

pkg/pushsync/pushsync.go
```diff
@@ -119,6 +123,15 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+		}
+		return ps.accounting.Debit(p.Address, ps.pricer.Price(chunk.Address()))
```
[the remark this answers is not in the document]
esadakar: I believe this was discussed already here https://beehive.ethswarm.org/swarm/pl/piq7cwuumpda3mzhjy7f8jah8w

pkg/pushsync/pushsync.go
esadakar: I am not sure what you mean.

pkg/pushsync/pushsync.go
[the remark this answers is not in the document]
esadakar: I tried to use the existing code base as a reference as much as possible.

pkg/pushsync/pushsync.go
```diff
@@ -119,6 +123,15 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 		return swarm.ErrInvalidChunk
 	}
 
+	//if p's address is closer to the chunk than my address, then simply store it and return
```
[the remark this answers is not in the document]
esadakar: To summarize, storer node, being closest to the chunk, finds the second closest peer for replication, and sends the chunk. To prevent infinite loops, nodes that are second closest, in other words, nodes that receive chunks from peers with closer addresses, simply store the chunk and exit early out of the handler method. 

Please help me understand if I am missing any details.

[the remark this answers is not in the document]
acud: @zelig the requirement you're setting here is problematic since to determine whether a chunk is in the neighborhood of the node we're getting the chunk from presupposes that node's connectivity which I think will not hold. The other problem is that we'd need to introduce neighborhood depth as a dependency here and I was trying to avoid that as much as possible, with the hope that the pricer actually detects whether some is or not within the NN, returning the correct price then.

pkg/pushsync/pushsync.go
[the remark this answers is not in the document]
zelig: yes it returns but without doing the accounting event in  132.

pkg/pushsync/pushsync.go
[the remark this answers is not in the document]
zelig: > To summarize, storer node, being closest to the chunk, finds the second closest peer for replication, and sends the chunk. To prevent infinite loops, nodes that are second closest, in other words, nodes that receive chunks from peers with closer addresses, simply store the chunk and exit early out of the handler method.
> 
> Please help me understand if I am missing any details.


If a node N receives  a push sync message of chunk C from its peer P:
- and N is itself the peer closest to the chunk C,  then it sends it to all its nearest neighours (which just store the chunk and do no forwarding  or  responding) and (concurrently) responds with  a receipt to P.
- and N is  not  the  closest peer, but the sending peer is within its neighbourhood, i.e., 
     - PO(N,P)  >= Depth(N), then
          - if P is the closest peer  to  C (ie., PO(N,C) > PO(N',C) for all N' peers of P), then N just stores the chunk, does no forwarding or responding
          - if N is closer to C than P (ie., PO(N,C) > PO(P,C)) then N forwards it  to the closest peer (can assume good faith and that P is not connected to N' which IS the node closest to the chunk) 
          - if N is not closer to C than P (i.e., PO(N,C) <= PO(P,C)), then N just stores the chunk, does no forwarding or responding (assuming in good faith that P is not connected to N' which IS the node closest to the chunk) 
- if P is not in N's neighbourhood then N forwards to a node

pkg/pushsync/pushsync.go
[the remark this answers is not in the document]
esadakar: If the call to Debit is required regardless of what Put returns, then the current storer logic is also a bug. If this is the case, we can address this in a separate issue. What do you think @acud ?

pkg/pushsync/pushsync.go
```diff
@@ -130,6 +143,60 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+
+					var err error
+					defer func() {
+						if err != nil {
+							ps.logger.Tracef("pushsync replication: %v", err)
+						}
+					}()
+
+					receiptPrice := ps.pricer.PeerPrice(peer, chunk.Address())
+
```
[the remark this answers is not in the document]
metacertain: So, this changed recently with communication of prices in headers.
In the new system, when we send out pricing headers before the request, we guess that peers who appear to be in the same neighborhood as us will charge us 0, however, when the response headers are returned, this could prove false (if the peer advertises prices for a higher than neighborhood depth, or just simply communicates a price anyway)

When we do in-neighborhood actions for push sync, we probably want to skip the peers who communicate a non-zero price in their response headers, as nobody will cover for costs arising from this, but engage in the attempt to the extent of exchanging price related headers anyway towards the peers who appear to be within the neighborhood

PR #1386, opened 2021-03-05
pullsync: coalesce batch writes

acud:
This should also improve the CPU usage, since leveldb doesn't need to do individual writes. The downside is that because it might take longer to do the actual write, we might end up wanting the same chunk from multiple peers since this creates a possible data race (we want the chunk from one peer which is slow, and another peer which is fast, resulting in deliveries from both).

Open for discussion about this. Not entirely sure if it's needed if we get the other change which buffers the triggers to the subscriptions, which should solve the problem.

### DIFF
--- a/pkg/pullsync/pullsync.go
+++ b/pkg/pullsync/pullsync.go
@@ -202,33 +202,50 @@ func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8
 	// thus, the following loop will not get executed and the method
 	// returns immediately with the topmost value on the offer, which
 	// will seal the interval and request the next one
+	err = nil
+	var chunksToPut []swarm.Chunk
 
 	for ; ctr > 0; ctr-- {
 		var delivery pb.Delivery
 		if err = r.ReadMsgWithContext(ctx, &delivery); err != nil {
-			return 0, ru.Ruid, fmt.Errorf("read delivery: %w", err)
+			// this is not a fatal error and we should write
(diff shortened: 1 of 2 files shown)


ldeffenb (commented):
If there was an error reading the delivery causing the for loop to terminate, but it was after at least one chunk was added to chunksToPut, then the storage.Put is going to hide the read error by resetting err.

zelig (commented):
improve error handling

janos (approved):
LGTM. It might be more clear with some appropriately named functions, but it is well commented, so maybe some other time. 👍

pkg/pullsync/pullsync.go
```diff
@@ -202,33 +202,53 @@ func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8
...
 			return 0, ru.Ruid, swarm.ErrInvalidChunk
 		}
-
-		if err = s.storage.Put(ctx, storage.ModePutSync, chunk); err != nil {
-			return 0, ru.Ruid, fmt.Errorf("delivery put: %w", err)
+		chunksToPut = append(chunksToPut, chunk)
+	}
+	if len(chunksToPut) > 0 {
+		s.metrics.DbOpsCounter.Inc()
+		if ierr := s.storage.Put(ctx, storage.ModePutSync, chunksToPut...); ierr != nil {
```
zelig, line 239: this is a bit convoluted, but ok

zelig (approved):

PR #1390, opened 2021-03-05
libp2p: wait for readyness before accepting incoming connections

acud:
Since the libp2p instance is created before the entire node goes online, some nodes might already be able to dial into the node, causing the handshake to pass, but to subsequent protocol bootstrapping to fail. This change adds a `Ready` call to libp2p abstraction which will be called once the node constructor finishes. It will block incoming connections before the node bootstrapping is done, preventing inconsistent state on other nodes in the meanwhile.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -110,7 +110,7 @@ type Options struct {
 	SwapEnable             bool
 }
 
-func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey, signer crypto.Signer, networkID uint64, logger logging.Logger, libp2pPrivateKey, pssPrivateKey *ecdsa.PrivateKey, o Options) (*Bee, error) {
+func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey, signer crypto.Signer, networkID uint64, logger logging.Logger, libp2pPrivateKey, pssPrivateKey *ecdsa.PrivateKey, o Options) (b *Bee, err error) {
 	tracer, tracerCloser, err := tracing.NewTracer(&tracing.Options{
 		Enabled:     o.TracingEnabled,
 		Endpoint:    o.TracingEndpoint,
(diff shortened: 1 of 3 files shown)


pkg/node/node.go
```diff
@@ -244,6 +244,7 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey,
 		return nil, fmt.Errorf("p2p service: %w", err)
 	}
 	b.p2pService = p2ps
+	defer p2ps.Ready()
```
ralph-pichler, line 247: do we really need the `Ready` call if startup fails midway?

ralph-pichler (approved):

pkg/node/node.go
acud: I prefer since when we fail midway no shutdown will be called (this is wrong, as some things might already have happened by that point), and so some goroutines might already be waiting on that select case, and they should at least be released. Ideally as I previously mentioned we should also add a `Close` method on the libp2p library, and add a `quit` channel on the select case

ralph-pichler: should we not perhaps cancel the `p2pCtx` in an error case? that would also trigger the select. also this context is passed into `libp2p.New` (which then passes it on to the real `libp2p.New` from libp2p), so this might also use this already to wait on something.

janos: If both ready is closed and context is canceled it is not deterministic which select case would happen in the handshake handler. That is not so problematic but still not the best thing to have. But, @ralph-pichler suggestion to cancel the context in error case is very good. I would add an argument to Ready function, for example bool and that the channel is of bool values, so that we can check if it is false, to return in the handler, and if it is true to unblock the handler. It is not so clear naming in that case, maybe to rename the method to something else, but Ready is ok. But what way the described problem with select case would not exist, as both context will be canceled and Ready would pass the false value, both resulting in the return in the handler.

janos (approved):
Very nice. The discussion about the error state has good points, but even without that this functionality is good.

pkg/node/node.go
acud: OK point taken :+1:

acud:
@janos @ralph-pichler I added a defer statement to handle the error in case there was one and cancel the p2p context in that case. Thanks for the suggestions :+1:

janos (approved):

PR #1392, opened 2021-03-05
statestore: add migration capabilities

acud:
This PR adds migration capabilities to our leveldb statestore implementation. Implementation was adapted from the localstore migration code

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -70,7 +70,6 @@ require (
 	golang.org/x/tools v0.0.0-20200626171337-aa94e735be7f // indirect
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
 	gopkg.in/ini.v1 v1.57.0 // indirect
-	gopkg.in/src-d/go-log.v1 v1.0.1
 	gopkg.in/yaml.v2 v2.3.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 // indirect
 	honnef.co/go/tools v0.0.1-2020.1.4 // indirect
--- a/go.mod
+++ b/go.mod
@@ -70,6 +70,7 @@ require (
 	golang.org/x/tools v0.0.0-20200626171337-aa94e735be7f // indirect
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
 	gopkg.in/ini.v1 v1.57.0 // indirect
+	gopkg.in/src-d/go-log.v1 v1.0.1
 	gopkg.in/yaml.v2 v2.3.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 // indirect
 	honnef.co/go/tools v0.0.1-2020.1.4 // indirect
--- a/pkg/statestore/leveldb/leveldb.go
+++ b/pkg/statestore/leveldb/leveldb.go
@@ -17,16 +17,16 @@ import (
 	"github.com/syndtr/goleveldb/leveldb/util"
 )
 
-var _ storage.StateStorer = (*Store)(nil)
+var _ storage.StateStorer = (*store)(nil)
 
-// Store uses LevelDB to store values.
-type Store struct {
+// store uses LevelDB to store values.
+type store struct {
 	db     *leveldb.DB
 	logger logging.Logger
 }
 
 // New creates a new persistent state storage.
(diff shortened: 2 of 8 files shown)


go.mod
```diff
@@ -74,6 +74,7 @@ require (
 	google.golang.org/protobuf v1.25.0 // indirect
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
 	gopkg.in/ini.v1 v1.57.0 // indirect
+	gopkg.in/src-d/go-log.v1 v1.0.1
```
ralph-pichler, line 77: why does this appear here again?

ralph-pichler: please rebase and go mod tidy if necessary. this branch still contains the bad logger stuff from the deploy pr.

acud: I reverted this change but I actually don't see this as a problem. Dunno where its coming from but the build is fine

pkg/statestore/leveldb/migration.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			return err
+		}
+		s.logger.Infof("statestore: successfully ran migration: id %d current schema: %s", i, schemaName)
+	}
+	return nil
+}
+
+// getMigrations returns an ordered list of migrations that need be executed
+// with no errors in order to bring the localstore to the most up-to-date
```
ralph-pichler, line 79: comment says localstore

pkg/statestore/leveldb/migration_test.go
```diff
@@ -0,0 +1,297 @@
+// Copyright 2019 The Swarm Authors
```
ralph-pichler, line 1: the copyright headers here are all different than what we have in bee. also wrong license.

pkg/statestore/leveldb/leveldb.go
```diff
@@ -17,16 +17,16 @@ import (
 	"github.com/syndtr/goleveldb/leveldb/util"
 )
 
-var _ storage.StateStorer = (*store)(nil)
+var _ storage.StateStorer = (*Store)(nil)
 
 // Store uses LevelDB to store values.
-type store struct {
+type Store struct {
```
ralph-pichler, line 23: why the export?

pkg/statestore/leveldb/migration_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+	dir, err := ioutil.TempDir("", "statestore-test")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dir)
+
+	logger := logging.New(ioutil.Discard, 0)
+
+	// start the fresh localstore with the sanctuary schema name
```
ralph-pichler, line 59: comment says localstore (same in several other places)

pkg/statestore/leveldb/leveldb.go
acud: because `leveldb_test`

janos: Maybe to export it in export_test.go? Or just to export there GetSchemaName and keep it unexported in the package.

pkg/statestore/leveldb/leveldb_test.go
```diff
@@ -48,3 +48,33 @@ func TestPersistentStateStore(t *testing.T) {
 		return store
 	})
 }
+
+func TestGetSchemaName(t *testing.T) {
+	dir, err := ioutil.TempDir("", "statestore_test")
```
janos, line 53: I would suggest to use t.TempDir for convenience. Not just here but in all new code in this PR (6 instances).

janos (commented):
I think that it is better to try to keep the `store` type unexported.

pkg/statestore/leveldb/leveldb.go
acud: I tried it now again (second attempt). The problem is that the unit test needs to verify the schema name is the correct one, but the statestore constructor returns an interface. The unit test is not a black-box test as a result. This is why it was not a black-box test in localstore either (probably the only places in the codebase where this is necessary). The other idea I have is to unexport this, and just have a helper in the test that loads a new leveldb, and just tries to load the specified key out of the instance, and verify that it matches the latest schema version

esadakar (approved):

pkg/statestore/leveldb/leveldb.go
acud: I didn't think of that :) Added a new commit, let me know if this is what you've meant

PR #1394, opened 2021-03-05
Soc refactor

r0qs:
This PR attempt to refactor the soc API to a more idiomatic and clean API.

The new API exposes the following:
```go
func New(id ID, ch swarm.Chunk) *Soc
func NewSigned(id ID, ch swarm.Chunk, owner, sig []byte) (*Soc, error)

func (s *Soc) Sign(signer crypto.Signer) (swarm.Chunk, error)
func (s *Soc) Chunk() (swarm.Chunk, error)
func (s *Soc) WrappedChunk() swarm.Chunk

func FromChunk(sch swarm.Chunk) (*Soc, error)
func CreateAddress(id ID, owner Owner) (swarm.Address, error) 
func Valid(ch swarm.Chunk) bool
```

### DIFF
--- a/pkg/soc/soc.go
+++ b/pkg/soc/soc.go
@@ -9,7 +9,6 @@ package soc
 import (
 	"bytes"
 	"errors"
-	"fmt"
 
 	"github.com/ethersphere/bee/pkg/cac"
 	"github.com/ethersphere/bee/pkg/crypto"
@@ -23,28 +22,17 @@ const (
 )
 
 var (
-	errNoOwner        = errors.New("soc: owner not set")
+	errInvalidAddress = errors.New("soc: invalid address")
 	errWrongChunkSize = errors.New("soc: chunk length is less than minimum")
 )
 
 // ID is a soc identifier
 type ID []byte
 
-// Owner is the address in bytes of soc owner.
-type Owner []byte
-
-// NewOwner ensures a valid length address of soc owner.
-func NewOwner(address []byte) (Owner, error) {
-	if len(address) != crypto.AddressSize {
-		return nil, fmt.Errorf("soc: invalid address %x", address)
-	}
-	return address, nil
-}
-
 // Soc wraps a single-owner chunk.
 type Soc struct {
 	id        ID
-	owner     Owner
+	owner     []byte // Owner is the address in bytes of soc owner.
 	signature []byte
 	chunk     swarm.Chunk // wrapped chunk.
 }
@@ -61,11 +49,10 @@ func New(id ID, ch swarm.Chunk) *Soc {
 // NewSignedSoc creates a single-owner chunk based on already signed data.
 func NewSigned(id ID, ch swarm.Chunk, owner, sig []byte) (*Soc, error) {
 	s := New(id, ch)
-	o, err := NewOwner(owner)
-	if err != nil {
-		return nil, err
+	if len(owner) != crypto.AddressSize {
+		return nil, errInvalidAddress
 	}
-	s.owner = o
+	s.owner = owner
 	s.signature = sig
 	return s, nil
 }
@@ -73,7 +60,7 @@ func NewSigned(id ID, ch swarm.Chunk, owner, sig []byte) (*Soc, error) {
 // address returns the soc chunk address.
 func (s *Soc) address() (swarm.Address, error) {
 	if len(s.owner) != crypto.AddressSize {
-		return swarm.ZeroAddress, errNoOwner
+		return swarm.ZeroAddress, errInvalidAddress
 	}
 	return CreateAddress(s.id, s.owner)
 }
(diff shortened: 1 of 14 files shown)


pkg/soc/soc.go
```diff
@@ -22,130 +22,131 @@ const (
...
 }
 
-// NewChunk is a convenience function to create a single-owner chunk ready to be sent
-// on the network.
-func NewSignedChunk(id Id, ch swarm.Chunk, owner, sig []byte) (swarm.Chunk, error) {
+// NewSignedSoc creates a single-owner chunk based on already signed data.
+func NewSigned(id ID, ch swarm.Chunk, owner, sig []byte) (*Soc, error) {
 	s := New(id, ch)
-	s.signature = sig
 	o, err := NewOwner(owner)
```
acud, line 64: I think that `Owner` type is redundant and unnecessary. We could just settle for a simple byte slice and just verify the length in the constructor with
```go
	if len(address) != crypto.AddressSize {
		return nil, fmt.Errorf("soc: invalid owner address %x", address)
	}
```

WDYT?

pkg/soc/soc.go
```diff
@@ -22,130 +22,131 @@ const (
...
-		return err
+		return nil, err
 	}
 	ownerAddress, err := NewOwner(ownerAddressBytes)
 	if err != nil {
-		return err
+		return nil, err
 	}
-	s.signer = signer
 	s.owner = ownerAddress
```
acud, line 120: I don't think that we need to mutate this field. The same goes for the signature field. If there is no subsequent usage of the `Soc` instance after signing, then this is probably not necessary.

pkg/soc/testing/soc.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// Chunk returns the soc chunk of the mocked soc.
+func (ms MockSoc) Chunk() swarm.Chunk {
+	return swarm.NewChunk(ms.Address(), append(ms.ID, append(ms.Signature, ms.WrappedChunk.Data()...)...))
+}
+
+// GenerateMockSoc generates a valid mocked soc from given data.
+// If data is nil it generates random data.
+func GenerateMockSoc(data []byte) *MockSoc {
```
acud, line 42: Ideally, this should also take a `t *testing.T` as a parameter, then first line of the function call `t.Helper()`.
I would also actually handle the errors, and once the `t.Helper` is called, you can just `t.Fatal(err)`  directly in the error checks

acud (commented):
This is really great :clap: I have just a few minor suggestions

pkg/soc/soc.go
r0qs: I agree. I'll do the change and remove the Owner type.

pkg/soc/soc.go
r0qs: The signature is used in the `Chunk` method to create the soc chunk in the return of the `Sign` function, but I can do it in the `Sign` method instead. 

The `owner` is used to generate the soc's address, but it is currently used only in the tests and in the `Valid` method. Removing the owner would require us to extract the signature from the chunk and make the recovery of the owner address in the `Valid` function and use it there to recreate the chunk address for comparison.

r0qs: For example, I could create a signed soc, and then call `Sign` again with a different signer, and if the sign does not mutate the owner, the newly signed soc will have another owner from the signer and result in a different soc address.

pkg/soc/soc.go
```diff
@@ -22,130 +21,118 @@ const (
...
-	}
-	return &Owner{
-		address: address,
-	}, nil
-}
+// ID is a soc identifier
+type ID []byte
 
-// Soc wraps a single soc.
+// Soc wraps a single-owner chunk.
```
acud, line 32: actually, `Soc` wraps a content-addressed chunk :)

acud (approved):
Great stuff @r0qs :clap: 
Much needed refactoring. Very nicely done!

pkg/soc/soc.go
r0qs: haha of course, sorry. Fixed.

acud:
2021-03-16-100506_955x473_scrot

Something came up on DeepSource

r0qs:
> 2021-03-16-100506_955x473_scrot
> 
> Something came up on DeepSource

I added this for convenience in the tests to generate random data for a chunk. I'm not doing any security-related operation with it. So we can ignore this warning.

But looking at my tests again now, I noticed that although I added the possibility of randomly generate the chunk data, I'm not using it anymore. So I'm removing this code since the caller of `GenerateMockSoc` can generate random data if needed.

janos (approved):
Nice.

I have one small suggestion, and that is to rename `Soc` to `SOC` everywhere as it is an abbreviation. It is also more clear regarding the pronunciation and the usage of indefinite article.

r0qs:
> as it is an abbreviation

While renaming, I found this [comment](https://github.com/ethersphere/bee/blob/master/pkg/api/file.go#L304), and I'm wondering what concept this comment is about?
```
// If none match header is set always send the reply as not modified
// TODO: when SOC comes, we need to revisit this concept
```

janos:
> what concept this comment is about?

@r0qs I believe that http.StatusNotModified should not be returned in case of SOC chunk, as browser caching may not be desired. But, the concept should be revisited. How to properly do the caching if possible.

acud:
I think that the comment might relate to feeds, not SOC chunks, but I might be wrong. I don't see why an SOC should change this logic.

PR #1399, opened 2021-03-08
kademlia, cmd, node: add bootnode-mode flag

acud:
As a result of #1352, nodes will limit the amount of incoming connections from other peers.
This is an undesirable for our bootnodes, which should be accepting more incoming connections so that new peers can get informed about the rest of the network.

In the future we should improve this by accepting new nodes in bootnode mode and either give them a list of peers to connect to, or just kick away old nodes that have been connected for a while (oldest, perhaps), so that we rotate the network that the bootnode sees, without prioritizing or preferring certain nodes over others.

### DIFF
--- a/cmd/bee/cmd/cmd.go
+++ b/cmd/bee/cmd/cmd.go
@@ -42,6 +42,7 @@ const (
 	optionNamePaymentTolerance          = "payment-tolerance"
 	optionNamePaymentEarly              = "payment-early"
 	optionNameResolverEndpoints         = "resolver-options"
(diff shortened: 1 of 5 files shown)


acud:
cc @vandot - once this is rolled out we must add the `--bootnode-mode` flag to our bootnodes

ralph-pichler (approved):
LGTM

metacertain (approved):

janos (approved):
LGTM

Issue #1332, opened 2021-02-24
Uploading empty file as part of a collection fails


agazso:
#### Summary
Uploading to the `/dirs` endpoint a tar collection which contains an empty file fails.

#### Steps to reproduce
Running Bee 0.5.1

```sh
$ touch x
$ tar cf ../test.tar ./x 
$ curl \                    
    -X POST \
    -H "Content-Type: application/x-tar" \
    --data-binary @../test.tar http://localhost:1633/dirs
{"message":"could not store dir","code":500}
```
#### Expected behavior
Empty files should work without problem. This used to work in 0.5.0 so this is a regression in 0.5.1.

Eknir:
Why do you want to upload an empty file?

acud:
This is caused because of my changes to the splitter, which no longer accepts empty data. I'l rectify this somehow tomorrow. 

@agazso what is the desired behavior on retrieval? HTTP 204?

agazso:
> @agazso what is the desired behavior on retrieval? HTTP 204?

I would say 200 with `content-length: 0`

PR #1406, opened 2021-03-09
pipeline: add special case for empty file

acud:
fixes a regression caused by #1267, which in turn caused #1332

### DIFF
--- a/pkg/file/pipeline/builder/builder_test.go
+++ b/pkg/file/pipeline/builder/builder_test.go
@@ -56,6 +56,27 @@ func TestHelloWorld(t *testing.T) {
 	}
 }
 
+// TestEmpty tests that a hash is generated for an empty file.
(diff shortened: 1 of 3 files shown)


acud:
cc @agazso

pkg/p2p/libp2p/static_resolver_test.go
```diff
@@ -16,6 +15,10 @@ import (
 )
 
 func TestStaticAddressResolver(t *testing.T) {
+	if runtime.GOOS == "windows" || runtime.GOOS == "plan9" {
```
zelig, line 18: this is in another pr

zelig (approved):
maybe  the empty file chunk should just be a swarm package constant and not burdden the logic here, but its ok

janos (approved):

PR #1416, opened 2021-03-10
pushsync: improve context handling

acud:
There's a lot of improper context handling on pushsync. Expiration on forwarding/reception is non existent, and on `pushToClosest` the receipt context might already be expired once it is created from the previous context, which shadows the original one. Also, defer calls are done incorrectly.

I am also raising TTL since it seems just too low value right now in the current state of the network.

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -224,9 +224,9 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
 		deferFuncs = append(deferFuncs, func() { go streamer.FullClose() })
 
 		w, r := protobuf.NewWriterAndReader(streamer)
-		ctx, cancel := context.WithTimeout(ctx, timeToWaitForReceipt)
-		defer cancel()
-		if err := w.WriteMsgWithContext(ctx, &pb.Delivery{
+		ctxd, cancel := context.WithTimeout(ctx, timeToWaitForReceipt)
+		deferFuncs = append(deferFuncs, func() { cancel() })
+		if err := w.WriteMsgWithContext(ctxd, &pb.Delivery{
 			Address: ch.Address().Bytes(),
 			Data:    ch.Data(),
 		}); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/pushsync/pushsync.go
```diff
@@ -189,13 +187,6 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
 
 		defersFn()
 
-		deferFuncs = append(deferFuncs, func() {
```
acud, line 192: this is needed since we might be logging the error unnecessarily when the error is returned from `ClosestPeer` below

pkg/pushsync/pushsync.go
```diff
@@ -224,9 +222,9 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
 		deferFuncs = append(deferFuncs, func() { go streamer.FullClose() })
 
 		w, r := protobuf.NewWriterAndReader(streamer)
-		ctx, cancel := context.WithTimeout(ctx, timeToWaitForReceipt)
-		defer cancel()
-		if err := w.WriteMsgWithContext(ctx, &pb.Delivery{
+		ctxd, canceld := context.WithTimeout(ctx, timeToLive)
```
janos, line 225: This context `ctxd` has a timeout of `timeToLive` which is the same as the inherited context created in the handler function. Basically, this timeout it is not needed if they are of the same value. The same goes for `ctxr`. Either to raise the `ctx` timeout or to just pass `ctx` directly to `WriteMsgWithContext` and `WriteMsgWithContext` functions without creating additional timeouts here.

acud: actually, one of them is needed but the other isn't. The problem is that this same method is shared both by `pusher` and `handler`. If the deadline is already set by handler then the context will expire before the new deadline is reached. I think we can live with this, at least until we refactor the package to wait for responses from already open streams instead of just closing and trying a different one

janos (approved):
Thanks!

esadakar (approved):

Eknir:
@acud, could you please connect this one to an issue? I recall that there were many bug reports about this It would be good to connect this, so we can ask the authors of the report to verify if the bug is fixed.

PR #1421, opened 2021-03-11
fix unconfigured cors origin

janos:
This PR solves a serious bug with serving CORS headers if no allowed origins are configured and an origin header is present in an HTTP request. If CORS headers are configured, the functionality is correct.

This PR adds two test cases in already existing debugapi test and adds the same test for the api package.

### DIFF
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -29,17 +29,18 @@ import (
 )
 
 type testServerOptions struct {
-	Storer          storage.Storer
-	Resolver        resolver.Interface
-	Pss             pss.Interface
-	Traversal       traversal.Service
-	WsPath          string
-	Tags            *tags.Tags
-	GatewayMode     bool
-	WsPingPeriod    time.Duration
-	Logger          logging.Logger
-	PreventRedirect bool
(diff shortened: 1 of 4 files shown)


agazso:
It seems that this fix worked for disabling CORS for the browser. However it only stops the browser from receiving the response, but not from making it completely. For example as a test if I open this link in Chrome, it sets the welcome message of my Bee node to empty string, never mind how `cors-allowed-origins` is set 😬 

https://test.felfele.org/cors.html?bee=http://localhost:1635/welcome-message&method=POST&body={%22welcome-message%22:%22cors%22}

janos:
@agazso Indeed, the browser does log the request as blocked because of CORS `Access to fetch at 'http://localhost:1635/welcome-message' from origin 'https://test.felfele.org' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.` but the request is actually made, for both POST and GET.

acud (approved):

esadakar (approved):

Issue #1340, opened 2021-02-25
index out of range panic


shpyntov:
#### Summary
I had a crash of node, logs attached

#### Steps to reproduce
https://pastebin.com/raw/rVnHAVt1
https://pastebee.com/?05a3568672967385ab32a1ea0154a53596cc50ad188cc2116d5d1754f1b022e0

#### Expected behavior
node isn't crashing

#### Actual behavior
node is crashing

Eknir:
Thanks for reporting this! We will have a look at this. Did you apply any special (non-default) configuration of your node? How did you install and run Bee? What version of Bee were you running?

acud:
```
45c179bbcd6ae0ae1df1d7d29be4004763af6a10f7b8: context deadline exceeded 
panic: runtime error: slice bounds out of range [:16] with capacity 0

goroutine 5550134 [running]:
github.com/ethersphere/bee/pkg/localstore.New.func4(0xc009e12639, 0x20, 0x27, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	github.com/ethersphere/bee/pkg/localstore/localstore.go:233 +0xd1
github.com/ethersphere/bee/pkg/shed.Index.Get(0xc00040f5f0, 0xc000295c30, 0x1, 0x1, 0xc000c4fd40, 0xc000c77770, 0x16fbf78, 0x16fbf80, 0xc009e12639, 0x20, ...)
	github.com/ethersphere/bee/pkg/shed/index.go:154 +0x26d
github.com/ethersphere/bee/pkg/localstore.(*DB).setSync(0xc00010d800, 0xc00a57f840, 0xc009e12639, 0x20, 0x27, 0x0, 0x20, 0x40, 0xc800000000000000)
	github.com/ethersphere/bee/pkg/localstore/mode_set.go:134 +0x178
github.com/ethersphere/bee/pkg/localstore.(*DB).set(0xc00010d800, 0x0, 0xc0088faec0, 0x1, 0x1, 0x0, 0x0)
	github.com/ethersphere/bee/pkg/localstore/mode_set.go:64 +0x1af
github.com/ethersphere/bee/pkg/localstore.(*DB).Set(0xc00010d800, 0x185bd60, 0xc009942d80, 0x0, 0xc0088faec0, 0x1, 0x1, 0x0, 0x0)
	github.com/ethersphere/bee/pkg/localstore/mode_set.go:38 +0xf3
[...]

PR #1321, opened 2021-02-24
stamp support in storage and protocols

zelig:
#898 
AP https://hackmd.io/2NgPR2CeQICKzWBAH7Mk0Q#

PR #1426, opened 2021-03-13
shed: return swallowed error

acud:
Rectifies a case where an error might be returned from leveldb but gets swallowed in the `if` statement, resulting in a `nil` `value` and eventually causing a panic in the calling context.

Should improve #1340


janos (approved):
Thanks!

zelig (approved):
I had also fixed this  one https://github.com/ethersphere/bee/pull/1321/files#diff-0b17498d28521cf5b22652e0df1a359d38ed2e219f1b60f1facd406128ead388  :)

PR #1431, opened 2021-03-15
receipt is signed by storer node and verified by origin node

esadakar:


pkg/pushsync/pb/pushsync.proto
```diff
@@ -15,4 +15,6 @@ message Delivery {
 
 message Receipt {
   bytes Address = 1;
+  bytes StorerAddress = 2;
```
acud, line 18: I think that the addition of the storer address in the message is not necessary, since the address can be recovered from the signature

acud (commented):
Looks good, but I think that the addition of the storer address in the protocol is unnecessary

pkg/pushsync/pb/pushsync.proto
esadakar: This check was meant to prevent nodes from signing on behalf of other nodes, eg. public key in the signature being different from the address of the node.

After this, we still have the problem of uncertainty about the signer being the actual storer.

acud: This does not provide much more protection. If you spoof a new bogus key to sign the message in another node's name, what's preventing you from setting the address field on the message too?

esadakar: Hmm. The node may indeed create a bogus key and a node address, and sign the receipt with the address and public key being the same. It does not provide much protection, like you mentioned. 👍

pkg/pusher/pusher.go
```diff
@@ -163,6 +167,13 @@ LOOP:
 						return
 					}
 				}
+
+				_, err = crypto.Recover(receipt.Signature, receipt.Address.Bytes())
+				if err != nil {
+					err = fmt.Errorf("pusher receipt recover: %w", err)
```
janos, line 173: The message should have a colon after the `pusher` - `"pusher: receipt recover: %w"`

pkg/pushsync/pushsync_test.go
```diff
@@ -34,25 +36,27 @@ const (
 )
 
 // TestSendChunkAndGetReceipt inserts a chunk as uploaded chunk in db. This triggers sending a chunk to the closest node
-// and expects a receipt. The message are intercepted in the outgoing stream to check for correctness.
+// and expects a receipt. The message are intercepted in the outgoing stream to check for correctness.cry
```
janos, line 39: The test should not make anyone cry :) at least in an ideal world.

pkg/crypto/mock/signer.go
```diff
@@ -18,6 +18,7 @@ type signerMock struct {
 	signTx          func(transaction *types.Transaction, chainID *big.Int) (*types.Transaction, error)
 	signTypedData   func(*eip712.TypedData) ([]byte, error)
 	ethereumAddress func() (common.Address, error)
+	signature       []byte
```
janos, line 21: It would be convenient to have `signature` as `signFunc func([]byte) ([]byte, error)` in order to be able to validate both the input argument and error handling in tests. This change would reflect on `WithSignature` function. Also it would be inline with already existing functions in this mock.

pkg/pushsync/pushsync_test.go
```diff
@@ -385,7 +394,46 @@ func TestHandler(t *testing.T) {
...
+	receipt, err := psPivot.PushChunkToClosest(context.Background(), chunk)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !bytes.Equal(chunk.Address().Bytes(), receipt.Address.Bytes()) {
+		t.Fatal("chunk address do not match")
+	}
+
+	fmt.Println(receipt.Signature)
```
janos, line 428: Use t.Log to log data in tests, or just to remove this line.

janos (commented):
Nice!

pkg/pushsync/pushsync_test.go
esadakar: I have no idea how that ended up there :)

pkg/pusher/pusher.go
```diff
@@ -163,6 +167,13 @@ LOOP:
 						return
 					}
 				}
+
+				_, err = crypto.Recover(receipt.Signature, receipt.Address.Bytes())
```
zelig, line 171: at least comment here that this is in no way aaa validation on any sort. ECrecover returns without an  error to whatever text you give, so the public key needs  checking I guess  this  can come in the subsequent PR but just comment temporary  or something to  indicate this

zelig (approved):

pkg/pusher/pusher.go
esadakar: Once we have the ability to check that the storer node is in the right neighborhood (ie depth oracle) with the public key, we will add it here. Thanks for the heads up.

janos (approved):
👍

pkg/pusher/pusher.go
acud: I think that would be a nice to have here, if we could actually use the `Recover` call to extract the node overlay address. There are helper functions in the crypto package to create a `NewOverlayAddress`. Then if we could somehow also log this as part of the information logged in `L146`, like: `pusher: pushed chunk %s to node %s`, logging the chunk address and the overlay address that signed the reciept. I think this could also help us increase our observability while debugging problems on clusters.

acud (approved):
Nice!

acud (approved):
Great! thanks!

acud:
@esadakar now that @metacertain's PR is merged with variable pricing - can you rebase and merge this one too?

acud (approved):
Great :) thank you!

PR #1446, opened 2021-03-18
add metrics to track number of cheques

ralph-pichler:
adds more metrics so we can easily tell how much cheque traffic is actually happening. hopefully this can give some insight on why there are so many `eth_Call` operations.

### DIFF
--- a/pkg/settlement/swap/metrics.go
+++ b/pkg/settlement/swap/metrics.go
@@ -38,13 +38,13 @@ func newMetrics() metrics {
 			Namespace: m.Namespace,
 			Subsystem: subsystem,
 			Name:      "cheques_received",
-			Help:      "Amount of cheques received from peers",
+			Help:      "Number of cheques received from peers",
 		}),
 		ChequesSent: prometheus.NewCounter(prometheus.CounterOpts{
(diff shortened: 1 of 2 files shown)


pkg/settlement/swap/metrics.go
```diff
@@ -32,6 +34,18 @@ func newMetrics() metrics {
...
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "cheques_received",
+			Help:      "Amount of cheques received from peers",
+		}),
+		ChequesSent: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "cheques_sent",
+			Help:      "Amount of cheques sent to peer",
```
acud, line 47: to peers

esadakar (approved):

pkg/settlement/swap/metrics.go
ralph-pichler: fixed. I also changed amount of to number of, as it was also that way in the existing metrics that track number of cheques.

PR #1449, opened 2021-03-19
kademlia: wait on peer before retrying after connect err

acud:
Adds a wait to wait before retrying a peer that we failed connecting to. Observed on staging.

### DIFF
--- a/pkg/kademlia/kademlia.go
+++ b/pkg/kademlia/kademlia.go
@@ -309,6 +309,11 @@ func (k *Kad) manage() {
 								}
 								k.logger.Debugf("peer not reachable from kademlia %s: %v", bzzAddr.String(), err)
 								k.logger.Warningf("peer not reachable when attempting to connect")
+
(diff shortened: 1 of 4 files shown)


ralph-pichler (approved):

esadakar (approved):

pkg/kademlia/kademlia.go
```diff
@@ -309,6 +309,11 @@ func (k *Kad) manage() {
 								}
 								k.logger.Debugf("peer not reachable from kademlia %s: %v", bzzAddr.String(), err)
 								k.logger.Warningf("peer not reachable when attempting to connect")
+
+								k.waitNextMu.Lock()
+								k.waitNext[peer.String()] = retryInfo{tryAfter: time.Now().Add(shortRetry)}
```
esadakar, line 314: maybe we can add this at the end of the loop because we also have a wait in the case that there is no error. this way we wont repeat the same wait logic

PR #1453, opened 2021-03-20
kademlia: balanced iterator erroneously exits but should continue

acud:
The balanced kademlia erroneously exits upon an error connecting to a peer. The comment maintains that we should `// continue to next` but it instead just returns, invoking the normal kademlia connection iterator below.


acud:
test that failed is fixed on a different PR

janos (approved):
Well spotted.

zelig (approved):

PR #1454, opened 2021-03-21
shed, localstore, cmd: expose leveldb configuration parameters to cli

acud:
We are running a standard open files limit which is very low compared with the recommended setting by google's leveldb [recommended](https://github.com/google/leveldb/blob/master/include/leveldb/options.h#L87) specifications of number of open files per size of working set. Realistic values should be in the thousands range, rather than hundreds (ideally - around 10K). This should be documented and communicated as part of the release. I am exposing these parameters so that users can experiment with these settings without having to rebuild the binary every time.

CCing @eknir @significance for documentation and communicating this with the release


esadakar (approved):

pkg/shed/db.go
```diff
@@ -50,16 +56,22 @@ type DB struct {
...
 		ldb, err = leveldb.Open(storage.NewMemStorage(), nil)
 	} else {
 		ldb, err = leveldb.OpenFile(path, &opt.Options{
-			OpenFilesCacheCapacity: openFileLimit,
-			BlockCacheCapacity:     blockCacheCapacity,
-			DisableSeeksCompaction: true,
-			WriteBuffer:            writeBuffer,
+			OpenFilesCacheCapacity: int(o.OpenFilesLimit),
+			BlockCacheCapacity:     int(o.BlockCacheCapacity),
+			WriteBuffer:            int(o.WriteBufferSize),
```
janos, line 74: This PR removes DisableSeeksCompaction option. If there are some noticeable additional i/o, one suggestion is to try to put it back.

janos (approved):

pkg/shed/db.go
acud: Indeed. This is intentional. Perhaps maybe best it to actually add another CLI flag for this... I'll do this right away

Issue #1450, opened 2021-03-20
"Out of Gas" transaction handling?


ldeffenb:
#### Summary
With the recent gas price fluctuations on the goerli test network, many of us are accumulating failed cashout transactions due to "out of gas".  Are these transactions reverted internal to bee or do we need to somehow reprocess them as was mentioned in bee-support?

#### Steps to reproduce
Do a bunch of cashouts and you'll likely hit an out-of-gas as the goerli network jumps the price to 20gwei instead of 1.

#### Expected behavior
I would like to believe that the bee node will detect the failed transaction and make the peer cashable again, but having examined the source code, I don't think that is currently the case.

#### Actual behavior
I haven't done a definitive test yet, but I suspect that the peer is no longer cashable until at least another threshold has been reached.  And even then, I'm not certain that the new check will include the settlement from the failed cashout transaction.

The following script was suggested in bee-support as being necessary to get the out-of-gas transactions to be reprocessed.  I can confirm that it lists my failed transactions, but I'm not sure if I can meet the requirement for a "geth compatible" keystore and password for my node's Ethereum account.

https://gist.github.com/syspulse/dd652cce3bf2e1ddc13272ef4e472726

ldeffenb:
Any comment from the developers on what can be done to recover these transactions?

kevin486:
I have many cash transactions went into out-of-gas error,  and desperately need help on this point.

Eknir:
If you can send a cheque with a same peer when you have received `new` cheques, then you will also cash out the old cheques. The reason for this can be found in how cheques are internally implemented (see [sw3](https://github.com/ethersphere/swap-swear-and-swindle/blob/master/contracts/ERC20SimpleSwap.sol#L224)). 

As you can see, what matters is the cumulative amount on the cheque. If you did not cash out a previous cheque, then the chequebook of your peer did not update the `paidout[beneficiary]` mapping, so you will just receive the payout as if the previous transaction did not happen.

While I understand this is a nuisance, we will not work on fixing this now. Especially since there is a way around this.

Thanks for reporting though and please let me know if you need any further explanations (can be here in a comment or PM).

PR #1455, opened 2021-03-21
swap,sctx,debugapi: add optional gas limit and gas price to cashout

acud:
Users are complaining about `out of gas` errors due to some problem in the gas estimator. This PR adds two optional headers that inject the values that users may add to the request into the context. The cashout is reading the context value, which falls back to the default values that kick in the estimator.

Should help with #1450

### DIFF
--- a/pkg/debugapi/balances.go
+++ b/pkg/debugapi/balances.go
@@ -16,10 +16,10 @@ import (
 )
 
 var (
-	errCantBalances  = "Cannot get balances"
-	errCantBalance   = "Cannot get balance"
-	errNoBalance     = "No balance for peer"
-	errInvaliAddress = "Invalid address"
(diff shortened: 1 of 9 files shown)


ralph-pichler (approved):
Thanks. Not sure if passing these values through context is really the best way though, some calls (e.g. create in postage stamp) might have two different txs sent with same context.

Alternatively we could just hardcode the gaslimits for the calls where estimations fail. There is an predictable upper bound for how much a cashout can cost.

ralph-pichler (approved):

janos (approved):
As @ralph-pichler mentioned, using context for this information may not be the best way, but certainly requires the least changes. It is fine for me, with a note to revaluate the approach when there is time for that.

esadakar (approved):

PR #1456, opened 2021-03-21
flipflop: fix goroutine leak on subscription termination

acud:
Embarrassing goroutine leak spotted by @ldeffenb.
On termination of a subscribe pull subscription, the detector might have a timer that is already pending, with the select case already selecting on the timer expiration, this results in an attempt to write to an unbuffered channel that is never read from, since the listening goroutine does not try to drain the channel. Also, draining the channel would be a bit ugly, since we'd have to potentially write one value to the `in` channel then wait for the incoming message after the buffer timeout expires. So I have instead added a buffer of one and a non-blocking trigger of the output.

### DIFF
--- a/pkg/flipflop/falling_edge.go
+++ b/pkg/flipflop/falling_edge.go
@@ -25,7 +25,7 @@ func NewFallingEdge(bufferTime, worstCase time.Duration) (in chan<- struct{}, ou
 		t:         bufferTime,
 		worstCase: worstCase,
 		buf:       make(chan struct{}, 1),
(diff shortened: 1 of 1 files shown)


janos (approved):

ldeffenb:
I put this fix in the node that pointed out the issue and can attest that it does fix it.

esadakar (approved):

acud:
I have simplified the fix, adding the buffer to the output channel is not needed. Simply selecting on the quit channel in the timer cases will suffice.

PR #1461, opened 2021-03-22
TestRetrievePreemptiveRetry fixed

esadakar:


acud (approved):
Great! :) thanks @esadakar

pkg/retrieval/retrieval_test.go
```diff
@@ -297,19 +299,19 @@ func TestRetrievePreemptiveRetry(t *testing.T) {
 			streamtest.WithMiddlewares(
 				func(h p2p.HandlerFunc) p2p.HandlerFunc {
 					return func(ctx context.Context, peer p2p.Peer, stream p2p.Stream) error {
+						failM.Lock()
+						defer failM.Unlock()
```
aloknerurkar, line 303: Would it be better to explicitly unlock here before the server handler is called?

pkg/retrieval/retrieval_test.go
```diff
@@ -333,15 +336,12 @@ func TestRetrievePreemptiveRetry(t *testing.T) {
 			streamtest.WithMiddlewares(
 				func(h p2p.HandlerFunc) p2p.HandlerFunc {
 					return func(ctx context.Context, peer p2p.Peer, stream p2p.Stream) error {
-						if serverAddress1.Equal(peer.Address) {
+						if failOnce {
```
aloknerurkar, line 339: Any reason why this isn't lock protected? Not sure if its required

aloknerurkar (approved):

PR #1465, opened 2021-03-22
transaction: add nonce-based monitor

ralph-pichler:
This is a pretty radical rewrite of how transactions are monitored for confirmations.

The primary aim is to be able to:
* have a somewhat constant number of backend requests over time independent of the number of pending transactions
* give up on transactions which are provably cancelled (while avoiding giving up on transactions which might still happen, even if they take a very long time to confirm)
* provide the base for watching transactions in the background
* provide the base for advanced transaction functionality such as gas boosting, replacing or rebroadcasting transactions

The way this works is as follows:

Instead of watching transactions individually, the senders nonce is monitored and transactions are checked based on this. The idea is that if the nonce is still lower than that of a pending transaction, there is no point in actually checking the transaction for a receipt. At the same time if the nonce was already used and this was a few blocks ago we can reasonably assume that it will never confirm.

A new component `transaction.Monitor` was introduced for that purpose (primarily because this could be instantiated even without a signer, and also to avoid the locks interfering with the locks of the transaction sending). However other components still wait using the `transaction.Service` as using the monitor directly requires knowing the nonce which only the transaction service does.

The `transaction.Service` is now keeping a record of all transactions it sent. In this PR this is only used to look up the nonce later, but in the future this will allow adding users to query past transaction info, query pending transactions and to enable the advanced functionality mentioned above. Because it now uses the monitor underneath the `transaction.Service` can now only wait for its own transactions, not arbitrary ones (but it was never used in that way anyway). 

In order to test the monitor (which queries many blocks and nonces at different blocks in sequence) a new `transaction.Backend` mock was added which can be used in a more declarative way.

### DIFF
--- a/pkg/settlement/swap/transaction/monitor.go
+++ b/pkg/settlement/swap/transaction/monitor.go
@@ -28,7 +28,7 @@ var ErrMonitorClosed = errors.New("monitor closed")
 type Monitor interface {
 	io.Closer
 	// WatchTransaction watches the transaction until either there is 1 confirmation or a competing transaction with cancellationDepth confirmations.
-	WatchTransaction(txHash common.Hash, nonce uint64) (chan types.Receipt, chan error, error)
+	WatchTransaction(txHash common.Hash, nonce uint64) (<-chan types.Receipt, <-chan error, error)
 }
 type transactionMonitor struct {
 	lock       sync.Mutex
@@ -78,7 +78,7 @@ func NewMonitor(logger logging.Logger, backend Backend, sender common.Address, p
 	return t
 }
 
-func (tm *transactionMonitor) WatchTransaction(txHash common.Hash, nonce uint64) (chan types.Receipt, chan error, error) {
+func (tm *transactionMonitor) WatchTransaction(txHash common.Hash, nonce uint64) (<-chan types.Receipt, <-chan error, error) {
 	tm.lock.Lock()
 	defer tm.lock.Unlock()
 
--- a/pkg/settlement/swap/transaction/monitor.go
+++ b/pkg/settlement/swap/transaction/monitor.go
@@ -21,6 +21,9 @@ import (
 var ErrTransactionCancelled = errors.New("transaction cancelled")
 
 // Monitor is a nonce-based watcher for transaction confirmations.
+// Instead of watching transactions individually, the senders nonce is monitored and transactions are checked based on this.
+// The idea is that if the nonce is still lower than that of a pending transaction, there is no point in actually checking the transaction for a receipt.
+// At the same time if the nonce was already used and this was a few blocks ago we can reasonably assume that it will never confirm.
 
 type Monitor interface {
 	io.Closer
--- a/pkg/settlement/swap/transaction/monitor.go
+++ b/pkg/settlement/swap/transaction/monitor.go
@@ -25,7 +25,6 @@ var ErrMonitorClosed = errors.New("monitor closed")
 // Instead of watching transactions individually, the senders nonce is monitored and transactions are checked based on this.
 // The idea is that if the nonce is still lower than that of a pending transaction, there is no point in actually checking the transaction for a receipt.
 // At the same time if the nonce was already used and this was a few blocks ago we can reasonably assume that it will never confirm.
-
 type Monitor interface {
 	io.Closer
 	// WatchTransaction watches the transaction until either there is 1 confirmation or a competing transaction with cancellationDepth confirmations.
@@ -119,8 +118,9 @@ func (tm *transactionMonitor) watchPending() {
 
 	var lastBlock uint64 = 0
 
+	var added bool // flag if this iteration was triggered by the watchAdded channel
 	for {
-		var added bool = false // flag if this iteration was triggered by the watchAdded channel
+		added = false
 		select {
 		// if a new watch has been added check again without waiting
 		case <-tm.watchAdded:
--- a/pkg/settlement/swap/transaction/monitor.go
+++ b/pkg/settlement/swap/transaction/monitor.go
@@ -19,6 +19,7 @@ import (
 )
 
 var ErrTransactionCancelled = errors.New("transaction cancelled")
+var ErrMonitorClosed = errors.New("monitor closed")
 
 // Monitor is a nonce-based watcher for transaction confirmations.
 // Instead of watching transactions individually, the senders nonce is monitored and transactions are checked based on this.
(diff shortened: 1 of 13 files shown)


zelig (approved):

pkg/settlement/swap/transaction/monitor.go
```diff
@@ -0,0 +1,245 @@
...
+
+		// if there are no watched transactions there is nothing to do
+		if !tm.hasWatches() {
+			continue
+		}
+
+		// switch to new head subscriptions once websockets are the norm
+		block, err := tm.backend.BlockNumber(tm.ctx)
+		if err != nil {
+			tm.logger.Errorf("could not get block number: %v", err)
```
ralph-pichler, line 141: add continue

pkg/settlement/swap/transaction/monitor.go
```diff
@@ -0,0 +1,245 @@
...
+		} else if block <= uint64(lastBlock) {
```
ralph-pichler, line 142: don't continue if triggered by watchAdded

pkg/settlement/swap/transaction/monitor.go
```diff
@@ -0,0 +1,245 @@
...
+			// if the block number is not higher than before there is nothing todo
+			// in the rare case where a block was reorged and the new one is the first to contain our tx we wait an extra block
+			continue
+		}
+
+		if err := tm.checkPending(block); err != nil {
+			tm.logger.Errorf("error while checking pending transactions: %v", err)
+		}
```
ralph-pichler, line 150: update lastBlock

metacertain (approved):
Looks good!

pkg/settlement/swap/transaction/monitor.go
```diff
@@ -0,0 +1,251 @@
...
+	receipt types.Receipt
+	watch   *transactionWatch
+}
+
+// potentiallyConfirmedWatches returns all watches with nonce less than what was specified
+func (tm *transactionMonitor) potentiallyConfirmedWatches(nonce uint64) (watches []*transactionWatch) {
+	tm.lock.Lock()
+	defer tm.lock.Unlock()
+
+	for watch := range tm.watches {
```
acud, line 170: this code block sent me thinking quite a lot and writing quite a few code examples in the playground to see that the for iterator here won't pass the same allocated pointer. it seems to be ok, but consider that others might spend time understanding whether this works as expected or not in the future...

pkg/settlement/swap/transaction/monitor_test.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+
+		select {
+		case receipt := <-receiptC:
+			if receipt.TxHash != txHash {
+				t.Fatal("got wrong receipt")
+			}
+		case err := <-errC:
+			t.Fatal(err)
+		case <-time.After(1 * time.Second):
```
acud, line 67: i wonder if this won't flake on the CI. consider higher values here

pkg/settlement/swap/transaction/monitor_test.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+
+		select {
+		case <-receiptC:
+			t.Fatal("got receipt")
+		case err := <-errC:
+			if !errors.Is(err, transaction.ErrTransactionCancelled) {
+				t.Fatalf("got wrong error. wanted %v, got %v", transaction.ErrTransactionCancelled, err)
+			}
+		case <-time.After(1 * time.Second):
```
acud, line 122: ditto

pkg/settlement/swap/transaction/monitor_test.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+
+		select {
+		case receipt := <-receiptC:
+			if receipt.TxHash != txHash {
+				t.Fatal("got wrong receipt")
+			}
+		case err := <-errC:
+			t.Fatalf("got wrong error. wanted %v, got %v", transaction.ErrTransactionCancelled, err)
+		case <-time.After(1 * time.Second):
```
acud, line 222: ditto

pkg/settlement/swap/transaction/monitor_test.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+
+		select {
+		case <-receiptC2:
+			t.Fatal("got receipt")
+		case err := <-errC2:
+			if !errors.Is(err, transaction.ErrTransactionCancelled) {
+				t.Fatalf("got wrong error. wanted %v, got %v", transaction.ErrTransactionCancelled, err)
+			}
+		case <-time.After(1 * time.Second):
```
acud, line 233: ...

pkg/settlement/swap/transaction/monitormock/monitor.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
acud, line 1: 21

acud (changes requested):
Looks good. Requesting some changes regarding test timeouts (consider raising all of them so that the CI doesn't flake), context propagation can also be improved, with the context fields removed from the struct. I think we've done this before in other places and regretted it, it's better to just pass context from caller and create a root context in the forever loop. Also, some things can be improve as a result in terms of shutdown.

acud (approved):

pkg/settlement/swap/transaction/monitor.go
```diff
@@ -0,0 +1,251 @@
...
+type transactionWatch struct {
+	receiptC chan types.Receipt // channel to which the receipt will be written once available
+	errC     chan error         // error channel (primarily for cancelled transactions)
+
+	txHash common.Hash // hash of the transaction to watch
+	nonce  uint64      // nonce of the transaction to watch
+}
+
+func NewMonitor(logger logging.Logger, backend Backend, sender common.Address, pollingInterval time.Duration, cancellationDepth uint64) Monitor {
+	ctx, cancelFunc := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
ralph-pichler: there is one shared loop for all watched transactions, so which context should it use for the `NonceAt` call which is not even related to any specific transaction? If it uses a context created inside `watchPending` and cancelled in its defer than shutdown might not work at all as it might get stuck waiting on `NonceAt` or `BlockNumber` calls.

Where do you foresee a deadlock happening in the current code?

PR #1483, opened 2021-03-23
fix chown after init

vandot:


acud (approved):

ldeffenb:
Any chance the existing 0.5.3 packages can be updated with this fix?  I just figured out a user's issue in bee-support last night that was likely due to this.  His statestore (or was it localstore) directory was owned by root and his node would not start because it could not create the LOCK file.  That 15 minutes times the number of new user installs would certainly justify a fix to the distributed packages IMHO.

PR #1487, opened 2021-03-24
transaction: add extra gas everywhere

ralph-pichler:
add an extra 20% gas to all transactions if the estimator is used.

deployment transactions can also run out of gas if multiple chequebooks are deployed in the same block.


ldeffenb (commented):
Granted this uses Integer math, but wouldn't gasLimit *= 1.2 be more readable?

ralph-pichler:
@ldeffenb more readable sure, but not compilable. an alternative would be `gasLimit = (gasLimit * 120) / 100`.

acud (approved):

Eknir (approved):
lgtm!

PR #1488, opened 2021-03-24
increase push sync delay to 10s

ralph-pichler:
increase push sync retry delay to 10s for clef run. this is way more than it should be but at least it increases the probability of ci passing until the situation is improved. unfortunately this has no effect on the chunks test as the flag is ignored on the beekeeper side.


acud (approved):

ralph-pichler:
no longer needed

PR #1494, opened 2021-03-25
pusher: receipt can only be nil for ErrWantSelf cases

esadakar:


pkg/pusher/pusher.go
```diff
@@ -165,21 +167,27 @@ LOOP:
...
 
-				storerPeer, err = crypto.NewOverlayAddress(*publicKey, s.networkID)
-				if err != nil {
-					err = fmt.Errorf("pusher: receipt storer address: %w", err)
+					storerPeer, err = crypto.NewOverlayAddress(*publicKey, s.networkID)
+					if err != nil {
+						err = fmt.Errorf("pusher: receipt storer address: %w", err)
+						return
+					}
+				} else if !wantSelf { // receipt is nil, so ErrWantSelf had to be returned
```
acud, line 189: not sure if it's necessary to check this. if I understand correctly, this case is `receipt == nil && !wantSelf` which shouldn't happen at all no?

esadakar: this is technically a part of the signature verification process. We should always get a receipt, and it should be signed. I think the added check is a good thing for cases where people run nodes with modified code bases. What do you think?

esadakar: ahh, the receipt is returned from PushChunkToClosest, the origin node constructs it from the stream, so yes, the extra check is not needed.

acud (approved):

ralph-pichler (approved):

Issue #1444, opened 2021-03-17
Eliminate collections package, remove /files endpoint


acud:
# Task
The `collection` package needs to be deprecated as it proxies every manifest on Swarm. This is remnant of a period where we were pivoting around the approach we wanted to use for manifests.

Right now, when a user uploads a directory or a file, the file(s) get processed with their arbitrary length content chunked and saved, yielding the content addressed hash for the content. Now, in order for us to know which type of file it is, so that browsers can correctly display the file, metadata is required, alongside with file names, directories, etc. This is used with the mantaray manifest, and is used in the `/dirs` endpoint (but not in the `/files`, where only a metadata entry is used from the collections package).

Arising from this change there are 2 problems:
- Existing content cannot be accessed
- `/files` endpoint becomes unusable, since we need to add support to metadata in order to display the file correctly

The solutions:
- For existing content, an external tool (open source) must be written, and that can allow us to remove this code out of the codebase, while allowing users to still get the correct content-addressed hash to use after this change is merged. The external tool should mimic the behavior of this existing manifest type which will be removed, and given an existing root hash, should give the user the correct hash which already exists in the system (but these days requires a few more retrievals to actually get to) (cc @agazso @vojtechsimetka, this might be a good opportunity for synergy)
- `/files` endpoint should be removed, and either `/bzz` or `/dirs` endpoint should support single file uploads. This can be nicely handled. Right now, `/dirs` needs an `x-tar` content type to upload a directory. We can catch all `"other"` content types and fall-back into a `"single-file mode"`, resulting in a manifest generated with just a single entry and the necessary metadata and content addressed hash needed to retrieve.

## Acceptance criteria

- `/files` api removed
- `/bzz` must accept single file upload
- integration tests on beekeeper must be adjusted
- when uploading a single file through the `/bzz` endpoint, create manifest, with a single entry with the _reference_ and metadata, path should be empty

Eknir:
suggestion: first do openAPI spec

agazso:
Regarding the proposed solutions:

1. With this tool given a `/files/<file-reference>` link (or just the `file-reference` part) it would be able to download the file?
[...]

PR #1501, opened 2021-03-26
Removing collections package and API cleanup

aloknerurkar:
#1444 #1543


pkg/api/bzz.go
```diff
@@ -5,34 +5,234 @@
...
+		logger.Errorf("bzz upload file: create manifest, file %q", fileName)
+		jsonhttp.InternalServerError(w, "could not create file manifest")
+		return
+	}
+
+	rootMetadata := map[string]string{
+		manifestWebsiteIndexDocumentSuffixKey: fileName,
+	}
+
+	err = m.Add(ctx, "/", manifest.NewEntry(swarm.ZeroAddress, rootMetadata))
```
zelig, line 179: why rootMetadata not sit on the manifest rootnode entry?

pkg/api/bzz.go
```diff
@@ -252,81 +398,84 @@ FETCH:
...
 	if err != nil {
-		logger.Debugf("bzz download: joiner read file entry %s: %v", address, err)
-		logger.Errorf("bzz download: joiner read file entry %s", address)
+		if errors.Is(err, storage.ErrNotFound) {
+			logger.Debugf("api download: not found %s: %v", reference, err)
+			logger.Error("api download: not found")
+			jsonhttp.NotFound(w, nil)
+			return
+		}
+		logger.Debugf("api download: invalid root chunk %s: %v", reference, err)
```
zelig, line 442: why? can it not be any other error?

zelig: and why notFound error returned on 444?

pkg/api/bzz_test.go
```diff
@@ -32,144 +30,392 @@ import (
...
 	var (
-		bzzDownloadResource = func(addr, path string) string { return "/bzz/" + addr + "/" + path }
-		storer              = smock.NewStorer()
-		ctx                 = context.Background()
-		mockStatestore      = statestore.NewStateStore()
-		logger              = logging.New(ioutil.Discard, 0)
-		client, _, _        = newTestServer(t, testServerOptions{
-			Storer: storer,
+		fileUploadResource   = "/bzz"
+		targets              = "0x222"
```
zelig, line 36: invalid hex?

pkg/api/bzz_test.go
```diff
@@ -32,144 +30,392 @@ import (
...
+
+func createRangeHeader(data []byte, ranges [][2]int) (header string, parts [][]byte) {
+	header = "bytes="
+	for i, r := range ranges {
+		if i > 0 {
+			header += ", "
+		}
+		if r[0] >= 0 && r[1] >= 0 {
+			parts = append(parts, data[r[0]:r[1]])
+			header += fmt.Sprintf("%v-%v", r[0], r[1]-1) // Range: <unit>=<range-start>-<range-end> // end is inclusive
```
zelig, line 382: comments above the relevant line would appear better no?

pkg/api/dirs.go
```diff
@@ -31,27 +30,47 @@ import (
...
+		return
+	}
+	contentType := r.Header.Get(contentTypeHeader)
+	mediaType, params, err := mime.ParseMediaType(contentType)
 	if err != nil {
-		logger.Errorf("dir upload, validate request")
-		logger.Debugf("dir upload, validate request err: %v", err)
+		logger.Errorf("dir upload, invalid content-type")
+		logger.Debugf("dir upload, invalid content-type err: %v", err)
 		jsonhttp.BadRequest(w, "could not validate request")
```
zelig, line 57: could not validate request -> invalid content-type

pkg/api/pin_files_test.go
```diff
@@ -62,27 +64,36 @@ func TestPinFilesHandler(t *testing.T) {
...
 		)
+		if len(actualResponse.Chunks) != len(hashes) {
+			t.Fatalf("Response chunk count mismatch Expected: %d Found: %d",
+				len(hashes), len(actualResponse.Chunks))
+		}
+		for _, v := range actualResponse.Chunks {
+			if counter, ok := hashes[v.Address.String()]; !ok {
+				t.Fatalf("Found unexpected hash %s", v.Address.String())
+			} else if uint64(counter) != v.PinCounter {
+				t.Fatalf("Found unexpected Pin counter Expected: %d Found: %d",
```
zelig, line 89: `found unexpected pin counter: Expected %d, Found %d`

pkg/api/tag_test.go
```diff
@@ -276,7 +278,7 @@ func TestTags(t *testing.T) {
 		if err != nil {
 			t.Fatal(err)
 		}
-		tagValueTest(t, uint32(tagId), 3, 3, 0, 0, 0, 3, expectedHash, client)
+		tagValueTest(t, uint32(tagId), 4, 4, 0, 0, 0, 4, expectedHash, client)
```
zelig, line 281: why are there 4 chunks here?

pkg/manifest/mantaray.go
```diff
@@ -149,7 +149,7 @@ func (m *mantarayManifest) IterateAddresses(ctx context.Context, fn swarm.Addres
 				}
 			}
 
-			if node.IsValueType() && node.Entry() != nil {
+			if node.IsValueType() && node.Entry() != nil && len(node.Entry()) > 0 {
```
zelig, line 152: `node.Entry() != nil` redundant, remove

zelig (changes requested):
great work! can we eliminate `dirs` endpoint?

pkg/api/bzz.go
aloknerurkar: This behavior was retained from the old stuff. I think "/" is the manifest rootnode entry. I will use the constant variable instead of hard coding.

pkg/api/bzz_test.go
aloknerurkar: I actually did not notice this. The test doesn't verify this. From the API perspective, all we need to check is that the value we are returned in the header is the same as the value we sent. This is why the test passes. Not really sure where the targets are being used. Again this test was moved from the files_test.go file. I have retained most of it. Previously they were using this value.

pkg/api/tag_test.go
aloknerurkar: This is according to the new behavior. For each file, we will create a new manifest. The manifest will be similar to a directory with a single file. There will be rootManifestNode entry which will have the index file metadata pointing to the file we are uploading. This will ensure we can serve on ```/bzz/{reference}``` endpoint. There will be a separate file entry with the file metadata. Due to this structure we will now have additional references.

You can check the "traversal_test.go" file. I have added comments about the extra references. For eg, file with single chunk will be stored as follows:
```
"ae16fb27474b41273c0deb355e4405d3cd0a6639f834285f97c75636c9e29df7", // root manifest                                                                                                                    
"0cc878d32c96126d47f63fbe391114ee1438cd521146fc975dea1546d302b6c0", // mainifest root metadata                                                                                                          
"05e34f11a0967e8c09968b69c4f486f569ef58a31a197992e01304a1e59f8e75", // manifest file entry                                                                                                              
"e94a5aadf259f008b7d5039420c65d692901846523f503d97d24e2f077786d9a", // bytes 
```

If user doesn't specify a file name, we use the file reference as the name. The manifest doesn't support empty paths. So we use file reference string as default name. As this is larger than what a single value in the manifest can hold, we actually add some edges. This makes the hashes count at 6. For eg:

```
"7e0a4b6cd542eb501f372438cbbbcd8a82c444740f00bdd54f4981f487bcf8b7", // root manifest                                                                                                                    
"0cc878d32c96126d47f63fbe391114ee1438cd521146fc975dea1546d302b6c0", // manifest root metadata                                                                                                           
"3f538c3b5225111a79b3b1dbb5e269ca2115f2a7caf0e6925b773457cdef7be5", // manifest file entry (Edge)                                                                                                       
"2f09e41846a24201758db3535dc6c42d738180c8874d4d40d4f2924d0091521f", // manifest file entry (Edge)                                                                                                       
"b2662d17d51ce734695d993b44c0e2df34c3f50d5889e5bc3b8718838658e6b0", // manifest file entry (Value)                                                                                                      
"f833c17be12d68aec95eca7f9d993f7d7aaa7a9c282eb2c3d79ab26a5aeaf384", // bytes        
```

pkg/api/bzz.go
aloknerurkar: The error handling at L435 is special handling for "storage.ErrNotFound". But the second one is for the actual error at L434.
This is retained from old changes.

aloknerurkar:
> great work! can we eliminate `dirs` endpoint?

Yes we can technically remove it now itself. It was listed as a separate work item, so was not sure if there is anything else we need to do.

Beekeeper tests need to change. Will be looking into that now.

pkg/api/bzz.go
```diff
@@ -5,34 +5,234 @@
...
+		logger.Errorf("bzz upload file: create manifest, file %q", fileName)
+		jsonhttp.InternalServerError(w, "could not create file manifest")
+		return
+	}
+
+	rootMetadata := map[string]string{
+		manifestWebsiteIndexDocumentSuffixKey: fileName,
+	}
+
+	err = m.Add(ctx, manifest.RootPath, manifest.NewEntry(swarm.ZeroAddress, rootMetadata))
```
acud, line 179: I wonder if some of the logic in this method can be simplified to use the `storeFile` (and `storeDir`?) from the directory upload method

pkg/api/bzz.go
acud: I think it would be better to return a `500` in `L443`. I would also change the error from `invalid root chunk` to at least something like `unexpected error` or something like that. It's a corner case and we should at least be able to filter out 500s and try to figure out what caused this in the first place.

pkg/api/bzz.go
```diff
@@ -252,81 +398,83 @@ FETCH:
...
-	_, err = file.JoinReadAll(ctx, j, buf)
-	if err != nil {
-		logger.Debugf("bzz download: read file metadata %s: %v", address, err)
-		logger.Errorf("bzz download: read file metadata %s", address)
-		jsonhttp.NotFound(w, nil)
-		return
+	// include additional headers
+	for name, values := range additionalHeaders {
+		var v string
+		for _, value := range values {
```
acud, line 450: Not sure if this is actually needed.
consider:
```go
for name, values := range additionalHeaders {
   w.Header().Set(name,values)
}
```

pkg/api/dirs.go
```diff
@@ -31,26 +30,46 @@ import (
...
+	}
+
+	var dReader dirReader
+	switch mediaType {
+	case contentTypeTar:
+		dReader = &tarReader{r: tar.NewReader(r.Body), logger: s.logger}
+	case multiPartFormData:
+		dReader = &multipartReader{r: multipart.NewReader(r.Body, params["boundary"]), logger: s.logger}
+	default:
+		logger.Error("dir upload, invalid content-type for directory upload")
```
acud, line 68: prefix log message with package context:
`api: dir upload: invalid...`

pkg/api/dirs.go
```diff
@@ -224,98 +216,116 @@ func storeDir(ctx context.Context, encrypt bool, reader io.ReadCloser, log loggi
...
-		// - for metadata (1 or more) -> we use estimation function
-		// - for collection entry (1)
-		estimatedTotalChunks := calculateNumberOfChunks(int64(len(metadataBytes)), encrypt)
-		err = tag.IncN(tags.TotalChunks, estimatedTotalChunks+1)
-		if err != nil {
-			return swarm.ZeroAddress, fmt.Errorf("increment tag: %w", err)
-		}
+	contentType := part.Header.Get(contentTypeHeader)
+	if contentType == "" {
+		m.logger.Error("content-type missing in multipart request")
```
acud, line 303: no need to log and error and return it, the calling context should decide whether to log the error or propagate it further up the call stack

pkg/api/dirs.go
```diff
@@ -224,98 +216,116 @@ func storeDir(ctx context.Context, encrypt bool, reader io.ReadCloser, log loggi
...
+		return nil, errors.New("content-type missing")
```
acud, line 304: please pull out the error definitions to the top of the file or to the `api.go` file (this applies to all errors defined in this file)

pkg/api/bzz.go
```diff
@@ -252,81 +398,83 @@ FETCH:
...
-		logger.Debugf("bzz download: unmarshal file entry %s: %v", address, err)
-		logger.Errorf("bzz download: unmarshal file entry %s", address)
-		jsonhttp.NotFound(w, nil)
-		return
+
+	s.downloadHandler(w, r, manifestEntry.Reference(), additionalHeaders, etag)
+}
+
+// downloadHandler contains common logic for dowloading Swarm file from API
+func (s *server) downloadHandler(w http.ResponseWriter, r *http.Request, reference swarm.Address, additionalHeaders http.Header, etag bool) {
```
acud, line 426: just a minor comment, I think that `downloadHandler` is not used anywhere else except `serveManifestEntry`, so I'm not sure we need it as an extra method. Up to you

pkg/api/bzz_test.go
acud: The API won't verify the hex value here, as it is passed to netstore

pkg/api/bzz_test.go
```diff
@@ -32,144 +30,395 @@ import (
...
-		var err error
-		var fileContentReference swarm.Address
-		var fileReference swarm.Address
-		var manifestFileReference swarm.Address
+	t.Run("invalid-content-type", func(t *testing.T) {
+		jsonhttptest.Request(t, client, http.MethodPost, fileUploadResource,
+			http.StatusBadRequest,
+			jsonhttptest.WithRequestBody(bytes.NewReader(simpleData)),
+			jsonhttptest.WithExpectedJSONResponse(jsonhttp.StatusResponse{
+				Message: "invalid content-type header",
```
acud, line 53: the pattern we usually use is to define the error as a package variable (unexported), then export it using `export_test`, then check for the exact error which is exported in the `export_test_ file in the tests, instead of checking for the hardcoded text value (this allows to change the error messages later on quite easily)

pkg/api/dirs_test.go
```diff
@@ -274,62 +271,18 @@ Disallow: /`),
...
-			_, err = file.JoinReadAll(context.Background(), j, buf)
-			if err != nil {
-				t.Fatal(err)
-			}
-			e := &entry.Entry{}
-			err = e.UnmarshalBinary(buf.Bytes())
-			if err != nil {
-				t.Fatal(err)
-			}
+			// if !tc.encrypt {
```
acud, line 277: do we want to remove this? if yes - why? if not - can we make it work? :)

pkg/api/dirs_test.go
```diff
@@ -361,28 +303,28 @@ Disallow: /`),
...
-				jsonhttptest.Request(t, client, http.MethodGet, bzzDownloadResource(resp.Reference.String(), fromPath), http.StatusPermanentRedirect,
+				jsonhttptest.Request(t, client, http.MethodGet,
+					bzzDownloadResource(resp.Reference.String(), fromPath),
+					http.StatusPermanentRedirect,
 					jsonhttptest.WithExpectedResponse([]byte(expectedResponse)),
 				)
 			}
 
-			validateBzzPath := func(t *testing.T, fromPath, toPath string) {
+			validateAltPath := func(t *testing.T, fromPath, toPath string) {
```
acud, line 316: do we need the rename? I think `valdiateBzzPath` captures it no?

acud (commented):
This is really great stuff :) I have a few minor comments, but otherwise this looks great! :pray:

pkg/api/dirs_test.go
aloknerurkar: It works actually... I had commented during some changes.

pkg/api/bzz_test.go
aloknerurkar: This is not required for the test logic. So keeping it as is for now.

pkg/api/bzz.go
aloknerurkar: I thought about it, but with the single file upload, we need to add additional root metadata entry. In case of directories, we only add the file entry. Only part common is the code to create fileMetadata and file entry. I feel its not worth saving the few lines vs having to browse through a separate function for this.

pkg/api/bzz.go
aloknerurkar: Set needs string whereas values is []string

pkg/api/dirs.go
aloknerurkar: I meant to change this. I was thinking of using the prefix "bzz upload dir:", this would follow the one I have used for file upload which is "bzz upload file:"

In order to use the "api:" prefix we should use a prefixed logger in the server itself. I think api is not prefixed in any handler.

pkg/api/dirs_test.go
aloknerurkar: This function validates that the file served on both the 'fromPath' and 'toPath' is the same (for redirects).

I named it 'AltPath' for 'alternate path'. I think this is better no?

pkg/api/bzz_test.go
aloknerurkar: I went down a rabbit hole here. There are too many. I tried to do everything in the bzz/dirs files.

pkg/api/bzz.go
aloknerurkar: Its actually required in bytes file also.

zelig (approved):

pkg/api/bzz.go
acud:
```go
for name, values := range additionalHeaders {
   w.Header().Set(name,strings.Join(values,"; "))
}
```

acud (approved):
Looks good :+1: minor non-blockers otherwise good to go.
Merge checklist:
- beekeeper adjusted tests pass on this cluster
- beekeeper PR merged and new version tagged
- this PR can be then merged

PR #1512, opened 2021-03-30
Pushsync: replicate with neighbors

esadakar:


pkg/pushsync/pushsync.go
```diff
@@ -126,19 +130,36 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 		return swarm.ErrInvalidChunk
 	}
 
-	span, _, ctx := ps.tracer.StartSpanFromContext(ctx, "pushsync-handler", ps.logger, opentracing.Tag{Key: "address", Value: chunk.Address().String()})
-	defer span.Finish()
+	fmt.Println(ps.address, p.Address)
```
esadakar, line 133: i will remove

pkg/pushsync/pushsync.go
```diff
@@ -126,19 +130,34 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 		return swarm.ErrInvalidChunk
 	}
 
-	span, _, ctx := ps.tracer.StartSpanFromContext(ctx, "pushsync-handler", ps.logger, opentracing.Tag{Key: "address", Value: chunk.Address().String()})
-	defer span.Finish()
-
-	// Get price we charge for upstream peer read at headler
```
metacertain, line 132: this wasnt a typo really, we read price from our own pricetable in headler, and use it in handler

pkg/pushsync/pushsync.go
```diff
@@ -126,19 +130,34 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+	// Get price we charge for upstream peer read at header.
 	responseHeaders := stream.ResponseHeaders()
 	price, err := headerutils.ParsePriceHeader(responseHeaders)
-
 	if err != nil {
-		// if not found in returned header, compute the price we charge for this chunk and
```
metacertain, line 137: maybe the 'and' can be removed from the sentence as it is unnecessary, but please keep the comment

pkg/pushsync/pushsync.go
```diff
@@ -147,10 +166,98 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+
+					returnedHeaders := streamer.Headers()
+					_, returnedPrice, _, err := headerutils.ParsePricingResponseHeaders(returnedHeaders)
+					if err != nil {
+						err = fmt.Errorf("push price headers read returned: %w", err)
+						return
+					}
+
+					// check if returned price matches presumed price, if not, return early
+					if returnedPrice != receiptPrice {
```
metacertain, line 217: if the returned price does not match the previously calculated price, we want some things to happen, but not to return early based on this condition.

- we want to notifypeerprice with the index included, as in other places.
- we want to set receiptprice to returnedprice (as in other occurences)

after this, because we are pushsyncing in the neighborhood, and nobody will cover our cost of this (as it is supposed to be free) if the receiptPrice after the header communication is larger than zero, we don't want to push to that peer (instead we want to return in this case and reset the stream).

So because of this, we don't need the reserve, release, credit, because if there was  non zero value to reserve/release/credit, we would return early

metacertain (changes requested):
Nice! Some comments from the bandwidth incentives perspective

zelig (approved):

pkg/pushsync/pushsync.go
```diff
@@ -147,10 +167,99 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+						return
+					}
+
+					// check if returned price matches presumed price, if not, return early.
+					if returnedPrice != receiptPrice {
+						err = ps.pricer.NotifyPeerPrice(peer, returnedPrice, returnedIndex) // save priceHeaders["price"] corresponding row for peer
+						return
+					}
+
+					err = ps.accounting.Reserve(ctx, peer, receiptPrice)
```
metacertain, line 223: Perhaps we can remove the calls to reserve and release here, and leave a comment that as price is confirmed 0 at this point we skip reserve/release

Same for credit in line 243

metacertain (approved):

pkg/pushsync/pushsync.go
esadakar: This also means I need to remove the Debit call at 148? @metacertain

pkg/pushsync/pushsync.go
```diff
@@ -126,19 +130,35 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
-
+	// if not found in returned header, compute the price we charge for this chunk.
 	if err != nil {
-		// if not found in returned header, compute the price we charge for this chunk and
-		ps.logger.Warningf("push sync: peer %v no price in previously issued response headers: %v", p.Address, err)
+		ps.logger.Warningf("pushsync: peer %v no price in previously issued response headers: %v", p.Address, err)
 		price = ps.pricer.PriceForPeer(p.Address, chunk.Address())
 	}
 
+	// if the peer is closer to the chunk, we were selected for replication. Return early.
```
metacertain, line 142: If the chunk does not fall within our neighborhood depth, this is probably not the right conclusion to make.
can we merge this into the iswithindepth block below?

pkg/pushsync/pushsync.go
metacertain: No, we need the debit there, because, in that case, we didn't explicitly check for price being 0 at all

metacertain: in that case, we need the debit, as there is no previous indication the price we charge would be 0

pkg/pushsync/pushsync.go
esadakar: Could the following be an issue:

The node replicating the chunk and the storer node have different ideas about the neighborhood, and the replicating node returns FALSE for IsWithinDepth. In this case, the node may forward it, instead of storing it. This may cause an infinite loop of some sorts.

pkg/pushsync/pushsync.go
esadakar: ok

pkg/pushsync/pushsync.go
esadakar: @metacertain

pkg/pushsync/pushsync.go
```diff
@@ -126,19 +130,35 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+	if dcmp, _ := swarm.DistanceCmp(chunk.Address().Bytes(), p.Address.Bytes(), ps.address.Bytes()); dcmp == 1 {
+		_, err = ps.storer.Put(ctx, storage.ModePutSync, chunk)
+		if err != nil {
+			ps.logger.Errorf("pushsync: chunk store: %w", err)
```
acud, line 146: not sure if `%w` is the verb to use here... since you're logging and not returning w/ wrapping. probably `%v` should be used

pkg/pushsync/pushsync.go
acud: @metacertain Esad is right and we cannot forward any chunk that came from a storer node which is farther away from us. The `IsWithinDepth` might miss that and lead to loops as Esad mentions. Let's leave this as is.

pkg/pushsync/pushsync.go
```diff
@@ -211,8 +307,15 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
 
 		defersFn()
 
-		// find next closest peer
-		peer, err := ps.peerSuggester.ClosestPeer(ch.Address(), skipPeers...)
+		deferFuncs = append(deferFuncs, func() {
```
acud, line 310: this is already done in the code block below `ClosestPeer` call and is unnecessary. Having it here means that we will also unnecessarily print the error twice in case there was an error in the last iteration _and_ and error from `ClosestPeer`

pkg/pushsync/pushsync_test.go
```diff
@@ -351,10 +441,13 @@ func TestPushChunkToNextClosest(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if ta2.Get(tags.StateSent) != 1 {
+	if ta2.Get(tags.StateSent) != 2 {
 		t.Fatalf("tags error")
 	}
 
+	t.Log(pivotAccounting.Balance(peer1))
```
acud, line 448: can  be removed, next line too

acud (changes requested):
Looks great! just a few minor suggestions

pkg/pushsync/pushsync.go
metacertain: why would the replicating node even store the chunk if it's not in it's neighborhood?

In general we dont forward to upstream if i understand correctly, but if there was a special node that would (for some reason) it would be able to spam our nodes with out-of-neighborhood content

pkg/pushsync/pushsync.go
```diff
@@ -147,10 +167,86 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+					returnedHeaders := streamer.Headers()
+					_, returnedPrice, returnedIndex, err := headerutils.ParsePricingResponseHeaders(returnedHeaders)
+					if err != nil {
+						err = fmt.Errorf("push price headers read returned: %w", err)
+						return
+					}
+
+					// check if returned price matches presumed price, if not, return early.
+					if returnedPrice != receiptPrice {
+						err = ps.pricer.NotifyPeerPrice(peer, returnedPrice, returnedIndex) // save priceHeaders["price"] corresponding row for peer
```
acud, line 219: dunno if the comment here is needed as it leaks internals of what happens in the pricer

pkg/pushsync/pushsync.go
```diff
@@ -147,10 +167,86 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+
+					w := protobuf.NewWriter(streamer)
+					ctx, cancel := context.WithTimeout(ctx, timeToWaitForPushsyncToNeighbor)
+					defer cancel()
+
+					err = w.WriteMsgWithContext(ctx, &pb.Delivery{
+						Address: chunk.Address().Bytes(),
+						Data:    chunk.Data(),
+					})
+					if err != nil {
```
acud, line 231: consider: adding metrics on successful and failed replication attempts

acud (approved):
LGTM. Nice

pkg/pushsync/pushsync.go
metacertain: so, if the chunk is not within our ND, we will definietly charge a price for forwarding it.
so, when the receipt issuer pushes towards all peers in neighborhood, those who have a different idea about neighborhood depth will return a price > 0 in headers, and the receipt issuer skips those peers, doesn't push towards them.

metacertain: on the other hand, if the chunk does not fall into our neighborhood depth, that also means we have connections to forward the chunk towards, and I think that's fine, so I would still merge this block into the IsWithinDepth block

esadakar: We can add a IsWithinDepth check inside this distance compare if statement to guarantee that it is the neighborhood. BUT, putting it inside the IsWithinDepth is not good and will cause the loop problem we mentioned above.

metacertain: the way I see it, there could be a loop, if there was a node that does not behave well and forwards upstream for some reason.
 but even in this case the injected forked node, that forwards to upstream, would have to pay for the forwarding each time so that's okay.
 
 on the other hand, if we don't forward to neighborhood such a strange request, we just act rougishly and keep the fee without having managed to get the chunk to it's destination, no?

acud: @metacertain if we merge this block with the next, we might get a case where a node which is farther away from the chunk sent us a chunk and we still forward it and this is undesirable. Your solution of merging both blocks just institutionalizes this, which I believe is incorrect.

metacertain: I understand, but, this shouldn't happen with normal nodes that don't forward to other nodes farther away. (outside of the neighborhood as mentioned before)
In case against the odds it does happen, we should forward the chunk towards it's destination anyway, because we are not the destination in this case (if the chunk does not belong to our neighborhood). 
Or what should we do? We can refuse the forwarding alternatively, reset the stream, and by that, we would systematically prohibit this 'forked node' to push through nodes further away.

metacertain: okay maybe you are right it would be better to reset the stream in strange cases for now

acud: > this shouldn't happen with normal nodes that don't forward to other nodes farther away

so your assumption will not hold for malicious nodes. If I start a node that deliberately does upstream requests in order to generate loops, what then?

metacertain: it will be costly for it, and if the loop persists it will cause the request to timeout and probably be retried on a different route.
Not saying this is not mildly disruptive (though its a bit unlikely), and yes, that's why I said you are probably right, it's better not to accomodate for this kind of behaviour

so let's check distancecmp, if the requester is closer check if its in neighborhood depth (and do the currrent store+debit) 
if its not in neighborhood depth please introduce a stream reset instead, we don't want to debit or store in that case

.github/workflows/beekeeper.yml
```diff
@@ -94,6 +94,12 @@ jobs:
       - name: Test pin remote chunks
         id: pin-remote-1
         run: ./beekeeper check pin-remote --api-scheme http --debug-api-scheme http --disable-namespace --debug-api-domain localhost --api-domain localhost --node-count "${REPLICA}"
+      - name: Test pss
```
acud, line 97: not sure if this is needed

acud (approved):

pkg/pushsync/pushsync.go
esadakar: @metacertain I added the extra neighborhood check after a discussion with @acud . Thanks for the feedback!

PR #1525, opened 2021-04-02
Adding file upload handler

aloknerurkar:
- Adding new file upload handler (POST /bzz)
- all unit tests passing

### DIFF
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -5,12 +5,19 @@
 package api
 
 import (
+	"bufio"
 	"context"
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"io"
+	"io/ioutil"
+	"mime"
+	"mime/multipart"
 	"net/http"
+	"os"
 	"path"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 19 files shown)


pkg/manifest/manifest.go
```diff
@@ -16,6 +16,14 @@ import (
 
 const DefaultManifestType = ManifestMantarayContentType
 
+const (
+	RootPath                      = "/"
```
aloknerurkar, line 20: @acud 
The reason I moved these here is because I need them in the traversal tests. The traversal package doesn't need this. I don't see any particular reason why any package apart from api would need this.

api package imports traversal, so I cannot import api in traversal pkg. Also if these change in future, the traversal tests need to change. So maybe enough reason to move these here? That way both packages can access these as they both depend on manifest

pkg/api/bzz.go
```diff
@@ -21,13 +28,255 @@ import (
...
+				return
+			}
+		}
+	}
+
+	// Add the tag to the context
+	ctx := sctx.SetTag(r.Context(), tag)
+
+	if mediaType == multiPartFormData {
+		mr := multipart.NewReader(r.Body, params["boundary"])
```
acud, line 106: i would remove the whole multipart section from the file upload part. cc @agazso

PR #1532, opened 2021-04-03
Add /tags routes to the debug API

### LINKED ISSUE #1530: Add all /tags routes to the debug API
# Task
Currently, only the `GET /tags/{uid}` route is available on the debug API. This is problematic, because when `bee` is ran in gateway mode, all `/tags` endpoints are disabled on the external API.

As far as I can tell, this leaves us with no way of getting tag information if all created tags aren't recorded when files are created.

Alternatively, maybe some `/tags` routes should still be accessible publicly in gateway mode. Having `GET /tags` and `GET /tags/{uid}` could prove very useful to create statistics for example.

## Acceptance criterea
Be able to list tags and fetch information about them when running in gateway mode.

filoozom:
Fixes #1530

### DIFF
--- a/pkg/debugapi/router.go
+++ b/pkg/debugapi/router.go
@@ -160,8 +160,20 @@ func (s *Service) newRouter() *mux.Router {
 		})
 	}
 
+	router.Handle("/tags", jsonhttp.MethodHandler{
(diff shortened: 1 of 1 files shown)


filoozom:
Nevermind, not as trivial as I thought, also don't want to merely copy `tag.go`.

PR #1535, opened 2021-04-05
api: endpoints for stamp issuers

acud:
This PR adds two API endpoints as discussed with the JS team cc @agazso. These return either all available batches or the information about a specific batch and can be used to further build upon.
I did not add the OpenAPI spec and realised that it also doesn't exist for the other API endpoint (POST). And considering the amount of changes of the openapi spec on `master` I would suggest to postpone this in a few days, so that we can first rebase the storage incentives branch over `master, then deal with this, if there's no objection.

### DIFF
--- a/pkg/postage/stampissuer.go
+++ b/pkg/postage/stampissuer.go
@@ -112,7 +112,10 @@ func toString(buf []byte) string {
 }
 
 func (st *StampIssuer) Utilization() uint32 {
-	return 0
+	top := 0
+
+	//TODO return top bucket value for now
+	return top
 }
 
 func (s *StampIssuer) ID() []byte {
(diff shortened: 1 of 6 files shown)


pkg/postage/stampissuer.go
```diff
@@ -110,3 +110,16 @@ func toString(buf []byte) string {
 	}
 	return string(buf[i:])
 }
+
+func (st *StampIssuer) Utilization() uint32 {
+	top := 0
+
+	//TODO return top bucket value for now
```
zelig, line 117: fullness percentage can then be calculated as:

     max_bucket_value / 2 ^ (batch_depth - bucket_depth)

zelig (approved):

Issue #1504, opened 2021-03-26
Wrongly concluding that I am the closest node when building up Kademlia


ldeffenb:
#### Summary
I had a node which was trying (and failing) to push a bunch of chunks to closest peers, possibly due to #1503 .  This node was retrying the pushes over and over again.  I restarted the node and saw the entire pending push queue dump to the first peer that was successfully connected.  I can only presume that the node had somehow different or better connections into the required destination neighborhood, but have no way to be certain.

#### Steps to reproduce
Build up a pusher retry queue in a node for whatever reason.  It may be #1503 or a bunch of non-responsive nodes causing "stream reset" or "context deadline exceeded", but for whatever reason, your node cannot deliver one or more chunks to its closest nodes.

Restart that node and you should see the entire queue of pending chunks be delivered through the first connected peer.

#### Expected behavior
If the chunks can be delivered after a restart, why aren't they able to be delivered before the restart?  Rather than continuously trying and failing the same set of 5 "closest" peers, should pusher retries retain the peers that were previously failed and try a different, or increasing, set of peers on subsequent push retries?

#### Actual behavior
Pusher tries over and over until the node restarts and then the chunks are delivered.

The full node --verbosity 5 log is at: https://pastebee.com/?f031416fcefd6082e5ae54a5c6d5e4a0eb139a9804f312cb2e84134c75301fe6

But here's the gist of it:
time="2021-03-26T08:33:37-04:00" level=info msg="version: 0.5.3-23baf60-dirty"
time="2021-03-26T08:33:37-04:00" level=warning msg="clef is not enabled; portability and security of your keys is sub optimal"
time="2021-03-26T08:33:38-04:00" level=info msg="using existing swarm network address: 978a4128246abb0283a440d21731f629e4cc59394decae2bb01679fcda879d6e"
time="2021-03-26T08:33:38-04:00" level=info msg="swarm public key 03faf2886ae088aba53ca7f8e6427842b9a2e5c2cfab0bafd71e05ca9e45137252"
time="2021-03-26T08:33:38-04:00" level=debug msg="using existing libp2p key"
time="2021-03-26T08:33:38-04:00" level=debug msg="using existing pss key"
time="2021-03-26T08:33:38-04:00" level=info msg="pss public key 02aeae9c39c66ed030e5c309fc1d1bd4c01b479c0c7cc1156662d7f24f07e5c22e"
[...]

PR #1537, opened 2021-04-05
Push/pull sync - neighborhood syncing

esadakar:


pkg/puller/puller.go
```diff
@@ -148,23 +139,8 @@ func (p *Puller) manage() {
...
-						}
-					} else {
-						// already syncing, recalc
-						peerEntry := peer{addr: peerAddr, po: po}
-						peersToRecalc = append(peersToRecalc, peerEntry)
-					}
-				} else {
-					// within depth, sync everything >= depth
+				if po >= depth {
+					// within depth, sync everything
```
ldeffenb, line 143: As much as I currently dislike pullsync and how it pulls chunks into my node's localstore that have no business being there (well outside my neighborhood), I believe this change will result in leaving some chunks destined to never arrive in their desired neighborhood.  Consider the following situation:

An uploading node with a pusher queue drops network connectivity for some reason.  All peers are dropped due to this network disruption.  When the network event is past, the node will start establishing new peer connections.  As soon as a single connection is established, the pusher will flush the queue entirely via the first connected peer.  And if the node itself is closer to the chunk than that peer, then those chunks will be stored locally (topology.ErrWantSelf from pushToClosest) and will not be pushed again (see #1504 )

Without the puller rummaging around in nodes outside its own neighborhood, these misplaced chunks will never move outside the node that invalidly adopted them and will therefore be very unlikely to be discovered by any retrieval attempt which would be routed to those chunks' preferred neighborhood nodes.

I wish I could offer a fix for this, but I'm still wrapping my head around the issues with pullsync in general and working on formulating an issue and a (better) PR to correct those issues.

acud: @ldeffenb thanks we will keep this in mind. We're making improvements across the board and pushsync will also undergo more improvements soon. Thanks for the feedback :+1:

acud (approved):

pkg/puller/puller.go
ldeffenb: I just tried this patch on my mutant 0.5.3 node and it is lacking one thing.  As the node starts up, the neighborhood is quite low and pullers are started for peers that will end up outside the neighborhood once a full complement of peers has been acquired and the neighborhood has increased.

I believe there needs to be an else clause to cancel peers that are now outside the neighborhood but already started syncing while the neighborhood of peers was being established.

I know this because I have debug logs showing me the peers going through SyncInterval and the first ones to appear are well out of my leading bits range and they continue to operate even as the connected peers is increasing the neighborhood bin index.  I added the following else to force a recalc on syncing peers that are no longer within the neighborhood depth.

				} else {
					if _, ok := bp[peerAddr.String()]; ok {	// Force a recalc on syncing peers no longer within depth
						// already syncing, recalc
						peerEntry := peer{addr: peerAddr, po: po}
						peersToRecalc = append(peersToRecalc, peerEntry)
					}
				}

But this is not enough because (at least my) recalcPeer routine still tries to keep some syncing for peers outside the current depth.  So I made sure that all of the bins for peers "outside of depth" end up in "dontWant" so that they will all be cancelled.

pkg/puller/puller.go
```diff
@@ -244,23 +220,7 @@ func (p *Puller) recalcPeer(ctx context.Context, peer swarm.Address, po, d uint8
...
-				continue
-			}
-			dontWant = append(dontWant, i)
-		}
-
-		if !syncCtx.isBinSyncing(want) {
-			p.syncPeerBin(ctx, syncCtx, peer, want, c[want])
-		}
-		syncCtx.cancelBins(dontWant...)
+		panic("shouldn't")
```
ldeffenb, line 223: Based on my comments above, I believe that out-of-neighborhood peers that are already syncing need to be cancelled.  That will cause this else to need to invoke syncCtx.cancelBins(dontWant...) where dontWant contains all possible bins.  This will shut down pullers that were started before the network depth was fully established.

ldeffenb: Something like:
	} else {
		// outside of depth, cancel them all
		var (
			dontWant = []uint8{0} // never want bin 0
		)

		for i := uint8(0); i < p.bins; i++ {
			dontWant = append(dontWant, i)
		}

		syncCtx.cancelBins(dontWant...)
	}

pkg/pushsync/pushsync.go
```diff
@@ -147,10 +176,89 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+			// Push the chunk to some peers in the neighborhood in parallel for replication.
+			// Any errors here should NOT impact the rest of the handler.
+			err = ps.topologyDriver.EachNeighbor(func(peer swarm.Address, po uint8) (bool, bool, error) {
+
+				// skip forwarding peer
+				if peer.Equal(p.Address) {
+					return false, false, nil
+				}
+
+				if count == nPeersToPushsync {
```
zelig, line 189: why hae this limit?

pkg/pushsync/pushsync.go
```diff
@@ -147,10 +176,89 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+							ps.metrics.TotalReplicatedError.Inc()
+						} else {
+							ps.metrics.TotalReplicated.Inc()
+						}
+					}()
+
+					// price for neighborhood replication
+					const receiptPrice uint64 = 0
+
+					headers, err := headerutils.MakePricingHeaders(receiptPrice, chunk.Address())
```
zelig, line 209: does this not shadow the `err` in defer func?

pkg/topology/mock/mock.go
```diff
@@ -85,7 +85,7 @@ func (d *mock) Peers() []swarm.Address {
 	return d.peers
 }
 
-func (d *mock) ClosestPeer(_ swarm.Address, skipPeers ...swarm.Address) (peerAddr swarm.Address, err error) {
+func (d *mock) ClosestPeer(adr swarm.Address, skipPeers ...swarm.Address) (peerAddr swarm.Address, err error) {
```
zelig, line 88: adr -> addr

pkg/topology/mock/mock.go
```diff
@@ -128,11 +138,27 @@ func (*mock) NeighborhoodDepth() uint8 {
...
 
+func (m *mock) IsWithinDepth(adr swarm.Address) bool {
+	return false
+}
+
+func (m *mock) EachNeighbor(f topology.EachPeerFunc) error {
+	return m.EachPeer(f)
+}
+
+func (*mock) EachNeighborRev(topology.EachPeerFunc) error {
```
zelig, line 149: where do we need this function?

pkg/topology/topology.go
```diff
@@ -51,6 +52,15 @@ type EachPeerer interface {
 	EachPeerRev(EachPeerFunc) error
 }
 
+type EachNeighbor interface {
+	// EachNeighbor iterates from closest bin to farthest within the neighborhood.
+	EachNeighbor(EachPeerFunc) error
+	// EachNeighborRev iterates from farthest bin to closest within the neighborhood.
+	EachNeighborRev(EachPeerFunc) error
```
zelig, line 59: i dont think we ever need this

pkg/puller/puller.go
acud: @ldeffenb you will not be pulling for farther away peers. Once kademlia depth changes, puller also updates and cancels all streams outside of depth

pkg/pushsync/pushsync.go
acud: I guess you mean we should sent to all peers coming out of the iterator right? Not sure why this is coming up now and not as part of the original PR review

pkg/pushsync/pushsync.go
acud: no

pkg/topology/mock/mock.go
acud: we'll remove that in a future PR since anatol is also working on this code :+1:

pkg/puller/puller.go
```diff
@@ -151,6 +151,12 @@ func (p *Puller) manage() {
 						peerEntry := peer{addr: peerAddr, po: po}
 						peersToRecalc = append(peersToRecalc, peerEntry)
 					}
+				} else {
```
ldeffenb, line 154: Yep, that will the no-longer-in-neighborhood peers to recalc and be cancelled.  But they will now be queued to cancel every time through the loop because they're never removed from p.syncPeers which is where bp comes from.

I had this code for a while with a logger message and couldn't figure out why my bin1 peer was continually recalcing even though the depth was 13.  I finally changed this else to:

				} else {
					if _, ok := bp[peerAddr.String()]; ok {	// Disconnect syncing peers no longer within depth
						// already syncing, recalc
						peerEntry := peer{addr: peerAddr, po: po}
						peersDisconnected[peerAddr.String()] = peerEntry
						p.logger.Debugf("puller cancelling peer %s po %d depth %d", peerAddr, po, depth)
					}
				}

ralph-pichler:
please rebase and remove the commits which change the workflow

metacertain (approved):

zelig (approved):

PR #1539, opened 2021-04-06
update default factory and add support for legacy factories

ralph-pichler:
* updates the expected factory contract to the new version (with proxy-based and create2 deployments)
* adds support for legacy factories from which we still accept chequebooks
* adds recipient to abi call mock for better testing

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1539)


acud (approved):
:+1:

zelig (approved):
i wish we could eliminate the `common` package

PR #1551, opened 2021-04-07
Storage incentives

acud:
Todos:
- [ ] OpenAPI spec
- [ ] allow data import with a batch ID that will stamp all imported chunks
- [ ] indexing scheme
- [ ] cli flag for capacity turns into cache size


acud:
superseded by #1562

Issue #1302, opened 2021-02-20
Get current pins


Eknir:
# User story

## Story
As a local pinner, I want to retrieve all pinned roots of the content which I am currently pinning, such that I can manage these pins (remove, validate if still available, etc).

## Acceptance criteria
- The API exposed to the user should interact with pinned roots only

## Background
Pinned can come into the node either by uploads using the `Swarm-Pin` header, or, by downloads using the traversal service.

Right now, the implementation of the pinning API is a bit skewed and misinterpreted as it exposes localstore internals through the API package, by leaking individual chunk pin counters. This is not very user-friendly, complicates user onboarding, as well as leaks undesired internals that could hinder storage guarantees given by storage incentives that use pinning under-the-hood

## Tasks
- An additional abstraction should be defined to cater for the pinned root interaction. 
- Basic CRUD operations should be defined on the interface. 
- Pinned roots should be persisted in statestore
- API changes could be confined to just an extra middleware that would intercept incoming upload requests that have the pin header, then add the root once the request returns.
- Current API endpoints should be scratched and reimplemented (GET,DELETE,POST) using the aforementioned abstraction 
- PinCounter methods removed from the `swarm.Chunk` interface
- Beekeeper integration tests adjusted

Revised API definitions:
- `GET /pins` - gets all pinned roots
- `DELETE /pins/<root>` - unpins a root and all of its associated chunks (manifests and actual data chunks)
- `POST /pins/<unpinned root>` - tries to pin an unpinned root, fetching missing chunks from the network

significance:
> Current API endpoints should be scratched and reimplemented (GET,DELETE,POST) using the aforementioned abstraction

i think it would be good to retain visibility of the individual chunks that are pinned at the api level. when we are ensuring data availability from a data stewards position (a la 1508) it is important to know that each and every chunk is alive and available from all sections of the network.

PR #1566, opened 2021-04-14
Change pins API

mrekucci:
#1302


pkg/api/bytes.go
```diff
@@ -66,6 +67,16 @@ func (s *server) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
+
+	if strings.ToLower(r.Header.Get(SwarmPinHeader)) == "true" {
```
acud, line 71: we need to add this logic also to:
`bzz POST` handler - this should be in both forks - file upload and dir upload handler
`chunk POST` handler
`feed POST` handler
`soc POST` handler

pkg/api/pin.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/gorilla/mux"
+)
+
+// pinChunk pin's the already created chunk given its address.
+// It fails if the chunk is not present in the local store.
+// It also increments a pin counter to keep track of how many
+// pin requests are originating for this chunk.
+func (s *server) pinChunk(w http.ResponseWriter, r *http.Request) {
```
acud, line 21: this is no longer needed, since we don't provide explicit chunk level pinning (only if this is done through the chunk upload service or through pinning unpinned roots which we can also use to potentially unpin a single chunk)

pkg/api/pin.go
```diff
@@ -0,0 +1,167 @@
...
+		jsonhttp.InternalServerError(w, nil)
+		return
+	}
+
+	jsonhttp.OK(w, nil)
+}
+
+// unpinChunk unpin's an already pinned chunk. If the chunk is
+// not present or if the pin counter is zero, it raises error.
+func (s *server) unpinChunk(w http.ResponseWriter, r *http.Request) {
```
acud, line 69: likewise. both should be handled with the traversal service. have a look at how the traversal service is used in `pin_bytes.go` and `pin_files.go`. we should be using it in a similar way, although, some minor changes to the traversal service might be needed to correctly accomplish the task

pkg/api/pin.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		s.logger.Debugf("pin chunk: unable to parse chunk address: %v", err)
+		s.logger.Error("pin chunk: unable to parse address")
+		jsonhttp.BadRequest(w, "bad address")
+		return
+	}
+
+	switch err := s.storer.Set(r.Context(), storage.ModeSetPin, addr); {
+	case errors.Is(err, storage.ErrNotFound):
+		ch, err := s.storer.Get(r.Context(), storage.ModeGetRequest, addr)
```
zelig, line 32: there should be a `ModeGetRequestPin`

pkg/api/pin.go
```diff
@@ -0,0 +1,167 @@
...
+		if err != nil {
+			s.logger.Debugf("pin chunk: unable to get chunk for address %q from netstore: %v", addr, err)
+			s.logger.Error("pin chunk: unable to get chunk for address %q from netstore")
+			jsonhttp.NotFound(w, nil)
+			return
+		}
+
+		_, err = s.storer.Put(r.Context(), storage.ModePutRequestPin, ch)
```
zelig, line 40: the getrequest already saves this to localstore, in fact the put will come back after a has call returns true

pkg/api/pin.go
```diff
@@ -0,0 +1,167 @@
...
+		jsonhttp.InternalServerError(w, http.StatusText(http.StatusInternalServerError))
+		return
+	}
+
+	if !has {
+		jsonhttp.NotFound(w, nil)
+		return
+	}
+
+	_, err = s.storer.PinCounter(addr)
```
zelig, line 91: why is this needed?

pkg/pinning/pinning_test.go
```diff
@@ -0,0 +1,61 @@
...
+func TestPinningService(t *testing.T) {
+	var (
+		svc  = NewService(mock.NewStateStore())
+		addr = swarm.NewAddress([]byte{0})
+	)
+
+	if err := svc.CreatePin(addr); err != nil {
+		t.Fatalf("CreatePin(...): unexpected error: %v", err)
+	}
+	// The second call is to test idempotency.
```
zelig, line 21: but...strictly speaking you only testing that the result is as  desired after 2 calls :)

pkg/pinning/pinning_test.go
```diff
@@ -0,0 +1,61 @@
...
+		t.Fatalf("DeletePin(...): unexpected error: %v", err)
+	}
+	has, err = svc.HasPin(addr)
+	if err != nil {
+		t.Fatalf("HasPin(...): unexpected error: %v", err)
+	}
+	if has {
+		t.Fatalf("HasPin(...): have %t; want %t", has, !has)
+	}
+	// The second call is to test idempotency.
```
zelig, line 56: now you only checking after the first call :)

zelig (commented):
superb code quality. it is the feeling that i am learning as i read you

mrekucci:
@acud @zelig thanks guys for your valuable feedback, it's still WIP so I'll notify you once it's ready for review

zelig (approved):
überneat

pkg/pinning/pinning.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+		case err != nil:
+			return fmt.Errorf("unable to set pin for leaf %q of root %q: %w", leaf, addr, err)
+		}
+		return nil
+	}
+
+	if traverse {
+		if err := s.traverser.Traverse(ctx, addr, iterFn); err != nil {
+			return fmt.Errorf("traversal of %q failed: %v", addr, err)
```
acud, line 87: `%v` should be `%w`

pkg/traversal/traversal.go
```diff
@@ -16,127 +16,63 @@ import (
...
-) error {
-	j, _, err := joiner.New(ctx, s.storer, reference)
-	if err != nil {
-		return fmt.Errorf("traversal: joiner: %s: %w", reference, err)
+	ls := loadsave.New(s.store, storage.ModePutRequest, false)
+	switch mf, err := manifest.NewDefaultManifestReference(addr, ls); {
+	case errors.Is(err, manifest.ErrInvalidManifestType):
+		break
+	case err != nil:
+		return fmt.Errorf("traversal: unable to read manifest for %q: %w", addr, err)
```
acud, line 59: `NewDefaultManifestReference` actually will not read anything until you start using the data structure somehow, so this error will never occur

acud (commented):
looks great :pray: minor few comments

pkg/api/chunk.go
```diff
@@ -111,6 +112,15 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set(SwarmTagHeader, fmt.Sprint(tag.Uid))
 	}
 
+	if strings.ToLower(r.Header.Get(SwarmPinHeader)) == "true" {
+		if err := s.pinning.CreatePin(ctx, chunk.Address(), false); err != nil {
+			s.logger.Debugf("chunk upload: creation of pin for %q failed: %v", chunk.Address(), err)
```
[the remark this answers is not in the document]
mrekucci: The following:

```golang
address := swarm.MustParseHexAddress("f30c0aa7e9e2a0ef4c9b1b750ebfeaeb7c7c24da700bb089da19a46e3677824b")
	logger.Debugf("testing address string: %q", swarm.NewChunk(address, []byte{0x1}).Address())
```

will print:

```sh
DEBU[2021-04-22T17:57:41+02:00] testing address string: "f30c0aa7e9e2a0ef4c9b1b750ebfeaeb7c7c24da700bb089da19a46e3677824b"
```

PR #1568, opened 2021-04-15
localstore: dont acquire batchMu for a single chunk that already exists

acud:
This PR optimizes the way we acquire the `batchMu` in localstore for a single chunk.

In `pushsync`, chunks may be resent over the network. However, if a chunk already exists, the lock must not be acquired. This should improve lock contention on chunks which are retried, but as well leave the current situation in which two copies of the same chunk which are competing over the same lock will not be aware of each other. In order to solve this edge case, some sort of detection of inflight writes would be necessary, with result propagation to potentially multiple writers.

### DIFF
--- a/pkg/localstore/mode_put.go
+++ b/pkg/localstore/mode_put.go
@@ -44,7 +44,7 @@ func (db *DB) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk)
 	return exist, err
 }
 
-// put stores Chunks to database and updates other indexes. It acquires lockAddr
+// put stores Chunks to database and updates other indexes. It acquires batchMu
 // to protect two calls of this function for the same address in parallel. Item
 // fields Address and Data must not be with their nil values. If chunks with the
 // same address are passed in arguments, only the first chunk will be stored,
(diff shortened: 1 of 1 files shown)


pkg/localstore/mode_put.go
```diff
@@ -44,14 +44,27 @@ func (db *DB) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk)
...
 func (db *DB) put(mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
+	// this is an optimization that tries to optimize on already existing chunks
+	// not needing to acquire batchMu. This is in order to reduce lock contention
+	// when chunks are retried across the network for whatever reason.
+	if len(chs) == 1 {
+		has, err := db.retrievalDataIndex.Has(chunkToItem(chs[0]))
+		if err != nil {
+			return []bool{false}, err
+		}
+		if has {
```
zelig, line 63: if `has` is false here, it can still happen that the chunk arrives after another routine but before the current lock is acquired. What issue what that cause?

zelig (approved):

pkg/localstore/mode_put.go
acud: if `has` is false, we still acquire the lock and try to put the chunk. it does not resolve the current situation which is described in the description:


> ...two copies of the same chunk which are competing over the same lock will not be aware of each other. In order to solve this edge case, some sort of detection of inflight writes would be necessary, with result propagation to potentially multiple writers

acud: so the result is that the same chunk is put twice (or more)

acud: i actually stand corrected. the two chunks dont know about each other, the one will acquire the mutex and be put, the second will acquire the mutex, only to find out it already exists then will return

pkg/localstore/mode_put.go
```diff
@@ -44,14 +44,27 @@ func (db *DB) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk)
...
 // slice. This is the same behaviour as if the same chunks are passed one by one
 // in multiple put method calls.
 func (db *DB) put(mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
+	// this is an optimization that tries to optimize on already existing chunks
+	// not needing to acquire batchMu. This is in order to reduce lock contention
+	// when chunks are retried across the network for whatever reason.
+	if len(chs) == 1 {
+		has, err := db.retrievalDataIndex.Has(chunkToItem(chs[0]))
+		if err != nil {
+			return []bool{false}, err
```
janos, line 61: Just a small suggestion.

```suggestion
			return nil, err
```

janos (approved):

Issue #1414, opened 2021-03-10
peers added from debugapi don't end up in kademlia


acud:
#### Summary

Adding a peer from the debugapi is not showing afterwards in `/topology`.

#### Steps to reproduce

Add the peer, then query `/topology`. The peer will not be found.

#### Expected behavior

Peer found in topology

#### Actual behavior

No peer found.

Sidenote: we used to have notification logic wired up into the p2p connect method, but this apparently got lost at some point when this got refactored.

r0qs:
I believe that adding the following line after the [s.p2p.Connect](https://github.com/ethersphere/bee/blob/master/pkg/debugapi/peer.go#L30) in the peer connect handler of the debugapi would fix the problem since the peer is only not added in the topology of the caller node:

```go
err = s.topologyDriver.AddPeers(r.Context(), bzzAddr.Overlay)
```

Or do you think that it is necessary to change the current behavior of the [Connect](https://github.com/ethersphere/bee/blob/master/pkg/p2p/p2p.go#L22) in the p2p service?

acud:
Actually no. Calling `AddPeers` will not cause the peer to show up in kademlia, unless it tries to actively dial the peer. We should call the `Connected` hook on kademlia either from the p2p abstraction or from the debugapi. Whatever works best

Eknir:
Is this already resolved?

r0qs:
Hi @Eknir ,

As far I'm aware, the problem is not solved yet.
As @acud mentioned, we should use `Connected` to notify kademlia about the new peer, but the topology driver does not have access to the connection notifier methods anymore.

The easy solution would be to undo part of what was done in the PR https://github.com/ethersphere/bee/pull/641 calling the Connected in debug api:
```go
s.TopologyDriver.Connected(r.Context(),  p2p.Peer{Address: bzzAddr.Overlay}, true)
```
And adding the `Notifier` interface from the p2p package back to the topology `Driver` interface, so the debug API will have access to the Connected method. Like so:

```go
type Driver interface {
	Notifier
	...
}

type Notifier interface {
	Connected(context.Context, p2p.Peer) error
	Disconnected(p2p.Peer)
}
```

[...]

PR #1585, opened 2021-04-20
Add notifier to topology driver in debug API

r0qs:
Fix https://github.com/ethersphere/bee/issues/1414

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1585)

### DIFF
--- a/pkg/debugapi/peer.go
+++ b/pkg/debugapi/peer.go
@@ -35,6 +35,14 @@ func (s *Service) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if err := s.topologyDriver.Connected(r.Context(), p2p.Peer{Address: bzzAddr.Overlay}, true); err != nil {
+		_ = s.p2p.Disconnect(bzzAddr.Overlay)
+		s.logger.Debugf("debug api: peer connect handler %s: %v", addr, err)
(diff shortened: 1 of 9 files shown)


acud:
@r0qs thanks for the PR! I think that there could be a data race here, if we connect to the peer using the p2p dependency. If the peer disconnects quickly, the `Connected` notifier will still be called, resulting in a rouge peer in kademlia which will never be removed. A potentially favorable approach would be to not call the p2p dial method at all from the debug api, but rather add a flag that forces dialing to `AddPeer` in kademlia, forcing kademlia to both add the peer to the known peers list but to also immediately connect to it, resulting in a consistent state also if the peer disconnects, since the connect call is under lock, it will not propagate until the `connect` returns. WDYT?

r0qs:
Hi @acud,

ah ok, I see now. Sorry for the misunderstanding of the flag approach. I'll revert the changes and apply the suggestions.

But how can I get the overlay address of the peer without attempt to dial? Or there is another way to get the overlay based on the multi-address?

acud:
@r0qs can we have a rebase? happy to merge this once conflicts are out of the way

acud:
closing due to inactivity

r0qs:
Hey, sorry for the late reply. I did a rebase and open the PR again https://github.com/ethersphere/bee/pull/2093

PR #1586, opened 2021-04-20
Storage incentives reservestate api

esadakar:


esadakar:
reservestate

zelig (commented):
consider adding a test?

pkg/postage/batchstore/store.go
```diff
@@ -60,6 +60,15 @@ func New(st storage.StateStorer, unreserveFunc unreserveFn) (postage.Storer, err
 	return &store{st, cs, rs, unreserveFunc}, nil
 }
 
+func (s *store) GetReserveState() *postage.Reservestate {
```
acud, line 63: bear in mind you're leaking the package internals here: `Outer` and `Inner` are pointers. It would be prudent to create a copy of them.

acud (approved):
+1 for a small test in `debugapi`

pkg/postage/batchstore/store.go
```diff
@@ -60,6 +60,15 @@ func New(st storage.StateStorer, unreserveFunc unreserveFn) (postage.Storer, err
...
+	return &postage.Reservestate{
+		Radius:    s.rs.Radius,
+		Available: s.rs.Available,
+		Outer:     *s.rs.Outer,
```
acud, line 67: `big.Int`s are meant to be passed as pointers, not as values, so it is needed to create a new `big.Int` and give it the value which is currently in the existing fields.
Also, there is a typo in the line below, where `Outer` is dereferenced as value for Inner

pkg/postage/reservestate.go
```diff
@@ -0,0 +1,10 @@
+package postage
+
+import "math/big"
+
+type Reservestate struct {
+	Radius    uint8   `json:"radius"`
+	Available int64   `json:"available"`
+	Outer     big.Int `json:"outer"` // lower value limit for outer layer = the further half of chunks
```
acud, line 8: `*big.Int`, for both

acud (changes requested):
some comments relating to passing of `big.Int`s

acud (approved):

pkg/postage/reservestate.go
```diff
@@ -0,0 +1,10 @@
+package postage
```
acud, line 1: copyright missing here. otherwise LGTM

pkg/debugapi/postage_test.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
acud, line 1: 2021

pkg/debugapi/postage.go
```diff
@@ -0,0 +1,15 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
acud, line 1: 21

AuHau:
Btw. @esadakar the new `GET /reservestate` endpoint is not documented in OpenAPI ;-)

PR #1587, opened 2021-04-21
Postage stamp indexing

zelig:
this PR extends postage stamps with indexes
- indexes enable replacing stored chunks at the discretion of the owner: storer nodes store  only the latest chunk stamped with the same stamp (same batch same index)
- offer an easy way to catch overissuance as part of stamp validity check
- makes it possible to limit spoofing density proofs by mining chunks to a batch
- makes it possible to transfer ownership of arbitrary indexes 

the index is a 64-bit serialisation of
 - bucket depth (uint8, 1st byte, <24)
 - bucket index (neighbourhood index, uint32 <2^depth, bytes 2-4)
 - and the within-bucket index (uint32 <2^(batchdepth-bucketdepth), bytes 5-8)


zelig:
superceded by #1625

PR #1594, opened 2021-04-23
batchstore, kademlia: add radius as depth upper bound

acud:
This PR wires in the postage radius of responsibility into kademlia as an upper bound of the neighborhood depth, paving the way for agreement between connected neighborhoods and data persistence responsibility.


acud:
deepsource complaints addressed in a different PR

zelig (approved):

PR #1595, opened 2021-04-23
pushsync: no forwarder retry, longer timeouts

acud:


zelig (approved):

pkg/pushsync/pushsync.go
```diff
@@ -310,84 +299,87 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
...
-			continue
-		}
+	streamer, err := ps.streamer.NewStream(ctx, peer, nil, protocolName, protocolVersion, streamName)
+	if err != nil {
+		return nil, fmt.Errorf("new stream for peer %s: %w", peer.String(), err)
+	}
+	defer streamer.Close()
+
+	// Reserve to see whether we can make the request
+	err = ps.accounting.Reserve(ctx, peer, receiptPrice)
```
aloknerurkar, line 342: Seems like this can fail. It would be better to move this before the `NewStream` call. That way if we fail here we don't waste additional effort.

PR #1597, opened 2021-04-26
fix cors preflight headers bug

### LINKED ISSUE #1376: CORS Preflight request: return supported headers
#### Summary

When `bee-js` is used in the browser it can sometimes trigger [CORS Preflight request](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#preflighted_requests), where the browser sends `OPTIONS` request to Bee to verify that it will comply with Bee's CORS policy.

One reason when the Preflight check is triggered is when a "non-standard" header is attached to the request. Then the response to the Preflight request must also contain all headers that the server deems valid (in the `Access-Control-Allow-Headers` header). 

This is not happening currently and it restricts the feature set available to browser users. For example for the `/dirs` endpoint, you support `Swarm-Index-Document`, `Swarm-Error-Document` and many others which are not returned in the `Access-Control-Allow-Headers` response headers.

#### Steps to reproduce

Run Bee with `--cors-allowed-origins="*"` and

```
$ curl 'http://localhost:1633/dirs' \
  -X 'OPTIONS' \
  -H 'Connection: keep-alive' \
  -H 'Accept: */*' \
  -H 'Access-Control-Request-Method: POST' \
  -H 'Access-Control-Request-Headers: content-type,swarm-index-document' \
  -H 'Origin: http://localhost:8080' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: same-site' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Referer: http://localhost:8080/' \
  --compressed
```

#### Expected behavior

Returned headers:

```
HTTP/1.1 200 OK
Access-Control-Allow-Headers: ..., swarm-index-document, swarm-error-document, [... and all the other supported headers]
...
```

#### Actual behavior
Returned headers:

```
HTTP/1.1 200 OK
Access-Control-Allow-Headers: Origin, Accept, Authorization, Content-Type, X-Requested-With, Access-Control-Request-Headers, Access-Control-Request-Method
...
```

AuHau:
Closes #1376.

It is not a completely clean solution, as currently, it returns all Bee's custom headers across all endpoints, while a better solution would be to return only those that the endpoint indeed supports. But that would require quite a more complex support structure in the code that I was not really keen to introduce. I think this will make the work done, but it will require having it in mind when new headers are added to propagate them in here as well.

### DIFF
--- a/pkg/api/router.go
+++ b/pkg/api/router.go
@@ -162,7 +162,7 @@ func (s *server) setupRouting() {
 				if o := r.Header.Get("Origin"); o != "" && s.checkOrigin(r) {
 					w.Header().Set("Access-Control-Allow-Credentials", "true")
 					w.Header().Set("Access-Control-Allow-Origin", o)
-					w.Header().Set("Access-Control-Allow-Headers", "Origin, Accept, Authorization, Content-Type, X-Requested-With, Access-Control-Request-Headers, Access-Control-Request-Method")
+					w.Header().Set("Access-Control-Allow-Headers", "Origin, Accept, Authorization, Content-Type, X-Requested-With, Access-Control-Request-Headers, Access-Control-Request-Method, Swarm-Tag, Swarm-Pin, Swarm-Encrypt, Swarm-Index-Document, Swarm-Error-Document, Swarm-Collection")
 					w.Header().Set("Access-Control-Allow-Methods", "GET, HEAD, OPTIONS, POST, PUT, DELETE")
 					w.Header().Set("Access-Control-Max-Age", "3600")
 				}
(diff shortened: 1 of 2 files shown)


AuHau:
Not sure what to do about the failing test.

acud (approved):
nice. thanks!

acud:
@AuHau the failing tests are because this PR is from a fork, not from the ethersphere repo. Will get someone else to review this too, and we can merge it nevertheless

aloknerurkar (approved):

PR #1600, opened 2021-04-26
Slack/api metrics

aloknerurkar:
Adding counter for error codes that are returned eg:
```
# TYPE bee_api_reponse_error_codes counter
bee_api_reponse_error_codes{code="404",method="GET"} 3
```

### DIFF
--- a/pkg/api/metrics.go
+++ b/pkg/api/metrics.go
@@ -5,7 +5,9 @@
 package api
 
 import (
+	// "bufio"
 	"fmt"
+	// "net"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/api/metrics.go
```diff
@@ -38,6 +40,15 @@ func newMetrics() metrics {
 			Help:      "Histogram of API response durations.",
 			Buckets:   []float64{0.01, 0.1, 0.25, 0.5, 1, 2.5, 5, 10},
 		}),
+		ErrorPageViews: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: m.Namespace,
+				Subsystem: subsystem,
+				Name:      "reponse_error_codes",
+				Help:      "How many 5XX HTTP errors were returned",
```
esadakar, line 48: is it 4xx or 5xx

esadakar (approved):

pkg/api/metrics.go
```diff
@@ -38,6 +40,15 @@ func newMetrics() metrics {
 			Help:      "Histogram of API response durations.",
 			Buckets:   []float64{0.01, 0.1, 0.25, 0.5, 1, 2.5, 5, 10},
 		}),
+		ErrorPageViews: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: m.Namespace,
+				Subsystem: subsystem,
+				Name:      "reponse_error_codes",
```
acud, line 47: typo in `reponse`

pkg/api/metrics.go
```diff
@@ -19,6 +20,7 @@ type metrics struct {
 	RequestCount     prometheus.Counter
 	ResponseDuration prometheus.Histogram
 	PingRequestCount prometheus.Counter
+	ErrorPageViews   *prometheus.CounterVec
```
acud, line 23: I would consider having a breakdown of all status code groups - 1xx 2xx 3xx 4xx 5xx

aloknerurkar: Just mentioning our discussion here for record.

We have decided to do the grouping on graphana side instead. It seems to be better to not have the extra lookups on the client-side.

acud (approved):

acud (approved):

PR #1604, opened 2021-04-27
localstore: simplify updategc logic, put to pullsync index when stamp valid

acud:
This PR adds the chunk to the pullIndex when the stamp is valid and received from a retrieve request. It also simplifies the logic in the updateGC method.

### DIFF
--- a/pkg/localstore/mode_put_test.go
+++ b/pkg/localstore/mode_put_test.go
@@ -64,6 +64,7 @@ func TestModePutRequest(t *testing.T) {
 				}
 
 				newItemsCountTest(db.gcIndex, tc.count)(t)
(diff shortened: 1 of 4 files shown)


zelig (approved):

pkg/localstore/mode_put_test.go
```diff
@@ -394,6 +397,18 @@ func TestModePut_sameChunk(t *testing.T) {
 				{
 					name:      "ModePutRequest",
 					mode:      storage.ModePutRequest,
+					pullIndex: true,
+					pushIndex: false,
+				},
+				{
+					name:      "ModePutRequestPin",
+					mode:      storage.ModePutRequest,
```
zelig, line 405: oops this slipped through...should be ModeRequestPin

PR #1611, opened 2021-04-28
Using the correct http response codes when creating a resource

anatollupacescu:
As per #1084 we need to return http response code 201 every time we create a new resource.

### DIFF
--- a/openapi/SwarmDebug.yaml
+++ b/openapi/SwarmDebug.yaml
@@ -482,7 +482,7 @@ paths:
       tags:
         - Chequebook
       responses:
-        "200":
+        "201":
           description: OK
           content:
             application/json:
@@ -553,7 +553,7 @@ paths:
       tags:
         - Chequebook
       responses:
-        "200":
+        "201":
           description: Transaction hash of the deposit transaction
           content:
             application/json:
@@ -579,7 +579,7 @@ paths:
       tags:
         - Chequebook
       responses:
-        "200":
+        "201":
           description: Transaction hash of the withdraw transaction
           content:
             application/json:
--- a/openapi/SwarmDebug.yaml
+++ b/openapi/SwarmDebug.yaml
@@ -236,7 +236,7 @@ paths:
           required: true
           description: Underlay address of peer
       responses:
-        "200":
+        "201":
           description: Returns overlay address of connected peer
           content:
             application/json:
@@ -318,7 +318,7 @@ paths:
           required: true
           description: Swarm address of peer
       responses:
-        "200":
+        "201":
           description: Returns round trip time for given peer
           content:
             application/json:
@@ -430,7 +430,7 @@ paths:
             schema:
               $ref: "SwarmCommon.yaml#/components/schemas/WelcomeMessage"
       responses:
-        "200":
+        "201":
           description: OK
           content:
             application/json:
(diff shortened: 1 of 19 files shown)


openapi/SwarmDebug.yaml
```diff
@@ -236,7 +236,7 @@ paths:
           required: true
           description: Underlay address of peer
       responses:
-        "200":
+        "201":
```
acud, line 239: this should stay 200, since when you call the API endpoint, it returns when you connect to the peer, so it is OK rather than Created (which might infer the operation not still complete)

openapi/SwarmDebug.yaml
```diff
@@ -318,7 +318,7 @@ paths:
           required: true
           description: Swarm address of peer
       responses:
-        "200":
+        "201":
```
acud, line 321: also can stay 200

openapi/SwarmDebug.yaml
```diff
@@ -430,7 +430,7 @@ paths:
             schema:
               $ref: "SwarmCommon.yaml#/components/schemas/WelcomeMessage"
       responses:
-        "200":
+        "201":
```
acud, line 433: since this sets the message, 200 is ok, since operation is complete

openapi/SwarmDebug.yaml
```diff
@@ -482,7 +482,7 @@ paths:
       tags:
         - Chequebook
       responses:
-        "200":
+        "201":
```
acud, line 485: unsure about this. @ralph-pichler is the cashout endpoint returning once the cashout is successful?

openapi/SwarmDebug.yaml
```diff
@@ -553,7 +553,7 @@ paths:
...
```
acud, line 556: this too, @ralph-pichler

openapi/SwarmDebug.yaml
```diff
@@ -579,7 +579,7 @@ paths:
...
```
acud, line 582: this too @ralph-pichler

acud (changes requested):
a few changes should be reverted as they are atomic operations, and need input from @ralph-pichler on a few issues

openapi/SwarmDebug.yaml
ralph-pichler: it returns once the signed tx has been sent to the eth backend (but not confirmed yet).

openapi/SwarmDebug.yaml
ralph-pichler: same, returns once the tx has been sent to the backend

openapi/SwarmDebug.yaml
ralph-pichler: same again

openapi/SwarmDebug.yaml
anatollupacescu: so should we keep it a 200 then?

mrekucci (approved):
Besides the requested changes LGTM.

openapi/SwarmDebug.yaml
acud: 201 is fine in this case :+1:

acud (approved):
Sweet! thanks!

aloknerurkar (approved):

Issue #1078, opened 2020-12-29
Inconsistent naming convention in API JSON responses


agazso:
#### Summary
The naming convention in the API JSON responses are inconsistent. For example there are cases when it's lowerCamelCase

https://github.com/ethersphere/bee/blob/b44ed2eb3c60de775f063ed58d15891dccd8ce37/pkg/api/pin_chunks.go#L111

There are cases when it's lowercase

https://github.com/ethersphere/bee/blob/b44ed2eb3c60de775f063ed58d15891dccd8ce37/pkg/debugapi/chequebook.go#L40

There are cases when it's kebab-case

https://github.com/ethersphere/bee/blob/b44ed2eb3c60de775f063ed58d15891dccd8ce37/pkg/debugapi/p2p.go#L22

I think the best would be to stick to one naming convention and use that everywhere. Assuming these are all equal from the perspective of the Go code, I would suggest using lowerCamelCase because that's the easiest to use from Javascript.

Eknir:
@paxthemax , this could be something

AuHau:
Even after #1613 there are still some inconsistencies. For example https://github.com/ethersphere/bee/blob/master/openapi/SwarmCommon.yaml#L108

PR #1613, opened 2021-04-28
Using lower camel case for field names in http responses

anatollupacescu:
Using lowerCamelCase because that's the easiest to use from Javascript.


acud (approved):
great! thank you!
@agazso this would probably break a few things, but addresses the changes you've mentioned earlier about inconsistent naming

agazso:
Thanks, this looks like a fix for #1078

acud:
correct @agazso

janos (commented):
Thanks. It would be good to have these changes reflected in the openapi/*.yaml files, as well.

agazso (approved):
👍

aloknerurkar (approved):

AuHau:
FYI unfortunately this still did not address all the issues. For example https://github.com/ethersphere/bee/blob/master/openapi/SwarmCommon.yaml#L108 is still all lowercased.

PR #1614, opened 2021-04-28
Fix OpenAPI specification inconsistencies

mrekucci:
Several inconsistencies declared in the pins and stamps endpoints
of the OpenAPI specification and the bee API server implementation
are removed. On top of that also fixes several declaration errors.

### DIFF
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -19,7 +19,7 @@ servers:
         default: "localhost"
         description: Base address of the local bee node main API
       port:
-        default: 1633
+        default: "1633"
         description: Service port provided in bee node config
   - url: "http://{apiRoot}:{port}"
     variables:
@@ -27,7 +27,7 @@ servers:
         default: "localhost"
         description: Base address of the local bee node main API
       port:
-        default: 1633
+        default: "1633"
         description: Service port provided in bee node config
 
 paths:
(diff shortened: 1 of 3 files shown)


anatollupacescu (approved):

AuHau:
👍

aloknerurkar (approved):

openapi/Swarm.yaml
```diff
@@ -443,7 +443,7 @@ paths:
       - in: path
         name: address
         schema:
-          $ref: "SwarmCommon.yaml#/components/schemas/SwarmAddress"
+          $ref: "SwarmCommon.yaml#/components/schemas/SwarmOnlyReference"
```
acud, line 446: not sure how this will play out with unpinning an encrypted entry

openapi/Swarm.yaml
```diff
@@ -723,6 +727,13 @@ paths:
           description: Default response
 
   "/stamps/{id}":
+    parameters:
+      - in: path
+        name: id
+        schema:
+          $ref: "SwarmCommon.yaml#/components/schemas/SwarmAddress"
+        required: true
+        description: Swarm address of the stamp
```
acud, line 736: actually the batchID is not an address, it is just a uuid we get from the chain AFAIK. not sure exactly how it is calculated. @ralph-pichler ?

ralph-pichler: it is a 32 byte hash based on the creator address and random nonce

acud: so maybe it would be good to have this as a 32 byte hex string (I think we have this sort type in the OpenAPI spec) :+1:

AuHau: `SwarmAddress` is just that ;-)

mrekucci: Created separate type for that

acud (approved):

PR #1616, opened 2021-04-28
Rename address to reference in pinning related code

mrekucci:
This PR is a follow-up to the #1614 where the semantics of address is changed to reference for the pinning API.


acud (approved):
great stuff. thank you

zelig (approved):

PR #1617, opened 2021-04-28
libp2p static address resolver test simplification

janos:
This PR removes mockdns dependency for libp2p static address resolver by mocking the LookupIP function only.


acud (approved):
very cool :) thanks janos <3

anatollupacescu (approved):

PR #1618, opened 2021-04-28
libp2p: improve cleanup on shutdown and tests

acud:
Adds a missing close to `natManager` and test cleanups.

### DIFF
--- a/pkg/p2p/libp2p/connections_test.go
+++ b/pkg/p2p/libp2p/connections_test.go
@@ -40,8 +40,9 @@ func TestConnectDisconnect(t *testing.T) {
 	defer cancel()
 
 	s1, overlay1 := newService(t, 1, libp2pServiceOpts{})
-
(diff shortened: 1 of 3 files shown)


pkg/p2p/libp2p/connections_test.go
```diff
@@ -40,8 +40,9 @@ func TestConnectDisconnect(t *testing.T) {
 	defer cancel()
 
 	s1, overlay1 := newService(t, 1, libp2pServiceOpts{})
-
 	s2, overlay2 := newService(t, 1, libp2pServiceOpts{})
+	defer s1.Close()
```
janos, line 44: All defers that are added in this PR are not needed as they are done in newService' t.Cleanup call https://github.com/ethersphere/bee/blob/master/pkg/p2p/libp2p/libp2p_test.go#L80, especially to avoid the need for defers.

PR #1622, opened 2021-04-29
libp2p: improve cleanup on shutdown

acud:
adds nat manager cleanup to libp2p close

### DIFF
--- a/pkg/p2p/libp2p/connections_test.go
+++ b/pkg/p2p/libp2p/connections_test.go
@@ -40,8 +40,9 @@ func TestConnectDisconnect(t *testing.T) {
 	defer cancel()
 
 	s1, overlay1 := newService(t, 1, libp2pServiceOpts{})
-
 	s2, overlay2 := newService(t, 1, libp2pServiceOpts{})
+	defer s1.Close()
+	defer s2.Close()
 
 	addr := serviceUnderlayAddress(t, s1)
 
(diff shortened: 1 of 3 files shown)


anatollupacescu (approved):

pkg/p2p/libp2p/connections_test.go
```diff
@@ -346,11 +340,11 @@ func TestBlocklisting(t *testing.T) {
 	expectPeers(t, s2)
 	expectPeersEventually(t, s1)
 
-	// s2 connects to s1, thus the notifier on s1 should be called on Connect
 	_, err = s1.Connect(context.Background(), addr2)
 	if err == nil {
 		t.Fatal("expected error during connection, got nil")
 	}
+	t.Logf("s1 connect threw error: %v", err)
```
mrekucci, line 347: r: In the tests, this message looks like a false positive:

```sh
=== RUN   TestBlocklisting
    connections_test.go:347: s1 connect threw error: connect full close stream reset
--- PASS: TestBlocklisting (2.41s)
PASS
```

It is really needed here? Maybe we should rather assert this error or rewrite the message in a clearer way.

mrekucci (approved):

pkg/p2p/libp2p/connections_test.go
acud: woops. leftovers :) removing. thanks!

PR #1627, opened 2021-04-30
fix correct bigint json marshalling

### LINKED ISSUE #756: Fix big.Int (incorrectly) marshals to a number instead of a string

### LINKED ISSUE #1561: JS Number overflow for balances, settlements and generally "money"
JavaScript's `number` type is able to maximally safely represent `2^53 - 1` (see [Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)) above that floating errors steps in and we can't assume accurate representations and calculations. This is especially problematic when money comes in, which is the exact case as we are running into this limitation with all the balance, settlement, and generally money APIs. With the number of decimals that BZZ uses we are able to safely represent inside JavaScript only values lower than `0.9 BZZ`, which I believe you agree is not sufficient.

**The main problem is that the fields containing these data (balances, settlements etc.) are currently typed and returned as `number`.** Which then if somebody uses thing like `JSON.parse()` to convert the response into JS native objects hits this problem.

The common approach to solve this is usually to pass the numeric value as a string in all the response/requests and on the JS-side use libraries or natives like [`BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) that are able to correctly work with these values.

We (Bee Gees) are working on correctly handling this inside bee-js, but we believe that this should be actually tackled on the Bee level. Possibly by changing the return types of these properties to string as that would correctly communicate this expectation to other API users.

AuHau:
Closes #1561 and closes #756.

It creates `BigInt` struct that is applied only to API layer, while all internal calculations are still kept in `big.Int`.

It introduces breaking changes for several DebugAPI endpoints where amounts are converted from numbers to strings.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1627)

### DIFF
--- a/pkg/bigint/bigint.go
+++ b/pkg/bigint/bigint.go
@@ -1,0 +1,36 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package bigint
+
+import (
+	"encoding/json"
+	"fmt"
+	"math/big"
+)
+
+type BigInt struct {
+	*big.Int
+}
+
+func (i *BigInt) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf(`"%s"`, i.String())), nil
+}
+
+func (i *BigInt) UnmarshalJSON(b []byte) error {
+	var val string
+	err := json.Unmarshal(b, &val)
+	if err != nil {
+		return err
+	}
+
+	i.SetString(val, 10)
+
+	return nil
+}
+
+//Wrap wraps big.Int pointer into BigInt struct.
+func Wrap(i *big.Int) *BigInt {
+	return &BigInt{i}
+}
--- a/pkg/bigint/bigint.go
+++ b/pkg/bigint/bigint.go
@@ -25,6 +25,10 @@ func (i *BigInt) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
+	if i.Int == nil {
+		i.Int = new(big.Int)
+	}
+
 	i.SetString(val, 10)
 
 	return nil
--- a/pkg/bigint/bigint.go
+++ b/pkg/bigint/bigint.go
@@ -36,5 +36,5 @@ func (i *BigInt) UnmarshalJSON(b []byte) error {
 
 //Wrap wraps big.Int pointer into BigInt struct.
 func Wrap(i *big.Int) *BigInt {
-	return &BigInt{i}
+	return &BigInt{Int: i}
 }
--- a/pkg/bigint/bigint_test.go
+++ b/pkg/bigint/bigint_test.go
@@ -1,0 +1,28 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package bigint_test
+
+import (
+	"encoding/json"
+	"github.com/ethersphere/bee/pkg/bigint"
+	"math"
+	"math/big"
+	"reflect"
+	"testing"
+)
+
+func TestMarshaling(t *testing.T) {
+	mar, err := json.Marshal(struct {
+		Bg *bigint.BigInt
+	}{
+		Bg: bigint.Wrap(new(big.Int).Mul(big.NewInt(math.MaxInt64), big.NewInt(math.MaxInt64))),
+	})
+	if err != nil {
+		t.Errorf("Marshaling failed")
+	}
+	if !reflect.DeepEqual(mar, []byte("{\"Bg\":\"85070591730234615847396907784232501249\"}")) {
+		t.Errorf("Wrongly marshaled data")
+	}
+}
--- a/pkg/bigint/bigint_test.go
+++ b/pkg/bigint/bigint_test.go
@@ -20,7 +20,7 @@ func TestMarshaling(t *testing.T) {
 		Bg: bigint.Wrap(new(big.Int).Mul(big.NewInt(math.MaxInt64), big.NewInt(math.MaxInt64))),
 	})
 	if err != nil {
-		t.Errorf("Marshaling failed")
+		t.Errorf("Marshaling failed: %v", err)
 	}
 	if !reflect.DeepEqual(mar, []byte("{\"Bg\":\"85070591730234615847396907784232501249\"}")) {
 		t.Errorf("Wrongly marshaled data")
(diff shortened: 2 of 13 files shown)


pkg/bigint/bigint.go
```diff
@@ -0,0 +1,37 @@
+package bigint
```
acud, line 1: code copyright missing

pkg/bigint/bigint.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"math/big"
+)
+
+type BigInt struct {
+	big.Int
```
acud, line 10: should be `*big.Int`. big ints are passed around as pointers, this has all sorts of implications but your type should probably wrap the pointer instead of an instance

pkg/bigint/bigint.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func (i BigInt) MarshalJSON() ([]byte, error) {
```
acud, line 13: this should have a pointer receiver type

pkg/bigint/bigint.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return err
+	}
+
+	i.SetString(val, 10)
+
+	return nil
+}
+
+func NewBigInt(x int64) *BigInt {
```
acud, line 29: since package name is already `bigint`, no need to repeat the name again.
compare `bigint.New(123)`  with `bigint.NewBigInt(123)`

pkg/bigint/bigint_test.go
```diff
@@ -0,0 +1,24 @@
+package bigint_test
```
acud, line 1: copyright missing

pkg/bigint/bigint_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"encoding/json"
+	"github.com/ethersphere/bee/pkg/bigint"
+	"reflect"
+	"testing"
+)
+
+type TestStruct struct {
+	Bg *bigint.BigInt `json:"bg"`
```
acud, line 11: since this type is used only within the test, you could inline the struct definition directly within `Marshal`:
```go
mar, err := json.Marshal(struct{
Bg *bigint.BigInt
}{ 
Bg: bigint.New(5)
})

File: pkg/bigint/bigint_test.go [hunk 7]
```diff
@@ -0,0 +1,24 @@
+package bigint_test
+
+import (
+	"encoding/json"
+	"github.com/ethersphere/bee/pkg/bigint"
+	"reflect"
+	"testing"
+)
+
+type TestStruct struct {
```
acud, line 10: exporting the type is not needed

pkg/bigint/bigint_test.go
```diff
@@ -0,0 +1,24 @@
...
+	Bg *bigint.BigInt `json:"bg"`
+}
+
+func TestMarshaling(t *testing.T) {
+	mar, err := json.Marshal(TestStruct{Bg: bigint.NewBigInt(5)})
```
acud, line 15: i would add tests for values which are higher than `math.MaxInt64`, since we are testing `big.Int`s here

pkg/bigint/bigint.go
acud: Also, I went through the rest of the CL, and it seems that this method is only used in tests. I would therefore remove it and just initialize a `big.Int` in the tests, and just call `Wrap` in case it is needed

pkg/bigint/bigint.go
```diff
@@ -0,0 +1,37 @@
...
+	b := new(BigInt)
+	b.SetInt64(x)
+	return b
+}
+
+func Wrap(i *big.Int) *BigInt {
```
acud, line 35: exposed functions should have a go doc comment on them

pkg/debugapi/postage.go
```diff
@@ -7,9 +7,24 @@ package debugapi
 import (
 	"net/http"
 
+	"github.com/ethersphere/bee/pkg/bigint"
 	"github.com/ethersphere/bee/pkg/jsonhttp"
 )
 
+type reserveStateResponse struct {
```
acud, line 14: since we already have two types for reserve state maybe it would be better to reuse the one at `pkg/postage/reservestate.go` which was created just for the api response? or remove that one and use this one?

acud (changes requested):
Thanks for the contribution @AuHau :pray: I have some minor comments

pkg/debugapi/postage.go
AuHau: Well, that depends on how the `pkg/postage/reservestate.go` is used, which I can't really assess. I was following the style of other API endpoints that have their own response types. What I am not sure is if this `pkg/postage/reservestate.go`'s type is used also internally for something else then API return type. Because I believe I would have to change the `big.Int` into this new `bigint.BigInt` type which is not meant for internal computation. 
So what do you think?

acud (approved):

esadakar (approved):

pkg/bigint/bigint_test.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+func TestMarshaling(t *testing.T) {
+	mar, err := json.Marshal(struct {
+		Bg *bigint.BigInt
+	}{
+		Bg: bigint.Wrap(new(big.Int).Mul(big.NewInt(math.MaxInt64), big.NewInt(math.MaxInt64))),
+	})
+	if err != nil {
+		t.Errorf("Marshaling failed")
```
janos, line 23: Provide the actual error in the message.

pkg/bigint/bigint.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	i.SetString(val, 10)
+
+	return nil
+}
+
+//Wrap wraps big.Int pointer into BigInt struct.
+func Wrap(i *big.Int) *BigInt {
+	return &BigInt{i}
```
janos, line 39: It is usually safer to construct with named fields in case any future field changes.

janos (approved):
👍  Only minor comments.

acud:
since this breaks our integration tests we'll have to wait for after the release

AuHau:
Closing in favor of #1735

PR #1628, opened 2021-05-01
Quick-fix unpinning error caused by a manifest serialization bug

mrekucci:
This quick-fix prevents the mantaray manifest IterateAddresses method
walker to fail when it encounters an empty address, otherwise, the unpin
traversal operation for the manifest fails.

### DIFF
--- a/pkg/manifest/mantaray.go
+++ b/pkg/manifest/mantaray.go
@@ -134,6 +134,7 @@ func (m *mantarayManifest) IterateAddresses(ctx context.Context, fn swarm.Addres
 		return ErrMissingReference
 	}
 
+	emptyAddr := swarm.NewAddress([]byte{31: 0})
 	walker := func(path []byte, node *mantaray.Node, err error) error {
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


mrekucci:
The state returned by `DebugIndices()` before the unpinning:

```sh
map[string]int{"gcExcludeIndex":11, "gcIndex":0, "gcSize":0, "pinIndex":11, "pullIndex":11, "pushIndex":11, "retrievalAccessIndex":0, "retrievalDataIndex":11}
```

and after the unpinning:

```sh
map[string]int{"gcExcludeIndex":11, "gcIndex":0, "gcSize":0, "pinIndex":0, "pullIndex":11, "pushIndex":11, "retrievalAccessIndex":0, "retrievalDataIndex":11}
```

zelig (commented):
should this not be fixed in the walk iterator within the mantaray subpkg?

acud (approved):
LGTM Thanks!

mrekucci:
> should this not be fixed in the walk iterator within the mantaray subpkg?

From the responsibility perspective, I think it would make more sense, but this closure is passed to the mantaray subpackage functions/methods and the functions/methods are eighter calling the closure directly or passing it to another function/method so the call graph is not that simple, that's why fixing it here, relatively close to the source, is the best tradeoff in terms of clarity.

acud:
agree with @mrekucci, let's not spend too much time on this. I'd like to get this through as is. We can make things pretty when we give enough time to sort out mantaray as a whole as previously discussed @zelig

esadakar (approved):

PR #1635, opened 2021-05-03
Add payment threshold validation

anatollupacescu:
Right now it will disconnect if the condition is not met

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1635)

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -432,9 +432,9 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey, 
 
 	pricer := pricer.NewFixedPricer(swarmAddress, 1000000000)
 
-	minThreashold := pricer.MostExpensive()
+	minThreshold := pricer.MostExpensive()
 
-	pricing := pricing.New(p2ps, logger, paymentThreshold, minThreashold)
+	pricing := pricing.New(p2ps, logger, paymentThreshold, minThreshold)
 
 	if err = p2ps.AddProtocol(pricing.Protocol()); err != nil {
 		return nil, fmt.Errorf("pricing service: %w", err)
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -432,7 +432,9 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey, 
 
 	pricer := pricer.NewFixedPricer(swarmAddress, 1000000000)
 
-	pricing := pricing.New(p2ps, logger, paymentThreshold)
+	minThreashold := pricer.MostExpensive()
+
+	pricing := pricing.New(p2ps, logger, paymentThreshold, minThreashold)
 
 	if err = p2ps.AddProtocol(pricing.Protocol()); err != nil {
 		return nil, fmt.Errorf("pricing service: %w", err)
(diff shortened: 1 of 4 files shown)


ralph-pichler (commented):
might be better to pass the maximum value into `pricing.New` as it will be the same value for all peers. I would not have it depend on our threshold, instead we can use `10*swarm.MaxPO*basePrice` where `basePrice` is the last value passed into `pricer.NewFixedPricer`. This guarantees we can at least reserve the most expensive chunk 10 times.

pkg/pricer/pricer.go
```diff
@@ -39,3 +39,7 @@ func (pricer *FixedPricer) PeerPrice(peer, chunk swarm.Address) uint64 {
 func (pricer *FixedPricer) Price(chunk swarm.Address) uint64 {
 	return pricer.PeerPrice(pricer.overlay, chunk)
 }
+
+func (pricer *FixedPricer) MostExpensive() int64 {
```
anatollupacescu, line 43: return type to be changed to *big.Int

pkg/node/node.go
```diff
@@ -432,7 +432,9 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey,
 
 	pricer := pricer.NewFixedPricer(swarmAddress, 1000000000)
 
-	pricing := pricing.New(p2ps, logger, paymentThreshold)
+	minThreashold := pricer.MostExpensive()
```
metacertain, line 435: minThreshold*

pkg/pricer/pricer.go
```diff
@@ -39,3 +41,8 @@ func (pricer *FixedPricer) PeerPrice(peer, chunk swarm.Address) uint64 {
 func (pricer *FixedPricer) Price(chunk swarm.Address) uint64 {
 	return pricer.PeerPrice(pricer.overlay, chunk)
 }
+
+func (pricer *FixedPricer) MostExpensive() *big.Int {
+	mostExp := int64(10 * uint64(swarm.MaxPO) * pricer.poPrice)
```
metacertain, line 46: to avoid overflow of the multiplication, you can do the calculation like 
new(big.Int).Mul( new(big.Int).Mul(big.NewInt(10), new(big.Int).SetUint64(uint64(swarm.MaxPO)), new(big.Int).SetUint64(pricer.poPrice) )

pkg/pricing/pricing.go
```diff
@@ -44,14 +50,16 @@ type Service struct {
 	streamer                 p2p.Streamer
 	logger                   logging.Logger
 	paymentThreshold         *big.Int
+	minPaymentThreashold     *big.Int
```
metacertain, line 53: minPaymentThreshold*

metacertain (changes requested):

pkg/pricing/pricing.go
anatollupacescu: fixed

pkg/pricer/pricer.go
anatollupacescu: thanks, I'll use this one instead

metacertain (approved):

ralph-pichler (approved):

PR #1637, opened 2021-05-03
kademlia: limit broadcasting peers

esadakar:
This PR should be a good foundation for rate limiting broadcasting peers going forward. We will need to investigate if there are better strategies for the selection of peers.

### DIFF
--- a/pkg/topology/kademlia/kademlia.go
+++ b/pkg/topology/kademlia/kademlia.go
@@ -10,9 +10,12 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"math/big"
 	"math/bits"
 	"sync"
 	"time"
+
+	random "crypto/rand"
 
 	"github.com/ethersphere/bee/pkg/addressbook"
 	"github.com/ethersphere/bee/pkg/discovery"
@@ -38,6 +41,7 @@ var (
 	shortRetry                 = 30 * time.Second
 	saturationPeers            = 4
 	overSaturationPeers        = 16
+	broadcastBinSize           = 4
 )
 
 type binSaturationFunc func(bin uint8, peers, connected *pslice.PSlice) (saturated bool, oversaturated bool)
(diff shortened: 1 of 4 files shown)


pkg/topology/kademlia/kademlia.go
```diff
@@ -668,28 +671,28 @@ func (k *Kad) connect(ctx context.Context, peer swarm.Address, ma ma.Multiaddr,
...
-		// be slow and since this function is called with the same context from kademlia connect
-		// function, this might result in the unfortunate situation where we end up on
-		// `err := k.discovery.BroadcastPeers(ctx, peer, addrs...)` with an already expired context
-		// indicating falsely, that the peer connection has timed out.
-		k.wg.Add(1)
-		go func(connectedPeer swarm.Address) {
-			defer k.wg.Done()
-			if err := k.discovery.BroadcastPeers(context.Background(), connectedPeer, peer); err != nil {
-				k.logger.Debugf("could not gossip peer %s to peer %s: %v", peer, connectedPeer, err)
+		connectedPeers := randomSubset(r, k.connectedPeers.BinPeers(bin), broadcastBinSize)
```
ldeffenb, line 678: It'd be great to see the outbound connector use something like this instead of the current BinRev sorted peerID selection of outbound peer to connect to.  Especially for low-numbered, distant bins!

ldeffenb: For details of the connection behavior, see the following capture.  Look at the first peers in buckets 0, 1, and even 2 ignoring the few that were randomly added by gossip after the restart.  Code to generate your own connection peer list is included in:

https://ipfs.io/ipfs/QmSKgFpoXZoGJHo2DXhPoxikXdfZc8ye5KPg7AjnHiYbtA?filename=peer-buckets.txt

esadakar: From the perspectives of two nodes that are close and share a neighborhood, there could be performance benefits to randomizing the peers list because they chances are, they share a similar kademlia topology, hence a similar peer list in each bin.

PR #1647, opened 2021-05-03
Inception

mrekucci:


aloknerurkar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @acud, @esadakar, @mrekucci, and @zelig)

___
*[pkg/topology/kademlia/kademlia.go, line 214 at r1](https://reviewable.io/reviews/ethersphere/bee/1647#-MZwWkbS1oRZ_XuLFsyO:-MZwWkbS1oRZ_XuLFsyP:be34azy) ([raw file](https://github.com/ethersphere/bee/blob/b5aedb5bb5e19342a4dd1bde328dfa6a565cb85a/pkg/topology/kademlia/kademlia.go#L214)):*
> ```Go
> type connectFn func(context.Context, swarm.Address, uint8) bool
> 
> // peerConnInfo groups unnecessary fields needed to create a connection.
> ```

unnecessary?
___
*[pkg/topology/kademlia/kademlia.go, line 226 at r1](https://reviewable.io/reviews/ethersphere/bee/1647#-MZwXnWO1phXp7dhTpGG:-MZwXnWO1phXp7dhTpGH:b-aw0mqh) ([raw file](https://github.com/ethersphere/bee/blob/b5aedb5bb5e19342a4dd1bde328dfa6a565cb85a/pkg/topology/kademlia/kademlia.go#L226)):*
> ```Go
> 		k.waitNextMu.Lock()
> 		defer k.waitNextMu.Unlock()
> 		next, ok := k.waitNext[peer.String()]
> ```

I see two operations on this `waitNext` map that are purely read and two which are write. Maybe a RW Lock would be better?

Although I don't see that much contention, it's something to consider.

aloknerurkar:
Great work!

I really liked the switch-case error handling pattern. Its refreshing to see such explicit error handling!

aloknerurkar:
___
*[pkg/topology/kademlia/kademlia.go, line 373 at r1](https://reviewable.io/reviews/ethersphere/bee/1647#-MZwcwpx-iw_UV0AJuG8:-MZwcwpx-iw_UV0AJuG9:bju19ap) ([raw file](https://github.com/ethersphere/bee/blob/b5aedb5bb5e19342a4dd1bde328dfa6a565cb85a/pkg/topology/kademlia/kademlia.go#L373)):*
> ```Go
> 						k.logger.Debug(peer.msg)
> 						k.notifyPeerSig()
> 					}
> ```

Would it be useful to have some metrics around this area? For eg. no of failed connection attempts. Not sure if there is any plan to add metrics in this pkg. It might be helpful to understand if the `manage` routine is doing its job correctly.

aloknerurkar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: all files reviewed, 3 unresolved discussions (waiting on @acud, @esadakar, @mrekucci, and @zelig)

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: all files reviewed, 3 unresolved discussions (waiting on @acud, @aloknerurkar, @esadakar, and @zelig)

___
*[pkg/topology/kademlia/kademlia.go, line 214 at r1](https://reviewable.io/reviews/ethersphere/bee/1647#-MZwWkbS1oRZ_XuLFsyO:-MZwm1EhAkH4u5qqZoX3:b-o7vuhc) ([raw file](https://github.com/ethersphere/bee/blob/b5aedb5bb5e19342a4dd1bde328dfa6a565cb85a/pkg/topology/kademlia/kademlia.go#L214)):*

unnecessary?
</blockquote></details>

Good catch! Done.
___
*[pkg/topology/kademlia/kademlia.go, line 226 at r1](https://reviewable.io/reviews/ethersphere/bee/1647#-MZwXnWO1phXp7dhTpGG:-MZwnQo45JMcjUH301Wa:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/b5aedb5bb5e19342a4dd1bde328dfa6a565cb85a/pkg/topology/kademlia/kademlia.go#L226)):*

I see two operations on this `waitNext` map that are purely read and two which are write. Maybe a RW Lock would be better?

Although I don't see that much contention, it's something to consider.
</blockquote></details>

Done.
___
*[pkg/topology/kademlia/kademlia.go, line 373 at r1](https://reviewable.io/reviews/ethersphere/bee/1647#-MZwcwpx-iw_UV0AJuG8:-MZwncoa9x5y-4f0O1xZ:b-xsnm36) ([raw file](https://github.com/ethersphere/bee/blob/b5aedb5bb5e19342a4dd1bde328dfa6a565cb85a/pkg/topology/kademlia/kademlia.go#L373)):*

Would it be useful to have some metrics around this area? For eg. no of failed connection attempts. Not sure if there is any plan to add metrics in this pkg. It might be helpful to understand if the `manage` routine is doing its job correctly.
</blockquote></details>

Yes, it'll be addressed in #1658.

esadakar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: 1 of 2 files reviewed, 5 unresolved discussions (waiting on @acud, @aloknerurkar, @mrekucci, and @zelig)

___
*[pkg/topology/kademlia/kademlia.go, line 290 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx0SVT2RYNB29Mdwdf:-MZx0SVT2RYNB29Mdwdg:b-a44yh4) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L290)):*
> ```Go
> 
> 	for i := 0; i < numberOfWorkers; i++ {
> 		g.Go(func() error {
> ```

If I understand correctly, we are launching 1000 go routines here to listen to a channel that sends peer info from code above. A better solution may be to have a buffered channel of size 1000, and have the loop here launch as many go-routines as needed. In many cases, that number will be WAY below 1000.

The same comment applies to the loop below.  
___
*[pkg/topology/kademlia/kademlia.go, line 313 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx1ZSd7dpkke1hOYU1:-MZx1ZSd7dpkke1hOYU2:b-gq890e) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L313)):*
> ```Go
> // connectLeftovers attempts to connect to the leftovers
> // which were left by the connectBalanced method.
> func (k *Kad) connectLeftovers(ctx context.Context, connect connectFn) error {
> ```

maybe a better name might be connectKnownPeeers

esadakar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: 1 of 2 files reviewed, 6 unresolved discussions (waiting on @acud, @aloknerurkar, @mrekucci, and @zelig)

___
*[pkg/topology/kademlia/kademlia.go, line 315 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx2SeTEnjh8AHEM07F:-MZx2SeTEnjh8AHEM07G:b-s4avt3) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L315)):*
> ```Go
> func (k *Kad) connectLeftovers(ctx context.Context, connect connectFn) error {
> 	leftoverPeerChan := make(chan *peerConnInfo)
> 	g, ctx := errgroup.WithContext(ctx)
> ```

What is the behavior with errgroup if a connection attempt fails? And why is it necessary?

esadakar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: 1 of 2 files reviewed, 7 unresolved discussions (waiting on @acud, @aloknerurkar, @mrekucci, and @zelig)

___
*[pkg/topology/kademlia/kademlia.go, line 393 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx3Tg193B6r5oWV89E:-MZx3Tg193B6r5oWV89F:b-oyx7vc) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L393)):*
> ```Go
> 		bzzAddr, err := k.addressBook.Get(peer)
> 		switch {
> 		case errors.Is(err, addressbook.ErrNotFound):
> ```

Not sure if I like the switch statement error checking but that's a personal opinion :)

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: 1 of 2 files reviewed, 7 unresolved discussions (waiting on @acud, @aloknerurkar, @esadakar, and @zelig)

___
*[pkg/topology/kademlia/kademlia.go, line 290 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx0SVT2RYNB29Mdwdf:-MZx6xK_1THItz-Fgqdi:be8o409) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L290)):*
<details><summary><i>Previously, esadakar (Esad Akar) wrote…</i></summary><blockquote>

If I understand correctly, we are launching 1000 go routines here to listen to a channel that sends peer info from code above. A better solution may be to have a buffered channel of size 1000, and have the loop here launch as many go-routines as needed. In many cases, that number will be WAY below 1000.

The same comment applies to the loop below.  
</blockquote></details>

I would object to that. In our current test net, we have a node with more than 70K peers:

```json
{
	"baseAddr": "95a5f57eee862b6abbebfdaed4a...c1fefc7d6b855b2f0e3c3c693811b8c416",
	"population": 75484,
	"connected": 2482,
	"timestamp": "2021-05-05T15:19:55.079566422Z",
	"nnLowWatermark": 2,
	"depth": 15,
	"bins": "..."
}
```

The current timeout, when creating a connection, is 5 sec. With that and the unlimited upper bound for goroutines, this can diverge to uncontrolled use of resources pretty quickly. The buffered channel of size 1000 won't stop from spinning less than 1000 goroutines if the consumer is slower than the producer. Also, the idea here is to recycle the goroutines and not creating new ones for every new entry. Even though they are relatively cheap, it makes difference to e.g. create 1000 and 100K goroutines.
___
*[pkg/topology/kademlia/kademlia.go, line 313 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx1ZSd7dpkke1hOYU1:-MZxDXtb3RRcpvLC8Uij:b-g4jd4w) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L313)):*
<details><summary><i>Previously, esadakar (Esad Akar) wrote…</i></summary><blockquote>

maybe a better name might be connectKnownPeeers
</blockquote></details>

This is kind of tricky because in the `connectBalanced` method you're also connecting known peers. Here you're connecting the peers which are not balanced and so were not connected in the first phase. Maybe `connectUnbalanced` (but I find it a little bit misleading). Maybe @acud may suggest better name?
___
*[pkg/topology/kademlia/kademlia.go, line 315 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx2SeTEnjh8AHEM07F:-MZxFX5zDo26A7LNZi93:b-m2vekk) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L315)):*
<details><summary><i>Previously, esadakar (Esad Akar) wrote…</i></summary><blockquote>

What is the behavior with errgroup if a connection attempt fails? And why is it necessary? 
</blockquote></details>

If a connection attempt fails it will be just logged and the goroutine will continue. Having it in the errgroup helps us managing the group, e.g. when the context will be canceled for some reasons (the first call to return a non-nil error cancels the group). The same can be achieved with the `sync.WaitGroup` but would be a little bit complex.
___
*[pkg/topology/kademlia/kademlia.go, line 393 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx3Tg193B6r5oWV89E:-MZxIlgg5zIiIeXyQXZ7:b-pcq4ix) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L393)):*
<details><summary><i>Previously, esadakar (Esad Akar) wrote…</i></summary><blockquote>

Not sure if I like the switch statement error checking but that's a personal opinion :)
</blockquote></details>

It helps save some space, you don't need to indent for the second if statement. But agree, it's a matter of taste.

zelig (approved):

esadakar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: 1 of 2 files reviewed, 7 unresolved discussions (waiting on @acud, @aloknerurkar, @esadakar, and @mrekucci)

___
*[pkg/topology/kademlia/kademlia.go, line 290 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx0SVT2RYNB29Mdwdf:-MZyTLj37EVEzHanTZTc:bhejwup) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L290)):*

I would object to that. In our current test net, we have a node with more than 70K peers:

```json
{
	"baseAddr": "95a5f57eee862b6abbebfdaed4a...c1fefc7d6b855b2f0e3c3c693811b8c416",
	"population": 75484,
	"connected": 2482,
	"timestamp": "2021-05-05T15:19:55.079566422Z",
	"nnLowWatermark": 2,
	"depth": 15,
	"bins": "..."
}
```

The current timeout, when creating a connection, is 5 sec. With that and the unlimited upper bound for goroutines, this can diverge to uncontrolled use of resources pretty quickly. The buffered channel of size 1000 won't stop from spinning less than 1000 goroutines if the consumer is slower than the producer. Also, the idea here is to recycle the goroutines and not creating new ones for every new entry. Even though they are relatively cheap, it makes difference to e.g. create 1000 and 100K goroutines.
</blockquote></details>

The reason why I mentioned that, generally, connections attemps should be less than 1000 is that we do saturation check on each bin before we connect to peers and that connected peers count per bin should not exceed overSaturationPeers (16). I belive there is a bug in the current concurrent solution. Imagine during bootup that we have a list of known peers and zero connected peers, and when kademlia starts up, we iterate over knownPeers, sure, we do check for saturation, but since connections are now attempted in parallel, we dont wait for connections to finish before we check for saturation, and the saturation check will return false as it relies on the connectedPeers list. So, for the most part, we may end up with all the bins being oversaturated. Please do correct me if I am mistaken.

If my concern has merit, my suggestion is that we limit worker count per bin.

ldeffenb:
> If my concern has merit, my suggestion is that we limit worker count per bin.

Or track inflight connection attempts per bin and consider that also in the saturation test.  Similar to how the pusher tracks inflights to avoid pushing the same pending chunk twice.

aloknerurkar (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @acud, @aloknerurkar, @esadakar, and @mrekucci)

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @acud, @aloknerurkar, and @esadakar)

___
*[pkg/topology/kademlia/kademlia.go, line 290 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx0SVT2RYNB29Mdwdf:-M_-EM0nAN_CIowNcnZk:biq566e) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L290)):*
<details><summary><i>Previously, esadakar (Esad Akar) wrote…</i></summary><blockquote>

The reason why I mentioned that, generally, connections attemps should be less than 1000 is that we do saturation check on each bin before we connect to peers and that connected peers count per bin should not exceed overSaturationPeers (16). I belive there is a bug in the current concurrent solution. Imagine during bootup that we have a list of known peers and zero connected peers, and when kademlia starts up, we iterate over knownPeers, sure, we do check for saturation, but since connections are now attempted in parallel, we dont wait for connections to finish before we check for saturation, and the saturation check will return false as it relies on the connectedPeers list. So, for the most part, we may end up with all the bins being oversaturated. Please do correct me if I am mistaken.

If my concern has merit, my suggestion is that we limit worker count per bin.
</blockquote></details>

Your concern is right. But I still don't see how the suggestion using a buffered channel of size 1000 helps solve this issue. Can you elaborate more on that?

esadakar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @acud, @aloknerurkar, @esadakar, and @mrekucci)

___
*[pkg/topology/kademlia/kademlia.go, line 290 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx0SVT2RYNB29Mdwdf:-M_-WxjZ5QCEjix4DApV:b-4erf8f) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L290)):*

Your concern is right. But I still don't see how the suggestion using a buffered channel of size 1000 helps solve this issue. Can you elaborate more on that?
</blockquote></details>

Honestly, simply switching to a buffered channel does not solve it. We need a mechanism of limiting connection attempts per bin.  
___
*[pkg/topology/kademlia/kademlia.go, line 315 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx2SeTEnjh8AHEM07F:-M_-Xwyo4EHyeL4tad1g:bc15l59) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L315)):*

If a connection attempt fails it will be just logged and the goroutine will continue. Having it in the errgroup helps us managing the group, e.g. when the context will be canceled for some reasons (the first call to return a non-nil error cancels the group). The same can be achieved with the `sync.WaitGroup` but would be a little bit complex.
</blockquote></details>

Ah okay :thumbsup: I thought we were returning an error from the `connect` function, and I thought that was cancelling the whole thing. Thanks.

esadakar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @acud, @aloknerurkar, @esadakar, and @mrekucci)

___
*[pkg/topology/kademlia/kademlia.go, line 290 at r2](https://reviewable.io/reviews/ethersphere/bee/1647#-MZx0SVT2RYNB29Mdwdf:-M_-YQwmCkJ8qPB3OIFW:b-2vnc3g) ([raw file](https://github.com/ethersphere/bee/blob/0f569f3ac7bdea2b1d7a2b8bcd824a73185e355b/pkg/topology/kademlia/kademlia.go#L290)):*
<details><summary><i>Previously, esadakar (Esad Akar) wrote…</i></summary><blockquote>

Honestly, simply switching to a buffered channel does not solve it. We need a mechanism of limiting connection attempts per bin.  
</blockquote></details>

And I don't really know what's the best approach here at the moment.

esadakar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: all files reviewed, 1 unresolved discussion (waiting on @acud and @mrekucci)

esadakar (commented):
> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud)

esadakar (commented):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud)

acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: all files reviewed, 13 unresolved discussions (waiting on @mrekucci)

___
*[pkg/topology/kademlia/kademlia.go, line 33 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4cCkX0aKAjO1uGIQt:-M_4cCkY6GcvNtfSN7yV:b-x1q5pb) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L33)):*
> ```Go
> 	maxBootnodeAttempts    = 3    // how many attempts to dial to bootnodes before giving up
> 	defaultBitSuffixLength = 2    // the number of bits used to create pseudo addresses for balancing
> 	numberOfWorkers        = 1000 // The number of workers which will be created in order to concurrently connect to peers..
> ```

I think that if we connect to a peer per bin vertically as discussed, then `swarm.MaxBins` should suffice
___
*[pkg/topology/kademlia/kademlia.go, line 76 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4cU0N6QTk3qfYGNoR:-M_4cU0N6QTk3qfYGNoS:bxgwp6w) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L76)):*
> ```Go
> 	manageC           chan struct{}        // trigger the manage forever loop to connect to new peers
> 	waitNext          map[string]retryInfo // sanction connections to a peer, key is overlay string and value is a retry information
> 	waitNextMu        sync.RWMutex         // guards waitNext map
> ```

Not sure if this is really necessary to switch to RWMutex as this mutex is never acquired for complex operations and the code paths that acquire it are not hot at all. We try to keep RWMutexes in this package to a minimum, since they can lead to deadlocks
___
*[pkg/topology/kademlia/kademlia.go, line 217 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4fLu9E8AWmMGYRyQb:-M_4fLu9E8AWmMGYRyQc:bb06d2g) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L217)):*
> ```Go
> type peerConnInfo struct {
> 	po   uint8
> 	msg  string // Message which will be printed on successful connection.
> ```

not sure if this is really needed as this is presentation logic and the data type maintains information about a peer connection. it is fine to duplicate the logger part and be explicit so that the code is easier to read
___
*[pkg/topology/kademlia/kademlia.go, line 222 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4iSVf-1UG5mHWO8tI:-M_4iSVf-1UG5mHWO8tJ:b-pjt0qt) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L222)):*
> ```Go
> 
> // connectBalanced attempts to connect to the balanced peers first.
> func (k *Kad) connectBalanced(ctx context.Context, connect connectFn) error {
> ```

the fact that the method returns an error is a bit misleading to the reader, since one might assume it will also communicate errors from connecting to a peer. this is also communicated in the log line which is printed after this method returns:
```go
			if err := k.connectBalanced(ctx, connect); err != nil {
				k.logger.Errorf("unexpected error while connecting to peers: %v", err)
			}
```
^^ which is incorrect, since you are not connecting to peers during the execution of the method, the producer does. Also, the closure just returns a boolean and prints the error in the actual closure. That is fine, but maybe the log line when returning with an error from `connectBalanced` should be something along the lines of `unexpected error while selecting balanced peers: %v`. Also, when logging please prefix the log lines with the package name so that we see the context in the logger, since this logger does not support displaying the point in which the log line has occured.
___
*[pkg/topology/kademlia/kademlia.go, line 277 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4g18W1nPBWuycz6iM:-M_4g18W1nPBWuycz6iN:bvsev23) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L277)):*
> ```Go
> 					return ctx.Err()
> 				default:
> 					balancedPeerChan <- &peerConnInfo{
> ```

as discussed, this still does not implement a vertical connection strategy that connects to a peer per bin in a vertical column. once you communicate a peer to connect to in a bin, in the case of `connectBalanced`, you should skip to the next bin i.e. `continue` on the outer `for`
___
*[pkg/topology/kademlia/kademlia.go, line 279 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4fnbJ8J_89k4oRy1r:-M_4fnbJ8J_89k4oRy1s:b4fkwrn) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L279)):*
> ```Go
> 					balancedPeerChan <- &peerConnInfo{
> 						po:   swarm.Proximity(k.base.Bytes(), closestKnownPeer.Bytes()),
> 						msg:  fmt.Sprintf("connected to peer: %q for bin: %d", closestKnownPeer, i),
> ```

`i` should be equal to `po` if i am not mistaken, so you are in fact already maintaining this information on the data structure
___
*[pkg/topology/kademlia/kademlia.go, line 317 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4g_pj7Gax6rg2Fjwm:-M_4g_pj7Gax6rg2Fjwn:bd7uf43) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L317)):*
> ```Go
> // connectNeighbours attempts to connect to the neighbours
> // which were not considered by the connectBalanced method.
> func (k *Kad) connectNeighbours(ctx context.Context, connect connectFn) error {
> ```

both `connectNeighbours` and `connectBalanced` can benefit from the shared logic of connecting to a peer. Therefore it is not needed to create a consumer in every one of them. The consumer can be created once for the whole `manage` loop, which will just shut down on `<-quit`, and the same channel that peers are passed onto the consumer with can be just communicated both to `connectNeighbours` and `connectBalanced`
___
*[pkg/topology/kademlia/kademlia.go, line 318 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4gTrj81IqJDSAeEPY:-M_4gTrj81IqJDSAeEPZ:b5o30ru) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L318)):*
> ```Go
> // which were not considered by the connectBalanced method.
> func (k *Kad) connectNeighbours(ctx context.Context, connect connectFn) error {
> 	leftoverPeerChan := make(chan *peerConnInfo)
> ```

`neighbourPeerChan`
___
*[pkg/topology/kademlia/kademlia.go, line 365 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4j_yy306unieJV32L:-M_4j_yy306unieJV32M:bldattg) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L365)):*
> ```Go
> 	})
> 
> 	for i := 0; i < numberOfWorkers; i++ {
> ```

as mentioned, this can be spawned once per the whole `manage` loop
___
*[pkg/topology/kademlia/kademlia.go, line 417 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4hZIT-Eqyf3ggAAOd:-M_4hZIU4-CmFVFgsHr7:bhf98tr) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L417)):*
> ```Go
> 				k.logger.Debugf("could not remove peer %q from addressbook", peer)
> 			}
> 			fallthrough
> ```

if there's an overlay mismatch then we remove the peer from the known peers list, but a `delete` of a possible entry from `waitNext` is needed, so that a memory leak does not occur, since `waitNext` is never pruned
___
*[pkg/topology/kademlia/kademlia.go, line 423 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4jsCT-5S5Fa1H4TcR:-M_4jsCT-5S5Fa1H4TcS:b-834so3) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L423)):*
> ```Go
> 
> 			k.waitNextMu.Lock()
> 			if _, ok := k.waitNext[peer.String()]; !ok {
> ```

this can be removed. if `err != nil`, `k.connect` will add the entry in `waitNext`
___
*[pkg/topology/kademlia/kademlia.go, line 442 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4k8Wi3SJUsO9CfEaT:-M_4k8Wi3SJUsO9CfEaU:bg7fsfk) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L442)):*
> ```Go
> 		k.depthMu.Unlock()
> 
> 		return true
> ```

as discussed, when a peer is connected to successfully, we try to write to `manageC`:
```go
select {
case k.manageC<-struct{}{}:
default:
}
```

___
*[pkg/topology/kademlia/kademlia.go, line 844 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4keC76PJtniGE00HA:-M_4keC76PJtniGE00HB:b16smp6) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L844)):*
> ```Go
> 	closest := swarm.ZeroAddress
> 	err := peers.EachBinRev(func(peer swarm.Address, po uint8) (bool, bool, error) {
> 		if peer.IsZero() {
> ```

this change is incorrect since you are no longer checking the skipPeers list, not sure why. also, `peer` should _never_ be the zero peer

acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: all files reviewed, 14 unresolved discussions (waiting on @mrekucci)

___
*[pkg/topology/kademlia/kademlia.go, line 289 at r5](https://reviewable.io/reviews/ethersphere/bee/1647#-M_4nEhG-1ig9CJDqWjJ:-M_4nEhG-1ig9CJDqWjK:bhumpup) ([raw file](https://github.com/ethersphere/bee/blob/855fefcd4272af80b3503a248fa52492349a6ba4/pkg/topology/kademlia/kademlia.go#L289)):*
> ```Go
> 	})
> 
> 	for i := 0; i < numberOfWorkers; i++ {
> ```

a note about concurrency: you're spawning 1000 goroutines here while most of them will never be used, the same goes for the part inside `connectNeighbours`. This might lead to unnecessary stress on the garbage collector

acud (approved):
great work! thank you very much for this one :pray :lgtm:

> Reviewed 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647#-:-M_Opcb6Cq-cxpEc8hiy:b-4sh5kb)* status: all files reviewed, 5 unresolved discussions (waiting on @mrekucci)

___
*[pkg/topology/kademlia/kademlia.go, line 345 at r6](https://reviewable.io/reviews/ethersphere/bee/1647#-M_On2S8DZDGFlVh38me:-M_On2S8DZDGFlVh38mf:b514sk8) ([raw file](https://github.com/ethersphere/bee/blob/74e4a70474196eba1ea57aafe9430c98de7f7a3e/pkg/topology/kademlia/kademlia.go#L345)):*
> ```Go
> 				delete(k.waitNext, peer.addr.String())
> 				k.waitNextMu.Unlock()
> 				k.knownPeers.Remove(peer.addr, peer.po)
> ```

r: to have a `logger.Debug` line in this switch case indicating that an overlay mismatch has occurred to an overlay `peer.Addr` with underlay `bzzAddr.Underlay`. Might be handy for debugging purposes
___
*[pkg/topology/kademlia/kademlia.go, line 385 at r6](https://reviewable.io/reviews/ethersphere/bee/1647#-M_OoMHc4x1NOf_yG0R1:-M_OoMHc4x1NOf_yG0R2:b-uonc6x) ([raw file](https://github.com/ethersphere/bee/blob/74e4a70474196eba1ea57aafe9430c98de7f7a3e/pkg/topology/kademlia/kademlia.go#L385)):*
> ```Go
> 				return
> 			case peer := <-peerConnChan:
> 				if peer == nil {
> ```

r: for clarity you can have 
```
case peer, ok := <-peerConnChan:
  if !ok {return}
```
since the channel is a channel that communicates pointers, it is not obvious at first sight whether you need to explicitly communicate a `nil` value into the channel or close it (which is what actually is done), however having the `ok` in the case disambiguates this at the first sight
___
*[pkg/topology/kademlia/kademlia.go, line 437 at r6](https://reviewable.io/reviews/ethersphere/bee/1647#-M_OpblnBGenc3ZvfqsQ:-M_OpblnBGenc3ZvfqsR:bb75ed8) ([raw file](https://github.com/ethersphere/bee/blob/e0765db3f1077b067d92fc05b675801e2b394a77/pkg/topology/kademlia/kademlia.go#L437)):*
> ```
> Depth, k.NeighborhoodDepth())
> ```

acud (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @acud and @mrekucci)

___
*[pkg/topology/kademlia/kademlia.go, line 385 at r6](https://reviewable.io/reviews/ethersphere/bee/1647#-M_OoMHc4x1NOf_yG0R1:-M_Pscrh6NQysaDzxvmq:b9v3s5d) ([raw file](https://github.com/ethersphere/bee/blob/74e4a70474196eba1ea57aafe9430c98de7f7a3e/pkg/topology/kademlia/kademlia.go#L385)):*

Done.
</blockquote></details>

this is still pending I think

mrekucci (commented):
> Reviewed 1 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mrekucci)

mrekucci (commented):
> Reviewed 1 of 1 files at r9.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mrekucci)

esadakar (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1647#-:-M_Qfb0Q9UFWUh-QoGp-:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mrekucci)

Issue #1638, opened 2021-05-03
pushsync: forwarder should retry in order to get one successful send


acud:
forwarders can end up in a situation where they pick a node to forward to, but the `accounting.Reserve` call may return an error when sending. We should detect whether a node made it through the `Reserve` call _and_ the actual sending of the delivery. If that indeed has occurred, the forwarding node should not forward to another peer and wait for the response.

aloknerurkar:
Only one ongoing send operation should be in progress
Retry only on failed send (connection is dropped)

PR #1662, opened 2021-05-06
refactor(pushsync): Don't account for failed retries due to accounting

aloknerurkar:
#1638

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1662)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -36,11 +36,15 @@ const (
 )
 
 const (
-	maxPeers = 5
+	maxPeers = 3
 )
 
 var (
 	ErrOutOfDepthReplication = errors.New("replication outside of the neighborhood")
+	ErrNoPush                = errors.New("could not push chunk")
+
+	// time to wait before next retry attempt (whenever applicable)
+	preemptiveRetryTime = 5 * time.Second
 )
 
 type PushSyncer interface {
@@ -69,7 +73,7 @@ type PushSync struct {
 	isFullNode     bool
 }
 
-var timeToLive = 5 * time.Second                      // request time to live
+var timeToLive = 20 * time.Second                     // request time to live
 var timeToWaitForPushsyncToNeighbor = 3 * time.Second // time to wait to get a receipt for a chunk
 var nPeersToPushsync = 3                              // number of peers to replicate to as receipt is sent upstream
 
@@ -174,7 +178,7 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 	span, _, ctx := ps.tracer.StartSpanFromContext(ctx, "pushsync-handler", ps.logger, opentracing.Tag{Key: "address", Value: chunk.Address().String()})
 	defer span.Finish()
 
-	receipt, err := ps.pushToClosest(ctx, chunk)
+	receipt, err := ps.pushToClosest(ctx, chunk, false)
 	if err != nil {
 		if errors.Is(err, topology.ErrWantSelf) {
 			_, err = ps.storer.Put(ctx, storage.ModePutSync, chunk)
@@ -300,7 +304,7 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 // a receipt from that peer and returns error or nil based on the receiving and
 // the validity of the receipt.
 func (ps *PushSync) PushChunkToClosest(ctx context.Context, ch swarm.Chunk) (*Receipt, error) {
-	r, err := ps.pushToClosest(ctx, ch)
+	r, err := ps.pushToClosest(ctx, ch, true)
 	if err != nil {
 		return nil, err
 	}
@@ -309,41 +313,26 @@ func (ps *PushSync) PushChunkToClosest(ctx context.Context, ch swarm.Chunk) (*Re
 		Signature: r.Signature}, nil
 }
 
-func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.Receipt, reterr error) {
+func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, retryAllowed bool) (*pb.Receipt, error) {
 	span, logger, ctx := ps.tracer.StartSpanFromContext(ctx, "push-closest", ps.logger, opentracing.Tag{Key: "address", Value: ch.Address().String()})
-	defer span.Finish()
 	var (
-		skipPeers []swarm.Address
-		lastErr   error
+		skipPeers      []swarm.Address
+		allowedRetries = 1
+		c              = make(chan *pb.Receipt)
+		done           = make(chan struct{})
 	)
 
-	stamp, err := ch.Stamp().MarshalBinary()
-	if err != nil {
-		return nil, err
+	defer span.Finish()
+	if retryAllowed {
+		// only originator retries
+		allowedRetries = maxPeers
 	}
 
-	deferFuncs := make([]func(), 0)
-	defersFn := func() {
-		if len(deferFuncs) > 0 {
-			for _, deferFn := range deferFuncs {
-				deferFn()
-			}
-			deferFuncs = deferFuncs[:0]
-		}
-	}
-	defer defersFn()
+	t := time.NewTicker(preemptiveRetryTime)
 
-	for i := 0; i < maxPeers; i++ {
-		select {
-		case <-ctx.Done():
-			return nil, ctx.Err()
-		default:
-		}
+	includeSelf := ps.isFullNode
 
-		defersFn()
-
-		includeSelf := ps.isFullNode
-
+	for i := allowedRetries; i > 0; i-- {
 		// find the next closest peer
 		peer, err := ps.topologyDriver.ClosestPeer(ch.Address(), includeSelf, skipPeers...)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


aloknerurkar:
Current changes include:
1. Rebasing @acud 's changes on the latest master and fixing conflicts
2. Adding logic to discount failed retry attempts (for both originator and forwarder)
3. Adding test to ensure above behavior

pkg/pushsync/pushsync.go
```diff
@@ -352,86 +336,108 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
...
+			r, attempted, err := ps.pushPeer(ctxd, peer, ch)
+			// attempted is true if we get past accounting and actually attempt
+			// to send the request to the peer. If we dont get past accounting, we
+			// should not count the retry and try with a different peer again
+			if attempted {
+				allowedRetries--
+			}
+			if err != nil {
+				logger.Debugf("could not push to peer %s: %v", peer.String(), err)
+				resultC <- &pushResult{err: err}
```
zelig, line 354: this error does not seem to be read anywhere. 
Consider using the errgroup package.

pkg/pushsync/pushsync.go
```diff
@@ -352,86 +336,108 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
...
+	defer streamer.Close()
+
+	w, r := protobuf.NewWriterAndReader(streamer)
+	if err := w.WriteMsgWithContext(ctx, &pb.Delivery{
+		Address: ch.Address().Bytes(),
+		Data:    ch.Data(),
+		Stamp:   stamp,
+	}); err != nil {
+		_ = streamer.Reset()
+		return nil, true, fmt.Errorf("chunk %s deliver to peer %s: %w", ch.Address().String(), peer.String(), err)
```
zelig, line 407: the `String()` call is not needed on %s arguments

pkg/pushsync/pushsync.go
```diff
@@ -352,86 +336,108 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk) (rr *pb.R
...
+	err = ps.accounting.Credit(peer, receiptPrice)
+	if err != nil {
+		return nil, true, err
 	}
 
-	return nil, topology.ErrNotFound
+	return &receipt, true, nil
+}
+
+type pushResult struct {
```
zelig, line 440: if you use errgroup or separate errs channel, you wont need this struct

zelig (approved):

acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1662)* status: all files reviewed, 5 unresolved discussions (waiting on @aloknerurkar, @metacertain, and @zelig)

___
*[pkg/pushsync/pushsync.go, line 354 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_-62nkMuFoYnhTHUmk-r2-354:-M_L2MF38DVr_hEBQQck:b83jvm9) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L354)):*

this error does not seem to be read anywhere. 
Consider using the errgroup package.
</blockquote></details>

errgroup is not needed since there is no concurrency here
___
*[pkg/pushsync/pushsync.go, line 365 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_L2T6nEIHpI8144ame:-M_L2T6nEIHpI8144amf:bxtbhwo) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L365)):*
> ```Go
> 		select {
> 		case r := <-resultC:
> 			if r.receipt != nil {
> ```

check if `r.err != nil`, log the error if not and `continue` to the next retry
___
*[pkg/pushsync/pushsync.go, line 396 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_L2zlM2tAqVtyk7GSc:-M_L2zlNEYA1HiA4KnfS:biqh52k) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L396)):*
> ```Go
> 	streamer, err := ps.streamer.NewStream(ctx, peer, nil, protocolName, protocolVersion, streamName)
> 	if err != nil {
> 		return nil, true, fmt.Errorf("new stream for peer %s: %w", peer.String(), err)
> ```

this should be `nil,false, error` since there was no attempt yet (nothing has been sent)
___
*[pkg/pushsync/pushsync.go, line 440 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_-62nkMuFoYnhTHUmk-r2-440:-M_L39U52gwLt5K-fWEK:bnqsdx8) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L440)):*

if you use errgroup or separate errs channel, you wont need this struct
</blockquote></details>

errgroup won't help, since it will result in an error if one error or more has occurred. but since subsequent retries might resolve the problem, we would not like to error. i would personally refrain from adding an errgroup here. also, errgroup helps when havin concurrent requests, which is not the case here, since we wait for one request to expire before trying the next i.e. no parallel requests here

aloknerurkar:
___
*[pkg/pushsync/pushsync.go, line 396 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_L2zlM2tAqVtyk7GSc:-M_L5NEP7n2XUy_oNp4T:bjp74dd) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L396)):*

this should be `nil,false, error` since there was no attempt yet (nothing has been sent)
</blockquote></details>

If NewStream returns an error, we are not able to connect to a peer right? I was assuming this might take some time to return in which case we can call it an attempt. No?

In the subsequent change, we will filter peers which return this error. Which should mitigate the issue (by not considering these peers) in my view. If we don't consider this an attempt, we will not consider this node to be blocked for future requests and retry this.

aloknerurkar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1662)* status: 1 of 2 files reviewed, 5 unresolved discussions (waiting on @acud, @metacertain, and @zelig)

___
*[pkg/pushsync/pushsync.go, line 354 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_-62nkMuFoYnhTHUmk-r2-354:-M_LSIe5Co1_d1Kn6l8q:bcj8p4q) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L354)):*

errgroup is not needed since there is no concurrency here
</blockquote></details>

Agree with @acud We only do 1 request at a time and only if it fails we retry.
___
*[pkg/pushsync/pushsync.go, line 365 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_L2T6nEIHpI8144ame:-M_LSaFR2n-3hx72yj__:b-werm2e) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L365)):*

check if `r.err != nil`, log the error if not and `continue` to the next retry
</blockquote></details>

This error is returned from pushPeer. The error is already logged there once. 
___
*[pkg/pushsync/pushsync.go, line 407 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_-62nkMuFoYnhTHUmk-r2-407:-M_LT3Q9Aw-ZXCU6DEO4:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L407)):*

the `String()` call is not needed on %s arguments
</blockquote></details>

Done.

acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1662)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @acud, @aloknerurkar, @metacertain, and @zelig)

___
*[pkg/pushsync/pushsync.go, line 365 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_L2T6nEIHpI8144ame:-M_P5OKA4Bjdl2AQa0F9:b-manwqd) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L365)):*

This error is returned from pushPeer. The error is already logged there once. 
</blockquote></details>

OK :+1: note here that the happy path is nested, usually you would check for an error and if there's an error do something, otherwise the happy path is assumed to be at the root of the method/block
___
*[pkg/pushsync/pushsync.go, line 396 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_L2zlM2tAqVtyk7GSc:-M_P5dmw5xHXgsS3MInJ:b-cmuq9y) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L396)):*
If `NewStream` returns an error it is most probably that we are no longer connected to this peer or that some other problem may exist. It should not take too long to return here.

> If we don't consider this an attempt, we will not consider this node to be blocked for future requests and retry this.

As far as I remember it was agreed that the boolean returned from the method signifies whether there was an actual successful attempt to communicate the chunk. It should not be `true` unless the chunk actually got sent successfully, which is not the case here. In any case, a failed attempt to send to the peer (i.e. if we did not manage to get around to sending the chunk to the peer) must always result in a failed attempt logged for the trace in the LRU cache and the peer added to the skip list
___
*[pkg/pushsync/pushsync.go, line 407 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_-62nkMuFoYnhTHUmk-r2-407:-M_P6XDH4wNldTzMeIWq:b-v52yd9) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L407)):*

Done.
</blockquote></details>

that actually doesn't work with chunk addresses in all cases. a manual test to confirm would be nice

aloknerurkar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1662)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @acud, @aloknerurkar, @metacertain, and @zelig)

___
*[pkg/pushsync/pushsync.go, line 365 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_L2T6nEIHpI8144ame:-M_PQXEn7ZE-ljaJwFrx:bxrm2j3) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L365)):*

OK :+1: note here that the happy path is nested, usually you would check for an error and if there's an error do something, otherwise the happy path is assumed to be at the root of the method/block
</blockquote></details>

There will be an error check once the next PR will be merged.

The comment on the bottom was meant to clarify that we will be continuing the loop. We will revisit this once I merge the next PR and if more details are required will add.
___
*[pkg/pushsync/pushsync.go, line 407 at r2](https://reviewable.io/reviews/ethersphere/bee/1662#-M_-62nkMuFoYnhTHUmk-r2-407:-M_PQES_1vFklaNykeWt:birmx21) ([raw file](https://github.com/ethersphere/bee/blob/654c69ef5b60be90296824ee9ae768261eb4c573/pkg/pushsync/pushsync.go#L407)):*

that actually doesn't work with chunk addresses in all cases. a manual test to confirm would be nice
</blockquote></details>

`DEBU[2021-05-11T13:25:40+05:30] could not push to peer 6000000000000000000000000000000000000000000000000000000000000000: new stream for peer 6000000000000000000000000000000000000000000000000000000000000000: peer not reachable`

This does work in the test. Not sure what you mean by 'all' cases?

aloknerurkar:
Just for documenting:

We have decided to classify the NewStream error as an attempt for now. Idea is that it should not happen often and if it does, we can know using the new metrics. The metrics once run on production env, should tell us how the cache is being utilized. Once we have a better understanding, we can tweak these things in the future. Will check in as-is for now.

PR #1680, opened 2021-05-10
feat(pushsync): Add LRU cache to remember failed requests and select better peers

aloknerurkar:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1680)

### DIFF
--- a/pkg/p2p/streamtest/streamtest.go
+++ b/pkg/p2p/streamtest/streamtest.go
@@ -30,16 +30,24 @@ var (
 )
 
 type Recorder struct {
-	base        swarm.Address
-	records     map[string][]*Record
-	recordsMu   sync.Mutex
-	protocols   []p2p.ProtocolSpec
-	middlewares []p2p.HandlerMiddleware
+	base               swarm.Address
+	records            map[string][]*Record
+	recordsMu          sync.Mutex
+	protocols          []p2p.ProtocolSpec
+	middlewares        []p2p.HandlerMiddleware
+	streamErr          func(swarm.Address, string, string, string) error
+	protocolsWithPeers map[string]p2p.ProtocolSpec
 }
 
 func WithProtocols(protocols ...p2p.ProtocolSpec) Option {
 	return optionFunc(func(r *Recorder) {
 		r.protocols = append(r.protocols, protocols...)
+	})
+}
+
+func WithProtocolsAndPeers(protocolsWithPeers map[string]p2p.ProtocolSpec) Option {
+	return optionFunc(func(r *Recorder) {
+		r.protocolsWithPeers = protocolsWithPeers
 	})
 }
 
@@ -52,6 +60,12 @@ func WithMiddlewares(middlewares ...p2p.HandlerMiddleware) Option {
 func WithBaseAddr(a swarm.Address) Option {
 	return optionFunc(func(r *Recorder) {
 		r.base = a
+	})
+}
+
+func WithStreamError(streamErr func(swarm.Address, string, string, string) error) Option {
+	return optionFunc(func(r *Recorder) {
+		r.streamErr = streamErr
 	})
 }
 
@@ -73,6 +87,12 @@ func (r *Recorder) SetProtocols(protocols ...p2p.ProtocolSpec) {
 }
 
 func (r *Recorder) NewStream(ctx context.Context, addr swarm.Address, h p2p.Headers, protocolName, protocolVersion, streamName string) (p2p.Stream, error) {
+	if r.streamErr != nil {
+		err := r.streamErr(addr, protocolName, protocolVersion, streamName)
+		if err != nil {
+			return nil, err
+		}
+	}
 	recordIn := newRecord()
 	recordOut := newRecord()
 	streamOut := newStream(recordIn, recordOut)
@@ -80,14 +100,18 @@ func (r *Recorder) NewStream(ctx context.Context, addr swarm.Address, h p2p.Head
 
 	var handler p2p.HandlerFunc
 	var headler p2p.HeadlerFunc
-	for _, p := range r.protocols {
-		if p.Name == protocolName && p.Version == protocolVersion {
-			for _, s := range p.StreamSpecs {
-				if s.Name == streamName {
-					handler = s.Handler
-					headler = s.Headler
-				}
+	peerHandlers, ok := r.protocolsWithPeers[addr.String()]
+	if !ok {
+		for _, p := range r.protocols {
+			if p.Name == protocolName && p.Version == protocolVersion {
+				peerHandlers = p
 			}
+		}
+	}
+	for _, s := range peerHandlers.StreamSpecs {
+		if s.Name == streamName {
+			handler = s.Handler
+			headler = s.Headler
 		}
 	}
 	if handler == nil {
--- a/pkg/p2p/streamtest/streamtest.go
+++ b/pkg/p2p/streamtest/streamtest.go
@@ -45,7 +45,7 @@ func WithProtocols(protocols ...p2p.ProtocolSpec) Option {
 	})
 }
 
-func WithProtocolsAndPeers(protocolsWithPeers map[string]p2p.ProtocolSpec) Option {
+func WithPeerProtocols(protocolsWithPeers map[string]p2p.ProtocolSpec) Option {
 	return optionFunc(func(r *Recorder) {
 		r.protocolsWithPeers = protocolsWithPeers
 	})
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -342,9 +342,11 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, retryAllo
 		}
 		if !ps.failedRequests.Useful(peer, ch.Address()) {
 			skipPeers = append(skipPeers, peer)
+			ps.metrics.FailedCacheHits.WithLabelValues(peer.String(), ch.Address().String()).Inc()
 			continue
 		}
 		skipPeers = append(skipPeers, peer)
+		ps.metrics.TotalSendAttempts.Inc()
 
 		go func(peer swarm.Address, ch swarm.Chunk) {
 			ctxd, canceld := context.WithTimeout(ctx, defaultTTL)
@@ -376,6 +378,7 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, retryAllo
 			}
 			if r.err != nil && r.attempted {
 				ps.failedRequests.RecordFailure(peer, ch.Address())
+				ps.metrics.TotalFailedSendAttempts.Inc()
 			}
 			// proceed to retrying if applicable
 		case <-ctx.Done():
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -464,7 +464,7 @@ type failedRequestCache struct {
 
 func newFailedRequestCache() *failedRequestCache {
 	// not necessary to check error here if we use constant value
-	cache, _ := lru.New(10000)
+	cache, _ := lru.New(1000)
 	return &failedRequestCache{cache: cache}
 }
 
@@ -498,8 +498,5 @@ func (f *failedRequestCache) Useful(peer swarm.Address, chunk swarm.Address) boo
 	if !found {
 		return true
 	}
-	if val.(int) < failureThreshold {
-		return true
-	}
-	return false
+	return val.(int) < failureThreshold
 }
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -10,7 +10,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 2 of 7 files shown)


acud (changes requested):
> Reviewed 3 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1680)* status: 3 of 5 files reviewed, 5 unresolved discussions (waiting on @aloknerurkar)

___
*[pkg/pushsync/pushsync.go, line 336 at r1](https://reviewable.io/reviews/ethersphere/bee/1680#-M_LBDGU-qOrrHL6GjVZ:-M_LBDGV4VLTLNG663Rd:b-7t91em) ([raw file](https://github.com/ethersphere/bee/blob/206dede04ce9c34a1da8bbb4ff0ffb1f1a0913cd/pkg/pushsync/pushsync.go#L336)):*
> ```Go
> 
> 	for i := maxAttempts; allowedRetries > 0 && i > 0; i-- {
> 		skipCheck := false
> ```

not sure what `skipCheck` means
___
*[pkg/pushsync/pushsync.go, line 475 at r1](https://reviewable.io/reviews/ethersphere/bee/1680#-M_LACgDEEEVDV_erDiG:-M_LACgDEEEVDV_erDiH:b-dh48z4) ([raw file](https://github.com/ethersphere/bee/blob/206dede04ce9c34a1da8bbb4ff0ffb1f1a0913cd/pkg/pushsync/pushsync.go#L475)):*
> ```Go
> func newFailedRequestCache() *failedRequestCache {
> 	// not necessary to check error here if we use constant value
> 	cache, _ := lru.New(128)
> ```

i would set this much much higher. more like 10k
___
*[pkg/pushsync/pushsync.go, line 480 at r1](https://reviewable.io/reviews/ethersphere/bee/1680#-M_LAK2iAzyQuIoGGv9x:-M_LAK2j9Kp2rVltIwfY:b-stqs7d) ([raw file](https://github.com/ethersphere/bee/blob/206dede04ce9c34a1da8bbb4ff0ffb1f1a0913cd/pkg/pushsync/pushsync.go#L480)):*
> ```Go
> 
> func keyForReq(peer swarm.Address, chunk swarm.Address) string {
> 	return strings.Join([]string{peer.String(), chunk.String()}, "/")
> ```

r: using `fmt.Sprintf` will make this more readable
___
*[pkg/pushsync/pushsync.go, line 484 at r1](https://reviewable.io/reviews/ethersphere/bee/1680#-M_LAYyg4kO2rA4p1Cg0:-M_LAYyg4kO2rA4p1Cg1:b5bkq8r) ([raw file](https://github.com/ethersphere/bee/blob/206dede04ce9c34a1da8bbb4ff0ffb1f1a0913cd/pkg/pushsync/pushsync.go#L484)):*
> ```Go
> 
> func (f *failedRequestCache) RecordFailure(peer swarm.Address, chunk swarm.Address) {
> 	val, found := f.cache.Get(keyForReq(peer, chunk))
> ```

you'll need a mutex here in order to avoid data races with parallel requests
___
*[pkg/pushsync/pushsync.go, line 505 at r1](https://reviewable.io/reviews/ethersphere/bee/1680#-M_LAoFR6dKVyzalZZMe:-M_LAoFR6dKVyzalZZMf:b-vh2nar) ([raw file](https://github.com/ethersphere/bee/blob/206dede04ce9c34a1da8bbb4ff0ffb1f1a0913cd/pkg/pushsync/pushsync.go#L505)):*
> ```Go
> 		return true
> 	}
> 	return false
> ```

r: usually happy path is in the root

aloknerurkar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1680)* status: 2 of 5 files reviewed, 5 unresolved discussions (waiting on @acud)

___
*[pkg/pushsync/pushsync.go, line 336 at r1](https://reviewable.io/reviews/ethersphere/bee/1680#-M_LBDGU-qOrrHL6GjVZ:-M_LRBH1E6jNR_sfb4C9:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/206dede04ce9c34a1da8bbb4ff0ffb1f1a0913cd/pkg/pushsync/pushsync.go#L336)):*

not sure what `skipCheck` means
</blockquote></details>

Done.
___
*[pkg/pushsync/pushsync.go, line 475 at r1](https://reviewable.io/reviews/ethersphere/bee/1680#-M_LACgDEEEVDV_erDiG:-M_LRHDn8KfLNreQfNux:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/206dede04ce9c34a1da8bbb4ff0ffb1f1a0913cd/pkg/pushsync/pushsync.go#L475)):*

i would set this much much higher. more like 10k
</blockquote></details>

Done.
___
*[pkg/pushsync/pushsync.go, line 480 at r1](https://reviewable.io/reviews/ethersphere/bee/1680#-M_LAK2iAzyQuIoGGv9x:-M_LRVv-AWYmamJSBGNI:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/206dede04ce9c34a1da8bbb4ff0ffb1f1a0913cd/pkg/pushsync/pushsync.go#L480)):*

r: using `fmt.Sprintf` will make this more readable
</blockquote></details>

Done.
___
*[pkg/pushsync/pushsync.go, line 484 at r1](https://reviewable.io/reviews/ethersphere/bee/1680#-M_LAYyg4kO2rA4p1Cg0:-M_LRIqc-a4u6jRXMkRa:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/206dede04ce9c34a1da8bbb4ff0ffb1f1a0913cd/pkg/pushsync/pushsync.go#L484)):*

you'll need a mutex here in order to avoid data races with parallel requests
</blockquote></details>

Done.
___
*[pkg/pushsync/pushsync.go, line 505 at r1](https://reviewable.io/reviews/ethersphere/bee/1680#-M_LAoFR6dKVyzalZZMe:-M_LRfFYDlVT_GDfXJzv:b-605n1i) ([raw file](https://github.com/ethersphere/bee/blob/206dede04ce9c34a1da8bbb4ff0ffb1f1a0913cd/pkg/pushsync/pushsync.go#L505)):*

r: usually happy path is in the root
</blockquote></details>

I did not understand this. The happy path here is 'true' as then we can go ahead with the request and prevent unnecessary check with threshold

pkg/pushsync/pushsync.go
```diff
@@ -438,6 +450,56 @@ func (ps *PushSync) pushPeer(ctx context.Context, peer swarm.Address, ch swarm.C
...
+	f.mtx.RLock()
+	val, found := f.cache.Get(keyForReq(peer, chunk))
+	f.mtx.RUnlock()
+	if !found {
+		return true
+	}
+	if val.(int) < failureThreshold {
+		return true
+	}
+	return false
```
janos, line 504:
```suggestion
	return val.(int) < failureThreshold
```

pkg/p2p/streamtest/streamtest.go
```diff
@@ -43,6 +45,12 @@ func WithProtocols(protocols ...p2p.ProtocolSpec) Option {
 	})
 }
 
+func WithProtocolsAndPeers(protocolsWithPeers map[string]p2p.ProtocolSpec) Option {
```
janos, line 48: This is a new functionality and it would be good to have tests for it.

pkg/p2p/streamtest/streamtest.go
```diff
@@ -55,6 +63,12 @@ func WithBaseAddr(a swarm.Address) Option {
 	})
 }
 
+func WithStreamError(streamErr func(swarm.Address, string, string, string) error) Option {
```
janos, line 66: This is a new functionality and it would be good to have tests for it.

zelig (commented):
> Reviewed 1 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1680)* status: 3 of 5 files reviewed, 11 unresolved discussions (waiting on @acud and @aloknerurkar)

___
*[pkg/pushsync/pushsync.go, line 501 at r2](https://reviewable.io/reviews/ethersphere/bee/1680#-M_Oa63W0DaLaBHhzW11:-M_Oa63X4E-FxxUGJnyX:b7ula30) ([raw file](https://github.com/ethersphere/bee/blob/499c1c02d2630f5542a8cd21b1ca8985d909bd47/pkg/pushsync/pushsync.go#L501)):*
> ```Go
> 		return true
> 	}
> 	if val.(int) < failureThreshold {
> ```

we could use a timestamp here and  expire 
___
*[pkg/pushsync/pushsync_test.go, line 870 at r2](https://reviewable.io/reviews/ethersphere/bee/1680#-M_OaxK240RK0x__xlQz:-M_OaxK240RK0x__xlR-:b3coj07) ([raw file](https://github.com/ethersphere/bee/blob/499c1c02d2630f5542a8cd21b1ca8985d909bd47/pkg/pushsync/pushsync_test.go#L870)):*
> ```Go
> 		}
> 
> 		cache.RecordFailure(peer, chunk)
> ```

check after 1,2 attempts also
___
*[pkg/pushsync/pushsync_test.go, line 899 at r2](https://reviewable.io/reviews/ethersphere/bee/1680#-M_ObILh57-bW8cvlYQA:-M_ObILh57-bW8cvlYQB:b-90vbmd) ([raw file](https://github.com/ethersphere/bee/blob/499c1c02d2630f5542a8cd21b1ca8985d909bd47/pkg/pushsync/pushsync_test.go#L899)):*
> ```Go
> 
> 		if !cache.Useful(peer, chunk) {
> 			t.Fatal("incorrect cache state after first failure")
> ```

error message correct?

acud (approved):
@janos bear in mind that since the cache entries keys are complied from the peer address _and_ the chunk address, they just guarantee that the same trace will not be tried subsequently after a number of attempts have been made. About cache size - we will have to tweak this but the problem is that if there are a large number of chunks being pushed, when the capacity is reached quickly the failed traces which are still not fully synced will be forgotten, resulting in new retries for the failed trace which might exceed what we've wanted to try for this peer and chunk combination.
@aloknerurkar I think maybe a thousand entries would suffice as a starting point for the LRU capacity, and I would also add some metrics to see how many times we actually hit the different cases here - how many times we add/delete from the map as well as how many success, failure and useful/unuseful events we get. This will allow us to tweak the LRU size to the correct size, for example: if many results add to the map and all calls to `Useful` result in `true` while the number of failures is large, it means that the cache size is potentially too small, otherwise you would expect to see more unuseful peers

:lgtm: minus the pending comments from @janos 

> Reviewed 1 of 5 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1680#-:-M_OzsNE2ndj0cC4ggYJ:b-62br8p)* status: all files reviewed, 7 unresolved discussions (waiting on @aloknerurkar)

___
*[pkg/pushsync/pushsync_test.go, line 435 at r2](https://reviewable.io/reviews/ethersphere/bee/1680#-M_MVxD0ASi0dRZV92WW:-M_MVxD1DdQf_2lwXQbu:b-atqpx4) ([raw file](https://github.com/ethersphere/bee/blob/499c1c02d2630f5542a8cd21b1ca8985d909bd47/pkg/pushsync/pushsync_test.go#L435)):*
> ```Go
> 					if fail {
> 						fail = false
> 						defer stream.Close()
> ```

it is not needed to use `defer` here

janos (approved):
@acud I am completely aware what is the cache key, of course. That is why I am talking about a chunk in my comment. I was describing a potential problem when no peers would be tried for a chunk indefinitely, even there may be a chance that some peer will be able to send it, maybe it is completely irrelevant. My suggestion is to have expiration on the failures with reached threshold. But I get from you comment that this is completely undesirable. All good.

acud:
So I had a chat about this with @aloknerurkar. The thing is that we will have as a result of the current strategy forwarder nodes that will try different traces, and with subsequent improvements to how we actually do the handling of the chunk in the neighborhood, the next closest peer will eventually conclude that it is the closest (since the closest was tried several times and did not give back an answer), therefore the next closest will conclude that it is the closest and assume the closest peer's duties (NN spreading and sending back the receipt). So for now ttl is implied by the size of the cache, and the problem in the failed trace should be resolved within NN's next closest peer to resolve the situation, however I think this should be done in subsequent PRs.

I would try to introduce as less moving parts on this cache at this point, since introducing TTLs already has a lot of inferred meanings in terms of how often should the originator try and what would be a reasonable TTL per request. Keeping this cache as dumb as possible would help us tweak those numbers without really having to deal with the cache policy, since the TTLs could then be tuned to get the best performance in a dynamic environment.

ldeffenb:
I can confirm that commit 0.6.0-0b0bb9e-dirty seems to be using this new LRU cache to prevent repeated pushes. However, the bee_pushsync_failed_cache_hits seem to keep incrementing after a chunk has failed enough times, even though no --verbosity 5 logs indicate that the chunk is still trying to push.  The bee_pusher_total_to_push metric is also continuing to increase even though no --verbosity 5 logs indicate any pusher activity.

You can see my logs for this activity at:
https://pastebee.com/?b20ea4728cd82dbf89ef5f555f4e263c46855960141f81c3d49d2de6db22b9a6

I have the entire --verbosity 5 log for this node (which is still apparently trying to push that chunk) available. Just ask me for bee6-5.out for my 0.6.0 node.

aloknerurkar:
We did not add an error trace in the skipping logic. Maybe we can add some warning trace? @acud

PR #1683, opened 2021-05-10
perf(pullstorage): share common iterators in pullstorage

acud:
This PR optimizes the way we open pull sync subscription to localstore in `pullstorage` package.

The assumptions are the following:
- when starting to sync, different peers will do individual, one-time-cost historical syncing which is relevant only to them
- when doing live-syncing, most peers will want the same intervals, since most of them are assumed to be tracking the tip of the binIDs they are interested in

So for example, two peers which are live-syncing from a third peer will both want anything from a given interval to `math.MaxUint64`. Since both peers are expected to track the tip of the bin more or less, and since some debouncing is introduced by the `flipflop` package, it is expected that in most cases, the `From` value will be shared between peers around the same timeframe.

Through the pull sync protocol handler, these independent requests will yield to two identical calls to `IntervalChunks` in `makeOffer`:
```go
// makeOffer tries to assemble an offer for a given requested interval.
func (s *Syncer) makeOffer(ctx context.Context, rn pb.GetRange) (o *pb.Offer, addrs []swarm.Address, err error) {
 -->	chs, top, err := s.storage.IntervalChunks(ctx, uint8(rn.Bin), rn.From, rn.To, maxPage)
	if err != nil {
		return o, nil, err
	}
```

Once the notification about new chunks in a bin comes through in `localstore`, these identical iterators will be triggered, resulting in duplicate work and possible CPU overhead syncing open iterators. 
The aim of this PR is to reuse existing, open subscriptions and leverage the same response to multiple requesters at the same time, lowering the amount of I/O and CPU we're doing significantly, moving the bottleneck to (hopefully) the transport level. 

This is also crucial to leverage due to:
1. Recent changes to kademlia upper bound which is set by the postage contract data. This might under certain circumstances result in very large neighborhoods 
2. The change of `swarm.MaxPO` to `32` instead of `16`. This may result in a lot more open subscriptions from peers on a lot more bins which would amplify the problem significantly


janos (approved):
Very nice! While reading, I got an impression that singleflight could have been used for the same purpose, but the implementation in this PR is fine.

zelig (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1683)* status: all files reviewed, 1 unresolved discussion (waiting on @acud, @aloknerurkar, @anatollupacescu, @esadakar, and @mrekucci)

___
*[pkg/pullsync/pullstorage/pullstorage.go, line 109 at r1](https://reviewable.io/reviews/ethersphere/bee/1683#-M_OdhAB-onNyxdBAbfV:-M_OdhAB-onNyxdBAbfW:b-y61x2k) ([raw file](https://github.com/ethersphere/bee/blob/19be3700085b0a5e6325c5eabc5e862f8ad2bb47/pkg/pullsync/pullstorage/pullstorage.go#L109)):*
> ```Go
> 		s.openSubsMu.Lock()
> 		for _, c := range s.openSubs[k] {
> 			c <- intervalChunks{chs: chs, topmost: topmost, err: err}
> ```

there must be a select default here or the channel be buffered, otherwise this leaks

acud (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1683)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @aloknerurkar, @anatollupacescu, @esadakar, @mrekucci, and @zelig)

___
*[pkg/pullsync/pullstorage/pullstorage.go, line 109 at r1](https://reviewable.io/reviews/ethersphere/bee/1683#-M_OdhAB-onNyxdBAbfV:-M_Ol_1-6YiKWG_3ZOec:b-5m9zpc) ([raw file](https://github.com/ethersphere/bee/blob/19be3700085b0a5e6325c5eabc5e862f8ad2bb47/pkg/pullsync/pullstorage/pullstorage.go#L109)):*

there must be a select default here or the channel be buffered, otherwise this leaks
</blockquote></details>

actually it will deadlock completely. I added a select default case here and some comments about why this is needed :+1:

aloknerurkar (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1683#-:-M_PVn7ZB0OVOPcgztuq:bnfp4nl)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @anatollupacescu, @esadakar, @mrekucci, and @zelig)

mrekucci (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1683#-:-M_R0s1nBIe4RjFlccP-:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @acud, @anatollupacescu, @esadakar, and @zelig)

PR #1693, opened 2021-05-12
fix(libp2p): remove disconnect call with empty overlay

esadakar:
When overlay address does not exist, no call to Disconnect should be made inside AddProtocol stream handler.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1693)

### DIFF
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -379,7 +379,6 @@ func (s *Service) AddProtocol(p p2p.ProtocolSpec) (err error) {
 			peerID := streamlibp2p.Conn().RemotePeer()
 			overlay, found := s.peers.overlay(peerID)
 			if !found {
-				_ = s.Disconnect(overlay)
 				_ = streamlibp2p.Reset()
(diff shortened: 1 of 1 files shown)


acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1693)* status: all files reviewed, 1 unresolved discussion (waiting on @esadakar)

___
*[pkg/p2p/libp2p/libp2p.go, line 382 at r1](https://reviewable.io/reviews/ethersphere/bee/1693#-M_VAb6S-dMkFqILwNlC:-M_VAb6T8sStZ61cVtqT:b-yfqp3e) ([raw file](https://github.com/ethersphere/bee/blob/0b0bb9e84ba3609b86f962157793f3f194a60ce3/pkg/p2p/libp2p/libp2p.go#L382)):*
> ```Go
> 			overlay, found := s.peers.overlay(peerID)
> 			if !found {
> 				_ = s.Disconnect(overlay)
> ```

but then, considering how the `Disconnect` method works, wouldn't we need to also move the part that checks whether the peer is found or not?
```go
	if !found {
		return p2p.ErrPeerNotFound
	}
```

PR #1697, opened 2021-05-12
feat: add block-time cli flag

acud:
Improve how many calls to the ethereum backend we're doing

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1697)

### DIFF
--- a/pkg/postage/listener/listener.go
+++ b/pkg/postage/listener/listener.go
@@ -25,9 +25,8 @@ import (
 )
 
 const (
-	blockPage = 10000 // how many blocks to sync every time
-	tailSize  = 4     // how many blocks to tail from the tip of the chain
-	pollEvery = tailSize + 1
+	blockPage = 5000 // how many blocks to sync every time we page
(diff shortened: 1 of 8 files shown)


pkg/postage/listener/listener.go
```diff
@@ -149,6 +152,8 @@ func (l *listener) Listen(from uint64, updater postage.EventUpdater) <-chan stru
 		cancel()
 	}()
 
+	chainUpdateInterval := time.Duration(l.blockTime) * pollEvery * time.Second
```
ralph-pichler, line 155: i don't think this formula makes sense. why would the update interval increase if tailsize goes up? it should either be `time.Duration(l.blockTime)` or `time.Duration(l.blockTime) / constantFactor`.

ralph-pichler (changes requested):

ralph-pichler (commented):
also use blocktime here instead of `pollingInterval `: https://github.com/ethersphere/bee/blob/master/pkg/node/chain.go#L57

vandot (approved):

PR #1706, opened 2021-05-14
fix: postage header for cors preflight

AuHau:
This is follow up PR after #1597 that adds the new Postage stamp batch id header for CORS Preflight

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1706)


acud (approved):
Nice, thanks!

esadakar (approved):

PR #1735, opened 2021-05-17
fix: marshal bigint as string

### LINKED ISSUE #756: Fix big.Int (incorrectly) marshals to a number instead of a string

### LINKED ISSUE #1561: JS Number overflow for balances, settlements and generally "money"

AuHau:
Closes #1561 and closes #756.

It creates BigInt struct that is applied only to API layer, while all internal calculations are still kept in big.Int.

It introduces breaking changes for several DebugAPI endpoints where amounts are converted from numbers to strings.

Succeeds #1627 

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1735)

### DIFF
--- a/.github/workflows/beekeeper.yml
+++ b/.github/workflows/beekeeper.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Prepare local cluster
         run: |
           printf ${{ secrets.CR_PAT }} | docker login ghcr.io -u bee-worker --password-stdin
-          make beekeeper BEEKEEPER_INSTALL_DIR=/usr/local/bin BEEKEEPER_USE_SUDO=true BEEKEEPER_BRANCH=bigint-serialization
+          make beekeeper BEEKEEPER_INSTALL_DIR=/usr/local/bin BEEKEEPER_USE_SUDO=true
           timeout 10m make beelocal OPTS='ci skip-vet'
       - name: Set kube config
         run: |
@@ -55,7 +55,7 @@ jobs:
       - name: Set local cluster
         run: |
           make beelocal ACTION=add-hosts
-          timeout 10m make deploylocal BEEKEEPER_CLUSTER=local-dns BEEKEEPER_BRANCH=bigint-serialization
+          timeout 10m make deploylocal BEEKEEPER_CLUSTER=local-dns
       - name: Test pingpong
         id: pingpong-1
         run: until beekeeper check --cluster-name local-dns --checks ci-pingpong; do echo "waiting for pingpong..."; sleep .3; done
--- a/.github/workflows/beekeeper.yml
+++ b/.github/workflows/beekeeper.yml
@@ -14,7 +14,6 @@ jobs:
       REPLICA: 3
       RUN_TYPE: "PR RUN"
       SETUP_CONTRACT_IMAGE_TAG: "0.1.0"
-      BEEKEEPER_BRANCH: "bigint-serialization"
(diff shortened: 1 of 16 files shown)


ralph-pichler:
this will most likely require changes to the beekeeper tests

AuHau:
Yeah, you are right. I will try to make those adjustments, but not sure if I have sufficient knowledge for this 😅

AuHau:
I tried to put something here https://github.com/ethersphere/beekeeper/pull/148 but not sure if it is working as I can't run it locally unfortunately.

zelig (commented):
> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1735)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @AuHau)

acud (approved):

.github/workflows/beekeeper.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       - name: Prepare local cluster
         run: |
           printf ${{ secrets.CR_PAT }} | docker login ghcr.io -u bee-worker --password-stdin
-          make beekeeper BEEKEEPER_INSTALL_DIR=/usr/local/bin BEEKEEPER_USE_SUDO=true
+          make beekeeper BEEKEEPER_INSTALL_DIR=/usr/local/bin BEEKEEPER_USE_SUDO=true BEEKEEPER_BRANCH=bigint-serialization
```
esadakar, line 49: Will this stay as `BEEKEEPER_BRANCH=bigint-serialization` or just something temporary?

esadakar (approved):

.github/workflows/beekeeper.yml
acud: this needs to be reverted

AuHau: done

PR #1752, opened 2021-05-18
refactor(openapi): swap EmptyNull with nullable

mrekucci:
Replaces user-defined type EmptyNull with the
nullable defined by the OpenAPI specification.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1752)

### DIFF
--- a/openapi/SwarmCommon.yaml
+++ b/openapi/SwarmCommon.yaml
@@ -93,11 +93,10 @@ components:
       type: object
       properties:
         lastcheques:
-          oneOf:
-            - $ref: '#/components/schemas/EmptyNull'
-            - type: array
-              items:
-                $ref: "#/components/schemas/ChequePeerResponse"
+          type: array
+          nullable: true
+          items:
+            $ref: "#/components/schemas/ChequePeerResponse"
 
     ChequePeerResponse:
       type: object
@@ -139,10 +138,6 @@ components:
       pattern: "^[A-Fa-f0-9]{40}$"
       example: "36b7efd913ca4cf880b8eeac5093fa27b0825906"
       
-    EmptyNull:
-      type: 'null'
-      description: When no entries 'null' is returned instead of empty array.
-
     FileName:
       type: string
 
@@ -214,11 +209,10 @@ components:
       type: object
       properties:
         tags:
-          oneOf:
-            - $ref: '#/components/schemas/EmptyNull'
-            - type: array
-              items:
-                $ref: "#/components/schemas/NewTagResponse"
+          type: array
+          nullable: true
+          items:
+            $ref: "#/components/schemas/NewTagResponse"
 
     P2PUnderlay:
       type: string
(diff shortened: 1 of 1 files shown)


acud (approved):

anatollupacescu (approved):

AuHau:
Hmm, honestly I am not such a fan of this approach. This does not explain at all when the property is `null` at the behavior (empty array = null) is not obvious (at least for non-Golang dev). If you want to go this path I would suggest adding into all descriptions of these properties info about when it is null.

mrekucci:
> Hmm, honestly I am not such a fan of this approach. This does not explain at all when the property is `null` and the behavior (empty array = null) is not obvious (at least for non-Golang dev). If you want to go this path I would suggest adding into all descriptions of these properties info about when it is null.

I'd disagree here since the specification is clear on that what the [`nullable`](https://github.com/OAI/OpenAPI-Specification/blob/main/proposals/003_Clarify-Nullable.md) express and the consumer should be aware of this specification. On top of that, the current description using the `EmptyNull` is not valid according to the https://editor.swagger.io/ it returns the following error:

```
Structural error at components.schemas.EmptyNull.type
should be equal to one of the allowed values
allowedValues: array, boolean, integer, number, object, string
Jump to line 143
```

AuHau:
> I'd disagree here since the specification is clear on that what the nullable express and the consumer should be aware of this specification.

I am not saying that the user should not understand what `nullable` means. Sure, for some generic objects that have an attribute with a potential `null` value that in its context makes sense, why not. But coming from JS (and few other langs) backgrounds, I would 100% expect you to return an empty array instead of `null` for the case when there are no entries. Few others team members from Bee Geezs (JS) were also quite surprised by this.

> On top of that, the current description using the EmptyNull is not valid according to the editor.swagger.io it returns the following error:

Heh, my IDE then confused me about allowed values 😅 Then I agree with dropping my approach but, as mentioned before, I would add a textual description that "Returns `null` if there are no entries instead of empty array".

mrekucci:
> > I'd disagree here since the specification is clear on that what the nullable express and the consumer should be aware of this specification.
> 
> I am not saying that the user should not understand what `nullable` means. Sure, for some generic objects that have an attribute with a potential `null` value that in its context makes sense, why not. But coming from JS (and few other langs) backgrounds, I would 100% expect you to return an empty array instead of `null` for the case when there are no entries. Few others team members from Bee Geezs (JS) were also quite surprised by this.
> 
> > On top of that, the current description using the EmptyNull is not valid according to the editor.swagger.io it returns the following error:
> 
> Heh, my IDE then confused me about allowed values 😅 Then I agree with dropping my approach but, as mentioned before, I would add a textual description that "Returns `null` if there are no entries instead of empty array".

I agree with returning an empty array instead of a null. That's best practice when designing an API. This PR doesn't address that, it just fixes the errors pointed out by the validator. Hopefully, the fix replacing the null values with the empty array will come soon, but some things need to be discussed before we make that change.

AuHau:
Yeah agree, I am only trying to be more "dummy developer-friendly" and propose to explain when there can be null value as I don't think it is obvious...

PR #1753, opened 2021-05-18
fix: still update balance on blocklisting

ralph-pichler:
* still update balance after time settle enforcement check even in case of blocklisting. this should avoid permanent imbalances arising related to enforcement issues.
* reduce the blocklisting time disconnect thresholds

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1753)


acud (approved):

metacertain (approved):
24 hour disconnection included in order to make future fixes for blocklisting behaviour take effect in a shorter while for already blocklisted peers

PR #1780, opened 2021-05-19
feat: Retrial behavior in retrieval

metacertain:
This pr attempts to mirror some of the push sync behaviour changes introduced in #1662 and to enable retrying until a chunk is attempted to be retrieved from at least one peer

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1780)

### DIFF
--- a/pkg/retrieval/retrieval.go
+++ b/pkg/retrieval/retrieval.go
@@ -130,7 +130,6 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin 
 		)
 
 		requestAttempt := 0
-		emptyRounds := 0
 
 		lastTime := time.Now().Unix()
 
@@ -159,7 +158,6 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin 
 				if res.retrieved {
 					if res.err != nil {
 						if !res.peer.IsZero() {
-							emptyRounds++
 							logger.Debugf("retrieval: failed to get chunk %s from peer %s: %v", addr, res.peer, res.err)
 						}
 						peersResults++
@@ -200,10 +198,6 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin 
 				}
 			}
 
-			if emptyRounds >= maxSelects {
-				requestAttempt++
-				emptyRounds = 0
-			}
 		}
 
 		// if we have not managed to get results after 5 rounds of peer selections, give up
--- a/pkg/retrieval/retrieval.go
+++ b/pkg/retrieval/retrieval.go
@@ -130,15 +130,15 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin 
 		)
 
 		requestAttempt := 0
+		emptyRounds := 0
+
 		lastTime := time.Now().Unix()
 
 		for requestAttempt < 5 {
 
 			if peerAttempt < maxSelects {
 				peerAttempt++
-
 				s.metrics.PeerRequestCounter.Inc()
-
 				go func() {
 					chunk, peer, requested, err := s.retrieveChunk(ctx, addr, sp)
 					resultC <- retrievalResult{
@@ -149,7 +149,7 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin 
 					}
 				}()
 			} else {
-				ticker.Stop()
+				resultC <- retrievalResult{}
 			}
 
 			select {
@@ -159,6 +159,7 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin 
 				if res.retrieved {
 					if res.err != nil {
 						if !res.peer.IsZero() {
+							emptyRounds++
 							logger.Debugf("retrieval: failed to get chunk %s from peer %s: %v", addr, res.peer, res.err)
 						}
 						peersResults++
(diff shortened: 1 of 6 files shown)


acud (approved):

janos (approved):
👍

aloknerurkar (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1780#-:-Ma8At04EZYMjoTdmrpC:bg7igk9)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @metacertain)

pkg/retrieval/retrieval.go
```diff
@@ -87,16 +94,21 @@ func (s *Service) Protocol() p2p.ProtocolSpec {
...
 	retrieveChunkTimeout = 10 * time.Second
 
 	retrieveRetryIntervalDuration = 5 * time.Second
 )
 
-func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address) (swarm.Chunk, error) {
+func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin bool) (swarm.Chunk, error) {
 	s.metrics.RequestCounter.Inc()
 
 	v, err, _ := s.singleflight.Do(addr.String(), func() (interface{}, error) {
```
ralph-pichler, line 105: if a chunk is already retrieved with origin false, does that singleflight here not imply that we would never try with `maxPeers=8`  if a origin true call happens at the same time?

janos: Good point, the origin flag should be part of the singleflight key.

Eknir:
Let's merge this after the upcoming release (0.6.2)

aloknerurkar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1780)* status: 2 of 6 files reviewed, 2 unresolved discussions (waiting on @aloknerurkar, @metacertain, and @ralph-pichler)

___
*[pkg/retrieval/retrieval.go, line 111 at r2](https://reviewable.io/reviews/ethersphere/bee/1780#-MbkDatnBQOcRUcfmebh:-MbkDatnBQOcRUcfmebi:b-jf9h42) ([raw file](https://github.com/ethersphere/bee/blob/150d81fe6e744435cde95c1e9b1de056fb377300/pkg/retrieval/retrieval.go#L111)):*
> ```Go
> 	}
> 
> 	v, err, _ := s.singleflight.Do(flightRoute, func() (interface{}, error) {
> ```

Problem with this singleFlight pattern is that the parent context cancellation does not cancel the singleFlight requests which are being shared, resulting in goroutine leaks. This was inspected during #1891 and we decided to use @janos 's lib which adds context semantics correctly to singleFlight. I would recommend using that here.
___
*[pkg/retrieval/retrieval.go, line 187 at r2](https://reviewable.io/reviews/ethersphere/bee/1780#-MbkTKA7A7RfHymB8cEP:-MbkTKA7A7RfHymB8cEQ:b-bs64ve) ([raw file](https://github.com/ethersphere/bee/blob/150d81fe6e744435cde95c1e9b1de056fb377300/pkg/retrieval/retrieval.go#L187)):*
> ```Go
> 				requestAttempt++
> 				timeNow := time.Now().Unix()
> 				if timeNow > lastTime {
> ```

When will this not be true?

metacertain:
> When will this not be true?

if we are in the same second still

pkg/retrieval/retrieval.go
```diff
@@ -108,51 +126,83 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address) (swarm.
...
-				s.metrics.PeerRequestCounter.Inc()
+		lastTime := time.Now().Unix()
 
-				go func() {
-					chunk, peer, err := s.retrieveChunk(ctx, addr, sp)
-					if err != nil {
-						if !peer.IsZero() {
-							logger.Debugf("retrieval: failed to get chunk %s from peer %s: %v", addr, peer, err)
-						}
+		for requestAttempt < 5 {
```
esadakar, line 133: maybe we should stay consistent and have this as a variable at the top of the file

esadakar (approved):

pkg/retrieval/retrieval.go
```diff
@@ -108,51 +123,87 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address) (swarm.
...
+
+			// if we have not counted enough successful attempts but out of selection amount, reset
+			if peerAttempt >= maxSelects {
+				if !origin {
+					return nil, storage.ErrNotFound
+				}
+
+				requestAttempt++
+				timeNow := time.Now().Unix()
+				if timeNow > lastTime {
```
esadakar, line 188: This would always be true, no?

metacertain: if we are in the same second still
we wait 600 miliseconds only so we should get into the next second within 2 rounds of requestAttempts

aloknerurkar (approved):
> Reviewed 1 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1780)* status: 3 of 6 files reviewed, 2 unresolved discussions (waiting on @aloknerurkar, @esadakar, and @ralph-pichler)

aloknerurkar:
___
*[pkg/retrieval/retrieval.go, line 188 at r3](https://reviewable.io/reviews/ethersphere/bee/1780#-Ma3_AGpKm9GMH1j3wbx-r3-188:-MboI4J_7dteCs7cEpea:b-dg1pmv) ([raw file](https://github.com/ethersphere/bee/blob/bd0e3fcf4b7bd73a6492f009b89927484bcb9dd6/pkg/retrieval/retrieval.go#L188)):*
<details><summary><i>Previously, metacertain wrote…</i></summary><blockquote>

if we are in the same second still
we wait 600 miliseconds only so we should get into the next second within 2 rounds of requestAttempts
</blockquote></details>

It would be good if you can add a comment here explaining this.

pkg/retrieval/retrieval.go
esadakar: Ok, can we add a test case that covers the area starting at `requestAttempt++` and down.

esadakar: I leave the judgement to you if this is necessary

PR #1793, opened 2021-05-20
feat: transact batchstore

acud:
This PR adds a change to prevent inconsistencies when syncing the postage data from the chain.

We've found that during shutdown, the block height might be persisted before an event (or a bunch of events) are processed. This sends the node into an inconsistent state since the next time it will boot up, syncing chain data will commence from `<previous persisted block>+1`.

Due to this, we've introduced a stopgap solution that detects whether a shutdown happened while updating a batch.
Once a dirty shutdown is detected on boot up sequence, the batch store keys are swiped and the node starts to resync the chain data from the start block number.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1793)

### DIFF
--- a/pkg/localstore/reserve.go
+++ b/pkg/localstore/reserve.go
@@ -40,13 +40,20 @@ func (db *DB) UnreserveBatch(id []byte, radius uint8) error {
 	oldRadius := i.Radius
 	var gcSizeChange int64 // number to add or subtract from gcSize
 	unpin := func(item shed.Item) (stop bool, err error) {
-		c, err := db.setUnpin(batch, swarm.NewAddress(item.Address))
+		addr := swarm.NewAddress(item.Address)
+		c, err := db.setUnpin(batch, addr)
 		if err != nil {
-			return false, fmt.Errorf("unpin: %w", err)
+			if !errors.Is(err, leveldb.ErrNotFound) {
+				return false, fmt.Errorf("unpin: %w", err)
+			} else {
+				// this is possible when we are resyncing chain data after
+				// a dirty shutdown
+				db.logger.Tracef("unreserve set unpin chunk %s: %v", addr.String(), err)
+			}
 		}
 
 		gcSizeChange += c
-		return false, err
+		return false, nil
 	}
 
 	// iterate over chunk in bins
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -362,7 +362,7 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey, 
 		eventListener := listener.New(logger, swapBackend, postageContractAddress, o.BlockTime)
 		b.listenerCloser = eventListener
 
-		batchSvc = batchservice.New(batchStore, logger, eventListener)
+		batchSvc = batchservice.New(stateStore, batchStore, logger, eventListener)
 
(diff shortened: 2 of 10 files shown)


anatollupacescu (changes requested):
> Reviewed 9 of 9 files at r1, 2 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1793)* status: all files reviewed, 2 unresolved discussions (waiting on @acud)

___
*[pkg/postage/batchservice/batchservice.go, line 129 at r1](https://reviewable.io/reviews/ethersphere/bee/1793#-Ma8Zbo57XRZOD5FC1na:-Ma8Zbo57XRZOD5FC1nb:bciowcr) ([raw file](https://github.com/ethersphere/bee/blob/9f5c8303e30fce5048ab54605683bcb902d005bc/pkg/postage/batchservice/batchservice.go#L129)):*
> ```
> 	if err != nil {
> 		if !errors.Is(err, storage.ErrNotFound) {
> ```

these can be inlined to reduce nesting
___
*[pkg/postage/batchservice/batchservice.go, line 132 at r1](https://reviewable.io/reviews/ethersphere/bee/1793#-Ma8Zq_vC4_7NGOeYmEx:-Ma8Z0sfAoSitT-Dx59B:b-5gywc6) ([raw file](https://github.com/ethersphere/bee/blob/9f5c8303e30fce5048ab54605683bcb902d005bc/pkg/postage/batchservice/batchservice.go#L132)):*
> ```Go
> 			return nil, err
> 		}
> 	} else {
> ```

I think it will be more readable without this `else` since in the condition above we return.

acud (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1793)* status: all files reviewed, 2 unresolved discussions (waiting on @anatollupacescu)

___
*[pkg/postage/batchservice/batchservice.go, line 129 at r1](https://reviewable.io/reviews/ethersphere/bee/1793#-Ma8Zbo57XRZOD5FC1na:-Ma8aDhIDy7cpIbdj7GS:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/9f5c8303e30fce5048ab54605683bcb902d005bc/pkg/postage/batchservice/batchservice.go#L129)):*
<details><summary><i>Previously, anatollupacescu (Anatolie Lupacescu) wrote…</i></summary><blockquote>

> ```
> 	if err != nil {
> 		if !errors.Is(err, storage.ErrNotFound) {
> ```

these can be inlined to reduce nesting
</blockquote></details>

Done.
___
*[pkg/postage/batchservice/batchservice.go, line 132 at r1](https://reviewable.io/reviews/ethersphere/bee/1793#-Ma8Zq_vC4_7NGOeYmEx:-Ma8aE9E-054sBVhljek:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/9f5c8303e30fce5048ab54605683bcb902d005bc/pkg/postage/batchservice/batchservice.go#L132)):*
<details><summary><i>Previously, anatollupacescu (Anatolie Lupacescu) wrote…</i></summary><blockquote>

I think it will be more readable without this `else` since in the condition above we return.
</blockquote></details>

Done.

ralph-pichler (approved):

esadakar (approved):

Issue #1556, opened 2021-04-09
`TestSendChunkToSyncWithTag` data race


acud:
```

ok  	github.com/ethersphere/bee/pkg/pullsync/pullstorage	11.736s
?   	github.com/ethersphere/bee/pkg/pullsync/pullstorage/mock	[no test files]
=== RUN   TestSendChunkToSyncWithTag
    pusher_test.go:115: Chunk not synced
==================
WARNING: DATA RACE
Write at 0x00c000034f00 by goroutine 34:
  internal/race.Write()
      /opt/hostedtoolcache/go/1.15.11/x64/src/internal/race/race.go:41 +0x125
  sync.(*WaitGroup).Wait()
      /opt/hostedtoolcache/go/1.15.11/x64/src/sync/waitgroup.go:128 +0x126
  github.com/ethersphere/bee/pkg/localstore.(*DB).Close.func1()
      /home/runner/work/bee/bee/pkg/localstore/localstore.go:451 +0x5a

Previous read at 0x00c000034f00 by goroutine 28:
  internal/race.Read()
      /opt/hostedtoolcache/go/1.15.11/x64/src/internal/race/race.go:37 +0x206
  sync.(*WaitGroup).Add()
      /opt/hostedtoolcache/go/1.15.11/x64/src/sync/waitgroup.go:71 +0x219
  github.com/ethersphere/bee/pkg/localstore.(*DB).SubscribePush()
      /home/runner/work/bee/bee/pkg/localstore/subscription_push.go:49 +0x2c6
  github.com/ethersphere/bee/pkg/pusher_test.(*Store).SubscribePush()
      <autogenerated>:1 +0x7a
  github.com/ethersphere/bee/pkg/pusher.(*Service).chunksWorker()
      /home/runner/work/bee/bee/pkg/pusher/pusher.go:224 +0xad8

Goroutine 34 (running) created at:
  github.com/ethersphere/bee/pkg/localstore.(*DB).Close()
      /home/runner/work/bee/bee/pkg/localstore/localstore.go:449 +0xc4
  github.com/ethersphere/bee/pkg/pusher_test.(*Store).Close()
      /home/runner/work/bee/bee/pkg/pusher/pusher_test.go:66 +0xf7
  runtime.call32()
      /opt/hostedtoolcache/go/1.15.11/x64/src/runtime/asm_amd64.s:540 +0x3d
  testing.(*common).Fatal()
[...]

PR #1800, opened 2021-05-20
fix(pusher): flaky test

janos:
This PR addresses two related issues with TestSendChunkToPushSyncWithoutTag reported in issue https://github.com/ethersphere/bee/issues/1556.

The first one is the test failure `pusher_test.go:166: Chunk not synced` which is due to too short time waiting for chunk to be synced in the retry loop. It is addressed by raising the value to 50ms from 10ms for every retry iteration. It is possible that with race detector scheduler is not able to execute all required goroutines for test to pass even with retries. This was validate locally and validated that it does not happen with this change by running the test tens of thousand times.

The second issue is with the Closing sequence when the failure above occurs. It is incorrect to have the pusher Close only at the end of the function as it will not be called on the failure (additional side effect is a memory leak). Beside that store Close is called in the defer function, which will be called even if the failure occurs. This creates a possible situation where the pusher is still calling SubscribePush and localstore's s.subscritionsWG.Add(1) after the test finishes with a failure, and the localstore's Close is calling db.subscritionsWG.Wait() in test's defer. This results with a Panic reported in the issue. This PR sets up the proper resource cleanup and termination.

Additionally, the closing sequence is changed in two other tests in the same file.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1800)

### DIFF
--- a/pkg/pusher/pusher_test.go
+++ b/pkg/pusher/pusher_test.go
@@ -89,6 +89,7 @@ func TestSendChunkToSyncWithTag(t *testing.T) {
 
 	mtags, p, storer := createPusher(t, triggerPeer, pushSyncService, mock.WithClosestPeer(closestPeer))
 	defer storer.Close()
+	defer p.Close()
 
 	ta, err := mtags.Create(1)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


esadakar:
Great! I was looking into this today and had the same fix :) Thanks for the PR.

esadakar (approved):

aloknerurkar (approved):

acud (approved):

PR #1811, opened 2021-05-21
pullstorage: fix goroutine leak

acud:
Fixes the goroutine leak in pullstorage. This is due to context cancellation of the main goroutine that is actually supposed to communicate the results to others. Results in the channels never being written to and contexts never being terminated.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1811)

### DIFF
--- a/pkg/puller/puller.go
+++ b/pkg/puller/puller.go
@@ -324,7 +324,6 @@ func (p *Puller) histSyncWorker(ctx context.Context, peer swarm.Address, bin uin
 			}
 			if ruid == 0 {
 				p.metrics.HistWorkerErrCounter.Inc()
-				return
 			}
 			if err := p.syncer.CancelRuid(ctx, peer, ruid); err != nil && logMore {
 				p.logger.Debugf("histSyncWorker cancel ruid: %v", err)
@@ -368,7 +367,6 @@ func (p *Puller) liveSyncWorker(ctx context.Context, peer swarm.Address, bin uin
 			}
 			if ruid == 0 {
 				p.metrics.LiveWorkerErrCounter.Inc()
-				return
 			}
 			if err := p.syncer.CancelRuid(ctx, peer, ruid); err != nil && logMore {
 				p.logger.Debugf("histSyncWorker cancel ruid: %v", err)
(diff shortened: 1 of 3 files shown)


janos (approved):
Looks good for a hotfix.

aloknerurkar:
___
*[pkg/pullsync/pullstorage/pullstorage.go, line 174 at r1](https://reviewable.io/reviews/ethersphere/bee/1811#-MaDzIBNEglmSDDrQbaL:-MaDzIBOD2L7KiO8tfj3:bpuil2) ([raw file](https://github.com/ethersphere/bee/blob/78c8376f4cf7f724b5f2b62a487c12d906516581/pkg/pullsync/pullstorage/pullstorage.go#L174)):*
> ```Go
> 			return chs, topmost, nil
> 		}
> 		_, _, _ = fn(context.Background())
> ```

Why are you using background ctx here?

If the parent context was canceled, SubscribePull should return early right. So if it takes time in SubscribePull, passing the same context here is better no?

anatollupacescu (approved):

acud (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1811)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @aloknerurkar)

___
*[pkg/pullsync/pullstorage/pullstorage.go, line 174 at r1](https://reviewable.io/reviews/ethersphere/bee/1811#-MaDzIBNEglmSDDrQbaL:-MaE5omgE9WQQJsv5xGB:babtcgz) ([raw file](https://github.com/ethersphere/bee/blob/78c8376f4cf7f724b5f2b62a487c12d906516581/pkg/pullsync/pullstorage/pullstorage.go#L174)):*

Why are you using background ctx here?

If the parent context was canceled, SubscribePull should return early right. So if it takes time in SubscribePull, passing the same context here is better no?
</blockquote></details>

no it should not. the problem is that because the singleflight affects other goroutines, one context cancellation can affect the results that other goroutines are seeing, this is why the leak appeared in the first place (the main goroutine that was supposed to hand out the results to the other ones disappeared, causing the others to dangle for ever)

aloknerurkar (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1811)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @acud)

janos (approved):

PR #1824, opened 2021-05-21
perf(pslice): optimize Add method

janos:
This PR addresses a performance problem found in profiling when adding new peer address to PSlice with a larger number of peers in it. This is manifested as a very high node start time with a large number of overlay addresses in the addressbook, directly impacting bootnode's performance.

Two improvements combined give significantly better results in newley added BenchmarkAdd test:

- optimized insertAddresses based on https://github.com/golang/go/wiki/SliceTricks#insertvector
- making the copied peers slice capacity larger by one to avoid extending the slice

The improvements are in execution time and also memory and allocations which is reducing the pressure on the gc.

The results are:

```
goos: darwin
goarch: amd64
pkg: github.com/ethersphere/bee/pkg/topology/pslice
cpu: Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz

// Code state on the mster branch (baseline)
BenchmarkAdd
BenchmarkAdd-8   	    9228	    381933 ns/op	  689114 B/op	       5 allocs/op

// Only using insertAddresses instead append function (one optimization)
BenchmarkAdd
BenchmarkAdd-8   	   10000	    278262 ns/op	  478792 B/op	       4 allocs/op

// insertAddresses and peerExtraCap in the copy method (both optimizations)
BenchmarkAdd
BenchmarkAdd-8   	   10000	    171411 ns/op	  239487 B/op	       3 allocs/op
```

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1824)

### DIFF
--- a/pkg/topology/pslice/pslice.go
+++ b/pkg/topology/pslice/pslice.go
@@ -173,12 +173,9 @@ func (s *PSlice) Add(addr swarm.Address, po uint8) {
 		return
 	}
 
-	peers, bins := s.copy()
+	peers, bins := s.copy(1)
 
-	head := peers[:s.bins[po]]
-	tail := append([]swarm.Address{addr}, peers[s.bins[po]:]...)
-
-	peers = append(head, tail...)
+	peers = insertAddresses(peers, int(s.bins[po]), addr)
 	s.peers = peers
 
 	incDeeper(bins, po)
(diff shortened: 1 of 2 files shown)


acud (approved):

acud:
@janos great! thank you :pray:

PR #1657, opened 2021-05-05
go 1.16

esadakar:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1657)

PR #1855, opened 2021-05-24
Revert "go: update to 1.16 (#1657)"

esadakar:
This reverts commit 17f7837b9eb1e535739958f9534614e17e275039.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1855)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/ethersphere/bee
 
-go 1.16
+go 1.15
 
 require (
 	github.com/btcsuite/btcd v0.21.0-beta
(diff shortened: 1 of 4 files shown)


aloknerurkar (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/ethersphere/bee
 
-go 1.16
+go 1.14
```
anatollupacescu, line 3: I know it's a revert but I wonder if we should stay on 1.15 since our workflows use this ver?

anatollupacescu (approved):

go.mod
esadakar: yeah it wouldn't hurt

PR #1856, opened 2021-05-24
feat: remove chunks that fail to sync after some time

esadakar:
After the first failed attempt at pushing a chunk, chunks will now have a 5 minute window to be uploaded successfully. Also, the PR replaces `inflight` map with `singleflight` package.

The PR is not a permanent solution. Future work is needed for a more proper solution. For example, instead of generic errors, we may only decide to remove chunks from storage based on specific errors from pushsync protocol. 

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1856)

### DIFF
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -27,6 +27,7 @@ import (
 	"github.com/opentracing/opentracing-go"
 
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/singleflight"
 )
 
 type Service struct {
@@ -39,10 +40,13 @@ type Service struct {
 	metrics           metrics
 	quit              chan struct{}
 	chunksWorkerQuitC chan struct{}
+	retryMap          map[string]time.Time
+	retryMux          sync.Mutex
 }
 
 var (
 	retryInterval  = 5 * time.Second // time interval between retries
+	expiresAfter   = 5 * time.Minute // time interval between retries
 	concurrentJobs = 10              // how many chunks to push simultaneously
 )
 
@@ -59,6 +63,7 @@ func New(networkID uint64, storer storage.Storer, peerSuggester topology.Closest
 		metrics:           newMetrics(),
 		quit:              make(chan struct{}),
 		chunksWorkerQuitC: make(chan struct{}),
+		retryMap:          make(map[string]time.Time),
 	}
 	go service.chunksWorker()
 	return service
@@ -75,10 +80,9 @@ func (s *Service) chunksWorker() {
 		cctx, cancel  = context.WithCancel(context.Background())
 		ctx           = cctx
 		sem           = make(chan struct{}, concurrentJobs)
-		inflight      = make(map[string]struct{})
-		mtx           sync.Mutex
 		span          opentracing.Span
 		logger        *logrus.Entry
+		requestGroup  singleflight.Group
 	)
 	defer timer.Stop()
 	defer close(s.chunksWorkerQuitC)
@@ -88,6 +92,7 @@ func (s *Service) chunksWorker() {
 	}()
 
 LOOP:
+
 	for {
 		select {
 		// handle incoming chunks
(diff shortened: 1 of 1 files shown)


aloknerurkar (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1856)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @acud, @anatollupacescu, @esadakar, and @mrekucci)

___
*[pkg/pusher/pusher.go, line 43 at r1](https://reviewable.io/reviews/ethersphere/bee/1856#-MaSiUAVBVim-6kG7h5Y:-MaSiUAVBVim-6kG7h5Z:b-bnbdw4) ([raw file](https://github.com/ethersphere/bee/blob/c0a54eec03322d869e347ab77ec56f8b0d5d03a3/pkg/pusher/pusher.go#L43)):*
> ```Go
> 	quit              chan struct{}
> 	chunksWorkerQuitC chan struct{}
> 	retryMap          map[string]time.Time
> ```

I would consider using sync.Map here. Lots of goroutines will be in contention for the lock here. sync.Map is much more optimized for this although it does not provide type safety. I don't see it being used in a lot of places in our code but this is a critical path for performance.

If not, maybe RWLock.
___
*[pkg/pusher/pusher.go, line 171 at r1](https://reviewable.io/reviews/ethersphere/bee/1856#-MaSnpar134gQq2U6F7I:-MaSnpar134gQq2U6F7J:bqi00qi) ([raw file](https://github.com/ethersphere/bee/blob/c0a54eec03322d869e347ab77ec56f8b0d5d03a3/pkg/pusher/pusher.go#L171)):*
> ```Go
> 							} else {
> 								clean()
> 								_ = s.storer.Set(ctx, storage.ModeSetRemove, ch.Address())
> ```

This will delete the chunk from localstore completely?

Not sure if this is desired. If we upload some file and we are offline for some reason and we don't manage to sync, the file will be deleted. As opposed to retrying automatically after connecting back?

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1856)* status: all files reviewed, 2 unresolved discussions (waiting on @acud, @anatollupacescu, and @esadakar)

___
*[pkg/pusher/pusher.go, line 49 at r1](https://reviewable.io/reviews/ethersphere/bee/1856#-MaSqNHM6yiWnBgxQGUl:-MaSqNHM6yiWnBgxQGUm:btrzpmc) ([raw file](https://github.com/ethersphere/bee/blob/c0a54eec03322d869e347ab77ec56f8b0d5d03a3/pkg/pusher/pusher.go#L49)):*
> ```Go
> var (
> 	retryInterval  = 5 * time.Second // time interval between retries
> 	expiresAfter   = 5 * time.Minute // time interval between retries
> ```

I'd rename `expiresAfter` to `retryExpireInterval` or `retryTTL`; also the description doesn't seem to fit.
___
*[pkg/pusher/pusher.go, line 287 at r1](https://reviewable.io/reviews/ethersphere/bee/1856#-MaSuCH9221rFJREr7QN:-MaSuCHA6jWcO9LM2X6F:b-76mu2g) ([raw file](https://github.com/ethersphere/bee/blob/c0a54eec03322d869e347ab77ec56f8b0d5d03a3/pkg/pusher/pusher.go#L287)):*
> ```Go
> 	defer s.retryMux.Unlock()
> 
> 	if expiration, ok := s.retryMap[addr.String()]; !ok {
> ```

Please use `ByteString()` for the map key, it's more efficient.

anatollupacescu (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1856)* status: all files reviewed, 2 unresolved discussions (waiting on @acud, @aloknerurkar, and @esadakar)

___
*[pkg/pusher/pusher.go, line 43 at r1](https://reviewable.io/reviews/ethersphere/bee/1856#-MaSiUAVBVim-6kG7h5Y:-MaTNQ0v2nZkSGlJR1xy:b-waq5vx) ([raw file](https://github.com/ethersphere/bee/blob/c0a54eec03322d869e347ab77ec56f8b0d5d03a3/pkg/pusher/pusher.go#L43)):*

I would consider using sync.Map here. Lots of goroutines will be in contention for the lock here. sync.Map is much more optimized for this although it does not provide type safety. I don't see it being used in a lot of places in our code but this is a critical path for performance.

If not, maybe RWLock.
</blockquote></details>

A mutex is already being used (defined bellow) but I agree we don't use `sync.Map` often in our codebase.
In this particular case - because in the `retry` func we do a read and then a write op - and we want both to happen atomically - I'd keep the mutex, at least for now.

esadakar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1856)* status: all files reviewed, 2 unresolved discussions (waiting on @acud and @aloknerurkar)

___
*[pkg/pusher/pusher.go, line 43 at r1](https://reviewable.io/reviews/ethersphere/bee/1856#-MaSiUAVBVim-6kG7h5Y:-MaTVtyF-v1HzqSPR-um:ba6ssqf) ([raw file](https://github.com/ethersphere/bee/blob/c0a54eec03322d869e347ab77ec56f8b0d5d03a3/pkg/pusher/pusher.go#L43)):*
<details><summary><i>Previously, anatollupacescu (Anatolie Lupacescu) wrote…</i></summary><blockquote>

A mutex is already being used (defined bellow) but I agree we don't use `sync.Map` often in our codebase.
In this particular case - because in the `retry` func we do a read and then a write op - and we want both to happen atomically - I'd keep the mutex, at least for now.
</blockquote></details>

Here's a nice article. Sync.maps are meant for "high-read scenarios with relatively few updates to the map or where many updates may occur in bursty scenario". This is not a good fit. In pusher, we are doing 10 concurrent tasks, and only the hit the retryMap if there is a failure. 
___
*[pkg/pusher/pusher.go, line 171 at r1](https://reviewable.io/reviews/ethersphere/bee/1856#-MaSnpar134gQq2U6F7I:-MaTYUA9BZT3yhnJ3inr:bk2fcwu) ([raw file](https://github.com/ethersphere/bee/blob/c0a54eec03322d869e347ab77ec56f8b0d5d03a3/pkg/pusher/pusher.go#L171)):*

This will delete the chunk from localstore completely?

Not sure if this is desired. If we upload some file and we are offline for some reason and we don't manage to sync, the file will be deleted. As opposed to retrying automatically after connecting back?
</blockquote></details>

What needs to be resolved is removal of chunks that are impossible to be uploaded, eg: chunks uploaded with expired postage batches. So for these cases, retrying at another time will simply not work. I think we need specific errors returned from pushsync protocol and decide if we should remove the chunk permanently based on that.

esadakar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1856)* status: all files reviewed, 2 unresolved discussions (waiting on @acud and @aloknerurkar)

___
*[pkg/pusher/pusher.go, line 43 at r1](https://reviewable.io/reviews/ethersphere/bee/1856#-MaSiUAVBVim-6kG7h5Y:-MaTZM9-CKsTCRsaLnT2:b-hq061y) ([raw file](https://github.com/ethersphere/bee/blob/c0a54eec03322d869e347ab77ec56f8b0d5d03a3/pkg/pusher/pusher.go#L43)):*
<details><summary><i>Previously, esadakar (Esad Akar) wrote…</i></summary><blockquote>

Here's a nice article. Sync.maps are meant for "high-read scenarios with relatively few updates to the map or where many updates may occur in bursty scenario". This is not a good fit. In pusher, we are doing 10 concurrent tasks, and only the hit the retryMap if there is a failure. 
</blockquote></details>

https://medium.com/@deckarep/the-new-kid-in-town-gos-sync-map-de24a6bf7c2c

aloknerurkar:
___
*[pkg/pusher/pusher.go, line 43 at r1](https://reviewable.io/reviews/ethersphere/bee/1856#-MaSiUAVBVim-6kG7h5Y:-MaTuHiD7TuQZWc7Ybm2:b-4zpv15) ([raw file](https://github.com/ethersphere/bee/blob/c0a54eec03322d869e347ab77ec56f8b0d5d03a3/pkg/pusher/pusher.go#L43)):*
> https://medium.com/@deckarep/the-new-kid-in-town-gos-sync-map-de24a6bf7c2c
https://golang.org/src/sync/map.go

The comment on L22.

Look at case (2). We have a similar use case, right? The read/updates are being made to disjoint keys (as chunks will most probably have different addresses)

As I understand, this should improve the case if our keys are different.

@anatollupacescu 
In the retry function, the write op is returned as a function call which the caller can execute at a later time. This is not atomic.

aloknerurkar:
___
*[pkg/pusher/pusher.go, line 171 at r1](https://reviewable.io/reviews/ethersphere/bee/1856#-MaSnpar134gQq2U6F7I:-MaTvZK9-uc4Gkv3XES5:b-gn18y6) ([raw file](https://github.com/ethersphere/bee/blob/c0a54eec03322d869e347ab77ec56f8b0d5d03a3/pkg/pusher/pusher.go#L171)):*
<details><summary><i>Previously, esadakar (Esad Akar) wrote…</i></summary><blockquote>

What needs to be resolved is removal of chunks that are impossible to be uploaded, eg: chunks uploaded with expired postage batches. So for these cases, retrying at another time will simply not work. I think we need specific errors returned from pushsync protocol and decide if we should remove the chunk permanently based on that.  
</blockquote></details>

But so till then, I think we should not be deleting the data right?

Not sure how much tolerance our users would have if their data is deleted locally as well after they upload.

zelig (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1856)* status: all files reviewed, 3 unresolved discussions (waiting on @acud, @aloknerurkar, @anatollupacescu, and @esadakar)

___
*[pkg/pusher/pusher.go, line 169 at r1](https://reviewable.io/reviews/ethersphere/bee/1856#-MaXShjd6SSYaehA9hh8:-MaXShjd6SSYaehA9hh9:b-7uh4lu) ([raw file](https://github.com/ethersphere/bee/blob/c0a54eec03322d869e347ab77ec56f8b0d5d03a3/pkg/pusher/pusher.go#L169)):*
> ```Go
> 							if retry, clean := s.retry(ch.Address()); retry {
> 								return
> 							} else {
> ```

no else needed here
___
*[pkg/pusher/pusher.go, line 171 at r1](https://reviewable.io/reviews/ethersphere/bee/1856#-MaSnpar134gQq2U6F7I:-MaZ9cBeEvgiidUo0mBg:b-h9h11a) ([raw file](https://github.com/ethersphere/bee/blob/c0a54eec03322d869e347ab77ec56f8b0d5d03a3/pkg/pusher/pusher.go#L171)):*

But so till then, I think we should not be deleting the data right?

Not sure how much tolerance our users would have if their data is deleted locally as well after they upload.
</blockquote></details>

yes this does not look right, you should never remove a chunk like this. Instead you can call a`ModeSetSync`, although that may not be semantically right. An alteranative is to require revalidation on retry and simply not allow this situation to persist long

PR #1862, opened 2021-05-24
fix: reduce bootnodes peers count

esadakar:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1862)

### DIFF
--- a/pkg/topology/kademlia/kademlia.go
+++ b/pkg/topology/kademlia/kademlia.go
@@ -40,7 +40,7 @@ const (
 var (
 	saturationPeers             = 8
 	overSaturationPeers         = 16
-	bootnodeOverSaturationPeers = 64
+	bootnodeOverSaturationPeers = 20
 	shortRetry                  = 30 * time.Second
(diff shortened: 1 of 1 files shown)


aloknerurkar (approved):

pkg/topology/kademlia/kademlia.go
```diff
@@ -40,7 +40,7 @@ const (
 var (
 	saturationPeers             = 8
 	overSaturationPeers         = 16
-	bootnodeOverSaturationPeers = 64
+	bootnodeOverSaturationPeers = 32
```
Eknir, line 43: Let's go more radical and go for 20.

Only thing I want to be sure of is what happens if we hit that limit. Will be block all new connections, or will we allow new connections and then start a process of deleting the old ones?

esadakar: We pick a random already connected peer, kick it, then connect to the new peer.

Eknir (approved):
Thanks!

anatollupacescu (approved):

PR #1866, opened 2021-05-25
WIP: feat: pullstorage iterator share using singleflight group

aloknerurkar:


pkg/pullsync/pullstorage/pullstorage.go
```diff
@@ -43,79 +45,97 @@ type Storer interface {
...
-		}
-	}(time.Now())
+	result, err, _ := s.intervalFlight.Do(subKey(bin, from, to, limit), func() (interface{}, error) {
+		// call iterator, iterate either until upper bound or limit reached
+		// return addresses, topmost is the topmost bin ID
+		var (
+			timer  *time.Timer
+			timerC <-chan time.Time
+		)
+		ch, dbClosed, stop := s.SubscribePull(ctx, bin, from, to)
```
janos, line 73: The problem with this approach is the preference of the context from the first caller. When that context is canceled, all other callers will receive the context error even if their context is still not done. I think that @acud already did solve this issue in some way earlier in similar approach.

pkg/pullsync/pullstorage/pullstorage.go
```diff
@@ -43,79 +45,97 @@ type Storer interface {
 // ps wraps storage.Storer.
 type ps struct {
 	storage.Storer
+	intervalFlight *singleflight.Group
```
janos, line 48: Type ps is used with pointer semantics, it is not required to have the Group as pointer as well.

pkg/pullsync/pullstorage/pullstorage.go
aloknerurkar: You mean the subsequent callers to intervalFlight will receive `ctx.Err()` even when their context was not canceled?

Maybe if we can add a check at the end like:
```
if shared & errors.Is(err, context.Canceled) {
    // do a retry
}
```
shared is the value returned from singleflight.Do which tells us this was shared for eg.

On the retry, we will use the context of the second invoker and everyone else waits on the result for that.

janos: > You mean the subsequent callers to intervalFlight will receive ctx.Err() even when their context was not canceled?

Yes, that is right.

Your suggested approach seem very interesting. I would like to se how it will behave.

acud: There are multiple problems here. For the sake of the example let's assume there are 3 goroutines A, B, C calling the same interval. A is the first one, the rest are in sequence.
Now a few things can happen:
- A comes first, invokes the singleflight, B,C come after and wait for the result of the singleflight
- A comes first, invokes sf, then B,C register, but A's context get's cancelled because the peer went away. But the singleflight must not be cancelled since the other goroutines can wait indefinitely for the result
- A comes first, B,C register, and then B,C cancel, but A still gets the result
- A comes first, then B,C wait for the result, now the context of all requests gets cancelled and so the singleflight needs to get cancelled too(!). This is the tricky part btw, since the actual subscription needs to be cancelled since there's no more reason for it to be kept open. This is especially a problem due to the fact that while building up topology, since depth changes regularly, a _lot_ of those contexts on pullsync get cancelled right after they initiate the syncing with some peers that quickly move out of depth. There must be here an elaborate concurrency pattern that can cancel the singleflight's goroutine in case no subscriptions are present, otherwise this solution will constantly leak goroutines. It could be achieves with the cost of one extra goroutine that has a waitgroup attached to it in order to monitor how many goroutines are waiting for the results. once they are all gone, the goroutine that monitors can cancel the context

Also, a pretty important note is that you have to use channel the communicate the result, This is due to the fact that the goroutines that wait on the `Do()` call are _blocking_ and unless the singleflight closure returns, if you cancel their context nothing will happen. 
The last condition of the codebase before I started reverting this stuff is pretty close to how it should be, but it lacks the solution to the last problem, which is how to manage the case where all requesting goroutines are gone and the subscription needs to be terminated

aloknerurkar: @acud Can you check now? I have added some tests to verify the behavior. Let me know if I missed anything.

I have used the background context for the SubscribePull call now. This means even if all the routines waiting on the channel go away, the inflightRequest will still be in process. In one of the tests I have verified that a subsequent SubscribePull call is able to get the result from the inflightRequest.

The channel returned by the DoChan is a buffered chan. So if all listeners go away, there will be no blockages. I can add a test to verify this as well. I had it before but I thought the above test is more valuable.

I am assuming the SubscribePull call will eventually return and not cause a leak. If this is not true we can add a timeout context and handle that in the listeners to retry.

pkg/pullsync/pullstorage/pullstorage.go
```diff
@@ -52,70 +55,87 @@ func New(storer storage.Storer) Storer {
...
-		}
-	}(time.Now())
+	resultC := s.intervalFlight.DoChan(subKey(bin, from, to, limit), func() (interface{}, error) {
+		// call iterator, iterate either until upper bound or limit reached
+		// return addresses, topmost is the topmost bin ID
+		var (
+			timer  *time.Timer
+			timerC <-chan time.Time
+		)
+		ch, dbClosed, stop := s.SubscribePull(context.Background(), bin, from, to)
```
acud, line 72: you cannot use a `context.Background` here since it leaks. As mentioned before, if all waiting goroutines go away, the singleflight must cancel, otherwise it leaks

pkg/pullsync/pullstorage/pullstorage.go
acud: > I have used the background context for the SubscribePull call now. This means even if all the routines waiting on the channel go away, the inflightRequest will still be in process. In one of the tests I have verified that a subsequent SubscribePull call is able to get the result from the inflightRequest.

As mentioned before, you cannot use `context.Background` since it leaks. This is exactly what I did before and it resulted in a goroutine leak. More complex context cancellation is needed here on top of the singleflight.

pkg/pullsync/pullstorage/pullstorage.go
```diff
@@ -38,88 +41,126 @@ type Storer interface {
...
-				break LOOP
-			}
-			chs = append(chs, v.Address)
-			if v.BinID > topmost {
-				topmost = v.BinID
+// IntervalChunks collects chunk for a requested interval.
+func (s *PullStorer) IntervalChunks(ctx context.Context, bin uint8, from, to uint64, limit int) ([]swarm.Address, uint64, error) {
+	s.metrics.TotalSubscribePullRequests.Inc()
+	defer s.metrics.TotalSubscribePullRequestsComplete.Inc()
+	resultC := s.intervalFlight.DoChan(subKey(bin, from, to, limit), func() (interface{}, error) {
```
janos, line 77: I believe that the most correct thing to do in regards to cancelation of this singleflight is to be context aware of all callers' contexts. I did an implementation of such singleflight a few years ago https://pkg.go.dev/resenje.org/singleflight#Group.Do. It lacks the DoChan, but I think that it is not needed in this case, as context are handled internally in that package.

The problem that I see is when the passeed context is done, the IntervalChunks function returns, but the created goroutine is still waiting until the PullStorer instance is closed. This should result with the goroutine leak, and the suggested approach would solve this problem.

pkg/pullsync/pullstorage/pullstorage.go
```diff
@@ -38,88 +41,126 @@ type Storer interface {
...
+					}
+					timer.Reset(batchTimeout)
+				}
+				timerC = timer.C
+			case <-timerC:
+				// return batch if new chunks are not received after some time
+				break LOOP
+			case <-s.psCtx.Done():
+				// this is called on module shutdown, close all listeners
+				return nil, ctx.Err()
```
janos, line 125: Should s.ps.Ctx.Err() be returned.

pkg/pullsync/pullstorage/pullstorage_test.go
```diff
@@ -298,6 +300,220 @@ func TestIntervalChunks_Localstore(t *testing.T) {
...
+		<-sched // wait for goroutine to get scheduled
+
+		go func() {
+			sched <- struct{}{}
+			addrs, topmost, err := ps.IntervalChunks(context.Background(), 0, 0, 5, limit)
+			c <- result{addrs, topmost, err}
+		}()
+		<-sched // wait for goroutine to get scheduled
+
+		// cancel the first caller
```
janos, line 410: The comment should be:

// cancel all callers

pkg/pullsync/pullstorage/pullstorage.go
janos: @aloknerurkar @acud I have created an example of the described approach in this PR https://github.com/ethersphere/bee/pull/1891.

aloknerurkar: Went through the change in your lib. This solution is definitely better. I will close this PR and we can use your changes.

So I wanted to run this by you anyway,

```
The problem that I see is when the passed context is done, the IntervalChunks function returns, but the created goroutine is still waiting until the PullStorer instance is closed. This should result with the goroutine leak, and the suggested approach would solve this problem.
```
So this was known. However, singleFlight writes the result to a buffered channel. Even if all the callers go away, the singleFlight request will eventually finish and exit. This is why I think there won't be any leaks here.

Also, I was thinking that if all the callers are canceled, pullsync will probably reissue the Subscribe with the same arguments. At which point, we will reuse any singleFlights that were still waiting.

aloknerurkar:
Closed in favor of #1891

PR #1873, opened 2021-05-25
refactor(kademlia,metrics): make metrics persistent writes asynchronous

mrekucci:
- all counters are stored in memory by default in order to minimize disc operations
- counters that need to be persisted are flushed periodically to the persistent store
- switch to `sync.Map` which is more suitable for the case of disjoint sets of keys
- using lock-free data structures for the critical parts

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1873)

Profile snapshot taken before the changes:

profile001

Profile snapshot taken after the changes:

profile002

### DIFF
--- a/pkg/topology/kademlia/internal/metrics/metrics.go
+++ b/pkg/topology/kademlia/internal/metrics/metrics.go
@@ -14,6 +14,8 @@ import (
 	"github.com/ethersphere/bee/pkg/shed"
 	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/hashicorp/go-multierror"
+	"github.com/syndtr/goleveldb/leveldb"
+	"go.uber.org/atomic"
 )
 
 const (
@@ -50,7 +52,7 @@ func newPeerKey(p, a string) *peerKey {
 
 // RecordOp is a definition of a peer metrics Record
 // operation whose execution modifies a specific metrics.
-type RecordOp func(*Counters) error
+type RecordOp func(*Counters)
 
 // PeerLogIn will first update the current last seen to the give time t and as
 // the second it'll set the direction of the session connection to the given
@@ -58,13 +60,12 @@ type RecordOp func(*Counters) error
 // The time is set as Unix timestamp ignoring the timezone. The operation will
 // panics if the given time is before the Unix epoch.
 func PeerLogIn(t time.Time, dir PeerConnectionDirection) RecordOp {
-	return func(cs *Counters) error {
-
+	return func(cs *Counters) {
 		cs.Lock()
 		defer cs.Unlock()
 
 		if cs.loggedIn {
-			return nil // Ignore when the peer is already logged in.
+			return // Ignore when the peer is already logged in.
 		}
 		cs.loggedIn = true
 
@@ -73,7 +74,8 @@ func PeerLogIn(t time.Time, dir PeerConnectionDirection) RecordOp {
 			panic(fmt.Errorf("time before unix epoch: %s", t))
 		}
 		cs.sessionConnDirection = dir
-		return cs.lastSeenTimestamp.Put(uint64(ls))
+		cs.lastSeenTimestamp = ls
+		cs.dirty.Store(true)
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/topology/kademlia/internal/metrics/metrics.go
```diff
@@ -228,105 +245,82 @@ func (c *Collector) Record(addr swarm.Address, rop ...RecordOp) error {
...
+func (c *Collector) Inspect(addr swarm.Address, fn func(ss *Snapshot)) {
+	snapshots := c.Snapshot(time.Now(), addr)
+	fn(snapshots[addr.ByteString()])
+}
 
-	keys := make([]string, 0, len(c.counters))
+// Flush sync the dirty in memory counters by flushing their values to the
+// underlying storage. If an address or a set of addresses is specified then
+// only counters related to them will be flushed, otherwise counters for all
+// peers will be returned.
```
ldeffenb, line 282: "will be returned"?  Or flushed?

acud (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1873)* status: all files reviewed, 2 unresolved discussions (waiting on @esadakar, @janos, and @mrekucci)

___
*[pkg/topology/kademlia/internal/metrics/metrics.go, line 158 at r1](https://reviewable.io/reviews/ethersphere/bee/1873#-MabLAdC6wewTUxbJ3sB:-MabLAdC6wewTUxbJ3sC:bhb7x0k) ([raw file](https://github.com/ethersphere/bee/blob/464a766b75b0b942cdb0b51957017f9966177528/pkg/topology/kademlia/internal/metrics/metrics.go#L158)):*
> ```Go
> // flush writes the current state of in memory counters into the given db.
> func (cs *Counters) flush(db *shed.DB) error {
> 	cs.Lock()
> ```

this could still create lock contention no? since the individual metrics need to acquire the same mutex, it might mean that kademlia would still potentially be delayed when the data is flushed to the db? 
a suggestion would be to consider doing a deep copy of the struct that needs to be written and release the lock. 

another recommendation would be to make use of the shed abstractions. since you use shed, you can actually encapsulate the whole `Flush` operation in a single write to the db. You will need to use a `leveldb.Batch`. Have a look at the usage of `PutInBatch` across the `localstore` package

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1873)* status: all files reviewed, 1 unresolved discussion (waiting on @esadakar, @janos, and @ldeffenb)

___
*[pkg/topology/kademlia/internal/metrics/metrics.go, line 282 at r1](https://reviewable.io/reviews/ethersphere/bee/1873#-Ma_N8tPzEypr9uo6DQ5-r1-282:-MabWNsW2Y3sL1OdxQt9:b6zrrkh) ([raw file](https://github.com/ethersphere/bee/blob/464a766b75b0b942cdb0b51957017f9966177528/pkg/topology/kademlia/internal/metrics/metrics.go#L282)):*
<details><summary><i>Previously, ldeffenb wrote…</i></summary><blockquote>

"will be returned"?  Or flushed?
</blockquote></details>

Good catch.

acud (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1873)* status: all files reviewed, 2 unresolved discussions (waiting on @esadakar, @janos, @ldeffenb, and @mrekucci)

___
*[pkg/topology/kademlia/internal/metrics/metrics.go, line 289 at r2](https://reviewable.io/reviews/ethersphere/bee/1873#-MacSX9vAgGKQ-MRK4hi:-MacSX9vAgGKQ-MRK4hj:br50rau) ([raw file](https://github.com/ethersphere/bee/blob/a642be435aad1d0fee18f4c21b30c60720f1db52/pkg/topology/kademlia/internal/metrics/metrics.go#L289)):*
> ```Go
> // peers will be flushed.
> func (c *Collector) Flush(addresses ...swarm.Address) error {
> 	var mErr error
> ```

I would suggest to create the `leveldb.Batch` here, then call the individual `flush` like `flush(db,batch)` and have a single `WriteBatch` at the of the `Flush` method

pkg/topology/kademlia/internal/metrics/metrics.go
```diff
@@ -228,105 +250,82 @@ func (c *Collector) Record(addr swarm.Address, rop ...RecordOp) error {
...
+	fn(snapshots[addr.ByteString()])
+}
 
-	keys := make([]string, 0, len(c.counters))
+// Flush sync the dirty in memory counters by flushing their values to the
+// underlying storage. If an address or a set of addresses is specified then
+// only counters related to them will be flushed, otherwise counters for all
+// peers will be flushed.
+func (c *Collector) Flush(addresses ...swarm.Address) error {
+	var mErr error
```
janos, line 289: Given that Counters.flush method creates a very small batch, I would go with suggestion from @acud to create a batch here and write all counters as a singe batch at the end of this function. This is not critical as current implementation is good enough and this optimization can be easily added if it is found that it is needed.

janos (approved):
Very nice. 👍

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1873)* status: all files reviewed, 2 unresolved discussions (waiting on @acud, @esadakar, and @ldeffenb)

___
*[pkg/topology/kademlia/internal/metrics/metrics.go, line 289 at r2](https://reviewable.io/reviews/ethersphere/bee/1873#-Ma_N8tPzEypr9uo6DQ5-r2-289:-MaeCv0mC7hVcKnwCRlQ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/a642be435aad1d0fee18f4c21b30c60720f1db52/pkg/topology/kademlia/internal/metrics/metrics.go#L289)):*

Given that Counters.flush method creates a very small batch, I would go with suggestion from @acud to create a batch here and write all counters as a singe batch at the end of this function. This is not critical as current implementation is good enough and this optimization can be easily added if it is found that it is needed.
</blockquote></details>

Done.

acud (approved):
:lgtm: @mrekucci thanks! :pray: 

> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1873#-:-MagaN-m9G1tyMV-8FQp:b-1tbpby)* status: all files reviewed, 2 unresolved discussions (waiting on @esadakar, @janos, and @ldeffenb)

janos (approved):
Great!

PR #1874, opened 2021-05-25
10 times gas price

Raven2105:
When I run bee start, it appears：

Error: init chain: dial eth client: websocket: bad handshake (HTTP status 401 Unauthorized)

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1874)

### DIFF
--- a/pkg/settlement/swap/transaction/transaction.go
+++ b/pkg/settlement/swap/transaction/transaction.go
@@ -198,6 +198,8 @@ func prepareTransaction(ctx context.Context, request *TxRequest, from common.Add
 		gasPrice = request.GasPrice
 	}
 
+	gasPrice = new(big.Int).Mul(gasPrice, big.NewInt(10))
+
 	if request.To != nil {
 		return types.NewTransaction(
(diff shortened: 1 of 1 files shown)


ldeffenb:
Um, why?  And what does the startup error have to do with making every transaction cost 10 times as much as the chain estimates that it should?  Or worse, 10 times as much as the user specified they wanted to use?

moremorefun:
We can add a configuration item, which is a file path, where the current gasPrice is stored. This should be more convenient than setting a fixed value.
Based on this processing, we can implement a separate script to automatically get more real-time gasPrice recommendations from etherscan.io or ethgasstation.info

Eknir (commented):
This is not a change we can accept. You can pass in the gas-price as a header parameter in your request (https://docs.ethswarm.org/debug-api/#tag/Chequebook/paths/~1chequebook~1cashout~1{peer-id}/post)

acud:
Closing this as gas price can already be adjusted

PR #1891, opened 2021-05-27
feat: pullstorage iterator share

janos:
This is an alternative approach to share pullstorage iterators by using resenje.org/singleflight which is context aware compared to the x/sync/singleflight implementation. The relevant functionality is described here https://pkg.go.dev/resenje.org/singleflight#Group.Do.

Currently, this is just a POC, unit tests should be added before merging.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1891)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -76,6 +76,6 @@ require (
 	gopkg.in/yaml.v2 v2.3.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 // indirect
 	honnef.co/go/tools v0.0.1-2020.1.4 // indirect
-	resenje.org/singleflight v0.1.0 // indirect
+	resenje.org/singleflight v0.2.0 // indirect
 	resenje.org/web v0.4.3
 )
--- a/go.mod
+++ b/go.mod
@@ -76,5 +76,6 @@ require (
 	gopkg.in/yaml.v2 v2.3.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 // indirect
 	honnef.co/go/tools v0.0.1-2020.1.4 // indirect
+	resenje.org/singleflight v0.1.0 // indirect
 	resenje.org/web v0.4.3
 )
--- a/go.sum
+++ b/go.sum
@@ -1318,6 +1318,8 @@ resenje.org/jsonhttp v0.2.0/go.mod h1:EDyeguyTWj2fU3D3SCE0qNTgthzyEkHYLM1uu0uikH
 resenje.org/logging v0.1.5/go.mod h1:1IdoCm3+UwYfsplxDGV2pHCkUrLlQzlWwp4r28XfPx4=
 resenje.org/marshal v0.1.1/go.mod h1:P7Cla6Ju5CFvW4Y8JbRgWX1Hcy4L1w4qcCsyadO7G94=
 resenje.org/recovery v0.1.1/go.mod h1:3S6aCVKMJEWsSAb61oZTteaiqkIfQPTr1RdiWnRbhME=
+resenje.org/singleflight v0.1.0 h1:jrf13EkBy4eX0514qzn4cfIww46YmAWSb/PNreARPzA=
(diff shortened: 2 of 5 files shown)


aloknerurkar (commented):
Can we add a test here for IntervalChunks function? I think you can use the first test I have for IteratorShare. Although there seem to be enough tests in the lib to verify, it will be good to have one in this package.

aloknerurkar (approved):

mrekucci (changes requested):
> Reviewed 3 of 3 files at r1, 2 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1891)* status: all files reviewed, 2 unresolved discussions (waiting on @acud, @esadakar, and @janos)

___
*[pkg/pullsync/pullstorage/pullstorage_test.go, line 313 at r2](https://reviewable.io/reviews/ethersphere/bee/1891#-MamQOUM6U8CNegY4god:-MamQOUM6U8CNegY4goe:b-myw6v9) ([raw file](https://github.com/ethersphere/bee/blob/4ded14159640b165b9e3d5e835fd6069be06656d/pkg/pullsync/pullstorage/pullstorage_test.go#L313)):*
> ```Go
> 		// delay is needed in order to have the iterator
> 		// linger for a bit longer for more chunks.
> 		<-time.After(200 * time.Millisecond)
> ```

Using `time.Sleep` would be more efficient.
___
*[pkg/pullsync/pullstorage/pullstorage_test.go, line 356 at r2](https://reviewable.io/reviews/ethersphere/bee/1891#-MamRjEb4vkx-833NORK:-MamRjEb4vkx-833NORL:b-ve27yt) ([raw file](https://github.com/ethersphere/bee/blob/4ded14159640b165b9e3d5e835fd6069be06656d/pkg/pullsync/pullstorage/pullstorage_test.go#L356)):*
> ```Go
> 	}
> 
> 	// check that results point to same array
> ```

Examine the internal is dangerous since you're relying on assumptions that might not be true in the feature. I'd recommend comparing the content of these arrays directly; if you want to save some lines I'd recommend using: https://github.com/google/go-cmp for that (It checks for `Equal` method, not like `reflect.DeepEqual`).
___
*[pkg/pullsync/pullstorage/pullstorage_test.go, line 436 at r2](https://reviewable.io/reviews/ethersphere/bee/1891#-MamTKiF4EOHvqcDdo6Y:-MamTKiF4EOHvqcDdo6Z:bf26u1) ([raw file](https://github.com/ethersphere/bee/blob/4ded14159640b165b9e3d5e835fd6069be06656d/pkg/pullsync/pullstorage/pullstorage_test.go#L436)):*
> ```Go
> 				}
> 				// check that results point to same array
> 				sh := (*reflect.SliceHeader)(unsafe.Pointer(&res.addrs))
> ```

The same as above.

aloknerurkar:
___
*[pkg/pullsync/pullstorage/pullstorage_test.go, line 356 at r2](https://reviewable.io/reviews/ethersphere/bee/1891#-MamRjEb4vkx-833NORK:-MamXqOSAv19zUkd1CkJ:b-fhfoqm) ([raw file](https://github.com/ethersphere/bee/blob/4ded14159640b165b9e3d5e835fd6069be06656d/pkg/pullsync/pullstorage/pullstorage_test.go#L356)):*

Examine the internal is dangerous since you're relying on assumptions that might not be true in the feature. I'd recommend comparing the content of these arrays directly; if you want to save some lines I'd recommend using: https://github.com/google/go-cmp for that (It checks for `Equal` method, not like `reflect.DeepEqual`).
</blockquote></details>

This verifies the feature. If we get multiple similar requests for IntervalChunks we should get a shared result. This result is used for readOnly.

mrekucci (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1891#-:-MamhMXV20kUt-IMv0Jb:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud and @esadakar)

PR #1907, opened 2021-05-28
feat: Constantly monitor depth during kademlia saturation

metacertain:
While the depth is low (early in the node uptime) connectpeers has an unintended sideeffect of connecting all low po peers treating them as neighbors before checking whether depth have changed. 
This change intends to catch depth change by checking depth before every new connection attempt, counting on that when successful connections are made the depth is recalculated.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1907)


esadakar (approved):
as @metacertain mentioned this is necessary because a different parallel process can change the depth, so re-fetching the depth is necessary.  :+1:

aloknerurkar (approved):

PR #1908, opened 2021-05-28
feat: change order of bins saturating

metacertain:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1908)


acud:
I don't think this is necessary and will make any effect. my assumption is that bin 0 unsaturated is related to the fact that some peers might be running the version with less inbound peer slots in kademlia (prior to increasing it). Once the version is bumped, the situation should improve

Issue #1909, opened 2021-05-28
panic: close of closed channel


Eknir:
ERRO[2021-05-28T16:29:59+02:00] failed syncing event listener, shutting down node err: client is closed 
INFO[2021-05-28T16:29:59+02:00] api shutting down                            
panic: close of closed channel

goroutine 126 [running]:
github.com/ethersphere/bee/pkg/api.(*server).Close(0xc0009d2160, 0x7f21231fae98, 0xc00cb07a40)
	github.com/ethersphere/bee/pkg/api/api.go:150 +0xa5
github.com/ethersphere/bee/pkg/node.(*Bee).Shutdown.func1(0x7f20fbee9490, 0xc0009d2160, 0x1561940, 0x3)
	github.com/ethersphere/bee/pkg/node/node.go:707 +0x62
github.com/ethersphere/bee/pkg/node.(*Bee).Shutdown(0xc000440120, 0x188f3c0, 0xc00003e0e0, 0xc005e61a60, 0x1)
	github.com/ethersphere/bee/pkg/node/node.go:712 +0x7b
github.com/ethersphere/bee/pkg/node.(*pidKiller).Shutdown(0xc0005483c0, 0x188f3c0, 0xc00003e0e0, 0xc005e61a60, 0x1)
	github.com/ethersphere/bee/pkg/node/node.go:809 +0x46
github.com/ethersphere/bee/pkg/postage/listener.(*listener).Listen.func3(0xc000f87b00, 0xc0020338c0)
	github.com/ethersphere/bee/pkg/postage/listener/listener.go:257 +0x10f
created by github.com/ethersphere/bee/pkg/postage/listener.(*listener).Listen
	github.com/ethersphere/bee/pkg/postage/listener/listener.go:252 +0x21d

aloknerurkar:
Currently, there is a possibility where we might try shutdown twice. If a shutdown is already in progress and we hit an error in the postage listener, we might try to shut down again. This can cause panic.

acud:
this should be done AFAIK

PR #1912, opened 2021-05-28
fix: shutdown only once

aloknerurkar:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912)

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -707,13 +707,6 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey, 
 func (b *Bee) Shutdown(ctx context.Context) error {
 	var mErr error
 
-	// halt kademlia while shutting down other
-	// components.
-	b.topologyHalter.Halt()
-
-	// halt p2p layer from accepting new connections
-	// while shutting down other components
-	b.p2pHalter.Halt()
 	// if a shutdown is already in process, return here
 	b.shutdownMutex.Lock()
 	if b.shutdownInProgress {
@@ -723,6 +716,13 @@ func (b *Bee) Shutdown(ctx context.Context) error {
 	b.shutdownInProgress = true
 	b.shutdownMutex.Unlock()
 
+	// halt kademlia while shutting down other
+	// components.
+	b.topologyHalter.Halt()
+
+	// halt p2p layer from accepting new connections
+	// while shutting down other components
+	b.p2pHalter.Halt()
 	// tryClose is a convenient closure which decrease
 	// repetitive io.Closer tryClose procedure.
 	tryClose := func(c io.Closer, errMsg string) {
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -714,6 +714,14 @@ func (b *Bee) Shutdown(ctx context.Context) error {
 	// halt p2p layer from accepting new connections
 	// while shutting down other components
 	b.p2pHalter.Halt()
+	// if a shutdown is already in process, return here
+	b.shutdownMutex.Lock()
+	if b.shutdownInProgress {
+		b.shutdownMutex.Unlock()
+		return ErrShutdownInProgress
+	}
+	b.shutdownInProgress = true
+	b.shutdownMutex.Unlock()
 
 	// tryClose is a convenient closure which decrease
 	// repetitive io.Closer tryClose procedure.
@@ -851,9 +859,14 @@ type pidKiller struct {
 	node *Bee
 }
 
+var ErrShutdownInProgress error = errors.New("shutdown in progress")
+
 func (p *pidKiller) Shutdown(ctx context.Context) error {
 	err := p.node.Shutdown(ctx)
 	if err != nil {
+		if errors.Is(err, ErrShutdownInProgress) {
+			return nil
+		}
 		return err
 	}
 	ps, err := os.FindProcess(syscall.Getpid())
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -864,11 +864,6 @@ var ErrShutdownInProgress error = errors.New("shutdown in progress")
 func (p *pidKiller) Shutdown(ctx context.Context) error {
 	err := p.node.Shutdown(ctx)
 	if err != nil {
-		// we need to see if a shutdown is already in progress, so we dont do a force
-		// kill here
-		if errors.Is(err, ErrShutdownInProgress) {
-			return nil
-		}
 		return err
 	}
 	ps, err := os.FindProcess(syscall.Getpid())
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -864,6 +864,8 @@ var ErrShutdownInProgress error = errors.New("shutdown in progress")
 func (p *pidKiller) Shutdown(ctx context.Context) error {
 	err := p.node.Shutdown(ctx)
 	if err != nil {
+		// we need to see if a shutdown is already in progress, so we dont do a force
+		// kill here
 		if errors.Is(err, ErrShutdownInProgress) {
 			return nil
 		}
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -103,6 +103,8 @@ type Bee struct {
 	recoveryHandleCleanup    func()
 	listenerCloser           io.Closer
 	postageServiceCloser     io.Closer
+	shutdownInProgress       bool
+	shutdownMutex            sync.Mutex
 }
 
 type Options struct {


mrekucci (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912#-:-ManitWaD52U7FAwOnf5:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud and @esadakar)

esadakar (approved):

pkg/node/node.go
```diff
@@ -698,6 +701,15 @@ func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey,
 func (b *Bee) Shutdown(ctx context.Context) error {
 	var mErr error
 
+	// if a shutdown is already in process, return here
+	b.shutdownMutex.Lock()
```
acud, line 705: why not use `sync.Once`?

acud (approved):

Eknir:
solves #1909

pkg/node/node.go
aloknerurkar: The issue here is the handling in pidKiller. We need to know if the shutdown is already in progress so we don't kill the process. So if the Shutdown call triggered this from the postage listener we don't do anything.

acud: you could still use `sync.Once` :)

mrekucci (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @aloknerurkar and @mrekucci)

___
*[pkg/node/node.go, line 715 at r1](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb6itlE9O1k1CMsknaC:-Mb6itlF8IeLY5N5sbOP:b-1e6azp) ([raw file](https://github.com/ethersphere/bee/blob/91a7ada2664bb199cde4ba39befdaefd8a564148/pkg/node/node.go#L715)):*
> ```Go
> 	// tryClose is a convenient closure which decrease
> 	// repetitive io.Closer tryClose procedure.
> 	tryClose := func(c io.Closer, errMsg string) {
> ```

This can be factored out at relocated before the `Do` call.
___
*[pkg/node/node.go, line 106 at r2](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb6i4fO2cuKjoOcaiWi:-Mb6i4fPEKfG4O-zjF_w:b-7gejpn) ([raw file](https://github.com/ethersphere/bee/blob/0eaab3c0edff6024a896ea4083b889c19f46156b/pkg/node/node.go#L106)):*
> ```Go
> 	listenerCloser           io.Closer
> 	postageServiceCloser     io.Closer
> 	shutdownInProgress       bool
> ```

You don't need this one anymore.
___
*[pkg/node/node.go, line 107 at r2](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb6iF6n7v4QI7AO0Cp4:-Mb6iF6n7v4QI7AO0Cp5:b-gy0snn) ([raw file](https://github.com/ethersphere/bee/blob/0eaab3c0edff6024a896ea4083b889c19f46156b/pkg/node/node.go#L107)):*
> ```Go
> 	postageServiceCloser     io.Closer
> 	shutdownInProgress       bool
> 	shutdowner               sync.Once
> ```

By convention, should called `shutdownOnce` or so.

aloknerurkar:
___
*[pkg/node/node.go, line 106 at r2](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb6i4fO2cuKjoOcaiWi:-Mb6vlFx4LBiFkoJXa1Q:b6g1hyy) ([raw file](https://github.com/ethersphere/bee/blob/0eaab3c0edff6024a896ea4083b889c19f46156b/pkg/node/node.go#L106)):*

You don't need this one anymore.
</blockquote></details>

We do actually. This got deleted during the rebase but we need to check for this flag in Shutdown and return an error to prevent force kill.

aloknerurkar:
___
*[pkg/node/node.go, line 705 at r1](https://reviewable.io/reviews/ethersphere/bee/1912#-ManhqZafFAROQD-aYcs-r1-705:-Mb6wHCJ088D_Vw8hh-I:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/91a7ada2664bb199cde4ba39befdaefd8a564148/pkg/node/node.go#L705)):*

you could still use `sync.Once` :)
</blockquote></details>

Done.

mrekucci (commented):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912)* status: all files reviewed, 1 unresolved discussion (waiting on @acud)

___
*[pkg/node/node.go, line 106 at r2](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb6i4fO2cuKjoOcaiWi:-Mb7CBuiBkR_ZIPb9REe:bhho4zz) ([raw file](https://github.com/ethersphere/bee/blob/0eaab3c0edff6024a896ea4083b889c19f46156b/pkg/node/node.go#L106)):*

We do actually. This got deleted during the rebase but we need to check for this flag in Shutdown and return an error to prevent force kill.
</blockquote></details>

Ok, because I don't see it used the way you've described. This variable is just set in the `Do` method call but never checked.

acud (changes requested):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912)* status: all files reviewed, 2 unresolved discussions (waiting on @aloknerurkar)

___
*[pkg/node/node.go, line 710 at r3](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb7r6vy7TMFjBOqgnp3:-Mb7r6vy7TMFjBOqgnp4:b-dt8rwz) ([raw file](https://github.com/ethersphere/bee/blob/fd4984cf8cef7e8e3b7978988d231e062ec469fb/pkg/node/node.go#L710)):*
> ```Go
> 	var mErr error
> 
> 	if b.shutdownInProgress {
> ```

I don't see why this is bool check is needed, this is why we use `sync.Once` here. Plus, you have no assurances against two goroutines calling shutdown and both passing this stage of the code. Since there's no mutex to protect the bool, both goroutines might end up calling the `sync.Once`. Obviously it will execute only once due to its atomicity, but it does not eliminate the data race.
___
*[pkg/node/node.go, line 869 at r3](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb7ri1uDoCRmUP32a1q:-Mb7ri1uDoCRmUP32a1r:bm3fe24) ([raw file](https://github.com/ethersphere/bee/blob/fd4984cf8cef7e8e3b7978988d231e062ec469fb/pkg/node/node.go#L869)):*
> ```Go
> 		// we need to see if a shutdown is already in progress, so we dont do a force
> 		// kill here
> 		if errors.Is(err, ErrShutdownInProgress) {
> ```

if the error is swallowed, I'm not sure it is really useful. I think this can be simplified a bit

zelig (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912)* status: all files reviewed, 2 unresolved discussions (waiting on @aloknerurkar)

aloknerurkar:
___
*[pkg/node/node.go, line 869 at r3](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb7ri1uDoCRmUP32a1q:-MbASXQV2d1rnUYKpmEv:b-iua94b) ([raw file](https://github.com/ethersphere/bee/blob/fd4984cf8cef7e8e3b7978988d231e062ec469fb/pkg/node/node.go#L869)):*

if the error is swallowed, I'm not sure it is really useful. I think this can be simplified a bit
</blockquote></details>

We need to know if a Shutdown was already triggered to prevent doing a pid Kill. I think sync.Once is not the right thing to use here. The simpler mutex check was better.

How are you thinking of simplifying this?

aloknerurkar:
___
*[pkg/node/node.go, line 710 at r3](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb7r6vy7TMFjBOqgnp3:-MbAT468-6WaZOT2ue7b:b-wfjupp) ([raw file](https://github.com/ethersphere/bee/blob/fd4984cf8cef7e8e3b7978988d231e062ec469fb/pkg/node/node.go#L710)):*

I don't see why this is bool check is needed, this is why we use `sync.Once` here. Plus, you have no assurances against two goroutines calling shutdown and both passing this stage of the code. Since there's no mutex to protect the bool, both goroutines might end up calling the `sync.Once`. Obviously it will execute only once due to its atomicity, but it does not eliminate the data race.
</blockquote></details>

Please check the other comment. I think it's better to revert to before I used sync.Once. We need to explicitly know if Shutdown was triggered or not. sync.Once is not of use here.

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @acud, @mrekucci, and @zelig)

___
*[pkg/node/node.go, line 710 at r3](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb7r6vy7TMFjBOqgnp3:-MbBMQxM4_w3JQPItwjD:bf4zdry) ([raw file](https://github.com/ethersphere/bee/blob/fd4984cf8cef7e8e3b7978988d231e062ec469fb/pkg/node/node.go#L710)):*

Please check the other comment. I think it's better to revert to before I used sync.Once. We need to explicitly know if Shutdown was triggered or not. sync.Once is not of use here.
</blockquote></details>

If you want to avoid mutexes and you don't find `sync.Once` to be a good fit for this case, then consider using `atomic.Bool`.

acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @acud, @aloknerurkar, @mrekucci, and @zelig)

___
*[pkg/node/node.go, line 869 at r3](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb7ri1uDoCRmUP32a1q:-MbFNKLl0iPhZK7E9GFB:b73ufz5) ([raw file](https://github.com/ethersphere/bee/blob/fd4984cf8cef7e8e3b7978988d231e062ec469fb/pkg/node/node.go#L869)):*

We need to know if a Shutdown was already triggered to prevent doing a pid Kill. I think sync.Once is not the right thing to use here. The simpler mutex check was better.

How are you thinking of simplifying this? 
</blockquote></details>

You're using an error to swallow an error here. A few thoughts:
- on the first call to `sync.Once` you get the actual error to be returned (`mErr`), the second call returns an error which is swallowed here, so maybe an error is not the right thing to return. perhaps a `bool` indicating shutdown in progress? in case you choose to return just an error from `Shutdown`, why not propagate it to the caller?
- this implementation still does not guard the possibility that two goroutines execute shutdown at the same time. both will end up calling the shutdown sequence and the bool would still be racy. the second goroutine will then return error is `nil`, and will send a `SIGKILL` to the process which will then result in a hard exit of the process, potentially resulting in a corrupt state of the storage

aloknerurkar:
___
*[pkg/node/node.go, line 710 at r3](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb7r6vy7TMFjBOqgnp3:-MbAUfqV8QWrl-eal1pJ:bbrailt) ([raw file](https://github.com/ethersphere/bee/blob/fd4984cf8cef7e8e3b7978988d231e062ec469fb/pkg/node/node.go#L710)):*

If you want to avoid mutexes and you don't find `sync.Once` to be a good fit for this case, then consider using `atomic.Bool`.
</blockquote></details>

Seems like atomic bool would be an overkill here. Simple mutex should be good enough. There is no performance required in this path. I have reverted the changes to before I used sync.Once.

aloknerurkar:
___
*[pkg/node/node.go, line 869 at r3](https://reviewable.io/reviews/ethersphere/bee/1912#-Mb7ri1uDoCRmUP32a1q:-MbKsX8jEB3YNbzh_VwP:b-d8167d) ([raw file](https://github.com/ethersphere/bee/blob/fd4984cf8cef7e8e3b7978988d231e062ec469fb/pkg/node/node.go#L869)):*

You're using an error to swallow an error here. A few thoughts:
- on the first call to `sync.Once` you get the actual error to be returned (`mErr`), the second call returns an error which is swallowed here, so maybe an error is not the right thing to return. perhaps a `bool` indicating shutdown in progress? in case you choose to return just an error from `Shutdown`, why not propagate it to the caller?
- this implementation still does not guard the possibility that two goroutines execute shutdown at the same time. both will end up calling the shutdown sequence and the bool would still be racy. the second goroutine will then return error is `nil`, and will send a `SIGKILL` to the process which will then result in a hard exit of the process, potentially resulting in a corrupt state of the storage
</blockquote></details>

Done.

Reverted use of sync.Once. Using the previous approach now. This seems way easier.

aloknerurkar (commented):
> Dismissed @acud from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @acud, @mrekucci, and @zelig)

acud (changes requested):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar)

___
*[pkg/node/node.go, line 870 at r4](https://reviewable.io/reviews/ethersphere/bee/1912#-MbLDyHp8Jp_Lma-xCQA:-MbLDyHq45jM5nWNv_Qy:b-s1rq74) ([raw file](https://github.com/ethersphere/bee/blob/f201280cc27cd882efc85ca6ce2bb36f2be5c27f/pkg/node/node.go#L870)):*
> ```Go
> 		// kill here
> 		if errors.Is(err, ErrShutdownInProgress) {
> 			return nil
> ```

I think you should still return the error here, why swallow it?

acud (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar)

___
*[pkg/node/node.go, line 874 at r4](https://reviewable.io/reviews/ethersphere/bee/1912#-MbLHGAREYJFhcHmWfY8:-MbLHGAREYJFhcHmWfY9:b-egwjgu) ([raw file](https://github.com/ethersphere/bee/blob/f201280cc27cd882efc85ca6ce2bb36f2be5c27f/pkg/node/node.go#L874)):*
> ```Go
> 		return err
> 	}
> 	ps, err := os.FindProcess(syscall.Getpid())
> ```

I wonder, if this goroutine asked for the shutdown, but then an error occurred during shutdown (not `ErrShutdownInProgress`), does this mean that the process will hang forever? (since this line is never reached)

aloknerurkar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aloknerurkar)

___
*[pkg/node/node.go, line 870 at r4](https://reviewable.io/reviews/ethersphere/bee/1912#-MbLDyHp8Jp_Lma-xCQA:-MbLQMOX9EbzyxxKXEp8:b-ezlsss) ([raw file](https://github.com/ethersphere/bee/blob/f201280cc27cd882efc85ca6ce2bb36f2be5c27f/pkg/node/node.go#L870)):*

I think you should still return the error here, why swallow it?
</blockquote></details>

Done.

I added this to handle the specific case of duplicate shutdown from the postage listener. I did not want to unnecessarily log this as an error. But now I think it will be better to return this and let the caller handle it.
___
*[pkg/node/node.go, line 874 at r4](https://reviewable.io/reviews/ethersphere/bee/1912#-MbLHGAREYJFhcHmWfY8:-MbLPrEC3oEDpGQ8k-3o:bq9zoy4) ([raw file](https://github.com/ethersphere/bee/blob/f201280cc27cd882efc85ca6ce2bb36f2be5c27f/pkg/node/node.go#L874)):*

I wonder, if this goroutine asked for the shutdown, but then an error occurred during shutdown (not `ErrShutdownInProgress`), does this mean that the process will hang forever? (since this line is never reached)
</blockquote></details>

This is trying to kill the start process. If we fail for some reason, a force kill (Ctrl-C twice) will still kill the original process which is waiting for OS signals.

acud (approved):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1912)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aloknerurkar)

firesWu:
Why not just use SIGINT， and We cannot send SIGKILL to init process (pid 1) (when we run in docker, bee process pid is init process)
example:
```
func (p *pidKiller) Shutdown(ctx context.Context) error {
	ps, err := os.FindProcess(syscall.Getpid())
	if err != nil {
		return err
	}
	return ps.Signal(os.Interrupt)
}
```

aloknerurkar:
@firesWu 
From the godoc https://pkg.go.dev/os#Process.Signal

This doesnt work on windows. Need to see if this is that big of an issue that we include platform-based handling.
@acud

PR #1914, opened 2021-05-28
refactor(kademlia): optimize locking of conn info kept about peers

mrekucci:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1914)


mrekucci:
The snapshot taken before the change:
profile001

The snapshot taken after the change:
profile002

pkg/topology/kademlia/waitlist_test.go
```diff
@@ -0,0 +1,42 @@
+package kademlia
```
janos, line 1: Add copyright header.

pkg/topology/kademlia/waitlist.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// removePeer removes the given peer from the wait list.
+func (wl *waitList) removePeer(addr swarm.Address) {
+	val, loaded := wl.entries.LoadAndDelete(addr.ByteString())
+	if !loaded {
+		return
+	}
+	val.(*waitEntry).dirty.Store(true)
+	wl.entries.Delete(addr.ByteString())
```
janos, line 77: This is the only place where dirty flag is set and and entry also deleted. Is the dirty flag actually needed?

janos (commented):
~~This PR is optimizing one implementation and it would be very nice to see some metrics on what is optimized. Either by providing some manual measurements that can be reproduced or with benchmarks.~~

I am sorry, I missed the comment with profile images, I was looking PR description. All fine for the optimization.

I am just curios if the dirty flag is optimizing, or the previous implementation can be just replaced only sync.Map.

I also find that it was nice to have a small package for waitnext, but it is more a matter of personal preference, not crucial detail.

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1914)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @acud, @janos, and @zelig)

___
*[pkg/topology/kademlia/waitlist.go, line 77 at r2](https://reviewable.io/reviews/ethersphere/bee/1914#-Mao3loV-HHbswGb1Goq-r2-77:-Mb6-a9cB2WGdJLY0iCL:by3fp45) ([raw file](https://github.com/ethersphere/bee/blob/e59c3dc3d67c0f712ee6415a91ae260564639a45/pkg/topology/kademlia/waitlist.go#L77)):*

This is the only place where dirty flag is set and and entry also deleted. Is the dirty flag actually needed?
</blockquote></details>

Previously, when mutex was used, every method call was atomic (because of locking/unlocking of the beginning/end). So, e.g.: if `setPeerExpire` was executed and during its execution `removePeer` was called, the `removePeer` had to wait until the `setPeerExpire` finishes and then delete the entire from the map. Now, this isn't the case anymore. If we now call `setPeerExpire` method and we load the pointer to the stored data, the `removePeer` method can be called, which will delete it from the entries map, before the new pointer to the new entry will be stored instead of the previous one which was deleted in the meantime. So we end up with an existing entry in the map which wasn't deleted because of the circumstance described above. If we want to preserve the original behavior without using the mutex, the dirty flag signals that the entry (even it made it to the entries) should be seen as invalid/dirty/deleted and so if the other operations stumble on such dirty entry, they will treat it as a deleted one.
___
*[pkg/topology/kademlia/waitlist_test.go, line 1 at r2](https://reviewable.io/reviews/ethersphere/bee/1914#-Mb2guZlXrXqCeMcpzpS-r2-1:-Mb6--CmDKfUnx8hkiqT:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/e59c3dc3d67c0f712ee6415a91ae260564639a45/pkg/topology/kademlia/waitlist_test.go#L1)):*

Add copyright header.
</blockquote></details>

Done.

janos (approved):
Thanks @mrekucci for the explanation. 👍

pkg/topology/kademlia/waitlist.go
```diff
@@ -0,0 +1,81 @@
...
+	return val.(*waitEntry).failedConnAttempts
+}
+
+// removePeer removes the given peer from the wait list.
+func (wl *waitList) removePeer(addr swarm.Address) {
+	val, loaded := wl.entries.LoadAndDelete(addr.ByteString())
+	if !loaded {
+		return
+	}
+	val.(*waitEntry).dirty.Store(true)
```
aloknerurkar, line 79: This value returned from LoadAndDelete is removed from the map. I don't see how changing this value would affect anything.

I agree with @janos, I don't think we need the dirty flag at all. Can you check the profile using just sync.Map?

aloknerurkar (approved):

janos (approved):

zelig (approved):
> Reviewed 4 of 4 files at r1, 2 of 3 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1914)* status: all files reviewed, 2 unresolved discussions (waiting on @acud and @janos)

aloknerurkar (approved):

acud (approved):
:lgtm:

> Reviewed 2 of 4 files at r1, 2 of 3 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1914#-:-MbAtvIP6DZj1pjUGGWC:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @janos)

janos (approved):

PR #1898, opened 2021-05-28
feat: limit number of light nodes

acud:
This PR limits the number of lightnodes we allow to connect inbound. This is to give some safety level to full nodes operators not to be resource overwhelmed by too many light node requests. This sets the initial limit to 100 above which the node will start kicking away other light nodes to accommodate for the new ones.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1898)

PR #1917, opened 2021-05-29
perf(pullsync): release reader buffer in handler

janos:
After profiling existing goroutines running on a node, it is found that pullsync.Syncer.handler goroutine number is larger with the number of connected peers. One approach to optimize was to reuse the iterators but the number of handler goroutines is still high as every peer needs to have a stream opened for pull syncing, most notably live pull syncing for proximity bins.

Profiling shows that a larger meemory allocated at protobuf reader

```
(pprof) top -cum
Showing nodes accounting for 134.09MB, 23.39% of 573.29MB total
Dropped 278 nodes (cum <= 2.87MB)
Showing top 10 nodes out of 205
      flat  flat%   sum%        cum   cum%
         0     0%     0%   221.12MB 38.57%  github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).SetStreamHandlerMatch.func1
         0     0%     0%   205.17MB 35.79%  github.com/ethersphere/bee/pkg/p2p/libp2p.(*Service).AddProtocol.func1
    3.60MB  0.63%  0.63%   154.62MB 26.97%  github.com/ethersphere/bee/pkg/pullsync.(*Syncer).handler
         0     0%  0.63%   128.49MB 22.41%  bufio.NewReader (inline)
  128.49MB 22.41% 23.04%   128.49MB 22.41%  bufio.NewReaderSize (inline)
         0     0% 23.04%   127.47MB 22.24%  github.com/ethersphere/bee/pkg/p2p/protobuf.NewWriterAndReader
         0     0% 23.04%   124.97MB 21.80%  github.com/ethersphere/bee/pkg/p2p/protobuf.NewReader (inline)
       2MB  0.35% 23.39%   124.97MB 21.80%  github.com/gogo/protobuf/io.NewDelimitedReader
         0     0% 23.39%    89.70MB 15.65%  github.com/libp2p/go-libp2p/p2p/protocol/identify.(*IDService).identifyConn
         0     0% 23.39%    87.70MB 15.30%  github.com/libp2p/go-libp2p/p2p/protocol/identify.(*IDService).consumeMessage
```

More precisely on line 263:

```
ROUTINE ======================== github.com/ethersphere/bee/pkg/pullsync.(*Syncer).handler in github.com/ethersphere/bee/pkg/pullsync/pullsync.go
    3.60MB   154.62MB (flat, cum) 26.97% of Total
         .          .    258:	return offer.Topmost, ru.Ruid, nil
         .          .    259:}
         .          .    260:
         .          .    261:// handler handles an incoming request to sync an interval
         .          .    262:func (s *Syncer) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) (err error) {
         .   126.47MB    263:	w, r := protobuf.NewWriterAndReader(stream)
         .          .    264:	defer func() {
         .          .    265:		if err != nil {
         .          .    266:			_ = stream.Reset()
         .          .    267:		} else {
         .          .    268:			_ = stream.FullClose()
         .          .    269:		}
         .          .    270:	}()
         .          .    271:	var ru pb.Ruid
         .        3MB    272:	if err := r.ReadMsgWithContext(ctx, &ru); err != nil {
         .          .    273:		return fmt.Errorf("send ruid: %w", err)
         .          .    274:	}
         .          .    275:
         .     5.50MB    276:	ctx, cancel := context.WithCancel(ctx)
         .          .    277:	s.ruidMtx.Lock()
    1.10MB     1.10MB    278:	s.ruidCtx[ru.Ruid] = cancel
         .          .    279:	s.ruidMtx.Unlock()
    2.50MB     2.50MB    280:	cc := make(chan struct{})
         .          .    281:	defer close(cc)
         .          .    282:	go func() {
         .          .    283:		select {
         .          .    284:		case <-s.quit:
         .          .    285:		case <-ctx.Done():
         .          .    286:		case <-cc:
         .          .    287:		}
         .          .    288:		cancel()
         .          .    289:		s.ruidMtx.Lock()
         .          .    290:		delete(s.ruidCtx, ru.Ruid)
         .          .    291:		s.ruidMtx.Unlock()
         .          .    292:	}()
         .          .    293:
         .          .    294:	select {
         .          .    295:	case <-s.quit:
         .          .    296:		return nil
         .          .    297:	default:
         .          .    298:	}
         .          .    299:
         .          .    300:	s.wg.Add(1)
         .          .    301:	defer s.wg.Done()
         .          .    302:
         .          .    303:	var rn pb.GetRange
         .     1.50MB    304:	if err := r.ReadMsgWithContext(ctx, &rn); err != nil {
         .          .    305:		return fmt.Errorf("read get range: %w", err)
         .          .    306:	}
         .          .    307:
         .          .    308:	// make an offer to the upstream peer in return for the requested range
         .     7.50MB    309:	offer, _, err := s.makeOffer(ctx, rn)
         .          .    310:	if err != nil {
         .          .    311:		return fmt.Errorf("make offer: %w", err)
         .          .    312:	}
         .          .    313:
         .          .    314:	if err := w.WriteMsgWithContext(ctx, offer); err != nil {
         .          .    315:		return fmt.Errorf("write offer: %w", err)
         .          .    316:	}
         .          .    317:
         .          .    318:	// we don't have any hashes to offer in this range (the
         .          .    319:	// interval is empty). nothing more to do
         .          .    320:	if len(offer.Hashes) == 0 {
         .          .    321:		return nil
         .          .    322:	}
         .          .    323:
         .          .    324:	var want pb.Want
         .          .    325:	if err := r.ReadMsgWithContext(ctx, &want); err != nil {
         .          .    326:		return fmt.Errorf("read want: %w", err)
         .          .    327:	}
         .          .    328:
         .     7.05MB    329:	chs, err := s.processWant(ctx, offer, &want)
         .          .    330:	if err != nil {
         .          .    331:		return fmt.Errorf("process want: %w", err)
         .          .    332:	}
         .          .    333:
         .          .    334:	for _, v := range chs {
```

The reader is allocating a buffer of 4096 bytes (bufio.Reader) and holding it from Go's CG while waiting for new chunks at line 329 on `makeOffer` call which is blocking until there are new chunks (expected behaviour). The allocated reader is not and could be released, but it is not as it can be used further down the function execution path. Wit higher number of goroutines these buffers take more memory, which is not needed during `makeOffer` call.

This PR allows for the reader to be garbage collected as a new one is created after the `makeOffer` call. This approach creates additional allocation of 4096 bytes, but allows Go's GC to garbage collect the buffer and reduce the total allocated memory, resulting the handler goroutine allocations to look like this:

```
ROUTINE ======================== github.com/ethersphere/bee/pkg/pullsync.(*Syncer).handler in github.com/ethersphere/bee/pkg/pullsync/pullsync.go
  512.05kB     8.03MB (flat, cum)  7.06% of Total
         .          .    267:		} else {
         .          .    268:			_ = stream.FullClose()
         .          .    269:		}
         .          .    270:	}()
         .          .    271:	var ru pb.Ruid
         .   512.05kB    272:	if err := r.ReadMsgWithContext(ctx, &ru); err != nil {
         .          .    273:		return fmt.Errorf("send ruid: %w", err)
         .          .    274:	}
         .          .    275:
         .        1MB    276:	ctx, cancel := context.WithCancel(ctx)
         .          .    277:	s.ruidMtx.Lock()
         .          .    278:	s.ruidCtx[ru.Ruid] = cancel
         .          .    279:	s.ruidMtx.Unlock()
  512.05kB   512.05kB    280:	cc := make(chan struct{})
         .          .    281:	defer close(cc)
         .          .    282:	go func() {
         .          .    283:		select {
         .          .    284:		case <-s.quit:
         .          .    285:		case <-ctx.Done():
         .          .    286:		case <-cc:
         .          .    287:		}
         .          .    288:		cancel()
         .          .    289:		s.ruidMtx.Lock()
         .          .    290:		delete(s.ruidCtx, ru.Ruid)
         .          .    291:		s.ruidMtx.Unlock()
         .          .    292:	}()
         .          .    293:
         .          .    294:	select {
         .          .    295:	case <-s.quit:
         .          .    296:		return nil
         .          .    297:	default:
         .          .    298:	}
         .          .    299:
         .          .    300:	s.wg.Add(1)
         .          .    301:	defer s.wg.Done()
         .          .    302:
         .          .    303:	var rn pb.GetRange
         .        1MB    304:	if err := r.ReadMsgWithContext(ctx, &rn); err != nil {
         .          .    305:		return fmt.Errorf("read get range: %w", err)
         .          .    306:	}
         .          .    307:
         .          .    308:	// make an offer to the upstream peer in return for the requested range
         .   512.02kB    309:	offer, _, err := s.makeOffer(ctx, rn)
         .          .    310:	if err != nil {
         .          .    311:		return fmt.Errorf("make offer: %w", err)
         .          .    312:	}
         .          .    313:
         .          .    314:	w, r := protobuf.NewWriterAndReader(stream)
         .          .    315:
         .          .    316:	if err := w.WriteMsgWithContext(ctx, offer); err != nil {
         .          .    317:		return fmt.Errorf("write offer: %w", err)
         .          .    318:	}
         .          .    319:
         .          .    320:	// we don't have any hashes to offer in this range (the
         .          .    321:	// interval is empty). nothing more to do
         .          .    322:	if len(offer.Hashes) == 0 {
         .          .    323:		return nil
         .          .    324:	}
         .          .    325:
         .          .    326:	var want pb.Want
         .          .    327:	if err := r.ReadMsgWithContext(ctx, &want); err != nil {
         .          .    328:		return fmt.Errorf("read want: %w", err)
         .          .    329:	}
         .          .    330:
         .     4.53MB    331:	chs, err := s.processWant(ctx, offer, &want)
         .          .    332:	if err != nil {
         .          .    333:		return fmt.Errorf("process want: %w", err)
         .          .    334:	}
         .          .    335:
         .          .    336:	for _, v := range chs {
```

No allocations are held by the first created reader.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1917)

### DIFF
--- a/pkg/pullsync/pullsync.go
+++ b/pkg/pullsync/pullsync.go
@@ -260,7 +260,7 @@ func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8
 
 // handler handles an incoming request to sync an interval
 func (s *Syncer) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) (err error) {
-	w, r := protobuf.NewWriterAndReader(stream)
+	r := protobuf.NewReader(stream)
 	defer func() {
 		if err != nil {
 			_ = stream.Reset()
@@ -310,6 +310,11 @@ func (s *Syncer) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) (er
 	if err != nil {
 		return fmt.Errorf("make offer: %w", err)
 	}
+
+	// recreate the reader to allow the first one to be garbage collected
+	// before the makeOffer function call, to reduce the total memory allocated
+	// while makeOffer is executing (waiting for the new chunks)
+	w, r := protobuf.NewWriterAndReader(stream)
 
 	if err := w.WriteMsgWithContext(ctx, offer); err != nil {
 		return fmt.Errorf("write offer: %w", err)


acud (approved):
:pray: great stuff @janos. Thanks!
Just to confirm, the PR says you are releasing the reader buffer but from the changeset it seems that it is the writer that is just being created later. So out of curiosity, I did not really understand how is it that the reader can be garbage collected. Is it a compiler optimization that makes the first reader be GCd?

Isn't it so that while `makeOffer` executes, since the `handler` stack frame still exists, then the original reader is still referenced from the stack frame, therefore to have it GCd wouldn't you need to explicitly set it to `nil` before `makeOffer`?

zelig (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1917)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aloknerurkar)

aloknerurkar (approved):

esadakar (approved):

esadakar:
this PR taught me how to use pprof better 👍 thanks!

janos:
Thank you all for reviews.

I did some more investigation on high memory consumption, mainly on bootnodes. And I've found two more sources of it. I am if I am writing here something that is already known or even being worked on.

### 1. Unlimited lightnode peers

There is no limit on how many lightondes are connected to a node. This would not be so bad if libp2p identify protocol could be disabled. It actually does not do anything crucial and it was discussed for removal in libp2p 1.0. Another point is that every lightonde is creating a Syncer to every full node. As there is no limit on connected light nodes, a full node calls a large number of pullsync handers. 

I am suggesting to limit the number of connected light nodes.

I was able to prune the lightondes on a node by issuing something like this:

```bash
curl http://bee-8-debug.bee-bootnode.core.internal/peers | jq -r ".peers[] | select( .fullNode == false ) | .address" | { while read line; do curl -XDELETE http://bee-8-debug.bee-bootnode.core.internal/peers/$line; done }
```

to validate changes in the used memory consumption and number of goroutines.

### 2. Kademlia knownPeers PSlice

It grows a bit large on bootnodes because of high number of peers connecting. As I can see it is around 280000 peers on bootnodes currently.

Kademlia should probably not load the complete addressbook on start and have some limit on knownPeers PSlice.

One side effect is also a pretty large /topology response, but I did not have any problems with it.

-----

Thanks @vandot for deploying this branch to bee-8 and bee-9 on bootnode namespace.

@acud Thanks.
> Is it a compiler optimization that makes the first reader be GCd?

That is correct. Since the first reader is no longer referenced by the shadowing of the later one (it does not have to be shadowed, other variable name would do the same) the first reader can be garbage collected during the long running makeOffer call.

> to have it GCd wouldn't you need to explicitly set it to nil before makeOffer?

There is no need to do that even if there is no need for the second reader. GC will find it unused.

@esadakar I am very happy to hear that.

significance:
@janos you can see  280000 live nodes?

janos:
> @janos you can see 280000 live nodes?

Actually, this is the number that a bootnode was aware of at some point and showing them in the /topology response of a bootnode. It is very likely that many of them do no longer exist. Some nodes have lower number of 160k, for example, it varies.

acud:
Thanks for the explanation @janos :pray: 
About limiting light node number, I already have an open PR on that from last week [here](https://github.com/ethersphere/bee/pull/1898) :)

janos:
> Thanks for the explanation @janos 🙏
> About limiting light node number, I already have an open PR on that from last week [here](https://github.com/ethersphere/bee/pull/1898) :)

Thanks @acud. I missed that one. :)

ldeffenb:
A large number of known nodes in the /topology response isn't unique to bootnodes.  My 0.6 nodes have anywhere from 104,205 through 130,000's up to 180,152.  And these address book sizes result in positively HUGE /topology API query responses.

I'm currently accessing my nodes via a VPN from a resort WiFi and just pulling the /topology response can take 10-15 seconds and >10Mbyte of data.

Issue #1819, opened 2021-05-21
Price oracle for bandwidth and initial costs


Eknir:
# Intro
There is a need for having a price oracle that quotes the exchange rate between honey (accounting credits) and BZZ. Nodes query this value before sending/receiving a cheque and hence value the BZZ cheque at a certain honey-worth. 

The price oracle is a smart-contract on the blockchain that may internally gets its price from the bonding curve or another source. The maintenance of the price oracle is done by an Ethereum smart-contract, which initially is controlled by a set of trusted keys, but may later transition to being controlled by BZZ token holders.

Initially, a price oracle was not great to have, as services that were rendered in the past might have a radically different value at the moment of settlement. With time based settlement, it is much more likely that a settlement comes at the time when the services are delivered.

# Architecture
## Smart-contract
Exposes:
- Exchange rate
        - How many accounting units does 1 PLUR give me?
 - Initial cost
        - Deducted from first payment from a peer *ever*. Reject payment if less.

## Reading price in bee
There is a service that polls the contract each X minutes sharp. All nodes poll at the same time. If there are minor time differences and the block with the price update is mined at the time that nodes poll, it may happen that part of the network accounts with a different price for X minutes.

## Using price information
The price information is used when sending a cheque:

1. Settler sends message header to receiver, signalling it wants to monetary-settle
2. Receiver returns current exchange rate and deduction if applicable
3. Settler validates exchange rate and deduction with it's own version of the truth
    - If corresponding
        - Settler issue cheque to Abel and updates accounting if write succeeds
        - Receiver receives cheque and updates accounting
    - If not corresponding
        - Settler reset stream

- consider timeout of streams
    - We may want to prevent a node to indefinitely keep a stream open and engage in option-buying behavior (may also be as an update at a later time)
        - timeout
        - If the receiving party, based on their version of time, receives the cheque too long after the exchange rate was issues => apply current exchange rate

acud:
this is already done and live on xdai

PR #1924, opened 2021-05-31
feat: Negotiate cheques

metacertain:
https://github.com/ethersphere/bee/issues/1819

The pr consists of three main parts:
- the priceoracle, polling the current values of the exchange rate and initial deduction at node start and later in every unix timestamp divisible by 300 (meaning every 5 minutes at approximately the same time on all nodes).
- the exchange rate and applicable deduction negotiation happening before issuing a cheque to a peer.
- converting between the amount of accounting credits attempted to be settled, and the amount of plur sent to cover for this settlement using the exchange rate learned from the price oracle / negotiated in swap protocol

The negotiation process consists of:
- the peer willing to issue a cheque begins opening a stream sending empty headers
- the accepting peer returns it's current known exchange rate, and the applicable deduction in headers
	+ the applicable deduction is the current known deduction, or 0 if the peer attempting to settle has already sent a cheque with negotiated deduction previously
- the peer compares the received exchange rate and deduction headers to his version of known truth, if both match, issues a cheque, otherwise cancels the cheque issuing and restrains from further attempts to send cheques to this peer for 10 seconds, to limit rate of errors in negative scenarios
- if the deduction was negotiated non zero, both peers save the other as already having a deduction sent to / received from, so that in the future they are able to agree on zero deduction based on this registry.

Negative scenarios can be:
- if there was an update of the oracle exchange rate or deduction between polling point A and A+1, any node started after the the point of update but before A+1 can be ahead of the network in his version of the truth until A+1.
- if there was an update of the rate or deduction, near the polling point when one of the peers might be ahead by getting or using the new version of truth sooner / faster 

The conversion logic between accounting credits and plur:
Cheque value is ( accounting amount * exchange rate ) + deduction
Thereby received/sent cheque is calculated as changing the balance with ( cheque amount - deduction ) / exchange rate

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1924)


zelig (commented):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1924)* status: all files reviewed, 1 unresolved discussion (waiting on @metacertain)

___
*[pkg/settlement/swap/swapprotocol/swapprotocol.go, line 208 at r1](https://reviewable.io/reviews/ethersphere/bee/1924#-Mb-zCwL0aQyNBYEHOlO:-Mb-zCwM56SRUvhYqP78:b-25tpd5) ([raw file](https://github.com/ethersphere/bee/blob/c58e21dbf91740e8baefe0e264215d2aa574d96c/pkg/settlement/swap/swapprotocol/swapprotocol.go#L208)):*
> ```Go
> 	checkPeer := false
> 
> 	if checkPeer {
> ```

always false

metacertain:
> > Reviewed 7 of 7 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1924)_ status: all files reviewed, 1 unresolved discussion (waiting on @metacertain)
> 
> _[pkg/settlement/swap/swapprotocol/swapprotocol.go, line 208 at r1](https://reviewable.io/reviews/ethersphere/bee/1924#-Mb-zCwL0aQyNBYEHOlO:-Mb-zCwM56SRUvhYqP78:b-25tpd5) ([raw file](https://github.com/ethersphere/bee/blob/c58e21dbf91740e8baefe0e264215d2aa574d96c/pkg/settlement/swap/swapprotocol/swapprotocol.go#L208)):_
> 
> > ```go
> > 	checkPeer := false
> > 
> > 	if checkPeer {
> > ```
> 
> always false

it is a work in progress, yet to introduce the component for registering whether a peer already had an initial deduction

acud (changes requested):
> Reviewed 1 of 7 files at r1, 29 of 30 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1924)* status: 30 of 31 files reviewed, 15 unresolved discussions (waiting on @metacertain and @zelig)

___
*[.github/workflows/beekeeper.yml, line 17 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-Mbpn_Jt4nbZmP5rQYLX:-Mbpn_Jt4nbZmP5rQYLY:b-58so6z) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/.github/workflows/beekeeper.yml#L17)):*
> ```YAML
>       RUN_TYPE: "PR RUN"
>       SETUP_CONTRACT_IMAGE_TAG: "0.1.0"
>       BEEKEEPER_BRANCH: "thresholds"
> ```

needed?
___
*[pkg/accounting/accounting.go, line 32 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-MbpoFGq9Ef2Pg3cxwkG:-MbpoFGr4j7qUI-UfEdL:bgn733f) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/accounting/accounting.go#L32)):*
> ```Go
> 	// this value is chosen so that tiny payments are prevented while still allowing small payments in environments with lower payment thresholds
> 	minimumPaymentDivisor    = int64(5)
> 	failedSettlementInterval = int64(10)
> ```

ten what? s/m/d/m/y? 
EDIT: seconds.
comment would be nice
___
*[pkg/accounting/accounting.go, line 333 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-Mbpoz980FATMorwmotf:-Mbpoz980FATMorwmotg:bvwdtwn) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/accounting/accounting.go#L333)):*
> ```Go
> 				// add settled amount to shadow reserve before sending it
> 				balance.shadowReservedBalance.Add(balance.shadowReservedBalance, paymentAmount)
> 				go a.payFunction(context.Background(), peer, paymentAmount)
> ```

there is no goroutine tracking here for shutdown, which means that payments made during shutdown might be lost
___
*[pkg/accounting/accounting_test.go, line 1100 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-MbppOlQCQUXOIEk06_H:-MbppOlQCQUXOIEk06_I:b7hxpzw) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/accounting/accounting_test.go#L1100)):*
> ```Go
> 	}
> 
> 	acc.Release(peer1Addr, 1)
> ```

is this needed at the end of the test?
___
*[pkg/settlement/swap/swap.go, line 101 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-MbpqbzN55XxvC_Xauw9:-MbpqbzN55XxvC_XauwA:b-7zqllp) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/settlement/swap/swap.go#L101)):*
> ```Go
> 
> 	if deduction.Cmp(big.NewInt(0)) > 0 {
> 		err = s.addressbook.AddDeductionFor(peer)
> ```

is this operation made under a peer lock? there might be a data race hiding here
___
*[pkg/settlement/swap/chequebook/chequestore_test.go, line 447 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-Mbptyzy0Q4Z3xeacWsI:-Mbptyzz-ErlU44DrnYG:b-c2omw4) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/settlement/swap/chequebook/chequestore_test.go#L447)):*
> ```Go
> 	sig := make([]byte, 65)
> 	chainID := int64(1)
> 	exchangeRate := big.NewInt(100)
> ```

can we have a short comment explaining the arithmetics and why the error should be returned?
___
*[pkg/settlement/swap/headers/utilities.go, line 24 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-MbpuJYn7aLDN0J8jJaB:-MbpuJYn7aLDN0J8jJaC:b-wqil8) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/settlement/swap/headers/utilities.go#L24)):*
> ```Go
> 	// ErrNoExchangeHeader denotes p2p.Header lacking specified field
> 	ErrNoExchangeHeader = errors.New("no exchange header")
> 	// ErrNoTargetHeader denotes p2p.Header lacking specified field
> ```

`ErrWrongComment`
___
*[pkg/settlement/swap/headers/utilities.go, line 45 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-MbpuVKV1aC9osf5uzam:-MbpuVKV1aC9osf5uzan:bepfk84) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/settlement/swap/headers/utilities.go#L45)):*
> ```Go
> 	deduction, err = ParseDeductionHeader(receivedHeaders)
> 	if err != nil {
> 		return exchangeRate, nil, err
> ```

Nit, but since the only error that can be returned from `ParseDeductionHeader` is `ErrNoDeductionHeader`, and since the calling contexts always fall back to deduction value `0`, then maybe it is not needed to return the error and just return a `0` from `ParseDeductionHeader`. However having it explicit as it is might be a nice option too, since it makes things a bit clearer in the calling contexts
___
*[pkg/settlement/swap/priceoracle/priceoracle.go, line 38 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-MbpvzNIDA_Z58kkefhb:-MbpvzNJ7CEDYFYWLHr4:b-v7rqmh) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/settlement/swap/priceoracle/priceoracle.go#L38)):*
> ```Go
> type Service interface {
> 	io.Closer
> 	// Deposit starts depositing erc20 token into the chequebook. This returns once the transactions has been broadcast.
> ```

wrong comment, missing comments on other methods in the interface
___
*[pkg/settlement/swap/priceoracle/priceoracle.go, line 45 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-Mbpw4QK7h1AMilBVroJ:-Mbpw4QK7h1AMilBVroK:bsi76ij) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/settlement/swap/priceoracle/priceoracle.go#L45)):*
> ```Go
> 
> var (
> 	priceOracleABI = transaction.ParseABIUnchecked(priceoracleabi.PriceOracleABIv0_1_0)
> ```

I think there's a `v0_2_0`
___
*[pkg/settlement/swap/priceoracle/priceoracle.go, line 81 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-MbpwrqW1Qq-QvPtOa9C:-MbpwrqW1Qq-QvPtOa9D:b-xbcvgc) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/settlement/swap/priceoracle/priceoracle.go#L81)):*
> ```Go
> 			ts := time.Now().Unix()
> 
> 			timeUntilNextPoll := time.Duration(s.timeDivisor-ts%s.timeDivisor) * time.Second
> ```

why not use a `const`?
___
*[pkg/settlement/swap/priceoracle/priceoracle.go, line 92 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-Mbpww-rERRu9lY3apZy:-Mbpww-rERRu9lY3apZz:b-ql0r4i) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/settlement/swap/priceoracle/priceoracle.go#L92)):*
> ```Go
> }
> 
> func (s *service) GetPrice(ctx context.Context) (*big.Int, *big.Int, error) {
> ```

since this wrapping method does not provide any further functionality, it can be removed and instead the unexported one can be exported
___
*[pkg/settlement/swap/priceoracle/priceoracle_test.go, line 22 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-Mbpx_rE1GEn4wJNKGFz:-Mbpx_rE1GEn4wJNKGG-:b-yveou2) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/settlement/swap/priceoracle/priceoracle_test.go#L22)):*
> ```Go
> 
> var (
> 	priceOracleABI = transaction.ParseABIUnchecked(priceoracleabi.PriceOracleABIv0_1_0)
> ```

here too, v2?
___
*[pkg/settlement/swap/swapprotocol/swapprotocol_test.go, line 31 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-Mbpxu09DxpwiLUWaKAh:-Mbpxu09DxpwiLUWaKAi:bguvgc7) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/settlement/swap/swapprotocol/swapprotocol_test.go#L31)):*
> ```Go
> )
> 
> func TestInit(t *testing.T) {
> ```

it would be good to document these tests also with a godoc style comment and also inline comments within the tests as they are not self describing and I find them hard to review

ralph-pichler:
___
*[pkg/settlement/swap/priceoracle/priceoracle.go, line 45 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-Mbpw4QK7h1AMilBVroJ:-MbqDdg368X5UNuzkrXK:bxkv7p4) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/settlement/swap/priceoracle/priceoracle.go#L45)):*

I think there's a `v0_2_0`
</blockquote></details>

that's a different price oracle, for this one there is only one version.

metacertain:
> >       BEEKEEPER_BRANCH: "thresholds"
> > ```
> 
> needed?

yes, this reduces thresholds by the magnitude of the newly introduced exchange rates
the beekeeper PR containing this should be merged at the same time as this to avoid breaks

> > 	failedSettlementInterval = int64(10)
> > ```
> 
> ten what? s/m/d/m/y?
> EDIT: seconds.

seconds indeed, added comment

> _[pkg/accounting/accounting.go, line 333 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-Mbpoz980FATMorwmotf:-Mbpoz980FATMorwmotg:bvwdtwn) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/accounting/accounting.go#L333)):_
> 
> > ```go
> > 				// add settled amount to shadow reserve before sending it
> > 				balance.shadowReservedBalance.Add(balance.shadowReservedBalance, paymentAmount)
> > 				go a.payFunction(context.Background(), peer, paymentAmount)
> > ```
> 
> there is no goroutine tracking here for shutdown, which means that payments made during shutdown might be lost

still to be fixed

> > 	acc.Release(peer1Addr, 1)
> > ```
> 
> is this needed at the end of the test?

we always do it consistently, not strictly needed

> > 		err = s.addressbook.AddDeductionFor(peer)
> > ```
> 
> is this operation made under a peer lock? there might be a data race hiding here

no need for lock in this case. this adds an empty struct to the specified address-index
if it is overwritten by another empty struct, wouldn't cause a problem
this is to remember deduction happened once towards a specific peer.
only one payment is ongoing at a time so no other payment is attempted without this saved yet

> > 	exchangeRate := big.NewInt(100)
> > ```
> 
> can we have a short comment explaining the arithmetics and why the error should be returned?

added comment to test
 
> `ErrWrongComment`

:)
 
> 
> Nit, but since the only error that can be returned from `ParseDeductionHeader` is `ErrNoDeductionHeader`, and since the calling contexts always fall back to deduction value `0`, then maybe it is not needed to return the error and just return a `0` from `ParseDeductionHeader`. However having it explicit as it is might be a nice option too, since it makes things a bit clearer in the calling contexts

this should be fine the way it is, than?
 
> > 	// Deposit starts depositing erc20 token into the chequebook. This returns once the transactions has been broadcast.
> > ```
> 
> wrong comment, missing comments on other methods in the interface

changed comment

 
> > var (
> > 	priceOracleABI = transaction.ParseABIUnchecked(priceoracleabi.PriceOracleABIv0_1_0)
> > ```
> 
> I think there's a `v0_2_0`

thats the other ABI for postage stamps, v0_1_0 is good

> > 			timeUntilNextPoll := time.Duration(s.timeDivisor-ts%s.timeDivisor) * time.Second
> > ```
> 
> why not use a `const`?

added comment explaining logic
 
> > func (s *service) GetPrice(ctx context.Context) (*big.Int, *big.Int, error) {
> > ```
> 
> since this wrapping method does not provide any further functionality, it can be removed and instead the unexported one can be exported

done
 
> > 	priceOracleABI = transaction.ParseABIUnchecked(priceoracleabi.PriceOracleABIv0_1_0)
> > ```
> 
> here too, v2?

v1 is ok

> > func TestInit(t *testing.T) {
> > ```
> 
> it would be good to document these tests also with a godoc style comment and also inline comments within the tests as they are not self describing and I find them hard to review

added comments

acud (approved):
LGTM abel. Thanks. :+1: 

> Reviewed 2 of 2 files at r3, 4 of 5 files at r4, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1924#-:-MbupF-V-gziUQ0I8LaT:brwroh3)* status: all files reviewed, 1 unresolved discussion (waiting on @zelig)

___
*[pkg/accounting/accounting.go, line 333 at r2](https://reviewable.io/reviews/ethersphere/bee/1924#-Mbpoz980FATMorwmotf:-MbuoqNa3c99mTk0RqFs:b-idflxc) ([raw file](https://github.com/ethersphere/bee/blob/6281c66051a3646047337747f5c06d651041d7b6/pkg/accounting/accounting.go#L333)):*
<details><summary><i>Previously, metacertain wrote…</i></summary><blockquote>

> > ```go
> > 				// add settled amount to shadow reserve before sending it
> > 				balance.shadowReservedBalance.Add(balance.shadowReservedBalance, paymentAmount)
> > 				go a.payFunction(context.Background(), peer, paymentAmount)
> > ```
> > 

> 
> there is no goroutine tracking here for shutdown, which means that payments made during shutdown might be lost
> 

still to be fixed


> > 	acc.Release(peer1Addr, 1)
> > ```
> > 

> 
> is this needed at the end of the test?
> 

we always do it consistently, not strictly needed


> > 		err = s.addressbook.AddDeductionFor(peer)
> > ```
> > 

> 
> is this operation made under a peer lock? there might be a data race hiding here
> 

no need for lock in this case. this adds an empty struct to the specified address-index
if it is overwritten by another empty struct, wouldn't cause a problem
this is to remember deduction happened once towards a specific peer.
only one payment is ongoing at a time so no other payment is attempted without this saved yet


> > 	exchangeRate := big.NewInt(100)
> > ```
> > 

> 
> can we have a short comment explaining the arithmetics and why the error should be returned?
> 

added comment to test
 


> `ErrWrongComment`
> 

:)
 


> 
> Nit, but since the only error that can be returned from `ParseDeductionHeader` is `ErrNoDeductionHeader`, and since the calling contexts always fall back to deduction value `0`, then maybe it is not needed to return the error and just return a `0` from `ParseDeductionHeader`. However having it explicit as it is might be a nice option too, since it makes things a bit clearer in the calling contexts
> 

this should be fine the way it is, than?
 


> > 	// Deposit starts depositing erc20 token into the chequebook. This returns once the transactions has been broadcast.
> > ```
> > 

> 
> wrong comment, missing comments on other methods in the interface
> 

changed comment

 


> > var (
> > 	priceOracleABI = transaction.ParseABIUnchecked(priceoracleabi.PriceOracleABIv0_1_0)
> > ```
> > 

> 
> I think there's a `v0_2_0`
> 

thats the other ABI for postage stamps, v0_1_0 is good


> > 			timeUntilNextPoll := time.Duration(s.timeDivisor-ts%s.timeDivisor) * time.Second
> > ```
> > 

> 
> why not use a `const`?
> 

added comment explaining logic
 


> > func (s *service) GetPrice(ctx context.Context) (*big.Int, *big.Int, error) {
> > ```
> > 

> 
> since this wrapping method does not provide any further functionality, it can be removed and instead the unexported one can be exported
> 

done
 


> > 	priceOracleABI = transaction.ParseABIUnchecked(priceoracleabi.PriceOracleABIv0_1_0)
> > ```
> > 

> 
> here too, v2?
> 

v1 is ok


> > func TestInit(t *testing.T) {
> > ```
> > 

> 
> it would be good to document these tests also with a godoc style comment and also inline comments within the tests as they are not self describing and I find them hard to review
> 

added comments

</blockquote></details>

thanks.

ldeffenb:
Is this a breaking change due to the new protocol streams?  In other words, if I update to the current master, can I still participate in the 0.6.2 swarm?

PR #1926, opened 2021-05-31
refactor: improve shutdown sequence

acud:
This PR improves the shutdown sequence by introducing a few changes to the way we do shutdowns:
1. kademlia and libp2p are now `Halt`ed before anything else happens in the shutdown, this will prevent _new_ outgoing and incoming connections to be established, however existing connections will remain untouched. This will also improve exit time of kademlia when it is finally shut down since the number of ongoing goroutines dailing will be smaller, so we would need to wait less.
2. more components are shut down in parallel
3. api shutdown timeout is decreased, since shutting down with running websockets is not critical, while slowing down the rest of the process _is_
4. kademlia `Close` is bumped _before_ statestore is closed. Since metrics need to be saved before shutting down, this can result in a panic
5. eth dependent components are shut down before the eth client is closed. On shutdown, it can result in an unnecessary panic right now, since the eth client might close while the listener is querying the backend, resulting in an error, therefore calling `Shutdown` on `node.go` twice, which might result in a panic
6. explicit check for `context.Cancelled` is therefore needed in `listener.go`, so that shutdown is not called twice

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1926)

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -157,7 +157,7 @@ func (s *server) Close() error {
 
 	select {
 	case <-done:
-	case <-time.After(5 * time.Second):
+	case <-time.After(1 * time.Second):
 		return errors.New("api shutting down with open websockets")
 	}
 
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"syscall"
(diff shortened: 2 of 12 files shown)


esadakar (approved):
Nice!

aloknerurkar (approved):

janos (approved):
Nice!

PR #1964, opened 2021-06-02
fix: added "format: int64" where required in openapi

### LINKED ISSUE #1726: Use "format: int64" into open api specification for integers where required
#### Summary
Currently OpenAPI specification declare some types as integers, but never declare the format, also where would be required.  
It's the case for example of `Cheque.payout` property schema, declared in `SwarmCommon.yaml` as.

```
    Cheque:
      type: object
      properties:
        beneficiary:
(...)
        payout:
          type: integer
```

where it should be:

```
    Cheque:
      type: object
      properties:
        beneficiary:
(...)
        payout:
          type: integer
          format: int64
```

Missing of `format: int64` implies that expected number will be an `int32` value. This lead for example into generating invalid client code, where this property is implemented as an `int` value instead of a `long`, generating errors on response parsing.

#### Expected behavior
Where API uses long numbers, schema should declare it as

```
type: integer
format: int64
```

#### Actual behavior
There is no declared differences between `int32` and `int64` values
```
type: integer
```

tmm360:
Fixes #1726.

I think to have identified all int64 values in open api definition, but please verify.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1964)

### DIFF
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -796,6 +796,7 @@ paths:
           name: amount
           schema:
             type: integer
+            format: int64
           required: true
           description: Amount of BZZ added that the postage batch will have.
         - in: path
@@ -824,4 +825,3 @@ paths:
           $ref: "SwarmCommon.yaml#/components/responses/500"
         default:
           description: Default response
-
--- a/openapi/SwarmCommon.yaml
+++ b/openapi/SwarmCommon.yaml
@@ -41,20 +41,26 @@ components:
           type: integer
         available:
           type: integer
+          format: int64
         outer:
           type: integer
+          format: int64
         inner:
           type: integer
+          format: int64
 
     ChainState:
       type: object
       properties:
         block:
           type: integer
+          format: int64
         totalAmount:
           type: integer
+          format: int64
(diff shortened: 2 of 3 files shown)


acud:
@tmm360 I think that after @AuHau's PR got merged some of these changes are no longer relevant. The types are `big.Int`s and actually `int64` overflows quite quickly with on-chain interactions, that is why our serialization has been converted to stringify the values instead of using the native javascript number that even overflows faster than int64 if I remember correctly. The general OpenAPI integer type is unbounded and probably should be kept this way so that users know that it is unbounded

tmm360:
I see... Maybe output strings from API should be better in these cases? The point is that a clear and complete API specification can be used for generate an maintain clients. I'm using OpenAPI documentation for maintain [Bee.Net](https://github.com/Etherna/bee-net), but I've to fix and merge any new version. If a clear and universal API specification can be defined, I think that should be really better. For me and others that have to implement clients.

acud:
This has been addressed in a separate PR :+1: The numbers are now represented as strings. Thanks for pointing this out :pray:

PR #1976, opened 2021-06-03
fix: check cancelled tx in checkPending

deaswang:
if transaction cancelled. old code won't check the transaction, and cancelledTxs process won't take effect.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1976)

### DIFF
--- a/pkg/transaction/monitor.go
+++ b/pkg/transaction/monitor.go
@@ -162,13 +162,13 @@ type confirmedTx struct {
 	watch   *transactionWatch
 }
 
-// potentiallyConfirmedWatches returns all watches with nonce less than what was specified
+// potentiallyConfirmedWatches returns all watches with nonce less than or equal what was specified
 func (tm *transactionMonitor) potentiallyConfirmedWatches(nonce uint64) (watches []*transactionWatch) {
 	tm.lock.Lock()
 	defer tm.lock.Unlock()
 
 	for watch := range tm.watches {
-		if watch.nonce < nonce {
+		if watch.nonce <= nonce {
 			watches = append(watches, watch)
 		}
 	}


pkg/transaction/monitor.go
```diff
@@ -162,13 +162,13 @@ type confirmedTx struct {
...
 
-// potentiallyConfirmedWatches returns all watches with nonce less than what was specified
+// potentiallyConfirmedWatches returns all watches with nonce less than or equal what was specified
 func (tm *transactionMonitor) potentiallyConfirmedWatches(nonce uint64) (watches []*transactionWatch) {
 	tm.lock.Lock()
 	defer tm.lock.Unlock()
 
 	for watch := range tm.watches {
-		if watch.nonce < nonce {
+		if watch.nonce <= nonce {
```
ralph-pichler, line 171: I think this was correct before. The nonce passed in here is the nonce at a block (which is the next nonce not yet used, e.g. 0 if the account was never used, 1 if only the tx with nonce 0 has been sent). So if the in-block nonce is 12, then a transaction with nonce 12 cannot have been included yet.

deaswang: Sometime, the transaction maybe stuck in the tx pool after send to swap endpoint. e.g. send an transaction with nonce 0, then it stuck. NonceAt will return 0. but in the leveldb, there is one transaction with nonce 0 which will not include into checkWatches.
If you deploy lots of nodes with low gas price by infura API, this situation will happen in some nodes.
It is better to develop nonce manage module, you can't believe other's nonce manage.

deaswang: https://medium.com/coinmonks/hyperledger-and-ethereum-nonce-management-890029d481b

ralph-pichler: we don't believe others nonce manager. `NonceAt` gets the real onchain nonce at the last block. If this equals the nonce of the tx we are interested in it cannot possibly have been confirmed or replaced (hence not "cancelled"). We cannot exclude the possibility that it will still be included in the future, hence marking it cancelled would be a mistake. If one ends up in such an uncertain situation it is best to attempt rebroadcasting (using the `POST /transactions/{hash}` endpoint) or cancelling it explicitly (using the `DELETE /transactions/{hash}` endpoint to double spend with a dummy tx).

PR #1983, opened 2021-06-03
feat: Always reset balances to zero after disconnection or blocklisting

metacertain:
This PR introduces a failsafe mechanism for unforeseeable problems that can result in accounting discrepancies and enabling nodes to be able to recover a state of symmetric balances afterwards while inhibiting nodes of taking advantage of the failsafe mechanism.
The failsafe mechanism is that both peers start with 0 balances upon a new establishment of the connection in a peer to peer relationship.
The safety of this mechanism is guaranteed by not allowing reconnections by the same peer until the refreshement rate would remove the outstanding debt plus one times paymentthreshold worth of credit space, in case a peer disconnects from us.

This pr also introduces ghost debt, meaning we account for requests that the node attempted to serve as a forwarder - meaning it successfully requested it and got a response for the same request from a third peer - but, for some reason, was unable to write back the result and thereby could not actually debit for these 'ghosted' requests. 

Once the cumulative amount of ghost balance reaches the disconnect limit, the peer in ghost debt is blocklisted by the node based on the time the refreshement rate would take to remove the outstanding debt plus one times paymentthreshold worth of credit space.

For all such blocklisting purposes, the outstanding debt is defined as the sum of:
- debt shown by the accounting balance, 
- shadow reserved debt (requests already served by downstream, which could be debited for once response was wrote back to upstream peer) 
- ghost debt

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/1983)

### DIFF
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -1037,9 +1037,24 @@ func (a *Accounting) blocklist(peer swarm.Address, multiplier int64) error {
 
 func (a *Accounting) Connect(peer swarm.Address) {
 	accountingPeer := a.getAccountingPeer(peer)
+	zero := big.NewInt(0)
 
 	accountingPeer.lock.Lock()
 	defer accountingPeer.lock.Unlock()
+
+	accountingPeer.shadowReservedBalance.Set(zero)
+	accountingPeer.ghostBalance.Set(zero)
+	accountingPeer.reservedBalance.Set(zero)
+
+	err := a.store.Put(peerBalanceKey(peer), zero)
+	if err != nil {
+		a.logger.Errorf("failed to persist balance: %w", err)
+	}
+
+	err = a.store.Put(peerSurplusBalanceKey(peer), zero)
+	if err != nil {
+		a.logger.Errorf("failed to persist surplus balance: %w", err)
+	}
 
 	if accountingPeer.reconnectAllowTimestamp != 0 {
 		timeNow := a.timeNow().Unix()
@@ -1101,27 +1116,8 @@ func (a *Accounting) decreaseOriginatedBalanceBy(peer swarm.Address, amount *big
 func (a *Accounting) Disconnect(peer swarm.Address) {
 	accountingPeer := a.getAccountingPeer(peer)
 
-	zero := big.NewInt(0)
 	accountingPeer.lock.Lock()
-	defer func() {
-
-		accountingPeer.shadowReservedBalance.Set(zero)
-		accountingPeer.ghostBalance.Set(zero)
-		accountingPeer.reservedBalance.Set(zero)
-
-		err := a.store.Put(peerBalanceKey(peer), zero)
-		if err != nil {
-			a.logger.Errorf("failed to persist balance: %w", err)
-		}
-
-		err = a.store.Put(peerSurplusBalanceKey(peer), zero)
-		if err != nil {
-			a.logger.Errorf("failed to persist surplus balance: %w", err)
-		}
-
-		accountingPeer.lock.Unlock()
-
-	}()
+	defer accountingPeer.lock.Unlock()
 
 	timeNow := a.timeNow().Unix()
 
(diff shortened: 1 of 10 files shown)


pkg/accounting/accounting.go
```diff
@@ -555,6 +562,36 @@ func (a *Accounting) PeerDebt(peer swarm.Address) (*big.Int, error) {
 	return peerDebt, nil
 }
 
+// peerLatentDebt returns the sum of the positive part of the outstanding balance, shadow reserve and the ghost balance
+func (a *Accounting) peerLatentDebt(peer swarm.Address) (*big.Int, error) {
+
+	accountingPeer := a.getAccountingPeer(peer)
+
+	balance := new(big.Int)
+	zero := big.NewInt(0)
```
anatollupacescu, line 571: are these two equivalent?

metacertain: no

pkg/accounting/accounting.go
```diff
@@ -862,10 +905,95 @@ func (d *debitAction) Cleanup() {
...
+
+	disconnectFor, err := a.blocklistUntil(peer, multiplier)
+	if err != nil {
+		return a.p2p.Blocklist(peer, 1*time.Hour)
+	}
+
+	return a.p2p.Blocklist(peer, time.Duration(disconnectFor)*time.Second)
+}
+
+func (a *Accounting) Connect(peer swarm.Address) {
```
esadakar, line 949: Maybe we can unit test these functions?

pkg/accounting/accounting.go
```diff
@@ -862,10 +905,95 @@ func (d *debitAction) Cleanup() {
...
+	if accountingPeer.reconnectAllowTimestamp != 0 {
+		timeNow := a.timeNow().Unix()
+		if timeNow < accountingPeer.reconnectAllowTimestamp {
+			disconnectFor := accountingPeer.reconnectAllowTimestamp - timeNow
+			_ = a.p2p.Blocklist(peer, time.Duration(disconnectFor)*time.Second)
+		}
+	}
+}
+
+func (a *Accounting) Disconnect(peer swarm.Address) {
```
esadakar, line 979: Maybe we can unit test these functions?

pkg/accounting/accounting.go
esadakar: Otherwise, LGTM. Nice work!

metacertain: yes, planning to add a unit test still, thanks!

janos (approved):

pkg/settlement/pseudosettle/pseudosettle.go
```diff
@@ -102,6 +102,7 @@ func (s *Service) init(ctx context.Context, p p2p.Peer) error {
 		s.peers[p.Address.String()] = peerData
 	}
 
+	s.accounting.Connect(p.Address)
```
metacertain, line 105: Question here is whether the connect reseting balances can race with first push / retrieve requests being reserved / credited / debited, so if we have traffic immediately from the peer, can it change any balances before they are reset creating discrepancies

zelig (approved):
> Reviewed 2 of 8 files at r1, 1 of 3 files at r2, 1 of 5 files at r3, 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/1983)* status: 9 of 10 files reviewed, 6 unresolved discussions (waiting on @anatollupacescu, @esadakar, and @metacertain)

___
*[pkg/accounting/accounting.go, line 94 at r4](https://reviewable.io/reviews/ethersphere/bee/1983#-McIKDv61c9rpcyrteFG:-McIKDv61c9rpcyrteFH:b-m7whyi) ([raw file](https://github.com/ethersphere/bee/blob/a4e56b29530d486799ca021cf0274be98474e955/pkg/accounting/accounting.go#L94)):*
> ```Go
> 	refreshTimestamp               int64      // last time we attempted time-based settlement
> 	paymentOngoing                 bool       // indicate if we are currently settling with the peer
> 	reconnectAllowTimestamp        int64
> ```

comments explaining the fields
___
*[pkg/settlement/swap/swap_test.go, line 86 at r4](https://reviewable.io/reviews/ethersphere/bee/1983#-McIPSYV9jCce2aYuTo4:-McIPSYV9jCce2aYuTo5:b7qt0zl) ([raw file](https://github.com/ethersphere/bee/blob/a4e56b29530d486799ca021cf0274be98474e955/pkg/settlement/swap/swap_test.go#L86)):*
> ```Go
> }
> 
> func (t *testObserver) Connect(peer swarm.Address) {
> ```

can we just introduce Connect and Disconnect in a Connecter interface only required where it is used

PR #2016, opened 2021-06-07
fix: delay concluding attempt in pushsync

metacertain:
This PR intends to make forwarder behavior ensure that chunk is successfully written to at least one further peer before concluding attempt has been made

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2016)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -454,6 +454,15 @@ func (ps *PushSync) pushPeer(ctx context.Context, peer swarm.Address, ch swarm.C
 
 	ps.metrics.TotalSent.Inc()
 
+	// if you manage to get a tag, just increment the respective counter
+	t, err := ps.tagger.Get(ch.TagID())
+	if err == nil && t != nil {
+		err = t.Inc(tags.StateSent)
+		if err != nil {
+			return nil, true, fmt.Errorf("tag %d increment: %v", ch.TagID(), err)
+		}
+	}
+
 	var receipt pb.Receipt
 	if err := r.ReadMsgWithContext(ctx, &receipt); err != nil {
 		_ = streamer.Reset()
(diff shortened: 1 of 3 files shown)


pkg/pushsync/pushsync_test.go
```diff
@@ -954,15 +952,25 @@ func TestPushChunkToClosestSkipFailed(t *testing.T) {
...
+			func(h p2p.HandlerFunc) p2p.HandlerFunc {
+				return func(ctx context.Context, peer p2p.Peer, stream p2p.Stream) error {
+					lock.Lock()
+					defer lock.Unlock()
 
-				return nil
+					if triggerCount < 9 {
+						triggerCount++
+						stream.Close()
+						return errors.New("fmt")
```
aloknerurkar, line 964: Use a better error here.

aloknerurkar (approved):

pkg/pushsync/pushsync.go
```diff
@@ -470,6 +461,15 @@ func (ps *PushSync) pushPeer(ctx context.Context, peer swarm.Address, ch swarm.C
 		return nil, true, err
 	}
 
+	// if you manage to get a tag, just increment the respective counter
+	t, err := ps.tagger.Get(ch.TagID())
```
acud, line 465: this is not correct... you should increment the sent count once the chunk is sent.

pkg/pushsync/pushsync_test.go
```diff
@@ -488,7 +488,7 @@ func TestPushChunkToNextClosest(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if ta2.Get(tags.StateSent) != 2 {
```
acud, line 491: this seems a bit wrong no? why is this change needed?

acud (changes requested):
tag increment needs to be reverted to previous state

pkg/pushsync/pushsync.go
metacertain: ok, it was a misunderstanding

pkg/pushsync/pushsync_test.go
metacertain: restored as well

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2016)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @acud and @metacertain)

___
*[pkg/pushsync/pushsync_test.go, line 944 at r3](https://reviewable.io/reviews/ethersphere/bee/2016#-MbkH6mRD0zCYWmXfmCO:-MbkH6mSE2NpaEjsWjCw:bg1io4a) ([raw file](https://github.com/ethersphere/bee/blob/9c237ed82b893bce65948a82f256bb357e1d2389/pkg/pushsync/pushsync_test.go#L944)):*
> ```Go
> 
> 	triggerCount := 0
> 	var lock sync.Mutex
> ```

Does this lock suppose to guard just the `triggerCount` variable? If so, I'd recommend to rename it to signal that those two are related.

esadakar (approved):

acud (approved):

PR #2018, opened 2021-06-07
feat: reduce required eth

ralph-pichler:
reduce required eth for initial deployment. value was still based on the pre proxy factory.

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2018)

### DIFF
--- a/pkg/settlement/swap/chequebook/init.go
+++ b/pkg/settlement/swap/chequebook/init.go
@@ -53,7 +53,7 @@ func checkBalance(
 			return err
 		}
 
-		minimumEth := gasPrice.Mul(gasPrice, big.NewInt(2000000))
+		minimumEth := gasPrice.Mul(gasPrice, big.NewInt(250000))
 
 		insufficientERC20 := erc20Balance.Cmp(swapInitialDeposit) < 0
 		insufficientETH := ethBalance.Cmp(minimumEth) < 0


metacertain (approved):

acud (approved):

Eknir:
Can't we do a minimum of 0? Before, this was required because otherwise the node wouldn't function. Now, it is completely up to the user.

ldeffenb:
> Can't we do a minimum of 0? Before, this was required because otherwise the node wouldn't function. Now, it is completely up to the user.

But wouldn't a minimum of zero cause bee to queue the checkbook deployment transaction even if there was zero gETH available in the account?  And that would just cause the transaction to fail over and over again.

Checkbook contract deployment does have a known gas cost, and I would say that it is reasonable for the bee node to ensure that there is at least that much gETH available to pay for it without sending a transaction that is guaranteed to fail.

But that's only my $0.02 (2 cents) and you can take it for whatever it's worth.

ldeffenb:
And isn't "up to the user" actually running a non-full-node?  Which shouldn't try to deploy the checkbook anyway.  But in a full-node: true, the checkbook is still required, isn't it?

ralph-pichler:
@Eknir if they want to run with 0, they can start with `swap-enable: false` which skips this check entirely. (downside is that if they are a full node and somebody wants to send them a cheque they will get blocklisted)

cppfuns:
Wouldn't it be better to manage this value through configuration?

PR #2019, opened 2021-06-07
feat: block on swap payment

ralph-pichler:
adds blocking (but non-locking) semantics to `Reserve` if the reserve is full but swap payment is already ongoing.

the logic is:
* if we are near enough the threshold we trigger time settlement and wait for it
* if that was not enough or we could not time settle at all we trigger swap settlement
* in the meantime we can continue until we burn through the rest of the reserve
* if swap settlement is not complete by the time the reserve is full, we wait for swap to complete (or the source request to time out)
* if afterwards we still cannot afford it (e.g. too little amount or payment failed) overdraft error is returned

this should also reduce the flakes seen in push-sync and pss test (however there is at least one other tag related issue in the push-sync test). 

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2019)

### DIFF
--- a/.github/workflows/beekeeper.yml
+++ b/.github/workflows/beekeeper.yml
@@ -14,6 +14,7 @@ jobs:
       REPLICA: 3
       RUN_TYPE: "PR RUN"
       SETUP_CONTRACT_IMAGE_TAG: "0.2.0"
+      BEEKEEPER_BRANCH: "normal_thresholds"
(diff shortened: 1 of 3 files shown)


Eknir:
Is this expected to speed up insert/retrieve?
How can we validate if this is indeed fixing something, or is working?

ralph-pichler:
closing this for now as it might not be needed

PR #2036, opened 2021-06-09
feat: verify chunk availability before concluding synced status

metacertain:
This is an experimental change to increase push reliability
The availability check is done by attempting to retrieve through a peer farther away from the chunk as the node, to minimize the probability of merging/rejoining routes used for the push and verification

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2036)

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -642,7 +642,8 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
 		b.recoveryHandleCleanup = pssService.Register(recovery.Topic, chunkRepairHandler)
 	}
 
-	pusherService := pusher.New(networkID, storer, kad, pushSyncProtocol, tagService, logger, tracer, warmupTime)
+	pusherService := pusher.New(networkID, storer, kad, pushSyncProtocol, retrieve, tagService, logger, tracer, warmupTime)
+
 	b.pusherCloser = pusherService
 
 	pullStorage := pullstorage.New(storer)
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -20,6 +20,7 @@ import (
 	"github.com/ethersphere/bee/pkg/crypto"
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/pushsync"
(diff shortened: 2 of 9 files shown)


zelig (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2036)* status: all files reviewed, 3 unresolved discussions (waiting on @metacertain)

*[a discussion](https://reviewable.io/reviews/ethersphere/bee/2036#-MboWGgEA5RFCLVJMsHS:-MboWGgEA5RFCLVJMsHT:b-ot43pm) (no related file):*
not sure why this would be needed

---
___
*[pkg/pusher/pusher.go, line 199 at r1](https://reviewable.io/reviews/ethersphere/bee/2036#-MboVM2b5I2n-h_8utlK:-MboVM2b5I2n-h_8utlL:b2pc0dh) ([raw file](https://github.com/ethersphere/bee/blob/428a5499be371a6632e864ca6e9ff4f89dbd4c44/pkg/pusher/pusher.go#L199)):*
> ```Go
> 
> 				if err = s.retrieval.CheckAvailableChunk(ctx, ch.Address()); err != nil {
> 					err = fmt.Errorf("independent verification of availability failed, %w", err)
> ```

use : instead of ,
___
*[pkg/retrieval/retrieval.go, line 462 at r1](https://reviewable.io/reviews/ethersphere/bee/2036#-MboVhSMCWjfxmoMPGuU:-MboVhSMCWjfxmoMPGuV:b8l6u3z) ([raw file](https://github.com/ethersphere/bee/blob/428a5499be371a6632e864ca6e9ff4f89dbd4c44/pkg/retrieval/retrieval.go#L462)):*
> ```Go
> }
> 
> func (s *Service) farthestPeer(addr swarm.Address) (swarm.Address, error) {
> ```

would be better to have this function in peerSuggester?

metacertain:
> not sure why this would be needed

because, out of neighborhood peers creating storage receipts could be filtered out by checking the availability of the chunk, 
so large files can be uploaded more reliably

> use : instead of ,

?
 
> would be better to have this function in peerSuggester?
no opinion, it was introduced here because closestPeer was used from local code as well, to maintain consistency

Eknir:
Thanks for this PR @metacertain . Let's first roll out what we have to the testnet and then validate the effect of this change.

PR #2049, opened 2021-06-10
pusher: retry shallow receipts

acud:
This change makes pusher push chunks for which we've received shallow receipts multiple times to try and get them stored deeper. This is a naive implementation that can be elaborate later on to attempt different traces somehow.

unit tests still TODO

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2049)


mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2049)* status: all files reviewed, 1 unresolved discussion (waiting on @acud)

___
*[pkg/pusher/pusher.go, line 161 at r1](https://reviewable.io/reviews/ethersphere/bee/2049#-Mbu3li4DHd27-VdDQzt:-Mbu3li4DHd27-VdDQzu:b-6y3wsc) ([raw file](https://github.com/ethersphere/bee/blob/24121ff5fddf91da6065a218de238ac019b55fcb/pkg/pusher/pusher.go#L161)):*
> ```Go
> 						po := swarm.Proximity(ch.Address().Bytes(), storerPeer.Bytes())
> 						s.metrics.ReceiptDepth.WithLabelValues(strconv.Itoa(int(po))).Inc()
> 						delete(retryCounter, ch.Address().ByteString())
> ```

Shouldn't be the delete operation also guarded with the `mtx` mutex?

mrekucci (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2049#-:-Mburjek44xXtEqAewew:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud)

pkg/pusher/pusher.go
```diff
@@ -144,19 +151,20 @@ LOOP:
 					storerPeer swarm.Address
 				)
 				defer func() {
+					mtx.Lock()
 					if err == nil {
 						s.metrics.TotalSynced.Inc()
 						s.metrics.SyncTime.Observe(time.Since(startTime).Seconds())
 						// only print this if there was no error while sending the chunk
 						logger.Tracef("pusher: pushed chunk %s to node %s", ch.Address().String(), storerPeer.String())
```
Eknir, line 159: "pusher: pushed chunk %s to node %s" with depth (...)"

pkg/pusher/pusher.go
```diff
@@ -190,6 +198,21 @@ LOOP:
 						err = fmt.Errorf("pusher: receipt storer address: %w", err)
 						return
 					}
+
+					po := swarm.Proximity(ch.Address().Bytes(), storerPeer.Bytes())
+					d := s.depther.NeighborhoodDepth()
+					if po < d {
+						mtx.Lock()
+						retryCounter[ch.Address().ByteString()]++
```
Eknir, line 206: increment after the check with the counter. Otherwise you will effectively only retry `retryCount-1`

pkg/pusher/pusher.go
```diff
@@ -190,6 +198,21 @@ LOOP:
...
+						if retryCounter[ch.Address().ByteString()] < retryCount {
+							mtx.Unlock()
+							err = fmt.Errorf("pusher: shallow receipt depth %d, want at least %d", po, d)
+							po := swarm.Proximity(ch.Address().Bytes(), storerPeer.Bytes())
+							s.metrics.ReceiptDepth.WithLabelValues(strconv.Itoa(int(po))).Inc()
```
Eknir, line 211: I suggest to keep the metrics of retried chunks because of shallow depths separate from chunks that were final. Perhaps just a counter that keeps track of shallow retries and then another one of shallow accepted (because we exhaused our retry counts)

pkg/pusher/pusher.go
```diff
@@ -190,6 +198,21 @@ LOOP:
 						err = fmt.Errorf("pusher: receipt storer address: %w", err)
 						return
 					}
+
+					po := swarm.Proximity(ch.Address().Bytes(), storerPeer.Bytes())
+					d := s.depther.NeighborhoodDepth()
+					if po < d {
+						mtx.Lock()
```
Eknir, line 205: under this branch, you should also add the peer that returned the shallow-depth receipt to the skiplist of this particular chunk (not the protocol wide skiplist). This way, you will reduce the change of hitting exactly the same route again.

Eknir: Alternatively, the function that choses which peer to forward to, might be good to have a bit of randomness there and not deterministically choose always the same for each request.

acud: there are no skip lists nor even notions of peers in this component, so this is a bit out of scope. trying to get a different trace should be done tweaking the retry count and the pushsync protocol behavior by @esadakar

pkg/pusher/pusher.go
acud: since we already retried, this is the correct thing to do. you can double check

esadakar (approved):

janos:
This PR somehow revealed a data race on the logger that is shared between goroutines https://github.com/ethersphere/bee/pull/2049/checks?check_run_id=2820412618#step:10:5875.

PR #2066, opened 2021-06-11
feat: pushsync sanctions closest node on push error

esadakar:


zelig (approved):

Eknir:
Hi Esad, thank you for the PR. Could you please update the PR description?

pkg/pushsync/pushsync.go
```diff
@@ -482,54 +495,62 @@ func (ps *PushSync) pushPeer(ctx context.Context, peer swarm.Address, ch swarm.C
...
+
+func (l *peerSkipList) PruneExpired() {
+	l.Lock()
+	defer l.Unlock()
+
+	now := time.Now()
+
+	for k, v := range l.skipExpiration {
+		if v.Before(now) {
+			fmt.Println("expire", k)
```
esadakar, line 552: will remove when ready to merge

pkg/pushsync/pushsync.go
```diff
@@ -491,54 +504,64 @@ func (ps *PushSync) pushPeer(ctx context.Context, peer swarm.Address, ch swarm.C
...
+	defer l.Unlock()
+
+	peerStr := peer.ByteString()
 
-	val, found := f.cache.Get(keyForReq(peer, chunk))
-	if !found {
-		f.cache.Add(keyForReq(peer, chunk), 1)
-		return
+	if exp, has := l.skipExpiration[peerStr]; has {
+		delete(l.skipExpiration, peerStr)
```
acud, line 535: you need to delete only in the case that `exp.After(time.Now()) == true`

pkg/pushsync/pushsync.go
```diff
@@ -380,12 +387,14 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, retryAllo
 			// if ErrWantSelf is returned, it means we are the closest peer.
 			return nil, fmt.Errorf("closest peer: %w", err)
 		}
-		if !ps.failedRequests.Useful(peer, ch.Address()) {
-			skipPeers = append(skipPeers, peer)
-			ps.metrics.TotalFailedCacheHits.Inc()
+
+		skipPeers = append(skipPeers, peer)
+
+		if ps.skipList.ShouldSkip(peer) {
```
aloknerurkar, line 393: Why do we need the chunk map? I dont see it being used to make any decisions. We are just storing and getting stuff from it?

acud (approved):

pkg/pushsync/pushsync.go
esadakar: We use the chunk map to check if a peer has been already skipped for that chunk. It's one way to make sure only one peer is skipped.

PR #2082, opened 2021-06-14
refactor(kademlia): add peers at start asynchronously

mrekucci:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2082)

### DIFF
--- a/pkg/topology/kademlia/kademlia.go
+++ b/pkg/topology/kademlia/kademlia.go
@@ -9,7 +9,6 @@ import (
 	random "crypto/rand"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"math"
 	"math/big"
 	"math/bits"
@@ -570,12 +569,25 @@ func (k *Kad) Start(_ context.Context) error {
 	k.wg.Add(1)
 	go k.manage()
 
-	addresses, err := k.addressBook.Overlays()
-	if err != nil {
-		return fmt.Errorf("addressbook overlays: %w", err)
-	}
+	go func() {
+		select {
+		case <-k.halt:
+			return
+		case <-k.quit:
+			return
+		default:
+		}
 
-	k.AddPeers(addresses...)
+		start := time.Now()
+		addresses, err := k.addressBook.Overlays()
+		if err != nil {
+			k.logger.Errorf("addressbook overlays: %w", err)
+			return
+		}
+		k.AddPeers(addresses...)
+		k.metrics.StartAddAddressBookOverlaysTime.Observe(float64(time.Since(start).Nanoseconds()))
+	}()
+
 	return nil
 }
 
--- a/pkg/topology/kademlia/kademlia_test.go
+++ b/pkg/topology/kademlia/kademlia_test.go
@@ -807,6 +807,8 @@ func TestAddressBookQuickPrune(t *testing.T) {
 	}
 	defer kad.Close()
 
+	time.Sleep(100 * time.Millisecond)
+
 	nonConnPeer, err := bzz.NewAddress(signer, nonConnectableAddress, test.RandomAddressAt(base, 1), 0)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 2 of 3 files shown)


ldeffenb:
This may or may not work.  In my experience, the /topology Debug API also doesn't get an answer until the AddPeers call completes.  So there seems to be some locking going on inside there that may still prevent the Kademlia from making peer connection progress.

Give it a try sometime.  Start up a peer with a large (300,000+ population) address book and try to get :1635/topology.  It will sit there for a long time (15-30 minutes for my nodes) until it responds.  And as soon as it responds, the peer connections start happening.

acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2082)* status: all files reviewed, 2 unresolved discussions (waiting on @anatollupacescu, @esadakar, and @mrekucci)

___
*[pkg/p2p/libp2p/libp2p.go, line 275 at r1](https://reviewable.io/reviews/ethersphere/bee/2082#-McDd8E26IhabjlXrHAE:-McDd8E38Oiok4KDLjJ_:b-cua512) ([raw file](https://github.com/ethersphere/bee/blob/3b68e76cfd67738c5f1ac312faea0c13d97e335a/pkg/p2p/libp2p/libp2p.go#L275)):*
> ```Go
> 	select {
> 	case <-s.ready:
> 		go func() { _ = stream.Reset() }()
> ```

you would need to `return` after this, since this would just continue execution here
___
*[pkg/topology/kademlia/kademlia.go, line 585 at r1](https://reviewable.io/reviews/ethersphere/bee/2082#-McDdJx60ZOWvj6zTD_F:-McDdJx60ZOWvj6zTD_G:bqy4ur) ([raw file](https://github.com/ethersphere/bee/blob/3b68e76cfd67738c5f1ac312faea0c13d97e335a/pkg/topology/kademlia/kademlia.go#L585)):*
> ```Go
> 		addresses, err := k.addressBook.Overlays()
> 		if err != nil {
> 			panic(fmt.Errorf("addressbook overlays: %w", err))
> ```

I'm not sure that panicking is the right thing to do here, as it would cause a dirty shutdown. Logging the error and returning should be fine

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2082)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @acud, @anatollupacescu, and @esadakar)

___
*[pkg/topology/kademlia/kademlia.go, line 585 at r1](https://reviewable.io/reviews/ethersphere/bee/2082#-McDdJx60ZOWvj6zTD_F:-McF_EZ97pxYIBg33As1:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/3b68e76cfd67738c5f1ac312faea0c13d97e335a/pkg/topology/kademlia/kademlia.go#L585)):*

I'm not sure that panicking is the right thing to do here, as it would cause a dirty shutdown. Logging the error and returning should be fine
</blockquote></details>

Done.
___
*[pkg/p2p/libp2p/libp2p.go, line 275 at r1](https://reviewable.io/reviews/ethersphere/bee/2082#-McDd8E26IhabjlXrHAE:-McFZmrG2aQvvgqKTy8K:b8suiow) ([raw file](https://github.com/ethersphere/bee/blob/3b68e76cfd67738c5f1ac312faea0c13d97e335a/pkg/p2p/libp2p/libp2p.go#L275)):*

you would need to `return` after this, since this would just continue execution here
</blockquote></details>

Getting rid of this since the nodes cannot connect to each other even with the `return` statement.

acud (approved):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2082)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anatollupacescu and @esadakar)

PR #2093, opened 2021-06-15
fix: add peer in topology through the debug api

r0qs:
This PR fixes: https://github.com/ethersphere/bee/issues/1414.

Peers added by using the debug API were not being connected in the kademlia, and thus were not shown in the topology. This PR adds the `p2p.Notifier` to the topology driver to allow peers to be connected in kademlia. It also adds a flag that enables connection to over-saturated peers through the debug API.

The flag is only true if a connection request is made using the `/connect` endpoint in the debug API.
The flag does not affect bootnodes. Bootnodes always follow the saturation function.

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2093)

### DIFF
--- a/pkg/debugapi/peer.go
+++ b/pkg/debugapi/peer.go
@@ -35,6 +35,14 @@ func (s *Service) peerConnectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if err := s.topologyDriver.Connected(r.Context(), p2p.Peer{Address: bzzAddr.Overlay}, true); err != nil {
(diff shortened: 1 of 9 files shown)


acud (approved):

mrekucci:
Thanks for the PR! Could you please fix the commit message according to [this guideline](https://github.com/ethersphere/bee/blob/master/CODING.md#commit-messages). Also, a description of what it does and why it's needed would be great to have in the body of this PR.

aloknerurkar (approved):
Nice work! LGTM

PR #2105, opened 2021-06-16
fix: the default listening address of api and debug-api is changed to…

15ho:
… localhost

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2105)


acud:
This is intentional so that we bind on all interfaces

PR #2108, opened 2021-06-16
feat: non-mineable overlays

esadakar:
We introduce non-mineable overlays which are overlays whose sha256 sum now also contains the block hash of the block that's one after the block that contains the node's chequebook transaction. 

Verification of an overlay is the process of confirming that the block hash that was used to produce the overlay addresses belongs to the block whose parent is the block that contains the transaction. 

Standalone nodes must also pass in transaction and block hash options.

After a discussion with @acud @Eknir, we have concluded that addressbook entries will contain the transaction hash, hive protocol will skip the overlay validation, pushsync receipts will contain the blockhash,  and the p2p will expose the peer public key so that swap handshake can be removed.

TODO:
- ~Debug api must start before chequebook deployment, use configure to inject the overlay address.~
- ~remove adding block hash to statestore~
- ~change handshake err not found~
- ~confirm crypto.NewOverlayAddress with nil blockhash does not panic~
- ~move WaitBlock to backend.go~
- ~create new issue for overlay migration~ https://github.com/ethersphere/bee/issues/2125

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2108)


cmd/bee/cmd/cmd.go
```diff
@@ -238,6 +239,7 @@ func (c *command) setAllFlags(cmd *cobra.Command) {
 	cmd.Flags().String(optionNamePostageContractAddress, "", "postage stamp contract address")
 	cmd.Flags().String(optionNamePriceOracleAddress, "", "price oracle contract address")
 	cmd.Flags().String(optionNameTransactionHash, "", "proof-of-identity transaction hash")
+	cmd.Flags().String(optionNameBlockHash, "", "block hash of the block whose parent is the block that contains the transaction hash")
```
Eknir, line 242: Why did you chose to include this as a separate parameter? The blockhash should be the blockhash from the block in which the `optionNameTransactionHash` was included in the chain (or the swap deployment transaction if the tx-hash is not passed in)

cmd/bee/cmd/deploy.go
```diff
@@ -94,6 +91,31 @@ func (c *command) initDeployCmd() error {
...
+
+			optionTrxHash := c.config.GetString(optionNameTransactionHash)
+			optionBlockHash := c.config.GetString(optionNameBlockHash)
+
+			txHash, err := node.GetTxHash(stateStore, logger, optionTrxHash)
+			if err != nil {
+				return fmt.Errorf("invalid transaction hash: %w", err)
+			}
+
+			blockHash, err := node.GetTrxNextBlock(ctx, logger, stateStore, swapBackend, transactionMonitor, txHash, optionBlockHash)
```
Eknir, line 106: Does it make sense to refactor, to do save in another place, so that it would be more visible what is happening?

cmd/bee/cmd/cmd.go
esadakar: Standalone nodes have no way to fetch the hash since they do not connect to the blockchain backend.

cmd/bee/cmd/cmd.go
```diff
@@ -238,6 +239,7 @@ func (c *command) setAllFlags(cmd *cobra.Command) {
 	cmd.Flags().String(optionNamePostageContractAddress, "", "postage stamp contract address")
 	cmd.Flags().String(optionNamePriceOracleAddress, "", "price oracle contract address")
 	cmd.Flags().String(optionNameTransactionHash, "", "proof-of-identity transaction hash")
```
Eknir, line 241: Somehow mention here or in a comment elsewhere that the order of precedence is to use the tx-hash that is already stored, then the tx-hash that is passed in and then the tx-hash that is acquired through deploying a new chequebook.

pkg/node/chain.go
```diff
@@ -257,3 +260,77 @@ func InitSwap(
...
+	}
+
+	block, err := monitor.WaitBlock(ctx, big.NewInt(0).Add(tx.BlockNumber, big.NewInt(1)))
+	if err != nil {
+		return nil, err
+	}
+
+	hash = block.Hash()
+
+	err = stateStore.Put(nextBlockHashKey, hash)
```
Eknir, line 330: We are storing both the transaction and blockhash in the state. Is one of these perhaps redundant?

pkg/transaction/sender_matcher.go
```diff
@@ -29,24 +31,39 @@ func NewMatcher(backend Backend, signer types.Signer) *Matcher {
...
+	}
+
+	nextBlock, err := m.backend.HeaderByNumber(ctx, big.NewInt(0).Add(receipt.BlockNumber, big.NewInt(1)))
+	if err != nil {
+		return false, nil, err
+	}
+
+	if !bytes.Equal(receipt.BlockHash.Bytes(), nextBlock.ParentHash.Bytes()) {
+		return false, nil, nil
 	}
```
Eknir, line 64: What does this test for?

pkg/settlement/swap/swap_test.go
```diff
@@ -600,6 +603,9 @@ func TestHandshakeNewPeer(t *testing.T) {
 }
 
 func TestHandshakeWrongBeneficiary(t *testing.T) {
+
+	t.Skip()
```
Eknir, line 607: why do you skip this?

pkg/settlement/swap/swap.go
```diff
@@ -237,13 +236,9 @@ func (s *Service) SettlementsReceived() (map[string]*big.Int, error) {
...
-	// check that the overlay address was derived from the beneficiary (implying they have the same private key)
-	// while this is not strictly necessary for correct functionality we need to ensure no two peers use the same beneficiary
-	// as long as we enforce this we might not need the handshake message if the p2p layer exposed the overlay public key
-	expectedOverlay := crypto.NewOverlayFromEthereumAddress(beneficiary[:], s.networkID)
-	if !expectedOverlay.Equal(peer) {
-		return ErrWrongBeneficiary
-	}
+
+	// TODO: beneficiary check is removed here,
+	// and the beneficiary will be derived from the public eth key of the peer
```
Eknir, line 241: WIP?

Eknir (commented):
Left a few comments. But overall looks good

cmd/bee/cmd/cmd.go
Eknir: Wouldn't it be better for standalone nodes to just use a default value for txHash and blockhash, so they don't need to deal with this and can just ignore it?

pkg/settlement/swap/swap_test.go
esadakar: this test can actually be removed after @aloknerurkar 's PR

pkg/transaction/sender_matcher.go
esadakar: The block hash of the block that contains the chequebook must be the parent of the block whose hash was used for the overlay

pkg/node/chain.go
zelig: > We are storing both the transaction and blockhash in the state. Is one of these perhaps redundant?

there can be several txs by the same account in the parent block while there is only a single child block, so the blockhash is the redundant one

pkg/transaction/sender_matcher.go
Eknir: How I see it:
You have a receipt (ln 52), calculate the nextBlock from that (57), and then compare the nextBlock parent's hash against the hash of the block for the receipt. Unless there is an error in libraries used by the backend, this should always be the case, as the root object (txHash) is used all along the way.

pkg/node/chain.go
esadakar: It stores it so that the next boot up does not need to call the blockchain to fetch it.

pkg/settlement/swap/swap.go
esadakar: @anatollupacescu is working on this

pkg/node/chain.go
Eknir: I think it's good as it is.

pkg/crypto/crypto.go
```diff
@@ -20,24 +20,34 @@ import (
...
-func NewOverlayAddress(p ecdsa.PublicKey, networkID uint64) (swarm.Address, error) {
+func NewOverlayAddress(p ecdsa.PublicKey, networkID uint64, blockHash []byte) (swarm.Address, error) {
+
 	ethAddr, err := NewEthereumAddress(p)
 	if err != nil {
 		return swarm.ZeroAddress, err
 	}
-	return NewOverlayFromEthereumAddress(ethAddr, networkID), nil
+
+	if blockHash == nil {
```
acud, line 37: better check `len`

pkg/hive/hive.go
```diff
@@ -112,7 +113,8 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
 		if err != nil {
 			_ = stream.Reset()
 		} else {
-			_ = stream.FullClose()
+			time.Sleep(time.Millisecond * 100)
```
acud, line 116: would be good to have a comment why this is needed, also, it should be in a separate goroutine as discussed

pkg/settlement/swap/swap.go
```diff
@@ -237,13 +236,9 @@ func (s *Service) SettlementsReceived() (map[string]*big.Int, error) {
...
 func (s *Service) Handshake(peer swarm.Address, beneficiary common.Address) error {
-	// check that the overlay address was derived from the beneficiary (implying they have the same private key)
-	// while this is not strictly necessary for correct functionality we need to ensure no two peers use the same beneficiary
-	// as long as we enforce this we might not need the handshake message if the p2p layer exposed the overlay public key
-	expectedOverlay := crypto.NewOverlayFromEthereumAddress(beneficiary[:], s.networkID)
-	if !expectedOverlay.Equal(peer) {
-		return ErrWrongBeneficiary
-	}
+
+	// TODO: beneficiary check is removed here,
```
acud, line 240: remove TODOs?

acud (approved):
nice. minor comments

pkg/debugapi/debugapi_test.go
```diff
@@ -162,15 +162,15 @@ func TestServer_Configure(t *testing.T) {
 	)
 	jsonhttptest.Request(t, client, http.MethodGet, "/addresses", http.StatusOK,
 		jsonhttptest.WithExpectedJSONResponse(debugapi.AddressesResponse{
-			Overlay:      o.Overlay,
+			Overlay:      swarm.ZeroAddress,
```
acud, line 165: can we not have the field `nil` here?

pkg/node/node.go
```diff
@@ -158,7 +156,7 @@ const (
 	basePrice   = 10000
 )
 
-func NewBee(addr string, swarmAddress swarm.Address, publicKey ecdsa.PublicKey, signer crypto.Signer, networkID uint64, logger logging.Logger, libp2pPrivateKey, pssPrivateKey *ecdsa.PrivateKey, o Options) (b *Bee, err error) {
+func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, networkID uint64, logger logging.Logger, libp2pPrivateKey, pssPrivateKey *ecdsa.PrivateKey, o *Options) (b *Bee, err error) {
```
acud, line 159: why change to pointer semantics?

acud (approved):
LGTM. In addition, I think it would be more correct to remove the `Overlay` from the debugapi response if it has not been set yet. Having a nullable json field supported by a `*swarm.Address` should do the trick

pkg/node/node.go
esadakar: there was some DeepSource (ci checks) complaints about passing in large objects without a pointer.

pkg/hive/hive.go
esadakar: the goroutine, for reasons I can't explain, makes this extremely flaky

Eknir (approved):
LGTM, thanks!

PR #2114, opened 2021-06-16
feat: add public key to handshake

anatollupacescu:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2114)


zelig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2114)* status: all files reviewed, 1 unresolved discussion (waiting on @anatollupacescu)

___
*[pkg/bzz/address.go, line 83 at r1](https://reviewable.io/reviews/ethersphere/bee/2114#-McMyZBC6HkeztnZJG5l:-McMyZBC6HkeztnZJG5m:bfgmw6k) ([raw file](https://github.com/ethersphere/bee/blob/0479445646d7b5cf763fa3f465ff872e15701326/pkg/bzz/address.go#L83)):*
> ```Go
> 	}
> 
> 	ethAddress, err := crypto.NewEthereumAddress(*recoveredPK)
> ```

as you see this value can be calculated and thus unnecessary to be put in the address struct

ralph-pichler (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2114)* status: all files reviewed, 1 unresolved discussion (waiting on @zelig)

___
*[pkg/bzz/address.go, line 83 at r1](https://reviewable.io/reviews/ethersphere/bee/2114#-McMyZBC6HkeztnZJG5l:-McOTJWpDL3Pbmj2uUTF:b4r4cjd) ([raw file](https://github.com/ethersphere/bee/blob/0479445646d7b5cf763fa3f465ff872e15701326/pkg/bzz/address.go#L83)):*

as you see this value can be calculated and thus unnecessary to be put in the address struct 
</blockquote></details>

the goal is to have access to this during the p2p handlers. if it is not stored somewhere this will have to be recomputed quite often.

pkg/bzz/address.go
```diff
@@ -28,10 +28,11 @@ var ErrInvalidAddress = errors.New("invalid address")
...
 type Address struct {
-	Underlay    ma.Multiaddr
-	Overlay     swarm.Address
-	Signature   []byte
-	Transaction []byte
+	Underlay        ma.Multiaddr
+	Overlay         swarm.Address
+	Signature       []byte
+	Transaction     []byte
+	EthereumAddress []byte
```
esadakar, line 35: Why isn't it sufficient to put into p2p.Peer? Why add it to the addressbook?

ralph-pichler: how will you recover it from the outside? recovering requires the `generateSignData` logic which is only available here. It could be exported, then we could do the same computation another time somewhere else.

esadakar: Isn't this PR about preparing the removal of beneficiary check we do in the swap protocol because the handshake should now pass the public key?

PR #2131, opened 2021-06-18
push index iterator and pusher simplified for more reliable push sync

zelig:
This PR provides a minimalistic simple implementation of the pusher for push sync.
The motivation is to eliminate the excessive complexity currently in the code.


pkg/pusher/pusher.go
```diff
@@ -25,294 +24,202 @@ import (
...
-	metrics           metrics
-	quit              chan struct{}
-	chunksWorkerQuitC chan struct{}
-}
-
 var (
-	retryInterval  = 5 * time.Second // time interval between retries
-	concurrentJobs = 10              // how many chunks to push simultaneously
-	retryCount     = 6
+	concurrentPushes = 1000 // how many chunks to push simultaneously
```
acud, line 30: we've seen massive amounts of errors with this number set at _20_

PR #2132, opened 2021-06-18
fix: experimental locks related to connectivity 

metacertain:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2132)

### DIFF
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -73,8 +73,12 @@ type Service struct {
 	lightNodes        lightnodes
 	lightNodeLimit    int
 	protocolsmu       sync.RWMutex
-	connectMuMap      map[string]sync.Mutex
+	connectMuMap      map[string]*connectPeer
 	connectMapMu      sync.Mutex
+}
+
+type connectPeer struct {
+	lock sync.Mutex
 }
 
 type lightnodes interface {
@@ -96,13 +100,13 @@ type Options struct {
 	Transaction    []byte
 }
 
-func (s *Service) getPeerMutex(peer string) sync.Mutex {
+func (s *Service) getPeerMutex(peer string) *connectPeer {
 	s.connectMapMu.Lock()
 	defer s.connectMapMu.Unlock()
 
 	peerMu, ok := s.connectMuMap[peer]
 	if !ok {
-		var peerMu sync.Mutex
+		peerMu = &connectPeer{}
 		s.connectMuMap[peer] = peerMu
 	}
 
@@ -256,7 +260,7 @@ func New(ctx context.Context, signer beecrypto.Signer, networkID uint64, overlay
 		connectionBreaker: breaker.NewBreaker(breaker.Options{}), // use default options
 		ready:             make(chan struct{}),
 		halt:              make(chan struct{}),
-		connectMuMap:      make(map[string]sync.Mutex),
+		connectMuMap:      make(map[string]*connectPeer),
 		lightNodes:        lightNodes,
 	}
 
@@ -570,8 +574,8 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (address *bzz.
 	}
 
 	peerMu := s.getPeerMutex(info.ID.String())
-	peerMu.Lock()
-	defer peerMu.Unlock()
+	peerMu.lock.Lock()
+	defer peerMu.lock.Unlock()
 
 	hostAddr, err := buildHostAddress(info.ID)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


mrekucci (changes requested):
For the `kademlia.go` and the `libp2p.go` I'd suggest using `sync.Map`, it's more suitable for this case because:
> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

and it'll simplify the code, e.g.:
```golang
func (k *Kad) getPeerMutex(peer string) *sync.Mutex {
	mu, _ := k.connectMuMap.LoadOrStore(peer, new(sync.Mutex))
	return mu.(*sync.Mutex)
}
```

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2132#-:-McU02M49x1n95rwHCVb:b-byiqat)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @metacertain)

___
*[pkg/p2p/libp2p/peer.go, line 65 at r1](https://reviewable.io/reviews/ethersphere/bee/2132#-McTxGC30R6WhHtz_gZ5:-McTxGC30R6WhHtz_gZ6:b-7qnhrt) ([raw file](https://github.com/ethersphere/bee/blob/fa778c6ec8453ecefc4a9a1edef9ad3c195fc04c/pkg/p2p/libp2p/peer.go#L65)):*
> ```Go
> 	// not eventusally newly created one for the same peer
> 	if _, ok := r.connections[peerID][c]; !ok {
> 		fmt.Println("removing on one channel only:", peerID)
> ```

Is this just a debugging print or a forgotten one? If the former then I suggest to inject and use logger here.
___
*[pkg/p2p/libp2p/peer.go, line 73 at r1](https://reviewable.io/reviews/ethersphere/bee/2132#-McTxtU882VqcRn02q5f:-McTxtU90vmnPZTgS5g9:bf26u1) ([raw file](https://github.com/ethersphere/bee/blob/fa778c6ec8453ecefc4a9a1edef9ad3c195fc04c/pkg/p2p/libp2p/peer.go#L73)):*
> ```Go
> 	delete(r.connections[peerID], c)
> 	if len(r.connections[peerID]) > 0 {
> 		fmt.Println("removed peer not on every channel yet:", peerID)
> ```

The same as above.
___
*[pkg/pusher/pusher.go, line 139 at r1](https://reviewable.io/reviews/ethersphere/bee/2132#-McTyD_B7ZgIHToC9NAc:-McTyD_B7ZgIHToC9NAd:bmp9e3q) ([raw file](https://github.com/ethersphere/bee/blob/fa778c6ec8453ecefc4a9a1edef9ad3c195fc04c/pkg/pusher/pusher.go#L139)):*
> ```Go
> 			case <-s.quit:
> 				if unsubscribe != nil {
> 					fmt.Println("\\\\\\|||///\\\\\\|||/// Unsubscribe at", time.Now().Unix())
> ```

Is this a forgotten one? If not use logger here.
___
*[pkg/pusher/pusher.go, line 264 at r1](https://reviewable.io/reviews/ethersphere/bee/2132#-McTyJWY2J2YiBF5CmIX:-McTyJWZ5Ng5JNhQ7PLm:bf26u1) ([raw file](https://github.com/ethersphere/bee/blob/fa778c6ec8453ecefc4a9a1edef9ad3c195fc04c/pkg/pusher/pusher.go#L264)):*
> ```Go
> 			// if subscribe was running, stop it
> 			if unsubscribe != nil {
> 				fmt.Println("\\\\\\|||///\\\\\\|||///\\\\\\|||/// Unsubscribe at", time.Now().Unix())
> ```

The same as above.
___
*[pkg/pusher/pusher.go, line 272 at r1](https://reviewable.io/reviews/ethersphere/bee/2132#-McTyLcVCAjgETCmGrdU:-McTyLcVCAjgETCmGrdV:bf26u1) ([raw file](https://github.com/ethersphere/bee/blob/fa778c6ec8453ecefc4a9a1edef9ad3c195fc04c/pkg/pusher/pusher.go#L272)):*
> ```Go
> 			// and start iterating on Push index from the beginning
> 			chunks, unsubscribe = s.storer.SubscribePush(ctx)
> 			fmt.Println("\\\\\\|||///\\\\\\|||///\\\\\\|||/// Subscribe at", time.Now().Unix())
> ```

The same as above.
___
*[pkg/pusher/pusher.go, line 285 at r1](https://reviewable.io/reviews/ethersphere/bee/2132#-McTyMh4BjbnjXaR3SKu:-McTyMh4BjbnjXaR3SKv:bf26u1) ([raw file](https://github.com/ethersphere/bee/blob/fa778c6ec8453ecefc4a9a1edef9ad3c195fc04c/pkg/pusher/pusher.go#L285)):*
> ```Go
> 		case <-s.quit:
> 			if unsubscribe != nil {
> 				fmt.Println("\\\\\\|||/// Unsubscribe at", time.Now().Unix())
> ```

The same as above.

PR #2147, opened 2021-06-19
add disclaimer

Eknir:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2147)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,13 @@ Welcome to the Swarm.... Bzzz Bzzzz Bzzzz
    |                '  ' '
 ```
 
+## DISCLAIMER
+This software is provided to you "as is", use at your own risk and without warranties of any kind.
+It is your responsibility to read and understand how Swarm works and the implications of running this software.
+The usage of Bee involves various risks, including, but not limited to:
+damage to hardware or loss of funds associated with the Ethereum account connected to your node.
+No developers or entity involved will be liable for any claims and damages whatsoever associated with your use,
+inability to use, or your interaction with other nodes or the software.
 
 Our documentation is hosted at https://docs.ethswarm.org.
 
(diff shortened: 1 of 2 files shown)


Eknir:
While the user is responsible, we are not careless or uninterested. bzzaar disclaimer proposes to say:

`
The code has been carefully tested, peer-reviewed and fully audited. Although this can be seen as a step forward in terms of security, it's recommended to use the protocol at your own risk.
`

Take inspiration from that.

cmd/bee/cmd/start.go
```diff
@@ -96,6 +96,18 @@ Welcome to the Swarm.... Bzzz Bzzzz Bzzzz
...
 			fmt.Println(beeASCII)
+			fmt.Print(`
+
+DISCLAIMER:
+This software is provided to you "as is", use at your own risk and without warranties of any kind.
+It is your responsibility to read and understand how Swarm works and the implications of running this software.
+The usage of Bee involves various risks, including, but not limited to:
+damage to hardware or loss of funds associated with the Ethereum account connected to your node.
+No developers or entity involved will be liable for any claims and damages whatsoever associated with your use,
+inabibility to use, or your interaction with other nodes or the software.
```
acud, line 107: inabibility

README.md
```diff
@@ -20,6 +20,13 @@ Welcome to the Swarm.... Bzzz Bzzzz Bzzzz
    |                '  ' '
 ```
 
+## DISCLAIMER
```
acud, line 23: there is already a disclaimer in the `LICENSE` file. I don't think this is necessary as the license file should already cover this legally. I would also advise using the disclaimer in the license file if you really want to print it in the terminal. It is not needed to be explicit on what can happen as a result of using `bee`.

cmd/bee/cmd/start.go
Eknir: fixed, thanks

PR #2162, opened 2021-06-21
feat(kademlia): AddPeers batch on Start

acud:
Since the process of adding peers can take a long time, I am adding an iterator that processes them in batches. It is just unclear whether this would result in increased CPU usage while this iterator is open in leveldb (@janos WDYT?).

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2162)

### DIFF
--- a/pkg/addressbook/addressbook.go
+++ b/pkg/addressbook/addressbook.go
@@ -26,6 +26,8 @@ type Interface interface {
 	Remover
 	// Overlays returns a list of all overlay addresses saved in addressbook.
 	Overlays() ([]swarm.Address, error)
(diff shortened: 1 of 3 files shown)


esadakar (approved):

aloknerurkar (approved):

janos (approved):
👍

janos:
@acud this is totally fine. There should not be larger cpu utilization, mostly because of the usage of the store where this iterator is mostly used without any other operations at the start time. This approach is better for memory consumption for sure.

anatollupacescu (approved):

PR #2168, opened 2021-06-21
fix: only greylist after error

ralph-pichler:
Only add to the greylist after actual errors. This is necessary because the handshake can and does frequently run mutiple times concurrently for the same peer.

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2168)

### DIFF
--- a/pkg/transaction/sender_matcher.go
+++ b/pkg/transaction/sender_matcher.go
@@ -71,35 +71,62 @@ func (m *Matcher) Matches(ctx context.Context, tx []byte, networkID uint64, send
 		return nil, ErrGreylisted
 	}
 
-	err = m.storage.Put(peerOverlayKey(senderOverlay, incomingTx), &overlayVerification{
-		TimeStamp: m.timeNow(),
-		Verified:  false,
-	})
-	if err != nil {
-		return nil, err
-	}
-
 	nTx, isPending, err := m.backend.TransactionByHash(ctx, incomingTx)
 	if err != nil {
+		err2 := m.storage.Put(peerOverlayKey(senderOverlay, incomingTx), &overlayVerification{
+			TimeStamp: m.timeNow(),
+			Verified:  false,
+		})
+		if err2 != nil {
+			return nil, err2
+		}
 		return nil, fmt.Errorf("%v: %w", err, ErrTransactionNotFound)
 	}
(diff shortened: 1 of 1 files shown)


acud (approved):

anatollupacescu (approved):

pkg/transaction/sender_matcher.go
```diff
@@ -71,35 +71,62 @@ func (m *Matcher) Matches(ctx context.Context, tx []byte, networkID uint64, send
...
-		TimeStamp: m.timeNow(),
-		Verified:  false,
-	})
-	if err != nil {
-		return nil, err
-	}
-
 	nTx, isPending, err := m.backend.TransactionByHash(ctx, incomingTx)
 	if err != nil {
+		err2 := m.storage.Put(peerOverlayKey(senderOverlay, incomingTx), &overlayVerification{
```
esadakar, line 76: maybe we can move this to a new function and call it once in a `defer` function in the case of an error?

ralph-pichler: then we cannot return the `err2` to anything and we don't have logger available here without changing more stuff, tests, etc.

pkg/transaction/sender_matcher.go
```diff
@@ -71,35 +71,62 @@ func (m *Matcher) Matches(ctx context.Context, tx []byte, networkID uint64, send
 		return nil, ErrGreylisted
```
esadakar, line 71: Do we use this `Timestamp` value anywhere? I don't see a greylisting mechanism anywhere?

pkg/transaction/sender_matcher.go
esadakar: we can still move this out
`err2 := m.storage.Put(peerOverlayKey(senderOverlay, incomingTx), &overlayVerification{
			TimeStamp: m.timeNow(),
			Verified:  false,
		})
		if err2 != nil {
			return nil, err2
		}`

pkg/transaction/sender_matcher.go
ralph-pichler: Timestamp is set to `time.Now` on error. Then this if will cause `ErrGreylisted` for 5 minutes.

PR #2183, opened 2021-06-22
fix: default networkd-id should be 1, not 10

moremorefun:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2183)


moremorefun:
Here is some confusion, the default configuration seems to want to match the test network?
If so, we need to change the comment in bee.yaml.

PR #2191, opened 2021-06-22
fix: set commit time to the current time when missing

mrekucci:
Sets the commit time to the current time when the
node is running directly from the source code
without compilation.


mrekucci:
> does this also help with the wrong time in release builds?

Yes, it should, even though the wrong time in release builds should be addressed separately since the time will be updated every time the node is re-started in this PR.

acud (approved):

anatollupacescu (approved):

PR #2227, opened 2021-06-28
feat: batchstore checksums

acud:
This PR adds a simple, flattened merkelized hash chain checksum implementation to the batch service which processes events from the postage event listener. It is meant to provide an extra layer of debugging information to compare reserve state across nodes.

It uses a simple sha3-256 hasher and works in the following manner:
0. current checksum of state is written into the hasher (if exists on component bootstrap)
1. transaction hash bytes of event which has been read from the chain is written into the hasher after being processed
2. the sum of the hasher is produced and:
2.1. written to disk
2.2. the hasher is reset to zero, and the sum written into it again
3. repeat steps 1-2 indefinitely

The checksums will be printed alongside each event log line in the terminal.

TODO:
- [x] Unit tests
- ~~metric that communicates the checksum periodically to a field that can be communicated through grafana~~

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2227)

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -75,6 +75,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	ma "github.com/multiformats/go-multiaddr"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/crypto/sha3"
 	"golang.org/x/sync/errgroup"
 )
 
@@ -416,7 +417,10 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
 		eventListener = listener.New(logger, swapBackend, postageContractAddress, o.BlockTime, &pidKiller{node: b})
 		b.listenerCloser = eventListener
 
-		batchSvc = batchservice.New(stateStore, batchStore, logger, eventListener, overlayEthAddress.Bytes(), post)
+		batchSvc, err = batchservice.New(stateStore, batchStore, logger, eventListener, overlayEthAddress.Bytes(), post, sha3.New256)
+		if err != nil {
+			return nil, err
+		}
 
 		erc20Address, err := postagecontract.LookupERC20Address(p2pCtx, transactionService, postageContractAddress)
 		if err != nil {
--- a/pkg/postage/batchservice/batchservice.go
+++ b/pkg/postage/batchservice/batchservice.go
@@ -46,13 +46,17 @@ func New(
 	listener postage.Listener,
 	owner []byte,
 	batchListener postage.BatchCreationListener,
+	checksumFunc func() hash.Hash,
 ) (Interface, error) {
+	if checksumFunc == nil {
+		checksumFunc = sha3.New256
+	}
 	var (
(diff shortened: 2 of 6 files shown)


acud:
scratching the metrics part since it contaminates too many interfaces

mrekucci (approved):
Just one nitpick, I'd rename `txhash` -> `txHash`. Otherwise :lgtm: 

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2227#-:-MdWGV3zACei8KCBv2L3:blpga14)* status: all files reviewed, 1 unresolved discussion (waiting on @acud)

___
*[pkg/postage/batchservice/batchservice.go, line 49 at r1](https://reviewable.io/reviews/ethersphere/bee/2227#-MdWDalE5eurcHADRJY8:-MdWDalFCaZ_HDGmpevo:b-9n7z35) ([raw file](https://github.com/ethersphere/bee/blob/5f7456097206460024cf4cd5252cdbcf36a5662d/pkg/postage/batchservice/batchservice.go#L49)):*
> ```Go
> 	owner []byte,
> 	batchListener postage.BatchCreationListener,
> 	checksumFunc func() hash.Hash,
> ```

Is this `checksumFunc` needed here as a param? It is only used in the `node.go` and it's set up to the same value as when it is `nil` (in tests), so wouldn't be better to just use the `sh3.New256` directly without passing it as a param?
___
*[pkg/postage/batchservice/batchservice.go, line 59 at r1](https://reviewable.io/reviews/ethersphere/bee/2227#-MdWEK5H9D_EsM_voRp8:-MdWEK5H9D_EsM_voRp9:b-upqfqq) ([raw file](https://github.com/ethersphere/bee/blob/5f7456097206460024cf4cd5252cdbcf36a5662d/pkg/postage/batchservice/batchservice.go#L59)):*
> ```Go
> 	)
> 
> 	if err := stateStore.Get(checksumDBKey, &b); err == nil {
> ```

Would be nice to log the `err` when it's not `nil`.

istae (approved):

acud (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2227)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @mrekucci)

___
*[pkg/postage/batchservice/batchservice.go, line 49 at r1](https://reviewable.io/reviews/ethersphere/bee/2227#-MdWDalE5eurcHADRJY8:-MdWeS5K1dXeAskM9Rwh:bn2v68u) ([raw file](https://github.com/ethersphere/bee/blob/5f7456097206460024cf4cd5252cdbcf36a5662d/pkg/postage/batchservice/batchservice.go#L49)):*

Is this `checksumFunc` needed here as a param? It is only used in the `node.go` and it's set up to the same value as when it is `nil` (in tests), so wouldn't be better to just use the `sh3.New256` directly without passing it as a param?
</blockquote></details>

it is used in the checksum unit test

PR #2229, opened 2021-06-28
fix: reduce retrieval result chan buffer

metacertain:
This PR intends to fix a memory overuse issue on oversized result channel buffers in retrieval 

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2229)

### DIFF
--- a/pkg/retrieval/retrieval.go
+++ b/pkg/retrieval/retrieval.go
@@ -130,7 +130,7 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin 
 		var (
 			peerAttempt  int
 			peersResults int
-			resultC      = make(chan retrievalResult, 1)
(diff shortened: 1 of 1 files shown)


pkg/retrieval/retrieval.go
```diff
@@ -126,7 +126,7 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin
 		var (
 			peerAttempt  int
 			peersResults int
-			resultC      = make(chan retrievalResult, maxSelects)
```
acud, line 129: why is a buffered channel larger than 1 needed here? the separate goroutines can just try to write to the channel with a select-default block, and then it is not needed. i find the current implementation a bit convoluted, maybe we could simplify it a bit? not sure why it is needed for example to communicate an empty result. these patterns can be simplified using waitgroups

PR #2230, opened 2021-06-28
feat: chunk stream upload endpoint

aloknerurkar:
Currently in the `/chunks` API we only support single chunk upload

Applications using the `/chunks` endpoint need to loop while sending multiple chunks. This can cause server to hit issues due to too many connections (or on client-side while re-using connections we might see an EOF).

As we already use the `gorilla/websocket` library, I think it would be better to have a client-side streaming endpoint to upload multiple chunks. This will be easier on the server side as well as would re-use all the resources initialised for the first chunk upload.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2230)

### DIFF
--- a/pkg/api/router.go
+++ b/pkg/api/router.go
@@ -63,6 +63,11 @@ func (s *server) setupRouting() {
 		),
 	})
 
+	handle("/stream/chunks", web.ChainHandlers(
+		s.gatewayModeForbidEndpointHandler,
+		web.FinalHandlerFunc(s.chunkUploadStreamHandler),
+	))
+
 	handle("/chunks/{addr}", jsonhttp.MethodHandler{
 		"GET": http.HandlerFunc(s.chunkGetHandler),
 	})
--- a/pkg/api/router.go
+++ b/pkg/api/router.go
@@ -63,8 +63,8 @@ func (s *server) setupRouting() {
 		),
 	})
 
-	handle("/stream/chunks", web.ChainHandlers(
-		s.gatewayModeForbidEndpointHandler,
+	handle("/chunks/stream", web.ChainHandlers(
+		s.newTracingHandler("chunks-stream-upload"),
 		web.FinalHandlerFunc(s.chunkUploadStreamHandler),
 	))
 
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -175,7 +175,7 @@ paths:
         - $ref: "SwarmCommon.yaml#/components/parameters/SwarmPostageBatchId"
       responses:
         "200":
-          description: "Returns a Websocket connection on which stream of chunks can be uploaded. Each chunk sent is acknowledged using a text response `success`. Chunks should be packaged as binary messages for uploading."
+          description: "Returns a Websocket connection on which stream of chunks can be uploaded. Each chunk sent is acknowledged using a binary response `0` which serves as confirmation of upload of single chunk. Chunks should be packaged as binary messages for uploading."
         "400":
           $ref: "SwarmCommon.yaml#/components/responses/400"
         default:
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -164,6 +164,22 @@ paths:
         default:
           description: Default response
 
+  "/chunks/stream":
+    get:
+      summary: "Upload stream of chunks"
+      tags:
+        - Chunk
+      parameters:
+        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmTagParameter"
+        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmPinParameter"
+        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmPostageBatchId"
+      responses:
+        "200":
+          description: "Returns a Websocket connection on which stream of chunks can be uploaded. Each chunk sent is acknowledged using a text response `success`. Chunks should be packaged as binary messages for uploading."
+        "400":
+          $ref: "SwarmCommon.yaml#/components/responses/400"
+        default:
+          description: Default response
   "/bzz":
     post:
       summary: "Upload file or a collection of files"
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -71,6 +71,7 @@ type testServerOptions struct {
 	PostageContract    postagecontract.Interface
 	Post               postage.Service
 	Steward            steward.Reuploader
+	WsHeaders          http.Header
 }
 
 func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.Conn, string) {
@@ -115,7 +116,7 @@ func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.
 
 	if o.WsPath != "" {
 		u := url.URL{Scheme: "ws", Host: ts.Listener.Addr().String(), Path: o.WsPath}
-		conn, _, err = websocket.DefaultDialer.Dial(u.String(), nil)
+		conn, _, err = websocket.DefaultDialer.Dial(u.String(), o.WsHeaders)
 		if err != nil {
 			t.Fatalf("dial: %v. url %v", err, u.String())
 		}
--- a/pkg/api/chunk.go
+++ b/pkg/api/chunk.go
@@ -57,7 +57,7 @@ func (s *server) processUploadRequest(
 
 	putter, err = newStamperPutter(s.storer, s.post, s.signer, batch)
 	if err != nil {
-		s.logger.Debugf("chunk upload: putter:%v", err)
+		s.logger.Debugf("chunk upload: putter: %v", err)
 		s.logger.Error("chunk upload: putter")
 		switch {
 		case errors.Is(err, postage.ErrNotFound):
@@ -65,7 +65,7 @@ func (s *server) processUploadRequest(
 		case errors.Is(err, postage.ErrNotUsable):
 			return nil, nil, nil, errors.New("batch not usable")
 		}
-		return nil, nil, nil, errors.New("")
+		return nil, nil, nil, err
 	}
 
 	return ctx, tag, putter, nil
--- a/pkg/api/chunk.go
+++ b/pkg/api/chunk.go
@@ -6,6 +6,7 @@ package api
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -29,27 +30,67 @@ type chunkAddressResponse struct {
 	Reference swarm.Address `json:"reference"`
 }
 
-func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
-	var (
-		tag *tags.Tag
-		ctx = r.Context()
-		err error
-	)
+func (s *server) processUploadRequest(
+	r *http.Request,
+) (ctx context.Context, tag *tags.Tag, putter storage.Putter, err error) {
 
 	if h := r.Header.Get(SwarmTagHeader); h != "" {
 		tag, err = s.getTag(h)
 		if err != nil {
 			s.logger.Debugf("chunk upload: get tag: %v", err)
 			s.logger.Error("chunk upload: get tag")
-			jsonhttp.BadRequest(w, "cannot get tag")
-			return
-
+			return nil, nil, nil, errors.New("cannot get tag")
 		}
 
 		// add the tag to the context if it exists
 		ctx = sctx.SetTag(r.Context(), tag)
 
 		// increment the StateSplit here since we dont have a splitter for the file upload
+		err = tag.Inc(tags.StateSplit)
+		if err != nil {
+			s.logger.Debugf("chunk upload: increment tag: %v", err)
+			s.logger.Error("chunk upload: increment tag")
+			return nil, nil, nil, errors.New("cannot increment tag")
+		}
+	} else {
+		ctx = r.Context()
+	}
+
+	batch, err := requestPostageBatchId(r)
+	switch {
+	case errors.Is(err, errSwarmPostageBatchIDHeaderNotFound) && s.post.DefaultIssuer() != nil:
+		batch = s.post.DefaultIssuer().ID()
+	case err != nil:
+		s.logger.Debugf("chunk upload: postage batch id: %v", err)
+		s.logger.Error("chunk upload: postage batch id")
+		return nil, nil, nil, errors.New("invalid postage batch id")
+	}
+
+	putter, err = newStamperPutter(s.storer, s.post, s.signer, batch)
+	if err != nil {
+		s.logger.Debugf("chunk upload: putter:%v", err)
+		s.logger.Error("chunk upload: putter")
+		switch {
+		case errors.Is(err, postage.ErrNotFound):
+			return nil, nil, nil, errors.New("batch not found")
+		case errors.Is(err, postage.ErrNotUsable):
+			return nil, nil, nil, errors.New("batch not usable")
+		}
+		return nil, nil, nil, errors.New("")
+	}
+
+	return ctx, tag, putter, nil
+}
+
+func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
+	fmt.Println("handling")
+	ctx, tag, putter, err := s.processUploadRequest(r)
+	if err != nil {
+		jsonhttp.BadRequest(w, err.Error())
+		return
+	}
+
+	if tag != nil {
 		err = tag.Inc(tags.StateSplit)
 		if err != nil {
 			s.logger.Debugf("chunk upload: increment tag: %v", err)
--- a/pkg/api/chunk.go
+++ b/pkg/api/chunk.go
@@ -133,7 +133,9 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 			jsonhttp.BadRequest(w, "increment tag")
 			return
 		}
-	} else if tag != nil {
+	}
+
+	if tag != nil {
 		// indicate that the chunk is stored
 		err = tag.Inc(tags.StateStored)
 		if err != nil {
@@ -149,9 +151,7 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 		if err := s.pinning.CreatePin(ctx, chunk.Address(), false); err != nil {
 			s.logger.Debugf("chunk upload: creation of pin for %q failed: %v", chunk.Address(), err)
 			s.logger.Error("chunk upload: creation of pin failed")
-			// since we already increment the pin counter because of the ModePut, we need
-			// to delete the pin here to prevent the pin counter from never going to 0
-			err = s.pinning.DeletePin(ctx, chunk.Address())
+			err = s.storer.Set(ctx, storage.ModeSetUnpin, chunk.Address())
 			if err != nil {
 				s.logger.Debugf("chunk upload: deletion of pin for %q failed: %v", chunk.Address(), err)
 				s.logger.Error("chunk upload: deletion of pin failed")
--- a/pkg/api/chunk.go
+++ b/pkg/api/chunk.go
@@ -57,10 +57,7 @@ func (s *server) processUploadRequest(
 	}
 
 	batch, err := requestPostageBatchId(r)
-	switch {
-	case errors.Is(err, errSwarmPostageBatchIDHeaderNotFound) && s.post.DefaultIssuer() != nil:
-		batch = s.post.DefaultIssuer().ID()
-	case err != nil:
+	if err != nil {
 		s.logger.Debugf("chunk upload: postage batch id: %v", err)
 		s.logger.Error("chunk upload: postage batch id")
 		return nil, nil, nil, errors.New("invalid postage batch id")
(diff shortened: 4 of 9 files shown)


aloknerurkar:
> why not support stream for files first?
> in other words why not use similar for the file endpoint

I actually implemented this as part of something else I was working on. I was using the `/chunks` endpoint as I am using all our other packages in the code. (https://github.com/aloknerurkar/bee-fs)

If we are okay with the approach I can implement this for files as well.

acud:
> why not support stream for files first?
> in other words why not use similar for the file endpoint

@zelig streaming of files can be done today on the `/bzz` endpoint using multipart AFAIK. This is indeed not the most up-to-date method of streaming. That would be using the [streams](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) API in conjunction with the Fetch API.
@alok I would consider to alter the implementation so that it is compatible with the streams and fetch API, since this is implemented in a non-standardized manner.
It would be also good to have the people from the JS team to comment which way to go, since it would be mainly leveraged and implemented by them as a primary stakeholder, at least at the beginning. cc @agazso @AuHau @nugaon

AuHau:
Hey,

yeah, to have a possibility to stream chunks upload seems like a crucial thing to have the more we will go into the direction of "lean Bee", so I definitely welcome this idea!

> That would be using the streams API in conjunction with the Fetch API.

@acud I believe what are you mentioning are JS native APIs that are used for streaming in JS, but do not directly affect the transport protocol. `Stream` provides an abstraction to handle streaming data and `fetch` is a tool for making HTTP requests. So here we still have to choose the transport protocol and its usage.

I see two ways here.

As @acud mentioned there is already support for `/bzz` using multipart. We could utilize the same principle here. But this does have disadvantages for us in browsers because browsers do not support streaming requests. 

From this point of view actually, @aloknerurkar's approach is kind of innovative and even though non-standard it would bypass this browser's limitation, which would allow us in the future to do for example uploading big files from the browser using only chunks (eq. chunking, BMT construction and/or manifests all on client side). So I am actually in favor of this approach!

nugaon:
> As @acud mentioned there is already support for /bzz using multipart. We could utilize the same principle here. But this does have disadvantages for us in browsers because browsers do not support streaming requests.

I don't completely understand this statement or it is just not true. `multipart/form-data` does have support for browsers, even for HTML forms.
Also, on browser-side with JS you can stream but the data that before uploading via any stream solution will be copied into the RAM AFAIK (see FileReader usages). [Though somehow the file stream reading should be possible as Youtube also solved somehow ;) if you know it how pls let me know].

Therefore

> From this point of view actually, @aloknerurkar's approach is kind of innovative and even though non-standard it would bypass this browser's limitation, which would allow us in the future to do for example uploading big files from the browser using only chunks (eq. chunking, BMT construction and/or manifests all on client side). So I am actually in favor of this approach!

is not connected to the transport protocol and its handsakes/headers question like in the case of the stream and fetch APIs that you pointed out correctly.
I mean by that with `multipart` you can also upload as many chunks as you want in one upload process by putting the chunks as files into the `FileList` and send it to the Bee client, but it still reading whole file(s) into the memory first, so I don't see it would solve the "browser issue". Can you make some clarification about it @AuHau?  

Nevertheless, the websocket implementation seems a good approach too, because it may save some unnecessary header data between messages, but the interpretation of the response messages about the successful chunk uploads may be cumbersome on client side. Because of the latter I don't see why would it be better than a traditional `multipart` uploading with the extension that I just described.

nugaon:
> [Though somehow the file stream reading should be possible as Youtube also solved somehow ;) if you know it how pls let me know].

Maybe the `File`'s `stream` or `slice` method? https://developer.mozilla.org/en-US/docs/Web/API/File#instance_methods

aloknerurkar:
> From this point of view actually, @aloknerurkar's approach is kind of innovative and even though non-standard it would bypass this browser's limitation, which would allow us in the future to do for example uploading big files from the browser using only chunks (eq. chunking, BMT construction and/or manifests all on client side). So I am actually in favor of this approach!

This resonates with why I implemented this in the first place. I am using bee manifests (in golang) outside of bee and hence when I interact with bee, I can already format the data I want to upload into a series of chunks. 

> Because of the latter I don't see why would it be better than a traditional multipart uploading with the extension that I just described.

Putting chunks into FileList seems hacky and deviating from the standard.

> but the interpretation of the response messages about the successful chunk uploads may be cumbersome on client side

The example code in JS that I have seen has callbacks for messages coming from upstream. So I don't understand why to would be complicated. The approach I have takes is a Request-Response scheme. So similar to how we do in libp2p protocols. Each response is read before we send the next request.

AuHau:
> Also, on browser-side with JS you can stream but the data that before uploading via any stream solution will be copied into the RAM AFAIK (see FileReader usages). [Though somehow the file stream reading should be possible as Youtube also solved somehow ;) if you know it how pls let me know].

@nugaon yes you are right browser can handle Streams but not completely. As you have guessed the problem is that it will load all the data to memory, which is the main problem here.

>  [Though somehow the file stream reading should be possible as Youtube also solved somehow ;) if you know it how pls let me know].

Downloading stream is not the problem here. That is already supported, only the upload is problematic. Or do you mean that they can upload big files? My understanding is that this limitation that I am talking about here is only about XHR/AJAX/fetch requests. I think that when submitting a normal form with large files, the browser handles it internally differently and actually stream the file.


To my knowledge, only Chrome recently shipped this feature but only as a closed experiment (you have to register somewhere for a given URL and only then Chrome actually enables it for your site). See for example [here](https://www.chromestatus.com/feature/5274139738767360), [here](https://web.dev/fetch-upload-streaming/) and open issue [here](https://github.com/whatwg/fetch/issues/966#issuecomment-608128154).

> Therefore ... is not connected to the transport protocol and its handsakes/headers question like in the case of the stream and fetch APIs that you pointed out correctly.
I mean by that with multipart you can also upload as many chunks as you want in one upload process by putting the chunks as files into the FileList and send it to the Bee client, but it still reading whole file(s) into the memory first, so I don't see it would solve the "browser issue". Can you make some clarification about it @AuHau?

Yes, with multipart you could upload multiple chunks, but you would have to load them into memory, which is the problem I am trying to raise here :-)

With the `websockets` you can assemble chunks on the fly and send them (=stream them) as you have a persistent connection with the server. You can do similar thing now, only making a new request for each chunk, which is something that @aloknerurkar is trying to fix. With `multipart` you would have to load all the data into memory before sending them.

---

I guess the question is what is the end-game here. As I see it, from the direction of "lean Bee", a lot of operations with data will in the future happen on the client-side and it is then crucial for the client-side to be able to work with data efficiently which requires for example not loading all the data into the memory. To my knowledge from JS API point of view there is already support for everything we need (like having streams, having a way to stream a file from FS) **except** stream the data to Bee node. 

I know that the focus of Swarm/Bee now is not long-term file storage for big data and that we are not promoting uploading big files, but I would argue that if/once we make this shift the tools should be ready for it. Or in another word, the tools/libraries should not be the problem for a task that is already possible but discouraged.

pkg/api/router.go
```diff
@@ -63,6 +63,11 @@ func (s *server) setupRouting() {
 		),
 	})
 
+	handle("/stream/chunks", web.ChainHandlers(
+		s.gatewayModeForbidEndpointHandler,
```
AuHau, line 67: Why is it supposed to be disabled on the gateway?

pkg/api/router.go
```diff
@@ -63,6 +63,11 @@ func (s *server) setupRouting() {
 		),
 	})
 
+	handle("/stream/chunks", web.ChainHandlers(
```
AuHau, line 66: I would maybe suggest `/chunks/stream` URL?

acud (changes requested):
Looks good, I have a few minor comments

> Reviewed 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2230#-:-MfcAJWZB7KIRfr-klgI:b-vur6uz)* status: all files reviewed, 6 unresolved discussions (waiting on @aloknerurkar)

___
*[pkg/api/chunk_stream.go, line 1 at r2](https://reviewable.io/reviews/ethersphere/bee/2230#-MfcAG7C-FXNH8o6JyEf:-MfcAG7C-FXNH8o6JyEg:b-dk8tp4) ([raw file](https://github.com/ethersphere/bee/blob/c3faa21e730ac4508bf228f2f2183d8e30dbcd28/pkg/api/chunk_stream.go#L1)):*
> ```Go
> package api
> ```

copyright missing
___
*[pkg/api/chunk_stream.go, line 164 at r2](https://reviewable.io/reviews/ethersphere/bee/2230#-Mfc9Zn49K46wlhgP5Pb:-Mfc9Zn49K46wlhgP5Pc:b-ix8jr5) ([raw file](https://github.com/ethersphere/bee/blob/c3faa21e730ac4508bf228f2f2183d8e30dbcd28/pkg/api/chunk_stream.go#L164)):*
> ```Go
> 		if tag != nil {
> 			// indicate that the chunk is stored
> 			err = tag.Inc(tags.StateStored)
> ```

if the chunk has been seen before, no need to increment this again
___
*[pkg/api/chunk_stream.go, line 166 at r2](https://reviewable.io/reviews/ethersphere/bee/2230#-Mfc9iTEExbCZFOtJG2I:-Mfc9iTEExbCZFOtJG2J:b-krb7zs) ([raw file](https://github.com/ethersphere/bee/blob/c3faa21e730ac4508bf228f2f2183d8e30dbcd28/pkg/api/chunk_stream.go#L166)):*
> ```Go
> 			err = tag.Inc(tags.StateStored)
> 			if err != nil {
> 				s.logger.Debugf("chunk upload: increment tag", err)
> ```

i would also add an Error here, as we do in the other endpoints. This applies also to the other error checks in this file
___
*[pkg/api/chunk_stream_test.go, line 1 at r2](https://reviewable.io/reviews/ethersphere/bee/2230#-Mfc9q1Q9ECTXsBt0WGl:-Mfc9q1Q9ECTXsBt0WGm:be55z33) ([raw file](https://github.com/ethersphere/bee/blob/c3faa21e730ac4508bf228f2f2183d8e30dbcd28/pkg/api/chunk_stream_test.go#L1)):*
> ```Go
> package api_test
> ```

code copyright missing

aloknerurkar (commented):
> Dismissed @AuHau and @acud from 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2230)* status: 5 of 8 files reviewed, all discussions resolved (waiting on @acud)

___
*[pkg/api/chunk_stream.go, line 1 at r2](https://reviewable.io/reviews/ethersphere/bee/2230#-MfcAG7C-FXNH8o6JyEf:-Mfh6CPmC4G9hf4eMR69:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c3faa21e730ac4508bf228f2f2183d8e30dbcd28/pkg/api/chunk_stream.go#L1)):*

copyright missing
</blockquote></details>

Done.
___
*[pkg/api/chunk_stream.go, line 164 at r2](https://reviewable.io/reviews/ethersphere/bee/2230#-Mfc9Zn49K46wlhgP5Pb:-Mfh4p-f7-fNwQ5IGqh9:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c3faa21e730ac4508bf228f2f2183d8e30dbcd28/pkg/api/chunk_stream.go#L164)):*

if the chunk has been seen before, no need to increment this again
</blockquote></details>

Done.
___
*[pkg/api/chunk_stream.go, line 166 at r2](https://reviewable.io/reviews/ethersphere/bee/2230#-Mfc9iTEExbCZFOtJG2I:-Mfh4oYw5AeLm9XYch-8:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c3faa21e730ac4508bf228f2f2183d8e30dbcd28/pkg/api/chunk_stream.go#L166)):*

i would also add an Error here, as we do in the other endpoints. This applies also to the other error checks in this file
</blockquote></details>

Done.
___
*[pkg/api/chunk_stream_test.go, line 1 at r2](https://reviewable.io/reviews/ethersphere/bee/2230#-Mfc9q1Q9ECTXsBt0WGl:-Mfh6Evi5KbK59IDQtZ1:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c3faa21e730ac4508bf228f2f2183d8e30dbcd28/pkg/api/chunk_stream_test.go#L1)):*

code copyright missing
</blockquote></details>

Done.
___
*[pkg/api/router.go, line 66 at r1](https://reviewable.io/reviews/ethersphere/bee/2230#-MdGxsSkGtcsdZBNbbMx-r1-66:-Mfh4tE82Th4pTwO9dSG:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/38cf3cc71344d1edda984ca8f86869ddec70df40/pkg/api/router.go#L66)):*
<details><summary><i>Previously, AuHau (Adam Uhlíř) wrote…</i></summary><blockquote>

I would maybe suggest `/chunks/stream` URL? 
</blockquote></details>

Done.
___
*[pkg/api/router.go, line 67 at r1](https://reviewable.io/reviews/ethersphere/bee/2230#-MdGxsSkGtcsdZBNbbMx-r1-67:-Mfh4scED3o8vX2buJuZ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/38cf3cc71344d1edda984ca8f86869ddec70df40/pkg/api/router.go#L67)):*
<details><summary><i>Previously, AuHau (Adam Uhlíř) wrote…</i></summary><blockquote>

Why is it supposed to be disabled on the gateway?
</blockquote></details>

Done.

acud (changes requested):
> Reviewed 2 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2230)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @acud and @aloknerurkar)

___
*[pkg/api/chunk_stream.go, line 185 at r3](https://reviewable.io/reviews/ethersphere/bee/2230#-Mg5AZTqB17olF2Jvdfv:-Mg5AZTqB17olF2Jvdfw:b-zhlxrg) ([raw file](https://github.com/ethersphere/bee/blob/b21ff6027fd795532edbee94420917b37fddfcc9/pkg/api/chunk_stream.go#L185)):*
> ```Go
> 			if err := s.pinning.CreatePin(ctx, chunk.Address(), false); err != nil {
> 				s.logger.Debugf("chunk stream handler: creation of pin for %q failed: %v", chunk.Address(), err)
> 				s.logger.Error("chunk stream handler: creation of pin failed")
> ```

for consistency consideration if this happens and if the pinning on localstore really did happen correctly, you would need to now unpin the chunk again from the localstore
___
*[pkg/api/chunk_stream.go, line 191 at r3](https://reviewable.io/reviews/ethersphere/bee/2230#-Mg5AjBYCscHB1Z9dfDc:-Mg5AjBYCscHB1Z9dfDd:b-5rwm57) ([raw file](https://github.com/ethersphere/bee/blob/b21ff6027fd795532edbee94420917b37fddfcc9/pkg/api/chunk_stream.go#L191)):*
> ```Go
> 		}
> 
> 		err = sendMsg(websocket.TextMessage, successWsMsg)
> ```

nit, but since the API is handling binary data, a string success message might not be needed. you could utilize some response code which is purely binary too, like `0x0` for clean exit (like exit codes), or `0x1` for error, etc

aloknerurkar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2230)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @acud)

___
*[pkg/api/chunk_stream.go, line 185 at r3](https://reviewable.io/reviews/ethersphere/bee/2230#-Mg5AZTqB17olF2Jvdfv:-Mg5VaZR-CmrFKMoyEjb:beg3y4m) ([raw file](https://github.com/ethersphere/bee/blob/b21ff6027fd795532edbee94420917b37fddfcc9/pkg/api/chunk_stream.go#L185)):*

for consistency consideration if this happens and if the pinning on localstore really did happen correctly, you would need to now unpin the chunk again from the localstore
</blockquote></details>

In this case, pinning has failed, right? Is there a way we could get an error from `CreatePin` even if pinning succeeded? Or you meant if we fail to send the response?
___
*[pkg/api/chunk_stream.go, line 191 at r3](https://reviewable.io/reviews/ethersphere/bee/2230#-Mg5AjBYCscHB1Z9dfDc:-Mg5Usk-2wjC02kErVSv:b5uzoku) ([raw file](https://github.com/ethersphere/bee/blob/b21ff6027fd795532edbee94420917b37fddfcc9/pkg/api/chunk_stream.go#L191)):*

nit, but since the API is handling binary data, a string success message might not be needed. you could utilize some response code which is purely binary too, like `0x0` for clean exit (like exit codes), or `0x1` for error, etc
</blockquote></details>

For the error, we use `sendErrorClose` which sends a code and a message. So we only need this message for confirmation of chunk upload. I have used a `0` byte now instead of the string message.

aloknerurkar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2230)* status: 4 of 8 files reviewed, all discussions resolved (waiting on @acud)

___
*[pkg/api/chunk_stream.go, line 185 at r3](https://reviewable.io/reviews/ethersphere/bee/2230#-Mg5AZTqB17olF2Jvdfv:-Mg5zavb7dJU9dOYQCUY:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/b21ff6027fd795532edbee94420917b37fddfcc9/pkg/api/chunk_stream.go#L185)):*

In this case, pinning has failed, right? Is there a way we could get an error from `CreatePin` even if pinning succeeded? Or you meant if we fail to send the response?
</blockquote></details>

Done.

nugaon:
> The approach I have taken is a Request-Response scheme. So similar to how we do in libp2p protocols. Each response is read before we send the next request.

I think a global WS endpoint with [RPC-like messages](https://www.jsonrpc.org/specification) would be more effective for all ws services as we need this streaming on the download part as well later.
By that, if we will have auth layer on HTTP requests (that we should have sonner than later) this authentication could happen once at the initialization of the websocket connection (like `new BeeWs('ws://localhost:1636/ws/api-key`). If every service gets its own ws endpoint later the things may go complex (understanding each EPs message syntax one-by-one, also handling the authorization in every handler on bee side) and not unified.

Also I don't see why requests-responses need to be blocked in a queue instead of simply just upload chunks independently from each other and handle the responses by their IDs

acud (approved):
> Reviewed 3 of 3 files at r4, 3 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2230)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar)

___
*[pkg/api/chunk.go, line 156 at r6](https://reviewable.io/reviews/ethersphere/bee/2230#-Mg7Hgg66gYT5ECI1iE-:-Mg7Hgg66gYT5ECI1iE0:b-zaa9hq) ([raw file](https://github.com/ethersphere/bee/blob/d8e97610df58f56da2ee078a1a74a6672e3a5388/pkg/api/chunk.go#L156)):*
> ```Go
> 			err = s.storer.Set(ctx, storage.ModeSetUnpin, chunk.Address())
> 			if err != nil {
> 				s.logger.Debugf("chunk upload: deletion of pin for %q failed: %v", chunk.Address(), err)
> ```

nit, not sure if %q will catch the Address to String method

mrekucci (changes requested):
> Reviewed 4 of 7 files at r2, 2 of 3 files at r4, 1 of 3 files at r6, 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2230)* status: all files reviewed, 2 unresolved discussions (waiting on @aloknerurkar)

___
*[pkg/api/chunk.go, line 60 at r7](https://reviewable.io/reviews/ethersphere/bee/2230#-Mg9rX7vE0oviAAXrezW:-Mg9rX7vE0oviAAXrezX:b-e7cu6x) ([raw file](https://github.com/ethersphere/bee/blob/2de3be2731a55052fea40b9917d4a2973736a370/pkg/api/chunk.go#L60)):*
> ```Go
> 	putter, err = newStamperPutter(s.storer, s.post, s.signer, batch)
> 	if err != nil {
> 		s.logger.Debugf("chunk upload: putter:%v", err)
> ```

Missing space after `:` in the debug message.
___
*[pkg/api/chunk.go, line 68 at r7](https://reviewable.io/reviews/ethersphere/bee/2230#-Mg9ruzBEGTXQ29PvKpv:-Mg9ruzBEGTXQ29PvKpw:b-aeffc7) ([raw file](https://github.com/ethersphere/bee/blob/2de3be2731a55052fea40b9917d4a2973736a370/pkg/api/chunk.go#L68)):*
> ```Go
> 			return nil, nil, nil, errors.New("batch not usable")
> 		}
> 		return nil, nil, nil, errors.New("")
> ```

The returned error `errors.New("")` doesn't tell much. Don't you wanna return `err` instead?
___
*[pkg/api/chunk_stream.go, line 106 at r7](https://reviewable.io/reviews/ethersphere/bee/2230#-Mg9oqH98JOL8_exAlEH:-Mg9oqH98JOL8_exAlEI:b-arnfux) ([raw file](https://github.com/ethersphere/bee/blob/2de3be2731a55052fea40b9917d4a2973736a370/pkg/api/chunk_stream.go#L106)):*
> ```Go
> 
> 	for {
> 		select {
> ```

You should also check for `ctx.Done()`.
___
*[pkg/api/pss.go, line 26 at r7](https://reviewable.io/reviews/ethersphere/bee/2230#-Mg9pzHhEdEbLQKt4bTh:-Mg9pzHhEdEbLQKt4bTi:b-nr4ng3) ([raw file](https://github.com/ethersphere/bee/blob/2de3be2731a55052fea40b9917d4a2973736a370/pkg/api/pss.go#L26)):*
> ```Go
> )
> 
> var (
> ```

You can change the variables into constants.

aloknerurkar:
___
*[pkg/api/chunk_stream.go, line 106 at r7](https://reviewable.io/reviews/ethersphere/bee/2230#-Mg9oqH98JOL8_exAlEH:-Mg9zb8y78F7Wwl01LbJ:byjo7o) ([raw file](https://github.com/ethersphere/bee/blob/2de3be2731a55052fea40b9917d4a2973736a370/pkg/api/chunk_stream.go#L106)):*

You should also check for `ctx.Done()`.
</blockquote></details>

This is not required as gorilla websocket library ignores the http request context. The `gone` channel will ensure we stop if client goes away and the `quit` channel will ensure we stop if the application is stopped. Apart from that we dont need any more synchronization.

The request context can be canceled prematurely. Ideally we should use a new context here, but none of the function calls which need the context actually check on `ctx.Done()` in this context.

aloknerurkar (commented):
> Dismissed @mrekucci from a discussion.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2230)* status: 7 of 9 files reviewed, all discussions resolved (waiting on @acud and @mrekucci)

___
*[pkg/api/chunk.go, line 68 at r7](https://reviewable.io/reviews/ethersphere/bee/2230#-Mg9ruzBEGTXQ29PvKpv:-Mg9z_-uE4BbrGgpDqdf:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/2de3be2731a55052fea40b9917d4a2973736a370/pkg/api/chunk.go#L68)):*

The returned error `errors.New("")` doesn't tell much. Don't you wanna return `err` instead?
</blockquote></details>

Done.

mrekucci (approved):
> Reviewed 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2230)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aloknerurkar)

PR #2236, opened 2021-06-29
feat: allow swap without chequebook

ralph-pichler:
* add a now flag `chequebook-enable` which is true by default.
* if `chequebook-enable` is false but `swap-enable` is true the node will still be able to receive and cash cheques but does not deploy a (or use a preexisting) chequebook and therefore does not send cheques. only time-based settlement will be used.
* transaction hash needs to be supplied manually if started with this and there has been no deployment in the past. as sending a tx manually from the overlay key can be tricky this is best done using the new alternative tx data type (#2409).
* several api endpoints are disabled in this mode.

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2236)

### DIFF
--- a/cmd/bee/cmd/cmd.go
+++ b/cmd/bee/cmd/cmd.go
@@ -60,6 +60,7 @@ const (
 	optionNameSwapLegacyFactoryAddresses = "swap-legacy-factory-addresses"
 	optionNameSwapInitialDeposit         = "swap-initial-deposit"
 	optionNameSwapEnable                 = "swap-enable"
+	optionNameChequebookEnable           = "chequebook-enable"
 	optionNameTransactionHash            = "transaction"
 	optionNameBlockHash                  = "block-hash"
 	optionNameSwapDeploymentGasPrice     = "swap-deployment-gas-price"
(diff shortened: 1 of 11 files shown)


zelig (commented):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2236)* status: all files reviewed, 1 unresolved discussion (waiting on @ralph-pichler)

___
*[pkg/debugapi/router.go, line 154 at r1](https://reviewable.io/reviews/ethersphere/bee/2236#-MdQI6bhEAkoPQA4eBdh:-MdQI6bhEAkoPQA4eBdi:bcj3ep7) ([raw file](https://github.com/ethersphere/bee/blob/e4af8b5747eac35de0912b53b276cf1629c26fda/pkg/debugapi/router.go#L154)):*
> ```Go
> 			"GET": http.HandlerFunc(s.peerSettlementsHandler),
> 		})
> 		router.Handle("/chequebook/cheque/{peer}", jsonhttp.MethodHandler{
> ```

so these are endpoints not requiring to have our own chequebook?

ldeffenb:
I thought received cheques that are subsequently cashed put the tokens in the local node's chequebook?  If the node doesn't have one, where do the cashed tokens end up?

PR #2237, opened 2021-06-29
fix: only in neighborhood replication

metacertain:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2237)

### DIFF
--- a/pkg/topology/mock/mock.go
+++ b/pkg/topology/mock/mock.go
@@ -164,7 +164,7 @@ func (m *mock) IsWithinDepth(addr swarm.Address) bool {
 	if m.isWithinFunc != nil {
 		return m.isWithinFunc(addr)
 	}
-	return true
+	return false
 }
 
 func (m *mock) EachNeighbor(f topology.EachPeerFunc) error {
--- a/pkg/topology/mock/mock.go
+++ b/pkg/topology/mock/mock.go
@@ -164,7 +164,7 @@ func (m *mock) IsWithinDepth(addr swarm.Address) bool {
 	if m.isWithinFunc != nil {
 		return m.isWithinFunc(addr)
 	}
-	return false
+	return true
 }
 
 func (m *mock) EachNeighbor(f topology.EachPeerFunc) error {
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -322,6 +322,10 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, retryAllo
 					return nil, ErrWarmup
 				}
 
+				if !ps.topologyDriver.IsWithinDepth(ch.Address()) {
+					return nil, ErrNoPush
+				}
+
 				count := 0
 				// Push the chunk to some peers in the neighborhood in parallel for replication.
 				// Any errors here should NOT impact the rest of the handler.
--- a/pkg/pushsync/pushsync_test.go
+++ b/pkg/pushsync/pushsync_test.go
@@ -129,13 +129,9 @@ func TestReplicateBeforeReceipt(t *testing.T) {
 	defer storerEmpty.Close()
 	emptyRecorder := streamtest.New(streamtest.WithProtocols(psEmpty.Protocol()), streamtest.WithBaseAddr(secondPeer))
 
-	wFunc := func(addr swarm.Address) bool {
-		return true
-	}
-
 	// node that is connected to closestPeer
 	// will receieve chunk from closestPeer
-	psSecond, storerSecond, _, secondAccounting := createPushSyncNode(t, secondPeer, defaultPrices, emptyRecorder, nil, defaultSigner, mock.WithPeers(emptyPeer), mock.WithIsWithinFunc(wFunc))
+	psSecond, storerSecond, _, secondAccounting := createPushSyncNode(t, secondPeer, defaultPrices, emptyRecorder, nil, defaultSigner, mock.WithPeers(emptyPeer))
 	defer storerSecond.Close()
 	secondRecorder := streamtest.New(streamtest.WithProtocols(psSecond.Protocol()), streamtest.WithBaseAddr(closestPeer))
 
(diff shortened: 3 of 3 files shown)


acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2237)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @metacertain)

___
*[pkg/pushsync/pushsync.go, line 302 at r1](https://reviewable.io/reviews/ethersphere/bee/2237#-MdMHTp1-C1Ftk334KtN:-MdMHTp29Y-IfyuRE3ek:b-6e9jdf) ([raw file](https://github.com/ethersphere/bee/blob/cad2c16b235860048eaa67c0f5c81aa0e9fcc505/pkg/pushsync/pushsync.go#L302)):*
> ```Go
> 				}
> 
> 				if ps.topologyDriver.IsWithinDepth(ch.Address()) {
> ```

it would be better to reduce nesting by checking `if !WithinDepth() { return ErrNoPush }`, also, the extra error is not needed, and existing package-level errors could be reused.

metacertain:
> > _[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2237)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @metacertain)
> 
> _[pkg/pushsync/pushsync.go, line 302 at r1](https://reviewable.io/reviews/ethersphere/bee/2237#-MdMHTp1-C1Ftk334KtN:-MdMHTp29Y-IfyuRE3ek:b-6e9jdf) ([raw file](https://github.com/ethersphere/bee/blob/cad2c16b235860048eaa67c0f5c81aa0e9fcc505/pkg/pushsync/pushsync.go#L302)):_
> 
> > ```go
> > 				}
> > 
> > 				if ps.topologyDriver.IsWithinDepth(ch.Address()) {
> > ```
> 
> it would be better to reduce nesting by checking `if !WithinDepth() { return ErrNoPush }`, also, the extra error is not needed, and existing package-level errors could be reused.

there would also be an extra else if condition to change err when not within depth but ErrWantSelf, so that pusher does not take it as self-stored / synced

zelig (approved):
> Reviewed 2 of 2 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2237)* status: all files reviewed, 1 unresolved discussion (waiting on @acud)

mrekucci (approved):
> Reviewed 2 of 2 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2237)* status: all files reviewed, 1 unresolved discussion (waiting on @acud)

acud (changes requested):
> Reviewed 1 of 1 files at r1, 2 of 2 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2237)* status: all files reviewed, 2 unresolved discussions (waiting on @metacertain)

___
*[pkg/pushsync/pushsync.go, line 369 at r4](https://reviewable.io/reviews/ethersphere/bee/2237#-MdSLmPq5T82LeCp2lRn:-MdSLmPq5T82LeCp2lRo:b-rws2e) ([raw file](https://github.com/ethersphere/bee/blob/096df1b3473d12423ceb6503889ba6e2b5eb3b40/pkg/pushsync/pushsync.go#L369)):*
> ```Go
> 
> 				// if the node has warmed up AND no other closer peer has been tried
> 				if time.Now().After(ps.warmupPeriod) && !ps.skipList.HasChunk(ch.Address()) && ps.topologyDriver.IsWithinDepth(ch.Address()) {
> ```

this change is outside of the scope of this PR, since it changes the skip list behavior. It would be at least nice to document this change in the PR description, about why it is needed and what it affects.
___
*[pkg/topology/mock/mock.go, line 167 at r4](https://reviewable.io/reviews/ethersphere/bee/2237#-MdSMJ7m6_FyVarcstUc:-MdSMJ7m6_FyVarcstUd:bptu4xi) ([raw file](https://github.com/ethersphere/bee/blob/096df1b3473d12423ceb6503889ba6e2b5eb3b40/pkg/topology/mock/mock.go#L167)):*
> ```Go
> 		return m.isWithinFunc(addr)
> 	}
> 	return true
> ```

this is changing the test setup for another test that uses this mock and used to rely on a `false` here. how is it that this needs to change?

pkg/topology/mock/mock.go
```diff
@@ -164,7 +164,7 @@ func (m *mock) IsWithinDepth(addr swarm.Address) bool {
 	if m.isWithinFunc != nil {
 		return m.isWithinFunc(addr)
 	}
-	return false
+	return true
```
istae, line 167: @acud The reason why the default is now `true` is because the chunk must be in the neighborhood of the peer with this change. The majority of the tests expect a valid receipt so the closest peers in the tests must be in the  neighborhood. As  of writing this, there is only one test that expects a peer to be out of depth.

istae (approved):

pkg/topology/mock/mock.go
istae: In short, the previous default `false` causes every test to fail.

acud: @istae then can we have an explicit change to the tests, in which a custom mocked `IsWithinFunc` is injected with a `true`? my concern is that with this change, also forwarder nodes (and there are forwarding tests) will see the chunk as within depth, leading to a change in the test cases' final state

acud: it is also counter-intuitive that we would now test everything with the following assumptions by default:
- originator node sees all chunks as within neighborhood
- fowarder node sees all chunks as within neighborhood
- storer node sees all chunks as within neighborhood
- replicating node sees all chunks as within neighborhood

a bit too permissive IMO. so I would vote for reverting this back to `false` and have an explicit `true` injected with a mock wherever needed

acud (changes requested):

acud (approved):
Great, thanks @istae :pray:

PR #2238, opened 2021-06-29
Fix version

ldeffenb:
Apparently when version.go was updated to include the commitTime, the actual commit hash was changed from "commit" to "commitHash".  But the Makefile was not updated to reflect this change resulting in the "bee version" command always showing "-dev" instead of the commit hash.

This one-word patch allows the Makefile to populate the new "commitHash" value which then corrects the "bee version" output to include the hash instead of "-dev".

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2238)


ldeffenb:
Still trying to get those blasted commit strings to be compliant!

PR #2240, opened 2021-06-29
fix: display version with commit hash

significance:
At @ldeffenb's request, [ref]( https://github.com/ethersphere/bee/pull/2239)

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2240)


mrekucci:
@significance @ldeffenb thanks for pointing this out, fixed in #2285

PR #2264, opened 2021-06-30
feat: retry push in neighborhood

metacertain:
This pr implements 
- Retrial of pushing / establishing storage for in-neighborhood content for the purposes of pushsync
- Preemptive receipt signing behaviour within neighborhood depth to chunk after 9 seconds of not receiving a storage receipt from neighborhood peers
- If the preemptive storage establishment is in progress, the pending pushed peer can still succeed in passing a receipt back first
- Increases the number of originator retrials for the same chunk to increase the possibility of finding separate routes to neighborhoods without relying on the expiring skiplist
- If an in-neighborhood peer times out while being pushed, it is added to the skiplist for 1 minute. This is not feasible for forwarders outside the neighborhood as those peers might be timed out on by their downstream without being able to preemptively sign a receipt for this case
- Only using peers further away from the chunk for in-neighborhood immediate replication, as other peers could take it as a regular push attempt towards them, and cascade the immediate replication

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2264)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -365,7 +365,7 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, retryAllo
 				ps.metrics.TotalFailedSendAttempts.Inc()
 
 				// if the node has warmed up AND no other closer peer has been tried
-				if time.Now().After(ps.warmupPeriod) && !ps.skipList.HasChunk(ch.Address()) {
+				if time.Now().After(ps.warmupPeriod) && !ps.skipList.HasChunk(ch.Address()) && ps.topologyDriver.IsWithinDepth(ch.Address()) {
 					ps.skipList.Add(peer, ch.Address(), skipPeerExpiration)
 				}
 			}
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -329,13 +329,13 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, originate
 			skipPeers = append(skipPeers, peer)
 
 			if ps.skipList.ShouldSkip(peer) {
-				// fmt.Println("skipping peer")
 				ps.metrics.TotalSkippedPeers.Inc()
 				continue
 			}
 
 			ps.metrics.TotalSendAttempts.Inc()
 			attempts++
+
 			go func(peer swarm.Address, ch swarm.Chunk, chunkInNeighbourhood bool) {
 				ctxd, canceld := context.WithTimeout(context.Background(), defaultTTL)
 				defer canceld()
@@ -344,9 +344,6 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, originate
 				// attempted is true if we get past accounting and actually attempt
 				// to send the request to the peer. If we dont get past accounting, we
 				// should not count the retry and try with a different peer again
-				if attempted {
-					allowedRetries--
-				}
 				if err != nil {
 
 					if time.Now().After(ps.warmupPeriod) && !ps.skipList.HasChunk(ch.Address()) && chunkInNeighbourhood && errors.Is(err, context.DeadlineExceeded) {
@@ -364,13 +361,14 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, originate
 				}
 			}(peer, ch, chunkInNeighbourhood)
 		}
+
 		select {
 		case r := <-resultC:
 			results++
 
-			// fmt.Printf("peer: %v\n", peer)
-			// fmt.Println(r.err)
-			// fmt.Println(r.attempted)
+			if r.attempted {
+				allowedRetries--
+			}
 
 			// receipt received for chunk
 			if r.receipt != nil {
@@ -383,47 +381,16 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, originate
 				if results > attempts {
 					return nil, ErrNoPush
 				}
-
-				// fmt.Println(chunkInNeighbourhood)
-				// fmt.Println(!ps.skipList.HasChunk(ch.Address()))
-				// fmt.Println(err)
-				// fmt.Println(time.Now().After(ps.warmupPeriod))
-
-				// if the node has warmed up AND no other closer peer has been tried
-				if time.Now().After(ps.warmupPeriod) && !ps.skipList.HasChunk(ch.Address()) && chunkInNeighbourhood && errors.Is(err, context.DeadlineExceeded) {
-					fmt.Println("skipping peer ", peer)
-					ps.skipList.Add(peer, ch.Address(), skipPeerExpiration)
-				}
-				if r.err != nil && r.attempted {
-					ps.metrics.TotalFailedSendAttempts.Inc()
-					// if the node has warmed up AND no other closer peer has been tried
-					if time.Now().After(ps.warmupPeriod) && !ps.skipList.HasChunk(ch.Address()) && chunkInNeighborhood && errors.Is(err, context.DeadlineExceeded) {
-						ps.skipList.Add(peer, ch.Address(), skipPeerExpiration)
-					}
-				}
 			}
 		case <-ctx.Done():
 			return nil, ctx.Err()
 		case <-sendReceipt:
 			if !signInProgress {
 				signInProgress = true
-				go ps.establishStorage(ch, origin, originated, resultC)
 				allowedRetries++
-			}
-		case <-ctx.Done():
-			fmt.Println(peer)
-			fmt.Println("timeout")
-			return nil, ctx.Err()
-		case <-sendReceipt:
-			fmt.Println(peer)
-			fmt.Println("sendReceipt")
-			if !signInProgress {
-				signInProgress = true
 				go ps.establishStorage(ch, origin, originated, resultC)
-				allowedRetries++
 			}
 		}
-
 	}
 
 	return nil, ErrNoPush
@@ -636,10 +603,18 @@ func (ps *PushSync) establishStorage(ch swarm.Chunk, origin swarm.Address, origi
 			return false, false, nil
 		}
 
+		// here we skip the peer if the peer is closer to the chunk than us
+		// we replicate with peers that are further away than us because we are the strorer
+		if dcmp, _ := swarm.DistanceCmp(ch.Address().Bytes(), peer.Bytes(), ps.address.Bytes()); dcmp == 1 {
+			return false, false, nil
+		}
+
 		if count == nPeersToPushsync {
 			return true, false, nil
 		}
+
 		count++
+
 		go ps.pushToNeighbour(peer, ch, originated)
 		return false, false, nil
 	})
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -124,6 +124,7 @@ func (s *PushSync) Protocol() p2p.ProtocolSpec {
 func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) (err error) {
 	w, r := protobuf.NewWriterAndReader(stream)
 	ctx, cancel := context.WithTimeout(ctx, defaultTTL)
+
 	defer cancel()
 	defer func() {
 		if err != nil {
@@ -205,7 +206,7 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 	span, _, ctx := ps.tracer.StartSpanFromContext(ctx, "pushsync-handler", ps.logger, opentracing.Tag{Key: "address", Value: chunk.Address().String()})
 	defer span.Finish()
 
-	receipt, err := ps.pushToClosest(ctx, chunk, false, p.Address)
+	receipt, err := ps.pushToClosest(ctx, chunk, retryAllow, false, p.Address)
 	if err != nil {
 		if errors.Is(err, topology.ErrWantSelf) {
 			if !storedChunk {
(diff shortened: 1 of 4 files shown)


zelig (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2264)* status: 0 of 1 files reviewed, 9 unresolved discussions (waiting on @metacertain)

___
*[pkg/pushsync/pushsync.go, line 82 at r1](https://reviewable.io/reviews/ethersphere/bee/2264#-MdUJMT-2leIo6MTChyB:-MdUJMT-2leIo6MTChyC:b-k2zdnf) ([raw file](https://github.com/ethersphere/bee/blob/8f347d5951940adb315acbad98fe2377ce073343/pkg/pushsync/pushsync.go#L82)):*
> ```Go
> 
> var defaultTTL = 15 * time.Second // request time to live
> var preemptiveTTL = 9 * time.Second
> ```

call it `sendReceiptDelay`
___
*[pkg/pushsync/pushsync.go, line 127 at r1](https://reviewable.io/reviews/ethersphere/bee/2264#-MdUGLho8ZrDlnDlpJPF:-MdUGLho8ZrDlnDlpJPG:b3kre) ([raw file](https://github.com/ethersphere/bee/blob/8f347d5951940adb315acbad98fe2377ce073343/pkg/pushsync/pushsync.go#L127)):*
> ```Go
> 	w, r := protobuf.NewWriterAndReader(stream)
> 	ctx, cancel := context.WithTimeout(ctx, defaultTTL)
> ```

?
___
*[pkg/pushsync/pushsync.go, line 196 at r1](https://reviewable.io/reviews/ethersphere/bee/2264#-MdUS25SEqO4gu4x4O6n:-MdUS25SEqO4gu4x4O6o:b-h1qxau) ([raw file](https://github.com/ethersphere/bee/blob/8f347d5951940adb315acbad98fe2377ce073343/pkg/pushsync/pushsync.go#L196)):*
> ```Go
> 	// forwarding replication
> 	storedChunk := false
> 	retryAllow := ps.topologyDriver.IsWithinDepth(chunk.Address())
> ```

call this var `inNeighbourhood`.
it becomes retryAllowed only in the context of the pushToClosest call
___
*[pkg/pushsync/pushsync.go, line 260 at r1](https://reviewable.io/reviews/ethersphere/bee/2264#-MdUT02T3H9DmRvsxt_5:-MdUT02T3H9DmRvsxt_6:b-i2ut5s) ([raw file](https://github.com/ethersphere/bee/blob/8f347d5951940adb315acbad98fe2377ce073343/pkg/pushsync/pushsync.go#L260)):*
> ```Go
> // the validity of the receipt.
> func (ps *PushSync) PushChunkToClosest(ctx context.Context, ch swarm.Chunk) (*Receipt, error) {
> 	r, err := ps.pushToClosest(ctx, ch, true, true, swarm.ZeroAddress)
> ```

i think the two types of retry logic should be explicit within the respective functions cos they are different
and pushToClosest should be a synchronous func without retry and with skipPeers optional list
___
*[pkg/pushsync/pushsync.go, line 281 at r1](https://reviewable.io/reviews/ethersphere/bee/2264#-MdUHC_37j0SopgXrnls:-MdUHC_37j0SopgXrnlt:b-9agc06) ([raw file](https://github.com/ethersphere/bee/blob/8f347d5951940adb315acbad98fe2377ce073343/pkg/pushsync/pushsync.go#L281)):*
> ```Go
> 	defer ps.skipList.PruneExpired()
> 
> 	preemptiveLimit := time.After(time.Second * 9)
> ```

you just initialise this as non-nil only if within neighbourhood 

```
var sendReceipt <-chan time.Time
chunkInNeighbourhood := ps.topologyDriver.IsWithinDepth(ch.Address())
if chunkInNeighbourhood {
    sendReceipt = time.After(sendReceiptDelay)
}
```
___
*[pkg/pushsync/pushsync.go, line 365 at r1](https://reviewable.io/reviews/ethersphere/bee/2264#-MdUJqVDDtB1sCWwIXMC:-MdUJqVDDtB1sCWwIXMD:b-7rcqc5) ([raw file](https://github.com/ethersphere/bee/blob/8f347d5951940adb315acbad98fe2377ce073343/pkg/pushsync/pushsync.go#L365)):*
> ```Go
> 		}(peer, ch)
> 
> 		if !chunkInNeighborhood {
> ```

no need for the two cases only what is in the else clause. in the first `sendReceipt` channel is nil and always blocks
___
*[pkg/pushsync/pushsync.go, line 377 at r1](https://reviewable.io/reviews/ethersphere/bee/2264#-MdUWqeA9uRJWgu9zLSX:-MdUWqeA9uRJWgu9zLSY:b-7ubuyc) ([raw file](https://github.com/ethersphere/bee/blob/8f347d5951940adb315acbad98fe2377ce073343/pkg/pushsync/pushsync.go#L377)):*
> ```Go
> 
> 					// if the node has warmed up AND no other closer peer has been tried
> 					if time.Now().After(ps.warmupPeriod) && !ps.skipList.HasChunk(ch.Address()) && chunkInNeighborhood && errors.Is(err, context.DeadlineExceeded) {
> ```

here you only need a local var to force alt routes
___
*[pkg/pushsync/pushsync.go, line 395 at r1](https://reviewable.io/reviews/ethersphere/bee/2264#-MdUWOWRENt_RjBvPlfu:-MdUWOWS-wwIDV8sIOL2:b-2taic) ([raw file](https://github.com/ethersphere/bee/blob/8f347d5951940adb315acbad98fe2377ce073343/pkg/pushsync/pushsync.go#L395)):*
> ```Go
> 					ps.metrics.TotalFailedSendAttempts.Inc()
> 					// if the node has warmed up AND no other closer peer has been tried
> 					if time.Now().After(ps.warmupPeriod) && !ps.skipList.HasChunk(ch.Address()) && chunkInNeighborhood && errors.Is(err, context.DeadlineExceeded) {
> ```

here we need a protocolwide skip list
___
*[pkg/pushsync/pushsync.go, line 607 at r1](https://reviewable.io/reviews/ethersphere/bee/2264#-MdUXH634_oRr-5yCIXM:-MdUXH634_oRr-5yCIXN:b-ql02ln) ([raw file](https://github.com/ethersphere/bee/blob/8f347d5951940adb315acbad98fe2377ce073343/pkg/pushsync/pushsync.go#L607)):*
> ```Go
> }
> 
> func (ps *PushSync) establishStorage(ch swarm.Chunk, origin swarm.Address, originated bool, resultC chan<- *pushResult) {
> ```

this is shared code with errwantself case above

zelig (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2264)* status: all files reviewed, 9 unresolved discussions (waiting on @metacertain)

zelig:
This is becoming a monster, lets  rewrite this properly
i dont want to formally block this, but in fact i want to block :)

metacertain:
> This is becoming a monster, lets rewrite this properly
> i dont want to formally block this, but in fact i want to block :)

this is a work in progress, not the intended state

zelig (commented):
> Reviewed 3 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2264)* status: 3 of 4 files reviewed, 9 unresolved discussions (waiting on @metacertain and @zelig)

___
*[pkg/pushsync/pushsync.go, line 313 at r5](https://reviewable.io/reviews/ethersphere/bee/2264#-Md_na3iBW521FjxH8ic:-Md_na3iBW521FjxH8id:b-8p2yen) ([raw file](https://github.com/ethersphere/bee/blob/13f46cfd18ad5e989c128012497a18bf57fc170f/pkg/pushsync/pushsync.go#L313)):*
> ```Go
> 				}
> 
> 				if !chunkInNeighbourhood {
> ```

here we cannot return with an error right? what if we are still waiting for another earlier roiutine to return?
___
*[pkg/pushsync/pushsync.go, line 318 at r5](https://reviewable.io/reviews/ethersphere/bee/2264#-Md_omVv1yREVY403zNL:-Md_omVv1yREVY403zNM:b-yzd465) ([raw file](https://github.com/ethersphere/bee/blob/13f46cfd18ad5e989c128012497a18bf57fc170f/pkg/pushsync/pushsync.go#L318)):*
> ```Go
> 
> 				signInProgress = true
> 				go ps.establishStorage(ch, origin, originated, resultC)
> ```

origin and originated seem redundant, origin == nil <=> originated == true
___
*[pkg/pushsync/pushsync.go, line 325 at r5](https://reviewable.io/reviews/ethersphere/bee/2264#-Md_oCPC4IN5_cqm-K8C:-Md_oCPC4IN5_cqm-K8D:bpojb24) ([raw file](https://github.com/ethersphere/bee/blob/13f46cfd18ad5e989c128012497a18bf57fc170f/pkg/pushsync/pushsync.go#L325)):*
> ```Go
> 		}
> 
> 		if !peer.Equal(swarm.Address{}) {
> ```

when is this condition  true
___
*[pkg/pushsync/pushsync.go, line 387 at r5](https://reviewable.io/reviews/ethersphere/bee/2264#-Md_qYC1BcEB9OYRGs3S:-Md_qYC1BcEB9OYRGs3T:b-lbo1s1) ([raw file](https://github.com/ethersphere/bee/blob/13f46cfd18ad5e989c128012497a18bf57fc170f/pkg/pushsync/pushsync.go#L387)):*
> ```Go
> 			}
> 		}
> ```

cleanup
___
*[pkg/pushsync/pushsync.go, line 397 at r5](https://reviewable.io/reviews/ethersphere/bee/2264#-Md_qUbZBjOS3da3gNKY:-Md_qUbZBjOS3da3gNKZ:b-lrp1cl) ([raw file](https://github.com/ethersphere/bee/blob/13f46cfd18ad5e989c128012497a18bf57fc170f/pkg/pushsync/pushsync.go#L397)):*
> ```Go
> 	receiptPrice := ps.pricer.PeerPrice(peer, ch.Address())
> 
> 	fmt.Printf("peer: %v\n", peer)
> ```

cleanup

metacertain:
> > Reviewed 3 of 4 files at r5.
> > _[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2264)_ status: 3 of 4 files reviewed, 9 unresolved discussions (waiting on @metacertain and @zelig)

added comments to implementation explaining early return cases and others

pkg/pushsync/pushsync.go
```diff
@@ -365,13 +395,23 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, retryAllo
...
+				if results > attempts {
+					return nil, ErrNoPush
 				}
 			}
 		case <-ctx.Done():
 			return nil, ctx.Err()
+		case <-sendReceipt:
+			// to make the loop wait for the one extra result because of this one time interruption, increase attempt exhaustion limits
+			allowedRetries++
+			i++
```
istae, line 409: maybe a `goto` statement to  a label immediately after the `for` loop could also work?

pkg/pushsync/pushsync.go
```diff
@@ -270,23 +272,42 @@ type pushResult struct {
...
+	chunkInNeighbourhood := ps.topologyDriver.IsWithinDepth(ch.Address())
+	if chunkInNeighbourhood {
+		sendReceipt = time.After(sendReceiptDelay)
+	}
+
 	var (
 		skipPeers      []swarm.Address
 		allowedRetries = 1
-		resultC        = make(chan *pushResult)
+		resultC        = make(chan *pushResult, 1)
```
istae, line 295: Why buffered channel here?

metacertain: it was intended for the concurrent storage establishment / pending push towards peer case, so that they can both write to this channel without blocking the other

pkg/pushsync/pushsync.go
metacertain: could work, intention was to avoid goto statement so loop flow is easier to reason about

istae: or `goto` to top of the `select` which gets called after `go ps.establishStorage(ch, origin, resultC)`

pkg/pushsync/pushsync.go
```diff
@@ -365,13 +397,23 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, retryAllo
...
 			return nil, ctx.Err()
+		case <-sendReceipt:
+			// to make the loop wait for the one extra result because of this one time interruption, increase attempt exhaustion limits
+			allowedRetries++
+			i++
+			// if not in progress yet, establish storage
+			if !signInProgress {
+				signInProgress = true
+				go ps.establishStorage(ch, origin, resultC)
+			}
```
istae, line 416: or `goto` to top of the `select` which gets called after `go ps.establishStorage(ch, origin, resultC)`

pkg/pushsync/pushsync.go
```diff
@@ -365,13 +395,23 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, retryAllo
 			if r.err != nil && r.attempted {
 				ps.metrics.TotalFailedSendAttempts.Inc()
 
-				// if the node has warmed up AND no other closer peer has been tried
-				if time.Now().After(ps.warmupPeriod) && !ps.skipList.HasChunk(ch.Address()) && ps.topologyDriver.IsWithinDepth(ch.Address()) {
-					ps.skipList.Add(peer, ch.Address(), skipPeerExpiration)
+				// if we have received 1 more results than peers attempted, return
+				// this means all attempts including establishing storage have already failed
+				if results > attempts {
```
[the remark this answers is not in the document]
metacertain: not sure that would be advantageous, as that bool is set once to avoid multiple attempts of trying to establish storage, 
and using that would require signinprogress to be set to false to be able to exit based on it? also we dont obviously know if the result was received by the select statement already

intention was to keep track of the attempts and results, and as we can only get 1 result from 1 peer attempt, and one extra from the self establishing, if we ever received more results than attempts, that would mean all attempts concluded already, we can exit

pkg/pushsync/pushsync.go
metacertain: we achieve the same result by skipping the push peer block and entering the select.
intention was to avoid goto to make reasoning about the sequence of events happening in the loop more accessible

pkg/pushsync/pushsync.go
```diff
@@ -191,7 +192,8 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 
 	// forwarding replication
 	storedChunk := false
-	if ps.topologyDriver.IsWithinDepth(chunk.Address()) {
+	inNeighborhood := ps.topologyDriver.IsWithinDepth(chunk.Address())
```
istae, line 195: `storedChunk` should be propagated to `pushToClosest` in the case that the node already stored it.

PR #2289, opened 2021-07-04
ci: auto build openapi spec

acud:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2289)

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -21,11 +21,9 @@ jobs:
       id: checkdocs
       run: |
           changed=false
-          git diff --name-only HEAD^ HEAD
           git diff --name-only HEAD^ HEAD > files.txt
           while IFS= read -r file
           do
-            echo $file
             if [[ $file == openapi/* ]]; then
               echo "detected openapi spec change"
               changed=true
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -10,24 +10,15 @@ on:
 
 jobs:
   build:
-    name: Build
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        go: [1.15]
-        os: [ubuntu-latest]
+    name: Preview
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


mrekucci (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2289)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud)

istae (approved):

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+          if [ $changed == true ]
+          then
+            echo "::set-output name=build_docs::true"
+          else
+            echo "::set-output name=build_docs::false"
+          fi
+    - name: Build the OpenAPI specs
+      if: steps.checkdocs.outputs.build_docs == 'true'
+      uses: acud/openapi-dockerized@v1
```
janos, line 41: This is a very nice action. I would suggest to paremetrize the endpoint and the bucket name for it. Or just to make it explicitly swarm specific by naming it with something that will make it obvious for that.

janos (approved):

.github/workflows/docs.yaml
acud: I agree... I wanted to have it fully parametrized, but it gets a bit more complicated... since I am changing the object ACL, comes the question whether the canned-ACL I'm using also needs to be parametrized, and also the aws CLI tool needs a few tweaks so that it works with DigitalOcean, for example a different endpoint URL:
`aws --endpoint-url https://fra1.digitaloceanspaces.com s3 cp $outfname s3://openapi-specs/`

I am not sure how it will react if a user tries to use it without any concrete endpoint (leading to the endpoint being set as explicitly null for an intended AWS usage). It can further be modularized but then the actual addition of the cli flag to the aws utility must be conditioned too, which is quite ugly. I will add a few additions there, as you suggest :+1:

acud: thanks @janos for the great suggestions. I implemented them as suggested :+1:

PR #2294, opened 2021-07-06
fix: retry in neighbourhood

istae:
Retrial of pushing / establishing storage for in-neighborhood content for the purposes of pushsync

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2294)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -305,8 +305,9 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, retryAllo
 		includeSelf    = ps.isFullNode
 	)
 
-	if retryAllowed {
-		// only originator retries
+	chunkInNeighbourhood := ps.topologyDriver.IsWithinDepth(ch.Address())
+
+	if retryAllowed || chunkInNeighbourhood {
 		allowedRetries = maxPeers
 	}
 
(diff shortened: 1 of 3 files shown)


metacertain (approved):

janos (approved):

acud (commented):
Not so happy about this change since it may cause delays in forwarding receipts up the forwarding chain to the originator. This will cause accounting discrepancies and excess errors when the receipts try to travel the cancelled chain

metacertain:
> Not so happy about this change since it may cause delays in forwarding receipts up the forwarding chain to the originator. This will cause accounting discrepancies and excess errors when the receipts try to travel the cancelled chain

So it is true that this can delay the in neighborhood traffic past original TTLs, but, this change mostly makes sense together with preemptive receipt signing after 9 seconds, so that the first forwarder from the neighborhood who do not have to worry about previous forwarders retrying can established storage itself in time.

Accounting discrepancies are not created when we try to write a result back after upstream has exceeded its TTL, because, if the writeback fails we don't debit, instead we keep track of such undebitable responses on a per-peer ghost balance to throttle such requests.

A good thing to add is to separate the downstream peers context timeout from the parent context, so the retried peers can still respond and debit for the work even if they got the request as a 2nd 3rd etc retry so when the parent context is already partly drained

acud:
> So it is true that this can delay the in neighborhood traffic past original TTLs, but, this change mostly makes sense together with preemptive receipt signing after 9 seconds

To me this `9 seconds` thing is really not a well thought through thing sorry. In a growing network this number is simply something we're gonna have to patch up sooner than later. I prefer to postpone this change to a later stage with a well thought through strategy.

acud (changes requested):
Blocking on this for now.

zelig (approved):
> Reviewed 1 of 3 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2294)* status: all files reviewed, 1 unresolved discussion (waiting on @istae)

___
*[pkg/pushsync/pushsync.go, line 335 at r3](https://reviewable.io/reviews/ethersphere/bee/2294#-Me3l6vCBqI9NW8ulGkR:-Me3l6vDEzv6EOsntUHd:b-2urwzr) ([raw file](https://github.com/ethersphere/bee/blob/814a6c1f9831041084865a3c3d45fe7d884a6bf5/pkg/pushsync/pushsync.go#L335)):*
> ```Go
> 				count := 0
> 				_ = ps.topologyDriver.EachNeighbor(func(peer swarm.Address, po uint8) (bool, bool, error) {
> 					// skip forwarding peer
> ```

misleading comment,  should read: skip peer sent us the request

PR #2296, opened 2021-07-06
fix: least authority gcSizeChange bug

acud:
as reported by the Least Authority audit team, we would be doing unnecessary increments to gcSize when uploading. A subsequent migration that reindexes the gcSize index is needed

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2296)

### DIFF
--- a/pkg/localstore/mode_put.go
+++ b/pkg/localstore/mode_put.go
@@ -137,8 +137,8 @@ func (db *DB) put(mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err e
 				if err != nil {
 					return nil, err
 				}
+				gcSizeChange += c
 			}
-			gcSizeChange += c
 		}
 
 	case storage.ModePutSync:
--- a/pkg/localstore/mode_put_test.go
+++ b/pkg/localstore/mode_put_test.go
@@ -260,6 +260,7 @@ func TestModePutUpload(t *testing.T) {
 				newPinIndexTest(db, ch, leveldb.ErrNotFound)(t)
 			}
 			newItemsCountTest(db.postageIndexIndex, tc.count)(t)
+			newIndexGCSizeTest(db)(t)
 		})
 	}
 }
@@ -298,6 +299,7 @@ func TestModePutUploadPin(t *testing.T) {
 				newPinIndexTest(db, ch, nil)(t)
 			}
 			newItemsCountTest(db.postageIndexIndex, tc.count)(t)
+			newIndexGCSizeTest(db)(t)
 		})
 	}
 }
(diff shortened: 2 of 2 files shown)


notanatol (approved):

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2296)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @acud, @aloknerurkar, @istae, and @zelig)

___
*[pkg/localstore/mode_put.go, line 140 at r1](https://reviewable.io/reviews/ethersphere/bee/2296#-MdwZJaXD2JP-Cff8bKy:-MdwZJaYARNdBi6R6Gkf:b-w0df4b) ([raw file](https://github.com/ethersphere/bee/blob/72599c47392431a771c59a54444af1b35ba88271/pkg/localstore/mode_put.go#L140)):*
> ```Go
> 					return nil, err
> 				}
> 				gcSizeChange += c
> ```

The `gcSizeChange` is already incremented before the `if` block; could you provide some context why it's also needed to be incremented inside the block for the second time?

acud (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2296)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @aloknerurkar, @istae, @mrekucci, and @zelig)

___
*[pkg/localstore/mode_put.go, line 140 at r1](https://reviewable.io/reviews/ethersphere/bee/2296#-MdwZJaXD2JP-Cff8bKy:-MdzbF_J0Qm-txkaNOag:b-9qp90e) ([raw file](https://github.com/ethersphere/bee/blob/72599c47392431a771c59a54444af1b35ba88271/pkg/localstore/mode_put.go#L140)):*

The `gcSizeChange` is already incremented before the `if` block; could you provide some context why it's also needed to be incremented inside the block for the second time?
</blockquote></details>

since `c` would be set to a different value from `setPin` - either `0` or `-1`. This means we need to increment `gcSizeChange` with that value again to reflect the current state of the gc index

aloknerurkar (approved):

mrekucci (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2296)* status: all files reviewed, 1 unresolved discussion (waiting on @acud, @istae, and @zelig)

___
*[pkg/localstore/mode_put.go, line 140 at r1](https://reviewable.io/reviews/ethersphere/bee/2296#-MdwZJaXD2JP-Cff8bKy:-Me--qDSA6k7FT3ngfCf:byd9ov7) ([raw file](https://github.com/ethersphere/bee/blob/72599c47392431a771c59a54444af1b35ba88271/pkg/localstore/mode_put.go#L140)):*

since `c` would be set to a different value from `setPin` - either `0` or `-1`. This means we need to increment `gcSizeChange` with that value again to reflect the current state of the gc index
</blockquote></details>

Yes, I'm sorry I missed that.

PR #2311, opened 2021-07-07
fix: add to skiplist on error in goroutine

istae:
1. Adding a peer to the skip list has been moved to the pushPeer goroutine because previously (this is the bug) peers were not added to the skiplist if the context timed out.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2311)


pkg/pushsync/pushsync.go
```diff
@@ -369,6 +371,12 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, retryAllo
 			}
 			if err != nil {
 				logger.Debugf("could not push to peer %s: %v", peer, err)
+
```
janos, line 374: It may be that this change affected the TestReplicateBeforeReceipt test.

istae: So we discovered that the granularity of the time pkg is not enough such that the warmup check returns false in some cases. We fixed it by changing the warmup time to some negative number. Tests should pass in the new commit.

pkg/pushsync/pushsync_test.go
```diff
@@ -929,16 +926,16 @@ func createPushSyncNodeWithAccounting(t *testing.T, addr swarm.Address, prices p
 		return ch, nil
 	}
 
-	return pushsync.New(addr, blockHash.Bytes(), recorderDisconnecter, storer, mockTopology, mtag, true, unwrap, validStamp, logger, acct, mockPricer, signer, nil, 0), storer, mtag
+	return pushsync.New(addr, blockHash.Bytes(), recorderDisconnecter, storer, mockTopology, mtag, true, unwrap, validStamp, logger, acct, mockPricer, signer, nil, -1), storer, mtag
 }
 
-func waitOnRecordAndTest(t *testing.T, peer swarm.Address, recorder *streamtest.Recorder, add swarm.Address, data []byte) {
+func waitOnRecordNAndTest(t *testing.T, peer swarm.Address, recorder *streamtest.Recorder, add swarm.Address, data []byte, n int) {
```
acud, line 932: i wonder why this is needed... there is no usage of this method anywhere. leftovers?

acud (changes requested):

pkg/pushsync/pushsync_test.go
istae: leftover, yes. I can remove :)

acud (approved):
Thanks @istae :pray:

mrekucci (approved):
> Reviewed 1 of 2 files at r2, 1 of 2 files at r3, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2311)* status: all files reviewed, 2 unresolved discussions (waiting on @acud, @aloknerurkar, @janos, and @metacertain)

___
*[pkg/topology/mock/mock.go, line 152 at r5](https://reviewable.io/reviews/ethersphere/bee/2311#-Me5XH7b-KlXu1QsP2oM:-Me5XH7b-KlXu1QsP2oN:bp3ecbd) ([raw file](https://github.com/ethersphere/bee/blob/0ac1eea91cae3b53c75e08ab83941c706cf582d7/pkg/topology/mock/mock.go#L152)):*
> ```Go
> 		if wantSelf {
> 			return peerAddr, topology.ErrWantSelf
> 		} else {
> ```

The else isn't needed here, you can return after the `if` block.

janos (approved):

Issue #829, opened 2020-10-10
http: panic serving 127.0.0.1:58978: runtime error: invalid memory address or nil pointer dereference


Baleba:
Hi there,

was trying to upload a 5.5Mb  video file. 

**My configuration:**

raspberry pi 4 
Linux ubuntu 5.8.0+ #1 SMP PREEMPT Fri Sep 25 22:26:37 UTC 2020 aarch64 aarch64 aarch64 GNU/Linux

**bee version**

```
ubuntu@ubuntu:~/.local$ bee version
0.3.2-9d7d98f

```

**upload console**

```
ubuntu@ubuntu:/mnt/beecachefs/ubuntu/Videos/BlickBassy$ curl -F file=@'BLICK BASSY   NLAL.m4a' http://localhost:8080/files
curl: (52) Empty reply from server
```
got the above on the upload console and the errors below on the bee console. Checked memory, there doesn't seem to be a problem with memory.

**Memory conf:**

```
ubuntu@ubuntu:~/.local$ free -h
              total        used        free      shared  buff/cache   available
Mem:          1.8Gi       647Mi       308Mi       0.0Ki       892Mi       1.2Gi
Swap:         2.0Gi        62Mi       1.9Gi
```
**bee console**

```
ERRO[2020-10-10T11:30:21Z] http: panic serving 127.0.0.1:58978: runtime error: invalid memory address or nil pointer dereference 
ERRO[2020-10-10T11:30:21Z] goroutine 22502 [running]:                   
ERRO[2020-10-10T11:30:21Z] net/http.(*conn).serve.func1(0x40020d8fa0)   
ERRO[2020-10-10T11:30:21Z] 	net/http/server.go:1801 +0xfc               
ERRO[2020-10-10T11:30:21Z] panic(0xcfc340, 0x1a8bb50)                   
ERRO[2020-10-10T11:30:21Z] 	runtime/panic.go:975 +0x370                 
ERRO[2020-10-10T11:30:21Z] math/big.(*Int).Mul(0x400188c638, 0x0, 0x400188c738, 0x0) 
ERRO[2020-10-10T11:30:21Z] 	math/big/int.go:168 +0xcc                   
ERRO[2020-10-10T11:30:21Z] github.com/btcsuite/btcd/btcec.recoverKeyFromSignature(0x1ac41a0, 0x400188c728, 0x4002616280, 0x20, 0x20, 0x2, 0x40033b1700, 0x1, 0x8, 0x70af14) 
[...]

PR #2316, opened 2021-07-08
chore: update dependency versions

acud:
Updating the libp2p version, among other dependencies

fixes https://github.com/ethersphere/bee/issues/829 

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2316)


notanatol (approved):

acud:
@istae I remember that you also went through some dependency update process and there was a problem... Do you remember what it was and if it could happen again in this changeset?

istae:
> @istae I remember that you also went through some dependency update process and there was a problem... Do you remember what it was and if it could happen again in this changeset?

I tried updating go-ethereum but @ralph-pichler reverted it because it was incompatible with our api client. The updates in this PR should be OK as far as I know.

istae (approved):

PR #2319, opened 2021-07-08
feat(hive, libp2p): check reachability while processing peer from hive

aloknerurkar:
Currently, we add all the addresses we get from hive to the addressbook and then signal kademlia to connect to them.

We see a lot of addresses passed through hive are not dialable/reachable. Each address could be tried multiple times before we discard it. Even if we discard the address, the statestore compaction may not happen and we see addressbooks bloating in size.

This change introduces the following:

- libp2p ping functionality
    libp2p has a default ping protocol. This is useful for us as we don't need to connect to the peer before we do this. Ping is exported as a new interface function to our libp2p abstraction.

- hive checking addresses using ping
     As a result, we can check the sanity of the addresses passed to us before we commit them into our addressbook using ping.

## Updates to implementation

- Use a separate libp2p dialer instance for doing ping
   Using the default libp2p instance for doing ping will eventually cause inconsistencies in the libp2p::peerstore and addressbook. As ping will open a new fresh stream, we would see peer in the other peer's peerstore prematurely. As a result, we will use a separate libp2p dialer instance (NoListenAddresses) which will use random keys to do the ping. This address is not dialable hence if it percolates into any peerstore, it should be subsequently cleaned up.

- Reduce kademlia routines used to dial out
   Currently, we start around `64 + 8` goroutines in kademlia to dial out to all known peers. Due to the ping check, we will reduce the addresses we add to knownPeers list. As a result, kademlia will have substantially less peers to try. So we would not require as many routines going ahead.

- Add dedicated routines for ping
   As we have divided some of the work of kademlia to hive to check peer addresses, we need to introduce some logic around the lifecycle of the checking to properly close when we shutdown the node. We should provision the routines on demand, however we should ensure we don't overload the system. Hence, semaphore is now global and with a bit more capacity.
   
## Metrics
Screenshot 2021-07-15 at 1 53 41 PM

As we see here, the third node is running the new build, known peers count has reduced significantly reduced (~80%)

Screenshot 2021-07-15 at 1 52 54 PM Screenshot 2021-07-15 at 1 53 12 PM

Address book size is consistently 50% of what it was.

Screenshot 2021-07-15 at 2 57 04 PM

Kademlia build-up time is not affected (blue => this change orange => old build restarted together)

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2319)

### DIFF
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -201,13 +201,16 @@ func (s *Service) checkAndAddPeers(peers pb.Peers) {
 	sem := make(chan struct{}, 5)
 	var peersToAdd []swarm.Address
 	mtx := sync.Mutex{}
+	wg := sync.WaitGroup{}
 
 	for i := range peers.Peers {
 		sem <- struct{}{}
 
+		wg.Add(1)
 		go func(idx int) {
 			defer func() {
 				<-sem
+				wg.Done()
 			}()
 
 			newPeer := peers.Peers[idx]
@@ -244,8 +247,9 @@ func (s *Service) checkAndAddPeers(peers pb.Peers) {
 		}(i)
 	}
 
-	if s.addPeersHandler != nil {
+	wg.Wait()
+
+	if s.addPeersHandler != nil && len(peersToAdd) > 0 {
 		s.addPeersHandler(peersToAdd...)
 	}
-
 }
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -205,8 +205,8 @@ func (s *Service) peersHandler(ctx context.Context, peer p2p.Peer, stream p2p.St
 
 	select {
 	case s.peersChan <- peersReq:
-	case <-ctx.Done():
-		return fmt.Errorf("failed to process peers, reason: %w", ctx.Err())
+	case <-s.quit:
+		return errors.New("failed to process peers, shutting down hive")
 	}
 
 	return nil
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -201,7 +201,7 @@ func (s *Service) checkAndAddPeers(peers pb.Peers) {
 	sem := make(chan struct{}, 5)
 	var peersToAdd []swarm.Address
 
-	for i, _ := range peers.Peers {
+	for i := range peers.Peers {
 		sem <- struct{}{}
 
 		go func(idx int) {
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -44,7 +44,7 @@ var (
 )
 
 type Service struct {
-	streamer        p2p.Streamer
+	streamer        p2p.StreamerPinger
 	addressBook     addressbook.GetPutter
 	addPeersHandler func(...swarm.Address)
 	networkID       uint64
@@ -55,7 +55,7 @@ type Service struct {
 	clearMtx        sync.Mutex
 }
 
-func New(streamer p2p.Streamer, addressbook addressbook.GetPutter, networkID uint64, logger logging.Logger) *Service {
+func New(streamer p2p.StreamerPinger, addressbook addressbook.GetPutter, networkID uint64, logger logging.Logger) *Service {
 	return &Service{
 		streamer:    streamer,
 		logger:      logger,
@@ -176,34 +176,7 @@ func (s *Service) peersHandler(ctx context.Context, peer p2p.Peer, stream p2p.St
 	// but we still want to handle not closed stream from the other side to avoid zombie stream
 	go stream.FullClose()
 
-	var peers []swarm.Address
-	for _, newPeer := range peersReq.Peers {
-
-		multiUnderlay, err := ma.NewMultiaddrBytes(newPeer.Underlay)
-		if err != nil {
-			s.logger.Errorf("hive: multi address underlay err: %v", err)
-			continue
-		}
-
-		bzzAddress := bzz.Address{
-			Overlay:     swarm.NewAddress(newPeer.Overlay),
-			Underlay:    multiUnderlay,
-			Signature:   newPeer.Signature,
-			Transaction: newPeer.Transaction,
-		}
-
-		err = s.addressBook.Put(bzzAddress.Overlay, bzzAddress)
-		if err != nil {
-			s.logger.Warningf("skipping peer in response %s: %v", newPeer.String(), err)
-			continue
-		}
-
-		peers = append(peers, bzzAddress.Overlay)
-	}
-
-	if s.addPeersHandler != nil {
-		s.addPeersHandler(peers...)
-	}
+	go s.checkAndAddPeers(peersReq)
 
 	return nil
 }
@@ -217,4 +190,59 @@ func (s *Service) disconnect(peer p2p.Peer) error {
 	s.outLimiter.Clear(peer.Address.ByteString())
 
 	return nil
+}
+
+const checkPeersTimeout = time.Minute * 15
+
+func (s *Service) checkAndAddPeers(peers pb.Peers) {
+	ctx, cancel := context.WithTimeout(context.Background(), checkPeersTimeout)
+	defer cancel()
+
+	sem := make(chan struct{}, 5)
+	var peersToAdd []swarm.Address
+
+	for i, _ := range peers.Peers {
+		sem <- struct{}{}
+
+		go func(idx int) {
+			defer func() {
+				<-sem
+			}()
+
+			newPeer := peers.Peers[idx]
+
+			multiUnderlay, err := ma.NewMultiaddrBytes(newPeer.Underlay)
+			if err != nil {
+				s.logger.Errorf("hive: multi address underlay err: %v", err)
+				return
+			}
+
+			// check if the underlay is usable by doing a raw ping using libp2p
+			_, err = s.streamer.Ping(ctx, multiUnderlay)
+			if err != nil {
+				s.logger.Errorf("hive: multi address underlay %s not reachable err: %w", multiUnderlay, err)
+				return
+			}
+
+			bzzAddress := bzz.Address{
+				Overlay:     swarm.NewAddress(newPeer.Overlay),
+				Underlay:    multiUnderlay,
+				Signature:   newPeer.Signature,
+				Transaction: newPeer.Transaction,
+			}
+
+			err = s.addressBook.Put(bzzAddress.Overlay, bzzAddress)
+			if err != nil {
+				s.logger.Warningf("skipping peer in response %s: %v", newPeer.String(), err)
+				return
+			}
+
+			peersToAdd = append(peersToAdd, bzzAddress.Overlay)
+		}(i)
+	}
+
+	if s.addPeersHandler != nil {
+		s.addPeersHandler(peersToAdd...)
+	}
+
 }
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -71,7 +71,7 @@ func New(streamer p2p.StreamerPinger, addressbook addressbook.GetPutter, network
 		outLimiter:  ratelimit.New(limitRate, limitBurst),
 		quit:        make(chan struct{}),
 		peersChan:   make(chan pb.Peers),
-		sem:         semaphore.NewWeighted(int64(15)),
+		sem:         semaphore.NewWeighted(int64(31)),
 	}
 	svc.startCheckPeersHandler()
 	return svc
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -225,7 +225,7 @@ func (s *Service) checkAndAddPeers(peers pb.Peers) {
 			_, err = s.streamer.Ping(ctx, multiUnderlay)
 			if err != nil {
 				s.metrics.UnreachablePeers.Inc()
-				s.logger.Errorf("hive: multi address underlay %s not reachable err: %w", multiUnderlay, err)
+				s.logger.Warningf("hive: multi address underlay %s not reachable err: %w", multiUnderlay, err)
 				return
 			}
 
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -256,20 +256,18 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
 	mtx := sync.Mutex{}
 	wg := sync.WaitGroup{}
 
-	for i := range peers.Peers {
+	for _, p := range peers.Peers {
 		err := s.sem.Acquire(ctx, 1)
 		if err != nil {
 			return
 		}
 
 		wg.Add(1)
-		go func(idx int) {
+		go func(newPeer *pb.BzzAddress) {
 			defer func() {
 				s.sem.Release(1)
 				wg.Done()
 			}()
-
-			newPeer := peers.Peers[idx]
 
 			multiUnderlay, err := ma.NewMultiaddrBytes(newPeer.Underlay)
 			if err != nil {
@@ -301,7 +299,7 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
 			mtx.Lock()
 			peersToAdd = append(peersToAdd, bzzAddress.Overlay)
 			mtx.Unlock()
-		}(i)
+		}(p)
 	}
 
 	wg.Wait()
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -14,6 +14,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"golang.org/x/sync/semaphore"
 	"sync"
 	"time"
 
@@ -53,10 +54,14 @@ type Service struct {
 	inLimiter       *ratelimit.Limiter
 	outLimiter      *ratelimit.Limiter
 	clearMtx        sync.Mutex
+	quit            chan struct{}
+	wg              sync.WaitGroup
+	peersChan       chan pb.Peers
+	sem             *semaphore.Weighted
 }
 
 func New(streamer p2p.StreamerPinger, addressbook addressbook.GetPutter, networkID uint64, logger logging.Logger) *Service {
-	return &Service{
+	svc := &Service{
 		streamer:    streamer,
 		logger:      logger,
 		addressBook: addressbook,
@@ -64,7 +69,12 @@ func New(streamer p2p.StreamerPinger, addressbook addressbook.GetPutter, network
 		metrics:     newMetrics(),
 		inLimiter:   ratelimit.New(limitRate, limitBurst),
 		outLimiter:  ratelimit.New(limitRate, limitBurst),
+		quit:        make(chan struct{}),
+		peersChan:   make(chan pb.Peers),
+		sem:         semaphore.NewWeighted(int64(15)),
 	}
+	svc.startCheckPeersHandler()
+	return svc
 }
 
 func (s *Service) Protocol() p2p.ProtocolSpec {
@@ -109,6 +119,23 @@ func (s *Service) BroadcastPeers(ctx context.Context, addressee swarm.Address, p
 
 func (s *Service) SetAddPeersHandler(h func(addr ...swarm.Address)) {
 	s.addPeersHandler = h
+}
+
+func (s *Service) Close() error {
+	close(s.quit)
+
+	stopped := make(chan struct{})
+	go func() {
+		defer close(stopped)
+		s.wg.Wait()
+	}()
+
+	select {
+	case <-stopped:
+		return nil
+	case <-time.After(time.Second * 5):
+		return errors.New("hive: waited 5 seconds to close active goroutines")
+	}
 }
 
 func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swarm.Address) (err error) {
@@ -176,7 +203,11 @@ func (s *Service) peersHandler(ctx context.Context, peer p2p.Peer, stream p2p.St
 	// but we still want to handle not closed stream from the other side to avoid zombie stream
 	go stream.FullClose()
 
-	go s.checkAndAddPeers(peersReq)
+	select {
+	case s.peersChan <- peersReq:
+	case <-ctx.Done():
+		return fmt.Errorf("failed to process peers, reason: %w", ctx.Err())
+	}
 
 	return nil
 }
@@ -192,24 +223,49 @@ func (s *Service) disconnect(peer p2p.Peer) error {
 	return nil
 }
 
-const checkPeersTimeout = time.Minute * 15
+func (s *Service) startCheckPeersHandler() {
+	ctx, cancel := context.WithCancel(context.Background())
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+		<-s.quit
+		cancel()
+	}()
 
-func (s *Service) checkAndAddPeers(peers pb.Peers) {
-	ctx, cancel := context.WithTimeout(context.Background(), checkPeersTimeout)
-	defer cancel()
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case newPeers := <-s.peersChan:
+				s.wg.Add(1)
+				go func() {
+					defer s.wg.Done()
+					s.checkAndAddPeers(ctx, newPeers)
+				}()
+			}
+		}
+	}()
+}
 
-	sem := make(chan struct{}, 5)
+func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
+
 	var peersToAdd []swarm.Address
 	mtx := sync.Mutex{}
 	wg := sync.WaitGroup{}
 
 	for i := range peers.Peers {
-		sem <- struct{}{}
+		err := s.sem.Acquire(ctx, 1)
+		if err != nil {
+			return
+		}
 
 		wg.Add(1)
 		go func(idx int) {
 			defer func() {
-				<-sem
+				s.sem.Release(1)
 				wg.Done()
 			}()
 
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -224,6 +224,7 @@ func (s *Service) checkAndAddPeers(peers pb.Peers) {
 			// check if the underlay is usable by doing a raw ping using libp2p
 			_, err = s.streamer.Ping(ctx, multiUnderlay)
 			if err != nil {
+				s.metrics.UnreachablePeers.Inc()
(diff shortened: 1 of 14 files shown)


pkg/hive/hive.go
```diff
@@ -218,3 +191,58 @@ func (s *Service) disconnect(peer p2p.Peer) error {
...
+			if err != nil {
+				s.logger.Warningf("skipping peer in response %s: %v", newPeer.String(), err)
+				return
+			}
+
+			peersToAdd = append(peersToAdd, bzzAddress.Overlay)
+		}(i)
+	}
+
+	if s.addPeersHandler != nil {
```
aloknerurkar, line 244: I think we could do this as a defer. So for some reason if the above attempts timeout after some are done, we would still add some addresses to the book.

pkg/hive/hive.go
```diff
@@ -218,3 +191,61 @@ func (s *Service) disconnect(peer p2p.Peer) error {
...
+				s.logger.Warningf("skipping peer in response %s: %v", newPeer.String(), err)
+				return
+			}
+
+			mtx.Lock()
+			peersToAdd = append(peersToAdd, bzzAddress.Overlay)
+			mtx.Unlock()
+		}(i)
+	}
+
```
janos, line 246: I think that a wait group is needed to wait here for all goroutines to finish before calling addPeersHandler or even exiting the function.

janos (commented):
Nice utilization of the libp2p ping protocol. 👍

pkg/hive/hive.go
aloknerurkar: Yes I just realized that, had come to comment on this... This is why the integration tests are also failing I believe.

I'll use an errgroup as I need it to obey the context as well.

janos (approved):
Great! Thanks for addressing my comment.

istae (approved):
Great idea! Thanks 👏

acud:
@janos, @aloknerurkar, it had occurred to me that there could be possible side-effects:
- adding a peer to the peerstore might invoke the `connected` hook notifier in the libp2p abstraction, leading to the peer being listed as connected to, while in fact it is not
- this might have all sorts of unintended consequences if kademlia also tries to connect at the same time, receiving an `ErrAlreadyConnected`, resulting in the peer being listed in the connected peers list in kademlia

If it would be possible to overrule this possible problem by checking how the libp2p notifies the registered hooks in case we manually add an entry to the peerstore. Otherwise we might have to do this check in some other way, like just opening a raw TCP socket to the target host+port and see if there's any response from the other side

aloknerurkar:
> * connected

@acud @janos 
I dont see where we are reacting to the `Connected` hook. I see handling for `Disconnected` hook in handshake and peerRegistry. Both seem to be using the NoopNotifiee for Connected. As long as peerRegistry has not seen the peer, we should be okay I think.

istae:
> @janos, @aloknerurkar, it had occurred to me that there could be possible side-effects:
> 
> * adding a peer to the peerstore might invoke the `connected` hook notifier in the libp2p abstraction, leading to the peer being listed as connected to, while in fact it is not
> * this might have all sorts of unintended consequences if kademlia also tries to connect at the same time, receiving an `ErrAlreadyConnected`, resulting in the peer being listed in the connected peers list in kademlia
> 
> If it would be possible to overrule this possible problem by checking how the libp2p notifies the registered hooks in case we manually add an entry to the peerstore. Otherwise we might have to do this check in some other way, like just opening a raw TCP socket to the target host+port and see if there's any response from the other side

Another issue, just realized, is peers that sit behind a NAT router and are not connectable with inbound connections?

acud:
That's why we want to check whether they are reachable before adding them to the addressbook

acud:
> > * connected
> 
> @acud @janos
> I dont see where we are reacting to the `Connected` hook. I see handling for `Disconnected` hook in handshake and peerRegistry. Both seem to be using the NoopNotifiee for Connected. As long as peerRegistry has not seen the peer, we should be okay I think.

True - we don't react to this event explicitly, however consider:
- peer A enters peer B into the peerstore to do the validation check (does not "officially" dial)
- peer B tries to dial to peer A while the check is happening
- handshake process starts (and even finishes) between the peers but then peer A finishes the reachability check

Now there are possible ways on how to finish the process:
- peer A should clean up the peer from the peerstore (use `ClosePeer`), since the reachability check does not have notion of whether we are connected to the peer. The peer might already bubble to higher abstraction layers and be used. This might lead to the handshake being terminated also (or an already established connection to do so), since we are effectively disconnecting from the peer, therefore contributing to network instability
- peer A should leave the peerstore as it is, leaving the cleanup to libp2p (and the question here would be - is this synchronous and if not - how often is it done?)

I think that cleaning up would be the right way to go, but it means we're gonna have to do a bit more in the reachability check so that it doesn't leak or do harm. Or a plain TCP socket :)

aloknerurkar:
> > > * connected
> > 
> > 
> > @acud @janos
> > I dont see where we are reacting to the `Connected` hook. I see handling for `Disconnected` hook in handshake and peerRegistry. Both seem to be using the NoopNotifiee for Connected. As long as peerRegistry has not seen the peer, we should be okay I think.
> 
> True - we don't react to this event explicitly, however consider:
> 
> * peer A enters peer B into the peerstore to do the validation check (does not "officially" dial)
> * peer B tries to dial to peer A while the check is happening
> * handshake process starts (and even finishes) between the peers but then peer A finishes the reachability check
> 
> Now there are possible ways on how to finish the process:
> 
> * peer A should clean up the peer from the peerstore (use `ClosePeer`), since the reachability check does not have notion of whether we are connected to the peer. The peer might already bubble to higher abstraction layers and be used. This might lead to the handshake being terminated also (or an already established connection to do so), since we are effectively disconnecting from the peer, therefore contributing to network instability
> * peer A should leave the peerstore as it is, leaving the cleanup to libp2p (and the question here would be - is this synchronous and if not - how often is it done?)
> 
> I think that cleaning up would be the right way to go, but it means we're gonna have to do a bit more in the reachability check so that it doesn't leak or do harm. Or a plain TCP socket :)

Plain TCP socket will not work right. Peer can have non-TCP addresses. These are perfectly valid.

acud:
> may even happen on the same stream

it would definitely not happen on the same stream, as each protocol has its own stream which are then multiplexed into one connection.

> If the reachability check succeeds, in all possibility the handshake will also go through

not exactly. one peer may want to check if another is reachable but would not dial to that peer for a while because that bin is saturated, and the other peer would like to dial to the peer since it needs a peer in the bin. in such case if the first peer does a cleanup then we might get more errors while trying to establish connections from some peers.

It would be good to have a look at how libp2p does the cleanup on peerstore, then we could make an educated guess (although race conditions would still be there)

aloknerurkar:
> > may even happen on the same stream
> 
> it would definitely not happen on the same stream, as each protocol has its own stream which are then multiplexed into one connection. if there is a cleanup though, that connection is at risk of being removed
> 

Okay yes... but then this is even better actually... we do a Close on the stream when we are done with the reachability check. This means if a handshake is attempted at the same time the close should not affect it right? There is no other cleanup done here.

> > If the reachability check succeeds, in all possibility the handshake will also go through
> 
> not exactly. one peer may want to check if another is reachable but would not dial to that peer for a while because that bin is saturated, and the other peer would like to dial to the peer since it needs a peer in the bin. in such case if the first peer does a cleanup then we might get more errors while trying to establish connections from some peers.
> 
> It would be good to have a look at how libp2p does the cleanup on peerstore, then we could make an educated guess (although race conditions would still be there)

libp2p in-mem peerstore which we use has a background thread that cleans up addresses every hour. However, the `Addrs()` call will not return expired entries. If the peer does not do handshake for some time and the peerstore entry is cleaned up, this is not a problem as the addressbook entry has the underlay information and it will use that to connect anyway.

janos:
@aloknerurkar @acud your points are all good, from what I can see. Adding peer to the libp2p peerstore may result with some undesired consequences, if not now, then in the future. That makes the code a bit less maintainable, mostly as we do not know which components will be listening on libp2p notifier notifications. Trying to make a partial or full TCP handshake to validate the connectivity is also not so great as other protocols are involved. I've got a strange idea to construct a second libp2p host that is not connected to swarm specific components but configured just enough to make a ping protocol stream to the peer and validate the connectivity for whatever transport it is using. There should be no listen addrs to the second host and my assumption that it should be possible as we will only make connections to other peers. Second libp2p host instance should provide full isolation. The idea is adding another important and complex instance, but I am currently out of other ideas.

acud:
> @aloknerurkar @acud your points are all good, from what I can see. Adding peer to the libp2p peerstore may result with some undesired consequences, if not now, then in the future. That makes the code a bit less maintainable, mostly as we do not know which components will be listening on libp2p notifier notifications. Trying to make a partial or full TCP handshake to validate the connectivity is also not so great as other protocols are involved. I've got a strange idea to construct a second libp2p host that is not connected to swarm specific components but configured just enough to make a ping protocol stream to the peer and validate the connectivity for whatever transport it is using. There should be no listen addrs to the second host and my assumption that it should be possible as we will only make connections to other peers. Second libp2p host instance should provide full isolation. The idea is adding another important and complex instance, but I am currently out of other ideas.

Thanks @janos, I think this would be the best idea, since if we use an ephemeral key store for the dialer libp2p instance, since the underlying encryption keys of the different instances would mean that the downstream peers will not keep dirty entries regarding the actual upstream peer (since the identities of the dialers are different). Some dirty entries are indeed possible, but they refer to the ephemeral node, so probably we're good :+1:

istae:
> That's why we want to check whether they are reachable before adding them to the addressbook

Funny enough, if a node of this type connects with an incoming connection, we put the peer in the address book. Maybe we can do pinging attempt there too?

acud (changes requested):
Lovely thank you! I have a few minor comments

> Reviewed 3 of 7 files at r1, 9 of 9 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2319#-:-MedSPvB1ST-6TBb1o1z:be9dp2z)* status: all files reviewed, 5 unresolved discussions (waiting on @aloknerurkar, @istae, and @janos)

___
*[pkg/hive/hive.go, line 244 at r3](https://reviewable.io/reviews/ethersphere/bee/2319#-MedQ6dm1BBUmd84FgGL:-MedQ6dm1BBUmd84FgGM:b-77ugwc) ([raw file](https://github.com/ethersphere/bee/blob/20fe011a728bc9ed71b33df931d0c68bb5b5873d/pkg/hive/hive.go#L244)):*
> ```Go
> 			case newPeers := <-s.peersChan:
> 				s.wg.Add(1)
> 				go func() {
> ```

since you are spawning a new goroutine on every peer which is received, this can increase stress on GC at some point. consider having a pool of workers spawned in advance similarly to kademlia. They can just directly read off the newPeers channel and do the check
___
*[pkg/hive/hive.go, line 260 at r3](https://reviewable.io/reviews/ethersphere/bee/2319#-MedQX8rALbuNH_wDret:-MedQX8sA58AWRVaFXRp:b-aiu9kf) ([raw file](https://github.com/ethersphere/bee/blob/20fe011a728bc9ed71b33df931d0c68bb5b5873d/pkg/hive/hive.go#L260)):*
> ```Go
> 
> 	for i := range peers.Peers {
> 		err := s.sem.Acquire(ctx, 1)
> ```

not sure why the weighted semaphore is needed as the resource and the load are linear. what are the benefits over a regular buffered channel?
___
*[pkg/hive/hive.go, line 282 at r3](https://reviewable.io/reviews/ethersphere/bee/2319#-MedREaw1HF609g_Gk4D:-MedREaw1HF609g_Gk4E:b-cm2lzz) ([raw file](https://github.com/ethersphere/bee/blob/20fe011a728bc9ed71b33df931d0c68bb5b5873d/pkg/hive/hive.go#L282)):*
> ```Go
> 			// check if the underlay is usable by doing a raw ping using libp2p
> 			_, err = s.streamer.Ping(ctx, multiUnderlay)
> 			if err != nil {
> ```

nit, it might be beneficial to check which types of errors can be returned here, since different errors could signify different assumptions about reachability

pkg/p2p/p2p.go
```diff
@@ -70,6 +70,18 @@ type StreamerDisconnecter interface {
 	Disconnecter
 }
 
+// Pinger interface is used to ping a underlay address which is not yet known to the bee node.
+// It uses libp2p's default ping protocol. This is different from the PinPong protocol as this
```
janos, line 74: PinPong -> PingPong

aloknerurkar:
___
*[pkg/hive/hive.go, line 246 at r2](https://reviewable.io/reviews/ethersphere/bee/2319#-Me5KfTl23N67XX098FC-r2-246:-MedVuVF01tee0jRCQwg:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f5f0331dbea1c98b4e88d3ef3a56d4c6e1e191f6/pkg/hive/hive.go#L246)):*

Yes I just realized that, had come to comment on this... This is why the integration tests are also failing I believe.

I'll use an errgroup as I need it to obey the context as well.
</blockquote></details>

Done.

aloknerurkar:
___
*[pkg/hive/hive.go, line 244 at r3](https://reviewable.io/reviews/ethersphere/bee/2319#-MedQ6dm1BBUmd84FgGL:-MedW9XS3sOTnZZW0KzX:b-wii2sy) ([raw file](https://github.com/ethersphere/bee/blob/20fe011a728bc9ed71b33df931d0c68bb5b5873d/pkg/hive/hive.go#L244)):*

since you are spawning a new goroutine on every peer which is received, this can increase stress on GC at some point. consider having a pool of workers spawned in advance similarly to kademlia. They can just directly read off the newPeers channel and do the check
</blockquote></details>

So we need these routines to spawn on-demand. We will only do the check when we get a message from hive. The rest of the time we would waste memory having these running. I was checking pprof and I see only 2 routines that are running when we dont get the broadcast. I think this is better.

Usually, the addPeers would be followed by kamelia using its worker pool, which has been reduced significantly. The semaphore caps the checking routines to 15 + 1 for the routine which will spawn the workers. So I dont think this would cause any more stress to the system.

Wdyt?

aloknerurkar:
___
*[pkg/hive/hive.go, line 260 at r3](https://reviewable.io/reviews/ethersphere/bee/2319#-MedQX8rALbuNH_wDret:-MedVKGQ3MU3VukrY1QR:bbrafjw) ([raw file](https://github.com/ethersphere/bee/blob/20fe011a728bc9ed71b33df931d0c68bb5b5873d/pkg/hive/hive.go#L260)):*

not sure why the weighted semaphore is needed as the resource and the load are linear. what are the benefits over a regular buffered channel?
</blockquote></details>

The Weighted semaphore provides context-based semantics. So if we have a shutdown or close for some reason, this should error out. Did not want to reimplement this as we already use `x/sync`

aloknerurkar (commented):
> Dismissed @janos from a discussion.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2319)* status: all files reviewed, 3 unresolved discussions (waiting on @acud, @aloknerurkar, and @istae)

pkg/p2p/streamtest/streamtest.go
```diff
@@ -153,6 +161,13 @@ func (r *Recorder) NewStream(ctx context.Context, addr swarm.Address, h p2p.Head
 	return streamOut, nil
 }
 
+func (r *Recorder) Ping(ctx context.Context, addr ma.Multiaddr) (rtt time.Duration, err error) {
```
janos, line 164: Given that the Recorder has a very good test coverage, a simple unit test for the new Ping method would be nice to have, even if it just calls the injected function. That would make the test even easier.

janos (approved):
Very nice PR. Thanks @aloknerurkar!

acud (dismissed):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2319)* status: 11 of 12 files reviewed, 8 unresolved discussions (waiting on @acud, @aloknerurkar, and @istae)

___
*[pkg/hive/hive.go, line 260 at r3](https://reviewable.io/reviews/ethersphere/bee/2319#-MedQX8rALbuNH_wDret:-Mee6K4n-6aU-J95W0L8:bj4h49b) ([raw file](https://github.com/ethersphere/bee/blob/20fe011a728bc9ed71b33df931d0c68bb5b5873d/pkg/hive/hive.go#L260)):*

The Weighted semaphore provides context-based semantics. So if we have a shutdown or close for some reason, this should error out. Did not want to reimplement this as we already use `x/sync`
</blockquote></details>

shutdown assumes that the libp2p will close before this component (it won't - look in `node.go#Shutdown`). So the context won't cancel because of the shutdown. Since they all share the same underlying context, this isn't really needed because it is just as good as the following, no?
```go
select {
case sem<-struct{}{}:
case <-ctx.Done():
case <s.quit:
}
```
___
*[pkg/hive/hive.go, line 208 at r4](https://reviewable.io/reviews/ethersphere/bee/2319#-Mee8GyM6FMmuXCbE5_u:-Mee8GyM6FMmuXCbE5_v:ba8kbub) ([raw file](https://github.com/ethersphere/bee/blob/8ba56ec892a9f2e8294e0e9fceeaf0f20b6b2eac/pkg/hive/hive.go#L208)):*
> ```Go
> 	select {
> 	case s.peersChan <- peersReq:
> 	case <-ctx.Done():
> ```

this context handling is a bit problematic. if the peer went away then you basically lose all the peers it sent you and that are still waiting in the validation pipeline
___
*[pkg/hive/hive.go, line 210 at r4](https://reviewable.io/reviews/ethersphere/bee/2319#-Mee8R3z3ZArVPwa5o9C:-Mee8R3z3ZArVPwa5o9D:bege0g6) ([raw file](https://github.com/ethersphere/bee/blob/8ba56ec892a9f2e8294e0e9fceeaf0f20b6b2eac/pkg/hive/hive.go#L210)):*
> ```Go
> 	case <-ctx.Done():
> 		return fmt.Errorf("failed to process peers, reason: %w", ctx.Err())
> 	}
> ```

need to add here:
```go
case <-s.quit:
```
___
*[pkg/p2p/libp2p/protocols_test.go, line 394 at r4](https://reviewable.io/reviews/ethersphere/bee/2319#-Mee8hNwC6C7ActCQlwY:-Mee8hNwC6C7ActCQlwZ:b-pc7kko) ([raw file](https://github.com/ethersphere/bee/blob/8ba56ec892a9f2e8294e0e9fceeaf0f20b6b2eac/pkg/p2p/libp2p/protocols_test.go#L394)):*
> ```Go
> 	addr := serviceUnderlayAddress(t, s1)
> 
> 	fmt.Printf("%v\n", addr)
> ```

can be removed

aloknerurkar:
___
*[pkg/p2p/p2p.go, line 74 at r3](https://reviewable.io/reviews/ethersphere/bee/2319#-Me5KfTl23N67XX098FH-r3-74:-MeeVAZ_70tOYIyt98qo:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/20fe011a728bc9ed71b33df931d0c68bb5b5873d/pkg/p2p/p2p.go#L74)):*

PinPong -> PingPong
</blockquote></details>

Done.

aloknerurkar:
___
*[pkg/p2p/libp2p/protocols_test.go, line 394 at r4](https://reviewable.io/reviews/ethersphere/bee/2319#-Mee8hNwC6C7ActCQlwY:-MeeVF4t6o22Rn9j2_vT:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/8ba56ec892a9f2e8294e0e9fceeaf0f20b6b2eac/pkg/p2p/libp2p/protocols_test.go#L394)):*

can be removed
</blockquote></details>

Done.

aloknerurkar:
___
*[pkg/p2p/streamtest/streamtest.go, line 164 at r3](https://reviewable.io/reviews/ethersphere/bee/2319#-Me5KfTl23N67XX098FI-r3-164:-MeeVHjv3U1bhDqI4H8k:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/20fe011a728bc9ed71b33df931d0c68bb5b5873d/pkg/p2p/streamtest/streamtest.go#L164)):*

Given that the Recorder has a very good test coverage, a simple unit test for the new Ping method would be nice to have, even if it just calls the injected function. That would make the test even easier.
</blockquote></details>

Done.

aloknerurkar:
___
*[pkg/hive/hive.go, line 208 at r4](https://reviewable.io/reviews/ethersphere/bee/2319#-Mee8GyM6FMmuXCbE5_u:-MeeWJMq3cQxv8MfPCS7:bqrv1q8) ([raw file](https://github.com/ethersphere/bee/blob/8ba56ec892a9f2e8294e0e9fceeaf0f20b6b2eac/pkg/hive/hive.go#L208)):*

this context handling is a bit problematic. if the peer went away then you basically lose all the peers it sent you and that are still waiting in the validation pipeline
</blockquote></details>

Ignoring the ctx.Done() signal and using s.quit instead.

aloknerurkar:
___
*[pkg/hive/hive.go, line 260 at r3](https://reviewable.io/reviews/ethersphere/bee/2319#-MedQX8rALbuNH_wDret:-MeeWVDc5lmFsoOHv2Br:b8g6ijw) ([raw file](https://github.com/ethersphere/bee/blob/20fe011a728bc9ed71b33df931d0c68bb5b5873d/pkg/hive/hive.go#L260)):*

shutdown assumes that the libp2p will close before this component (it won't - look in `node.go#Shutdown`). So the context won't cancel because of the shutdown. Since they all share the same underlying context, this isn't really needed because it is just as good as the following, no?
```go
select {
case sem<-struct{}{}:
case <-ctx.Done():
case <s.quit:
}
```
</blockquote></details>

As discussed, using the lib for better readability and in-built context handling.

aloknerurkar:
___
*[pkg/hive/hive.go, line 282 at r3](https://reviewable.io/reviews/ethersphere/bee/2319#-MedREaw1HF609g_Gk4D:-MeeXCo3CF9t7lk466c3:b-n4ljt2) ([raw file](https://github.com/ethersphere/bee/blob/20fe011a728bc9ed71b33df931d0c68bb5b5873d/pkg/hive/hive.go#L282)):*

nit, it might be beneficial to check which types of errors can be returned here, since different errors could signify different assumptions about reachability
</blockquote></details>

Adding a TODO comment for this. Will look into this more later in a separate change.

aloknerurkar:
___
*[pkg/hive/hive.go, line 210 at r4](https://reviewable.io/reviews/ethersphere/bee/2319#-Mee8R3z3ZArVPwa5o9C:-MeeWRGa69sUlI3KiJ-k:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/8ba56ec892a9f2e8294e0e9fceeaf0f20b6b2eac/pkg/hive/hive.go#L210)):*

need to add here:
```go
case <-s.quit:
```
</blockquote></details>

Done.

aloknerurkar (commented):
> Dismissed @acud and @janos from 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2319)* status: 11 of 12 files reviewed, 1 unresolved discussion (waiting on @acud and @istae)

aloknerurkar:
___
*[pkg/hive/hive.go, line 244 at r3](https://reviewable.io/reviews/ethersphere/bee/2319#-MedQ6dm1BBUmd84FgGL:-MeeXwHS2D1JbRnT-QTO:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/20fe011a728bc9ed71b33df931d0c68bb5b5873d/pkg/hive/hive.go#L244)):*

So we need these routines to spawn on-demand. We will only do the check when we get a message from hive. The rest of the time we would waste memory having these running. I was checking pprof and I see only 2 routines that are running when we dont get the broadcast. I think this is better.

Usually, the addPeers would be followed by kamelia using its worker pool, which has been reduced significantly. The semaphore caps the checking routines to 15 + 1 for the routine which will spawn the workers. So I dont think this would cause any more stress to the system.

Wdyt?
</blockquote></details>

Done.

aloknerurkar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2319)* status: 8 of 13 files reviewed, all discussions resolved (waiting on @acud and @istae)

pkg/hive/hive.go
```diff
@@ -218,3 +222,91 @@ func (s *Service) disconnect(peer p2p.Peer) error {
...
+		}
+
+		wg.Add(1)
+		go func(idx int) {
+			defer func() {
+				s.sem.Release(1)
+				wg.Done()
+			}()
+
+			newPeer := peers.Peers[idx]
```
acud, line 272: just a minor suggestion - since you are already iterating over `peers.Peers`, you could already:
```go
for _, p := range peers.Peers {
   go func(p bzz.Address){ ... } (p)
}
```

pkg/p2p/libp2p/libp2p.go
```diff
@@ -816,3 +832,23 @@ func (s *Service) Ready() {
 func (s *Service) Halt() {
 	close(s.halt)
 }
+
+func (s *Service) Ping(ctx context.Context, addr ma.Multiaddr) (rtt time.Duration, err error) {
+	ctx, cancel := context.WithCancel(ctx)
```
acud, line 837: not sure if this extra context is needed, since there is actually no cases where it is being cancelled. on shutdown, the hive will cancel the context which is propagated to this call, so probably this is not needed

acud (approved):
Nice! few more nitpicks :)

aloknerurkar (commented):
> Dismissed @acud from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2319)* status: 6 of 13 files reviewed, all discussions resolved (waiting on @acud and @istae)

___
*[pkg/hive/hive.go, line 272 at r5](https://reviewable.io/reviews/ethersphere/bee/2319#-Me5KfTl23N67XX098FC-r5-272:-MehxpIK2-GikHc8G_kz:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/1d14053b4876c9231094a180a3afe655acff5214/pkg/hive/hive.go#L272)):*

just a minor suggestion - since you are already iterating over `peers.Peers`, you could already:
```go
for _, p := range peers.Peers {
   go func(p bzz.Address){ ... } (p)
}
```
</blockquote></details>

Done.
___
*[pkg/p2p/libp2p/libp2p.go, line 837 at r5](https://reviewable.io/reviews/ethersphere/bee/2319#-Me5KfTl23N67XX098FF-r5-837:-MehxqUu6Iz5uQ3JVwHZ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/1d14053b4876c9231094a180a3afe655acff5214/pkg/p2p/libp2p/libp2p.go#L837)):*

not sure if this extra context is needed, since there is actually no cases where it is being cancelled. on shutdown, the hive will cancel the context which is propagated to this call, so probably this is not needed
</blockquote></details>

Done.

PR #2326, opened 2021-07-12
fix: streamtest

istae:
@janos @acud As I understand, the behavior of streamtest does not fully mock libp2p. `stream.` `Close` and `FullClose` functions for example close the stream in both directions which is not accurate. This has caused us to insert artificial `time.Sleep()`s in the codebase, like the hive protocol, just to make the unit tests pass, which is far from ideal. 

In this draft PR, direction of streams can be closed separately with `ReadClose()` and `WriteClose()`, let me know how you like this approach, and how we can amend some of the tests that are now failing, (mostly streamtest_test.go).

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2326)

### DIFF
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -121,8 +121,6 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
 		if err != nil {
 			_ = stream.Reset()
 		} else {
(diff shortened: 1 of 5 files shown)


janos (commented):
I think that the splitting close with this approach is a good idea. Thanks. 👍

acud:
should we revive this PR? or close it?

PR #2327, opened 2021-07-12
fix(pushsync, retrieval): avoid deadlocks on context done

janos:
This PR addresses a deadlock issue discovered with @acud and @aloknerurkar on production cluster.

It handles the context cancelation in the minimal way to solve blocking on pushToClosest:

```
goroutine profile: total 31463
16690 @ 0x43a885 0x405aea 0x405895 0xe06266 0x470b81
#    0xe06265    github.com/ethersphere/bee/pkg/pushsync.(*PushSync).pushToClosest.func2+0x265    github.com/ethersphere/bee/pkg/pushsync/pushsync.go:345

2992 @ 0x43a885 0x44a9ef 0x11f0025 0x470b81
#    0x11f0024    github.com/ethersphere/bee/pkg/pullsync.(*Syncer).handler.func2+0x104    github.com/ethersphere/bee/pkg/pullsync/pullsync.go:283

...
```

Additionally, the similar fix is done in the retrieval protocol preemptive.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2327)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -381,7 +381,10 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, retryAllo
 					ps.skipList.Add(peer, ch.Address(), skipPeerExpiration)
 				}
 
-				resultC <- &pushResult{err: err, attempted: attempted}
+				select {
+				case resultC <- &pushResult{err: err, attempted: attempted}:
+				case <-ctx.Done():
+				}
 				return
 			}
 			select {
--- a/pkg/retrieval/retrieval.go
+++ b/pkg/retrieval/retrieval.go
@@ -154,16 +154,22 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin 
 					defer cancel()
 
 					chunk, peer, requested, err := s.retrieveChunk(ctx, addr, sp, origin)
-					resultC <- retrievalResult{
+					select {
+					case resultC <- retrievalResult{
 						chunk:     chunk,
 						peer:      peer,
 						err:       err,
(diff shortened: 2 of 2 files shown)


mrekucci (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2327#-:-MePpEliA1tT-XAzoZQX:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud, @aloknerurkar, and @metacertain)

aloknerurkar (approved):

acud (approved):
LGTM @janos.
I would just add: it seems that not in all places a context cancellation actually takes place. It is explicitly done in the `handler` in pushsync, but for example not in the `PushChunkToClosest` method. So in fact it could happen that some dangling goroutines might still deadlock because no one is draining the channel. It might be good to enclose `	r, err := ps.pushToClosest(ctx, ch, true, swarm.ZeroAddress)` with a `context.WithCancel` so that the context gets cancelled anyway when the request returns.

These same problems exist equally in the retrieval protocol so it might be worth to add some explicit context cancellation there too.

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2327#-:-MePqRFp5NkcrWzyY3SP:ba5m0xe)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aloknerurkar and @metacertain)

janos:
@acud context is canceled when the handler returns in removeStream call on defer at https://github.com/ethersphere/bee/blob/master/pkg/p2p/libp2p/libp2p.go#L453 for every protocol handler call. It provides context cancelation without the need to close it in protocol implementations. I think that this should be good enough.

Issue #2343, opened 2021-07-16
New "hive: multi address underlay ... not reachable" badly formatted log


ldeffenb:
#### Summary
PR #2319 introduced peer reachability testing.  When a peer fails reachability, hive.go logs an error including the error returned from s.stream.ping.  This error, however, appears to be mal-formed as you can see in the following example.

>time="2021-07-16T13:42:23-04:00" level=warning msg="hive: multi address underlay /ip4/136.243.31.122/tcp/1640/p2p/16Uiu2HAmKJ6j8oURfABNNTYZYdkZbDFCcKxggP3TagHLWTLuaaoR not reachable err: %!w(*swarm.DialError=&{\x00%\b\x02\x12!\x03b\xa7\xd2S\xb9Ö'\x12\xbb\xf2\x89<T\xb6\x06\xe9\xa5\x15n'8\xd6\xdb\xc4f\x06\x12w\x9d\x05\xe7p [{0xc00e977980 0xc00310c3c0}] <nil> 0})"

#### Steps to reproduce
Just start a node with an existing address book containing a bunch of non-reachable peers and you'll have plenty of examples.

#### Expected behavior
I would expect the log to be more readable.  And possibly not a "Warning" level, but something less because this is an all-to-common situation for peers in the swarm.

#### Actual behavior
See quoted example log message above.

PR #2349, opened 2021-07-20
fix: flaky ping test and log correction

aloknerurkar:
- Adding a new HostFactory option to libp2p to enable test host creation for unit tests.
- Ping test now works
-  Correcting log msg #2343

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2349)

### DIFF
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -93,6 +93,7 @@ type Options struct {
 	LightNodeLimit int
 	WelcomeMessage string
 	Transaction    []byte
+	HostFactory    func(context.Context, ...libp2p.Option) (host.Host, error)
 }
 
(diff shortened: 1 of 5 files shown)


pkg/p2p/libp2p/libp2p.go
```diff
@@ -93,6 +93,7 @@ type Options struct {
 	LightNodeLimit int
 	WelcomeMessage string
 	Transaction    []byte
+	HostFactory    func(context.Context, ...libp2p.Option) (host.Host, error)
```
janos, line 96: Somehow, this field does not seems useful for the exposed package api, only for the tests.

Would you consider having an unexported field in `hostFactory func(context.Context, ...libp2p.Option) (host.Host, error)` instead to be set only by a new helper function defined in export_test.go. This is just a suggestion, not a hard requirement for the PR approval.

janos (approved):
Nice, I have only one suggestion, only if you think that it would improve exposed package api.

acud (approved):
Nice! thanks!

janos:
Thank you @aloknerurkar for additional change. That was exactly what I meant. 🙏

PR #2351, opened 2021-07-21
feat: announce to lightnodes

acud:
Right now we announce nodes to light nodes only when they connect. This does not allow them to know about new nodes as they join the network. This also causes the integration tests to fail, as light nodes may connect early to the bootnode, never knowing anything about any other node in the network, leaving them only with a connection to the bootnode, which results in the tests to flake much more, since they overdraft with the bootnode very quickly, which then eventually results in errors in retrieval.

This PR rectifies this by notifying light nodes about new full nodes as they join the network.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2351)


notanatol (approved):

janos (approved):
👍

PR #2369, opened 2021-07-27
feat: dont push all docker image tags to the dockerhub for rc

vandot:
When tagging RC push only `rc` and `latest` tag to the dockerhub
@acud I think this makes sense, only when releasing stable version set all other tags (`major`, `major.minor`, `major.minor.patch` and `stable`) for RC only `major.minor.patch-rc` and `latest` (rc tag will not be set for stable build)

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2369)


aleksandarknezevic (approved):

acud (approved):
nice! thanks

PR #2371, opened 2021-07-28
refactor: rename postage stamp response exists field to expired

mrekucci:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2371)

### DIFF
--- a/openapi/SwarmCommon.yaml
+++ b/openapi/SwarmCommon.yaml
@@ -333,7 +333,7 @@ components:
           type: integer
         immutableFlag:
           type: boolean
-        exists:
+        expired:
           type: boolean
         batchTTL:
           description: The time (in seconds) remaining until the batch expires; -1 signals that the batch never expires; 0 signals that the batch has already expired.
--- a/pkg/debugapi/postage.go
+++ b/pkg/debugapi/postage.go
@@ -101,7 +101,7 @@ type postageStampResponse struct {
 	BucketDepth   uint8          `json:"bucketDepth"`
 	BlockNumber   uint64         `json:"blockNumber"`
 	ImmutableFlag bool           `json:"immutableFlag"`
-	Exists        bool           `json:"exists"`
+	Expired       bool           `json:"expired"`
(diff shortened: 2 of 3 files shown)


acud:
The problem with this change is that there could be some time in which this will be a bit misleading.
When a batch is bought, the stamp issuer will be created, however until the listener picks up the batch from the chain, the postage package will not find the stamp, leading to an `Expired: true` here. So it might be a bit misleading. Not sure whether we should do the change as a result

AuHau:
@acud hmm but during that time `usable` should be `false` right? So it could be indicating that if `usable = false` then `expired` is not really applicable?

agazso:
> The problem with this change is that there could be some time in which this will be a bit misleading.
> When a batch is bought, the stamp issuer will be created, however until the listener picks up the batch from the chain, the postage package will not find the stamp, leading to an `Expired: true` here. So it might be a bit misleading. Not sure whether we should do the change as a result

To me this suggests that the logic for the `expired` property should include a `&& usable` so that it's not true when it's created. Otherwise this logic has to be done on the client side, and it will lead to very cryptic logic of `expired = !exists && usable` for every user of the postage stamps.

acud:
@AuHau this is a debug endpoint and this specific field is needed to know whether a batch that we bought got evicted earlier than expected. since this is a debug endpoint i wouldn't try to cast everything into UX considerations and however higher-order presentation layers choose to interpret this field (if at all) is not really in scope for the endpoint to handle. on a later thought i might say it is just better to ignore this field and that we use it for internal debugging purposes only

acud:
Using the `usable` field should suffice here, and we (the bee dev team) should take care that as long as usable is true, the `exists` shoud always be true too

notanatol (approved):

aloknerurkar (approved):

AuHau:
Could we get this in before the release? Otherwise it will be breaking change later on...

mrekucci:
After internal discussion, we came to the conclusion that the `expired` doesn't reflect the property of this field and that `Usable` field should be used for that case #2390

PR #2375, opened 2021-07-29
fix: postage stamp usable flag

### LINKED ISSUE #2373: Querying not usable Stamp returns `400`
#### Summary
When querying stamp by ID that was newly created and is still not usable the HTTP request fails with 400 instead of returning the Stamp object with `usable: false`.

This happens because there is a check in [`GetStampIssuer`](https://github.com/ethersphere/bee/blob/5aa39499e52563071cee1e559a533841d9c20768/pkg/postage/service.go#L132) for usability which fails because of it.

#### Steps to reproduce

 1. Create stamp
 2. Query the stamp by ID before it becomes usable

#### Expected behavior

Returned object with `usable: false`

#### Actual behavior

HTTP 400 with
```
time="2021-07-29T07:14:50Z" level=debug msg="get stamp issuer: get issuer: not usable"
time="2021-07-29T07:14:50Z" level=error msg="get stamp issuer: get issuer"
time="2021-07-29T07:14:50Z" level=info msg="debug api access" duration=0.0004778 ip=172.18.0.1 method=GET proto=HTTP/1.1 size=44 status=400 uri=/stamps/b75f5afe387b19057aed57f68f3704a0920b0b0bfab856bc905e6f76e327d61d user-agent=curl/7.64.1
```

mrekucci:
Fixes #2373

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2375)

### DIFF
--- a/pkg/debugapi/postage.go
+++ b/pkg/debugapi/postage.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/ethersphere/bee/pkg/bigint"
 	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/postage"
 	"github.com/ethersphere/bee/pkg/postage/postagecontract"
 	"github.com/ethersphere/bee/pkg/sctx"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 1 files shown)


aloknerurkar (approved):

acud:
not sure if this actually solves the problem... the thing is that the batch estimation also fails because it isn't able to retrieve the batch from the store...

mrekucci:
> not sure if this actually solves the problem... the thing is that the batch estimation also fails because it isn't able to retrieve the batch from the store...

This is solved here #2376

acud (approved):

PR #2379, opened 2021-07-29
feat: limit concurrent access to some parts of the api resources

mrekucci:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2379)

### DIFF
--- a/pkg/debugapi/chequebook.go
+++ b/pkg/debugapi/chequebook.go
@@ -235,6 +235,14 @@ func (s *Service) swapCashoutHandler(w http.ResponseWriter, r *http.Request) {
 		ctx = sctx.SetGasLimit(ctx, l)
 	}
 
+	if !s.cashOutChequeSem.TryAcquire(1) {
+		s.logger.Debug("debug api: simultaneous on-chain operations not supported")
+		s.logger.Error("debug api: simultaneous on-chain operations not supported")
+		jsonhttp.TooManyRequests(w, "simultaneous on-chain operations not supported")
+		return
+	}
+	defer s.cashOutChequeSem.Release(1)
+
 	txHash, err := s.swap.CashCheque(ctx, peer)
 	if err != nil {
 		s.logger.Debugf("debug api: cashout peer: cannot cash %s: %v", addr, err)
--- a/pkg/debugapi/postage.go
+++ b/pkg/debugapi/postage.go
@@ -66,6 +66,14 @@ func (s *Service) postageCreateHandler(w http.ResponseWriter, r *http.Request) {
 		immutable, _ = strconv.ParseBool(val[0])
 	}
 
+	if !s.postageCreateSem.TryAcquire(1) {
+		s.logger.Debug("create batch: simultaneous on-chain operations not supported")
+		s.logger.Error("create batch: simultaneous on-chain operations not supported")
+		jsonhttp.TooManyRequests(w, "simultaneous on-chain operations not supported")
+		return
+	}
+	defer s.postageCreateSem.Release(1)
+
 	batchID, err := s.postageContract.CreateBatch(ctx, amount, uint8(depth), immutable, label)
 	if err != nil {
 		if errors.Is(err, postagecontract.ErrInsufficientFunds) {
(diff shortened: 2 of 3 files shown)


pkg/debugapi/postage.go
```diff
@@ -66,6 +66,14 @@ func (s *Service) postageCreateHandler(w http.ResponseWriter, r *http.Request) {
 		immutable, _ = strconv.ParseBool(val[0])
 	}
 
+	if !s.postageCreateSem.TryAcquire(1) {
+		s.logger.Debug("create batch: resource is busy")
```
acud, line 70: I would add an Error here too, for the user to see even if running not in debug log level

pkg/debugapi/postage.go
```diff
@@ -66,6 +66,14 @@ func (s *Service) postageCreateHandler(w http.ResponseWriter, r *http.Request) {
...
+		jsonhttp.InternalServerError(w, "resource is busy")
+		return
+	} else {
```
acud, line 73: else is not needed, since you `return` in the if block

pkg/debugapi/chequebook.go
```diff
@@ -235,6 +235,14 @@ func (s *Service) swapCashoutHandler(w http.ResponseWriter, r *http.Request) {
 		ctx = sctx.SetGasLimit(ctx, l)
 	}
 
+	if !s.cashOutChequeSem.TryAcquire(1) {
+		s.logger.Debug("debug api: cashout resource is busy")
+		jsonhttp.InternalServerError(w, "cashout resource is busy")
```
acud, line 240: for the message logged in the terminal I would add something like "simultaneous on-chain operations not supported" . for the error code I would consider to return either `HTTP 503 Service Unavailable` or `HTTP 429 Too many requests`

acud (changes requested):

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2379)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @acud)

___
*[pkg/debugapi/chequebook.go, line 240 at r1](https://reviewable.io/reviews/ethersphere/bee/2379#-MfmjqBVvybdm9nvsvHv-r1-240:-Mfmmnac5k32g5-nAXjZ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/a5f421b583333775af4b0e9f334ec6d7234df878/pkg/debugapi/chequebook.go#L240)):*

for the message logged in the terminal I would add something like "simultaneous on-chain operations not supported" . for the error code I would consider to return either `HTTP 503 Service Unavailable` or `HTTP 429 Too many requests`
</blockquote></details>

Done.
___
*[pkg/debugapi/postage.go, line 70 at r1](https://reviewable.io/reviews/ethersphere/bee/2379#-MfmjqBVvybdm9nvsvHx-r1-70:-Mfmmo_t31dcCzq6AxJr:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/a5f421b583333775af4b0e9f334ec6d7234df878/pkg/debugapi/postage.go#L70)):*

I would add an Error here too, for the user to see even if running not in debug log level
</blockquote></details>

Done.
___
*[pkg/debugapi/postage.go, line 73 at r1](https://reviewable.io/reviews/ethersphere/bee/2379#-MfmjqBVvybdm9nvsvHx-r1-73:-MfmnECNBig75YWXUmvK:bl09kba) ([raw file](https://github.com/ethersphere/bee/blob/a5f421b583333775af4b0e9f334ec6d7234df878/pkg/debugapi/postage.go#L73)):*

else is not needed, since you `return` in the if block
</blockquote></details>

It is needed since `TryAcquire` returns `true` on success so you need to release the acquired weight after you finish. Changed the order for more clarity.

acud (approved):

pkg/debugapi/postage.go
mrekucci: Done.

aloknerurkar (approved):

AuHau:
This should be also documented on OpenAPI level!

mrekucci:
> This should be also documented on OpenAPI level!

Thanks for noticing it! Fixed in #2387

PR #2380, opened 2021-07-29
fix(mantaray): mantaray lookup issue

aloknerurkar:
The fix added earlier to fix the nodeType information on root nodes actually overrides the values in the case of other nodes, resulting in incorrect nodeType information on Lookup. Due to this, we will not process the node returned as a valid path and hence return `404` although we have the reference.

The correct fix for the above issue was actually sitting in the original manifest repository as an open PR. This PR was approved but never merged. Pulling in the changes here.

https://github.com/ethersphere/manifest/pull/20/files

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2380)

### DIFF
--- a/pkg/manifest/mantaray/marshal.go
+++ b/pkg/manifest/mantaray/marshal.go
@@ -270,13 +270,6 @@ func (n *Node) UnmarshalBinary(data []byte) error {
 
 		n.entry = append([]byte{}, data[nodeHeaderSize:nodeHeaderSize+refBytesSize]...)
 		offset := nodeHeaderSize + refBytesSize // skip entry
-		// Currently we don't persist the root nodeType when we marshal the manifest, as a result
-		// the root nodeType information is lost on Unmarshal. This causes issues when we want to
-		// perform a path 'Walk' on the root. If there is more than 1 fork, the root node type
-		// is an edge, so we will deduce this information from index byte array
-		if !bytes.Equal(data[offset:offset+32], make([]byte, 32)) {
-			n.nodeType = nodeTypeEdge
-		}
 		n.forks = make(map[byte]*fork)
 		bb := &bitsForBytes{}
 		bb.fromBytes(data[offset:])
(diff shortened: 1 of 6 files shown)


agazso:
I tested it and it works correctly with the files that failed before.

zelig (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2380)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud)

pkg/manifest/mantaray/marshal.go
```diff
@@ -270,13 +270,6 @@ func (n *Node) UnmarshalBinary(data []byte) error {
 
 		n.entry = append([]byte{}, data[nodeHeaderSize:nodeHeaderSize+refBytesSize]...)
 		offset := nodeHeaderSize + refBytesSize // skip entry
-		// Currently we don't persist the root nodeType when we marshal the manifest, as a result
-		// the root nodeType information is lost on Unmarshal. This causes issues when we want to
-		// perform a path 'Walk' on the root. If there is more than 1 fork, the root node type
-		// is an edge, so we will deduce this information from index byte array
-		if !bytes.Equal(data[offset:offset+32], make([]byte, 32)) {
-			n.nodeType = nodeTypeEdge
-		}
```
nugaon, line 279: IMO this edge case handling could remain here, just instead of overwriting the `n.nodeType`, the `makeEdgeType` method of `n` should be called, so `n.nodeType = nodeTypeEdge` -> `n.makeEdge()`

zelig: true, plus actually an all zeros check does not new allocation each time, just compare to `swarm.ZeroAddress.Bytes()` or do it with a loop

aloknerurkar: Agreed. I have actually added tests now to verify the behavior. Most of the bugs we saw were after serialization and deserialization and the tests don't test this behavior.

For all the relevant tests I have added a `with load store` variant which will do the serialization-deserialization and re-run the tests.

@zelig We don't have the notion of `swarm.Address` in the mantaray pkg. Instead, I have created a new constant that is local to this pkg.

acud (approved):
nice! thanks!

PR #2393, opened 2021-08-03
feat: only check once per transaction

ralph-pichler:
don't do multiple receipt lookups for the same tx. that way we cannot get inconsistent results.
also increase cancellation depth as there might be lagging nodes behind some balancers.

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2393)


notanatol (approved):

acud (approved):

ralph-pichler:
superceded by #2496

PR #2400, opened 2021-08-09
feat: protected APIs

notanatol:


agazso:
> * do we leave `/chunks/stream`, `/stewardship/{address}` and `/consumed/*` endpoints unrestricted?

In restricted mode I would not leave any endpoints unrestricted.

pkg/auth/auth.go
```diff
@@ -0,0 +1,162 @@
+package auth
+
+import (
+	sha "crypto/sha512"
```
notanatol, line 4: @acud what is the best library to generate access tokens? should we leave `sha512` or is there a better fit?

acud: not sure if we need a library for that... might be that using `urandom` might be enough. or guid. not sure. would be happy for input from the JS team here

agazso: From the perspective of the API user it does not matter how it is created, this is just an opaque string that needs to be kept in memory so that the requests are authorized. It helps if it is base64 encoded so that it can be easily copy-pasted and used from CLI tools without escaping.

pkg/auth/auth.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if err := applyPolicies(e); err != nil {
+		return nil, err
+	}
+
+	auth := Authenticator{
+		user:     os.Getenv("AUTH_USERNAME"),
```
janos, line 54: The New constructor should receive variables for user and pass as arguments, instead to put a hard requirement on env variable configuration. The configuration should be handled in the cmd package and passed to other components as variables. With this approach, configuration is leaking into components. Another thing is that variables are names pretty generically. This can cause a miss-configuration, I would suppose to prefix them with BEE_, or just to make them as struct fields in the existing cmd configuration and handle them just as any other variables.

pkg/auth/auth.go
```diff
@@ -0,0 +1,162 @@
...
+		pass:     os.Getenv("AUTH_PASSWORD"),
```
janos, line 55: I would suggest to try to avoid or discourage users to configure plain text passwords. Instead to require to pass the passwords in the form of a hash, which they can generate either by some tool (online or offline) or with a dedicated bee command. This is the same as apache or nginx are requiring to have auth passwords configured as hashed with htpasswd or any other tool that is able to hash with supported hashers. I am suggesting golang.org/x/crypto/bcrypt.

pkg/auth/auth.go
```diff
@@ -0,0 +1,162 @@
...
+		apiKeys:  make(apiKeys),
+		enforcer: e,
+	}
+
+	return &auth, nil
+}
+
+func (a *Authenticator) Authorize(u, p string) bool {
+	return a.user == u && a.pass == p
```
janos, line 64: I am suggesting to have the password hashed and checked with, for example, bcrypt.GenerateFromPassword.

pkg/auth/auth.go
```diff
@@ -0,0 +1,162 @@
+package auth
```
janos, line 1: Add copyright header.

janos (commented):
I would just point out that tests are important for this functionality, even though I know that the status is still in progress.

pkg/auth/auth.go
notanatol: fixed

pkg/auth/auth.go
notanatol: swapped for a simpler solution, please review.

pkg/api/api.go
```diff
@@ -240,6 +251,35 @@ func requestPostageBatchId(r *http.Request) ([]byte, error) {
 	return nil, errInvalidPostageBatch
 }
 
+func (s *server) authHandler(w http.ResponseWriter, r *http.Request) {
```
janos, line 254: This handler is returning responses in plain text, while other handlers are returning json-encoded responses. Is this what it intended?

pkg/api/api.go
```diff
@@ -263,6 +303,43 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
 	}
 }
 
+func (s *server) permissionCheckHandler() func(h http.Handler) http.Handler {
```
janos, line 306: This middleware should not return plain text responses as it is part of the json-encoded endpoints. I think that there should be consistency, as clients may assume json encoded response even without checking content-type.

pkg/api/api.go
```diff
@@ -263,6 +303,43 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
...
+				h.ServeHTTP(w, r)
+			})
+		}
+	}
+
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			reqToken := r.Header.Get("Authorization")
+			if !strings.HasPrefix(reqToken, "Bearer ") {
+				http.Error(w, "Missing bearer token", http.StatusForbidden)
```
janos, line 319: Use json encoded response.

pkg/api/api.go
```diff
@@ -263,6 +303,43 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
...
+				return
+			}
+
+			keys := strings.Split(reqToken, "Bearer ")
+
+			if len(keys) != 2 || strings.Trim(keys[1], " ") == "" {
+				http.Error(w, "Missing API Key", http.StatusForbidden)
```
janos, line 326: JSON encoded response should be more aligned with other responses.

pkg/api/api.go
```diff
@@ -263,6 +303,43 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
...
+				return
+			}
+
+			apiKey := keys[1]
+
+			allowed := s.auth.Enforce(apiKey, r.URL.Path, r.Method)
+			if !allowed {
+				w.WriteHeader(http.StatusForbidden)
```
janos, line 334: I think that this response deserves a body, json encoded, explaining that the auth key is not allowed.

pkg/api/api.go
```diff
@@ -240,6 +251,35 @@ func requestPostageBatchId(r *http.Request) ([]byte, error) {
...
+	}
+
+	if err = json.Unmarshal(body, &role); err != nil {
+		http.Error(w, "Unmarshal json body", http.StatusBadRequest)
+		return
+	}
+
+	key := s.auth.AddKey(user, role.Role)
+
+	_, _ = w.Write([]byte(key))
```
janos, line 280: For extensibility, a json object with field "key" would be better.

pkg/debugapi/router.go
```diff
@@ -229,3 +302,40 @@ func (s *Service) setRouter(router http.Handler) {
 
 	s.handler = h
 }
+
+func (s *Service) permissionCheckHandler() func(h http.Handler) http.Handler {
```
janos, line 306: The same comments on the api package applies here.

pkg/node/devnode.go
```diff
@@ -110,6 +113,18 @@ func NewDevBee(logger logging.Logger, o *DevOptions) (b *DevBee, err error) {
 		return nil, fmt.Errorf("eth address: %w", err)
 	}
 
+	var authenticator *auth.Authenticator
+
+	username := os.Getenv("BEE_AUTH_USERNAME")
+	password := os.Getenv("BEE_AUTH_PASSWORD")
```
janos, line 119: The node package does not handle env variables, it explicitly expect input arguments to functions. Currently all configuration is done in the cmd/bee package and I would suggest to make the same with these parameters also. Just adding string flags as with any other configuration parameters will also provide the possibility to configure them with env variables, the same way as with any other parameter.

janos (commented):
Thanks @notanatol for implementing bcrypt. There are a few more comments regarding configuration handling and plain text http responses.

mrekucci (commented):
> Reviewed 2 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400)* status: 2 of 14 files reviewed, 14 unresolved discussions (waiting on @acud, @agazso, @istae, @janos, and @notanatol)

___
*[pkg/api/api.go, line 280 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r3-280:-Mh8h-QEDN_J0DoEeUcc:b9pbpve) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/api/api.go#L280)):*

For extensibility, a json object with field "key" would be better.
</blockquote></details>

I agree, we should use `jsonhttp.Created` for response with a JSON object.
___
*[pkg/api/router.go, line 31 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8iVq28IXVBINaeUTc:-Mh8iVq28IXVBINaeUTd:badxny) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/api/router.go#L31)):*
> ```Go
> 
> 	// handle is a helper closure which simplifies the router setup.
> 	handle := func(path string, handler http.Handler) {
> ```

Since almost every path should be protected I'd recommend implementing the permission check directly into the `handle` closure and at the same time create new closure (which will be used by the `handle` closure) for unprotected endpoints.
___
*[pkg/auth/auth.go, line 25 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8lNk8CNBWiZXA8azx:-Mh8lNkC9_xSlRJ8WeAu:b-p5jdkt) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/auth/auth.go#L25)):*
> ```Go
> 
> type Authenticator struct {
> 	user     string
> ```

If the `user` isn't an object it's usually referred to as `username` and `password` instead of pass.
___
*[pkg/auth/auth.go, line 75 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8mZrP-7pFZky_fuV5:-Mh8mZrP-7pFZky_fuV6:b3xu1g5) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/auth/auth.go#L75)):*
> ```Go
> }
> 
> func isValidHash(password string, hash []byte) bool {
> ```

Since it is used just in the `Authorize` I'd inline this there.
___
*[pkg/auth/auth.go, line 84 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8mvKu4XvIy5kr3diX:-Mh8mvKu4XvIy5kr3diY:b-gk1ol1) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/auth/auth.go#L84)):*
> ```Go
> }
> 
> func (a *Authenticator) AddKey(user, role string) string {
> ```

The `user` isn't used.
___
*[pkg/auth/auth.go, line 93 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8n8Mx1k1jm4xdZ2n9:-Mh8n8Mx1k1jm4xdZ2nA:b-ubvsni) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/auth/auth.go#L93)):*
> ```Go
> 
> 	b := make([]byte, 16)
> 	_, _ = rand.Read(b)
> ```

The error shouldn't be silenced.
___
*[pkg/auth/auth.go, line 114 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8nmdbAGtgYCwOSCuZ:-Mh8nmdbAGtgYCwOSCu_:b-txbpss) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/auth/auth.go#L114)):*
> ```Go
> 	sub := ar.role // the user that wants to access a resource.
> 
> 	if allow, _ := a.enforcer.Enforce(sub, obj, act); !allow {
> ```

The `!allow` variable can be returned directly.
___
*[pkg/debugapi/router.go, line 32 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8kI0m-gEW1RnWRuzQ:-Mh8kI0m-gEW1RnWRuzR:b-8imm33) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/debugapi/router.go#L32)):*
> ```Go
> 	router := mux.NewRouter()
> 	router.NotFoundHandler = http.HandlerFunc(jsonhttp.NotFoundHandler)
> ```

I'd recommend creating the same closure as in the regular API router and implementing the permission check directly into the `handle` closure similar as was recommended in the regular router API.

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400)* status: 0 of 16 files reviewed, 14 unresolved discussions (waiting on @acud, @agazso, @istae, @janos, @mrekucci, and @notanatol)

___
*[pkg/api/api.go, line 254 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r3-254:-MhDAdyH6hvBeMh7zzbS:b-qi6ie1) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/api/api.go#L254)):*

This handler is returning responses in plain text, while other handlers are returning json-encoded responses. Is this what it intended?
</blockquote></details>

No, by mistake, now fixed.
___
*[pkg/api/api.go, line 280 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r3-280:-MhDA_WtCHH140WhVnWx:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/api/api.go#L280)):*

I agree, we should use `jsonhttp.Created` for response with a JSON object.
</blockquote></details>

Done.
___
*[pkg/api/api.go, line 306 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r3-306:-MhDHvj2BZBQ1GV4b2o0:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/api/api.go#L306)):*

This middleware should not return plain text responses as it is part of the json-encoded endpoints. I think that there should be consistency, as clients may assume json encoded response even without checking content-type.
</blockquote></details>

fixed
___
*[pkg/api/api.go, line 319 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r3-319:-MhD9Tsq-p9-yc3tFmje:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/api/api.go#L319)):*

Use json encoded response.
</blockquote></details>

fixed
___
*[pkg/api/api.go, line 326 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r3-326:-MhD9SZn1ltgscTr1TWr:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/api/api.go#L326)):*

JSON encoded response should be more aligned with other responses.
</blockquote></details>

fixed
___
*[pkg/api/api.go, line 334 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r3-334:-MhD9RIMEtt1nD37jkKX:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/api/api.go#L334)):*

I think that this response deserves a body, json encoded, explaining that the auth key is not allowed.
</blockquote></details>

fixed
___
*[pkg/auth/auth.go, line 54 at r1](https://reviewable.io/reviews/ethersphere/bee/2400#-MgjdFCVazrR8Ks1CKzI-r1-54:-MhEPpk5Em5uVO0EnS8x:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/d0b09f3864c6adf21536d448a4ad7be47021e5c5/pkg/auth/auth.go#L54)):*

The New constructor should receive variables for user and pass as arguments, instead to put a hard requirement on env variable configuration. The configuration should be handled in the cmd package and passed to other components as variables. With this approach, configuration is leaking into components. Another thing is that variables are names pretty generically. This can cause a miss-configuration, I would suppose to prefix them with BEE_, or just to make them as struct fields in the existing cmd configuration and handle them just as any other variables.
</blockquote></details>

Done.
___
*[pkg/auth/auth.go, line 55 at r1](https://reviewable.io/reviews/ethersphere/bee/2400#-MgjdFCVazrR8Ks1CKzI-r1-55:-MhDDunl-31MjcUQlEhl:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/d0b09f3864c6adf21536d448a4ad7be47021e5c5/pkg/auth/auth.go#L55)):*

I would suggest to try to avoid or discourage users to configure plain text passwords. Instead to require to pass the passwords in the form of a hash, which they can generate either by some tool (online or offline) or with a dedicated bee command. This is the same as apache or nginx are requiring to have auth passwords configured as hashed with htpasswd or any other tool that is able to hash with supported hashers. I am suggesting golang.org/x/crypto/bcrypt.
</blockquote></details>

fixed
___
*[pkg/auth/auth.go, line 64 at r1](https://reviewable.io/reviews/ethersphere/bee/2400#-MgjdFCVazrR8Ks1CKzI-r1-64:-MhDEdm04jSgRNxk9Mnf:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/d0b09f3864c6adf21536d448a4ad7be47021e5c5/pkg/auth/auth.go#L64)):*

I am suggesting to have the password hashed and checked with, for example, bcrypt.GenerateFromPassword.
</blockquote></details>

fixed
___
*[pkg/auth/auth.go, line 25 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8lNk8CNBWiZXA8azx:-MhDDm6j-4jGgH9Kf4CD:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/auth/auth.go#L25)):*

If the `user` isn't an object it's usually referred to as `username` and `password` instead of pass.
</blockquote></details>

fixed
___
*[pkg/auth/auth.go, line 75 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8mZrP-7pFZky_fuV5:-MhDEcGF23tum7USn8zl:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/auth/auth.go#L75)):*

Since it is used just in the `Authorize` I'd inline this there.
</blockquote></details>

fixed
___
*[pkg/auth/auth.go, line 84 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8mvKu4XvIy5kr3diX:-MhDEke11C2igO8fgin5:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/auth/auth.go#L84)):*

The `user` isn't used.
</blockquote></details>

fixed
___
*[pkg/auth/auth.go, line 93 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8n8Mx1k1jm4xdZ2n9:-MhDFYer5Jr6lWxQ2n6K:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/auth/auth.go#L93)):*

The error shouldn't be silenced.
</blockquote></details>

fixed
___
*[pkg/auth/auth.go, line 114 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8nmdbAGtgYCwOSCuZ:-MhDHT1p39Oyot6Hofxp:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/auth/auth.go#L114)):*

The `!allow` variable can be returned directly.
</blockquote></details>

fixed
___
*[pkg/debugapi/router.go, line 306 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-MgjdFCVazrR8Ks1CKzL-r3-306:-MhEPlZgBCKe7Y-ulMkf:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/debugapi/router.go#L306)):*

The same comments on the api package applies here.
</blockquote></details>

Done.

pkg/api/api.go
```diff
@@ -240,6 +252,46 @@ func requestPostageBatchId(r *http.Request) ([]byte, error) {
 	return nil, errInvalidPostageBatch
 }
 
+type AuthKeyResponse struct {
+	Key string `json:"key"`
+}
+
+type RoleRequest struct {
+	Role string `json:"role"`
+}
```
janos, line 261: These types does not need to be exported, and can be provided through export_test.go for tests.

pkg/api/api.go
```diff
@@ -240,6 +252,46 @@ func requestPostageBatchId(r *http.Request) ([]byte, error) {
...
+	if err = json.Unmarshal(body, &role); err != nil {
+		jsonhttp.BadRequest(w, "Unmarshal json body")
+		return
+	}
+
+	key, err := s.auth.AddKey(role.Role)
+	if err != nil {
+		jsonhttp.InternalServerError(w, err)
+		return
+	}
```
janos, line 288: Usually errors are logged as well.

pkg/api/api.go
```diff
@@ -263,6 +315,48 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
...
+
+			if len(keys) != 2 || strings.Trim(keys[1], " ") == "" {
+				jsonhttp.Forbidden(w, "Missing security token")
+				return
+			}
+
+			apiKey := keys[1]
+
+			allowed, err := s.auth.Enforce(apiKey, r.URL.Path, r.Method)
+			if err != nil {
```
janos, line 345: This error can be logged also.

pkg/auth/mock.go
```diff
@@ -0,0 +1,16 @@
+package auth
```
janos, line 1: Usually, mocks are in a separate package beside the main implementation. Mainly to keep the package api clean from testing code.

pkg/debugapi/router.go
```diff
@@ -229,3 +302,45 @@ func (s *Service) setRouter(router http.Handler) {
...
+			if len(keys) != 2 || strings.Trim(keys[1], " ") == "" {
+				jsonhttp.Forbidden(w, "Missing security token")
+				return
+			}
+
+			apiKey := keys[1]
+
+			allowed, err := s.auth.Enforce(apiKey, r.URL.Path, r.Method)
+			if err != nil {
+				jsonhttp.InternalServerError(w, "Validate security token")
```
janos, line 334: This error can also be logged.

janos (commented):
Thanks @notanatol for addressing my comments. I have a few more, which are mostly opinions that are also part of the existing codebase.

notanatol (commented):
> Reviewed 1 of 8 files at r5.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400)* status: 0 of 19 files reviewed, 18 unresolved discussions (waiting on @acud, @istae, @janos, @mrekucci, and @notanatol)

___
*[pkg/api/api.go, line 261 at r5](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r5-261:-MhIG-044xtZn5kKJoMn:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/6379c3129b36abb4a76e64306aec4892c3c775d9/pkg/api/api.go#L261)):*

These types does not need to be exported, and can be provided through export_test.go for tests.
</blockquote></details>

fixed
___
*[pkg/api/api.go, line 288 at r5](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r5-288:-MhIIxJJ0d27I9_p-u01:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/6379c3129b36abb4a76e64306aec4892c3c775d9/pkg/api/api.go#L288)):*

Usually errors are logged as well.
</blockquote></details>

fixed
___
*[pkg/api/api.go, line 345 at r5](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r5-345:-MhIM2zw94ThEhLQtBlH:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/6379c3129b36abb4a76e64306aec4892c3c775d9/pkg/api/api.go#L345)):*

This error can be logged also.
</blockquote></details>

Done.
___
*[pkg/api/router.go, line 31 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-Mh8iVq28IXVBINaeUTc:-MhIWrKDEZjjNvIzLadI:b-7tfmjp) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/api/router.go#L31)):*

Since almost every path should be protected I'd recommend implementing the permission check directly into the `handle` closure and at the same time create new closure (which will be used by the `handle` closure) for unprotected endpoints.
</blockquote></details>

added
___
*[pkg/debugapi/router.go, line 334 at r5](https://reviewable.io/reviews/ethersphere/bee/2400#-MgjdFCVazrR8Ks1CKzL-r5-334:-MhIL4dHCcW_0iox-JIT:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/6379c3129b36abb4a76e64306aec4892c3c775d9/pkg/debugapi/router.go#L334)):*

This error can also be logged.
</blockquote></details>

fixed
___
*[pkg/auth/mock.go, line 1 at r5](https://reviewable.io/reviews/ethersphere/bee/2400#-MhEPW_4SnxH-Bvc815y-r5-1:-MhILABQ-SauBvxREgBG:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/6379c3129b36abb4a76e64306aec4892c3c775d9/pkg/auth/mock.go#L1)):*

Usually, mocks are in a separate package beside the main implementation. Mainly to keep the package api clean from testing code.
</blockquote></details>

Done.

notanatol (commented):
> Reviewed 3 of 13 files at r1, 1 of 4 files at r2, 3 of 7 files at r4, 2 of 8 files at r5, 9 of 10 files at r6, 1 of 7 files at r7.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400)* status: 13 of 20 files reviewed, 18 unresolved discussions (waiting on @acud, @istae, @janos, and @notanatol)

___
*[pkg/node/devnode.go, line 119 at r3](https://reviewable.io/reviews/ethersphere/bee/2400#-MgjdFCVazrR8Ks1CKzM-r3-119:-MhJI0uiA7Q5BhLtgtpE:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/9d1a32f439f9c4738ce41e1274965def1bfbdc75/pkg/node/devnode.go#L119)):*

The node package does not handle env variables, it explicitly expect input arguments to functions. Currently all configuration is done in the cmd/bee package and I would suggest to make the same with these parameters also. Just adding string flags as with any other configuration parameters will also provide the possibility to configure them with env variables, the same way as with any other parameter.
</blockquote></details>

Done.

notanatol (commented):
should be addressed now, thanks.

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400#-:-MhJKneQ0Vi-DE7ToU6z:b-nvnpab)* status: 13 of 20 files reviewed, 18 unresolved discussions (waiting on @acud, @istae, @janos, and @notanatol)

pkg/api/api.go
```diff
@@ -240,6 +252,54 @@ func requestPostageBatchId(r *http.Request) ([]byte, error) {
...
+
+type roleRequest struct {
+	Role string `json:"role"`
+}
+
+func (s *server) authHandler(w http.ResponseWriter, r *http.Request) {
+	user, pass, ok := r.BasicAuth()
+
+	if !ok || !s.auth.Authorize(user, pass) {
+		s.logger.Debug("unauthorized")
```
janos, line 267: These log messages would be better with more information. Like with a prefix indicating that it is from the api and which handler.

pkg/api/api.go
```diff
@@ -240,6 +252,54 @@ func requestPostageBatchId(r *http.Request) ([]byte, error) {
...
+		s.logger.Error("unauthorized")
+		w.Header().Set("WWW-Authenticate", `Basic realm="Restricted"`)
+		jsonhttp.Unauthorized(w, "Unauthorized")
+		return
+	}
+
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		s.logger.Debugf("read auth request body: %v", err)
+		s.logger.Errorf("read auth request body: %v", err)
```
janos, line 277: Error messages should not reveal internal implementations, where the err passed here may do so. It should be descriptive to the user about the problem and debug log messages should provide more information about the internals for developers.

pkg/auth/mock/auth.go
```diff
@@ -0,0 +1,22 @@
+package mock
```
janos, line 1: Add copyright header.

pkg/auth/auth.go
```diff
@@ -0,0 +1,180 @@
...
+	e, err := casbin.NewEnforcer(m)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := applyPolicies(e); err != nil {
+		return nil, err
+	}
+
+	passwordHash, err := bcrypt.GenerateFromPassword([]byte(password), 14)
```
janos, line 59: I had in mind to avoid keeping in configuration password in plain text by using bcrypt, but this is not doing much for that as it is still required to set password in plain text in config in order to hash it here. The hash should be set in the configuration, instead the plain password. The hash may be generated by a separate command or a tool like htpasswd or any online bcrypt site. I think that the cli command would be nice. It does add a ux complexity, but I think for a justified security reason.

notanatol: should we maybe add it as a separate command?
`/cmd/bee` <- current entry point
`/cmd/passwordgen` <- new command that takes a plain text and returns its hash

or maybe just document so the user could use existing tools?

pkg/api/api.go
notanatol: I'm not sure how to achieve that, is there a relevant example somewhere?

notanatol (commented):
> Reviewed 1 of 10 files at r6, 6 of 7 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400)* status: all files reviewed, 22 unresolved discussions (waiting on @istae, @janos, and @notanatol)

___
*[pkg/api/api.go, line 267 at r7](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r7-267:-MhN1qSiBjNJlCmDEdGt:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/b7d13cd94b58961490565c5bb3725dbd6a9855b0/pkg/api/api.go#L267)):*

These log messages would be better with more information. Like with a prefix indicating that it is from the api and which handler.
</blockquote></details>

Done.
___
*[pkg/auth/mock/auth.go, line 1 at r7](https://reviewable.io/reviews/ethersphere/bee/2400#-MhIZeZDc5_eSGNU47VN-r7-1:-MhN1nN9BoISMkxADTuv:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/b7d13cd94b58961490565c5bb3725dbd6a9855b0/pkg/auth/mock/auth.go#L1)):*

Add copyright header.
</blockquote></details>

Done.

pkg/api/api.go
janos: Log messages in other handlers as for example these ones https://github.com/ethersphere/bee/blob/master/pkg/api/bytes.go#L32 can reveal a pattern. Also the reasoning is explained in coding.md file in the root dir.

pkg/auth/auth.go
janos: My opinion is that a separate command would provide a direct way to obtain the hash, and to have in documentation how a password can be generated. But for this PR, the command is not necessary, only documenting where the hash can be  generated. For example, nginx relied on apache's htpasswd tool for many years and clearly documented how to use it in relation to nginx. So I would not rush with the command, but I think that it provides a nice convenience.

notanatol (commented):
> Reviewed 1 of 2 files at r9.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400)* status: 12 of 20 files reviewed, 22 unresolved discussions (waiting on @istae, @janos, and @notanatol)

___
*[pkg/api/api.go, line 277 at r7](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r7-277:-MhO1aav1OefIUACicA-:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/b7d13cd94b58961490565c5bb3725dbd6a9855b0/pkg/api/api.go#L277)):*

Log messages in other handlers as for example these ones https://github.com/ethersphere/bee/blob/master/pkg/api/bytes.go#L32 can reveal a pattern. Also the reasoning is explained in coding.md file in the root dir.
</blockquote></details>

Done.

janos (approved):
Thank you @notanatol for addressing all of my comments! 👍

mrekucci (commented):
> Reviewed 1 of 7 files at r4, 1 of 8 files at r5, 5 of 10 files at r6, 2 of 7 files at r7, 4 of 8 files at r8, 2 of 2 files at r9, 6 of 6 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400)* status: 21 of 22 files reviewed, 22 unresolved discussions (waiting on @acud, @istae, @janos, and @notanatol)

___
*[.github/workflows/beekeeper.yml, line 50 at r10](https://reviewable.io/reviews/ethersphere/bee/2400#-MhmVLzP9H_D2OjJvja5:-MhmVLzP9H_D2OjJvja6:b-3s8gd6) ([raw file](https://github.com/ethersphere/bee/blob/eb76d6165ebe60434b1e756dfb5ad46375e1a64d/.github/workflows/beekeeper.yml#L50)):*
> ```YAML
>         run: |
>           printf ${{ secrets.CR_PAT }} | docker login ghcr.io -u bee-worker --password-stdin
>           make beekeeper BEEKEEPER_INSTALL_DIR=/usr/local/bin BEEKEEPER_USE_SUDO=true BEEKEEPER_BRANCH=auth-health-check 
> ```

Remove the two trailing spaces at the end.
___
*[pkg/api/router.go, line 47 at r1](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8g-r1-47:-MhmWUTZ58GEEckKguRL:b-ugwiiu) ([raw file](https://github.com/ethersphere/bee/blob/d0b09f3864c6adf21536d448a4ad7be47021e5c5/pkg/api/router.go#L47)):*

this is a bit tricky, since there is no check about the identity of the caller here, any malicious dapp can try to query the API, then figure it is restricted, and then just add itself using this API endpoint as a legitimate API user
</blockquote></details>

Agree, this should be protected by username/password.
___
*[pkg/auth/auth.go, line 36 at r10](https://reviewable.io/reviews/ethersphere/bee/2400#-MhmX4Cj98ecrxpOWv2n:-MhmX4Cj98ecrxpOWv2o:b-day9p4) ([raw file](https://github.com/ethersphere/bee/blob/eb76d6165ebe60434b1e756dfb5ad46375e1a64d/pkg/auth/auth.go#L36)):*
> ```Go
> 	[request_definition]
> 	r = sub, obj, act
> 	
> ```

Remove trailing spaces, also below.

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400)* status: 19 of 22 files reviewed, 22 unresolved discussions (waiting on @acud, @istae, @janos, @mrekucci, and @notanatol)

___
*[pkg/api/router.go, line 47 at r1](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8g-r1-47:-MhmiegS9cl1XTBMkmzb:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/d0b09f3864c6adf21536d448a4ad7be47021e5c5/pkg/api/router.go#L47)):*

Agree, this should be protected by username/password.
</blockquote></details>

Done.
___
*[pkg/auth/auth.go, line 59 at r7](https://reviewable.io/reviews/ethersphere/bee/2400#-MgjdFCVazrR8Ks1CKzI-r7-59:-MhmnPWYBlekZXt_Gmgo:bg41svl) ([raw file](https://github.com/ethersphere/bee/blob/b7d13cd94b58961490565c5bb3725dbd6a9855b0/pkg/auth/auth.go#L59)):*

My opinion is that a separate command would provide a direct way to obtain the hash, and to have in documentation how a password can be generated. But for this PR, the command is not necessary, only documenting where the hash can be  generated. For example, nginx relied on apache's htpasswd tool for many years and clearly documented how to use it in relation to nginx. So I would not rush with the command, but I think that it provides a nice convenience.
</blockquote></details>

I'll make sure to document how to hash the password.

mrekucci (approved):
:lgtm: 

> Reviewed 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400#-:-MhmxOCq2_5CcW_JBlim:b593w57)* status: all files reviewed, 22 unresolved discussions (waiting on @acud, @istae, and @janos)

.github/workflows/beekeeper.yml
```diff
@@ -56,89 +56,17 @@ jobs:
...
-      - name: Test retrieval
-        id: retrieval-1
-        run: beekeeper check --cluster-name local-dns --checks=ci-retrieval
-      - name: Test manifest
-        id: manifest-1
-        run: beekeeper check --cluster-name local-dns --checks=ci-manifest
-      - name: Test content availability
-        id: content-availability
-        run: beekeeper check --cluster-name local-dns --checks=ci-content-availability
+      - name: Test authenticated health
```
janos, line 63: Should we wait for new beekeeper version to be released with the ci-auth check and to get back ?

janos (commented):
All looks good. I just have a question about the removed checks in ci beekeeper configuration.

.github/workflows/beekeeper.yml
notanatol: yes, I will mark it as draft and merge it only after BK changes are in.

acud (approved):
bravisimo... looks very nice :clap: I have a few very small suggestions

> Reviewed 1 of 7 files at r4, 5 of 10 files at r6, 1 of 7 files at r7, 3 of 8 files at r8, 2 of 2 files at r9, 3 of 6 files at r10, 1 of 3 files at r11, 5 of 6 files at r12, 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400#-:-Mho1B4O8z9V6IqRtZsw:b-31m786)* status: all files reviewed, 22 unresolved discussions (waiting on @istae and @janos)

___
*[pkg/api/api.go, line 267 at r7](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r7-267:-Mho-nvbCcq8QKnAU0yb:blhijob) ([raw file](https://github.com/ethersphere/bee/blob/b7d13cd94b58961490565c5bb3725dbd6a9855b0/pkg/api/api.go#L267)):*

Done.
</blockquote></details>

same here and in the rest of the PR :)
___
*[pkg/api/api.go, line 267 at r13](https://reviewable.io/reviews/ethersphere/bee/2400#-Mho-WTrAj6vFsxE-IDp:-Mho-WTrAj6vFsxE-IDq:bozu1r6) ([raw file](https://github.com/ethersphere/bee/blob/89227ffd8cf22e66728c35805f196630c24b404f/pkg/api/api.go#L267)):*
> ```Go
> 
> 	if !ok {
> 		s.logger.Debug("api: auth handler: missing basic auth")
> ```

if the `Debug` does not provide any additional information over `Error`, it can be safely omitted

notanatol (commented):
> Reviewed 3 of 8 files at r8, 1 of 2 files at r9, 3 of 6 files at r10, 1 of 3 files at r11, 5 of 6 files at r12, 1 of 1 files at r13.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400)* status: 19 of 22 files reviewed, 22 unresolved discussions (waiting on @acud, @istae, @janos, @mrekucci, and @notanatol)

___
*[pkg/api/api.go, line 267 at r7](https://reviewable.io/reviews/ethersphere/bee/2400#-MgeAf-90HXtWj0MIx8f-r7-267:-Mhr6oSx6JIxJsRUf85a:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/b7d13cd94b58961490565c5bb3725dbd6a9855b0/pkg/api/api.go#L267)):*

same here and in the rest of the PR :)
</blockquote></details>

Done.

notanatol (commented):
> Reviewed 3 of 8 files at r20, 1 of 2 files at r21, 2 of 2 files at r22, 2 of 2 files at r25, 2 of 3 files at r26, 2 of 2 files at r29, 1 of 1 files at r30, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2400)* status: all files reviewed, 24 unresolved discussions (waiting on @istae, @janos, and @notanatol)

pkg/auth/auth.go
```diff
@@ -0,0 +1,290 @@
...
+
+	apiKey = base64.StdEncoding.EncodeToString(encryptedBytes)
+
+	return apiKey, nil
+}
+
+func (a *Authenticator) Enforce(apiKey, obj, act string) (bool, error) {
+	decoded, err := base64.StdEncoding.DecodeString(apiKey)
+	if err != nil {
+		a.log.Error("decode token", err)
```
janos, line 156: I would avoid both logging and returning errors in this function.

janos (approved):
Nice 👍. I have only one minor comment.

pkg/auth/auth.go
notanatol: I'm going to reduce logging in a future PR but right now we need detailed information when certain cases occur, like token expiration.

PR #2401, opened 2021-08-09
feat(debugapi, postage): topup batch handling

aloknerurkar:
- Add new API endpoint for topup of batches
- Add handling in postagecontract.Interface for TopUp
- Update stampissuer details on topup

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2401)


acud:
@aloknerurkar could you please also update the openapi spec to reflect these changes?

aloknerurkar:
> @aloknerurkar could you please also update the openapi spec to reflect these changes?

Yes. I was hoping @zelig to review once and approve the endpoint path and format.

pkg/postage/service.go
```diff
@@ -103,6 +104,24 @@ func (ps *service) Handle(b *Batch) {
...
+// issuer was not created initially, we will create it here.
+func (ps *service) HandleTopUp(batchID []byte, newValue *big.Int, blockNumber uint64) {
+	ps.lock.Lock()
+	defer ps.lock.Unlock()
+
+	for _, v := range ps.issuers {
+		if bytes.Equal(batchID, v.data.BatchID) {
+			if blockNumber > v.data.BlockNumber {
+				v.data.BatchAmount = newValue
+				v.data.BlockNumber = blockNumber
```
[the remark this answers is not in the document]
acud: this is not necessary. if the topup doesn't propagate on time and the batch expires there's no use of waiting anyway and the contract will refuse to top up an expired batch. a user that chooses to top-up a batch last-minute should be aware of the risk (and it should be documented)

pkg/api/postage.go
```diff
@@ -166,3 +166,55 @@ func (s *server) postageGetStampHandler(w http.ResponseWriter, r *http.Request)
...
+		}
+		ctx = sctx.SetGasPrice(ctx, p)
+	}
+
+	err = s.postageContract.TopUpBatch(ctx, id, amount)
+	if err != nil {
+		if errors.Is(err, postagecontract.ErrInsufficientFunds) {
+			s.logger.Debugf("topup batch: out of funds: %v", err)
+			s.logger.Error("topup batch: out of funds")
+			jsonhttp.BadRequest(w, "out of funds")
```
acud, line 208: consider returning a PaymentRequired http code here

pkg/postage/batchservice/batchservice_test.go
```diff
@@ -211,6 +233,41 @@ func TestBatchServiceTopUp(t *testing.T) {
 		if got.Value.Cmp(want) != 0 {
 			t.Fatalf("topped up amount: got %v, want %v", got.Value, want)
 		}
+		if testBatchListener.topupCount != 1 {
+			t.Fatalf("unexpected batch listener count, exp %d found %d", 1, testBatchListener.topupCount)
+		}
+	})
+
+	t.Run("passes without update", func(t *testing.T) {
```
acud, line 241: not sure what this means?

pkg/postage/postagecontract/contract.go
```diff
@@ -204,6 +242,42 @@ func (c *postageContract) CreateBatch(ctx context.Context, initialBalance *big.I
...
 }
 
+func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topUpAmount *big.Int) error {
+
+	batch, err := c.postageStorer.Get(batchID)
+	if err != nil {
+		return err
+	}
+
+	totalAmount := big.NewInt(0).Mul(topUpAmount, big.NewInt(int64(1<<batch.Depth)))
```
acud, line 252: since `topUpAmount` is never used below, you can just call `topUpAmount.Mul(big.NewInt(int64(1<<batch.Depth)))`

pkg/api/router.go
```diff
@@ -182,6 +182,13 @@ func (s *server) setupRouting() {
 		})),
 	)
 
+	handle("/stamps/topup/{id}/{amount}", web.ChainHandlers(
```
acud, line 185: this should _not_ be in the api. all postage APIs have moved to the debugAPI

pkg/api/postage.go
```diff
@@ -166,3 +166,55 @@ func (s *server) postageGetStampHandler(w http.ResponseWriter, r *http.Request)
 	}
 	jsonhttp.OK(w, &resp)
 }
+
+func (s *server) postageTopUpHandler(w http.ResponseWriter, r *http.Request) {
+	idStr := mux.Vars(r)["id"]
```
acud, line 171: once this is moved to the debugapi, you;ll need to acquire the semaphore protecting multiple postage contract events

pkg/postage/service.go
acud, line 118: this is not necessary

acud (changes requested):
the endpoint is on the wrong api :)

pkg/postage/postagecontract/contract.go
aloknerurkar: Mul takes 2 arguments and puts result into the calling object.

pkg/api/router.go
aloknerurkar: Why do we still have the postage file in api package?

pkg/postage/batchservice/batchservice_test.go
aloknerurkar: This is to test that if the batch topup event is for a different owner, we should not see any topup events in the batchService.

pkg/postage/postagecontract/contract_test.go
```diff
@@ -192,3 +198,157 @@ func TestLookupERC20Address(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		si, err := postageMock.GetStampIssuer(batch.ID)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if si == nil {
```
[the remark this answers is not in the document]
aloknerurkar: We dont have the listener in this test. So we wont get the event to update the stamp issuer. This flow is tested in batch service so ignoring.

openapi/SwarmDebug.yaml
```diff
@@ -871,3 +871,40 @@ paths:
...
+      description: Be aware, this endpoint creates on-chain transactions and transfers BZZ from the node's Ethereum account and hence directly manipulates the wallet balance!
+      tags:
+        - Postage Stamps
+      parameters:
+        - in: path
+          name: id
+          schema:
+            $ref: "SwarmCommon.yaml#/components/schemas/BatchID"
+          required: true
+          description: Swarm address of the stamp
```
acud, line 887: "batch ID to top up"

openapi/SwarmDebug.yaml
```diff
@@ -871,3 +871,40 @@ paths:
...
+        - in: path
+          name: amount
+          schema:
+            type: integer
+          required: true
+          description: Amount of BZZ to top up to an existing postage batch.
```
acud, line 893: amount per chunk to top up

pkg/debugapi/postage.go
```diff
@@ -321,3 +321,63 @@ func (s *Service) estimateBatchTTL(id []byte) (int64, error) {
...
+		p, ok := big.NewInt(0).SetString(price[0], 10)
+		if !ok {
+			s.logger.Error("topup batch: bad gas price")
+			jsonhttp.BadRequest(w, errBadGasPrice)
+			return
+		}
+		ctx = sctx.SetGasPrice(ctx, p)
+	}
+
+	if !s.postageCreateSem.TryAcquire(1) {
```
acud, line 358: consider renaming the semaphore field to just `postageSem`. also, another suggestion would be to have a new middleware for all the postage APIs that do the on-chain ops and just acquire and `defer` the release of the  resource in the middleware instead of inlining it in the different endpoints

acud (commented):
Very nice, I have a few minor nitpicks.
Also, a beekeeper integration test that checks that the batch value increases by the amount that was requested would be needed too

pkg/debugapi/postage.go
aloknerurkar: This seems cleaner. Will do this as a part of the dilution PR.

aloknerurkar:
@acud Will add beekeeper tests for both topup and dilute in separate beekeeper PR.

acud (approved):

pkg/postage/postagecontract/contract.go
```diff
@@ -123,6 +129,38 @@ func (c *postageContract) sendCreateBatchTransaction(ctx context.Context, owner
...
+	if err != nil {
+		return nil, err
+	}
+
+	request := &transaction.TxRequest{
+		To:       &c.postageContractAddress,
+		Data:     callData,
+		GasPrice: sctx.GetGasPrice(ctx),
+		GasLimit: 160000,
+		Value:    big.NewInt(0),
```
ralph-pichler, line 144: please set the `Description` field. ideally also add it to the create batch and approve request.

pkg/postage/postagecontract/contract.go
```diff
@@ -204,6 +242,42 @@ func (c *postageContract) CreateBatch(ctx context.Context, initialBalance *big.I
...
+		return err
+	}
+
+	receipt, err := c.sendTopUpBatchTransaction(ctx, batch.ID, totalAmount)
+	if err != nil {
+		return err
+	}
+
+	for _, ev := range receipt.Logs {
+		if ev.Address == c.postageContractAddress && ev.Topics[0] == batchTopUpTopic {
```
ralph-pichler, line 273: this will panic if there is an event. with no topics. should never happen with the real contract but that should be handled nonetheless. Ideally use. `transaction.FindSingleEvent` which already has this.

ralph-pichler (changes requested):

pkg/postage/postagecontract/contract.go
aloknerurkar: So the `FindSingleEvent` parses the event as well. I did not want to do that as then we would have to define a struct locally for the event and we dont really need to check the contents of the event here.

So I have added a check for `len(ev.Topics) > 0`

ralph-pichler (approved):

PR #2409, opened 2021-08-15
feat: allow transaction from another sender for identity tx

ralph-pichler:
allow transactions whose transaction data matches the overlay eth address exactly (a 32-byte data), regardless of sender. the first 4 bytes are required to be 0 to distinguish it from a 32-byte data transaction from the overlay. That way every tx can only be valid for one overlay.

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2409)

### DIFF
--- a/pkg/transaction/sender_matcher.go
+++ b/pkg/transaction/sender_matcher.go
@@ -94,7 +94,7 @@ func (m *Matcher) Matches(ctx context.Context, tx []byte, networkID uint64, send
 		return nil, ErrTransactionPending
 	}
 
-	sender, err := types.Sender(m.signer, nTx)
+	type1sender, err := types.Sender(m.signer, nTx)
 	if err != nil {
 		err2 := m.storage.Put(peerOverlayKey(senderOverlay, incomingTx), &overlayVerification{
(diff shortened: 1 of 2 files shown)


pkg/transaction/sender_matcher.go
```diff
@@ -106,6 +106,12 @@ func (m *Matcher) Matches(ctx context.Context, tx []byte, networkID uint64, send
 		return nil, fmt.Errorf("%v: %w", err, ErrTransactionSenderInvalid)
 	}
 
+	var type2sender common.Address
+
+	if len(nTx.Data()) == 20 {
```
acud, line 111: can we have a small comment about how this works? can't really understand when this data is supposed to be `nil` or not, and the variable names could also be improved. what does type 1 mean and what does type 2 mean?

ralph-pichler: logic changed a bit. names should now be clearer and comments were added.

acud (approved):
Thanks for addressing the comments :pray:

aloknerurkar (approved):
Not exactly familiar with this but the change LGTM from go perspective

PR #2418, opened 2021-08-17
fix: only flush dirty metrics

### LINKED ISSUE #2417: Kademlia peer metrics flush incorrect
#### Summary
While investigating the slowness of node shutdown and discovering that it was mainly due to how long the kademlia takes to flush peer metrics, I noticed a seemingly incorrect test at:

https://github.com/ethersphere/bee/blob/5a6ddbdfd10ada5fdb1513253e97c098dfca746e/pkg/topology/kademlia/internal/metrics/metrics.go#L167

According to the comments earlier in that file:

```
	// Watches in-memory counters which has to be persisted.
	// 	3 - dirty, need to be persisted
	//	2 - snapshot of counters in progress
	//	1 - batched for persistent write
	//	0 - persisted
	dirty atomic.Int32
```

And looking at the references to dirty through the flush function, it appears that:

```
	if cs.dirty.Load() > 1 {
		return nil
	}
```

should be

```
	if cs.dirty.Load() < 3 {
		return nil
	}
```

This change will then cause only those that are "dirty, need to be persisted" to be flushed rather than only flushing those that are either already persisted and/or already batched for writing.

#### Steps to reproduce
To see this in action, you need to add a bunch of logging to your node and observe the dirty states and flush activities of the various metrics.

#### Expected behavior
Only skip flushing metrics that are not dirty allowing the actual dirty ones to take the processing time.

#### Actual behavior
The current code flushes kademlia metrics that do not need to be flushed wasting processing resources.

ldeffenb:
Fixes #2417

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2418)

### DIFF
--- a/pkg/topology/kademlia/internal/metrics/metrics.go
+++ b/pkg/topology/kademlia/internal/metrics/metrics.go
@@ -164,7 +164,7 @@ type Counters struct {
 
 // flush writes the current state of in memory counters into the given db.
 func (cs *Counters) flush(db *shed.DB, batch *leveldb.Batch) error {
-	if cs.dirty.Load() > 1 {
+	if cs.dirty.Load() < 3 {
 		return nil
 	}
 	cs.dirty.CAS(3, 2)


pkg/topology/kademlia/internal/metrics/metrics.go
```diff
@@ -164,7 +164,7 @@ type Counters struct {
 
 // flush writes the current state of in memory counters into the given db.
 func (cs *Counters) flush(db *shed.DB, batch *leveldb.Batch) error {
-	if cs.dirty.Load() > 1 {
+	if cs.dirty.Load() < 3 {
 		return nil
 	}
 	cs.dirty.CAS(3, 2)
```
acud, line 170: the CAS call here is now wrong since dirty will never be 3 anymore. it might be useful to sweep through the entire usage of this field to see that everything is correct

acud (changes requested):

pkg/topology/kademlia/internal/metrics/metrics.go
ldeffenb: How will it never be 3?  The earlier return is for those < 3, so only 3s will actually be coming through here.

acud (approved):
sorry @ldeffenb I got confused by the diff. Could I please ask to double check the the other usages of `cs.dirty` are correct?

ldeffenb:
> sorry @ldeffenb I got confused by the diff. Could I please ask to double check the the other usages of `cs.dirty` are correct?

Been there, done that.  cs.dirty goes from 3 to 2, then 2 to 1, and 1 to 0 as it progresses down through saving.  And it is set to 3 regardless of its current value when the metric is updated.  I did all of those checks back when I was trying to understand that initial "if" statement.

istae (approved):

Issue #2419, opened 2021-08-17
Inefficient, redundant, schema saving


ldeffenb:
#### Summary
While investigating the slowness of node shutdown and discovering that it was mainly due to how long the kademlia takes to flush peer metrics, I noticed that the metrics were actually updating the statestore schema for each field of each peer that was being saved.  In this process, the current schema is first loaded from the database and unmarshalled into RAM.  The new field is then added to this which is then marshalled and written to the DB.

Given that 100s or 1,000s of keys are added in a single kademlia metrics flush, this results in a lot of unnecessary overhead showing up as copious CPU consumption during metrics flushing.

#### Steps to reproduce
Run a new (or even existing) node for a while to accumulated new peer metrics.  When you shut down the node, you will see a large CPU consumption for a long time (read: seconds, to minutes, to hours).  Inspecting the active goroutines and their stacks, you will see that the kademlia metrics flusher is active and bounding around in the shed doing these schema modifications.

Note: I didn't detect the issue using the stacks, but with iterative additions of logging down through the various layers.

#### Expected behavior
Given that locks are already in place to ensure a single schema modification is active at any point in time, why re-read the schema?  According to my testing logs, the schema size is not large enough to impact memory utilization, so simply cache the current version between updates.

#### Actual behavior
Bee wastes lots of CPU resources loading and unmarshalling the schema during kademlia peer metrics flushing.  By caching the current schema version, this time is easily reduced by 50%.

mrekucci:
Fixed in #2452

PR #2420, opened 2021-08-17
feat: cache schema between updates

ldeffenb:
Implements #2419

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2420)

### DIFF
--- a/pkg/shed/db.go
+++ b/pkg/shed/db.go
@@ -51,6 +51,7 @@ type Options struct {
 // information about naming and types.
 type DB struct {
 	ldb     *leveldb.DB
+	sch     *schema
 	metrics metrics
 	quit    chan struct{} // Quit channel to stop the metrics collection before closing the database
 }
@@ -95,6 +96,7 @@ func NewDBWrap(ldb *leveldb.DB) (db *DB, err error) {
 
 	db = &DB{
 		ldb:     ldb,
+		sch:     nil,
 		metrics: newMetrics(),
 	}
 
--- a/pkg/shed/schema.go
+++ b/pkg/shed/schema.go
@@ -147,17 +147,22 @@ func (db *DB) schemaIndexPrefix(name string) (id byte, err error) {
 // getSchema retrieves the complete schema from
 // the database.
 func (db *DB) getSchema() (s schema, err error) {
+	if db.sch != nil {
+		return *db.sch, nil
+	}
 	b, err := db.Get(keySchema)
 	if err != nil {
 		return s, err
 	}
 	err = json.Unmarshal(b, &s)
+	db.sch = &s
 	return s, err
 }
 
 // putSchema stores the complete schema to
 // the database.
 func (db *DB) putSchema(s schema) (err error) {
+	db.sch = &s
 	b, err := json.Marshal(s)
 	if err != nil {
 		return err


pkg/shed/db.go
```diff
@@ -95,6 +96,7 @@ func NewDBWrap(ldb *leveldb.DB) (db *DB, err error) {
 
 	db = &DB{
 		ldb:     ldb,
+		sch:     nil,
```
acud, line 99: no need to initialize a field that is a pointer to a `nil` value as this is the default value

pkg/shed/schema.go
```diff
@@ -147,17 +147,22 @@ func (db *DB) schemaIndexPrefix(name string) (id byte, err error) {
 // getSchema retrieves the complete schema from
 // the database.
 func (db *DB) getSchema() (s schema, err error) {
+	if db.sch != nil {
```
acud, line 150: since there's no mutex here, this will result in a panic

pkg/shed/schema.go
```diff
@@ -147,17 +147,22 @@ func (db *DB) schemaIndexPrefix(name string) (id byte, err error) {
...
+		return *db.sch, nil
+	}
 	b, err := db.Get(keySchema)
 	if err != nil {
 		return s, err
 	}
 	err = json.Unmarshal(b, &s)
+	db.sch = &s
```
acud, line 158: same here

pkg/shed/schema.go
```diff
@@ -147,17 +147,22 @@ func (db *DB) schemaIndexPrefix(name string) (id byte, err error) {
...
 	return s, err
 }
 
 // putSchema stores the complete schema to
 // the database.
 func (db *DB) putSchema(s schema) (err error) {
+	db.sch = &s
```
acud, line 165: same here

acud (changes requested):

acud:
Also, I am not sure how this will behave when we run schema migrations. @janos?

pkg/shed/schema.go
ldeffenb: How will testing if the schema is nil cause a panic?  We're not de-referencing db.sch here, but simply testing if it is populated.

pkg/shed/schema.go
ldeffenb: Again, since we're just remembering the schema, I don't see any chance for a panic here.

pkg/shed/schema.go
ldeffenb: And ditto to this one.  sch is a pointer to the schema, not the schema itself.  No panics are possible AFAIK.

pkg/shed/db.go
ldeffenb: I included that just to make it clear that the initial state is that no schema is currently cached.

acud: it is not needed. a comment should suffice

pkg/shed/schema.go
acud: a pointer is still a memory address and has a value, that value is stored in the memory address. the fact that we are not dereferencing the value does not mean we are not accessing the same memory address (in this case of the pointer) and this needs synchronization. afaik when the runtime tries to read and write at the same time to the same memory location it will panic. the fact that you haven't seen the panic does not mean it will not happen. in any case, when accessing a shared resource across different goroutines it is accustomed to use a mutex in order to prevent data races

janos:
@ldeffenb thanks fo this PR, but I think that the problem described in should be actually fixed in kademlia metrics where shed is used. I think that shed is not needed at all there as it is used with such a big overhead on changing schema on every update. I am not sure if the shed.Uint64Field is the correct type to be used within the current Flush implementation and that https://github.com/ethersphere/bee/blob/master/pkg/shed/field_struct.go#L28 can be used much more efficiently while keeping the shed for metrics storage.

I think that caching the schema may lead to some unexpected and hidden issues with shed schemas as @acud already pointed about migrations.

ldeffenb:
@janos - I concur that the proper long-term fix is to change the way kademlia persists the peer metrics.  But the node operators need some sooner relief when their nodes take literally hours to shut down cleanly, during which time they are undetectably inoperable.

@acud - I appreciate your concerns about unexpected and hidden issues with this use of the shed, but if you consider the exposed API consisting only of load and save of the entire schema, you will begin to see that this change is completely safe within that architecture.  The window would seem to already exist that multiple routines can fetch the schema, make their own modifications to it and the last one to put the schema wins.

However, I did examine all of the callers of these routines and determined that they all provide the locking protection (or suitable single-threadedness) that you and I know would need to exist to keep this approach "safe".  Otherwise there would already be issues occurring with losing pieces of the schema if there existed any unprotected use of schema get/put.

The choice is ultimately your's on whether or not you accept the PR.  If it helps, I've been running this on my nodes for two or three months now because I restart them too often to deal with the delays.  And none of those nodes have crashed for any reason, let alone this one.  And I figure it's better to have this fix running rather than just aborting the node while it is still trying to write out the metrics which risks even more DB corruption.

acud:
> Oh, and migrations occur before the node starts anything, so they would not be subject to any concurrent access or there'd be many many more issues with db consistency in the face of migrations, especially with the single get/put API to the schema.

The problem with migrations is not one of concurrency, but one of caching. When the db schema changes during a migration the actual schema changes and since you are caching it, you are caching the old schema, and it is a bit unclear how this will behave when we do need to run a migration

janos:
@ldeffenb it is correct that the problem is very severe, but this solution may present even bigger ones. One of which is caching the schema can result in invalid schema information as it may change on every metrics flush where keys for counters are updated. I do not see that changing to use StructField would be so complex or difficult. On the other hand, I would even question the need for persisting metrics as they are part of the instrumentation, persisted by the infrastructure.

ldeffenb:
> The problem with migrations is not one of concurrency, but one of caching. When the db schema changes during a migration the actual schema changes and since you are caching it, you are caching the old schema, and it is a bit unclear how this will behave when we do need to run a migration
> 
> Just to reiterate, when different `shed.Index`es keep different shadow copies of the schema and the schema changes they will not update their schema.

If the different indexes are keeping shadow copies of the schema, and if they eventually put an updated version of that, they will have instantly overwritten changes to the schema made by other users.  There is only a get and a put on the schema, there is no "merge" nor "individual update".  So, if a cached copy exists outside of the shed, it is hopefully only for local reference and not subsequent update.  And if this is true, then they obviously are not interested in any outside updates because if they were, they'd have to be getting the schema anyway in which case the proposed change is perfectly safe.

ldeffenb:
> @ldeffenb it is correct that the problem is very severe, but this solution may present even bigger ones. One of which is caching the schema can result in invalid schema information as it may change on every metrics flush where keys for counters are updated. I do not see that changing to use StructField would be so complex or difficult. On the other hand, I would even question the need for persisting metrics as they are part of the instrumentation, persisted by the infrastructure.

It is precisely the schema change on metrics flushing that drove me to propose this change.  Because of the volume of schema changes caused by a single metrics flush, there is a severe performance drain on the bee node.  My first thought was "why, oh why are they persisting the metrics with schema updates", and then it was "ok, they're doing it that way, so let's make it less of an impact" which led to this proposal.

Changing the actual peer metrics persistence approach and/or eliminating the persistence entirely goes beyond my current golang abilities, so I'll have to leave that to someone more capable than myself.

Hopefully this performance issue can be raised internally and a decision made, because the current implementation is certainly hurting bee performance and the swarm availability when swap-endpoint issues arise and nodes go invisibly offline for hours as they work to persist their peer metrics during kademlia shutdown.  (Although my other two PRs will mitigate that somewhat)

ldeffenb:
It seems there is strong resistance to accepting this PR, even as a performance stopgap until a more proper peer metrics persistence approach can be conceived and implemented.  I am perfectly fine if you simply close it now, but I personally will continue to run with this in my nodes as it truly does deliver quite the performance boost / resource reduction.

acud:
Yep I would scratch this PR and I would suggest the following: instead of waiting for the Finalize method to finish, I would pass an expiring context to it that expires with a reasonable amount of time (say, 5 seconds), and just exit the finalize method in case the context expires. Since this information is not mission critical, we can live with a partial dirty shutdown here.

acud:
proposed solution here https://github.com/ethersphere/bee/pull/2428

PR #2424, opened 2021-08-17
feat: flush metrics in separate goroutine

### LINKED ISSUE #2423: Kademlia doesn't save peer metrics
#### Summary
While investigating the slowness of node shutdown and discovering that it was mainly due to how long the kademlia takes to flush peer metrics, I noticed that the metrics were not actually being saved every 15 seconds as the code would seem to indicate.  This is due to how long and how many pending requests there are for the connection requirements of MangeC.

And worse, when/if the peer metrics do attempt to flush, they take a very long time to complete during which time the connection features of the ManageC loop do not execute resulting in lost peer connections not being recovered even to the point of losing kademlia depth which has serious ramifications on the pullsync activity.  This is actually what led me to this research when I saw my pullsync activity ramp up only to notice that my topology depth had dropped and then to notice that kademlia had not logged any connections attempts for several hours.

#### Steps to reproduce
Just look at the metrics of any long-running active mainnet node.  The following node has been running for weeks and has logged a SINGLE `bee_kademlia_internal_metrics_flush_time_count` which took 5.4702293483513e+13 of whatever units this time is specified in, but it's a LONG time.
```
# HELP bee_kademlia_internal_metrics_flush_time The time spent flushing the internal metrics about peers to the state-store.
# TYPE bee_kademlia_internal_metrics_flush_time histogram
bee_kademlia_internal_metrics_flush_time_bucket{le="0.005"} 0
<snip>
bee_kademlia_internal_metrics_flush_time_bucket{le="10"} 0
bee_kademlia_internal_metrics_flush_time_bucket{le="+Inf"} 1
bee_kademlia_internal_metrics_flush_time_sum 5.4702293483513e+13
bee_kademlia_internal_metrics_flush_time_count 1
# HELP bee_kademlia_internal_metrics_flush_total_errors Number of total errors occurred during flushing the internal metrics to the state-store.
# TYPE bee_kademlia_internal_metrics_flush_total_errors counter
bee_kademlia_internal_metrics_flush_total_errors 0
```

#### Expected behavior
I would expect the kademlia peer metrics to flush out periodically reducing the number that need to flush at process shutdown.  Further, I would expect connection attempts to continue even though metrics were taking a long time to flush.

#### Actual behavior
Per the metrics, the kademlia peer metrics are not flushing while the node is running but do a full flush at node shutdown.

ldeffenb:
Fixes #2423 

This PR moves the kademlia metrics flushing out of the main ManageC loop and into a dedicated goroutine.  This solves the issue of the 15 second timer never (or rarely) actually firing and the metrics never actually flushing until the node is terminated.  The new goroutine is added to the wait group and marks itself done when it exits so that the kademlia shutdown logic remains consistent.

I also changed the timer to 5 minutes since the metrics do actually get flushed now without competing with the connector.  The shutdown logic in Close still Finalizes the metrics collector which will ensure that the most recent updates are still persisted.

You will also notice a new Tracef log showing that the flush happened and how long it took to execute. This is similar to the existing log in the connector loop that shows activity along with the new depth.  At least with this log, we can be sure that the metrics are flushing periodically.

I did consider if this is "safe" to write the metrics from an independent goroutine for locking and consistency purposes, and given the "dirty" CAS counter on the individual metrics, I'm certain that metrics consistency will be eventually attained even if a given peer metric changes while being persisted.  This seems to be the only reason for the dirty counter on the metrics anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2424)

### DIFF
--- a/pkg/topology/kademlia/kademlia.go
+++ b/pkg/topology/kademlia/kademlia.go
@@ -506,19 +506,32 @@ func (k *Kad) manage() {
 	peerConnChan2 := make(chan *peerConnInfo)
 	go k.connectionAttemptsHandler(ctx, &wg, peerConnChan, peerConnChan2)
 
+	k.wg.Add(1)
+	go func() {
+		defer k.wg.Done()
+		for {
+			select {
+			case <-k.halt:
+				return
+			case <-k.quit:
+				return
+			case <-time.After(5 * time.Minute):
+				start := time.Now()
+				if err := k.collector.Flush(); err != nil {
+					k.metrics.InternalMetricsFlushTotalErrors.Inc()
+					k.logger.Debugf("kademlia: took %s unable to flush metrics counters to the persistent store: %v", time.Since(start), err)
+				} else {
+					k.metrics.InternalMetricsFlushTime.Observe(float64(time.Since(start).Nanoseconds()))
+					k.logger.Tracef("kademlia took %s to flush", time.Since(start))
+				}
+			}
+		}
+	}()
+
 	for {
 		select {
 		case <-k.quit:
 			return
-		case <-time.After(15 * time.Second):
-			start := time.Now()
-			if err := k.collector.Flush(); err != nil {
-				k.metrics.InternalMetricsFlushTotalErrors.Inc()
-				k.logger.Debugf("kademlia: unable to flush metrics counters to the persistent store: %v", err)
-			} else {
-				k.metrics.InternalMetricsFlushTime.Observe(float64(time.Since(start).Nanoseconds()))
-			}
-			k.notifyManageLoop()
 		case <-k.manageC:
 			start := time.Now()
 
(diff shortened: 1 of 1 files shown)


acud:
This looks ok. But there's no guarantee that a flushing operation does not occur twice on shutdown. Could you please somehow ensure that?

ldeffenb:
> This looks ok. But there's no guarantee that a flushing operation does not occur twice on shutdown. Could you please somehow ensure that?

That is why the dirty flag is set and checked on each peer's metrics.  If the metrics are not dirty, they won't redundantly flush, either on the periodic or Finalize.  The same collector.flush() method is used in both instances and it will skip over metrics that are clean, provided, of course, that #2418 is adopted.

ldeffenb:
I noticed in my final review and testing of this PR that I inadvertently removed the 15 second k.notifyManageLoop() reducing the interval at which connections are attempted outside of receiving new peer notifications.  I restored this timer to notify, but retained the 5 minute timer in the new goroutine as it doesn't make sense IMHO to poll all of the peer metrics every 15 seconds just to notify the connection loop.

acud (approved):

istae (approved):

PR #2425, opened 2021-08-17
fix: random generation of tag uids

metacertain:
This PR attempts to solve a problem with fixed seed generation of new tags

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2425)

### DIFF
--- a/pkg/tags/tags.go
+++ b/pkg/tags/tags.go
@@ -49,16 +49,25 @@ type Tags struct {
 	tags       *sync.Map
 	stateStore storage.StateStorer
 	logger     logging.Logger
+	rand       *rand.Rand
+	randM      sync.Mutex
 }
 
 // NewTags creates a tags object
 func NewTags(stateStore storage.StateStorer, logger logging.Logger) *Tags {
-	rand.Seed(time.Now().UnixNano())
+
 	return &Tags{
 		tags:       &sync.Map{},
 		stateStore: stateStore,
 		logger:     logger,
+		rand:       rand.New(rand.NewSource(time.Now().UnixNano())),
 	}
+}
+
+func (ts *Tags) TagUidFunc() uint32 {
+	ts.randM.Lock()
+	defer ts.randM.Unlock()
+	return ts.rand.Uint32()
 }
 
 // Create creates a new tag, stores it by a not yet in use UID and returns it
@@ -69,7 +78,7 @@ func (ts *Tags) Create(total int64) (*Tag, error) {
 	var uid uint32
 
 	for exists {
-		uid = TagUidFunc()
+		uid = ts.TagUidFunc()
 		if _, loaded := ts.tags.Load(uid); !loaded {
 			exists = false
 		}
--- a/pkg/tags/tags.go
+++ b/pkg/tags/tags.go
@@ -53,7 +53,7 @@ type Tags struct {
 
 // NewTags creates a tags object
 func NewTags(stateStore storage.StateStorer, logger logging.Logger) *Tags {
-	rand.Seed(time.Now().Unix())
+	rand.Seed(time.Now().UnixNano())
 	return &Tags{
 		tags:       &sync.Map{},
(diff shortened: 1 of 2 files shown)


zelig (commented):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2425)* status: all files reviewed, 1 unresolved discussion (waiting on @metacertain)

___
*[pkg/tags/tags_test.go, line 102 at r1](https://reviewable.io/reviews/ethersphere/bee/2425#-MhJtSRE9Yb6XIzQcXUs:-MhJtSRE9Yb6XIzQcXUt:bme811k) ([raw file](https://github.com/ethersphere/bee/blob/bed1d8a0076ca38f29b1253d6605329542ff8fc5/pkg/tags/tags_test.go#L102)):*
> ```Go
> 	// this highlights that upon tags.Create(), already existing values are only checked in sync.Map but not in statestore
> 
> 	time.Sleep(1 * time.Second)
> ```

take the seed from unixnano instead

pkg/tags/tags.go
```diff
@@ -53,17 +53,29 @@ type Tags struct {
 
 // NewTags creates a tags object
 func NewTags(stateStore storage.StateStorer, logger logging.Logger) *Tags {
+	rand.Seed(time.Now().Unix())
```
aloknerurkar, line 56:
```suggestion
	rand.Seed(time.Now().UnixNano())
```
After this, you should be able to remove the sleep from the test as well.

metacertain: Well, the sleep is still needed for some reason, though now only for a nanosecond

aloknerurkar (approved):

pkg/tags/tags.go
```diff
@@ -53,17 +53,29 @@ type Tags struct {
 
 // NewTags creates a tags object
 func NewTags(stateStore storage.StateStorer, logger logging.Logger) *Tags {
+	rand.Seed(time.Now().UnixNano())
```
acud, line 56: this is not so great as it sets the global rand seed. we should generally refrain from doing such things. it would be better to just pass a new source here as a parameter and do the same within the test. it should simplify things and clean this up

acud (changes requested):

acud:
@metacertain can we wrap this up?

pkg/tags/tags.go
metacertain: well, we can use a different source, it needs a lock, as only the global rand is safe to use in concurrency
I will replace it

pkg/tags/tags.go
```diff
@@ -49,21 +49,42 @@ type Tags struct {
...
 }
 
 // NewTags creates a tags object
 func NewTags(stateStore storage.StateStorer, logger logging.Logger) *Tags {
+
 	return &Tags{
 		tags:       &sync.Map{},
 		stateStore: stateStore,
 		logger:     logger,
+		rand:       rand.New(rand.NewSource(time.Now().UnixNano())),
```
acud, line 63: this is the same as it was before just that it has been moved a few lines :) the idea was that the `Source` can be passed to the constructor as a parameter, allowing you to create two `NewTags` with different sources on the test level, avoid the needs for sleeps

pkg/tags/tags_test.go
```diff
@@ -121,7 +128,7 @@ func TestListAll(t *testing.T) {
 	}
 
 	if len(tagList3) != 5 {
-		t.Fatalf("want %d tags but got %d", 5, len(tagList2))
+		t.Fatalf("want %d tags but got %d", 5, len(tagList3))
```
acud, line 131: how did this change?

metacertain: Variable logged didn't match one in preceding check, which came to light when the test failed with "want 0, got 0" during some earlier state.

acud (approved):

PR #2428, opened 2021-08-18
fix: allow dirty metrics shutdown

acud:
this PR introduces an expiring context to the metrics finalizer in kademlia. as noted by @ldeffenb the shutdown takes too long and we should not wait indefinitely there. this is a stopgap solution until we come up with a better persistence solution that doesn't burden the shed abstraction with too much IO with regards to these metrics. We also need metrics pruning, since this can just burden other components with statestore size at some point.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2428)

### DIFF
--- a/pkg/topology/kademlia/internal/metrics/metrics.go
+++ b/pkg/topology/kademlia/internal/metrics/metrics.go
@@ -7,6 +7,7 @@
 package metrics
 
 import (
+	"context"
 	"fmt"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


ldeffenb:
I'm not sure I would have gone with as short as 5 seconds, but this will certainly solve the issue for those folks complaining about there Docker-based nodes not terminating.  With this, their node will actually exit allowing the implementation of detection and auto-restart of the failed node, especially those failing due to swap-endpoint glitches.

acud:
> I'm not sure I would have gone with as short as 5 seconds, but this will certainly solve the issue for those folks complaining about there Docker-based nodes not terminating. With this, their node will actually exit allowing the implementation of detection and auto-restart of the failed node, especially those failing due to swap-endpoint glitches.

we need to be cautious with restart times. kubernetes, for example, will kill a process that cannot terminate within a 30s timeout. so 5s might even be too permissive depending on the performance of the node's statestore. so if a process got a SIGQUIT or SIGTERM and does not exit in time will be killed (and with it the risk for all sorts of inconsistencies)

ldeffenb:
Copy that and understood.  I'm assuming that's only when kubernetes is directing the termination and not when the process is shutting itself down from inside?  But you are correct, it's better to have the process miss persisting the peer metrics than to have the carpet pulled out from under it!

acud:
> I'm assuming that's only when kubernetes is directing the termination

That is correct and can happen for a variety of reasons (pod rescheduling on a different node, or a simple upgrade to a new docker version). When the node shuts down it might actually kill it too; I think that the `readiness` and `health` endpoints are some of the first things to shut down when the node is brought down (we take down the APIs first so that new requests don't come in).

metacertain (approved):

aloknerurkar (approved):

janos (approved):

PR #2430, opened 2021-08-19
feat: balancer connector skips unpopulated, out of depth bins

istae:
PR intends to limit the balanced connector's domain of work to bins that could be balanced.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2430)

### DIFF
--- a/pkg/topology/kademlia/kademlia.go
+++ b/pkg/topology/kademlia/kademlia.go
@@ -248,7 +248,20 @@ func (k *Kad) connectBalanced(wg *sync.WaitGroup, peerConnChan chan<- *peerConnI
 		return false
 	}
 
+	depth := k.NeighborhoodDepth()
+
 	for i := range k.commonBinPrefixes {
+
+		binPeersLength := k.knownPeers.BinPeersLength(uint8(i))
+
+		// balancer should skip on bins where neighborhood connector would connect to peers anyway
+		// and there are not enough peers in known addresses to properly balance the bin
+		if i >= int(depth) && binPeersLength < len(k.commonBinPrefixes[i]) {
+			continue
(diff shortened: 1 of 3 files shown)


pkg/topology/kademlia/kademlia.go
```diff
@@ -967,6 +981,46 @@ func closestPeer(peers *pslice.PSlice, addr swarm.Address, spf sanctionedPeerFun
 	return closest, nil
 }
 
+func closestPeerList(peers []swarm.Address, addr swarm.Address, spf sanctionedPeerFunc) (swarm.Address, error) {
```
acud, line 984: since similar function already exists, can we somehow have it refactored to cater for this functionality also?

istae: Done.

pkg/topology/kademlia/kademlia.go
```diff
@@ -306,13 +319,13 @@ func (k *Kad) connectNeighbours(wg *sync.WaitGroup, peerConnChan chan<- *peerCon
 
 	sent := 0
 	var currentPo uint8 = 0
-	_ = k.knownPeers.EachBinRev(func(addr swarm.Address, po uint8) (bool, bool, error) {
+
+	_ = k.knownPeers.EachBin(func(addr swarm.Address, po uint8) (bool, bool, error) {
```
acud, line 323: this changes the iterator to be from deepest to closest, not sure if it is needed (we'd like to connect to shallower peers first, so that depth can gradually go deeper)

istae: good catch, this is not needed, a typo.

PR #2438, opened 2021-08-23
fix: chain ID error message

mrekucci:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2438)

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -233,7 +233,7 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
 	b.transactionMonitorCloser = transactionMonitor
 
 	if o.ChainID != -1 && o.ChainID != chainID {
-		return nil, fmt.Errorf("connected to wrong ethereum network: got chainID %d, want %d", chainID, o.ChainID)
+		return nil, fmt.Errorf("connected to wrong ethereum network; network chainID %d; configured chainID %d", chainID, o.ChainID)
 	}
 
 	var debugAPIService *debugapi.Service


notanatol (approved):

pkg/node/node.go
```diff
@@ -233,7 +233,7 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
 	b.transactionMonitorCloser = transactionMonitor
 
 	if o.ChainID != -1 && o.ChainID != chainID {
-		return nil, fmt.Errorf("connected to wrong ethereum network: got chainID %d, want %d", chainID, o.ChainID)
+		return nil, fmt.Errorf("connected to wrong ethereum network: got chainID %d, want %d", o.ChainID, chainID)
```
ralph-pichler, line 236: this seems wrong. `chainID` is what we get from the backend, so why would that be `want` instead of `got`?.

mrekucci: That depends on the perspective and context. E.g.: If (in the config) we set the network ID to 1, this will result to set the config chain ID value to 100 (see `getConfigByNetworkID` function). When we connect to network ID 5 instead of 1, this will result in the following error: `Error: connected to wrong ethereum network: got chainID 5, want 100`. So if our intention was to connect to the network ID 5 but we (by mistake) set the network ID to 1, from that perspective we want chain ID to be 5 but got 100. So I find the `got` and `want` bit misleading. With that being said, I'll change at least the phrasing of that message.

ralph-pichler (approved):

PR #2443, opened 2021-08-24
refactor(loadsave): simplify interface

acud:
this PR simplifies the `loadsave` abstraction to remove state like encryption and mode. those are now injected through a pipeline factory. It also handles more explicitly the cases where a readonly loadsaver is needed, so that readonly operations on manifest can happen. this is now possible by requesting explicitly a read-only loadsaver.

also adds missing unit test coverage and documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2443)

### DIFF
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -142,7 +142,8 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request, store
 	}
 
 	encrypt := requestEncrypt(r)
-	l := loadsave.New(storer, requestModePut(r), encrypt)
+	factory := requestPipelineFactory(ctx, storer, r)
+	l := loadsave.New(storer, factory)
 
 	m, err := manifest.NewDefaultManifest(l, encrypt)
 	if err != nil {
@@ -238,7 +239,7 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request, store
 
 func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
 	logger := tracing.NewLoggerWithTraceID(r.Context(), s.logger)
-	ls := loadsave.New(s.storer, storage.ModePutRequest, false)
+	ls := loadsave.NewReadonly(s.storer)
 	feedDereferenced := false
 
 	targets := r.URL.Query().Get("targets")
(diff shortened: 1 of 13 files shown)


janos (commented):
Looks good, but I do not know why tags test is failing.

acud:
should be ok now. the tag pointer on context propagation is biting back. this needs to be tackled as part of technical debt

janos (approved):
👍

pkg/api/bzz.go
```diff
@@ -210,6 +211,7 @@ func (s *server) fileUploadHandler(w http.ResponseWriter, r *http.Request, store
 	logger.Debugf("bzz upload file: manifest reference: %s", manifestReference.String())
 
 	if created {
+		fmt.Println("donesplit")
```
aloknerurkar, line 214: remove debug print

pkg/file/loadsave/loadsave.go
```diff
@@ -21,16 +31,23 @@ type PutGetter interface {
...
-	storer    PutGetter
-	mode      storage.ModePut
-	encrypted bool
+	storer     PutGetter
+	pipelineFn func() pipeline.Interface
 }
 
-func New(storer PutGetter, mode storage.ModePut, enc bool) file.LoadSaver {
+// New returns a new read-write load-saver.
+func New(storer PutGetter, pipelineFn func() pipeline.Interface) file.LoadSaver {
```
aloknerurkar, line 39: Seems like we already import `pipeline/builder here`. Would it be better to just instantiate the pipeline within this constructor?

That way we don't do a breaking change to the loadsave API.

acud: unfortunately this is not possible (at least not right now). every time you do a `save` you save an arbitrary length octet stream and so every save operation will result in a different root hash. a manifest can execute `save` more than once with the same loadsave instance. therefore, in order to reuse the same instance we need to implement a `reset` method on the pipeline. it would be nice to tackle this as a part of technical debt (as it is indeed a waste of resources to create a new pipeline every time. but then again - unless you're constantly uploading stuff it is a one-time-cost)

aloknerurkar (approved):

PR #2450, opened 2021-08-26
fix: cors expose swarm-tag header for bzz dir upload

AuHau:
I was tracking down why in browser I would not get created `Swarm-Tag` header for uploaded collection, but I would get it for uploaded file. 

I added some checks for tests for this which did not catch the problem as it was CORS related, but I think I can leave the tests checks for completnes.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2450)

### DIFF
--- a/pkg/api/tag_test.go
+++ b/pkg/api/tag_test.go
@@ -123,7 +123,7 @@ func TestTags(t *testing.T) {
 			jsonhttptest.WithRequestHeader(api.SwarmTagHeader, strconv.FormatUint(uint64(tr.Uid), 10)),
 		)
 
-		IsTagFoundInResponse(t, rcvdHeaders, &tr)
+		isTagFoundInResponse(t, rcvdHeaders, &tr)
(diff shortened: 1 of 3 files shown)


pkg/api/tag_test.go
```diff
@@ -410,9 +410,9 @@ func TestTags(t *testing.T) {
 	})
 }
 
-// isTagFoundInResponse verifies that the tag id is found in the supplied HTTP headers
+// IsTagFoundInResponse verifies that the tag id is found in the supplied HTTP headers
 // if an API tag response is supplied, it also verifies that it contains an id which matches the headers
-func isTagFoundInResponse(t *testing.T, headers http.Header, tr *api.TagResponse) uint32 {
+func IsTagFoundInResponse(t *testing.T, headers http.Header, tr *api.TagResponse) uint32 {
```
acud, line 415: why is it needed to export this function? it is in the same namespace as the tests

AuHau: Sorry about that. I originally did not notice that they are in the same package and then forgot to revert that. Not I reverted it.

acud (approved):

PR #2452, opened 2021-08-26
fix: inefficient kademlia metrics saving

mrekucci:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2452)


pkg/topology/kademlia/internal/metrics/metrics.go
```diff
@@ -135,77 +109,56 @@ func (ss *Snapshot) HasAtMaxOneConnectionAttempt() bool {
 	return ss.LastSeenTimestamp == 0 && ss.SessionConnectionRetry <= 1
 }
 
+// persistentCounters is a helper struct used for persisting selected counters.
+type persistentCounters struct {
+	PeerAddress       *swarm.Address `json:"peerAddress"`
```
acud, line 114: why pointer?

pkg/topology/kademlia/internal/metrics/metrics.go
```diff
@@ -135,77 +109,56 @@ func (ss *Snapshot) HasAtMaxOneConnectionAttempt() bool {
...
-	ls, ok := cs.persistentLastSeenTimestamp.Load().(*shed.Uint64Field)
-	if !ok {
-		mk := newPeerKey(peerLastSeenTimestamp, key)
-		field, err := db.NewUint64Field(mk.String())
-		if err != nil {
-			return fmt.Errorf("field initialization for %q failed: %w", mk, err)
-		}
-		ls = &field
-		cs.persistentLastSeenTimestamp.Store(ls)
+	if cs != nil {
```
acud, line 142: is this meant to protect against a panic? if so, it also swallows an error and does not indicate that a programmer error has occurred. i would suggest to remove the if statement and directly access the fields. is there any case in which this is `nil`?

pkg/topology/kademlia/internal/metrics/metrics.go
```diff
@@ -272,108 +248,59 @@ func (c *Collector) Snapshot(t time.Time, addresses ...swarm.Address) map[string
...
-// Finalize tries to logs out all ongoing peer sessions.
+// Finalize tries to log out all ongoing peer sessions.
 func (c *Collector) Finalize(ctx context.Context, t time.Time) error {
-	var (
-		mErr  error
-		batch = new(leveldb.Batch)
-	)
-
 	c.counters.Range(func(_, val interface{}) bool {
-		select {
```
acud, line 353: since this can take a while, it might make the shutdown process take much too long than it is expected to take. this was added explicitly for this reason. unless we can guarantee that by no circumstances this `Range` operation is not going to be time consuming.

pkg/topology/kademlia/internal/metrics/metrics.go
```diff
@@ -272,108 +248,59 @@ func (c *Collector) Snapshot(t time.Time, addresses ...swarm.Address) map[string
...
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+		if err := c.Flush(); err != nil {
+			return err
 		}
 	}
 
 	c.counters.Range(func(_, val interface{}) bool {
```
acud, line 299: since we are shutting down here, i'm not sure why it would be needed to remove entries from an in-memory map

acud (changes requested):

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2452)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @acud)

___
*[pkg/topology/kademlia/internal/metrics/metrics.go, line 114 at r1](https://reviewable.io/reviews/ethersphere/bee/2452#-Mi2MjDdiKh5PN_Hudgf-r1-114:-MiQ1v-fEKcGN5r9NBzU:bt2vr1b) ([raw file](https://github.com/ethersphere/bee/blob/e686dede48dc66251a9570b73129751dd807f113/pkg/topology/kademlia/internal/metrics/metrics.go#L114)):*

why pointer?
</blockquote></details>

Fixed.
___
*[pkg/topology/kademlia/internal/metrics/metrics.go, line 142 at r1](https://reviewable.io/reviews/ethersphere/bee/2452#-Mi2MjDdiKh5PN_Hudgf-r1-142:-MiQ2Bm2B7o_Z3CyqGj7:bt2vr1b) ([raw file](https://github.com/ethersphere/bee/blob/e686dede48dc66251a9570b73129751dd807f113/pkg/topology/kademlia/internal/metrics/metrics.go#L142)):*

is this meant to protect against a panic? if so, it also swallows an error and does not indicate that a programmer error has occurred. i would suggest to remove the if statement and directly access the fields. is there any case in which this is `nil`?
</blockquote></details>

Fixed.
___
*[pkg/topology/kademlia/internal/metrics/metrics.go, line 299 at r1](https://reviewable.io/reviews/ethersphere/bee/2452#-Mi2MjDdiKh5PN_Hudgf-r1-299:-MiQ2VWr7T1P37KCVTZz:b-ym4wr4) ([raw file](https://github.com/ethersphere/bee/blob/e686dede48dc66251a9570b73129751dd807f113/pkg/topology/kademlia/internal/metrics/metrics.go#L299)):*

since we are shutting down here, i'm not sure why it would be needed to remove entries from an in-memory map
</blockquote></details>

It's just that the `Finalize` is used in the shutdown procedure, I'm adding a boolean flag that will allow skipping this.

pkg/topology/kademlia/internal/metrics/metrics.go
mrekucci: Fixed.

acud (approved):

pkg/localstore/migration_kademlia_metrics.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"time"
+)
+
+// DBSchemaKademliaMetrics is the bee schema identifier for kademlia metrics.
+const DBSchemaKademliaMetrics = "kademlia-metrics"
+
+// migrateKademliaMetrics removes all old existing
+// kademlia metrics database content.
+func migrateKademliaMetrics(db *DB) error {
```
acud, line 16: this is in the wrong package.... the migration should be in the statestore package, not localstore, as the statestore stores the kademlia metrics, not localstore

acud (changes requested):

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2452)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @acud)

___
*[pkg/localstore/migration_kademlia_metrics.go, line 16 at r5](https://reviewable.io/reviews/ethersphere/bee/2452#-MiWrao7ipUlSNFh7qI0-r5-16:-Mj-MEjy4rs7C3z0N1r3:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/d7e4d9db5e1b8a83b8bcaa9c76e628f77ac58390/pkg/localstore/migration_kademlia_metrics.go#L16)):*

this is in the wrong package.... the migration should be in the statestore package, not localstore, as the statestore stores the kademlia metrics, not localstore
</blockquote></details>

Done.

mrekucci:
> Is there going to be a migration to remove all of the obsolete keys from the statestore?

Yes, it is already included.

acud (approved):
nice, thanks!

acud (approved):
> Reviewed 1 of 5 files at r1, 2 of 3 files at r3, 6 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2452)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mrekucci)

PR #2453, opened 2021-08-26
feat: pslice 2d

istae:
2D array implementation of the pslice that has shown great improvements on cpu and memory usage. The benchmark results are below. 

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2453)


istae:
image

acud (changes requested):
> Reviewed 1 of 5 files at r1, 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2453)* status: 4 of 5 files reviewed, 9 unresolved discussions (waiting on @istae)

___
*[pkg/topology/pslice/pslice.go, line 1 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLIlJl6Hp_CGsFXAFL:-MiLIlJl6Hp_CGsFXAFM:bz05wjz) ([raw file](https://github.com/ethersphere/bee/blob/12081169ca5ef3be0fbbb9c9bde5a3f05066193a/pkg/topology/pslice/pslice.go#L1)):*
> ```Go
> // Copyright 2020 The Swarm Authors. All rights reserved.
> ```

remove copyright?
___
*[pkg/topology/pslice/pslice.go, line 14 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLIpeL5EEXFdlS68xL:-MiLIpeL5EEXFdlS68xM:bimnasd) ([raw file](https://github.com/ethersphere/bee/blob/12081169ca5ef3be0fbbb9c9bde5a3f05066193a/pkg/topology/pslice/pslice.go#L14)):*
> ```Go
> )
> 
> // PSlice maintains a list of addresses, indexing them by their different proximity orders.
> ```

why remove the godoc?
___
*[pkg/topology/pslice/pslice.go, line 23 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLIxc75JCDvlp97xl6:-MiLIxc75JCDvlp97xl7:bw6uilo) ([raw file](https://github.com/ethersphere/bee/blob/79126cbe7abf1eb47a7bd311a0256bfd40085ab5/pkg/topology/pslice/pslice.go#L23)):*
> ```Go
> }
> 
> func (s *PSlice) Add(addrs ...swarm.Address) {
> ```

add godoc
___
*[pkg/topology/pslice/pslice.go, line 25 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLIrtJDHXp6xYLFls_:-MiLIrtK9yL9-Vaq3gLh:b-a0dsqa) ([raw file](https://github.com/ethersphere/bee/blob/12081169ca5ef3be0fbbb9c9bde5a3f05066193a/pkg/topology/pslice/pslice.go#L25)):*
> ```Go
> }
> 
> // New creates a new PSlice.
> ```

why remove godoc?
___
*[pkg/topology/pslice/pslice.go, line 34 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLJ3sPE7XNiFXYsdJ_:-MiLJ3sPE7XNiFXYsdJa:b-zehcug) ([raw file](https://github.com/ethersphere/bee/blob/79126cbe7abf1eb47a7bd311a0256bfd40085ab5/pkg/topology/pslice/pslice.go#L34)):*
> ```Go
> 		}
> 
> 		s.peers[po] = append(s.peers[po], addr)
> ```

this is fairly unoptimized and would result in a lot of allocations for a lot of `Add`s. It might be beneficial to categorize the addresses by POs and grow every PO of the peers slice once per `Add` operation
___
*[pkg/topology/pslice/pslice.go, line 37 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLJmCnCtlFqExp-dT1:-MiLJmCnCtlFqExp-dT2:bvuxg32) ([raw file](https://github.com/ethersphere/bee/blob/12081169ca5ef3be0fbbb9c9bde5a3f05066193a/pkg/topology/pslice/pslice.go#L37)):*
> ```Go
> func (s *PSlice) EachBin(pf topology.EachPeerFunc) error {
> 	s.RLock()
> 	peers, bins := s.peers, s.bins
> ```

Copy-on-write is still needed here since the callback might take long to execute, blocking other consumers of PSlice from accessing the shared resource
___
*[pkg/topology/pslice/pslice.go, line 68 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLJuJWAMoEDGpeiMHA:-MiLJuJWAMoEDGpeiMHB:bb7od0r) ([raw file](https://github.com/ethersphere/bee/blob/12081169ca5ef3be0fbbb9c9bde5a3f05066193a/pkg/topology/pslice/pslice.go#L68)):*
> ```Go
> func (s *PSlice) EachBinRev(pf topology.EachPeerFunc) error {
> 	s.RLock()
> 	peers, bins := s.peers, s.bins
> ```

same
___
*[pkg/topology/pslice/pslice.go, line 130 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLK9kd7Met3aTyh6HG:-MiLK9kd7Met3aTyh6HH:bemqhb5) ([raw file](https://github.com/ethersphere/bee/blob/79126cbe7abf1eb47a7bd311a0256bfd40085ab5/pkg/topology/pslice/pslice.go#L130)):*
> ```Go
> 
> 	po := swarm.Proximity(s.baseBytes, peer)
> 	if int(po) >= len(s.peers) {
> ```

pretty sure that `swarm.Proximity` should do the same operation (rounding down to `swarm.MaxPO`)
___
*[pkg/topology/pslice/pslice.go, line 171 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLKVSpBTfs0_y9GVFk:-MiLKVSpBTfs0_y9GVFl:bi27zxq) ([raw file](https://github.com/ethersphere/bee/blob/79126cbe7abf1eb47a7bd311a0256bfd40085ab5/pkg/topology/pslice/pslice.go#L171)):*
> ```Go
> 
> 	// replace index with last element
> 	s.peers[po][i] = s.peers[po][len(s.peers[po])-1]
> ```

there's a way to [delete](https://github.com/golang/go/wiki/SliceTricks#delete) elements if a slice

istae (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2453)* status: 4 of 5 files reviewed, 9 unresolved discussions (waiting on @acud)

___
*[pkg/topology/pslice/pslice.go, line 1 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLIlJl6Hp_CGsFXAFL:-MiLnDe11OC45yslvBKL:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/12081169ca5ef3be0fbbb9c9bde5a3f05066193a/pkg/topology/pslice/pslice.go#L1)):*

remove copyright?
</blockquote></details>

Done.
___
*[pkg/topology/pslice/pslice.go, line 14 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLIpeL5EEXFdlS68xL:-MiLuYF3EKLd6ENoRBzQ:b-lc1pez) ([raw file](https://github.com/ethersphere/bee/blob/12081169ca5ef3be0fbbb9c9bde5a3f05066193a/pkg/topology/pslice/pslice.go#L14)):*

why remove the godoc?
</blockquote></details>

copy paste error
___
*[pkg/topology/pslice/pslice.go, line 23 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLIxc75JCDvlp97xl6:-MiLogqW8G_QLlyxKt_1:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/79126cbe7abf1eb47a7bd311a0256bfd40085ab5/pkg/topology/pslice/pslice.go#L23)):*

add godoc
</blockquote></details>

Done.
___
*[pkg/topology/pslice/pslice.go, line 25 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLIrtJDHXp6xYLFls_:-MiLuZQL2WmeNyVOlkuh:b-ip63jx) ([raw file](https://github.com/ethersphere/bee/blob/12081169ca5ef3be0fbbb9c9bde5a3f05066193a/pkg/topology/pslice/pslice.go#L25)):*
> copy paste error
copy paste error
___
*[pkg/topology/pslice/pslice.go, line 34 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLJ3sPE7XNiFXYsdJ_:-MiLuWrqCRoSGUQowpDP:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/79126cbe7abf1eb47a7bd311a0256bfd40085ab5/pkg/topology/pslice/pslice.go#L34)):*

this is fairly unoptimized and would result in a lot of allocations for a lot of `Add`s. It might be beneficial to categorize the addresses by POs and grow every PO of the peers slice once per `Add` operation
</blockquote></details>

Done.
___
*[pkg/topology/pslice/pslice.go, line 37 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLJmCnCtlFqExp-dT1:-MiLoexTALgzXnEgUiyB:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/12081169ca5ef3be0fbbb9c9bde5a3f05066193a/pkg/topology/pslice/pslice.go#L37)):*

Copy-on-write is still needed here since the callback might take long to execute, blocking other consumers of PSlice from accessing the shared resource
</blockquote></details>

Done.
___
*[pkg/topology/pslice/pslice.go, line 68 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLJuJWAMoEDGpeiMHA:-MiLofdi8hb9h_oXG7Bd:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/12081169ca5ef3be0fbbb9c9bde5a3f05066193a/pkg/topology/pslice/pslice.go#L68)):*

same
</blockquote></details>

Done.
___
*[pkg/topology/pslice/pslice.go, line 130 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLK9kd7Met3aTyh6HG:-MiLuGDe0TQN5CPJDqEu:b-aj9dv4) ([raw file](https://github.com/ethersphere/bee/blob/79126cbe7abf1eb47a7bd311a0256bfd40085ab5/pkg/topology/pslice/pslice.go#L130)):*

pretty sure that `swarm.Proximity` should do the same operation (rounding down to `swarm.MaxPO`)
</blockquote></details>

this is normally true, but some unit tests pass in a `maxBins` that is less than MaxPO
___
*[pkg/topology/pslice/pslice.go, line 171 at r2](https://reviewable.io/reviews/ethersphere/bee/2453#-MiLKVSpBTfs0_y9GVFk:-MiLo-pF2YM7On7hQaes:b-k3djti) ([raw file](https://github.com/ethersphere/bee/blob/79126cbe7abf1eb47a7bd311a0256bfd40085ab5/pkg/topology/pslice/pslice.go#L171)):*

there's a way to [delete](https://github.com/golang/go/wiki/SliceTricks#delete) elements if a slice
</blockquote></details>

Since order in the array does not matter, this trick removes indices with O(1) complexity. The method in the PR does not use `append` so it's very efficient.

pkg/topology/pslice/pslice.go
```diff
@@ -16,36 +16,66 @@ import (
...
-	peers, bins := s.peers, s.bins
-	s.RUnlock()
 
-	if len(peers) == 0 {
-		return nil
-	}
+	for i := s.maxBins - 1; i >= 0; i-- {
 
-	var binEnd = uint(len(peers))
+		s.RLock()
```
istae, line 73: @acud This seems the only way to prevent data races, unless I lock the whole function at the beginning.

istae:
unoptimized Add()
561639550 ns/op	 5390080 B/op	     253 allocs/op

optimized Add() (commit  45141cae8c51b105606b0b88e223fc4957b4773a)
564776350 ns/op	 3870976 B/op	      34 allocs/op

acud (changes requested):
> Reviewed 1 of 4 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2453)* status: all files reviewed, 4 unresolved discussions (waiting on @acud, @istae, and @janos)

___
*[pkg/topology/pslice/pslice.go, line 73 at r3](https://reviewable.io/reviews/ethersphere/bee/2453#-Mi2N6eQ0m1BIJUjd0lY-r3-73:-MiQx8lg78h5svHZYRTi:b31wtb4) ([raw file](https://github.com/ethersphere/bee/blob/55d4eff02e1721ae2dec2006b98bbc2c85f3dfc0/pkg/topology/pslice/pslice.go#L73)):*

@acud This seems the only way to prevent data races, unless I lock the whole function at the beginning.
</blockquote></details>

not the only way :) the other way would be to create a deep copy at the beginning of the function:
the old COW was simple since there was only one slice involved (the peers slice). if the peers slice grow, the slice pointer would change (the copy would still use the old pointer).
since the new slice is a slice of slices, then copying the first slice is useless since the slice instances of the second dimension would have the same address, while their underlying slice pointers would be changing every time the arrays would be reallocated.
so the second solution is to create a new slice of slices, and do an assignment of every one of the cells like:
```go
s.Rlock()
pcopy := make([][]swarm.Address,s.maxBins)
for i,_ := range pcopy { pcopy[i] = s.peers[i] }
s.RUnlock()
```
___
*[pkg/topology/pslice/pslice.go, line 120 at r3](https://reviewable.io/reviews/ethersphere/bee/2453#-MiQza4i6RwMV_V8oVg4:-MiQza4i6RwMV_V8oVg5:b-nv133d) ([raw file](https://github.com/ethersphere/bee/blob/55d4eff02e1721ae2dec2006b98bbc2c85f3dfc0/pkg/topology/pslice/pslice.go#L120)):*
> ```Go
> }
> 
> // BinPeers returns the slice of peers in a certain bin
> ```

period at end of sentence (godoc). same for all

istae (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2453)* status: 3 of 5 files reviewed, 3 unresolved discussions (waiting on @acud and @janos)

___
*[pkg/topology/pslice/pslice.go, line 73 at r3](https://reviewable.io/reviews/ethersphere/bee/2453#-Mi2N6eQ0m1BIJUjd0lY-r3-73:-MiSL3JCDxOO7D9gCaK9:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/55d4eff02e1721ae2dec2006b98bbc2c85f3dfc0/pkg/topology/pslice/pslice.go#L73)):*

not the only way :) the other way would be to create a deep copy at the beginning of the function:
the old COW was simple since there was only one slice involved (the peers slice). if the peers slice grow, the slice pointer would change (the copy would still use the old pointer).
since the new slice is a slice of slices, then copying the first slice is useless since the slice instances of the second dimension would have the same address, while their underlying slice pointers would be changing every time the arrays would be reallocated.
so the second solution is to create a new slice of slices, and do an assignment of every one of the cells like:
```go
s.Rlock()
pcopy := make([][]swarm.Address,s.maxBins)
for i,_ := range pcopy { pcopy[i] = s.peers[i] }
s.RUnlock()
```

</blockquote></details>

Done.
___
*[pkg/topology/pslice/pslice.go, line 120 at r3](https://reviewable.io/reviews/ethersphere/bee/2453#-MiQza4i6RwMV_V8oVg4:-MiSL4TiEJpI51uxXZVV:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/55d4eff02e1721ae2dec2006b98bbc2c85f3dfc0/pkg/topology/pslice/pslice.go#L120)):*

period at end of sentence (godoc). same for all
</blockquote></details>

Done.

janos:
> image
> 
> The benchmark tests with 2d suffix is the new implementation.

I am curious what would the benchmarks with the most recent code would produce. I am running:

```
go test -v -run none -bench . -benchmem ./pkg/topology/pslice
```

and getting:

```
goos: darwin
goarch: amd64
pkg: github.com/ethersphere/bee/pkg/topology/pslice
cpu: Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
Benchmark2DAdd
Benchmark2DAdd-8                   10000           2232472 ns/op         3967503 B/op        160 allocs/op
Benchmark2DAddBatch
Benchmark2DAddBatch-8                100         866211958 ns/op          239134 B/op          3 allocs/op
Benchmark2DAddReverse
Benchmark2DAddReverse-8            10000           2299042 ns/op         3967503 B/op        160 allocs/op
Benchmark2DRemove
Benchmark2DRemove-8                10000           1593827 ns/op         3957420 B/op         32 allocs/op
Benchmark2DRemoveReverse
Benchmark2DRemoveReverse-8         10000           1603883 ns/op         3957420 B/op         32 allocs/op
PASS
```

Where is no benchmarks with the old implementation to compare.

And the number of allocations raised from 32 to 160.

I am afraid that there is too much copying of the slices in this implementation and that benefits of CoW are removed, without addressing the issues that @acud already raised or addressing them by adding allocations.

I would also like to see the benchmarks of EachBin and EachBinRev functions as they are most frequently used and they benefit the most from CoW. I think that they now allocate much more than before.

I am suggesting to try to incorporate the new 2d slice with the copy on write approach.

istae:
In the codebase, the most general use case of `Add()` is with a single address so I've included in the latest commit an optimization that bypasses unnecessary allocations when addrs count is 1. 

```
goos: linux
goarch: amd64
pkg: github.com/ethersphere/bee/pkg/topology/pslice
Benchmark2DAdd
Benchmark2DAdd-4                   10000            409571 ns/op            4458 B/op         32 allocs/op
Benchmark2DAddBatch
Benchmark2DAddBatch-4                100         553828299 ns/op          239134 B/op          3 allocs/op
Benchmark2DAddReverse
Benchmark2DAddReverse-4            10000            407857 ns/op            4458 B/op         32 allocs/op
Benchmark2DRemove
Benchmark2DRemove-4                10000           1261347 ns/op         3957421 B/op         32 allocs/op
Benchmark2DRemoveReverse
Benchmark2DRemoveReverse-4         10000           1276244 ns/op         3957421 B/op         32 allocs/op
```

janos:
Thanks @istae for the optimization on case when addrs count is 1 for Add. It is a very specific optimization, in general, allocations on Add can be higher.

But, even with more or less optimized Add method, the most frequently used EachBin and EachBinRev are not benchmarked. It would be very good to see if they are more performant.

istae:
Pslice 2D
BenchmarkEachBin-4         15338             78406 ns/op             768 B/op          1 allocs/op

Pslice (master)
BenchmarkEachBin-4         15232             78670 ns/op               0 B/op          0 allocs/op

pkg/topology/pslice/pslice_test.go
```diff
@@ -422,22 +421,102 @@ func chkNotExists(t *testing.T, ps *pslice.PSlice, addrs ...swarm.Address) {
...
-	var (
-		base = test.RandomAddress()
-		ps   = pslice.New(16, base)
-	)
+	ps := pslice.New(bins, base)
 
-	for i := 0; i < 16; i++ {
-		for j := 0; j < 300; j++ {
+	for i := 0; i < bins; i++ {
+		for j := 0; j < b.N; j++ {
```
janos, line 433: This is not correct way to write benchmarks. The benchmark iteration is under another loop.

pkg/topology/pslice/pslice_test.go
```diff
@@ -422,22 +421,102 @@ func chkNotExists(t *testing.T, ps *pslice.PSlice, addrs ...swarm.Address) {
...
+	for j := 0; j < b.N; j++ {
+		ps.Add(addrs...)
+	}
+}
+
+func BenchmarkAddReverse(b *testing.B) {
+	ps := pslice.New(bins, base)
+
+	for i := bins - 1; i >= 0; i-- {
+		for j := 0; j < b.N; j++ {
```
janos, line 459: Same here.

pkg/topology/pslice/pslice_test.go
```diff
@@ -422,22 +421,102 @@ func chkNotExists(t *testing.T, ps *pslice.PSlice, addrs ...swarm.Address) {
...
+	}
+}
+
+func BenchmarkRemove(b *testing.B) {
+	ps := pslice.New(bins, base)
+
+	var addrs []swarm.Address
+
+	for i := 0; i < bins; i++ {
+		for j := 0; j < b.N; j++ {
```
janos, line 471: This is not correct benchmark. The `j` limit should not be b.N.

pkg/topology/pslice/pslice_test.go
```diff
@@ -422,22 +421,102 @@ func chkNotExists(t *testing.T, ps *pslice.PSlice, addrs ...swarm.Address) {
...
+			addrs = append(addrs, addr)
+			ps.Add(addr)
+		}
+	}
+
+	b.ResetTimer()
+
+	for _, addr := range addrs {
+		ps.Remove(addr)
+	}
```
janos, line 482: This is adding only one address, without iterating for measuements.

pkg/topology/pslice/pslice_test.go
```diff
@@ -422,22 +421,102 @@ func chkNotExists(t *testing.T, ps *pslice.PSlice, addrs ...swarm.Address) {
...
+
+}
+
+func BenchmarkRemoveReverse(b *testing.B) {
+	ps := pslice.New(bins, base)
+
+	var addrs []swarm.Address
+
+	for i := bins - 1; i >= 0; i-- {
+		for j := 0; j < b.N; j++ {
```
janos, line 492: Same here.

pkg/topology/pslice/pslice_test.go
```diff
@@ -422,22 +421,102 @@ func chkNotExists(t *testing.T, ps *pslice.PSlice, addrs ...swarm.Address) {
...
+			addrs = append(addrs, addr)
+			ps.Add(addr)
+		}
+	}
+
+	b.ResetTimer()
+
+	for _, addr := range addrs {
+		ps.Remove(addr)
+	}
```
janos, line 503: Same here.

janos (commented):
I have noticed some important issues with the current benchmarks which make results hard to interpret and compare impact of the changes to different functions.

In essence, the main question if the optimization is useful is whether the reduction of allocations on `Add` and `Remove` will compensate additional allocations on `EachBin`.

istae:
@janos The tests should now be fixed.  Thanks for pointing it out :)

New results:
```
BenchmarkAdd-4                        28          39338543 ns/op           56155 B/op         12 allocs/op
BenchmarkAddBatch-4                 5324            220597 ns/op            2309 B/op          3 allocs/op
BenchmarkAddReverse-4                 30          39104950 ns/op           52411 B/op         11 allocs/op
BenchmarkRemove-4                     15          76943993 ns/op               0 B/op          0 allocs/op
BenchmarkRemoveReverse-4             696           1512997 ns/op          580726 B/op         45 allocs/op
BenchmarkEachBin-4                 15616             76526 ns/op             768 B/op          1 allocs/op
BenchmarkAddOld-4                      1        5181641400 ns/op        12423398016 B/op           64151 allocs/op
BenchmarkAddBatchOld-4            108356             11283 ns/op           49408 B/op          2 allocs/op
BenchmarkAddReverseOld-4               1        6092022100 ns/op        12423417952 B/op           64366 allocs/op
BenchmarkRemoveOld-4                   1        2857826500 ns/op               0 B/op          0 allocs/op
BenchmarkRemoveReverseOld-4            1        5258139000 ns/op        12423407616 B/op           64251 allocs/op
BenchmarkEachBinOld-4              15759             77433 ns/op               0 B/op          0 allocs/op
PASS
```

istae:
The single allocation in the new iterators should not be an issue. The bottleneck there is calling the callback function for each address. The signification improvement in Add and Remove should compensate for this.

janos:
> The single allocation in the new iterators should not be an issue. The bottleneck there is calling the callback function for each address. The signification improvement in Add and Remove should compensate for this.

It might, it is a guess as we do not have the information about frequency calls of these functions are. In general, this change can be tested in integration. I am of the opinion that CoW is better than CoR, but that has to be checked with profiling. I assume that this change was inspired by some profiling results, and that the same can be performed to validate the impact in the application by these changes.

acud:
> I am of the opinion that CoW is better than CoR

@janos, I might be getting it wrong, but AFAIK we are still doing CoW. The thing is that we need a small allocation on reads, because we need to copy the slice metadata that points to the second dimensions that actually store the peers on every PO. It is not a deep copy but only a shallow copy of the 1st level of the array. The actual copy and resize of the slice of the peers is done still only on write. Part of the motivation here was that we need to copy and shift the whole thing when inserting items (copy all references to all addresses), while when we separate to POs, we only need to reallocate the actual POs which are modified. Again, I might be getting it wrong, so just correct me :)

janos:
@acud yes that is correct what you wrote. To be more clear what I wanted to say is that avoiding any CoR is better than having it even slightly. But as I wrote, this is up to the profiling to show, whether reducing allocations on write and adding small allocations on read is solving the observed performance problem. Benchmarks on its own can validate improvements if changes do not introduce some, even slight, regressions. In this case, there is a regression on a frequently used functions and benchmarks cannot show the sum of all impacts. I am encouraging profiling or metrics to validate the improvements, well firstly to validate the bottleneck and that the improvement can significantly improve performance or reduce resource utilization. If this task is chosen without profiling first, then it can be merged without profile validation and just checked if there is a significant resource utilization change on the next release. But, in general, I would not suggest spending time on optimizations of the code that is not identified as the bottleneck.

acud:
I see @janos. Then we can revert a step back to not having the extra allocation of the whole slice on read, but just keeping the reference of the slice when inside the for loop within the iterators (as it was prior to the shallow copy).

acud:
can we have another round of benchmarks @istae

istae:
```
goos: linux
goarch: amd64
pkg: github.com/ethersphere/bee/pkg/topology/pslice
BenchmarkAdd-4                                 3         462245300 ns/op         1085957 B/op         45 allocs/op
BenchmarkAddBatch-4                          100         454527320 ns/op           73703 B/op          3 allocs/op
BenchmarkRemove-4                              1        1002807400 ns/op               0 B/op          0 allocs/op
BenchmarkEachBin-4                         15380             78251 ns/op               0 B/op          0 allocs/op
BenchmarkEachBinParallel-4                 60043             19726 ns/op               0 B/op          0 allocs/op
BenchmarkEachBinRev-4                      15357             78474 ns/op               0 B/op          0 allocs/op
BenchmarkAddOld-4                              1        6029349400 ns/op        12423406200 B/op           64207 allocs/op
BenchmarkAddBatchOld-4                         1        1635543100 ns/op          770304 B/op          2 allocs/op
BenchmarkRemoveOld-4                           1        3217701400 ns/op               0 B/op          0 allocs/op
BenchmarkEachBinOld-4                      14781             80953 ns/op               0 B/op          0 allocs/op
BenchmarkEachBinParallelOld-4              60906             19779 ns/op               0 B/op          0 allocs/op
BenchmarkEachBinRevOld-4                   14800             81065 ns/op               0 B/op          0 allocs/op
```

istae:
```
BenchmarkAdd-4                         3         462643933 ns/op         1085925 B/op         45 allocs/op
BenchmarkAddBatch-4                  100         454821119 ns/op           73703 B/op          3 allocs/op
BenchmarkRemove-4                      1        1003735200 ns/op               0 B/op          0 allocs/op
BenchmarkEachBin-4                 15319             77850 ns/op               0 B/op          0 allocs/op
BenchmarkEachBinRev-4              15283             78512 ns/op               0 B/op          0 allocs/op
BenchmarkAddOld-4                      1        6339416400 ns/op        12423401160 B/op           64170 allocs/op
BenchmarkAddBatchOld-4                 1        1624457500 ns/op          770304 B/op          2 allocs/op
BenchmarkRemoveOld-4                   1        3176552300 ns/op               0 B/op          0 allocs/op
BenchmarkEachBinOld-4              14847             81066 ns/op               0 B/op          0 allocs/op
BenchmarkEachBinRevOld-4           14781             81100 ns/op               0 B/op          0 allocs/op
```

acud (approved):
LGTM! Thanks for the patience @istae

janos (commented):
Thanks @istae for improvements and addressing my comments.

I think that, it is only needed to cleanup the PR from old code to be merged.

janos (approved):
Great, thanks @istae.

PR #2456, opened 2021-08-30
fix: build custom beekeeper

vandot:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2456)

To build, install and use custom beekeeper branch you can set env variable globally or set it in make command
```
jobs:
  beekeeper:
    name: Infra tests
    env:
      REPLICA: 3
      RUN_TYPE: "PR RUN"
      SETUP_CONTRACT_IMAGE_TAG: "0.2.0"
      BEEKEEPER_BRANCH: "CUSTOM-BRANCH"
```

`Line 51` of `beekeeper.yaml`
```
make beekeeper BEEKEEPER_INSTALL_DIR=/usr/local/bin BEEKEEPER_USE_SUDO=true BEEKEEPER_BRANCH=CUSTOM-BRANCH
```

### DIFF
--- a/.github/workflows/beekeeper.yml
+++ b/.github/workflows/beekeeper.yml
@@ -14,7 +14,6 @@ jobs:
       REPLICA: 3
       RUN_TYPE: "PR RUN"
       SETUP_CONTRACT_IMAGE_TAG: "0.2.0"
-      BEEKEEPER_BRANCH: "auth-health-check"
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/.github/workflows/beekeeper.yml
+++ b/.github/workflows/beekeeper.yml
@@ -14,6 +14,7 @@ jobs:
       REPLICA: 3
       RUN_TYPE: "PR RUN"
       SETUP_CONTRACT_IMAGE_TAG: "0.2.0"
+      BEEKEEPER_BRANCH: "master"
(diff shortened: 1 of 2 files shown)


.github/workflows/beekeeper.yml
```diff
@@ -14,6 +14,7 @@ jobs:
       REPLICA: 3
       RUN_TYPE: "PR RUN"
       SETUP_CONTRACT_IMAGE_TAG: "0.2.0"
+      BEEKEEPER_BRANCH: "auth-health-check"
```
notanatol, line 17: should we use `master` as default value?

vandot: I actually reverted it now, this was only to test if it works.
Not bad idea to have a placeholder so that you just change it there.

.github/workflows/beekeeper.yml
```diff
@@ -14,6 +14,7 @@ jobs:
       REPLICA: 3
       RUN_TYPE: "PR RUN"
       SETUP_CONTRACT_IMAGE_TAG: "0.2.0"
+      BEEKEEPER_BRANCH: "master"
```
vandot, line 17: @notanatol like this?

notanatol (approved):

acud (approved):
Great. Thanks!

PR #2458, opened 2021-08-30
fix: use correct prefix for ens entries

aloknerurkar:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2458)

### DIFF
--- a/pkg/resolver/client/ens/ens.go
+++ b/pkg/resolver/client/ens/ens.go
@@ -20,7 +20,7 @@ import (
 
 const (
 	defaultENSContractAddress = "00000000000C2E074eC69A0dFb2997BA6C7d2e1e"
-	swarmContentHashPrefix    = "/swarm/"
+	swarmContentHashPrefix    = "bzz://"
 )
 
 // Address is the swarm bzz address.
--- a/pkg/resolver/client/ens/ens.go
+++ b/pkg/resolver/client/ens/ens.go
@@ -120,7 +120,7 @@ func (c *Client) Resolve(name string) (Address, error) {
 	}
 
 	// Ensure that the content hash string is in a valid format, eg.
-	// "/swarm/<address>".
+	// "bzz://<address>".
 	if !strings.HasPrefix(hash, swarmContentHashPrefix) {
 		return swarm.ZeroAddress, fmt.Errorf("contenthash %s: %w", hash, ErrInvalidContentHash)
 	}
--- a/pkg/resolver/client/ens/ens.go
+++ b/pkg/resolver/client/ens/ens.go
@@ -163,7 +163,7 @@ func wrapDial(endpoint, contractAddr string) (*ethclient.Client, *goens.Registry
 	return ethCl, registry, nil
 }
 
-func wrapResolve(registry *goens.Registry, addr common.Address, name string) (string, error) {
+func wrapResolve(registry *goens.Registry, _ common.Address, name string) (string, error) {
 	// Ensure the name is registered.
 	ownerAddress, err := registry.Owner(name)
 	if err != nil {


agazso:
It would be good to add an integration test if possible to catch when this is changed again in the dependency.

aloknerurkar:
@agazso It would be difficult to manage the test as it would depend on an existing ENS entry to be present. Not sure if we want that.
@acud

acud:
@agazso the spec hasn't changed in [three years](https://eips.ethereum.org/EIPS/eip-1577) so I don't see a risk of this happening anytime soon. Adding an integration test here is also not so simple as it would involve using a production ENS with ethereum backend and would either need to do real ens updates or perform hardcoded checks to see whether a given ens entry resolves to some static value that should never change. The latter would be relatively simple, but would also depend on either exposing ENS resolution as a separate endpoint in `bee` to make sure that it resolves an entry correctly, or just testing the ens library we use separately. Not sure if this is time well spent

acud (approved):

ldeffenb:
> @agazso the spec hasn't changed in [three years](https://eips.ethereum.org/EIPS/eip-1577) so I don't see a risk of this happening anytime soon. 

Just FYI, there is this (hopefully) coming down the ENS pipeline that may impact the resolver library.  It will give more cost-effective, xdai-native, ENS update capability.

https://medium.com/the-ethereum-name-service/a-general-purpose-bridge-for-ethereum-layer-2s-e28810ec1d88

agazso:
> @agazso the spec hasn't changed in [three years](https://eips.ethereum.org/EIPS/eip-1577) so I don't see a risk of this happening anytime soon. 

It's true that the spec hasn't changed but the dependency Bee uses changed two months ago and that completely broke the ENS functionality. This is a regression and this is why I propose to have some testing facility that would avoid this in the future.

> Adding an integration test here is also not so simple as it would involve using a production ENS with ethereum backend and would either need to do real ens updates or perform hardcoded checks to see whether a given ens entry resolves to some static value that should never change. The latter would be relatively simple, but would also depend on either exposing ENS resolution as a separate endpoint in `bee` to make sure that it resolves an entry correctly, or just testing the ens library we use separately. Not sure if this is time well spent

It would be good if adding integration tests like this were relatively simple and cheap, because I see more integration being built in Bee, not less. Then there will be cross-dependencies, like in this case, where it seems `go-ens` was bumped because `go-ethereum` was bumped as well. So the number of things that can possibly break in the future will increase and one of the most efficient way to avoid these issues is automated testing.

acud:
Yeah it's a fair point. I also inspected the commit that changed the `go-ens` library. In fact, it was never implemented according to the EIP standard until its recent update. Also, their version bump was a minor version bump and not a major. I have commented to the maintainer about this. Having said that I'm not entirely sure how to test this. Doing an integration test separately does not guarantee that we catch the regression (since the dependency would be vendored externally) and exposing the actual ENS resolution through bee implies more "features" to maintain. We could consider exposing this as a debug api functionality and just do an integration test through that API endpoint to see that the resolution is correct.

cc @janos would be happy for your input here. 

also, another idea would be to ditch the library and just do the decoding ourselves. then just a unit test will do (the library is vendored to do 10 lines of code which unmarshal the multicodec to a swarm hash - definitely not rocket science)

pkg/resolver/client/ens/ens.go
```diff
@@ -20,7 +20,7 @@ import (
 
 const (
 	defaultENSContractAddress = "00000000000C2E074eC69A0dFb2997BA6C7d2e1e"
-	swarmContentHashPrefix    = "/swarm/"
+	swarmContentHashPrefix    = "bzz://"
```
acud, line 23: comment on `L123` needs to be amended

aloknerurkar:
@agazso So we actually have an integration test in the codebase which was not enabled in the CI. The test uses the endpoint `https://goerli.infura.io/v3/59d83a5a4be74f86b9851190c802297b` and checks for some entries which were probably created before. But this endpoint no longer works.

We have decided that this change can go in as-is and then we will work on enabling this integration test in CI.

mrekucci (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2458)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar and @vandot)

acud (approved):

PR #2460, opened 2021-08-31
fix: replace readAll with limit readers

notanatol:
**SWA-01-005 WP2: Usage of ioutil.ReadAll in client API can result in DoS** _(Low)_

During a source code review of the bee repository, it was spotted that various client API
handler routines are using  ioutil.ReadAll  for reading user-input transmitted as part of
HTTP requests within the body. The usage of ioutil.ReadAll is dangerous2 and can result
in a DoS situation as this function continues to read data into a buffer allocated on the
system heap until it has received EOF. A malicious user could leverage this behavior by
sending HTTP requests to a bee node in order to cause an out-of-memory situation.

This is closed due to the fact that all the http routes that receive content are wrapped in a `NewMaxBodyBytesHandler` that makes sure that the content size is limited.

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2460)

### DIFF
--- a/pkg/api/chunk.go
+++ b/pkg/api/chunk.go
@@ -10,7 +10,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"strings"
 
@@ -88,8 +87,13 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	data, err := ioutil.ReadAll(r.Body)
-	if err != nil {
+	reader := io.LimitReader(r.Body, swarm.ChunkSize+swarm.SpanSize)
+
+	defer r.Body.Close()
+
+	var buf bytes.Buffer
+
+	if _, err := io.Copy(&buf, reader); err != nil {
 		if jsonhttp.HandleBodyReadError(err, w) {
 			return
 		}
@@ -98,6 +102,8 @@ func (s *server) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 		jsonhttp.InternalServerError(w, "cannot read chunk data")
 		return
 	}
+
+	data := buf.Bytes()
 
 	if len(data) < swarm.SpanSize {
 		s.logger.Debug("chunk upload: not enough data")
--- a/pkg/api/pss.go
+++ b/pkg/api/pss.go
@@ -5,11 +5,12 @@
 package api
 
 import (
+	"bytes"
 	"context"
 	"crypto/ecdsa"
 	"encoding/hex"
 	"errors"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"strings"
 	"time"
@@ -71,8 +72,13 @@ func (s *server) pssPostHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	payload, err := ioutil.ReadAll(r.Body)
-	if err != nil {
+	reader := io.LimitReader(r.Body, swarm.ChunkSize+swarm.SpanSize)
+
+	defer r.Body.Close()
+
+	var buf bytes.Buffer
+
+	if _, err := io.Copy(&buf, reader); err != nil {
 		s.logger.Debugf("pss read payload: %v", err)
 		s.logger.Error("pss read payload")
 		jsonhttp.InternalServerError(w, nil)
@@ -101,7 +107,7 @@ func (s *server) pssPostHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	stamper := postage.NewStamper(i, s.signer)
 
-	err = s.pss.Send(r.Context(), topic, payload, stamper, recipient, targets)
+	err = s.pss.Send(r.Context(), topic, buf.Bytes(), stamper, recipient, targets)
 	if err != nil {
 		s.logger.Debugf("pss send payload: %v. topic: %s", err, topicVar)
 		s.logger.Error("pss send payload")
--- a/pkg/api/soc.go
+++ b/pkg/api/soc.go
@@ -5,9 +5,10 @@
 package api
 
 import (
+	"bytes"
 	"encoding/hex"
 	"errors"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"strings"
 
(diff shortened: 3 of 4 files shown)


mrekucci (commented):
> Reviewed 3 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2460)* status: all files reviewed, 2 unresolved discussions (waiting on @notanatol)

___
*[pkg/api/chunk.go, line 90 at r2](https://reviewable.io/reviews/ethersphere/bee/2460#-MiRQ_u4AIEGn7K7czgc:-MiRQ_u4AIEGn7K7czgd:bdof0an) ([raw file](https://github.com/ethersphere/bee/blob/1cde0eb83f4b1c8f3162a0e1a8d36f689cc3e135/pkg/api/chunk.go#L90)):*
> ```Go
> 	}
> 
> 	reader := io.LimitReader(r.Body, swarm.ChunkSize+swarm.SpanSize)
> ```

Nice optimization. I'd suggest to use `io.ReadFull(r.Body, payload)`with an allocated buffer `payload := make([]byte, swarm.ChunkSize+swarm.SpanSize)` instead.
___
*[pkg/api/soc.go, line 61 at r2](https://reviewable.io/reviews/ethersphere/bee/2460#-MiRQL-WDFeJE-pR34Ns:-MiRQL-WDFeJE-pR34Nt:b-17qjie) ([raw file](https://github.com/ethersphere/bee/blob/1cde0eb83f4b1c8f3162a0e1a8d36f689cc3e135/pkg/api/soc.go#L61)):*
> ```Go
> 	}
> 
> 	reader := io.LimitReader(r.Body, swarm.ChunkSize+swarm.SpanSize)
> ```

Same here, I'd consider using `io.ReadFull(r.Body, payload)`with an allocated buffer `payload := make([]byte, swarm.ChunkSize+swarm.SpanSize)` instead.

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2460)* status: all files reviewed, 2 unresolved discussions (waiting on @acud and @notanatol)

___
*[pkg/api/chunk.go, line 90 at r2](https://reviewable.io/reviews/ethersphere/bee/2460#-MiRQ_u4AIEGn7K7czgc:-MiUmA8N9tW2wKOSvlv1:btbsn79) ([raw file](https://github.com/ethersphere/bee/blob/1cde0eb83f4b1c8f3162a0e1a8d36f689cc3e135/pkg/api/chunk.go#L90)):*

Nice optimization. I'd suggest to use `io.ReadFull(r.Body, payload)`with an allocated buffer `payload := make([]byte, swarm.ChunkSize+swarm.SpanSize)` instead.
</blockquote></details>

good point, but I wander if in our case we want to allow for reading of "less than" the size of the buffer.

`io.ReadAll` will return an error if fewer bytes were read - @acud would this behaviour fit here?

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2460)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @acud)

___
*[pkg/api/chunk.go, line 90 at r2](https://reviewable.io/reviews/ethersphere/bee/2460#-MiRQ_u4AIEGn7K7czgc:-MiWRJFe7Q9ZH_fznkJX:bwdyqcl) ([raw file](https://github.com/ethersphere/bee/blob/1cde0eb83f4b1c8f3162a0e1a8d36f689cc3e135/pkg/api/chunk.go#L90)):*

good point, but I wander if in our case we want to allow for reading of "less than" the size of the buffer.

`io.ReadAll` will return an error if fewer bytes were read - @acud would this behaviour fit here? 
</blockquote></details>

*correction: I meant that `io.ReadFull` will error if the payload is smaller than allocated buffer.

acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2460)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @mrekucci and @notanatol)

___
*[pkg/api/chunk.go, line 90 at r2](https://reviewable.io/reviews/ethersphere/bee/2460#-MiRQ_u4AIEGn7K7czgc:-MiWvG5T4mRE34w4oUbO:b-68n06q) ([raw file](https://github.com/ethersphere/bee/blob/1cde0eb83f4b1c8f3162a0e1a8d36f689cc3e135/pkg/api/chunk.go#L90)):*

*correction: I meant that `io.ReadFull` will error if the payload is smaller than allocated buffer.
</blockquote></details>

we need to support reading of less than a full chunk definitely
___
*[pkg/api/pss.go, line 75 at r3](https://reviewable.io/reviews/ethersphere/bee/2460#-MiWvVqYBGXIJAzStSUm:-MiWvVqYBGXIJAzStSUn:b4dfeov) ([raw file](https://github.com/ethersphere/bee/blob/c2a7a6bd759671652a7cb30773a6530e80fc5a17/pkg/api/pss.go#L75)):*
> ```Go
> 	}
> 
> 	reader := io.LimitReader(r.Body, swarm.ChunkSize+swarm.SpanSize)
> ```

since there's some repetition (also the static analyzer is complaining), maybe it would be good to extract the part between here and the Copy to a separate function?

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2460)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @acud)

___
*[pkg/api/pss.go, line 75 at r3](https://reviewable.io/reviews/ethersphere/bee/2460#-MiWvVqYBGXIJAzStSUm:-MiWwmoy6Sy5XJozyjk4:b-soz29b) ([raw file](https://github.com/ethersphere/bee/blob/c2a7a6bd759671652a7cb30773a6530e80fc5a17/pkg/api/pss.go#L75)):*

since there's some repetition (also the static analyzer is complaining), maybe it would be good to extract the part between here and the Copy to a separate function?
</blockquote></details>

I need to confirm it first but it seems that we have a router middleware that already takes care of this (ReadMaxBytes).
If this is the case I will close this PR altogether...

notanatol:
All the http routes that receive content are wrapped in a `NewMaxBodyBytesHandler` that makes sure that the content size is limited.

PR #2462, opened 2021-08-31
feat(libp2p): add custom user agent and log it on connection

janos:
This PR add a custom User Agent string to libp2p host which is communicated between peers over the internal libp2p Identify protocol. On every connection, the user agent is logged if it exists.

The current proposal for the user agent is in this form: `bee/<version> <goversion> <GOOS>/<GOARCH>`. For example `bee/1.1.1-dev go1.15.14 darwin/amd64`. It is debatable if all of these information should be included. It is encouraged to give your opinions on this subject.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2462)

### DIFF
--- a/pkg/p2p/libp2p/connections_test.go
+++ b/pkg/p2p/libp2p/connections_test.go
@@ -875,10 +875,11 @@ func testUserAgentLogLine(t *testing.T, logs bytes.Buffer, substring string) {
 					t.Errorf("log line %q does not contain an expected User Agent %q", line, wantUserAgent)
 				}
 			}
+			t.Log(line)
 		}
 	}
 	if !foundLogLine {
-		t.Errorf("log line with %s string was not found", logLineMarker)
+		t.Errorf("log line with %q string was not found", logLineMarker)
 	}
 }
 
--- a/pkg/p2p/libp2p/connections_test.go
+++ b/pkg/p2p/libp2p/connections_test.go
@@ -828,13 +828,13 @@ func TestUserAgentLogging(t *testing.T) {
 	s1Logs := new(buffer)
 	s2Logs := new(buffer)
 
-	s1, overlay1 := newService(t, 1, libp2pServiceOpts{
+	s1, _ := newService(t, 1, libp2pServiceOpts{
 		libp2pOpts: libp2p.Options{
 			FullNode: true,
 		},
 		Logger: logging.New(s1Logs, 5),
 	})
-	s2, overlay2 := newService(t, 1, libp2pServiceOpts{
+	s2, _ := newService(t, 1, libp2pServiceOpts{
 		Logger: logging.New(s2Logs, 5),
 	})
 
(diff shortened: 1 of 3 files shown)


acud (approved):
Very cool @janos. Nicely done. I would add that having a reverse parser that could already split the user agent and increment a new metric that would track different versions by labels would be very helpful and would remove the need to go through the logs manually.

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2462#-:-MiVL9BS0JReLUAHBmfX:b2sfa3d)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aloknerurkar)

aloknerurkar (approved):

mrekucci (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2462)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @janos)

___
*[pkg/p2p/libp2p/libp2p.go, line 892 at r1](https://reviewable.io/reviews/ethersphere/bee/2462#-MiVTjBEBMfDRpqvCLgB:-MiVTjBF2ViODG_VkllJ:bwrn6eb) ([raw file](https://github.com/ethersphere/bee/blob/144521ba38b51e05f84fd21c272a71571bf1c9a0/pkg/p2p/libp2p/libp2p.go#L892)):*
> ```Go
> 		return ""
> 	}
> 	if v == nil {
> ```

This check can be avoided since the `ok` value from type assertion covers that.

janos:
> Very cool @janos. Nicely done. I would add that having a reverse parser that could already split the user agent and increment a new metric that would track different versions by labels would be very helpful and would remove the need to go through the logs manually.
> 
> > Reviewed 3 of 3 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2462#-:-MiVL9BS0JReLUAHBmfX:b2sfa3d)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aloknerurkar)

Current format of the User Agent string can be parsed, but the question is should it be in some more formal format so that other implementations can use it? Or to stick with such more or less similar approach as http user agent strings are created? I believe that parsing http browser user agents is an art of its own because of the lack of format from the beginning. I am totally fine with informal format, as for example json-encoded user agent does not look user friendly. Maybe just to separate sections with `;` like `bee/<version>; <goversion>; <GOOS>/<GOARCH>`, or in general `<client>/<version>; <language>; <os>/<arch>`?

ldeffenb:
> Very cool @janos. Nicely done. I would add that having a reverse parser that could already split the user agent and increment a new metric that would track different versions by labels would be very helpful and would remove the need to go through the logs manually.

As long as the added keyed metric is constrained to only recognizable item, maybe it would be a good idea.  However, it does expose the node to a DOS attack if someone were to start generating random, but "good" looking User Agent strings.   The metrics page could grow quite large in a short span of time.

acud:
@janos I think that the user agent in its current form is fine, as it is in line with other standards, like [RFC7231](https://datatracker.ietf.org/doc/html/rfc7231#section-5.5.3) or [RFC1945](http://www.ietf.org/rfc/rfc1945.txt). I would just think about @ldeffenb's proposal that we log the version string only for known versions. However this would be difficult for dev versions (with commit version). Maybe we could have catchalls per version, like `v1.1.0-aabbcc` falls under `v1.1.0-dev`

janos:
That is a good point @ldeffenb. The metrics aggregation should be considered about that, but I think that it is up to the aggregation to choose which information to store and which to discard (in this case would be a long tail in the histogram of user agent counters). The user agent string cannot be trusted at all, it can just provide some insight by assuming that most of the nodes are using the correct one.

PR #2464, opened 2021-09-01
fix: read correct underlay

notanatol:


pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -144,7 +144,7 @@ func (s *Service) Handshake(ctx context.Context, stream p2p.Stream, peerMultiadd
 		return nil, fmt.Errorf("read synack message: %w", err)
 	}
 
-	observedUnderlay, err := ma.NewMultiaddrBytes(resp.Syn.ObservedUnderlay)
+	observedUnderlay, err := ma.NewMultiaddrBytes(fullRemoteMABytes)
```
janos, line 147: ObservedUnderlay exists to communicate the real underlay that a peer is connected to another peer. This may be different than the fullRemoteMA if the peer is under NAT. Hence the name observable. I do not think that this change is correct as it makes the ObservedUnderlay irrelevant and could be removed completely. With this change observedUnderlay and fullRemoteMA are always the same and can be even more simplified, but with losing the information about the observed underlay by the peer.

If there is investigation that observable underlay is not needed here or in the Handle method, it would be good to add to the PR description.

notanatol: Thanks, I will add the relevant info to the description. The problem we are trying to tackle is that `Syn.ObservableUnderlay` value belongs to the unsigned part of the handshake payload and can be manipulated. With this PR we are trying to mitigate this risk.

janos: I am suggesting to sign observable underly instead to just ignore it. Or, at least, to analyze the impact of removing this information.

notanatol: After the discussion we had we decided to put a warning in place.

acud (approved):
Maybe just good to amend the PR title

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -149,6 +150,16 @@ func (s *Service) Handshake(ctx context.Context, stream p2p.Stream, peerMultiadd
...
 	}
 
+	observedUnderlayAddrInfo, err := libp2ppeer.AddrInfoFromP2pAddr(observedUnderlay)
+	if err != nil {
+		return nil, fmt.Errorf("extract addr from P2P: %w", err)
+	}
+
+	if s.libp2pID != observedUnderlayAddrInfo.ID {
+		//NOTE eventually we will return error here, but for now we want to gather some statistics
+		s.logger.Warningf("received peer ID %s does not match ours: %s", observedUnderlayAddrInfo.ID, s.libp2pID)
```
notanatol, line 160: @acud normally we should return error here but we thought about maybe leaving it as a warning to gather some statistics about how often it happens.
Do you think this kind of data would be valuable or should we error here right away?

mrekucci (commented):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2464)* status: all files reviewed, 2 unresolved discussions (waiting on @acud and @janos)

___
*[pkg/p2p/libp2p/internal/handshake/handshake.go, line 160 at r2](https://reviewable.io/reviews/ethersphere/bee/2464#-MiVb_Bqt1xbt5ML5qXu-r2-160:-MivTFzF6V_WFesQJz_4:b5lf4fh) ([raw file](https://github.com/ethersphere/bee/blob/12e81a3ce876688d7169fbc1edc202bdafbfd07b/pkg/p2p/libp2p/internal/handshake/handshake.go#L160)):*

@acud normally we should return error here but we thought about maybe leaving it as a warning to gather some statistics about how often it happens.
Do you think this kind of data would be valuable or should we error here right away?
</blockquote></details>

I think you can return an error after the logline!?

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2464)* status: all files reviewed, 2 unresolved discussions (waiting on @acud and @janos)

___
*[pkg/p2p/libp2p/internal/handshake/handshake.go, line 160 at r2](https://reviewable.io/reviews/ethersphere/bee/2464#-MiVb_Bqt1xbt5ML5qXu-r2-160:-MivYQbzBHo2c6Hepd4-:bm9x9fl) ([raw file](https://github.com/ethersphere/bee/blob/12e81a3ce876688d7169fbc1edc202bdafbfd07b/pkg/p2p/libp2p/internal/handshake/handshake.go#L160)):*

I think you can return an error after the logline!?
</blockquote></details>

Yes, technically it's doable, but it's just a general concern when it comes to introducing such a change to the behavior of the handshake protocol.
We are not sure how often this condition would be met "in the wild" where some peers might be behind NAT and the underlays might not match.

pkg/p2p/libp2p/internal/handshake/handshake.go
acud: sure :+1:

acud (approved):

janos (approved):
LGTM.

PR #2467, opened 2021-09-01
refactor: simplify kademlia announce to

istae:
AnnounceTo is only called by libp2p package in cases where a newly connected peer is a full node which then gets announced to already connected light nodes. The function itself has an unnecessary full-node check, which this PR removes.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2467)


acud:
The reason why I did it this way was because I just wouldn't like to create a situation where a light node is announced. Having the `fullnode` flag makes sure that attention is paid to which type of node is announced. The change is fine, but then I wonder if we should change the method name to be something a bit more explicit, like `AnnounceFull` or something along those lines

PR #2474, opened 2021-09-03
build: add -buildid='' so that the binaries are reproducible

attila-lendvai:
ideally, this should be tested by the CI so that it doesn't break.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2474)


attila-lendvai:
so, `buildid` is stable on the same host, but not sure about different hosts.

it's being discussed on https://github.com/ethereum/go-ethereum/pull/23527

marking it as draft until clarified.

acud:
let's reopen this once we have more concrete idea whether this solves the problem or not

PR #2482, opened 2021-09-07
fix: retry user agent get from libp2p peerstore

janos:
This pr fixes the flaky test that checks for libp2p user agent in logs in high load environments like CI. It also ensures higher probability that the user agent is logged in high load environments.

The core problem is that the peerstore may not contain all keys and values right after the connections is created.
This retry mechanism ensures more reliable user agent propagation.

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2482)

### DIFF
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -437,10 +437,13 @@ func (s *Service) handleIncoming(stream network.Stream) {
 		return
 	}
 
-	peerUserAgent := appendSpace(s.peerUserAgent(peerID))
+	peerUserAgent, err := s.peerUserAgent(peerID)
+	if err != nil {
+		s.logger.Debugf("stream handler: inbound peer %s user agent: %w", err)
+	}
 
-	s.logger.Debugf("stream handler: successfully connected to peer %s%s%s (inbound)", i.BzzAddress.ShortString(), i.LightString(), peerUserAgent)
-	s.logger.Infof("stream handler: successfully connected to peer %s%s%s (inbound)", i.BzzAddress.Overlay, i.LightString(), peerUserAgent)
+	s.logger.Debugf("stream handler: successfully connected to peer %s%s%s (inbound)", i.BzzAddress.ShortString(), i.LightString(), appendSpace(peerUserAgent))
+	s.logger.Infof("stream handler: successfully connected to peer %s%s%s (inbound)", i.BzzAddress.Overlay, i.LightString(), appendSpace(peerUserAgent))
 }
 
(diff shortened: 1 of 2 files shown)


acud:
I think this needs a rebase

pkg/p2p/libp2p/libp2p.go
```diff
@@ -884,7 +884,19 @@ func (s *Service) Ping(ctx context.Context, addr ma.Multiaddr) (rtt time.Duratio
...
 // "github.com/libp2p/go-libp2p" and returns empty string in that case.
 func (s *Service) peerUserAgent(peerID libp2ppeer.ID) string {
-	v, err := s.host.Peerstore().Get(peerID, "AgentVersion")
+	var (
+		v   interface{}
+		err error
+	)
+	// Peerstore may not contain all keys and values right after the connections is created.
+	// This retry mechanism ensures more reliable user agent propagation.
+	for deadline := time.Now().Add(2 * time.Second); time.Now().Before(deadline); {
```
acud, line 893: my only question would be: could it be that the retry mechanism would make some context with deadline in the enclosing context expire?
example:
```go
func (k *Kad) connect(ctx context.Context, peer swarm.Address, ma ma.Multiaddr) error {
	ctx, cancel := context.WithTimeout(ctx, peerConnectionAttemptTimeout)
	defer cancel()

	switch i, err := k.p2p.Connect(ctx, ma); {
```
kademlia dial has a 5 seconds timeout in this case. so the user agent printing might delay this and make it expire. I wonder if we could either decrease this timespan to be shorter so that it becomes insignificant comparing to the total connection context deadline or simply do this polling in a different goroutine and print it eventually in an unrelated logline that would have the overlay and user agent together.

janos: Good suggestion. I've added context handling to the function.

acud (approved):

pkg/p2p/libp2p/libp2p.go
```diff
@@ -883,8 +886,26 @@ func (s *Service) Ping(ctx context.Context, addr ma.Multiaddr) (rtt time.Duratio
...
-func (s *Service) peerUserAgent(peerID libp2ppeer.ID) string {
-	v, err := s.host.Peerstore().Get(peerID, "AgentVersion")
+func (s *Service) peerUserAgent(ctx context.Context, peerID libp2ppeer.ID) string {
+	ctx, cancel := context.WithTimeout(ctx, peerUserAgentTimeout)
+	defer cancel()
+	var (
+		v   interface{}
+		err error
+	)
+	// Peerstore may not contain all keys and values right after the connections is created.
```
notanatol, line 896: typo: *connection

notanatol (approved):

PR #2486, opened 2021-09-08
chore: remove deepsource for good

vandot:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2486)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -49,6 +49,9 @@ jobs:
     - name: Vet
       if: matrix.os == 'ubuntu-latest'
       run: make vet
+    - name: Coverage
+      if: matrix.os == 'ubuntu-latest'
+      run: go test -coverprofile=cover.out ./...
     - name: Test with Race Detector
       # do not tests with race detector on windows
       # until the issue with badgerdb memory allocation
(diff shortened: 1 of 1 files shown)


mrekucci (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2486)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud and @istae)

.github/workflows/go.yml
```diff
@@ -49,6 +49,9 @@ jobs:
     - name: Vet
       if: matrix.os == 'ubuntu-latest'
       run: make vet
+    - name: Coverage
+      if: matrix.os == 'ubuntu-latest'
+      run: go test -coverprofile=cover.out ./...
```
acud, line 54: i think this step is not needed. you can just add `-coverprofile=cover.out` to the test step below (Test with Race detector). the coveralls only run on the ubuntu build (the one that uses the test with race detector). not sure if `make` will like the extra params though, we might need to change the makefile

vandot: @acud I thought that but `make` doesn't accept additional argument (unless I do some changes) and we will create `cover.out` on mac as well (which is not a real issue), if we do it like this tests will be cached and we will not slow down whole pipeline

acud: cool :+1:

acud (approved):

PR #2487, opened 2021-09-08
chore: improve metrics

acud:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2487)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -127,7 +127,7 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 	defer cancel()
 	defer func() {
 		if err != nil {
-			ps.metrics.TotalErrors.Inc()
+			ps.metrics.TotalHandlerErrors.Inc()
 			_ = stream.Reset()
 		} else {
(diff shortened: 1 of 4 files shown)


pkg/pushsync/pushsync.go
```diff
@@ -127,7 +127,7 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 	defer cancel()
 	defer func() {
 		if err != nil {
-			ps.metrics.TotalErrors.Inc()
+			ps.metrics.TotalHandlerErrors.Inc()
```
janos, line 130: Should TotalHandlerErrors be added alongside TotalErrors as this error should be counted in to TotalErrors as well?

acud: so we were going through this yesterday while creating KPIs for pushsync and we wanted to separate success rates for different functions. the problem is that TotalErrors if added here would be potentially incremented twice per forwarding operation, which means we cannot use it to build a success rate of a certain code section

notanatol (approved):

istae (approved):

janos (approved):

PR #2490, opened 2021-09-09
feat: add handshake protocol messages metrics

mrekucci:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2490)

### DIFF
--- a/pkg/p2p/libp2p/internal/handshake/handshake.go
+++ b/pkg/p2p/libp2p/internal/handshake/handshake.go
@@ -77,6 +77,7 @@ type Service struct {
 	receivedHandshakesMu  sync.Mutex
 	logger                logging.Logger
 	libp2pID              libp2ppeer.ID
+	metrics               metrics
 	network.Notifiee      // handshake service can be the receiver for network.Notify
 }
 
@@ -111,6 +112,7 @@ func New(signer crypto.Signer, advertisableAddresser AdvertisableAddressResolver
 		receivedHandshakes:    make(map[libp2ppeer.ID]struct{}),
 		libp2pID:              ownPeerID,
 		logger:                logger,
+		metrics:               newMetrics(),
 		Notifiee:              new(network.NoopNotifiee),
 	}
 	svc.welcomeMessage.Store(welcomeMessage)
@@ -244,8 +246,10 @@ func (s *Service) Handle(ctx context.Context, stream p2p.Stream, remoteMultiaddr
 
 	var syn pb.Syn
 	if err := r.ReadMsgWithContext(ctx, &syn); err != nil {
+		s.metrics.SynRxFailed.Inc()
 		return nil, fmt.Errorf("read syn message: %w", err)
 	}
+	s.metrics.SynRx.Inc()
 
 	observedUnderlay, err := ma.NewMultiaddrBytes(syn.ObservedUnderlay)
 	if err != nil {
@@ -285,13 +289,17 @@ func (s *Service) Handle(ctx context.Context, stream p2p.Stream, remoteMultiaddr
 			WelcomeMessage: welcomeMessage,
 		},
 	}); err != nil {
+		s.metrics.SynAckTxFailed.Inc()
 		return nil, fmt.Errorf("write synack message: %w", err)
 	}
+	s.metrics.SynAckTx.Inc()
 
 	var ack pb.Ack
 	if err := r.ReadMsgWithContext(ctx, &ack); err != nil {
+		s.metrics.AckRxFailed.Inc()
 		return nil, fmt.Errorf("read ack message: %w", err)
 	}
+	s.metrics.AckRx.Inc()
 
 	if ack.NetworkID != s.networkID {
 		return nil, ErrNetworkIDIncompatible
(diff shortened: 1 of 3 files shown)


pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -111,6 +112,7 @@ func New(signer crypto.Signer, advertisableAddresser AdvertisableAddressResolver
 		receivedHandshakes:    make(map[libp2ppeer.ID]struct{}),
 		libp2pID:              ownPeerID,
 		logger:                logger,
+		metrics:               newMetrics(),
```
acud, line 115: exposing metrics from this package might be tricky... since the package is internal and metrics need to be registered inside the `node.go` file

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2490)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @acud and @notanatol)

___
*[pkg/p2p/libp2p/internal/handshake/handshake.go, line 115 at r1](https://reviewable.io/reviews/ethersphere/bee/2490#-MjAELJr5JnRMejTGMNZ-r1-115:-MjTGdxi8IGp4O5mEed1:bcbqrnc) ([raw file](https://github.com/ethersphere/bee/blob/07bc463e6eda9e5a7f48260469df40494dace7c8/pkg/p2p/libp2p/internal/handshake/handshake.go#L115)):*

exposing metrics from this package might be tricky... since the package is internal and metrics need to be registered inside the `node.go` file
</blockquote></details>

Nice catch! Since the handshake is an internal part of the libp2p, I'll instantiate it as a part of the libp2p `Matrics()` call.

acud (changes requested):
I would suggest to split this into two: (1) metrics only PR (2) adding logging for specific debugging sessions which should not end up on `master` (and don't need a PR per-se)

> Reviewed 2 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2490#-:-MjUWMpv-zpov8nc8ZmM:b-64qqhr)* status: all files reviewed, 3 unresolved discussions (waiting on @mrekucci and @notanatol)

___
*[pkg/p2p/libp2p/internal/handshake/handshake.go, line 253 at r2](https://reviewable.io/reviews/ethersphere/bee/2490#-MjUVnj80kLVSMm6xYOn:-MjUVnj92H8WAciclDUY:b-bbq996) ([raw file](https://github.com/ethersphere/bee/blob/a92bb46728484bac31cd990afc0092d22b79e7c8/pkg/p2p/libp2p/internal/handshake/handshake.go#L253)):*
> ```Go
> 	if err := r.ReadMsgWithContext(ctx, &syn); err != nil {
> 		s.metrics.SynRxFailed.Inc()
> 		s.logger.Tracef("handshake: read syn msg error; took: %s", time.Since(beginning))
> ```

you are logging and returning an error, which is not needed since the error would normally be logged anyway by the calling context. if timing info is needed you can instrument the error with it in the line below
___
*[pkg/p2p/libp2p/internal/handshake/handshake.go, line 257 at r2](https://reviewable.io/reviews/ethersphere/bee/2490#-MjUW1X13fWNXhI7ovT0:-MjUW1X13fWNXhI7ovT1:b-exkf5) ([raw file](https://github.com/ethersphere/bee/blob/a92bb46728484bac31cd990afc0092d22b79e7c8/pkg/p2p/libp2p/internal/handshake/handshake.go#L257)):*
> ```Go
> 	}
> 	s.metrics.SynRx.Inc()
> 	s.logger.Tracef("handshake: read syn msg took: %s", time.Since(beginning))
> ```

perhaps this can be added to an already existing logline, so we don't clutter the terminal with too much information? if specific debugging information is needed in order to find a bug this is fine, but then maybe we don't need to have it on `master`
___
*[pkg/p2p/libp2p/internal/handshake/handshake.go, line 299 at r2](https://reviewable.io/reviews/ethersphere/bee/2490#-MjUWDGDBmkF5YM-MTOk:-MjUWDGDBmkF5YM-MTOl:bq8f1ue) ([raw file](https://github.com/ethersphere/bee/blob/a92bb46728484bac31cd990afc0092d22b79e7c8/pkg/p2p/libp2p/internal/handshake/handshake.go#L299)):*
> ```Go
> 	}); err != nil {
> 		s.metrics.SynAckTxFailed.Inc()
> 		s.logger.Tracef("handshake: write syn ack msg error; took: %s; total: %s", time.Since(start), time.Since(beginning))
> ```

same comment here about logging and returning an error. it is fine if it doesn't end up on `master` though)

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2490)* status: all files reviewed, 3 unresolved discussions (waiting on @acud, @mrekucci, and @notanatol)

___
*[pkg/p2p/libp2p/internal/handshake/handshake.go, line 253 at r2](https://reviewable.io/reviews/ethersphere/bee/2490#-MjUVnj80kLVSMm6xYOn:-MjYNq1W3MPJIW090gs2:bctq28a) ([raw file](https://github.com/ethersphere/bee/blob/a92bb46728484bac31cd990afc0092d22b79e7c8/pkg/p2p/libp2p/internal/handshake/handshake.go#L253)):*
> you are logging and returning an error

The trace logging is about knowing how much of time each step took in the protocol execution not about logging an error.

> if timing info is needed you can instrument the error with it in the line below

It can be inferred but not with such granularity or when the step is successful.

I agree though that I can move the traces into a separate branch.

acud (approved):

istae (approved):

notanatol (approved):

PR #2491, opened 2021-09-10
feat: skip validation if peer is known

istae:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2491)

### DIFF
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -66,9 +66,10 @@ type Service struct {
 	peersChan       chan pb.Peers
 	sem             *semaphore.Weighted
 	lru             *lru.Cache // cache for unreachable peers
+	bootnode        bool
 }
 
-func New(streamer p2p.StreamerPinger, addressbook addressbook.GetPutter, networkID uint64, logger logging.Logger) (*Service, error) {
+func New(streamer p2p.StreamerPinger, addressbook addressbook.GetPutter, networkID uint64, bootnode bool, logger logging.Logger) (*Service, error) {
 	lruCache, err := lru.New(cacheSize)
 	if err != nil {
 		return nil, err
@@ -86,6 +87,7 @@ func New(streamer p2p.StreamerPinger, addressbook addressbook.GetPutter, network
 		peersChan:   make(chan pb.Peers),
 		sem:         semaphore.NewWeighted(int64(31)),
 		lru:         lruCache,
+		bootnode:    bootnode,
 	}
 	svc.startCheckPeersHandler()
 	return svc, nil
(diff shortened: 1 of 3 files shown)


acud (approved):

acud:
@istae just a suggestion, but we might also want not to check the same peer twice, in case the result was a negative. for this we can reuse the LRU cache used in the chainsync PR, to basically cache the results (true/false) for an address - no entry means we didn't try, false means unreachable (dont retry) and true means reachable. having an upper bound of addresses in the cache would mean we could potentially retry after a while (but it should be well into the thousands range).

i would also consider not to have this validation on the bootnode at all since it does not actively dial

mrekucci (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2491)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @istae)

___
*[pkg/hive/hive.go, line 268 at r1](https://reviewable.io/reviews/ethersphere/bee/2491#-MjSx9hJEkcO_Qn3r32d:-MjSx9hJEkcO_Qn3r32e:b-n00s3l) ([raw file](https://github.com/ethersphere/bee/blob/442b315d0061da5912a85d537d039adf06905c6f/pkg/hive/hive.go#L268)):*
> ```Go
> 
> 		// if peer exists already, skip validation
> 		if _, err := s.addressBook.Get(swarm.NewAddress(p.Overlay)); err == nil {
> ```

The errors thrown by the `Get` method should be handled (at least logged).

aloknerurkar (approved):
If this is happening a lot, I would also suggest an LRU cache here to speed it up more as this is going to disk every time.

pkg/hive/hive.go
```diff
@@ -260,57 +272,102 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
...
 	mtx := sync.Mutex{}
-	wg := sync.WaitGroup{}
 
-	for _, p := range peers.Peers {
-		err := s.sem.Acquire(ctx, 1)
-		if err != nil {
-			return
-		}
+	if s.bootnode {
+		for _, p := range peers.Peers {
```
acud, line 277: not sure what is the desirable outcome here. this basically means - if we are a bootnode we persist every peer we get over discovery. if anything, we should not persist _any_ node which we are not connected to

pkg/hive/hive.go
```diff
@@ -260,57 +272,102 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
...
-				Transaction: newPeer.Transaction,
+			// mark peer as seen
+			_ = s.lru.Add(cacheOverlay, nil)
+
+			// if peer exists already in the addressBook, skip
+			_, err := s.addressBook.Get(overlay)
+			if err == nil {
+				continue
+			} else {
+				s.logger.Errorf("hive: addressbook %v", err)
```
acud, line 320: this means that you return an error here if the peer is _not_ in the address book, so you basically fail for every peer you don't know

istae: @mrekucci's request was that we should at least log it, but I agree, logging here is does not add much value.

pkg/hive/hive.go
istae: because bootnodes do not dial, the newest commit discards the received peers list without storing or validating it.

mrekucci (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2491)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @acud and @istae)

pkg/hive/hive.go
```diff
@@ -39,6 +40,8 @@ const (
 	maxBatchSize           = 30
 	pingTimeout            = time.Second * 5 // time to wait for ping to succeed
 	batchValidationTimeout = 5 * time.Minute // prevent lock contention on peer validation
+	cacheSize              = 25000
+	cachePrefix            = 4 // enough bytes (32 bits) to uniquely identify a peer
```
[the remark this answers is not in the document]
istae: So I increased the cache size to 100,000 and kept the cachePrefix, it's best if cache is big for hive.

pkg/hive/hive.go
```diff
@@ -39,6 +40,8 @@ const (
 	maxBatchSize           = 30
 	pingTimeout            = time.Second * 5 // time to wait for ping to succeed
 	batchValidationTimeout = 5 * time.Minute // prevent lock contention on peer validation
+	cacheSize              = 100000
+	cachePrefix            = swarm.MaxBins / 8 // enough bytes (32 bits) to uniquely identify a peer
```
acud, line 44: having a `const` here would be more readable

pkg/hive/hive.go
```diff
@@ -263,6 +291,23 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
...
+		overlay := swarm.NewAddress(p.Overlay)
+		cacheOverlay := overlay.ByteString()[:cachePrefix]
+
+		// cached peer, skip
+		if _, ok := s.lru.Get(cacheOverlay); ok {
+			continue
+		}
+
+		// mark peer as seen
+		_ = s.lru.Add(cacheOverlay, nil)
```
acud, line 304: this is suboptimal to do here and better done once the `Ping` returns, since `context` can expire in the meanwhile, we might not even get to the point where we try to ping the peer, resulting in some peers marked as seen when in fact we haven't even checked them yet

acud (changes requested):

pkg/hive/hive.go
```diff
@@ -62,9 +66,24 @@ type Service struct {
...
 
-func New(streamer p2p.StreamerPinger, addressbook addressbook.GetPutter, networkID uint64, logger logging.Logger) *Service {
+func New(streamer p2p.StreamerPinger, addressbook addressbook.GetPutter, networkID uint64, bootnode bool, logger logging.Logger) (*Service, error) {
+
+	var (
+		lruCache *lru.Cache
+		err      error
+	)
+
+	if !bootnode {
```
acud, line 80: i think it is fine to initialize this even in bootnode mode. the cache size is negligible when empty, and we don't risk any panics on bootnodes

acud (approved):

PR #2492, opened 2021-09-10
feat: ping incoming peers

istae:
Incoming peers may not be reachable so they should be pinged before being added to the addressbook.  

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2492)

### DIFF
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -361,10 +361,11 @@ func (s *Service) handleIncoming(stream network.Stream) {
 	if i.FullNode {
 
 		ctx, cancel := context.WithTimeout(context.Background(), pingTimeout)
-		defer cancel()
+		_, err := s.Ping(ctx, i.BzzAddress.Underlay)
+		cancel()
 
 		// only insert to addressbook if peer is reachable
-		if _, err := s.Ping(ctx, i.BzzAddress.Underlay); err == nil {
+		if err == nil {
 			err = s.addressbook.Put(i.BzzAddress.Overlay, *i.BzzAddress)
 			if err != nil {
 				s.logger.Debugf("stream handler: addressbook put error %s: %v", peerID, err)
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -358,6 +358,7 @@ func (s *Service) handleIncoming(stream network.Stream) {
 		return
 	}
 
+	var reachable bool
 	if i.FullNode {
 
 		ctx, cancel := context.WithTimeout(context.Background(), pingTimeout)
@@ -366,6 +367,7 @@ func (s *Service) handleIncoming(stream network.Stream) {
 
 		// only insert to addressbook if peer is reachable
 		if err == nil {
+			reachable = true
 			err = s.addressbook.Put(i.BzzAddress.Overlay, *i.BzzAddress)
 			if err != nil {
 				s.logger.Debugf("stream handler: addressbook put error %s: %v", peerID, err)
@@ -373,6 +375,8 @@ func (s *Service) handleIncoming(stream network.Stream) {
 				_ = s.Disconnect(i.BzzAddress.Overlay, "unable to persist peer in addressbook")
 				return
 			}
+		} else {
+			s.logger.Errorf("stream handler: peer ping %s: %v", peerID, err)
 		}
 	}
 
@@ -414,7 +418,7 @@ func (s *Service) handleIncoming(stream network.Stream) {
 				}
 			}
 		} else {
-			if err := s.notifier.Connected(s.ctx, peer, false); err != nil {
+			if err := s.notifier.Connected(s.ctx, peer, reachable, false); err != nil {
 				s.logger.Debugf("stream handler: notifier.Connected: peer disconnected: %s: %v", i.BzzAddress.Overlay, err)
 				// note: this cannot be unit tested since the node
 				// waiting on handshakeStream.FullClose() on the other side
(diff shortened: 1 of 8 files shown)


mrekucci (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2492)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @istae)

___
*[pkg/p2p/libp2p/libp2p.go, line 368 at r1](https://reviewable.io/reviews/ethersphere/bee/2492#-MjSvMS93k85Sr8976-V:-MjSvMSB9KB30KexsT9l:b1t4zp0) ([raw file](https://github.com/ethersphere/bee/blob/3284186872fa9941a8a3459b345679158f339691/pkg/p2p/libp2p/libp2p.go#L368)):*
> ```Go
> 
> 		// only insert to addressbook if peer is reachable
> 		if err == nil {
> ```

The error should be handled (at least logged) for the non-nil case.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -358,12 +359,20 @@ func (s *Service) handleIncoming(stream network.Stream) {
...
-			_ = s.Disconnect(i.BzzAddress.Overlay, "unable to persist peer in addressbook")
-			return
+
+		ctx, cancel := context.WithTimeout(context.Background(), pingTimeout)
+		_, err := s.Ping(ctx, i.BzzAddress.Underlay)
+		cancel()
+
+		// only insert to addressbook if peer is reachable
+		if err == nil {
+			err = s.addressbook.Put(i.BzzAddress.Overlay, *i.BzzAddress)
```
aloknerurkar, line 369: The issue here is although the peer's underlay address is not reachable, we have an active connection with it. If we dont put the address in the addressbook, we have an overlay address in the known peers list and not in the addressbook. Is this okay? Can this cause issues elsewhere?

istae: In this case that this peer disconnects, because this peer is in the knownpeers list, kademlia could possibly attempt to dial for a connection again in the future, so good catch. 

I will fix this.

acud: the other thing is that we won't be announcing this peer since the addressbook entry could not be found. so this peer will be picked up and would return an error when we try to announce. this could result in discovery issues, since we gossip less peers than expected

pkg/topology/kademlia/kademlia.go
```diff
@@ -883,22 +883,24 @@ func (k *Kad) Connected(ctx context.Context, peer p2p.Peer, forceConnection bool
...
 
-func (k *Kad) onConnected(ctx context.Context, addr swarm.Address) error {
+func (k *Kad) onConnected(ctx context.Context, addr swarm.Address, reachable bool) error {
 	if err := k.Announce(ctx, addr, true); err != nil {
 		return err
 	}
 
-	k.knownPeers.Add(addr)
+	if reachable {
+		k.knownPeers.Add(addr)
```
acud, line 902: not sure about this as it changes the semantics of `knownPeers`. so it is no longer `knownPeers` but `knownAndReachable` peers. this will make connectivity issues more difficult to detect using the debug apis

acud (changes requested):
not sure about this change, it might be a bit too much

pkg/topology/kademlia/kademlia.go
istae: I'm ok with removing this change. It was added to deal with the case that if the unreachable peer connects, then some time later disconnects, kademlia could possibly attempt to reconnect to it. But this is not a common case and we can let kademlia fail to dial to this peer, it's not a big issue.

pkg/p2p/libp2p/libp2p.go
istae: The current hive logic skips the peer if the addressbook entry is not found and continues with other peers. 
In general, afaik unreachable peers should not be announced.

acud: hive will skip but kademlia gives hive a peer which is not consistent with the address book state. i.e. kademlia assumes that the subset of peers given for discovery are all in the address book but this is not the case.

acud: what i'm saying here is that you're adding more dependencies within kademlia when you do the announcements, since you can't just pick `n` nodes to gossip randomly, but you must ensure they are all in the address book prior to communicating them to hive

istae: Ideally, all entries in the addressbook should be reachable. This is why we do pinging now before we insert peers into the addressbook.
Since not all connected peers in kademlia may be reachable,  kademlia cannot make that assumption.

acud: this is not about reachability, please have a look at `kademlia.go#Announce`:
```go
// Announce a newly connected peer to our connected peers, but also
// notify the peer about our already connected peers
func (k *Kad) Announce(ctx context.Context, peer swarm.Address, fullnode bool) error {
	var addrs []swarm.Address

	for bin := uint8(0); bin < swarm.MaxBins; bin++ {

		connectedPeers, err := randomSubset(k.connectedPeers.BinPeers(bin), broadcastBinSize)
		if err != nil {
			return err
		}

		for _, connectedPeer := range connectedPeers {
			if connectedPeer.Equal(peer) {
				continue
			}

			addrs = append(addrs, connectedPeer)

			if !fullnode {
				// we continue here so we dont gossip
				// about lightnodes to others.
				continue
			}
			go func(connectedPeer swarm.Address) {
				if err := k.discovery.BroadcastPeers(ctx, connectedPeer, peer); err != nil {
					k.logger.Debugf("could not gossip peer %s to peer %s: %v", peer, connectedPeer, err)
				}
			}(connectedPeer)
		}
	}
```
here, `randomSubset` picks a random few connected peers, however with your change it is no longer guaranteed that any of these peers are in the address book, resulting in peers not being gossiped to other peers

aloknerurkar (commented):
I dont think these changes are needed at all tbh. The risk to reward ratio is too high here.

Issue #1876, opened 2021-05-26
Segmentation violation: unexpect fault address


stevensu1977:
#### Summary
I use v0.6.0 , when I stop my docker container and restart always get error , I try use normal binary not in docker , get same error.

#### Error Info

INFO[2021-05-26T01:22:50Z] debug api address: [::]:1635
unexpected fault address 0x3939373134934c
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x1 addr=0x3939373134934c pc=0x4d2d00]

goroutine 1 [running]:
runtime.throw(0xe14787, 0x5)
	runtime/panic.go:1117 +0x54 fp=0x40003cfc30 sp=0x40003cfc00 pc=0x44914
runtime.sigpanic()
	runtime/signal_unix.go:741 +0x230 fp=0x40003cfc70 sp=0x40003cfc30 pc=0x5c450
github.com/golang/snappy.encodeBlock(0x4000a58a02, 0x1494, 0x1494, 0x4000a7c000, 0x1186, 0x1b53, 0x901)
	github.com/golang/snappy@v0.0.2/encode_arm64.s:666 +0x360 fp=0x40003d7d10 sp=0x40003cfc80 pc=0x4d2d00
github.com/golang/snappy.Encode(0x4000a58a00, 0x1496, 0x1496, 0x0, 0x0, 0x0, 0x879, 0x879, 0x40003d7e28)
	github.com/golang/snappy@v0.0.2/encode.go:39 +0x17c fp=0x40003d7d60 sp=0x40003d7d10 pc=0x4d223c
github.com/syndtr/goleveldb/leveldb/table.(*Writer).writeBlock(0x40008646c0, 0x4000864718, 0x2, 0x0, 0x7, 0x10, 0x4000a081f0)
	github.com/syndtr/goleveldb@v1.0.1-0.20200815110645-5c35d600f0ca/leveldb/table/writer.go:171 +0xb0 fp=0x40003d7de0 sp=0x40003d7d60 pc=0x4daa20
github.com/syndtr/goleveldb/leveldb/table.(*Writer).finishBlock(0x40008646c0, 0x4004fa33c9, 0x9)
	github.com/syndtr/goleveldb@v1.0.1-0.20200815110645-5c35d600f0ca/leveldb/table/writer.go:222 +0x4c fp=0x40003d7e30 sp=0x40003d7de0 pc=0x4daebc
[...]

Issue #1806, opened 2021-05-21
Bee 0.6.0-67cc8f24 segfaults on arm64 (Raspberry Pi 4, 8GB)


mfw78:
#### Summary
After following the upgrade instructions, installing bee-clef `0.4.12` and bee `0.6.0`, `bee` fails to start - segfaulting.

The `/var/lib/bee` folder was sanitised to contain only as follows:

```
root@raspberrypi:/var/lib/bee# ls -alh
total 20K
drwxr-xr-x  4 root root 4.0K May 21 10:54 .
drwxr-xr-x 46 root root 4.0K May 21 10:55 ..
drwx------  2 bee  bee  4.0K Apr 28 03:03 keys
-rw-------  1 bee  bee    32 Apr 28 03:02 password
drwxr-xr-x  2 bee  bee  4.0K May 21 10:55 statestore
```

#### Steps to reproduce
Apply standard upgrade instructions.

#### Expected behavior
Bee to start gracefully and connect to the networking, collecting some of that sweet, sweet pollen to make some gBZZ.

#### Actual behavior

```
May 21 10:56:00 raspberrypi systemd[1]: Started Bee - Ethereum Swarm node.
May 21 10:56:00 raspberrypi bee[19827]: Welcome to the Swarm.... Bzzz Bzzzz Bzzzz
May 21 10:56:00 raspberrypi bee[19827]:                .-.         .--''-.
May 21 10:56:00 raspberrypi bee[19827]:              .'   '.     /'       '
May 21 10:56:00 raspberrypi bee[19827]:              '.     '. ,'          |
May 21 10:56:00 raspberrypi bee[19827]:    _        o    '.o   ,'        _.-'
May 21 10:56:00 raspberrypi bee[19827]:  .\ /.       \.--./'. /.:. :._:.'
May 21 10:56:00 raspberrypi bee[19827]: < ~O~ >    .'    '._-': ': ': ': ':
May 21 10:56:00 raspberrypi bee[19827]:  '/_\'     :(.) (.) :  ': ': ': ': ':>-
May 21 10:56:00 raspberrypi bee[19827]:  \ | /      ' ____ .'_.:' :' :' :' :'
May 21 10:56:00 raspberrypi bee[19827]:   \|/        '\<>/'/ | | :' :' :'
May 21 10:56:00 raspberrypi bee[19827]:    |               \  \ \
May 21 10:56:00 raspberrypi bee[19827]:    |                '  ' '
[...]

PR #2495, opened 2021-09-13
chore: update golang and ldb

acud:
leveldb has updated the snappy compression that caused a segfault on arm64 cpus. as of such we can update to go 1.16 (although 1.17 is already out)

related to #1876 #1806 

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2495)


janos (approved):

istae (approved):
Thanks!

PR #2498, opened 2021-09-14
feat: ping time metric

istae:
This should make network latency observable.

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2498)

### DIFF
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -288,9 +288,9 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
 
 			// check if the underlay is usable by doing a raw ping using libp2p
 			if _, err = s.streamer.Ping(ctx, multiUnderlay); err != nil {
+				s.metrics.PingTime.Observe(float64(time.Since(start).Milliseconds()))
 				s.metrics.UnreachablePeers.Inc()
 				s.logger.Debugf("hive: peer %s: underlay %s not reachable", hex.EncodeToString(newPeer.Overlay), multiUnderlay)
-				s.metrics.PingTime.Observe(float64(time.Since(start).Milliseconds()))
 				return
 			}
 			s.metrics.PingTime.Observe(float64(time.Since(start).Milliseconds()))
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -288,7 +288,7 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
 
 			// check if the underlay is usable by doing a raw ping using libp2p
 			if _, err = s.streamer.Ping(ctx, multiUnderlay); err != nil {
-				s.metrics.PingTime.Observe(float64(time.Since(start).Milliseconds()))
+				s.metrics.PingFailureTime.Observe(float64(time.Since(start).Milliseconds()))
 				s.metrics.UnreachablePeers.Inc()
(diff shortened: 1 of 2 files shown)


pkg/hive/hive.go
```diff
@@ -284,12 +284,16 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
 			ctx, cancel := context.WithTimeout(ctx, pingTimeout)
 			defer cancel()
 
+			start := time.Now()
+
 			// check if the underlay is usable by doing a raw ping using libp2p
 			if _, err = s.streamer.Ping(ctx, multiUnderlay); err != nil {
+				s.metrics.PingTime.Observe(float64(time.Since(start).Milliseconds()))
```
acud, line 291: wouldn't it be useful to split this into an error metric for the timing?

pkg/hive/hive.go
```diff
@@ -284,12 +284,16 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
 			ctx, cancel := context.WithTimeout(ctx, pingTimeout)
 			defer cancel()
 
+			start := time.Now()
+
 			// check if the underlay is usable by doing a raw ping using libp2p
 			if _, err = s.streamer.Ping(ctx, multiUnderlay); err != nil {
+				s.metrics.PingFailureTime.Observe(float64(time.Since(start).Milliseconds()))
```
acud, line 291: it is not needed to convert to milliseconds. grafana supports nanoseconds like the default golang time representation

pkg/hive/hive.go
```diff
@@ -284,12 +284,16 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
...
 				s.metrics.UnreachablePeers.Inc()
 				s.logger.Debugf("hive: peer %s: underlay %s not reachable", hex.EncodeToString(newPeer.Overlay), multiUnderlay)
 				return
 			}
+			s.metrics.PingTime.Observe(float64(time.Since(start).Milliseconds()))
```
acud, line 296: same

pkg/hive/metrics.go
```diff
@@ -59,6 +62,18 @@ func newMetrics() metrics {
 			Name:      "unreachable_peers_count",
 			Help:      "Number of peers that are unreachable.",
 		}),
+		PingTime: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "ping_time",
+			Help:      "The time spent for pings.",
```
acud, line 69: maybe worth mentioning that it expects ms. also, histogram needs buckets I think. it might be useful to define them in advance (not sure if it does automatic bucket division on its own)

pkg/hive/metrics.go
```diff
@@ -59,6 +62,18 @@ func newMetrics() metrics {
...
+		}),
+		PingFailureTime: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      "fail_ping_time",
+			Help:      "The time spent for unsuccessful pings.",
+		}),
```
acud, line 76: and here

aloknerurkar (approved):

pkg/hive/metrics.go
istae: `DefBuckets = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}` default buckets if one is not specified.

acud (approved):

PR #2509, opened 2021-09-15
feat: metrics for non closest storer

metacertain:
Metrics to show how many times replication request came from any other node than the closest connected peer to chunk

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2509)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -170,6 +170,7 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 	if p.FullNode {
 		bytes := chunkAddress.Bytes()
 		if dcmp, _ := swarm.DistanceCmp(bytes, p.Address.Bytes(), ps.address.Bytes()); dcmp == 1 {
+			ps.metrics.TotalReplication.Inc()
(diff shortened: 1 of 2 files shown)


pkg/pushsync/pushsync.go
```diff
@@ -168,6 +168,13 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 				ctxd, canceld := context.WithTimeout(context.Background(), timeToWaitForPushsyncToNeighbor)
 				defer canceld()
 
+				realClosestPeer, err := ps.topologyDriver.ClosestPeer(chunk.Address(), false, swarm.ZeroAddress)
+				if err == nil {
+					if !realClosestPeer.Equal(p.Address) {
+						ps.metrics.TotalReplicationFromDistantPeer.Inc()
```
acud, line 174: what about having another metric which is the reverse? i.e. have an `else` clause with a replication metric for the number of correct replications we got. also having an extra metric at the top between L165 and L166 to count how many total replication calls we get from other full nodes? this way we can have percentage rates of how many calls were correct, how many calls were not, etc

acud (commented):
looks good, i have minor suggestions

acud (approved):
thanks @metacertain

mrekucci (approved):

PR #2512, opened 2021-09-16
 feat(kademlia): static nodes

aloknerurkar:
Add configuration for adding protected nodes on the bootnodes. This change will prevent kicking out of the protected nodes to ensure better connectivity.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2512)

### DIFF
--- a/cmd/bee/cmd/cmd.go
+++ b/cmd/bee/cmd/cmd.go
@@ -75,6 +75,7 @@ const (
 	optionNameResync                     = "resync"
 	optionNamePProfBlock                 = "pprof-profile"
 	optionNamePProfMutex                 = "pprof-mutex"
+	optionNameProtectedNodes             = "protected-nodes"
 )
 
 func init() {
@@ -260,6 +261,7 @@ func (c *command) setAllFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool(optionNameResync, false, "forces the node to resync postage contract data")
 	cmd.Flags().Bool(optionNamePProfBlock, false, "enable pprof block profile")
 	cmd.Flags().Bool(optionNamePProfMutex, false, "enable pprof mutex profile")
+	cmd.Flags().StringSlice(optionNameProtectedNodes, []string{}, "protect nodes from getting kicked out on bootnode")
 }
 
 func newLogger(cmd *cobra.Command, verbosity string) (logging.Logger, error) {
--- a/cmd/bee/cmd/cmd.go
+++ b/cmd/bee/cmd/cmd.go
@@ -75,7 +75,7 @@ const (
 	optionNameResync                     = "resync"
 	optionNamePProfBlock                 = "pprof-profile"
 	optionNamePProfMutex                 = "pprof-mutex"
-	optionNameProtectedNodes             = "protected-nodes"
+	optionNameStaticNodes                = "static-nodes"
 )
 
 func init() {
@@ -261,7 +261,7 @@ func (c *command) setAllFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool(optionNameResync, false, "forces the node to resync postage contract data")
 	cmd.Flags().Bool(optionNamePProfBlock, false, "enable pprof block profile")
 	cmd.Flags().Bool(optionNamePProfMutex, false, "enable pprof mutex profile")
-	cmd.Flags().StringSlice(optionNameProtectedNodes, []string{}, "protect nodes from getting kicked out on bootnode")
+	cmd.Flags().StringSlice(optionNameStaticNodes, []string{}, "protect nodes from getting kicked out on bootnode")
 }
 
 func newLogger(cmd *cobra.Command, verbosity string) (logging.Logger, error) {
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -31,6 +31,7 @@ import (
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/node"
 	"github.com/ethersphere/bee/pkg/resolver/multiresolver"
+	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/kardianos/service"
 	"github.com/spf13/cobra"
 )
@@ -137,7 +138,16 @@ func (c *command) initStartCmd() (err error) {
 				tracingEndpoint = strings.Join([]string{c.config.GetString(optionNameTracingHost), c.config.GetString(optionNameTracingPort)}, ":")
 			}
 
-			staticNodes := c.config.GetStringSlice(optionNameStaticNodes)
+			var staticNodes []swarm.Address
+
+			for _, p := range c.config.GetStringSlice(optionNameStaticNodes) {
+				addr, err := swarm.ParseHexAddress(p)
+				if err != nil {
+					return errors.New("invalid swarm address configured for static node")
+				}
+
+				staticNodes = append(staticNodes, addr)
+			}
 			if len(staticNodes) > 0 && !bootNode {
 				return errors.New("static nodes can only be configured on bootnodes")
 			}
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -137,6 +137,11 @@ func (c *command) initStartCmd() (err error) {
 				tracingEndpoint = strings.Join([]string{c.config.GetString(optionNameTracingHost), c.config.GetString(optionNameTracingPort)}, ":")
 			}
 
+			protectedNodes := c.config.GetStringSlice(optionNameProtectedNodes)
+			if len(protectedNodes) > 0 && !bootNode {
+				return errors.New("protected nodes can only be configured on bootnodes")
+			}
+
 			b, err := node.NewBee(c.config.GetString(optionNameP2PAddr), signerConfig.publicKey, signerConfig.signer, networkID, logger, signerConfig.libp2pPrivateKey, signerConfig.pssPrivateKey, &node.Options{
 				DataDir:                    c.config.GetString(optionNameDataDir),
 				CacheCapacity:              c.config.GetUint64(optionNameCacheCapacity),
@@ -182,6 +187,7 @@ func (c *command) initStartCmd() (err error) {
 				Resync:                     c.config.GetBool(optionNameResync),
 				BlockProfile:               c.config.GetBool(optionNamePProfBlock),
 				MutexProfile:               c.config.GetBool(optionNamePProfMutex),
+				ProtectedNodes:             protectedNodes,
 			})
 			if err != nil {
 				return err
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -137,9 +137,9 @@ func (c *command) initStartCmd() (err error) {
 				tracingEndpoint = strings.Join([]string{c.config.GetString(optionNameTracingHost), c.config.GetString(optionNameTracingPort)}, ":")
 			}
 
-			protectedNodes := c.config.GetStringSlice(optionNameProtectedNodes)
-			if len(protectedNodes) > 0 && !bootNode {
-				return errors.New("protected nodes can only be configured on bootnodes")
+			staticNodes := c.config.GetStringSlice(optionNameStaticNodes)
+			if len(staticNodes) > 0 && !bootNode {
+				return errors.New("static nodes can only be configured on bootnodes")
 			}
 
 			b, err := node.NewBee(c.config.GetString(optionNameP2PAddr), signerConfig.publicKey, signerConfig.signer, networkID, logger, signerConfig.libp2pPrivateKey, signerConfig.pssPrivateKey, &node.Options{
@@ -187,7 +187,7 @@ func (c *command) initStartCmd() (err error) {
 				Resync:                     c.config.GetBool(optionNameResync),
 				BlockProfile:               c.config.GetBool(optionNamePProfBlock),
 				MutexProfile:               c.config.GetBool(optionNamePProfMutex),
-				ProtectedNodes:             protectedNodes,
+				StaticNodes:                staticNodes,
 			})
 			if err != nil {
 				return err
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -143,7 +143,7 @@ func (c *command) initStartCmd() (err error) {
 			for _, p := range c.config.GetStringSlice(optionNameStaticNodes) {
 				addr, err := swarm.ParseHexAddress(p)
 				if err != nil {
-					return errors.New("invalid swarm address configured for static node")
+					return fmt.Errorf("invalid swarm address %q configured for static node", p)
 				}
 
 				staticNodes = append(staticNodes, addr)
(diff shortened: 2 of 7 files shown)


acud (changes requested):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2512)* status: all files reviewed, 5 unresolved discussions (waiting on @aloknerurkar and @mrekucci)

___
*[cmd/bee/cmd/cmd.go, line 78 at r1](https://reviewable.io/reviews/ethersphere/bee/2512#-MjhXhsa5zOuU4nRWYkQ:-MjhXhsa5zOuU4nRWYkR:bbksrjv) ([raw file](https://github.com/ethersphere/bee/blob/48e06795c7dd382d6c54139a5262bca98785c806/cmd/bee/cmd/cmd.go#L78)):*
> ```Go
> 	optionNamePProfBlock                 = "pprof-profile"
> 	optionNamePProfMutex                 = "pprof-mutex"
> 	optionNameProtectedNodes             = "protected-nodes"
> ```

nit, consider `static-nodes` or `static-node-overlays`
___
*[cmd/bee/cmd/start.go, line 141 at r1](https://reviewable.io/reviews/ethersphere/bee/2512#-MjhXtqZ2TAydF2l7uCm:-MjhXtqZ2TAydF2l7uCn:b1t0fuh) ([raw file](https://github.com/ethersphere/bee/blob/48e06795c7dd382d6c54139a5262bca98785c806/cmd/bee/cmd/start.go#L141)):*
> ```Go
> 
> 			protectedNodes := c.config.GetStringSlice(optionNameProtectedNodes)
> 			if len(protectedNodes) > 0 && !bootNode {
> ```

we might want to apply this for other nodes too
___
*[pkg/node/node.go, line 525 at r1](https://reviewable.io/reviews/ethersphere/bee/2512#-MjhYW2H8vwDNZGuJXIv:-MjhYW2H8vwDNZGuJXIw:bglnt03) ([raw file](https://github.com/ethersphere/bee/blob/48e06795c7dd382d6c54139a5262bca98785c806/pkg/node/node.go#L525)):*
> ```Go
> 	var protectedNodes []swarm.Address
> 
> 	for _, p := range o.ProtectedNodes {
> ```

we should probably do this parsing part before calling `NewBee` and just have the slice of `swarm.Address` passed on `Options`. Also, I would return the error in case there was a parsing error. If the provided address isn't correct it might cause us not to list some nodes as static which is something we shouldn't be permissive about
___
*[pkg/topology/kademlia/kademlia.go, line 893 at r1](https://reviewable.io/reviews/ethersphere/bee/2512#-MjhZ76i0mMXC2Yzl2Hp:-MjhZ76i0mMXC2Yzl2Hq:b-tmnsuw) ([raw file](https://github.com/ethersphere/bee/blob/48e06795c7dd382d6c54139a5262bca98785c806/pkg/topology/kademlia/kademlia.go#L893)):*
> ```Go
> 	if _, overSaturated := k.saturationFunc(po, k.knownPeers, k.connectedPeers); overSaturated {
> 		if k.bootnode {
> 			for i := 0; i < bootNodeOverSaturationPeers; i++ {
> ```

I find this loop a bit confusing. Maybe it would be clearer to just change the `randomPeer` to do what we want here? which is:
- take all of the bin peers
- remove all static overlays from the set
- pick a random one
___
*[pkg/topology/kademlia/kademlia_test.go, line 1290 at r1](https://reviewable.io/reviews/ethersphere/bee/2512#-Mjh_5Cg4fYgF-0xEcM5:-Mjh_5Ch502cfzxlfFml:bmg6ocv) ([raw file](https://github.com/ethersphere/bee/blob/48e06795c7dd382d6c54139a5262bca98785c806/pkg/topology/kademlia/kademlia_test.go#L1290)):*
> ```Go
> 	*kademlia.BootnodeOverSaturationPeers = 1
> 
> 	defer func(p int) {
> ```

no need for 4 defer statements, can all be condensed into one closure

aloknerurkar:
___
*[cmd/bee/cmd/start.go, line 141 at r1](https://reviewable.io/reviews/ethersphere/bee/2512#-MjhXtqZ2TAydF2l7uCm:-Mjhl6grAQeutUZP6snd:b-5vw0mz) ([raw file](https://github.com/ethersphere/bee/blob/48e06795c7dd382d6c54139a5262bca98785c806/cmd/bee/cmd/start.go#L141)):*

we might want to apply this for other nodes too
</blockquote></details>

In the current flow, the kicking-out behavior is only applicable for bootnodes. Which is why I added the check. This then becomes wrong configuration from user which is why I am returning error.

aloknerurkar (commented):
> Dismissed @acud from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2512)* status: all files reviewed, 1 unresolved discussion (waiting on @acud and @mrekucci)

___
*[pkg/node/node.go, line 525 at r1](https://reviewable.io/reviews/ethersphere/bee/2512#-MjhYW2H8vwDNZGuJXIv:-MjhmiDx9iSFWKIzdABu:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/48e06795c7dd382d6c54139a5262bca98785c806/pkg/node/node.go#L525)):*

we should probably do this parsing part before calling `NewBee` and just have the slice of `swarm.Address` passed on `Options`. Also, I would return the error in case there was a parsing error. If the provided address isn't correct it might cause us not to list some nodes as static which is something we shouldn't be permissive about
</blockquote></details>

Done.
___
*[pkg/topology/kademlia/kademlia.go, line 893 at r1](https://reviewable.io/reviews/ethersphere/bee/2512#-MjhZ76i0mMXC2Yzl2Hp:-MjhlZzzAAaBMRVTHkMz:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/48e06795c7dd382d6c54139a5262bca98785c806/pkg/topology/kademlia/kademlia.go#L893)):*

I find this loop a bit confusing. Maybe it would be clearer to just change the `randomPeer` to do what we want here? which is:
- take all of the bin peers
- remove all static overlays from the set
- pick a random one
</blockquote></details>

Done.
___
*[pkg/topology/kademlia/kademlia_test.go, line 1290 at r1](https://reviewable.io/reviews/ethersphere/bee/2512#-Mjh_5Cg4fYgF-0xEcM5:-MjhlaP26xW1dyjQuXuZ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/48e06795c7dd382d6c54139a5262bca98785c806/pkg/topology/kademlia/kademlia_test.go#L1290)):*

no need for 4 defer statements, can all be condensed into one closure
</blockquote></details>

Done.

acud (changes requested):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2512)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar and @mrekucci)

___
*[pkg/topology/kademlia/kademlia.go, line 1356 at r2](https://reviewable.io/reviews/ethersphere/bee/2512#-MjhoifADa38mSD3dJfm:-MjhoifADa38mSD3dJfn:b-g2q2i8) ([raw file](https://github.com/ethersphere/bee/blob/f1522df649c3a258a035bd366e304c1e220783a1/pkg/topology/kademlia/kademlia.go#L1356)):*
> ```Go
> 		// do not consider protected peers
> 		if k.isProtected(p) {
> 			peers = append(peers[:idx], peers[idx+1:]...)
> ```

not sure how this plays along with the fact you're inside the for loop that iterates over the same slice you're modifying, but this is about the implementation of the `range` statement. You might want to `for ..., i<len(peers), i++ {} ` manually to make this clearer, but then you'll have to `i++` only in case you did not remove an element (since once you remove an element, you basically need to inspect the same index next time you iterate)

acud:
also, test is flaking:
```
=== RUN   TestBootnodeProtectedNodes
    kademlia_test.go:1336: expected error oversaturated , got empty bin
--- FAIL: TestBootnodeProtectedNodes (0.08s)
```

aloknerurkar:
___
*[pkg/topology/kademlia/kademlia.go, line 1356 at r2](https://reviewable.io/reviews/ethersphere/bee/2512#-MjhoifADa38mSD3dJfm:-MjhsHuk9kx0O7LdOEvg:b-6g8tz9) ([raw file](https://github.com/ethersphere/bee/blob/f1522df649c3a258a035bd366e304c1e220783a1/pkg/topology/kademlia/kademlia.go#L1356)):*

not sure how this plays along with the fact you're inside the for loop that iterates over the same slice you're modifying, but this is about the implementation of the `range` statement. You might want to `for ..., i<len(peers), i++ {} ` manually to make this clearer, but then you'll have to `i++` only in case you did not remove an element (since once you remove an element, you basically need to inspect the same index next time you iterate)
</blockquote></details>

Yes, you are right, this can definitely cause issues. Fixed.

acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2512)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar and @mrekucci)

___
*[pkg/topology/kademlia/kademlia.go, line 896 at r3](https://reviewable.io/reviews/ethersphere/bee/2512#-Mji2x7s-Zky9MDw8QFx:-Mji2x7t7tgQCIdNf5d9:bszcept) ([raw file](https://github.com/ethersphere/bee/blob/40eecb2d2c9a8f1ef9cb0b53a09ee8f6efc67e8e/pkg/topology/kademlia/kademlia.go#L896)):*
> ```Go
> 			randPeer, err := k.randomPeer(po)
> 			if err != nil {
> 				return fmt.Errorf("failed to get random peer to kick-out: %w", topology.ErrOversaturated)
> ```

i think that if the bin is empty for whatever reason, this should constitute as an error, we still want the peer no?

mrekucci (approved):
> Reviewed 1 of 6 files at r1, 4 of 5 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2512)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar)

___
*[cmd/bee/cmd/start.go, line 146 at r3](https://reviewable.io/reviews/ethersphere/bee/2512#-Mji9y1q3BI1yLR4QN6G:-Mji9y1r62Pmf-Qh9LPS:b-ixt95t) ([raw file](https://github.com/ethersphere/bee/blob/40eecb2d2c9a8f1ef9cb0b53a09ee8f6efc67e8e/cmd/bee/cmd/start.go#L146)):*
> ```Go
> 				addr, err := swarm.ParseHexAddress(p)
> 				if err != nil {
> 					return errors.New("invalid swarm address configured for static node")
> ```

It would be helpful to also print the invalid address.

acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2512)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar)

___
*[pkg/topology/kademlia/kademlia.go, line 896 at r3](https://reviewable.io/reviews/ethersphere/bee/2512#-Mji2x7s-Zky9MDw8QFx:-MjiFYjFC5Q2m4g7kTON:bqys8up) ([raw file](https://github.com/ethersphere/bee/blob/40eecb2d2c9a8f1ef9cb0b53a09ee8f6efc67e8e/pkg/topology/kademlia/kademlia.go#L896)):*

i think that if the bin is empty for whatever reason, this should constitute as an error, we still want the peer no?
</blockquote></details>

also, it is misleading that an error of oversaturation is being returned here...

aloknerurkar:
___
*[pkg/topology/kademlia/kademlia.go, line 896 at r3](https://reviewable.io/reviews/ethersphere/bee/2512#-Mji2x7s-Zky9MDw8QFx:-MjiMY3h10j0927HeLoe:b29fqg0) ([raw file](https://github.com/ethersphere/bee/blob/40eecb2d2c9a8f1ef9cb0b53a09ee8f6efc67e8e/pkg/topology/kademlia/kademlia.go#L896)):*

also, it is misleading that an error of oversaturation is being returned here...
</blockquote></details>

We enter this loop only if the bin is overSaturated. The only reason why randomPeer would return empty bin error is if all the nodes are protected. This still means the bin is oversaturated, which is why I dont use the emptyBin error. Actually I was not sure why we would get empty bin here in the first place?

aloknerurkar:
___
*[pkg/topology/kademlia/kademlia.go, line 896 at r3](https://reviewable.io/reviews/ethersphere/bee/2512#-Mji2x7s-Zky9MDw8QFx:-Mjj7ZXpAtXRnMYLwphN:bslfisw) ([raw file](https://github.com/ethersphere/bee/blob/40eecb2d2c9a8f1ef9cb0b53a09ee8f6efc67e8e/pkg/topology/kademlia/kademlia.go#L896)):*

so... i am trying to think about some edge cases... imagine we have by coincidence few static nodes in one bin, it could be (even in the test) that you end up with no peers to remove. it is a bit of an edge case (it would require that the bin is oversaturated with static nodes). not completely unlikely but would require us to have a lot of bootnodes
</blockquote></details>

but then even in this case, I think returning oversaturated error makes more sense as randomPeer would fail with empty bin when we have saturated bin with protected nodes right?

acud (changes requested):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2512)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar)

___
*[pkg/topology/kademlia/kademlia.go, line 896 at r3](https://reviewable.io/reviews/ethersphere/bee/2512#-Mji2x7s-Zky9MDw8QFx:-MjjJwKU33CNCHdbxYwU:bpcqhso) ([raw file](https://github.com/ethersphere/bee/blob/40eecb2d2c9a8f1ef9cb0b53a09ee8f6efc67e8e/pkg/topology/kademlia/kademlia.go#L896)):*

but then even in this case, I think returning oversaturated error makes more sense as randomPeer would fail with empty bin when we have saturated bin with protected nodes right?
</blockquote></details>

well depends... lets assume that the bin is full of only `oversaturationPeers` bootnodes, then you would still like to accept the new peer since we don't want to account for the bootnodes...

acud (approved):
:lgtm:

> Reviewed 1 of 1 files at r5, 2 of 3 files at r6, 1 of 1 files at r7, 4 of 4 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2512#-:-Mk2za6M5TmWTRUB8vFU:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar)

aloknerurkar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2512)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aloknerurkar)

___
*[pkg/topology/kademlia/kademlia.go, line 896 at r3](https://reviewable.io/reviews/ethersphere/bee/2512#-Mji2x7s-Zky9MDw8QFx:-Mk30yeg6TqtAyt5L9GD:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/40eecb2d2c9a8f1ef9cb0b53a09ee8f6efc67e8e/pkg/topology/kademlia/kademlia.go#L896)):*

well depends... lets assume that the bin is full of only `oversaturationPeers` bootnodes, then you would still like to accept the new peer since we don't want to account for the bootnodes...
</blockquote></details>

Done.

PR #2520, opened 2021-09-16
refactor(pushsync): better tracing

acud:
this PR adds traces for replication (on the origin and target nodes) and moves the handler span creation to the right place

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2520)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -141,6 +141,10 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 
 	chunk := swarm.NewChunk(swarm.NewAddress(ch.Address), ch.Data)
 	chunkAddress := chunk.Address()
+
+	span, _, ctx := ps.tracer.StartSpanFromContext(ctx, "pushsync-handler", ps.logger, opentracing.Tag{Key: "address", Value: chunkAddress.String()})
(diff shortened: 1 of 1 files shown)


pkg/pushsync/pushsync.go
```diff
@@ -480,9 +484,18 @@ func (ps *PushSync) pushToNeighbour(peer swarm.Address, ch swarm.Chunk, origin b
 	// price for neighborhood replication
 	receiptPrice := ps.pricer.PeerPrice(peer, ch.Address())
 
+	// decouple the span data from the original context so it doesn't get
+	// cancelled, then glue the stuff on the new context
+	span := tracing.FromContext(ctx)
+
 	ctx, cancel := context.WithTimeout(context.Background(), timeToWaitForPushsyncToNeighbor)
```
janos, line 491: Since ctx is now passed to pushToNeighbour, could here context.Background() be replaced with ctx?

janos (approved):
LGTM. One small comment on possible additional improvement.

pkg/pushsync/pushsync.go
acud: I'm passing it just so I can extract the OT span info out of it. The reason we're using context.Background is so that the parent context doesn't expire while we do the replication

mrekucci (approved):

PR #2522, opened 2021-09-18
feat(kademlia): peer latencies

acud:
This PR adds the pingpong protocol to kademlia as a dependency, such that the latter uses the former to record peer latencies with EWMA smoothing. Tested to be working on the mainnet nicely


notanatol (approved):

janos (approved):
LGTM

pkg/topology/kademlia/kademlia.go
```diff
@@ -437,6 +441,25 @@ func (k *Kad) manage() {
...
+	k.wg.Add(1)
+	go func() {
+		defer k.wg.Done()
+		for {
+			select {
+			case <-k.halt:
+				return
+			case <-k.quit:
+				return
+			case <-time.After(1 * time.Second):
```
istae, line 453: this might be a bit aggressive, maybe wait longer here?

mrekucci (approved):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2522)* status: all files reviewed, 1 unresolved discussion (waiting on @acud)

___
*[pkg/topology/kademlia/kademlia.go, line 1403 at r1](https://reviewable.io/reviews/ethersphere/bee/2522#-Mk1N7sD-8WzGPQbXCTT:-Mk1N7sEDLUwDXybmrIF:b-bj4e0q) ([raw file](https://github.com/ethersphere/bee/blob/9edd87993d089d8020f9539250f82055f7545061/pkg/topology/kademlia/kademlia.go#L1403)):*
> ```Go
> // createMetricsSnapshotView creates new topology.MetricSnapshotView from the
> // given metrics.Snapshot and rounds all the timestamps and durations to its
> // nearest second.
> ```

Since the `LatencyEWMA` is in milliseconds this description should be adjusted accordingly.

pkg/topology/kademlia/kademlia.go
acud: since the metric is averaged, it takes a while for it to actually add up and show up in the measurement, and if the peer has a high latency it will anyway take a long time until it is actually reflected in the smoothed metric (it has a delay), so if we poll more slowly the delay will become much more significant (it would be a few seconds as it is, which is already quite slow). also, the actual pinging is not a costly operation (it is just a couple of kilobytes for 200 peers) also on the downstream peer, so I think it is ok as is

PR #2524, opened 2021-09-20
fix: continue forwarding if replication fails

istae:
Forwarding nodes that replicate should be warmed up and, in the case of a replication error, should continue forwarding it.

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2524)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -218,14 +218,13 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 		chunk, err = ps.validStamp(chunk, ch.Stamp)
 		if err != nil {
 			ps.metrics.InvalidStampErrors.Inc()
-			return fmt.Errorf("pushsync valid stamp: %w", err)
-		}
-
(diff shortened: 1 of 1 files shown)


pkg/pushsync/pushsync.go
```diff
@@ -213,19 +213,18 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 
 	// forwarding replication
 	storedChunk := false
-	if ps.topologyDriver.IsWithinDepth(chunkAddress) {
+	if ps.warmedUp() && ps.topologyDriver.IsWithinDepth(chunkAddress) {
```
acud, line 216: where did the `warmedUp` come from? it is used to mitigate sending back receipts when the node has not yet warmed up and I think it is not relevant in this context

istae: checking depth depends on the kademlia topology of the node. If it has not warmed up, the node may accept chunks that are not normally in it's neighborhood.

acud (approved):
LGTM

mrekucci (approved):

PR #2526, opened 2021-09-20
fix: removed neighborhood check for replication

istae:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2526)


acud (changes requested):

pkg/pushsync/pushsync_test.go
```diff
@@ -211,102 +211,6 @@ func TestReplicateBeforeReceipt(t *testing.T) {
 	}
 }
 
-func TestFailToReplicateBeforeReceipt(t *testing.T) {
```
acud, line 214: why is the test removed? we still do replication no?

istae: This test is the same as one before it, the only difference is that this test checks that the replication fails because the replication node is out of depth. Since this requirement no longer exists, this test is invalid.

acud (approved):

mrekucci (approved):
> Reviewed 1 of 2 files at r3, 2 of 3 files at r4, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2526)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @notanatol)

PR #2527, opened 2021-09-20
refactor: closer func

istae:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2527)

### DIFF
--- a/pkg/swarm/swarm.go
+++ b/pkg/swarm/swarm.go
@@ -119,6 +119,15 @@ func (a Address) MarshalJSON() ([]byte, error) {
 	return json.Marshal(a.String())
 }
 
+// Closer returns if x is closer to a than y
+func (x Address) Closer(a Address, y Address) bool {
+	cmp, err := DistanceCmp(a.b, x.b, y.b)
+	if err != nil {
+		return false
+	}
+	return cmp == 1
+}
+
 // ZeroAddress is the address that has no value.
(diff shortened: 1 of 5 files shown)


pkg/swarm/swarm.go
```diff
@@ -119,6 +119,15 @@ func (a Address) MarshalJSON() ([]byte, error) {
 	return json.Marshal(a.String())
 }
 
+// Closer returns if x is closer to a than y
+func (x Address) Closer(a Address, y Address) bool {
+	cmp, err := DistanceCmp(a.b, x.b, y.b)
+	if err != nil {
+		return false
```
acud, line 126: i wouldn't swallow the error here. it assumes that x is not closer to y and that is not the correct assumption IMO. it would lead to flow control errors. better to propagate

pkg/swarm/swarm.go
```diff
@@ -119,6 +119,15 @@ func (a Address) MarshalJSON() ([]byte, error) {
 	return json.Marshal(a.String())
 }
 
+// Closer returns if x is closer to a than y
+func (x Address) Closer(a Address, y Address) bool {
```
acud, line 123: I think that if you add this function, it would be good to sweep through the entire codebase and apply this, instead of just in one place

acud (approved):

mrekucci (approved):

PR #2530, opened 2021-09-21
feat: kademlia oversaturation based on an exponential decay formula

istae:
Bin oversaturation count is calculated using an exponential decay formula.
The goal is to give more weight to shallower bins (relative to how they are now) as they make up the majority of known peers.
This should improve connectivity and decrease hops in syncing protocols.
 

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2530)

### DIFF
--- a/pkg/topology/kademlia/kademlia.go
+++ b/pkg/topology/kademlia/kademlia.go
@@ -138,9 +138,7 @@ func New(
 ) (*Kad, error) {
 
 	if o.OverSaturationCalc == nil {
-		o.OverSaturationCalc = func(bin uint8) int {
-			return int(math.Exp(-float64(bin)/8.0)) * 50
-		}
+		o.OverSaturationCalc = defaultOverSaturationCalc
 	}
 
 	o.SaturationFunc = binSaturated(o.OverSaturationCalc, isStaticPeer(o.StaticNodes))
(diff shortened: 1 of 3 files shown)


pkg/topology/kademlia/kademlia.go
```diff
@@ -118,17 +122,20 @@ func New(
...
+
+	o.SaturationFunc = binSaturated(o.OverSaturationCalc, isStaticPeer(o.StaticNodes))
+
 	if o.BitSuffixLength == 0 {
 		o.BitSuffixLength = defaultBitSuffixLength
 	}
 
+	for i := 0; i < int(swarm.MaxBins); i++ {
+	}
+
```
ldeffenb, line 138: An empty loop?  To what purpose?

pkg/topology/kademlia/kademlia.go
```diff
@@ -173,6 +181,12 @@ func New(
 	return k, nil
 }
 
+// produces oversaturation values using an exponential decay formula, sequence is as follows:
+// bin 0 -> 50, 44, 39, 34, 30, 27
+func defaultOverSaturationCalc(bin uint8) int {
+	return int(math.Round(50 * math.Exp(-float64(bin)/8.0)))
+}
+
```
ldeffenb, line 189: Should the 50 and the 8.0 be constants declared above?  Otherwise this completely hides the actual bin limits inside this function.

pkg/topology/kademlia/kademlia.go
istae: hence the 'in progress' label :) thank you!

PR #2535, opened 2021-09-22
feat: multiple retries at each hop

istae:
PR introduces retries at each hop for the pushsync protocol.

Peer-chunk skip-listing by @acud prevents repeating network traversals so we can have this feature again.

One benefit, which is counterintuitive, is that this change theoretically **reduces** the total number of hops to the closest peer.

The reason is the protocol allows for multiple deeper depth-first searches at each hop where a successful traversal may be discovered at some second or third attempt starting from a node far away from the origin node.

Without retries after the origin node, if a traversal returns with error, the next traversal has to start beginning at the origin which means all the previous traversal progress has been lost.

This was proved to be theoretically true using the bee-sync-sim tool.

image

In the example above, without retries, reaching the destination node would've required three full traversals starting from the origin. Instead, the node '1b944a' retries with his own peers and finds the destination node. '1b91da' is also aware of closer nodes but because those retries fail, he becomes the storer node.

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2535)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -38,8 +38,8 @@ const (
 )
 
 const (
-	maxPeers    = 3
-	maxAttempts = 16
+	maxPeers               = 3
+	maxAttemptsWithoutPush = 16
 )
 
 var (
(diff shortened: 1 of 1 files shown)


acud:
nice @istae, i only have one concern about this: retries are problematic because they don't take into account the context expiration. so the problem is that when you retry the contexts that are supposed to allow the receipt to come back up to the uploader expire, not allowing the receipt to propagate back up again. this has proven to be very problematic in the past because then receipts cannot propagate and originators keep trying to upload a chunk that has already been pushed before

PR #2537, opened 2021-09-23
fix: fix swap migraton state store corruption

ralph-pichler:
prevents the corruption of the swarm peer address and already migrated keys during the migration. Note that this does not help nodes which already ran the migration.

* both `swarm.Address` and `common.Address` are treated as string in their state store representation, so just loading and storing preserves the correct value.
* addresses which are not length 20 strings are assumed to be already migrated
* in case of a conflict, the newer value takes precedence

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2537)

### DIFF
--- a/pkg/statestore/leveldb/migration.go
+++ b/pkg/statestore/leveldb/migration.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethersphere/bee/pkg/storage"
 )
 
 var (
@@ -143,6 +144,9 @@ func migrateSwap(s *store) error {
 					return err
 				}
 				continue
+			}
+			if !errors.Is(err, storage.ErrNotFound) {
+				return err
 			}
 
 			if err = s.Get(key, &val); err != nil {
--- a/pkg/statestore/leveldb/migration.go
+++ b/pkg/statestore/leveldb/migration.go
@@ -129,7 +129,7 @@ func migrateSwap(s *store) error {
 			}
 
 			if len(split[1]) != 20 {
-				s.logger.Debugf("skipping already migrated key %s", string(key))
+				s.logger.Debugf("skipping already migrated key %s", key)
 				continue
 			}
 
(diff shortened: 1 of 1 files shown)


acud (approved):

notanatol (approved):

mrekucci (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2537)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @ralph-pichler)

___
*[pkg/statestore/leveldb/migration.go, line 140 at r2](https://reviewable.io/reviews/ethersphere/bee/2537#-MkH1YXi9aCGzKKls06K:-MkH1YXi9aCGzKKls06L:by2jwg3) ([raw file](https://github.com/ethersphere/bee/blob/6bb202a47709a2df8ffe21706cb6f2938ada81a5/pkg/statestore/leveldb/migration.go#L140)):*
> ```Go
> 
> 			var val string
> 			if err = s.Get(fixed, &val); err == nil {
> ```

Do we want to ignore all errors? If so, would be nice to leave a comment here on why.

ralph-pichler (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2537)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mrekucci)

___
*[pkg/statestore/leveldb/migration.go, line 140 at r2](https://reviewable.io/reviews/ethersphere/bee/2537#-MkH1YXi9aCGzKKls06K:-MkH8Ttj6T8KCcOpvc-r:b-n82qls) ([raw file](https://github.com/ethersphere/bee/blob/6bb202a47709a2df8ffe21706cb6f2938ada81a5/pkg/statestore/leveldb/migration.go#L140)):*

Do we want to ignore all errors? If so, would be nice to leave a comment here on why.
</blockquote></details>

we want to ignore only errnotfound. I added a check for other errors.

mrekucci (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2537)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ralph-pichler)

Issue #1439, opened 2021-03-16
Internal network configuration leaking


mfw78:
# Leaking Network Configuration

When `bee` establishes connections, it leaks internal networking configuration (IP address, port) that some cases, may not be needed. It's been observed on `bee` node logs that other bee nodes are advertising private IP addresses that are non-routable, causing network overhead, as well as leaking potentially private information / network configuration.

# Suggested Fix

It's suggested that logic be applied to the underlay addresses _exported_ such that:

1. Blacklist all [private CIDRs](https://en.wikipedia.org/wiki/IPv4#Special-use_addresses) by default.
2. Send private CIDRs only to whitelisted nodes (such as bootstrap nodes?). This would assist in the event of running multiple nodes on a Kubernetes cluster, allowing them to communicate amongst one another without the need for NAT reflection firewall rules.

Eknir:
cc @vandot

DigiDr:
+1 for this. I've had a cloud provider disconnect a box for persistent attempts to connect to private ip ranges. In my case, docker makes this challenging to resolve with iptables.

janos:
I would suggest to have a configuration option to enable advertisement of private CIDRs (current behaviour) and by default not to advertise them. There are cases where a private swarm network is created and they operate only in the private CIDRs, such are testing environments and organisations' private networks.

Skjold:
+1 more for this.
I've temporarily solved the issue by applying the settings from https://github.com/chaifeng/ufw-docker and then doing:
`curl --retry 20 --retry-delay 1 -o bogons https://www.team-cymru.org/Services/Bogons/bogon-bn-agg.txt
while read -r bogon; do
    ufw route deny from any to "$bogon";
done < bogons
rm bogons`
It does work to cut all Docker traffic to bogon addresses, you just need to insert your own `ufw route allow` rules before these ones to explicitly allow outgoing connections to any RFC1918 networks you actually use.
It would be nice to also log a warning for users to set the nat-address in bee.yaml or as an environment variable if Bee is advertising a bogon IP.

mrekucci:
Fixed in #2538

mrekucci:
Fixed in #2538

PR #2538, opened 2021-09-23
fix: avoid advertising private IP addresses to the public network

mrekucci:
related to #1439 

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2538)

### DIFF
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -30,6 +30,7 @@ import (
 	"github.com/ethersphere/bee/pkg/swarm"
 	lru "github.com/hashicorp/golang-lru"
 	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


janos (approved):
Nice. It would be good to have a test to ensure that the feature is not broken in the future by accident.

aloknerurkar (approved):

pkg/hive/hive.go
```diff
@@ -187,6 +194,11 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
 			return err
 		}
 
+		// Don't advertise private addresses to the public network.
+		if isPeerPublic && manet.IsPrivateAddr(addr.Underlay) {
```
istae, line 198: maybe it's better to have this check under `BroadcastPeers` so that you can send full batches instead of skipping some here.

istae: can we also have this check before we attempt to ping the address under `checkAndAddPeers`

acud (changes requested):
@mrekucci this might be a bit tricky... i'm not sure what's the cluster configuration in ci flow or in the large cluster tests. since private CIDRs may be used to advertise the nodes in these cases, the cluster might become completely defunct. we need to expose this via a CLI flag probably, in order to allow this to work in larger test networks. the reason that the beekeeper integration test is passing, is probably because all nodes explicitly dial to the bootnode on an internal network, and the cluster size is small, so full connectivity is preserved and no gossiping is needed

mrekucci:
> Nice. It would be good to have a test to ensure that the feature is not broken in the future by accident.

Done.

mrekucci:
> @mrekucci this might be a bit tricky... i'm not sure what's the cluster configuration in ci flow or in the large cluster tests. since private CIDRs may be used to advertise the nodes in these cases, the cluster might become completely defunct. we need to expose this via a CLI flag probably, in order to allow this to work in larger test networks. the reason that the beekeeper integration test is passing, is probably because all nodes explicitly dial to the bootnode on an internal network, and the cluster size is small, so full connectivity is preserved and no gossiping is needed

Done.

janos (approved):
Great!. LGTM.

acud (approved):
Nice! We need this flag to be added also to the helm chart and to beekeeper before this gets merged

PR #2541, opened 2021-09-26
opt: handshake handle checks with kademlia pick before overlay verification

istae:
Oversaturated bins reject new incoming peers, so in the life time of a node, after the warmup period, almost all of connection attempts are rejected. 
The PR moves the Pick check before the overlay verification step. As a result, the number of eth backend calls are drastically reduced.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2541)

### DIFF
--- a/pkg/p2p/libp2p/internal/handshake/handshake.go
+++ b/pkg/p2p/libp2p/internal/handshake/handshake.go
@@ -54,7 +54,7 @@ var (
 	ErrWelcomeMessageLength = fmt.Errorf("handshake welcome message longer than maximum of %d characters", MaxWelcomeMessageLength)
 
 	// ErrPicker is returned if the picker (kademlia) rejects the peer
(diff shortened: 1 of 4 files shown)


acud (commented):
Nice. It would also be nice to have unit tests for this inside this package, since the logic moved between packages (and it is tested in the previous package)

acud (approved):

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -307,6 +315,12 @@ func (s *Service) Handle(ctx context.Context, stream p2p.Stream, remoteMultiaddr
 
 	overlay := swarm.NewAddress(ack.Address.Overlay)
 
+	if s.picker != nil {
```
aloknerurkar, line 318: Maybe we can remove the extra call to Pick in `handleIncoming`?

aloknerurkar (approved):

PR #2564, opened 2021-10-04
feat: update reachability state of node

aloknerurkar:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2564)

### DIFF
--- a/pkg/p2p/p2p.go
+++ b/pkg/p2p/p2p.go
@@ -49,6 +49,7 @@ type Halter interface {
 type PickyNotifier interface {
 	Picker
 	Notifier
+	ReachabilityTracker
 }
 
 type Picker interface {
@@ -62,6 +63,26 @@ type ReachableNotifier interface {
 type Reacher interface {
 	Connected(swarm.Address, ma.Multiaddr)
 	Disconnected(swarm.Address)
+}
+
+type Reachability int
+
+const (
+	Unknown Reachability = iota
+	Private
+	Public
+)
+
+func (r Reachability) String() string {
+	str := [...]string{"Unknown", "Public", "Private"}
+	if r < 0 || int(r) >= len(str) {
+		return "(unrecognized)"
+	}
+	return str[r]
+}
+
+type ReachabilityTracker interface {
+	UpdateReachability(Reachability)
 }
 
 type Notifier interface {
--- a/pkg/p2p/p2p.go
+++ b/pkg/p2p/p2p.go
@@ -82,7 +82,7 @@ func (r Reachability) String() string {
 }
 
 type ReachabilityTracker interface {
-	UpdateReachability(Reachability)
+	UpdateReachability(string)
 }
 
 type Notifier interface {
--- a/pkg/p2p/p2p.go
+++ b/pkg/p2p/p2p.go
@@ -80,22 +80,6 @@ type Reacher interface {
 	Disconnected(swarm.Address)
 }
 
-type Reachability int
-
-const (
-	Unknown Reachability = iota
-	Private
-	Public
-)
-
-func (r Reachability) String() string {
-	str := [...]string{"Unknown", "Public", "Private"}
-	if r < 0 || int(r) >= len(str) {
-		return "(unrecognized)"
-	}
-	return str[r]
-}
-
 type ReachabilityUpdater interface {
 	UpdateReachability(ReachabilityStatus)
 }
--- a/pkg/p2p/p2p.go
+++ b/pkg/p2p/p2p.go
@@ -49,7 +49,7 @@ type Halter interface {
 type PickyNotifier interface {
 	Picker
 	Notifier
-	ReachabilityTracker
+	ReachabilityUpdater
 }
 
 type Picker interface {
@@ -81,7 +81,7 @@ func (r Reachability) String() string {
 	return str[r]
 }
 
-type ReachabilityTracker interface {
+type ReachabilityUpdater interface {
 	UpdateReachability(string) error
 }
 
--- a/pkg/p2p/p2p.go
+++ b/pkg/p2p/p2p.go
@@ -8,12 +8,52 @@ package p2p
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"time"
 
 	"github.com/ethersphere/bee/pkg/bzz"
 	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/libp2p/go-libp2p-core/network"
 	ma "github.com/multiformats/go-multiaddr"
+)
+
+// ReachabilityStatus represents the node reachability status.
+type ReachabilityStatus int
+
+// String implements the fmt.Stringer interface.
+func (rs ReachabilityStatus) String() string {
+	str := [...]string{reachabilityUnknown, reachabilityPublic, reachabilityPrivate}
+	if rs < 0 || int(rs) >= len(str) {
+		return "(unrecognized)"
+	}
+	return str[rs]
+}
+
+// ParseReachabilityStatus tries to parse reachability status from the given string.
+func ParseReachabilityStatus(s string) (ReachabilityStatus, error) {
+	switch s {
+	case reachabilityUnknown:
+		return ReachabilityStatusUnknown, nil
+	case reachabilityPublic:
+		return ReachabilityStatusPublic, nil
+	case reachabilityPrivate:
+		return ReachabilityStatusPrivate, nil
+	}
+	return -1, fmt.Errorf("unrecognized reachability status: %q", s)
+}
+
+const (
+	ReachabilityStatusUnknown = ReachabilityStatus(network.ReachabilityUnknown)
+	ReachabilityStatusPublic  = ReachabilityStatus(network.ReachabilityPublic)
+	ReachabilityStatusPrivate = ReachabilityStatus(network.ReachabilityPrivate)
+)
+
+var (
+	// String representations of the ReachabilityStatus.
+	reachabilityUnknown = network.ReachabilityUnknown.String()
+	reachabilityPublic  = network.ReachabilityPublic.String()
+	reachabilityPrivate = network.ReachabilityPrivate.String()
 )
 
 // Service provides methods to handle p2p Peers and Protocols.
@@ -82,7 +122,7 @@ func (r Reachability) String() string {
 }
 
 type ReachabilityUpdater interface {
-	UpdateReachability(string) error
+	UpdateReachability(ReachabilityStatus)
 }
 
 type Notifier interface {
--- a/pkg/p2p/p2p.go
+++ b/pkg/p2p/p2p.go
@@ -82,7 +82,7 @@ func (r Reachability) String() string {
 }
 
 type ReachabilityTracker interface {
-	UpdateReachability(string)
+	UpdateReachability(string) error
 }
 
 type Notifier interface {
--- a/pkg/p2p/p2p.go
+++ b/pkg/p2p/p2p.go
@@ -8,7 +8,6 @@ package p2p
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"time"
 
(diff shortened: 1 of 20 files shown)


pkg/topology/kademlia/internal/metrics/metrics.go
```diff
@@ -15,17 +15,27 @@ import (
...
 	PeerConnectionDirectionInbound  PeerConnectionDirection = "inbound"
 	PeerConnectionDirectionOutbound PeerConnectionDirection = "outbound"
+)
+
+// PeerReachabilityStatus represents the peer reachability status.
+type PeerReachabilityStatus string
 
-	ewmaSmoothing = 0.1
+var (
+	PeerReachabilityStatusUnknown = PeerReachabilityStatus(network.ReachabilityUnknown.String())
```
aloknerurkar, line 36: Not sure if we want to have dependency on `go-libp2p-core/network` in `kademlia`

It would be better to use the actual string values here and mention these are directly related to the libp2p network reachability

mrekucci: Fixed.

pkg/topology/kademlia/internal/metrics/metrics.go
```diff
@@ -18,14 +18,50 @@ import (
...
+	case reachabilityPublic:
+		return PeerReachabilityStatusPublic, nil
+	case reachabilityPrivate:
+		return PeerReachabilityStatusPrivate, nil
+	}
+	return -1, fmt.Errorf("unrecognized reachability status: %q", s)
+}
+
+const (
+	PeerReachabilityStatusUnknown PeerReachabilityStatus = 0 // Mirrors the network.ReachabilityUnknown.
```
acud, line 57: have a look at https://github.com/ethersphere/bee/blob/master/pkg/storage/store.go#L26 as an example of a bit more simplified use case (using `iota` keyword). consider having the method on the status type but perhaps named `UnmarshalText`. also, consider defining the type inside the `p2p` package (where all the p2p contracts are defined) or in the `topology` package

pkg/topology/kademlia/internal/metrics/metrics_test.go
```diff
@@ -130,13 +130,13 @@ func TestPeerMetricsCollector(t *testing.T) {
...
+	if have, want := ss.ReachabilityStatus, metrics.PeerReachabilityStatusUnknown; have != want {
+		t.Fatalf("Snapshot(%q, ...): is eachability status mismatch: have %q; want %q", addr, have, want)
 	}
-	mc.Record(addr, metrics.PeerReachability(true))
+	mc.Record(addr, metrics.PeerReachability(metrics.PeerReachabilityStatusPublic))
 	ss = snapshot(t, mc, t2, addr)
-	if have, want := ss.IsReachable, true; have != want {
-		t.Fatalf("Snapshot(%q, ...): is reachable status mismatch: have %t; want %t", addr, have, want)
+	if have, want := ss.ReachabilityStatus, metrics.PeerReachabilityStatusPublic; have != want {
+		t.Fatalf("Snapshot(%q, ...): is eachability status mismatch: have %q; want %q", addr, have, want)
```
acud, line 139: `has reachability`

pkg/topology/kademlia/kademlia.go
```diff
@@ -1203,16 +1203,26 @@ func (k *Kad) EachPeerRev(f topology.EachPeerFunc) error {
...
 			return false, false, nil
 		}
 		return f(addr, po)
 	}
 }
 
 // SetPeerReachability sets the peer reachability status.
-func (k *Kad) SetPeerReachability(addr swarm.Address, isReachable bool) {
-	k.collector.Record(addr, im.PeerReachability(isReachable))
+func (k *Kad) SetPeerReachability(addr swarm.Address, status string) error {
```
acud, line 1214: passing the translated type could simplify things here (signature does not need to return error)

pkg/topology/kademlia/internal/metrics/metrics.go
```diff
@@ -208,7 +244,7 @@ func (cs *Counters) snapshot(t time.Time) *Snapshot {
 		SessionConnectionDuration:  sessionConnDuration,
 		SessionConnectionDirection: cs.sessionConnDirection,
 		LatencyEWMA:                cs.latencyEWMA,
-		IsReachable:                cs.isReachable,
+		ReachabilityStatus:         cs.ReachabilityStatus,
```
acud, line 247: nitpick, for the same of uniformity (like below), maybe `Reachability` is enough?

pkg/topology/kademlia/internal/metrics/metrics.go
```diff
@@ -18,14 +18,50 @@ import (
...
+	case reachabilityPrivate:
+		return PeerReachabilityStatusPrivate, nil
+	}
+	return -1, fmt.Errorf("unrecognized reachability status: %q", s)
+}
+
+const (
+	PeerReachabilityStatusUnknown PeerReachabilityStatus = 0 // Mirrors the network.ReachabilityUnknown.
+	PeerReachabilityStatusPublic  PeerReachabilityStatus = 1 // Mirrors the network.ReachabilityPublic.
+	PeerReachabilityStatusPrivate PeerReachabilityStatus = 2 // Mirrors the network.ReachabilityPrivate.
```
acud, line 59: nitpick, consider removing `Peer` from the type name and then we can repurpose the same type to indicate both the node's reachability or peer reachability

pkg/p2p/p2p.go
```diff
@@ -49,12 +49,17 @@ type Halter interface {
...
 	Picker
 	Notifier
+	ReachabilityTracker
 }
 
 type Picker interface {
 	Pick(Peer) bool
 }
 
+type ReachabilityTracker interface {
```
acud, line 59: nitpicking, but usually single-method interfaces are named like `<function-name>-er`so `Pick()` -> `Picker`. it also helps to find more condensed names :) so maybe `ReachabilityUpdater`?

pkg/topology/kademlia/kademlia.go
```diff
@@ -1203,16 +1203,26 @@ func (k *Kad) EachPeerRev(f topology.EachPeerFunc) error {
...
+	if err != nil {
+		return err
+	}
+	k.collector.Record(addr, im.PeerReachability(rs))
+	return nil
+}
+
+// UpdateReachability updates our own reachability status.
+func (k *Kad) UpdateReachability(status string) error {
+	return k.SetPeerReachability(k.base, status)
```
acud, line 1225: this is also misleading with the `Peer` in the naming convention. Since you're setting the node's reachability status and not other peer's reachability status I would suggest to just keep this as a field on kademlia service, and then also expose this information through the topology endpoint

acud (changes requested):
very nice. might be nice to have a comment that

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2564)* status: 0 of 8 files reviewed, 8 unresolved discussions (waiting on @acud, @aloknerurkar, and @mrekucci)

___
*[pkg/topology/kademlia/kademlia.go, line 1214 at r4](https://reviewable.io/reviews/ethersphere/bee/2564#-MlA2hXiwZyejm9YahxB-r4-1214:-MlAlz-07HYpLHEZQ51t:bmtmoaf) ([raw file](https://github.com/ethersphere/bee/blob/5fd5c12886890285c24ce665c486050a1c9928fc/pkg/topology/kademlia/kademlia.go#L1214)):*

passing the translated type could simplify things here (signature does not need to return error)
</blockquote></details>

Yes, but the `metrics` package is internal so the type cannot be used outside of the `kademlia` package level.
___
*[pkg/topology/kademlia/kademlia.go, line 1225 at r4](https://reviewable.io/reviews/ethersphere/bee/2564#-MlA2hXiwZyejm9YahxB-r4-1225:-MlApXN06VA901dKreDp:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/5fd5c12886890285c24ce665c486050a1c9928fc/pkg/topology/kademlia/kademlia.go#L1225)):*

this is also misleading with the `Peer` in the naming convention. Since you're setting the node's reachability status and not other peer's reachability status I would suggest to just keep this as a field on kademlia service, and then also expose this information through the topology endpoint
</blockquote></details>

Done.
___
*[pkg/topology/kademlia/internal/metrics/metrics.go, line 59 at r4](https://reviewable.io/reviews/ethersphere/bee/2564#-MlA890WPd7TjC4C2ErZ-r4-59:-MlAqFrID62ySREsk1Ql:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/5fd5c12886890285c24ce665c486050a1c9928fc/pkg/topology/kademlia/internal/metrics/metrics.go#L59)):*

nitpick, consider removing `Peer` from the type name and then we can repurpose the same type to indicate both the node's reachability or peer reachability
</blockquote></details>

Done.
___
*[pkg/topology/kademlia/internal/metrics/metrics.go, line 247 at r4](https://reviewable.io/reviews/ethersphere/bee/2564#-MlA890WPd7TjC4C2ErZ-r4-247:-MlAqTyC0qvin-P72XXZ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/5fd5c12886890285c24ce665c486050a1c9928fc/pkg/topology/kademlia/internal/metrics/metrics.go#L247)):*

nitpick, for the same of uniformity (like below), maybe `Reachability` is enough?
</blockquote></details>

Done.
___
*[pkg/topology/kademlia/internal/metrics/metrics_test.go, line 139 at r4](https://reviewable.io/reviews/ethersphere/bee/2564#-MlAQ_TQiO8d71fQAsM2-r4-139:-MlAqcv67_Mi9dTn8JlQ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/5fd5c12886890285c24ce665c486050a1c9928fc/pkg/topology/kademlia/internal/metrics/metrics_test.go#L139)):*

`has reachability`
</blockquote></details>

Done.

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2564)* status: 0 of 9 files reviewed, 8 unresolved discussions (waiting on @acud and @aloknerurkar)

___
*[pkg/topology/kademlia/internal/metrics/metrics.go, line 57 at r4](https://reviewable.io/reviews/ethersphere/bee/2564#-MlA890WPd7TjC4C2ErZ-r4-57:-MlArSoHD0TvJCAeDBmt:b-9hg39r) ([raw file](https://github.com/ethersphere/bee/blob/5fd5c12886890285c24ce665c486050a1c9928fc/pkg/topology/kademlia/internal/metrics/metrics.go#L57)):*

have a look at https://github.com/ethersphere/bee/blob/master/pkg/storage/store.go#L26 as an example of a bit more simplified use case (using `iota` keyword). consider having the method on the status type but perhaps named `UnmarshalText`. also, consider defining the type inside the `p2p` package (where all the p2p contracts are defined) or in the `topology` package
</blockquote></details>

I intentionally didn't want to use `iota` here. We're relying on the `network.Reachability` type so the number must reflect the order of the enums there. If we use also `iota` here and someone unintentionally reorders the enums here, then the statuses wouldn't match.

acud (commented):
> Reviewed 1 of 6 files at r4, 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2564)* status: 7 of 9 files reviewed, 3 unresolved discussions (waiting on @acud, @aloknerurkar, and @mrekucci)

___
*[pkg/topology/kademlia/kademlia.go, line 1214 at r4](https://reviewable.io/reviews/ethersphere/bee/2564#-MlA2hXiwZyejm9YahxB-r4-1214:-MlAxS1w1gJVImU_DUrt:b-zbi8jt) ([raw file](https://github.com/ethersphere/bee/blob/5fd5c12886890285c24ce665c486050a1c9928fc/pkg/topology/kademlia/kademlia.go#L1214)):*

Yes, but the `metrics` package is internal so the type cannot be used outside of the `kademlia` package level.
</blockquote></details>

but it can still use types defined outside, e.g. define it in the `p2p` or `topology` package, no?

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2564)* status: 2 of 9 files reviewed, 3 unresolved discussions (waiting on @acud, @aloknerurkar, and @mrekucci)

___
*[pkg/topology/kademlia/kademlia.go, line 1214 at r4](https://reviewable.io/reviews/ethersphere/bee/2564#-MlA2hXiwZyejm9YahxB-r4-1214:-MlBGqvC0Vr3GM6Xxrf3:bwmg3di) ([raw file](https://github.com/ethersphere/bee/blob/5fd5c12886890285c24ce665c486050a1c9928fc/pkg/topology/kademlia/kademlia.go#L1214)):*

but it can still use types defined outside, e.g. define it in the `p2p` or `topology` package, no?
</blockquote></details>

Ok, I misunderstood this, fixed.

acud (approved):
> Reviewed 1 of 2 files at r3, 1 of 6 files at r4, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2564)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar and @mrekucci)

notanatol (approved):

pkg/topology/kademlia/kademlia_test.go
```diff
@@ -1485,6 +1485,92 @@ func TestAnnounceBgBroadcast(t *testing.T) {
...
+		totalPeers++
+		if randBool.Bool() {
+			kad.SetPeerReachability(addr, p2p.ReachabilityStatusPublic)
+			reachable[addr.ByteString()] = struct{}{}
+			totalReachable++
+		}
+		return false, false, nil
+	})
+
+	fmt.Println(totalPeers, totalReachable)
```
aloknerurkar, line 1519: Remove print

pkg/topology/kademlia/kademlia_test.go
```diff
@@ -1485,6 +1485,92 @@ func TestAnnounceBgBroadcast(t *testing.T) {
...
+
+	t.Run("EachPeer reachable", func(t *testing.T) {
+		count := 0
+		err := kad.EachPeer(func(addr swarm.Address, _ uint8) (bool, bool, error) {
+			if _, notExists := reachable[addr.ByteString()]; !notExists {
```
aloknerurkar, line 1524: variable should be exists

acud:
btw @aloknerurkar @mrekucci the codecov tool complains that the p2p.go file coverage is down 50%, my guess is that unit tests for the reachability type are missing

PR #2569, opened 2021-10-05
fix(pusher): non-fatal tag error

### LINKED ISSUE #2568: Pusher fails if tag is missing
#### Summary
I have a node with LOTS of chunks waiting in a push queue with associated tags.  One of those tags got deleted (a different story) causing the pusher to believe it failed to push the chunk when the tag was not found for incrementing.

#### Steps to reproduce
Get a bunch of chunks in a push queue and delete the tag that they were uploaded with.  When the pusher gets to them, you'll get the following log:
'''
level=trace msg="pusher: cannot push chunk d28d0072b8aa438e7d66a456462956f6bfadb4317fafbd8f620c47b0879c287d: tag not found"
'''
#### Expected behavior
Given that the comments in the code even describe how a tag might not be present, I expected the pusher to think the chunk was pushed successfully even though the tag was not incremented.

#### Actual behavior
The err value from the tag lookup was exposed to the defer code (https://github.com/ethersphere/bee/blob/a14ebf53c3a2fe9c46313b69affc31063c5e8f75/pkg/pusher/pusher.go#L191) causing the pusher to believe it failed.

The correction should be to add the following code to clear the error if a tag was not found.  This code goes immediately after the tag increment if found at https://github.com/ethersphere/bee/blob/a14ebf53c3a2fe9c46313b69affc31063c5e8f75/pkg/pusher/pusher.go#L280

```
					t, err = s.tag.Get(ch.TagID())
					if err == nil && t != nil {
<snip>
					} else {
                                              err = nil   // Allow defer to know that the chunk was pushed
                                        }
```

acud:
fixes #2568 

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2569)

### DIFF
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -267,16 +267,22 @@ LOOP:
 					if err == nil && t != nil {
 						err = t.Inc(tags.StateSynced)
 						if err != nil {
-							err = fmt.Errorf("pusher: increment synced: %v", err)
+							err = fmt.Errorf("increment synced: %v", err)
+							logger.Tracef("pusher: chunk %s tag error: %v", ch.Address().String(), err)
+							err = nil
 							return
 						}
 						if wantSelf {
 							err = t.Inc(tags.StateSent)
 							if err != nil {
-								err = fmt.Errorf("pusher: increment sent: %w", err)
+								err = fmt.Errorf("increment sent: %v", err)
+								logger.Tracef("pusher: chunk %s tag error: %v", ch.Address().String(), err)
+								err = nil
 								return
 							}
 						}
+					} else {
+						err = nil // don't fail because of a tag error
 					}
 				}
 


ldeffenb:
Wow, good catch on the other two!  They'd be a lot harder to cause and/or notice than simply deleting the tag of a chunk that's still in the push queue.

notanatol (approved):

mrekucci (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2569)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ldeffenb)

ldeffenb (approved):
This seems to solve the issue I found this morning.

acud:
Thanks for reporting @ldeffenb !

PR #2570, opened 2021-10-05
fix: bootnode connection direction

mrekucci:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2570)


acud (approved):

aloknerurkar (approved):
Can we add the comment in `debugapi/peer.go` file to mention that this will show up as an inbound connection in the topology endpoint?

PR #2574, opened 2021-10-06
feat: protocols use only reachable peers

istae:


acud (approved):
Looks good but I think it is not needed to provide the argument as a pointer, value semantics should be fine

notanatol (approved):

PR #2597, opened 2021-10-13
refactor: remove QUIC protocol support from libp2p

mrekucci:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2597)


acud:
@mrekucci would be maybe still test for the `ws` transport (though it is also not used)

PR #2620, opened 2021-10-25
feat: direct uploads

acud:
This PR introduces the possibility to push data directly onto the network by default while still allowing buffered uploads as it was traditionally allowed. With the new mode of operation, the root hash of an upload will only be received once all chunks have been safely uploaded into the network. 

TODO:
- [x] test vectors on pusher (for the apiC)
- - [x] create a simple channel and inject it into the new pusher, feed a chunk and check that the pushsync mock was called
- [x] API test vectors
- [x] modify all endpoints to push to network by default
- [x] add header to allow still normal deferred uploads
- - [x] `Swarm-Deferred-Upload` header
- [x] figure out the right amount of retries we need to do (and where)
- [x] resolve how to do re-stamping of content/repetitive upload of the same content
- - repetitive upload of the same content using the same postage stamp while pushing directly into the network will result in the uniformity assuption being invalid on the stamp level
- - same chunk address gets stamped multiple times over the same stamp, causing some buckets to be overcluttered with same collisions, which would render the stamp useless much faster
- - one possible solution is to keep tabs on which addresses were stamped on every batch (out of scope for this task) (or maybe a cuckoo/bloom filter to allow for a small amount of errors, the memory footprint would be far smaller)
- - another solution would be to implement the resetting of collision buckets on a batch (workaround for users, needs explaining on the functional level)
- - document thoroughly (documentation)
- - turn on encryption (return to the uniformity assumption,  no matter how many times the chunk was uploaded using the same batch)
- [x] openAPI spec (define the new header across the endpoints)

supersedes https://github.com/ethersphere/bee/pull/2579

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2620)


pkg/api/api.go
```diff
@@ -442,12 +454,91 @@ func equalASCIIFold(s, t string) bool {
...
+		stamp, err := p.stamper.Stamp(c.Address())
+		if err != nil {
+			return nil, err
+		}
+
+		func(ch swarm.Chunk) {
+			p.sem <- struct{}{}
+			p.eg.Go(func() error {
+				errc := make(chan error, 1)
+				// note: shutdown might be tricky, we need to pass the quit channel
```
aloknerurkar, line 520: Would it be better to have a separate `chan` for each `pushStamperPutter`. Since the `smugler` can take in multiple channels, we can handle closing of each channel separately and handle the global channel closing better?

So for every upload we will try to enqueue a new `pusher.Op` channel to the smugler. So we can wait till the channel is enqueued till the request context is closed. Once it is enqueued, a separate routine will read from the channel uninterrupted. This way a naked write here would be okay.

AuHau:
Hey guys,

I believe @agazso already shared with you our concern about what this change will do to the user experience of uploading data. I know that you plan to support the current behavior for now as well but to my understanding, you plan to change the default to this new blocking upload.

I would like to strongly suggest not to do that as I perceive it as a breaking change. Yes, it does not break the schema of the API, but it breaks the behavior of it which is to me on the same level of importance. If I heard correctly, your idea is that in the future the syncing will be fast enough so the resulting overhead will be small, but I don't think we are there yet. Also, this is quite a major change to the mechanism of uploading data and it would make sense if it goes through the deprecation cycle in a way that the current behavior stays default and the new blocking upload would be hidden behind a flag. In this way users can start to slowly prepare for this change.

ldeffenb:
It's called "backward compatibility" by some and "principle of least surprise" by others.

acud:
Your points are taken @AuHau, @ldeffenb, but please be mindful that we're not proceeding with this change just because we think this makes the most sense in terms of UX. Part of the problem here is that upload-intensive nodes will have their localstores significantly grow, slowing down _all_ storage abstraction operations across all protocols. I have mentioned this to @agazso multiple times. This change is indeed a significant UX change, but it is needed so that cascading effects across the network due to a degradation of service of the storage abstraction is mitigated as much as possible.

Or in other words, since less data goes into the nodes' local storage by default, better quality of service is to be preserved across all protocols. Once nodes start ingesting large amounts of data (like @ldeffenb's OSM archive), they are expected to grind down all syncing and retrieval operations due to storage overgrowth.

Now that we have the discussion refocused, I am happy to hear out your arguments. Also, the current design and implementation was thoroughly discussed with @agazso and he gave a nod to keeping deferred uploads possible through a header. so :shrug:

ldeffenb:
Well, with the recent crash of the mainnet depth due to the EIP???? upgrade impacting apparently lots of nodes' swap-endpoint, everyone's localstore has been growing huge since last Friday anyway.  Instead of pullsync just pulling depth 10-12, it's now pulling depth 3-4.  But it is what it is, I guess.  Until the localstore abstraction gets more efficient for larger data (which is inevitable if the swarm grows the way I believe it is intended), then I guess we'll have to live with temporary patches and work-arounds that may prove sub-optimal in the long run as well.

ldeffenb:
Well, with the recent crash of the mainnet depth due to the EIP???? upgrade impacting apparently lots of nodes' swap-endpoint, everyone's localstore has been growing huge since last Friday anyway.  Instead of pullsync just pulling depth 10-12, it's now pulling depth 3-4.  But it is what it is, I guess.  Until the localstore abstraction gets more efficient for larger data (which is inevitable if the swarm grows the way I believe it is intended), then I guess we'll have to live with temporary patches and work-arounds that may prove sub-optimal in the long run as well.

acud:
This is not intended to be a temporary patch or a work-around.

ldeffenb:
Maybe not intended to be temporary, and maybe a good thing to have, but I suspect it would not be happening right now if the localstore performed better with large content.  Just a comment from an outsider looking in.

pkg/api/api_test.go
```diff
@@ -336,6 +338,73 @@ func TestPostageHeaderError(t *testing.T) {
...
+			)
+			var body struct {
+				Reference swarm.Address `json:"reference"`
+			}
+			if err := json.Unmarshal(responseBytes, &body); err != nil {
+				t.Fatal("unmarshal response body:", err)
+			}
+			if found, _ := mockStorer.Has(context.Background(), body.Reference); !found {
+				t.Fatal("chunk not found in the store")
+			}
```
acud, line 381: i would also check that the chan store does not have the chunk (for whatever reason). the inverse also in the next test

openapi/SwarmCommon.yaml
```diff
@@ -738,6 +738,15 @@ components:
...
         $ref: "#/components/schemas/SwarmAddress"
 
+    SwarmDeferredUpload:
+      in: header
+      name: swarm-deferred-upload
+      schema:
+        type: boolean
+      required: false
+      description: >
+        Determines if the uploaded data should be sent to the network immediately or in a deferred fashion.
```
acud, line 748: Maybe a note about the default fallback value here

pkg/api/api.go
acud: I think it is a nice idea, but do you think it makes a difference? you still need to invoke a goroutine for every chunk through the error group, and you still need to do the individual allocations for the channels here while just adding more complexity on the pusher side, no? I would create a `sync.Pool` of channels and just pool the individual operation channels, so that channel allocations become less wasteful. Also, the naked write should be replaced with a select case as you pointed out :+1:

zelig:
I think this is a super important feature which needs wider discussion.

I am adamant that this feature cannot be used well without having continuous saving of the current root hash and support for resume upload functionality which would use the last recorded top hash in the swarm file tree to append to a tree and read starting from an offset

pkg/node/node.go
```diff
@@ -760,14 +760,16 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
...
 		feedFactory := factory.New(ns)
 		steward := steward.New(storer, traversalService, retrieve, pushSyncProtocol)
-		apiService = api.New(tagService, ns, multiResolver, pssService, traversalService, pinningService, feedFactory, post, postageContractService, steward, signer, authenticator, logger, tracer, api.Options{
+		apiService, chunkC = api.New(tagService, ns, multiResolver, pssService, traversalService, pinningService, feedFactory, post, postageContractService, steward, signer, authenticator, logger, tracer, api.Options{
 			CORSAllowedOrigins: o.CORSAllowedOrigins,
 			GatewayMode:        o.GatewayMode,
 			WsPingPeriod:       60 * time.Second,
 			Restricted:         o.Restricted,
 		})
+		pusherService.SetApiC(chunkC)
```
aloknerurkar, line 772: So context from my previous context regarding enqueuing a separate channel for every putter...

I find this a bit confusing. We call this SetApiC() function here. This looks like it will tell the pusher service to start taking chunks from this channel. But what we are doing is creating a separate goroutine to read from this channel.

So the SetApiC function can be called multiple times, and it is not being used that way. There is nothing wrong here from a code perspective, just less clarity on the use of this.

pkg/pusher/pusher.go
```diff
@@ -129,43 +170,51 @@ func (s *Service) chunksWorker(warmupTime time.Duration, tracer *tracing.Tracer)
...
-		startTime := time.Now()
-		wg.Add(1)
-		go func(ctx context.Context, ch swarm.Chunk) {
-			defer func() {
-				wg.Done()
-				<-s.sem
+		case apiC := <-s.smugler:
+			go func() {
+				for op := range apiC {
+					cc <- op
```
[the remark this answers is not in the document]
aloknerurkar: this still needs to be handled, right?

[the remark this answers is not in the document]
acud: :+1:

pkg/node/node.go
acud: I think that the `SetApiC` can be renamed into something more like `FeedFrom(chan *Op)` and this would resolve the contention here. Am I correct?

aloknerurkar: `AddFeed` sounds better. As the interface to pusher specifies that more feeds can possibly be added.

aloknerurkar (approved):

notanatol (approved):

PR #2628, opened 2021-10-27
Feat: Time based upgrade of payment thresholds

metacertain:
This PR introduces a mechanism, that favors long-standing connections by gradually upgrading payment thresholds after an amount of traffic is served for and settled by a peer.

The milestone for upgrading a payment threshold is after each 100 seconds worth of refreshment rate being settled (by either time based or monetary settlements).
The upgrade of the payment threshold is an increase of 1 second worth of refreshment rate per upgrade.
After the first 18 milestones the increase slows down to a logarithmic rate, by setting the next milestone at twice of the current milestones accumulated traffic settled.

When a peer is newly connected, it receives a default payment threshold once again, so that disconnections / reconnections are not incentivized.

The PR introduces individual payment thresholds given to peers.
The PR also includes relaying the information about whether a peer is a full node to accounting from pseudosettlements, in order to differentiate traffic milestones for threshold upgrades between light nodes and full nodes.
The PR introduces differentiation of initial payment thresholds given to light nodes and full nodes.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2628)


pkg/accounting/accounting.go
```diff
@@ -92,11 +92,16 @@ type accountingPeer struct {
...
-	lastSettlementFailureTimestamp int64 // time of last unsuccessful attempt to issue a cheque
+	earlyPayment                   *big.Int   // individual early payment threshold calculated from from payment threshold and early payment percentage
+	paymentThresholdForPeer        *big.Int   // individual payment threshold at which the peer is expected to pay
+	disconnectLimit                *big.Int   // individual disconnect threshold calculated from tolerance and payment threshold for peer
+	refreshTimestamp               int64      // last time we attempted time-based settlement
+	paymentOngoing                 bool       // indicate if we are currently settling with the peer
+	lastSettlementFailureTimestamp int64      // time of last unsuccessful attempt to issue a cheque
 	connected                      bool
+	fullNode                       bool
+	totalDebtRepay                 *big.Int
```
ralph-pichler, line 103: add comments to the new fields

pkg/accounting/accounting.go
```diff
@@ -150,23 +163,34 @@ func NewAccounting(
...
+		logger:                   Logger,
+		store:                    Store,
+		pricing:                  Pricing,
+		metrics:                  newMetrics(),
+		refreshRate:              new(big.Int).Set(refreshRate),
+		lightRefreshRate:         new(big.Int).Div(refreshRate, big.NewInt(lightFactor)),
+		timeNow:                  time.Now,
+		minimumPayment:           new(big.Int).Div(refreshRate, big.NewInt(minimumPaymentDivisor)),
+		p2p:                      p2pService,
+		thresholdGrowChange:      new(big.Int).Mul(refreshRate, big.NewInt(180)),
```
ralph-pichler, line 188: there are some numbers here (and below) which should be extracted to constants

pkg/accounting/accounting.go
```diff
@@ -524,20 +548,52 @@ func (a *Accounting) getAccountingPeer(peer swarm.Address) *accountingPeer {
...
+			totalDebtRepay:          big.NewInt(0),
+			paymentThreshold:        new(big.Int).Set(a.paymentThreshold),
+			paymentThresholdForPeer: new(big.Int).Set(a.paymentThreshold),
+			disconnectLimit:         new(big.Int).Set(a.disconnectLimit),
+			thresholdGrowAt:         new(big.Int).Set(a.thresholdGrowStep),
 			// initially assume the peer has the same threshold as us
-			paymentThreshold: new(big.Int).Set(a.paymentThreshold),
-			earlyPayment:     new(big.Int).Div(new(big.Int).Mul(a.paymentThreshold, big.NewInt(100-a.earlyPayment)), big.NewInt(100)),
-			connected:        false,
+			earlyPayment: new(big.Int).Div(new(big.Int).Mul(a.paymentThreshold, big.NewInt(100-a.earlyPayment)), big.NewInt(100)),
```
ralph-pichler, line 560: as we have this increase or decrease by some percent in multiple places can we make a helper for that. might help with readability.

pkg/accounting/accounting.go
```diff
@@ -524,20 +548,52 @@ func (a *Accounting) getAccountingPeer(peer swarm.Address) *accountingPeer {
...
+			connected:    false,
 		}
 		a.accountingPeers[peer.String()] = peerData
 	}
 
 	return peerData
 }
 
+func (a *Accounting) NotifyPaymentThresholdUpgrade(peer swarm.Address, accountingPeer *accountingPeer) {
```
ralph-pichler, line 569: why is this exported?

pkg/accounting/accounting.go
```diff
@@ -524,20 +548,52 @@ func (a *Accounting) getAccountingPeer(peer swarm.Address) *accountingPeer {
...
+
+	thresholdGrowChange := new(big.Int).Set(a.thresholdGrowChange)
+	if !accountingPeer.fullNode {
+		thresholdGrowChange.Set(a.lightThresholdGrowChange)
+	}
+
+	if accountingPeer.thresholdGrowAt.Cmp(thresholdGrowChange) >= 0 {
```
ralph-pichler, line 576: add comments

pkg/accounting/accounting.go
ralph-pichler, line 569: also add a comment that `accountingPeer` must be locked already

pkg/accounting/accounting.go
```diff
@@ -524,20 +548,52 @@ func (a *Accounting) getAccountingPeer(peer swarm.Address) *accountingPeer {
...
+
+	refreshRate := new(big.Int).Set(a.refreshRate)
+	if !accountingPeer.fullNode {
+		refreshRate = new(big.Int).Set(a.lightRefreshRate)
+	}
+
+	accountingPeer.paymentThresholdForPeer = new(big.Int).Add(accountingPeer.paymentThresholdForPeer, refreshRate)
+	accountingPeer.disconnectLimit = new(big.Int).Div(new(big.Int).Mul(accountingPeer.paymentThresholdForPeer, big.NewInt(100+a.paymentTolerance)), big.NewInt(100))
+
+	_ = a.pricing.AnnouncePaymentThreshold(context.Background(), peer, accountingPeer.paymentThresholdForPeer)
```
ralph-pichler, line 594: should we not at least log this? or is this already done elsewhere?

pkg/accounting/accounting.go
```diff
@@ -120,14 +125,22 @@ type Accounting struct {
...
+	minimumPayment      *big.Int
+	pricing             pricing.Interface
+	metrics             metrics
+	wg                  sync.WaitGroup
+	p2p                 p2p.Service
+	timeNow             func() time.Time
+	thresholdGrowStep   *big.Int
+	thresholdGrowChange *big.Int
+	// light node counterparts
+	lightPaymentThreshold    *big.Int
```
ralph-pichler, line 140: maybe not for this PR but it might be useful to have some kind of threshold "profile" struct, one for full nodes, one for light nodes. then each accounting peer can be given the appropriate one and just get stuff like `thresholdGrowChange` or the step from there. this may simplify the logic a bit and avoid some duplication.

ralph-pichler (changes requested):

pkg/accounting/accounting.go
ralph-pichler: not addressed yet

metacertain: it is

pkg/accounting/accounting.go
metacertain: lets address this in a separate PR

pkg/accounting/accounting.go
metacertain: you can only see it in the actual files changed tab, for some reason Conversation page displays a snippet of a different earlier version

mrekucci (changes requested):
> Reviewed 9 of 11 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2628)* status: all files reviewed, 10 unresolved discussions (waiting on @metacertain and @ralph-pichler)

___
*[pkg/accounting/accounting.go, line 106 at r3](https://reviewable.io/reviews/ethersphere/bee/2628#-MrNMHqK3f1OHKk-c23D:-MrNMHqL-A971HukIgfe:b-q0seqp) ([raw file](https://github.com/ethersphere/bee/blob/b7236adbb20e99aeba7bbbb3318fcb0e3f752767/pkg/accounting/accounting.go#L106)):*
> ```Go
> 	paymentOngoing                 bool       // indicate if we are currently settling with the peer
> 	lastSettlementFailureTimestamp int64      // time of last unsuccessful attempt to issue a cheque
> 	connected                      bool       //
> ```

Missing comment?
___
*[pkg/accounting/accounting.go, line 611 at r3](https://reviewable.io/reviews/ethersphere/bee/2628#-MrNOb5503Ph0ipCp4nQ:-MrNOb5503Ph0ipCp4nR:b-htp5qs) ([raw file](https://github.com/ethersphere/bee/blob/b7236adbb20e99aeba7bbbb3318fcb0e3f752767/pkg/accounting/accounting.go#L611)):*
> ```Go
> 	err := a.pricing.AnnouncePaymentThreshold(context.Background(), peer, accountingPeer.paymentThresholdForPeer)
> 	if err != nil {
> 		a.logger.Errorf("announcing increased payment threshold of %d to peer %v: %w", accountingPeer.paymentThresholdForPeer, peer, err)
> ```

The `%w` verb isn't supported by the logger, use `%v` instead.
___
*[pkg/accounting/accounting_test.go, line 1597 at r3](https://reviewable.io/reviews/ethersphere/bee/2628#-MrNPDtUB034AE78kfXA:-MrNPDtUB034AE78kfXB:bdgkl5b) ([raw file](https://github.com/ethersphere/bee/blob/b7236adbb20e99aeba7bbbb3318fcb0e3f752767/pkg/accounting/accounting_test.go#L1597)):*
> ```Go
> 
> 			if pricing.paymentThreshold.Cmp(checkPaymentThreshold) != 0 {
> 				t.Fatalf("wrong threshold. got %d wanted %d", pricing.paymentThreshold, checkPaymentThreshold)
> ```

Don't use punctuation in error strings: https://github.com/golang/go/wiki/CodeReviewComments#error-strings
The same is below.
___
*[pkg/accounting/accounting_test.go, line 1719 at r3](https://reviewable.io/reviews/ethersphere/bee/2628#-MrNQ4iU7kj8-UUaJDnz:-MrNQ4iU7kj8-UUaJDo-:b7uo9na) ([raw file](https://github.com/ethersphere/bee/blob/b7236adbb20e99aeba7bbbb3318fcb0e3f752767/pkg/accounting/accounting_test.go#L1719)):*
> ```Go
> 	err = acc.NotifyRefreshmentReceived(peer1Addr, big.NewInt(debitRefresh))
> 	if err != nil {
> 		t.Fatal("Unexpected error from NotifyRefreshmentReceived", err)
> ```

Error strings should not be capitalized: https://github.com/golang/go/wiki/CodeReviewComments#error-strings
Also, you should use the `Fatalf` method to add the `err` into the context.
The same is below.
___
*[pkg/pricing/pricing_test.go, line 286 at r3](https://reviewable.io/reviews/ethersphere/bee/2628#-MrNRVx60EJ2lfj6EKwH:-MrNRVx60EJ2lfj6EKwI:bhpw8eu) ([raw file](https://github.com/ethersphere/bee/blob/b7236adbb20e99aeba7bbbb3318fcb0e3f752767/pkg/pricing/pricing_test.go#L286)):*
> ```Go
> 
> 	if observer.paymentThreshold.Cmp(testLightThreshold) != 0 {
> 		t.Fatalf("observer called with wrong paymentThreshold. got %d, want %d", observer.paymentThreshold, testLightThreshold)
> ```

Use `%v` or `%s` formatting verb instead of `%d` for `big.Int` values.

mrekucci (approved):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2628)* status: all files reviewed, 10 unresolved discussions (waiting on @metacertain and @ralph-pichler)

pkg/accounting/accounting.go
```diff
@@ -524,20 +553,63 @@ func (a *Accounting) getAccountingPeer(peer swarm.Address) *accountingPeer {
...
+func (a *Accounting) notifyPaymentThresholdUpgrade(peer swarm.Address, accountingPeer *accountingPeer) {
+
+	// get appropriate linear growth limit based on whether the peer is a full node or a light node
+	thresholdGrowChange := new(big.Int).Set(a.thresholdGrowChange)
+	if !accountingPeer.fullNode {
+		thresholdGrowChange.Set(a.lightThresholdGrowChange)
+	}
+
+	// if current milestone already passed linear growth limit, set next milestone exponentially
+	if accountingPeer.thresholdGrowAt.Cmp(thresholdGrowChange) >= 0 {
```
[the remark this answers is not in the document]
metacertain: In the warmup period, connections have their inital, safe-low payment thresholds. The nodes do self throttle, meaning they don't reserve (queue) more requests towards a peer than the payment threshold would allow. As long as the requests are served / fail efficiently (within reasonable time) this throttling effect is negligible. If the initial thresholds are smaller, timing-out / hanging requests can have a more pronounced effect combined with this self throttling.

Later on, the long standing connections that see significant amount of traffic get these self-throttling limits extended, meaning they can work through larger bursts of requests enabling more throughput

The intention is to consecutively decrease the default configured payment threshold to the minimal accepted threshold which is 10 times less than the current default.
In this case, warmup can be considered the time it takes to get a 10x higher payment threshold than the safe-low.
This is the linear phase of threshold growth, which is after having served and get settled for a total of 1800 seconds worth of refreshment rate.
Without cheques this presumes having at least half an hour intensive traffic through the connection.
If cheques are present, this can be shorter, because the same volume of traffic can be served and settled in less time, as the debts are decreased faster than the time based refreshments, and achieving checkpoints earlier are only limited by the speed of the network combined with the self-throttling of concurrently queued requests.


Throughput expected on a single p2p connection with safe-low initial payment threshold:

Reserve limit / number of seconds average request is served in 
 
where
reserve limit ~ minimal payment threshold accepted = 2 * refresh rate ~ 33+ chunks (depending on PO more for higher PO connections)

pkg/accounting/accounting.go
```diff
@@ -92,11 +97,16 @@ type accountingPeer struct {
...
-	paymentOngoing                 bool  // indicate if we are currently settling with the peer
-	lastSettlementFailureTimestamp int64 // time of last unsuccessful attempt to issue a cheque
-	connected                      bool
+	earlyPayment                   *big.Int   // individual early payment threshold calculated from from payment threshold and early payment percentage
+	paymentThresholdForPeer        *big.Int   // individual payment threshold at which the peer is expected to pay
+	disconnectLimit                *big.Int   // individual disconnect threshold calculated from tolerance and payment threshold for peer
+	refreshTimestamp               int64      // last time we attempted time-based settlement
+	paymentOngoing                 bool       // indicate if we are currently settling with the peer
+	lastSettlementFailureTimestamp int64      // time of last unsuccessful attempt to issue a cheque
+	connected                      bool       //
```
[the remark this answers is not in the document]
metacertain: stale version displayed, it was added recently

ralph-pichler (approved):

metacertain:
as part of #2904

PR #2638, opened 2021-11-02
feat: add bee open API versions info to the health endpoint

mrekucci:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2638)

### DIFF
--- a/pkg/api/version.go
+++ b/pkg/api/version.go
@@ -5,4 +5,4 @@
 package api
 
 // Version is set in the build process.
-var Version string
+var Version = "0.0.0"
--- a/pkg/api/version.go
+++ b/pkg/api/version.go
@@ -1,0 +1,8 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package api
+
+// Version is set in the build process.
+var Version string
--- a/pkg/debugapi/version.go
+++ b/pkg/debugapi/version.go
@@ -1,0 +1,8 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package debugapi
+
+// Version is set in the build process.
+var Version string
--- a/pkg/debugapi/version.go
+++ b/pkg/debugapi/version.go
@@ -5,4 +5,4 @@
 package debugapi
 
 // Version is set in the build process.
-var Version string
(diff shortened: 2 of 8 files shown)


mrekucci:
@AuHau

notanatol (approved):

AuHau:
WOW! What a speed! Thanks a lot @mrekucci 🙏

acud (approved):
nice!

acud:
my only question would be - what would the endpoint give out when you just use `go build cmd/bee` and not `make binary`? and would it make the api panic or the libraries that read it break or throw an exception while parsing? maybe it would be beneficial to set the fallback to semver `0.0.0` so that parsing succeeds and that the version in this case is always _below_ some version that is expected, so that the libraries can issue the right warnings to the users. just an idea

pkg/api/version.go
```diff
@@ -0,0 +1,8 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package api
+
+// Version is set in the build process.
+var Version string
```
acud, line 8: it is actually not needed to make the member public. the build tags can also set a package scoped variable.

pkg/debugapi/version.go
```diff
@@ -0,0 +1,8 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package debugapi
+
+// Version is set in the build process.
+var Version string
```
acud, line 8: here too

pkg/api/version.go
mrekucci: It is actually used it in the `debugapi` package so it must be exported.

pkg/debugapi/version.go
mrekucci: This doesn't need to but just for the symmetry with the `api` package, I'd also export it.

mrekucci:
> my only question would be - what would the endpoint give out when you just use `go build cmd/bee` and not `make binary`? and would it make the api panic or the libraries that read it break or throw an exception while parsing? maybe it would be beneficial to set the fallback to semver `0.0.0` so that parsing succeeds and that the version in this case is always _below_ some version that is expected, so that the libraries can issue the right warnings to the users. just an idea

It'll return `{"status":"ok","version":"1.3.0-dev","apiVersion":"","debugApiVersion":""}`. We can do `0.0.0` as a fallback.

PR #2639, opened 2021-11-03
feat!: change to percentage based accounting tolerances

metacertain:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2639)


ldeffenb (commented):
Would it be better for migration/upgrade purposes to change the actual names of the configuration parameters so that obsolete values aren't interpreted as really huge percentages?  That way existing .yaml configuration files would just get the new defaults until an operator explicitly provided a percentage value.

ldeffenb (commented):
@metacertain Thank you!

packaging/docker/env
```diff
@@ -59,12 +59,12 @@ BEE_CLEF_SIGNER_ENABLE=true
 # BEE_PASSWORD=
 ## path to a file that contains password for decrypting keys
 # BEE_PASSWORD_FILE=
-## amount in BZZ below the peers payment threshold when we initiate settlement (default 10000000)
-# BEE_PAYMENT_EARLY=10000000
+## percentage below the peers payment threshold when we initiate settlement (default 25)
+# BEE_PAYMENT_EARLY=50
```
acud, line 63: this is no longer the flag name (you need to add `_PERCENT`)

pkg/accounting/accounting.go
```diff
@@ -106,11 +107,11 @@ type Accounting struct {
 	store             storage.StateStorer
 	// The payment threshold in BZZ we communicate to our peers.
 	paymentThreshold *big.Int
-	// The amount in BZZ we let peers exceed the payment threshold before we
+	// The amount in % we let peers exceed the payment threshold before we
```
acud, line 110: percent

mrekucci (approved):
> Reviewed 8 of 10 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2639)* status: all files reviewed, 3 unresolved discussions (waiting on @metacertain)

___
*[.github/workflows/beekeeper.yml, line 17 at r3](https://reviewable.io/reviews/ethersphere/bee/2639#-Moc1FvqDP46-9bCBdBm:-Moc1Fvs7AliyzKiZEjK:b-26k295) ([raw file](https://github.com/ethersphere/bee/blob/9184a4ffb8e81322e52cbff58f597230903a63ff/.github/workflows/beekeeper.yml#L17)):*
> ```YAML
>       RUN_TYPE: "PR RUN"
>       SETUP_CONTRACT_IMAGE_TAG: "0.2.0"
>       BEEKEEPER_BRANCH: "tolerance_percentages"
> ```

I guess this was valid when you tested it against this beekeeper branch, shouldn't be this reverted back to `master` if the beekeeper changes were merged (also the same below)?

acud (approved):

PR #2649, opened 2021-11-08
feat(debugapi): node ep

### LINKED ISSUE #1668: Get info about Bee node on the API
# User story

Recently the Bee node can start either as a full node, light client or in gateway mode. Their _active services_ (that you can call and get successful response from, via HTTP API) differ from each other and it is not possible right now to tell about the services of the target node whether those are active or not _from code_ (unless you know which services won't work in a particular mode and you handle the exception in that way, but it is not the best alternative).

## Story

Therefore would be nice to have an endpoint on the Bee API which tells what paths of the node are active and which ones are turned off so that the apps which communicates with bee node(s) could handle properly these differences. 

## Acceptance criteria

`/info` (or with some other name) API EP, which tells information about the Bee node, and one property of this gives back what services are turned off (like `/pin/*` paths) as an array 

## Background

We had to handle somehow in `swarm-cli` the pinning is not available in gateway mode, but we [did not do in a nice way](https://github.com/ethersphere/swarm-cli/issues/97)

## Tasks

nugaon:
New `/node` GET endpoint on the API.

Returns the mode in which the node was started as `light`, `full` or `dev`.
Also gives back whether the node started in `gateway` mode or not.

TODO: 
- [x] tests
- [x] OpenAPI doc 

Closes #1668

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2649)

### DIFF
--- a/pkg/debugapi/info.go
+++ b/pkg/debugapi/info.go
@@ -1,4 +1,4 @@
-package api
+package debugapi
 
 import (
 	"net/http"
@@ -31,10 +31,10 @@ func (b BeeNodeMode) String() string {
 	return "unknown"
 }
 
-// bytesGetHandler handles retrieval of raw binary data of arbitrary length.
-func (s *server) infoGetHandler(w http.ResponseWriter, r *http.Request) {
+// infoGetHandler gives back information about the Bee node configuration.
+func (s *Service) infoGetHandler(w http.ResponseWriter, r *http.Request) {
 	jsonhttp.OK(w, infoResponse{
-		BeeMode:     s.BeeMode.String(),
-		GatewayMode: s.GatewayMode,
+		BeeMode:     s.beeMode.String(),
+		GatewayMode: s.gatewayMode,
 	})
 }
--- a/pkg/debugapi/info.go
+++ b/pkg/debugapi/info.go
@@ -1,3 +1,7 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 package debugapi
 
 import (
--- a/pkg/debugapi/node.go
+++ b/pkg/debugapi/node.go
@@ -18,7 +18,7 @@ const (
 	DevMode
 )
 
-type infoResponse struct {
+type nodeResponse struct {
 	BeeMode     string `json:"beeMode"`
 	GatewayMode bool   `json:"gatewayMode"`
 }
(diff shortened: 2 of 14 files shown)


mrekucci:
Thanks for the PR. I'd suggest moving this information to the `/health` endpoint.

acud (changes requested):
Thanks for the contribution @nugaon. My comments would be that this information should not be exposed on the `api` but rather on the `debug api` for now. The normal API does not dive into all sorts of node configuration details and tries to be strictly opinionated about content related operations. Node configuration is exposed through the endpoints in the debug api. Also, I think that there should not be a type representing all different bee node operation mode permutations, but rather expose all the configuration states through the debug api. So for example, you would expose whether the node is a bootnode, whether it is in gateway mode, whether it is a light node etc separately, rather than just trying to represent them all under one label.

pkg/debugapi/info.go
```diff
@@ -0,0 +1,40 @@
...
+	case DevMode:
+		return "dev"
+	}
+	return "unknown"
+}
+
+// infoGetHandler gives back information about the Bee node configuration.
+func (s *Service) infoGetHandler(w http.ResponseWriter, r *http.Request) {
+	jsonhttp.OK(w, infoResponse{
+		BeeMode:     s.beeMode.String(),
```
acud, line 37: why not use the existing api endpoint that shows the versions?

pkg/debugapi/info.go
```diff
@@ -0,0 +1,40 @@
+package debugapi
```
acud, line 1: code copyright missing

acud (changes requested):
comment from 5 days ago still pending

pkg/debugapi/info.go
acud: this is still not addressed, also, `info` is very general name for such an endpoint. not sure why a new endpoint is needed

nugaon: well, the JS team hadn't found corresponding existing endpoint for these properties, moreover the `apiversion`, `debugApiVersion` and the `version` properties also could be placed on this new EP. Thereby, the `health` EP could act as a healthcheck only and all other informational parameters could be served from the `info` EP.

pkg/debugapi/router.go
```diff
@@ -63,7 +63,11 @@ func (s *Service) newBasicRouter() *mux.Router {
 		router.Handle(path, handler)
 	}
 
-	handle("/addresses", jsonhttp.MethodHandler{
+	router.Handle("/info", jsonhttp.MethodHandler{
```
acud, line 66: consider `/node`

acud (approved):

notanatol (approved):

pkg/debugapi/node.go
```diff
@@ -0,0 +1,44 @@
...
+	FullMode
+	DevMode
+)
+
+type nodeResponse struct {
+	BeeMode     string `json:"beeMode"`
+	GatewayMode bool   `json:"gatewayMode"`
+}
+
+func (b BeeNodeMode) String() string {
```
aloknerurkar, line 26: It would be good to add a test for this. You can look at [this](https://github.com/ethersphere/bee/blob/39cda82e11d57d3146ea53e8ac7bc2b8e44080a2/pkg/p2p/p2p_test.go#L23) for examples.

acud:
@nugaon please have a separate PR for the missing unit test. Merging this for now so that we can roll it out

PR #2651, opened 2021-11-08
fix: checks at time based refreshment

metacertain:
The aim of this PR is to make time based settlements more safe, by verifying timestamps from both sides to conclude whether 1 second passed since the last refreshment happened.

As an implication of this, the refreshments can be deintensified to once every two seconds in case there is an NTP difference observed on the two sides.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2651)

### DIFF
--- a/pkg/settlement/pseudosettle/pseudosettle.go
+++ b/pkg/settlement/pseudosettle/pseudosettle.go
@@ -272,7 +272,7 @@ func (s *Service) Pay(ctx context.Context, peer swarm.Address, amount, checkAllo
 	}
 
 	currentTime := s.timeNow().Unix()
-	if currentTime == lastTime.CheckTimestamp {
+	if currentTime == lastTime.CheckTimestamp || currentTime == lastTime.Timestamp {
 		return nil, 0, ErrSettlementTooSoon
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/settlement/pseudosettle/pseudosettle.go
```diff
@@ -272,7 +272,7 @@ func (s *Service) Pay(ctx context.Context, peer swarm.Address, amount, checkAllo
 	}
 
 	currentTime := s.timeNow().Unix()
-	if currentTime == lastTime.CheckTimestamp {
+	if currentTime == lastTime.CheckTimestamp || currentTime == lastTime.Timestamp {
```
acud, line 275: I know that previously this check was here but @ralph-pichler replaced it with the `CheckTimestamp` field stating that checking against `Timestamp` was incorrect. @ralph-pichler can you elaborate?

metacertain: ( comment from review: peer timestamp can be preemptively checked in accounting as well to inhibit ErrSettlementTooSoon)

pkg/settlement/pseudosettle/pseudosettle.go
```diff
@@ -272,7 +272,7 @@ func (s *Service) Pay(ctx context.Context, peer swarm.Address, amount, checkAllo
 	}
 
 	currentTime := s.timeNow().Unix()
-	if currentTime == lastTime.CheckTimestamp {
+	if currentTime <= lastTime.CheckTimestamp || currentTime <= lastTime.Timestamp {
```
acud, line 275: Would be really nice to have an elaborate comment on this line, since it is so important

acud (approved):

notanatol (approved):

PR #2654, opened 2021-11-09
ci: trigger bee-factory builds on release

AuHau:
This PR is part of an automatization initiative to stream-line updating Bee versions across our JS land. This action triggers automatically build of Bee Factory images that will be then propagated to our JS repos with PRs.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2654)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -52,16 +52,3 @@ jobs:
           SCOOP_PAT: ${{ secrets.SCOOP_PAT }}
           GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
           GPG_FINGERPRINT: ${{ secrets.GPG_FINGERPRINT }}
-
-      - name: Get the released tags version
-        id: get-version
-        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
-
-      - name: Trigger Bee Factory image build
-        uses: peter-evans/repository-dispatch@v1
-        with:
-          token: ${{ secrets.HOMEBREW_TAP_PAT }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -52,3 +52,16 @@ jobs:
...
           GPG_FINGERPRINT: ${{ secrets.GPG_FINGERPRINT }}
+
+      - name: Get the released tags version
+        id: get-version
+        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+
+      - name: Trigger Bee Factory image build
+        uses: peter-evans/repository-dispatch@v1
+        with:
+          token: ${{ secrets.HOMEBREW_TAP_PAT }}
```
AuHau, line 63: This to my understanding should be the Bee Worker's PAT hence I used it even though it is not "home brew" related.

acud (approved):

mrekucci (approved):

vandot:
This will also trigger bee-factory for prereleases (rc builds), is that the behavior we want?

AuHau:
From our end - yes.

vandot:
Maybe it would be nice to include it in separate file (`release_triggers.yaml` or something like that) with event [release](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release) to keep release file on its own and it looks that you already have tag available as `GITHUB_REF`

AuHau:
@vandot like this?

I originally thought it is not possible to do this as I need to have all the Docker images published by the time the Bee Factory workflow is triggered, but in the end, I think I found a way.

PR #2655, opened 2021-11-09
chore: upload and download api metrics middleware

istae:
related to https://github.com/ethersphere/grafana-dashboards/issues/54

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2655)

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -398,21 +398,17 @@ func (s *server) contentLengthMetricMiddleware() func(h http.Handler) http.Handl
 			h.ServeHTTP(w, r)
 			switch r.Method {
 			case http.MethodGet:
-				{
-					contentLength, err := strconv.Atoi(w.Header().Get("Content-Length"))
-					if err != nil {
-						s.logger.Debugf("api: content length int conversation failed: %v", err)
-						return
-					}
-					if contentLength > 0 {
-						s.metrics.ContentApiDuration.WithLabelValues(fmt.Sprintf("%d", toFileSizeBucket(int64(contentLength))), r.Method).Observe(time.Since(now).Seconds())
-					}
+				contentLength, err := strconv.Atoi(w.Header().Get("Content-Length"))
+				if err != nil {
+					s.logger.Debugf("api: content length int conversation failed: %v", err)
+					return
+				}
+				if contentLength > 0 {
+					s.metrics.ContentApiDuration.WithLabelValues(fmt.Sprintf("%d", toFileSizeBucket(int64(contentLength))), r.Method).Observe(time.Since(now).Seconds())
 				}
 			case http.MethodPost:
-				{
-					if r.ContentLength > 0 {
-						s.metrics.ContentApiDuration.WithLabelValues(fmt.Sprintf("%d", toFileSizeBucket(r.ContentLength)), r.Method).Observe(time.Since(now).Seconds())
-					}
+				if r.ContentLength > 0 {
+					s.metrics.ContentApiDuration.WithLabelValues(fmt.Sprintf("%d", toFileSizeBucket(r.ContentLength)), r.Method).Observe(time.Since(now).Seconds())
 				}
 			}
 		})
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -398,7 +398,7 @@ func (s *server) contentLengthMetricMiddleware() func(h http.Handler) http.Handl
 			h.ServeHTTP(w, r)
 			switch r.Method {
 			case http.MethodGet:
-				contentLength, err := strconv.Atoi(w.Header().Get("Content-Length"))
+				contentLength, err := strconv.Atoi(w.Header().Get("Decompressed-Content-Length"))
 				if err != nil {
 					s.logger.Debugf("api: content length int conversation failed: %v", err)
 					return
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -391,23 +391,28 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
 	}
 }
 
-func (s *server) contentLengthMetricMiddleware(method string) func(h http.Handler) http.Handler {
+func (s *server) contentLengthMetricMiddleware() func(h http.Handler) http.Handler {
 	return func(h http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			now := time.Now()
 			h.ServeHTTP(w, r)
-			if method == "GET" {
-				contentLength, err := strconv.Atoi(w.Header().Get("Content-Length"))
-				if err != nil {
-					s.logger.Debugf("api: content length int conversation failed: %v", err)
-					return
+			switch r.Method {
+			case http.MethodGet:
+				{
+					contentLength, err := strconv.Atoi(w.Header().Get("Content-Length"))
+					if err != nil {
+						s.logger.Debugf("api: content length int conversation failed: %v", err)
+						return
+					}
+					if contentLength > 0 {
+						s.metrics.ContentApiDuration.WithLabelValues(fmt.Sprintf("%d", toFileSizeBucket(int64(contentLength))), r.Method).Observe(time.Since(now).Seconds())
+					}
 				}
-				if contentLength > 0 {
-					s.metrics.ContentApiDuration.WithLabelValues(fmt.Sprintf("%d", toFileSizeBucket(int64(contentLength))), method).Observe(time.Since(now).Seconds())
-				}
-			} else if method == "POST" {
-				if r.ContentLength > 0 {
-					s.metrics.ContentApiDuration.WithLabelValues(fmt.Sprintf("%d", toFileSizeBucket(r.ContentLength)), method).Observe(time.Since(now).Seconds())
+			case http.MethodPost:
+				{
+					if r.ContentLength > 0 {
+						s.metrics.ContentApiDuration.WithLabelValues(fmt.Sprintf("%d", toFileSizeBucket(r.ContentLength)), r.Method).Observe(time.Since(now).Seconds())
+					}
 				}
 			}
 		})
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -391,33 +391,27 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
 	}
 }
 
-func (s *server) bzzUploadDurationMiddleware() func(h http.Handler) http.Handler {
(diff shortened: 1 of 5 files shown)


mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2655)* status: all files reviewed, 2 unresolved discussions (waiting on @istae)

___
*[pkg/api/api.go, line 272 at r1](https://reviewable.io/reviews/ethersphere/bee/2655#-Mo3u45e0oHnbSDK0B8V:-Mo3u45e0oHnbSDK0B8W:b9z6ryf) ([raw file](https://github.com/ethersphere/bee/blob/4a17dafaa7effcec1e3ce82030aa7bc1c514ed2e/pkg/api/api.go#L272)):*
> ```Go
> 			now := time.Now()
> 			h.ServeHTTP(w, r)
> 			s.metrics.UploadDuration.WithLabelValues(fmt.Sprintf("%d", r.ContentLength)).Observe(time.Since(now).Seconds())
> ```

I'd suggest checking whether the content length is known (`r.ContentLength > 0`) before recording the metrics.
___
*[pkg/api/api.go, line 282 at r1](https://reviewable.io/reviews/ethersphere/bee/2655#-Mo3uZZH1QfKsH0aDY1m:-Mo3uZZH1QfKsH0aDY1n:b-8l4rah) ([raw file](https://github.com/ethersphere/bee/blob/4a17dafaa7effcec1e3ce82030aa7bc1c514ed2e/pkg/api/api.go#L282)):*
> ```Go
> 			now := time.Now()
> 			h.ServeHTTP(w, r)
> 			s.metrics.DownloadDuration.WithLabelValues(w.Header().Get("Content-Length")).Observe(time.Since(now).Seconds())
> ```

Also here, I'd check whether the `"Content-Length"` header exists before recording the metrics.

ldeffenb (commented):
Do you have any idea how many different "content-length" values this is going to shove into the metrics?  Maybe consider rounding/truncating them into exponentially sized buckets?   Maybe since the chunk size is 4K, that would make a good bucket increment?  Then the metric labels would approximate (ignoring index overhead) the number of chunks that were uploaded or downloaded.

acud (changes requested):
> Reviewed 1 of 3 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2655)* status: all files reviewed, 7 unresolved discussions (waiting on @istae and @mrekucci)

___
*[pkg/api/api.go, line 272 at r1](https://reviewable.io/reviews/ethersphere/bee/2655#-Mo3u45e0oHnbSDK0B8V:-Mo4nm8F1AhHidOMv89Y:b32v57) ([raw file](https://github.com/ethersphere/bee/blob/4a17dafaa7effcec1e3ce82030aa7bc1c514ed2e/pkg/api/api.go#L272)):*

I'd suggest checking whether the content length is known (`r.ContentLength > 0`) before recording the metrics.
</blockquote></details>

+1
___
*[pkg/api/api.go, line 277 at r2](https://reviewable.io/reviews/ethersphere/bee/2655#-Mo4oLCDAujRACHUMIJr:-Mo4oLCDAujRACHUMIJs:b400tjb) ([raw file](https://github.com/ethersphere/bee/blob/af101f9c5bd114b496f3aabbb5975b282d827986/pkg/api/api.go#L277)):*
> ```Go
> }
> 
> func (s *server) bzzDownloadDurationMiddleware() func(h http.Handler) http.Handler {
> ```

Since `content-length` header is written to both on uploads and downloads, you might get away with having just one middleware for all different endpoints that do upload and download, and just create another dimension in the histogram for the HTTP verb used (GET,POST,etc)
___
*[pkg/api/api.go, line 292 at r2](https://reviewable.io/reviews/ethersphere/bee/2655#-Mo4ofmN8IJpDctjSh2s:-Mo4ofmN8IJpDctjSh2t:b-pc7kko) ([raw file](https://github.com/ethersphere/bee/blob/af101f9c5bd114b496f3aabbb5975b282d827986/pkg/api/api.go#L292)):*
> ```Go
> }
> 
> func fileSizeBuckets() {
> ```

can be removed
___
*[pkg/api/metrics.go, line 18 at r2](https://reviewable.io/reviews/ethersphere/bee/2655#-Mo4p9DdDZBUuIxPdixH:-Mo4p9DdDZBUuIxPdixI:bcb8o9j) ([raw file](https://github.com/ethersphere/bee/blob/af101f9c5bd114b496f3aabbb5975b282d827986/pkg/api/metrics.go#L18)):*
> ```Go
> const bytesInKB = 1000
> 
> var fileSizeBucketsKBytes = []int64{50, 100, 500, 1000, 2000, 3000, 5000, 10000, 20000, 30000, 50000, 100000, 200000, 500000, 1000000}
> ```

not sure if so many histograms are needed, we could start simple and then expand the list (this means we need to create 15 different panels for all the different file sizes). consider just having a few histograms for now: perhaps `<100k, 100k<500k, 500k<2.5m,2.5m<4999999b`, one for exactly `5000000 bytes` for the integration test, and one for everything above that
___
*[pkg/api/metrics.go, line 64 at r2](https://reviewable.io/reviews/ethersphere/bee/2655#-Mo4oqzY0801gaG23OHK:-Mo4oqzZDUrWAcXlfDp8:bo3xyd3) ([raw file](https://github.com/ethersphere/bee/blob/af101f9c5bd114b496f3aabbb5975b282d827986/pkg/api/metrics.go#L64)):*
> ```Go
> 			Name:      "upload_duration",
> 			Help:      "Histogram of file upload API response durations.",
> 			Buckets:   []float64{0.5, 1, 2, 5, 10, 30, 60, 120, 180, 300},
> ```

not sure if explicit initialization needs to be done, as the default histogram buckets are meant to measure time
___
*[pkg/api/router.go, line 104 at r2](https://reviewable.io/reviews/ethersphere/bee/2655#-Mo4qHN8C5bToOXQwRoP:-Mo4qHN9Dh8701lxVAhf:bdxluxu) ([raw file](https://github.com/ethersphere/bee/blob/af101f9c5bd114b496f3aabbb5975b282d827986/pkg/api/router.go#L104)):*
> ```Go
> 	handle("/bzz/{address}/{path:.*}", jsonhttp.MethodHandler{
> 		"GET": web.ChainHandlers(
> 			s.bzzDownloadDurationMiddleware(),
> ```

missing also in `bytes` endpoint (upload and download)

mrekucci (approved):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2655)* status: all files reviewed, 7 unresolved discussions (waiting on @istae and @mrekucci)

*[a discussion](https://reviewable.io/reviews/ethersphere/bee/2655#-MoDBdxZ7yNA7rvPDr_7:-MoDBdxZ7yNA7rvPDr_8:bbo569c) (no related file):*
Just a nitpick, the HTTP method strings "GET", "POST", etc can be replaced with its constants defined in the `http` package `http.Method*`.

---
___
*[pkg/api/api.go, line 399 at r3](https://reviewable.io/reviews/ethersphere/bee/2655#-MoDAn1eD2pdz8Wnvx9f:-MoDAn1eD2pdz8Wnvx9g:b-l0zq6i) ([raw file](https://github.com/ethersphere/bee/blob/7e17bf42631b6809d26ed28330c200d0e16512fc/pkg/api/api.go#L399)):*
> ```Go
> 			now := time.Now()
> 			h.ServeHTTP(w, r)
> 			if method == "GET" {
> ```

I'd suggesst: "GET" -> `http.MethodGet`.
___
*[pkg/api/api.go, line 408 at r3](https://reviewable.io/reviews/ethersphere/bee/2655#-MoDB3PMAJGc648o3ofQ:-MoDB3PMAJGc648o3ofR:b4xdfam) ([raw file](https://github.com/ethersphere/bee/blob/7e17bf42631b6809d26ed28330c200d0e16512fc/pkg/api/api.go#L408)):*
> ```Go
> 					s.metrics.ContentApiDuration.WithLabelValues(fmt.Sprintf("%d", toFileSizeBucket(contentLength)), method).Observe(time.Since(now).Seconds())
> 				}
> 			} else if method == "POST" {
> ```

Also here: "POST" -> `http.MethodPost`.

pkg/api/api.go
```diff
@@ -391,6 +391,29 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
 	}
 }
 
+func (s *server) contentLengthMetricMiddleware(method string) func(h http.Handler) http.Handler {
```
acud, line 394: it is not needed to provide the `method`, since you can always get the `request.Method` from the actual request object inside the middleware

acud (changes requested):

pkg/api/api.go
```diff
@@ -391,6 +391,34 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
...
 }
 
+func (s *server) contentLengthMetricMiddleware() func(h http.Handler) http.Handler {
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			now := time.Now()
+			h.ServeHTTP(w, r)
+			switch r.Method {
+			case http.MethodGet:
+				{
```
acud, line 401: the curly braces on this level of nesting are not needed

pkg/api/api.go
```diff
@@ -391,6 +391,34 @@ func (s *server) newTracingHandler(spanName string) func(h http.Handler) http.Ha
...
+					if err != nil {
+						s.logger.Debugf("api: content length int conversation failed: %v", err)
+						return
+					}
+					if contentLength > 0 {
+						s.metrics.ContentApiDuration.WithLabelValues(fmt.Sprintf("%d", toFileSizeBucket(int64(contentLength))), r.Method).Observe(time.Since(now).Seconds())
+					}
+				}
+			case http.MethodPost:
+				{
```
acud, line 412: ditto

acud (approved):
Due to the bucket behavior, we cannot build a dashboard with the response times for the exact 5mb file upload (everything between 5-10mb will fall into the same bucket). you'll need to add another bucket of 5000001 bytes between 5 and 10mb

pkg/api/api.go
istae: i added for visual clarity, can remove it

acud (approved):
nice!

mrekucci (approved):
> Reviewed 1 of 2 files at r5, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2655)* status: all files reviewed, 11 unresolved discussions (waiting on @acud, @istae, and @mrekucci)

PR #2663, opened 2021-11-15
feat: extra linters command

notanatol:
It is meant to be run manually
```make lint-local```
 after you've committed your code. It will only take into account one last commit, but will engage an extended set of linters, potentially picking-up some older issues that you might want to take care of.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2663)

### CODE REVIEW

File: .golangci.local.yml [hunk 1]
```diff
@@ -0,0 +1,428 @@
...
+    - whitespace
+    - govet
+    - stylecheck
+    - wrapcheck
+    - gofumpt
+    - nlreturn
+    - gomnd
+    - wsl
+    - godot
+    - misspell
```
mrekucci, line 321: Don't we want to check for misspelling?

notanatol: I was trying to get the least opinionated set of linters in terms of style, but I think it makes sense to include `mispell`

mrekucci (approved):

janos (approved):

PR #2664, opened 2021-11-15
feat: make forwarder storing non-blocking

metacertain:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2664)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -232,7 +232,7 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 	// forwarding replication
 	storedChunk := false
 	defer func(stored *bool) {
-		if !stored {
+		if !*stored {
 			if ps.warmedUp() && ps.topologyDriver.IsWithinDepth(chunkAddress) {
 				verifiedChunk, err := ps.validStamp(chunk, ch.Stamp)
 				if err != nil {
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -231,39 +231,40 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 
 	// forwarding replication
 	storedChunk := false
-	if ps.warmedUp() && ps.topologyDriver.IsWithinDepth(chunkAddress) {
-
-		verifiedChunk, err := ps.validStamp(chunk, ch.Stamp)
-		if err != nil {
-			ps.metrics.InvalidStampErrors.Inc()
-			ps.logger.Warningf("pushsync: forwarder, invalid stamp for chunk %s", chunkAddress.String())
-		} else {
-			chunk = verifiedChunk
-			_, err = ps.storer.Put(ctx, storage.ModePutSync, chunk)
-			if err != nil {
-				ps.logger.Warningf("pushsync: within depth peer's attempt to store chunk failed: %v", err)
-			} else {
-				storedChunk = true
+	defer func(stored *bool) {
+		if !stored {
+			if ps.warmedUp() && ps.topologyDriver.IsWithinDepth(chunkAddress) {
+				verifiedChunk, err := ps.validStamp(chunk, ch.Stamp)
+				if err != nil {
+					ps.metrics.InvalidStampErrors.Inc()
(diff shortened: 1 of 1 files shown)


aloknerurkar (approved):

pkg/pushsync/pushsync.go
```diff
@@ -231,40 +231,41 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
-			ps.metrics.InvalidStampErrors.Inc()
-			ps.logger.Warningf("pushsync: forwarder, invalid stamp for chunk %s", chunkAddress.String())
-		} else {
-			chunk = verifiedChunk
-			_, err = ps.storer.Put(ctx, storage.ModePutSync, chunk)
-			if err != nil {
-				ps.logger.Warningf("pushsync: within depth peer's attempt to store chunk failed: %v", err)
-			} else {
-				storedChunk = true
+	defer func(stored *bool) {
```
acud, line 234: not sure why it is needed to do all this pointer juggling. if the variable is defined before the closure, you can still modify its value later and just inspect the variable value. no need for pointers

pkg/pushsync/pushsync.go
```diff
@@ -231,40 +231,41 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
+		if !*stored {
+			if ps.warmedUp() && ps.topologyDriver.IsWithinDepth(chunkAddress) {
+				verifiedChunk, err := ps.validStamp(chunk, ch.Stamp)
+				if err != nil {
+					ps.metrics.InvalidStampErrors.Inc()
+					ps.logger.Warningf("pushsync: forwarder, invalid stamp for chunk %s", chunkAddress.String())
+				} else {
+					chunk = verifiedChunk
```
acud, line 242: this was intended to be verified before forwarding with NN. with putting this in a defer, you're intervening in the sequence, and changing the protocol behavior. I think that having this code section in a defer is suboptimal. Having it somehow done at the end of the handler after all network operations have been finalized might be better.

acud (changes requested):

pkg/pushsync/pushsync.go
metacertain: so far this was verified in sequence, however, with the only consequence that the error is logged. so this seems only to defer the storing without much impact for behaviour?

acud (approved):

pkg/pushsync/pushsync.go
```diff
@@ -231,40 +231,41 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
-		} else {
-			chunk = verifiedChunk
-			_, err = ps.storer.Put(ctx, storage.ModePutSync, chunk)
-			if err != nil {
-				ps.logger.Warningf("pushsync: within depth peer's attempt to store chunk failed: %v", err)
-			} else {
-				storedChunk = true
+	defer func() {
+		if !storedChunk {
+			if ps.warmedUp() && ps.topologyDriver.IsWithinDepth(chunkAddress) {
```
istae, line 236: sad

istae (changes requested):
Nice PR, just one idea.

pkg/pushsync/pushsync.go
metacertain: I think this is a correction mechanism in itself, the in-neighborhood forwarders storing the chunk (and making it available thereby) in case of downstream peer errors

istae: If we are okay with storing the chunk multiple times as a forwarder node as a result of multiple attempts to sync a chunk, then the PR is good.

istae (approved):

pkg/pushsync/pushsync.go
metacertain: its a good question whether store.Put works efficiently in that sense, whether it avoids at least write operations in case the chunk is already stored and then attempted to be stored again. 
on a higher level, with stewardship it is not guaranteeable that the chunk doesn't get reuploaded later on, so this probably needs to be handled in the store put efficiency

acud: no chunk will be stored twice if it already exists in the store. this is ensured by localstore

mrekucci (approved):

PR #2665, opened 2021-11-16
feat: async writes to store for netstore retrieval

aloknerurkar:
This PR aims to optimize the `netstore` Get operation by doing the store Put operation in the background. This allows the chunk to be returned early for processing and unblock the client to request more chunks.

TODO:
- [ ] post measurements of download rates before+after this change cherry-picked over `v1.3.0` on production

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2665)

### DIFF
--- a/pkg/netstore/netstore.go
+++ b/pkg/netstore/netstore.go
@@ -70,6 +70,7 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 				go s.recoveryCallback(addr, targets)
 				return nil, ErrRecoveryAttempt
 			}
+			s.wg.Add(1)
 			s.put(ch, mode)
 			return ch, nil
 		}
@@ -80,7 +81,6 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 
 // put will store the chunk into storage asynchronously
 func (s *store) put(ch swarm.Chunk, mode storage.ModeGet) {
-	s.wg.Add(1)
 	go func() {
 		defer s.wg.Done()
 
@@ -96,7 +96,7 @@ func (s *store) put(ch swarm.Chunk, mode storage.ModeGet) {
 
 		stamp, err := ch.Stamp().MarshalBinary()
 		if err != nil {
-			s.logger.Errorf("netstore: failed to marshal stamp from chunk %s Err:%s", ch.Address(), err.Error())
+			s.logger.Errorf("netstore: failed to marshal stamp from chunk %s err:%s", ch.Address(), err.Error())
 			return
 		}
 
@@ -115,7 +115,7 @@ func (s *store) put(ch swarm.Chunk, mode storage.ModeGet) {
 
 		_, err = s.Storer.Put(s.sCtx, putMode, cch)
 		if err != nil {
-			s.logger.Errorf("netstore: failed to put chunk %s Err: %s", cch.Address(), err.Error())
+			s.logger.Errorf("netstore: failed to put chunk %s err: %s", cch.Address(), err.Error())
 		}
 	}()
 }
@@ -126,14 +126,14 @@ func (s *store) Close() error {
 
 	stopped := make(chan struct{})
 	go func() {
-		defer close(stopped)
 		s.wg.Wait()
+		close(stopped)
 	}()
 
 	select {
 	case <-stopped:
 		return nil
-	case <-time.After(time.Second * 5):
+	case <-time.After(5 * time.Second):
 		return errors.New("netstore: waited 5 seconds to close active goroutines")
 	}
 }
--- a/pkg/netstore/netstore.go
+++ b/pkg/netstore/netstore.go
@@ -30,7 +30,6 @@ type store struct {
 	logger           logging.Logger
 	validStamp       postage.ValidStampFn
 	recoveryCallback recovery.Callback // this is the callback to be executed when a chunk fails to be retrieved
-	opChan           chan putOp
 	bgWorkers        chan struct{}
 	sCtx             context.Context
 	sCancel          context.CancelFunc
@@ -44,7 +43,8 @@ var (
 // New returns a new NetStore that wraps a given Storer.
 func New(s storage.Storer, validStamp postage.ValidStampFn, rcb recovery.Callback, r retrieval.Interface, logger logging.Logger) storage.Storer {
 	ns := &store{Storer: s, validStamp: validStamp, recoveryCallback: rcb, retrieval: r, logger: logger}
-	ns.backgroundPutter()
+	ns.sCtx, ns.sCancel = context.WithCancel(context.Background())
+	ns.bgWorkers = make(chan struct{}, 16)
 	return ns
 }
 
@@ -66,10 +66,44 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 				go s.recoveryCallback(addr, targets)
 				return nil, ErrRecoveryAttempt
 			}
-			select {
-			case <-ctx.Done():
-			case s.opChan <- putOp{ch: ch, mode: mode}:
-			}
+			s.wg.Add(1)
+			go func() {
+				defer s.wg.Done()
+
+				select {
+				case <-s.sCtx.Done():
+					s.logger.Debug("netstore: stopping netstore")
+					return
+				case s.bgWorkers <- struct{}{}:
+				}
+				defer func() {
+					<-s.bgWorkers
+				}()
+
+				stamp, err := ch.Stamp().MarshalBinary()
+				if err != nil {
+					s.logger.Errorf("netstore: failed to marshal stamp from chunk %s Err:%s", ch.Address(), err.Error())
+					return
+				}
+
+				putMode := storage.ModePutRequest
+				if mode == storage.ModeGetRequestPin {
+					putMode = storage.ModePutRequestPin
+				}
+
+				cch, err := s.validStamp(ch, stamp)
+				if err != nil {
+					// if a chunk with an invalid postage stamp was received
+					// we force it into the cache.
+					putMode = storage.ModePutRequestCache
+					cch = ch
+				}
+
+				_, err = s.Storer.Put(s.sCtx, putMode, cch)
+				if err != nil {
+					s.logger.Errorf("netstore: failed to put chunk %s Err: %s", cch.Address(), err.Error())
+				}
+			}()
 			return ch, nil
 		}
 		return nil, fmt.Errorf("netstore get: %w", err)
@@ -98,67 +132,4 @@ func (s *store) Close() error {
 	case <-time.After(time.Second * 5):
 		return errors.New("netstore: waited 5 seconds to close active goroutines")
 	}
-}
-
-type putOp struct {
-	ch   swarm.Chunk
-	mode storage.ModeGet
-}
-
-func (s *store) backgroundPutter() {
-	s.sCtx, s.sCancel = context.WithCancel(context.Background())
-	s.bgWorkers = make(chan struct{}, 16)
-	s.opChan = make(chan putOp)
-
-	s.wg.Add(1)
-	go func() {
-		defer close(s.bgWorkers)
-		defer s.wg.Done()
-
-		for {
-			select {
-			case <-s.sCtx.Done():
-				s.logger.Debug("netstore: stopping netstore background putter")
-				return
-			case op := <-s.opChan:
-				select {
-				case <-s.sCtx.Done():
-					s.logger.Debug("netstore: stopping netstore background putter")
-					return
-				case s.bgWorkers <- struct{}{}:
-					s.wg.Add(1)
-					go func(op putOp) {
-						defer func() {
-							<-s.bgWorkers
-							s.wg.Done()
-						}()
-
-						stamp, err := op.ch.Stamp().MarshalBinary()
-						if err != nil {
-							s.logger.Errorf("netstore: failed to marshal stamp from chunk %s Err:%s", op.ch.Address(), err.Error())
-							return
-						}
-
-						putMode := storage.ModePutRequest
-						if op.mode == storage.ModeGetRequestPin {
-							putMode = storage.ModePutRequestPin
-						}
-
-						cch, err := s.validStamp(op.ch, stamp)
-						if err != nil {
-							// if a chunk with an invalid postage stamp was received
-							// we force it into the cache.
-							putMode = storage.ModePutRequestCache
-							cch = op.ch
-						}
-
-						_, err = s.Storer.Put(s.sCtx, putMode, cch)
-						if err != nil {
-							s.logger.Errorf("netstore: failed to put chunk %s Err: %s", cch.Address(), err.Error())
-						}
-					}(op)
-				}
-			}
-		}
-	}()
 }
--- a/pkg/netstore/netstore.go
+++ b/pkg/netstore/netstore.go
@@ -12,6 +12,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
+	"time"
 
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/postage"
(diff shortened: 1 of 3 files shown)


pkg/netstore/netstore.go
```diff
@@ -57,31 +66,99 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
+		return errors.New("netstore: waited 5 seconds to close active goroutines")
+	}
+}
+
+type putOp struct {
+	ch   swarm.Chunk
+	mode storage.ModeGet
+}
+
+func (s *store) backgroundPutter() {
```
acud, line 108: I think that if you went as far as introducing an extra worker that does the `Put`s in the background, wouldn't it make more sense to just do write coalescing? If you're spawning a goroutine for every `Put`, you could just avoid the whole complexity of the background worker in favor of a simple async write within the body of the `Get`(potentially with a semaphore embedded into it), which simply ignores the error:
```go
go func() { storer.Put() } ()
```
Write coalescing in this context would be just a period flushing of chunks collected to be put over a period of time.

pkg/netstore/netstore_test.go
```diff
@@ -77,6 +93,12 @@ func TestNetstoreRetrieval(t *testing.T) {
 // whenever it is found locally.
 func TestNetstoreNoRetrieval(t *testing.T) {
 	retrieve, store, nstore := newRetrievingNetstore(nil, noopValidStamp)
+	defer func() {
+		err := nstore.Close()
```
acud, line 97: i would add this directly inside `newRetrievingNetstore` inside a `t.Cleanup()` (you'll need to pass the test helper). Then this change becomes redundant across all the individual tests

acud (changes requested):

pkg/netstore/netstore.go
```diff
@@ -57,31 +66,70 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 				go s.recoveryCallback(addr, targets)
 				return nil, ErrRecoveryAttempt
 			}
-			stamp, err := ch.Stamp().MarshalBinary()
-			if err != nil {
-				return nil, err
-			}
+			s.wg.Add(1)
+			go func() {
+				defer s.wg.Done()
```
acud, line 71: wouldn't it make more sense to factor this out to a separate method on `netstore`?

pkg/netstore/netstore.go
```diff
@@ -36,7 +42,10 @@ var (
 
 // New returns a new NetStore that wraps a given Storer.
 func New(s storage.Storer, validStamp postage.ValidStampFn, rcb recovery.Callback, r retrieval.Interface, logger logging.Logger) storage.Storer {
-	return &store{Storer: s, validStamp: validStamp, recoveryCallback: rcb, retrieval: r, logger: logger}
+	ns := &store{Storer: s, validStamp: validStamp, recoveryCallback: rcb, retrieval: r, logger: logger}
+	ns.sCtx, ns.sCancel = context.WithCancel(context.Background())
+	ns.bgWorkers = make(chan struct{}, 16)
```
acud, line 47: please factor out the size of the semaphore to a `const`

pkg/netstore/netstore.go
```diff
@@ -57,31 +66,70 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
 		}
 		return nil, fmt.Errorf("netstore get: %w", err)
 	}
 	return ch, nil
 }
+
+// The underlying store is not the netstore's responsibility to close
+func (s *store) Close() error {
+	// for some reason if the backgroundPutter was not started, this should be a noop
+	if s.sCancel == nil {
```
acud, line 117: how and why is this possible, if it is set in the constructor?

pkg/netstore/netstore_test.go
```diff
@@ -158,10 +168,20 @@ func TestInvalidPostageStamp(t *testing.T) {
...
 	}
 
-	// store should have the chunk now
-	d, err := store.Get(context.Background(), storage.ModeGetRequest, addr)
-	if err != nil {
-		t.Fatal(err)
+	// store should have the chunk once the background PUT is complete
+	var d swarm.Chunk
+	start := time.Now()
+	for {
```
acud, line 174: it would be nice to factor this out to a test helper, that can do the retries, pass to it the test helper, and `t.Fatal` directly in the helper method. we do this stuff in different tests, look for "eventually" across the codebase for examples

acud (changes requested):

mrekucci (commented):
> Reviewed 2 of 3 files at r1, 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2665)* status: 2 of 3 files reviewed, 6 unresolved discussions (waiting on @acud, @aloknerurkar, @metacertain, and @mrekucci)

___
*[pkg/netstore/netstore.go, line 99 at r3](https://reviewable.io/reviews/ethersphere/bee/2665#-MocsdhA6n2BcTxlHevg:-MocsdhA6n2BcTxlHevh:b-qsnwwu) ([raw file](https://github.com/ethersphere/bee/blob/c8edb8fa103f326d14f31cd53cf0d6120f622ca5/pkg/netstore/netstore.go#L99)):*
> ```Go
> 		stamp, err := ch.Stamp().MarshalBinary()
> 		if err != nil {
> 			s.logger.Errorf("netstore: failed to marshal stamp from chunk %s Err:%s", ch.Address(), err.Error())
> ```

The `Err` shouldn't be capitalized.
___
*[pkg/netstore/netstore.go, line 118 at r3](https://reviewable.io/reviews/ethersphere/bee/2665#-MocsogQ49rDoQL96rxW:-MocsogQ49rDoQL96rxX:b-9dadgx) ([raw file](https://github.com/ethersphere/bee/blob/c8edb8fa103f326d14f31cd53cf0d6120f622ca5/pkg/netstore/netstore.go#L118)):*
> ```Go
> 		_, err = s.Storer.Put(s.sCtx, putMode, cch)
> 		if err != nil {
> 			s.logger.Errorf("netstore: failed to put chunk %s Err: %s", cch.Address(), err.Error())
> ```

Also here, the `Err` shouldn't be capitalized.
___
*[pkg/netstore/netstore.go, line 129 at r3](https://reviewable.io/reviews/ethersphere/bee/2665#-MocsImNB4OLOpBsOGNf:-MocsImO9ZsoDeuxnj5C:b-9zo6ey) ([raw file](https://github.com/ethersphere/bee/blob/c8edb8fa103f326d14f31cd53cf0d6120f622ca5/pkg/netstore/netstore.go#L129)):*
> ```Go
> 	stopped := make(chan struct{})
> 	go func() {
> 		defer close(stopped)
> ```

The `defer` call doesn't add much value here and it can be rather misleading, I'd suggest to simply calling `close` after the `Wait`.
___
*[pkg/netstore/netstore.go, line 136 at r3](https://reviewable.io/reviews/ethersphere/bee/2665#-Moct8yC9jCiphjJB9u2:-Moct8yC9jCiphjJB9u3:bdt330t) ([raw file](https://github.com/ethersphere/bee/blob/c8edb8fa103f326d14f31cd53cf0d6120f622ca5/pkg/netstore/netstore.go#L136)):*
> ```Go
> 	case <-stopped:
> 		return nil
> 	case <-time.After(time.Second * 5):
> ```

Just a nitpick: it's usually the other way around: `5 * time.Second`.
___
*[pkg/netstore/netstore_test.go, line 65 at r1](https://reviewable.io/reviews/ethersphere/bee/2665#-MoctPsOA8ySGE-rCmGT:-MoctPsOA8ySGE-rCmGU:b-gv9jux) ([raw file](https://github.com/ethersphere/bee/blob/847b76e7166c83562ee7325b528b8ec8e3594e36/pkg/netstore/netstore_test.go#L65)):*
> ```Go
> 		d, err = store.Get(context.Background(), storage.ModeGetRequest, addr)
> 		if err != nil {
> 			if time.Since(start) > time.Second*3 {
> ```

Just a nitpick: it's usually the other way around: `3*time.Second`.

pkg/netstore/netstore.go
```diff
@@ -57,31 +70,70 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
-				// we force it into the cache.
-				putMode = storage.ModePutRequestCache
-				cch = ch
-			}
-
-			_, err = s.Storer.Put(ctx, putMode, cch)
-			if err != nil {
-				return nil, fmt.Errorf("netstore retrieve put: %w", err)
-			}
+			s.put(ch, mode)
```
acud, line 73: nitpicking, but writing this as:
```go
s.wg.Add(1)
go s.put(ch, mode)
```
makes it easier to understand that there's something asynchronous which is happening here without having to navigate to the method definition.

pkg/netstore/netstore.go
```diff
@@ -57,31 +70,70 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
 }
+
+// put will store the chunk into storage asynchronously
+func (s *store) put(ch swarm.Chunk, mode storage.ModeGet) {
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+
+		select {
+		case <-s.sCtx.Done():
```
istae, line 88: you can create a new context in this function and when the quit chan is closed, the context can be canceled.

pkg/netstore/netstore.go
```diff
@@ -36,7 +46,10 @@ var (
 
 // New returns a new NetStore that wraps a given Storer.
 func New(s storage.Storer, validStamp postage.ValidStampFn, rcb recovery.Callback, r retrieval.Interface, logger logging.Logger) storage.Storer {
-	return &store{Storer: s, validStamp: validStamp, recoveryCallback: rcb, retrieval: r, logger: logger}
+	ns := &store{Storer: s, validStamp: validStamp, recoveryCallback: rcb, retrieval: r, logger: logger}
+	ns.sCtx, ns.sCancel = context.WithCancel(context.Background())
```
istae, line 50: instead of having context in the struct, you could have a quit channel instead. continued....

aloknerurkar: This is definitely one way of doing it. But there is a context required for the operation in the routines created. This is why I prefer using a background context. Achieves the same purpose and is pretty clear from the code.

aloknerurkar (commented):
> Dismissed @acud and @istae from 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2665)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @istae, @metacertain, and @mrekucci)

___
*[pkg/netstore/netstore.go, line 108 at r1](https://reviewable.io/reviews/ethersphere/bee/2665#-Moc0RnM6oZQbVajDzzA-r1-108:-MogcVZZ4yH3qPWLUctm:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/847b76e7166c83562ee7325b528b8ec8e3594e36/pkg/netstore/netstore.go#L108)):*

I think that if you went as far as introducing an extra worker that does the `Put`s in the background, wouldn't it make more sense to just do write coalescing? If you're spawning a goroutine for every `Put`, you could just avoid the whole complexity of the background worker in favor of a simple async write within the body of the `Get`(potentially with a semaphore embedded into it), which simply ignores the error:
```go
go func() { storer.Put() } ()
```
Write coalescing in this context would be just a period flushing of chunks collected to be put over a period of time.
</blockquote></details>

Done.
___
*[pkg/netstore/netstore.go, line 47 at r2](https://reviewable.io/reviews/ethersphere/bee/2665#-Moc0RnM6oZQbVajDzzA-r2-47:-MogcT9b2T4no0OUeeKU:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/e88499855b3a4b5dc5188f0a7b0f24fcc3ba46dd/pkg/netstore/netstore.go#L47)):*

please factor out the size of the semaphore to a `const`
</blockquote></details>

Done.
___
*[pkg/netstore/netstore.go, line 71 at r2](https://reviewable.io/reviews/ethersphere/bee/2665#-Moc0RnM6oZQbVajDzzA-r2-71:-MogcYdA9sO_6mICMD5A:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/e88499855b3a4b5dc5188f0a7b0f24fcc3ba46dd/pkg/netstore/netstore.go#L71)):*

wouldn't it make more sense to factor this out to a separate method on `netstore`?
</blockquote></details>

Done.
___
*[pkg/netstore/netstore.go, line 117 at r2](https://reviewable.io/reviews/ethersphere/bee/2665#-Moc0RnM6oZQbVajDzzA-r2-117:-MogciYB3nXDnYGMlLeF:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/e88499855b3a4b5dc5188f0a7b0f24fcc3ba46dd/pkg/netstore/netstore.go#L117)):*

how and why is this possible, if it is set in the constructor?
</blockquote></details>

Done.
___
*[pkg/netstore/netstore.go, line 73 at r3](https://reviewable.io/reviews/ethersphere/bee/2665#-Moc0RnM6oZQbVajDzzA-r3-73:-MogcW_cASDZV8Xfze7d:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c8edb8fa103f326d14f31cd53cf0d6120f622ca5/pkg/netstore/netstore.go#L73)):*

nitpicking, but writing this as:
```go
s.wg.Add(1)
go s.put(ch, mode)
```
makes it easier to understand that there's something asynchronous which is happening here without having to navigate to the method definition.
</blockquote></details>

Done.
___
*[pkg/netstore/netstore.go, line 88 at r3](https://reviewable.io/reviews/ethersphere/bee/2665#-Moc0RnM6oZQbVajDzzA-r3-88:-MogcZGa88JoZGkV4G0S:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c8edb8fa103f326d14f31cd53cf0d6120f622ca5/pkg/netstore/netstore.go#L88)):*

you can create a new context in this function and when the quit chan is closed, the context can be canceled.  
</blockquote></details>

Done.
___
*[pkg/netstore/netstore_test.go, line 97 at r1](https://reviewable.io/reviews/ethersphere/bee/2665#-Moc0RnM6oZQbVajDzzB-r1-97:-MogcySv7KGqtZqhY0Ai:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/847b76e7166c83562ee7325b528b8ec8e3594e36/pkg/netstore/netstore_test.go#L97)):*

i would add this directly inside `newRetrievingNetstore` inside a `t.Cleanup()` (you'll need to pass the test helper). Then this change becomes redundant across all the individual tests
</blockquote></details>

Done.
___
*[pkg/netstore/netstore_test.go, line 174 at r2](https://reviewable.io/reviews/ethersphere/bee/2665#-Moc0RnM6oZQbVajDzzB-r2-174:-MogczMcEFyEV41yYEI-:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/e88499855b3a4b5dc5188f0a7b0f24fcc3ba46dd/pkg/netstore/netstore_test.go#L174)):*

it would be nice to factor this out to a test helper, that can do the retries, pass to it the test helper, and `t.Fatal` directly in the helper method. we do this stuff in different tests, look for "eventually" across the codebase for examples
</blockquote></details>

Done.

pkg/netstore/netstore.go
```diff
@@ -57,31 +70,70 @@ func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
...
 		return nil, fmt.Errorf("netstore get: %w", err)
 	}
 	return ch, nil
 }
+
+// put will store the chunk into storage asynchronously
+func (s *store) put(ch swarm.Chunk, mode storage.ModeGet) {
+	go func() {
+		defer s.wg.Done()
+
```
aloknerurkar, line 86: @acud 
Should we create a new `context.WithTimeout` here? In the case of gateway nodes, retrieving a lot of files, we might see a pile-up here.

acud: let's do it if the problem indeed occurs. i don't think that the gateways are that active

acud (approved):
OK. But a correction - what is being done here is not write coalescing. Write coalescing means combining multiple writes into one, saving on individual operations. What is done here is just async writes.

istae (approved):

mrekucci (approved):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2665)* status: all files reviewed, 2 unresolved discussions (waiting on @aloknerurkar, @istae, and @metacertain)

PR #2668, opened 2021-11-17
refactor: chi router

istae:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2668)

### DIFF
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -251,17 +251,15 @@ func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
 
 	nameOrHex := chi.URLParam(r, "address")
 
-	url := strings.Split(strings.TrimSpace(r.URL.Path), "/")
-	if url[1] == "bzz" {
+	// extract the file or dir path from URL
+	url := strings.Split(r.URL.Path, "/")
+	if url[1] == "bzz" { // /bzz/...
 		url = url[3:]
-	} else {
+	} else { // /v1/bzz/...
 		url = url[4:]
 	}
 	pathVar := strings.Join(url, "/")
-
-	if strings.HasSuffix(pathVar, "/") {
-		pathVar = strings.TrimRight(pathVar, "/")
-		// NOTE: leave one slash if there was some
+	if pathVar != "" && strings.HasSuffix(r.URL.Path, "/") {
 		pathVar += "/"
 	}
 
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -251,8 +251,14 @@ func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
 
 	nameOrHex := chi.URLParam(r, "address")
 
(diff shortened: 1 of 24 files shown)


pkg/api/bzz.go
```diff
@@ -249,8 +249,16 @@ func (s *server) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
...
+	nameOrHex := chi.URLParam(r, "address")
+
+	url := strings.Split(strings.TrimSpace(r.URL.Path), "/")
+	if url[1] == "bzz" {
+		url = url[3:]
+	} else {
+		url = url[4:]
+	}
+	pathVar := strings.Join(url, "/")
+
```
notanatol, line 261: Generally I think `chi` offers a cleaner approach to routing, but I'm not sure what the new logic above is doing?

istae: aha! so the regex url variable only works for one subroute, meaning if you have a url like this:
`/bzz/aaa/lorem/ipsum.txt` and your path is like this `/bzz/{address}/{path:.*}`, the chi regex variable path only captures the lorem part, this worked right out of the box for gorilla/mux, but not chi.

pkg/api/router.go
```diff
@@ -25,198 +24,108 @@ func (s *server) setupRouting() {
...
 	})
 
-	handle("/chunks", jsonhttp.MethodHandler{
-		"POST": web.ChainHandlers(
-			jsonhttp.NewMaxBodyBytesHandler(swarm.ChunkWithSpanSize),
-			web.FinalHandlerFunc(s.chunkUploadHandler),
-		),
+	api.Route("/chunks", func(r chi.Router) {
+		r.With(jsonhttp.NewMaxBodyBytesHandler(swarm.ChunkWithSpanSize)).Post("/", s.chunkUploadHandler)
+		r.Get("/{addr}", s.chunkGetHandler)
```
acud, line 74: not sure if i really like this structure. I find it a bit confusing and harder to read

acud (commented):
Not sure about whether this change is needed. Personally I find it more difficult to read (I might not be well-versed in reading this sort router format)

pkg/api/router.go
istae: we could inline it, instead of nesting it like this. The idea is for complex REST apis, this type of structuring simplifies things.

pkg/api/bzz_test.go
```diff
@@ -34,9 +34,9 @@ import (
 
 func TestBzzFiles(t *testing.T) {
 	var (
-		fileUploadResource   = "/bzz"
+		fileUploadResource   = "/v1/bzz"
```
acud, line 37: in the current implementation, both schemes are supported (you can call either `/bzz` or `/v1/bzz`). i would suggest to keep it this way

istae: so some of our unit tests does not test if /v1 works, so I just changed one of them to test it. If I had removed the /v1 routing, the beekeper tests would all be broken.

Issue #2521, opened 2021-09-16
"Balanced" connector in Kademia isn't balanced


ldeffenb:
#### Summary
While monitoring my 6 nodes in the mainnet, I noticed that 2 of them consistently have full bin zero connections (20) while the other 4 get up to 4 or 8 connections in bin zero and hover there.  For the node that only gets up to 4, it sometimes drops the depth from 12 to 0 if one of them disconnects and then the depth goes back to 12 once the bin 0 recovers to at least 4 connections.

Hint on a possible fix for this behavior is at the very bottom...

#### Steps to reproduce
Monitor the /topology low numbered bin connections across a random collection of peer IDs and you'll likely see the pattern.  You will want to expand their peer IDs to binary to see the string of zero bits (or not) beyond the first 3 or 4 bits of their peer IDs.

#### Expected behavior
I would expect the /topology bin connection counts to be independent of the peer ID, especially for low numbered bins in a large network like mainnet.  I mean, for bin 0, one half of the total swarm size are valid connection candidates!

#### Actual behavior
Because of the way the less-than-depth outbound connections are handled in kademlia.go's connectBalanced routine based on the proximity to commonBinPrefixes, there is a definite preference for connecting to peers with more zeroes just after their leading bits.  Peers with non-zero (one) bits in the most significant bits (after 3 or 4 bits) of their peer ID will not see as many incoming connections causing them to hover at or under saturationPeers and never come close to overSaturationPeers.

But peers with zeros in their most significant bits (after 3 or 4 bits) will nearly see a DOS attack as other nodes continually select them for "balanced" connections, even when they have long since reached their overSaturationPeers limit.  This can be observed by watching the "don't want incoming peer" messages in their logs.

My peer ID that starts with 6127 (0110 0001 0010) rejects 30 or more incoming peers every minute, 24x7:
```
time="2021-09-16T21:59:01Z" level=warning msg="stream handler: don't want incoming peer 44a499abfd9b214bf8171702bd2318f371ea76846ab76182faa3239e19455788. disconnecting"
time="2021-09-16T21:59:06Z" level=warning msg="stream handler: don't want incoming peer 735c85ea0cbb67e3af73269d11e8364800aa1e72538261da51bae0c1d3114915. disconnecting"
time="2021-09-16T21:59:07Z" level=warning msg="stream handler: don't want incoming peer 5821b89185402fceb01dcb0590c4862555e5ae44b6d5ea57f867464fbe91a3db. disconnecting"
[...]

PR #2677, opened 2021-11-23
feat: relax balancing

metacertain:
### Description

This PR intends to change the selection of balanced connections by relaxing the criteria from selecting a closest peer to selecting a close-enough peer for a pseudo-address

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/2521

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2677)

### DIFF
--- a/pkg/topology/kademlia/kademlia.go
+++ b/pkg/topology/kademlia/kademlia.go
@@ -231,40 +231,24 @@ func (k *Kad) connectBalanced(wg *sync.WaitGroup, peerConnChan chan<- *peerConnI
 		}
 
 		binPeers := k.knownPeers.BinPeers(uint8(i))
+		binConnectedPeers := k.connectedPeers.BinPeers(uint8(i))
 
 		for j := range k.commonBinPrefixes[i] {
 			pseudoAddr := k.commonBinPrefixes[i][j]
 
-			closestConnectedPeer, err := closestPeer(k.connectedPeers, pseudoAddr, noopSanctionedPeerFn)
-			if err != nil {
-				if errors.Is(err, topology.ErrNotFound) {
-					break
-				}
-				k.logger.Errorf("closest connected peer: %v", err)
-				continue
-			}
-
-			closestConnectedPO := swarm.ExtendedProximity(closestConnectedPeer.Bytes(), pseudoAddr.Bytes())
-			if int(closestConnectedPO) >= i+k.bitSuffixLength+1 {
-				continue
-			}
-
 			// Connect to closest known peer which we haven't tried connecting to recently.
-			closestKnownPeer, err := closestPeerInSlice(binPeers, pseudoAddr, skipPeers)
-			if err != nil {
-				if errors.Is(err, topology.ErrNotFound) {
-					break
-				}
-				k.logger.Errorf("closest known peer: %v", err)
+
+			_, exists := nClosePeerInSlice(binConnectedPeers, pseudoAddr, noopSanctionedPeerFn, uint8(i+k.bitSuffixLength+1))
+			if exists {
+				continue
+			}
+
+			closestKnownPeer, exists := nClosePeerInSlice(binPeers, pseudoAddr, skipPeers, uint8(i+k.bitSuffixLength+1))
+			if !exists {
 				continue
 			}
 
 			if k.connectedPeers.Exists(closestKnownPeer) {
-				continue
-			}
-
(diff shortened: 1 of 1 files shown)


istae (approved):
perfect, thanks,.

pkg/topology/kademlia/kademlia.go
```diff
@@ -231,43 +231,35 @@ func (k *Kad) connectBalanced(wg *sync.WaitGroup, peerConnChan chan<- *peerConnI
...
 
-			closestConnectedPeer, err := closestPeer(k.connectedPeers, pseudoAddr, noopSanctionedPeerFn)
+			// Connect to closest known peer which we haven't tried connecting to recently.
+
+			_, err := nClosePeerInSlice(binConnectedPeers, pseudoAddr, noopSanctionedPeerFn, uint8(i+k.bitSuffixLength+1))
 			if err != nil {
-				if errors.Is(err, topology.ErrNotFound) {
-					break
+				if !errors.Is(err, topology.ErrNotFound) {
+					break // not possible
```
istae, line 244: technically if there is an error here that's not topology.ErrNotFound, we should return, instead of break.

istae (approved):

pkg/topology/kademlia/kademlia.go
```diff
@@ -231,43 +231,35 @@ func (k *Kad) connectBalanced(wg *sync.WaitGroup, peerConnChan chan<- *peerConnI
...
 
-			// Connect to closest known peer which we haven't tried connecting to recently.
-			closestKnownPeer, err := closestPeerInSlice(binPeers, pseudoAddr, skipPeers)
+			closestKnownPeer, err := nClosePeerInSlice(binPeers, pseudoAddr, skipPeers, uint8(i+k.bitSuffixLength+1))
 			if err != nil {
 				if errors.Is(err, topology.ErrNotFound) {
-					break
+					continue
+				} else {
+					return
```
istae, line 255: return here?

istae (approved):

pkg/topology/kademlia/kademlia.go
metacertain: yes, no other error than topology.ErrNotFound is expected here

pkg/topology/kademlia/kademlia.go
```diff
@@ -231,43 +231,35 @@ func (k *Kad) connectBalanced(wg *sync.WaitGroup, peerConnChan chan<- *peerConnI
...
 
-			closestConnectedPeer, err := closestPeer(k.connectedPeers, pseudoAddr, noopSanctionedPeerFn)
+			// Connect to closest known peer which we haven't tried connecting to recently.
+
+			_, err := nClosePeerInSlice(binConnectedPeers, pseudoAddr, noopSanctionedPeerFn, uint8(i+k.bitSuffixLength+1))
 			if err != nil {
-				if errors.Is(err, topology.ErrNotFound) {
-					break
+				if !errors.Is(err, topology.ErrNotFound) {
+					return
```
acud, line 244: there was not a return here, why the change?

metacertain: this function does not return any other error than ErrNotFound, so this is non-consequential, I can remove this

acud: that's incorrect. there are 2 levels of `for` loops here, the `break` statement means that the immediate enclosing for loop breaks, but the enclosing loop continues

pkg/topology/kademlia/kademlia.go
metacertain: this return is also non-consequential, it can be removed

acud: why return and not `break`? please see the comment on the other change

pkg/topology/kademlia/kademlia.go
metacertain: the break was in a different context. 
closestpeer to any pseudoaddress always existed if there was at least some connections in the bins.
Now, closeenoughPeer can return a close-enough-peer if there is one, or nothing. 
if there's a close enough connected peer, we skip to next pseudoaddress (continue).
if there is not one, we select one in line 250. 
if there is not one, we get the topology.ErrNotFound error. if any other error is received, probably something is off, so return. this can never happen with the current nClosePeer function

acud: interactive review please

acud (approved):

PR #2680, opened 2021-11-23
perf: preemptive pushsync retries

istae:


mrekucci (commented):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2680)* status: all files reviewed, 1 unresolved discussion (waiting on @istae)

___
*[pkg/pushsync/pushsync.go, line 89 at r2](https://reviewable.io/reviews/ethersphere/bee/2680#-MpG37Sy5g5m6qxeEyHX:-MpG37T2DVjjvFQqvJmO:br07r01) ([raw file](https://github.com/ethersphere/bee/blob/1cf4e29315ecf6af130093ac56fb5eefd09993bd/pkg/pushsync/pushsync.go#L89)):*
> ```Go
> 
> type receiptResult struct {
> 	t         time.Time
> ```

The `t` doesn't tell much here I'd suggest renaming it to `pushDuration` or something similar.

mrekucci (approved):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2680)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @istae)

pkg/pushsync/pushsync.go
```diff
@@ -352,129 +365,173 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
+			time.AfterFunc(defaultTTL, cancel)
+
+			go ps.pushPeer(ctxd, resultChan, peer, ch, origin)
+
+			if retries <= 0 {
+				return nil, ErrNoPush
 			}
+
+			// retry
+			timer.Reset(p90TTL)
```
acud, line 411: why not a simple ticker?

acud (commented):
nice, i have a few minor comments

pkg/pushsync/pushsync.go
istae: If an attempt fails under 5 seconds, we shouldn't have to wait till the next tick.
With a timer its easier to control when the next request should fire.

acud: makes sense :+1:

pkg/pushsync/pushsync.go
```diff
@@ -352,129 +368,175 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
+	var (
+		err       error
+		receipt   pb.Receipt
+		attempted bool
+		now       = time.Now()
+	)
+
+	defer func() {
+		select {
+		case resultChan <- receiptResult{pushTime: now, peer: peer, err: err, attempted: attempted, receipt: &receipt}:
```
aloknerurkar, line 482: So I am not able to see where we limit the no of in-flight requests to two. Could it happen that we do more than 2 requests and then lose some results here?

istae: We could actually have many inflight requests. It's limited by allowedAttempts.
In the case that a receipt is received from some request, the results from the other requests are ignored.

acud (approved):
Good stuff! thanks!

aloknerurkar (approved):

PR #2682, opened 2021-11-24
Sharky integration.0

aloknerurkar:
### Checklist


pkg/localstore/localstore.go
```diff
@@ -204,7 +205,7 @@ type Options struct {
 // New returns a new DB.  All fields and indexes are initialized
 // and possible conflicts with schema from existing database is checked.
 // One goroutine for writing batches is created.
-func New(path string, baseKey []byte, ss storage.StateStorer, o *Options, logger logging.Logger) (db *DB, err error) {
+func New(ctx context.Context, path string, baseKey []byte, ss storage.StateStorer, o *Options, logger logging.Logger) (db *DB, err error) {
```
aloknerurkar, line 208: Not sure about this. This would then mean that the Localstore would then have to adhere to context semantics. This is not the case. We have `db.Close` which should be used to close the localstore. So for `collectGarbage` you can use the background context or a cancellable context which is canceled on localstore.Close.

pkg/localstore/gc.go
```diff
@@ -218,6 +227,11 @@ func (db *DB) collectGarbage() (collectedCount uint64, done bool, err error) {
 		db.metrics.GCErrorCounter.Inc()
 		return 0, false, err
 	}
+
+	for _, loc := range locations {
```
aloknerurkar, line 231: Instead of passing a context here, you can create a context that is canceled when localstore.Close is called (which could close some new channel to signal it's called). Which would stop the garbage collection when the user calls a close.

pkg/localstore/localstore.go
```diff
@@ -562,6 +623,7 @@ func (db *DB) Close() (err error) {
 	// wait for all handlers to finish
 	done := make(chan struct{})
 	go func() {
+		db.cancel()
```
aloknerurkar, line 626: This can be canceled at the start of the Close function itself.

acud (changes requested):
Very nicely done. I have some minor comments!

> Reviewed 4 of 10 files at r3, 1 of 1 files at r5, 8 of 13 files at r7, 11 of 11 files at r8, 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2682#-:-MpubCp-8WWf3Zk1FNQo:b-hdm70u)* status: all files reviewed, 9 unresolved discussions (waiting on @aloknerurkar, @janos, and @zelig)

___
*[pkg/localstore/export.go, line 68 at r9](https://reviewable.io/reviews/ethersphere/bee/2682#-MpuVYE49ukMvqd8k4Vv:-MpuVYE49ukMvqd8k4Vw:b-4f324w) ([raw file](https://github.com/ethersphere/bee/blob/df91e66d779a9b9e76ef8131b432f0ca4275ec98/pkg/localstore/export.go#L68)):*
> ```Go
> 		}
> 
> 		ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
> ```

you can create just one context.Background() before the Iterate call and reuse it. Maybe a timeout is not needed here, since the process should bubble up the context cancellation on a `SIGKILL/QUIT`
___
*[pkg/localstore/localstore.go, line 626 at r8](https://reviewable.io/reviews/ethersphere/bee/2682#-MpH9djfAe_ey-j_mA5e-r8-626:-MpuWdDK-BjlmkcjHhe0:b-84wvtf) ([raw file](https://github.com/ethersphere/bee/blob/b122050248c62793747a22ad2eec3fa9879bb2d0/pkg/localstore/localstore.go#L626)):*

This can be canceled at the start of the Close function itself.
</blockquote></details>

:+1: no need to wait for the goroutine to get scheduled in order to call the cancel
___
*[pkg/localstore/localstore.go, line 496 at r9](https://reviewable.io/reviews/ethersphere/bee/2682#-MpuW7qB8QyhujwNabCw:-MpuW7qB8QyhujwNabCx:b-srerv9) ([raw file](https://github.com/ethersphere/bee/blob/df91e66d779a9b9e76ef8131b432f0ca4275ec98/pkg/localstore/localstore.go#L496)):*
> ```Go
> 		},
> 		EncodeValue: func(fields shed.Item) (value []byte, err error) {
> 			value = make([]byte, 40+sharky.LocationSize)
> ```

note this might change with viktor's PR
___
*[pkg/localstore/mode_put.go, line 53 at r9](https://reviewable.io/reviews/ethersphere/bee/2682#-MpuX_3KClCm0-x9zOe_:-MpuX_3KClCm0-x9zOea:b-hnca5s) ([raw file](https://github.com/ethersphere/bee/blob/df91e66d779a9b9e76ef8131b432f0ca4275ec98/pkg/localstore/mode_put.go#L53)):*
> ```Go
> }
> 
> type releaseLocations []*sharky.Location
> ```

not sure why `sharky.Location` is a pointer everywhere. It should be an instance, since it gets dereferenced everywhere anyway no?
___
*[pkg/localstore/mode_put.go, line 216 at r9](https://reviewable.io/reviews/ethersphere/bee/2682#-MpuYuq198EhdcaBYob8:-MpuYuq198EhdcaBYob9:b-3aeteb) ([raw file](https://github.com/ethersphere/bee/blob/df91e66d779a9b9e76ef8131b432f0ca4275ec98/pkg/localstore/mode_put.go#L216)):*
> ```Go
> 	}
> 
> 	for _, v := range *releaseLocs {
> ```

doesn't this mean that we essentially releasing (and therefore deleting) all the chunks that we just referenced from the retrieval data index? isn't this supposed to be executed when `WriteBatch` fails in the next block up?
___
*[pkg/sharky/shard.go, line 39 at r9](https://reviewable.io/reviews/ethersphere/bee/2682#-MpuZufz1HsBHlD3AaCW:-MpuZufz1HsBHlD3AaCX:b-36syzd) ([raw file](https://github.com/ethersphere/bee/blob/df91e66d779a9b9e76ef8131b432f0ca4275ec98/pkg/sharky/shard.go#L39)):*
> ```Go
> func (l *Location) UnmarshalBinary(buf []byte) error {
> 	l.Shard = buf[0]
> 	l.Offset, _ = binary.Varint(buf[1:9])
> ```

errors should be checked
___
*[pkg/sharky/shard.go, line 44 at r9](https://reviewable.io/reviews/ethersphere/bee/2682#-Mpu_JVVDfMdcNFmHrth:-Mpu_JVVDfMdcNFmHrti:b5z8l0s) ([raw file](https://github.com/ethersphere/bee/blob/df91e66d779a9b9e76ef8131b432f0ca4275ec98/pkg/sharky/shard.go#L44)):*
> ```Go
> }
> 
> func LocationFromBinary(buf []byte) (*Location, error) {
> ```

this should not return a pointer, since it is dereferenced everywhere

aloknerurkar:
___
*[pkg/localstore/mode_put.go, line 216 at r9](https://reviewable.io/reviews/ethersphere/bee/2682#-MpuYuq198EhdcaBYob8:-MpukTyk4O6YbWg3xWZi:b-iwm4j5) ([raw file](https://github.com/ethersphere/bee/blob/df91e66d779a9b9e76ef8131b432f0ca4275ec98/pkg/localstore/mode_put.go#L216)):*

doesn't this mean that we essentially releasing (and therefore deleting) all the chunks that we just referenced from the retrieval data index? isn't this supposed to be executed when `WriteBatch` fails in the next block up?
</blockquote></details>

The committedLocations has the new locations that were committed. These are released if we fail.

The releaseLocations are the locations that were overridden due to the postage stamp logic. These are released only after the WriteBatch is successful.

aloknerurkar (commented):
> Dismissed @acud from 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2682)* status: 16 of 27 files reviewed, 2 unresolved discussions (waiting on @acud, @aloknerurkar, @janos, and @zelig)

___
*[pkg/localstore/export.go, line 68 at r9](https://reviewable.io/reviews/ethersphere/bee/2682#-MpuVYE49ukMvqd8k4Vv:-MpukCNZ6PFLyY4gjeC4:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/df91e66d779a9b9e76ef8131b432f0ca4275ec98/pkg/localstore/export.go#L68)):*

you can create just one context.Background() before the Iterate call and reuse it. Maybe a timeout is not needed here, since the process should bubble up the context cancellation on a `SIGKILL/QUIT`
</blockquote></details>

Done.
___
*[pkg/localstore/mode_put.go, line 53 at r9](https://reviewable.io/reviews/ethersphere/bee/2682#-MpuX_3KClCm0-x9zOe_:-MpukSPI8OL4cSfrrQ-9:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/df91e66d779a9b9e76ef8131b432f0ca4275ec98/pkg/localstore/mode_put.go#L53)):*

not sure why `sharky.Location` is a pointer everywhere. It should be an instance, since it gets dereferenced everywhere anyway no?
</blockquote></details>

Done.
___
*[pkg/sharky/shard.go, line 39 at r9](https://reviewable.io/reviews/ethersphere/bee/2682#-MpuZufz1HsBHlD3AaCW:-MpukoouCKFnl7Py3O7z:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/df91e66d779a9b9e76ef8131b432f0ca4275ec98/pkg/sharky/shard.go#L39)):*

errors should be checked
</blockquote></details>

Done.
___
*[pkg/sharky/shard.go, line 44 at r9](https://reviewable.io/reviews/ethersphere/bee/2682#-Mpu_JVVDfMdcNFmHrth:-MpukpuF0IAEYPFvBeBc:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/df91e66d779a9b9e76ef8131b432f0ca4275ec98/pkg/sharky/shard.go#L44)):*

this should not return a pointer, since it is dereferenced everywhere
</blockquote></details>

Done.

acud (changes requested):
> Reviewed 10 of 10 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2682)* status: all files reviewed, 3 unresolved discussions (waiting on @aloknerurkar, @janos, and @zelig)

___
*[pkg/localstore/mode_put.go, line 216 at r9](https://reviewable.io/reviews/ethersphere/bee/2682#-MpuYuq198EhdcaBYob8:-MpvJKkl3Zrbbr6zAhBA:b5i3igo) ([raw file](https://github.com/ethersphere/bee/blob/df91e66d779a9b9e76ef8131b432f0ca4275ec98/pkg/localstore/mode_put.go#L216)):*

The committedLocations has the new locations that were committed. These are released if we fail.

The releaseLocations are the locations that were overridden due to the postage stamp logic. These are released only after the WriteBatch is successful.
</blockquote></details>

hmmm... i might have to re-read then...
___
*[pkg/localstore/mode_put.go, line 55 at r11](https://reviewable.io/reviews/ethersphere/bee/2682#-MpvJ1e-32O6Pq7BvTM-:-MpvJ1e-32O6Pq7BvTM0:bpnxb74) ([raw file](https://github.com/ethersphere/bee/blob/6ce1336c756287111838cd862016da53926859ec/pkg/localstore/mode_put.go#L55)):*
> ```Go
> type releaseLocations []sharky.Location
> 
> func (r *releaseLocations) add(loc sharky.Location) {
> ```

I wonder whether this type is needed... also the confusing dereferencing of the pointer on L216. Wouldn't it be clearer to just pass a normal slice?

aloknerurkar:
___
*[pkg/localstore/mode_put.go, line 55 at r11](https://reviewable.io/reviews/ethersphere/bee/2682#-MpvJ1e-32O6Pq7BvTM-:-Mpvb_ytE1zY9r9eWB-G:bjw6rmw) ([raw file](https://github.com/ethersphere/bee/blob/6ce1336c756287111838cd862016da53926859ec/pkg/localstore/mode_put.go#L55)):*

I wonder whether this type is needed... also the confusing dereferencing of the pointer on L216. Wouldn't it be clearer to just pass a normal slice?
</blockquote></details>

The releaseLocations array is not deterministic. Most cases we might not have anything here. This is why we need to use `append`. As `append` mutates the slice pointer, we cannot pass a slice directly. We tested this on go playground. Which is why we made this type to prevent use of `*[]*sharky.Location`. I think this is more readable.

pkg/sharky/shard.go
```diff
@@ -5,18 +5,79 @@
...
 	Shard  uint8
 	Offset int64
 	Length int64
 }
 
+// MarshalBinary returns byte representation of location. We use binary.PutVarint
+// which can take upto 10bytes to marshal int64. Realistically, we will never have
+// values here that would require extra bytes to pack. The offset is currently capped
+// and the length is deterministic. In order to save the extra bytes, we use the min
+// required bytes here which is 8
```
janos, line 29: To be completely correct that capping extra two bytes does not make provide any possibility for error, I would suggest to use uint64 for Location.Offset and Location.Length and to encode them with PutUint64 which already requires 8 bytes and both offset and length values should not be negative.

pkg/localstore/localstore.go
```diff
@@ -292,9 +331,28 @@ func New(path string, baseKey []byte, ss storage.StateStorer, o *Options, logger
...
+	}
+	db.sharky, err = sharky.New(sharkyBase, sharkyNoOfShards, sharkyPerShardLimit)
+	if err != nil {
+		return nil, err
+	}
+
 	// Index storing actual chunk address, data and bin id.
 	headerSize := 16 + postage.StampSize
-	db.retrievalDataIndex, err = db.shed.NewIndex("Address->StoreTimestamp|BinID|BatchID|BatchIndex|Sig|Data", shed.IndexFuncs{
+	db.retrievalDataIndex, err = db.shed.NewIndex("Address->StoreTimestamp|BinID|BatchID|BatchIndex|Sig|Location", shed.IndexFuncs{
```
janos, line 355: Just to mention that this (and gcIndex) change requires a migration. Probably for the followup PR.

janos (commented):
Looks very nice.

pkg/localstore/localstore.go
acud: thanks @janos, there's already a migration PR in progress. But why is a migration to gcIndex required? I might have overlooked something

janos: gcIndex is also changed  to include the location in value. The location should be appended to every value in that index.

notanatol (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2682)* status: all files reviewed, 5 unresolved discussions (waiting on @acud, @aloknerurkar, and @zelig)

pkg/localstore/gc.go
```diff
@@ -165,6 +166,8 @@ func (db *DB) collectGarbage() (collectedCount uint64, done bool, err error) {
 		return 0, false, err
 	}
 
+	locations := make([]sharky.Location, 0, len(candidates))
+
 	// get rid of dirty entries
 	for _, item := range candidates {
 		if swarm.NewAddress(item.Address).MemberOf(db.dirtyAddresses) {
```
acud, line 173: `L177`:
```go
i, err := db.retrievalDataIndex.Get(item)
if err ....
item.Location = i.Location
```

notanatol: fixed

acud (changes requested):
> Reviewed 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2682)* status: all files reviewed, 4 unresolved discussions (waiting on @acud, @aloknerurkar, and @zelig)

___
*[pkg/localstore/localstore.go, line 481 at r12](https://reviewable.io/reviews/ethersphere/bee/2682#-MqEmjXYBZTOmUqCHGR8:-MqEmjXYBZTOmUqCHGR9:b-bdiw13) ([raw file](https://github.com/ethersphere/bee/blob/6c5b250d45138accad3a26565d596e1ea6ce679e/pkg/localstore/localstore.go#L481)):*
> ```Go
> 	db.pushTriggers = make([]chan<- struct{}, 0)
> 	// gc index for removable chunk ordered by ascending last access time
> 	db.gcIndex, err = db.shed.NewIndex("AccessTimestamp|BinID|Hash->BatchID|BatchIndex|Location", shed.IndexFuncs{
> ```

the addition of location here also needs to be reverted

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2682)* status: 26 of 27 files reviewed, 4 unresolved discussions (waiting on @acud, @aloknerurkar, and @zelig)

___
*[pkg/localstore/localstore.go, line 481 at r12](https://reviewable.io/reviews/ethersphere/bee/2682#-MqEmjXYBZTOmUqCHGR8:-MqEt-e-14IaSg_nMmOi:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/6c5b250d45138accad3a26565d596e1ea6ce679e/pkg/localstore/localstore.go#L481)):*

the addition of location here also needs to be reverted
</blockquote></details>

fixed

mrekucci (commented):
> Reviewed 2 of 11 files at r8, 1 of 10 files at r10, 1 of 1 files at r12, 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2682)* status: all files reviewed, 4 unresolved discussions (waiting on @acud, @aloknerurkar, and @zelig)

___
*[pkg/localstore/export.go, line 60 at r13](https://reviewable.io/reviews/ethersphere/bee/2682#-MqPgRjKBzzDUDdcSlUN:-MqPgRjLCzvFP8cMGnpE:b-r9pwzd) ([raw file](https://github.com/ethersphere/bee/blob/6634fd147d7b38adb8c5c63f7f3a12cdec17c68f/pkg/localstore/export.go#L60)):*
> ```Go
> 	}
> 
> 	ctx := context.Background()
> ```

The `db.ctx` should be used as a parent.
___
*[pkg/localstore/export.go, line 83 at r13](https://reviewable.io/reviews/ethersphere/bee/2682#-MqPgnGD7DpIfP0Un4L8:-MqPgnGD7DpIfP0Un4L9:b-l9lz5j) ([raw file](https://github.com/ethersphere/bee/blob/6634fd147d7b38adb8c5c63f7f3a12cdec17c68f/pkg/localstore/export.go#L83)):*
> ```Go
> 			return false, err
> 		}
> 		if _, err := tw.Write(item.BatchID); err != nil {
> ```

This may be simplified as:
```golang
err = nil
write := func(buf []byte) {
	if err != nil {
		return
	}
	_, err = tw.Write(buf)
}
write(item.BatchID)
write(item.Index)
write(item.Timestamp)
write(item.Sig)
write(data)
if err != nil {
	return false, err
}
```
See: https://go.dev/blog/errors-are-values

janos (approved):

pkg/localstore/localstore.go
```diff
@@ -57,11 +62,20 @@ var (
 	flipFlopWorstCaseDuration = 10 * time.Second
 )
 
+const (
+	// 32 * 312500 chunks = 1000000 chunks (40GB)
```
aloknerurkar, line 66: This should be `10000000`

mrekucci: Done.

mrekucci (approved):

PR #2694, opened 2021-11-29
fix: storer node forwarder replication edge case

istae:
Current implementation allows forwarding replication logic to run when the storer node fails to validate or store the chunk. PR fixes that.

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2694)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -248,11 +248,11 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 			if err != nil {
 				ps.metrics.InvalidStampErrors.Inc()
 				ps.logger.Warningf("pushsync: forwarder, invalid stamp for chunk %s", chunkAddress.String())
-			} else {
-				_, err = ps.storer.Put(ctx, storage.ModePutSync, verifiedChunk)
-				if err != nil {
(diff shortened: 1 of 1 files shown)


acud:
@istae this is the intended behavior

pkg/pushsync/pushsync.go
```diff
@@ -241,20 +241,17 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
 	defer func() {
-		if !storedChunk {
-			if ps.warmedUp() && ps.topologyDriver.IsWithinDepth(chunkAddress) {
-				verifiedChunk, err := ps.validStamp(chunk, ch.Stamp)
+		if !storerNode && ps.warmedUp() && ps.topologyDriver.IsWithinDepth(chunkAddress) {
+			verifiedChunk, err := ps.validStamp(chunk, ch.Stamp)
+			if err != nil {
+				ps.metrics.InvalidStampErrors.Inc()
+				ps.logger.Warningf("pushsync: forwarder, invalid stamp for chunk %s", chunkAddress.String())
+			} else {
```
acud, line 251: since we are in a closure, you can just `return` inside the error block, and remove the `else` clause to reduce nesting

acud (approved):

aloknerurkar (approved):

aloknerurkar:
Good catch!

PR #2701, opened 2021-11-30
feat: verify chunk availability before concludin synced status

metacertain:
### Checklist

- [x] Quick rebase of previous verify_independent branch

### Description
 After receiving an acceptable storage receipt for a chunk in pusher, before concluding synced status, try retrieving the chunk through far away peers to get some measure of chunk availability
 
Reliability improvement

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2701)


acud:
I'd would like to have measurements on how this affects the syncing speed before reviewing this.

metacertain:
So syncing is expected to be slower or slightly slower (though it can be probably mediated by raising the pusher concurrency limits), but another measure, actual content availability is also relevant, as the main intention of this is not to increase syncing speed (that's not necessarily relevant from a user experience perspective) but to decrease the time for the content to become reliably retrievable from anywhere in the network without having to resort to user-intervened stewardship efforts

PR #2703, opened 2021-11-30
fix(pusher): feed channel respects warmup

### LINKED ISSUE #2702: v1.4.1-RC1 doesn't connect peers until after warmup.
### Context 
Running v1.4.1-RC1

### Summary
When a node starts, it is supposed to connect to the peers to establish the depth before the pusher and puller kick off after the warmup period.  Starting with v1.4.1-RC1, peers are no longer connected until after the warmup.

### Expected behavior
Peers should begin being connected to establish a depth during the warmup period.

### Actual behavior
Kademlia doesn't try to connect to any peers until after the warmup period.

### Steps to reproduce
I added a bunch of logs to node.to to track this down and you can see the results at:
https://pastebee.com/?4a6d72f1e4a9c4f80cc92091e32cc0797c94c874ffd1f1efa9205d918eac8a18

### Possible solution
It appears that the invocation of pusherService.AddFeed from node.go is the problem.  That routine tries to write to a channel which is not being read until after the warmup period has cleared in the pusher.

https://github.com/ethersphere/bee/blob/9a1162ff3f44297ac69421dfd7e6ba04d8c992dc/pkg/node/node.go#L772
https://github.com/ethersphere/bee/blob/9a1162ff3f44297ac69421dfd7e6ba04d8c992dc/pkg/pusher/pusher.go#L312

Somehow the write to s.smugler in AddFeed needs to be deferred to allow node.go's startup to complete without waiting for the warmup period.

acud:
fixes #2702 

This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2703)

### DIFF
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -310,7 +310,13 @@ func (s *Service) valid(ch swarm.Chunk) error {
 }
 
 func (s *Service) AddFeed(c <-chan *Op) {
-	s.smugler <- c
+	go func() {
+		select {
+		case s.smugler <- c:
+		case <-s.quit:
+			// if we're quitting: don't do anything
+		}
+	}()
 }
 
 func (s *Service) Close() error {


ldeffenb (commented):
I couldn't have coded it that fast, but that's pretty much what I thought would fix it!  I would have neglected to watch for the quit though!

ldeffenb:
Pulled this into my build and tested in my environment and can positively confirm the peers are now being connected during the warmup period.  Thanks for the quick fix @acud !

PR #2704, opened 2021-12-01
Sharky db migration

mrekucci:
### Checklist

- [x] verify that `newRetrievalDataIndex` location serialization hasn't changed
- [x] graceful shutdown on migration error + sharky location release
- [x] check mem usage with current batch size
- [x] check cost of compactions and run in different intervals
- [x] document that if the migration did not execute correctly you cannot downgrade to the old version

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2704)

### DIFF
--- a/pkg/localstore/migration_sharky.go
+++ b/pkg/localstore/migration_sharky.go
@@ -1,0 +1,167 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package localstore
+
+import (
+	"context"
+	"encoding/binary"
+	"fmt"
+	"time"
+
+	"github.com/ethersphere/bee/pkg/postage"
+	"github.com/ethersphere/bee/pkg/sharky"
+	"github.com/ethersphere/bee/pkg/shed"
+	"github.com/syndtr/goleveldb/leveldb"
+)
+
+// DBSchemaSharky is the bee schema identifier for sharky.
+const DBSchemaSharky = "sharky"
+
+// migrateDeadPush cleans up dangling push index entries that make the pusher stop pushing entries
+func migrateSharky(db *DB) error {
+	db.logger.Debug("starting sharky migration. have patience, this might take a while...")
+	var (
+		start      = time.Now()
+		batch      = new(leveldb.Batch)
+		count      = 0
+		batchSize  = 10000
+		headerSize = 16 + postage.StampSize
+	)
+
+	retrievalDataIndex, err := db.shed.NewIndex("Address->StoreTimestamp|BinID|BatchID|BatchIndex|Sig|Data", shed.IndexFuncs{
+		EncodeKey: func(fields shed.Item) (key []byte, err error) {
+			return fields.Address, nil
+		},
+		DecodeKey: func(key []byte) (e shed.Item, err error) {
+			e.Address = key
+			return e, nil
+		},
+		EncodeValue: func(fields shed.Item) (value []byte, err error) {
+			b := make([]byte, headerSize)
+			binary.BigEndian.PutUint64(b[:8], fields.BinID)
+			binary.BigEndian.PutUint64(b[8:16], uint64(fields.StoreTimestamp))
+			stamp, err := postage.NewStamp(fields.BatchID, fields.Index, fields.Timestamp, fields.Sig).MarshalBinary()
+			if err != nil {
+				return nil, err
+			}
+			copy(b[16:], stamp)
+			value = append(b, fields.Data...)
+			return value, nil
+		},
+		DecodeValue: func(keyItem shed.Item, value []byte) (e shed.Item, err error) {
+			e.StoreTimestamp = int64(binary.BigEndian.Uint64(value[8:16]))
+			e.BinID = binary.BigEndian.Uint64(value[:8])
+			stamp := new(postage.Stamp)
+			if err = stamp.UnmarshalBinary(value[16:headerSize]); err != nil {
+				return e, err
+			}
+			e.BatchID = stamp.BatchID()
+			e.Index = stamp.Index()
+			e.Timestamp = stamp.Timestamp()
+			e.Sig = stamp.Sig()
+			e.Data = value[headerSize:]
+			return e, nil
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	newRetrievalDataIndex, err := db.shed.NewIndex("Address->StoreTimestamp|BinID|BatchID|BatchIndex|Sig|Location", shed.IndexFuncs{
+		EncodeKey: func(fields shed.Item) (key []byte, err error) {
+			return fields.Address, nil
+		},
+		DecodeKey: func(key []byte) (e shed.Item, err error) {
+			e.Address = key
+			return e, nil
+		},
+		EncodeValue: func(fields shed.Item) (value []byte, err error) {
+			b := make([]byte, headerSize)
+			binary.BigEndian.PutUint64(b[:8], fields.BinID)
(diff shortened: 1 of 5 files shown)


pkg/localstore/migration_sharky.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/ethersphere/bee/pkg/sharky"
+	"github.com/ethersphere/bee/pkg/shed"
+	"github.com/syndtr/goleveldb/leveldb"
+)
+
+// DBSchemaSharky is the bee schema identifier for sharky.
+const DBSchemaSharky = "sharky"
+
+// migrateDeadPush cleans up dangling push index entries that make the pusher stop pushing entries
+func migrateSharky(db *DB) error {
```
aloknerurkar, line 23: So we have also changed the gcIndex. Do we need to migrate that as well?

zelig (commented):
> Reviewed 6 of 10 files at r3, 5 of 5 files at r4, 1 of 1 files at r5, 2 of 5 files at r6, 2 of 3 files at r7, 1 of 1 files at r8, 20 of 23 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2704)* status: 24 of 27 files reviewed, 3 unresolved discussions (waiting on @acud, @aloknerurkar, @mrekucci, and @zelig)

___
*[pkg/localstore/localstore.go, line 254 at r9](https://reviewable.io/reviews/ethersphere/bee/2704#-MspvTd7CA15lI67_Mka:-MspvTd7CA15lI67_Mkb:b-sldvqg) ([raw file](https://github.com/ethersphere/bee/blob/3853906bca6f156a2473efa15454431c222a8b77/pkg/localstore/localstore.go#L254)):*
> ```Go
> 		baseKey:         baseKey,
> 		tags:            o.Tags,
> 		ctx:             ctx,
> ```

suspicious, antipattern
___
*[pkg/sharky/sharky_test.go, line 25 at r9](https://reviewable.io/reviews/ethersphere/bee/2704#-Msq7zWaCPnAy5FGPkm-:-Msq7zWaCPnAy5FGPkm0:brr49sx) ([raw file](https://github.com/ethersphere/bee/blob/3853906bca6f156a2473efa15454431c222a8b77/pkg/sharky/sharky_test.go#L25)):*
> ```Go
> )
> 
> type dirFS struct {
> ```

this is also in a non-test package, should get rid of one

zelig:
this PR contains (some of) the changes from my PR, maybe it should be rebased on a branch that has those changes, no?

zelig:
why do we need the dirFS abstraction for sharky?

aloknerurkar:
> why do we need the dirFS abstraction for sharky?

@zelig 

One of the work items for sharky integration was to handle the `dev` mode. The `dev` mode currently uses an in-mem localstore. Even in all the unit tests we use the inmem leveldb. After sharky, this was broken as sharky needed access to files on disk. The tests would create local directories when we ran them. This is not desirable.

In order to handle this, we decided to give sharky access to a file abstraction that includes all the operations we do on the file. This way we could use an in-mem file implementation with sharky and have a corresponding in-mem implementation.

Some benefits:
- We would re-use the sharky logic entirely in the in-mem implementation. This should re-use the code in all unit tests as well.
- This could possibly be used in the future to write different implementations. Again preserving the basic fixed-length shard logic. This could also be useful in testing if there ever was a need.

mrekucci:
Thanks @zelig for your input. The relative code you've commented on are changes from the sharky integration PR: #2682 (done by @aloknerurkar @notanatol) on which this PR is based. I believe that those should be addressed there.

aloknerurkar (approved):

acud (approved):
> Reviewed 2 of 5 files at r6, 1 of 23 files at r9, 21 of 24 files at r10, 3 of 3 files at r11.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2704)* status: all files reviewed (commit messages unreviewed), 4 unresolved discussions (waiting on @aloknerurkar, @mrekucci, and @zelig)

___
*[pkg/localstore/migration_sharky.go, line 23 at r7](https://reviewable.io/reviews/ethersphere/bee/2704#-MppVmWEzcTS_m37xS3P-r7-23:-Mt9iJtk8q0h9tZixiwU:bhl0x2p) ([raw file](https://github.com/ethersphere/bee/blob/c05f473cc44c7357b9df910a84e91c3520bb1d7c/pkg/localstore/migration_sharky.go#L23)):*

So we have also changed the gcIndex. Do we need to migrate that as well?
</blockquote></details>

where has it changed? i remember a change being reverted...
___
*[pkg/localstore/migration_sharky.go, line 121 at r11](https://reviewable.io/reviews/ethersphere/bee/2704#-Mt9j3Ov89diZm4ItiUT:-Mt9j3Ov89diZm4ItiUU:bb2lxqv) ([raw file](https://github.com/ethersphere/bee/blob/439ed64d90f0ad39c55268ed5af4442a213ee90d/pkg/localstore/migration_sharky.go#L121)):*
> ```Go
> 		return err
> 	}
> ```

@mrekucci just a suggestion: perhaps initializing a copy of `reserveSize` and `gcSize` like initialized on the `DB` type in localstore might be useful here, since you could check `reserveSize+gcSize` and check that it doesn't exceed the maximum allowed chunks for sharky 
___
*[pkg/sharky/sharky_test.go, line 25 at r9](https://reviewable.io/reviews/ethersphere/bee/2704#-Msq7zWaCPnAy5FGPkm-:-Mt9iiN92a4PBKHAiA2K:b-4qx44m) ([raw file](https://github.com/ethersphere/bee/blob/3853906bca6f156a2473efa15454431c222a8b77/pkg/sharky/sharky_test.go#L25)):*

this is also in a non-test package, should get rid of one
</blockquote></details>

what's the problem again?

janos (approved):
👍

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2704)* status: 25 of 27 files reviewed, 4 unresolved discussions (waiting on @acud, @aloknerurkar, @mrekucci, and @zelig)

___
*[pkg/localstore/migration_sharky.go, line 121 at r11](https://reviewable.io/reviews/ethersphere/bee/2704#-Mt9j3Ov89diZm4ItiUT:-MtDYKm22WBrWCHIzTh9:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/439ed64d90f0ad39c55268ed5af4442a213ee90d/pkg/localstore/migration_sharky.go#L121)):*

@mrekucci just a suggestion: perhaps initializing a copy of `reserveSize` and `gcSize` like initialized on the `DB` type in localstore might be useful here, since you could check `reserveSize+gcSize` and check that it doesn't exceed the maximum allowed chunks for sharky 
</blockquote></details>

Done.

aloknerurkar (approved):

pkg/localstore/migration_sharky.go
```diff
@@ -119,6 +120,30 @@ func migrateSharky(db *DB) error {
...
+		return err
+	}
+
+	reserveSize, err := db.reserveSize.Get()
+	if err != nil {
+		return err
+	}
+
+	if gcSize+reserveSize > sharkyMaxTotalChunks {
+		return errors.New("maximum allowed chunks exceeded")
```
ldeffenb, line 144: So the answer to my question about what will happen to my massively pinned chunk node is that the sharky migration will fail leaving me with a node that no longer works until I build my own dirty version with a dramatically increased limit.  Ug.

acud: Yes you would need to do that. We will try to work around it so that it is not needed. Will keep you in the loop

acud: Another option is that we will just up the "limit" (or make it unbounded)

PR #2705, opened 2021-12-01
chore(pushsync): tweak p90 to mainnet

istae:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2705)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -38,11 +38,10 @@ const (
 )
 
 const (
-	defaultTTL     = 30 * time.Second // request time to live
-	p90TTL         = 5 * time.Second  // P90 request time to live
+	defaultTTL     = 30 * time.Second        // request time to live
+	p90TTL         = 1500 * time.Millisecond // P90 request time to live
 	sanctionWait   = 5 * time.Minute
 	replicationTTL = 5 * time.Second // time to live for neighborhood replication
-
 )
 
 const (


pkg/pushsync/pushsync.go
```diff
@@ -38,11 +38,10 @@ const (
 )
 
 const (
-	defaultTTL     = 30 * time.Second // request time to live
-	p90TTL         = 5 * time.Second  // P90 request time to live
+	defaultTTL     = 30 * time.Second        // request time to live
+	p90TTL         = 1500 * time.Millisecond // P90 request time to live
```
acud, line 42: is this the p90 for requests at origin? or for forwarding too?

istae: The previous 5 seconds was just a placeholder.
The new value is based on p90 of requests to peers from bin 0 and 1. In short, it's for requests at origin since we don't have retries at forwarders.

istae: This one https://github.com/ethersphere/bee/commit/f3f8e6e3a97467ec558e6b4d5126754e7988b59b introduces a new metric for origin nodes ONLY.

acud (approved):

pkg/pushsync/pushsync.go
acud: Having thought about this a bit more, this might be a bit tricky to change. If the network starts to slow down, this might have the inverse effect of generating even more traffic and more load over the network, causing it to degrade even more due to a cascade of service degradation. I would suggest to keep this value intentionally higher, at least for now, and lower it carefully.

istae:
ON HOLD until low network volatility is achieved.

mrekucci (approved):

PR #2706, opened 2021-12-01
feat: preemptive retrieval based on mainnet p90

istae:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2706)

### DIFF
--- a/pkg/retrieval/retrieval.go
+++ b/pkg/retrieval/retrieval.go
@@ -49,7 +49,7 @@ type retrievalResult struct {
 	chunk     swarm.Chunk
 	peer      swarm.Address
 	err       error
-	retrieved bool
+	requested bool
 }
 
 type Service struct {
@@ -98,9 +98,9 @@ func (s *Service) Protocol() p2p.ProtocolSpec {
 
 const (
 	defaultTTL   = 10 * time.Second
-	p90TTL       = time.Second
+	p90TTL       = 2 * time.Second
 	maxRetries   = 32
-	maxRetrieved = 8
+	maxRequests  = 8
 	originSuffix = "_origin"
 )
 
@@ -112,17 +112,21 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin 
 		flightRoute = addr.String() + originSuffix
 	}
 
-	v, _, err := s.singleflight.Do(ctx, flightRoute, func(_ context.Context) (interface{}, error) {
+	// topCtx is passing the tracing span to the first singleflight call
+	topCtx := ctx
+
+	v, _, err := s.singleflight.Do(ctx, flightRoute, func(ctx context.Context) (interface{}, error) {
 
 		sp := newSkipPeers()
 
+		// two counters, one for total number of peers tried, other number of retrieval requests fired
 		var (
-			allowedRetries   = maxRetries
-			allowedRetrieves = 1
+			allowedRetries  = maxRetries
+			allowedRequests = 1
 		)
 
 		if origin {
-			allowedRetrieves = maxRetrieved
+			allowedRequests = maxRequests
 		}
 
 		timer := time.NewTimer(0)
@@ -140,13 +144,13 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin 
 			case <-timer.C:
 
 				allowedRetries--
-				allowedRetrieves--
+				allowedRequests--
 
 				s.metrics.PeerRequestCounter.Inc()
 
 				// create a new context without cancelation but
 				// set the tracing span to the new context from the context of the first caller
-				ctx := tracing.WithContext(context.Background(), tracing.FromContext(ctx))
+				ctx := tracing.WithContext(context.Background(), tracing.FromContext(topCtx))
 
 				// get the tracing span
 				span, _, ctx := s.tracer.StartSpanFromContext(ctx, "retrieve-chunk", s.logger, opentracing.Tag{Key: "address", Value: addr.String()})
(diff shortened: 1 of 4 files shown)


pkg/pushsync/pushsync.go
```diff
@@ -88,11 +88,11 @@ type PushSync struct {
 }
 
 type receiptResult struct {
-	pushTime  time.Time
-	peer      swarm.Address
-	receipt   *pb.Receipt
-	attempted bool
-	err       error
+	pushTime time.Time
```
acud, line 91: maybe good to split these into a separate PR?

pkg/retrieval/retrieval.go
```diff
@@ -248,8 +202,23 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin
...
+		chunk      swarm.Chunk
+		retrieved  bool
+		startTimer = time.Now()
+	)
+
+	defer func() {
+		select {
+		case resultChan <- retrievalResult{peer: peer, err: err, retrieved: retrieved, chunk: chunk}:
+		case <-doneChan:
+		}
```
acud, line 219: maybe also add a similar duplicate chunk metric here, similar to what we do in pushsync

pkg/retrieval/retrieval.go
```diff
@@ -112,134 +112,88 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin
 		flightRoute = addr.String() + originSuffix
 	}
 
-	// topCtx is passing the tracing span to the first singleflight call
-	topCtx := ctx
```
acud, line 116: the context handling here is very delicate and fragile. why is the change needed?

pkg/retrieval/retrieval.go
```diff
@@ -112,134 +112,88 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin
...
+
+				allowedRetries--
+				allowedRetrieves--
+
+				s.metrics.PeerRequestCounter.Inc()
 
 				// create a new context without cancelation but
 				// set the tracing span to the new context from the context of the first caller
-				ctx := tracing.WithContext(context.Background(), tracing.FromContext(topCtx))
+				ctx := tracing.WithContext(context.Background(), tracing.FromContext(ctx))
```
acud, line 149: this should not be changed

pkg/retrieval/retrieval.go
```diff
@@ -112,134 +112,88 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin
...
-					}
-
+					s.retrieveChunk(ctx, resultChan, doneChan, addr, sp, origin)
 				}()
-			} else {
-				select {
-				case resultC <- retrievalResult{}:
-				case <-ctx.Done():
+
+				if allowedRetries <= 0 || allowedRetrieves <= 0 {
```
acud, line 162: this looks a bit odd... if we have no more attempts left then we try to do another attempt?

pkg/retrieval/retrieval.go
```diff
@@ -112,134 +112,88 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin
...
 				}
+
 				if res.retrieved {
-					if res.err != nil {
-						if !res.peer.IsZero() {
-							s.logger.Debugf("retrieval: failed to get chunk %s from peer %s: %v", addr, res.peer, res.err)
-						}
-						peersResults++
-					} else {
+					if res.err == nil {
```
acud, line 179: maybe just checking whether the error is nil would be enough? can't `retrieved` be assumed from that?

pkg/retrieval/retrieval.go
```diff
@@ -112,134 +112,88 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin
...
-				return nil, storage.ErrNotFound
-			}
 
-			// if we have not counted enough successful attempts but out of selection amount, reset
-			if peerAttempt >= maxSelects {
-
-				// forwarder gives up after first round of request attempts,
-				// rather than holding up the forwarding chain by waiting to be able to reserve downstream peer
-				if !origin {
+				if allowedRetries <= 0 || allowedRetrieves <= 0 {
```
acud, line 189: here again the same conditional like before, but this time causes an error to be propagated

pkg/retrieval/retrieval.go
```diff
@@ -248,8 +202,24 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin
...
 
-func (s *Service) retrieveChunk(ctx context.Context, addr swarm.Address, sp *skipPeers, originated bool) (chunk swarm.Chunk, peer swarm.Address, requested bool, err error) {
-	startTimer := time.Now()
+func (s *Service) retrieveChunk(ctx context.Context, resultChan chan retrievalResult, doneChan chan struct{}, addr swarm.Address, sp *skipPeers, originated bool) {
+
+	var (
+		peer       swarm.Address
+		err        error
+		chunk      swarm.Chunk
+		retrieved  bool
```
acud, line 211: can't this be assumed by checking `err == nil`?

pkg/retrieval/retrieval.go
```diff
@@ -248,8 +202,24 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin
...
+		startTimer = time.Now()
+	)
+
+	defer func() {
```
acud, line 215: i wonder if this is needed here, maybe it is more readable if we inline this in the same place you're spawning the goroutine? then just keep this method the same with the normal return values?

acud (changes requested):

acud (approved):
nice, thanks!

PR #2707, opened 2021-12-02
fix(pushsync): close done channel with a defer

istae:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2707)

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -88,11 +88,11 @@ type PushSync struct {
 }
 
 type receiptResult struct {
-	pushTime  time.Time
-	peer      swarm.Address
-	receipt   *pb.Receipt
-	attempted bool
-	err       error
+	pushTime time.Time
+	peer     swarm.Address
+	receipt  *pb.Receipt
+	pushed   bool
+	err      error
 }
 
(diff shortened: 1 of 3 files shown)


pkg/pushsync/pushsync.go
```diff
@@ -351,6 +351,7 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
 
 	resultChan := make(chan receiptResult, 1)
 	doneChan := make(chan struct{})
+	defer close(doneChan)
```
acud, line 354: you cannot do this. please see https://go.dev/play/p/YO9nL9sKMuy
the dangling goroutines would panic

acud (changes requested):

pkg/pushsync/pushsync.go
istae: this is the done channel, it's read only.

acud: oh sorry, looked at the wrong line. so we have a deadlock on the testnet without this, right?

acud (approved):

pkg/pushsync/pushsync.go
istae: yeah, if pushToClosest returns with error, there is a goroutine leak.

mrekucci (approved):

PR #2714, opened 2021-12-06
chore(ci): adjust block threshold for ci

ralph-pichler:
block threshold in ci is too too low leading to unnecessary push sync fails due to unsynced batch stores. it might take up to 5 blocks for a batch to become valid (4 tail size, 2 batch factor), so I used 6 as threshold here. 

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2714)

### DIFF
--- a/.github/patches/postageservice.patch
+++ b/.github/patches/postageservice.patch
@@ -1,4 +1,4 @@
 21c21
 < 	blockThreshold = 10
 ---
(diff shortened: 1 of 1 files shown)


acud (approved):

acud:
@istae could you sync with @ralph-pichler about the changes you made to the listener polling to see if this is in line?

PR #2723, opened 2021-12-08
feat: multiple preemptive retries

metacertain:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2723)


mrekucci (approved):

notanatol (approved):

metacertain:
> @metacertain I don't really understand this change and I think that it makes the protocol to be even more unreadable. I also don't really understand why only one preemptive retry is done today and why a Reset is needed on the ticker. A ticket will expire anyway, why do you need to reset it (which effectively only postpones the retry)

yeah the ticker reset is actually unnecessary

PR #2729, opened 2021-12-09
fix: unblocking write to channel in retrieval

metacertain:
This change improves a rare edge case where the channel for results would be filled by asynchronous retrieves, also removing the buffer for the channel

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2729)

### DIFF
--- a/pkg/retrieval/retrieval.go
+++ b/pkg/retrieval/retrieval.go
@@ -129,7 +129,7 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr swarm.Address, origin 
 		var (
 			peerAttempt  int
 			peersResults int
-			resultC      = make(chan retrievalResult, maxSelects)
+			resultC      = make(chan retrievalResult)
 		)
 
 		requestAttempt := 0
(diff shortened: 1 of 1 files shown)


acud (changes requested):
This also opens up another edge case, where because the polling goroutine is not yet listening to the response on the `select` on `L178`, the response is forever lost due to the `default` case being hit. I would advise against this

metacertain:
> This also opens up another edge case, where because the polling goroutine is not yet listening to the response on the `select` on `L178`, the response is forever lost due to the `default` case being hit. I would advise against this

When we receive a response, we always write that back in L161, and that happens in a goroutine, so blocking until the channel is free is not a problem there.

The empty result being wrote to the channel that gets a default case now only attempts to write to this channel to make sure the select in L178 doesn't block waiting as no new attempt was started in this iteration. 
If meanwhile there was a result wrote by one of the previous goroutines, this is not necessary, as the select will take that result immediately.

acud (approved):

notanatol (approved):

mrekucci (approved):

PR #2732, opened 2021-12-13
fix: cashout to overlay address if no chequebook

ralph-pichler:
the last PR did not handle cashouts correctly if running without a chequebook. this prevents a go panic and uses the nodes overlay eth address as a destination instead.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2732)


ldeffenb:
This is the code I expected to see in the previous PR!  Great job!

acud (approved):

PR #2746, opened 2022-01-03
fix: remove subdirectories, not root path for migration command

istae:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2746)

### DIFF
--- a/cmd/bee/cmd/db.go
+++ b/cmd/bee/cmd/db.go
@@ -191,7 +191,7 @@ func dbNukeCmd(cmd *cobra.Command) {
 			logger.Warningf("Proceeding with database nuke...")
 
 			localstorePath := filepath.Join(dataDir, "localstore")
-			err = os.RemoveAll(localstorePath)
+			err = removeContent(localstorePath)
 			if err != nil {
 				return fmt.Errorf("localstore delete: %w", err)
 			}
@@ -204,7 +204,7 @@ func dbNukeCmd(cmd *cobra.Command) {
 			}
 
 			if forgetOverlay {
-				err = os.RemoveAll(statestorePath)
+				err = removeContent(statestorePath)
 				if err != nil {
 					return fmt.Errorf("statestore delete: %w", err)
 				}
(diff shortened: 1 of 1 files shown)


cmd/bee/cmd/db.go
```diff
@@ -230,3 +230,24 @@ func dbNukeCmd(cmd *cobra.Command) {
 	c.Flags().Duration(optionNameSleepAfter, time.Duration(0), "time to sleep after the operation finished")
 	cmd.AddCommand(c)
 }
+
+func removeContent(path string) error {
+	dir, err := os.Open(path)
```
acud, line 235: this is a bit misleading. i believe that if you use this you would then get everything that sits in the directory in which `path` is contained. did you try this manually? i would consider using [filepath.Walk](https://pkg.go.dev/path/filepath#Walk)

istae: yeah, tried manually, root folder is preserved

acud (approved):
thanks!

notanatol (approved):

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2746)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @acud, @istae, and @mrekucci)

___
*[cmd/bee/cmd/db.go, line 248 at r2](https://reviewable.io/reviews/ethersphere/bee/2746#-Ms_URqWETfpaAbeGRin:-Ms_URqWETfpaAbeGRio:b-152phs) ([raw file](https://github.com/ethersphere/bee/blob/b76e53ff61f21b6a6b3d838b3cf068d1ccb260c0/cmd/bee/cmd/db.go#L248)):*
> ```Go
> 	for _, sub := range subpaths {
> 		err = os.RemoveAll(filepath.Join(path, sub))
> 		if err != nil {
> ```

It is desirable to return on the first error? Wouldn't be better to collect all the errors (and continue the execution) and return flattened multierror?

acud (approved):

PR #2751, opened 2022-01-06
fix: remove old overlay delete

istae:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2751)

Now that all nodes in the network are using > 1.0.0, we can remove this bit of old code.


notanatol (approved):

mrekucci (approved):

acud:
there's a lot of this stuff and not just here... locastore, statestore and other places. btw the assumption that all nodes are running over 1.0 is one thing, but you're effectively blocking old nodes of a migration path to upgrade (conceptually. whether it is really needed is a different story).

PR #2756, opened 2022-01-12
chore(localstore): decrease debug logging for pull subs

acud:
Decreases the amount of logging we do on pull subs in localstore

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2756)

### DIFF
--- a/pkg/localstore/subscription_pull.go
+++ b/pkg/localstore/subscription_pull.go
@@ -29,6 +29,8 @@ import (
 	"github.com/syndtr/goleveldb/leveldb"
 )
 
+const logMore = false
+
 // SubscribePull returns a channel that provides chunk addresses and stored times from pull syncing index.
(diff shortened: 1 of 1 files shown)


mrekucci (approved):

acud:
Logging is such an interesting problem... There are really such diverse ways of executing on logging. For instance, in the leveldb package, by default, the logger method is set to some noop function that just ignores the input. In the go-ethereum library, for example, you could even set the logger to filter out data from a specific package (but it would require a CLI argument, so a binary restart would always be needed). There could also be possibilities of toggling the log level on runtime (but would require an adequate central abstraction that would deal with that). Or otherwise use some ELK/Grafana+Loki stack (which have their own drawbacks and problems)...

notanatol (approved):

PR #2762, opened 2022-01-16
feat: bootstrap new nodes to sync batch events from the network snapshot

istae:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2762)

### DIFF
--- a/pkg/postage/listener/listener.go
+++ b/pkg/postage/listener/listener.go
@@ -116,7 +116,8 @@ func (l *listener) filterQuery(from, to *big.Int) ethereum.FilterQuery {
 	}
 }
 
-func ProcessEvent(e types.Log, updater postage.EventUpdater) error {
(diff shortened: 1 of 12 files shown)


pkg/postage/listener/listener.go
```diff
@@ -177,13 +177,24 @@ func (l *listener) processEvent(e types.Log, updater postage.EventUpdater) error
...
 
-func (l *listener) Listen(from uint64, updater postage.EventUpdater) <-chan struct{} {
+func (l *listener) Listen(from uint64, updater postage.EventUpdater, initState *postage.BatchSnapshot) <-chan struct{} {
 	ctx, cancel := context.WithCancel(context.Background())
 	go func() {
 		<-l.quit
 		cancel()
 	}()
 
+	if initState != nil {
```
acud, line 187: this section is not correct btw, not is it unit tested. when processing events you need to update the block height after every event. please see how the individual events are being processed normally by the listener

PR #2764, opened 2022-01-17
fix: remove sharky size limit

notanatol:
### Checklist

### Description
This change removes the check for the upper size limit of sharky.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2764)

### DIFF
--- a/pkg/sharky/recovery.go
+++ b/pkg/sharky/recovery.go
@@ -19,12 +19,14 @@ type Recovery struct {
 	shards []*slots
 }
 
+var ErrShardNotFound = errors.New("shard not found")
+
 func NewRecovery(dir string, shardCnt int, datasize int) (*Recovery, error) {
 	shards := make([]*slots, shardCnt)
 	for i := 0; i < shardCnt; i++ {
-		file, err := os.OpenFile(path.Join(dir, fmt.Sprintf("shard_%03d", i)), os.O_RDWR|os.O_CREATE, 0666)
+		file, err := os.OpenFile(path.Join(dir, fmt.Sprintf("shard_%03d", i)), os.O_RDONLY, 0666)
 		if errors.Is(err, fs.ErrNotExist) {
-			break
+			return nil, fmt.Errorf("index %d: %w", i, ErrShardNotFound)
 		}
 		if err != nil {
 			return nil, err
--- a/pkg/sharky/recovery.go
+++ b/pkg/sharky/recovery.go
@@ -5,7 +5,9 @@
 package sharky
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"path"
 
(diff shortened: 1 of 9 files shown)


istae (approved):

aloknerurkar (approved):

pkg/sharky/recovery.go
```diff
@@ -17,10 +19,13 @@ type Recovery struct {
 	shards []*slots
 }
 
-func NewRecovery(dir string, shardCnt int, shardSize uint32, datasize int) (*Recovery, error) {
+func NewRecovery(dir string, shardCnt int, datasize int) (*Recovery, error) {
 	shards := make([]*slots, shardCnt)
 	for i := 0; i < shardCnt; i++ {
 		file, err := os.OpenFile(path.Join(dir, fmt.Sprintf("shard_%03d", i)), os.O_RDWR|os.O_CREATE, 0666)
+		if errors.Is(err, fs.ErrNotExist) {
```
aloknerurkar, line 26: Why is this handled differently? If we dont have `shardCnt` no of shards... we should fail the recovery I feel

pkg/sharky/shard_only_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestShard(t *testing.T) {
```
aloknerurkar, line 15: I dont understand what this test is doing. Can we have more comments? Why do we need to test these separately?

aloknerurkar (changes requested):

pkg/sharky/shard_only_test.go
notanatol: this is needed because the sharky package is under tested.
there is a high level acceptance test `sharky_test.go` but it does not cover the functionality related to releasing slots and making them available for write, on a shard level.
this test covers that part, but you are right, I will add comments to clarify its purpose.

pkg/sharky/recovery.go
notanatol: thanks, this is a good catch, I fixed it to fail if shard is not found, added test.

acud (approved):
nice, thanks

aloknerurkar (approved):

zelig (commented):
i think the process loops can be simplified now.
I think the shard test is unnecessary
overall i am not 100% that this change is bringing much of value

> Reviewed 5 of 8 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2764#-:-MtzELURAu7wwmcQQf4S:b-pbxg83)* status: all files reviewed, 2 unresolved discussions (waiting on @aloknerurkar and @istae)

ldeffenb:
> i think the process loops can be simplified now. I think the shard test is unnecessary overall i am not 100% that this change is bringing much of value

I believe moving the chunks out of the localstore will buy performance gains for large storage nodes.  You haven't experienced performance pain until you've watched a 50-100GB localstore try to serve pullsync requests.  The disk I/O caused by the interleaving of small index data with 4K chunk data is positively painful.

PR #2779, opened 2022-01-20
fix: gc over collecting

notanatol:
### Checklist

### Description
Increase the limit of the number of chunks in a single transaction on garbage collection from 2k to 10k.
Changes the approach in regards to chunk collection - previously the strategy was to start with a minimal amount of chunks and work our way up, in the new approach we try to collect as many as possible, only stopping when the target is achieved.

#### Motivation and context
Currently garbage collection tends to lag behind the influx of chunks, leading to poor eviction performance.
This change takes a different approach - specifically - trying to over-collect chunks pending for eviction, minimizing the lag.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2779)

### DIFF
--- a/pkg/localstore/gc.go
+++ b/pkg/localstore/gc.go
@@ -37,7 +37,7 @@ var (
 	gcTargetRatio = 0.9
 	// gcBatchSize limits the number of chunks in a single
 	// transaction on garbage collection.
-	gcBatchSize uint64 = 2000
+	gcBatchSize uint64 = 10_000
 
 	// reserveCollectionRatio is the ratio of the cache to evict from
 	// the reserve every time it hits the limit. If the cache size is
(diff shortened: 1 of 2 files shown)


acud (approved):
Looks good. Thanks!

pkg/localstore/gc.go
```diff
@@ -165,13 +162,21 @@ func (db *DB) collectGarbage() (collectedCount uint64, done bool, err error) {
...
+	var totalChunksEvicted uint64
+
 	// get rid of dirty entries
 	for _, item := range candidates {
 		if swarm.NewAddress(item.Address).MemberOf(db.dirtyAddresses) {
-			collectedCount--
 			continue
 		}
 
+		if gcSize-totalChunksEvicted <= target {
```
aloknerurkar, line 173: I am not able to understand this check. I see the comment on the function `gcTarget` needs to be fixed?

aloknerurkar (approved):

pkg/localstore/gc.go
notanatol: the comment has been fixed, thanks.
we also added another comment over this check that explains its purpose.

PR #2788, opened 2022-01-31
feat: crypt hasher

notanatol:
### Checklist

### Description
- We now return `403 Forbidden` in only one case - when the provided token is valid but does not have sufficient privileges to access the requested resource.
All the other cases will result in a `401 Unauthorized` return code.

- Added a new command  `bcrypt` that can generate a bcrypt hash for a given plain text password. The returned value can be used for `admin-password` configuration parameter. The tool can also validate that the provided hash matches a plain text password (using the `--check` option). Note: when checking the hash please make sure to quote it, the shell does not take kindly to the dollar signs in the hash text.

#### Motivation and context
The bcrypt command is a handy offline alternative to https://bcrypt-generator.com/

#### Usage
```sh
> bee bcrypt super$ecret
$2a$10$eZP5YuhJq2k8DFmj9UJGWOIjDtXu6NcAQMrz7Zj1bgIVBcHA3bU5u
> bee bcrypt --check super$ecret '$2a$10$eZP5YuhJq2k8DFmj9UJGWOIjDtXu6NcAQMrz7Zj1bgIVBcHA3bU5u'
OK: password hash matches provided plain text
```

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2788)


cmd/hasher/main.go
```diff
@@ -0,0 +1,27 @@
...
+		fmt.Println("usage:", "hasher", "your-plain-text-password")
+		return
+	}
+
+	plainText := os.Args[1]
+
+	hashed, err := bcrypt.GenerateFromPassword([]byte(plainText), bcrypt.DefaultCost)
+
+	if err != nil {
+		panic(err)
```
mrekucci, line 23: I wouldn't panic here, just print a user-friendly error to the stderr and return a non-zero OS exit code.

mrekucci (approved):

AuHau:
LGTM, but actually I would like to bring up why we actually even need to hash the password? What is the goal? 

I guess protection of the password, but I am not sure if it is not overkill? This just seems to me like a bad UX...

notanatol:
> LGTM, but actually I would like to bring up why we actually even need to hash the password? What is the goal?
> 
> I guess protection of the password, but I am not sure if it is not overkill? This just seems to me like a bad UX...

I don't remember the exact reason (except not having plain text passwords) but there was an explanation given by Janos about this in one of the comments here: https://github.com/ethersphere/bee/pull/2400

janos:
> LGTM, but actually I would like to bring up why we actually even need to hash the password? What is the goal?
> 
> I guess protection of the password, but I am not sure if it is not overkill? This just seems to me like a bad UX...

Yes, as @notanatol said more details are in the previous pr comments, but basic the reason is the same as passwords are not stored in plain text in databases in common web applications. Or, why web servers like apache2 and nginx are doing the same for basic auth configurations. Apache2 and nginx are very popular web servers.

acud (commented):
@notanatol since we are using viper-cobra wouldn't it make more sense to use the existing infrastructure to ship the new command? also, since you're de-facto creating a new binary here, it will not be built nor shipped by default, which means that we're gonna have to have separate artifacts on releases for this. This involves a lot of changes to the release process and I'm not sure this is entirely necessary or desirable (we were already burnt by maintaining these sort of extra util in the past).
So I wonder whether it would not make sense to pack this as a subcommand in `bee`. Ideally we would not have this sort of stuff in bee, not sure why this cannot be a command as part of `swarm-cli`. @agazso input please?

notanatol:
> > So I wonder whether it would not make sense to pack this as a subcommand in `bee`. Ideally we would not have this sort of stuff in bee, not sure why this cannot be a command as part of `swarm-cli`. @agazso input please?
> 
> I agree that creating a new binary can make things cumbersome. Alternatively it makes sense to add this functionality to `swarm-cli`.

OK so in this case I'm going to split this PR in two - one with the fixes (HTTP codes) requested by @AuHau and another one with the embedded hasher.

Then we'll have all the time we need to make a decision if we want this hasher in the binary or a separate js tool.

mrekucci:
Maybe it will be worth considering just using something existing as `htpasswd`: https://unix.stackexchange.com/questions/307994/compute-bcrypt-hash-from-command-line

AuHau:
Oh, I haven't noticed that it is a standalone command. I originally envisioned it as `bee`'s subcommand, so I agree that makes more sense. 

Although I would strongly suggest incorporating this functionality directly into `bee`. The @mrekucci mentioned `htpasswd` command requires `apache-utils` package which is a dependency completely unrelated to Bee's stack and requires the user to actually understand what command and what parameters it should use to produce hash compatible with Bee's requirements.

Also, I am not in favor of outsourcing this to `swarm-cli` as it is directly related to the Bee's configuration and not "Swarm functionality" (which is the domain of `swarm-cli`). For example, I am doing this hashing as part of Bee Factory setup for the restricted API support and this would require also to install `swarm-cli` to make it work...

notanatol:
> Oh, I haven't noticed that it is a standalone command. I originally envisioned it as `bee`'s subcommand, so I agree that makes more sense.
> 
> Although I would strongly suggest incorporating this functionality directly into `bee`. The @mrekucci mentioned `htpasswd` command requires `apache-utils` package which is a dependency completely unrelated to Bee's stack and requires the user to actually understand what command and what parameters it should use to produce hash compatible with Bee's requirements.
> 
> Also, I am not in favor of outsourcing this to `swarm-cli` as it is directly related to the Bee's configuration and not "Swarm functionality" (which is the domain of `swarm-cli`). For example, I am doing this hashing as part of Bee Factory setup for the restricted API support and this would require also to install `swarm-cli` to make it work...

@AuHau I moved it to be a `bee` sub-command, feel free to give it a try.

cmd/bee/cmd/hasher.go
```diff
@@ -0,0 +1,73 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
```
acud, line 1: 22

cmd/bee/cmd/hasher.go
```diff
@@ -0,0 +1,73 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"golang.org/x/crypto/bcrypt"
+)
+
+func (c *command) initHasherCmd() (err error) {
+	cmd := &cobra.Command{
+		Use:   "hasher",
```
acud, line 17: maybe just `bcrypt`? hasher has some historical connotation related to file-hashing

cmd/hasher/main.go
```diff
@@ -0,0 +1,64 @@
+package main
```
acud, line 1: code copy...

acud (commented):
the usual naming nitpicking :)

cmd/bee/cmd/hasher.go
notanatol: fixed

cmd/bee/cmd/hasher.go
notanatol: fixed

cmd/hasher/main.go
notanatol: fixed

acud:
> Also, I am not in favor of outsourcing this to `swarm-cli` as it is directly related to the Bee's configuration and not "Swarm functionality" (which is the domain of `swarm-cli`). For example, I am doing this hashing as part of Bee Factory setup for the restricted API support and this would require also to install `swarm-cli` to make it work...

I guess here we are not 100% agreeing (which is fine). IMO it is not in `bee`'s responsibility to hash passwords just as much as it does not have an explicit command that generates you with a V3 ethereum key (you use external libraries for that). It also also raises the question on what is exactly the responsibility of `swarm-cli`. For me `swarm-cli` can easily be a sort of `apache-utils` suite. Just some thoughts.

AuHau:
Ups sorry 😅  Mis-click 🙈
Just wanted to write that I have tried it and LGTM! 😅

AuHau:
And ad @acud:
I guess I agree with you, except that until now "configuration of Bee" is not really domain of `swarm-cli` and that was why I have though of Bee's CLI as better place to put it. If `swarm-cli` will start to be in the configuration business of Bee then I agree there it would make more sense...

cmd/bee/cmd/bcrypt.go
```diff
@@ -0,0 +1,73 @@
...
+			if len(args) != 1 {
+				return cmd.Help()
+			}
+
+			hashed, err := bcrypt.GenerateFromPassword([]byte(args[0]), bcrypt.DefaultCost)
+			if err != nil {
+				return errors.New("failed to generate password hash")
+			}
+
+			fmt.Println(string(hashed))
```
acud, line 61: Be aware that you're print not just the actual string here, but also a line feed. It would be good to check that it doesn't make some cli checks fail as a result (e.g. if the hashed password is stored in a file then simply `bee bcrypt --check playtextpass | diff hash_pass -` to see that they match)

notanatol: I think it works:
```
$> go run $(pwd)/cmd/bee bcrypt super$ecret > plaintext
$> go run $(pwd)/cmd/bee bcrypt --check super$ecret `cat plaintext` 
OK: password hash matches provided plain text
```
but I will remove the new line just in case

notanatol:
> go run $(pwd)/cmd/bee bcrypt --check super$ecret `go run $(pwd)/cmd/bee bcrypt super$ecret
> OK: password hash matches provided plain text

cmd/bee/cmd/bcrypt.go
acud: they will always pass in this case, because the command generated both outputs (hence both have exactly the same characters, line feed included) :)

PR #2794, opened 2022-02-07
refactor: allow duplicate handshakes

acud:
As discussed while looking into various errors we've been running into on the libp2p abstraction level, it has been concluded that removing the duplicate handshake constraint is not beneficial and should be removed. It does not guard against anything, and duplicate handshakes are allowed even _with_ this code section (since we can dial to a node and a node can dial back and there's no keeping track of the bidirectional connections). Furthermore, libp2p allows for multiple underlying connections and will use the best one when sending a message according to its internal heuristics of connection selection.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2794)


mrekucci (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2794)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud)

aloknerurkar (approved):

PR #2796, opened 2022-02-07
fix(p2p): handle header stream reset explicitly

acud:
This PR adds explicit handling to a stream reset on the header exchange with the peer. We've found that this is caused by an inconsistent state of the peerstore, in which apparently disconnections do no propagate correctly to our abstraction, so we seem to be connected to a peer, while the other peer does not seem to accept the same assumption. This results in an immediate stream reset from the other side, resolving the long mystery under which we were operating.

Note that this is a stopgap solution and creates an unnecessary dependency in the headers exchange in order to determine connectivity status. Not exactly textbook solution, but allows for a clear disambiguation of the case since this error should definitely _never ever_ happen in this code block, and will give us a bit of leeway to rectify the situation by inspecting this in depth.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2796)

### DIFF
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -35,6 +35,7 @@ import (
 	crypto "github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/event"
 	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/mux"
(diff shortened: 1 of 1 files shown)


janos (approved):
This is just the first stream communication that is affected and because of that, probably the predominant. But, the same error can happen on any read or write on the stream. A general solution for handling this error will be better.

notanatol (approved):

PR #2806, opened 2022-02-09
refactor: standardize stream closure

metacertain:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2806)

### DIFF
--- a/pkg/chainsync/chainsync.go
+++ b/pkg/chainsync/chainsync.go
@@ -87,7 +87,13 @@ func (s *ChainSync) Prove(ctx context.Context, peer swarm.Address, blockheight u
 	if err != nil {
 		return nil, fmt.Errorf("new stream: %w", err)
 	}
-	defer stream.Close()
+	defer func() {
+		if err != nil {
+			_ = stream.Reset()
+		} else {
+			go stream.FullClose()
+		}
+	}()
 
 	ctx, cancel := context.WithTimeout(ctx, messageTimeout)
 	defer cancel()
@@ -114,7 +120,7 @@ func (s *ChainSync) syncHandler(ctx context.Context, peer p2p.Peer, stream p2p.S
 		if err != nil {
 			_ = stream.Reset()
 		} else {
-			_ = stream.Close()
+			_ = stream.FullClose()
 		}
 	}()
 	if !s.inLimiter.Allow(peer.Address.ByteString(), 1) {
--- a/pkg/chainsync/chainsync.go
+++ b/pkg/chainsync/chainsync.go
@@ -91,7 +91,7 @@ func (s *ChainSync) Prove(ctx context.Context, peer swarm.Address, blockheight u
 		if err != nil {
 			_ = stream.Reset()
 		} else {
-			go stream.FullClose()
+			go stream.Close()
 		}
 	}()
 
@@ -120,7 +120,7 @@ func (s *ChainSync) syncHandler(ctx context.Context, peer p2p.Peer, stream p2p.S
 		if err != nil {
 			_ = stream.Reset()
 		} else {
-			_ = stream.FullClose()
+			_ = stream.Close()
 		}
 	}()
 	if !s.inLimiter.Allow(peer.Address.ByteString(), 1) {
--- a/pkg/chainsync/chainsync.go
+++ b/pkg/chainsync/chainsync.go
@@ -79,7 +79,7 @@ func New(s p2p.Streamer, backend transaction.Backend) (*ChainSync, error) {
 
 // Prove asks a peer to prove they know a certain block height on the
 // current used eth backend.
-func (s *ChainSync) Prove(ctx context.Context, peer swarm.Address, blockheight uint64) ([]byte, error) {
+func (s *ChainSync) Prove(ctx context.Context, peer swarm.Address, blockheight uint64) (hash []byte, err error) {
 	if !s.outLimiter.Allow(peer.ByteString(), 1) {
 		return nil, errRateLimitExceeded
 	}
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -180,7 +180,7 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
 		} else {
 			// added this because Recorder (unit test) emits an unnecessary EOF when Close is called
 			time.Sleep(time.Millisecond * 50)
-			go stream.Close()
+			_ = stream.Close()
 		}
 	}()
 	w, _ := protobuf.NewWriterAndReader(stream)
@@ -218,8 +218,6 @@ func (s *Service) peersHandler(ctx context.Context, peer p2p.Peer, stream p2p.St
 	defer func() {
 		if err != nil {
 			_ = stream.Reset()
-		} else {
-			go stream.Close()
 		}
 	}()
 	s.metrics.PeersHandler.Inc()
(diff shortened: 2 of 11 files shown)


notanatol (approved):

mrekucci (approved):

notanatol (approved):

pkg/chainsync/chainsync.go
```diff
@@ -114,7 +120,7 @@ func (s *ChainSync) syncHandler(ctx context.Context, peer p2p.Peer, stream p2p.S
 		if err != nil {
 			_ = stream.Reset()
 		} else {
-			_ = stream.Close()
+			_ = stream.FullClose()
```
acud, line 123: it is not needed to full close here (and i don't think we should use this too much anyway...)

metacertain: this PR aims to standardize our usage of streams. should we change FullClose to Close in all protocols within the scope of this PR in retrieval pushsync etc?

acud: so the standardization here (not just in this case) is not necessarily needed and if indeed you want to standardize it might be useful to understand how streams behave. if it is to prevent the case of stream leakage then I argue we should resolve this on the abstraction level

metacertain: my impression is there is practically no difference between the protocols (perhaps with the exception of handshake) on how we use streams. we open them, write and read, and close them at the end or reset on error. this is done mostly consistently in the core protocols. the reason I think it would be nice to have it the same way everywhere is it would decrease uncertainty, and make behaviour more predictable regardless of which context we are in

pkg/hive/hive.go
```diff
@@ -214,29 +214,32 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
...
 
 	if !s.inLimiter.Allow(peer.Address.ByteString(), len(peersReq.Peers)) {
-		_ = stream.Reset()
 		return ErrRateLimitExceeded
 	}
 
 	// close the stream before processing in order to unblock the sending side
 	// fullclose is called async because there is no need to wait for confirmation,
 	// but we still want to handle not closed stream from the other side to avoid zombie stream
-	go stream.FullClose()
```
acud, line 238: this should not be moved

pkg/chainsync/chainsync.go
```diff
@@ -79,15 +79,21 @@ func New(s p2p.Streamer, backend transaction.Backend) (*ChainSync, error) {
...
 	stream, err := s.streamer.NewStream(ctx, peer, nil, protocolName, protocolVersion, syncStreamName)
 	if err != nil {
 		return nil, fmt.Errorf("new stream: %w", err)
 	}
-	defer stream.Close()
+	defer func() {
+		if err != nil {
+			_ = stream.Reset()
+		} else {
+			go stream.FullClose()
```
acud, line 94: `Close` suffices

pkg/pushsync/pushsync.go
```diff
@@ -501,7 +501,13 @@ func (ps *PushSync) pushPeer(ctx context.Context, resultChan chan<- receiptResul
 		err = fmt.Errorf("new stream for peer %s: %w", peer, err)
 		return
 	}
-	defer streamer.Close()
+	defer func() {
+		if err != nil {
+			_ = streamer.Reset()
+		} else {
+			go streamer.FullClose()
```
acud, line 508: fullclose is not needed

pkg/pushsync/pushsync.go
```diff
@@ -620,7 +626,7 @@ func (ps *PushSync) pushToNeighbour(ctx context.Context, peer swarm.Address, ch
 		if err != nil {
 			_ = streamer.Reset()
 		} else {
-			_ = streamer.FullClose()
+			go streamer.FullClose()
```
acud, line 629: fullclose not needed neither is a new goroutine

acud:
While this attempts to standardize the use of stream closure, it ignores a lot of the considerations of how they have been used so far. As said, we should probably start with closing streams on the abstraction layer so that the handlers could be relieved from the responsibility to close them.

pkg/chainsync/chainsync.go
```diff
@@ -79,15 +79,21 @@ func New(s p2p.Streamer, backend transaction.Backend) (*ChainSync, error) {
...
 		return nil, errRateLimitExceeded
 	}
 	stream, err := s.streamer.NewStream(ctx, peer, nil, protocolName, protocolVersion, syncStreamName)
 	if err != nil {
 		return nil, fmt.Errorf("new stream: %w", err)
 	}
-	defer stream.Close()
+	defer func() {
+		if err != nil {
+			_ = stream.Reset()
```
acud, line 92: To the best of my knowledge, when a read/write fro/to a stream returns an error there is no need to reset the stream, since it is already in an errored state, therefore the reset is ineffectual. This is why the original `Close` suffices.

pkg/pingpong/pingpong.go
```diff
@@ -69,7 +69,11 @@ func (s *Service) Ping(ctx context.Context, address swarm.Address, msgs ...strin
 		return 0, fmt.Errorf("new stream: %w", err)
 	}
 	defer func() {
-		go stream.FullClose()
+		if err != nil {
+			_ = stream.Reset()
```
acud, line 73: Here again, we only do read and write operations, so if they errored, the stream is in an errored state and the reset is ineffectual

pkg/pingpong/pingpong.go
```diff
@@ -95,9 +99,15 @@ func (s *Service) Ping(ctx context.Context, address swarm.Address, msgs ...strin
 	return time.Since(start), nil
 }
 
-func (s *Service) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) error {
+func (s *Service) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) (err error) {
 	w, r := protobuf.NewWriterAndReader(stream)
-	defer stream.FullClose()
+	defer func() {
+		if err != nil {
+			_ = stream.Reset()
```
acud, line 106: same here, with an additional missed edge case that the error may be EOF, in which case the loop would exit and the stream would erroneously be reset (it shouldn't).

pkg/pushsync/pushsync.go
```diff
@@ -501,7 +501,13 @@ func (ps *PushSync) pushPeer(ctx context.Context, resultChan chan<- receiptResul
 		err = fmt.Errorf("new stream for peer %s: %w", peer, err)
 		return
 	}
-	defer streamer.Close()
+	defer func() {
+		if err != nil {
+			_ = streamer.Reset()
```
acud, line 506: if you reset here, it is expected you would delete all of the resets below no?

acud (changes requested):

pkg/pingpong/pingpong.go
metacertain: on EOF we break and return nil, so no reset happens.

isn't the reset effectual in a sense that it avoids a dangling-until-timeout read on a stream while the other side is not writing because it encountered some error?

acud: Reset is used to indicate to the other peer that there was an error. actually at a second look, the EOF won't be considered as an error since an explicit nil is returned. In any case, this reset is ineffectual since there is no point of resetting an errored stream. I don't know how libp2p propagates errored streams across nodes but I won't jump into any conclusions without a careful inspection of their source code.

acud (changes requested):
still some issues left unaddressed

PR #2814, opened 2022-02-15
feat: bootstrap node from postage snapshot

aloknerurkar:
### Checklist

### Related Issue (Optional)
[Issue](https://github.com/ethersphere/bee-backlog/issues/17)

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2814)

### DIFF
--- a/pkg/storage/inmemstore/store.go
+++ b/pkg/storage/inmemstore/store.go
@@ -1,3 +1,7 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 package inmemstore
 
 import (
@@ -19,7 +23,7 @@ func New() *Store {
 	}
 }
 
-func (s *Store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Address) (ch swarm.Chunk, err error) {
+func (s *Store) Get(_ context.Context, _ storage.ModeGet, addr swarm.Address) (ch swarm.Chunk, err error) {
 	s.mtx.Lock()
 	defer s.mtx.Unlock()
 
@@ -30,7 +34,7 @@ func (s *Store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 	return nil, storage.ErrNotFound
 }
 
-func (s *Store) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
+func (s *Store) Put(_ context.Context, _ storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
 	s.mtx.Lock()
 	defer s.mtx.Unlock()
 
@@ -43,31 +47,31 @@ func (s *Store) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chun
 	return exist, err
 }
 
-func (s *Store) GetMulti(ctx context.Context, mode storage.ModeGet, addrs ...swarm.Address) (ch []swarm.Chunk, err error) {
+func (s *Store) GetMulti(_ context.Context, _ storage.ModeGet, _ ...swarm.Address) (ch []swarm.Chunk, err error) {
 	panic("not implemented")
 }
 
-func (s *Store) Has(ctx context.Context, addr swarm.Address) (yes bool, err error) {
+func (s *Store) Has(_ context.Context, _ swarm.Address) (yes bool, err error) {
 	panic("not implemented")
 }
 
-func (s *Store) HasMulti(ctx context.Context, addrs ...swarm.Address) (yes []bool, err error) {
+func (s *Store) HasMulti(_ context.Context, _ ...swarm.Address) (yes []bool, err error) {
 	panic("not implemented")
 }
 
-func (s *Store) Set(ctx context.Context, mode storage.ModeSet, addrs ...swarm.Address) (err error) {
+func (s *Store) Set(_ context.Context, _ storage.ModeSet, _ ...swarm.Address) (err error) {
 	panic("not implemented")
 }
 
-func (s *Store) LastPullSubscriptionBinID(bin uint8) (id uint64, err error) {
+func (s *Store) LastPullSubscriptionBinID(_ uint8) (id uint64, err error) {
 	panic("not implemented")
 }
 
-func (s *Store) SubscribePull(ctx context.Context, bin uint8, since uint64, until uint64) (c <-chan storage.Descriptor, closed <-chan struct{}, stop func()) {
+func (s *Store) SubscribePull(_ context.Context, _ uint8, _ uint64, _ uint64) (c <-chan storage.Descriptor, closed <-chan struct{}, stop func()) {
 	panic("not implemented")
 }
 
-func (s *Store) SubscribePush(ctx context.Context, skipf func([]byte) bool) (c <-chan swarm.Chunk, repeat func(), stop func()) {
+func (s *Store) SubscribePush(_ context.Context, _ func([]byte) bool) (c <-chan swarm.Chunk, repeat func(), stop func()) {
 	panic("not implemented")
 }
 
--- a/pkg/storage/inmemstore/store.go
+++ b/pkg/storage/inmemstore/store.go
@@ -33,10 +33,13 @@ func (s *Store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Addres
 func (s *Store) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
 	s.mtx.Lock()
 	defer s.mtx.Unlock()
+
 	for _, ch := range chs {
 		s.store[ch.Address().ByteString()] = ch
 	}
+
 	exist = make([]bool, len(chs))
+
 	return exist, err
 }
 
--- a/pkg/storage/inmemstore/store.go
+++ b/pkg/storage/inmemstore/store.go
@@ -12,6 +12,9 @@ import (
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
+// Store implements a simple Putter and Getter which can be used to temporarily cache
+// chunks. Currently this is used in the bootstrapping process of new nodes where
+// we sync the postage events from the swarm network.
 type Store struct {
 	mtx   sync.Mutex
 	store map[string]swarm.Chunk
--- a/pkg/storage/inmemstore/store.go
+++ b/pkg/storage/inmemstore/store.go
@@ -1,0 +1,73 @@
+package inmemstore
+
+import (
+	"context"
+	"sync"
+
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+type Store struct {
+	mtx   sync.Mutex
+	store map[string]swarm.Chunk
+}
+
+func New() *Store {
+	return &Store{
+		store: make(map[string]swarm.Chunk),
+	}
+}
+
+func (s *Store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Address) (ch swarm.Chunk, err error) {
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	if ch, ok := s.store[addr.ByteString()]; ok {
+		return ch, nil
+	}
+
+	return nil, storage.ErrNotFound
+}
+
+func (s *Store) Put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk) (exist []bool, err error) {
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	for _, ch := range chs {
+		s.store[ch.Address().ByteString()] = ch
+	}
+	exist = make([]bool, len(chs))
+	return exist, err
+}
+
+func (s *Store) GetMulti(ctx context.Context, mode storage.ModeGet, addrs ...swarm.Address) (ch []swarm.Chunk, err error) {
+	panic("not implemented")
+}
+
+func (s *Store) Has(ctx context.Context, addr swarm.Address) (yes bool, err error) {
+	panic("not implemented")
+}
+
+func (s *Store) HasMulti(ctx context.Context, addrs ...swarm.Address) (yes []bool, err error) {
+	panic("not implemented")
+}
+
+func (s *Store) Set(ctx context.Context, mode storage.ModeSet, addrs ...swarm.Address) (err error) {
+	panic("not implemented")
+}
+
+func (s *Store) LastPullSubscriptionBinID(bin uint8) (id uint64, err error) {
+	panic("not implemented")
+}
+
+func (s *Store) SubscribePull(ctx context.Context, bin uint8, since uint64, until uint64) (c <-chan storage.Descriptor, closed <-chan struct{}, stop func()) {
+	panic("not implemented")
+}
+
+func (s *Store) SubscribePush(ctx context.Context, skipf func([]byte) bool) (c <-chan swarm.Chunk, repeat func(), stop func()) {
+	panic("not implemented")
+}
+
+func (s *Store) Close() error {
+	return nil
+}
--- a/cmd/bee/cmd/cmd.go
+++ b/cmd/bee/cmd/cmd.go
@@ -82,6 +82,7 @@ const (
 	optionNameRestrictedAPI              = "restricted"
 	optionNameTokenEncryptionKey         = "token-encryption-key"
 	optionNameAdminPasswordHash          = "admin-password"
+	optionNameUsePostageSnapshot         = "use-postage-snapshot"
 )
 
 func init() {
@@ -277,6 +278,7 @@ func (c *command) setAllFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool(optionNameRestrictedAPI, false, "enable permission check on the http APIs")
 	cmd.Flags().String(optionNameTokenEncryptionKey, "", "admin username to get the security token")
 	cmd.Flags().String(optionNameAdminPasswordHash, "", "bcrypt hash of the admin password to get the security token")
+	cmd.Flags().Bool(optionNameUsePostageSnapshot, false, "bootstrap node using postage snapshot from the network")
 }
 
 func newLogger(cmd *cobra.Command, verbosity string) (logging.Logger, error) {
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -202,6 +202,7 @@ func (c *command) initStartCmd() (err error) {
 				Restricted:                 c.config.GetBool(optionNameRestrictedAPI),
 				TokenEncryptionKey:         c.config.GetString(optionNameTokenEncryptionKey),
 				AdminPasswordHash:          c.config.GetString(optionNameAdminPasswordHash),
+				UsePostageSnapshot:         c.config.GetBool(optionNameUsePostageSnapshot),
 			})
 			if err != nil {
 				return err
--- a/pkg/node/bootstrap.go
+++ b/pkg/node/bootstrap.go
@@ -1,0 +1,431 @@
+package node
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"errors"
+	"fmt"
+	"io"
+	"math/big"
+	"os"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethersphere/bee/pkg/accounting"
+	"github.com/ethersphere/bee/pkg/addressbook"
+	"github.com/ethersphere/bee/pkg/crypto"
+	"github.com/ethersphere/bee/pkg/file/joiner"
+	"github.com/ethersphere/bee/pkg/hive"
+	"github.com/ethersphere/bee/pkg/logging"
+	"github.com/ethersphere/bee/pkg/netstore"
+	"github.com/ethersphere/bee/pkg/p2p/libp2p"
+	"github.com/ethersphere/bee/pkg/pricer"
+	"github.com/ethersphere/bee/pkg/pricing"
+	"github.com/ethersphere/bee/pkg/retrieval"
+	"github.com/ethersphere/bee/pkg/settlement/pseudosettle"
+	"github.com/ethersphere/bee/pkg/settlement/swap"
+	"github.com/ethersphere/bee/pkg/settlement/swap/chequebook"
+	"github.com/ethersphere/bee/pkg/settlement/swap/priceoracle"
+	"github.com/ethersphere/bee/pkg/shed"
+	"github.com/ethersphere/bee/pkg/storage/inmemstore"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/ethersphere/bee/pkg/topology"
+	"github.com/ethersphere/bee/pkg/topology/kademlia"
+	"github.com/ethersphere/bee/pkg/topology/lightnode"
+	"github.com/ethersphere/bee/pkg/tracing"
+	"github.com/ethersphere/bee/pkg/transaction"
+	ma "github.com/multiformats/go-multiaddr"
+	"github.com/sirupsen/logrus"
+)
+
+var snapshotReference = swarm.MustParseHexAddress("b933cd26548ae992e0ee6ba6fabdf9ab7769663d8ddac84b65b1a527ea0734e7")
+
+func NewBeeBootstrapper(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, networkID uint64, logger logging.Logger, libp2pPrivateKey, pssPrivateKey *ecdsa.PrivateKey, o *Options) (b *Bee, err error) {
+	tracer, tracerCloser, err := tracing.NewTracer(&tracing.Options{
+		Enabled:     o.TracingEnabled,
+		Endpoint:    o.TracingEndpoint,
+		ServiceName: o.TracingServiceName,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("tracer: %w", err)
+	}
+
+	p2pCtx, p2pCancel := context.WithCancel(context.Background())
+	defer func() {
+		// if there's been an error on this function
+		// we'd like to cancel the p2p context so that
+		// incoming connections will not be possible
+		if err != nil {
+			p2pCancel()
+		}
+	}()
+
+	b = &Bee{
+		p2pCancel:      p2pCancel,
+		errorLogWriter: logger.WriterLevel(logrus.ErrorLevel),
+		tracerCloser:   tracerCloser,
+	}
+
+	stateStore, err := InitStateStore(logger, o.DataDir)
+	if err != nil {
+		return nil, err
+	}
+	b.stateStoreCloser = stateStore
+
+	addressbook := addressbook.New(stateStore)
+
+	var (
+		swapBackend        transaction.Backend
+		overlayEthAddress  common.Address
+		chainID            int64
+		transactionService transaction.Service
+		transactionMonitor transaction.Monitor
+		chequebookFactory  chequebook.Factory
+		chequebookService  chequebook.Service
+		chequeStore        chequebook.ChequeStore
+		cashoutService     chequebook.CashoutService
+		pollingInterval    = time.Duration(o.BlockTime) * time.Second
+	)
+	swapBackend, overlayEthAddress, chainID, transactionMonitor, transactionService, err = InitChain(
+		p2pCtx,
+		logger,
+		stateStore,
+		o.SwapEndpoint,
+		signer,
+		pollingInterval,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("init chain: %w", err)
+	}
+	b.ethClientCloser = swapBackend.Close
+	b.transactionCloser = tracerCloser
+	b.transactionMonitorCloser = transactionMonitor
+
+	if o.ChainID != -1 && o.ChainID != chainID {
+		return nil, fmt.Errorf("connected to wrong ethereum network; network chainID %d; configured chainID %d", chainID, o.ChainID)
+	}
+
+	// Sync the with the given Ethereum backend:
+	isSynced, _, err := transaction.IsSynced(p2pCtx, swapBackend, maxDelay)
+	if err != nil {
+		return nil, fmt.Errorf("is synced: %w", err)
+	}
+	if !isSynced {
+		logger.Infof("waiting to sync with the Ethereum backend")
+
+		err := transaction.WaitSynced(p2pCtx, logger, swapBackend, maxDelay)
+		if err != nil {
+			return nil, fmt.Errorf("waiting backend sync: %w", err)
+		}
+	}
+
+	if o.SwapEnable {
+		chequebookFactory, err = InitChequebookFactory(
+			logger,
+			swapBackend,
+			chainID,
+			transactionService,
+			o.SwapFactoryAddress,
+			o.SwapLegacyFactoryAddresses,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		if err = chequebookFactory.VerifyBytecode(p2pCtx); err != nil {
+			return nil, fmt.Errorf("factory fail: %w", err)
+		}
+
+		if o.ChequebookEnable {
+			chequebookService, err = InitChequebookService(
+				p2pCtx,
+				logger,
+				stateStore,
+				signer,
+				chainID,
+				swapBackend,
+				overlayEthAddress,
+				transactionService,
+				chequebookFactory,
+				o.SwapInitialDeposit,
+				o.DeployGasPrice,
+			)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		chequeStore, cashoutService = initChequeStoreCashout(
+			stateStore,
+			swapBackend,
+			chequebookFactory,
+			chainID,
+			overlayEthAddress,
+			transactionService,
+		)
+	}
+
+	pubKey, _ := signer.PublicKey()
+	if err != nil {
+		return nil, err
+	}
+
+	var (
+		blockHash []byte
+		txHash    []byte
+	)
+
+	txHash, err = GetTxHash(stateStore, logger, o.Transaction)
+	if err != nil {
+		return nil, fmt.Errorf("invalid transaction hash: %w", err)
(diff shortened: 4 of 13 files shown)


pkg/storage/inmemstore/store.go
```diff
@@ -0,0 +1,80 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package inmemstore
```
aloknerurkar, line 5: @acud Maybe we should combine this and the mock store. Both are essentially the same.

mrekucci (commented):
> Reviewed 7 of 10 files at r2, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2814)* status: all files reviewed, 3 unresolved discussions (waiting on @acud, @aloknerurkar, and @istae)

___
*[pkg/node/bootstrap.go, line 53 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2JT0U01W3jrovP-kC:-Mw2JT0VEIaa3Xyd6l3J:b-c7l7q5) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L53)):*
> ```Go
> var snapshotFeed = swarm.MustParseHexAddress("b36f03d995a04df1757c3a5ddbb795f48d279c532b11803864503f6b97fb20e1")
> 
> func bootstrapNode(addr string,
> ```

The `publicKey` and `pssPrivateKey` params are not used. I'd suggest having every param on a new line not just partially. Also, this function takes way too many parameters, would be nice to break it down or use a complex object to pass the params.
___
*[pkg/node/bootstrap.go, line 88 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2Ia0U5Dv6F_krLQUn:-Mw2Ia0U5Dv6F_krLQUo:bu50k27) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L88)):*
> ```Go
> 		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
> 		defer cancel()
> 		err := multierror.Append(new(multierror.Error), retErr, b.Shutdown(ctx))
> ```

This can be simplified to `retErr = multierror.Append(retErr, b.Shutdown(ctx)).ErrorOrNil()`.
___
*[pkg/node/node.go, line 220 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2Kdk-9-fxxg-oeFuA:-Mw2Kdk-9-fxxg-oeFuB:bmly14d) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/node.go#L220)):*
> ```Go
> 
> 	newStateStore := true
> 	if err := stateStore.Get(secureOverlayKey, new(swarm.Address)); err == nil {
> ```

Why is the error ignored?
___
*[pkg/node/node.go, line 452 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2LN7FE1QRHmrqA8LK:-Mw2LN7FE1QRHmrqA8LL:bn3mhb8) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/node.go#L452)):*
> ```Go
> 	if newStateStore || o.Resync {
> 		start := time.Now()
> 		logger.Infof("cold postage start detected. fetching postage stamp snapshot from swarm")
> ```

`Infof` -> `Info`
___
*[pkg/node/statestore.go, line 26 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2LstABesUEb_JJG-o:-Mw2LstABesUEb_JJG-p:btk7nuu) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/statestore.go#L26)):*
> ```Go
> 		ret := mock.NewStateStore()
> 		log.Warning("using in-mem state store, no node state will be persisted")
> 		return ret, nil
> ```

You can return the `mock.NewStateStore()` directly without creating new value.
___
*[pkg/postage/interface.go, line 28 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2MmiQ8Tiz4onjj9zj:-Mw2MmiQ8Tiz4onjj9zk:ba1jzzx) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/postage/interface.go#L28)):*
> ```Go
> }
> 
> type BatchSnapshot struct {
> ```

A description of the struct would help to understand the context but probably not the best name since `Batch` has more and mostly different values. Maybe `ChainEvents` or `ChainSnapshot`... Also, since the `FirstBlockNumber` and `Timestamp` are only set in the test and never used, are they necessary?

acud (changes requested):
> Reviewed 7 of 10 files at r2, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2814)* status: all files reviewed, 16 unresolved discussions (waiting on @aloknerurkar, @istae, and @mrekucci)

___
*[pkg/node/bootstrap.go, line 79 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2zaOvAaDlA9hoVsbM:-Mw2zaOvAaDlA9hoVsbN:b8yudds) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L79)):*
> ```Go
> 	p2pCtx, p2pCancel := context.WithCancel(context.Background())
> 
> 	b := &Bee{
> ```

nit, I wonder whether this is strictly needed, and whether simple `defer` calls won't do the job?
___
*[pkg/node/bootstrap.go, line 103 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2zvRTCrgVz_UN6m2T:-Mw2zvRTCrgVz_UN6m2U:b-s8hpb2) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L103)):*
> ```Go
> 		return nil, fmt.Errorf("p2p service: %w", err)
> 	}
> 	b.p2pService = p2ps
> ```

the same here about `defer`s
___
*[pkg/node/bootstrap.go, line 131 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw3-EQY-UD-BYAEsLVo:-Mw3-EQY-UD-BYAEsLVp:b2sky0b) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L131)):*
> ```Go
> 	p2ps.SetPickyNotifier(kad)
> 
> 	minThreshold := big.NewInt(2 * refreshRate)
> ```

would be good to extract this stuff (from here and the main node setup phase) to a constant that can be reused, same for line below and other examples below
___
*[pkg/node/bootstrap.go, line 157 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw3-P72C9-m80UVTOEa:-Mw3-P72C9-m80UVTOEb:b3pr8o6) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L157)):*
> ```Go
> 	b.accountingCloser = acc
> 
> 	enforcedRefreshRate := big.NewInt(lightRefreshRate)
> ```

pull out to const
___
*[pkg/node/bootstrap.go, line 185 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw3-W5v9DqhVo_U4ycq:-Mw3-W5v9DqhVo_U4ycr:b-1q0st4) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L185)):*
> ```Go
> 	b.priceOracleCloser = priceOracle
> 
> 	if o.ChequebookEnable {
> ```

I think that it might be a safer bet to assume we don't want to do any chequebook related operations at this stage
___
*[pkg/node/bootstrap.go, line 199 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw3-mywAGN3r-JLzZ1j:-Mw3-mywAGN3r-JLzZ1k:b-5z2ciz) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L199)):*
> ```Go
> 	ns := netstore.New(storer, noopValidStamp, nil, retrieve, logger)
> 
> 	retrieveProtocolSpec := retrieve.Protocol()
> ```

this is not used anywhere else and can be inlined in the line below
___
*[pkg/node/bootstrap.go, line 280 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw30Gpn0dGul7JyrCdS:-Mw30Gpn0dGul7JyrCdT:b-dzd743) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L280)):*
> ```Go
> }
> 
> func (p *noopPinger) Ping(_ context.Context, _ swarm.Address, _ ...string) (time.Duration, error) {
> ```

cosmetics, but I think that if none of the parameters are used, you can even omit `_`
___
*[pkg/node/bootstrap.go, line 284 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw30eKF1APW2Jo5HANN:-Mw30eKF1APW2Jo5HANO:brm1k04) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L284)):*
> ```Go
> }
> 
> func parseFeedUpdate(
> ```

name is a bit misleading, since it is not just parsing but also looking up the latest feed update
___
*[pkg/node/bootstrap.go, line 338 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw30Wlj975rY83nnafn:-Mw30Wlj975rY83nnafo:b-prgfvs) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L338)):*
> ```Go
> 	}
> 
> 	u, _, _, err := l.At(ctx, time.Now().Unix(), 0)
> ```

how lo
___
*[pkg/node/node.go, line 220 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2Kdk-9-fxxg-oeFuA:-Mw30ovn9F5wavg-tYRL:bb74r0b) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/node.go#L220)):*

Why is the error ignored?
</blockquote></details>

because if the error is `nil` it means that the entry was found (not a new state store), if it was not, it most probably means we have a fresh statestore. You could explicitly check all three cases, not a blocker for me, but a comment would be nice
___
*[pkg/node/node.go, line 424 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw31NCfEf4B9mjw7l58:-Mw31NCfEf4B9mjw7l59:bu5lhxc) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/node.go#L424)):*
> ```Go
> 	// Perform checks related to payment threshold calculations here to not duplicate
> 	// the checks in bootstrap process
> 	minThreshold := big.NewInt(2 * refreshRate)
> ```

would be nice to extract this stuff
___
*[pkg/node/node.go, line 467 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw31SxR-v-176Rp6dWw:-Mw31SxS3cPHRUlqAQOA:b-p3yes4) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/node.go#L467)):*
> ```Go
> 			transactionService,
> 			stateStore, publicKey, signer, networkID, logger, libp2pPrivateKey, pssPrivateKey, o)
> 		logger.Infof("bootstrapper done, took", time.Since(start))
> ```

missing formatting verb in declaration (add `%s`)
___
*[pkg/node/node.go, line 469 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw31bb3-4tYVJq3Rg4p:-Mw31bb4EMyXFJNb9exT:bz9395z) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/node.go#L469)):*
> ```Go
> 		logger.Infof("bootstrapper done, took", time.Since(start))
> 		if err != nil {
> 			return nil, fmt.Errorf("bootstrapper failed to fetch batch state %w", err)
> ```

missing semicolon before formatting verb
___
*[pkg/node/statestore.go, line 26 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2LstABesUEb_JJG-o:-Mw31qHzC9hM-Raf7byx:b-mi3ws8) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/statestore.go#L26)):*

You can return the `mock.NewStateStore()` directly without creating new value.
</blockquote></details>

are we using this `mock` in the bootstrapping mode? if yes, it should not be so, we should use the real statestore
___
*[pkg/postage/interface.go, line 28 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2MmiQ8Tiz4onjj9zj:-Mw321Gm0ZuN2qOzoKAE:b-icwnk3) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/postage/interface.go#L28)):*

A description of the struct would help to understand the context but probably not the best name since `Batch` has more and mostly different values. Maybe `ChainEvents` or `ChainSnapshot`... Also, since the `FirstBlockNumber` and `Timestamp` are only set in the test and never used, are they necessary?
</blockquote></details>

It would be nice to have the block numbers for our own orientation, also timestamp, so we could know when a snapshot was generated
___
*[pkg/postage/listener/listener.go, line 203 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw32nue05FuFDVKNpew:-Mw32nue05FuFDVKNpex:bb0g1ov) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/postage/listener/listener.go#L203)):*
> ```Go
> 					return err
> 				}
> 				l.logger.Debugf("listener: %v", err)
> ```

maybe a bit more information here would be nice
___
*[pkg/postage/listener/listener.go, line 221 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw32xPS2gglSxOl_c3x:-Mw32xPS2gglSxOl_c3y:b-r2urty) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/postage/listener/listener.go#L221)):*
> ```Go
> 
> 	if initState != nil {
> 		err := processEvents(initState.Events, initState.LastBlockNumber+1)
> ```

why is it `LastBlockNumber+1` and not just `LastBlockNumber`? seems like a mistake. this will cause the next iteration to start from `LastBlockNumber+2` no?
___
*[pkg/storage/inmemstore/store.go, line 5 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-MvwjH9MGCphMdZoa_8N-r4-5:-Mw33KoH0Cl8RZNNFdnP:b-sclop6) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/storage/inmemstore/store.go#L5)):*

@acud Maybe we should combine this and the mock store. Both are essentially the same.
</blockquote></details>

I would prefer not to mix mocks that are used for testing with production code. It should be clearly delineated which package needs to be used for production code and which isn't. I am fine with eliminating the mock store in favor of this one

aloknerurkar (commented):
> Dismissed @acud from 10 discussions.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2814)* status: 2 of 11 files reviewed, 4 unresolved discussions (waiting on @acud, @istae, and @mrekucci)

___
*[pkg/node/bootstrap.go, line 79 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2zaOvAaDlA9hoVsbM:-MwBnph117ffyEXxG9Jq:b-f4o6tx) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L79)):*

nit, I wonder whether this is strictly needed, and whether simple `defer` calls won't do the job?
</blockquote></details>

I think the Shutdown function adds a lot of value with aggregating the errors as well. I feel this way is cleaner.
___
*[pkg/node/bootstrap.go, line 103 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw2zvRTCrgVz_UN6m2T:-MwBoCtm4mnOT3B7fuIY:blzt8mg) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L103)):*

the same here about `defer`s
</blockquote></details>

Done. Please check the previous response.
___
*[pkg/node/bootstrap.go, line 131 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw3-EQY-UD-BYAEsLVo:-MwBoImzBfpTnSJi8SWQ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L131)):*

would be good to extract this stuff (from here and the main node setup phase) to a constant that can be reused, same for line below and other examples below
</blockquote></details>

Done.
___
*[pkg/node/bootstrap.go, line 157 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw3-P72C9-m80UVTOEa:-MwBoKPB0KpiXhEHQlc5:bumf7up) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L157)):*

pull out to const
</blockquote></details>

This cannot be pulled out, unfortunately. This is a variable in the main `node.go` file.
___
*[pkg/node/bootstrap.go, line 185 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw3-W5v9DqhVo_U4ycq:-MwBoUWG9I_GDQ3Erhwv:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L185)):*

I think that it might be a safer bet to assume we don't want to do any chequebook related operations at this stage
</blockquote></details>

Done.
___
*[pkg/node/bootstrap.go, line 199 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw3-mywAGN3r-JLzZ1j:-MwBoVTf8KippndGqg07:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L199)):*

this is not used anywhere else and can be inlined in the line below
</blockquote></details>

Done.
___
*[pkg/node/bootstrap.go, line 280 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw30Gpn0dGul7JyrCdS:-MwBoW772qRNrcCOraG-:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L280)):*

cosmetics, but I think that if none of the parameters are used, you can even omit `_`
</blockquote></details>

Done.
___
*[pkg/node/bootstrap.go, line 284 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw30eKF1APW2Jo5HANN:-MwBoX6s7GpkzkdTgioK:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L284)):*

name is a bit misleading, since it is not just parsing but also looking up the latest feed update
</blockquote></details>

Done.
___
*[pkg/node/bootstrap.go, line 338 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw30Wlj975rY83nnafn:-MwQ9rg0DQWjhofz5SLj:b-xcyau1) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L338)):*

how lo
</blockquote></details>

As is? Do you mean the last parameter in this lookup? So I am not sure how to determine this initial value.
___
*[pkg/node/node.go, line 424 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw31NCfEf4B9mjw7l58:-MwBxYmy8HIH_BLpH0j_:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/node.go#L424)):*

would be nice to extract this stuff
</blockquote></details>

Done.
___
*[pkg/node/node.go, line 467 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw31SxR-v-176Rp6dWw:-MwC61ZAB15-ThLhtVEs:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/node.go#L467)):*

missing formatting verb in declaration (add `%s`)
</blockquote></details>

Done.
___
*[pkg/node/node.go, line 469 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw31bb3-4tYVJq3Rg4p:-MwC6K3M9tje-5Dmf2jn:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/node.go#L469)):*

missing semicolon before formatting verb
</blockquote></details>

Done.
___
*[pkg/postage/listener/listener.go, line 203 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw32nue05FuFDVKNpew:-MwCBwHBCPyWpYOpA346:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/postage/listener/listener.go#L203)):*

maybe a bit more information here would be nice
</blockquote></details>

Done.
___
*[pkg/postage/listener/listener.go, line 221 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw32xPS2gglSxOl_c3x:-MwCBXmE0kmHPW-JiJ71:b-gh8945) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/postage/listener/listener.go#L221)):*

why is it `LastBlockNumber+1` and not just `LastBlockNumber`? seems like a mistake. this will cause the next iteration to start from `LastBlockNumber+2` no?
</blockquote></details>

If you see the test, this behavior seems right. We provide events 496...499 and then the last `BlockNumber` call is for 500 which then stops the test. This is correct? If I remove the `+1` I get 2 calls for 499 block number.
___
*[pkg/storage/inmemstore/store.go, line 5 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-MvwjH9MGCphMdZoa_8N-r4-5:-MwCAWbP8GG720f4YUym:by6n6pw) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/storage/inmemstore/store.go#L5)):*

I would prefer not to mix mocks that are used for testing with production code. It should be clearly delineated which package needs to be used for production code and which isn't. I am fine with eliminating the mock store in favor of this one
</blockquote></details>

I went through the MockStorer implementation and it has a bunch of additional things. I saw the test file and assumed they are the same but there are a bunch of other test options which are in their place. I have added some small comments in the inmemstore to explain its use.

istae (changes requested):
> Reviewed 2 of 4 files at r4, 7 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2814)* status: 9 of 11 files reviewed, 6 unresolved discussions (waiting on @acud, @aloknerurkar, and @mrekucci)

___
*[pkg/node/bootstrap.go, line 235 at r5](https://reviewable.io/reviews/ethersphere/bee/2814#-MwSHsrq31JT5OIH5Blj:-MwSHsrr9EMa1NTVuRHZ:b-w3ab1v) ([raw file](https://github.com/ethersphere/bee/blob/e380aff1f3318bead1ad5f44048fbc1dd805fbd8/pkg/node/bootstrap.go#L235)):*
> ```Go
> 	items := 0
> 	c := make(chan struct{})
> 	go func() {
> ```

What is the purpose of waiting 10 seconds here? We can also use `time.NewTimer` package function to simplify this.
___
*[pkg/node/statestore.go, line 26 at r5](https://reviewable.io/reviews/ethersphere/bee/2814#-MwSGzF39ipB9UzcM0i3:-MwSGzF39ipB9UzcM0i4:b-1uxb99) ([raw file](https://github.com/ethersphere/bee/blob/e380aff1f3318bead1ad5f44048fbc1dd805fbd8/pkg/node/statestore.go#L26)):*
> ```Go
> 		stateStoreDir = filepath.Join(dataDir, "statestore")
> 	} else {
> 		log.Warning("using in-mem state store, no node state will be persisted")
> ```

Is there return of an in-memory state store missing here?

mrekucci (commented):
> Reviewed 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2814)* status: all files reviewed, 6 unresolved discussions (waiting on @acud and @aloknerurkar)

___
*[pkg/node/node.go, line 224 at r5](https://reviewable.io/reviews/ethersphere/bee/2814#-MwSPUkl1ZL270l2AoS7:-MwSPUkl1ZL270l2AoS8:b-fq06jy) ([raw file](https://github.com/ethersphere/bee/blob/e380aff1f3318bead1ad5f44048fbc1dd805fbd8/pkg/node/node.go#L224)):*
> ```Go
> 	// Check if the overlay is found in the statestore. If not, we can assume it has
> 	// not been created yet and treat this as a fresh install.
> 	if err := stateStore.Get(secureOverlayKey, new(swarm.Address)); errors.Is(err, storage.ErrNotFound) {
> ```

We're still ignoring other errors returned by the `Get` method. E.g.: the unmarshaling can fail and the error will be silently ignored.

aloknerurkar (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2814)* status: all files reviewed, 6 unresolved discussions (waiting on @acud and @istae)

___
*[pkg/node/bootstrap.go, line 235 at r5](https://reviewable.io/reviews/ethersphere/bee/2814#-MwSHsrq31JT5OIH5Blj:-MwVOTWf-1H7xTQmoysd:b-11bksb) ([raw file](https://github.com/ethersphere/bee/blob/e380aff1f3318bead1ad5f44048fbc1dd805fbd8/pkg/node/bootstrap.go#L235)):*

What is the purpose of waiting 10 seconds here? We can also use `time.NewTimer` package function to simplify this.
</blockquote></details>

I dont think this is needed at all. Missed this before. Removed it now. We already wait 30 secs in the worst case. Best case we dont need to wait 10 secs I believe.
___
*[pkg/node/node.go, line 224 at r5](https://reviewable.io/reviews/ethersphere/bee/2814#-MwSPUkl1ZL270l2AoS7:-MwVLo5L74iqIoPjUNMB:b4jiv5f) ([raw file](https://github.com/ethersphere/bee/blob/e380aff1f3318bead1ad5f44048fbc1dd805fbd8/pkg/node/node.go#L224)):*

We're still ignoring other errors returned by the `Get` method. E.g.: the unmarshaling can fail and the error will be silently ignored.
</blockquote></details>

So we don't care about the errors here. All other cases should be handled by `CheckOverlayWithStore` function. Check the comment.
___
*[pkg/node/statestore.go, line 26 at r5](https://reviewable.io/reviews/ethersphere/bee/2814#-MwSGzF39ipB9UzcM0i3:-MwVNQWyBaD6sdzebVEo:bq7nriu) ([raw file](https://github.com/ethersphere/bee/blob/e380aff1f3318bead1ad5f44048fbc1dd805fbd8/pkg/node/statestore.go#L26)):*

Is there return of an in-memory state store missing here?
</blockquote></details>

`stateStoreDir` is initialized to `""`. So this will create an in-memory levelDB instance if no path is given. We were using the mock store which was incorrect.

istae (approved):

acud (changes requested):
few minor things but looks great

> Reviewed 8 of 9 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2814#-:-MwX_PKk6aEt3nurzTgK:b-968igx)* status: all files reviewed, 3 unresolved discussions (waiting on @aloknerurkar, @istae, and @mrekucci)

___
*[pkg/node/bootstrap.go, line 338 at r4](https://reviewable.io/reviews/ethersphere/bee/2814#-Mw30Wlj975rY83nnafn:-MwXYzpS5YqhNKvwMxU0:b-zc3jdg) ([raw file](https://github.com/ethersphere/bee/blob/aabf96a6ce3984b54228a70c66f60736a63e29b4/pkg/node/bootstrap.go#L338)):*

As is? Do you mean the last parameter in this lookup? So I am not sure how to determine this initial value.
</blockquote></details>

sorry for the partial comment, I missed it in the previous review. Started writing something but then took it back as usual :) I think I meant to ask how long is this supposed to take and about the context cancellation, but I remember looking at the context expiration for the whole operation and it should be 30 seconds or something, which is fine.
___
*[pkg/node/bootstrap.go, line 235 at r5](https://reviewable.io/reviews/ethersphere/bee/2814#-MwSHsrq31JT5OIH5Blj:-MwXYLEw7VAHS6jJfxtx:b-8re9w7) ([raw file](https://github.com/ethersphere/bee/blob/e380aff1f3318bead1ad5f44048fbc1dd805fbd8/pkg/node/bootstrap.go#L235)):*

I dont think this is needed at all. Missed this before. Removed it now. We already wait 30 secs in the worst case. Best case we dont need to wait 10 secs I believe.
</blockquote></details>

I originally wrote this so that we error if we don't get any peers in kademlia after ten seconds. The idea was that you are supposed to connect to some peers rather quickly and if you don't then it is better to error the operation. I'm pretty sure there used to be some select here too... but it might have gotten deleted :) anyway, it is fine to remove as was done.
___
*[pkg/node/statestore.go, line 26 at r5](https://reviewable.io/reviews/ethersphere/bee/2814#-MwSGzF39ipB9UzcM0i3:-MwXZj3r0W09FT0EFRp4:b-7xnd3v) ([raw file](https://github.com/ethersphere/bee/blob/e380aff1f3318bead1ad5f44048fbc1dd805fbd8/pkg/node/statestore.go#L26)):*

`stateStoreDir` is initialized to `""`. So this will create an in-memory levelDB instance if no path is given. We were using the mock store which was incorrect.
</blockquote></details>

I think that if it is in memory you need to use the `NewInMemoryStateStore` constructor in that case no?

aloknerurkar (commented):
> Dismissed @acud and @istae from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2814)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @acud, @istae, and @mrekucci)

___
*[pkg/node/bootstrap.go, line 235 at r5](https://reviewable.io/reviews/ethersphere/bee/2814#-MwSHsrq31JT5OIH5Blj:-Mx-1sis6xJL2_SibdWZ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/e380aff1f3318bead1ad5f44048fbc1dd805fbd8/pkg/node/bootstrap.go#L235)):*

I originally wrote this so that we error if we don't get any peers in kademlia after ten seconds. The idea was that you are supposed to connect to some peers rather quickly and if you don't then it is better to error the operation. I'm pretty sure there used to be some select here too... but it might have gotten deleted :) anyway, it is fine to remove as was done.
</blockquote></details>

Done.
___
*[pkg/node/statestore.go, line 26 at r5](https://reviewable.io/reviews/ethersphere/bee/2814#-MwSGzF39ipB9UzcM0i3:-Mx-1uvEEJaaKJlL-rbT:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/e380aff1f3318bead1ad5f44048fbc1dd805fbd8/pkg/node/statestore.go#L26)):*

I think that if it is in memory you need to use the `NewInMemoryStateStore` constructor in that case no?
</blockquote></details>

Done.

acud (approved):
:lgtm:

> Reviewed 10 of 11 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2814#-:-MxKpkNN4EFhv5ikTQgu:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar and @mrekucci)

___
*[pkg/node/node.go, line 460 at r10](https://reviewable.io/reviews/ethersphere/bee/2814#-MxKpcSI9DneUROdbTyA:-MxKpcSI9DneUROdbTyB:b-bz5y8c) ([raw file](https://github.com/ethersphere/bee/blob/865a883aa357df279d633d06a9a1d03b0c7a7992/pkg/node/node.go#L460)):*
> ```Go
> 	var initBatchState *postage.ChainSnapshot
> 	// only use bootstrapping for mainnet nodes
> 	if networkID == 1 && (newStateStore || o.Resync) {
> ```

nit, consider pulling those network IDs to some `const` somewhere.... totally not a blocker tho

mrekucci (approved):
> Reviewed 9 of 11 files at r9, 4 of 4 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2814)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar)

mrekucci (approved):
> Reviewed 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2814)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aloknerurkar)

PR #2821, opened 2022-02-18
fix: return empty array instead of null

notanatol:
### Checklist

### Description
our collection responses are not aligned when it comes to empty arrays - we get mixed `[]` and `null`

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2821)

### DIFF
--- a/openapi/SwarmDebug.yaml
+++ b/openapi/SwarmDebug.yaml
@@ -1,6 +1,6 @@
 openapi: 3.0.3
 info:
-  version: 1.2.1
+  version: 1.2.2
   title: Bee Debug API
   description: "A list of the currently provided debug interfaces to interact with the bee node"
 
--- a/openapi/SwarmDebug.yaml
+++ b/openapi/SwarmDebug.yaml
@@ -1,6 +1,6 @@
 openapi: 3.0.3
 info:
-  version: 1.2.2
+  version: 2.2.1
   title: Bee Debug API
   description: "A list of the currently provided debug interfaces to interact with the bee node"
 
--- a/openapi/SwarmDebug.yaml
+++ b/openapi/SwarmDebug.yaml
@@ -1,6 +1,6 @@
 openapi: 3.0.3
 info:
-  version: 2.2.1
+  version: 2.0.0
   title: Bee Debug API
   description: "A list of the currently provided debug interfaces to interact with the bee node"
 
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 
 info:
-  version: 2.0.0
+  version: 2.0.1
   title: Bee API
   description: "A list of the currently provided Interfaces to interact with the swarm, implementing file operations and sending messages"
 
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 
 info:
-  version: 2.0.1
+  version: 3.0.0
   title: Bee API
   description: "A list of the currently provided Interfaces to interact with the swarm, implementing file operations and sending messages"
 
--- a/openapi/SwarmCommon.yaml
+++ b/openapi/SwarmCommon.yaml
@@ -287,7 +287,7 @@ components:
       properties:
         peers:
           type: array
-          nullable: true
+          nullable: false
           items:
             $ref: "#/components/schemas/Address"
 
@@ -321,7 +321,7 @@ components:
       properties:
         stamps:
           type: array
-          nullable: true
+          nullable: false
           items:
             $ref: "#/components/schemas/DebugPostageBatch"
 
@@ -531,7 +531,7 @@ components:
       properties:
         addresses:
           type: array
-          nullable: true
+          nullable: false
           items:
             $ref: "#/components/schemas/SwarmOnlyReference"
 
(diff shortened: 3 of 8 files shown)


acud (approved):

acud:
just worth to make sure that this is reflected in the openapi spec

AuHau:
Awesome that there is work on this! I have to remarks though.

First of all, just through searching of the OpenAPI spec, I have found 9 places where this is a problem, yet this PR tackles only three. I would suggest tackling all of those.

Second of all, be aware that this is a breaking change for the API, so please signal that with a correct bump of OpenAPI versions.

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2821)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @AuHau)

___
*[openapi/SwarmCommon.yaml, line 532 at r1](https://reviewable.io/reviews/ethersphere/bee/2821#-MwRiAA9BhSoW2APRt22:-MwRiAA9BhSoW2APRt23:b-jm42q) ([raw file](https://github.com/ethersphere/bee/blob/fe658d3adb9d97737efafde960fcd45474f422e9/openapi/SwarmCommon.yaml#L532)):*
> ```YAML
>       type: object
>       properties:
>         addresses:
> ```

This should be also renamed to `references` in order to reflect the real response.
___
*[pkg/api/pin.go, line 131 at r1](https://reviewable.io/reviews/ethersphere/bee/2821#-MwRcxLDCXyqkSqrjd0r:-MwRcxLDCXyqkSqrjd0s:blcdwh4) ([raw file](https://github.com/ethersphere/bee/blob/fe658d3adb9d97737efafde960fcd45474f422e9/pkg/api/pin.go#L131)):*
> ```Go
> 	}
> 
> 	if pinned == nil {
> ```

This doesn't make much sense, because the `Pins()` method already returns an empty array if there was no error.
___
*[pkg/debugapi/peer.go, line 104 at r1](https://reviewable.io/reviews/ethersphere/bee/2821#-MwRdusK8coP4fqpBqV7:-MwRdusLE9S9uOAzGpk7:b-fzave9) ([raw file](https://github.com/ethersphere/bee/blob/fe658d3adb9d97737efafde960fcd45474f422e9/pkg/debugapi/peer.go#L104)):*
> ```Go
> 
> func mapPeers(peers []p2p.Peer) (out []Peer) {
> 	out = make([]Peer, 0)
> ```

This can be optimized by specifying the `len(peers)` in the `make` call.
___
*[pkg/debugapi/postage.go, line 188 at r1](https://reviewable.io/reviews/ethersphere/bee/2821#-MwRezv0-j7ieGmi5Hy4:-MwRezv0-j7ieGmi5Hy5:b-6soam7) ([raw file](https://github.com/ethersphere/bee/blob/fe658d3adb9d97737efafde960fcd45474f422e9/pkg/debugapi/postage.go#L188)):*
> ```Go
> 	}
> 	if resp.Stamps == nil {
> 		resp.Stamps = make([]postageStampResponse, 0)
> ```

The stamps can be pre-allocated at beginning with `len(s.post.StampIssuers())`.

notanatol (commented):
version bumped!
I made sure all the other `nullable` responses use empty arrays instead of `null`.
Also, `/batches` endpoint was returning a bare array, not it's wrapped in an object as described by the spec.

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2821#-:-MwSS4gBBltF8WMMXFmv:b-s9iapz)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @AuHau and @mrekucci)

___
*[pkg/api/pin.go, line 131 at r1](https://reviewable.io/reviews/ethersphere/bee/2821#-MwRcxLDCXyqkSqrjd0r:-MwRqTzSBXebE5U0hz-5:b-rhqxct) ([raw file](https://github.com/ethersphere/bee/blob/fe658d3adb9d97737efafde960fcd45474f422e9/pkg/api/pin.go#L131)):*

This doesn't make much sense, because the `Pins()` method already returns an empty array if there was no error.
</blockquote></details>

I don't know, in `pinning.go:142` there only declaration, but it's not allocated
but point taken, I can move it there

mrekucci (approved):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2821)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @AuHau)

AuHau:
> I made sure all the other nullable responses use empty arrays instead of null.

Sweet!

> version bumped!

Well, since it is a breaking change I would argue the major version of the versions should be bumped instead of patch version.

notanatol (commented):
bumped now

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2821#-:-MwT15QP6YhQm9YX67e6:b14crps)* status: 3 of 7 files reviewed, all discussions resolved (waiting on @AuHau and @mrekucci)

openapi/SwarmDebug.yaml
```diff
@@ -1,6 +1,6 @@
 openapi: 3.0.3
 info:
-  version: 1.2.1
+  version: 2.2.1
```
AuHau, line 3:
```suggestion
  version: 2.0.0
```

"higher-level" version bump resets the "lower-level" versions ;-)

AuHau (changes requested):

AuHau (approved):
LGTM! Thanks

PR #2825, opened 2022-02-22
refactor: batchstore migration

istae:
### Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2825)

### DIFF
--- a/pkg/postage/batchstore/store.go
+++ b/pkg/postage/batchstore/store.go
@@ -286,10 +286,7 @@ func (s *store) Reset() error {
 
 	const prefix = "batchstore_"
 	if err := s.store.Iterate(prefix, func(k, _ []byte) (bool, error) {
-		if err := s.store.Delete(string(k)); err != nil {
-			return false, err
-		}
-		return false, nil
+		return false, s.store.Delete(string(k))
 	}); err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


pkg/postage/batchstore/store.go
```diff
@@ -288,6 +305,80 @@ func (s *store) Reset() error {
 	return nil
 }
 
+// migrate converts batchstore to the refactored version, removes unused indexes, computes radius, and
+// and calculates available capacity.
+func (s *store) migrate() error {
```
acud, line 310: not sure why this is needed.... i would recommend to just delete all batchstore keys as part of a statestore migration and force the batchstore to resync all chain data.

istae: fair enough, my thought process was that we could get away with not having to force users to a full resync, but it is the simplest and the least risky method.

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2825)* status: all files reviewed, 1 unresolved discussion (waiting on @acud)

___
*[pkg/postage/batchstore/store.go, line 92 at r1](https://reviewable.io/reviews/ethersphere/bee/2825#-Mwb5yhq1fPy9FbJtCvg:-Mwb5yhs6kQXj-5jFECR:b-liimwu) ([raw file](https://github.com/ethersphere/bee/blob/94644d03183b7c2789546502c76562ea4b821afc/pkg/postage/batchstore/store.go#L92)):*
> ```Go
> 	err = st.Get(batchstoreVersion, &migrated)
> 	if err != nil {
> 		if err == storage.ErrNotFound {
> ```

I'd suggest using `errors.Is(err, storage.ErrNotFound)` here, since it'll also capture nested errors.

mrekucci (approved):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2825)* status: all files reviewed, 1 unresolved discussion (waiting on @acud)

PR #2827, opened 2022-02-23
feat: handy http client scripts

notanatol:
### Description

A non-comprehensive list of http requests for api/debug

### Installation

you will need this plugin (VS Code)

REST Client
v0.24.6
Huachao Mao

Not sure if it's available for other IDEs though...

### Motivation
I keep this scripts outside of the repo but maybe it makes sense to share them so we could collaboratively enrich them with all the missing requests.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2827)

### DIFF
--- a/scripts/bee-api.http
+++ b/scripts/bee-api.http
@@ -1,0 +1,60 @@
+# api endpoints
+@api = http://localhost:1633
+@token = fijSzdudi8j4PYkWGMsqwdeKyCy/ykGTcvltoLQElHDQG/ULgZd4iJj6iOdzEtlm3K1mGSrucBjl5/Dixp0jUyqJmVbUXkvAolBZEg5npwKnHZNDUGUZFw==
+
+### auth dGVzdDp0ZXN0=test:test OmhlbGxv=:hello
+POST {{api}}/auth
+Content-Type: application/json
+
+{
+    "role": "maintainer",
+    "expiry": 315360000
+}
+
+### auth dGVzdDp0ZXN0=test:test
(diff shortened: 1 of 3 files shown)


mrekucci:
In general this is a good idea, but the problem I have with this approach is the additional burden of keeping in sync with our Open API specification and our Golang implementation of the specification. In my case (Goland IDE) I am able to generate per-click HTTP requests from the Open API specification we already have, maybe such a plugin exists for VS Code and other IDEs.

notanatol:
> In general this is a good idea, but the problem I have with this approach is the additional burden of keeping in sync with our Open API specification and our Golang implementation of the specification. In my case (Goland IDE) I am able to generate per-click HTTP requests from the Open API specification we already have, maybe such a plugin exists for VS Code and other IDEs.

that is a fair point
however what I was thinking is that there is a broader community of contributors to whom these scripts could be useful:
it's more lightweight/expressive than having a postman collection, you don't have to do extra steps like generating from openAPI and even before installing any plugins they already give the user an overview of what the general shape of our APIs is.

aloknerurkar:
I don't think we need this in the main codebase. I am sure there are a lot of developers like me who might not use VSCode or other popular IDEs. The risk to reward ratio is too high for me. These will be added now and we might think we will enrich it over time, but in most cases, it will go unnoticed until someone unearths it later at which point this library or plugin might not be maintained or no longer exists etc.

Having these in a separate repo makes more sense. Some documentation can be added that points to this separate repo from our docs website.

PR #2835, opened 2022-03-03
chore(deps): update go-ethereum

aloknerurkar:
### Checklist

### Description
Updating the `karalabe/usb` dependency to fix m1 build warnings

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2835)

### DIFF
--- a/pkg/crypto/eip712/typeddata.go
+++ b/pkg/crypto/eip712/typeddata.go
@@ -7,15 +7,15 @@ package eip712
 import (
 	"fmt"
 
-	"github.com/ethereum/go-ethereum/signer/core"
+	"github.com/ethereum/go-ethereum/signer/core/apitypes"
 )
 
 // type aliases to avoid importing "core" everywhere
-type TypedData = core.TypedData
-type TypedDataDomain = core.TypedDataDomain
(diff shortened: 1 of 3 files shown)


vandot (approved):
All warnings are gone

mrekucci (approved):

pkg/crypto/eip712/typeddata.go
```diff
@@ -7,15 +7,15 @@ package eip712
 import (
 	"fmt"
 
-	"github.com/ethereum/go-ethereum/signer/core"
+	"github.com/ethereum/go-ethereum/signer/core/apitypes"
```
mrekucci, line 10: How does this relate to the m1 usb?

aloknerurkar: So the dep in question is `github.com/karalabe/usb` which is an indirect dependency because of `go-ethereum`. If we update `go-ethereum`, this package was changed and we need to update.

So seems like it will be better to do this in the next release as the RC is already out. I will hold this merge till we tag the next release.

notanatol (approved):

PR #2836, opened 2022-03-04
fix: runtime panic

notanatol:
### Checklist

### Description
Fixes the `/chequebook` runtime panic.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2836)

### DIFF
--- a/pkg/debugapi/debugapi.go
+++ b/pkg/debugapi/debugapi.go
@@ -8,7 +8,6 @@
 package debugapi
 
 import (
-	"context"
 	"crypto/ecdsa"
 	"math/big"
 	"net/http"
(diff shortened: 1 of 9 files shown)


pkg/debugapi/debugapi.go
```diff
@@ -125,6 +126,9 @@ func (s *Service) Configure(overlay swarm.Address, p2p p2p.DebugService, pingpon
 	s.chequebook = chequebook
 	s.swapEnabled = swapEnabled
 	s.swap = swap
+	if !chainEnabled {
```
aloknerurkar, line 129: This extra parameter can be avoided by either initialising swap to `noOpSwap` before passing here or maybe checking if the `swap == nil`.

pkg/debugapi/debugapi.go
```diff
@@ -145,3 +149,54 @@ func (s *Service) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	h.ServeHTTP(w, r)
 }
+
+type noOpSwap struct {
```
aloknerurkar, line 153: So I went through the APIs to see how they are handled. Seems like there are a few errors like ErrNoChequebook which are being handled. I wonder if we should retain the errors here instead of returning empty responses? If API expects to send No Chequebook error if there is no chequebook, then this would sort of break that no?

acud (approved):

PR #2848, opened 2022-03-16
fix: incorrect release of sharky locations

### LINKED ISSUE #2845: 1.5.0rc3 (and rc2) store invalid chunks on Windows hosts
### Context
1.5.0rc3 and rc2 executing on a native Windows host.  The issue was not reproducible on a Linux host.

### Summary
When uploading a large (28,000+ files over 122,000+ chunks), the pusher is observing "stream reset" errors on some chunks.  The peers in those resets were logging "invalid chunk" errors from the PushSync handler.  This occurs on a very small fraction of the chunks, but seemingly the same subset of chunks when uploading through a given node.

This behavior was observed on two different nodes on two different Windows hosts using the same upload dataset.

The dataset is zoom levels 0 through 8 of my OSM dataset composed of mostly small (256x256 pixel) PNG tile files.

### Expected behavior
The push should be clean and invalid chunks should not exist.

### Actual behavior
Invalid chunks are happening on a duplicable basis, at least on Windows hosts.

I added the following code to pushsync's PushChunkToClosest to reliably detect the issue directly in the uploading/pushing node to avoid the complexity of needing to look at the target peer logs.

```
	if !cac.Valid(ch) {
		if !soc.Valid(ch) {
			ps.logger.Debugf("PushChunkToClosest:Invalid chunk %s", ch.Address())
			return nil, swarm.ErrInvalidChunk
		}
	}
```

### Steps to reproduce
Here's the steps that I sent to Alok on Discord (https://discord.com/channels/799027393297514537/834408095689343016/953066113019813929):

The following link contains zooms 0 through 8 of my OSM data set.  I have used this data to duplicate the Invalid Chunks in the push queue on a freshly db-nuked uploading node.  I'm still working on the tool you can use to upload this directory tree of files to duplicate the issue.
https://ldeffenb.dnsalias.net/swarm/OSM-Zoom-0-8.7z

The next thing you'll need is my work-in-progress of a directory manifest pusher for the swarm.  I'm working on this for the Wikipedia WAM bounty.
https://github.com/ldeffenb/Swarming-Wikipedia

Of course, I have no idea if this will actually work on anyone's machine but my own.  And you'll need to edit src/index.ts to customize some variables.

1) Near the top, there are two objects (nsTestnet1 and oldLaptopBee6).  Make a copy of one of these and update it appropriately for your testnet node and a valid stamp within that node.  My stamp depth 28 stamp gets 0.22% utilized on this upload.  YMMV.

2) Set the useNode to point to your newly created object.

3) Near the very bottom you should find a definition of srcDir.  This should point to the directory you unzipped from the 7z above.  Point inside the OSM-Zoom-0-8 directory where the 0 through 8 directories are and a few other files.

4) I run this with "npm run run", but you likely need to do an "npm install" or something similar first to get all of the dependencies installed.

And of course, you'll need the checker code in your bee node just inside the pushsync's PushChunkToClosest method.  It should look like:
    if !cac.Valid(ch) {
        if !soc.Valid(ch) {
            ps.logger.Debugf("PushChunkToClosest:Invalid chunk %s", ch.Address())
            return nil, swarm.ErrInvalidChunk
        }
    }

Then fire up your newly built (dirty) bee node with at least --verbosity 4 into a log file.  Then I tail / grep that log file for "PushChunkToClosest".  Wait for the pusher startup delay, then run the Swarming-Wikipedia and you should shortly see the invalid chunk messages.

To re-do the test, just db nuke your bee node, restart it, wait the 20 minute pusher warmup period, and re-run the Swarming-Wikipedia script.  You should see more invalid chunk messages.  At least, I do!

### Possible solution
Still researching that one!

aloknerurkar:
Fixes #2845

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2848)

### DIFF
--- a/pkg/localstore/mode_put.go
+++ b/pkg/localstore/mode_put.go
@@ -295,7 +295,12 @@ func (db *DB) putRequest(
 			return false, 0, 0, err
 		}
 
-		l, err := sharky.LocationFromBinary(previous.Location)
+		previousIdx, err := db.retrievalDataIndex.Get(previous)
+		if err != nil {
+			return false, 0, 0, fmt.Errorf("could not fetch previous item: %w", err)
+		}
+
+		l, err := sharky.LocationFromBinary(previousIdx.Location)
 		if err != nil {
 			return false, 0, 0, err
 		}
@@ -383,7 +388,12 @@ func (db *DB) putUpload(
 			return false, 0, fmt.Errorf("same slot remove: %w", err)
 		}
 
-		l, err := sharky.LocationFromBinary(previous.Location)
+		previousIdx, err := db.retrievalDataIndex.Get(previous)
+		if err != nil {
+			return false, 0, fmt.Errorf("could not fetch previous item: %w", err)
+		}
+
+		l, err := sharky.LocationFromBinary(previousIdx.Location)
 		if err != nil {
 			return false, 0, err
 		}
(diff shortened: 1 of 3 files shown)


notanatol (approved):

janos (approved):
Excelent catch.

PR #2857, opened 2022-03-22
fix: blocklisting peers after waking up from hibernation

mrekucci:
Stops the internal clock of the blocker instance during hibernation so that the peer list is not blocked upon wakeup.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2857)

### DIFF
--- a/pkg/blocker/blocker.go
+++ b/pkg/blocker/blocker.go
@@ -5,25 +5,31 @@
 package blocker
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/p2p"
 	"github.com/ethersphere/bee/pkg/swarm"
+	"go.uber.org/atomic"
 )
 
+// sequencerResolution represents monotonic sequencer resolution.
+// It must be in the time.Duration base form without a multiplier.
+var sequencerResolution = time.Second
+
 type peer struct {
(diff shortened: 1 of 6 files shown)


pkg/blocker/blocker.go
```diff
@@ -33,11 +39,17 @@ type Blocker struct {
 }
 
 func New(dis p2p.Blocklister, flagTimeout, blockDuration, wakeUpTime time.Duration, callback func(swarm.Address), logger logging.Logger) *Blocker {
+	if flagTimeout <= resolution {
+		panic(fmt.Errorf("flag timeout %v cannot be equal to or lower then the clock resolution %v", flagTimeout, resolution))
+	}
+	if wakeUpTime < resolution {
+		panic(fmt.Errorf("wakeup time %v cannot be lower then the clock resolution %v", wakeUpTime, resolution))
```
notanatol, line 46: wouldn't panicking here potentially cause inconsistencies in the store?

mrekucci: These panics basically mean programmer error. It makes no sense to check for flagged peers more often than the sequencer resolution rate.

notanatol: right, but would panicking trigger the shutdown sequence? or would it just quick the process?

mrekucci: Will crash the process.

notanatol (approved):

janos (approved):
👍

aloknerurkar (approved):

PR #2859, opened 2022-03-23
feat: default configs for mainnet

istae:
### Checklist


vandot:
@istae we discussed a lot of placing cloudflare ens as a default and we agreed not to do it
For xdaichain rpc maybe it makes sense to specify it but I think new url is https://rpc.gnosischain.com/
This two changes needs to be confirmed from other members as well

mrekucci (approved):

notanatol (approved):

PR #2870, opened 2022-04-05
fix: hotfix for v1.5.0 download race

aloknerurkar:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2870)

### DIFF
--- a/pkg/netstore/netstore.go
+++ b/pkg/netstore/netstore.go
@@ -15,11 +15,13 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ethersphere/bee/pkg/cac"
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/postage"
 	"github.com/ethersphere/bee/pkg/recovery"
 	"github.com/ethersphere/bee/pkg/retrieval"
 	"github.com/ethersphere/bee/pkg/sctx"
+	"github.com/ethersphere/bee/pkg/soc"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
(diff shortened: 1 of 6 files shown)


janos:
It looks like that a test needs to be adjusted for new netstore changes https://github.com/ethersphere/bee/runs/5835696394?check_suite_focus=true#step:11:5530.

mrekucci (approved):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2870)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar and @istae)

janos (approved):

notanatol (approved):

pkg/netstore/netstore.go
```diff
@@ -58,6 +60,17 @@ func New(s storage.Storer, validStamp postage.ValidStampFn, rcb recovery.Callbac
 // local-store.
 func (s *store) Get(ctx context.Context, mode storage.ModeGet, addr swarm.Address) (ch swarm.Chunk, err error) {
 	ch, err = s.Storer.Get(ctx, mode, addr)
+	if err == nil {
+		// ensure the chunk we get locally is valid. If not, retrieve the chunk
+		// from network. If there is any corruption of data in the local storage,
+		// this would ensure it is retrieved again from network and added back with
+		// the correct data
+		if !cac.Valid(ch) && !soc.Valid(ch) {
+			err = errors.New("invalid chunk")
```
zelig, line 69: problem is, this error will not let this go to retrieval (see below L75-91) but wraps it as netstore error and returns (L92)

zelig (dismissed):

pkg/netstore/netstore.go
aloknerurkar: Sorry for that. Thanks for catching this!

The current master is running with this fix and passing all iterations. So ideally we don't need this additional check. But we all think this is currently nice to have as this will ensure download doesnt return garbage. The hit in performance is not so much.

aloknerurkar: I have added a test to verify this as well.

zelig: In fact I would like to not have this check at all. We trust our own DB and chunker. 
The circumstance that we introduced a bug does not justify a check in the production code.

aloknerurkar: @zelig    cc @metacertain @janos 
The problem is this issue was very difficult to identify by going through the code. There are multiple things that led to this together. Secondly there are other reasons why a disk could potentially get corrupted without us introducing any code changes as well.

Given all this, we thought it would be okay to not trust local storage as well. This was thought to be in-line with other assumptions that we make. For eg. we have this check in the protocols already. So during download this would be happening either way. In the interest of doing the release at the earliest, would it be okay to keep this change now and maybe have a discussion internally whether or not to keep it in the upcoming release?

We are already ready with the changes for the next version. This can be done immediately.

PR #2874, opened 2022-04-07
chore: consolidate all secrets variables

vandot:
This PR consolidates all secrets variables, we defined some org wide variables and bee will use them, we will rotate and manage those variables more easily

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2874)


istae (approved):
thanks!

mrekucci (approved):

PR #2887, opened 2022-04-13
feat: improve resilience on hibernation and network failure

mrekucci:
- don't blocklist peers after waking up from hibernation
- don't blocklist peers as a consequence of a network outage
- keep the known peers list stable during the network outage
- expose network status indicator on the kademlia topology endpoint

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2887)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -74,18 +74,15 @@ require github.com/libp2p/go-libp2p-yamux v0.6.0
 require (
 	github.com/Knetic/govaluate v3.0.1-0.20171022003610-9aa49832a739+incompatible // indirect
 	github.com/StackExchange/wmi v0.0.0-20210224194228-fe8f1750fd46 // indirect
-	github.com/VictoriaMetrics/fastcache v1.6.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cespare/xxhash/v2 v2.1.1 // indirect
 	github.com/cheekybits/genny v1.0.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/davidlazar/go-crypto v0.0.0-20200604182044-b73af7476f6c // indirect
 	github.com/deckarep/golang-set v1.8.0 // indirect
-	github.com/edsrzf/mmap-go v1.0.0 // indirect
 	github.com/flynn/noise v1.0.0 // indirect
 	github.com/francoispqt/gojay v1.2.13 // indirect
 	github.com/fsnotify/fsnotify v1.4.9 // indirect
-	github.com/gballet/go-libpcsclite v0.0.0-20191108122812-4678299bea08 // indirect
 	github.com/go-ole/go-ole v1.2.5 // indirect
 	github.com/go-stack/stack v1.8.0 // indirect
 	github.com/go-task/slim-sprig v0.0.0-20210107165309-348f09dbbbc0 // indirect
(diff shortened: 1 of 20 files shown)


notanatol (approved):

janos (approved):

acud (changes requested):
Nice and thorough! I have a small nitpick and another concern about online-ness assumptions.

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2887#-:-N-dmsTyE6WPTqZkxVTG:b-64bid3)* status: all files reviewed, 2 unresolved discussions (waiting on @mrekucci and @ralph-pichler)

___
*[pkg/p2p/libp2p/libp2p.go, line 604 at r1](https://reviewable.io/reviews/ethersphere/bee/2887#-N-dlUsc4nkV4wCyk3Q4:-N-dlUsc4nkV4wCyk3Q5:b-7h3xko) ([raw file](https://github.com/ethersphere/bee/blob/b1e4d03206469342a98148c1910d9b89b9b3f541/pkg/p2p/libp2p/libp2p.go#L604)):*
> ```Go
> func (s *Service) Blocklist(overlay swarm.Address, duration time.Duration, reason string) error {
> 	if s.NetworkStatus() != p2p.NetworkStatusAvailable {
> 		return nil
> ```

nit, but I'm not sure that silencing the error here is the preferred way to go. Perhaps returning an explicit error that an attempt to blocklist has been made while offline and allowing it to bubble and logging wherever necessary could be preferable, even if for the sake of debugging corner cases. Up to you...
___
*[pkg/p2p/libp2p/libp2p.go, line 1004 at r1](https://reviewable.io/reviews/ethersphere/bee/2887#-N-dmDiN0cTe-NLxnUfu:-N-dmDiN0cTe-NLxnUfv:b-dwrss5) ([raw file](https://github.com/ethersphere/bee/blob/b1e4d03206469342a98148c1910d9b89b9b3f541/pkg/p2p/libp2p/libp2p.go#L1004)):*
> ```Go
> 			err = p2p.ErrNetworkUnavailable
> 		}
> 	case isNetworkOrHostUnreachableError(err):
> ```

careful... will this be also triggered if we are online and the target peer is offline? a boolean might not shine here very well... that's why libp2p's autonat uses a scoring system that adds and subtracts points from a counter in order to achieve more certainty about a node's reachability status

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2887)* status: all files reviewed, 1 unresolved discussion (waiting on @acud and @ralph-pichler)

___
*[pkg/p2p/libp2p/libp2p.go, line 604 at r1](https://reviewable.io/reviews/ethersphere/bee/2887#-N-dlUsc4nkV4wCyk3Q4:-N01E7erEUdwYG_GI98M:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/b1e4d03206469342a98148c1910d9b89b9b3f541/pkg/p2p/libp2p/libp2p.go#L604)):*

nit, but I'm not sure that silencing the error here is the preferred way to go. Perhaps returning an explicit error that an attempt to blocklist has been made while offline and allowing it to bubble and logging wherever necessary could be preferable, even if for the sake of debugging corner cases. Up to you...
</blockquote></details>

Done.
___
*[pkg/p2p/libp2p/libp2p.go, line 1004 at r1](https://reviewable.io/reviews/ethersphere/bee/2887#-N-dmDiN0cTe-NLxnUfu:-N01ELpsBSw3z2Qh2zSB:b-j1b0vz) ([raw file](https://github.com/ethersphere/bee/blob/b1e4d03206469342a98148c1910d9b89b9b3f541/pkg/p2p/libp2p/libp2p.go#L1004)):*

careful... will this be also triggered if we are online and the target peer is offline? a boolean might not shine here very well... that's why libp2p's autonat uses a scoring system that adds and subtracts points from a counter in order to achieve more certainty about a node's reachability status
</blockquote></details>

This function specifically checks for errors that are raised by the underlying operating system when a network failure occurs. There are two specific cases captured:
- when there is no network connection available on your local machine
- when your local network is not connected to the Internet

As for the scoring system, we already have something similar in the place, see `internal/breaker`. I originally started tuning this component, but it wasn't fast enough to reliably detect a network outage. I think as a first iteration this is good enough and we can iterate on it if new cases come up.

acud (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2887)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ralph-pichler)

metacertain (approved):

Issue #2713, opened 2021-12-03
DELETE /pins/{reference} Fails with 500 Internal Server Error


ldeffenb:
### Context 
Old bee nodes that have been running both in mainnet and testnet for a long time having supported a large volume of uploads over time.  See also #2712 

### Summary
When trying to use DELETE /pins/{reference} where the references are retrieved with a GET /pins query, many of them are failing with 500 Internal Server Error and the pin is not deleted.

### Expected behavior
If I ask to delete a pin, I would expect it to be deleted.

### Actual behavior
The pin is not deleted and an error is returned.  The --verbosity 5 logs show that the traversal of the specified reference is failing because some secondary chunk below the reference is not found.  Given that the pin exists, one would have hoped that all of the chunks for that reference would also exist, except for #2712 

### Steps to reproduce
Again, difficult to reproduce, but if you can create an existing pin for which not all of the traversal chunks are actually pinned and are now not retrievable from the swarm, then you will be unable to delete that pin due to traversal failure.

### Possible solution
I'm planning to disable the traversal in DeletePin since my pins were all created with /bytes uploads with Swarm-Pin specified which, as #2712 describes, actually only creates the statestore key and doesn't actually pin anything in localstore.  This will hopefully get my nodes cleaned up, but is not a general solution.

asabya:
Facing the same issue on `v1.4.3-1213e063`

I'm calling "POST" on `/bytes` with `pin` and `encrypt` headers `true`. After getting the reference, if I call "DELETE" on `/pins/<encrypted_reference>` I get `"Internal Server Error"` with `500`

The file size is  ~190kb

logs
```
DEBU[2022-03-11T17:39:32+05:30] unpin root hash: deletion of pin for "68fd6487cbaf2a08d9436256646359319a02d25c8fcd46604a654813384b25b808fbe2398155b0b3aae32b4c425e513a6c96bd74f35a008482f41a3e93dbf711" failed: 3 errors occurred:
	* traversal iteration failed
	* get value: leveldb: not found
[...]

PR #2896, opened 2022-04-21
fix: unpinning of encrypted references

acud:
### Description
Fixes a bug where unpinning of encrypted content resulted in an error, this is due to the traversal service reporting long references that include the decryption key, which in turn resulted in the localstore trying to unpin an address that doesn't exist. This is due to the fact that localstore does not perform a bounds check on the input of `swarm.Address` and the fact that `swarm.Address` is an unbounded type that does not enforce input validation on the size of the address/reference. Therefore, the localstore was looking for a chunk with 64 byte long address which never existed in the pin index. 

I conducted some tests to verify that no chunks are stored with 64 byte long addresses on `Put` but it seems that this doesn't happen. Plus, the `cac` and `soc` validators make sure that the address length is kept to 32 bytes on chunks that come from the network, so there's no risk of this sort of leakage from the network. 

The solution is some middle ground that tries to avoid further complexity in the traversal service and instead just makes sure that localstore receives the correct input. Further improvements on the output of the iterate addresses functionality of the joiner has been also added.

Also, some error handling in the traversal service was improved to provide more informative errors in addition to some minor refinements and error decorating in localstore.

### Related Issue (Optional)
related to https://github.com/ethersphere/bee/issues/2713 but does not solve the problem of `HTTP 500` on unencrypted content.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2896)

### DIFF
--- a/pkg/file/joiner/joiner.go
+++ b/pkg/file/joiner/joiner.go
@@ -13,6 +13,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/ethersphere/bee/pkg/encryption"
 	"github.com/ethersphere/bee/pkg/encryption/store"
 	"github.com/ethersphere/bee/pkg/file"
 	"github.com/ethersphere/bee/pkg/storage"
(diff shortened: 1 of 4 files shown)


mrekucci (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2896)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud)

notanatol (approved):

PR #2900, opened 2022-04-22
feat: wallet endpoint

istae:
### Checklist

Untitled

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2900)


pkg/debugapi/debugapi_test.go
```diff
@@ -162,7 +167,7 @@ func TestServer_Configure(t *testing.T) {
 	transaction := transactionmock.New(o.TransactionOpts...)
 	gatewayMode := false
 	beeMode := debugapi.FullMode
-	s := debugapi.New(o.PublicKey, o.PSSPublicKey, o.EthereumAddress, logging.New(io.Discard, 0), nil, nil, big.NewInt(2), transaction, nil, false, nil, gatewayMode, beeMode)
+	s := debugapi.New(o.PublicKey, o.PSSPublicKey, o.EthereumAddress, logging.New(io.Discard, 0), nil, nil, big.NewInt(2), transaction, nil, false, nil, gatewayMode, beeMode, 1)
```
notanatol, line 170: I think here we can use `api.FullMode`

notanatol (approved):

notanatol:
would be good to add this endpoint to `pkg/auth/auth.go` list of endpoints, with its according role, which I guess would be `accountant`

pkg/debugapi/debugapi_test.go
istae: the `1` is chainID

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2900)* status: all files reviewed, 1 unresolved discussion (waiting on @notanatol)

___
*[`pkg/node/devnode.go` line 388 at r1](https://reviewable.io/reviews/ethersphere/bee/2900#-N0a_HHj5R-YT1auDTcL:-N0a_HHj5R-YT1auDTcM:b-f3345t) ([raw file](https://github.com/ethersphere/bee/blob/a1e6b7e0fc2111acdd0932dbcb3b6f99ea9e6562/pkg/node/devnode.go#L388)):*
> ```Go
> 
> 		// inject dependencies and configure full debug api http path routes
> 		debugAPIService.Configure(swarmAddress, p2ps, pingPong, kad, lightNodes, storer, tagService, acc, pseudoset, true, true, mockSwap, mockChequebook, batchStore, post, postageContract, traversalService, nil)
> ```

The `/wallet` endpoint will panic in the `dev` mode, maybe consider adding a mock instead of nil.

notanatol (approved):

pkg/debugapi/wallet.go
```diff
@@ -14,10 +14,10 @@ import (
 )
 
 type walletResponse struct {
-	BZZ             float64        `json:"bzz"`
-	XDai            float64        `json:"xDai"`
-	ChainID         int64          `json:"chainID"`
-	ContractAddress common.Address `json:"contractAddress"`
+	BZZ             float64        `json:"bzz"`             // the BZZ balance of the wallet associated with the eth address of the node
```
agazso, line 17: In general it's not a good idea to use `float` for representing monetary values, because at small amounts it starts to lose precision. There are some languages which has `bigint` or `decimal` types for storing that, unfortunately in JSON there is no such types.

Because of this previously with the `/balance` endpoint we used `string` type to be able encode the value of the internal `bigint` type for expressing that in a lossless way: https://docs.ethswarm.org/debug-api/#tag/Balance/paths/~1balances/get

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2900)* status: all files reviewed, 2 unresolved discussions (waiting on @agazso and @notanatol)

___
*[`pkg/debugapi/wallet.go` line 17 at r2](https://reviewable.io/reviews/ethersphere/bee/2900#-N0Ge2Jygx46ckvaMXC8-r2-17:-N0epcoi7_tdJYuS8srp:biwf4rc) ([raw file](https://github.com/ethersphere/bee/blob/ebf5e20d786e92d151db81268b3695eeed2a0f20/pkg/debugapi/wallet.go#L17)):*
<details><summary><i>Previously, agazso (Attila Gazso) wrote…</i></summary><blockquote>

In general it's not a good idea to use `float` for representing monetary values, because at small amounts it starts to lose precision. There are some languages which has `bigint` or `decimal` types for storing that, unfortunately in JSON there is no such types.

Because of this previously with the `/balance` endpoint we used `string` type to be able encode the value of the internal `bigint` type for expressing that in a lossless way: https://docs.ethswarm.org/debug-api/#tag/Balance/paths/~1balances/get
</blockquote></details>

Agree with @agazso here.

notanatol (approved):

pkg/debugapi/wallet.go
istae: This raises another question: we can return the balance with either big or small units, meaning: if the balance is 17 BZZ, we can return 17 or 170000000000000000 (which is what the erc20 service returns). I've already wrote the logic for converting the small units balance for better UX.

ldeffenb: I would put my vote in for consistency and always return amounts as integers (or strings) relative to the smallest possible unit of measurement.  In other words, mirror what the erc20 and other blockchain APIs do.

mrekucci (approved):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2900)* status: all files reviewed, 1 unresolved discussion (waiting on @agazso and @ldeffenb)

pkg/debugapi/wallet.go
AuHau: Agree with @ldeffenb here.

acud (approved):

AuHau:
Just for future reference `contractAddress` naming does not really convey what address it points to. Something like `chequebookContractAddress` would have been better ;-)

istae:
we just released this feature, we can sneak in an change if you'd like. PR's welcome!

PR #2910, opened 2022-04-27
fix: node start up to continue on bootstrapper error

istae:
### Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2910)

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -481,7 +481,7 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
 		)
 		logger.Infof("bootstrapper took %s", time.Since(start))
 		if err != nil {
-			logger.Errorf("bootstrapper failed to fetch batch state: %w", err)
+			logger.Errorf("bootstrapper failed to fetch batch state: %v", err)
 		}
 	}
 
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -479,9 +479,9 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
 			libp2pPrivateKey,
 			o,
 		)
-		logger.Infof("bootstrapper done, took %s", time.Since(start))
+		logger.Infof("bootstrapper took %s", time.Since(start))
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/node/node.go
```diff
@@ -479,9 +479,9 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
 			libp2pPrivateKey,
 			o,
 		)
-		logger.Infof("bootstrapper done, took %s", time.Since(start))
+		logger.Infof("bootstrapper took %s", time.Since(start))
 		if err != nil {
-			return nil, fmt.Errorf("bootstrapper failed to fetch batch state: %w", err)
+			logger.Errorf("bootstrapper failed to fetch batch state: %w", err)
```
mrekucci, line 484: Replace %w` with `%v`.

mrekucci (approved):

acud (approved):

pkg/node/node.go
```diff
@@ -479,9 +479,9 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
 			libp2pPrivateKey,
 			o,
 		)
-		logger.Infof("bootstrapper done, took %s", time.Since(start))
+		logger.Infof("bootstrapper took %s", time.Since(start))
```
ldeffenb, line 482: Should this be bootstrapper or BatchState initialization?

pkg/node/node.go
```diff
@@ -479,9 +479,9 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
...
 		if err != nil {
-			return nil, fmt.Errorf("bootstrapper failed to fetch batch state: %w", err)
+			logger.Errorf("bootstrapper failed to fetch batch state: %v", err)
```
ldeffenb, line 484: Ditto here.

ldeffenb (changes requested):
Questioning "bootstrapper" as being incorrect/misleading?

pkg/node/node.go
agazso: I tend to agree that `bootstrapper` is hard to understand without context. My suggestion would be `postage snapshot syncing took %` because that would be consistent how the feature is called in the config (`use-postage-snapshot`).

PR #2914, opened 2022-04-28
fix: postage service initialization bug

istae:
### Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2914)

### DIFF
--- a/pkg/postage/service.go
+++ b/pkg/postage/service.go
@@ -57,21 +57,17 @@ func NewService(store storage.StateStorer, postageStore Storer, chainID int64) (
 		chainID:      chainID,
 	}
 
-	n := 0
-	if err := s.store.Iterate(s.key(), func(_, _ []byte) (stop bool, err error) {
-		n++
+	if err := s.store.Iterate(s.key(), func(_, value []byte) (bool, error) {
+		st := &StampIssuer{}
+		if err := st.UnmarshalBinary(value); err != nil {
+			return false, err
+		}
+		_ = s.add(st)
 		return false, nil
 	}); err != nil {
 		return nil, err
 	}
-	for i := 0; i < n; i++ {
-		st := &StampIssuer{}
-		err := s.store.Get(s.keyForIndex(i), st)
-		if err != nil {
-			return nil, err
-		}
-		_ = s.add(st)
-	}
+
 	return s, nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/postage/service.go
```diff
@@ -196,11 +192,11 @@ func (ps *service) Close() error {
 
 // keyForIndex returns the statestore key for an issuer
 func (ps *service) keyForIndex(i int) string {
-	return fmt.Sprintf(ps.key()+"%d", i)
+	return fmt.Sprintf(ps.key()+"_%d", i)
```
acud, line 195: not sure why this is needed... it means we make people lose the batch utilization and collision bucket data and it would be good to avoid this if possible. can we not resolve the problem without changing the keys?

ldeffenb: Ouch!  I really really like tracking my major stamp utilization and have jumped through some pretty big hoops to avoid losing it.  I'd really hate for something like this to just sneak into an updated version.

istae: Since the new release will delete the keys anyways, I thought I would sneak in a change for how the keys look just as a visualization. It's not important and can be removed

ldeffenb: > Since the new release will delete the keys anyways, ...

So all of our in-use keys (like the following 23.71% utilized) will go back to being 0% utilized at the source, but may become over-utilized in remote nodes that are already storing chunks using that stamp?  Not very good practice for backward-compatibility.  Any chance of a migration instead of an outright delete?

      "batchID": "0e8366a6fdac185b6f0327dc89af99e67d9d3b3f2af22432542dc5971065c1df",
      "utilization": 3884,
      "usable": true,
      "label": "20210916-OSM-Map",
      "depth": 30,
      "amount": "441429710",
      "bucketDepth": 16,
      "blockNumber": 18125458,
      "immutableFlag": false,
      "exists": true,
      "batchTTL": 525217652

acud: > Since the new release will delete the keys anyways

initially it was not supposed to delete the issuer keys, only the batchstore keys

istae: @ldeffenb No data will be lost, with this release, batches will be resynced at startup, so issuers will be restored as well.

acud (approved):

PR #2929, opened 2022-05-04
fix: add ignore radius flag

notanatol:
### Checklist

### Description
In chain-less mode the topology depth is not updated, staying at 0.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2929)

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -636,7 +636,7 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
 	}
 
 	kad, err := kademlia.New(swarmAddress, addressbook, hive, p2ps, pingPong, metricsDB, logger,
-		kademlia.Options{Bootnodes: bootnodes, BootnodeMode: o.BootnodeMode, StaticNodes: o.StaticNodes, IgnoreRadius: o.ChainEnable})
+		kademlia.Options{Bootnodes: bootnodes, BootnodeMode: o.BootnodeMode, StaticNodes: o.StaticNodes, IgnoreRadius: !o.ChainEnable})
 	if err != nil {
 		return nil, fmt.Errorf("unable to create kademlia: %w", err)
 	}
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -636,7 +636,7 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
 	}
 
 	kad, err := kademlia.New(swarmAddress, addressbook, hive, p2ps, pingPong, metricsDB, logger,
-		kademlia.Options{Bootnodes: bootnodes, BootnodeMode: o.BootnodeMode, StaticNodes: o.StaticNodes})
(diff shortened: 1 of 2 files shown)


pkg/node/node.go
```diff
@@ -636,7 +636,7 @@ func NewBee(addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, netwo
 	}
 
 	kad, err := kademlia.New(swarmAddress, addressbook, hive, p2ps, pingPong, metricsDB, logger,
-		kademlia.Options{Bootnodes: bootnodes, BootnodeMode: o.BootnodeMode, StaticNodes: o.StaticNodes})
+		kademlia.Options{Bootnodes: bootnodes, BootnodeMode: o.BootnodeMode, StaticNodes: o.StaticNodes, IgnoreRadius: o.ChainEnable})
```
acud, line 639: isn't it the other way?
- if we have chain enabled we don't want to ignore the radius since the postage data is going to be there
- if we have the chain disabled then we want to ignore the radius

notanatol: yes, it seems I flipped it the other way around, fixing...

vojtechsimetka:
Just tested and it solves the problem for me:

```JSON
"population":2526,
"connected":108,
"timestamp":"2022-05-04T14:20:06.923154+02:00",
"nnLowWatermark":3,
"depth":8,
...
```

Here is the config just for a reference
```yaml
api-addr: 127.0.0.1:1633
debug-api-addr: 127.0.0.1:1635
debug-api-enable: true
password: <censored>
swap-enable: false
swap-initial-deposit: 0
mainnet: true
full-node: false
chain-enable: false
cors-allowed-origins: '*'
use-postage-snapshot: true
data-dir: <censored>
transaction: 0x01208d6ba7ab04515219befb64e286d13efbd7fa6483892a6f26e9c1de5157cb
block-hash: ae5e61f258e8420a1d3c5e75dd82c0f1c86b51a2986e6c49ba457ed592c34d0b% 
```

mrekucci (approved):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2929)* status: all files reviewed, 2 unresolved discussions (waiting on @acud and @notanatol)

___
*[`pkg/topology/kademlia/kademlia.go` line 845 at r2](https://reviewable.io/reviews/ethersphere/bee/2929#-N1E8FyS7BHgy3TiVghc:-N1E8FyS7BHgy3TiVghd:b-k9s08t) ([raw file](https://github.com/ethersphere/bee/blob/cf9a4b7223e590a636252723af78eae9f429e37d/pkg/topology/kademlia/kademlia.go#L845)):*
> ```Go
> 
> // recalcDepth calculates and returns the kademlia depth.
> func (k *Kad) recalcDepth(peers *pslice.PSlice, radius uint8, filter peerFilterFunc) uint8 {
> ```

Does it need the input parameters when it was turned into a method?

vojtechsimetka:
Checked the latest version [cf9a4b7](https://github.com/ethersphere/bee/pull/2929/commits/cf9a4b7223e590a636252723af78eae9f429e37d) and it also works

Issue #2909, opened 2022-04-27
Bee API stops to return data on GET requests without evident reason


tmm360:
### Context
Using Bee 1.5.1 with Docker.

### Summary
I've a node that at some point started to timeout get requests, even if content should be cached.

This is an extract of logs:
```
time="2022-04-27T15:59:58Z" level=info msg="api access" duration=0.013540011 ip=10.0.8.247 method=GET proto=HTTP/1.0 referrer="https://app.etherna.io/" size=96387 status=200 uri=/bzz/a1384bd71fcecf3a706d777075b56e3d6063dd876bba5617c985ba4eff3311e3/ user-agent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.127 Safari/537.36" x-forwarded-for="185.159.157.9, 151.139.95.145, 10.0.0.74"
time="2022-04-27T16:00:08Z" level=info msg="api access" duration=1.003372393 ip=10.0.8.247 method=GET proto=HTTP/1.0 referrer="https://app.etherna.io/" size=93 status=200 uri="/feeds/F60Fbc6A72dde156f24A08C414197AE3CF5399E0/40daa16a08bb37fc91dcfb3402af8193e327e741209d6f86a8127cbd4b1befaa?type=sequence" user-agent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.127 Safari/537.36" x-forwarded-for="185.159.157.9, 151.139.35.201, 10.0.0.74"
time="2022-04-27T16:00:09Z" level=info msg="api access" duration=0.01309673 ip=10.0.8.11 method=GET proto=HTTP/1.0 referrer="https://app.etherna.io/" size=96387 status=200 uri=/bzz/a1384bd71fcecf3a706d777075b56e3d6063dd876bba5617c985ba4eff3311e3/ user-agent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.127 Safari/537.36" x-forwarded-for="185.159.157.9, 151.139.95.145, 10.0.0.73"
[...]

PR #2932, opened 2022-05-05
fix: sharky read deadlock on context cancelled

acud:
### Description

Fixes a bug where a cancelled context would result in a deadlock on the sharky read which tries to write the error back to the `sh.errc` channel

Caught in a goroutine dump from @tmm360 on #2909.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2932)

### DIFF
--- a/pkg/sharky/shard.go
+++ b/pkg/sharky/shard.go
@@ -73,6 +73,7 @@ type entry struct {
 
 // read models the input to read operation (the output is an error)
 type read struct {
+	ctx  context.Context
 	buf  []byte // variable size read buffer
 	slot uint32 // slot to read from
 }
@@ -114,6 +115,7 @@ LOOP:
 			for {
 				select {
 				case sh.errc <- sh.read(op):
+				case <-op.ctx.Done():
 				case <-sh.quit:
 					return
 				}
--- a/pkg/sharky/shard_slots_test.go
+++ b/pkg/sharky/shard_slots_test.go
@@ -82,7 +82,7 @@ func readFromLocation(t *testing.T, shard *shard, loc Location) []byte {
 	buf := make([]byte, loc.Length)
 
 	select {
-	case shard.reads <- read{buf[:loc.Length], loc.Slot}:
+	case shard.reads <- read{ctx: context.Background(), buf: buf[:loc.Length], slot: loc.Slot}:
 		if err := <-shard.errc; err != nil {
 			t.Fatal("read", err)
 		}
--- a/pkg/sharky/store.go
+++ b/pkg/sharky/store.go
@@ -119,7 +119,7 @@ func (s *Store) create(index uint8, maxDataSize int, basedir fs.FS) (*shard, err
 func (s *Store) Read(ctx context.Context, loc Location, buf []byte) (err error) {
 	sh := s.shards[loc.Shard]
 	select {
-	case sh.reads <- read{buf[:loc.Length], loc.Slot}:
+	case sh.reads <- read{ctx: ctx, buf: buf[:loc.Length], slot: loc.Slot}:
 		s.metrics.TotalReadCalls.Inc()
 	case <-ctx.Done():
 		return ctx.Err()
@@ -133,8 +133,6 @@ func (s *Store) Read(ctx context.Context, loc Location, buf []byte) (err error) 
 		return err
 	case <-s.quit:
 		return ErrQuitting
-	case <-ctx.Done():
-		return ctx.Err()
 	}
 }
 
--- a/pkg/sharky/store.go
+++ b/pkg/sharky/store.go
@@ -125,6 +125,10 @@ func (s *Store) Read(ctx context.Context, loc Location, buf []byte) (err error) 
 		return ctx.Err()
 	}
 
+	// it is important that this select would NEVER respect the context
+	// cancellation. this would result in a deadlock on the shard, since
+	// the result of the operation must be drained from errc, allowing the
+	// shard to be able to handle new operations (#2932).
 	select {
 	case err = <-sh.errc:
 		if err != nil {


notanatol (approved):

janos (approved):
Quite interesting. 👏

pkg/sharky/shard.go
```diff
@@ -114,6 +115,7 @@ LOOP:
 			for {
 				select {
 				case sh.errc <- sh.read(op):
```
acud, line 117: @janos I am puzzled by how this specific syntax usage would unfold. So I assume that if there is a reader on `errc` on the other side the specific case will be selected, blocking the possible selects on the other cases?
Or does `sh.read()` have to first return for the case to actually be selected?
I am concerned that `sh.read` might execute halfway while the context expires, resulting in the shard again not being able to drain the channel since nothing is written to it.

acud: ok question answered here: https://go.dev/play/p/-8VDC_2ACD7
it is enough to just have the reader on the channel, and the case will already be selected

acud: In which case it might make sense to even get rid of the context altogether and just not respect it when we get to the actual read part (we could respect it until the task is handed over to the shard, but not after).
Would be happy for some thoughts on this.

janos: That is right. The read function will not block the select case, as the channel will be written to only after the function has returned.

Offloading cancelation to the Read caller, but I think that currently, the already existing channels handling is convenient to be used for particular reads cancelation as well. As file operations could not be canceled with context, there is always a possibility to have some goroutines blocking.

janos: I see context here more like an insurance that cancelation will happen regardless if some code further away is changed. Code always can be more readable and improvements in that way here may result in better understanding of the flow. Somehow, I tend to keep operations on a single channel as close in the code as possible, preferably within one function with clear read only channels returned, if needed. If channels are more sparse in the code, it is a bit harder to follow.

acud: :+1: I added a small comment about the change in `store.go` just so that there's a breadcrumb trail in the code, and I shall keep the change as-is. Thanks for the feedback

acud:
so as a result of the discussion, I've had a look into the part of sharky that does the writes and it seems that there's a similar problem there too (it does not result in a deadlock fortunately due to channel buffering):
```go
// Write stores a new blob and returns its location to be used as a reference
// It can be given to a Read call to return the stored blob.
func (s *Store) Write(ctx context.Context, data []byte) (loc Location, err error) {
	if len(data) > s.maxDataSize {
		return loc, ErrTooLong
	}
	s.wg.Add(1)
	defer s.wg.Done()

	c := make(chan entry, 1) // buffer the channel to avoid blocking in shard.process on quit or context done

	select {
	case s.writes <- write{data, c}:
		s.metrics.TotalWriteCalls.Inc()
	case <-s.quit:
		return loc, ErrQuitting
	case <-ctx.Done():
		return loc, ctx.Err()
	}

	select {
	case e := <-c:
		if e.err == nil {
			shard := strconv.Itoa(int(e.loc.Shard))
			s.metrics.CurrentShardSize.WithLabelValues(shard).Inc()
			s.metrics.ShardFragmentation.WithLabelValues(shard).Add(float64(s.maxDataSize - int(e.loc.Length)))
		} else {
			s.metrics.TotalWriteCallsErr.Inc()
		}
		return e.loc, e.err
	case <-s.quit:
		return loc, ErrQuitting
	case <-ctx.Done():
		return loc, ctx.Err() <-- if context is respected here, the entry will be written to sharky, and since the result channel is buffered, it basically means that there's a stale entry in the db that will never be used (the reader never reads the result, and the calling context returns, allowing the chunk to stay in the db indefinitely (or until a recovery happens)). The previous `case` of `ErrQuitting` can actually result in the same problem on shutdown.
	}
}
```

I would handle this in a separate PR, but the gist is to remove the channel buffering on the result and not to respect the context cancellation on this scope for the same reasons as above. We could add the context to the write op struct and respect it by adding a select in `shard.go`:
```go
		case op := <-writes:
			op.res <- sh.write(op.buf, slot) <-- can respect context here with a select
			free = sh.slots.out // reenable popping a free slot next time we can write
			writes = nil        // disable popping a write operation until there is a free slot
```

Feedback welcome

janos:
> so as a result of the discussion, I've had a look into the part of sharky that does the writes and it seems that there's a similar problem there too (it does not result in a deadlock fortunately due to channel buffering):
> 
> ```go
> // Write stores a new blob and returns its location to be used as a reference
> // It can be given to a Read call to return the stored blob.
> func (s *Store) Write(ctx context.Context, data []byte) (loc Location, err error) {
> 	if len(data) > s.maxDataSize {
> 		return loc, ErrTooLong
> 	}
> 	s.wg.Add(1)
> 	defer s.wg.Done()
> 
> 	c := make(chan entry, 1) // buffer the channel to avoid blocking in shard.process on quit or context done
> 
> 	select {
> 	case s.writes <- write{data, c}:
> 		s.metrics.TotalWriteCalls.Inc()
> 	case <-s.quit:
> 		return loc, ErrQuitting
> 	case <-ctx.Done():
> 		return loc, ctx.Err()
> 	}
> 
> 	select {
> 	case e := <-c:
> 		if e.err == nil {
> 			shard := strconv.Itoa(int(e.loc.Shard))
> 			s.metrics.CurrentShardSize.WithLabelValues(shard).Inc()
> 			s.metrics.ShardFragmentation.WithLabelValues(shard).Add(float64(s.maxDataSize - int(e.loc.Length)))
> 		} else {
> 			s.metrics.TotalWriteCallsErr.Inc()
> 		}
> 		return e.loc, e.err
> 	case <-s.quit:
> 		return loc, ErrQuitting
> 	case <-ctx.Done():
> 		return loc, ctx.Err() <-- if context is respected here, the entry will be written to sharky, and since the result channel is buffered, it basically means that there's a stale entry in the db that will never be used (the reader never reads the result, and the calling context returns, allowing the chunk to stay in the db indefinitely (or until a recovery happens)). The previous `case` of `ErrQuitting` can actually result in the same problem on shutdown.
> 	}
> }
> ```
> 
> I would handle this in a separate PR, but the gist is to remove the channel buffering on the result and not to respect the context cancellation on this scope for the same reasons as above. We could add the context to the write op struct and respect it by adding a select in `shard.go`:
> 
> ```go
> 		case op := <-writes:
> 			op.res <- sh.write(op.buf, slot) <-- can respect context here with a select
> 			free = sh.slots.out // reenable popping a free slot next time we can write
> 			writes = nil        // disable popping a write operation until there is a free slot
> ```
> 
> Feedback welcome

Correct. The buffer of 1 to channel was added exactly to avoid a deadlock as reported by one of the users and validated by profiling from him. I am completely for a more clear solution that does not require buffering and improves context cancelation.

aloknerurkar (approved):

PR #2936, opened 2022-05-08
feat: generate and verify inclusion proofs for chunks

aloknerurkar:
### Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2936)

### DIFF
--- a/pkg/bmt/inclusionproof.go
+++ b/pkg/bmt/inclusionproof.go
@@ -1,0 +1,103 @@
+package bmt
+
+import "fmt"
+
+const (
+	DefaultMaxPayload = 4096
+	SegmentSize       = 32
+	SegmentPairSize   = 2 * SegmentSize
+)
+
+func constructBMT(payload []byte) ([][]byte, error) {
+	if len(payload) > DefaultMaxPayload {
+		return nil, fmt.Errorf("invalid payload length %d", len(payload))
+	}
+
+	// segments contains the data segments at each level of the BMT as we build it up
+	// if we have less the 4096 bytes, we will pad the payload with zero bytes
+	segments := append(payload, make([]byte, DefaultMaxPayload-len(payload))...)
+	// levels will contain each level of the BMT as a byte array concatenated with all
+	// the segments of that level up to the root hash of the BMT
+	var levels [][]byte
+	for len(segments) != SegmentSize {
+		levels = append(levels, segments)
+		// next will contain the segments obtained by hashing the contents of the
+		// previous level. This will be the next level for the calculation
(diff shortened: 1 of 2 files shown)


pkg/bmt/inclusionproof_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"time"
+
+	"github.com/ethersphere/bee/pkg/bmt"
+	"github.com/ethersphere/bee/pkg/bmt/reference"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+func TestInclusionProofs(t *testing.T) {
+
+	testDatas := []struct {
```
acud, line 18: i believe that due to the importance of this feature, it would be prudent to have a manually calculated test that checks every byte of the proof is the way it should be. i.e. generating something and then silently verifying it goes only half-way. i'd expect to have tests that cover:
- what happens when the proof is on the left end of the tree
- what happens when its on the right
- middle
- encrypted chunks work correctly
- ideally you'll also calculate "by-hand" the tree with `keccak()` to see that the results add up

pkg/bmt/inclusionproof.go
```diff
@@ -0,0 +1,134 @@
+package bmt
```
mrekucci, line 1: Missing header.

pkg/bmt/inclusionproof_test.go
aloknerurkar: So I added the cases now with hard-coded values which were calculated manually.

I am not sure if we need to add separate case for encrypted data though. The hasher doesnt seem to care if the payload is encrypted. We already have tests with random data below. So I think we should be covered now.

zelig:
@aloknerurkar 
here is a rather different approach:
https://github.com/ethersphere/bee/blob/SIP-BMT/pkg/bmt/proof_test.go
https://github.com/ethersphere/bee/blob/SIP-BMT/pkg/bmt/proof.go

aloknerurkar:
> @aloknerurkar here is a rather different approach: https://github.com/ethersphere/bee/blob/SIP-BMT/pkg/bmt/proof_test.go https://github.com/ethersphere/bee/blob/SIP-BMT/pkg/bmt/proof.go

@zelig This definitely looks better but then it assumes the hasher has already constructed the BMT. I guess I am not yet sure how this will be used in the bigger picture to make a call. Current approach with package level functions allow loose coupling of this functionality with the hasher.

I guess my question is when we are participating in the lottery and we are 'asked' to prove a chunk, how do we construct the hasher BMT?

PR #2941, opened 2022-05-09
chore: bump to Go 1.18

mrekucci:
### Checklist

- [x] I have read the [coding guide](https://github.com/ethersphere/bee/blob/master/CODING.md)
- [ ] My change requires a documentation update and I have done it
- [ ] I have added tests to cover my changes.

### Description
This PR brings the following changes:
- upgrade Golang to version 1.18
- upgrade Golang linter to version 1.46
- upgrade go-libp2p to  version v0.19.1 and its relevant dependencies

### Related Issue (Optional)
In go-libp2p v0.18.0, the resource manager was introduced with [predefined limits](https://github.com/libp2p/go-libp2p-resource-manager#default-limits), and so some limits tweaking might be necessary.

### Screenshots (if appropriate):

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2941)

PR #2942, opened 2022-05-09
fix: autonat dependency

mrekucci:
### Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2942)

### DIFF
--- a/pkg/p2p/libp2p/connections_test.go
+++ b/pkg/p2p/libp2p/connections_test.go
@@ -9,7 +9,6 @@ import (
 	"context"
 	"errors"
 	"io"
-	"math/rand"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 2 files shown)


mrekucci:
> I am not sure if it is safe to just remove MaxIncomingStreams limit. It was put there to protect the node after investigation on large number of streams. Is this handled implicitly in new libp2p versions?

In version #2934, we're switching from v0.16.0 to v0.19.0, and in v0.18.0 [Resource Manager](https://github.com/libp2p/go-libp2p-resource-manager) was introduced with [Default Limits](https://github.com/libp2p/go-libp2p-resource-manager#default-limits). Due to the dynamic nature of the limiter, the test does not add much value and was failing. The resource manager also provides a static implementation of the limiter, so we can switch if this causes any problems.

mrekucci:
Closing in favor of #2941

PR #2944, opened 2022-05-10
fix: shutdown deadlock

### LINKED ISSUE #2926: Terminate process when kademlia shutdown, or other critical issues
#### Summary
Currently process is kept alive, but this doesn't permit to handle the error without an active check.
Please also see this issue: https://github.com/ethersphere/bee/issues/2556

#### Motivation
Container orchestrators can try to restart the container automatically, but it has to terminate first.
Any blocking issue should terminate the process, so IF the process is recoverable with a restart, the orchestrator can perform automatically.
Otherwise, a script have to actively check if node is reachable, and eventually that it isn't try to kill and restart. Not much user friendly, it add complexity, and I don't see rationale under choose to not kill the process.

#### Implementation
Simply kill the process when a critical issue come on an active node.

#### Drawbacks
Don't see any.

### LINKED ISSUE #2931: node shutdown while chain syncing causes deadlock
node will not shut down cleanly. see `pkg/node/node.go#L661`

solution would be to initialize the interrupt channel earlier and pass it to `NewBee` so that we can select on it on the problematic line in `661`:
```go
			// Wait for termination or interrupt signals.
			// We want to clean up things at the end.
			interruptChannel := make(chan os.Signal, 1)
			signal.Notify(interruptChannel, syscall.SIGINT, syscall.SIGTERM)

```

also the returning of the error should preferably also cause a call to `Shutdown` to close up all open resources (specifically local and state stores)

notanatol:
### Checklist

### Description
Closes #2931
Closes #2926 

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2944)


pkg/topology/kademlia/kademlia.go
```diff
@@ -454,7 +455,11 @@ func (k *Kad) manage() {
 	go func() {
 		<-k.quit
 		if !timer.Stop() {
-			<-timer.C
+			select {
+			case <-timer.C:
+			case <-time.After(1 * time.Second):
+				k.logger.Debugf("kademlia timer not drained after 1 second, goroutine dump...\n%s\n*** end\n", goroutine.Dump())
```
acud, line 461: why not use `pprof.Lookup("goroutine").WriteTo(os.Stdout, 1)`?

notanatol: fixed it to use pprof

pkg/node/devnode.go
```diff
@@ -419,6 +419,9 @@ func (b *DevBee) Shutdown(ctx context.Context) error {
 
 	tryClose(b.apiCloser, "api")
 
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
acud, line 422: this was 15s and now is 5s. it needs to be evaluated whether this is the right value, and what are the side effects of a non-clean api shutdown (since this propagates to other components that are later shut-down.

pkg/goroutine/goroutine.go
```diff
@@ -0,0 +1,21 @@
...
+	"io"
+	"runtime"
+	"runtime/pprof"
+)
+
+func Dump(out io.Writer) {
+	_ = pprof.Lookup("goroutine").WriteTo(out, 2)
+}
+
+func Stack() string {
```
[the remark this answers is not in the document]
notanatol: not needed, just syntactically looked nice.

removed

pkg/node/node.go
```diff
@@ -969,6 +980,9 @@ func (b *Bee) Shutdown(ctx context.Context) error {
 
 	tryClose(b.apiCloser, "api")
 
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
acud, line 983: this was 15s and now is 5s. it needs to be evaluated whether this is the right value, and what are the side effects of a non-clean api shutdown (since this propagates to other components that are later shut-down.

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2944)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @acud)

___
*[`pkg/node/node.go` line 673 at r3](https://reviewable.io/reviews/ethersphere/bee/2944#-N1guT2PtB8cF7o-g8Z8-r3-673:-N1qs3_X371Prf17_qeX:b-u0hht) ([raw file](https://github.com/ethersphere/bee/blob/d070c6ac8caa5cd1800b757a701e02275673e0ee/pkg/node/node.go#L673)):*

this is actually an abnormal state but the select would just cause the software to continue booting, and only later shutting down. i think that actually returning an error here and causing the defer statement to initiate the shutdown would be more correct, no?
</blockquote></details>

yes, returning an error now
___
*[`pkg/node/node.go` line 983 at r3](https://reviewable.io/reviews/ethersphere/bee/2944#-N1guT2PtB8cF7o-g8Z8-r3-983:-N1qs6FQAvxIK4EMXF2W:bvth56w) ([raw file](https://github.com/ethersphere/bee/blob/d070c6ac8caa5cd1800b757a701e02275673e0ee/pkg/node/node.go#L983)):*

this was 15s and now is 5s. it needs to be evaluated whether this is the right value, and what are the side effects of a non-clean api shutdown (since this propagates to other components that are later shut-down.
</blockquote></details>

reverted

acud (approved):
> Reviewed 6 of 10 files at r1, 2 of 4 files at r3, 2 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2944)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @notanatol)

acud:
@notanatol can you push another commit so that the static analyzer can run again? I adjusted the rules. Should pass now

notanatol (commented):
done

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2944#-:-N205M6jAKLOfT2wheBW:bb7f2mf)* status: 10 of 11 files reviewed, all discussions resolved (waiting on @acud and @mrekucci)

notanatol (commented):
> Reviewed 6 of 10 files at r1, 1 of 4 files at r3, 1 of 2 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2944)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mrekucci)

acud (approved):
looks great, thanks

PR #2951, opened 2022-05-12
fix: allow nested paths for multipart upload

aloknerurkar:
### Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2951)

### DIFF
--- a/pkg/api/dirs.go
+++ b/pkg/api/dirs.go
@@ -305,13 +305,15 @@ func (m *multipartReader) Next() (*FileInfo, error) {
 		return nil, err
 	}
 
-	fileName := part.FileName()
-	if fileName == "" {
-		fileName = part.FormName()
+	filePath := part.FileName()
+	if filePath == "" {
+		filePath = part.FormName()
 	}
-	if fileName == "" {
-		return nil, errors.New("filename missing")
+	if filePath == "" {
+		return nil, errors.New("filepath missing")
 	}
+
+	fileName := filepath.Base(filePath)
 
 	contentType := part.Header.Get(contentTypeHeader)
 	if contentType == "" {
@@ -327,12 +329,8 @@ func (m *multipartReader) Next() (*FileInfo, error) {
 		return nil, errors.New("invalid file size")
 	}
 
-	if filepath.Dir(fileName) != "." {
-		return nil, errors.New("multipart upload supports only single directory")
-	}
-
 	return &FileInfo{
-		Path:        fileName,
+		Path:        filePath,
 		Name:        fileName,
 		ContentType: contentType,
 		Size:        fileSize,
--- a/pkg/api/dirs_test.go
+++ b/pkg/api/dirs_test.go
@@ -134,6 +134,7 @@ func TestDirs(t *testing.T) {
 		},
 		{
 			name:              "nested files with extension",
+			doMultipart:       true,
 			expectedReference: swarm.MustParseHexAddress("4c9c76d63856102e54092c38a7cd227d769752d768b7adc8c3542e3dd9fcf295"),
 			files: []f{
 				{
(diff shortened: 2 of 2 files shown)


notanatol (approved):

AuHau:
This LGTM, but if I correctly recall, @acud was concerned about making sure this works correctly, so maybe add some tests that the nested paths are correctly handled?

aloknerurkar:
> This LGTM, but if I correctly recall, @acud was concerned about making sure this works correctly, so maybe add some tests that the nested paths are correctly handled?

So the tests are enabled in this PR. The tests upload using both tar and multipart once the `doMultipart` is set.

The problem was w.r.t tar, by virtue of the tar format being correct we know that the directory structure is valid. In the case of multipart, users can specify incorrect paths and there is no way to actually verify this. With this PR, if used correctly the feature will be enabled. If this assumption is okay we dont need any more changes I believe.

AuHau:
Ok sounds good to me. 

Regarding bad paths. To my understanding, we support UNIX-like paths in Bee, right? If so, then I think there are a few checks that can be done here, which in the `tar` upload use-case are most probably handled by the tar implementation itself, but with multipart are not treated. Things like:

 - calling [`path.Clean`](https://pkg.go.dev/path#Clean) on the path
 - verification if it is a relative or absolute path - what should be the required format?
 - verification if the path does not contain invalid characters - see for example [here](https://www.cyberciti.biz/faq/linuxunix-rules-for-naming-file-and-directory-names/)
 - verification if the folder/file names are not longer than 255 characters 

And maybe more, these are just from top of my head.

mrekucci (approved):

aloknerurkar:
> Ok sounds good to me.
> 
> Regarding bad paths. To my understanding, we support UNIX-like paths in Bee, right? If so, then I think there are a few checks that can be done here, which in the `tar` upload use-case are most probably handled by the tar implementation itself, but with multipart are not treated. Things like:
> 
> * calling [`path.Clean`](https://pkg.go.dev/path#Clean) on the path
> * verification if it is a relative or absolute path - what should be the required format?
> * verification if the path does not contain invalid characters - see for example [here](https://www.cyberciti.biz/faq/linuxunix-rules-for-naming-file-and-directory-names/)
> * verification if the folder/file names are not longer than 255 characters
> 
> And maybe more, these are just from top of my head.

So this was the problem why we disallowed the multipart upload in the first place. Handling all the corner cases would be too many checks in my opinion. The manifest is a key-value store and should work fine for any unique path names. This way if there are any developer-induced errors for some reason, this would manifest when they try to access the files. If the feature is used correctly everything should work.

If we are okay with this we can use this change as is. I am not really sure what are all the cases covered by the tar format. But since it needs a filesystem to create, we can rest assured it is some valid structure. Unlike this case.

acud:
So I agree with @aloknerurkar that it can be a bit more tricky than it might seem in the first place. Don't forget that stuff like `/somefolder/././././././somesubfolder/.././././somesubfolder/somefile.html` is a valid *nix path too. It would be good to follow something more concrete than a linux Q&A website... I tried to find some more official POSIX path specification but could not find anything meaningful so far... I'll be on the lookout for this and would prefer not to implement something by a hunch here. There's a lot of corner cases and it is also not 100% clear to me how for example UTF8 paths fit into this, etc

AuHau:
Sooo, this is up to you guys if you want to merge it in the current form, but here are my 5 cents.
While yes Swarm can be used to store generic data, it is still very much related to the file system's concepts like files, folders, paths and it support these and uses these in its terminology. I think it is generally expected to follow *NIX style paths in project like this. I think it is quite important to stay consistent in the rules that applies for those, because as @Elad Acud mentioned in the PR you can have for example some/path and /some/./path which suddenly points to the same file and this should be handled. And dismissing these problems with saying that "we use manifests which are key-value store" is IMHO shortsighted, because the user will expect the behavior of paths and folders as we are using this terminology. If we would like to be free of these restrictions different names should had been chosen for these concepts.
While I agree that there might be some weird edge-cases that we might not cover, I still think it is fairly simple to cover like 80% of those with the few rules that I have wrote down in the PR and they are easy to implement. 
My prediction is that if we won't do input sanitation (which is a basic think to do) we can expect fairly soon bug reports like "file after uploading disappeared" or something similar. 
Btw. I would like to work at some point on hack project writing FUSE mount for Swarm and then that is another part where proper path handling is essential and I have no idea how I would handle manifests with invalid paths

acud (approved):

AuHau:
So the outcome of our API Guild call yesterday is that `bzz` endpoint does not follow any conventions and as such does not require any path validation or similar. Therefore it is good to merge as-is.

agazso (approved):

PR #2952, opened 2022-05-12
feat(bmt): segment inclusion Merkle proofs

aloknerurkar:
### Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2952)

### DIFF
--- a/pkg/bmt/proof.go
+++ b/pkg/bmt/proof.go
@@ -1,3 +1,7 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 package bmt
 
 // Prover wraps the Hasher to allow Merkle proof functionality
@@ -14,6 +18,9 @@ type Proof struct {
 
 // Proof returns the inclusion proof of the i-th data segment
 func (p Prover) Proof(i int) Proof {
+	if i < 0 || i > 127 {
+		panic("segment index can only lie between 0-127")
+	}
 	i = i / 2
 	n := p.bmt.leaves[i]
 	isLeft := n.isLeft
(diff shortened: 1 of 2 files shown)


zelig (approved):

pkg/bmt/proof.go
```diff
@@ -0,0 +1,64 @@
+package bmt
```
acud, line 1: code copy missing for both files

pkg/bmt/proof.go
```diff
@@ -0,0 +1,64 @@
...
+// Proof represents a Merkle proof of segment
+type Proof struct {
+	Section []byte
+	Sisters [][]byte
+	Span    []byte
+}
+
+// Proof returns the inclusion proof of the i-th data segment
+func (p Prover) Proof(i int) Proof {
+	i = i / 2
```
acud, line 17: I would add some input validation here to check that we never have more than 128 segments here, otherwise we will panic

pkg/bmt/proof.go
```diff
@@ -0,0 +1,64 @@
...
+	n := p.bmt.leaves[i]
+	isLeft := n.isLeft
+	var sisters [][]byte
+	for n = n.parent; n != nil; n = n.parent {
+		sisters = append(sisters, n.getSister(isLeft))
```
acud, line 22: I had some suspicion that there might be some weird slice bug awaiting us [here](https://go.dev/play/p/LMWJw3cKerO) but that might not be the case. However, at first sight this looks suspicious, because `sisters` is a slice of slices, which means that when you do an `append` you're appending the slices, not the data, which also means that if for some reason another goroutine modifies those slices, the proof would immediately be affected and subject to an unintentional slice array sharing. To make things worse, the `bmt` instances are put in a pool once usage is done, so if we go with this approach, the correctness of the proof is at the mercy of whether we do or don't do segment or tree node allocations (if we would optimize on allocations this would be a no-go).

For the sake of clarity I would go with a `copy()` of the node data into the new slice (read: make a copy of the sister nodes before `append`ing)

acud (changes requested):

acud (approved):

PR #2961, opened 2022-05-17
fix: prepare credit context argument

### LINKED ISSUE #2959: `PrepareDebit` and `PrepareCredit` should take a context
We've been seeing lock contention in the accounting package on these method calls from the protocols. While it is not clear why the elevated amount of calls is happening, it is clear that the goroutines that try to acquire the lock don't back off in case the context expires. We should respect the context when trying to do these operations, and potentially return an error and return in case the context expires _while_ trying to acquire the lock.

notanatol:
### Checklist

### Description
In order to minimize lock contention for the accounting peer we will reject the goroutines that have an expired context as well as the goroutines that do not receive the lock within the context deadline.
This PR introduces a context aware mutex for that purpose.

Closes #2959

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2961)

### DIFF
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -21,7 +21,6 @@ import (
 	"github.com/ethersphere/bee/pkg/settlement/pseudosettle"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
-	"golang.org/x/sync/semaphore"
 )
 
 var (
@@ -86,43 +85,35 @@ type PayFunc func(context.Context, swarm.Address, *big.Int)
 // RefreshFunc is the function used for sync time-based settlement
 type RefreshFunc func(context.Context, swarm.Address, *big.Int, *big.Int) (*big.Int, int64, error)
 
-// SemMutex is a drop in replacement for the sync.Mutex
+// Mutex is a drop in replacement for the sync.Mutex
 // it will not lock if the context is expired
-type SemMutex struct {
-	sem *semaphore.Weighted
-}
-
-func NewMutex() SemMutex {
-	return SemMutex{
-		sem: semaphore.NewWeighted(1),
-	}
+type Mutex struct {
+	mu sync.Mutex
 }
 
 var ErrFailToLock = errors.New("failed to lock")
 
-func (m *SemMutex) Lock(ctx context.Context) error {
-	if err := m.sem.Acquire(ctx, 1); err != nil {
-		return fmt.Errorf("%v: %w", ErrFailToLock, err)
+func (m *Mutex) TryLock(ctx context.Context) error {
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("%v: %w", ctx.Err(), ErrFailToLock)
+	default:
+		m.mu.Lock()
+		return nil
 	}
-
-	return nil
 }
 
-func (m *SemMutex) TryLock() error {
-	if !m.sem.TryAcquire(1) {
-		return ErrFailToLock
-	}
-
-	return nil
+func (m *Mutex) Lock() {
+	m.mu.Lock()
 }
 
-func (m *SemMutex) Unlock() {
-	m.sem.Release(1)
+func (m *Mutex) Unlock() {
+	m.mu.Unlock()
 }
 
 // accountingPeer holds all in-memory accounting information for one peer.
 type accountingPeer struct {
-	SemMutex                                // lock to be held during any accounting action for this peer
+	lock                           Mutex    // lock to be held during any accounting action for this peer
 	reservedBalance                *big.Int // amount currently reserved for active peer interaction
 	shadowReservedBalance          *big.Int // amount potentially to be debited for active peer interaction
 	ghostBalance                   *big.Int // amount potentially could have been debited for but was not
@@ -233,11 +224,11 @@ func (a *Accounting) getIncreasedExpectedDebt(peer swarm.Address, accountingPeer
 func (a *Accounting) PrepareCredit(ctx context.Context, peer swarm.Address, price uint64, originated bool) (Action, error) {
 	accountingPeer := a.getAccountingPeer(peer)
 
-	if err := accountingPeer.Lock(ctx); err != nil {
+	if err := accountingPeer.lock.TryLock(ctx); err != nil {
 		a.logger.Errorf("prepare credit: failed to acquire lock: %v", err)
 		return nil, err
 	}
-	defer accountingPeer.Unlock()
+	defer accountingPeer.lock.Unlock()
 
 	if !accountingPeer.connected {
 		return nil, errors.New("connection not initialized yet")
@@ -306,10 +297,8 @@ func (a *Accounting) PrepareCredit(ctx context.Context, peer swarm.Address, pric
 }
 
 func (c *creditAction) Apply() error {
-	if err := c.accountingPeer.TryLock(); err != nil {
-		return err
-	}
-	defer c.accountingPeer.Unlock()
+	c.accountingPeer.lock.Lock()
+	defer c.accountingPeer.lock.Unlock()
 
 	currentBalance, err := c.accounting.Balance(c.peer)
 	if err != nil {
@@ -381,10 +370,8 @@ func (c *creditAction) Cleanup() {
 		return
 	}
 
-	if err := c.accountingPeer.TryLock(); err != nil {
-		return
-	}
-	defer c.accountingPeer.Unlock()
+	c.accountingPeer.lock.Lock()
+	defer c.accountingPeer.lock.Unlock()
 
 	if c.price.Cmp(c.accountingPeer.reservedBalance) > 0 {
 		c.accounting.logger.Error("attempting to release more balance than was reserved for peer")
@@ -586,7 +573,7 @@ func (a *Accounting) getAccountingPeer(peer swarm.Address) *accountingPeer {
 	peerData, ok := a.accountingPeers[peer.String()]
 	if !ok {
 		peerData = &accountingPeer{
-			SemMutex:              NewMutex(),
+			lock:                  Mutex{},
 			reservedBalance:       big.NewInt(0),
 			shadowReservedBalance: big.NewInt(0),
 			ghostBalance:          big.NewInt(0),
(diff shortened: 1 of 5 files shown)


mrekucci (commented):
It would be good if this scenario was also covered by a test.

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2961#-:-N2HBUwV37_CoEiyhWes:bvkghkg)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @notanatol)

___
*[`pkg/accounting/accounting.go` line 40 at r1](https://reviewable.io/reviews/ethersphere/bee/2961#-N2H3YySDTwvG77zZaP1:-N2H3YySDTwvG77zZaP2:b-11cf2y) ([raw file](https://github.com/ethersphere/bee/blob/1ceba41f2cee5ad94db33dbf73ad1e9db4fa0c5a/pkg/accounting/accounting.go#L40)):*
> ```Go
> // Interface is the Accounting interface.
> type Interface interface {
> 	// Credit action to prevent overspending in case of concurrent requests.
> ```

Comment should have the following format 'PrepareCredit ...'
___
*[`pkg/accounting/accounting.go` line 93 at r1](https://reviewable.io/reviews/ethersphere/bee/2961#-N2HB6oaDpwCs9N1jBtj:-N2HB6oaDpwCs9N1jBtk:bswawqo) ([raw file](https://github.com/ethersphere/bee/blob/1ceba41f2cee5ad94db33dbf73ad1e9db4fa0c5a/pkg/accounting/accounting.go#L93)):*
> ```Go
> 	// unlocked = 0 (default)
> 	// locked = 1
> 	locked uint32
> ```

You can use Uber's [`atomic.Bool`](https://github.com/uber-go/atomic).
___
*[`pkg/accounting/accounting.go` line 96 at r1](https://reviewable.io/reviews/ethersphere/bee/2961#-N2H5jNo45YjDwEl0gDz:-N2H5jNo45YjDwEl0gE-:bzviq8) ([raw file](https://github.com/ethersphere/bee/blob/1ceba41f2cee5ad94db33dbf73ad1e9db4fa0c5a/pkg/accounting/accounting.go#L96)):*
> ```Go
> }
> 
> func (m *AtomicMutex) Locked() bool {
> ```

This method is not used.

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2961)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @notanatol)

___
*[`pkg/accounting/accounting.go` line 96 at r2](https://reviewable.io/reviews/ethersphere/bee/2961#-N2HLTZXAXB-LwBsKjuD:-N2HLQEe4vAFR1o_rSkE:b-1y7gve) ([raw file](https://github.com/ethersphere/bee/blob/701200767e9968bd80683c3c635148f27534bbe5/pkg/accounting/accounting.go#L96)):*
> ```Go
> }
> 
> func (m *AtomicMutex) Lock() (out chan struct{}) {
> ```

Creating a new channel on every successful `Lock` call (where most of the time the channel is not used) will put unnecessary pressure on the GC. Maybe consider replacing the whole lock with a channel (wrapped internally) or use a [semaphore](https://pkg.go.dev/golang.org/x/sync/semaphore):
```
sem := semaphore.NewWeighted(1)
// ...
if err := sem.Acquire(ctx, 1) != nil {
	return err
}
defer sem.Release(1)
```

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2961)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @mrekucci)

___
*[`pkg/accounting/accounting.go` line 96 at r2](https://reviewable.io/reviews/ethersphere/bee/2961#-N2HLTZXAXB-LwBsKjuD:-N2I5qsv58jRZAwJRl58:bof9t1h) ([raw file](https://github.com/ethersphere/bee/blob/701200767e9968bd80683c3c635148f27534bbe5/pkg/accounting/accounting.go#L96)):*

Creating a new channel on every successful `Lock` call (where most of the time the channel is not used) will put unnecessary pressure on the GC. Maybe consider replacing the whole lock with a channel (wrapped internally) or use a [semaphore](https://pkg.go.dev/golang.org/x/sync/semaphore):
```
sem := semaphore.NewWeighted(1)
// ...
if err := sem.Acquire(ctx, 1) != nil {
	return err
}
defer sem.Release(1)
```
</blockquote></details>

swapped for a no-allocation implementation

mrekucci (approved):
> Reviewed 12 of 12 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2961)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud, @aloknerurkar, and @istae)

___
*[`pkg/accounting/accounting.go` line 112 at r4](https://reviewable.io/reviews/ethersphere/bee/2961#-N4Xkpxm8FLHIuZ0UbVj:-N4Xkpxm8FLHIuZ0UbVk:b-a9rmv9) ([raw file](https://github.com/ethersphere/bee/blob/d9725de81a29d4024a2cfba3d86c92285d28348a/pkg/accounting/accounting.go#L112)):*
> ```Go
> }
> 
> func (m *Mutex) TryLock() {
> ```

I think this can be renamed to `Lock` (since it locks) and the one above with context to `TryLock` (since it might lock depending on the context).

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2961)* status: 11 of 12 files reviewed, all discussions resolved (waiting on @acud, @aloknerurkar, @istae, and @mrekucci)

___
*[`pkg/accounting/accounting.go` line 112 at r4](https://reviewable.io/reviews/ethersphere/bee/2961#-N4Xkpxm8FLHIuZ0UbVj:-N4XuNJm4BmYB3XcKEs_:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/d9725de81a29d4024a2cfba3d86c92285d28348a/pkg/accounting/accounting.go#L112)):*

I think this can be renamed to `Lock` (since it locks) and the one above with context to `TryLock` (since it might lock depending on the context).
</blockquote></details>

fixed

mrekucci (approved):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2961)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud, @aloknerurkar, and @istae)

pkg/accounting/accounting.go
```diff
@@ -923,10 +958,14 @@ func (a *Accounting) NotifyRefreshmentReceived(peer swarm.Address, amount *big.I
 }
 
 // PrepareDebit prepares a debit operation by increasing the shadowReservedBalance
-func (a *Accounting) PrepareDebit(peer swarm.Address, price uint64) (Action, error) {
+func (a *Accounting) PrepareDebit(ctx context.Context, peer swarm.Address, price uint64) (Action, error) {
 	accountingPeer := a.getAccountingPeer(peer)
 
-	accountingPeer.lock.Lock()
+	if err := accountingPeer.lock.TryLock(ctx); err != nil {
```
istae, line 964: I am bit confused, why cant we check if context is done directly inside PrepareDebit and Credit? Why the need for a custom mutex?

notanatol: We want to wait to acquire the lock but only for as long as the context permits us, this is why we have the select construct in the `TryLock` method.
Just checking for context expiration in the method would weed out some expired goroutines but we want to go further and reject those goroutines which were not able to get the lock within the deadline.

istae: I see, so we are essentially racing the context expiration with lock acquiral, which means we have seen cases where these functions were under lock contention, for longer periods than one second, judging from the code.

istae (approved):

pkg/accounting/accounting.go
```diff
@@ -195,10 +226,13 @@ func (a *Accounting) getIncreasedExpectedDebt(peer swarm.Address, accountingPeer
...
 }
 
-func (a *Accounting) PrepareCredit(peer swarm.Address, price uint64, originated bool) (Action, error) {
+func (a *Accounting) PrepareCredit(ctx context.Context, peer swarm.Address, price uint64, originated bool) (Action, error) {
 	accountingPeer := a.getAccountingPeer(peer)
 
-	accountingPeer.lock.Lock()
+	if err := accountingPeer.lock.TryLock(ctx); err != nil {
+		a.logger.Errorf("prepare credit: failed to acquire lock: %v", err)
+		return nil, err
```
istae, line 234: this would be a good place to add a metric to monitor lock contention

istae: and also inside the other Prepare function

acud: it would be difficult to measure lock contention per-se... that's what you should use the mutex/block profile for. the metric would not give much more info here, because it would just be an aggregate of those values. you can measure the amount of time spend for the entire function vs. time spent under lock in a similar manner like what we do in the localstore, but it is a very different use case - if there are a lot of goroutines waiting you'll have astronomical values for waiting (time for all of them in aggregate) vs very low values under lock (since only some will actually execute the code)

pkg/accounting/accounting.go
```diff
@@ -85,13 +85,44 @@ type PayFunc func(context.Context, swarm.Address, *big.Int)
 // RefreshFunc is the function used for sync time-based settlement
 type RefreshFunc func(context.Context, swarm.Address, *big.Int, *big.Int) (*big.Int, int64, error)
 
+// mutex is a drop in replacement for the sync.mutex
+// it will not lock if the context is expired
+type mutex struct {
```
aloknerurkar, line 90: Why are there no tests for this?

aloknerurkar (changes requested):
So I don't see any tests for the new behavior. Neither the mutex nor the context timeout in PrepareXXX methods is being tested.

pkg/accounting/accounting.go
notanatol: added

notanatol:
> So I don't see any tests for the new behavior. Neither the mutex nor the context timeout in PrepareXXX methods is being tested.

I exported the mutex construct and added a test for it. As for the Prepare methods, they are tested implicitly in other tests but also covering this particular scenario adds a lot of overhead.

pkg/accounting/accounting.go
```diff
@@ -85,13 +85,44 @@ type PayFunc func(context.Context, swarm.Address, *big.Int)
 // RefreshFunc is the function used for sync time-based settlement
 type RefreshFunc func(context.Context, swarm.Address, *big.Int, *big.Int) (*big.Int, int64, error)
 
+// Mutex is a drop in replacement for the sync.Mutex
+// it will not lock if the context is expired
+type Mutex struct {
```
aloknerurkar, line 90: Not sure about the best practices in go. But would it make sense to not export this and have a separate test file that could lie in the same package? That way we dont need to export this. Just a thought.

aloknerurkar (approved):

pkg/accounting/accounting.go
notanatol: I think it can be argued that its functionality falls under a "general purpose" category - a context aware mutex. Since it's not opinionated towards any business rules - I think it's fine to export it.

PR #2971, opened 2022-05-19
fix(kademlia): increase bit suffix length to improve bin population

istae:
### Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2971)

### DIFF
--- a/pkg/topology/kademlia/kademlia.go
+++ b/pkg/topology/kademlia/kademlia.go
@@ -37,7 +37,7 @@ import (
 const (
 	maxConnAttempts        = 1 // when there is maxConnAttempts failed connect calls for a given peer it is considered non-connectable
 	maxBootNodeAttempts    = 3 // how many attempts to dial to boot-nodes before giving up
-	defaultBitSuffixLength = 3 // the number of bits used to create pseudo addresses for balancing
+	defaultBitSuffixLength = 4 // the number of bits used to create pseudo addresses for balancing
 
 	addPeerBatchSize = 500
 
--- a/pkg/topology/kademlia/export_test.go
+++ b/pkg/topology/kademlia/export_test.go
@@ -16,6 +16,7 @@ var (
 	}
 	GenerateCommonBinPrefixes = generateCommonBinPrefixes
 	PeerPingPollTime          = &peerPingPollTime
+	BitSuffixLength           = defaultBitSuffixLength
 )
 
 type PeerFilterFunc = peerFilterFunc
--- a/pkg/topology/kademlia/kademlia_test.go
+++ b/pkg/topology/kademlia/kademlia_test.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"math/rand"
 	"reflect"
 	"sync"
@@ -445,6 +446,11 @@ func TestManage(t *testing.T) {
 		})
 	)
 
+	defer func(p int) {
+		*kademlia.SaturationPeers = p
+	}(*kademlia.SaturationPeers)
+	*kademlia.SaturationPeers = 4
+
 	if err := kad.Start(context.Background()); err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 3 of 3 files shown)


pkg/topology/kademlia/kademlia.go
```diff
@@ -281,18 +281,15 @@ func (k *Kad) connectBalanced(wg *sync.WaitGroup, peerConnChan chan<- *peerConnI
 	}
 }
 
-// connectNeighbours attempts to connect to the neighbours
-// which were not considered by the connectBalanced method.
-func (k *Kad) connectNeighbours(wg *sync.WaitGroup, peerConnChan chan<- *peerConnInfo) {
+// connectUntilSaturation attempts to connect to peers until bin oversaturation is reached.
+func (k *Kad) connectUntilSaturation(wg *sync.WaitGroup, peerConnChan chan<- *peerConnInfo) {
```
mrekucci, line 285: I recommend calling it `connectUntilOversaturation`.

acud:
don't understand.... didn't we say we're just changing the saturation and oversaturation parameter values? btw inside the neighborhood you will always try to connect to all neighbors, so concepts of saturation and oversaturation are largely irrelevant. saturation is only relevant **outside** of neighborhood

istae:
@acud please ignore the draft PR for now 👍

istae:
> don't understand.... didn't we say we're just changing the saturation and oversaturation parameter values? btw inside the neighborhood you will always try to connect to all neighbors, so concepts of saturation and oversaturation are largely irrelevant. saturation is only relevant **outside** of neighborhood

having looked at kademlia, we actually don't use the `saturationPeers` anywhere meaningfully. This PR changes that and also increases the saturation limits.

pkg/topology/kademlia/kademlia.go
```diff
@@ -281,23 +281,22 @@ func (k *Kad) connectBalanced(wg *sync.WaitGroup, peerConnChan chan<- *peerConnI
...
 	var currentPo uint8 = 0
 
-	_ = k.knownPeers.EachBinRev(func(addr swarm.Address, po uint8) (bool, bool, error) {
-		depth := k.NeighborhoodDepth()
+	_ = k.knownPeers.EachBinRev(func(addr swarm.Address, bin uint8) (bool, bool, error) {
 
-		// out of depth, skip bin
-		if po < depth {
+		// if the bin is already saturated AND the bin is out of the depth, skip bin
+		if bin < k.NeighborhoodDepth() && k.connectedPeers.BinSize(bin) >= saturationPeers {
```
acud, line 294: again, there is no notion of saturation within depth and we always want to connect to all peers

acud (changes requested):

pkg/topology/kademlia/kademlia.go
acud: I find this change unnecessary. We're mixing things up here. The parameter change should be enough, please confirm @istae

acud: If the connector method handles both bins which are inside and outside of depth there can not exist any notion of saturation in the name

istae: So how's `connectToPeers` ?

acud: this is unfortunately not a naming issue, but an issue of single responsibility. you're not just renaming but mixing up responsibilities. the change you've done here basically overrides the responsibility of the bin balancer by connecting to peers outside of depth, which would mess up the balancing inside the bins.

acud: > the neighborhood connecter ignores bins that are out of depth

that is incorrect.

have a look at this conditional that you wrote:
```go
		if bin < k.NeighborhoodDepth() && k.connectedPeers.BinSize(bin) >= saturationPeers {
```
this means that if the bin is out of depth and unsaturated then this connector will indeed cause more peers to be connected outside of depth and this is not what we want. we always want the balancer to do connections out of depth. no reinforcements needed there. if you want more peers out of depth you need to change the number of bits the balancer is trying to brute force.

istae: Without a new connector function, the issue of some of the shallower bins never filling up won't be resolved.

istae: > the neighborhood connecter ignores bins that are out of depth
yes, `connectNeighbours` in master is doing this,.

mrekucci (approved):

acud (approved):

PR #2974, opened 2022-05-20
fix: pusher attempt delete

istae:
### Checklist

### DIFF
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -301,6 +301,7 @@ func (s *Service) checkReceipt(receipt *pushsync.Receipt) error {
 	}
 	s.logger.Tracef("pusher: pushed chunk %s to node %s, receipt depth %d", addr, peer, po)
 	s.metrics.ReceiptDepth.WithLabelValues(strconv.Itoa(int(po))).Inc()
+	s.attempts.delete(addr)
 	return nil
 }
 
--- a/pkg/pusher/inflight.go
+++ b/pkg/pusher/inflight.go
@@ -28,14 +28,16 @@ func (i *inflight) delete(ch swarm.Chunk) {
 }
 
 func (i *inflight) set(addr []byte) bool {
+
 	i.mtx.Lock()
+	defer i.mtx.Unlock()
+
 	key := string(addr)
 	if _, ok := i.inflight[key]; ok {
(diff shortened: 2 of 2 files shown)


acud:
I'm not entirely sure that this is desirable behavior. At some point we've turned on the skip list for successful pushes too, which should in theory result in (some) disjoint path traversal for repeated successful pushes of the same chunk. The key doesn't leak since it has a TTL and there are timely cleanups on the skip-list.

istae:
> I'm not entirely sure that this is desirable behavior. At some point we've turned on the skip list for successful pushes too, which should in theory result in (some) disjoint path traversal for repeated successful pushes of the same chunk. The key doesn't leak since it has a TTL and there are timely cleanups on the skip-list.

@acud I think you were looking at a closed PR. This PR is related to the pusher package :)

acud:
oops wrong pacakge. sorry you're right. i'll review again

acud (approved):
I find this change small but significant. Just spent a considerable amount of time trying to understand what it exactly does and how. It would be nice to have more comments about this, since this is not trivial.

aloknerurkar (approved):

pkg/pusher/pusher.go
```diff
@@ -295,12 +295,15 @@ func (s *Service) checkReceipt(receipt *pushsync.Receipt) error {
...
 	d := s.depther.NeighborhoodDepth()
+
+	// if the receipt po is out of depth AND the receipt has not yet hit the maximum retry limit, reject the receipt.
 	if po < d && s.attempts.try(addr) {
 		s.metrics.ShallowReceiptDepth.WithLabelValues(strconv.Itoa(int(po))).Inc()
 		return fmt.Errorf("pusher: shallow receipt depth %d, want at least %d", po, d)
 	}
 	s.logger.Tracef("pusher: pushed chunk %s to node %s, receipt depth %d", addr, peer, po)
 	s.metrics.ReceiptDepth.WithLabelValues(strconv.Itoa(int(po))).Inc()
+	s.attempts.delete(addr)
```
acud, line 306: since checking the receipt is not synchronized with setting the chunk as synced, this might result in the chunk being pushed again in the timeframe between the delete and the setting of the chunk. not critical but can result in a bit of redundant operations

PR #2975, opened 2022-05-20
refactor: introduce structural logging

mrekucci:
### Checklist

### Description
This new logger borrows some of the ideas described here: https://github.com/go-logr/logr#background
with the exception of excluding warning and debug log levels, for compatibility reasons. In addition, it adds V levels to the debug log level, so that messages with larger V levels can be logged if needed. The status of loggers can be queried via the HTTP endpoint, and verbosity can also be set for single or multiple loggers (based on regex matching) via the HTTP endpoint at runtime.

During debugging sessions, it is often necessary to reduce or increase/decrease the range of log messages for individual packages (or groups) in order to efficiently focus on the relevant log messages. It is also very important to have this ability to turn on/off/change the verbosity of the logs at runtime without restarting individual nodes. Finally, log messages that have structure (structural logging) are a crucial component in parsing logs efficiently and in automated log processing.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/2975)

### DIFF
--- a/pkg/api/logger.go
+++ b/pkg/api/logger.go
@@ -112,6 +112,7 @@ func (s *Service) loggerSetVerbosityHandler(w http.ResponseWriter, r *http.Reque
 		s.logger.Debugf("loggers set verbosity: parse verbosity level failed: %v", err)
 		s.logger.Error("loggers set verbosity: parse verbosity level failed")
 		jsonhttp.BadRequest(w, err)
+		return
 	}
 
 	exp, err := base64.URLEncoding.DecodeString(mux.Vars(r)["exp"])
@@ -119,6 +120,7 @@ func (s *Service) loggerSetVerbosityHandler(w http.ResponseWriter, r *http.Reque
 		s.logger.Debugf("loggers set verbosity: query parameter decoding failed: %v", err)
 		s.logger.Error("loggers set verbosity: query parameter decoding failed")
 		jsonhttp.BadRequest(w, err)
+		return
 	}
 
 	if err := logSetVerbosityByExp(string(exp), verbosity); err != nil {
--- a/pkg/api/logger.go
+++ b/pkg/api/logger.go
@@ -1,0 +1,129 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package api
+
+import (
+	"encoding/base64"
+	"fmt"
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/log"
+	"github.com/gorilla/mux"
+)
+
+// The following variables exist only to be mocked in tests.
+var (
+	logRegistryIterate   = log.RegistryIterate
+	logSetVerbosityByExp = log.SetVerbosityByExp
+)
+
+type (
+	data struct {
+		Next  node     `json:"/,omitempty"`
+		Names []string `json:"+,omitempty"`
+	}
+
+	node map[string]*data
+
+	logger struct {
+		Logger    string `json:"logger"`
+		Verbosity string `json:"verbosity"`
+		Subsystem string `json:"subsystem"`
+		ID        string `json:"id"`
+	}
+
+	loggerResult struct {
+		Tree    node     `json:"tree"`
+		Loggers []logger `json:"loggers"`
+	}
+)
+
+// loggerGetHandler returns all available loggers that match the specified expression.
+func (s *Service) loggerGetHandler(w http.ResponseWriter, r *http.Request) {
+	var (
+		exp string
+		err error
+	)
+
+	if val, ok := mux.Vars(r)["exp"]; ok {
+		buf, err := base64.URLEncoding.DecodeString(val)
+		if err != nil {
+			s.logger.Debugf("loggers get: query parameter decoding failed: %v", err)
+			s.logger.Error("loggers get: query parameter decoding failed")
+			jsonhttp.BadRequest(w, err)
+		}
+		exp = string(buf)
+	}
+
+	rex, err := regexp.Compile(exp)
+
+	result := loggerResult{Tree: make(node)}
+	logRegistryIterate(func(id, name string, verbosity log.Level, v uint) bool {
+		if exp == "" || exp == id || (rex != nil && rex.MatchString(id)) {
+			if int(verbosity) == int(v) {
+				verbosity = log.VerbosityAll
+			}
+
+			// Tree structure.
+			curr := result.Tree
+			path := strings.Split(name, "/")
+			last := len(path) - 1
+			for i, n := range path {
+				if curr[n] == nil {
+					curr[n] = &data{Next: make(node)}
+				}
+				if i == last {
+					name := fmt.Sprintf("%s|%s", verbosity, id)
+					curr[n].Names = append(curr[n].Names, name)
+				}
+				curr = curr[n].Next
+			}
+
+			// Flat structure.
+			result.Loggers = append(result.Loggers, logger{
+				Logger:    name,
+				Verbosity: verbosity.String(),
+				Subsystem: id,
+				ID:        base64.URLEncoding.EncodeToString([]byte(id)),
+			})
+		}
+		return true
+	})
+
+	if len(result.Loggers) == 0 && err != nil {
+		s.logger.Debugf("loggers get: regexp compilation failed: %v", err)
+		s.logger.Error("loggers get: regexp compilation failed")
+		jsonhttp.BadRequest(w, err)
+	} else {
+		jsonhttp.OK(w, result)
+	}
+}
+
+// loggerSetVerbosityHandler sets logger(s) verbosity level based on
+// the specified expression or subsystem that matches the logger(s).
+func (s *Service) loggerSetVerbosityHandler(w http.ResponseWriter, r *http.Request) {
+	verbosity, err := log.ParseVerbosityLevel(mux.Vars(r)["verbosity"])
+	if err != nil {
+		s.logger.Debugf("loggers set verbosity: parse verbosity level failed: %v", err)
+		s.logger.Error("loggers set verbosity: parse verbosity level failed")
+		jsonhttp.BadRequest(w, err)
+	}
+
+	exp, err := base64.URLEncoding.DecodeString(mux.Vars(r)["exp"])
+	if err != nil {
+		s.logger.Debugf("loggers set verbosity: query parameter decoding failed: %v", err)
+		s.logger.Error("loggers set verbosity: query parameter decoding failed")
+		jsonhttp.BadRequest(w, err)
+	}
+
+	if err := logSetVerbosityByExp(string(exp), verbosity); err != nil {
+		jsonhttp.BadRequest(w, err)
+	} else {
+		jsonhttp.OK(w, nil)
+	}
+}
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -263,6 +263,5 @@ func WithLevelHooks(l Level, hooks ...Hook) Option {
 		default:
 			opts.levelHooks[l] = append(opts.levelHooks[l], hooks...)
 		}
-
 	}
 }
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -108,9 +108,17 @@ type Builder interface {
 	// to be logged with each log line.
 	WithValues(keysAndValues ...interface{}) Builder
 
-	// Build returns new or existing Logger
+	// Build returns a new or existing Logger
 	// instance, if such instance already exists.
+	// The new instance is not registered in the
+	// logger registry.
 	Build() Logger
+
+	// Register returns new or existing Logger
+	// instance, if such instance already exists.
+	// If a new instance is created, it is also
+	// registered in the logger registry.
+	Register() Logger
 }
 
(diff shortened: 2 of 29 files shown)


acud (changes requested):
Thanks @mrekucci. There's quite a bit to review here and I am not sure I can go in depth here since there's a lot of details which I might not really understand why they are needed in the first place. My main concern is with regards to allocations, specifically whether our existing logger somehow optimizes on that. We do a lot of logging and therefore this will mean we will be doing a lot of allocations. It would be good to know if the current logger optimizes before switching to this one. Also noticed is usage of `runtime.Caller`, which executes a serious amount of operations under the hood. I would try to avoid such things by default and allow to feature-flag them as they can be quite burdensome.

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2975#-:-N4fNRwwEJcTSBxrShaN:b-v2d7qi)* status: all files reviewed, 6 unresolved discussions (waiting on @janos, @mrekucci, and @zelig)

___
*[`pkg/log/formatter.go` line 97 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fIKFUC20FVmqkMpha:-N4fIKFUC20FVmqkMphb:bvwcoqy) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/formatter.go#L97)):*
> ```Go
> // never escaped; the opposite is true for args.
> func (f *formatter) render(base, args []interface{}) []byte {
> 	buf := bytes.NewBuffer(make([]byte, 0, 1024))
> ```

this might/will be heavy on allocations no? does the previous logger we've been using optimize on this in any way? with the amount of logging we do this will pile up quite a bit
___
*[`pkg/log/formatter.go` line 162 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fIjRs5se-Zgx0YVOA:-N4fIjRs5se-Zgx0YVOB:bfl09c4) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/formatter.go#L162)):*
> ```Go
> 
> 	if depth > f.opts.maxLogDepth {
> 		return `"<max-log-depth-exceeded>"`
> ```

you probably want to pull this out to a const
___
*[`pkg/log/formatter.go` line 217 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fIp-j5unkILw2dI1Z:-N4fIp-j5unkILw2dI1_:b-zb6wli) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/formatter.go#L217)):*
> ```Go
> 		return `"` + strconv.FormatComplex(v, 'f', -1, 128) + `"`
> 	case PseudoStruct:
> 		buf := bytes.NewBuffer(make([]byte, 0, 1024))
> ```

same question here as above
___
*[`pkg/log/formatter.go` line 446 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fJFjNEYNzMpcy-qa3:-N4fJFjNEYNzMpcy-qa4:bji9y6k) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/formatter.go#L446)):*
> ```Go
> 		return strconv.Quote(s)
> 	}
> 	b := bytes.NewBuffer(make([]byte, 0, 1024))
> ```

and this
___
*[`pkg/log/log.go` line 88 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fKU6F5D0bijgHw4wO:-N4fKU6F5D0bijgHw4wP:bt69e4d) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/log.go#L88)):*
> ```Go
> // Note, the call must be non-blocking.
> type Hook interface {
> 	Fire(Level) error
> ```

not sure i understand this... is this hook called on every message on the loglevel? if so... why
___
*[`pkg/log/logger.go` line 119 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fJl1C4Ze8hoXg1TNQ:-N4fJl1C4Ze8hoXg1TNR:b-4ltptb) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/logger.go#L119)):*
> ```Go
> 	if l.verbosity.get() >= VerbosityError {
> 		if err := l.log(VerbosityError, CategoryError, err, msg, keysAndValues...); err != nil {
> 			fmt.Println(err)
> ```

isn't this supposed to write to stderr?

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2975)* status: all files reviewed, 6 unresolved discussions (waiting on @acud, @janos, and @zelig)

___
*[`pkg/log/formatter.go` line 97 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fIKFUC20FVmqkMpha:-N4m1YnC4evy3BCkns5s:bd6y6hj) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/formatter.go#L97)):*

this might/will be heavy on allocations no? does the previous logger we've been using optimize on this in any way? with the amount of logging we do this will pile up quite a bit
</blockquote></details>

Most of the formatter code was taken from logr, there is definitely room for optimization (and it's in the pipeline, see the TODOs) but that requires writing benchmarks before doing it (which might be time-consuming). The current logger we're using (logrus) allocates quite heavily, see: http://hackemist.com/logbench/
___
*[`pkg/log/formatter.go` line 162 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fIjRs5se-Zgx0YVOA:-N4m5VmkEf8alteQf-Db:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/formatter.go#L162)):*

you probably want to pull this out to a const
</blockquote></details>

Done.
___
*[`pkg/log/formatter.go` line 217 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fIp-j5unkILw2dI1Z:-N4m5XnuDie4sFObf6M1:b-96utdj) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/formatter.go#L217)):*

same question here as above
</blockquote></details>

The same applies here as above.
___
*[`pkg/log/formatter.go` line 446 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fJFjNEYNzMpcy-qa3:-N4m5ayY7UN3J77h0CzU:b-hbcorw) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/formatter.go#L446)):*

and this
</blockquote></details>

The same.
___
*[`pkg/log/log.go` line 88 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fKU6F5D0bijgHw4wO:-N4m6-LL2BBm0GUrnRfw:b-kj0c27) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/log.go#L88)):*

not sure i understand this... is this hook called on every message on the loglevel? if so... why
</blockquote></details>

It is called on every log message of the associated log level. This is the same as in the current logger, where we have hooks that count (using metrics) how many log messages have been written at each log level.
___
*[`pkg/log/logger.go` line 119 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fJl1C4Ze8hoXg1TNQ:-N4m7x7u10AdMKyiM2Hd:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/logger.go#L119)):*

isn't this supposed to write to stderr?
</blockquote></details>

Done.

acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2975)* status: all files reviewed, 3 unresolved discussions (waiting on @acud, @janos, @mrekucci, and @zelig)

___
*[`pkg/log/logger.go` line 93 at r2](https://reviewable.io/reviews/ethersphere/bee/2975#-N542vzZ51xZp5260LpP:-N542vzZ51xZp5260LpQ:bb0xkwv) ([raw file](https://github.com/ethersphere/bee/blob/eaf7ca4599599153bc1b9ddd245b3f669d8acdd2/pkg/log/logger.go#L93)):*
> ```Go
> 	if int(l.verbosity.get()) >= int(l.v) {
> 		if err := l.log(VerbosityDebug, CategoryDebug, nil, msg, keysAndValues...); err != nil {
> 			fmt.Fprintln(os.Stderr, err)
> ```

I think that only errors (and potentially warnings) should be printed to stderr and the rest (debug etc) should be in stdout, as they are not errors.

acud (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2975)* status: all files reviewed, 3 unresolved discussions (waiting on @acud, @janos, @mrekucci, and @zelig)

___
*[`pkg/log/logger.go` line 119 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fJl1C4Ze8hoXg1TNQ:-N543aAJ-YZonK9GYwDd:bik392u) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/logger.go#L119)):*

Done.
</blockquote></details>

i meant only errors, not the rest

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2975)* status: all files reviewed, 3 unresolved discussions (waiting on @acud, @janos, and @zelig)

___
*[`pkg/log/logger.go` line 119 at r1](https://reviewable.io/reviews/ethersphere/bee/2975#-N4fJl1C4Ze8hoXg1TNQ:-N54B18u1AaBgl9OlR63:b9thr77) ([raw file](https://github.com/ethersphere/bee/blob/d1f6497db64afc193aad5cdc4e25414761a166d4/pkg/log/logger.go#L119)):*

i meant only errors, not the rest
</blockquote></details>

See the comment above.
___
*[`pkg/log/logger.go` line 93 at r2](https://reviewable.io/reviews/ethersphere/bee/2975#-N542vzZ51xZp5260LpP:-N54AL4201QQQisO29Qo:b18fl72) ([raw file](https://github.com/ethersphere/bee/blob/eaf7ca4599599153bc1b9ddd245b3f669d8acdd2/pkg/log/logger.go#L93)):*

I think that only errors (and potentially warnings) should be printed to stderr and the rest (debug etc) should be in stdout, as they are not errors.
</blockquote></details>

This does not print a log message, but an error message when it failed to write a debug, info, warning, or error message to the sink.

acud (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/2975#-:-N59V2W04xjH1kUZRudR:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @janos and @zelig)

notanatol (approved):

pkg/api/logger.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// loggerSetVerbosityHandler sets logger(s) verbosity level based on
+// the specified expression or subsystem that matches the logger(s).
+func (s *Service) loggerSetVerbosityHandler(w http.ResponseWriter, r *http.Request) {
+	verbosity, err := log.ParseVerbosityLevel(mux.Vars(r)["verbosity"])
+	if err != nil {
+		s.logger.Debugf("loggers set verbosity: parse verbosity level failed: %v", err)
+		s.logger.Error("loggers set verbosity: parse verbosity level failed")
+		jsonhttp.BadRequest(w, err)
```
janos, line 114: Missing return.

pkg/api/logger.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	exp, err := base64.URLEncoding.DecodeString(mux.Vars(r)["exp"])
+	if err != nil {
+		s.logger.Debugf("loggers set verbosity: query parameter decoding failed: %v", err)
+		s.logger.Error("loggers set verbosity: query parameter decoding failed")
+		jsonhttp.BadRequest(w, err)
```
janos, line 121: Missing return.

pkg/api/logger.go
mrekucci: Good catch, fixed.

pkg/api/logger.go
mrekucci: Done.

pkg/log/log.go
```diff
@@ -0,0 +1,268 @@
...
+	// VerbosityDebug allows only error, warning, info and debug messages to be printed.
+	VerbosityDebug
+	// VerbosityAll allows to print all messages up to and including level V.
+	// It is a placeholder for the maximum possible V level verbosity within the logger.
+	VerbosityAll = Level(1<<31 - 1)
+)
+
+// Hook that is fired when logging
+// on the associated severity log level.
+// Note, the call must be non-blocking.
```
aloknerurkar, line 86: So although I see that there could be some use for this, I would reconsider having this hook functionality. This could be easily abused, given that this is called during the happy path. I think with the options and the builder we have enough ways to configure things.

aloknerurkar (approved):

janos (approved):

PR #3006, opened 2022-06-06
fix: check for SOC while doing traversal operations

### LINKED ISSUE #2999: Unpinning SOC fails with 500 Internal Server Error
1. I upload an SOC
- POST on `http://localhost:1633/soc/ef552ff649ebca68c37ddc28e25a27a55657c916/e1987b7e93640469a116a2327de50599afc2c977935ddd189259b540753b65e2?sig=8843051fefa23a4f6fca446ce7e66b730731119f52753bc7a13c8ad6717cbd244bad9280f41ca179b5b07b336f1e7a5a0f59c3b7340c14afe16a35ebef2630741c
` with `swarm-pin` header. this returns the reference `bd760322e2d9c86bf7346f1ca319315cc81e69584f77d30a449beee3c5b39fee`
2. Now if I call `DELETE` on `http://localhost:1633/pins/bd760322e2d9c86bf7346f1ca319315cc81e69584f77d30a449beee3c5b39fee` I get 
```
{
    "message": "Internal Server Error",
    "code": 500
}
``` 
with the above log.

If I pin the reference again and try `DELETE` I get the same response.

_Originally posted by @asabya in https://github.com/ethersphere/bee/issues/2713#issuecomment-1138485291_

aloknerurkar:
Fixes #2999 

Ideally, joiner and other file-related operations don't need to know about SOC. But the problem is we might call Traversal from the API layer with all kinds of chunks. We expect traversal to traverse any and all kinds. I feel this safeguard in Joiner would be the easiest way to understand and also prevent this type of use of Joiner in future.

The problem is when the problem occurs, we will see `storage: not found` errors which are false positives and we might end up wasting effort looking at all sorts of things, like I did here.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3006)

### DIFF
--- a/pkg/traversal/traversal.go
+++ b/pkg/traversal/traversal.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ethersphere/bee/pkg/file/loadsave"
 	"github.com/ethersphere/bee/pkg/manifest"
 	"github.com/ethersphere/bee/pkg/manifest/mantaray"
+	"github.com/ethersphere/bee/pkg/soc"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
@@ -54,6 +55,15 @@ func (s *service) Traverse(ctx context.Context, addr swarm.Address, iterFn swarm
 			return fmt.Errorf("traversal: iterate chunk address error for %q: %w", ref, err)
 		}
 		return nil
+	}
+
+	ch, err := s.store.Get(ctx, storage.ModeGetRequest, addr)
+	if err != nil {
+		return fmt.Errorf("traversal: failed to get root chunk %s: %w", addr.String(), err)
+	}
+	if soc.Valid(ch) {
+		// if this is a SOC, the traversal will be just be the single chunk
+		return iterFn(addr)
 	}
 
 	ls := loadsave.NewReadonly(s.store)
--- a/pkg/traversal/traversal_test.go
+++ b/pkg/traversal/traversal_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/ethersphere/bee/pkg/file/pipeline"
 	"github.com/ethersphere/bee/pkg/file/pipeline/builder"
 	"github.com/ethersphere/bee/pkg/manifest"
+	testingsoc "github.com/ethersphere/bee/pkg/soc/testing"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/storage/mock"
 	"github.com/ethersphere/bee/pkg/swarm"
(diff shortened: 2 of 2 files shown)


pkg/traversal/traversal.go
```diff
@@ -57,27 +57,22 @@ func (s *service) Traverse(ctx context.Context, addr swarm.Address, iterFn swarm
 	}
 
 	ls := loadsave.NewReadonly(s.store)
-	switch mf, err := manifest.NewDefaultManifestReference(addr, ls); {
-	case errors.Is(err, manifest.ErrInvalidManifestType):
-		break
-	case err != nil:
-		return fmt.Errorf("traversal: unable to create manifest reference for %q: %w", addr, err)
+	mf, _ := manifest.NewDefaultManifestReference(addr, ls)
```
janos, line 60: I am not sure if it is safe to remove handling of ErrInvalidManifestType error.

aloknerurkar: So currently this function cannot return an error. The New function only returns an error if the type is incorrect and `DefaultManifestReference` gives the Mantaray type. Also the `NewXXXReference` call doesn't do any get till other operations are done.

janos: I would even remove error return value from NewDefaultManifestReference to make that clear. But that is probably for a different PR.

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3006)* status: all files reviewed, 1 unresolved discussion (waiting on @acud, @aloknerurkar, and @zelig)

___
*[`pkg/file/joiner/joiner_test.go` line 948 at r1](https://reviewable.io/reviews/ethersphere/bee/3006#-N3usKmD4pIIVAFcOkWv:-N3usKmEApQhnKdQfTVP:b-w8mlfw) ([raw file](https://github.com/ethersphere/bee/blob/2f59b05f44600f7873a944116f4f7e329231489c/pkg/file/joiner/joiner_test.go#L948)):*
> ```Go
> 	_, _, err = joiner.New(ctx, store, sch.Address())
> 	if !errors.Is(err, joiner.ErrInvalidRootChunk) {
> 		t.Fatal("expected invalid root chunk error")
> ```

Would be nice to print the errors in the case of mismatch `t.Fatalf("New(...):\n\thave error: %v\n\twant error: %v", err, joiner.ErrInvalidRootChunk)`.

acud (approved):

aloknerurkar:
> @aloknerurkar as a sidenote I don't think that putting this logic in the joiner is the ideal solution. I would try to pull this out potentially. This can be done with a single `Get` operation on the root chunk when traversing and then doing the guesswork in the calling context rather than the joiner doing this. That being said, if you wish to continue in the current approach you are free to do so.

So the reason I put it in the joiner is that the joiner is not designed to work if the chunk is an SOC. If someone tries to do a Read on a joiner initialized with an SOC, this leads to unspecified behaviour. As I mentioned in the issue notes, if it finds some valid span header, it might actually try to Get addresses which are non-existent. Also, in this case, loadsave using the joiner was causing the problem. I feel if we do it here then we could potentially cut these problems at the source. Also, in the logs it will manifest like `storage: not found` which might lead us down some rabbithole.

mrekucci (approved):

PR #3011, opened 2022-06-09
fix: dont remove stampissuers on db nuke

acud:


notanatol (approved):

pkg/statestore/leveldb/migration.go
```diff
@@ -312,12 +312,17 @@ func deleteKeys(s *Store, keys []string) error {
...
-func (s *Store) Nuke() error {
+func (s *Store) Nuke(forgetStamps bool) error {
 	var (
 		keys     []string
 		prefixes = []string{"accounting", "pseudosettle", "swap", "non-mineable-overlay"}
 		err      error
 	)
+
+	if forgetStamps {
+		prefixes = append(prefixes, "postage")
```
istae, line 323: Aren't these prefixes the keys we SKIP when nuking?

aloknerurkar: so if the option is set we SHOULD forget, if not we should not... I think this is correct no?

aloknerurkar (approved):

pkg/statestore/leveldb/migration.go
istae: `collectKeysExcept` does not collect keys with these prefixes as the implies.

istae:
```
func collectKeysExcept(s *Store, prefix []string) (keys []string, err error) {
	if err := s.Iterate("", func(k, v []byte) (bool, error) {
		stk := string(k)
		has := false
		for _, v := range prefix {
			if strings.HasPrefix(stk, v) {
				has = true
				break
			}
		}
		if !has {
			keys = append(keys, stk)
		}
		return false, nil
	}); err != nil {
		return nil, err
	}
	return keys, nil
}
```

istae: then we call `deleteKeys` on these collected keys

istae: oh its' `dont remove stampissuers on db nuke`, got it, approved!

istae (approved):

pkg/statestore/leveldb/migration.go
acud: @istae is right. sorry. will fix!

significance:
❤️❤️❤️

PR #3012, opened 2022-06-09
feat: enable chainstore endpoint on debug

### LINKED ISSUE #2977: Debug API is not available during postage syncing
1.6.0

I wanted to use the `/chainstate` endpoint to see the status on postage batch syncing, but kept getting 404:

```
curl "http://localhost:1635/chainstate"
{"message":"Not Found","code":404}
```

After the sync was done, I executed the same command again:

```
curl "http://localhost:1635/chainstate"
{"chainTip":22293963,"block":22293955,"totalAmount":"22818108","currentPrice":"4"}
```

Expected:

Debug API (or at least the `/chainstate` endpoint) should be available during postage batch syncing

notanatol:
### Checklist

### Description
Closes #2977

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -236,6 +236,17 @@ func NewBee(interrupt chan os.Signal, addr string, publicKey *ecdsa.PublicKey, s
 		return nil, err
 	}
 
+	var unreserveFn func([]byte, uint8) (uint64, error)
+	var evictFn = func(b []byte) error {
+		_, err := unreserveFn(b, swarm.MaxPO+1)
+		return err
+	}
+
+	batchStore, err := batchstore.New(stateStore, evictFn, logger)
+	if err != nil {
+		return nil, fmt.Errorf("batchstore: %w", err)
+	}
+
 	addressbook := addressbook.New(stateStore)
 
 	var (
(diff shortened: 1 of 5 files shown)


pkg/node/node.go
```diff
@@ -236,6 +236,17 @@ func NewBee(interrupt chan os.Signal, addr string, publicKey *ecdsa.PublicKey, s
 		return nil, err
 	}
 
+	var unreserveFn func([]byte, uint8) (uint64, error)
+	var evictFn = func(b []byte) error {
+		_, err := unreserveFn(b, swarm.MaxPO+1)
+		return err
+	}
+
+	batchStore, err := batchstore.New(stateStore, evictFn, logger)
```
acud, line 245: This might cause some problems. If the node is started in a mode where there is no chain backend then we end up with this component dangling. It would be good to make sure that the component can finally shut down. Possibly also to initialize a mock of the component when the node needs to start without a chain backend

notanatol: replaced for a `noOp` implementation when in chainless mode, not sure what to do about the shutdown part...

mrekucci (approved):
> Reviewed 4 of 5 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3012)* status: all files reviewed, 1 unresolved discussion (waiting on @acud)

pkg/node/node.go
```diff
@@ -254,6 +255,20 @@ func NewBee(interrupt chan os.Signal, addr string, publicKey *ecdsa.PublicKey, s
 
 	chainEnabled := isChainEnabled(o, logger)
 
+	var batchStore postage.Storer = bsm.NoOp()
```
aloknerurkar, line 258: So I dont think we should be using mock code in production. This should be for testing only.

It would be better to define a Noop batchstore here or in a new package in postage.

aloknerurkar (changes requested):

pkg/node/node.go
notanatol: moved to a new package

PR #3022, opened 2022-06-10
fix: single point of shutdown

notanatol:
### Checklist

### Description
Listener triggering the shutdown sometimes leaves the main thread hanging, this PR replaces the pidKiller with a signal channel.


pkg/postage/listener/listener.go
```diff
@@ -327,19 +321,25 @@ func (l *listener) Listen(from uint64, updater postage.EventUpdater, initState *
 
 	go func() {
 		err := listenf()
+		defer func() {
```
acud, line 324: few question:
- whats the benefit of a `defer` statement here?
- naked writes are usually not recommended. the synced channel is only read off of during node startup, so if this fails later on you get a deadlock on shutdown

pkg/postage/listener/listener.go
```diff
@@ -327,19 +321,25 @@ func (l *listener) Listen(from uint64, updater postage.EventUpdater, initState *
...
 				return
 			}
 			l.logger.Errorf("failed syncing event listener, shutting down node err: %v", err)
-			if l.shutdowner != nil {
-				err = l.shutdowner.Shutdown()
-				if err != nil {
-					l.logger.Errorf("failed shutting down node: %v", err)
-				}
-			}
+			close(l.running) // trigger shutdown in start.go
```
acud, line 342: there are two `close` calls for this channel (one here and one on the actual shutdown method). this will panic at one point or another

notanatol: removed the leftover `close`

pkg/node/node.go
```diff
@@ -113,6 +112,7 @@ type Bee struct {
 	chainSyncerCloser        io.Closer
 	shutdownInProgress       bool
 	shutdownMutex            sync.Mutex
+	syncing                  chan struct{}
```
aloknerurkar, line 115: So I am not sure if we should name this `syncing`

I got confused while going through the PR about what this is meant for. Maybe `postageSyncingFailure` would be better? or along those lines?

notanatol: it's basically going to be open as long as there's a live listening goroutine, it's not related to failure handling, just blocking the main thread from exiting.
maybe can be named to something like `syncingInProgress`

aloknerurkar (approved):

pkg/node/node.go
aloknerurkar: I leave it up to you to rename this or not. Its just that we will in the code we will exit when this channel is closed. So it looks more like a signal to stop which is why I would prefer it to be some error type string.
`syncingStopped` maybe. But its up to you.

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3022)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on @acud and @aloknerurkar)

___
*[`pkg/node/node.go` line 115 at r2](https://reviewable.io/reviews/ethersphere/bee/3022#-N4DZs4yeOdkonjErdfC-r2-115:-N4WawBd3ws7htrCYgQC:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4d78df3c6c3d65fcb42f4f307d34735f67c1b02b/pkg/node/node.go#L115)):*

I leave it up to you to rename this or not. Its just that we will in the code we will exit when this channel is closed. So it looks more like a signal to stop which is why I would prefer it to be some error type string.
`syncingStopped` maybe. But its up to you.
</blockquote></details>

Done.

istae (approved):

Issue #3024, opened 2022-06-13
Add support for HTTP HEAD and OPTIONS methods


agazso:
#### Summary
During the EthPrague event there was a hack project to run the static wikipedia on swarm, based on the [static-wiki](https://github.com/segfall/static-wiki) project. That requires support for HEAD and OPTIONS methods.

#### Implementation
A workaround was that instead of adding this to Bee the support was added to the [gateway-proxy in this branch](https://github.com/ethersphere/gateway-proxy/compare/feat/head-method-support). 

Another problem was that Bee supports the `accept-encoding` header and browsers by default try to use `gzip` and pass in this header, but that breaks either the HEAD or the OPTIONS methods, so the proxy just removes this header.

agazso:
Related: #2709

acud:
@agazso do we still need `OPTIONS`?

agazso:
> @agazso do we still need `OPTIONS`?

For the static.wiki use case: yes.

PR #3027, opened 2022-06-16
feat(api): allow HEAD requests on the bytes endpoint

acud:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3027)

### DIFF
--- a/pkg/api/router.go
+++ b/pkg/api/router.go
@@ -15,7 +15,10 @@ import (
 	"github.com/ethersphere/bee/pkg/jsonhttp"
 	"github.com/ethersphere/bee/pkg/logging/httpaccess"
 	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/gorilla/handlers"
+	"github.com/gorilla/mux"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/sirupsen/logrus"
 	"resenje.org/web"
 )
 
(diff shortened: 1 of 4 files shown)


mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3027)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @acud)

___
*[`pkg/api/bytes.go` line 137 at r1](https://reviewable.io/reviews/ethersphere/bee/3027#-N4lHPpS6_VAc0VNmyxv:-N4lHPpS6_VAc0VNmyxw:bxym9it) ([raw file](https://github.com/ethersphere/bee/blob/b0fb5ad7bce91b1aa63fd9c6bcc678e387b4fab7/pkg/api/bytes.go#L137)):*
> ```Go
> 		logger.Debugf("bytes: parse address %s: %v", nameOrHex, err)
> 		logger.Error("bytes: parse address error")
> 		w.WriteHeader(400) // HEAD requests do not write a body
> ```

I'd recommend using predefined HTTP constants (to improve readability), like: `http.StatusBadRequest` instead of `400`, `http.StatusNotFound` instead of `404`, and `http.StatusOK` instead of `200` below.
___
*[`pkg/api/bytes.go` line 151 at r1](https://reviewable.io/reviews/ethersphere/bee/3027#-N4lJ-O63gTrRipesKwi:-N4lJ-O63gTrRipesKwj:b-3oos8z) ([raw file](https://github.com/ethersphere/bee/blob/b0fb5ad7bce91b1aa63fd9c6bcc678e387b4fab7/pkg/api/bytes.go#L151)):*
> ```Go
> 	w.Header().Add("Content-Type", "application/octet-stream")
> 	span := int64(binary.LittleEndian.Uint64(ch.Data()[:swarm.SpanSize]))
> 	w.Header().Add("Content-Length", fmt.Sprintf("%d", span))
> ```

Using `strconv.FormatUint(span, 10)` would be more effective; also no need for the `in64` casting above.

agazso:
Can you please make sure that the data is not gzipped when you send back the response?

I found out that sending gzipped response to the `HEAD` and `OPTIONS` requests break applications (as mentioned in https://github.com/ethersphere/bee/issues/3024).

acud:
> Can you please make sure that the data is not gzipped when you send back the response?
> 
> I found out that sending gzipped response to the `HEAD` and `OPTIONS` requests break applications (as mentioned in #3024).

So I'm not sure what you're asking me for... Please make a distinction between the nginx proxy between the node and the user, it might do the compression anyway depending on the headers the client may attach to the request. The bee node does not compress the data.

ldeffenb:
> The bee node **does** compress the data in case there is an `Accept-encoding: gzip` header passed to it and it causes problems with the `HEAD` requests.

Seems easy enough to me.  Just make sure you don't put the offending header on.  That means it's the client's choice and server only complies with what is considered acceptable.

agazso:
> Seems easy enough to me. Just make sure you don't put the offending header on. That means it's the client's choice and server only complies with what is considered acceptable.

There are a bunch of problems with this, so trying to enumerate them:

1. I made a mistake in my previous examples, because I did not use `HEAD` requests. So the correct example is this (please note the `-I` switch for curl that make it a `HEAD` request).

This is the `HEAD` request without the `Accept-encoding` header:
```
curl -v http://localhost:1633/bytes/aaf55d27c4597c5b9a9ab569e278af555a4391eb4f8ec6a31bc75685331c6966 -I
...output omitted...
Content-Length: 43240423424
```

This returns the correct length as `43240423424`.

If you pass the `Accept-encoding` header:
```
curl -v -H 'Accept-encoding: gzip' http://localhost:1633/bytes/aaf55d27c4597c5b9a9ab569e278af555a4391eb4f8ec6a31bc75685331c6966 -I
...output omitted...
Content-Length: 23
```

So the issue is that when the `Accept-encoding` header is used then Bee returns an obviously invalid `Content-length`!

2. The most often used use-case of the `HEAD` support is to get the size of a resource without downloading the whole resource. The problem is that browsers by default support gzip encoding so by default they add the `Accept-encoding: gzip ` header to requests by default, including `HEAD` requests, but that defeats the purpose of using such request. 

3. It is also not trivial to remove the `Accept-encoding` header, at least I could not figure out how to do it. This is especially a problem if you are using a library that uses `HEAD` requests, maybe even coming from another dependency. 

Because of these issues, my suggestion would be to ignore the `Accept-encoding` header in case of `HEAD` requests (which is a perfectly compliant thing to do according to the HTTP standard).

agazso:
> > Seems easy enough to me. Just make sure you don't put the offending header on. That means it's the client's choice and server only complies with what is considered acceptable.

> 3. It is also not trivial to remove the `Accept-encoding` header, at least I could not figure out how to do it. This is especially a problem if you are using a library that uses `HEAD` requests, maybe even coming from another dependency.

It turns out that it is impossible to remove certain headers from the requests made by the browser and `Accept-encoding` is one of those headers:
https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name

acud:
right... I'll look into this :+1:

acud:
so @agazso I looked into it and indeed the golang HTTP server does the compression under-the-hood when attaching the header from the client. I have no control over this and cannot switch this on and off. Any setting to `Content-Length` headers from the request handler is ineffectual since the HTTP server does the compression under the hood and replaces the content length header with the size after compression. My assumption is that `23 bytes` is an empty gzip stream that the server generates. This is probably an edge case in the golang HTTP server and I'll open an issue in their repo about this.

My guess would be that the following behavior should be the correct one (and probably the best we can get to right now without: A. writing our own HTTP server or B. migrate the entire HTTP API to a server implementation to one that would allow us control over whether we compress data back when the header exists, since intentional ignoring of the header would be needed):
- get the size of the data we are about to write:
- - if it's a range query - get the range
- - for normal query it is the whole thing
- check the bounds of the request in case it is a range query and make sure we're not out of bound. if we are, the content size is `<size of byte stream>-<beginning of range query>`
- set the return header
- - if the accept encoding header is set, set `Decompressed-Content-Length` header to be the size of the returned content
- - if it is a normal request - set the `Content-Length` header

Let me know if this makes sense and if so I can continue implementing it

pkg/api/router.go
```diff
@@ -122,6 +122,11 @@ func (s *Service) mountAPI() {
 			s.newTracingHandler("bytes-download"),
 			web.FinalHandlerFunc(s.bytesGetHandler),
 		),
+		"HEAD": web.ChainHandlers(
+			s.contentLengthMetricMiddleware(),
```
aloknerurkar, line 126: So `contentLengthMetricMiddleware` seems to be like a no-op for HEAD requests?

notanatol (approved):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3027)* status: all files reviewed, 1 unresolved discussion (waiting on @acud, @aloknerurkar, and @mrekucci)

mrekucci (approved):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3027)* status: all files reviewed, 1 unresolved discussion (waiting on @acud and @aloknerurkar)

aloknerurkar (approved):

PR #3036, opened 2022-06-22
fix: pushsync perpetual loop when all peers are unavailable

### LINKED ISSUE #2946: Direct upload gets stuck forever
### Context
Uploading a file with the `swarm-deferred-upload` header set to `false` (also known as direct upload) may get stuck forever depending on the data to be uploaded. Uploading the same data with `swarm-deferred-upload` set to `true` (which is the default) succeeds and it seems to be cached after, because once it was successfully uploaded then the following uploads are going to be successful no matter how the header is set. 

### Steps to reproduce
There is an attached zip file which contains a file, which has 16384 bytes size and has `4d36fd00cd3d1eb376c2633a10a27287c90f48e696e401f8dec23548031a6f16` as content-hash.

Use the contained file and upload it to the `/bytes` endpoint with the `swarm-deferred-upload` header set to false:
```
curl --data-binary @4d36fd00cd3d1eb376c2633a10a27287c90f48e696e401f8dec23548031a6f16.bin -H "swarm-deferred-upload: false" -H "swarm-postage-batch-id: <use-your-postage-batch-id>" http://localhost:1633/bytes
```
[4d36fd00cd3d1eb376c2633a10a27287c90f48e696e401f8dec23548031a6f16.zip](https://github.com/ethersphere/bee/files/8662972/4d36fd00cd3d1eb376c2633a10a27287c90f48e696e401f8dec23548031a6f16.zip)

mrekucci:
### Checklist

### Description
This PR fixes the perpetual pushsync loop when all peers are unavailable, in which case it will attempt to contact the peers a predefined number of times with a predefined wait time between attempts and terminate when it fails.

### Related Issue (Optional)
Closes #2946 

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3036)

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -610,26 +610,28 @@ func (s *Service) newStamperPutter(r *http.Request) (storage.Storer, func() erro
 		p, err := newStoringStamperPutter(s.storer, s.post, s.signer, batch)
 		return p, noopWaitFn, err
 	}
-	p, err := newPushStamperPutter(s.storer, s.post, s.signer, batch, s.chunkPushC)
+	p, err := newPushStamperPutter(s.storer, s.logger, s.post, s.signer, batch, s.chunkPushC)
 	return p, p.eg.Wait, err
 }
 
 type pushStamperPutter struct {
 	storage.Storer
+
+	logger  logging.Logger
 	stamper postage.Stamper
 	eg      errgroup.Group
 	c       chan *pusher.Op
 	sem     chan struct{}
 }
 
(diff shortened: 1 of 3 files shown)


acud:
> terminate when it fails

can you explain this in a bit more detail? what does this mean?

mrekucci:
> > terminate when it fails
> 
> can you explain this in a bit more detail? what does this mean?

E.g.: in the case of upload, this will result in an error. Basically, this means that the perpetual pushsync loop will be terminated (the chunk(s) would not be resent on errors).

mrekucci:
Closed in favor of #3041

PR #3041, opened 2022-06-23
fix: timeout stuck upload on bytes endpoint

### LINKED ISSUE #2946: Direct upload gets stuck forever

mrekucci:
### Checklist

### Related Issue (Optional)
Closes #2946

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3041)

### DIFF
--- a/pkg/api/bytes.go
+++ b/pkg/api/bytes.go
@@ -5,11 +5,13 @@
 package api
 
 import (
+	"context"
 	"encoding/binary"
 	"errors"
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/ethersphere/bee/pkg/cac"
 	"github.com/ethersphere/bee/pkg/jsonhttp"
@@ -19,6 +21,7 @@ import (
 	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/ethersphere/bee/pkg/tags"
 	"github.com/ethersphere/bee/pkg/tracing"
+	"github.com/ethersphere/bee/pkg/util/ioutil"
 	"github.com/gorilla/mux"
 )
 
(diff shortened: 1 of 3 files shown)


notanatol (approved):

pkg/api/util.go
```diff
@@ -0,0 +1,68 @@
...
+package api
+
+import (
+	"context"
+	"io"
+	"time"
+)
+
+// ProgressReader tracks the progress on io.Reader Read method.
+type ProgressReader struct {
```
janos, line 14: I would suggest not to export types and functions unless that is required. Currently, ProgressReader and its constructor is only used in the api package. Maybe there is even a better place for ProgressReader than api package, as it is more general type, not specific to the api package.

janos (approved):

acud (approved):
LGTM minus the comment from @janos

pkg/api/util.go
mrekucci: Agree, I exported the types and functions because of the `api_test` package. Moved to its own general package.

aloknerurkar (approved):
Nicely done!

istae (approved):

PR #3055, opened 2022-06-29
feat: depth reporting

istae:
### Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3055)

### DIFF
--- a/pkg/depthmonitor/depthmonitor.go
+++ b/pkg/depthmonitor/depthmonitor.go
@@ -7,12 +7,14 @@ import (
 
 	"github.com/ethersphere/bee/pkg/logging"
 	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/topology"
 )
 
 const (
-	depthKey         string  = "storage_depth"
-	adaptationWindow float64 = 60 * 60
-	manageWait               = 5 * time.Minute
+	depthKey                  string  = "storage_depth"
+	adaptationWindowSeconds   float64 = 60 * 60
+	adaptationRollbackSeconds         = 5
+	manageWait                        = 5 * time.Minute
 )
 
 // ReserveReporter interface defines the functionality required from the local storage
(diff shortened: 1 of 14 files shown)


pkg/depthmonitor/depthmonitor.go
```diff
@@ -128,60 +134,44 @@ func (s *Service) manage(warmupTime time.Duration) {
...
 			adaptationPeriod = true
 			adaptationStart = time.Now()
 		}
 
-		// if we have crossed 50% utilization, dont do anything
-		if !reserveNotHalfFull {
-			adaptationPeriod = false
+		// if we have crossed the adaptation window, roll it back a little to allow sync to fill the reserve
+		if time.Since(adaptationStart).Seconds() > adaptationWindowSeconds {
+			adaptationStart = time.Now().Add(-time.Second * adaptationRollbackSeconds)
```
aloknerurkar, line 155: I dont understand why we need this rollback? 5 seconds are not going to affect the calculations, so I think we should not add this new rollback concept, instead just start a new adaptation period and let the next part use the entire new window to make the calculations.

We can just add a case to the if loop above.

pkg/depthmonitor/depthmonitor.go
```diff
@@ -128,60 +134,44 @@ func (s *Service) manage(warmupTime time.Duration) {
...
-		if adaptationPeriod && isNotHalfFull(expectedSize, capacity) {
+		if isNotHalfFull(expectedSize) {
 			s.depthLock.Lock()
 			s.storageDepth--
-			s.topo.SetRadius(s.storageDepth)
-			s.logger.Infof("depthmonitor: reducing storage depth to better utilize reserve, new depth: %d", s.storageDepth)
+			s.topology.SetStorageDepth(s.storageDepth)
 			s.depthLock.Unlock()
-			adaptationPeriod = false
+			s.logger.Infof("depthmonitor: reducing storage depth to better utilize reserve, new depth: %d", s.storageDepth)
```
aloknerurkar, line 170: The reason for this reset was that once we change the depth and allow the protocols to do their thing for `manageWait` time. Then, we should let the loop again determine if we need a new adaptation period. If we do, we should ideally calculate the new expectedSize based on the entire window and not just the remaining time in the window. This takes into account that the depth has already changed once.

If there is need to decrease further, I think we will anyway end up in another adaptation period and then check the expected size. If this is improved then no need to further change anything.

Otherwise, we would aggressively decrease the depth and then might need to increase it again causing some oscillation.

PR #3081, opened 2022-07-08
fix: improve bootstrapper reliability

istae:
### Checklist

We increase the timeout of the context for fetching the snapshot to 2min because the previous 30s is too aggressive. PR also includes some simple restructuring of the retry mechanism.  

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3081)

### DIFF
--- a/pkg/node/bootstrap.go
+++ b/pkg/node/bootstrap.go
@@ -58,7 +58,7 @@ var (
 const (
 	getSnapshotRetries = 5
 	retryWait          = time.Second * 10
-	timeout            = time.Minute * 2
+	timeout            = time.Minute
 )
 
 func bootstrapNode(
@@ -91,9 +91,6 @@ func bootstrapNode(
 	if err != nil {
 		return nil, fmt.Errorf("tracer: %w", err)
 	}
-
-	passedInLogger := logger
-	logger = logging.New(io.Discard, 0)
 
 	p2pCtx, p2pCancel := context.WithCancel(context.Background())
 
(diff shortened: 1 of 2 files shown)


pkg/node/bootstrap.go
```diff
@@ -214,7 +219,12 @@ func bootstrapNode(
 		eventsJSON        []byte
 	)
 
+	logger = passedInLogger
```
acud, line 222: i find the logger swapping confusing... why not toggle which logger to use by either passing the discarding logger or just `logger` in the constructor like it was before?

pkg/node/bootstrap.go
```diff
@@ -214,7 +219,12 @@ func bootstrapNode(
...
+
 	for i := 0; i < getSnapshotRetries; i++ {
+		if err != nil {
+			time.Sleep(retryWait)
```
acud, line 226: why do you think that retrying would make any difference? the snapshot is 1mb... we should not even run into any time-settlements issues afaik but i might be wrong

pkg/node/bootstrap.go
```diff
@@ -227,6 +237,9 @@ func bootstrapNode(
 	}
 
 	for i := 0; i < getSnapshotRetries; i++ {
+		if err != nil {
+			time.Sleep(retryWait)
```
acud, line 241: ditto

pkg/node/bootstrap.go
```diff
@@ -261,7 +274,7 @@ func bootstrapNode(
 
 // wait till some peers are connected. returns true if all is ok
 func waitPeers(kad *kademlia.Kad) bool {
-	for i := 0; i < 60; i++ {
+	for i := 0; i < 2*60; i++ { // max 2 minutes
```
acud, line 277: it might be useful to understand if this is indeed the underlying issue that is creating the instability. 2 minutes is a lot of time to establish connections to 25 peers. there might be some other underlying issue that is causing the bootstrapper instability.

acud (commented):
it would be nice to know what is causing the bootstrapper instability... it appears we are adding sleeps and longer timeouts when in fact the process should be very fast and seamless, i.e. we're not solving the problem. I would recommend to add simple instrumentation to know _why_ the context deadline exceeded is reached i.e. break down the operations to -> log how long it took to establish the connections in kademlia and log how long it took to get the snapshot from the network. rinse and repeat. once the measurements are there we know whats the problem and fix it (potentially).

pkg/node/bootstrap.go
istae: this is not needed 👍

pkg/node/bootstrap.go
```diff
@@ -56,7 +56,9 @@ var (
 )
 
 const (
-	getSnapshotRetries = 3
+	getSnapshotRetries = 5
+	retryWait          = time.Second * 10
+	timeout            = time.Minute * 2
```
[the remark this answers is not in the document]
istae: this timeout starts after peer connection

acud:
> > it would be nice to know what is causing the bootstrapper instability... it appears we are adding sleeps and longer timeouts when in fact the process should be very fast and seamless, i.e. we're not solving the problem. I would recommend to add simple instrumentation to know _why_ the context deadline exceeded is reached i.e. break down the operations to -> log how long it took to establish the connections in kademlia and log how long it took to get the snapshot from the network. rinse and repeat. once the measurements are there we know whats the problem and fix it (potentially).
> 
> I agree with you. From @agazso 's logs, the issue was the 30 second download context timeout, which we increase to 2 min with this PR, which is the main fix. The other changes are just hardenings to potentially improve the reliability of the whole thing.

It would be even better to know what exactly with the download is causing the error. Is it the feed lookup? is it the actual download?

I would do the following:
- add log for feed lookup time
- add log for kademlia buildup time
- add log for download error/success time

run the whole thing in a loop over and over again and then figure out where's the problem. if its a feed lookup problem then we should probably just hardcode the latest update address and get rid of the feed for now. if it's the kademlia which is acting up then we should figure out why, etc...

istae:
With the increased timeout, dev-bee-storage bee-0 was able to retrieve the snapshot 25 times successfully with no errors.

agazso:
> With the increased timeout, dev-bee-storage bee-0 was able to retrieve the snapshot 25 times successfully with no errors.

Did you stop Bee and delete the data dir between subsequent runs? Otherwise it is possible that it was successful for the first time and then it was successful because it returned the chunks from the cache.

Usually when I debug it I start from scratch in ultra-light mode, because that is also what is happening in the Desktop app.

istae:
@agazso The localstore was dumped in each run. The initial fetch took 1m33s, and slowly went down each time to a low ~45s, which I presume happened because the retrieval cache of the network was warmed up with the chunks of the snapshot.

pkg/node/bootstrap.go
```diff
@@ -202,61 +203,55 @@ func bootstrapNode(
...
-	var (
-		snapshotReference swarm.Address
-		reader            file.Joiner
-		l                 int64
-		eventsJSON        []byte
-	)
-
-	for i := 0; i < getSnapshotRetries; i++ {
-		snapshotReference, err = getLatestSnapshot(ctx, ns, snapshotFeed)
+		snapshotReference, err := getLatestSnapshot(ctx, ns, snapshotFeed)
```
aloknerurkar, line 212: If we got past this stage in the first try, we should ideally not do this again. The reference will be the same and there might be more lookups triggered because of this.

Maybe we can separate these two operations and retry each one separately if they fail.

aloknerurkar (changes requested):

pkg/node/bootstrap.go
aloknerurkar: So I am guessing the difference could be that we have some more peers connected?

aloknerurkar (approved):

notanatol (approved):

PR #3103, opened 2022-07-20
fix: gofmt replace interface with any

notanatol:
### Checklist

### Description
In the spirit of go1.18 the old `interface{}` becomes `any`.

It also looks like there's a change in syntax of the build tags, I think it's good to keep up with these changes.

commands ran:

```
go fix ./pkg/... ./cmd/... 
gofmt -r 'interface{} -> any' -w .
```

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3103)

### DIFF
--- a/cmd/bee/cmd/start_windows.go
+++ b/cmd/bee/cmd/start_windows.go
@@ -3,7 +3,6 @@
 // license that can be found in the LICENSE file.
 
 //go:build windows
-// +build windows
 
 package cmd
 
--- a/cmd/bee/cmd/start_windows.go
+++ b/cmd/bee/cmd/start_windows.go
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+//go:build windows
 // +build windows
 
 package cmd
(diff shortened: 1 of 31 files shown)


cmd/bee/cmd/start_windows.go
```diff
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+//go:build windows
```
acud, line 5: is this needed?

acud (approved):
LGTM minus the build flag change?

cmd/bee/cmd/start_windows.go
notanatol: It seems that starting go1.18 the syntax for build constraints has changed https://pkg.go.dev/cmd/go#hdr-Build_constraints. It's likely that the old syntax will be phased out in the future so we need to get on-board at some point, might as well be now. @vandot  wdyt?

notanatol:
```
In Go 1.18, go fix [go tool fix] now removes the now-obsolete // +build lines in modules declaring go 1.18 or later in their go.mod files.
```

istae (approved):

cmd/bee/cmd/start_windows.go
vandot: As @notanatol said, it started in go1.17 and probably they will deprecate old syntax at some point

acud (approved):

PR #3125, opened 2022-07-27
chore: remove global var from blocklist,breaker packages

vladopajic:
### Checklist

### Description
`blocklist` and `breaker` packages had a hacky way for mocking current time in tests using global variable. 
This PR provides cleaner solution for mocking current time in tests.

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3125)

### DIFF
--- a/pkg/p2p/libp2p/internal/breaker/breaker.go
+++ b/pkg/p2p/libp2p/internal/breaker/breaker.go
@@ -21,9 +21,6 @@ const (
 var (
 	_ Interface = (*breaker)(nil)
 
-	// timeNow is used to deterministically mock time.Now() in tests.
-	timeNow = time.Now
-
 	// ErrClosed is the special error type that indicates that breaker is closed and that is not executing functions at the moment.
 	ErrClosed = errors.New("breaker closed")
 )
@@ -38,6 +35,8 @@ type Interface interface {
 	ClosedUntil() time.Time
 }
 
+type currentTimeFn = func() time.Time
+
 type breaker struct {
 	limit                int // breaker will not execute any more tasks after limit number of consecutive failures happen
 	consFailedCalls      int // current number of consecutive fails
@@ -46,6 +45,7 @@ type breaker struct {
 	backoff              time.Duration // initial backoff duration
 	maxBackoff           time.Duration
 	failInterval         time.Duration // consecutive failures are counted if they happen within this interval
+	currentTimeFn        currentTimeFn
 	mtx                  sync.Mutex
 }
 
@@ -57,11 +57,16 @@ type Options struct {
 }
 
 func NewBreaker(o Options) Interface {
+	return newBreakerWithCurrentTimeFn(o, time.Now)
+}
+
+func newBreakerWithCurrentTimeFn(o Options, currentTimeFn currentTimeFn) Interface {
 	breaker := &breaker{
-		limit:        o.Limit,
-		backoff:      o.StartBackoff,
-		maxBackoff:   o.MaxBackoff,
-		failInterval: o.FailInterval,
+		limit:         o.Limit,
+		backoff:       o.StartBackoff,
+		maxBackoff:    o.MaxBackoff,
+		failInterval:  o.FailInterval,
+		currentTimeFn: currentTimeFn,
 	}
 
 	if o.Limit == 0 {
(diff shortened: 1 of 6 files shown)


mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3125)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @vladopajic)

___
*[`pkg/p2p/libp2p/internal/blocklist/blocklist.go` line 18 at r1](https://reviewable.io/reviews/ethersphere/bee/3125#-N7zLpERDiq7WFG3nH5p:-N7zLpERDiq7WFG3nH5q:bjzb5vk) ([raw file](https://github.com/ethersphere/bee/blob/374674194e40d4c794047ca16d50258de9158369/pkg/p2p/libp2p/internal/blocklist/blocklist.go#L18)):*
> ```Go
> var keyPrefix = "blocklist-"
> 
> type CurrentTimeFn = func() time.Time
> ```

This type doesn't need to be exported.
___
*[`pkg/p2p/libp2p/internal/blocklist/blocklist.go` line 29 at r1](https://reviewable.io/reviews/ethersphere/bee/3125#-N7zLP-wAGkbDWPjVJfi:-N7zLP-wAGkbDWPjVJfj:b-ol126o) ([raw file](https://github.com/ethersphere/bee/blob/374674194e40d4c794047ca16d50258de9158369/pkg/p2p/libp2p/internal/blocklist/blocklist.go#L29)):*
> ```Go
> }
> 
> func NewBlocklistWithCurrentTimeFn(store storage.StateStorer, currentTimeFn CurrentTimeFn) *Blocklist {
> ```

Since this function is leveraged only in tests I'd recommend putting it into the `export_test.go` file.

notanatol (approved):

vladopajic:
@mrekucci 

1) I've changed type so it's not being exported anymore.

2) I don't think that moving `NewBlocklistWithCurrentTimeFn` function to separate file is necessary and furthermore I think that this practice makes bad code overall. Here are few cases:
- Mainly because code of specific type may end up scatters in different files witch makes  difficult to read all functionality that this type has.
- In large packages this practice may end up with `export_test.go` file holding vary different and unrelatable code. 
- Any code should be designed uniformly whether it's going to be used in test, production, etc. 
- What happens when user invokes method in non test files that is defined in `export_test.go` file? Do you have procedures in place that will forbid this or make lint error?

mrekucci (commented):
> Mainly because code of specific type may end up scatters in different files witch makes difficult to read all functionality that this type has.

It won't, this function will end up in a specific file named `export_test.go`, which is a common practice to utilize this file when you don't want to mix public API with test only API. I recommend you to check for this file in the Golang repository: https://github.com/golang/go/find/master

> In large packages this practice may end up with export_test.go file holding vary different and unrelatable code.

The code in the `export_test.go` file is related to the (extended) test API of the package that contains this file.

> Any code should be designed uniformly whether it's going to be used in test, production, etc.

That doesn't mean it cannot live in the `export_test.go` file.

> What happens when user invokes method in non test files that is defined in export_test.go file? Do you have procedures in place that will forbid this or make lint error?

That won't happen since `export_test.go` file will only be included when you run `go test`, so it does not pollute your public API, and users never access it.

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3125#-:-N7zffn0408a_TOhrkSH:be4lz5w)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @vladopajic)

vladopajic:
Thanks for clarification @mrekucci, I have updated PR.

mrekucci (approved):
No problem, thanks for the PR!

> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3125#-:-N8--5VSAWvrbeaES99l:b2wrwgr)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @vladopajic)

___
*[`pkg/p2p/libp2p/internal/blocklist/export_test.go` line 17 at r3](https://reviewable.io/reviews/ethersphere/bee/3125#-N7zzmPKAK4n1MGvG2KP:-N7zzmPKAK4n1MGvG2KQ:b-d3j9n4) ([raw file](https://github.com/ethersphere/bee/blob/d9349030845e01cd7d7fc01e107cc9362c25c0dd/pkg/p2p/libp2p/internal/blocklist/export_test.go#L17)):*
> ```Go
> 		currentTimeFn: currentTimeFn,
> 	}
> }
> ```

Missing newline at the end of the file.

aloknerurkar (approved):
Thanks for this!

vladopajic:
Gents could you review again? I have made similar changes in the `breaker` package as well.

aloknerurkar:
> Gents could you review again? I have made similar changes in the `breaker` package as well.

Next time, it would be better to do these changes in a separate PR. Usually, if PR is approved we don't add stuff to it unless it's some issue we found etc.

pkg/p2p/libp2p/internal/breaker/breaker.go
```diff
@@ -46,22 +45,25 @@ type breaker struct {
...
 type Options struct {
-	Limit        int
-	FailInterval time.Duration
-	StartBackoff time.Duration
-	MaxBackoff   time.Duration
+	Limit         int
+	FailInterval  time.Duration
+	StartBackoff  time.Duration
+	MaxBackoff    time.Duration
+	CurrentTimeFn currentTimeFn
```
aloknerurkar, line 57: This is not right. This will expose this to all users of this package, which is not what we want.

This should be handled similar to the blocklist change. You can have a new constructor which initializes this timeFunc in export_test file. This will only make it available to the tests.

Issue #3094, opened 2022-07-19
Feature request: light node without postage syncing


agazso:
#### Summary
It would be good to have a light node without postage syncing, because it is what takes the most time during Desktop setup and also it is error prone. 

Also when using the Desktop app the postage syncing is not needed, because it does not store chunks and it does not have initial postage stamps either.

notanatol:
The most intuitive way to achieve this is to have boolean flag that we can toggle to enable/disable postage syncing.

We just need to decide on the default behavior, what do you tihink?

notanatol:
> If there are no downsides we can try this. As far as I know there is already a config variable that enables the postage snapshot syncing (`use-postage-snapshot`). Then this logic would disable that altogether, right?

that is correct, this new flag would disable syncing altogether, but if enabled - the `use-postage-snapshot` would be responsible for fetching the snapshot from the network or syncing from scratch.

Question - let's say
```
postage-syncing-enable: false
use-postage-snapshot: true
```
should we error because of conflicting flags or override the latter and log a message "ignoring postage snapshot flag because syncing is disabled"?

also @acud wdyt?

agazso:
Ultra light node by definition cannot have the syncing enabled, because that would require `chain-enabled` to be `true`.

Regarding the config values logic: I would ignore postage snapshot if `postage-syncing-enable` is `true` and log a message, just how you suggested.

Btw I would prefer to have consistent naming in the config variables. I have seen variations of `*-enable` and I think it is not necessary at all, e.g. `postage-syncing: <true|false>` conveys the same meaning and that would be my preference. However that is also not consistent with the existing `use-postage-snapshot`, so for the sake of consistency we can consider `use-postage-syncing`.

PR #3127, opened 2022-07-27
feat(postage): background postage sync for light node

notanatol:
### Checklist

### Description
Foreground postage batch syncing results in a bad user experience because of the time it takes to complete.
This PR moves it to a background process - and while it's ongoing - the relevant endpoints will return `503`

Closes: #3094

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3127)

### DIFF
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -151,8 +151,14 @@ func (c *command) initStartCmd() (err error) {
 
 			// Wait for termination or interrupt signals.
 			// We want to clean up things at the end.
-			interruptChannel := make(chan os.Signal, 1)
-			signal.Notify(interruptChannel, syscall.SIGINT, syscall.SIGTERM)
+			sysInterruptChannel := make(chan os.Signal, 1)
+			signal.Notify(sysInterruptChannel, syscall.SIGINT, syscall.SIGTERM)
+
+			interruptChannel := make(chan struct{})
+			go func() {
+				<-sysInterruptChannel
+				close(interruptChannel)
+			}()
 
 			b, err := node.NewBee(interruptChannel, c.config.GetString(optionNameP2PAddr), signerConfig.publicKey, signerConfig.signer, networkID, logger, signerConfig.libp2pPrivateKey, signerConfig.pssPrivateKey, &node.Options{
 				DataDir:                    c.config.GetString(optionNameDataDir),
@@ -213,8 +219,8 @@ func (c *command) initStartCmd() (err error) {
 				start: func() {
 					// Block main goroutine until it is interrupted or stopped
 					select {
-					case sig := <-interruptChannel:
-						logger.Debugf("received signal: %v", sig)
+					case <-interruptChannel:
+						logger.Debug("received interrupt signal")
 					case <-b.SyncingStopped():
 					}
 
@@ -234,8 +240,8 @@ func (c *command) initStartCmd() (err error) {
 					// If shutdown function is blocking too long,
 					// allow process termination by receiving another signal.
 					select {
-					case sig := <-interruptChannel:
-						logger.Debugf("received signal: %v", sig)
+					case <-interruptChannel:
+						logger.Debug("received interrupt signal")
 					case <-done:
 					}
 				},
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -155,10 +155,6 @@ func (c *command) initStartCmd() (err error) {
 			signal.Notify(sysInterruptChannel, syscall.SIGINT, syscall.SIGTERM)
 
 			interruptChannel := make(chan struct{})
-			go func() {
-				<-sysInterruptChannel
-				close(interruptChannel)
-			}()
 
 			b, err := node.NewBee(interruptChannel, c.config.GetString(optionNameP2PAddr), signerConfig.publicKey, signerConfig.signer, networkID, logger, signerConfig.libp2pPrivateKey, signerConfig.pssPrivateKey, &node.Options{
 				DataDir:                    c.config.GetString(optionNameDataDir),
@@ -219,8 +215,9 @@ func (c *command) initStartCmd() (err error) {
 				start: func() {
 					// Block main goroutine until it is interrupted or stopped
 					select {
-					case <-interruptChannel:
+					case <-sysInterruptChannel:
 						logger.Debug("received interrupt signal")
+						close(interruptChannel)
 					case <-b.SyncingStopped():
 					}
 
@@ -240,7 +237,7 @@ func (c *command) initStartCmd() (err error) {
 					// If shutdown function is blocking too long,
 					// allow process termination by receiving another signal.
 					select {
-					case <-interruptChannel:
+					case <-sysInterruptChannel:
 						logger.Debug("received interrupt signal")
 					case <-done:
 					}
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -93,7 +93,6 @@ type testServerOptions struct {
 	PostageContract    postagecontract.Interface
 	Post               postage.Service
 	Steward            steward.Interface
-	BatchSyncStatus    postage.SyncStatus
 	WsHeaders          http.Header
 	Authenticator      *mockauth.Auth
 	DebugAPI           bool
@@ -111,9 +110,11 @@ type testServerOptions struct {
 	AccountingOpts  []accountingmock.Option
 	ChequebookOpts  []chequebookmock.Option
 	SwapOpts        []swapmock.Option
-	BatchStore      postage.Storer
 	TransactionOpts []transactionmock.Option
 	Traverser       traversal.Traverser
+
+	BatchStore        postage.Storer
+	BatchEventUpdater postage.SyncStatus
 
 	BackendOpts []backendmock.Option
 	Erc20Opts   []erc20mock.Option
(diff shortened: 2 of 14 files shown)


pkg/node/node.go
```diff
@@ -696,24 +696,49 @@ func NewBee(interrupt chan os.Signal, addr string, publicKey *ecdsa.PublicKey, s
...
+				}
+			case <-interrupt:
+				return nil, ErrInterruped
 			}
-		case <-interrupt:
-			return nil, ErrInterruped
+		} else {
+			go func() {
+				logger.Info("started postage contract data sync in the background...")
+				syncedChan, err := batchSvc.Start(postageSyncStart, initBatchState)
```
acud, line 725: this is almost identical to the code block above. i'd refactor this into a local, function scoped function (i.e. `syncFn := func() { ....}`) and just call it either from the goroutine or from the main thread.

pkg/postage/interface.go
```diff
@@ -11,9 +11,16 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
+// SyncStatus interface definitions reflect the status of the postage syncing progress.
+type SyncStatus interface {
+	Get() (bool, error)
```
acud, line 16: `GetSyncStatus` otherwise this is just a getter setter naming which is not very comprehensible

pkg/postage/interface.go
```diff
@@ -11,9 +11,16 @@ import (
...
+	Set(error)
```
acud, line 17: `SetError` maybe but this is also a bit bizarre... an external component can set the state of an internal component via getter/setter?

pkg/node/node.go
```diff
@@ -696,24 +696,49 @@ func NewBee(interrupt chan os.Signal, addr string, publicKey *ecdsa.PublicKey, s
...
 
-		// arguably this is not a very nice solution since we dont support
-		// interrupts at this stage of the application lifecycle. some changes
-		// would be needed on the cmd level to support context cancellation at
-		// this stage
-		select {
-		case err = <-syncedChan:
 			if err != nil {
-				return nil, err
+				batchSvc.Set(err)
```
acud, line 703: not sure i understand why this `Set` is needed

pkg/api/api_test.go
```diff
@@ -93,6 +93,7 @@ type testServerOptions struct {
 	PostageContract    postagecontract.Interface
 	Post               postage.Service
 	Steward            steward.Interface
+	BatchSyncStatus    postage.SyncStatus
```
acud, line 96: not sure whether an fully fledged interface is needed here. perhaps a pluggable function would do

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3127)* status: 0 of 11 files reviewed, 6 unresolved discussions (waiting on @acud and @notanatol)

___
*[`pkg/api/api_test.go` line 96 at r1](https://reviewable.io/reviews/ethersphere/bee/3127#-N7zQKp45TUVVUZeqpyj-r1-96:-N89PmbKCDEJMrrfAAUm:bbq239p) ([raw file](https://github.com/ethersphere/bee/blob/deb822f6e9eeed340223a6bfa74f336511df1cea/pkg/api/api_test.go#L96)):*

not sure whether an fully fledged interface is needed here. perhaps a pluggable function would do
</blockquote></details>

not sure how to access the plugged function, because we are passing around the `EventUpdater` interface and I would like to avoid casting
___
*[`pkg/node/node.go` line 703 at r1](https://reviewable.io/reviews/ethersphere/bee/3127#-N7zQKp45TUVVUZeqpyn-r1-703:-N89Pk141vRZ_AbwRrNC:bkmd0cn) ([raw file](https://github.com/ethersphere/bee/blob/deb822f6e9eeed340223a6bfa74f336511df1cea/pkg/node/node.go#L703)):*

not sure i understand why this `Set` is needed
</blockquote></details>

removed by refactoring the listener
___
*[`pkg/node/node.go` line 709 at r1](https://reviewable.io/reviews/ethersphere/bee/3127#-N7zQKp45TUVVUZeqpyn-r1-709:-N892BnY6ExFeKeqFsYP:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/deb822f6e9eeed340223a6bfa74f336511df1cea/pkg/node/node.go#L709)):*

this comment is no longer relevant, since interrupts _are_ supported
</blockquote></details>

fixed
___
*[`pkg/node/node.go` line 725 at r1](https://reviewable.io/reviews/ethersphere/bee/3127#-N7zQKp45TUVVUZeqpyn-r1-725:-N89PiXn0M9-zLCqoUsW:b4sy2au) ([raw file](https://github.com/ethersphere/bee/blob/deb822f6e9eeed340223a6bfa74f336511df1cea/pkg/node/node.go#L725)):*

this is almost identical to the code block above. i'd refactor this into a local, function scoped function (i.e. `syncFn := func() { ....}`) and just call it either from the goroutine or from the main thread.
</blockquote></details>

refactored
___
*[`pkg/postage/interface.go` line 16 at r1](https://reviewable.io/reviews/ethersphere/bee/3127#-N7zQKp45TUVVUZeqpyq-r1-16:-N89PhDS47JaLA7eSySf:b-hff6wv) ([raw file](https://github.com/ethersphere/bee/blob/deb822f6e9eeed340223a6bfa74f336511df1cea/pkg/postage/interface.go#L16)):*

`GetSyncStatus` otherwise this is just a getter setter naming which is not very comprehensible
</blockquote></details>

renamed
___
*[`pkg/postage/interface.go` line 17 at r1](https://reviewable.io/reviews/ethersphere/bee/3127#-N7zQKp45TUVVUZeqpyq-r1-17:-N89Pg648fbg9Yzssfpw:b-hfpv0l) ([raw file](https://github.com/ethersphere/bee/blob/deb822f6e9eeed340223a6bfa74f336511df1cea/pkg/postage/interface.go#L17)):*

`SetError` maybe but this is also a bit bizarre... an external component can set the state of an internal component via getter/setter?
</blockquote></details>

removed

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3127)* status: 0 of 11 files reviewed, 6 unresolved discussions (waiting on @acud)

___
*[`pkg/api/api_test.go` line 96 at r1](https://reviewable.io/reviews/ethersphere/bee/3127#-N7zQKp45TUVVUZeqpyj-r1-96:-N8DRcG0DCTGIwFl-31_:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/deb822f6e9eeed340223a6bfa74f336511df1cea/pkg/api/api_test.go#L96)):*

not sure how to access the plugged function, because we are passing around the `EventUpdater` interface and I would like to avoid casting
</blockquote></details>

fixed

aloknerurkar (approved):
Nice! LGTM

pkg/api/postage.go
```diff
@@ -37,6 +37,26 @@ func (s *Service) postageAccessHandler(h http.Handler) http.Handler {
 	})
 }
 
+func (s *Service) postageSyncStatusCheckHandler(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		done, err := s.syncStatus()
+		if err != nil {
+			s.logger.Debugf("postage access: %v", err)
+			s.logger.Error("postage access: syncing failed")
+			jsonhttp.ServiceUnavailable(w, "batch unavailable: syncing failed")
```
acud, line 46: the error message here is a bit nebulous, also, not sure why "batch unavailable".

pkg/postage/batchservice/batchservice.go
```diff
@@ -261,7 +264,14 @@ func (svc *batchService) Start(startBlock uint64, initState *postage.ChainSnapsh
 		startBlock = initState.LastBlockNumber
 	}
 
-	return svc.listener.Listen(startBlock+1, svc, initState), nil
+	syncedChan := svc.listener.Listen(startBlock+1, svc, initState)
+
+	select {
+	case err = <-syncedChan:
+		return err
+	case <-interrupt:
```
acud, line 272: something to bear in mind here is that once the `os.Signal` is read off the channel, you must make sure that the error bubbles up correctly. it might be a bit risky to pass this channel down the stack, since there are multiple points where we potentially read from it. the signal is sent once, rather than a closure of the channel, so if you read from it here, some other component might not be aware of the signal. so for example, since this in a light node will be done in a separate goroutine, the syncing process fails, but the process has been sent a kill signal, however (maybe) this goroutine picks up the signal, but the error does not bubble up the stack, and the process never exists (except for when you signal it with the kill or term _again_, leading the value to be read again from the same channel by a different goroutine). i would suggest not to pass this channel between components, but rather wrap it in `cmd/start.go` in a separate goroutine with a closure that would just `close()` a fresh `chan struct{}` which you could pass down into the new bee constructor.

acud (approved):
looks nice. one minor concern :)

pkg/api/postage.go
notanatol: fixed

pkg/postage/batchservice/batchservice.go
notanatol: Implemented a `struct{}` based channel listening on the system calls, as you described.

cmd/bee/cmd/start.go
```diff
@@ -151,8 +151,14 @@ func (c *command) initStartCmd() (err error) {
...
 			// We want to clean up things at the end.
-			interruptChannel := make(chan os.Signal, 1)
-			signal.Notify(interruptChannel, syscall.SIGINT, syscall.SIGTERM)
+			sysInterruptChannel := make(chan os.Signal, 1)
+			signal.Notify(sysInterruptChannel, syscall.SIGINT, syscall.SIGTERM)
+
+			interruptChannel := make(chan struct{})
+			go func() {
+				<-sysInterruptChannel
+				close(interruptChannel)
```
acud, line 160: nice. just one last suggestion. the thing is that the structure of the interrupt interception below was structured such that:
- user presses ctrl-c to quit
- process gets stuck(??) / shutdown commences
- user presses ctrl-c again to force the shutdown and not wait for a clean shutdown OR wait for a clean shutdown

that's why there's two reads from the signal channel later down in the file. so my suggestion would be, to keep those as they are, and just move the `close` here to line 223 (and get rid of the extra goroutine). the other consequence is that right now you will never wait for the `done` to select on line 245, since the interrupt channel is already closed and will most probably always select before the `done` gets closed. details... :)

acud (approved):

PR #3131, opened 2022-07-28
fix: incorrect batch depth

### LINKED ISSUE #3092: Stamp diluting doesn't change the depth
### Context
bee 1.6.2 (mode: dev & restricted)

### Summary
After diluting a stamp the TTL is reduced but the depth remains unchanged

### Expected behavior
The stamp depth should change to the one specified in the api call

### Actual behavior
The depth doesn't change

### Steps to reproduce
- Create a stamp (eg: amount: 10, depth 16)
- Dilute the stamp with depth 17 (`/stamps/dilute/[batchid]/17`)

### Possible solution
🤷‍♂️

swarmHaseeb:
### Checklist

### Description

This PR fixes #3092 and enables the change in depth after diluting

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3131)

### DIFF
--- a/pkg/postage/mock/service.go
+++ b/pkg/postage/mock/service.go
@@ -5,6 +5,7 @@
 package mock
 
 import (
+	"bytes"
 	"errors"
 	"math/big"
 	"sync"
@@ -93,7 +94,20 @@ func (m *mockPostage) HandleCreate(_ *postage.Batch) error { return nil }
 
 func (m *mockPostage) HandleTopUp(_ []byte, _ *big.Int) {}
 
-func (m *mockPostage) HandleDepthIncrease(_ []byte, _ uint8, _ *big.Int) {}
+func (m *mockPostage) HandleDepthIncrease(batchID []byte, newDepth uint8, newBalance *big.Int) {
+	m.issuerLock.Lock()
+	defer m.issuerLock.Unlock()
+
+	for k, v := range m.issuersMap {
+		if bytes.Equal(batchID, v.ID()) {
+			if newDepth > v.Depth() {
+				newStamp := postage.NewStampIssuer(v.Label(), k, v.ID(), v.Amount(), newDepth, v.BucketDepth(), v.BlockNumber(), v.ImmutableFlag())
+				m.issuersMap[k] = newStamp
+			}
+			return
+		}
+	}
+}
 
 func (m *mockPostage) Close() error {
 	return nil


pkg/postage/mock/service.go
```diff
@@ -93,7 +94,20 @@ func (m *mockPostage) HandleCreate(_ *postage.Batch) error { return nil }
 
 func (m *mockPostage) HandleTopUp(_ []byte, _ *big.Int) {}
 
-func (m *mockPostage) HandleDepthIncrease(_ []byte, _ uint8, _ *big.Int) {}
```
aloknerurkar, line 96: So this is a mock implementation. This is created mainly for using in tests. If we start adding code here for features for dev mode, we will need to start writing tests for these separately.

I believe this bug for dev mode should not be fixed at all.

swarmHaseeb: @istae @aloknerurkar how should I proceed with from here?

aloknerurkar: So if you want to provide the functionality for dev mode, I would suggest either implementing a different package which does things similar to mock but it has unit tests for the entire code.

Then we can use this pkg instead of the mock one in dev mode init. The same goes for the other PR.

pkg/postage/mock/service.go
```diff
@@ -114,7 +114,9 @@ func (m *mockPostage) HandleTopUp(batchID []byte, amount *big.Int) {
 	}
 }
 
-func (m *mockPostage) HandleDepthIncrease(_ []byte, _ uint8) {}
+func (m *mockPostage) HandleDepthIncrease(_ []byte, _ uint8) {
+	panic("not implemented")
```
aloknerurkar, line 118: So I think we decided to return error on the API. This is not the same as putting a panic here.

You need to change the `devnode.go` file, the mockcontract to return error on these things.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3131)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @acud, @aloknerurkar, and @istae)

___
*[`pkg/postage/mock/service.go` line 118 at r1](https://reviewable.io/reviews/ethersphere/bee/3131#-N84-q-MYZalCvN5ip80-r1-118:-NBHDU-n58Fr0wILnTen:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/451f008150998b0d462eeac991589779a2ae8cc1/pkg/postage/mock/service.go#L118)):*

So I think we decided to return error on the API. This is not the same as putting a panic here.

You need to change the `devnode.go` file, the mockcontract to return error on these things.

```
mockPostContract.WithTopUpBatchFunc
mockPostContract.WithDiluteBatchFunc
```
</blockquote></details>

Done.

pkg/node/devnode.go
```diff
@@ -278,7 +278,10 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 				}
 				topUpAmount := big.NewInt(0).Div(batch.Value, big.NewInt(int64(1<<(batch.Depth))))
 
-				post.HandleTopUp(batch.ID, topUpAmount)
+				err = post.HandleTopUp(batch.ID, topUpAmount)
```
aloknerurkar, line 281: This is wrong. This function doesnt have to change.

All you need to do is return error here. Not do any calculations. Same for HandleDepthIncrease.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3131)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @acud, @aloknerurkar, and @istae)

___
*[`pkg/node/devnode.go` line 281 at r2](https://reviewable.io/reviews/ethersphere/bee/3131#-NBHCztmp7gHRtUsa2EB-r2-281:-NBHQ1CS6EDjMCz_jrig:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/ff6fb289c1e0ae7951f0b7b5ba0ae6e277db2b1c/pkg/node/devnode.go#L281)):*

This is wrong. This function doesnt have to change.

All you need to do is return error here. Not do any calculations. Same for HandleDepthIncrease.
</blockquote></details>

Done.

pkg/node/devnode.go
```diff
@@ -301,7 +302,8 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 				}
 
 				post.HandleDepthIncrease(batch.ID, newDepth)
-				return nil
+
+				return mockPost.ErrNotImplemented
```
aloknerurkar, line 306: We should ONLY return this error. Why do we need L288-L304?

Same for the above function. These functions should just be 

```
func(ctx context.Context, batchID []byte, newDepth uint8) error {
        return ErrNotImplemented
}
```

Also, the error should not be defined in the mock package. Mock package is not expected to implement things, so housing the NotImplemented error there doesnt make sense. There could possible be a ErrNotImplemented already defined somewhere. If not you should ideally define it in the `node` package.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3131)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @acud, @aloknerurkar, and @istae)

___
*[`pkg/node/devnode.go` line 306 at r3](https://reviewable.io/reviews/ethersphere/bee/3131#-NBHCztmp7gHRtUsa2EB-r3-306:-NBHs7Ue1aj5id_vlDrv:b-gft78l) ([raw file](https://github.com/ethersphere/bee/blob/57ed8ac70919fcc46834609eb0cc68e27f47c55c/pkg/node/devnode.go#L306)):*

We should ONLY return this error. Why do we need L288-L304?

Same for the above function. These functions should just be 

```
func(ctx context.Context, batchID []byte, newDepth uint8) error {
        return ErrNotImplemented
}
```

Also, the error should not be defined in the mock package. Mock package is not expected to implement things, so housing the NotImplemented error there doesnt make sense. There could possible be a ErrNotImplemented already defined somewhere. If not you should ideally define it in the `node` package.

</blockquote></details>

got it, though all ErrNotImplemented are implemented within the mock package which is why i also implemented it, just within the right file. But ill create one within node pkg.

aloknerurkar (approved):

pkg/api/api.go
```diff
@@ -104,6 +104,7 @@ var (
 	errFileStore            = errors.New("could not store file")
 	errInvalidPostageBatch  = errors.New("invalid postage batch id")
 	errBatchUnusable        = errors.New("batch not usable")
+	ErrNotImplemented       = errors.New("not implemented")
```
mrekucci, line 107: The error doesn't need to be exported.

pkg/api/postage.go
```diff
@@ -487,6 +487,12 @@ func (s *Service) postageTopUpHandler(w http.ResponseWriter, r *http.Request) {
 			jsonhttp.PaymentRequired(w, "out of funds")
 			return
 		}
+		if errors.Is(err, ErrNotImplemented) {
+			s.logger.Debug("topup batch: not implemented", "error", err)
+			s.logger.Error(nil, "topup batch: not implemented")
+			jsonhttp.NotImplemented(w, "not implemented")
```
mrekucci, line 493: You can pass the error here directly and the `jsonhttp` will take care of it. The same is below.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3131)* status: 0 of 9 files reviewed, 5 unresolved discussions (waiting on @acud, @aloknerurkar, @istae, @mrekucci, and @notanatol)

___
*[`pkg/api/api.go` line 107 at r4](https://reviewable.io/reviews/ethersphere/bee/3131#-NBM5TEgIu-j8Gos2d6c-r4-107:-NBMqnJN9lClV2Baxthr:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/af78354f9975843fb9bd307959be56157be93846/pkg/api/api.go#L107)):*

The error doesn't need to be exported.
</blockquote></details>

Done.
___
*[`pkg/api/postage.go` line 493 at r4](https://reviewable.io/reviews/ethersphere/bee/3131#-NBHCztmp7gHRtUsa2EA-r4-493:-NBMqnzM3NMbG3cnR_A-:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/af78354f9975843fb9bd307959be56157be93846/pkg/api/postage.go#L493)):*

You can pass the error here directly and the `jsonhttp` will take care of it. The same is below.
</blockquote></details>

Done.

pkg/api/postage.go
mrekucci: If you do `jsonhttp.NotImplemented(w, nil)`, this will automatically translate using the `http.StatusText(http.StatusNotImplemented)` to "Not Implemented".

pkg/api/postage.go
```diff
@@ -487,6 +487,12 @@ func (s *Service) postageTopUpHandler(w http.ResponseWriter, r *http.Request) {
 			jsonhttp.PaymentRequired(w, "out of funds")
 			return
 		}
+		if errors.As(err, &errNotImplemented) {
```
mrekucci, line 490: You should use `errors.Is` instead of `errors.As`. The same is below.

pkg/api/api.go
```diff
@@ -104,6 +104,7 @@ var (
 	errFileStore            = errors.New("could not store file")
 	errInvalidPostageBatch  = errors.New("invalid postage batch id")
 	errBatchUnusable        = errors.New("batch not usable")
+	errNotImplemented       = errors.New("not implemented")
```
mrekucci, line 107: This error is not the same type as you've defined in the `devnode.go` file. You should drop this error, export the `errNotImplemented` from the `devnode.go` file, and check against this exported error.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3131)* status: 0 of 11 files reviewed, 7 unresolved discussions (waiting on @acud, @aloknerurkar, @istae, @mrekucci, and @notanatol)

___
*[`pkg/api/api.go` line 107 at r5](https://reviewable.io/reviews/ethersphere/bee/3131#-NBM5TEgIu-j8Gos2d6c-r5-107:-NBS-soa4aMQtdlbUmY5:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/9199df99843a04184e9720559050228ad75e30b6/pkg/api/api.go#L107)):*

This error is not the same type as you've defined in the `devnode.go` file. You should drop this error, export the `errNotImplemented` from the `devnode.go` file, and check against this exported error.
</blockquote></details>

Done.
___
*[`pkg/api/postage.go` line 493 at r4](https://reviewable.io/reviews/ethersphere/bee/3131#-NBHCztmp7gHRtUsa2EA-r4-493:-NBS-uCu7SC5NafVpVC1:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/af78354f9975843fb9bd307959be56157be93846/pkg/api/postage.go#L493)):*

If you do `jsonhttp.NotImplemented(w, nil)`, this will automatically translate using the `http.StatusText(http.StatusNotImplemented)` to "Not Implemented".
</blockquote></details>

Done.
___
*[`pkg/api/postage.go` line 490 at r5](https://reviewable.io/reviews/ethersphere/bee/3131#-NBHCztmp7gHRtUsa2EA-r5-490:-NBS-tRR9JYTRq6H_HNk:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/9199df99843a04184e9720559050228ad75e30b6/pkg/api/postage.go#L490)):*

You should use `errors.Is` instead of `errors.As`. The same is below.
</blockquote></details>

Done.

mrekucci (approved):
Thanks!

PR #3132, opened 2022-07-28
feat: remove storage depth

istae:
### Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3132)

### DIFF
--- a/pkg/topology/depthmonitor/depthmonitor.go
+++ b/pkg/topology/depthmonitor/depthmonitor.go
@@ -6,16 +6,14 @@ package depthmonitor
 
 import (
 	"errors"
-	"sync"
 	"time"
 
 	"github.com/ethersphere/bee/pkg/logging"
-	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/postage"
 	"github.com/ethersphere/bee/pkg/topology"
 )
 
 const (
-	depthKey             string  = "storage_depth"
 	adaptationFullWindow float64 = 2 * 60 * 60 // seconds allowed to fill half of the fully empty reserve
 	adaptationRollback           = 5 * 60      // seconds to slightly roll back the adaption window in case half capacity is not reached
 )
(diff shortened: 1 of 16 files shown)


metacertain:
might be a missing lock in one of the dm/export_test functions?
otherwise looks great

metacertain (approved):
LGTM, thanks!

pkg/topology/depthmonitor/depthmonitor.go
```diff
@@ -96,27 +89,21 @@ func (s *Service) manage(warmupTime time.Duration) {
...
 	}
 
-	// check if we have already saved a depth value before shutting down previously.
-	var initialDepth uint8
-	err := s.st.Get(depthKey, &initialDepth)
-	if err != nil && !errors.Is(err, storage.ErrNotFound) {
-		s.logger.Errorf("depthmonitor: failed reading stored depth: %s", err.Error())
-	}
+	// wire up batchstore to start reporting storage radius to kademlia
+	s.bs.SetStorageRadiusSetter(topology)
```
aloknerurkar, line 93: Although this makes sense, I would prefer not doing this here. This should be part of batchstore initialization in node.go and not depthmonitor. In case of light nodes we could initialize it with a dummy setter which does nothing.

This can be easily missed if we put it here.

mrekucci (approved):
> Reviewed 7 of 15 files at r1, 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3132)* status: all files reviewed, 1 unresolved discussion (waiting on @istae)

___
*[`pkg/topology/depthmonitor/depthmonitor.go` line 104 at r2](https://reviewable.io/reviews/ethersphere/bee/3132#-N8SHmQqEmzZy8yHElyW:-N8SHmQr-jq4AaUPZyPM:b-9v0q9r) ([raw file](https://github.com/ethersphere/bee/blob/571077a43ea27cfc8e93b5b91204c6bdd4cb1ba3/pkg/topology/depthmonitor/depthmonitor.go#L104)):*
> ```Go
> 	})
> 	if err != nil {
> 		s.logger.Errorf("depthmonitor: batchstore set storage radius: %w", err)
> ```

The `%w` verb doesn't apply in the case of the logger, use `%v` instead.
___
*[`pkg/topology/depthmonitor/depthmonitor.go` line 179 at r2](https://reviewable.io/reviews/ethersphere/bee/3132#-N8SIREt51I4BDPDSAsf:-N8SIREt51I4BDPDSAsg:b-3vhoyk) ([raw file](https://github.com/ethersphere/bee/blob/571077a43ea27cfc8e93b5b91204c6bdd4cb1ba3/pkg/topology/depthmonitor/depthmonitor.go#L179)):*
> ```Go
> 			})
> 			if err != nil {
> 				s.logger.Errorf("depthmonitor: batchstore set storage radius: %w", err)
> ```

The same as above, use `%v` instead of `%w`.

pkg/topology/depthmonitor/depthmonitor.go
istae: so storage radius is zero initially. Because right under this line we change the storage radius in the case that it's zero, this is the exact place we need to wire kademlia to batchstore.

istae: a zero storage radius must not be reported to kademlia, in short

aloknerurkar (approved):

PR #3138, opened 2022-08-01
historical syncing to dictate storage radius decrease

istae:
### Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3138)

### DIFF
--- a/pkg/pullsync/pullsync.go
+++ b/pkg/pullsync/pullsync.go
@@ -13,6 +13,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"sync"
 	"time"
 
@@ -41,6 +42,7 @@ const (
 )
 
 const logMore = false // enable this for more logging
+const MaxCursor = math.MaxUint64
 
 var (
 	ErrUnsolicitedChunk = errors.New("peer sent unsolicited chunk")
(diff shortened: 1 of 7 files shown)


pkg/topology/kademlia/kademlia.go
```diff
@@ -844,6 +844,11 @@ func binSaturated(oversaturationAmount int, staticNode staticPeerFunc, k *Kad) b
 // recalcDepth calculates, assigns the new depth, and returns if depth has changed
 func (k *Kad) recalcDepth() {
 
+	if k.storageRadius != swarm.MaxPO {
```
metacertain, line 847: depth should be minimum of storage radius and connection depth, this early return can set depth to higher than connection depth

pkg/topology/kademlia/kademlia.go
```diff
@@ -1340,6 +1345,17 @@ func (k *Kad) EachPeerRev(f topology.EachPeerFunc, filter topology.Filter) error
 		return f(addr, po)
 	})
 }
+func (k *Kad) PeersCount(filter topology.Filter) int {
+	count := 0
+	_ = k.connectedPeers.EachBinRev(func(addr swarm.Address, _ uint8) (bool, bool, error) {
```
aloknerurkar, line 1350: I think there is a function called `Length` on connectedPeers which gives complete length of pslice.

pkg/pullsync/pullsync.go
```diff
@@ -128,6 +130,7 @@ func (s *Syncer) Protocol() p2p.ProtocolSpec {
 // If the requested interval is too large, the downstream peer has the liberty to
 // provide less chunks than requested.
 func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8, from, to uint64) (topmost uint64, ruid uint32, err error) {
+	isLiveSync := to == MaxCursor
```
aloknerurkar, line 133: Is this always true?

istae: currently, yes

pkg/topology/kademlia/kademlia.go
istae: the reason this change is required, because storage radius starts at reserve radius, if the connection depth is below the radius, then syncing happens at connection depth, and not at storage radius.

PR #3147, opened 2022-08-02
refactor: migrate legacy loggers on package level

### LINKED ISSUE #3134: Replace existing logger with the newly introduced structural logger
Recently, a new structural logger has been introduced to the codebase. This structural logger currently acts as a sink for the existing logger. To take advantage of all the new functionality it brings, replace the existing logger(s) in the codebase with this new style of structural logging and register the selected logger(s) on which it is desirable to manipulate the verbosity levels.

mrekucci:
### Checklist

### Description
Migrates all loggers in the packages that log to the new structural logging style.
Closes #3134

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3147)


notanatol (approved):

notanatol (approved):

istae (approved):

mrekucci (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3147)* status: 0 of 162 files reviewed, 5 unresolved discussions (waiting on @aloknerurkar)

___
*[`pkg/api/api.go` line 226 at r1](https://reviewable.io/reviews/ethersphere/bee/3147#-N8iXoSChvZMSj1lVpCS-r1-226:-N95pAbk1CBT57E3TB0L:b-x83dtm) ([raw file](https://github.com/ethersphere/bee/blob/71df9d102c78d47c59987d98c31b928264fce153/pkg/api/api.go#L226)):*

So I dont see this being used. I see that we are initializing this in different places. Maybe you meant to remove this?
</blockquote></details>

Actually, it is used in the `router.go` file on lines: 523, 535, and 540.
___
*[`pkg/api/api.go` line 325 at r1](https://reviewable.io/reviews/ethersphere/bee/3147#-N8iXoSChvZMSj1lVpCS-r1-325:-N95pkTPCN1fZZtWssrl:b-l0t6ei) ([raw file](https://github.com/ethersphere/bee/blob/71df9d102c78d47c59987d98c31b928264fce153/pkg/api/api.go#L325)):*

Why do we need to do this? Why not use the same logger?
</blockquote></details>

Nice catch! Fixed.
___
*[`pkg/api/api.go` line 340 at r1](https://reviewable.io/reviews/ethersphere/bee/3147#-N8iXoSChvZMSj1lVpCS-r1-340:-N95q3HZCmNOID2x6fMP:bvyjksx) ([raw file](https://github.com/ethersphere/bee/blob/71df9d102c78d47c59987d98c31b928264fce153/pkg/api/api.go#L340)):*

At places we are using the newly built logger whereas there are places where we use the main logger. How do we distinguish the two?
</blockquote></details>

Basically, I created and registered a new logger for each service, which more or less results in almost every package having its own logger that inherits from the main logger (called "node"). Sometimes some packages use the same component. For example, `chainsyncer` uses `blocker` and so does `kademlia`. So instead of registering a single `blocker` logger that is used in both of these components, there is a `chainsyncer/blocker` logger and a `kademlia/chainsyncer` logger that can be independently requested to change the verbosity level of the log. In general, there is no hard rule for this, and we can only register loggers that we need to change their verbosity levels during a debugging session. So this is just an initial setup that I think may be beneficial for us, and which we can adjust later according to the cases we encounter. BTW, if you're interested, this is the current log tree (printed by the `/loggers` endpoint):
```json
{
  "tree": {
    "node": {
      "/": {
        "accounting": {
          "+": [
            "info|node/accounting[0][]>>824633786392"
          ]
        },
        "api": {
          "+": [
            "info|node/api[1][]>>824633786392",
            "info|node/api[0][]>>824633786392"
          ]
        },
        "batchservice": {
          "+": [
            "info|node/batchservice[0][]>>824633786392"
          ]
        },
        "batchstore": {
          "+": [
            "info|node/batchstore[0][]>>824633786392"
          ]
        },
        "chainsyncer": {
          "/": {
            "blocker": {
              "+": [
                "info|node/chainsyncer/blocker[0][]>>824633786392"
              ]
            }
          },
          "+": [
            "info|node/chainsyncer[1][]>>824633786392",
            "info|node/chainsyncer[0][]>>824633786392"
          ]
        },
        "chequebook": {
          "+": [
            "info|node/chequebook[0][]>>824633786392"
          ]
        },
        "handshake": {
          "+": [
            "info|node/handshake[0][]>>824633786392",
            "info|node/handshake[1][]>>824633786392"
          ]
        },
        "hive": {
          "+": [
            "info|node/hive[0][]>>824633786392"
          ]
        },
        "kademlia": {
          "/": {
            "blocker": {
              "+": [
                "info|node/kademlia/blocker[0][]>>824633786392"
              ]
            }
          },
          "+": [
            "info|node/kademlia[1][]>>824633786392",
            "info|node/kademlia[0][]>>824633786392"
          ]
        },
        "leveldb": {
          "+": [
            "info|node/leveldb[0][]>>824633786392"
          ]
        },
        "libp2p": {
          "+": [
            "info|node/libp2p[1][]>>824633786392",
            "info|node/libp2p[0][]>>824633786392"
          ]
        },
        "listener": {
          "+": [
            "info|node/listener[0][]>>824633786392"
          ]
        },
        "localstore": {
          "+": [
            "info|node/localstore[0][]>>824633786392"
          ]
        },
        "multiresolver": {
          "+": [
            "info|node/multiresolver[0][]>>824633786392"
          ]
        },
        "netstore": {
          "+": [
            "info|node/netstore[0][]>>824633786392"
          ]
        },
        "pinpong": {
          "+": [
            "info|node/pinpong[0][]>>824633786392"
          ]
        },
        "priceoracle": {
          "+": [
            "info|node/priceoracle[1][]>>824633786392",
            "info|node/priceoracle[0][]>>824633786392"
          ]
        },
        "pricing": {
          "+": [
            "info|node/pricing[0][]>>824633786392",
            "info|node/pricing[1][]>>824633786392"
          ]
        },
        "pseudosettle": {
          "+": [
            "info|node/pseudosettle[0][]>>824633786392"
          ]
        },
        "pss": {
          "+": [
            "info|node/pss[0][]>>824633786392"
          ]
        },
        "puller": {
          "+": [
            "info|node/puller[2][]>>824633786392",
            "info|node/puller[0][]>>824633786392"
          ]
        },
        "pullsync": {
          "+": [
            "info|node/pullsync[0][]>>824633786392",
            "info|node/pullsync[2][]>>824633786392"
          ]
        },
        "pusher": {
          "+": [
            "info|node/pusher[0][]>>824633786392"
          ]
        },
        "pushsync": {
          "+": [
            "info|node/pushsync[0][]>>824633786392"
          ]
        },
        "retrieval": {
          "+": [
            "info|node/retrieval[0][]>>824633786392"
          ]
        },
        "swap": {
          "+": [
            "info|node/swap[0][]>>824633786392",
            "info|node/swap[1][]>>824633786392"
          ]
        },
        "swapprotocol": {
          "+": [
            "info|node/swapprotocol[0][]>>824633786392"
          ]
        },
        "tags": {
          "+": [
            "info|node/tags[0][]>>824633786392"
          ]
        },
        "transaction": {
          "+": [
            "info|node/transaction[1][]>>824633786392",
            "info|node/transaction[0][]>>824633786392"
          ]
        }
      },
      "+": [
        "info|node[0][]>>824633786392"
      ]
    }
  }
}
```

I'm not sure if that answered your question, though.
___
*[`pkg/api/api_test.go` line 303 at r1](https://reviewable.io/reviews/ethersphere/bee/3147#-N8cyJcekkvOn7eGV1EC-r1-303:-N95wHB5-GpGbm-c5qmO:b-uipo81) ([raw file](https://github.com/ethersphere/bee/blob/71df9d102c78d47c59987d98c31b928264fce153/pkg/api/api_test.go#L303)):*

Would it be possible to set a default logger for the tests based on flags?

In the go test command we can parse some new flag which when provided initializes a logger which writes to stdout, otherwise we use the noop logger.

When I run the unit tests I sometimes need to see the logs. So in this case I would have to make some code changes everytime to enable the logs. I would like to avoid that.

</blockquote></details>

I'm sure we can do it, but I would not mix it with this PR. I'd recommend creating new PR for this.
___
*[`pkg/api/bzz.go` line 263 at r1](https://reviewable.io/reviews/ethersphere/bee/3147#-N8iXoSChvZMSj1lVpCW-r1-263:-N95wa9L8aNd6ptdIgtY:b-bnjoiw) ([raw file](https://github.com/ethersphere/bee/blob/71df9d102c78d47c59987d98c31b928264fce153/pkg/api/bzz.go#L263)):*

So I guess this is used for the `Trace` messages. loggerV1 doesnt really sound right in these cases. The verbosity level chosen is unclear as to what it means. 
</blockquote></details>

The main idea is that we no longer use a trace verbosity, but rather think in terms of V-levels [described here](https://github.com/go-logr/logr#why-v-levels). The advantage is that you can create debug messages that differ by the V-level (marked `v=x` in logs, where x is a number from 1 to max int32). So the granularity of your log messages can be very fine in some places. And when you register these V-level loggers, then you can enable or disable the output of a particular logger during a debugging session. With reference to that, the name `V1logger` makes sense here to let you know at which V-level the message will be logged. For legacy reasons, we consider the V1 level to be a trace.

Issue #2026, opened 2021-06-08
Requests often return 500 without any error message


agazso:
#### Summary
There are many requests that return 500 without any error message in the response. That makes it impossible to tell what the error was in a client application and also it makes difficult to debug issues without having access to an error log.

One particular example is when sending a PSS message, there are two cases when the node can return 500 error without further explanation:

https://github.com/ethersphere/bee/blob/1890f323bd002f43805601ace2f342c5c7242d00/pkg/api/pss.go#L76

and

https://github.com/ethersphere/bee/blob/1890f323bd002f43805601ace2f342c5c7242d00/pkg/api/pss.go#L99

There are many more similar cases like this (see below).

#### Expected behavior

It would be good if the 500 error would only be used as a last effort and if the problem is with the request itself return 4XX codes (for example in pss.go line 76 as seen above). Also it would be good if there were some explanation of the error in the response so that problems can be debugged more easily.

#### Examples in the Bee code

```
% grep -rn "InternalServerError(w, nil" .
./pkg/debugapi/pingpong.go:44:		jsonhttp.InternalServerError(w, nil)
./pkg/debugapi/peer.go:80:		jsonhttp.InternalServerError(w, nil)
./pkg/api/dirs.go:67:		jsonhttp.InternalServerError(w, nil)
./pkg/api/dirs.go:97:			jsonhttp.InternalServerError(w, nil)
./pkg/api/dirs.go:106:			jsonhttp.InternalServerError(w, nil)
./pkg/api/pin.go:32:		jsonhttp.InternalServerError(w, nil)
./pkg/api/pin.go:47:		jsonhttp.InternalServerError(w, nil)
./pkg/api/pin.go:68:		jsonhttp.InternalServerError(w, nil)
./pkg/api/pin.go:79:		jsonhttp.InternalServerError(w, nil)
./pkg/api/pin.go:84:		jsonhttp.InternalServerError(w, nil)
./pkg/api/pin.go:105:		jsonhttp.InternalServerError(w, nil)
./pkg/api/pin.go:127:		jsonhttp.InternalServerError(w, nil)
./pkg/api/chunk.go:135:			jsonhttp.InternalServerError(w, nil)
./pkg/api/feed.go:181:		jsonhttp.InternalServerError(w, nil)
./pkg/api/feed.go:188:		jsonhttp.InternalServerError(w, nil)
./pkg/api/feed.go:196:			jsonhttp.InternalServerError(w, nil)
[...]

PR #3148, opened 2022-08-02
fix: add error body to 500 responses

notanatol:
### Checklist

### Description
Closes: #2026

---
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3148)

### DIFF
--- a/pkg/api/bytes.go
+++ b/pkg/api/bytes.go
@@ -43,7 +43,7 @@ func (s *Service) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 
 	if strings.Contains(strings.ToLower(r.Header.Get(contentTypeHeader)), "multipart/form-data") {
 		logger.Error("bytes upload: multipart uploads are not supported on this endpoint")
-		jsonhttp.BadRequest(w, nil)
+		jsonhttp.BadRequest(w, "multipart uploads not supported")
 		return
 	}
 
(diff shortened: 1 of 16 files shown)


istae (approved):

pkg/api/bytes.go
```diff
@@ -37,13 +37,13 @@ func (s *Service) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		logger.Debugf("bytes upload: get putter:%v", err)
 		logger.Error("bytes upload: putter")
-		jsonhttp.BadRequest(w, nil)
+		jsonhttp.BadRequest(w, fmt.Sprintf("new stamper putter: %v", err))
```
aloknerurkar, line 40: The reason why this was not there in the first place was that we did not want to leak all these implementation details to the users. For eg, stamper putter or split write etc don't mean anything to the users. I am not sure if we want to expose these things.

Instead, we should put some messages here that don't leak the implementation details. This is the same reason why the error logs dont log the actual errors. As these errors are not meant for users. This is why they only show in debug logs.

notanatol: reverted

mrekucci (changes requested):
Exposing internal errors directly to the consumer is bad practice. Errors are not sanitized and can leak sensitive information. Also, exposing internal errors to the consumer does not provide much value; instead, messages should be worded in such a way that they do not go into unnecessary detail and multiple 500 API errors returned by the same endpoint can be easily distinguished: https://cloud.google.com/apis/design/errors

> Reviewed 13 of 16 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3148#-:-N8XWGGf6y3eOy7jM-GC:b-dlibhu)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar)

notanatol:
> Exposing internal errors directly to the consumer is bad practice. Errors are not sanitized and can leak sensitive information. Also, exposing internal errors to the consumer does not provide much value; instead, messages should be worded in such a way that they do not go into unnecessary detail and multiple 500 API errors returned by the same endpoint can be easily distinguished: https://cloud.google.com/apis/design/errors
> 
> > Reviewed 13 of 16 files at r1, 3 of 3 files at r2, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3148#-:-N8XWGGf6y3eOy7jM-GC:b-dlibhu)_ status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar)

that's a valid comment but I believe this discussion should be had here: https://github.com/ethersphere/bee/issues/2026

mrekucci (approved):
> Reviewed 16 of 16 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3148)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar)

PR #3152, opened 2022-08-04
chore: reduce locking extent in ratelimit package

vladopajic:
### Checklist

### Description
This PR fixes overlocking in `Allow` method which can arise when `Allow` is concurrently being called with two or more different keys. 

Instead of locking  whole functionality of `Allow` method we are now locking only part that changes state of limiters map, leaving individual limiter to handle synchronization individually.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3152)

### DIFF
--- a/pkg/ratelimit/ratelimit.go
+++ b/pkg/ratelimit/ratelimit.go
@@ -32,15 +32,17 @@ func New(r time.Duration, burst int) *Limiter {
 
 // Allow checks if the limiter that belongs to 'key' has not exceeded the limit.
 func (l *Limiter) Allow(key string, count int) bool {
-
 	l.mtx.Lock()
-	defer l.mtx.Unlock()
 
 	limiter, ok := l.limiter[key]
 	if !ok {
 		limiter = rate.NewLimiter(l.rate, l.burst)
 		l.limiter[key] = limiter
 	}
+
+	// We are intentionally not defer calling Unlock in order to reduce locking extent.
+	// Individual limiter is capable for handling concurrent calls.
+	l.mtx.Unlock()
 
 	return limiter.AllowN(time.Now(), count)
 }


notanatol (approved):

aloknerurkar:
@vladopajic 

So have you seen any problems with this? If this is a problem, I would like to see some pprof data to point to this. Also, we will need data on how this change improves it. For eg, if this is not a problem in the first place, we dont need the overhead of RWLock instead of normal Lock. This may not always improve things.

vladopajic:
What overhead of `RWLock` do you have in mind?

The overlocking does not have anything to do with `RWLock` (that's additional improvement); it happens because previously this code would lock for getting/creating `Limiter` and `AllowN` method call, where later op does not need to be included in this lock.

I would also like to see performance implications of this change, but it would require a lot of time to make those. Not to mention that results could greatly differ based on node load and type of request. 

Reason why I decided to make this PR was because it was trivial change that  theoretically couldn't degrade performance, but in worst case scenario could have some positive impact.

aloknerurkar:
> What overhead of `RWLock` do you have in mind?
> 
> The overlocking does not have anything to do with `RWLock` (that's additional improvement); it happens because previously this code would lock for getting/creating `Limiter` and `AllowN` method call, where later op does not need to be included in this lock.
> 
> I would also like to see performance implications of this change, but it would require a lot of time to make those. Not to mention that results could greatly differ based on node load and type of request.
> 

So RWLock achieves its features by consuming additional memory and synchronization primitives. Nothing comes for free.  If there is no problem right now, I dont see why we should do this.

There are multiple places in the code where a lock is used this way. So, theoretically we should replace all instances of Mutex we have in the code with this, but there is no guarantee this will result in an improvement in performance.

> Reason why I decided to make this PR was because it was trivial change that theoretically couldn't degrade performance, but in worst case scenario could have some positive impact.

These are dangerous assumptions when it comes to concurrency. I would consider backing these with numbers.

vladopajic:
@aloknerurkar 

Ok, I realized that utility of `RWMutex` is that it gives access to all concurrent reads for some cost of complexity, but I am not fully aware of those tradeoffs (I will investigate those).

Based on this input I have riverted `RWMutex` change leaving only overlocking part.

pkg/ratelimit/ratelimit.go
```diff
@@ -30,24 +30,28 @@ func New(r time.Duration, burst int) *Limiter {
...
 
-	limiter, ok := l.limiter[key]
-	if !ok {
-		limiter = rate.NewLimiter(l.rate, l.burst)
-		l.limiter[key] = limiter
+	if limiter, exists := l.limiter[key]; exists {
+		return limiter
 	}
 
-	return limiter.AllowN(time.Now(), count)
```
aloknerurkar, line 45: So it would be easily achieved by just doing an `l.mtx.Unlock` here after the if case instead of a different function no?

vladopajic: That would have same effect as fix for overlocking.

mrekucci (commented):
I agree with @aloknerurkar. It is generally bad practice to make performance improvements based on assumptions. You should benchmark the code and make an informed decision with comments. It doesn't take much time to write a simple benchmark (see https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go, https://dave.cheney.net/2013/06/19/stress-test-your-go-packages). As for performance optimizations, I recommend you take a look at this: https://github.com/dgryski/go-perfbook

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3152#-:-N8wGsuR2263Mpr2nhw6:bydcve7)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @mrekucci)

vladopajic:
@mrekucci 

Only intention behind this PR is to reduce unnecessary overlocking. Since this could be easily improved I decided to make this PR following [boy scout rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html).

With these changes I claim that code will keep same functionality while _potentially_  improving performance (theoretically it couldn't degrade). I didn't want to go down the benchmark rabbit hole since performance was not key objective and different test sets could produce results with neglectable performance gains.

mrekucci:
@vladopajic

> Only intention behind this PR is to reduce unnecessary overlocking. Since this could be easily improved I decided to make this PR following [boy scout rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html).

Nobody denies 

> @mrekucci
> 
> Only intention behind this PR is to reduce unnecessary overlocking. Since this could be easily improved I decided to make this PR following [boy scout rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html).
> 
> With these changes I claim that code will keep same functionality while _potentially_ improving performance (theoretically it couldn't degrade). I didn't want to go down the benchmark rabbit hole since performance was not key objective and different test sets could produce results with neglectable performance gains.

I apologize, you stated your intention but I'm confused about the key objective you want to achieve. When it's not performance then what is it?

vladopajic:
@mrekucci 

Cleaner and better code. As PR title would suggest this change is `chore`.

mrekucci (changes requested):
Please, change the commit message. We're using the imperative mood for commit messages. The commit message is also misleading because the PR does not fix anything, but rather changes the way locking is done.

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3152#-:-N90Fe6n-tIRghRqXmqO:bmxklzk)* status: all files reviewed, 1 unresolved discussion (waiting on @vladopajic)

___
*[`pkg/ratelimit/ratelimit.go` line 35 at r2](https://reviewable.io/reviews/ethersphere/bee/3152#-N90E6ynCAklWCtNCbhe:-N90E6yo0yXAi0-4AjwS:b-z4jcp5) ([raw file](https://github.com/ethersphere/bee/blob/68cfa607388279adfefa330707164c3f071de099/pkg/ratelimit/ratelimit.go#L35)):*
> ```Go
> // Allow checks if the limiter that belongs to key has not exceeded the limit.
> func (l *Limiter) Allow(key string, count int) bool {
> 	limiter := l.limiterForKey(key)
> ```

There is no need to factor this out into a new limiter method. This will do:
```golang
	l.mtx.Lock()
	limiter, ok := l.limiter[key]
	if !ok {
		limiter = rate.NewLimiter(l.rate, l.burst)
		l.limiter[key] = limiter
	}
	l.mtx.Unlock()
	return limiter.AllowN(time.Now(), count)
```

vladopajic:
@mrekucci 

Is there issue having this function? 

This function is also part of code improvement because it makes this code 
1) easier to read - it requires less cognitive load to read `Allow` function as it separates functionality for getting/creating limiter 
2) makes future code changes less prone to overlocking

mrekucci (changes requested):
> 1. easier to read - it requires less cognitive load to read Allow function as it separates functionality for getting/creating limiter

The code is not so complex that it should be abstracted into a separate method/function. The second argument for a separate method/function is when it is called in multiple places, which is not the case here.

> 2. makes future code changes less prone to overlocking

It's bad practice to anticipate what will be needed in the future, if it will be needed then it might be added/refactored then.

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3152#-:-N91Gt302CIhtbxVtP9A:b1q1acu)* status: all files reviewed, 1 unresolved discussion (waiting on @vladopajic)

vladopajic:
@mrekucci 


> The code is not so complex that it should be abstracted into a separate method/function.

This function makes it even easier to read regardless of  current complexity - hence it's better code. 1*

In all cases this is desirable practice, since it makes better code (as per 1*),  why would you want to suffocate it? 

> The second argument for a separate method/function is when it is called in multiple places, which is not the case here.

This is not true at all. There is vast number of places where function is defined even when it's being called in just one place. 

> It's bad practice to anticipate what will be needed in the future, if it will be needed then it might be added/refactored then.

This can lead to open ended debate. 

Undeniably one needs to anticipate future (future in general; as well as code changes in this case). 
Even for statements 'X will not be changed in the future'  you are actually anticipating future where X looks the same in future as it looks in present.

mrekucci (changes requested):
> This function makes it even easier to read regardless of current complexity - hence it's better code. 1

As you can see, better and more readable code is somewhat subjective to whoever is creating or reviewing it.

> In all cases this is desirable practice, since it makes better code (as per 1*), why would you want to suffocate it?

This is simply not true. For example, there are reasons why it is desirable to have functions inline: escape-analysis (keeping all values in the stack), performance, etc... I'm not claiming that this is the case, I'm just showing that your claim is not true. In this case, as I already mentioned, I'd argue that it doesn't help with the readability, since the original code isn't complex and is compact enough.

> This is not true at all. There is vast number of places where function is defined even when it's being called in just one place.

It is true in this case.

> Undeniably one needs to anticipate future (future in general; as well as code changes in this case).
Even for statements 'X will not be changed in the future' you are actually anticipating future where X looks the same in future as it looks in present.

No one is saying that it will not change in the future. Features should be added when they are needed, not when you think they will be needed. For example, you created this PR based on certain assumptions that may or may not hold true in the future. I would argue that a "better" choice would be to get rid of the lock and use `sync.Map`, since we might have problems with locks-congestion in the future and the code would be much shorter and we wouldn't need this extra function that you've created:

```golang
type Limiter struct {
	limiter *sync.Map
	rate    rate.Limit
	burst   int
}

// New returns a new Limiter object with refresh rate and burst amount.
func New(r time.Duration, burst int) *Limiter {
	return &Limiter{
		limiter: new(sync.Map),
		rate:    rate.Every(r),
		burst:   burst,
	}
}

// Allow checks if the limiter that belongs to 'key' has not exceeded the limit.
func (l *Limiter) Allow(key string, count int) bool {
	val, _ := l.limiter.LoadOrStore(key, rate.NewLimiter(l.rate, l.burst))
	return val.(*rate.Limiter).AllowN(time.Now(), count)
}

// Clear deletes the limiter that belongs to 'key'.
func (l *Limiter) Clear(key string) {
	l.limiter.Delete(key)
}
```

Would it be shorter? Yes. Would it increase readability? It depends on who is reading the code and how familiar they are with `sync.Map`, as it has its own preferred use cases that the reader should be familiar with. But I wouldn't do that, because I don't know the read/write ratio on that map in this case. I would also have to write a benchmark first to support my claims, but why would I put the effort to do that when it works just fine and the code is not complicated at all? As I said, it's not good practice to change the code based on what you think might happen in the future.

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3152#-:-N91_L8H5oS6esjVTC59:b-y5jqqd)* status: all files reviewed, 1 unresolved discussion (waiting on @vladopajic)

vladopajic:
> As you can see, better and more readable code is somewhat subjective to whoever is creating or reviewing it.

For some practices it requires one to adhere to it before one can see results (benefits/drawbacks). 

> This is simply not true. For example, there are reasons why it is desirable to have functions inline: escape-analysis (keeping all values in the stack), performance, etc... I'm not claiming that this is the case, I'm just showing that your claim is not true. In this case, as I already mentioned, I'd argue that it doesn't help with the readability, since the original code isn't complex and is compact enough.

For this case you would use `inline` mechanism form compiler to optimize execution. But you would still have code isolated in function.

> sync.Map solution

This solution is bad because it introduces memory trashing because it always creates new limiters, `rate.NewLimiter(l.rate, l.burst)` is called every time. If this was not the case I would prefer that solution more.

> Would it increase readability? It depends on who is reading the code and how familiar they are with sync.Map

One should get familiar with tools one needs to accomplish job. It's not shame being unfamiliar with better tools / approaches, professionals are learning and getting better everyday.

mrekucci (changes requested):
> For this case you would use inline mechanism form compiler to optimize execution. But you would still have code isolated in function.

This wasn't the point. Your claim was that "In all cases this is desirable practice" and this just demonstrated that this is not true in general!

> This solution is bad because it introduces memory trashing because it always creates new limiters, rate.NewLimiter(l.rate, l.burst) is called every time. If this was not the case I would prefer that solution more.

Again, that wasn't the point. I didn't say it was good, but the opposite! Also, the flaw you've mentioned can be easily solved by accepting some tradeoffs.

> One should get familiar with tools one needs to accomplish job. It's not shame being unfamiliar with better tools / approaches, professionals are learning and getting better everyday.

I'm not sure about this comment either, it also wasn't the point, the point was regarding the "readability" and "better" code that you claim.

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3152#-:-N91uV7yBmoaQfG0xele:ba71ob)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar and @vladopajic)

vladopajic:
We keep going in circles here. At this point we will not be able to proceed without going through mediation.

@zelig Would you be able to be our mediator or alternatively assign one for this case?

aloknerurkar:
@vladopajic 

I agree with Peter that there is no need for this function. This existing function is compact enough and just moving the unlock below the if loop would be the best change.

If not that, I would say we don't need this change at all as there are no indications to say that this is a problem. Not sure if we need any mediation here.

vladopajic:
@aloknerurkar 

I can't comprehend why would you decide to block change because it encapsulated functionality in individual function. 

Certainly we have deadlock situation here so mediator would be needed to help us effectively resolve and proceed.

mrekucci:
> We keep going in circles here. At this point we will not be able to proceed without going through mediation.
> 
> @zelig Would you be able to be our mediator or alternatively assign one for this case?

We are not going in circles. As a reviewer, I have clearly stated what I require to be changed and why in order to give it green. You've made some arguments that I did not consider valid, and vice versa. It's okay if we disagree, but the conversation should be constructive, which I don't think it is at this point. Feel free to ask other team members for their review. Thanks for your effort and the PR 🙏

aloknerurkar:
> @aloknerurkar 
> 
> I can't comprehend why would you decide to block change because it encapsulated functionality in individual function. 
> 
> Certainly we have deadlock situation here so mediator would be needed to help us effectively resolve and proceed. 
> 
> 
> 
> 
> 

So as I mentioned in the start. The right way to do this is to first report an issue in the bee repo about some problem you are seeing. If it's related to concurrency, you should include the necessary pprof data to show the problem. Eye-balling is not the right way to go about these things. If this had been done, we would not have questioned the validity of the change.

As it is right now, I don't see this as something we should spend time on. Next time please try to create an issue first and then go about making any changes. If we think this is not a problem we would indicate it on the issue itself. That way no one's time is wasted.

I don't want to discourage you from making changes in the future. Just that we have some sort of structure to it.

vladopajic:
@mrekucci 

Thank you for your review effort.

Certainly I have closely followed all suggestions that you guys have pointed out in this process. 

I would also like to draw attention to fact that I also felt that conversation slip down from constructive path even before. Original version of [this comment](https://github.com/ethersphere/bee/pull/3152#pullrequestreview-1066815481) had exclamation marks and took my response out of context forcibly trying to prove it's own agenda. Which made me to think that mediator would be needed.

---- 


I have not taken anything personally and I am still looking to work with you closely.

mrekucci (approved):
Thanks!

aloknerurkar (approved):

PR #3153, opened 2022-08-04
fix: full nodes use batchstore, light nodes use kad

istae:
### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3153)


pkg/topology/kademlia/kademlia.go
```diff
@@ -844,11 +844,6 @@ func binSaturated(oversaturationAmount int, staticNode staticPeerFunc, k *Kad) b
 // recalcDepth calculates, assigns the new depth, and returns if depth has changed
 func (k *Kad) recalcDepth() {
 
-	if k.storageRadius != swarm.MaxPO {
-		k.depth = k.storageRadius
```
istae, line 848: Previously, we made this change to force pullsync to sync at storage radius, not connection depth.

@metacertain pointed out that this is not possible because at a high storage radius, there may no peers in the neighborhood.

So this draft PR makes the protocols use batchstore directly (depending on if its's a full node or not).

istae: @aloknerurkar @metacertain  ^^^

pkg/node/depth.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/ethersphere/bee/pkg/topology"
+)
+
+type batchStoreDepther struct {
+	rs   postage.ReserveStateGetter
+	base swarm.Address
+}
+
+func (d *batchStoreDepther) NeighborhoodDepth() uint8 {
+	return d.rs.GetReserveState().StorageRadius
```
metacertain, line 19: This is not correct. 
The depth for a full node is minimum of connection depth and storage radius

metacertain: What we actually need is for pull sync to use storage radius instead of depth, not to always show storage radius as depth, exactly for the aforementioned reason, e.g. in protocols it would lead to a too narrow understanding of neighborhood, no immediate replication, etc.

istae: > pointed out that this is not possible because at a high storage radius, there may no peers in the neighborhood.

This is why the depthmonitor has the decreasing mechanism to find the optimal radius.  
The storage radius is zero for light nodes, so using the minimum of both for light nodes, won't work. This is why for full nodes we should use the storage radius directly, and for light nodes, we use the connection depth.

> pointed out that this is not possible because at a high storage radius, there may no peers in the neighborhood.

The reverse can also be true where the connection depth is too low where at this depth there's too much content to put in our reserve.

metacertain (approved):

aloknerurkar (approved):

PR #3155, opened 2022-08-04
refactor: adding EntityStore as utility abstraction on top of StateStorer

vladopajic:
### Checklist

### Description
This PR adds `EntityStore` as utility abstraction when `StateStorer` is used for mapping on fixed key to value types. This will help to eliminate some of boilerplate code when dealing with this use case and futher improve readability. Once when codebase starts supporting go version >=1.8 we could benefit event more from such abstraction (for example we would benefit from `PluckValues` and `PluckKeys` functions which now would be messy to implement). 

As utility example this PR also has refactored `addressbook` to use this type. After this PR gets approved I'll refactor other places with this use case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3155)

### DIFF
--- a/pkg/addressbook/addressbook.go
+++ b/pkg/addressbook/addressbook.go
@@ -7,7 +7,6 @@ package addressbook
 import (
 	"errors"
 	"fmt"
-	"strings"
 
 	"github.com/ethersphere/bee/pkg/bzz"
 	"github.com/ethersphere/bee/pkg/storage"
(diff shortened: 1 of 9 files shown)


aloknerurkar:
@vladopajic 

Also this is not a bad idea, this is a major change. I would suggest next time, before making such changes, you open an issue in the bee repository describing what you are planning to do and get comments from the team before proceeding. They have bigger implications and we have to maintain them.

We are in the process of refactoring our localstore right now. As part of these changes we will also be refactoring the statestore. Which is why I don't think we need these changes right now. If you had filed an issue we could have communicated this in advance and saved you the efforts.

Issue #3158, opened 2022-08-08
Resync the batchstore from the events of the new contract 


agazso:
Related to storage incentives contract change issue (https://github.com/ethersphere/bee/issues/3139) there is a task to 
> add logic to accomodate for resyncing the batchstore from the events of the new contract (after disposing of obsolete data). This probably needs to happen for the first time running the new version, if previous version has ever ran, similar to a migration.

PR #3159, opened 2022-08-09
fix: resync the batchstore

notanatol:
### Checklist

- [x] Add new snapshot feed, contract (maybe a test contract) address
- [x] update handshake version

### Description
Closes: #3158

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3159)


pkg/postage/batchstore/reserve.go
```diff
@@ -122,6 +122,46 @@ func (s *store) cleanup() error {
...
+
+		return false, nil
+	})
+	if err != nil {
+		return err
+	}
+
+	for _, b := range evictions {
+
+		err := s.evictFn(b.ID)
```
istae, line 148: we might as well nuke the whole localstore in one call, instead of evicting content one batch at a time

istae:
we can also use `/pkg/statestore/migration`  to nuke the batchstore during bootup, not sure if we need this separate versioning,

pkg/postage/batchstore/reserve.go
ldeffenb: What about nodes that have substantial numbers of chunks explicitly pinned?  IMHO, nuking the whole localstore would be a severe dis-service to anyone expecting to actually retain data locally!

aloknerurkar (approved):

PR #3162, opened 2022-08-10
fix: expired stamps can be distinguished from unexpired stamps

### LINKED ISSUE #3067: The stamp after buying cannot be distinguished from an expired stamp
### Context
Bee 1.6.2, macOS

### Summary
It is not possible to distinguish a stamp that you just bought from an expired stamp because all their relevant properties are the same.

There are certain properties of a stamp to tell the state of it. `usable`, `exists` and `batchTTL` are such properties. However these properties are the same for a stamp you just bought and a stamp that is expired, therefore it's impossible to tell the difference between them. This may cause problems in applications because it is not trivial to filter out expired postage stamps, but most of the time the users are not interested in those. The workaround for now for applications is to keep track of the id of a recently bought postage stamp and not filter out those as expired, but this makes the API more difficult to use than necessary. Also it can lead to subtle bugs.

It would be better to distinguish between these use case. This could be possibly done either by adding an extra field to the postage stamp (e.g. `expired`), which would only be true in the case of the stamp were expired.

Previously the expired stamp just disappeared from the list of stamps, but this behaviour has changed at some point. That behaviour would also solve this issue, but I am not sure why (and when) this was changed and what would break if this would be reverted.

### More detailed explanation:

Immediately after buying a stamp, it looks like this (only keeping the relevant properties):
```
"usable": false,
"exists": false,
"batchTTL": -1
```

After ~30 secs, it looks like this:
```
"usable": false,
"exists": true,
"batchTTL": 1125
```

After a minute or two:
```
"usable": true,
"exists": true,
"batchTTL": 1075
```

After expiry (this is the same as right after buying):
```
"usable": false,
"exists": false,
"batchTTL": -1
```

swarmHaseeb:
### Checklist

### Description

Please refer to issue for description of issue. This PR adds an `expired` property when fetching postage stamps

This PR fixes #3067 
### Related Issue (Optional)

This PR fixes #3067 

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3162)

### DIFF
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -351,6 +351,7 @@ func (s *Service) postageGetStampHandler(w http.ResponseWriter, r *http.Request)
 		resp.BucketDepth = issuer.BucketDepth()
 		resp.BlockNumber = issuer.BlockNumber()
 		resp.ImmutableFlag = issuer.ImmutableFlag()
+		resp.Expired = issuer.Expired()
 	}
 
 	jsonhttp.OK(w, &resp)
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -151,6 +151,7 @@ type postageStampResponse struct {
 	ImmutableFlag bool           `json:"immutableFlag"`
 	Exists        bool           `json:"exists"`
 	BatchTTL      int64          `json:"batchTTL"`
+	Expired       bool           `json:"expired"`
 }
 
 type postageStampsResponse struct {
@@ -214,6 +215,7 @@ func (s *Service) postageGetStampsHandler(w http.ResponseWriter, r *http.Request
 				ImmutableFlag: v.ImmutableFlag(),
 				Exists:        exists,
 				BatchTTL:      batchTTL,
+				Expired:       v.Expired(),
 			})
 		}
 	}
--- a/pkg/node/devnode.go
+++ b/pkg/node/devnode.go
@@ -112,7 +112,10 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 	}
 	b.stateStoreCloser = stateStore
 
+	post := mockPost.New()
+
 	batchStore, err := batchstore.New(stateStore, func(b []byte) error { return nil }, logger)
+	batchStore.SetBatchExpiryHandler(post.HandleStampExpiry)
 	if err != nil {
 		return nil, fmt.Errorf("batchstore: %w", err)
 	}
@@ -237,7 +240,6 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 
 	traversalService := traversal.New(storer)
 
-	post := mockPost.New()
 	postageContract := mockPostContract.New(
 		mockPostContract.WithCreateBatchFunc(
 			func(ctx context.Context, amount *big.Int, depth uint8, immutable bool, label string) ([]byte, error) {
@@ -255,7 +257,7 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 					return nil, err
 				}
 
-				stampIssuer := postage.NewStampIssuer(label, string(overlayEthAddress.Bytes()), id, amount, batch.Depth, 0, 0, immutable, batchStore)
+				stampIssuer := postage.NewStampIssuer(label, string(overlayEthAddress.Bytes()), id, amount, batch.Depth, 0, 0, immutable)
 				_ = post.Add(stampIssuer)
 
 				return id, nil
--- a/pkg/node/devnode.go
+++ b/pkg/node/devnode.go
@@ -115,7 +115,6 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 	post := mockPost.New()
 
 	batchStore, err := batchstore.New(stateStore, func(b []byte) error { return nil }, logger)
-	batchStore.SetBatchExpiryHandler(post)
 	if err != nil {
 		return nil, fmt.Errorf("batchstore: %w", err)
 	}
--- a/pkg/node/devnode.go
+++ b/pkg/node/devnode.go
@@ -112,8 +112,6 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 	}
 	b.stateStoreCloser = stateStore
 
-	post := mockPost.New()
-
 	batchStore, err := batchstore.New(stateStore, func(b []byte) error { return nil }, logger)
 	if err != nil {
 		return nil, fmt.Errorf("batchstore: %w", err)
@@ -239,6 +237,7 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 
 	traversalService := traversal.New(storer)
 
+	post := mockPost.New()
 	postageContract := mockPostContract.New(
 		mockPostContract.WithCreateBatchFunc(
 			func(ctx context.Context, amount *big.Int, depth uint8, immutable bool, label string) ([]byte, error) {
--- a/pkg/node/devnode.go
+++ b/pkg/node/devnode.go
@@ -115,7 +115,7 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 	post := mockPost.New()
 
 	batchStore, err := batchstore.New(stateStore, func(b []byte) error { return nil }, logger)
-	batchStore.SetBatchExpiryHandler(post.HandleStampExpiry)
+	batchStore.SetBatchExpiryHandler(post)
 	if err != nil {
 		return nil, fmt.Errorf("batchstore: %w", err)
 	}
--- a/pkg/node/devnode.go
+++ b/pkg/node/devnode.go
@@ -255,7 +255,7 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 					return nil, err
 				}
 
-				stampIssuer := postage.NewStampIssuer(label, string(overlayEthAddress.Bytes()), id, amount, batch.Depth, 0, 0, immutable)
+				stampIssuer := postage.NewStampIssuer(label, string(overlayEthAddress.Bytes()), id, amount, batch.Depth, 0, 0, immutable, batchStore)
 				_ = post.Add(stampIssuer)
 
 				return id, nil
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -273,11 +273,6 @@ func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, si
 			return nil, fmt.Errorf("batchstore: %w", err)
 		}
 	}
-	post, err := postage.NewService(stateStore, batchStore, chainID)
-	if err != nil {
-		return nil, fmt.Errorf("postage service load: %w", err)
-	}
-	batchStore.SetBatchExpiryHandler(post)
 
 	chainBackend, overlayEthAddress, chainID, transactionMonitor, transactionService, err = InitChain(
 		p2pCtx,
@@ -606,8 +601,12 @@ func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, si
 	unreserveFn = storer.UnreserveBatch
 
 	validStamp := postage.ValidStamp(batchStore)
-
+	post, err := postage.NewService(stateStore, batchStore, chainID)
+	if err != nil {
+		return nil, fmt.Errorf("postage service load: %w", err)
+	}
 	b.postageServiceCloser = post
+	batchStore.SetBatchExpiryHandler(post)
 
 	var (
 		postageContractService postagecontract.Interface
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -278,7 +278,7 @@ func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, si
 		return nil, fmt.Errorf("postage service load: %w", err)
 	}
 	batchStore.SetBatchExpiryHandler(post)
-	
+
 	chainBackend, overlayEthAddress, chainID, transactionMonitor, transactionService, err = InitChain(
 		p2pCtx,
 		logger,
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -653,7 +653,6 @@ func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, si
 		transactionService,
 		post,
 		batchStore,
-		batchSvc,
 		chainEnabled,
 	)
 
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -263,6 +263,11 @@ func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, si
 	var batchStore postage.Storer = new(postage.NoOpBatchStore)
 	var unreserveFn func([]byte, uint8) (uint64, error)
 
+	post, err := postage.NewService(stateStore, batchStore, chainID)
+	if err != nil {
+		return nil, fmt.Errorf("postage service load: %w", err)
+	}
+
 	if chainEnabled {
 		var evictFn = func(b []byte) error {
 			_, err := unreserveFn(b, swarm.MaxPO+1)
@@ -272,6 +277,8 @@ func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, si
 		if err != nil {
 			return nil, fmt.Errorf("batchstore: %w", err)
 		}
+
+		batchStore.SetBatchExpiryHandler(post.HandleStampExpiry)
 	}
 
 	chainBackend, overlayEthAddress, chainID, transactionMonitor, transactionService, err = InitChain(
@@ -601,10 +608,7 @@ func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, si
 	unreserveFn = storer.UnreserveBatch
 
 	validStamp := postage.ValidStamp(batchStore)
-	post, err := postage.NewService(stateStore, batchStore, chainID)
-	if err != nil {
-		return nil, fmt.Errorf("postage service load: %w", err)
-	}
+
 	b.postageServiceCloser = post
 
 	var (
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -263,11 +263,6 @@ func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, si
 	var batchStore postage.Storer = new(postage.NoOpBatchStore)
 	var unreserveFn func([]byte, uint8) (uint64, error)
 
-	post, err := postage.NewService(stateStore, batchStore, chainID)
-	if err != nil {
-		return nil, fmt.Errorf("postage service load: %w", err)
-	}
-
 	if chainEnabled {
 		var evictFn = func(b []byte) error {
 			_, err := unreserveFn(b, swarm.MaxPO+1)
@@ -277,10 +272,13 @@ func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, si
 		if err != nil {
 			return nil, fmt.Errorf("batchstore: %w", err)
 		}
-
-		batchStore.SetBatchExpiryHandler(post.HandleStampExpiry)
 	}
-
+	post, err := postage.NewService(stateStore, batchStore, chainID)
+	if err != nil {
+		return nil, fmt.Errorf("postage service load: %w", err)
+	}
+	batchStore.SetBatchExpiryHandler(post)
+	
 	chainBackend, overlayEthAddress, chainID, transactionMonitor, transactionService, err = InitChain(
 		p2pCtx,
 		logger,
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -649,6 +649,7 @@ func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, si
 		transactionService,
 		post,
 		batchStore,
+		batchSvc,
 		chainEnabled,
 	)
 
--- a/pkg/postage/batchservice/batchservice.go
+++ b/pkg/postage/batchservice/batchservice.go
@@ -99,7 +99,6 @@ func New(
 // Create will create a new batch with the given ID, owner value and depth and
 // stores it in the BatchStore.
 func (svc *batchService) Create(id, owner []byte, totalAmout, normalisedBalance *big.Int, depth, bucketDepth uint8, immutable bool, txHash []byte) error {
-
 	// don't add batches which have value which equals total cumulative
 	// payout or that are going to expire already within the next couple of blocks
 	val := big.NewInt(0).Add(svc.storer.GetChainState().TotalAmount, svc.storer.GetChainState().CurrentPrice)
--- a/pkg/postage/batchservice/batchservice.go
+++ b/pkg/postage/batchservice/batchservice.go
@@ -106,7 +106,6 @@ func (svc *batchService) Create(id, owner []byte, totalAmout, normalisedBalance 
 		// don't do anything
 		return fmt.Errorf("batch service: batch %x: %w", id, ErrZeroValueBatch)
 	}
-
 	batch := &postage.Batch{
 		ID:          id,
 		Owner:       owner,
--- a/pkg/postage/batchservice/batchservice.go
+++ b/pkg/postage/batchservice/batchservice.go
@@ -99,10 +99,6 @@ func New(
 // Create will create a new batch with the given ID, owner value and depth and
 // stores it in the BatchStore.
 func (svc *batchService) Create(id, owner []byte, totalAmout, normalisedBalance *big.Int, depth, bucketDepth uint8, immutable bool, txHash []byte) error {
-	batch, err := svc.storer.Get(id)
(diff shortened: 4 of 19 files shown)


pkg/api/postage.go
```diff
@@ -173,6 +175,14 @@ func (s *Service) postageGetStampsHandler(w http.ResponseWriter, r *http.Request
...
 			return
 		}
+		if !exists {
+			state := s.batchStore.GetChainState()
+			if state.TotalAmount.Cmp(v.Amount()) == -1 {
+				isExpired = true
+			} else {
+				isExpired = false
+			}
+		}
```
notanatol, line 185: I think we can reduce the scope of the `isExpired` var and get rid of the else cond?
```suggestion
		var isExpired bool
		if !exists {
			state := s.batchStore.GetChainState()
			if state.TotalAmount.Cmp(v.Amount()) == -1 {
				isExpired = true
			}
		}
```

pkg/api/postage.go
```diff
@@ -164,6 +165,7 @@ type bucketData struct {
 func (s *Service) postageGetStampsHandler(w http.ResponseWriter, r *http.Request) {
 	isAll := strings.ToLower(r.URL.Query().Get("all")) == "true"
 	resp := postageStampsResponse{}
+	var isExpired bool
```
aloknerurkar, line 168: We should move this variable inside the for loop. The scope of this variable is wrong.

pkg/api/postage.go
```diff
@@ -173,6 +175,14 @@ func (s *Service) postageGetStampsHandler(w http.ResponseWriter, r *http.Request
 			jsonhttp.InternalServerError(w, "unable to check batch")
 			return
 		}
+		if !exists {
+			state := s.batchStore.GetChainState()
+			if state.TotalAmount.Cmp(v.Amount()) == -1 {
+				isExpired = true
+			} else {
+				isExpired = false
```
aloknerurkar, line 183: This is not required if you define the variable inside the loop. The default will always be false.

pkg/api/postage.go
```diff
@@ -321,6 +334,16 @@ func (s *Service) postageGetStampHandler(w http.ResponseWriter, r *http.Request)
 	}
 
 	if issuer != nil {
+
+		if !exists {
+			state := s.batchStore.GetChainState()
+			if state.TotalAmount.Cmp(issuer.Amount()) == -1 {
+				isExpired = true
+			} else {
+				isExpired = false
```
aloknerurkar, line 343: This else is not required.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3162)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @aloknerurkar, @notanatol, and @swarmHaseeb)

___
*[`pkg/api/postage.go` line 168 at r1](https://reviewable.io/reviews/ethersphere/bee/3162#-N96NPn17cVznBnBZ27g-r1-168:-N96cSn34-Uef5nAAY85:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/569163517d179c76493870feb862dc2aca56bb90/pkg/api/postage.go#L168)):*

We should move this variable inside the for loop. The scope of this variable is wrong.
</blockquote></details>

Done.
___
*[`pkg/api/postage.go` line 183 at r1](https://reviewable.io/reviews/ethersphere/bee/3162#-N96NPn17cVznBnBZ27g-r1-183:-N96cUPa--fX_SZHnG2o:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/569163517d179c76493870feb862dc2aca56bb90/pkg/api/postage.go#L183)):*

This is not required if you define the variable inside the loop. The default will always be false.
</blockquote></details>

Done.
___
*[`pkg/api/postage.go` line 185 at r1](https://reviewable.io/reviews/ethersphere/bee/3162#-N96NPn17cVznBnBZ27g-r1-185:-N96cVzY8O3AWsnfxAPU:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/569163517d179c76493870feb862dc2aca56bb90/pkg/api/postage.go#L185)):*

I think we can reduce the scope of the `isExpired` var and get rid of the else cond?
```suggestion
		var isExpired bool
		if !exists {
			state := s.batchStore.GetChainState()
			if state.TotalAmount.Cmp(v.Amount()) == -1 {
				isExpired = true
			}
		}
```
</blockquote></details>

Done.
___
*[`pkg/api/postage.go` line 343 at r1](https://reviewable.io/reviews/ethersphere/bee/3162#-N96NPn17cVznBnBZ27g-r1-343:-N96cX9D9XKBjtdqhRgo:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/569163517d179c76493870feb862dc2aca56bb90/pkg/api/postage.go#L343)):*

This else is not required.
</blockquote></details>

Done.

pkg/api/postage.go
```diff
@@ -164,15 +165,23 @@ type bucketData struct {
...
 		exists, err := s.batchStore.Exists(v.ID())
 		if err != nil {
 			s.logger.Debugf("get stamp issuer: check batch: %v", err)
 			s.logger.Error("get stamp issuer: check batch")
 			jsonhttp.InternalServerError(w, "unable to check batch")
 			return
 		}
+		if !exists {
+			state := s.batchStore.GetChainState()
+			if state.TotalAmount.Cmp(v.Amount()) == -1 {
```
aloknerurkar, line 181: The value to be checked should be <= 0

Even if it is 0, the batch is expired.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3162)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @aloknerurkar, @notanatol, and @swarmHaseeb)

___
*[`pkg/api/postage.go` line 181 at r2](https://reviewable.io/reviews/ethersphere/bee/3162#-N96NPn17cVznBnBZ27g-r2-181:-N9NJHf62L99EFLOTcRm:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4fd700f40226bf846b9e83cc93ec9f34a90ec0f7/pkg/api/postage.go#L181)):*

The value to be checked should be <= 0

Even if it is 0, the batch is expired.
</blockquote></details>

Done.

mrekucci (approved):
> Reviewed 1 of 1 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3162)* status: all files reviewed, 5 unresolved discussions (waiting on @aloknerurkar and @notanatol)

notanatol (approved):

pkg/api/postage.go
```diff
@@ -164,15 +165,23 @@ type bucketData struct {
...
 		exists, err := s.batchStore.Exists(v.ID())
 		if err != nil {
 			s.logger.Debugf("get stamp issuer: check batch: %v", err)
 			s.logger.Error("get stamp issuer: check batch")
 			jsonhttp.InternalServerError(w, "unable to check batch")
 			return
 		}
+		if !exists {
+			state := s.batchStore.GetChainState()
+			if v.Amount().Cmp(state.TotalAmount) <= 0 {
```
istae, line 181: `Amount` is the amount per chunk and cannot be compared with `TotalAmount` which is the last incremented value of the contract. If the batch does not exist in the batchstore, then it's expired OR we have not received the event of a newly created batch from the chain yet.

istae (changes requested):

pkg/api/postage.go
```diff
@@ -164,15 +165,23 @@ type bucketData struct {
...
+				isExpired = true
+			}
+		}
 
 		batchTTL, err := s.estimateBatchTTLFromID(v.ID())
```
istae, line 186: an expired bath can be inferred from the batchTTL

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3162)* status: all files reviewed, 7 unresolved discussions (waiting on @aloknerurkar, @istae, and @notanatol)

___
*[`pkg/api/postage.go` line 181 at r3](https://reviewable.io/reviews/ethersphere/bee/3162#-N96NPn17cVznBnBZ27g-r3-181:-N9W3Dav1oymERf5g1Yo:b-iex9v4) ([raw file](https://github.com/ethersphere/bee/blob/2caf9bb1782b25ff2999fb18d569c155d12568cb/pkg/api/postage.go#L181)):*

`Amount` is the amount per chunk and cannot be compared with `TotalAmount` which is the last incremented value of the contract. If the batch does not exist in the batchstore, then it's expired OR we have not received the event of a newly created batch from the chain yet. 
</blockquote></details>

yes and a part of problem is that we do not receive newly created batch and both newly created and expired look same, which is why this comparison. any suggestions?
___
*[`pkg/api/postage.go` line 186 at r3](https://reviewable.io/reviews/ethersphere/bee/3162#-N96NPn17cVznBnBZ27g-r3-186:-N9W3fSNCZht9_gejKGh:b15l1g3) ([raw file](https://github.com/ethersphere/bee/blob/2caf9bb1782b25ff2999fb18d569c155d12568cb/pkg/api/postage.go#L186)):*

an expired bath can be inferred from the batchTTL 
</blockquote></details>

batchttl is -1 when creating the batch so if we use it it may show newly created batches as expired.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3162)* status: all files reviewed, 7 unresolved discussions (waiting on @aloknerurkar, @istae, and @notanatol)

___
*[`pkg/api/postage.go` line 181 at r3](https://reviewable.io/reviews/ethersphere/bee/3162#-N96NPn17cVznBnBZ27g-r3-181:-N9W4T0F9LbGgFo6ngCj:b-z1mag2) ([raw file](https://github.com/ethersphere/bee/blob/2caf9bb1782b25ff2999fb18d569c155d12568cb/pkg/api/postage.go#L181)):*
<details><summary><i>Previously, swarmHaseeb (Haseeb Raja) wrote…</i></summary><blockquote>

yes and a part of problem is that we do not receive newly created batch and both newly created and expired look same, which is why this comparison. any suggestions?
</blockquote></details>

also, umm so if dilution happens, the Amount per chunk exceeds the TotalAmount as per test cases which basically means that the batch is expired.

istae:
@swarmHaseeb it's hard to distinguish a newly created batch from an expired one because until we receive the batch from the chain backend, it does not exist in the batchstore. What you can do is create the batch inside the `CreateBatch` func in the postage package which the api calls.

istae (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3162)* status: all files reviewed, 7 unresolved discussions (waiting on @aloknerurkar, @istae, @notanatol, and @swarmHaseeb)

___
*[`pkg/api/postage.go` line 181 at r3](https://reviewable.io/reviews/ethersphere/bee/3162#-N96NPn17cVznBnBZ27g-r3-181:-N9WVjGW1XDY3OEQ8Ie7:btu6c6u) ([raw file](https://github.com/ethersphere/bee/blob/2caf9bb1782b25ff2999fb18d569c155d12568cb/pkg/api/postage.go#L181)):*
<details><summary><i>Previously, swarmHaseeb (Haseeb Raja) wrote…</i></summary><blockquote>

also, umm so if dilution happens, the Amount per chunk exceeds the TotalAmount as per test cases which basically means that the batch is expired.
</blockquote></details>

if we know the comparison is invalid, why did it make it's way into the PR?

pkg/postage/service.go
```diff
@@ -62,6 +62,13 @@ func NewService(store storage.StateStorer, postageStore Storer, chainID int64) (
 		if err := st.UnmarshalBinary(value); err != nil {
 			return false, err
 		}
+		exists, err := s.postageStore.Exists(st.ID())
+		if err != nil {
+			return false, err
+		}
+		if !exists {
+			st.data.Expired = true
```
aloknerurkar, line 70: Why not call `st.SetExpired` here?

pkg/postage/batchservice/batchservice.go
```diff
@@ -99,13 +99,15 @@ func New(
 // Create will create a new batch with the given ID, owner value and depth and
 // stores it in the BatchStore.
 func (svc *batchService) Create(id, owner []byte, totalAmout, normalisedBalance *big.Int, depth, bucketDepth uint8, immutable bool, txHash []byte) error {
+
```
aloknerurkar, line 102: Remove this extra line.

pkg/node/node.go
```diff
@@ -263,6 +263,11 @@ func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, si
 	var batchStore postage.Storer = new(postage.NoOpBatchStore)
 	var unreserveFn func([]byte, uint8) (uint64, error)
 
+	post, err := postage.NewService(stateStore, batchStore, chainID)
```
aloknerurkar, line 266: This is wrong. This will initialize the postage service with NoOpBatchStore.

What you should do is call batchStore.SetBatchExpiryHandler after the postage service is created instead of moving this here.

pkg/postage/interface.go
```diff
@@ -97,4 +99,7 @@ type BatchEventListener interface {
 	HandleCreate(*Batch, *big.Int) error
 	HandleTopUp(id []byte, newBalance *big.Int)
 	HandleDepthIncrease(id []byte, newDepth uint8)
+	HandleStampExpiry(id []byte)
```
aloknerurkar, line 102: This is actually not correct. Because of this change we had to change the BatchListener which is a different interface.

I think the better option here is to define a new interface and not function.

```go
type BatchExpiryHandler interface {
    HandleStampExpiry([]byte)
}
```

In batch store instead of passing in ExpiredFn you will pass in this interface. And in `cleanup` you would call this `HandleStampExpiry` function using the interface if it is Set by the Setter. As an example you can look at the RadiusSetter defined above in this file.

pkg/postage/batchservice/batchservice_test.go
```diff
@@ -47,6 +47,10 @@ type mockBatchListener struct {
 	diluteCount int
 }
 
+func (m *mockBatchListener) HandleStampExpiry(id []byte) {
```
aloknerurkar, line 50: Check my other comment. This should be removed. We should not be changing the BatchEventListener interface

aloknerurkar (changes requested):

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3162)* status: 0 of 16 files reviewed, 12 unresolved discussions (waiting on @aloknerurkar, @istae, @mrekucci, and @notanatol)

___
*[`pkg/api/postage.go` line 181 at r3](https://reviewable.io/reviews/ethersphere/bee/3162#-N96NPn17cVznBnBZ27g-r3-181:-NAJSxVK8KDnKQvKn6jB:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/2caf9bb1782b25ff2999fb18d569c155d12568cb/pkg/api/postage.go#L181)):*

if we know the comparison is invalid, why did it make it's way into the PR?
</blockquote></details>

Done.
___
*[`pkg/node/node.go` line 266 at r4](https://reviewable.io/reviews/ethersphere/bee/3162#-NA3_2GH25wARXMJqH1S-r4-266:-NAJSzDMAiJ6oBwqH6MO:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f9047ba7a4ab9a1d16b5827d894eb4c657b650a2/pkg/node/node.go#L266)):*

This is wrong. This will initialize the postage service with NoOpBatchStore.

What you should do is call batchStore.SetBatchExpiryHandler after the postage service is created instead of moving this here.
</blockquote></details>

Done.
___
*[`pkg/postage/interface.go` line 102 at r4](https://reviewable.io/reviews/ethersphere/bee/3162#-NA9sBNTc7YjgXz05VGD-r4-102:-NAJT0Hx7Oulm28WZzyr:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f9047ba7a4ab9a1d16b5827d894eb4c657b650a2/pkg/postage/interface.go#L102)):*

This is actually not correct. Because of this change we had to change the BatchListener which is a different interface.

I think the better option here is to define a new interface and not function.

```go
type BatchExpiryHandler interface {
    HandleStampExpiry([]byte)
}
```

In batch store instead of passing in ExpiredFn you will pass in this interface. And in `cleanup` you would call this `HandleStampExpiry` function using the interface if it is Set by the Setter. As an example you can look at the RadiusSetter defined above in this file.
</blockquote></details>

Done.
___
*[`pkg/postage/service.go` line 70 at r4](https://reviewable.io/reviews/ethersphere/bee/3162#-N9k1YuoKY_-7mT1JfXD-r4-70:-NAJT11w7Dr94S9saaGp:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f9047ba7a4ab9a1d16b5827d894eb4c657b650a2/pkg/postage/service.go#L70)):*

Why not call `st.SetExpired` here?
</blockquote></details>

Done.
___
*[`pkg/postage/batchservice/batchservice.go` line 102 at r4](https://reviewable.io/reviews/ethersphere/bee/3162#-NA3_2GH25wARXMJqH1T-r4-102:-NAJT1uF0faqCHKOM2m8:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f9047ba7a4ab9a1d16b5827d894eb4c657b650a2/pkg/postage/batchservice/batchservice.go#L102)):*

Remove this extra line.
</blockquote></details>

Done.
___
*[`pkg/postage/batchservice/batchservice_test.go` line 50 at r4](https://reviewable.io/reviews/ethersphere/bee/3162#-NA9sBNTc7YjgXz05VGB-r4-50:-NAJT2wSDFAUcqfBpalN:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f9047ba7a4ab9a1d16b5827d894eb4c657b650a2/pkg/postage/batchservice/batchservice_test.go#L50)):*

Check my other comment. This should be removed. We should not be changing the BatchEventListener interface
</blockquote></details>

Done.

istae (approved):
Great work, thanks!

pkg/postage/stampissuer_test.go
```diff
@@ -40,3 +40,16 @@ func newTestStampIssuer(t *testing.T, block uint64) *postage.StampIssuer {
 	}
 	return postage.NewStampIssuer("label", "keyID", id, big.NewInt(3), 16, 8, block, true)
 }
+
+func TestGetUnexpiredStampIssuer(t *testing.T) {
+	t.Helper()
```
aloknerurkar, line 45: This is not a helper function for tests. Please check why this is used.

pkg/postage/stampissuer_test.go
```diff
@@ -40,3 +40,16 @@ func newTestStampIssuer(t *testing.T, block uint64) *postage.StampIssuer {
...
+	id := make([]byte, 32)
+	_, err := io.ReadFull(crand.Reader, id)
```
aloknerurkar, line 47: There is already a testing function to create a new Batch. It should be in postage testing pkg.

pkg/postage/stampissuer_test.go
```diff
@@ -40,3 +40,16 @@ func newTestStampIssuer(t *testing.T, block uint64) *postage.StampIssuer {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	si := postage.NewStampIssuer("label", "keyID", id, big.NewInt(3), 16, 8, 1, true)
+	if si.Expired() != false {
+		t.Fatalf(" stampIssuer.Expired() != false , required %v, is %v \n", false, si.Expired())
+	}
```
aloknerurkar, line 54: You will need to call SetExpire and check if Expired is set or not.

pkg/node/devnode.go
```diff
@@ -112,7 +112,10 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 	}
 	b.stateStoreCloser = stateStore
 
+	post := mockPost.New()
```
aloknerurkar, line 115: This is not required. We dont have to set any expiry handler. If it is nil, we already handle the case.

pkg/postage/batchservice/batchservice.go
```diff
@@ -106,6 +106,7 @@ func (svc *batchService) Create(id, owner []byte, totalAmout, normalisedBalance
 		// don't do anything
 		return fmt.Errorf("batch service: batch %x: %w", id, ErrZeroValueBatch)
 	}
+
```
aloknerurkar, line 109: Remove extra line

pkg/node/node.go
```diff
@@ -273,6 +273,11 @@ func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, si
 			return nil, fmt.Errorf("batchstore: %w", err)
 		}
 	}
+	post, err := postage.NewService(stateStore, batchStore, chainID)
```
aloknerurkar, line 276: Why is this moved? This should be moved back to where it was originally. The Set needs to happen after this is done.

pkg/postage/postagecontract/contract.go
```diff
@@ -66,6 +66,7 @@ func New(
 	transactionService transaction.Service,
 	postageService postage.Service,
 	postageStorer postage.Storer,
+
```
aloknerurkar, line 69: Please remember to do a self-review once you post your diff. All these changes not pertaining to your changeset need to be reverted once you are done.

aloknerurkar (changes requested):
I think one more test is required. In BatchStore we need to add a test where we do cleanup, we need to make sure this BatchExpiryHandler is called if its set.

swarmHaseeb (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3162#-:-NAK5fFh0AyKhBPJ5Fyo:bnfp4nl)* status: 0 of 18 files reviewed, 19 unresolved discussions (waiting on @aloknerurkar, @istae, @mrekucci, @notanatol, and @swarmHaseeb)

___
*[`pkg/node/devnode.go` line 115 at r6](https://reviewable.io/reviews/ethersphere/bee/3162#-NA3_2GH25wARXMJqH1R-r6-115:-NAJlJe0-39G5cpWcbBm:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f611adf5ed79a3635b4ce132f29801e80dfb8a99/pkg/node/devnode.go#L115)):*

This is not required. We dont have to set any expiry handler. If it is nil, we already handle the case.
</blockquote></details>

Done.
___
*[`pkg/node/node.go` line 276 at r6](https://reviewable.io/reviews/ethersphere/bee/3162#-NA3_2GH25wARXMJqH1S-r6-276:-NAJoU4_5kBEpEHML5jm:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f611adf5ed79a3635b4ce132f29801e80dfb8a99/pkg/node/node.go#L276)):*

Why is this moved? This should be moved back to where it was originally. The Set needs to happen after this is done. 
</blockquote></details>

Done.
___
*[`pkg/postage/stampissuer_test.go` line 45 at r6](https://reviewable.io/reviews/ethersphere/bee/3162#-NAJZFpsa_MemJe6XNJO-r6-45:-NAJnqpJ4TDmXXBClUSt:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f611adf5ed79a3635b4ce132f29801e80dfb8a99/pkg/postage/stampissuer_test.go#L45)):*

This is not a helper function for tests. Please check why this is used.
</blockquote></details>

Done.
___
*[`pkg/postage/stampissuer_test.go` line 47 at r6](https://reviewable.io/reviews/ethersphere/bee/3162#-NAJZFpsa_MemJe6XNJO-r6-47:-NAJmHYnC-nBzZyrHDIi:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f611adf5ed79a3635b4ce132f29801e80dfb8a99/pkg/postage/stampissuer_test.go#L47)):*

There is already a testing function to create a new Batch. It should be in postage testing pkg.
</blockquote></details>

Done.
___
*[`pkg/postage/stampissuer_test.go` line 54 at r6](https://reviewable.io/reviews/ethersphere/bee/3162#-NAJZFpsa_MemJe6XNJO-r6-54:-NAJmN7RCF-KqCwpGK1l:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f611adf5ed79a3635b4ce132f29801e80dfb8a99/pkg/postage/stampissuer_test.go#L54)):*

You will need to call SetExpire and check if Expired is set or not.
</blockquote></details>

Done.
___
*[`pkg/postage/batchservice/batchservice.go` line 109 at r6](https://reviewable.io/reviews/ethersphere/bee/3162#-NA3_2GH25wARXMJqH1T-r6-109:-NAJnrvs9kA8spUc24wq:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f611adf5ed79a3635b4ce132f29801e80dfb8a99/pkg/postage/batchservice/batchservice.go#L109)):*

Remove extra line
</blockquote></details>

Done.
___
*[`pkg/postage/postagecontract/contract.go` line 69 at r6](https://reviewable.io/reviews/ethersphere/bee/3162#-NA3_2GH25wARXMJqH1V-r6-69:-NAJntKK5frgRm6nV-_a:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f611adf5ed79a3635b4ce132f29801e80dfb8a99/pkg/postage/postagecontract/contract.go#L69)):*

Please remember to do a self-review once you post your diff. All these changes not pertaining to your changeset need to be reverted once you are done.
</blockquote></details>

Done.

aloknerurkar (approved):

istae (approved):

PR #3168, opened 2022-08-10
fix: switch listeners to tcp6

### LINKED ISSUE #3167: API should listen also on IPv6 `::1`
Node 18 changed the default behavior of translating DNS names into IPv6 instead of IPv4. That means that `localhost:1633` gets translated into `::1:1633`. I think this is a quite strong signal for push to IPv6 support and also it creates a lot of problems for us as suddenly a lot of our projects do not work on Node 18 as we use `localhost` in almost all our default settings.

Would it be possible for the Bee's API to listen also on the IPv6's localhost `::1`?

notanatol:
### Checklist

### Considerations
On linux it successfully resolves for both tcp4 and tcp6 but probably needs checking it on mac and win.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3168)

### DIFF
--- a/pkg/node/devnode.go
+++ b/pkg/node/devnode.go
@@ -178,7 +178,7 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 	var debugApiService *api.Service
 
 	if o.DebugAPIAddr != "" {
-		debugAPIListener, err := net.Listen("tcp", o.DebugAPIAddr)
+		debugAPIListener, err := net.Listen("tcp6", o.DebugAPIAddr)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ldeffenb (commented):
But if you do that, does it leave IP4 users in the dust?  Or does tcp6 listen on both 6 and 4?

notanatol:
> But if you do that, does it leave IP4 users in the dust? Or does tcp6 listen on both 6 and 4?

On linux the switch is transparent, my `bee-dashboard` works without need to make any changes to it.

But this needs validation for other operating systems.

AuHau:
Hmm yeah, I share a concern with @ldeffenb here that this might resolves in IPv6 only listening, which should definitely not be the case 😅

PR #3170, opened 2022-08-11
chore: validation checks bzz and stewardship

### LINKED ISSUE #1803: Improve input validation for BZZ Reupload
When I try to input all sort of invalid or valid-ish data, I pretty much always get `500` which seems to me a bit extreme as for most of these I would either expect `400` or `404`. About `404` I did not manage to actually get that one at all, not sure when I should receive it?

Example inputs:

```bash
$ curl -X PATCH -s localhost:1633/bzz/111/

// Non existing reference
$ curl -X PATCH -s localhost:1633/bzz/1e477b015af480e387fbf5edd90f1685a30c0e3ba88eeb3871b326b816a542da/
```

umerm-work:
### Checklist

### Description
if the reference is incorrect (not 64 or 128 hex characters), return 400
if the reference is correct but not found, then return 404
Remove bzz patch endpoint as it is deprecated.

Closes #1803 

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3170)

### DIFF
--- a/pkg/api/stewardship.go
+++ b/pkg/api/stewardship.go
@@ -5,7 +5,10 @@
 package api
 
 import (
+	"errors"
 	"net/http"
+
+	"github.com/ethersphere/bee/pkg/resolver"
 
 	"github.com/ethersphere/bee/pkg/jsonhttp"
 	"github.com/gorilla/mux"
@@ -16,10 +19,31 @@ import (
 func (s *Service) stewardshipPutHandler(w http.ResponseWriter, r *http.Request) {
 	nameOrHex := mux.Vars(r)["address"]
 	address, err := s.resolveNameOrAddress(nameOrHex)
-	if err != nil {
+	switch {
+	case errors.Is(err, resolver.ErrParse):
 		s.logger.Debug("stewardship put: parse address string failed", "string", nameOrHex, "error", err)
 		s.logger.Error(nil, "stewardship put: parse address string failed")
-		jsonhttp.NotFound(w, nil)
+		jsonhttp.BadRequest(w, "invalid address")
+		return
+	case errors.Is(err, resolver.ErrNotFound):
+		s.logger.Debug("stewardship put: address not found", "string", nameOrHex, "error", err)
+		s.logger.Error(nil, "stewardship put: address not found")
+		jsonhttp.NotFound(w, "address not found")
+		return
+	case errors.Is(err, resolver.ErrServiceNotAvailable):
+		s.logger.Debug("stewardship put: service unavailable", "string", nameOrHex, "error", err)
+		s.logger.Error(nil, "stewardship put: service unavailable")
+		jsonhttp.ServiceUnavailable(w, nil)
+		return
+	case errors.Is(err, resolver.ErrInvalidContentHash):
+		s.logger.Debug("stewardship put: invalid content hash", "string", nameOrHex, "error", err)
+		s.logger.Error(nil, "stewardship put: invalid content hash")
+		jsonhttp.Respond(w, 422, "stewardship put: invalid content hash")
+		return
+	case err != nil:
+		s.logger.Debug("stewardship put: resolve address or name string failed", "string", nameOrHex, "error", err)
+		s.logger.Error(nil, "stewardship put: resolve address or name string failed")
+		jsonhttp.InternalServerError(w, "stewardship put: resolve name or address")
 		return
 	}
 	err = s.steward.Reupload(r.Context(), address)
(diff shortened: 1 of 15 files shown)


mrekucci (changes requested):
Please reflect the API changes also in the openapi specifications.

umerm-work:
Done

pkg/api/stewardship.go
```diff
@@ -16,10 +19,31 @@ import (
...
+		return
+	case errors.Is(err, resolver.ErrServiceNotAvailable):
+		s.logger.Debug("stewardship put: service unavailable", "string", nameOrHex, "error", err)
+		s.logger.Error(nil, "stewardship put: service unavailable")
+		jsonhttp.ServiceUnavailable(w, nil)
+		return
+	case errors.Is(err, resolver.ErrInvalidContentHash):
+		s.logger.Debug("stewardship put: invalid content hash", "string", nameOrHex, "error", err)
+		s.logger.Error(nil, "stewardship put: invalid content hash")
+		jsonhttp.Respond(w, http.StatusUnprocessableEntity, "stewardship put: invalid content hash")
```
agazso, line 41: Why do we use a quite obscure HTTP error number (422 Unprocessable Entity) here when there is a simple well-known alternative to it (400 Bad Request)?

As far as I understand we don't even use 422 correctly and that may be also because it's obscure.
> The 422 (Unprocessable Content) status code indicates that the server understands the content type of the request content (hence a [415 (Unsupported Media Type)](https://httpwg.org/specs/rfc9110.html#status.415) status code is inappropriate), and the syntax of the request content is correct, but it was unable to process the contained instructions. For example, this status code can be sent if an XML request content contains well-formed (i.e., syntactically correct), but semantically erroneous XML instructions.
https://httpwg.org/specs/rfc9110.html#status.422

mrekucci (commented):
> Reviewed 5 of 6 files at r1, 2 of 2 files at r3, 5 of 5 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3170)* status: 14 of 15 files reviewed, 2 unresolved discussions (waiting on @umerm-work)

___
*[`pkg/api/stewardship.go` line 36 at r5](https://reviewable.io/reviews/ethersphere/bee/3170#-N9fjRCK46XiQwDjhZXj:-N9fjRCK46XiQwDjhZXk:b-ednssq) ([raw file](https://github.com/ethersphere/bee/blob/1522616e953d1a6fa76086736c28fe7843917599/pkg/api/stewardship.go#L36)):*
> ```Go
> 		s.logger.Debug("stewardship put: service unavailable", "string", nameOrHex, "error", err)
> 		s.logger.Error(nil, "stewardship put: service unavailable")
> 		jsonhttp.ServiceUnavailable(w, nil)
> ```

This error code can be misleading because it is not clear from the response what service is not available. We also do not try to be specific in the response code. I would suggest returning a simple code 500 with the following message, "stewardship put: resolver service unavailable".

umerm-work (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3170)* status: 14 of 15 files reviewed, 1 unresolved discussion (waiting on @agazso and @umerm-work)

___
*[`pkg/api/stewardship.go` line 41 at r5](https://reviewable.io/reviews/ethersphere/bee/3170#-N9BmQX5RH0DBHYexaKr-r5-41:-N9kSUgF0GD_gSGlYaeF:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/1522616e953d1a6fa76086736c28fe7843917599/pkg/api/stewardship.go#L41)):*
<details><summary><i>Previously, agazso (Attila Gazso) wrote…</i></summary><blockquote>

Why do we use a quite obscure HTTP error number (422 Unprocessable Entity) here when there is a simple well-known alternative to it (400 Bad Request)?

As far as I understand we don't even use 422 correctly and that may be also because it's obscure.
> The 422 (Unprocessable Content) status code indicates that the server understands the content type of the request content (hence a [415 (Unsupported Media Type)](https://httpwg.org/specs/rfc9110.html#status.415) status code is inappropriate), and the syntax of the request content is correct, but it was unable to process the contained instructions. For example, this status code can be sent if an XML request content contains well-formed (i.e., syntactically correct), but semantically erroneous XML instructions.
https://httpwg.org/specs/rfc9110.html#status.422
</blockquote></details>

Done.

mrekucci (approved):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3170)* status: all files reviewed, 1 unresolved discussion (waiting on @agazso)

aloknerurkar (approved):

agazso (approved):

Issue #3174, opened 2022-08-12
Test suite for the data abstraction layer interfaces


aloknerurkar:
Test suite for the new store.

Test suite should focus on correctness at this point. This includes testing all the features outlined in the core abstractions.

Should provide enough coverage for the in-mem implementation.

Issue #3173, opened 2022-08-12
In-memory implementation of the data abstraction layer


aloknerurkar:
Write an in-mem implementation of the data abstraction layer. This can be used for testing and validation of the testsuites.

PR #3190, opened 2022-08-15
feat: in-mem implementation and testsuite

### LINKED ISSUE #3172: Define the core abstractions for the storage package
Define the core chunk related abstractions

Define the data abstraction layer interfaces for indexes.

aloknerurkar:
### Checklist

### Description
This PR implements the in-mem Store. This is part of the localstore rewrite epic.

### Related Issue (Optional)
The PR closes #3172 #3173 #3174 #3180 

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3190)

### DIFF
--- a/pkg/storagev2/inmemstore/inmemstore.go
+++ b/pkg/storagev2/inmemstore/inmemstore.go
@@ -1,0 +1,190 @@
+package inmem
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/armon/go-radix"
+	storage "github.com/ethersphere/bee/pkg/storagev2"
+	"github.com/hashicorp/go-multierror"
+)
+
+const (
+	separator = "/"
+)
+
+// store implements an in-memory Store. We need lexicographic ordering of keys for the
+// store. So if we use the go in-built map, we might end up complicating this implementation.
+// Instead we will use the hashicorp/go-radix implementation. This pkg provides a mutable radix
+// which gives O(k) lookup and ordered iteration
+type store struct {
+	st *radix.Tree
+	mu sync.RWMutex
+}
+
+func getKeyString(i storage.Key) string {
+	return strings.Join([]string{i.Namespace(), i.ID()}, separator)
+}
+
+func (s *store) Get(i storage.Item) error {
+	key := getKeyString(i)
+
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	val, found := s.st.Get(key)
+	if !found {
+		return storage.ErrNotFound
+	}
+
+	err := i.Unmarshal(val.([]byte))
+	if err != nil {
+		return fmt.Errorf("failed unmarshaling item %w", err)
+	}
+
+	return nil
+}
+
+func (s *store) Has(k storage.Key) (bool, error) {
+	key := getKeyString(k)
+
+	s.mu.RLock()
+	_, found := s.st.Get(key)
+	s.mu.RUnlock()
+
+	return found, nil
+}
+
+func (s *store) GetSize(k storage.Key) (int, error) {
+	key := getKeyString(k)
+
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	val, found := s.st.Get(key)
+	if !found {
+		return 0, storage.ErrNotFound
+	}
+
+	return len(val.([]byte)), nil
+}
+
+func (s *store) Put(i storage.Item) error {
+	key := getKeyString(i)
+
+	val, err := i.Marshal()
+	if err != nil {
+		return fmt.Errorf("failed marshaling item %w", err)
+	}
+
+	s.mu.Lock()
+	s.st.Insert(key, val)
+	s.mu.Unlock()
+
+	return nil
+}
+
+func (s *store) Delete(k storage.Key) error {
+	key := getKeyString(k)
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	_, deleted := s.st.Delete(key)
+	if !deleted {
+		return storage.ErrNotFound
+	}
+	return nil
+}
+
+func (s *store) Count(k storage.Key) (int, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	count := 0
+	s.st.WalkPrefix(k.Namespace(), func(_ string, _ interface{}) bool {
+		count++
+		return false
+	})
+	return count, nil
+}
+
+func (s *store) Iterate(q storage.Query, fn storage.IterateFn) error {
+	if err := q.Validate(); err != nil {
+		return fmt.Errorf("invalid query %w", err)
+	}
+
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	var retErr *multierror.Error
+
+	getNext := func(k string, v interface{}) (*storage.Result, error) {
+		for _, filter := range q.Filters {
+			if filter(k, v.([]byte)) {
+				return nil, nil
+			}
+		}
+		var res *storage.Result
+		switch q.ItemAttribute {
+		case storage.QueryItemID, storage.QueryItemSize:
+			res = &storage.Result{ID: k, Size: len(v.([]byte))}
+		case storage.QueryItem:
+			newItem := q.Factory()
+			err := newItem.Unmarshal(v.([]byte))
+			if err != nil {
+				return nil, fmt.Errorf("failed unmarshaling %w", err)
+			}
+			res = &storage.Result{Entry: newItem}
+		}
+		return res, nil
+	}
+
+	prefix := q.Factory().Namespace()
+	switch q.Order {
+	case storage.KeyAscendingOrder:
+		s.st.WalkPrefix(prefix, func(k string, v interface{}) bool {
+			res, err := getNext(k, v)
+			if err != nil {
+				retErr = multierror.Append(retErr, err)
+				return true
+			}
+			if res != nil {
+				stop, err := fn(*res)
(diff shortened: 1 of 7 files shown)


pkg/storagev2/testsuite/testsuite.go
```diff
@@ -0,0 +1,488 @@
...
+		t.Fatalf("failed marshaling %s", err.Error())
+	}
+
+	buf2, err := b.Marshal()
+	if err != nil {
+		t.Fatalf("failed marshaling %s", err.Error())
+	}
+
+	if bytes.Compare(buf1, buf2) != 0 {
+		t.Fatal("bytes not equal for item %s/%s", a.Namespace(), a.ID())
```
aloknerurkar, line 87: Fatalf

mrekucci (approved):
> Reviewed 2 of 4 files at r1, 4 of 4 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3190)* status: all files reviewed, 1 unresolved discussion (waiting on @mrekucci)

___
*[`pkg/storagev2/inmemstore/inmemstore.go` line 1 at r2](https://reviewable.io/reviews/ethersphere/bee/3190#-N9VMsfy7_QpEiW0RSx9:-N9VMsfy7_QpEiW0RSxA:b-zhdpvi) ([raw file](https://github.com/ethersphere/bee/blob/5f5cf7302068374da06209364de2350d42bccc16/pkg/storagev2/inmemstore/inmemstore.go#L1)):*
> ```Go
> package inmem
> ```

Missing header.
___
*[`pkg/storagev2/inmemstore/inmemstore.go` line 20 at r2](https://reviewable.io/reviews/ethersphere/bee/3190#-N9VNbua40ixMJAdH6IM:-N9VNbua40ixMJAdH6IN:b-hsg5tx) ([raw file](https://github.com/ethersphere/bee/blob/5f5cf7302068374da06209364de2350d42bccc16/pkg/storagev2/inmemstore/inmemstore.go#L20)):*
> ```Go
> // store. So if we use the go in-built map, we might end up complicating this implementation.
> // Instead we will use the hashicorp/go-radix implementation. This pkg provides a mutable radix
> // which gives O(k) lookup and ordered iteration
> ```

This is an unnecessary implementation detail. Just mentioning using the radix tree instead of the map should be enough. E.g.: "Since we need a lexicographic arrangement of keys for the store, this implementation uses a radix tree instead of a map. This makes the implementation much simpler and maintainable with O(k) lookup and ordered iteration".



_Suggestion:_
```Go
// ... We need lexicographic ordering of keys for the
// store. So if we use the go in-built map, we might end up complicating this implementation.
// Instead we will use the hashicorp/go-radix implementation. This pkg provides a mutable radix
// which gives O(k) lookup and ordered iteration
```
___
*[`pkg/storagev2/inmemstore/inmemstore.go` line 50 at r2](https://reviewable.io/reviews/ethersphere/bee/3190#-N9VPRHv3gNk7ECLHhEi:-N9VPRHv3gNk7ECLHhEj:bqpgvw7) ([raw file](https://github.com/ethersphere/bee/blob/5f5cf7302068374da06209364de2350d42bccc16/pkg/storagev2/inmemstore/inmemstore.go#L50)):*
> ```Go
> 	err := i.Unmarshal(val.([]byte))
> 	if err != nil {
> 		return fmt.Errorf("failed unmarshaling item %w", err)
> ```

Adding `:` before the `%w` would make it more readable.
___
*[`pkg/storagev2/inmemstore/inmemstore.go` line 84 at r2](https://reviewable.io/reviews/ethersphere/bee/3190#-N9VPxVOEVQdpsamTlKu:-N9VPxVOEVQdpsamTlKv:bf26u1) ([raw file](https://github.com/ethersphere/bee/blob/5f5cf7302068374da06209364de2350d42bccc16/pkg/storagev2/inmemstore/inmemstore.go#L84)):*
> ```Go
> 	val, err := i.Marshal()
> 	if err != nil {
> 		return fmt.Errorf("failed marshaling item %w", err)
> ```

The same as above.
___
*[`pkg/storagev2/inmemstore/inmemstore.go` line 120 at r2](https://reviewable.io/reviews/ethersphere/bee/3190#-N9VQQ0O9w8tHCpdz1h2:-N9VQQ0O9w8tHCpdz1h3:b-b2rgbe) ([raw file](https://github.com/ethersphere/bee/blob/5f5cf7302068374da06209364de2350d42bccc16/pkg/storagev2/inmemstore/inmemstore.go#L120)):*
> ```Go
> func (s *store) Iterate(q storage.Query, fn storage.IterateFn) error {
> 	if err := q.Validate(); err != nil {
> 		return fmt.Errorf("invalid query %w", err)
> ```

The `Validate` already returns "invalid query" string. To avoid stuttering, I would suggest returning something like: `fmt.Errorf("iteration failed: %w", err)`.
___
*[`pkg/storagev2/inmemstore/inmemstore.go` line 142 at r2](https://reviewable.io/reviews/ethersphere/bee/3190#-N9VRjjX3PPTL2n1Et45:-N9VRjjX3PPTL2n1Et46:bbnybev) ([raw file](https://github.com/ethersphere/bee/blob/5f5cf7302068374da06209364de2350d42bccc16/pkg/storagev2/inmemstore/inmemstore.go#L142)):*
> ```Go
> 			err := newItem.Unmarshal(v.([]byte))
> 			if err != nil {
> 				return nil, fmt.Errorf("failed unmarshaling %w", err)
> ```

Missing `:`.
___
*[`pkg/storagev2/inmemstore/inmemstore.go` line 161 at r2](https://reviewable.io/reviews/ethersphere/bee/3190#-N9VS4WECYr9BlX90c5z:-N9VS4WF6j5S11S81jxu:bf26u1) ([raw file](https://github.com/ethersphere/bee/blob/5f5cf7302068374da06209364de2350d42bccc16/pkg/storagev2/inmemstore/inmemstore.go#L161)):*
> ```Go
> 				stop, err := fn(*res)
> 				if err != nil {
> 					retErr = multierror.Append(retErr, fmt.Errorf("failed in iteration %w", err))
> ```

The same as above.
___
*[`pkg/storagev2/inmemstore/inmemstore_test.go` line 1 at r2](https://reviewable.io/reviews/ethersphere/bee/3190#-N9VSmxL0TUXDMf-vTZP:-N9VSmxL0TUXDMf-vTZQ:b-zhdpvi) ([raw file](https://github.com/ethersphere/bee/blob/5f5cf7302068374da06209364de2350d42bccc16/pkg/storagev2/inmemstore/inmemstore_test.go#L1)):*
> ```Go
> package inmem_test
> ```

Missing header.
___
*[`pkg/storagev2/testsuite/testsuite.go` line 1 at r2](https://reviewable.io/reviews/ethersphere/bee/3190#-N9VSr2k1_3hTlXKrpz2:-N9VSr2k1_3hTlXKrpz3:b-zhdpvi) ([raw file](https://github.com/ethersphere/bee/blob/5f5cf7302068374da06209364de2350d42bccc16/pkg/storagev2/testsuite/testsuite.go#L1)):*
> ```Go
> package storetesting
> ```

Missing header.
___
*[`pkg/storagev2/testsuite/testsuite.go` line 151 at r2](https://reviewable.io/reviews/ethersphere/bee/3190#-N9VTrTEEUjeIUm70fBL:-N9VTrTEEUjeIUm70fBM:b-qqst0v) ([raw file](https://github.com/ethersphere/bee/blob/5f5cf7302068374da06209364de2350d42bccc16/pkg/storagev2/testsuite/testsuite.go#L151)):*
> ```Go
> 			err := s.Put(i)
> 			if err != nil {
> 				t.Fatalf("failed to add new entry: %s", err.Error())
> ```

No need to call `Error()` when you use `%v` instead of `%s`. The same is in the test cases below.

mrekucci (approved):
Apart from the comments LGTM.

notanatol (approved):

pkg/storagev2/inmemstore/inmemstore.go
```diff
@@ -0,0 +1,209 @@
...
+// license that can be found in the LICENSE file.
+
+package inmem
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/armon/go-radix"
```
istae, line 12: is this the hashicorp one?

istae (approved):
This is great!

The iterator function could be greatly simplified if we just returned the key and byte value pair for the callback. It would get rid of the factory, item attribute, etc.... I understand that it passes the responsibility to the user to do the unmarshalling but it simplifies everything else.

aloknerurkar:
> The iterator function could be greatly simplified if we just returned the key and byte value pair for the callback. It would get rid of the factory, item attribute, etc.... I understand that it passes the responsibility to the user to do the unmarshalling but it simplifies everything else.

So this would add up I think. For each instance in the codebase if we do Unmarshal and the error check, it would add a lot of unnecessary code. This is unnecessary as the store already knows how to interpret the bytes because of the Serializable interface.

This could be improved later once we have generics. We can get rid of the factory if we just infer the type from the method itself.

You can get access to the bytes directly using the filters. This would allow you to prevent unnecessary unmarshaling if you know can check on the bytes directly. There is a test case for this as well.

pkg/storagev2/inmemstore/inmemstore.go
aloknerurkar: Yes. Hashicorp has a immutable-radix implementation. In that they have linked this as the mutable version. This is what we need. Its a very good and stable library I think.

janos (approved):
👍

PR #3191, opened 2022-08-15
fix: fix beekeeper_artifacts.sh script

vandot:
There was a change in beekeeper flag to suppress log output, this PR fixes that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3191)


gacevicljubisa (approved):
To turn off verbose logging, possible values for **--log-verbosity** are both **0** and **"silent"**.

notanatol (approved):

PR #3198, opened 2022-08-16
feat: lock-base transaction implementation

### LINKED ISSUE #3195: Transaction implementation for in-memmory store

mrekucci:
### Checklist

### Description
A simple lock-base transaction implementation.
Closes #3195


pkg/storagev2/transaction.go
```diff
@@ -0,0 +1,248 @@
...
+	Rollback() error
+}
+
+// TxStore represents a Tx Store where all
+// operations are completed in a transaction.
+type TxStore interface {
+	Tx
+	Store
+
+	NewTx(*TxState) TxStore
```
aloknerurkar, line 36: So I think the `NewTx` function should not be here. The problem being we already have the object implementing Tx and Store. If users dont use NewTx, this could cause panics as the TxState is not initialised. In any case, self instantiating new objects seems not good.

pkg/storagev2/transaction.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+// TxChunkStore represents a Tx ChunkStore where
+// all operations are completed in a transaction.
+type TxChunkStore interface {
+	Tx
+	ChunkStore
+
+	NewTx(*TxState) TxChunkStore
```
aloknerurkar, line 45: Same as above. This function should ideally not be here. Not sure how that changes the code.

pkg/storagev2/leveldbstore/transaction.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// Bookkeeping of invasive operations executed
+	// on the Store to support rollback functionality.
+	batchMu sync.Mutex
+	batch   *leveldb.Batch
+}
+
+// Put implements the Store interface.
+func (s *TxStore) Put(item storage.Item) error {
+	err := s.TxStoreBase.Put(item)
```
aloknerurkar, line 31: So one problem with this approach is we are updating the underlying store directly. This means we are on a slippery slope as if there is some unexpected problem and we fail to rollback, we are screwed. I was initially trying to see if we could write the updates in a batch, but the problem is if we write it in the batch, we wont be able to read the latest state of the keys that were updated in the batch. And this is required. So I guess we cannot get around this.

With my approach, problem was we will have to create some hybrid read operations (Get, Has... etc) which could read from the batch. But then iteration is not easy to do and it starts getting complicated. And seems like we depend on this, the way we have the components. For eg in pinning we iterate and delete in batches. So the delete has to prevent the iteration from reading the same entries again. And this would be slightly challenging. So if we go with the current approach we dont have all these complications but the risk of going into some invalid state. We should also test if we can handle context cancellations properly. If not we should maybe write the undo batch in some recovery file and replay it once we are up. This comment is mainly to highlight this fact. I have no perfect solution here.

aloknerurkar (commented):
I will review properly once everything is done. These comments are just initial thoughts I have.

pkg/storagev2/transaction_test.go
```diff
@@ -0,0 +1,116 @@
...
+		default:
+			t.Error("AwaitDone() is blocking")
+		}
+	})
+
+	t.Run("lifecycle-done-by-parent-ctx", func(t *testing.T) {
+		t.Parallel()
+
+		ctx, cancel := context.WithCancel(context.Background())
+		t.Cleanup(cancel)
```
aloknerurkar, line 72: This is not required I think.

aloknerurkar (approved):

PR #3203, opened 2022-08-17
feat: leveldb store

### LINKED ISSUE #3176: LevelDB implementation of the data abstraction layer
Implement the Store on levelDB.

- Use in-mem store as reference implementation to look at all the features required
- Make sure testsuite passes
- Add any additional tests if required

notanatol:
### Checklist

### Description
Closes #3176


pkg/storagev2/leveldb/store.go
```diff
@@ -0,0 +1,292 @@
...
+)
+
+type Store struct {
+	DB      *leveldb.DB
+	path    string
+	closeLk sync.RWMutex
+}
+
+var _ storageV2.Store = (*Store)(nil)
+var _ storageV2.Tx = (*transaction)(nil)
```
aloknerurkar, line 29: Let's remove the transaction part from here. We should do it separately. I am not sure yet if the transaction will only be concerned about levelDB.

pkg/storagev2/leveldb/store.go
```diff
@@ -0,0 +1,292 @@
...
+
+// NewDatastore returns a new datastore backed by leveldb
+// for path == "", an in memory backend will be chosen (TODO)
+func NewLevelDBStore(path string, opts *opt.Options) (storageV2.Store, error) {
```
aloknerurkar, line 33: Right now someone using it will do:
`st, err := leveldbstore.NewLevelDBStore()`

I think this can be improved:
`st, err := leveldb.NewStore()` or `st, err := leveldbstore.New()`

pkg/storagev2/leveldb/store.go
```diff
@@ -0,0 +1,292 @@
...
+
+	ds := Store{
+		DB:   db,
+		path: path,
+	}
+	return &ds, nil
+}
+
+func (s *Store) Count(key storageV2.Key) (c int, err error) {
+	keys := util.BytesPrefix([]byte(key.Namespace() + "/"))
```
aloknerurkar, line 58: I think you will need the lock here as well. If someone closes the DB while count is in process this could be a problem.

pkg/storagev2/leveldb/store.go
```diff
@@ -0,0 +1,292 @@
...
+	return
+}
+
+func (s *Store) Close() (err error) {
+	s.closeLk.Lock()
+	defer s.closeLk.Unlock()
+	return s.DB.Close()
+}
+
+// A leveldb transaction embedding the accessor backed by the transaction.
```
aloknerurkar, line 238: Lets remove this as well.

pkg/storagev2/leveldb/store.go
```diff
@@ -0,0 +1,292 @@
...
+
+	if err := iter.Error(); err != nil {
+		retErr = multierror.Append(retErr, err)
+	}
+
+	return retErr.ErrorOrNil()
+}
+
+func apply(k string, v []byte, q storageV2.Query, fn func(storageV2.Result) (bool, error)) (stop bool, err error) {
+	namespace := q.Factory().Namespace()
```
aloknerurkar, line 196: So this can be avoided. Basically here for each iteration we are allocating 2 objects. The namespace is constant, so we dont need to use Factory here.

pkg/storagev2/leveldb/store.go
```diff
@@ -0,0 +1,292 @@
...
+	prefix := namespace + "/"
+
+	keys := util.BytesPrefix([]byte(prefix))
+
+	iter := s.DB.NewIterator(keys, nil)
+
+	switch q.Order {
+	case storageV2.KeyAscendingOrder:
+		for iter.Next() {
```
aloknerurkar, line 159: So I think this can be simplified. Both the `Prev` and `Next` functions have same signature. So we can have just 1 for loop basically.
```
nextF := iter.Next

if q.Order == KeyOrderDescending {
    nextF = func() bool {
        nextF = iter.Prev
        return iter.Last()
   }
}
```

aloknerurkar: I think this way we dont need a separate function for `apply` maybe.

pkg/storagev2/leveldb/store.go
```diff
@@ -0,0 +1,234 @@
...
+
+	for iter.Next() {
+		c++
+	}
+
+	return c, iter.Error()
+}
+
+func (s *Store) Put(item storageV2.Item) error {
+	s.closeLk.RLock()
```
istae, line 69: Why are we locking the db for every operation?

mrekucci (commented):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3203)* status: all files reviewed, 8 unresolved discussions (waiting on @notanatol)

___
*[`pkg/storagev2/leveldb/store.go` line 55 at r2](https://reviewable.io/reviews/ethersphere/bee/3203#-NA3CMwG8nkWWa08A8vG:-NA3CMwG8nkWWa08A8vH:b-qleppx) ([raw file](https://github.com/ethersphere/bee/blob/fe7d9dd96b9cbe260b23cbc784d679d2d7fe5991/pkg/storagev2/leveldb/store.go#L55)):*
> ```Go
> }
> 
> func (s *Store) Count(key storageV2.Key) (c int, err error) {
> ```

Just a nitpick, avoid naked returns if possible.
___
*[`pkg/storagev2/leveldb/store.go` line 74 at r2](https://reviewable.io/reviews/ethersphere/bee/3203#-NA3EjAH95n-D5cXtkCk:-NA3EjAH95n-D5cXtkCl:bis3q8u) ([raw file](https://github.com/ethersphere/bee/blob/fe7d9dd96b9cbe260b23cbc784d679d2d7fe5991/pkg/storagev2/leveldb/store.go#L74)):*
> ```Go
> 	value, err := item.Marshal()
> 	if err != nil {
> 		return fmt.Errorf("failed serializing %w", err)
> ```

Add `:` before the `%w` verb to distinguish where the error starts.
___
*[`pkg/storagev2/leveldb/store.go` line 90 at r2](https://reviewable.io/reviews/ethersphere/bee/3203#-NA3CuVJEvPii5L_HcK3:-NA3CuVKCyINy24b4Lcs:bnxi3yd) ([raw file](https://github.com/ethersphere/bee/blob/fe7d9dd96b9cbe260b23cbc784d679d2d7fe5991/pkg/storagev2/leveldb/store.go#L90)):*
> ```Go
> 	val, err := s.DB.Get(key, nil)
> 	if err != nil {
> 		if err == leveldb.ErrNotFound {
> ```

Check for errors with `errors.Is`, since the error that you're looking for can be nested.
___
*[`pkg/storagev2/leveldb/store.go` line 103 at r2](https://reviewable.io/reviews/ethersphere/bee/3203#-NA3DFGeCBTxbD2k5zpH:-NA3DFGeCBTxbD2k5zpI:b-fugsz6) ([raw file](https://github.com/ethersphere/bee/blob/fe7d9dd96b9cbe260b23cbc784d679d2d7fe5991/pkg/storagev2/leveldb/store.go#L103)):*
> ```Go
> }
> 
> func (s *Store) Has(sKey storageV2.Key) (exists bool, err error) {
> ```

Not sure why are these return parameters named here and also in the methods below.
___
*[`pkg/storagev2/leveldb/store.go` line 120 at r2](https://reviewable.io/reviews/ethersphere/bee/3203#-NA3DuKM7T7VOVCavTyl:-NA3DuKM7T7VOVCavTym:b29kvco) ([raw file](https://github.com/ethersphere/bee/blob/fe7d9dd96b9cbe260b23cbc784d679d2d7fe5991/pkg/storagev2/leveldb/store.go#L120)):*
> ```Go
> 	val, err := s.DB.Get(key, nil)
> 	if err != nil {
> 		if err == leveldb.ErrNotFound {
> ```

The same as above, check for errors with `errors.Is`.

pkg/storagev2/leveldb/store.go
notanatol: We want to wait for all operations to complete before closing the database - hence the `RLock` on methods and `Lock` on the Close method.
However it looks like only the Iterate/Count operations would need this locking, because they execute multiple operations in the scope of one call, so closing the db amidst the op could lead to inconsistent behavior.
In all the other cases we could leverage the built in concurrency mechanism provided by levelDB, this is my impression after looking into library code, wdyt @aloknerurkar ?

pkg/storagev2/leveldb/store.go
notanatol: we're allocating on the stack, since the var does not escape func scope, so no pressure on gc.
but yes, I can pass it in as a value

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3203)* status: 2 of 3 files reviewed, 7 unresolved discussions (waiting on @aloknerurkar, @mrekucci, and @notanatol)

___
*[`pkg/storagev2/leveldb/store.go` line 29 at r1](https://reviewable.io/reviews/ethersphere/bee/3203#-N9fXTMYLUieIzVgmOzD-r1-29:-NA5UB6ZEIfKbFTHkOHG:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/527a962dab391188831b946c9d14f930f0432cd6/pkg/storagev2/leveldb/store.go#L29)):*

Let's remove the transaction part from here. We should do it separately. I am not sure yet if the transaction will only be concerned about levelDB.
</blockquote></details>

Done.
___
*[`pkg/storagev2/leveldb/store.go` line 33 at r1](https://reviewable.io/reviews/ethersphere/bee/3203#-N9fXTMYLUieIzVgmOzD-r1-33:-NA5UCQZ2MfiHS0LIlUX:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/527a962dab391188831b946c9d14f930f0432cd6/pkg/storagev2/leveldb/store.go#L33)):*

Right now someone using it will do:
`st, err := leveldbstore.NewLevelDBStore()`

I think this can be improved:
`st, err := leveldb.NewStore()` or `st, err := leveldbstore.New()`
</blockquote></details>

Done.
___
*[`pkg/storagev2/leveldb/store.go` line 58 at r1](https://reviewable.io/reviews/ethersphere/bee/3203#-N9fXTMYLUieIzVgmOzD-r1-58:-NA5Uc6k7xry6TTUlSQb:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/527a962dab391188831b946c9d14f930f0432cd6/pkg/storagev2/leveldb/store.go#L58)):*

I think you will need the lock here as well. If someone closes the DB while count is in process this could be a problem.
</blockquote></details>

Done.
___
*[`pkg/storagev2/leveldb/store.go` line 159 at r1](https://reviewable.io/reviews/ethersphere/bee/3203#-N9fXTMYLUieIzVgmOzD-r1-159:-NA5VEabD154zeEnnUHS:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/527a962dab391188831b946c9d14f930f0432cd6/pkg/storagev2/leveldb/store.go#L159)):*

I think this way we dont need a separate function for `apply` maybe.
</blockquote></details>

Done.

pkg/storagev2/leveldb/store.go
aloknerurkar: So for the `Close` function we need to take the Lock.

But for all the other functions we would need the RLock. This is to prevent Closing of the DB while any operations are happening. RLock can be held by multiple routines, so I dont think there would be any problem here.

pkg/storagev2/leveldb/store.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	return &ds, nil
+}
+
+func (s *Store) Count(key storageV2.Key) (int, error) {
+	s.closeLk.RLock()
+	defer s.closeLk.RUnlock()
+
+	keys := util.BytesPrefix([]byte(key.Namespace() + "/"))
```
aloknerurkar, line 61: So I would make the `/` constant. It is highly unlikely but its easier if we want to change this in future.

pkg/storagev2/leveldb/store.go
```diff
@@ -0,0 +1,238 @@
...
+	iter := s.DB.NewIterator(keys, nil)
+
+	var c int
+	for iter.Next() {
+		c++
+	}
+
+	iter.Release()
```
aloknerurkar, line 69: So I am not sure how this works. Is calling `iter.Error()` valid after release?

Might be better to do it in defer.

notanatol: the example in the official documentation from goleveldb is like so
```
iter.Release()
err = iter.Error()
```
so deferring would go against the recommended approach

pkg/storagev2/leveldb/store.go
notanatol: fixed

pkg/storagev2/leveldb/store.go
```diff
@@ -0,0 +1,238 @@
...
+	storageV2 "github.com/ethersphere/bee/pkg/storagev2"
+	"github.com/hashicorp/go-multierror"
+	"github.com/syndtr/goleveldb/leveldb"
+	ldbErrors "github.com/syndtr/goleveldb/leveldb/errors"
+	"github.com/syndtr/goleveldb/leveldb/opt"
+	"github.com/syndtr/goleveldb/leveldb/storage"
+	"github.com/syndtr/goleveldb/leveldb/util"
+)
+
+type Store struct {
```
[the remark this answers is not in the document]
notanatol: fixed

pkg/storagev2/leveldb/store.go
```diff
@@ -0,0 +1,240 @@
...
+			err = newItem.Unmarshal(nextVal)
+			res = &storageV2.Result{Entry: newItem}
+		}
+
+		if err != nil {
+			retErr = multierror.Append(retErr, fmt.Errorf("failed unmarshaling: %w", err))
+			break
+		}
+
+		if res == nil {
```
aloknerurkar, line 202: Not sure if this will ever happen.

pkg/storagev2/leveldb/store.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/syndtr/goleveldb/leveldb/util"
+)
+
+type store struct {
+	DB      *leveldb.DB
+	path    string
+	closeLk sync.RWMutex
+}
+
+const forwardSlash = "/"
```
aloknerurkar, line 28: I guess calling it `forwardSlash` is restrictive. I would rather just say `separator`

pkg/storagev2/leveldb/store.go
notanatol: this is just a future protection in case we add a new `Query` kind that doesn't match any switch cases.
this way it will return a meaningful error instead of panicking on line `:207`

mrekucci (commented):
> Reviewed 1 of 2 files at r5, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3203)* status: all files reviewed, 14 unresolved discussions (waiting on @aloknerurkar, @istae, and @notanatol)

___
*[`pkg/storagev2/leveldb/store.go` line 18 at r7](https://reviewable.io/reviews/ethersphere/bee/3203#-NA9OpOrEz9frUE-WAMG:-NA9OpOrEz9frUE-WAMH:b-n0wqjs) ([raw file](https://github.com/ethersphere/bee/blob/f5fe70bd518db34fcdf75b36f96a5718fed30355/pkg/storagev2/leveldb/store.go#L18)):*
> ```Go
> 	ldbErrors "github.com/syndtr/goleveldb/leveldb/errors"
> 	"github.com/syndtr/goleveldb/leveldb/opt"
> 	"github.com/syndtr/goleveldb/leveldb/storage"
> ```

Since we're already using `ldbError`, I'd suggest also renaming the leveldb storage import to `ldbStorage` and drop the `storageV2` so we're consistent.
___
*[`pkg/storagev2/leveldb/store.go` line 23 at r7](https://reviewable.io/reviews/ethersphere/bee/3203#-NA9PJwdC0IoZgyl0r6E:-NA9PJwdC0IoZgyl0r6F:b-olz4km) ([raw file](https://github.com/ethersphere/bee/blob/f5fe70bd518db34fcdf75b36f96a5718fed30355/pkg/storagev2/leveldb/store.go#L23)):*
> ```Go
> 
> type store struct {
> 	DB      *leveldb.DB
> ```

Not sure why the `DB` field is exported?
___
*[`pkg/storagev2/leveldb/store.go` line 208 at r7](https://reviewable.io/reviews/ethersphere/bee/3203#-NA9PqUa-7w_V3GlPnyl:-NA9PqUbB3X01LTEoEE4:b2t44c) ([raw file](https://github.com/ethersphere/bee/blob/f5fe70bd518db34fcdf75b36f96a5718fed30355/pkg/storagev2/leveldb/store.go#L208)):*
> ```Go
> 
> 		if stop, err := fn(*res); err != nil {
> 			retErr = multierror.Append(retErr, fmt.Errorf("failed in iterate function: %w", err))
> ```

Can the error message be restructured so it makes more sense?

mrekucci (approved):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3203)* status: all files reviewed, 13 unresolved discussions (waiting on @aloknerurkar, @istae, and @notanatol)

___
*[`pkg/storagev2/leveldb/store.go` line 20 at r8](https://reviewable.io/reviews/ethersphere/bee/3203#-NA9ZE8wAX0YO46fNqwO:-NA9ZE8wAX0YO46fNqwP:b-fm2wrq) ([raw file](https://github.com/ethersphere/bee/blob/07a5c5d4e96e29692923e0fbcf616c639a2aaa5a/pkg/storagev2/leveldb/store.go#L20)):*
> ```Go
> 	"github.com/syndtr/goleveldb/leveldb/util"
> 
> 	storage "github.com/ethersphere/bee/pkg/storagev2"
> ```

So the storage doesn't need to be renamed since the package is already named `storage` but it just resigns in the `storagev2` directory 'till we don't decide to do the refactoring.

notanatol (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3203)* status: all files reviewed, 13 unresolved discussions (waiting on @aloknerurkar, @istae, and @mrekucci)

___
*[`pkg/storagev2/leveldb/store.go` line 238 at r1](https://reviewable.io/reviews/ethersphere/bee/3203#-N9fXTMYLUieIzVgmOzD-r1-238:-NA9X_cuC-XJzomN4UVE:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/527a962dab391188831b946c9d14f930f0432cd6/pkg/storagev2/leveldb/store.go#L238)):*

Lets remove this as well.
</blockquote></details>

Done.
___
*[`pkg/storagev2/leveldb/store.go` line 28 at r6](https://reviewable.io/reviews/ethersphere/bee/3203#-N9fXTMYLUieIzVgmOzD-r6-28:-NA9asJk7YC8jjWv3n_Y:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/362bb166e4db004f24a77f0aa8a024a62022de3a/pkg/storagev2/leveldb/store.go#L28)):*

I guess calling it `forwardSlash` is restrictive. I would rather just say `separator`
</blockquote></details>

Done.
___
*[`pkg/storagev2/leveldb/store.go` line 18 at r7](https://reviewable.io/reviews/ethersphere/bee/3203#-NA9OpOrEz9frUE-WAMG:-NA9XFNG8jFZbUxLuRF5:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/f5fe70bd518db34fcdf75b36f96a5718fed30355/pkg/storagev2/leveldb/store.go#L18)):*

Since we're already using `ldbError`, I'd suggest also renaming the leveldb storage import to `ldbStorage` and drop the `storageV2` so we're consistent.
</blockquote></details>

fixed
___
*[`pkg/storagev2/leveldb/store.go` line 23 at r7](https://reviewable.io/reviews/ethersphere/bee/3203#-NA9PJwdC0IoZgyl0r6E:-NA9aPfNCGxk15EGPWts:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/f5fe70bd518db34fcdf75b36f96a5718fed30355/pkg/storagev2/leveldb/store.go#L23)):*

Not sure why the `DB` field is exported?
</blockquote></details>

fixed
___
*[`pkg/storagev2/leveldb/store.go` line 208 at r7](https://reviewable.io/reviews/ethersphere/bee/3203#-NA9PqUa-7w_V3GlPnyl:-NA9ap2wESTHFHcdzwCm:brmdyec) ([raw file](https://github.com/ethersphere/bee/blob/f5fe70bd518db34fcdf75b36f96a5718fed30355/pkg/storagev2/leveldb/store.go#L208)):*

Can the error message be restructured so it makes more sense?
</blockquote></details>

I gave it another try



_Code quote:_
```Go
fmt.Errorf("failed in iterate function: %w", err))
```
___
*[`pkg/storagev2/leveldb/store.go` line 20 at r8](https://reviewable.io/reviews/ethersphere/bee/3203#-NA9ZE8wAX0YO46fNqwO:-NA9aO8g3wEBnnYEHZp1:b-7ql1td) ([raw file](https://github.com/ethersphere/bee/blob/07a5c5d4e96e29692923e0fbcf616c639a2aaa5a/pkg/storagev2/leveldb/store.go#L20)):*

So the storage doesn't need to be renamed since the package is already named `storage` but it just resigns in the `storagev2` directory 'till we don't decide to do the refactoring.
</blockquote></details>

fixed

pkg/storagev2/leveldb/store.go
istae: The reading of the code is strange. We are doing read-locks everywhere and then a write lock on Close for Close to wait on all locks, I get the idea but it's unconventional.

Why not go with something more readable like:

```
db.wg.Add(1)
defer db.wg.Done()
```

and then in `Close()`
```
db.wg.Wait()
```

istae: Sure, til the WUnlock is done, any RLocks are disallowed BUT if there are db operations waiting for the lock to be released by the Close WUnlock, then, the db operations will occur on a closed db after they acquire the lock, which may lead to a panic.

The correct implementation here is that packages relying on the localstore are closed first such that when the localstore is closed, there are no on the fly requests on the db, in which case, the wait group waiting for requests to finish may be what we need.

aloknerurkar: > Sure, til the WUnlock is done, any RLocks are on hold BUT if there are db operations waiting for the lock to be released by the Close WUnlock, then, the db operations will occur on a closed db after they acquire the lock, which may lead to a panic.

This would not lead to panic as Close does not set the DB to nil. It should return an error I think.

> 
> The correct implementation here is that packages relying on the localstore are closed first such that when the localstore is closed, there are no on the fly requests on the db, in which case, the wait group waiting for requests to finish may be what we need. With the R and W locks, we don't have that guarantee of order of operations. 

So the case that you mentioned could happen with the Wg as well. Even after close is done an op could come in and it would increment the Wg counter and access a closed DB right? 

I mean we can never prevent user errors in using the package. If the Close is not used correctly both cases have issues in my opinion.

istae: > This would not lead to panic as Close does not set the DB to nil. It should return an error I think.
 Sure, don't really know.
 
> So the case that you mentioned could happen with the Wg as well. Even after close is done an op could come in and it would increment the Wg counter and access a closed DB right?

Given that we we do the rewrite correctly, and that no db operations should occurs after the close. you could have inflight requests when the db Close is called. The mutex version does not protect you like the wait group does. With the wait group, you wait for all operations to finish first at least.

aloknerurkar: >  > This would not lead to panic as Close does not set the DB to nil. It should return an error I think.
>  
> Sure, don't really know. But the final writes on shutdown to the db could be critical, like the stamp issuers data we write on Close. So not panicking is not good enough.
>  

But so the Wg solution does not fix this. I mean this is concurrency, nothing is guaranteed. For eg when user initiates a Close it could happen that no operation was in progress so DB is closed immediately. But just after the instant Close returns, some writes are initiated. This would cause the same errors.

> > So the case that you mentioned could happen with the Wg as well. Even after close is done an op could come in and it would increment the Wg counter and access a closed DB right?
> 
> Given that we we do the rewrite correctly, and that no db operations should occurs after the close. you could have inflight requests when the db Close is called. The mutex version does not protect you like the wait group does. With the wait group, you wait for all operations to finish first at least. 
> 
> It's also more readable :)

So as I mentioned above, both solutions don't solve the problem you are referring to. If the usage is correct I would say both solutions would give you the right answer and if not both would be wrong.

In terms of readability, I am not sure. Maybe we can ask others. I would usually not use waitgroups unless I am working with go routines. Also the Close would be blocking and could cause deadlocks. So we would need to handle that again like we do all around the code right now by Waiting for the waitgroup in a different routine and then returning early if it doesn't finish in time otherwise we can hang in shutdown.

PR #3214, opened 2022-08-22
refactor: simplify input validation

### LINKED ISSUE #3163: Simplify HTTP API input parameters check
Currently, a parameters check is performed at the beginning of each HTTP handler. This parameter check is very repetitive and unnecessarily increases the size of the handler's body. Find a way to reduce the repetitive code. Consider using the https://github.com/go-playground/validator library or any other library that accomplishes this task.

umerm-work:
### Checklist

### Description
Simplify input to remove repetitiveness.

Closes #3163 

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3214)


mrekucci (changes requested):
So a couple of things that will make it better:

Logging should not be part of the methods you've attached to the `Service` type. As a result, you don't need to attach all the methods to the `Service` type. 

The `parseAndValidate` method should be a combination of two steps:
1) parse
2) validate

Consider the following:
```Golang
func (s *Service) parseAndValidate(input, output interface{}) error {
	if err := parse(input, output); err != nil {
		return err
	}
	return validate(output)
}
```

The `parse` function should exist on its own and should be tested. The signature of the `parse` function should be: `func parse(input, output interface{}) error`.

For validation, use the library that was recommended: https://github.com/go-playground/validator

Parse input: you should expect three types for which you need to perform type assertion:
1) `mux.Vars`, which is `map[string]string`
2) `url.Values`, which is `map[string][]string`
3) `http.Header`, which is `map[stirng][]string`

Parse output (and validate input): should be a structure whose fields contain the `parse` (and `validate` if used) tag(s). The first value of the `parse` structure tag should be the name of the key input map whos value will be filled into this structure field. The second value should be a post parsing hook, which needs to be applied before the value is set. E.g.:

```Golang
pathParams := struct {
	Amount  int64          `parse:"amount"`
	Depth   uint8          `parse:"depth" validate:"min=0,max=31"`
	TxHash  common.Hash    `parse:"hash"`
	Address common.Address `parse:"address"`
	BatchID []byte         `parse:"id,hexToBytes" validate:"len=64"`
}{}
```

Based on this, you can generate errors that have a certain structure, e.g. "parsing `depth` failed: invalid syntax", etc... I recommend that you create your own error type for this purpose, see [this](https://github.com/golang/go/blob/master/src/strconv/atoi.go#L23-L34) for an inspiration.

umerm-work (commented):
In most cases, validation is applied before parsing in code. https://github.com/ethersphere/bee/blob/a2be1d7b816aa14cdff3125ebf66265e898ad9f7/pkg/api/postage.go#L509

This will require a lot of modification in test cases. I tried to keep the test cases intact. 

So if we want to apply validation, we must do it before parsing. 

The validation library doesn't support our custom types, so it's better to have a custom implementation.

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3214#-:-NC5jgoaDpKAvyykV6LZ:b-aebw1t)* status: 0 of 15 files reviewed, all discussions resolved (waiting on @aloknerurkar and @notanatol)

mrekucci:
> This will require a lot of modification in test cases. I tried to keep the test cases intact.

It will require changing the returned test results, which is perfectly fine since we want to unify the responses anyway.

> So if we want to apply validation, we must do it before parsing.

Before parsing you only have length information and this is only relevant in some cases, otherwise, you need to parse the value successfully first and then perform validation, so validation after parsing makes more sense.

> The validation library doesn't support our custom types, so it's better to have a custom implementation.

You can use custom validators from the validation library: "Cross Field and Cross Struct validations by using validation tags or custom validators."

mrekucci:
Closed in favor of #3344

PR #3216, opened 2022-08-23
chore: simplyfy formatting for byte slice

### LINKED ISSUE #3157: Simplyfy logging of `fmt.Sprintf("%x", value)` values
Currently, most `[]byte` values are logged using the `fmt.Sprintf("%x", value)` call, review these records and try to remove the formatting of these values. At the same time, check that the output remains the same, if not, then implement the `log.Marshaler` interface for these types, which will produce the same output as in the case of `fmt.Sprintf`.

vladopajic:
### Checklist

### Description
Closes #3157

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3216)

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -9,7 +9,6 @@ package api
 import (
 	"context"
 	"crypto/ecdsa"
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -364,12 +363,12 @@ func requestDeferred(r *http.Request) (bool, error) {
 	return true, nil
 }
 
-func requestPostageBatchId(r *http.Request) ([]byte, error) {
+func requestPostageBatchId(r *http.Request) (swarm.BatchID, error) {
 	if h := strings.ToLower(r.Header.Get(SwarmPostageBatchIdHeader)); h != "" {
 		if len(h) != 64 {
 			return nil, errInvalidPostageBatch
 		}
-		b, err := hex.DecodeString(h)
+		b, err := swarm.HexToBatchID(h)
 		if err != nil {
 			return nil, errInvalidPostageBatch
 		}
--- a/pkg/swarm/swarm.go
+++ b/pkg/swarm/swarm.go
@@ -161,9 +161,22 @@ type Chunk interface {
 	Equal(Chunk) bool
 }
 
+// BatchID type represents unique batch identifier.
+type BatchID []byte
+
+// String implements fmt.Stringer.
+func (b BatchID) String() string {
+	return fmt.Sprintf("%x", b.Bytes())
+}
+
+// Bytes returns raw []bytes value for this BatchID.
+func (b BatchID) Bytes() []byte {
+	return []byte(b)
+}
+
 // Stamp interface for postage.Stamp to avoid circular dependency
 type Stamp interface {
-	BatchID() []byte
+	BatchID() BatchID
 	Index() []byte
 	Sig() []byte
 	Timestamp() []byte
--- a/pkg/swarm/swarm.go
+++ b/pkg/swarm/swarm.go
@@ -166,7 +166,7 @@ type BatchID []byte
 
 // String implements fmt.Stringer.
 func (b BatchID) String() string {
-	return fmt.Sprintf("%x", b.Bytes())
+	return hex.EncodeToString(b.Bytes())
 }
 
 // Bytes returns raw []bytes value for this BatchID.
--- a/pkg/swarm/swarm.go
+++ b/pkg/swarm/swarm.go
@@ -174,6 +174,16 @@ func (b BatchID) Bytes() []byte {
 	return []byte(b)
 }
 
+// HexToBatchID returns BatchID form supplied hex string.
+func HexToBatchID(s string) (BatchID, error) {
+	b, err := hex.DecodeString(s)
+	if err != nil {
+		return nil, err
+	}
+
+	return BatchID(b), nil
+}
+
 // Stamp interface for postage.Stamp to avoid circular dependency
 type Stamp interface {
 	BatchID() BatchID
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -443,7 +443,7 @@ func (c *command) configureSigner(cmd *cobra.Command, logger log.Logger) (config
 	if err != nil {
 		return nil, err
 	}
-	logger.Info("using ethereum address", "address", fmt.Sprintf("%x", overlayEthAddress))
+	logger.Info("using ethereum address", "address", overlayEthAddress)
 
 	return &signerConfig{
 		signer:           signer,
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"crypto/ecdsa"
 	_ "embed"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"os"
@@ -414,7 +415,7 @@ func (c *command) configureSigner(cmd *cobra.Command, logger log.Logger) (config
 		publicKey = &swarmPrivateKey.PublicKey
 	}
 
-	logger.Info("swarm public key", "public_key", fmt.Sprintf("%x", crypto.EncodeSecp256k1PublicKey(publicKey)))
+	logger.Info("swarm public key", "public_key", hex.EncodeToString(crypto.EncodeSecp256k1PublicKey(publicKey)))
 
 	libp2pPrivateKey, created, err := keystore.Key("libp2p", password)
 	if err != nil {
@@ -436,7 +437,7 @@ func (c *command) configureSigner(cmd *cobra.Command, logger log.Logger) (config
 		logger.Debug("using existing pss key")
 	}
 
-	logger.Info("pss public key", "public_key", fmt.Sprintf("%x", crypto.EncodeSecp256k1PublicKey(&pssPrivateKey.PublicKey)))
+	logger.Info("pss public key", "public_key", hex.EncodeToString(crypto.EncodeSecp256k1PublicKey(&pssPrivateKey.PublicKey)))
 
 	// postinst and post scripts inside packaging/{deb,rpm} depend and parse on this log output
 	overlayEthAddress, err := signer.EthereumAddress()
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -230,7 +230,7 @@ func (s *Service) postageGetAllStampsHandler(w http.ResponseWriter, _ *http.Requ
 		}
 
 		batches = append(batches, postageBatchResponse{
-			BatchID:       b.ID,
+			BatchID:       b.ID.Bytes(),
 			Value:         bigint.Wrap(b.Value),
 			Start:         b.Start,
 			Owner:         b.Owner,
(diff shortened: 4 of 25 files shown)


pkg/swarm/swarm.go
```diff
@@ -161,9 +161,32 @@ type Chunk interface {
 	Equal(Chunk) bool
 }
 
+// BatchID type represents unique batch identifier.
+type BatchID []byte
```
notanatol, line 165: really nice! we should encourage more proper use of go types, thanks

notanatol (approved):

pkg/swarm/swarm.go
vladopajic: Thanks!

Yes, I feel that project would benefit form using types and types aliases. This would lead to better types safety and better code readability.

mrekucci (changes requested):
> Reviewed 19 of 24 files at r1, 1 of 2 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3216)* status: all files reviewed, 2 unresolved discussions (waiting on @notanatol and @vladopajic)

___
*[`pkg/swarm/swarm.go` line 178 at r3](https://reviewable.io/reviews/ethersphere/bee/3216#-NAARsr0ECPYkyEOoYLA:-NAARsr0ECPYkyEOoYLB:b-walox) ([raw file](https://github.com/ethersphere/bee/blob/cd56099ed1ffdcfa119919b95aa5355e566cb3ab/pkg/swarm/swarm.go#L178)):*
> ```Go
> 
> // HexToBatchID returns BatchID form supplied hex string.
> func HexToBatchID(s string) (BatchID, error) {
> ```

I'd suggest renaming the function to `ParseBatchID` since this is also the convention in the stdlib.
___
*[`pkg/swarm/swarm_test.go` line 13 at r3](https://reviewable.io/reviews/ethersphere/bee/3216#-NAASEiV-b5SQAl0oOWc:-NAASEiV-b5SQAl0oOWd:buh0ro0) ([raw file](https://github.com/ethersphere/bee/blob/cd56099ed1ffdcfa119919b95aa5355e566cb3ab/pkg/swarm/swarm_test.go#L13)):*
> ```Go
> 	"testing"
> 
> 	"github.com/stretchr/testify/assert"
> ```

Can you drop this library and use the Golang standard testing package so the tests are consistent across the codebase.
___
*[`pkg/swarm/swarm_test.go` line 115 at r3](https://reviewable.io/reviews/ethersphere/bee/3216#-NAASmeo7qMaNRbSBrOC:-NAASmeo7qMaNRbSBrOD:b-88dah3) ([raw file](https://github.com/ethersphere/bee/blob/cd56099ed1ffdcfa119919b95aa5355e566cb3ab/pkg/swarm/swarm_test.go#L115)):*
> ```Go
> }
> 
> func TestBatchID(t *testing.T) {
> ```

The test should state what you are testing. If it is the `HexToBatchID` function, then the test should include that in its name. If the goal of the test is to test the entire `BatchID` implementation, then I would suggest using named subtests in which you test different parts.

pkg/api/api.go
```diff
@@ -364,12 +363,12 @@ func requestDeferred(r *http.Request) (bool, error) {
...
 }
 
-func requestPostageBatchId(r *http.Request) ([]byte, error) {
+func requestPostageBatchId(r *http.Request) (swarm.BatchID, error) {
 	if h := strings.ToLower(r.Header.Get(SwarmPostageBatchIdHeader)); h != "" {
 		if len(h) != 64 {
 			return nil, errInvalidPostageBatch
 		}
-		b, err := hex.DecodeString(h)
+		b, err := swarm.HexToBatchID(h)
```
aloknerurkar, line 371: I dont understand why this function is needed?

Original function is simple, readable enough. Secondly, it is doing the exact same thing.

pkg/swarm/swarm.go
```diff
@@ -161,9 +161,32 @@ type Chunk interface {
...
+		return nil, err
+	}
+
+	return BatchID(b), nil
+}
+
 // Stamp interface for postage.Stamp to avoid circular dependency
 type Stamp interface {
-	BatchID() []byte
+	BatchID() BatchID
```
aloknerurkar, line 189: I dont see why this is required. This additional type would impose some additional code for no reason. I dont believe we need to define this type at all.

BatchID is not a new type. It is the same hash that we use everywhere. So I dont see why we would need a new type for this. Going to string and back is simple one-liner with hex package

aloknerurkar (commented):
Again, I dont really understand why we need this big change set. From the PR I see the only change we need is in `cmd/start.go` file. I dont see why we need this BatchID type. There is no separate use case for it apart from converting to and from string. Which is a one-liner anyway. Having additional types adds bytes to the build. I think we should avoid this if it is not required.

pkg/swarm/swarm.go
vladopajic: Defining domain specific types gives you type safety, enables better abstraction, encapsulation and makes code more readable. 

This kind of mechanism was added to language for vary this reason. 

----

This type does impose few lines of additional code. But at the same time it removes a lot of boilerplate code where developers need to format it when writing to logs.

pkg/api/api.go
vladopajic: `swarm.HexToBatchID`  returns `BatchID` from provided hex string, while 
`hex.DecodeString` returns `[]byte` from provided hex string.

vladopajic:
@mrekucci 

> I'd suggest renaming the function to ParseBatchID since this is also the convention in the stdlib.

Convention for `HexToBatchID` name was chosen analogously to ethereum functions `HexToHash` and `HextToAddress`.
I personally prefer this convention, because it implies that hex is converted to `BatchID`.

Also semantically this functionality is not `parsing` but `converting`. Conversion is process of translating object from one context to another where both object could be considered equal - eg. `123` and `"123"`. Parsing is process where we translate object from one context to another, such that both object could be vary different thing.

I would consider avoiding explicit connotation in this case since.

---

> Can you drop this library and use the Golang standard testing package so the tests are consistent across the codebase.

I would consider adopting this library for writing our tests because it provides many handy tools for writing tests. Also support is vary good (vary popular and has many contributors). 

---

On the side note, it would be easier for use to communicate if comments are made separately for each point using github's comment functionality.

pkg/swarm/swarm.go
aloknerurkar: > Defining domain specific types gives you type safety, enables better abstraction, encapsulation and makes code more readable. 
> 
> This kind of mechanism was added to language for vary this reason. 
> 

So I do think there could be some use of it. But in this specific instance I don't think this gives us anything additional. 

The only use for it is to print in string format and hex.EncodeToString is already a one-liner and simple way to understand.

Adding more types increases the size of binary. Having dependencies for types across packages is also not favourable. I don't see any rewards in this case. When I look at some new BatchID type, someone might think this is some new construct, which it is not.

> ----
> 
> This type does impose few lines of additional code. But at the same time it removes a lot of boilerplate code where developers need to format it when writing to logs.

The actual change in `cmd/start.go` is the same. I don't see any boiler plate code saved?

aloknerurkar: BatchID is not a new type. It is essentially a swarm.Address in byte format. So adding a new type to swarm package is definitely not the right thing to do.

vladopajic: Alright. In that case `BatchID` could be type alias for `swarm.Address`. 

Although I would add that type alias later since currently `swarm.Address` is unnecessary boxed in a structure which would introduce more changes.

vladopajic: > Adding more types increases the size of binary. 
Please, provide me with info on exact increase of the binary in with this type. Also please calculate reduction of the binary where boilerplate calls for `hex.EncodeToString` function is not removed. Lastly we would need to consider current binary size to know impact.

> Having dependencies for types across packages is also not favourable. 
There is no problem importing right types at right places.

>  I don't see any rewards in this case
1) type safety - it's preventing developers to mix frogs with apples
2) provides additional functionality - in this case removes boilerplate code where developer needs to call `hex.EncodeToString` each time when this type is written to the logs
3) makes code easier to read - consider case when you have `b []bytes` and `b BatchID` what is more readable?

> When I look at some new BatchID type, someone might think this is some new construct, which it is not.
It's vary common practice in go to have types for such cases.

vladopajic: > The actual change in cmd/start.go is the same. I don't see any boiler plate code saved? 

`cmd/start.go` is bootstrap code, right? 

Boilerplate code is any code that gets repeated in various places with little to no variation.

aloknerurkar: > > Adding more types increases the size of binary. 
> 
> Please, provide me with info on exact increase of the binary in with this type. Also please calculate reduction of the binary where boilerplate calls for `hex.EncodeToString` function is removed. Lastly we would need to consider current binary size to know impact.

I am not sure if you understand what I mean.  Boilerplate code doesnt decrease binary size as such. Compiler anyway would optimize these things. Aliases add more symbols to the codebase. In any case this may not be significant.
> 
> > Having dependencies for types across packages is also not favourable. 
> 
> There is no problem importing right types at right places.
> 
> >  I don't see any rewards in this case
> 
> 1) type safety - it's preventing developers to mix frogs with apples
> 2) provides additional functionality - in this case removes boilerplate code where developer needs to call `hex.EncodeToString` each time when this type is written to the logs
> 3) makes code easier to read and adds more semantics - consider case when you have `b []bytes` and `b BatchID`, later code is much clearer to the developer then first.
> 
> > When I look at some new BatchID type, someone might think this is some new construct, which it is not.
> 
> It's vary common practice in go to have types for such cases. 
> 

It is very common practice to not do unnecessary refactorings on large codebases.

The only reason you were able to make these changes is, the code is already simple enough for anyone to read and understand in the parts that you have touched. As such the changes to add more types adds complexity as developers have to know about this additional type which has no advantage. []byte slice is already used widely and has strong stdlib support for any operations.

Secondly there is not benefit of your changes. Hex.EncodeToString is a one-liner. Anyone who has worked in golang would know about this function and how to use it.

Secondly the boilerplate code you mentioned is actually doing the same thing. The Decode function has no benefit as it doesn't save any lines of code. Developer just has to know about this additional function when he can and knows how to use the hex.Decode function.

Finally, I mentioned in your last PR as well. The correct process to do refactorings is to first create an issue in the bee repo to explain what you are planning to do and get comments from the team. We are the ones who have to maintain the code and solve the bugs. So such large refactorings without prior discussions should be avoided.

vladopajic: > It is very common practice to not do unnecessary refactorings on large codebases.

Code is either refactored or new project is started from scratch. Personally, I wouldn't consider this as large codebase.

> Secondly there is not benefit of your changes. Hex.EncodeToString is a one-liner. Anyone who has worked in golang would know about this function and how to use it.

With this change it is not necessary to to even do that (type does it for you 🙂). If we look at the root issue we could see that developers actually used `fmt.Sprintf` instead of `hex.EncodToString`. Certainly there are benefit to this (already mentioned).

vladopajic:
After seeing concerns for introduction of new type `swarm.BatchID` this PR will be closed in flavor for #3220 where `fmt.Sprintf` is replaced with ` hex.EncodeToString` and where possible formatting is left to type.

New issue [#3221](https://github.com/ethersphere/bee/issues/3221) was opened to to serve as panel / proposal for introduction of such type.

PR #3226, opened 2022-08-26
chore: format int numbers using strconv

vladopajic:
### Checklist

### Description
Occurrences of `fmt.Sprintf("%d",...)` were replaced with `strconv.FormatInt(..)`; later approach is already suggested in [go style guide](https://github.com/ethersphere/bee/blob/master/CODINGSTYLE.md#prefer-strconv-over-fmt).

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3226)

### DIFF
--- a/pkg/tags/tags.go
+++ b/pkg/tags/tags.go
@@ -22,7 +22,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"math/rand"
 	"sort"
 	"strconv"
@@ -162,11 +161,11 @@ func (ts *Tags) Delete(k interface{}) {
 func (ts *Tags) MarshalJSON() (out []byte, err error) {
 	m := make(map[string]*Tag)
 	ts.Range(func(k, v interface{}) bool {
-		key := fmt.Sprintf("%d", k)
 		val := v.(*Tag)
 
 		// don't persist tags which were already done
 		if !val.Done(StateSynced) {
+			key := strconv.Itoa(int(k.(uint32)))
 			m[key] = val
 		}
 		return true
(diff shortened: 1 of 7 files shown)


pkg/tags/tags.go
```diff
@@ -162,11 +161,11 @@ func (ts *Tags) Delete(k interface{}) {
 func (ts *Tags) MarshalJSON() (out []byte, err error) {
 	m := make(map[string]*Tag)
 	ts.Range(func(k, v interface{}) bool {
-		key := fmt.Sprintf("%d", k)
 		val := v.(*Tag)
 
 		// don't persist tags which were already done
 		if !val.Done(StateSynced) {
+			key := strconv.Itoa(int(k.(uint32)))
```
mrekucci, line 168: This conversion to `int` has a potential problem. Since `int` is a platform-dependent type (typically 32 bits wide on 32-bit systems and 64 bits wide on 64-bit systems), if the value of `k` is equal to `math.MaxUint32`, then converting `k` to `int` on 32-bit systems where `int` has the same size as `int32` will convert the value of `k` to `-1`. Therefore, I recommend using `strconv.FormatInt(int64(k.(uint32)), 10)` instead.

mrekucci (changes requested):

pkg/tags/tags.go
vladopajic: Yes, this is better way to format. I was thinking that this edge case could not be possible, since this formatting was already used just like in `tagKey` function. 

I have pushed fix correcting formatting in both places.

notanatol (approved):

mrekucci (approved):
Nice!

PR #3229, opened 2022-08-29
fix: response for slash in website index document

### LINKED ISSUE #2661: Upload: Having a slash `/` in `website-index-document` yields 500 instead of 400
This could be a 400 Bad Request instead, and the error message from the logs ("err: index document suffix must not include slash character") could appear in the response.

Also, when it is set to a nonexistent file, the response is 200 OK. This may be fine, more of a design choice.

swarmHaseeb:
### Checklist

### Description

Please refer to tagged issue for details. This pr sends correct response if there is a slash in website index document
### Related Issue (Optional)

This pr fixes #2661 

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3229)

### DIFF
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -147,7 +147,7 @@ func (s *Service) fileUploadHandler(w http.ResponseWriter, r *http.Request, stor
 		return
 	}
 
-	// filename cannot contain a "/" in prefix
+	// filename cannot contain a "/" in prefix because the specification is that we cannot start with slash but can have a slash at a later position
 	if strings.HasPrefix(fileName, "/") {
 		fileName = strings.TrimLeft(fileName, "/")
 	}
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -149,7 +149,7 @@ func (s *Service) fileUploadHandler(w http.ResponseWriter, r *http.Request, stor
 
 	// filename cannot contain a "/" in prefix because the specification is that we cannot start with slash but can have a slash at a later position
 	if strings.HasPrefix(fileName, "/") {
-		logger.Debug("bzz upload file: / in prefix not allowed ", "file_name", fileName, "error", err)
+		logger.Debug("bzz upload file: / in prefix not allowed", "file_name", fileName, "error", err)
 		logger.Error(nil, "bzz upload file: / in prefix not allowed", "file_name", fileName)
 		jsonhttp.BadRequest(w, "bzz upload file: / in prefix not allowed")
 		return
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -147,10 +147,12 @@ func (s *Service) fileUploadHandler(w http.ResponseWriter, r *http.Request, stor
 		return
 	}
 
+	if strings.HasPrefix(fileName, "/") {
+		fileName = strings.TrimLeft(fileName, "/")
+	}
 	rootMetadata := map[string]string{
 		manifest.WebsiteIndexDocumentSuffixKey: fileName,
 	}
-
 	err = m.Add(ctx, manifest.RootPath, manifest.NewEntry(swarm.ZeroAddress, rootMetadata))
 	if err != nil {
 		logger.Debug("bzz upload file: adding metadata to manifest failed", "file_name", fileName, "error", err)
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -149,7 +149,10 @@ func (s *Service) fileUploadHandler(w http.ResponseWriter, r *http.Request, stor
 
 	// filename cannot contain a "/" in prefix because the specification is that we cannot start with slash but can have a slash at a later position
 	if strings.HasPrefix(fileName, "/") {
-		fileName = strings.TrimLeft(fileName, "/")
+		logger.Debug("bzz upload file: / in prefix not allowed ", "file_name", fileName, "error", err)
+		logger.Error(nil, "bzz upload file: / in prefix not allowed", "file_name", fileName)
+		jsonhttp.BadRequest(w, "bzz upload file: / in prefix not allowed")
+		return
 	}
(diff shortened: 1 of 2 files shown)


aloknerurkar:
Not sure if this is correct. We can have `/` in the index document. The way it works is for each folder you can configure the index document.

So you can have 2 index files:
index.html
someFolder/index.html

I guess this is why the tests are failing.

swarmHaseeb (commented):
Yup, i have made changes as per the discussion. Can you please review again? 

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3229#-:-NBWR2683bmLRUOwGqnk:bw1i7i)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @agazso, @aloknerurkar, and @mrekucci)

mrekucci (commented):
A short note on why this is needed would be fine. Also, is this case covered by test?

swarmHaseeb (commented):
yeah that sounds good, ill add a comment to it. So like there is no specific test written for it but the existing tests do cover this because with previous approach some tests would fail with a slash / as i would directly throw an error if slash occured which did occur in some cases. Also this case is a non-blocker so if there is a problem it fixes it and then continues.

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3229#-:-NBXGG4cCF7Q3s2gB4eF:butoxmh)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @agazso and @aloknerurkar)

pkg/api/bzz.go
```diff
@@ -147,10 +147,13 @@ func (s *Service) fileUploadHandler(w http.ResponseWriter, r *http.Request, stor
 		return
 	}
 
+	// filename cannot contain a "/" in prefix
```
mrekucci, line 150: Can you also add the reason "why" it cannot contain a "/" in prefix.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3229)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @agazso, @aloknerurkar, and @mrekucci)

___
*[`pkg/api/bzz.go` line 150 at r2](https://reviewable.io/reviews/ethersphere/bee/3229#-NAdC0Y7zcqkSviewaXk-r2-150:-NBkxrJk09gJL5SxCbbA:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/d88fdbca8f21502c29590def66aae676b8dfcf2e/pkg/api/bzz.go#L150)):*

Can you also add the reason "why" it cannot contain a "/" in prefix.
</blockquote></details>

Done.

mrekucci (commented):
I think the requests should return `400` instead of trying to fix the names.

pkg/api/bzz.go
```diff
@@ -247,7 +253,9 @@ func (s *Service) bzzDownloadHandler(w http.ResponseWriter, r *http.Request) {
 		// NOTE: leave one slash if there was some
 		pathVar += "/"
 	}
-
+	if strings.HasPrefix(nameOrHex, "/") {
```
mrekucci, line 256: You should also return `400` in this case. Also, add a test case.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3229)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @agazso, @aloknerurkar, and @mrekucci)

___
*[`pkg/api/bzz.go` line 256 at r4](https://reviewable.io/reviews/ethersphere/bee/3229#-NAdC0Y7zcqkSviewaXk-r4-256:-NCsgaEu2Ps6ar2e1KEp:bwrghh2) ([raw file](https://github.com/ethersphere/bee/blob/19777289e743caf45325b82989e542b611c33e7c/pkg/api/bzz.go#L256)):*

You should also return `400` in this case. Also, add a test case.
</blockquote></details>

i have noticed that we do not require a logic for that here since mux.Vars removes any slash in prefix.

mrekucci (approved):

pkg/api/bzz.go
```diff
@@ -147,10 +147,16 @@ func (s *Service) fileUploadHandler(w http.ResponseWriter, r *http.Request, stor
 		return
 	}
 
+	// filename cannot contain a "/" in prefix because the specification is that we cannot start with slash but can have a slash at a later position
+	if strings.HasPrefix(fileName, "/") {
+		logger.Debug("bzz upload file: / in prefix not allowed", "file_name", fileName, "error", err)
+		logger.Error(nil, "bzz upload file: / in prefix not allowed", "file_name", fileName)
```
vladopajic, line 153: Is `logger.Error` necessary here? It looks like this is normal flow for handling requests which has early return for this case when filename starts with `/`. In other words error has not occurred in the system so `logger.Error` seams to be escalating severity.

vladopajic (approved):
LGTM; apart from minor comment

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3229)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @agazso, @aloknerurkar, @mrekucci, @umerm-work, and @vladopajic)

___
*[`pkg/api/bzz.go` line 153 at r5](https://reviewable.io/reviews/ethersphere/bee/3229#-NAdC0Y7zcqkSviewaXk-r5-153:-NCt3Rhy-1ij54oIaJwz:bi2a2ag) ([raw file](https://github.com/ethersphere/bee/blob/edadca5fb1117c584c1e142ba880800dc0d2483d/pkg/api/bzz.go#L153)):*

Is `logger.Error` necessary here? It looks like this is normal flow for handling requests which has early return for this case when filename starts with `/`. In other words error has not occurred in the system so `logger.Error` seams to be escalating severity.
</blockquote></details>

i think it is kind of a convention where every response is preceded by a debug log and an error log so i was following that.

Issue #3175, opened 2022-08-12
Benchmark suite for the data abstraction layer interfaces


aloknerurkar:
Write a benchmark suite using the Store interface. This can be used by people writing the implementations to quickly check benchmarks.

For examples:
https://github.com/dgraph-io/badger-bench

Some of the considerations:
- Sequential and Random Gets need to be benchmarked separately.
- Iterations and Count operations need to be benchmarked correctly as these include user passed functions.

End product should be similar to the testsuite implementation which can be used by stores in their unit test files.

PR #3237, opened 2022-08-31
feat: localstore benchmarks

notanatol:
### Checklist

- [X] I have added tests to cover my changes.

### Description
Ported an existing benchmarking solution for leveldb, adapted it to target in memory implementation.

Added the results for comparison

Closes: #3175

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3237)


pkg/storagev2/testsuite/benchmark.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+func (w *singularDBWriter) Put(key, value []byte) {
+	item := &obj1{
+		Id:  string(key),
+		Buf: value,
+	}
+	err := w.db.Put(item)
+	if err != nil {
+		panic(err)
```
aloknerurkar, line 256: Instead of panic maybe we can use the t.Fatal by have a reference to the test in the writer. I think failing a test that way is better.

aloknerurkar (commented):
Great! These look better than I imagined!

pkg/storagev2/testsuite/benchmark.go
notanatol: fixed

aloknerurkar (approved):

pkg/storagev2/inmemstore/inmemstore_test.go
```diff
@@ -14,3 +14,49 @@ import (
...
+BenchmarkInMem/ReadRandomMissing
+BenchmarkInMem/ReadRandomMissing-8                 	 1000000	      1118 ns/op	      96 B/op	       3 allocs/op
+BenchmarkInMem/ReadReverse
+BenchmarkInMem/ReadReverse-8                       	 1587060	       915.5 ns/op	     152 B/op	       5 allocs/op
+BenchmarkInMem/ReadRedHot
+BenchmarkInMem/ReadRedHot-8                        	 1934049	       877.2 ns/op	     152 B/op	       5 allocs/op
+BenchmarkInMem/IterateSequential
+BenchmarkInMem/IterateSequential-8                 	  612537	      1683 ns/op	     573 B/op	      12 allocs/op
+BenchmarkInMem/IterateReverse
+BenchmarkInMem/IterateReverse-8                    	       1	1311065959 ns/op	695364040 B/op	 7757261 allocs/op
```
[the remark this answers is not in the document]
notanatol: yes but now we have some numbers to improve against, so no big deal

pkg/storagev2/storagetest/benchmark.go
```diff
@@ -0,0 +1,297 @@
...
+	valueSize        = flag.Int("value_size", 100, "Size of each value")
+	compressionRatio = flag.Float64("compression_ratio", 0.5, "")
+)
+
+const (
+	hitKeyFormat     = "%016d+"
+	missingKeyFormat = "%016d-"
+)
+
+func init() {
```
mrekucci, line 29: I am confused about this `init` function, why is it needed? The `hitKeyFormat` and `missingKeyFormat` are constants and so is the `math.MaxInt32`, doesn't it lead to the same result (17) every time the `init` is called?

pkg/storagev2/leveldbstore/store_test.go
```diff
@@ -18,3 +20,53 @@ func TestStoreTestSuite(t *testing.T) {
...
+func BenchmarkLevelDB(b *testing.B) {
+	st, err := ldb.New("", new(opt.Options))
+	if err != nil {
+		b.Fatal(err)
+	}
+	b.Cleanup(func() { _ = st.Close() })
+	storagetest.RunBenchmarkTests(b, st)
+}
+
+/* Benchmark results
```
mrekucci, line 33: I think this result should be part of the PR description. Also, the results are relative, so it makes sense to compare them on the same computer.

pkg/storagev2/storagetest/benchmark.go
```diff
@@ -0,0 +1,297 @@
+package storagetest
```
mrekucci, line 1: Missing header.

pkg/storagev2/inmemstore/inmemstore_test.go
```diff
@@ -14,3 +14,49 @@ import (
 func TestStore(t *testing.T) {
 	storagetest.TestStore(t, inmem.New())
 }
+
+func BenchmarkInMem(b *testing.B) {
+	st := inmem.New()
+	storagetest.RunBenchmarkTests(b, st)
+}
+
+/* Benchmark results
```
mrekucci, line 23: The same as in the leveldbstore bench results.

pkg/storagev2/inmemstore/inmemstore_test.go
```diff
@@ -14,3 +14,49 @@ import (
 func TestStore(t *testing.T) {
 	storagetest.TestStore(t, inmem.New())
 }
+
+func BenchmarkInMem(b *testing.B) {
```
mrekucci, line 18: I would recommend naming it `BenchmarkStore` to be consistent with `TestStore`.

mrekucci (commented):
It would be good if you could document the benchmark file so the user knows if all the values are pre-generated, how the different values of `valueSize` and `compressionRatio` will affect the benchmark, etc...

pkg/storagev2/leveldbstore/store_test.go
notanatol: let's start a conversation on this topic, there's Alok's first comment to this PR is exactly about that, please reply  there and we'll find a compromise.

PR #3238, opened 2022-08-31
fix: handle mantaray malformed manifest contents

vladopajic:
### Checklist

### Description
Adding error handling when invalid data is provided to `Node.UnmarshalBinary` function.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3238)


notanatol (approved):

pkg/manifest/mantaray/marshal.go
```diff
@@ -83,6 +83,8 @@ var (
 	ErrInvalidInput = errors.New("input invalid")
 	// ErrInvalidVersionHash signals unknown version of hash.
 	ErrInvalidVersionHash = errors.New("invalid version hash")
+	// ErrInvalidOutput signals when malformed manifest contnet is supplied to Unmarshal function
+	ErrInvalidOutput = errors.New("malformed manifest contents")
```
aloknerurkar, line 87: I would rather call it `ErrInvalidManifest`. Output doesnt make much sense.

aloknerurkar (approved):

pkg/manifest/mantaray/marshal_test.go
```diff
@@ -201,3 +202,54 @@ func TestMarshal(t *testing.T) {
 	// 	}
 	// }
 }
+
+func Test_UnmarshalBinary(t *testing.T) {
+	decode := func(s string) []byte {
```
mrekucci, line 207: Adding `t.Helper()` to the `decode` closure will help to accurately identify the problematic string.

pkg/manifest/mantaray/marshal_test.go
```diff
@@ -201,3 +202,54 @@ func TestMarshal(t *testing.T) {
...
+		{ // fork with metadata, where metadata size value exceeds actual metadata size (size is 96, want 93)
+			data: decode("00000000000000000000000000000000000000000000000000000000000000005768b3b6a7db56d21d1abff40d41cebfc83448fed8d7e9b06ec0d3b073f28f200000000000000000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000016012f0000000000000000000000000000000000000000000000000000000000e87f95c3d081c4fede769b6c69e27b435e525cbd25c6715c607e7c531e32963900607b22776562736974652d696e6465782d646f63756d656e74223a2233356561656538316262363338303436393965633637316265323736326465626665346662643330636461646139303232393239646131613965366134366436227d0a"),
+			err:  ErrInvalidOutput,
+		},
+	}
+
+	for _, tc := range tests {
+		n := New()
+		err := n.UnmarshalBinary(tc.data)
+
```
mrekucci, line 252: Since the test cases aren't named, executing every test case as a sub-test (`t.Run("", func(t *testing.T) {...`) will help to accurately identify the problematic test case.

pkg/manifest/mantaray/marshal_test.go
```diff
@@ -201,3 +202,58 @@ func TestMarshal(t *testing.T) {
...
+		data, err := hex.DecodeString(s)
+		if err != nil {
+			t.Fatalf("expected no error, got %v", err)
+		}
+		return data
+	}
+
+	tests := []struct {
+		data []byte // node binary data
+		err  error  // tells if UnmarshalBinary should return error
```
mrekucci, line 217: Naming it `wantErr` would be more indicative.

pkg/manifest/mantaray/marshal.go
mrekucci: Agree, also the error string should match the error name.

pkg/manifest/mantaray/marshal_test.go
```diff
@@ -201,3 +202,58 @@ func TestMarshal(t *testing.T) {
...
+		},
+		{ // fork with metadata, where metadata size value exceeds actual metadata size (size is 96, want 93)
+			data: decode("00000000000000000000000000000000000000000000000000000000000000005768b3b6a7db56d21d1abff40d41cebfc83448fed8d7e9b06ec0d3b073f28f200000000000000000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000016012f0000000000000000000000000000000000000000000000000000000000e87f95c3d081c4fede769b6c69e27b435e525cbd25c6715c607e7c531e32963900607b22776562736974652d696e6465782d646f63756d656e74223a2233356561656538316262363338303436393965633637316265323736326465626665346662643330636461646139303232393239646131613965366134366436227d0a"),
+			err:  ErrInvalidOutput,
+		},
+	}
+
+	for _, tc := range tests {
+		n := New()
+		err := n.UnmarshalBinary(tc.data)
```
mrekucci, line 251: Naming it `haveErr` or `gotErr` would be more indicative.

mfw78:
From my understanding of the Mantaray spec, if the test case were correct, the metadata bytes size should be 94, as the metadata JSON is padding with new line characters (`0a`). Interestingly, the `mantaray-js` implementation doesn't do the newline padding (which is however present in the `bee` implementation).

pkg/manifest/mantaray/marshal.go
vladopajic: @notanatol, do you agree to use @mrekucci and @aloknerurkar suggestion for this error variable?

pkg/manifest/mantaray/marshal_test.go
vladopajic: I my option `err` is equally indicative here and both suggestions seem verbose. Variable name `err` is usual convention used to name error result from last function calls. 

However I have updated `haveErr`.

vladopajic:
@mfw78 Thanks for pointing out on this. We will let you know what is correct approach.

pkg/manifest/mantaray/marshal.go
notanatol: sure, sounds good

pkg/manifest/mantaray/marshal_test.go
```diff
@@ -201,3 +203,61 @@ func TestMarshal(t *testing.T) {
...
+			wantErr: ErrInvalidManifest,
+		},
+		{ // fork with metadata, where metadata size value exceeds actual metadata size (size is 96, want 93)
+			data:    decode("00000000000000000000000000000000000000000000000000000000000000005768b3b6a7db56d21d1abff40d41cebfc83448fed8d7e9b06ec0d3b073f28f200000000000000000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000016012f0000000000000000000000000000000000000000000000000000000000e87f95c3d081c4fede769b6c69e27b435e525cbd25c6715c607e7c531e32963900607b22776562736974652d696e6465782d646f63756d656e74223a2233356561656538316262363338303436393965633637316265323736326465626665346662643330636461646139303232393239646131613965366134366436227d0a"),
+			wantErr: ErrInvalidManifest,
+		},
+	}
+
+	for i, tc := range tests {
+		t.Run(fmt.Sprintf("case #%d", i+1), func(t *testing.T) {
```
mrekucci, line 252: You don't need this `fmt.Sprintf("case #%d", i+1)`, if `""` is given, the numbers will be inserted automatically by the Golang testing library.

mrekucci (approved):

agazso:
> From my understanding of the Mantaray spec, if the test case were correct, the metadata bytes size should be 94, as the metadata JSON is padded with a new line character (`0a`). Interestingly, the `mantaray-js` implementation doesn't do the newline padding (which is however present in the `bee` implementation).

The mantaray-js has an issue in that it is not compatible with Bee. There is already a fix for that: https://github.com/ethersphere/mantaray-js/pull/37

Issue #2956, opened 2022-05-15
Expose "health" and "readiness" values on metrics


tmm360:
#### Summary
Expose current value of health and readiness on `/metrics` endpoint.

#### Motivation
This is useful for implement automatic alarms on node availability using tools based on metrics like Prometheus.

#### Implementation
Needs only to include these values on metrics.

#### Drawbacks
Nothing.

notanatol:
This can be done as part of the current effort to [refactor the APIs](https://github.com/ethersphere/bee/pull/2846)

Issue #2793, opened 2022-02-06
Debug API "/health" and "/readiness" need more distinction in documentation


tmm360:
I don't get differences between these two apis in documentation, and they look to be the same.

tmm360:
.

PR #3248, opened 2022-09-05
feat(api): expose readiness and health endpoints

vladopajic:
### Checklist

### Description
Closes: #2956,  #2793

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3248)

### DIFF
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -881,7 +881,7 @@ paths:
     get:
       summary: Get node overall health Status
       description: |
-        Health Status will indicate node healthiness (healthiness is k8s equivalent for liveness).
+        Health Status will indicate node healthiness.
 
         If node is unhealthy please check node logs for errors.
       tags:
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -881,8 +881,8 @@ paths:
     get:
       summary: Get node overall health Status
       description: |
-        Health Status will indicate node healthiness (healthiness is k8s equivalent for liveness). 
-        
+        Health Status will indicate node healthiness (healthiness is k8s equivalent for liveness).
+
         If node is unhealthy please check node logs for errors.
       tags:
         - Status
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -880,7 +880,21 @@ paths:
   "/health":
     get:
       summary: Get health of node
-      description: This endpoint is available on the main API only if the node is spawned with the `--restricted` flag.
+      tags:
+        - Status
+      responses:
+        "200":
+          description: Health State of node
+          content:
+            application/json:
+              schema:
+                $ref: "SwarmCommon.yaml#/components/schemas/Status"
+        default:
+          description: Default response
+
+  "/readiness":
+    get:
+      summary: Get readiness state of node
       tags:
         - Status
       responses:
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -894,16 +894,16 @@ paths:
 
   "/readiness":
     get:
-      summary: Get readiness state of node
+      summary: Readiness endpoint indicates if node is ready to start accepting traffic
       tags:
         - Status
       responses:
         "200":
-          description: Health State of node
-          content:
-            application/json:
-              schema:
-                $ref: "SwarmCommon.yaml#/components/schemas/Status"
+          description: Indicates that node is ready
+          $ref: "SwarmCommon.yaml#/components/responses/200"
+        "400":
+          description: Indicates that node is not ready
+          $ref: "SwarmCommon.yaml#/components/responses/400"
         default:
           description: Default response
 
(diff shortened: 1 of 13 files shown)


pkg/api/probe.go
```diff
@@ -0,0 +1,44 @@
+package api
```
mrekucci, line 1: Missing header.

So, I'd suggest making the zero value of `*Probe` usable so you don't need to check if the probe is `nil`. Also, `probeStatusPending` state is unnecessary. I'd recommend doing the following:
```Golang
// ProbeStatus is the status of a probe.
// ProbeStatus is treated as a sync/atomic int32.
type ProbeStatus int32

// get returns the value of the ProbeStatus.
func (ps *ProbeStatus) get() ProbeStatus {
	return ProbeStatus(atomic.LoadInt32((*int32)(ps)))
}

// set updates the value of the ProbeStatus.
func (ps *ProbeStatus) set(v ProbeStatus) {
	atomic.StoreInt32((*int32)(ps), int32(v))
}

// String implements the fmt.Stringer interface.
func (ps ProbeStatus) String() string {
	switch ps.get() {
	case ProbeStatusOK:
		return "ok"
	case ProbeStatusNOK:
		return "nok"
	}
	return "unknown"
}

const (
	// ProbeStatusOK indicates positive ProbeStatus status.
	ProbeStatusOK ProbeStatus = 1

	// ProbeStatusNOK indicates negative ProbeStatus status.
	ProbeStatusNOK ProbeStatus = 0
)

// Probe structure holds flags which indicate node healthiness (sometimes refert also as liveness) and readiness.
type Probe struct {
	// Healthy probe indicates if node, due to any reason, needs to restarted.
	healthy ProbeStatus
	// Ready probe indicates that node is ready to start accepting traffic.
	ready ProbeStatus
}

// Healthy returns the value of the healthy status.
func (p *Probe) Healthy() ProbeStatus {
	if p == nil {
		return ProbeStatusNOK
	}
	return p.healthy.get()

}

// SetHealthy updates the value of the healthy status.
func (p *Probe) SetHealthy(ps ProbeStatus) {
	p.healthy.set(ps)
}

// Ready returns the value of the ready status.
func (p *Probe) Ready() ProbeStatus {
	if p == nil {
		return ProbeStatusNOK
	}
	return p.ready.get()
}

// SetReady updates the value of the ready status.
func (p *Probe) SetReady(ps ProbeStatus) {
	p.ready.set(ps)
}
```

pkg/api/probe.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (p *Probe) Ready() bool {
+	return atomic.LoadInt32(&p.ready) == ProbeStatusOK
+}
+
+func (p *Probe) SetReady(s int32) {
+	atomic.StoreInt32(&p.ready, s)
+}
+
+func (s *Service) SetProbe(probe *Probe) {
```
mrekucci, line 42: I'd recommend moving it to the `api.go` file where all the methods of `*Service` are defined.

openapi/Swarm.yaml
```diff
@@ -853,20 +853,38 @@ paths:
 
   "/health":
     get:
-      summary: Get health of node
-      description: This endpoint is available on the main API only if the node is spawned with the `--restricted` flag.
+      summary: Get node overall health Status
+      description: |
+        Health Status will indicate node healthiness (healthiness is k8s equivalent for liveness).
```
mrekucci, line 858: I find the "(healthiness is k8s equivalent for liveness)" part to be irrelevant since it assumes that the consumer of the API is familiar with the k8s, which might not be the case.

openapi/SwarmDebug.yaml
```diff
@@ -395,21 +380,40 @@ paths:
         default:
           description: Default response
 
-  "/readiness":
+  "/health":
     get:
-      summary: Get readiness state of node
+      summary: Get node overall health Status
+      description: |
+        Health Status will indicate node healthiness (healthiness is k8s equivalent for liveness).
```
mrekucci, line 387: The same as above.

openapi/Swarm.yaml
vladopajic: I assume that main purpose behind these two endpoints is for orchestration solutions (like k8s). That's why I wanted to be more descriptive with this description.

However, that part from the docs has been removed in recent change.

pkg/api/probe.go
```diff
@@ -1,44 +1,80 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
```
mrekucci, line 1: The year should be 2022.

pkg/api/probe.go
```diff
@@ -1,44 +1,80 @@
...
-	healthy int32
+	// Healthy probe indicates if node, due to any reason, needs to restarted.
+	healthy ProbeStatus
 	// Ready probe indicates that node is ready to start accepting traffic.
-	ready int32
+	ready ProbeStatus
 }
 
+// NewProbe returns new Probe.
 func NewProbe() *Probe {
```
mrekucci, line 51: The constructor can be dropped and replaced with `new(Probe)`, since it doesn't add much value.

mrekucci (approved):

notanatol (approved):

PR #3250, opened 2022-09-05
feat: add upload store

mrekucci:
### Checklist

### Description
Adds push index.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3250)

### DIFF
--- a/pkg/storagev2/internal/internal.go
+++ b/pkg/storagev2/internal/internal.go
@@ -1,0 +1,18 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package internal
+
+import (
+	"context"
+
+	"github.com/ethersphere/bee/pkg/storagev2"
+)
+
+// Storage groups the storage.Store and storage.ChunkStore interfaces with context..
+type Storage interface {
+	Ctx() context.Context
+	Storage() storage.Store
+	ChunkStore() storage.ChunkStore
+}
--- a/pkg/storagev2/internal/internal.go
+++ b/pkg/storagev2/internal/internal.go
@@ -13,6 +13,6 @@ import (
 // Storage groups the storage.Store and storage.ChunkStore interfaces with context..
 type Storage interface {
 	Ctx() context.Context
-	Storage() storage.Store
+	Store() storage.Store
 	ChunkStore() storage.ChunkStore
 }
--- a/pkg/storagev2/internal/upload/uploadstore.go
+++ b/pkg/storagev2/internal/upload/uploadstore.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package internal
+package upload
 
 import (
 	"context"
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/ethersphere/bee/pkg/storagev2"
+	"github.com/ethersphere/bee/pkg/storagev2/internal"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
@@ -132,8 +133,8 @@ func (i *pushItem) Unmarshal(bytes []byte) error {
 	return nil
 }
 
-// Upload returns a storage.Putter which will store the given chunk.
-func Upload(s storage.Storage, tag uint64) (storage.Putter, error) {
+// Chunk returns a storage.Putter which will store the given chunk.
+func Chunk(s internal.Storage, tag uint64) (storage.Putter, error) {
 	return storage.PutterFunc(func(ctx context.Context, chunk swarm.Chunk) (bool, error) {
 		ati := &tagIDAddressItem{
 			Address: chunk.Address(),
--- a/pkg/storagev2/internal/upload/uploadstore.go
+++ b/pkg/storagev2/internal/upload/uploadstore.go
@@ -9,6 +9,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"path"
 	"time"
 
 	"github.com/ethersphere/bee/pkg/storagev2"
@@ -42,12 +43,12 @@ type tagIDAddressItem struct {
 
 // ID implements the storage.Item interface.
 func (i tagIDAddressItem) ID() string {
-	return fmt.Sprintf("%d_%s", i.TagID, i.Address.ByteString())
+	return i.Address.ByteString()
 }
 
 // Namespace implements the storage.Item interface.
 func (i tagIDAddressItem) Namespace() string {
-	return "TagIDAddressItem"
+	return fmt.Sprintf("TagIDAddressItem/%d", i.TagID)
 }
 
 // Marshal implements the storage.Item interface.
@@ -75,6 +76,11 @@ func (i *tagIDAddressItem) Unmarshal(bytes []byte) error {
 	return nil
 }
 
+// String implements the fmt.Stringer interface.
+func (i tagIDAddressItem) String() string {
+	return path.Join(i.Namespace(), i.ID())
+}
+
 var (
 	// errPushItemMarshalAddressIsZero is returned when trying
 	// to marshal a pushItem with an address that is zero.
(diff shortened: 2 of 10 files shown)


pkg/storagev2/internal/uploadstore.go
```diff
@@ -0,0 +1,162 @@
...
+	ni := new(pushItem)
+	ni.Timestamp = binary.LittleEndian.Uint64(bytes)
+	ni.Address = swarm.NewAddress(append(make([]byte, 0, swarm.HashSize), bytes[8:8+swarm.HashSize]...))
+	ni.TagID = binary.LittleEndian.Uint64(bytes[8+swarm.HashSize:])
+	*i = *ni
+	return nil
+}
+
+// Upload returns a storage.Putter which will store the given chunk.
+func Upload(s storage.Storage, tag uint64) (storage.Putter, error) {
```
aloknerurkar, line 136: So I think we should add tests for this function as well. We should put chunks using this and makes sure all the indexes etc are correct. I have done something similar in the pinstore PR.

pkg/storagev2/storage.go
```diff
@@ -147,3 +148,10 @@ type Store interface {
 	// Delete removes the given Item form the store.
 	Delete(Item) error
 }
+
+// Storage TODO: remove when merged to master.
+type Storage interface {
```
aloknerurkar, line 153: So I think it would be best if we define this in the internal packages themselves. This need not be a core interface. This allows better flexibility I think.

pkg/storagev2/internal/uploadstore.go
```diff
@@ -0,0 +1,162 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package internal
```
aloknerurkar, line 5: So I am not sure about these. For the pinning store I have created a separate package in internal. Mostly because I think these things are completely independent. So it would be better to have it in their own package with their own test logic etc.

We could have one internal package to house the common interfaces required for these components. Like for eg the `Storage` interface that we have defined could be here.

pkg/storagev2/internal/uploadstore.go
```diff
@@ -0,0 +1,162 @@
...
+	"time"
+
+	"github.com/ethersphere/bee/pkg/storagev2"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+var (
+	// errTagIDAddressItemMarshalAddressIsZero is returned when trying
+	// to marshal a tagIDAddressItem with an address that is zero.
+	errTagIDAddressItemMarshalAddressIsZero = errors.New("marshal tagIDAddressItem: address is zero")
```
aloknerurkar, line 21: So I think these errors need to be exported right?

The localstore might wrap the error differently based on these. But for that we would have to know about these I think.

mrekucci: Maybe, but I wouldn't do it now but postpone it when needed.

pkg/storagev2/internal/uploadstore.go
mrekucci: Done.

pkg/storagev2/internal/upload/uploadstore.go
```diff
@@ -0,0 +1,166 @@
...
+
+// tagIDAddressItem is an store.Item that stores addresses of already seen chunks.
+type tagIDAddressItem struct {
+	TagID   uint64
+	Address swarm.Address
+}
+
+// ID implements the storage.Item interface.
+func (i tagIDAddressItem) ID() string {
+	return fmt.Sprintf("%d_%s", i.TagID, i.Address.ByteString())
```
aloknerurkar, line 45: The TagID can be moved to the Namespace function. This way, we don't need to have this in the value. Check the chunkStampItem in the chunkstore PR. As the tag is the same for all the chunks in this tag, it can be part of the prefix as we would know this before we query the item.

This way we dont need to store the duplicate tagIDs in the store for each chunk. Will save some space.

pkg/storagev2/internal/upload/uploadstore_test.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+		if !has {
+			t.Fatalf("Has(...): item not found")
+		}
+
+		have, err := ts.ChunkStore().Get(context.TODO(), chunk.Address())
+		if err != nil {
+			t.Fatalf("Get(...): unexpected error: %v", err)
+		}
+		if diff := cmp.Diff(chunk, have); diff != "" {
```
aloknerurkar, line 263: We have a chunk.Equal function which would be better here.

mrekucci: Done.

pkg/storagev2/internal/upload/uploadstore_test.go
```diff
@@ -0,0 +1,267 @@
...
+			t.Errorf("ChunkStore().Close(): unexpected error: %v", err)
+		}
+	})
+
+	tag := uint64(1)
+	putter, err := upload.ChunkPutter(ts, tag)
+	if err != nil {
+		t.Fatalf("upload.ChunkPutter(...): unexpected error: %v", err)
+	}
+	chunk := swarm.NewChunk(swarm.NewAddress([]byte("1234")), []byte("data"))
```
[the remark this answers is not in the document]
mrekucci: Done.

pkg/storagev2/internal/upload/uploadstore.go
```diff
@@ -0,0 +1,166 @@
...
+// ChunkPutter returns a storage.Putter which will store the given chunk.
+func ChunkPutter(s internal.Storage, tag uint64) (storage.Putter, error) {
+	return storage.PutterFunc(func(ctx context.Context, chunk swarm.Chunk) (bool, error) {
+		tai := &tagIDAddressItem{
+			Address: chunk.Address(),
+			TagID:   tag,
+		}
+		switch exists, err := s.Storage().Has(tai); {
+		case err != nil:
+			return false, err
```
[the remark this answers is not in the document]
mrekucci: Done.

aloknerurkar (approved):
Nice!

pkg/storagev2/internal/internal.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"context"
+
+	"github.com/ethersphere/bee/pkg/storagev2"
+)
+
+// Storage groups the storage.Store and storage.ChunkStore interfaces with context..
+type Storage interface {
+	Ctx() context.Context
+	Storage() storage.Store
```
notanatol, line 16: shouldn't this be called `Store() storage.Store`?

pkg/storagev2/internal/upload/uploadstore_test.go
```diff
@@ -0,0 +1,272 @@
...
+			Item: &storagetest.ItemStub{
+				MarshalBuf:   []byte{0xFF},
+				UnmarshalBuf: []byte{0xFF},
+			},
+			Factory:      func() storage.Item { return new(upload.PushItem) },
+			UnmarshalErr: upload.ErrPushItemUnmarshalInvalidSize,
+		},
+	}}
+
+	for _, tc := range tests {
```
notanatol, line 176:
```
for tc := range tt {
  tc := tc // must not forget this
  t.Run(tc.name, func(t *testing.T) {
    t.Parallel()
```

pkg/storagev2/internal/internal.go
mrekucci: Done.

pkg/storagev2/internal/upload/uploadstore_test.go
mrekucci: Nice catch! Done.

pkg/storagev2/internal/upload/uploadstore.go
```diff
@@ -0,0 +1,259 @@
...
+// that restricts its operation to the specific tagID.
+type getterDeleter struct {
+	storage internal.Storage
+	tagID   uint64
+}
+
+// Close implements the io.Closer interface.
+func (gd *getterDeleter) Close() error {
+	return multierror.Append(
+		gd.storage.Store().Close(),
```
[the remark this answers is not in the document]
mrekucci: Done.

pkg/storagev2/internal/upload/uploadstore.go
```diff
@@ -0,0 +1,259 @@
...
+
+		pi := &pushItem{
+			Timestamp: now().Unix(),
+			Address:   chunk.Address(),
+			TagID:     tagID,
+		}
+		if err := s.Store().Put(pi); err != nil {
+			return false, fmt.Errorf("store put item %q call failed: %w", pi, err)
+		}
+		exists, err := s.ChunkStore().Put(ctx, chunk)
```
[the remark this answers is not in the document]
mrekucci: Ok, I'll leave it for now and we'll revisit this later.

notanatol (approved):

PR #3253, opened 2022-09-07
fix: interrupt mining

### LINKED ISSUE #3251: Can not interrupt while finding the new overlay
The process does not response to `syscall.SIG*` signals during the process of finding the new overlay.

notanatol:
### Checklist

### Description
At the moment the process of finding the new overlay is not interruptible, the user can only stop it by killing the process.

This PR makes the mining loop to be `os.Signal` aware and cleanly exit when the signal is received.

Closes #3251

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3253)

### DIFF
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -169,7 +169,7 @@ func (c *command) initStartCmd() (err error) {
 
 			interruptChannel := make(chan struct{})
 
-			b, err := node.NewBee(interruptChannel, c.config.GetString(optionNameP2PAddr), signerConfig.publicKey, signerConfig.signer, networkID, logger, signerConfig.libp2pPrivateKey, signerConfig.pssPrivateKey, &node.Options{
+			b, err := node.NewBee(interruptChannel, sysInterruptChannel, c.config.GetString(optionNameP2PAddr), signerConfig.publicKey, signerConfig.signer, networkID, logger, signerConfig.libp2pPrivateKey, signerConfig.pssPrivateKey, &node.Options{
 				DataDir:                    c.config.GetString(optionNameDataDir),
 				CacheCapacity:              c.config.GetUint64(optionNameCacheCapacity),
 				DBOpenFilesLimit:           c.config.GetUint64(optionNameDBOpenFilesLimit),
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -20,6 +20,7 @@ import (
 	"math/big"
 	"net"
 	"net/http"
+	"os"
 	"path/filepath"
 	"runtime"
 	"sync"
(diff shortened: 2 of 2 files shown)


cmd/bee/cmd/start.go
```diff
@@ -169,7 +169,7 @@ func (c *command) initStartCmd() (err error) {
 
 			interruptChannel := make(chan struct{})
 
-			b, err := node.NewBee(interruptChannel, c.config.GetString(optionNameP2PAddr), signerConfig.publicKey, signerConfig.signer, networkID, logger, signerConfig.libp2pPrivateKey, signerConfig.pssPrivateKey, &node.Options{
+			b, err := node.NewBee(interruptChannel, sysInterruptChannel, c.config.GetString(optionNameP2PAddr), signerConfig.publicKey, signerConfig.signer, networkID, logger, signerConfig.libp2pPrivateKey, signerConfig.pssPrivateKey, &node.Options{
```
aloknerurkar, line 172: So I remember in [this PR](https://github.com/ethersphere/bee/pull/3127), we decided not to pass the sysInterruptChannel down stream. Although in this case it seems to be okay as we are only using it during the NewBee function. This can be easily misused.

One way would be to wrap the checking of sysInterruptChan in a closure which can be used by NewBee. If we plan to remove this in future, we should create an issue in the backlog for this.

aloknerurkar (approved):

cmd/bee/cmd/start.go
notanatol: you are right, this is a fix that would live only until v1.9, added the issue here: https://github.com/ethersphere/bee/issues/3254

pkg/node/node.go
```diff
@@ -187,7 +188,7 @@ const (
 	mainnetNetworkID              = uint64(1)                 //
 )
 
-func NewBee(interrupt chan struct{}, addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, networkID uint64, logger log.Logger, libp2pPrivateKey, pssPrivateKey *ecdsa.PrivateKey, o *Options) (b *Bee, err error) {
+func NewBee(interrupt chan struct{}, sysInterrupt chan os.Signal, addr string, publicKey *ecdsa.PublicKey, signer crypto.Signer, networkID uint64, logger log.Logger, libp2pPrivateKey, pssPrivateKey *ecdsa.PrivateKey, o *Options) (b *Bee, err error) {
```
vladopajic, line 191: `NewBee` has two interrupt signal channels `interrupt` and `sysInterrupt`. Would it be possible to have only one interrupt signal channel?

mrekucci (approved):
As a quick fix it looks good. Please link this PR to the issue you've created #3254.

PR #3266, opened 2022-09-09
fix: remove redirection and throw error

### LINKED ISSUE #3199: Permanent redirect when data is not found
### Context
Bee 1.7.0, macOS

### Summary
I was trying to open the swarm website on my Bee node. It was not found (that's a different issue) however I ended up in a redirect loop.

### Expected behavior
I expect that when I open data that is not found then I don't get a redirect loop, but rather an error message.

### Actual behavior
```
% curl -v http://localhost:1633/bzz/swarm.eth/
*   Trying 127.0.0.1:1633...
* Connected to localhost (127.0.0.1) port 1633 (#0)
> GET /bzz/swarm.eth/ HTTP/1.1
> Host: localhost:1633
> User-Agent: curl/7.83.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 308 Permanent Redirect
< Content-Type: text/html; charset=utf-8
< Location: /bzz/swarm.eth/
< Date: Tue, 16 Aug 2022 15:40:25 GMT
< Content-Length: 51
< 
<a href="/bzz/swarm.eth/">Permanent Redirect</a>.
```
Screenshot 2022-08-16 at 17 45 12

Most likely this problem is coming from here:
https://github.com/ethersphere/bee/blob/5c02037e045a9dc30d7d5d3ab7ddc95f2bc43248/pkg/api/bzz.go#L364

swarmHaseeb:
### Checklist

### Description

Remove redirection and throw error when resource is not found

Api refactoring
### Related Issue (Optional)

fixes #3199

### DIFF
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -360,9 +360,10 @@ FETCH:
 					u.Path += "/"
 					redirectURL := u.String()
 
-					logger.Debug("bzz download: redirecting failed", "url", redirectURL, "error", err)
+					s.logger.Debug("bzz download: redirecting failed: not found", "url", redirectURL, "error", err)
+					s.logger.Error(nil, "bzz download: redirecting failed: not found")
 
-					http.Redirect(w, r, redirectURL, http.StatusPermanentRedirect)
+					jsonhttp.NotFound(w, "bzz download: redirecting failed: not found")
 					return
 				}
 			}


pkg/api/bzz.go
```diff
@@ -360,9 +360,10 @@ FETCH:
 					u.Path += "/"
 					redirectURL := u.String()
 
-					logger.Debug("bzz download: redirecting failed", "url", redirectURL, "error", err)
+					s.logger.Debug("bzz download: redirecting failed: not found", "url", redirectURL, "error", err)
+					s.logger.Error(nil, "bzz download: redirecting failed: not found")
```
vladopajic, line 364: to me this debug message looks confusing because it indicates that redirection has failed, but there is no redirect operation.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3266)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mrekucci, @umerm-work, and @vladopajic)

___
*[`pkg/api/bzz.go` line 364 at r1](https://reviewable.io/reviews/ethersphere/bee/3266#-NBWbqtR7WMGH4QBgMJx-r1-364:-NBknyBl9dlUMzcjYnWJ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/94e242c85c0b00e2fc235eb83f4c598dafad8499/pkg/api/bzz.go#L364)):*

to me this debug message looks confusing because it indicates that redirection has failed, but there is no redirect operation.
</blockquote></details>

Done.

vladopajic (commented):
do we really need this `u` variable? It looks like we could just log `url` parameter as `r.URL`, right?

pkg/api/bzz.go
```diff
@@ -360,9 +360,10 @@ FETCH:
 					u.Path += "/"
 					redirectURL := u.String()
```
vladopajic, line 361: do we really need this `u` variable? It looks like we could just log `url` parameter as `r.URL`, right?

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3266)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @mrekucci, @umerm-work, and @vladopajic)

___
*[`pkg/api/bzz.go` line 361 at r2](https://reviewable.io/reviews/ethersphere/bee/3266#-NBWbqtR7WMGH4QBgMJx-r2-361:-NBlctrz4pCV8k-6E0pk:b43vdok) ([raw file](https://github.com/ethersphere/bee/blob/40777107e84284f81218a546ce029ada603bb401/pkg/api/bzz.go#L361)):*

do we really need this `u` variable? It looks like we could just log `url` parameter as `r.URL`, right?
</blockquote></details>

i think the reason why we're doing using u var is bec we want to add a slash to the path, without variable we would be adding it directly in r.URL and my understanding here is that we may need it somewhere without a slash which is why i left it as is. But then as you said in other message, if there is no redirection then why log a redirect url in first place which is why i removed it.



_Code snippet:_
```Go
u.Path += "/"
redirectURL := u.String()
```

vladopajic (commented):
LGMT

vladopajic (approved):

pkg/api/bzz.go
```diff
@@ -355,14 +355,10 @@ FETCH:
 				dirPath := pathVar + "/"
 				exists, err := m.HasPrefix(ctx, dirPath)
 				if err == nil && exists {
-					// redirect to directory
```
aloknerurkar, line 358: This is a feature being disabled. The problem mentioned in the issue is the redirect loop and not the redirect. First, we need to understand how this feature is being used. I think the way this is used is that it would redirect you to the folder in case you entered an incorrect path. This way it would hit the `/bzz` with empty path and this would enter the previous if loop on L330.

pkg/api/dirs_test.go
```diff
@@ -305,16 +305,16 @@ func TestDirs(t *testing.T) {
 				)
 			}
 
-			validateIsPermanentRedirect := func(t *testing.T, fromPath, toPath string) {
```
aloknerurkar, line 308: So before disabling features, you should understand how and where they are used. The presence of this test should be clear that this is a feature.

aloknerurkar (changes requested):

agazso:
Looks like you just removed the redirect. Although that might fix the redirect loop, but it may break other use-cases where the redirect were put in deliberately.

PR #3279, opened 2022-09-14
feat: localstore batching

### LINKED ISSUE #3278: Batch interface and implementation
In order to implement the transaction functionality over a generic Store that we have, we need a batching functionality. This will be part of the core storage interface and can be implemented by different store implementations. The Batch will be able to stage multiple write (Put/Delete) operations before committing them atomically. Different implementations can achieve this in different ways.

## Tasks
1. Add new core storage interface for batch
```go
type Batch interface {
        // We will only need to batch write operations
        Put(Item) error
        Delete(Key) error

        Commit() error 
}

type BatchedStore interface {
        Batch(context.Context) (Batch, error)
}
```

2. Implement the BatchedStore interface for inmemstore implementation. This can be done using some inmem datastructure to maintain the ops.

3. Implement the BatchedStore interface for leveldbstore implementation. This will use the leveldb.Batch object.

notanatol:
### Checklist

- [X] I have filled out the description and linked the related issues.

### Description
Closes #3278

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3279)

### DIFF
--- a/pkg/storagev2/batch.go
+++ b/pkg/storagev2/batch.go
@@ -1,0 +1,19 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package storage
+
+import "context"
+
+type Batch interface {
+	// We will only need to batch write operations
+	Put(Item) error
+	Delete(Key) error
+
+	Commit() error
+}
+
+type BatchedStore interface {
+	Batch(context.Context) (Batch, error)
+}
--- a/pkg/storagev2/batch.go
+++ b/pkg/storagev2/batch.go
@@ -15,5 +15,6 @@ type Batch interface {
 }
 
 type BatchedStore interface {
+	Store
 	Batch(context.Context) (Batch, error)
 }
--- a/pkg/storagev2/inmemstore/inmembatch.go
+++ b/pkg/storagev2/inmemstore/inmembatch.go
@@ -12,21 +12,25 @@ import (
 	storage "github.com/ethersphere/bee/pkg/storagev2"
 )
 
-func (s *Store) Batch(ctx context.Context) (storage.Batch, error) {
-	return &Batch{
-		mu:    new(sync.Mutex),
-		ctx:   ctx,
-		store: s,
-	}, nil
+type batchOp struct {
+	delete bool
+	item   storage.Item
 }
 
 type Batch struct {
-	mu     *sync.Mutex
-	ctx    context.Context
-	put    []storage.Item
-	delete []storage.Key
-	store  *Store
-	done   bool
+	mu    sync.Mutex
+	ctx   context.Context
+	ops   map[string]batchOp
+	store *Store
+	done  bool
+}
+
+func (s *Store) Batch(ctx context.Context) (storage.Batch, error) {
+	return &Batch{
+		ctx:   ctx,
+		ops:   make(map[string]batchOp),
+		store: s,
+	}, nil
 }
 
 func (i *Batch) Put(item storage.Item) error {
@@ -37,13 +41,9 @@ func (i *Batch) Put(item storage.Item) error {
 	i.mu.Lock()
 	defer i.mu.Unlock()
 
-	for _, put := range i.put {
-		if put.ID() == item.ID() {
-			return nil // disallow duplicates
-		}
+	i.ops[item.ID()] = batchOp{
+		item: item,
 	}
-
-	i.put = append(i.put, item)
 
 	return nil
 }
@@ -56,20 +56,11 @@ func (i *Batch) Delete(key storage.Key) error {
 	i.mu.Lock()
 	defer i.mu.Unlock()
 
-	for index, put := range i.put {
-		if put.ID() == key.ID() {
-			i.put = append(i.put[:index], i.put[index+1:]...)
-			return nil
-		}
-	}
+	sKey := getKeyString(key)
 
-	for _, del := range i.delete {
-		if del.ID() == key.ID() {
-			return nil // disallow duplicates
-		}
+	i.ops[sKey] = batchOp{
+		delete: true,
 	}
-
-	i.delete = append(i.delete, key)
 
 	return nil
 }
@@ -86,19 +77,19 @@ func (i *Batch) Commit() error {
 		return errors.New("already committed")
 	}
 
-	for _, item := range i.put {
-		if err := i.store.put(item); err != nil {
+	defer func() { i.done = true }()
+
+	for key, ops := range i.ops {
+		if ops.delete {
+			if err := i.store.delete(key); err != nil {
+				return err
+			}
+			continue
+		}
+		if err := i.store.put(ops.item); err != nil {
 			return err
 		}
 	}
-
-	for _, item := range i.delete {
-		if err := i.store.delete(item); err != nil {
-			return err
-		}
-	}
-
-	i.done = true
 
 	return nil
 }
--- a/pkg/storagev2/inmemstore/inmembatch.go
+++ b/pkg/storagev2/inmemstore/inmembatch.go
@@ -1,0 +1,90 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package inmemstore
+
+import (
+	"context"
+	"errors"
+	"sync"
+
+	storage "github.com/ethersphere/bee/pkg/storagev2"
+)
+
+func (s *Store) Batch(ctx context.Context) (storage.Batch, error) {
+	return &Batch{
+		mu:    new(sync.Mutex),
+		ctx:   ctx,
+		store: s,
+	}, nil
+}
+
+type Batch struct {
+	mu     *sync.Mutex
+	ctx    context.Context
+	put    []storage.Item
+	delete []storage.Key
+	store  *Store
+	done   bool
+}
+
+func (i *Batch) Put(item storage.Item) error {
+	if i.ctx.Err() != nil {
+		return i.ctx.Err()
+	}
+
+	i.mu.Lock()
+	defer i.mu.Unlock()
+
+	for _, put := range i.put {
+		if put.ID() == item.ID() {
+			return nil // disallow duplicates
+		}
+	}
+
+	i.put = append(i.put, item)
+
+	return nil
+}
+
+func (i *Batch) Delete(key storage.Key) error {
+	if i.ctx.Err() != nil {
+		return i.ctx.Err()
+	}
+
+	i.mu.Lock()
+	defer i.mu.Unlock()
+
+	for index, put := range i.put {
+		if put.ID() == key.ID() {
+			i.put = append(i.put[:index], i.put[index+1:]...)
+			return nil
+		}
+	}
+
+	i.delete = append(i.delete, key)
+
+	return nil
+}
+
+func (i *Batch) Commit() error {
+	if i.ctx.Err() != nil {
+		return i.ctx.Err()
+	}
+
+	i.store.mu.Lock()
+	defer i.store.mu.Unlock()
+
+	if i.done {
+		return errors.New("already committed")
+	}
+
+	for _, item := range i.put {
+		i.store.Put(item)
+	}
+
+	i.done = true
+
+	return nil
+}
--- a/pkg/storagev2/inmemstore/inmembatch.go
+++ b/pkg/storagev2/inmemstore/inmembatch.go
@@ -63,6 +63,12 @@ func (i *Batch) Delete(key storage.Key) error {
 		}
 	}
 
+	for _, del := range i.delete {
+		if del.ID() == key.ID() {
+			return nil // disallow duplicates
+		}
+	}
+
 	i.delete = append(i.delete, key)
 
 	return nil
@@ -81,7 +87,15 @@ func (i *Batch) Commit() error {
 	}
 
 	for _, item := range i.put {
-		i.store.Put(item)
+		if err := i.store.put(item); err != nil {
+			return err
+		}
+	}
+
+	for _, item := range i.delete {
+		if err := i.store.delete(item); err != nil {
+			return err
+		}
 	}
 
 	i.done = true
--- a/pkg/storagev2/inmemstore/inmembatch.go
+++ b/pkg/storagev2/inmemstore/inmembatch.go
@@ -9,7 +9,7 @@ import (
 	"errors"
 	"sync"
 
-	storage "github.com/ethersphere/bee/pkg/storagev2"
+	"github.com/ethersphere/bee/pkg/storagev2"
 )
 
 type batchOp struct {
@@ -41,7 +41,9 @@ func (i *Batch) Put(item storage.Item) error {
 	i.mu.Lock()
 	defer i.mu.Unlock()
 
-	i.ops[item.ID()] = batchOp{
+	key := getKeyString(item)
+
+	i.ops[key] = batchOp{
 		item: item,
 	}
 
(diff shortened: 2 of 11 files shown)


pkg/storagev2/batch.go
```diff
@@ -0,0 +1,19 @@
...
+type Batch interface {
+	// We will only need to batch write operations
+	Put(Item) error
+	Delete(Key) error
+
+	Commit() error
+}
+
+type BatchedStore interface {
+	Batch(context.Context) (Batch, error)
```
notanatol, line 18: it seems that for current implementation there's no case in which creation of batch returns error, should we change the signature or should we keep it like this for future cases @aloknerurkar ?

pkg/storagev2/inmemstore/inmembatch.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (i *Batch) Put(item storage.Item) error {
+	if i.ctx.Err() != nil {
+		return i.ctx.Err()
+	}
+
+	i.mu.Lock()
+	defer i.mu.Unlock()
+
+	for _, put := range i.put {
```
aloknerurkar, line 40: So I see potential issues here. If the batch is big, the Put operation will be very slow as it has to iterate through all the ops every time. We should avoid this bottleneck in the Put operation.

What I would do is just have a map that is keyed on the key. That way duplicates will just be overwritten in the batch. This will shift the responsibility of correctly using the batch to the user.

Also, this can be done with just a single map.
```go
type batchOp struct {
    delete bool
    item storage.Item
}

type batch struct {
    ops map[string]batchOp
}
```
Even so only one op will be allowed for each key. If it is a duplicate op it is overwritten. This is the same as the behavior a user would see if he using the Store.

pkg/storagev2/inmemstore/inmembatch.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (i *Batch) Delete(key storage.Key) error {
+	if i.ctx.Err() != nil {
+		return i.ctx.Err()
+	}
+
+	i.mu.Lock()
+	defer i.mu.Unlock()
+
+	for index, put := range i.put {
```
aloknerurkar, line 59: Same as the previous comment. Lets not do these iterations.

pkg/storagev2/leveldbstore/batch.go
```diff
@@ -0,0 +1,86 @@
...
+	return &Batch{
+		ctx:   ctx,
+		mu:    new(sync.Mutex),
+		batch: new(ldb.Batch),
+		store: s,
+	}, nil
+}
+
+type Batch struct {
+	mu  *sync.Mutex
```
aloknerurkar, line 27: I dont think this also needs to be a pointer.

pkg/storagev2/inmemstore/inmembatch.go
```diff
@@ -0,0 +1,104 @@
...
+
+	i.store.mu.Lock()
+	defer i.store.mu.Unlock()
+
+	if i.done {
+		return errors.New("already committed")
+	}
+
+	for _, item := range i.put {
+		if err := i.store.put(item); err != nil {
```
aloknerurkar, line 90: I dont think we need these separate functions. The mutexes are different so we should be able to use the exported functions itself.

pkg/storagev2/storagetest/storage.go
```diff
@@ -741,3 +782,28 @@ func BenchmarkDeleteSequential(b *testing.B, db storage.Store) {
...
+	resetBenchmark(b)
+	batch, _ := db.Batch(context.Background())
+	for i := 0; i < b.N; i++ {
+		item := &obj1{
+			Id: string(g.Key(i)),
+		}
+		if err := batch.Delete(item); err != nil {
+			b.Fatalf("delete key '%s': %v", string(g.Key(i)), err)
+		}
+	}
```
aloknerurkar, line 798: We need to do the commit here right?

pkg/storagev2/storagetest/batch.go
```diff
@@ -0,0 +1,88 @@
...
+
+func TestBatch(t *testing.T, s storage.Store) {
+	t.Run("duplicates are rejected", func(t *testing.T) {
+		b, _ := s.Batch(context.Background())
+
+		item1 := &obj1{
+			Id: "id1",
+		}
+
+		_ = b.Put(item1)
```
aloknerurkar, line 23: Why are these errors not checked?

pkg/storagev2/batch.go
```diff
@@ -0,0 +1,19 @@
...
+
+type Batch interface {
+	// We will only need to batch write operations
+	Put(Item) error
+	Delete(Key) error
+
+	Commit() error
+}
+
+type BatchedStore interface {
```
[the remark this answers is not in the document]
notanatol: added

pkg/storagev2/storagetest/batch.go
notanatol: fixed

pkg/storagev2/leveldbstore/batch.go
notanatol: fixed

pkg/storagev2/inmemstore/inmembatch.go
```diff
@@ -0,0 +1,104 @@
...
+func (s *Store) Batch(ctx context.Context) (storage.Batch, error) {
+	return &Batch{
+		mu:    new(sync.Mutex),
+		ctx:   ctx,
+		store: s,
+	}, nil
+}
+
+type Batch struct {
+	mu     *sync.Mutex
```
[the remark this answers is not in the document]
notanatol: fixed

pkg/storagev2/storagetest/batch.go
notanatol: fixed

pkg/storagev2/inmemstore/inmembatch.go
notanatol: fixed

pkg/storagev2/inmemstore/inmembatch.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (i *Batch) Put(item storage.Item) error {
+	if i.ctx.Err() != nil {
+		return i.ctx.Err()
+	}
+
+	i.mu.Lock()
+	defer i.mu.Unlock()
+
+	i.ops[item.ID()] = batchOp{
```
vladopajic, line 44: For delete operation, key in `ops` map is generated using ` getKeyString(key)` method. Here in put operation, key in `ops` map is just `item.ID()` value. Is there reason why ` getKeyString(key)` is not used here?

pkg/storagev2/inmemstore/inmemstore.go
```diff
@@ -105,6 +119,15 @@ func (s *Store) Delete(i storage.Item) error {
 	return nil
 }
 
+// for batching purpose
+func (s *Store) delete(key string) error {
+	if _, deleted := s.st.Delete(key); !deleted {
+		return storage.ErrNotFound
```
vladopajic, line 125: Returning error when attempting to remove item which is not present in store may not be ideal design decision. Handing this error will propagate all the way up to user which form their perspective can be vary valid case.

Error should indicate that operation was not performed successfully and not that zero elements were affected.

pkg/storagev2/leveldbstore/batch.go
```diff
@@ -0,0 +1,85 @@
...
+	done  bool
+}
+
+func (i *Batch) Put(item storage.Item) error {
+	if i.ctx.Err() != nil {
+		return i.ctx.Err()
+	}
+
+	i.mu.Lock()
+	defer i.mu.Unlock()
```
vladopajic, line 41: could locking scope be reduce to not include item marshaling?

pkg/storagev2/storagetest/batch.go
```diff
@@ -0,0 +1,114 @@
...
+	"testing"
+
+	storage "github.com/ethersphere/bee/pkg/storagev2"
+)
+
+func TestBatch(t *testing.T, s storage.Store) {
+	t.Run("duplicates are rejected", func(t *testing.T) {
+		b, _ := s.Batch(context.Background())
+
+		item1 := &obj1{
```
vladopajic, line 19: could `item1` and `item2` variables be defined in common scope for all test cases, so those could be reused?

pkg/storagev2/inmemstore/inmemstore.go
notanatol: how should this be refactored?

pkg/storagev2/inmemstore/inmembatch.go
```diff
@@ -0,0 +1,95 @@
...
+
+	storage "github.com/ethersphere/bee/pkg/storagev2"
+)
+
+type batchOp struct {
+	delete bool
+	item   storage.Item
+}
+
+type Batch struct {
```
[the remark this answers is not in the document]
notanatol: We could but I think it's more valuable to keep them separate and let them evolve along their own logical lines, which might diverge in the future. A little copy is better than a little dependency, after all.

pkg/storagev2/leveldbstore/batch.go
```diff
@@ -0,0 +1,85 @@
...
+
+	dbKey := []byte(key.Namespace() + separator + key.ID())
+
+	i.batch.Delete(dbKey)
+
+	return nil
+}
+
+func (i *Batch) Commit() error {
```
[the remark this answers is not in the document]
notanatol: not for level db, it has an internal locking mechanism

pkg/storagev2/inmemstore/inmembatch.go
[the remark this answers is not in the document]
vladopajic: There is problem when code starts do diverge. Imagine have any two (or more) similar functionalities which could benefit from same abstraction. Somewhere down do road, developer may find issue in one and fix it only in for this case, leaving other(s) vulnerable to same bug. 

Why would dependency to same batching mechanism be a problem?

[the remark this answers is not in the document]
notanatol: I wasn't able to figure out the commonalities of the two, besides, at this point we might want to leave this as it is now and optimize it later once we have the bandwidth.

pkg/storagev2/inmemstore/inmemstore.go
vladopajic: In this case of `delete` and `Delete` methods no error should be returned if nothing was removed.

aloknerurkar: I don't think this is the right way to go about it. The error should be returned and the user can choose how to handle it. He can very well ignore it. But if there is a case when the user cares about this, it should be sent up.

pkg/storagev2/leveldbstore/batch.go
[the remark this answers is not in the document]
vladopajic: If `db.Write` has  internal locking, then `Delete` or `Put` can still modify same `i.batch` object while `db.Write` is execution on this batch. If this is right then locking on `Commit` would be needed to guard against this.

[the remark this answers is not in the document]
notanatol: fixed

pkg/storagev2/inmemstore/inmembatch.go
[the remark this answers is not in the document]
vladopajic: All what batch does is it collects operations to be preformed in `Commit`. Commit operation could have different implementations for `inmem` and `lvldb` stores. Pretty much everything just like in `inmemstore.Batch` currently does, except that `Commit` needs to support different implementations.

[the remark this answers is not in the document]
aloknerurkar: The leveldb batch is writing the operations to `*leveldb.Batch` and the inmem one is writing to inmem map. The inmem batch is just a reference implementation. It doesnt provide the same level of functionality as `*leveldb.Batch`. The `WriteBatch` function is atomic.

Batch will be in place of a Store when user is using it. So it needs to provide the same API as the write operations. Just the additional Commit function. Different implementations can have different staging mechanisms. So the batch interface is flexible enough right now.

pkg/storagev2/inmemstore/inmemstore.go
vladopajic: It add unnecessary complexity because user need to check if error is because operation has failed or nothing was removed. 

After executing `delete` operation all what user should care is that this element is not present in the state, considering operation was successful.

What should happen happen when this error was returned when batch is committed? Should whole batch fail? Since user can't deal with this error, can't ignore it, their whole batch is going to fail unwillingly. It looks like we made decision for user in this case

pkg/storagev2/inmemstore/inmembatch.go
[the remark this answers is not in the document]
vladopajic: All I am saying is that `inmemstore.Batch` and `leveldbstore.Batch` could use same mechanism because they are essentially just collecting what operations to execute. Only `Commit` step is different, which can be delegated to different implementations.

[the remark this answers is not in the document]
notanatol: I think this is one of those cases where it's easier to show than to tell.
Anyway, as Alok was saying - the inmem implementation is for reference, so there's no real and tangible benefit from optimizing things around it, at least right now.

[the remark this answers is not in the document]
vladopajic: https://github.com/ethersphere/bee/pull/3314/files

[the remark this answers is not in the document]
aloknerurkar: So I don't think the approach in this PR is better. Typically in a Batch, you would batch operations. So the implementation of how to batch the operations will be specific to the implementation. So the staging of the ops is also implementation dependent.

This is modeled after the levelDB.Batch interface. This also provides the same set of API (Put, Delete, Commit). This is also similar to what bolt and badger (and other KV stores) provide. The goal of the Batch interface here is to provide a consistent interface on top of these so that we can use any backend. Also, it needs to provide the Write APIs of the Store, so it can be used in place of that.

Also, these need to synchronize with the Store implementations as well. For eg. in the inmem implementation, we need access to the main mutex in order to push all the updates together. But this is not required for levelDB. But also in levelDB, we can't have multiple batches at a time, so we would need some sort of synchronization. There will be more such dependencies.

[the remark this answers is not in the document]
vladopajic: This was draft PR just to show that single batch implementation would be sufficient for any specific implementation. I am not sure what are exact cons for this implementation? Because, synchronization shouldn't be a problem.

Either way, it's your call.

pkg/storagev2/storagetest/batch.go
```diff
@@ -0,0 +1,100 @@
...
+		}, func(r storage.Result) (bool, error) {
+			t.Fatalf("expected empty store, got %v", r.Entry)
+			return true, nil
+		})
+		if err != nil {
+			t.Fatal("iterate", err)
+		}
+	})
+
+	t.Run("batche not reusable after commit", func(t *testing.T) {
```
aloknerurkar, line 79: Do we need the 'e' at the end? Same for the next test.

pkg/storagev2/inmemstore/inmembatch.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return nil
+}
+
+func (i *Batch) Commit() error {
+	if i.ctx.Err() != nil {
+		return i.ctx.Err()
+	}
+
```
aloknerurkar, line 74: So in order to be sure 100% we need to also have the lock on the batch mutex. During commit we cannot handle any Put or Delete operations.

pkg/storagev2/storagetest/batch.go
notanatol: fixed

pkg/storagev2/inmemstore/inmembatch.go
notanatol: fixed

aloknerurkar (approved):

Issue #3255, opened 2022-09-07
code linting


vladopajic:
1. have single source for linting rules symmetrical in CI/CD and local environments
Bee code currently has three different rules and only one of those is enforced by CI. Problems with this way of linting: a) it causes confusion what are rules to be followed, b) rules which are not enforced by CI are going to be forgotten  c) rules executed in CI/CD should be symmetrical to local (so devs can check code locally)
What could be done: have single source for all linting rules in `.golangci.yml` and single `make lint` command to run those lints on entire code base. 

2. trailing whitespace check only works in CI
This check seems to be only working in CI on diff changes. It would be easier for devs if this check is symmetrical in CI/CD and local environments.
What could be done: implement command `make whitespace-check` which would check entire code base.

3. increase `golangci-lint` strictness 
In order to ensure higher code quality we should consider including more linters. Below 

```
// compliant rules list - this list includes rules which bee can start using right away because code is compliant with them

    - asciicheck
    - bidichk
    - deadcode
    - depguard
    - errcheck
    - gofmt
    - goheader
    - gomoddirectives
    - goprintffuncname
    - gosimple
    - govet
    - importas
    - ineffassign
    - megacheck
    - misspell
    - misspell
    - nakedret
    - nilerr
    - promlinter
    - rowserrcheck
    - sqlclosecheck
    - staticcheck
    - structcheck
    - typecheck
    - unconvert
    - unconvert
    - unused
    - varcheck
  ```
  
  ```
//  easy to fix list - list of rules which could be easily fixed (there are few problems in the code witch needs to be addressed)

    - errorlint ##important
    - errname
    - dogsled
    - bodyclose
    - exportloopref
    - forbidigo
    - goconst
    - gochecknoinits
    - gofumpt
    - nilnil
    - nlreturn
    - noctx
    - prealloc
    - predeclared
    - wastedassign
    - whitespace
  ```
  
  ```
// test related - this list includes rules which affect how tests are written. these seem to be relatively easy to be fixed

   - paralleltest (important)
   - testpackage (easy)
   - thelper #easy
  ```

mrekucci:
> - megacheck
 
 Is deprecated.

> - rowserrcheck
> - sqlclosecheck

Only relevant to SQL.

> - bodyclose

Not needed for server-side since: https://github.com/golang/go/blob/master/src/net/http/request.go#L174-L177

> - gofumpt

We should stick with `gofmt`.

> - nilnil

[...]

PR #3293, opened 2022-09-15
chore(lint): make test execute in parallel (part #1)

vladopajic:
### Checklist

### Description
This PR is the first part of effort to improve static code analysis (issue https://github.com/ethersphere/bee/issues/3255)

* Enabeld `paralleltest` with configuration temporally disabling it on packages which remain to be resolved. Lint is enabled so that currently fixed packages won't drift away from this rule.
* This PR just set some test to execute in parallel. There will be more effort split into many PRs which fix and enable lint in currently disabled packages

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3293)

### DIFF
--- a/pkg/jsonhttp/jsonhttp_test.go
+++ b/pkg/jsonhttp/jsonhttp_test.go
@@ -17,6 +17,8 @@ import (
 )
 
 func TestRespond_defaults(t *testing.T) {
+	t.Parallel()
+
 	w := httptest.NewRecorder()
 
 	jsonhttp.Respond(w, 0, nil)
@@ -46,6 +48,8 @@ func TestRespond_defaults(t *testing.T) {
 }
 
 func TestRespond_statusResponse(t *testing.T) {
+	t.Parallel()
+
 	for _, tc := range []struct {
 		code int
 	}{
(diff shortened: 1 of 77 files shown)


pkg/jsonhttp/jsonhttp_test.go
```diff
@@ -125,6 +129,8 @@ func TestRespond_statusResponse(t *testing.T) {
 }
```
notanatol, line 129: I think the whole body within the loop can be inside a parallel `t.Run`, no?

pkg/traversal/traversal_test.go
```diff
@@ -147,6 +149,8 @@ func TestTraversalBytes(t *testing.T) {
 	for _, tc := range testCases {
```
notanatol, line 149: does this also needs the `tc := tc`? maybe in the example bellow also

pkg/jsonhttp/jsonhttp_test.go
vladopajic: At this point I am trying to just add `t.Parallel()` to functions, without further optimising tests. Because that itself is different rabbit hole.

pkg/traversal/traversal_test.go
vladopajic: correct; fixed in all places

notanatol (approved):

istae (approved):

PR #3321, opened 2022-09-21
feat: max out gas limit

metacertain:
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3321)

### DIFF
--- a/pkg/postage/postagecontract/contract.go
+++ b/pkg/postage/postagecontract/contract.go
@@ -118,7 +118,7 @@ func (c *postageContract) sendTransaction(ctx context.Context, callData []byte, 
 		To:          &c.postageContractAddress,
 		Data:        callData,
 		GasPrice:    sctx.GetGasPrice(ctx),
-		GasLimit:    30000000,
+		GasLimit:    14000000,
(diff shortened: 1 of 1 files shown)


istae (approved):

notanatol (approved):

mrekucci (approved):

pkg/postage/postagecontract/contract.go
```diff
@@ -118,7 +118,7 @@ func (c *postageContract) sendTransaction(ctx context.Context, callData []byte,
 		To:          &c.postageContractAddress,
 		Data:        callData,
 		GasPrice:    sctx.GetGasPrice(ctx),
-		GasLimit:    1600000,
+		GasLimit:    30000000,
```
ralph-pichler, line 121: not a good idea. this tx can only be included if it is the only tx in the block (also you need to be able to afford the entire transaction even if it is only a lot less).

swarmHaseeb (approved):

aloknerurkar:
Can we close this? Now that #3352 is merged?

PR #3322, opened 2022-09-21
fix: reserve size to return chunks with po greater than and equal to storage radius

istae:
### Checklist

### Description
When the depthmonitor changes the storage radius, this does not automatically alter the radius that the localstore keeps in it's reserve, so the `ReserveSize` does not return an accurate number. With this PR, we iterate on the pull index to manually count all chunks whose PO is greater than or equal to the storage radius.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3322)

### DIFF
--- a/pkg/localstore/reserve.go
+++ b/pkg/localstore/reserve.go
@@ -164,3 +164,37 @@ func withinRadius(db *DB, item shed.Item) bool {
 	po := db.po(swarm.NewAddress(item.Address))
 	return po >= item.Radius
 }
+
+func (db *DB) ComputeReserveSize(startPO uint8) (uint64, error) {
+
+	var count uint64
+
+	err := db.pullIndex.Iterate(func(item shed.Item) (stop bool, err error) {
+		count++
+		return false, nil
+	}, &shed.IterateOptions{
+		StartFrom: &shed.Item{
+			Address: generateAddressAt(db.baseKey, int(startPO)),
+		},
+	})
+
+	return count, err
+}
+
+func generateAddressAt(baseBytes []byte, prox int) []byte {
+
+	addr := make([]byte, 32)
+
+	for po := 0; po < prox; po++ {
+		index := po % 8
+		if baseBytes[po/8]&(1<<(7-index)) > 0 { // if baseBytes bit is 1
+			addr[po/8] |= 1 << (7 - index) // set addr bit to 1
+		}
+	}
+
+	if baseBytes[prox/8]&(1<<(7-(prox%8))) == 0 { // if baseBytes PO bit is zero
+		addr[prox/8] |= (1 << (7 - (prox % 8))) // set addr bit to 1
+	}
+
+	return addr
+}
--- a/pkg/localstore/reserve.go
+++ b/pkg/localstore/reserve.go
@@ -165,6 +165,9 @@ func withinRadius(db *DB, item shed.Item) bool {
 	return po >= item.Radius
 }
 
+// ComputeReserveSize iterates on the pull index to count all chunks
+// starting at some proximity order with an generated address whose PO
+// is used as a starting prefix by the index.
 func (db *DB) ComputeReserveSize(startPO uint8) (uint64, error) {
 
 	var count uint64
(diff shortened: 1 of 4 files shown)


pkg/localstore/reserve.go
```diff
@@ -164,3 +164,37 @@ func withinRadius(db *DB, item shed.Item) bool {
...
+
+func (db *DB) ComputeReserveSize(startPO uint8) (uint64, error) {
+
+	var count uint64
+
+	err := db.pullIndex.Iterate(func(item shed.Item) (stop bool, err error) {
+		count++
+		return false, nil
+	}, &shed.IterateOptions{
+		StartFrom: &shed.Item{
```
aloknerurkar, line 176: Lets add some comments here on why this gives the whole reserve in this iteration.

aloknerurkar (approved):

pkg/localstore/reserve.go
```diff
@@ -164,3 +164,37 @@ func withinRadius(db *DB, item shed.Item) bool {
...
+
+	for po := 0; po < prox; po++ {
+		index := po % 8
+		if baseBytes[po/8]&(1<<(7-index)) > 0 { // if baseBytes bit is 1
+			addr[po/8] |= 1 << (7 - index) // set addr bit to 1
+		}
+	}
+
+	if baseBytes[prox/8]&(1<<(7-(prox%8))) == 0 { // if baseBytes PO bit is zero
+		addr[prox/8] |= (1 << (7 - (prox % 8))) // set addr bit to 1
```
mrekucci, line 196: Redundant parentheses, can be transformed to: `1 << (7 - (prox % 8))`.

mrekucci (approved):

pkg/localstore/reserve.go
istae: see the function comment

pkg/localstore/reserve.go
```diff
@@ -164,3 +164,40 @@ func withinRadius(db *DB, item shed.Item) bool {
...
+	}, &shed.IterateOptions{
+		StartFrom: &shed.Item{
+			Address: generateAddressAt(db.baseKey, int(startPO)),
+		},
+	})
+
+	return count, err
+}
+
+func generateAddressAt(baseBytes []byte, prox int) []byte {
```
vladopajic, line 187: This function definition seems to be more suitable to be in `swarm` package, together with other `swarm.Address` utilities. 

Also, use of `swarm.Address` type instead of raw `[]byte` would be more indicative.

pkg/localstore/reserve_test.go
```diff
@@ -685,6 +685,43 @@ func TestReserveSize(t *testing.T) {
 	})
 }
 
+func TestComputeReserveSize(t *testing.T) {
+
```
mrekucci, line 689: Adding `t.Parallel()` would be nice.

pkg/localstore/reserve.go
ldeffenb: For nodes with lots of locally pinned data, iterating an index just to count things is very expensive.  I know this from the experience of watching how long it takes the pull cursors to be calculated on my 1.3TB node.  This node is currently taking just under an HOUR to handle a cursors request.  I have added a throttle to only allow a single request to be handled at a time to protect the CPU and disk from extreme overload.  I don't recall what index the cursor determination uses, but it's in the chain from pullsync to cursorHandler to the Cursors() invocation.

ldeffenb: I added an elapsed time log around ComputeReserveSize on my 1.3TB node and here's the first two passes:
```"time"="2022-09-23 14:46:48.126496" "level"="debug" "logger"="node/localstore" "v"=1 "msg"="ComputeReserveSize executed" "po"=12 "elapsed"="3m28.282063599s" "count"=82552
"time"="2022-09-23 14:46:48.127805" "level"="debug" "logger"="node/depthmonitor" "msg"="depthmonitor: state" "current size"=82552 "radius"=12 "chunks/sec rate"=8.024555244444445
"time"="2022-09-23 14:54:02.156451" "level"="debug" "logger"="node/localstore" "v"=1 "msg"="ComputeReserveSize executed" "po"=12 "elapsed"="2m14.028099138s" "count"=82555
"time"="2022-09-23 14:54:02.156488" "level"="debug" "logger"="node/depthmonitor" "msg"="depthmonitor: state" "current size"=82555 "radius"=12 "chunks/sec rate"=4.99```

PR #3335, opened 2022-09-23
feat: reserve commitment sampler

aloknerurkar:
### Checklist

### Description
Implements the reserve commitment sampler. Specs can be found [here](https://hackmd.io/@NivJ0VAwTB2QRu3weL9S2g/BJR1t_Ulj)

Based on the prototype in the `commit_sampler` branch.

### Testing results
So this branch was run on the `dev-bee-storage` cluster. This cluster runs 5 nodes on 1 pod. The nodes were actively taking part in syncing and normal activities during this time.

Ran different images with worker counts 4, 6 and 8. With count 6, we see some slight improvement. However with 8 workers we start seeing diminishing returns. So the final worker count, for now, will be 6.

- 4 workers
```
{
Sample: {
Items: [
"00000162adf137b15d526b6b23981fdb9b24ec431afe485c0c74d6c5ab676047",
"00000aafed3df08bcbf0347587114936bdebe0ac775a1ce276dc2ae362d7a966",
"00000f60ba8962c1ba516d24578fe69d1f4a0eceb4831d30d00ea5c5a82024bd",
"00001ca8cbe2e2604b8d3e7571c3e64b18696a765a0dae1468ba5cae81dc92b3",
"00001df91752d01639951d3858e3a8996702b05de4352323fc8a1e1fe0e1aa02",
"00001f260c90631df8844e20c83e1f4eda7250db26dcfafe8a85062c44b76784",
"00002194572c16894df13dcf1e9cc4ea04a12b519bd1b72604ee4dbfe7901037",
"0000273a17cf713c2d53d9f4d79c8d87a4125515bd3e0ee02f799ce8644e3a2a",
"00002e8dc06a613f3e143644b25f69ebcfd99345e0fc4016ce72a80620de226d",
"000034f5fef7eb20d43e20d47480136fb924597b08d8665abbac275f016ba4b0",
"0000354ae3130b42ab8dd31af29faf702c2ffdbbe93ab8adf75a0f1e80a52e67",
"00003682a2ec0079f9bd0a2ceea775eda4f0717daa8039bba7246f217a07d894",
"0000376e6ec6ce6702bd56f1b817a3a79da25fdc35156c04f0bb7e371d83701d",
"00003911e878d2868c54d4fb074098e2949d88f88ecf614de5d730f498d20601",
"00003d5f10bb85d7d424a17255cd02e8d7de6616a6a726c06abbf236d6f31885",
"00004e0f0915926eed1234091cd4f4f9634c6a2a58619d7493b7f08a199e2bc3"
],
Hash: "f4e4ff22644fadd85a7ab461ded34c7c4dd9ebf1a85f567cc102499c31baddea"
},
Time: "3m6.645507847s"
}
```

- 6 workers
```
{
Sample: {
Items: [
"00000d785fd83a0084feac1eced20a7f38085264f96680a7e7a301ebe32c856f",
"00000edd6778d93b7599819bdc4c7986cbb8059c35099f6bf4fb16aea29ef2e8",
"00001052d2bc5cfb6f3db94f8ddec0eef957469b2879bbcab633407d47932697",
"0000144dc51e4265528534cd4c2bb53a75c28d8f4b619c3c0e883a8b4d81d637",
"00001d058976a434bebbf79ac49c554b41f686256dc768e40a0e7b521778bb5c",
"00001dd311846d3cf09d0b66c4f9478c747601cb7dc408114d6588ff5fc56589",
"00001e95f945388298e5fba6061654a7d8053834f6d4b7dcab908c0a6f4a862e",
"00001ea973facc6aa159d9abb883b808c57d38c300cea138756cfc25866840fd",
"000023504665cdd41cfb8ff97eb39ed84ac7226592a3d091ec47fe5e1166917c",
"00002390d63d6b475db3aefe7cb9fb418cabba108041cfc2aba0bc4ed0b1cf9e",
"000024a979b9f257b2179437dc6371d3270aad28bb1c36015435efd3948306d2",
"000026106a14ef0b6abf8c0a79e22a80b9ac8d540a23e989b92f1c97acd606a3",
"00002ca846d080459062d51dad70a8a95d718637b8c7037388f0486837d5bd0d",
"00003699f83a549a0c636436cfaf5a732e1b8d2abca08bdd062036744fed895f",
"00003c20b9ac93f37dec41995d24a0188bc1715a69720f8322e60c678c7ffc8f",
"00003f2867110829800cd6cba41f9687813af0805e345f2040a773cc8f4be5ec"
],
Hash: "575dfcf935420b69b50da2f4f15ff0505a21b7a2f0ea41b793fbab59b638e295"
},
Time: "2m28.464781893s"
}
```

- 8 workers
```
{
Sample: {
Items: [
"000002a24038edfb43e7b6cb400e126a6265c4f09301e7110ff28c53f260a0ec",
"0000074e9fa265e239a8b0ff9beffaa42796fb88d3d9f0b9cd5cf4abd77f3096",
"00000b2c4d809588cde9e85785f1d20c794645aa1480e3bb9914c29f0a88dca1",
"00000b47eeb2f2893c0d72f291c95e5f1caef3f9dcaef5481ed0a434f862ba90",
"00000f3029dcc17a1776ced8275f644b18d200f6e8b59b2c7967c7062522d8a1",
"000013e5102771a6139aaad6ab44994e19fe26a614ab0f20d800d57cd48a563f",
"0000140ac5d6c19ce21f8857e2f12518e0ad8d807440f8776d5422cd130443e2",
"000015c785101e7311c9c10a2c23e8157b7567b563f7a47b07e1b78ddf62ef0e",
"00001dc2c178bcdfdb7f21a11d537aa1c5d67679424fee7acd9dc3883c694225",
"00001ecc618d7eebfd1081c469760f59e35f8b0aa2adbebdc70798a09e38b2fc",
"0000231cb2e2e73a515fd347e0f8736de98e563758fbfc86b9834a80103e91ba",
"000023c5ce3c6ff5f63b829e73a41aaa5b00f7b7ca6dcf8c2053976afce6c440",
"000025eb1eb3361367d9bc0bb89a4c3c2d6d91569b8031acbe9ee4a358ea8345",
"000026a14f9f8b2b09b8be4c4569d78fa93e6fa23c053e01ead4128f73e7dd0a",
"00002f9df9a581305a77f5af72747414b08ef5b9810e5299ab2ccd044fa18cae",
"0000324e9f37e68c607da3b020d409e37418bbb7ee5783688af0da69162d102e"
],
Hash: "d2bb8aa66507868a2a052636be09c3658149bd594633c35ba440c152340acf59"
},
Time: "2m51.176189251s"
}
```

Dashboards show spikes but things normalizing. Ran multiple times on multiple nodes sharing the same pod. We are able to get consistent results. Things go back to normal after the sampling is done.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3335)

### DIFF
--- a/pkg/api/rchash.go
+++ b/pkg/api/rchash.go
@@ -1,0 +1,47 @@
+package api
+
+import (
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/gorilla/mux"
+)
+
+type rchash struct {
+	Sample storage.Sample
+	Time   string
+}
+
+func (s *Service) rchasher(w http.ResponseWriter, r *http.Request) {
+
+	start := time.Now()
+	depthStr := mux.Vars(r)["depth"]
+
+	depth, err := strconv.ParseUint(depthStr, 10, 8)
+	if err != nil {
+		s.logger.Error(err, "reserve commitment hasher: invalid depth")
+		jsonhttp.BadRequest(w, "invalid depth")
+		return
+	}
+
+	if depth > 255 {
+		depth = 255
+	}
+
+	anchorStr := mux.Vars(r)["anchor"]
+	anchor := []byte(anchorStr)
+
+	sample, err := s.storer.ReserveSample(r.Context(), anchor, uint8(depth))
+	if err != nil {
+		jsonhttp.BadRequest(w, "invalid")
+		return
+	}
+
+	jsonhttp.OK(w, rchash{
+		Sample: sample,
+		Time:   time.Since(start).String(),
+	})
+}
--- a/pkg/api/rchash.go
+++ b/pkg/api/rchash.go
@@ -19,6 +19,10 @@ type rchash struct {
 	Time   string
 }
 
+// TODO: Remove this API before next release. This API is kept mainly for testing
+// the sampler till the storage incentives agent falls into place. As a result,
+// no documentation or tests are added here. This should be removed before next
+// breaking release.
 func (s *Service) rchasher(w http.ResponseWriter, r *http.Request) {
 
 	start := time.Now()
@@ -40,7 +44,8 @@ func (s *Service) rchasher(w http.ResponseWriter, r *http.Request) {
 
 	sample, err := s.storer.ReserveSample(r.Context(), anchor, uint8(depth))
 	if err != nil {
-		jsonhttp.BadRequest(w, "invalid")
+		s.logger.Error(err, "reserve commitment hasher: failed generating sample")
+		jsonhttp.InternalServerError(w, "failed generating sample")
 		return
 	}
 
--- a/pkg/api/rchash.go
+++ b/pkg/api/rchash.go
@@ -1,3 +1,7 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 package api
 
 import (
--- a/pkg/localstore/sampler.go
+++ b/pkg/localstore/sampler.go
@@ -15,6 +15,8 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const sampleSize = 16
+
 type sampleStat struct {
 	TotalIterated     atomic.Int64
 	NotFound          atomic.Int64
@@ -110,7 +112,6 @@ func (db *DB) ReserveSample(ctx context.Context, anchor []byte, storageDepth uin
 		close(sampleChan)
 	}()
 
-	const sampleSize = 16
 	samples := make([]storage.SampleItem, 0, sampleSize)
 	insert := func(item storage.SampleItem) {
 		added := false
--- a/pkg/localstore/sampler.go
+++ b/pkg/localstore/sampler.go
@@ -77,7 +77,7 @@ func (db *DB) ReserveSample(ctx context.Context, anchor []byte, storageDepth uin
 
 			for addr := range addrChan {
 				getStart := time.Now()
-				ch, err := db.Get(ctx, storage.ModeGetSync, addr)
+				chItem, err := db.get(ctx, storage.ModeGetSync, addr)
 				stat.GetDuration.Add(time.Since(getStart).Microseconds())
 				if err != nil {
 					stat.NotFound.Inc()
@@ -85,7 +85,7 @@ func (db *DB) ReserveSample(ctx context.Context, anchor []byte, storageDepth uin
 				}
 
 				hmacrStart := time.Now()
-				hmacr.Write(ch.Data())
+				hmacr.Write(chItem.Data)
 				taddr := hmacr.Sum(nil)
 				hmacr.Reset()
 				stat.HmacrDuration.Add(time.Since(hmacrStart).Microseconds())
--- a/pkg/localstore/sampler.go
+++ b/pkg/localstore/sampler.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/ethersphere/bee/pkg/bmtpool"
 	"github.com/ethersphere/bee/pkg/shed"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
@@ -69,7 +70,7 @@ func (db *DB) ReserveSample(ctx context.Context, anchor []byte, storageDepth uin
 		return nil
 	})
 
-	sampleChan := make(chan storage.SampleItem)
+	sampleItemChan := make(chan swarm.Address)
 	const workers = 4
 	for i := 0; i < workers; i++ {
 		g.Go(func() error {
@@ -91,10 +92,7 @@ func (db *DB) ReserveSample(ctx context.Context, anchor []byte, storageDepth uin
 				stat.HmacrDuration.Add(time.Since(hmacrStart).Microseconds())
 
 				select {
-				case sampleChan <- storage.SampleItem{
-					Address:            addr,
-					TransformedAddress: swarm.NewAddress(taddr),
-				}:
+				case sampleItemChan <- swarm.NewAddress(taddr):
 					// continue
 				case <-ctx.Done():
 					return ctx.Err()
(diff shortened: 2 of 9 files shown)


pkg/localstore/sampler.go
```diff
@@ -0,0 +1,163 @@
...
+
+	sampleChan := make(chan storage.SampleItem)
+	const workers = 4
+	for i := 0; i < workers; i++ {
+		g.Go(func() error {
+			hmacr := hmac.New(swarm.NewHasher, anchor)
+
+			for addr := range addrChan {
+				getStart := time.Now()
+				chItem, err := db.get(ctx, storage.ModeGetSync, addr)
```
zelig, line 80: make sure using db.get should not have any sideeffects.

pkg/localstore/sampler.go
```diff
@@ -0,0 +1,163 @@
...
+
+				hmacrStart := time.Now()
+				hmacr.Write(chItem.Data)
+				taddr := hmacr.Sum(nil)
+				hmacr.Reset()
+				stat.HmacrDuration.Add(time.Since(hmacrStart).Microseconds())
+
+				select {
+				case sampleChan <- storage.SampleItem{
+					Address:            addr,
```
zelig, line 95: why is the address needed in the items?

pkg/localstore/sampler.go
```diff
@@ -0,0 +1,163 @@
...
+			return nil
+		})
+	}
+
+	go func() {
+		g.Wait()
+		close(sampleChan)
+	}()
+
+	samples := make([]storage.SampleItem, 0, sampleSize)
```
zelig, line 115: samples is not good name, it is one sample

pkg/localstore/sampler.go
```diff
@@ -0,0 +1,163 @@
...
+		if len(samples) > sampleSize {
+			samples = samples[:sampleSize]
+		}
+		if len(samples) < sampleSize && !added {
+			samples = append(samples, item)
+		}
+	}
+
+	for sample := range sampleChan {
+		lowestAddr := swarm.NewAddress(make([]byte, 32))
```
zelig, line 135: this is not the best name. should be `sampleMax` or something

pkg/localstore/sampler.go
```diff
@@ -0,0 +1,163 @@
...
+		if len(samples) > 0 {
+			lowestAddr = samples[len(samples)-1].TransformedAddress
+		}
+		if leq(sample.TransformedAddress.Bytes(), lowestAddr.Bytes()) || len(samples) < sampleSize {
+			insert(sample)
+		}
+	}
+
+	hasher := swarm.NewHasher()
+	defer hasher.Reset()
```
zelig, line 145: no need to call reset on return, the hasher is not gonnabe reused

pkg/localstore/sampler.go
```diff
@@ -0,0 +1,163 @@
...
+
+	for _, s := range samples {
```
zelig, line 147: here we should be making the assumption that samplesize <=128

pkg/localstore/sampler.go
```diff
@@ -0,0 +1,163 @@
...
+		lowestAddr := swarm.NewAddress(make([]byte, 32))
+		if len(samples) > 0 {
+			lowestAddr = samples[len(samples)-1].TransformedAddress
+		}
+		if leq(sample.TransformedAddress.Bytes(), lowestAddr.Bytes()) || len(samples) < sampleSize {
+			insert(sample)
+		}
+	}
+
+	hasher := swarm.NewHasher()
```
zelig, line 144: this needs to be the BMT hasher, since we might need inclusion proofs on the sample

pkg/localstore/sampler.go
```diff
@@ -0,0 +1,163 @@
...
+	sample := storage.Sample{
+		Items: samples,
+		Hash:  swarm.NewAddress(hash),
+	}
+	logger.Info("Sampler done", "Stats", stat.String(), "Sample", sample)
+
+	return sample, nil
+}
+
+func leq(a, b []byte) bool {
```
zelig, line 161: `leq` stands for less or equal, this is just a less  function, so `le`

zelig (commented):
some early comments, hope you dont mind

pkg/api/rchash.go
```diff
@@ -0,0 +1,47 @@
...
+	if depth > 255 {
+		depth = 255
+	}
+
+	anchorStr := mux.Vars(r)["anchor"]
+	anchor := []byte(anchorStr)
+
+	sample, err := s.storer.ReserveSample(r.Context(), anchor, uint8(depth))
+	if err != nil {
+		jsonhttp.BadRequest(w, "invalid")
```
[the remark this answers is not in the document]
aloknerurkar: I am not sure if we should have this API? Right now I have it so we can test this on the cluster nodes which already have some significant reserve. I plan to remove this before checking in.

pkg/localstore/sampler.go
aloknerurkar: It doesnt. In fact this was the problem when we initially implemented it during the hackathon. As we were using the storer provided in the API which is actually a netstore wrapped storer, the node was trying to retrieve chunks which it did not have. With the unexported get, we will only use the local state.

pkg/localstore/sampler.go
aloknerurkar: We had it during testing. I guess its not required. Removed now.

pkg/localstore/sampler.go
aloknerurkar: So the insert function guarantees that we will never have samples exceed the sampleSize. It will pop any item beyond it. So I guess we dont need to check this here unless you think you want to have this extra safety.

Also, currently the sample size is 16. Not sure if this is the agreed number, but during our discussion with abel, we decided this no.

pkg/localstore/sampler.go
```diff
@@ -0,0 +1,195 @@
...
+	addrChan := make(chan swarm.Address)
+	var stat sampleStat
+	logger := db.logger.WithName("sampler").V(1).Register()
+
+	// Phase 1: Iterate chunk addresses
+	g.Go(func() error {
+		defer close(addrChan)
+		iterationStart := time.Now()
+		for bin := storageDepth; bin <= swarm.MaxPO; bin++ {
+			err := db.pullIndex.Iterate(func(item shed.Item) (bool, error) {
```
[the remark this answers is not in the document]
aloknerurkar: So I checked using the StartFrom option. The IterationTime is almost the same. The reason I kept it this way is we can further optimize this by having more workers here iterating on different bins.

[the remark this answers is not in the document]
istae: also the concept of swarm.MaxPO is an optimization in general, it is entirely possible that you might have chunks with PO's greater than the MaxPO (albeit very few). The correct thing to do is iterate on the whole thing.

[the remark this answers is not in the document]
istae: so if the performance is the same, then I'd go with iterating without bins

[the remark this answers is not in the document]
ldeffenb: And it may be approximately the same for small pullIndex populations, but it's a huge difference if you have lots of out-of-neighborhood chunks pinned in a node!   The fewer times we iterate the index, the better.

istae (approved):

pkg/localstore/sampler.go
```diff
@@ -0,0 +1,197 @@
...
+				if err != nil {
+					stat.NotFound.Inc()
+					continue
+				}
+
+				hmacrStart := time.Now()
+				_, err = hmacr.Write(chItem.Data)
+				if err != nil {
+					return err
+				}
```
zelig, line 111: I think this may be like this but if we ever want to inclusion proof sample items it may be better to use Single Segment Transform with BMT instead of HMAC

zelig (approved):

pkg/localstore/sampler.go
aloknerurkar: So I thought with the BMT hasher we should be able to verify 32byte segments (hmac results) no?

PR #3341, opened 2022-09-26
feat(stakingAgent): implemented api and contract comm for staking

### LINKED ISSUE #3071: Staking Agent
This component's aim is to provide a convenient user experience for staking as a participant in the storage incentives redistribution process.

The most simple form would take some user defined value from an API providing amount to stake, which is then consequentially deposited in the staking contract along with the node's overlay using the 'register'-like method of the contract.

expected smart contract interface for interactions:
**Deposit(bytes32 overlay, uint256 value)**

swarmHaseeb:
### Checklist

- [x] My change requires updating the Open API specification and/or changing its version, and I've done it.

### Description

implement api for staking so that user can stake bzz
### Related Issue (Optional)

closes #3071 

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3341)


pkg/api/staking_test.go
```diff
@@ -0,0 +1,100 @@
...
+	k, err := crypto.GenerateSecp256k1Key()
+	if err != nil {
+		t.Fatal(err)
+	}
+	addr, err := crypto.NewOverlayAddress(k.PublicKey, 1, common.HexToHash("0x1").Bytes())
+	if err != nil {
+		t.Fatal(err)
+	}
+	minStake := big.NewInt(1).String()
+	minStakedAmount := big.NewInt(1)
```
vladopajic, line 32: `minStakedAmount` value should come from constant `stakingcontract.MinimumStakeAmount`

openapi/Swarm.yaml
```diff
@@ -1927,6 +1927,33 @@ paths:
         default:
           description: Default response
 
+  "/stake/{address}/{amount}":
+        post:
+          summary: Deposit some amount for staking.
+          description: Be aware, this endpoint creates an on-chain transactions and transfers BZZ from the node's Ethereum account and hence directly manipulates the wallet balance!
```
vladopajic, line 1933: exclamation mark `!` might not be needed in this description

openapi/SwarmDebug.yaml
```diff
@@ -1077,3 +1077,30 @@ paths:
           $ref: "SwarmCommon.yaml#/components/responses/400"
         default:
           description: Default response
+
+  "/stake/{address}/{amount}":
+      post:
+        summary: Deposit some amount for staking.
+        description: Be aware, this endpoint creates an on-chain transactions and transfers BZZ from the node's Ethereum account and hence directly manipulates the wallet balance!
```
vladopajic, line 1084: exclamation mark `!` might not be needed in this description

pkg/api/api_test.go
```diff
@@ -91,6 +92,7 @@ type testServerOptions struct {
 	Feeds              feeds.Factory
 	CORSAllowedOrigins []string
 	PostageContract    postagecontract.Interface
+	StakingContract    stakingcontract.StakingContract
```
vladopajic, line 95: does `StakingContract` variable needs to be exported?

pkg/api/staking_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethersphere/bee/pkg/crypto"
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/jsonhttp/jsonhttptest"
+	"github.com/ethersphere/bee/pkg/staking/stakingcontract"
+	stakingContractMock "github.com/ethersphere/bee/pkg/staking/stakingcontract/mock"
+)
+
+func TestDepositStake(t *testing.T) {
```
vladopajic, line 22: Please, use `t.Parallel()` in tests.

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,191 @@
...
+	//stakeUpdatedTopic = stakingABI.Events["StakeUpdated"].ID
+
+	ErrInvalidStakeAmount = errors.New("invalid stake amount")
+	ErrInsufficientFunds  = errors.New("insufficient token balance")
+	ErrNotImplemented     = errors.New("not implemented")
+
+	depositStakeDescription = "Deposit Stake"
+)
+
+type StakingContract interface {
```
vladopajic, line 39: `stakingcontract.StakingContract` seem very verbose, could it be `staking.Contract`?

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,191 @@
...
+	DepositStake(ctx context.Context, stakedAmount *big.Int, overlay []byte) error
```
vladopajic, line 40: 1) Using `swarm.Address` would be more indicative instead of using `[]byte` for overlay type

2) Since this method shouldn't mutate `stakedAmount` value, it would be better to pass non-pointer value

vladopajic (commented):
added some early comments

pkg/api/staking.go
```diff
@@ -0,0 +1,75 @@
...
+		if errors.Is(err, stakingcontract.ErrInvalidStakeAmount) {
+			s.logger.Debug("deposit stake: invalid stake amount", "error", err)
+			s.logger.Error(nil, "deposit stake: invalid stake amount")
+			jsonhttp.BadRequest(w, "minimum 1 BZZ required for staking")
+			return
+		}
+		if errors.Is(err, stakingcontract.ErrNotImplemented) {
+			s.logger.Debug("deposit stake: not implemented", "error", err)
+			s.logger.Error(nil, "deposit stake: not implemented")
+			jsonhttp.NotImplemented(w, "not implemented")
```
istae, line 61: missing return

pkg/api/staking.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (s *Service) stakingDepositHandler(w http.ResponseWriter, r *http.Request) {
+	overlayAddr, err := hex.DecodeString(mux.Vars(r)["address"])
+	if err != nil {
+		s.logger.Debug("get stake: decode overlay address failed", "string", overlayAddr, "error", err)
+		s.logger.Error(nil, "get stake: decode overlay address failed")
+		jsonhttp.BadRequest(w, "invalid address")
+		return
+	}
+	if len(overlayAddr) == 0 {
```
istae, line 40: what does this mean?

pkg/api/staking.go
```diff
@@ -0,0 +1,75 @@
...
+	if !ok {
+		s.logger.Error(nil, "deposit stake: invalid amount")
+		jsonhttp.BadRequest(w, "invalid staking amount")
+		return
+	}
+	err = s.stakingContract.DepositStake(r.Context(), stakedAmount, overlayAddr)
+	if err != nil {
+		if errors.Is(err, stakingcontract.ErrInvalidStakeAmount) {
+			s.logger.Debug("deposit stake: invalid stake amount", "error", err)
+			s.logger.Error(nil, "deposit stake: invalid stake amount")
```
istae, line 54: passing err instead of nil is better

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,191 @@
...
+	balance, err := s.getBalance(ctx)
+	if err != nil {
+		return err
+	}
+
+	if balance.Cmp(stakedAmount) <= 0 {
+		return ErrInsufficientFunds
+	}
+
+	nonce := make([]byte, 32)
```
istae, line 149: is the nonce some random value or is it nonce used to generate the overlay address? @metacertain @zelig ?

istae (changes requested):

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3341)* status: 0 of 12 files reviewed, 12 unresolved discussions (waiting on @aloknerurkar, @istae, @swarmHaseeb, @vladopajic, and @zelig)

___
*[`openapi/Swarm.yaml` line 1933 at r1](https://reviewable.io/reviews/ethersphere/bee/3341#-NCthyPqHF8bmAAjVRUZ-r1-1933:-NCxuvU8D41l-_tuD1Qh:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/6c174018c5bc2ce0f7adf9b2925815e42eff99ee/openapi/Swarm.yaml#L1933)):*

exclamation mark `!` might not be needed in this description
</blockquote></details>

Done.
___
*[`openapi/SwarmDebug.yaml` line 1084 at r1](https://reviewable.io/reviews/ethersphere/bee/3341#-NCthyPqHF8bmAAjVRU_-r1-1084:-NCxuwCqDaPqc813Ht-_:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/6c174018c5bc2ce0f7adf9b2925815e42eff99ee/openapi/SwarmDebug.yaml#L1084)):*

exclamation mark `!` might not be needed in this description
</blockquote></details>

Done.
___
*[`pkg/api/api_test.go` line 95 at r1](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL4--r1-95:-NCxv6Zv4nnLokRCqmh9:b-6lkf5s) ([raw file](https://github.com/ethersphere/bee/blob/6c174018c5bc2ce0f7adf9b2925815e42eff99ee/pkg/api/api_test.go#L95)):*

does `StakingContract` variable needs to be exported?
</blockquote></details>

yes because otherwise it will not be accessible from Extra ops in node and devnode.
___
*[`pkg/api/staking.go` line 40 at r1](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL42-r1-40:-NCxveZd7cG3SOjbuUY9:bersp9e) ([raw file](https://github.com/ethersphere/bee/blob/6c174018c5bc2ce0f7adf9b2925815e42eff99ee/pkg/api/staking.go#L40)):*

what does this mean?
</blockquote></details>

that's a typo, its removed. I was working on some strategy so that if we get empty address then we default to overlay in service but still working on it. 
___
*[`pkg/api/staking.go` line 54 at r1](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL42-r1-54:-NCxw0H431ecEC4RtPko:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/6c174018c5bc2ce0f7adf9b2925815e42eff99ee/pkg/api/staking.go#L54)):*

passing err instead of nil is better
</blockquote></details>

Done.
___
*[`pkg/api/staking.go` line 61 at r1](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL42-r1-61:-NCxw2te5ymoqvcH-Jiv:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/6c174018c5bc2ce0f7adf9b2925815e42eff99ee/pkg/api/staking.go#L61)):*

missing return
</blockquote></details>

Done.
___
*[`pkg/api/staking.go` line 75 at r1](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL42-r1-75:-NCxw3i_25a6087rAbN5:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/6c174018c5bc2ce0f7adf9b2925815e42eff99ee/pkg/api/staking.go#L75)):*

We are missing a `GET` staking amount endpoint
</blockquote></details>

Done.
___
*[`pkg/api/staking_test.go` line 32 at r1](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL43-r1-32:-NCxwz4zEoCDGG1-b1sd:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/6c174018c5bc2ce0f7adf9b2925815e42eff99ee/pkg/api/staking_test.go#L32)):*

`minStakedAmount` value should come from constant `stakingcontract.MinimumStakeAmount` 
</blockquote></details>

Done.
___
*[`pkg/staking/stakingcontract/contract.go` line 39 at r1](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r1-39:-NCxxHt28f8B1tbhSJx7:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/6c174018c5bc2ce0f7adf9b2925815e42eff99ee/pkg/staking/stakingcontract/contract.go#L39)):*

`stakingcontract.StakingContract` seem very verbose, could it be `staking.Contract`?
</blockquote></details>

Done.
___
*[`pkg/staking/stakingcontract/contract.go` line 40 at r1](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r1-40:-NCy0W0S9hqLc7AEiVz-:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/6c174018c5bc2ce0f7adf9b2925815e42eff99ee/pkg/staking/stakingcontract/contract.go#L40)):*

1) Using `swarm.Address` would be more indicative instead of using `[]byte` for overlay type

2) Since this method shouldn't mutate `stakedAmount` value, it would be better to pass non-pointer value
</blockquote></details>

Done.

pkg/api/staking.go
```diff
@@ -0,0 +1,98 @@
...
+		h.ServeHTTP(w, r)
+	})
+}
+
+type getStakeResponse struct {
+	StakedAmount *big.Int `json:"stakedAmount"`
+}
+
+func (s *Service) stakingDepositHandler(w http.ResponseWriter, r *http.Request) {
+	overlayAddr, err := hex.DecodeString(mux.Vars(r)["address"])
```
vladopajic, line 38: if `swarm.ParseHexAddress` is used instead of `hex.DecodeString` it would give you `swarm.Address` instead of raw bytes

pkg/api/staking.go
```diff
@@ -0,0 +1,98 @@
...
+		s.logger.Debug("deposit stake: deposit failed", "error", err)
+		s.logger.Error(nil, "deposit stake: deposit failed")
+		jsonhttp.InternalServerError(w, "cannot stake")
+		return
+	}
+	jsonhttp.OK(w, nil)
+}
+
+func (s *Service) getStakedAmountHandler(w http.ResponseWriter, r *http.Request) {
+	overlayAddr, err := hex.DecodeString(mux.Vars(r)["address"])
```
vladopajic, line 81: if `swarm.ParseHexAddress` is used instead of `hex.DecodeString` it would give you `swarm.Address` instead of raw bytes

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3341)* status: 0 of 13 files reviewed, 12 unresolved discussions (waiting on @aloknerurkar, @istae, @swarmHaseeb, @vladopajic, and @zelig)

___
*[`pkg/api/staking_test.go` line 22 at r1](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL43-r1-22:-NCy5E-49KMM-byCbTat:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/6c174018c5bc2ce0f7adf9b2925815e42eff99ee/pkg/api/staking_test.go#L22)):*

Please, use `t.Parallel()` in tests.
</blockquote></details>

Done.

pkg/api/staking_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+func TestGetStake(t *testing.T) {
+	k, err := crypto.GenerateSecp256k1Key()
+	if err != nil {
+		t.Fatal(err)
+	}
+	addr, err := crypto.NewOverlayAddress(k.PublicKey, 1, common.HexToHash("0x1").Bytes())
+	if err != nil {
+		t.Fatal(err)
+	}
```
vladopajic, line 127: could you create address from some hardcoded value?

for example:
`
addr := swarm.MustParseHexAddress("f30c0aa7e9e2a0ef4c9b1b750ebfeaeb7c7c24da700bb089da19a46e3677824b")
`

istae (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3341)* status: 0 of 13 files reviewed, 13 unresolved discussions (waiting on @aloknerurkar, @istae, @swarmHaseeb, @vladopajic, and @zelig)

___
*[`pkg/api/staking.go` line 62 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCySzklB4IfjQ2q2NNR:-NCySzkmDw3X8l8jZdtl:b2kvrly) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/api/staking.go#L62)):*
> ```Go
> 		if errors.Is(err, stakingcontract.ErrNotImplemented) {
> 			s.logger.Debug("deposit stake: not implemented", "error", err)
> 			s.logger.Error(nil, "deposit stake: not implemented")
> ```

missing err
___
*[`pkg/api/staking.go` line 68 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCyT0oU4LI7q9PpcYzQ:-NCyT0oU4LI7q9PpcYzR:b2kvrly) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/api/staking.go#L68)):*
> ```Go
> 		if errors.Is(err, stakingcontract.ErrInsufficientFunds) {
> 			s.logger.Debug("deposit stake: out of funds", "error", err)
> 			s.logger.Error(nil, "deposit stake: out of funds")
> ```

missing err
___
*[`pkg/api/staking.go` line 73 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCyT3KJ0xy_tKQf8dhA:-NCyT3KJ0xy_tKQf8dhB:b2kvrly) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/api/staking.go#L73)):*
> ```Go
> 		}
> 		s.logger.Debug("deposit stake: deposit failed", "error", err)
> 		s.logger.Error(nil, "deposit stake: deposit failed")
> ```

missing err
___
*[`pkg/staking/stakingcontract/contract.go` line 96 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCyTZf_-XF7H66SJBs8:-NCyTZf_-XF7H66SJBs9:bc8phl1) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/staking/stakingcontract/contract.go#L96)):*
> ```Go
> func (s *contract) sendDepositStakeTransaction(ctx context.Context, owner common.Address, stakedAmount big.Int, nonce common.Hash) (*types.Receipt, error) {
> 
> 	callData, err := stakingABI.Pack("depositStake", owner, stakedAmount, nonce)
> ```

you might need to pass the address of stakedAmount, this is usually what we do

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,202 @@
...
+	prevStakedAmount, err := s.sendGetStakeTransaction(ctx, overlay.Bytes())
+	if err != nil {
+		return err
+	}
+
+	if prevStakedAmount.Cmp(big.NewInt(0)) == -1 {
+		if stakedAmount.Cmp(MinimumStakeAmount) == -1 {
+			return ErrInvalidStakeAmount
+		}
+	}
```
vladopajic, line 141:
```
prevStakedAmount.Cmp(big.NewInt(0)) == -1
```

this is always false (considering that `prevStakedAmount` is in range of `[0, N)` )?

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,202 @@
...
+
+	balance, err := s.getBalance(ctx)
+	if err != nil {
+		return err
+	}
+
+	if balance.Cmp(&stakedAmount) <= 0 {
+		return ErrInsufficientFunds
+	}
```
vladopajic, line 150: `balance.Cmp(&stakedAmount) <= 0` 

what if user wants to stake entire balance? do we want to forbid this?

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil, err
+	}
+	return abi.ConvertType(results[0], new(big.Int)).(*big.Int), nil
+}
+
+func (s *contract) DepositStake(ctx context.Context, stakedAmount big.Int, overlay swarm.Address) error {
+	prevStakedAmount, err := s.sendGetStakeTransaction(ctx, overlay.Bytes())
+	if err != nil {
+		return err
+	}
```
vladopajic, line 135: calling `GetStake` here would be beneficial because method would return same `ErrGetStakeFailed` in case of error

pkg/api/staking.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		s.logger.Debug("get stake: decode overlay string failed", "string", overlayAddr, "error", err)
+		s.logger.Error(nil, "get stake: decode overlay string failed")
+		jsonhttp.BadRequest(w, "invalid address")
+		return
+	}
+
+	stakedAmount, err := s.stakingContract.GetStake(r.Context(), swarm.NewAddress(overlayAddr))
+	if err != nil {
+		s.logger.Debug("get stake: get staked amount failed", "overlayAddr", hex.EncodeToString(overlayAddr), "error", err)
```
vladopajic, line 91: `swarm.Address` implements `Stringer`, so it can be logged using without `hex.EncodeToString()` - this only applies if address is created early on with `swarm.ParseHexAddress`

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,202 @@
...
+var (
+	MinimumStakeAmount = big.NewInt(1)
+
+	erc20ABI = parseABI(sw3abi.ERC20ABIv0_3_1)
+	//TODO: get ABI for staking contract and replace it below
+	stakingABI = parseABI(sw3abi.ERC20ABIv0_3_1)
+	//TODO: enable below mentioned topic for receiving receipts
+	//stakeUpdatedTopic = stakingABI.Events["StakeUpdated"].ID
+
+	ErrInvalidStakeAmount = errors.New("invalid stake amount")
```
vladopajic, line 33: `ErrInsufficientStakeAmont ` would be more appropriate name for `ErrInvalidStakeAmount` - because first indicates that more should be staked (thus conveying more information), later indicate that something is wrong with amount (it might mislead that amount value is not correctly formatted)

pkg/staking/stakingcontract/contract.go
swarmHaseeb: i was thinking maybe some bzz would be used for transaction (maybe gas fee?) which is why i implemented it this way.

pkg/staking/stakingcontract/contract.go
swarmHaseeb: i have slightly changed the logic in newer code, assumption is that if previous staked amount is zero, then we will check for minimum stake amount for current call and then add stake. If we have staked previously, then we can stake amount less than 1 bzz.

@istae please take a look at it.

pkg/staking/stakingcontract/contract.go
swarmHaseeb: i have modified the code accordingly. Please check

pkg/staking/stakingcontract/contract.go
vladopajic: 1) gas fees should come from native token of network
2) even with current logic we still have problem if gas fees are applied to bzz balance (case: balance is 20bzz, stake amount is 19bzz, fees are 2bzz).

based on both points we should allow stake amount to be up to and including entire balance

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,198 @@
...
+	transactionService     transaction.Service
+	overlayNonce           []byte
+}
+
+func New(
+	owner common.Address,
+	stakingContractAddress common.Address,
+	bzzTokenAddress common.Address,
+	transactionService transaction.Service,
+	nonce []byte,
```
vladopajic, line 58: `nonce` variable should have `common.Hash` type

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,198 @@
...
+	DepositStake(ctx context.Context, stakedAmount big.Int, overlay swarm.Address) error
+	GetStake(ctx context.Context, overlay swarm.Address) (big.Int, error)
+}
+
+type contract struct {
+	owner                  common.Address
+	stakingContractAddress common.Address
+	bzzTokenAddress        common.Address
+	transactionService     transaction.Service
+	overlayNonce           []byte
```
vladopajic, line 50: `overlayNonce` variable should have `common.Hash` type

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,198 @@
...
+
+	receipt, err := s.sendTransaction(ctx, callData, depositStakeDescription)
+	if err != nil {
+		return nil, fmt.Errorf("deposit stake: stakedAmount %d: %w", &stakedAmount, err)
+	}
+
+	return receipt, nil
+}
+
+func (s *contract) sendGetStakeTransaction(ctx context.Context, overlay []byte) (*big.Int, error) {
```
vladopajic, line 111: lets use `swarm.Address` for `overlay` argument (this type should be always convert from/to bytes as early/late as possible in order to have better type safety in interfaces/)

pkg/api/staking.go
```diff
@@ -0,0 +1,97 @@
...
+		s.logger.Error(nil, "get stake: decode overlay address failed")
+		jsonhttp.BadRequest(w, "invalid address")
+		return
+	}
+
+	stakedAmount, ok := big.NewInt(0).SetString(mux.Vars(r)["amount"], 10)
+	if !ok {
+		s.logger.Error(nil, "deposit stake: insufficient amount")
+		jsonhttp.BadRequest(w, "insufficient staking amount")
+		return
```
vladopajic, line 49: previous message `invalid staking amount` was correct at this point, because at this step we are converting string value to integer

pkg/staking/stakingcontract/contract.go
vladopajic: new logic seems better

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,198 @@
...
+	if balance.Cmp(&stakedAmount) < 0 {
+		return ErrInsufficientFunds
+	}
+
+	_, err = s.sendDepositStakeTransaction(ctx, s.owner, stakedAmount, common.BytesToHash(s.overlayNonce))
+	if err != nil {
+		return err
+	}
+	//TODO: verify if we need receipt as well as service for staking
+	//TODO: logic for receipt would be added here
```
vladopajic, line 159: `DepositStake` currently does wait receipt - this method calls `sendDepositStakeTransaction`, which calls `sendTransaction` which calls `WaitForReceipt`.

Please, consider removing this TODO.

swarmHaseeb: umm so this todo was left in case there was some code op that i missed which needed the receipt, ill remove it after my sync session with Esad.

pkg/staking/stakingcontract/contract.go
swarmHaseeb: thankyou

pkg/staking/stakingcontract/contract.go
vladopajic: Alright, generally please consider not checking in any `TODO`s in code because it should always get resolved before code is merged. In case there is technical dept there should be opened issue because that's only way technical depth is tracked.

In this manner please consider removing other `TODO`s from the code (there are some in the beginning of this file)

pkg/api/staking_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	t.Run("ok", func(t *testing.T) {
+		t.Parallel()
+
+		contract := stakingContractMock.New(
+			stakingContractMock.WithDepositStake(func(ctx context.Context, stakedAmount big.Int, overlay swarm.Address) error {
+				if stakedAmount.Cmp(minStakedAmount) == -1 {
+					return stakingcontract.ErrInsufficentStakeAmount
+				}
+				return nil
```
vladopajic, line 41: consider simplifying this mock function by always returning `nil`

pkg/api/staking_test.go
```diff
@@ -0,0 +1,152 @@
...
+	t.Run("with invalid stake amount", func(t *testing.T) {
+		t.Parallel()
+
+		invalidMinStake := big.NewInt(0).String()
+		invalidMinStakedAmount := big.NewInt(1)
+		contract := stakingContractMock.New(
+			stakingContractMock.WithDepositStake(func(ctx context.Context, stakedAmount big.Int, overlay swarm.Address) error {
+				if stakedAmount.Cmp(invalidMinStakedAmount) == -1 {
+					return stakingcontract.ErrInsufficentStakeAmount
+				}
```
vladopajic, line 57: consider simplifying this mock function by always returning `ErrInsufficentStakeAmount`

openapi/Swarm.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 
 info:
-  version: 3.1.0
+  version: 3.2.0
```
aloknerurkar, line 4: Revert version change.

openapi/SwarmDebug.yaml
```diff
@@ -1,6 +1,6 @@
 openapi: 3.0.3
 info:
-  version: 2.1.0
+  version: 2.2.0
```
aloknerurkar, line 3: Revert version change. Also, revert all these changes from the `SwarmDebug` doc. The DebugAPI will be deprecated soon, so no point in adding any new endpoints here.

pkg/api/api.go
```diff
@@ -213,6 +216,7 @@ type ExtraOptions struct {
 	FeedFactory      feeds.Factory
 	Post             postage.Service
 	PostageContract  postagecontract.Interface
+	StakingContract  stakingcontract.Contract
```
aloknerurkar, line 219: Why not keep the naming consistent?

contract repeating twice doesnt look or sound good. It would be better to call the `Contract` interface as `Interface`, similar to the postage package. This way it would read `stakingcontract.Interface` which makes more sense.

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,198 @@
...
+
+	ErrInsufficentStakeAmount = errors.New("insufficient stake amount")
+	ErrInsufficientFunds      = errors.New("insufficient token balance")
+	ErrNotImplemented         = errors.New("not implemented")
+	ErrGetStakeFailed         = errors.New("get stake failed")
+
+	depositStakeDescription = "Deposit Stake"
+)
+
+type Contract interface {
```
aloknerurkar, line 40: Refer to the comment in the `api.go` file. Keep naming consistent. This should be `Interface`. As it is based on the postage contract implementation.

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,198 @@
...
+	DepositStake(ctx context.Context, stakedAmount big.Int, overlay swarm.Address) error
+	GetStake(ctx context.Context, overlay swarm.Address) (big.Int, error)
```
aloknerurkar, line 42: Not sure why the `big.Int` in this signature is not a pointer. You get it back from geth as a pointer, then you return it as object, and then in the API you use the reference again. Instead, this could just be `*big.Int` and we can avoid two type conversions.

pkg/api/staking_test.go
```diff
@@ -0,0 +1,152 @@
...
+	depositStake := func(address string, amount string) string {
+		return fmt.Sprintf("/stake/deposit/%s/%s", address, amount)
+	}
+
+	t.Run("ok", func(t *testing.T) {
+		t.Parallel()
+
+		contract := stakingContractMock.New(
+			stakingContractMock.WithDepositStake(func(ctx context.Context, stakedAmount big.Int, overlay swarm.Address) error {
+				if stakedAmount.Cmp(minStakedAmount) == -1 {
```
aloknerurkar, line 38: So adding logic of the actual contracts in the Mock function to test some logic is not required in API tests. These tests should only be concerned about the code in the `api` package. The stake amount check is not in API so we should not add tests for these things.

So basically you only need to test cases like:
- Send some invalid address and check parsing returns error
- Send some invalid amount and ensure parsing of amount fails
- Return different errors using the mock functions to test the if cases in your error handling scenarios

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,198 @@
...
+		overlayNonce:           nonce,
+	}
+}
+
+func (s *contract) sendTransaction(ctx context.Context, callData []byte, desc string) (*types.Receipt, error) {
+	request := &transaction.TxRequest{
+		To:          &s.stakingContractAddress,
+		Data:        callData,
+		GasPrice:    sctx.GetGasPrice(ctx),
+		GasLimit:    1600000,
```
aloknerurkar, line 74: We should also use the gas limit. Please check the open PR for gas limit and add same support here.

zelig (commented):
> Reviewed 1 of 12 files at r1, 4 of 10 files at r2, 3 of 5 files at r3, 1 of 4 files at r4, 1 of 2 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3341)* status: 12 of 14 files reviewed, 34 unresolved discussions (waiting on @istae, @swarmHaseeb, @vladopajic, and @zelig)

___
*[`pkg/staking/stakingcontract/contract.go` line 141 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r3-141:-ND3wHgU-iAinbrXJaQO:bddhjuq) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/staking/stakingcontract/contract.go#L141)):*
<details><summary><i>Previously, swarmHaseeb (Haseeb Raja) wrote…</i></summary><blockquote>

thankyou
</blockquote></details>

I thought the minimum stake requirement is not part of the staking logic, but only enforced by the lottery. That said the API could warn you that if you stake less than minimum you wont be able to play, but nonetheless should be possible. Or am i wrong about requirements?

pkg/staking/stakingcontract/contract.go
swarmHaseeb: it was a suggestion by @vladopajic to avoid using pointers if the value isn't supposed to change.

pkg/staking/stakingcontract/contract.go
swarmHaseeb: it was also a suggestion to differentiate naming but yeah makes sense.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3341)* status: 6 of 14 files reviewed, 34 unresolved discussions (waiting on @aloknerurkar, @istae, @swarmHaseeb, @vladopajic, and @zelig)

___
*[`openapi/Swarm.yaml` line 4 at r8](https://reviewable.io/reviews/ethersphere/bee/3341#-NCthyPqHF8bmAAjVRUZ-r8-4:-ND6Sc4UE7EpPP7pnzHV:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f38bcdeecbe24c1d8725def3aa88ce47e1edc01c/openapi/Swarm.yaml#L4)):*

Revert version change.
</blockquote></details>

Done.
___
*[`pkg/api/api.go` line 219 at r8](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL3z-r8-219:-ND6Scwb-7dcKRACIAIM:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f38bcdeecbe24c1d8725def3aa88ce47e1edc01c/pkg/api/api.go#L219)):*

Why not keep the naming consistent?

contract repeating twice doesnt look or sound good. It would be better to call the `Contract` interface as `Interface`, similar to the postage package. This way it would read `stakingcontract.Interface` which makes more sense.
</blockquote></details>

Done.
___
*[`pkg/api/staking.go` line 40 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL42-r3-40:-ND6Sei-16kNPLfaekbd:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/api/staking.go#L40)):*

log field name: `"string"` -> `"overlayAddr"`
</blockquote></details>

Done.
___
*[`pkg/api/staking.go` line 62 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCySzklB4IfjQ2q2NNR:-ND6SgOZ2CI7pYBB1aQ-:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/api/staking.go#L62)):*

missing err
</blockquote></details>

Done.
___
*[`pkg/api/staking.go` line 68 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCyT0oU4LI7q9PpcYzQ:-ND6Sh-3Du7_per5siTS:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/api/staking.go#L68)):*

missing err
</blockquote></details>

Done.
___
*[`pkg/api/staking.go` line 73 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCyT3KJ0xy_tKQf8dhA:-ND6ShTF-TgodxNGfhFn:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/api/staking.go#L73)):*

missing err
</blockquote></details>

Done.
___
*[`pkg/api/staking.go` line 83 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL42-r3-83:-ND6ShxEDqY-ddjIvvsB:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/api/staking.go#L83)):*

log field name: `"string"` -> `"overlayAddr"`
</blockquote></details>

Done.
___
*[`pkg/api/staking.go` line 91 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL42-r3-91:-ND6SiOu7LxEt0YEkXyI:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/api/staking.go#L91)):*

`swarm.Address` implements `Stringer`, so it can be logged without `hex.EncodeToString()` - this only applies if address is created early on with `swarm.ParseHexAddress`
</blockquote></details>

Done.
___
*[`pkg/api/staking.go` line 49 at r5](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL42-r5-49:-ND6SfVo9bWxXLAZ6LNE:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/390dc9c6c211c043ddc9e708c285e0e21e11d622/pkg/api/staking.go#L49)):*

previous message `invalid staking amount` was correct at this point, because at this step we are converting string value to integer 
</blockquote></details>

Done.
___
*[`pkg/api/staking_test.go` line 141 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL43-r3-141:-ND6SnSXBouyFLh_MH6A:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/api/staking_test.go#L141)):*

stake amount from response data needs to be checked 
</blockquote></details>

Done.
___
*[`pkg/api/staking_test.go` line 38 at r8](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL43-r8-38:-ND6Sj3x0DB17Wix4sqk:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f38bcdeecbe24c1d8725def3aa88ce47e1edc01c/pkg/api/staking_test.go#L38)):*

So adding logic of the actual contracts in the Mock function to test some logic is not required in API tests. These tests should only be concerned about the code in the `api` package. The stake amount check is not in API so we should not add tests for these things.

So basically you only need to test cases like:
- Send some invalid address and check parsing returns error
- Send some invalid amount and ensure parsing of amount fails
- Return different errors using the mock functions to test the if cases in your error handling scenarios
</blockquote></details>

Done.
___
*[`pkg/api/staking_test.go` line 41 at r8](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL43-r8-41:-ND6Sjfd1q8Kzb9aCiDO:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f38bcdeecbe24c1d8725def3aa88ce47e1edc01c/pkg/api/staking_test.go#L41)):*

consider simplifying this mock function by always returning `nil` (idea is not to test mock functionality but to test that handlers are returning appropriate responses)
</blockquote></details>

Done.
___
*[`pkg/api/staking_test.go` line 57 at r8](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL43-r8-57:-ND6Slrl1Yl3WSwEPXjD:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f38bcdeecbe24c1d8725def3aa88ce47e1edc01c/pkg/api/staking_test.go#L57)):*

consider simplifying this mock function by always returning `ErrInsufficentStakeAmount` 
</blockquote></details>

Done.
___
*[`pkg/staking/stakingcontract/contract.go` line 33 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r3-33:-ND6SoSm9XOv0SRjc7Xv:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/staking/stakingcontract/contract.go#L33)):*

`ErrInsufficientStakeAmont ` would be more appropriate name for `ErrInvalidStakeAmount` - because first indicates that more should be staked (thus conveying more information), later indicate that something is wrong with amount (it might mislead that amount value is not correctly formatted)
</blockquote></details>

Done.
___
*[`pkg/staking/stakingcontract/contract.go` line 96 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCyTZf_-XF7H66SJBs8:-ND6SxfI-2d4PZeDxCBp:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/staking/stakingcontract/contract.go#L96)):*

you might need to pass the address of stakedAmount, this is usually what we do
</blockquote></details>

Done.
___
*[`pkg/staking/stakingcontract/contract.go` line 135 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r3-135:-ND6T026EygTzKusGTmU:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/staking/stakingcontract/contract.go#L135)):*

calling `GetStake` here would be beneficial because method would return same `ErrGetStakeFailed` in case of error
</blockquote></details>

Done.
___
*[`pkg/staking/stakingcontract/contract.go` line 141 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r3-141:-ND6T8Dd3OP2T7JsVFLi:b-ibpoph) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/staking/stakingcontract/contract.go#L141)):*
Thank you for reviewing this. 
> Checks if the amount given as parameter is not less than the minimum stake amount which is 1 BZZ.
This requirement in document makes it look like the api should not allow anything less than 1 BZZ. @istae maybe if you could offer some insight kindly.
I personally agree with what you say, that is that apart from playing, we should be able to stake any amount. 
___
*[`pkg/staking/stakingcontract/contract.go` line 150 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r3-150:-ND6Tppz4LQ7IjnMX0W5:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/staking/stakingcontract/contract.go#L150)):*

1) gas fees should come from native token of network
2) even with current logic we still have problem if gas fees are applied to bzz balance (case: balance is 20bzz, stake amount is 19bzz, fees are 2bzz).

based on both points we should allow stake amount to be up to and including entire balance
</blockquote></details>

Done.
___
*[`pkg/staking/stakingcontract/contract.go` line 50 at r5](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r5-50:-ND6SuQA21PkU43BoKRO:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/390dc9c6c211c043ddc9e708c285e0e21e11d622/pkg/staking/stakingcontract/contract.go#L50)):*

`overlayNonce` variable should have `common.Hash` type
</blockquote></details>

Done.
___
*[`pkg/staking/stakingcontract/contract.go` line 58 at r5](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r5-58:-ND6SvQ9EUz-bxayB64o:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/390dc9c6c211c043ddc9e708c285e0e21e11d622/pkg/staking/stakingcontract/contract.go#L58)):*

`nonce` variable should have `common.Hash` type
</blockquote></details>

Done.
___
*[`pkg/staking/stakingcontract/contract.go` line 111 at r5](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r5-111:-ND6SyN75AMg88qUs7AU:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/390dc9c6c211c043ddc9e708c285e0e21e11d622/pkg/staking/stakingcontract/contract.go#L111)):*

lets use `swarm.Address` for `overlay` argument (this type should be always convert from/to bytes as early/late as possible in order to have better type safety in interfaces/)
</blockquote></details>

Done.
___
*[`pkg/staking/stakingcontract/contract.go` line 40 at r8](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r8-40:-ND6SqdD4aEVAXJZB9BN:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f38bcdeecbe24c1d8725def3aa88ce47e1edc01c/pkg/staking/stakingcontract/contract.go#L40)):*

Refer to the comment in the `api.go` file. Keep naming consistent. This should be `Interface`. As it is based on the postage contract implementation.
</blockquote></details>

Done.
___
*[`openapi/SwarmDebug.yaml` line 3 at r8](https://reviewable.io/reviews/ethersphere/bee/3341#-NCthyPqHF8bmAAjVRU_-r8-3:-ND6TqkiAerDZER76IBm:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f38bcdeecbe24c1d8725def3aa88ce47e1edc01c/openapi/SwarmDebug.yaml#L3)):*

Revert version change. Also, revert all these changes from the `SwarmDebug` doc. The DebugAPI will be deprecated soon, so no point in adding any new endpoints here.
</blockquote></details>

Done.

swarmHaseeb (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3341)* status: 6 of 14 files reviewed, 34 unresolved discussions (waiting on @aloknerurkar, @istae, @swarmHaseeb, @vladopajic, and @zelig)

___
*[`pkg/staking/stakingcontract/contract.go` line 74 at r8](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r8-74:-ND6ityA0mZrYAMdXa_a:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/f38bcdeecbe24c1d8725def3aa88ce47e1edc01c/pkg/staking/stakingcontract/contract.go#L74)):*

We should also use the gas limit. Please check the open PR for gas limit and add same support here.
</blockquote></details>

Done.

pkg/staking/stakingcontract/contract.go
aloknerurkar: Hmm.... Not sure if this makes any sense. Even if it's a pointer or object, once it's returned the caller is the custodian. It doesn't matter either way.

It makes sense to return the pointer in this case as we save 2 type castings.

vladopajic: Pass by value should be preferred way so any changes made inside function does not affect original value. [playground example](https://go.dev/play/p/KI4zhru4S6c) 

What I haven't had in mind is how internals of `big.Int` are implemented - which can't make it immutable if value copied.

> 2 type castings
 
 (kindly drawing attention) getting value from pointer is dereference operation

istae: @swarmHaseeb let pass the pointer, let's not make this a blocker

istae (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3341)* status: 6 of 14 files reviewed, 28 unresolved discussions (waiting on @aloknerurkar, @metacertain, @swarmHaseeb, @vladopajic, and @zelig)

___
*[`pkg/staking/stakingcontract/contract.go` line 141 at r3](https://reviewable.io/reviews/ethersphere/bee/3341#-NCtGPaS2pN7rYiBgL44-r3-141:-ND7JOg33tHmYYa--y3X:b-1guq0q) ([raw file](https://github.com/ethersphere/bee/blob/0da2bde59f3369dfbb40713fc520b1756e77d75a/pkg/staking/stakingcontract/contract.go#L141)):*
<details><summary><i>Previously, swarmHaseeb (Haseeb Raja) wrote…</i></summary><blockquote>

Thank you for reviewing this. 
> Checks if the amount given as parameter is not less than the minimum stake amount which is 1 BZZ.
This requirement in document makes it look like the api should not allow anything less than 1 BZZ. @istae maybe if you could offer some insight kindly.
I personally agree with what you say, that is that apart from playing, we should be able to stake any amount. 
</blockquote></details>

It unlikely the contract will change soon, let's keep the minimum in the API and will update as we go.

pkg/api/staking_test.go
```diff
@@ -0,0 +1,192 @@
...
+			StakingContract: contract,
+		})
+
+		jsonhttptest.Request(t, ts, http.MethodPost, depositStake(addr.String(), minStake), http.StatusOK,
+			jsonhttptest.WithRequestHeader("Gas-Limit", "2000000"),
+		)
+
+		if gasLimit != 2000000 {
+			t.Fatalf("want 2000000, got %d", gasLimit)
+		}
```
vladopajic, line 123: instead of asserting gas limit here, it should be asserted in mock deposit method (right after `sctx.GetGasLimit(ctx)`)

pkg/api/staking.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	stakedAmount, ok := big.NewInt(0).SetString(mux.Vars(r)["amount"], 10)
+	if !ok {
+		s.logger.Error(nil, "deposit stake: invalid amount")
+		jsonhttp.BadRequest(w, "invalid staking amount")
+		return
+	}
+	err = s.stakingContract.DepositStake(ctx, *stakedAmount, overlayAddr)
```
istae, line 74: pass the pointer, not the value of stakedAmount

pkg/api/staking_test.go
```diff
@@ -0,0 +1,192 @@
...
+			jsonhttptest.WithExpectedJSONResponse(&jsonhttp.StatusResponse{Code: http.StatusInternalServerError, Message: "get staked amount failed"}))
+	})
+	t.Run("gas limit header", func(t *testing.T) {
+		t.Parallel()
+
+		var gasLimit uint64
+		contract := stakingContractMock.New(
+			stakingContractMock.WithGetStake(func(ctx context.Context, overlay swarm.Address) (big.Int, error) {
+				gasLimit = sctx.GetGasLimit(ctx)
+				return *big.NewInt(1), nil
```
vladopajic, line 176: could `gasLimit` be asserted here in mock stake method (right after `sctx.GetGasLimit(ctx)`) ?

pkg/api/staking_test.go
```diff
@@ -0,0 +1,192 @@
...
+	})
+
+	t.Run("gas limit header", func(t *testing.T) {
+		t.Parallel()
+
+		var gasLimit uint64
+		contract := stakingContractMock.New(
+			stakingContractMock.WithDepositStake(func(ctx context.Context, stakedAmount big.Int, overlay swarm.Address) error {
+				gasLimit = sctx.GetGasLimit(ctx)
+				return nil
```
vladopajic, line 109: could `gasLimit` be asserted here in mock deposit method (right after `sctx.GetGasLimit(ctx)`) ?

pkg/api/staking.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		if errors.Is(err, stakingcontract.ErrInsufficientStakeAmount) {
+			s.logger.Debug("deposit stake: insufficient stake amount", "error", err)
+			s.logger.Error(nil, "deposit stake: insufficient stake amount")
+			jsonhttp.BadRequest(w, "minimum 1 BZZ required for staking")
+			return
+		}
+		if errors.Is(err, stakingcontract.ErrNotImplemented) {
```
istae, line 82: the contract does not seem to return this error anywhere, please remove

pkg/api/staking.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if limit, ok := r.Header[gasLimitHeader]; ok {
+		l, err := strconv.ParseUint(limit[0], 10, 64)
+		if err != nil {
+			s.logger.Error(err, "deposit stake: bad gas limit")
+			jsonhttp.BadRequest(w, errBadGasLimit)
+			return
+		}
+		ctx = sctx.SetGasLimit(ctx, l)
+	}
```
vladopajic, line 58: could we set `GasLimit` and `GasPrice` to request's context on the middleware level? this way we wouldn't need to do this for each and every handler

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	if receipt.Status == 0 {
+		return nil, transaction.ErrTransactionReverted
+	}
+
+	return receipt, nil
+}
+
+func (s *contract) sendDepositStakeTransaction(ctx context.Context, owner common.Address, stakedAmount big.Int, nonce common.Hash) (*types.Receipt, error) {
```
istae, line 99: stakedAmound should be a pointer

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	return nil
+}
+
+func (s *contract) GetStake(ctx context.Context, overlay swarm.Address) (big.Int, error) {
+	stakedAmount, err := s.sendGetStakeTransaction(ctx, overlay)
+	if err != nil {
+		return *big.NewInt(0), fmt.Errorf("%w:%v", ErrGetStakeFailed, err)
+	}
+	return *stakedAmount, nil
```
istae, line 173: return pointer

istae (changes requested):

pkg/staking/stakingcontract/contract.go
aloknerurkar: > Pass by value should be preferred way so any changes made inside function does not affect original value. [playground example](https://go.dev/play/p/KI4zhru4S6c)

So I was talking about the return value of the `GetStake` function. The returned pointer will never change. Passing it in the DepositStake as a value seems better I agree.

> (kindly drawing attention) getting value from pointer is dereference operation

Yes sorry my bad, dereferencing. Either way, I would prefer not to do unnecessary dereferencing. Having the Deposit call signature with value is okay, but in the API code, the return value of GetStake can be used as is.

aloknerurkar (commented):
Apart from some of the existing comments from others, LGTM!

pkg/api/staking.go
swarmHaseeb: this is for the debug mode. it returns this error.

pkg/api/staking.go
swarmHaseeb: yes we can but since the scope of this task is related to staking agent i think it would be better if we use this and then have a separate PR for that if needed. That being said there is another pr which is focusing on this issue so maybe it would be better to have this suggestion implemented there since that is focused on exact this issue.

pkg/api/staking_test.go
```diff
@@ -171,7 +171,7 @@ func TestGetStake(t *testing.T) {
 		contract := stakingContractMock.New(
 			stakingContractMock.WithGetStake(func(ctx context.Context, overlay swarm.Address) (*big.Int, error) {
 				gasLimit = sctx.GetGasLimit(ctx)
-				assert.Equal(t, gasLimit, uint64(2000000), fmt.Sprintf("wrong gas limit"))
+				assert.Equal(t, gasLimit, uint64(2000000), "wrong gas limit")
```
vladopajic, line 174: order of parameters should be expected value followed by actual - eg  `assert.Equal(t, uint64(2000000), gasLimit, "wrong gas limit")` 

same applies above

pkg/api/staking.go
vladopajic: In that case I would suggest not committing code related to setting gas price and gas limit to context in this PR. And wait for other PR to implement this.

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if receipt.Status == 0 {
+		return nil, transaction.ErrTransactionReverted
+	}
+
+	return receipt, nil
+}
+
+func (s *contract) sendDepositStakeTransaction(ctx context.Context, owner common.Address, stakedAmount *big.Int, nonce common.Hash) (*types.Receipt, error) {
+	callData, err := stakingABI.Pack("depositStake", owner, &stakedAmount, nonce)
```
aloknerurkar, line 100: This should use `stakedAmount` not reference to it now.

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	receipt, err := s.sendTransaction(ctx, callData, depositStakeDescription)
+	if err != nil {
+		return nil, fmt.Errorf("deposit stake: stakedAmount %d: %w", &stakedAmount, err)
```
aloknerurkar, line 107: Same

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,203 @@
...
+	MinimumStakeAmount = big.NewInt(1)
+
+	erc20ABI = parseABI(sw3abi.ERC20ABIv0_3_1)
+	//TODO: get ABI for staking contract and replace it below
+	stakingABI = parseABI(sw3abi.ERC20ABIv0_3_1)
+
+	ErrInsufficientStakeAmount = errors.New("insufficient stake amount")
+	ErrInsufficientFunds       = errors.New("insufficient token balance")
+	ErrNotImplemented          = errors.New("not implemented")
+	ErrGetStakeFailed          = errors.New("get stake failed")
```
aloknerurkar, line 33: I dont think we need this error.

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (s *contract) GetStake(ctx context.Context, overlay swarm.Address) (*big.Int, error) {
+	stakedAmount, err := s.sendGetStakeTransaction(ctx, overlay)
+	if err != nil {
+		return big.NewInt(0), fmt.Errorf("%w:%v", ErrGetStakeFailed, err)
```
aloknerurkar, line 171: Dont see why we need to wrap this error. Instead you can just wrap err like:
`fmt.Errorf("staking contract: failed to get stake: %w", err)`

There is not need for this ErrGetStakeFailed. This has no information.

pkg/api/staking.go
istae: this is fine for now, let's merge as is, and create a PR for the middleware in another PR. 

there are other endpoints that needs the same treatment as well

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,203 @@
...
+		transactionService:     transactionService,
+		overlayNonce:           nonce,
+	}
+}
+
+func (s *contract) sendTransaction(ctx context.Context, callData []byte, desc string) (*types.Receipt, error) {
+	gasLimit := sctx.GetGasLimit(ctx)
+	if gasLimit == 0 {
+		// if gas limit is not set, use the default limit.
+		gasLimit = 9_000_000
```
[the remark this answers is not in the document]
istae: this is 1/3 of the max gas limit, not sure we need this much, maybe lower it to like a 2-3 million

istae (approved):

pkg/staking/stakingcontract/contract.go
```diff
@@ -0,0 +1,202 @@
...
+
+	receipt, err := s.sendTransaction(ctx, callData, depositStakeDescription)
+	if err != nil {
+		return nil, fmt.Errorf("deposit stake: stakedAmount %d: %w", stakedAmount, err)
+	}
+
+	return receipt, nil
+}
+
+func (s *contract) sendGetStakeTransaction(ctx context.Context, overlay swarm.Address) (*big.Int, error) {
```
aloknerurkar, line 112: So this function naming is not correct. This is not a transaction. This is similar to the helper `getBalance`, so it should be called just `getStake`.

aloknerurkar (approved):
Just a minor nitpick. Rest looks good. Either you can fix it before checkin or @istae can fix it in the follow up PR.

PR #3345, opened 2022-09-26
fix: rename swap

### LINKED ISSUE #3263: Rename `swap-endpoint` to better name
`swap-endpoint` represent Bee's blockchain connectivity and nowadays is not linked anymore only to SWAP protocol, but also postage-stamps, in close future to the storage loterry etc. It is quite confusing naming, so I would suggest renaming it to something like `rpc-endpoint`. It could be done with deprecation, eq. supporting for some time (until another breaking release?) both options with priority over the new name, communicating this to users and eventually removing `swap-endpoint`.

umerm-work:
### Checklist

### Description
Rename swap-endpoint to blockchain-rpc-endpoint.Closes  #3263

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3345)

### DIFF
--- a/packaging/bee.yaml
+++ b/packaging/bee.yaml
@@ -60,8 +60,10 @@ password-file: /var/lib/bee/password
 # resolver-options: []
 ## enable swap (default true)
 # swap-enable: true
-## swap ethereum blockchain endpoint (default "")
+## swap blockchain endpoint (default "") [deprecated]
 # swap-endpoint: ""
+## blockchain endpoint (default "")
+# blockchain-rpc-endpoint: ""
 ## swap factory address
 # swap-factory-address: ""
 ## legacy swap factory addresses
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -153,7 +153,7 @@ type Options struct {
 	ResolverConnectionCfgs        []multiresolver.ConnectionConfig
 	RetrievalCaching              bool
 	BootnodeMode                  bool
-	SwapEndpoint                  string
+	BlockchainRpcEndpoint         string
 	SwapFactoryAddress            string
 	SwapLegacyFactoryAddresses    []string
 	SwapInitialDeposit            string
@@ -269,7 +269,7 @@ func NewBee(interrupt chan struct{}, sysInterrupt chan os.Signal, addr string, p
 		erc20Service       erc20.Service
 	)
 
-	chainEnabled := isChainEnabled(o, o.SwapEndpoint, logger)
+	chainEnabled := isChainEnabled(o, o.BlockchainRpcEndpoint, logger)
 
 	var batchStore postage.Storer = new(postage.NoOpBatchStore)
 	var unreserveFn func([]byte, uint8) (uint64, error)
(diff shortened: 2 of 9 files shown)


notanatol (approved):

pkg/node/node.go
```diff
@@ -149,7 +149,8 @@ type Options struct {
 	ResolverConnectionCfgs     []multiresolver.ConnectionConfig
 	RetrievalCaching           bool
 	BootnodeMode               bool
-	SwapEndpoint               string
+	SwapEndpoint               string // deprecated: use rpc endpoint instead
+	RpcEndpoint                string
```
vladopajic, line 153: Could `SwapEndpoint` be removed from `Options` as well?

If we are deprecating this option we should stop using it in all packages other then `cmd` were we need to keep it for backward compatibility.

mrekucci (commented):
Not a good idea, especially renaming it to `rpc-endpoint` as it has a broad meaning. The prefix `swap` is narrower, though I agree it's not the best name either. It would be better to come up with a more specific and suitable name.

mfw78:
There may additionally be the time where the resolver-options​ actually specifies an rpc-endpoint​, and this would be an ethereum mainnet​ endpoint, as opposed to Gnosis Chain.

Perhaps something like settlement-endpoint​, or incentives-endpoint​ to highlight the use of said endpoint?

------- Original Message -------

pkg/node/node.go
umerm-work: Yes, we can.

vladopajic (commented):
LGTM; apart from different opinions on option name

umerm-work:
Now the endpoint name will be `blockchain-rpc-endpoint` as finalized in #3263

mrekucci (changes requested):
This PR contains 106 files, most of which are unrelated. Also, some checks are failing.

umerm-work:
> This PR contains 106 files, most of which are unrelated. Also, some checks are failing.

Done

mrekucci (commented):
There are other places such as homebrew, packaging, etc. where this needs to be reflected. Look at other PRs where a parameter has been added or removed to see all the places.

umerm-work (commented):
Done

> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3345#-:-NEqEBRM4CMMxht6Al1G:bb6un1j)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @aloknerurkar and @vladopajic)

packaging/bee.yaml
```diff
@@ -60,8 +60,10 @@ password-file: /var/lib/bee/password
 # resolver-options: []
 ## enable swap (default true)
 # swap-enable: true
-## swap ethereum blockchain endpoint (default "")
+## swap ethereum blockchain endpoint (default "") [deprecated]
```
mrekucci, line 63: I would also remove "ethereum". The same is in other places.

umerm-work: Done

mrekucci (approved):

AuHau:
@umerm-work can you please also update this in Docs?
https://github.com/ethersphere/bee-docs/blob/master/docs/working-with-bee/configuration.md#--swap-endpoint
(there might be other places where this is mentioned as well)

umerm-work:
> @umerm-work can you please also update this in Docs? https://github.com/ethersphere/bee-docs/blob/master/docs/working-with-bee/configuration.md#--swap-endpoint (there might be other places where this is mentioned as well)

Sure will update it as well.

AuHau:
Thanks, and actually now looking at the code, could there be also a warning in logs that the user is using the old and deprecated flag?

PR #3347, opened 2022-09-26
feat: storage incentives agent

istae:
### Checklist

### TODO
 - [ ] only staked full nodes must run the agent

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)


zelig (changes requested):
> Reviewed 4 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)* status: all files reviewed, 20 unresolved discussions (waiting on @istae)

___
*[`pkg/incentives/incentives.go` line 65 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-sgwq050JFkVCY1ao:-ND-sgwq050JFkVCY1ap:b-21juyw) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L65)):*
> ```Go
> 
> type IncentivesContract interface {
> 	RandomSeedAnchor(context.Context) ([]byte, error)
> ```

this should be called `ReserveSalt`
___
*[`pkg/incentives/incentives.go` line 66 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-s7eg4UYaVMpv_KrD:-ND-s7eg4UYaVMpv_KrE:bm11tom) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L66)):*
> ```Go
> type IncentivesContract interface {
> 	RandomSeedAnchor(context.Context) ([]byte, error)
> 	RandomSeedNeighbourhood(context.Context) ([]byte, error)
> ```

Instead of this there should just be the `isPlaying` call (callable outside the reveal phase)
___
*[`pkg/incentives/incentives.go` line 68 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-s1E6BXzLQPH3zkrE:-ND-s1E6BXzLQPH3zkrF:bvkk14d) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L68)):*
> ```Go
> 	RandomSeedNeighbourhood(context.Context) ([]byte, error)
> 	IsWinner(context.Context) (bool, bool, error)
> 	ClaimWin(context.Context) error
> ```

just Claim
___
*[`pkg/incentives/incentives.go` line 80 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-rw4z9MtaCyqbsdEu:-ND-rw4z9MtaCyqbsdEv:b9yjf9a) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L80)):*
> ```Go
> 	monitor Monitor
> 
> 	incentivesConract IncentivesContract
> ```

incentivesConract -> incentivesContract
___
*[`pkg/incentives/incentives.go` line 119 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-rS3u7TA5kPNbQcQc:-ND-rS3u7TA5kPNbQcQd:b-xobjly) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L119)):*
> ```Go
> }
> 
> func (s *Service) manage(blockTime time.Duration, startBlock, blocksPerRound, blocksPerPhase uint64) {
> ```

manage -> start or run?
___
*[`pkg/incentives/incentives.go` line 144 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-qXOM-JQGkdaPNsXM:-ND-qXOM-JQGkdaPNsXN:b66wfcw) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L144)):*
> ```Go
> 			case <-s.quit:
> 				return
> 			case <-time.After(blockTime):
> ```

does this not leak? I think it is better to use a ticker. 
___
*[`pkg/incentives/incentives.go` line 152 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-r1xO3DqQ3pDsDDn0:-ND-r1xO3DqQ3pDsDDn1:blkukrp) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L152)):*
> ```Go
> 			}
> 
> 			block, err := s.backend.BlockNumber(context.Background())
> ```

do we really neef to query the backend roughly once every block ?
___
*[`pkg/incentives/incentives.go` line 160 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-rgGm1hn5-0FA1tIY:-ND-rgGm1hn5-0FA1tIZ:bgx7h2b) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L160)):*
> ```Go
> 			blocks := (block - startBlock)
> 
> 			round = blocks / blocksPerRound
> ```

this might not be like this. Probs there should be a an accessor for the round number
___
*[`pkg/incentives/incentives.go` line 177 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-pitT5zKWCWUAD6Z2:-ND-pitUDJy_XqspCWiD:b-lcfuhb) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L177)):*
> ```Go
> 				if currentPhase == commit {
> 					select {
> 					case s.newRoundC <- struct{}{}: // on commit phase, we enter a new round, cancel all previus executions
> ```

i dont want to cancel my sampler too early, i can afford it inishihng just leaving me enough time to commit.
the new round really starts with the reveal phase
___
*[`pkg/incentives/incentives.go` line 182 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-o5OO-7_J3dYxpLc6:-ND-o5OO-7_J3dYxpLc7:bbtz5k8) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L182)):*
> ```Go
> 				}
> 				select {
> 				case phaseC <- phase{round: round, phase: currentPhase}:
> ```

this is only going to work if the receiving side is blocking idle, if it is alllowed to run beyond the phase end,  this way some events will be skipped..
in particular sampling should be allowed to go welll into the commit phase and we can stilll win. 
Maybe make this `phaseC` channel buffered with capacity 1 but then that may lead to incorrectly think it is still possible to commit/reveal 
___
*[`pkg/incentives/incentives.go` line 229 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-tuN6DG2VMN0UlZW6:-ND-tuN6DG2VMN0UlZW7:bx37ryl) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L229)):*
> ```Go
> 				if p.round == revealRound { // to claim, previous reveal must've happened in the same round
> 					err = s.attempClaim()
> 					if err != nil {
> ```

specifically if the error here is `ErrSlashed` we should quit the entire process
___
*[`pkg/incentives/incentives.go` line 235 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-nCdw8FWH86RDVZ_1:-ND-nCdx94Yf3PRs_4zn:bn7lc5k) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L235)):*
> ```Go
> 					}
> 				}
> 				storageRadius, sample, err = s.attempMakeSample()
> ```

if this call is syncronous what if it lasts beyond the next commit phase. I think we assume here that this call takes less time than the claim phase no?
___
*[`pkg/incentives/incentives.go` line 255 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-jPoUDKilgd37M6o2:-ND-jPoUDKilgd37M6o3:b-hafxdl) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L255)):*
> ```Go
> }
> 
> func (s *Service) attempClaim() error {
> ```

can we just call this function `claim`?
BTW? /
___
*[`pkg/incentives/incentives.go` line 265 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-uGJz8WLrlDzP2cmZ:-ND-uGJz8WLrlDzP2cm_:b-1kpc5i) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L265)):*
> ```Go
> 	}
> 	if isSlashed {
> 		return ErrSlashed
> ```

this error is not checked upstream but it should be b
___
*[`pkg/incentives/incentives.go` line 279 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-jrs66xbA0ndNjWJU:-ND-jrs66xbA0ndNjWJV:b-90cd7u) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L279)):*
> ```Go
> }
> 
> func (s *Service) attempMakeSample() (uint8, []byte, error) {
> ```

can we just call this function `play`, `participate` or `makeSample`
___
*[`pkg/incentives/incentives.go` line 284 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-imAfEznrBsANmVYD:-ND-imAfEznrBsANmVYE:bemf1r1) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L284)):*
> ```Go
> 	defer cancel()
> 
> 	neighbourhoodRnd, err := s.incentivesConract.RandomSeedNeighbourhood(ctx)
> ```

this is not a seed, lets call this `OverlayAnchor`
___
*[`pkg/incentives/incentives.go` line 289 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-gKvME4sARDkP1wjS:-ND-gKvME4sARDkP1wjT:b-wayrew) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L289)):*
> ```Go
> 	}
> 
> 	storageRadius := s.reserve.GetReserveState().StorageRadius
> ```

is this not available from the depthmonitor?
___
*[`pkg/incentives/incentives.go` line 297 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-gUHR-k4bYMIFepRk:-ND-gUHR-k4bYMIFepRl:bwiy8o9) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L297)):*
> ```Go
> 	s.logger.Info("neighbourhood chosen")
> 
> 	salt, err := s.incentivesConract.RandomSeedAnchor(ctx)
> ```

so this is not a seed and not an anchor, lets call it `ReserveSalt`
___
*[`pkg/incentives/incentives.go` line 320 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-hBML8C5uJvLODsw5:-ND-hBML8C5uJvLODsw6:b-s698do) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L320)):*
> ```Go
> 	}
> 
> 	orc, err := wrapCommit(storageRadius, sample, s.overlay.Bytes(), key)
> ```

for consensual serialisation, the incentive contract provides wrapCommit as a public read-only function 
___
*[`pkg/incentives/incentives.go` line 353 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-i9x198wXsPWmAJR3:-ND-i9x198wXsPWmAJR4:b-8vizc9) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L353)):*
> ```Go
> // wrapCommit concatenates the byte serialisations of all the data needed to apply to
> // the lottery and obfuscates it with a nonce that is to be revealed in the subsequent phase
> // This should be a contract accessor taking storage radius, and reserve sample and overlay and the obfuscater nonce
> ```

it is provided so remove from here

istae (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)* status: 2 of 4 files reviewed, 20 unresolved discussions (waiting on @istae and @zelig)

___
*[`pkg/incentives/incentives.go` line 65 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-sgwq050JFkVCY1ao:-ND0AGFJ0U3AJriu7QgJ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L65)):*

this should be called `ReserveSalt`
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 66 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-s7eg4UYaVMpv_KrD:-ND0AIrwBVaaHjORi0BN:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L66)):*

Instead of this there should just be the `isPlaying` call (callable outside the reveal phase)
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 68 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-s1E6BXzLQPH3zkrE:-ND0AJL2EUFtYMu2pCZf:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L68)):*

just Claim
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 80 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-rw4z9MtaCyqbsdEu:-ND0AJsQ8hEuczSWwto_:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L80)):*

incentivesConract -> incentivesContract
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 119 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-rS3u7TA5kPNbQcQc:-ND0AKjQ3_P7dHp5SJMh:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L119)):*

manage -> start or run?
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 144 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-qXOM-JQGkdaPNsXM:-ND0AM8QDtJ5AZp8y7lG:b-5oltam) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L144)):*

does this not leak? I think it is better to use a ticker. 
</blockquote></details>

no leak :)
___
*[`pkg/incentives/incentives.go` line 152 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-r1xO3DqQ3pDsDDn0:-ND0ARY68ZkCP3hcu4kH:b-tfqyto) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L152)):*

do we really neef to query the backend roughly once every block ?
</blockquote></details>

changed to checking every five blocks, so roughly every 25 seconds
___
*[`pkg/incentives/incentives.go` line 255 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-jPoUDKilgd37M6o2:-ND0AcU11s2fmTSFCinl:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L255)):*

can we just call this function `claim`?
BTW? /
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 279 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-jrs66xbA0ndNjWJU:-ND0Ae4e1y20CKXuXmBg:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L279)):*

can we just call this function `play`, `participate` or `makeSample`
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 284 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-imAfEznrBsANmVYD:-ND0Af0A4KUj2zYqWJXj:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L284)):*

this is not a seed, lets call this `OverlayAnchor`
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 289 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-gKvME4sARDkP1wjS:-ND0Afku3YjE2pREbiAx:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L289)):*

is this not available from the depthmonitor?
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 297 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-gUHR-k4bYMIFepRk:-ND0AgXHEexD-_y2nys0:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L297)):*

so this is not a seed and not an anchor, lets call it `ReserveSalt`
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 320 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-hBML8C5uJvLODsw5:-ND0Ak_N15y4mFkpAUT_:bde3d1c) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L320)):*

for consensual serialisation, the incentive contract provides wrapCommit as a public read-only function 
</blockquote></details>

so completely remove this from this pkg and call the contract for the wrapped commit ? as such we need to a new function to the interface.

istae (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)* status: 2 of 4 files reviewed, 20 unresolved discussions (waiting on @istae and @zelig)

___
*[`pkg/incentives/incentives.go` line 177 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-pitT5zKWCWUAD6Z2:-ND0BBPgCeBiPhhp3z1T:bpl04hb) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L177)):*

i dont want to cancel my sampler too early, i can afford it inishihng just leaving me enough time to commit.
the new round really starts with the reveal phase
</blockquote></details>

in this case, it'd be easier if we had separate block counts for every phase, where commit phase has n number of blocks, reveal has n, and claim something like 2 * n.
___
*[`pkg/incentives/incentives.go` line 229 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-tuN6DG2VMN0UlZW6:-ND0CJ320dD-xBqgefJ0:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L229)):*

specifically if the error here is `ErrSlashed` we should quit the entire process
</blockquote></details>

Done.

istae (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)* status: 2 of 4 files reviewed, 20 unresolved discussions (waiting on @istae and @zelig)

___
*[`pkg/incentives/incentives.go` line 177 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-pitT5zKWCWUAD6Z2:-ND0CVVFDmpDPEmAY48z:bobc8al) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L177)):*

in this case, it'd be easier if we had separate block counts for every phase, where commit phase has n number of blocks, reveal has n, and claim something like 2 * n.
</blockquote></details>

where blocksPerRound is n + n + 2*n

zelig (changes requested):
> Reviewed 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)* status: 3 of 4 files reviewed, 25 unresolved discussions (waiting on @istae and @zelig)

___
*[`pkg/incentives/incentives.go` line 320 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-hBML8C5uJvLODsw5:-ND0lWxR6vMclevrmJsq:bb7s9zc) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L320)):*

so completely remove this from this pkg and call the contract for the wrapped commit ? as such we need to a new function to the interface.
</blockquote></details>

yes
___
*[`pkg/incentives/incentives.go` line 87 at r4](https://reviewable.io/reviews/ethersphere/bee/3347#-ND0lxBeAYth_abvILrj:-ND0lxBeAYth_abvILrk:b-no3zo3) ([raw file](https://github.com/ethersphere/bee/blob/81d06737c591269449d6d0afc76103a5a6b515b4/pkg/incentives/incentives.go#L87)):*
> ```Go
> 
> 	cancelC chan struct{}
> 	stopC   chan struct{}
> ```

why needed?
___
*[`pkg/incentives/incentives.go` line 185 at r4](https://reviewable.io/reviews/ethersphere/bee/3347#-ND0m72VEfp6j9T--qfk:-ND0m72VEfp6j9T--qfl:b-fpf894) ([raw file](https://github.com/ethersphere/bee/blob/81d06737c591269449d6d0afc76103a5a6b515b4/pkg/incentives/incentives.go#L185)):*
> ```Go
> 			if currentPhase != prevPhase {
> 				s.logger.Info("entering phase", "phase", currentPhase.string(), "round", round, "block", block)
> 				select {
> ```

not sure i get this
___
*[`pkg/incentives/incentives.go` line 191 at r4](https://reviewable.io/reviews/ethersphere/bee/3347#-ND0mM62Ews6NBZR0kJp:-ND0mM62Ews6NBZR0kJq:b3kre) ([raw file](https://github.com/ethersphere/bee/blob/81d06737c591269449d6d0afc76103a5a6b515b4/pkg/incentives/incentives.go#L191)):*
> ```Go
> 				select {
> 				case phaseC <- phase{round: round, phase: currentPhase}:
> 				case <-s.quit:
> ```

?
___
*[`pkg/incentives/incentives.go` line 247 at r4](https://reviewable.io/reviews/ethersphere/bee/3347#-ND0kaUq4at8Zm1cNey1:-ND0kaUq4at8Zm1cNey2:be9yox2) ([raw file](https://github.com/ethersphere/bee/blob/81d06737c591269449d6d0afc76103a5a6b515b4/pkg/incentives/incentives.go#L247)):*
> ```Go
> 				if err != nil {
> 					s.logger.Error(err, "make sample")
> 					if errors.Is(err, ErrSlashed) {
> ```

this is returned by `claim` not `play`
___
*[`pkg/incentives/incentives.go` line 249 at r4](https://reviewable.io/reviews/ethersphere/bee/3347#-ND0k4cP-f1poxAPd08q:-ND0k4cP-f1poxAPd08r:b56673) ([raw file](https://github.com/ethersphere/bee/blob/81d06737c591269449d6d0afc76103a5a6b515b4/pkg/incentives/incentives.go#L249)):*
> ```Go
> 					if errors.Is(err, ErrSlashed) {
> 						s.logger.Info("slashed error, quiting incentives agent")
> 						close(s.stopC)
> ```

why not just use `s.Close()`

zelig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)* status: 3 of 4 files reviewed, 15 unresolved discussions (waiting on @istae and @zelig)

___
*[`pkg/incentives/incentives.go` line 152 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-r1xO3DqQ3pDsDDn0:-ND0viqt93MjT7eLF7pt:b-hf0j7r) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L152)):*

changed to checking every five blocks, so roughly every 25 seconds
</blockquote></details>

not sure

istae (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)* status: 3 of 4 files reviewed, 15 unresolved discussions (waiting on @istae and @zelig)

___
*[`pkg/incentives/incentives.go` line 182 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-o5OO-7_J3dYxpLc6:-ND2JERv8cP1kwabQYsT:bdnsh6c) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L182)):*

this is only going to work if the receiving side is blocking idle, if it is alllowed to run beyond the phase end,  this way some events will be skipped..
in particular sampling should be allowed to go welll into the commit phase and we can stilll win. 
Maybe make this `phaseC` channel buffered with capacity 1 but then that may lead to incorrectly think it is still possible to commit/reveal 
</blockquote></details>

we have a context cancellation strategy so if we enter a new phase and the previous phase is not complete, the context is canceled, so the work is terminated. As such, the reads phaseC will not lag behind 
___
*[`pkg/incentives/incentives.go` line 265 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-uGJz8WLrlDzP2cmZ:-ND2GOZF4I6VuvNQz3Nf:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L265)):*

this error is not checked upstream but it should be b
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 320 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-hBML8C5uJvLODsw5:-ND2Gz8V3h3PE9q8TBdu:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L320)):*

yes
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 353 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-i9x198wXsPWmAJR3:-ND2H-bs4uSPgLXJHrIM:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L353)):*

it is provided so remove from here
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 87 at r4](https://reviewable.io/reviews/ethersphere/bee/3347#-ND0lxBeAYth_abvILrj:-ND2Itjb9GRDZG9Hxkmn:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/81d06737c591269449d6d0afc76103a5a6b515b4/pkg/incentives/incentives.go#L87)):*

why needed?
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 185 at r4](https://reviewable.io/reviews/ethersphere/bee/3347#-ND0m72VEfp6j9T--qfk:-ND2IzJc8LEU_GrqS3qv:biuq15c) ([raw file](https://github.com/ethersphere/bee/blob/81d06737c591269449d6d0afc76103a5a6b515b4/pkg/incentives/incentives.go#L185)):*

not sure i get this
</blockquote></details>

this signals the cancellation of the contexts created by the different phases, see `newContext()`
___
*[`pkg/incentives/incentives.go` line 191 at r4](https://reviewable.io/reviews/ethersphere/bee/3347#-ND0mM62Ews6NBZR0kJp:-ND2HNQf1o82dlXlUQIM:byfxib2) ([raw file](https://github.com/ethersphere/bee/blob/81d06737c591269449d6d0afc76103a5a6b515b4/pkg/incentives/incentives.go#L191)):*

?
</blockquote></details>

on shutdown when the quit channel is closed, the next loop might have stopped reading on the phaseC channel 



_Suggestion:_
```Go
on shutdown the reader on this channel might have left already, this is to prevent this goroutine from hanging. 
I added a defer at the top to drain the channel on shutdown, so this `case <-s.quit` has been removed`
```
___
*[`pkg/incentives/incentives.go` line 247 at r4](https://reviewable.io/reviews/ethersphere/bee/3347#-ND0kaUq4at8Zm1cNey1:-ND2H1dcEE88i6My9Dlh:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/81d06737c591269449d6d0afc76103a5a6b515b4/pkg/incentives/incentives.go#L247)):*

this is returned by `claim` not `play`
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 249 at r4](https://reviewable.io/reviews/ethersphere/bee/3347#-ND0k4cP-f1poxAPd08q:-ND2G9jw35sNxDZFp2Dz:b-vzrvwz) ([raw file](https://github.com/ethersphere/bee/blob/81d06737c591269449d6d0afc76103a5a6b515b4/pkg/incentives/incentives.go#L249)):*

why not just use `s.Close()`
</blockquote></details>

close a channel can only be done once, so when the node calls Close() during shutdown again, it will cause a panic

istae (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)* status: 1 of 4 files reviewed, 15 unresolved discussions (waiting on @zelig)

___
*[`pkg/incentives/incentives.go` line 152 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-r1xO3DqQ3pDsDDn0:-ND2aQPFBTYbCOVHtlAI:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L152)):*

not sure 
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 160 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-rgGm1hn5-0FA1tIY:-ND2aR8cEC14dXfQDK5B:brgxmgn) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L160)):*

this might not be like this. Probs there should be a an accessor for the round number
</blockquote></details>

the point of this round tracker is to make sure that before performing an action in the current phase, we confirm that a previous required action has occurred already. For example, for the commit phase, the produced sample must come from previous round. The reveal  obfuscationKey must be produced in the same round number by the commit phase, and the claim for the win must happen in the same reveal round, and so on .... 
___
*[`pkg/incentives/incentives.go` line 235 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-nCdw8FWH86RDVZ_1:-ND2beFv5-Zd7e69sA53:bmq86km) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L235)):*

if this call is syncronous what if it lasts beyond the next commit phase. I think we assume here that this call takes less time than the claim phase no?
</blockquote></details>

yes, we manage this by canceling contexts when there is a phase change

zelig (changes requested):
> Reviewed 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)* status: all files reviewed, 2 unresolved discussions (waiting on @aloknerurkar, @istae, @metacertain, and @swarmHaseeb)

___
*[`pkg/incentives/incentives.go` line 229 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-tuN6DG2VMN0UlZW6:-ND3RQa_1ed5ttexCMaa:b-qafywl) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L229)):*

Done.
</blockquote></details>

whats up with slashing quits the process?
___
*[`pkg/incentives/incentives.go` line 288 at r5](https://reviewable.io/reviews/ethersphere/bee/3347#-ND3S75MCWaG-xuJVWbm:-ND3S75MCWaG-xuJVWbn:b-bs23x1) ([raw file](https://github.com/ethersphere/bee/blob/803f165a683c210e3ec1dd0c9db87653df74311d/pkg/incentives/incentives.go#L288)):*
> ```Go
> 	storageRadius := s.reserve.GetReserveState().StorageRadius
> 
> 	isPlaying, err := s.contract.IsPlaying(ctx, storageRadius)
> ```

whats happening when node is slashed?

pkg/incentives/incentives_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+func (m *mockContract) Commit(context.Context, []byte) error {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	m.commitCalls.Inc()
+	m.previousCall = commitCall
+	return nil
+}
+
+func (m *mockContract) Reveal(context.Context, uint8, []byte, []byte) error {
```
notanatol, line 212: do we want a `m.t.Helper()` on any of these methods?

notanatol (approved):

pkg/incentives/events.go
```diff
@@ -0,0 +1,119 @@
...
+
+const (
+	commit phaseType = iota
+	reveal
+	claim
+	sample
+	sampleEnd
+)
+
+func (p phaseType) string() string {
```
mrekucci, line 22: The signature should be `String` to implement the `fmt.Stringer` interface.

pkg/incentives/events.go
```diff
@@ -0,0 +1,119 @@
...
+	default:
+		return "unknown"
+	}
+}
+
+type events struct {
+	on       map[phaseType][]func(context.Context)
+	ctx      map[phaseType]context.Context
+	cancelF  map[phaseType]context.CancelFunc
+	mtx      sync.Mutex
```
mrekucci, line 43: Would be good to state which struct fields the mutex is guarding.

pkg/incentives/events.go
```diff
@@ -0,0 +1,119 @@
...
+
+import (
+	"context"
+	"sync"
+)
+
+type phaseType int
+
+const (
+	commit phaseType = iota
```
mrekucci, line 15: Maybe the first one can be `none` so you don't have to initialize it to `-1` in the constructor.

pkg/incentives/events.go
```diff
@@ -0,0 +1,119 @@
...
+		return "sample"
+	case sampleEnd:
+		return "sampleEnd"
+	default:
+		return "unknown"
+	}
+}
+
+type events struct {
+	on       map[phaseType][]func(context.Context)
```
mrekucci, line 40: Not sure why the three separate maps are needed. If you create `event` struct with three fields, that will do:
```golang
type event struct {
	on         []func(context.Context)
	ctx        context.Context
	cancelFunc context.CancelFunc
}
```
And then simply `map[phaseType]*event`.

pkg/incentives/incentives.go
```diff
@@ -0,0 +1,371 @@
...
+	ReserveSalt(context.Context) ([]byte, error)
+	IsPlaying(context.Context, uint8) (bool, error)
+	IsWinner(context.Context) (bool, bool, error)
+	Claim(context.Context) error
+	Commit(context.Context, []byte) error
+	Reveal(context.Context, uint8, []byte, []byte) error
+	WrapCommit(uint8, []byte, []byte, []byte) ([]byte, error)
+}
+
+const loggerName = "incentives"
```
mrekucci, line 46: That should be first after the imports with comments on it as it is done in the other packages.

pkg/incentives/incentives.go
```diff
@@ -0,0 +1,371 @@
...
+
+	// when the sample finishes, if we are in the commit phase, run commit
+	sampleEvents.On(sampleEnd, func(ctx context.Context) {
+		if phaseEvents.Last() == commit {
+			commitF(ctx)
+		}
+	})
+
+	// when we enter the commit phase, if the sample is already finished, run commit
+	phaseEvents.On(commit, func(ctx context.Context) {
```
aloknerurkar, line 148: It could happen that both this and the above event would be triggered and we have 2 go routines running the commitF. If we have these separate events, I am not sure if this can be done cleanly.

pkg/incentives/events.go
istae: good suggestion <3

zelig (commented):
> Reviewed 1 of 4 files at r7, 6 of 6 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)* status: all files reviewed, 15 unresolved discussions (waiting on @istae, @metacertain, @mrekucci, and @swarmHaseeb)

___
*[`pkg/storageincentives/agent.go` line 41 at r8](https://reviewable.io/reviews/ethersphere/bee/3347#-NDAghGm1MLeSToDskcz:-NDAghGm1MLeSToDskd-:b-wu6j8m) ([raw file](https://github.com/ethersphere/bee/blob/36b1ccc5f1432b8865c84f243686510d4cd24d5d/pkg/storageincentives/agent.go#L41)):*
> ```Go
> 	ReserveSalt(context.Context) ([]byte, error)
> 	IsPlaying(context.Context, uint8) (bool, error)
> 	IsWinner(context.Context) (bool, bool, error)
> ```

sorry about this but after talking to AbeI changed my mind about this. we decided that until some automated mechanism for restaking is implemented, being  slashed is not so 'relevant' and returning one bool for winning and the error is more  appropriate for the following reasons. 
- the agent needs to run if you are a full node and play whenever you are staked and Isplaying returns true. This way the agent picks up the game no matter how and what tool you staked with or restaked with.
- since we are waiting for the Isplaying to be true, you dont want to switch off the agent even if the node got slashed.  so I suggest getting rid of errslashed and the slashed boolean.
___
*[`pkg/storageincentives/agent.go` line 56 at r8](https://reviewable.io/reviews/ethersphere/bee/3347#-NDAisEu26_wt8vvhKCD:-NDAisEu26_wt8vvhKCE:byr4ffv) ([raw file](https://github.com/ethersphere/bee/blob/36b1ccc5f1432b8865c84f243686510d4cd24d5d/pkg/storageincentives/agent.go#L56)):*
> ```Go
> 	sampler  Sampler
> 	overlay  swarm.Address
> 	slashedC chan struct{}
> ```

see above
___
*[`pkg/storageincentives/agent.go` line 190 at r8](https://reviewable.io/reviews/ethersphere/bee/3347#-NDAjuUB9AttLuNaOMfH:-NDAjuUB9AttLuNaOMfI:b-l47pcl) ([raw file](https://github.com/ethersphere/bee/blob/36b1ccc5f1432b8865c84f243686510d4cd24d5d/pkg/storageincentives/agent.go#L190)):*
> ```Go
> 			err := s.claim(ctx)
> 			if err != nil {
> 				s.logger.Error(err, "claim")
> ```

 slashing logic not needed for now.

pkg/storageincentives/agent.go
```diff
@@ -0,0 +1,368 @@
...
+		if round-1 == sampleRound { // the sample has to come from previous round to be able to commit it
+			obf, err := s.commit(ctx, storageRadius, reserveSample)
+			if err != nil {
+				s.logger.Error(err, "commit")
+			} else {
+				mtx.Lock()
+				obfuscationKey = obf
+				commitRound = round
+				mtx.Unlock()
+				s.logger.Debug("commit phase")
```
aloknerurkar, line 129: "commit phase complete" right? Maybe its better to have entry/exit logs for all the events? It will be useful for debugging.

pkg/storageincentives/agent.go
```diff
@@ -0,0 +1,368 @@
...
+)
+
+const loggerName = "incentives"
+
+type ChainBackend interface {
+	BlockNumber(context.Context) (uint64, error)
+}
+
+type Sampler interface {
+	ReserveSample(context.Context, []byte, uint8) ([]byte, error)
```
aloknerurkar, line 29: I think it will be better to use the actual signature of the sampler. If not we would have to write some adapter later in the node init code.

pkg/storageincentives/incentives_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+			<-wait
+
+			time.Sleep(time.Millisecond * 500)
+
+			if int(contract.commitCalls.Load()) != tc.expectedCalls {
+				t.Fatalf("got %d, want %d", contract.commitCalls.Load(), tc.expectedCalls)
+			}
+
+			if int(contract.revealCalls.Load()) != tc.expectedCalls {
```
aloknerurkar, line 101: The test does not check the sample/sampleEnd events?

Also, it would be good to test both the cases of commit. I think right now only one path is tested where you get to commit and you already have sampleEnd.

pkg/storageincentives/agent.go
istae: good call, this was written before your mege

pkg/storageincentives/agent.go
istae: sure

istae (commented):
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)* status: 4 of 8 files reviewed, 18 unresolved discussions (waiting on @aloknerurkar, @metacertain, @mrekucci, @notanatol, @swarmHaseeb, and @zelig)

___
*[`pkg/incentives/events.go` line 5 at r7](https://reviewable.io/reviews/ethersphere/bee/3347#-ND55cPvQNn8lw6a19vm-r7-5:-NDCfxGA6auIelHbXYAi:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c6d8838319ff9ece586a9c5993ba38f4e238d4c9/pkg/incentives/events.go#L5)):*

Although it is longer, I would prefer we use `storageincentives` here. We have other types of incentives. Also the folder name should be storageincentives.
</blockquote></details>

Done.
___
*[`pkg/incentives/events.go` line 15 at r7](https://reviewable.io/reviews/ethersphere/bee/3347#-ND55cPvQNn8lw6a19vm-r7-15:-NDCfvWI4uaXw93RPmFQ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c6d8838319ff9ece586a9c5993ba38f4e238d4c9/pkg/incentives/events.go#L15)):*

Maybe the first one can be `none` so you don't have to initialize it to `-1` in the constructor.
</blockquote></details>

Done.
___
*[`pkg/incentives/events.go` line 22 at r7](https://reviewable.io/reviews/ethersphere/bee/3347#-ND55cPvQNn8lw6a19vm-r7-22:-NDCftyw-WywoZzrE8EM:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c6d8838319ff9ece586a9c5993ba38f4e238d4c9/pkg/incentives/events.go#L22)):*

The signature should be `String` to implement the `fmt.Stringer` interface.
</blockquote></details>

Done.
___
*[`pkg/incentives/events.go` line 39 at r7](https://reviewable.io/reviews/ethersphere/bee/3347#-ND55cPvQNn8lw6a19vm-r7-39:-NDCfwV220BW4gZ70-7y:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c6d8838319ff9ece586a9c5993ba38f4e238d4c9/pkg/incentives/events.go#L39)):*

Personally, I don't prefer the name `events` here. It makes it slightly confusing as we only maintain the last state locally. This is more like a `state machine`. Each publish is a state transition. This would make it more clear I think.

If you model it as a state machine, you can have the phases as states, as they are fixed. Also, you can define some type for the states to cancel them and ensure they are stopped before transitioning etc. I would prefer this over externally setting this type of context and cancel.
</blockquote></details>

Done.
___
*[`pkg/incentives/events.go` line 43 at r7](https://reviewable.io/reviews/ethersphere/bee/3347#-ND55cPvQNn8lw6a19vm-r7-43:-NDCfvHZ3QNUCjmWLveE:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c6d8838319ff9ece586a9c5993ba38f4e238d4c9/pkg/incentives/events.go#L43)):*

Would be good to state which struct fields the mutex is guarding.
</blockquote></details>

Done.
___
*[`pkg/incentives/events.go` line 80 at r7](https://reviewable.io/reviews/ethersphere/bee/3347#-ND55cPvQNn8lw6a19vm-r7-80:-NDCfwzI0woQz14MD4pq:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c6d8838319ff9ece586a9c5993ba38f4e238d4c9/pkg/incentives/events.go#L80)):*

Before assigning it, should we check if the phase change is valid?

This is why I would prefer this to be like a state machine. The transitions should be checked I think.
</blockquote></details>

Done.
___
*[`pkg/incentives/events.go` line 93 at r7](https://reviewable.io/reviews/ethersphere/bee/3347#-ND55cPvQNn8lw6a19vm-r7-93:-NDCfxzA9-X1LkKPzxWz:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c6d8838319ff9ece586a9c5993ba38f4e238d4c9/pkg/incentives/events.go#L93)):*

This doesn't really guarantee that the handler was stopped. To guarantee we should have some sort of signal from the handler that it is stopped.
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 229 at r1](https://reviewable.io/reviews/ethersphere/bee/3347#-ND-tuN6DG2VMN0UlZW6:-NDCfdTR2W-2x4d12vjR:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/4214a44b815a69b0f28aea37d02cb10bf52a33bc/pkg/incentives/incentives.go#L229)):*

whats up with slashing quits the process?
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 46 at r7](https://reviewable.io/reviews/ethersphere/bee/3347#-NCvtsD-kSUOq0AD8Ccq-r7-46:-NDCfZt8ElaF6zUUt5Xw:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c6d8838319ff9ece586a9c5993ba38f4e238d4c9/pkg/incentives/incentives.go#L46)):*

That should be first after the imports with comments on it as it is done in the other packages.
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives.go` line 148 at r7](https://reviewable.io/reviews/ethersphere/bee/3347#-NCvtsD-kSUOq0AD8Ccq-r7-148:-NDCfbkg--Pcrpp4bQFZ:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c6d8838319ff9ece586a9c5993ba38f4e238d4c9/pkg/incentives/incentives.go#L148)):*

It could happen that both this and the above event would be triggered and we have 2 go routines running the commitF. If we have these separate events, I am not sure if this can be done cleanly.
</blockquote></details>

Done.
___
*[`pkg/incentives/incentives_test.go` line 212 at r7](https://reviewable.io/reviews/ethersphere/bee/3347#-NCzD3w-ZuqiXWZ7lyxB-r7-212:-NDCfsqn3tnDjBvWZ44P:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/c6d8838319ff9ece586a9c5993ba38f4e238d4c9/pkg/incentives/incentives_test.go#L212)):*

do we want a `m.t.Helper()` on any of these methods?
</blockquote></details>

Done.
___
*[`pkg/storageincentives/agent.go` line 41 at r8](https://reviewable.io/reviews/ethersphere/bee/3347#-NDAghGm1MLeSToDskcz:-NDCfzF3AuGMqLkK2A9Q:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/36b1ccc5f1432b8865c84f243686510d4cd24d5d/pkg/storageincentives/agent.go#L41)):*

sorry about this but after talking to AbeI changed my mind about this. we decided that until some automated mechanism for restaking is implemented, being  slashed is not so 'relevant' and returning one bool for winning and the error is more  appropriate for the following reasons. 
- the agent needs to run if you are a full node and play whenever you are staked and Isplaying returns true. This way the agent picks up the game no matter how and what tool you staked with or restaked with.
- since we are waiting for the Isplaying to be true, you dont want to switch off the agent even if the node got slashed.  so I suggest getting rid of errslashed and the slashed boolean.
</blockquote></details>

Done.
___
*[`pkg/storageincentives/agent.go` line 56 at r8](https://reviewable.io/reviews/ethersphere/bee/3347#-NDAisEu26_wt8vvhKCD:-NDCfYwD57_I_DdnyRTM:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/36b1ccc5f1432b8865c84f243686510d4cd24d5d/pkg/storageincentives/agent.go#L56)):*

see above
</blockquote></details>

Done.
___
*[`pkg/storageincentives/agent.go` line 190 at r8](https://reviewable.io/reviews/ethersphere/bee/3347#-NDAjuUB9AttLuNaOMfH:-NDCfcHfDjQy7kLp_FHL:b-896fix) ([raw file](https://github.com/ethersphere/bee/blob/36b1ccc5f1432b8865c84f243686510d4cd24d5d/pkg/storageincentives/agent.go#L190)):*

 slashing logic not needed for now.
</blockquote></details>

Done.

aloknerurkar (approved):

zelig (approved):
> Reviewed 4 of 5 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3347)* status: 7 of 8 files reviewed, 15 unresolved discussions (waiting on @aloknerurkar, @metacertain, @mrekucci, @notanatol, @swarmHaseeb, and @zelig)

PR #3352, opened 2022-09-27
feat(postage): use gas limit for transactions

aloknerurkar:
### Checklist

### Description
Allow user to set gas limit for postage transactions using HTTP Header. This is existing functionality that is being used in the swap package. We will enable it for the postage contract as well.

As part of the changes, I have also updated the OpenAPI spec. Seems like the current spec did not show the `Gas-Price` header parameter. Updated it now to show both the options.

### DIFF
--- a/pkg/postage/postagecontract/contract.go
+++ b/pkg/postage/postagecontract/contract.go
@@ -114,11 +114,16 @@ func (c *postageContract) sendApproveTransaction(ctx context.Context, amount *bi
 }
 
 func (c *postageContract) sendTransaction(ctx context.Context, callData []byte, desc string) (*types.Receipt, error) {
+	gasLimit := sctx.GetGasLimit(ctx)
+	if gasLimit == 0 {
+		// if gas limit is not set, use the default limit.
+		gasLimit = 9_000_000
+	}
 	request := &transaction.TxRequest{
 		To:          &c.postageContractAddress,
 		Data:        callData,
 		GasPrice:    sctx.GetGasPrice(ctx),
-		GasLimit:    1600000,
+		GasLimit:    gasLimit,
 		Value:       big.NewInt(0),
 		Description: desc,
 	}
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -1783,6 +1783,7 @@ paths:
             type: boolean
           required: false
         - $ref: "SwarmCommon.yaml#/components/parameters/GasPriceParameter"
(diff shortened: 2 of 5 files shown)


agazso (approved):

vladopajic:
@aloknerurkar 

Instead of setting gas price and gas limit to context in each handler, we should do it in the middleware level so it becomes available to all handlers at once.

pkg/postage/postagecontract/contract.go
```diff
@@ -114,11 +114,16 @@ func (c *postageContract) sendApproveTransaction(ctx context.Context, amount *bi
 }
 
 func (c *postageContract) sendTransaction(ctx context.Context, callData []byte, desc string) (*types.Receipt, error) {
+	gasLimit := sctx.GetGasLimit(ctx)
+	if gasLimit == 0 {
+		// if gas limit is not set, use the default limit.
+		gasLimit = 9_000_000
+	}
```
vladopajic, line 121: would it be beneficial if new method is added to replace this usage with single call  `sctx.GetGasLimitDefaultTo(ctx, 9_000_000)`

pkg/api/api.go
```diff
@@ -560,6 +561,37 @@ func (s *Service) contentLengthMetricMiddleware() func(h http.Handler) http.Hand
 	}
 }
 
+// gasConfigMiddleware can be used by the APIs that allow block chain transactions to set
+// gas price and gas limit through the HTTP API headers.
+func (s *Service) gasConfigMiddleware(handlerName string) func(h http.Handler) http.Handler {
```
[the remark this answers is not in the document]
aloknerurkar: The handlerName is used for logging. Currently we dont log the exact http endpoint, instead we log using strings. This is better I feel. Keeps the logs consistent.

istae (approved):

istae:
this is great, thanks

pkg/sctx/sctx.go
```diff
@@ -67,6 +67,14 @@ func GetGasLimit(ctx context.Context) uint64 {
 	return 0
 }
 
+func GetGasLimitWithDefault(ctx context.Context, defaultLimit uint64) uint64 {
```
vladopajic, line 70: would having similar method for getting gas price (with default value) would also be helpful?

vladopajic (approved):
LGTM 

note: maybe update PR title to reflect change better eg. `feat(api): set gas price and gas limit in the middleware`

PR #3357, opened 2022-09-28
fix: move upload store

aloknerurkar:
### Checklist

### Description
So the upload store should be in the localstorev2 package and not the storage package. Missed this in the review.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3357)


pkg/localstorev2/internal/upload/uploadstore.go
```diff
@@ -12,8 +12,8 @@ import (
 	"path"
 	"time"
 
-	"github.com/ethersphere/bee/pkg/storagev2"
-	"github.com/ethersphere/bee/pkg/storagev2/internal"
+	"github.com/ethersphere/bee/pkg/localstorev2/internal"
+	storage "github.com/ethersphere/bee/pkg/storagev2"
```
mrekucci, line 16: These import overrides are not necessary. The package name is already `storage`. When we rename the directory it's in (when swapping the current implementation), then we end up with `storage "github.com/ethersphere/bee/pkg/storage"`, which doesn't make sense. The same is below.

mrekucci (approved):
Except for the linter issues, looks good.

pkg/localstorev2/internal/upload/uploadstore.go
aloknerurkar: So this is being added by my editor automatically when I remove it. Seems to me like gofmt is adding this. I think once the name changes, it will not need to add it?

PR #3362, opened 2022-09-30
refactor(depthmonitor): avoid configuration using global vars

vladopajic:
### Checklist

### Description
Refactored `depthmonitor` so that custom `minimumRadius` is set using setter, instead of hacky approach using global vars.

### DIFF
--- a/pkg/topology/depthmonitor/depthmonitor.go
+++ b/pkg/topology/depthmonitor/depthmonitor.go
@@ -19,7 +19,9 @@ const loggerName = "depthmonitor"
 // for the depth monitor wake-up interval.
 const DefaultWakeupInterval = 5 * time.Minute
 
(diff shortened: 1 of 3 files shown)


notanatol (approved):

istae (changes requested):
What is the purpose of this PR? The minimum depth is now a field in the struct but there is no way to configure it with the constructer. 
If we have a field that is not configurable, why make it part of the struct?

istae:
I think having const 'global' variables are fine

vladopajic:
@istae 

It's configurable with setter method exported only for tests files.

Having global `const` values are fine, however having `var` values which are changed via pointers in test in not best practice because it makes potential to cause race condition on that variable.

istae (approved):

PR #3364, opened 2022-09-30
chore: full nodes run the incentives agent

istae:
### Checklist


pkg/node/node.go
```diff
@@ -955,9 +956,15 @@ func NewBee(interrupt chan struct{}, sysInterrupt chan os.Signal, addr string, p
 		return nil, fmt.Errorf("pullsync protocol: %w", err)
 	}
 
+	staking := staking.New(overlayEthAddress, chainCfg.StakingContract, erc20Address, transactionService, common.BytesToHash(nonce))
```
aloknerurkar, line 959: I would prefer to use a different variable name here. This clashes with the package name.

pkg/node/node.go
```diff
@@ -955,9 +956,15 @@ func NewBee(interrupt chan struct{}, sysInterrupt chan os.Signal, addr string, p
...
+
 	if o.FullNodeMode {
 		depthMonitor := depthmonitor.New(kad, pullSyncProtocol, storer, batchStore, logger, warmupTime, depthmonitor.DefaultWakeupInterval)
 		b.depthMonitorCloser = depthMonitor
+
+		// TODO: inject incentives contract, blocksRound, blocksPerPhase
+		agent := storageincentives.New(swarmAddress, chainBackend, logger, depthMonitor, nil, batchStore, storer, o.BlockTime, storageincentives.DefaultBlocksPerRound, storageincentives.DefaultBlocksPerPhase)
```
aloknerurkar, line 966: This PR is complete?

This agent will run only if we are staked right?

Also, this is sort of a chicken and egg problem. We allow user to stake using the API, but the agent is only started on startup. So if user stakes using API later he will need to restart the node right?

Also, this `IsStaked` function that you have implemented, how is this going to be used?

pkg/staking/contract.go
```diff
@@ -156,11 +157,24 @@ func (s *contract) DepositStake(ctx context.Context, stakedAmount *big.Int, over
 func (s *contract) GetStake(ctx context.Context, overlay swarm.Address) (*big.Int, error) {
 	stakedAmount, err := s.getStake(ctx, overlay)
 	if err != nil {
-		return big.NewInt(0), fmt.Errorf("staking contract: failed to get stake: %w", err)
+		return nil, fmt.Errorf("staking contract: failed to get stake: %w", err)
 	}
 	return stakedAmount, nil
 }
 
+func (s *contract) IsStaked(ctx context.Context, overlay swarm.Address) (bool, error) {
```
aloknerurkar, line 165: Where is this used?

istae: not used anywhere atm

pkg/node/node.go
istae: @zelig stated the same problem. The agent runs on full-node if the node has staked or not. So this solves the chicken and egg problem.

istae: see https://github.com/ethersphere/bee/pull/3364#discussion_r984896894

aloknerurkar: Should we use the IsStaked function in the agent worker? Or we dont need it?

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -45,27 +45,27 @@ func New(
 	logger log.Logger,
 	txService transaction.Service,
 	incentivesContractAddress common.Address,
-) *Service {
+) Contract {
```
notanatol, line 48: although this is fine here, but if it's not necessary then generally it's recommended to lean towards returning concrete types

notanatol (approved):

swarmHaseeb (approved):
LGTM

pkg/storageincentives/agent_test.go
```diff
@@ -135,14 +135,19 @@ func createService(
...
+	block          float64
+	limit          float64
+	limitCallback  func()
 }
 
 func (m *mockchainBackend) BlockNumber(context.Context) (uint64, error) {
 
+	if m.incrementSleep != 0 {
+		time.Sleep(m.incrementSleep)
+	}
```
vladopajic, line 149: It would be helpful to explain why `time.Sleep` is added here.

Generally, `time.Sleep` in tests should be always properly commented because their intention could not be straightforward to understand. Assuming that purpose for these is to fix race condition (we want to add delay to test execution so other execution has time to finish), then it would be better to solve this using message passing via channels.

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -24,7 +24,7 @@ var (
 
 const loggerName = "redistributionContract"
 
-type Interface interface {
+type Contract interface {
```
vladopajic, line 27: lovin' this :)

pkg/storageincentives/staking/contract.go
```diff
@@ -34,7 +34,7 @@ var (
 	depositStakeDescription = "Deposit Stake"
 )
 
-type Interface interface {
+type Contract interface {
```
vladopajic, line 37: lovin' this :)

PR #3382, opened 2022-10-04
feat: decrease lower bound

metacertain:
Removing minimum storage radius 4 for depthmonitor

The aim of this PR is to retire an artificial safeguard that kept the storage radius from descending below 4 to avoid a fully connected topology in case of the storage radius decrease erronously never stopped. With the given period of testing this doesn't seem necessary anymore, as well as inhibiting the full scalability of the network


istae (approved):

mrekucci:
It would be good to have a PR description of why this change is needed.

PR #3344, opened 2022-09-26
feat(api): simplify parsing and validation of http input params

mrekucci:
### Checklist

- [x] I have read the [coding guide](https://github.com/ethersphere/bee/blob/master/CODING.md).
- [ ] My change requires a documentation update, and I have done it.
- [x] I have added tests to cover my changes.
- [x] I have filled out the description and linked the related issues.

### Description
Closes #3163 #3165 #3371

### Open API Spec Version Changes (if applicable)
Swarm.yaml: minor
SwarmDebug.yaml: minor
SwarmCommon.yaml: minor

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3344)

PR #3249, opened 2022-09-05
fix: review internal server errors

swarmHaseeb:
### Checklist

- [x] I have read the [coding guide](https://github.com/ethersphere/bee/blob/master/CODING.md)
- [ ] My change requires a documentation update and I have done it
- [x] I have added tests to cover my changes.
- [x] I have filled out the description and linked the related issues

### Description

Please refer to issue for description
#### Motivation and context (Optional)

This pr is focused on improving error handling and writing better response for errors.
### Screenshots (if appropriate):

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3249)

Issue #3166, opened 2022-08-10
Automate/review consistency between HTTP API endpoints and the OpenAPI specification


mrekucci:
In our codebase, it is difficult to maintain consistency between HTTP API endpoints and the OpenAPI specification. Try to think of ways to make it simpler and more automated. Consider generating the OpenAPI specification from the Golang source code, etc... If there is no simple or elegant way to achieve this, manually review that our OpenAPI specification matches one-to-one with our HTTP endpoints.

martinconic:
Duplicate issue , will be fixed in [#4974 ](https://github.com/ethersphere/bee/issues/4974)

Issue #3165, opened 2022-08-10
Review HTTP API responses


mrekucci:
Focus on setting standards for HTTP API endpoints so that the way they respond is consistent across all endpoints and so that consumers of our API are surprised as little as possible. Specifically:

- check if the response messages are meaningful to API consumers
- check if the log messages are consistent with the response messages
- check that response messages and log messages are consistent across all endpoints (maintain the same style)
- check that response messages do not leak unnecessary information, e.g.: errors, etc...
- consider using the logger `WithName(...)` build method to identify handlers
- document the style chosen for log messages and responses

Also schedule a meeting with the JS team to get their feedback on the usability and ergonomics of our HTTP API, as they are big users of it. Process the feedback and engage with it.

Resources:
  - https://cloud.google.com/apis/docs/http
  - https://cloud.google.com/apis/design/errors
  - https://cloud.google.com/blog/products/api-management/restful-api-design-what-about-errors
  - https://cloud.google.com/apigee/resources/ebook/web-api-design-register/index.html

PR #3388, opened 2022-10-05
refactor: parsing and validation

### LINKED ISSUE #3371: Add tests to cover the negative API requests

swarmHaseeb:
### Checklist

### Description

parsing and validation of headers and changes in tests 
w thanks to Vlado
extension of #3344 

### Related Issue (Optional)

related #3163 #3165 #3166 
blocking #3249 #3344 
closes #3371

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -84,7 +84,7 @@ var (
 		Name: "Swarm-Pin",
 		Validate: func(v string, _ *Service) error {
 			v = strings.ToLower(v)
-			if v == "true" || v == "false" || v == "" {
+			if v == boolHeaderSetValue || v == "false" || v == "" {
 				return nil
 			}
 			return fmt.Errorf("expected bool")
@@ -114,7 +114,7 @@ var (
 		Name: "Swarm-Encrypt",
 		Validate: func(v string, _ *Service) error {
 			v = strings.ToLower(v)
-			if v == "true" || v == "false" || v == "" {
+			if v == boolHeaderSetValue || v == "false" || v == "" {
 				return nil
 			}
 			return fmt.Errorf("expected bool")
@@ -132,7 +132,7 @@ var (
 	swarmCollectionHeader = headerDescriptor{
 		Name: "Swarm-Collection",
 		Validate: func(v string, _ *Service) error {
-			if v == "true" || v == "false" || v == "" {
+			if v == boolHeaderSetValue || v == "false" || v == "" {
 				return nil
 			}
 			return fmt.Errorf("expected bool")
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -67,19 +67,146 @@ import (
 // loggerName is the tree path name of the logger for this package.
 const loggerName = "api"
 
-type Headers struct {
-	SwarmPinHeader            string   `map:"Swarm-Pin"`
-	SwarmTagHeader            string   `map:"Swarm-Tag"`
-	SwarmEncryptHeader        string   `map:"Swarm-Encrypt"`
-	SwarmIndexDocumentHeader  string   `map:"Swarm-Index-Document"`
-	SwarmErrorDocumentHeader  string   `map:"Swarm-Error-Document"`
-	SwarmFeedIndexHeader      string   `map:"Swarm-Feed-Index"`
-	SwarmFeedIndexNextHeader  string   `map:"Swarm-Feed-Index-Next"`
-	SwarmCollectionHeader     string   `map:"Swarm-Collection"`
-	SwarmPostageBatchIdHeader []byte   `map:"Swarm-Postage-Batch-Id"`
-	SwarmDeferredUploadHeader string   `map:"Swarm-Deferred-Upload"`
-	GasPrice                  *big.Int `map:"Gas-Price"`
-	GasLimit                  uint64   `map:"Gas-Limit"`
+type headerDescriptor struct {
+	// name of header parameter
+	Name string
+
+	// Validate will validate parameter value or
+	// return error if value is not in correct format.
+	Validate func(v interface{}, s *Service) error
+
+	// If this function is defined it will set value to request context
+	SetToContext func(ctx context.Context, parsedValue interface{}) context.Context
+}
+
+var (
+	swarmPinHeader = headerDescriptor{
+		Name: "Swarm-Pin",
+		Validate: func(v interface{}, _ *Service) error {
+			v = strings.ToLower(v.(string))
+			switch v {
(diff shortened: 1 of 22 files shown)


swarmHaseeb:
Require input for current approach. If this looks good, i can continue on it and adjust code as necessary and needed.
@mrekucci apart from validation approach can you please check if the TODOs were fixed correctly?
@vladopajic @umerm-work for suggestions/reviews

pkg/api/api.go
```diff
@@ -67,6 +67,148 @@ import (
 // loggerName is the tree path name of the logger for this package.
 const loggerName = "api"
 
+type headerDescriptor struct {
+	// name of header parameter
+	Name string
+
+	// Validate will validate parameter value or
+	// return error if value is not in correct format.
+	Validate func(v interface{}, s *Service) error
```
mrekucci, line 76: Why is the `v` interface{} type when headers are always strings?

pkg/api/api.go
```diff
@@ -67,6 +67,148 @@ import (
...
+var (
+	swarmPinHeader = headerDescriptor{
+		Name: "Swarm-Pin",
+		Validate: func(v interface{}, _ *Service) error {
+			v = strings.ToLower(v.(string))
+			switch v {
+			case v.(*bool) == nil || !v.(bool) || v.(bool):
+				return nil
+			default:
+				return fmt.Errorf("%w: expected bool", errInvalidHeaderParams)
```
mrekucci, line 91: What is the purpose of wrapping the `errInvalidHeaderParams`, where it is unwrapped? The same is below.

pkg/api/api.go
```diff
@@ -567,25 +703,32 @@ func (s *Service) contentLengthMetricMiddleware() func(h http.Handler) http.Hand
...
-			}{}
-			if response := s.mapStructure(r.Header, &headers); response != nil {
-				response("invalid header params", logger, w)
-				return
-			}
 			ctx := r.Context()
-			ctx = sctx.SetGasPrice(ctx, headers.GasPrice)
-			ctx = sctx.SetGasLimit(ctx, headers.GasLimit)
 
+			for _, hd := range allHeaderDescriptors {
```
mrekucci, line 714: The iteration, validation, and context setting of all headers must be reflected in each endpoint described in the Open API specification, including a description of what the headers are used for. Since most endpoints do not use all of these headers, this is not the correct approach. The Open API specification should be on pair with the implementation.

pkg/api/util.go
```diff
@@ -146,7 +149,7 @@ var preMapHooks = map[string]func(v string) (string, error){
 //
 // In case of parsing error, a new parseError is returned to the caller.
 // The caller can use the Unwrap method to get the original error.
-func mapStructure(input, output interface{}) (err error) {
+func mapStructure(input, output interface{}, s *Service) (err error) {
```
mrekucci, line 152: This is a generic function, it should not take a `Service` object as a param, but the object should be wrapped in a pre-hook closure.

mrekucci (commented):
I find the validators to be very limiting, not only is it hard to configure or indicate that some headers are required and some are not, but the output errors are very generic, they don't tell why the validation failed, e.g.: odd hexadecimal characters, not a hexadecimal string, etc...

pkg/api/util.go
swarmHaseeb: thankyou, yeah makes sense, on it already

pkg/api/api.go
swarmHaseeb: i had a discussion about this same issue with @vladopajic on Friday or Sat and thanks to him there is an approach we can use to validate that and it also helps with actually enforcing some rules for open api docs and based on this approach i also have found a solution for it will also be sharing that later today probably, maybe we can explore it slightly more and see if it sorts out all those problems we had prev cuz like an extension of this solution will also fix the research part that we were figuring out before.
I am improving the error messages to be more user/dev friendly.

pkg/api/api.go
swarmHaseeb: sorry it's still a draft pr so i'm playing around with different things. Hopefully will have some proper implementation by tomorrow.

pkg/api/api.go
swarmHaseeb: sorry, i was playing around and focused on other part and didn't give it much thought before. Will be fixed in next commit.

pkg/api/api.go
```diff
@@ -67,6 +67,145 @@ import (
...
+}
+
+var (
+	swarmPinHeader = headerDescriptor{
+		Name: "Swarm-Pin",
+		Validate: func(v string, _ *Service) error {
+			v = strings.ToLower(v)
+			if v == boolHeaderSetValue || v != boolHeaderSetValue || v == "" {
+				return nil
+			}
```
vladopajic, line 89: `v != boolHeaderSetValue` is not correct

--- 
condition should be:

v == "" || v == "true" || v == "false" 

---


this validate function seam se to repeating, could we reuse it in other places?

swarmHaseeb: linter was not letting this repeat func go through but maybe i can write another func for it

vladopajic:
```
func validateBool(v string, _*Service) error {
 // todo
 return nil
}


...
 headerDescriptor{
	Name: "Swarm-Pin",
	Validate: validateBool,
}
...

```

would this work?

pkg/api/bzz.go
```diff
@@ -128,10 +128,14 @@ func (s *Service) fileUploadHandler(logger log.Logger, w http.ResponseWriter, r
...
-		// TODO: return standard validation error.
+		valErr := &validationError{
+			Entry: "queries.FileName",
+			Value: queries.FileName,
+			Cause: err,
+		}
 		logger.Debug("bzz upload file: / in prefix not allowed", "file_name", queries.FileName, "error", err)
 		logger.Error(nil, "bzz upload file: / in prefix not allowed", "file_name", queries.FileName)
-		jsonhttp.BadRequest(w, "/ in prefix not allowed")
+		jsonhttp.BadRequest(w, valErr)
```
notanatol, line 138: maybe for `bad request` we should not just pass on the technical error but use a plain string instead

swarmHaseeb: If there is a validation error, if would write a resp like this
```
got json response "{\"code\":400,\"message\":\"invalid query params\",\"reasons\":[{\"field\":\"name\",\"error\":\"want startsnotwith:/\"}]}"
```

notanatol (approved):

vladopajic (approved):

PR #3399, opened 2022-10-07
feat: add consensus time to sampling

aloknerurkar:
### Checklist


notanatol (approved):

pkg/localstore/sampler.go
```diff
@@ -104,6 +111,11 @@ func (db *DB) ReserveSample(ctx context.Context, anchor []byte, storageDepth uin
 					continue
 				}
 
+				if chItem.StoreTimestamp > consensusTime {
```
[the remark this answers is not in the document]
aloknerurkar: Its in Nanosecs. Thanks for pointing this out.

So now we will use the timestamp from the stamp which is also in Nanosecs. This is the same as what abel's PR will use to pass from the agent.

pkg/api/router.go
```diff
@@ -312,12 +312,6 @@ func (s *Service) mountAPI() {
 		web.FinalHandlerFunc(s.healthHandler),
 	))
 
-	handle("/rchash/{depth}/{anchor}", web.ChainHandlers(
```
istae, line 315: can we keep this around for now, it's useful.

istae (changes requested):

pkg/localstore/sampler.go
```diff
@@ -104,6 +112,13 @@ func (db *DB) ReserveSample(ctx context.Context, anchor []byte, storageDepth uin
 					continue
 				}
 
+				// check if the timestamp on the postage stamp is not later than
+				// the consensus time.
+				if time.Duration(binary.BigEndian.Uint64(chItem.Timestamp))*time.Nanosecond > consensusTime {
```
metacertain, line 117: doesnt this lead to an extra 2-4 million conversions to time.Duration?

metacertain: also if time.duration is an alias for int64 isn't this conversion theoretically unsafe?

istae: Storing timestamps with int 64 bits is safe for decades to come, so no danger here.

metacertain (approved):

mrekucci (approved):

istae (approved):

PR #3400, opened 2022-10-07
feat: add consensus time for sampler

metacertain:
Adding a consensus of deterministic time-delimiter for reserve sample inclusion of chunks based on timestamp of postage stamp

### DIFF
--- a/pkg/storageincentives/agent.go
+++ b/pkg/storageincentives/agent.go
@@ -29,7 +29,6 @@ const loggerName = "storageincentives"
 const (
 	DefaultBlocksPerRound = 152
 	DefaultBlocksPerPhase = DefaultBlocksPerRound / 4
-	MaxInt64              = 9223372036854775807
 )
 
 type ChainBackend interface {
@@ -327,7 +326,7 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
 	}
 
 	t := time.Now()
-	block, err := a.backend.BlockNumber(context.Background())
+	block, err := a.backend.BlockNumber(ctx)
 	if err != nil {
 		return 0, nil, err
 	}
--- a/pkg/storageincentives/agent.go
+++ b/pkg/storageincentives/agent.go
@@ -338,7 +338,7 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
 		return 0, nil, err
 	}
 
-	timeLimiterNanoseconds := time.Second * (timeLimiterBlock.Time) / time.Nanosecond
+	timeLimiterNanoseconds := timeLimiterBlock.Time * 1000000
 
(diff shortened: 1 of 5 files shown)


notanatol (approved):

pkg/storageincentives/agent.go
```diff
@@ -321,7 +327,21 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
...
+	}
+
+	previousRoundNumber := (block / a.blocksPerRound) - 1
+
+	timeLimiterBlock, err := a.backend.HeaderByNumber(ctx, new(big.Int).SetUint64(previousRoundNumber*a.blocksPerRound))
+	if err != nil {
+		return 0, nil, err
+	}
+
+	timeLimiter := timeLimiterBlock.Time
```
aloknerurkar, line 342: Do we need to convert this to Nanosecs?

pkg/storageincentives/agent.go
```diff
@@ -27,10 +29,12 @@ const loggerName = "storageincentives"
 const (
 	DefaultBlocksPerRound = 152
 	DefaultBlocksPerPhase = DefaultBlocksPerRound / 4
+	MaxInt64              = 9223372036854775807
```
mrekucci, line 32: The max int64 is already defined in the `math` package as `math.MaxInt64`. On top of that, the constant isn't used.

pkg/storageincentives/agent.go
```diff
@@ -321,7 +327,21 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
 	}
 
 	t := time.Now()
-	sample, err := a.sampler.ReserveSample(ctx, salt, storageRadius, uint64(time.Now().UTC().UnixNano()))
+	block, err := a.backend.BlockNumber(context.Background())
```
mrekucci, line 330: Why `context.Background()` is used instead of existing context?

pkg/storageincentives/agent.go
```diff
@@ -321,7 +326,21 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
...
+	}
+
+	previousRoundNumber := (block / a.blocksPerRound) - 1
+
+	timeLimiterBlock, err := a.backend.HeaderByNumber(ctx, new(big.Int).SetUint64(previousRoundNumber*a.blocksPerRound))
+	if err != nil {
+		return 0, nil, err
+	}
+
+	timeLimiterNanoseconds := timeLimiterBlock.Time * 1000000
```
istae, line 341: assuming timeLimiterBlock.Time is in seconds, the following is more clear
`time.Second * (timeLimiterBlock.Time) / time.Nanosecond`

istae (changes requested):

istae (approved):

Issue #3416, opened 2022-10-10
Flaky test `TestBlocksAfterFlagTimeout`


vladopajic:
Test `TestBlocksAfterFlagTimeout` seems to be flaky [CI attempt #1](https://github.com/ethersphere/bee/actions/runs/3219554211/jobs/5265069741#step:12:10440)

```
     blocker_test.go:75: address should be blocked
--- FAIL: TestBlocksAfterFlagTimeout (0.51s)
```

PR #3417, opened 2022-10-10
fix(blocker): fix flaky tests

vladopajic:
### Checklist

### Description
Closes:  #3416

### DIFF
--- a/pkg/blocker/blocker_test.go
+++ b/pkg/blocker/blocker_test.go
@@ -6,7 +6,6 @@ package blocker_test
 
 import (
 	"os"
-	"sync"
 	"testing"
 	"time"
 
@@ -18,12 +17,14 @@ import (
 	"github.com/ethersphere/bee/pkg/blocker"
 )
 
-var (
+const (
 	flagTime  = time.Millisecond * 25
-	checkTime = time.Millisecond * 100
 	blockTime = time.Second
-	addr      = test.RandomAddress()
-	logger    = log.Noop
+)
+
+var (
+	addr   = test.RandomAddress()
+	logger = log.Noop
 )
 
 func TestMain(m *testing.M) {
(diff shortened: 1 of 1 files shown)


pkg/blocker/blocker_test.go
```diff
@@ -18,144 +17,116 @@ import (
...
+	mock := mockBlockLister(func(a swarm.Address, d time.Duration, r string) error {
+		t.Fatalf("address should not be blocked")
+
+		return nil
+	})
+
+	b := blocker.New(mock, flagTime, blockTime, time.Millisecond, nil, logger)
 	defer b.Close()
 
+	// Flagging address shouldn't block it imidietly
```
notanatol, line 119: comment typo `immediately`

notanatol:
is there a way to confirm that these fixes remove flakiness? Or do we just run them with a large `-count` option?

vladopajic:
@notanatol 

All fixes which involve `time.Sleep` as tool for waiting for some async background task (aka some goroutine), are almost never real solution. All they do is increase chance for this task to happen sooner then goroutine being blocked - but this is never guaranteed.

Imagine having two goroutines A (currently blocked by `time.Sleep`) and B (doing some task and result of this task is expected by A). In normal circumstances B might finish way sooner then A attempts to read B's result (A block interval > B work interval). Problem arises when B's goroutine gets suspended (for whichever reason); in which case B's work interval gets increased (sometimes undesirable more then A's block interval).

Afaik there is not way to tests this as all goroutines scheduling is done by go runtime and no mechanism is exposed to the user. This wouldn't be surprise because it was never expected of user to write synchronization in such way. Instead, B should communicated back (via channels) result to the A.

Only thing we can do is increase likelihood for suspending any goroutine (by making tests parallel with high count)

notanatol (approved):

notanatol:
> @notanatol
> 
> All fixes which involve `time.Sleep` as tool for waiting for some async background task (aka some goroutine), are almost never real solution. All they do is increase chance for this task to happen sooner then goroutine being blocked - but this is never guaranteed.
> 
> Imagine having two goroutines A (currently blocked by `time.Sleep`) and B (doing some task and result of this task is expected by A). In normal circumstances B might finish way sooner then A attempts to read B's result (A block interval > B work interval). Problem arises when B's goroutine gets suspended (for whichever reason); in which case B's work interval gets increased (sometimes undesirable more then A's block interval).
> 
> Afaik there is not way to tests this as all goroutines scheduling is done by go runtime and no mechanism is exposed to the user. This wouldn't be surprise because it was never expected of user to write synchronization in such way. Instead, B should communicated back (via channels) result to the A.
> 
> Only thing we can do is increase likelihood for suspending any goroutine (by making tests parallel with high count)

yeah that's fair enough

Issue #3402, opened 2022-10-07
Migration helpers for store item


aloknerurkar:
Implement data migration helpers (from [proposal](https://hackmd.io/@ethswarm/H1_sxXOMs)).

vladopajic:
Implemented in #3421.

PR #3421, opened 2022-10-12
feat(migration): store migration helpers

vladopajic:
### Checklist

### Description
Closes:  #3402

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3421)

### DIFF
--- a/pkg/storagev2/migration/index.go
+++ b/pkg/storagev2/migration/index.go
@@ -1,3 +1,7 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 package migration
 
 import (
--- a/pkg/storagev2/migration/index.go
+++ b/pkg/storagev2/migration/index.go
@@ -1,0 +1,124 @@
+package migration
+
+import (
+	storage "github.com/ethersphere/bee/pkg/storagev2"
+)
+
+type (
+	// ItemDeleteFn is callback function called in migration step
+	// to check if Item should be removed in this step.
+	ItemDeleteFn func(storage.Item) (delete bool)
+
+	// ItemUpdateFn is callback function called in migration step
+	// to check if Item should be updated in this step.
+	ItemUpdateFn func(storage.Item) (updateditem storage.Item, hasChanged bool)
+)
+
+// WithItemDeleteFn return option with ItemDeleteFn set.
+func WithItemDeleteFn(fn ItemDeleteFn) option {
+	return func(o *options) {
+		o.deleteFn = fn
+	}
+}
+
+// WithItemUpdaterFn return option with ItemUpdateFn set.
+func WithItemUpdaterFn(fn ItemUpdateFn) option {
+	return func(o *options) {
+		o.updateFn = fn
+	}
+}
+
+type option func(*options)
+
+type options struct {
+	deleteFn ItemDeleteFn
+	updateFn ItemUpdateFn
+}
+
+func newOptions() *options {
+	return &options{
+		deleteFn: func(storage.Item) bool { return false },
+		updateFn: func(i storage.Item) (storage.Item, bool) { return i, false },
+	}
+}
+
+func (o *options) applyAll(opts []option) {
+	for _, opt := range opts {
+		opt(o)
+	}
+}
+
+// NewStepOnIndex creates new migration step with update and/or delete operation.
+// Migration will iterate on all elements selected by query and delete or update items
+// based on supplied callback functions.
+func NewStepOnIndex(query storage.Query, opts ...option) StepFn {
+	o := newOptions()
+	o.applyAll(opts)
+
+	return func(s storage.Store) error {
+		itemsForDelete := make([]storage.Item, 0)
+		itemsForUpdate := make([]storage.Item, 0)
+
+		err := s.Iterate(query, func(r storage.Result) (bool, error) {
+			item := r.Entry
+
+			if delete := o.deleteFn(item); delete {
+				itemsForDelete = append(itemsForDelete, newKey(item.ID(), item.Namespace()))
+				return false, nil
+			}
+
+			oldID := item.ID()
+			if updatedItem, hadChanged := o.updateFn(item); hadChanged {
+				// If ID of item has changed - we will need to remove old
+				// item from store, because saving new item will not overwrite old
+				if oldID != updatedItem.ID() {
+					// We can't append item(r.Entry) nor updatedItem to the slice
+					// because IDs have change (we need old ID). So we need to create new item which will mock
+					// storage.Item interface by returning Key of item we want to remove
+					itemsForDelete = append(itemsForDelete, newKey(oldID, item.Namespace()))
+				}
+
+				itemsForUpdate = append(itemsForUpdate, updatedItem)
+
+				return false, nil
+			}
+
+			return false, nil
+		})
+		if err != nil {
+			return err
+		}
+
+		for _, key := range itemsForDelete {
+			if err := s.Delete(key); err != nil {
+				return err
+			}
+		}
+
+		for _, i := range itemsForUpdate {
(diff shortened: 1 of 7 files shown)


pkg/storagev2/migration/index.go
```diff
@@ -0,0 +1,124 @@
...
+func (o *options) applyAll(opts []option) {
+	for _, opt := range opts {
+		opt(o)
+	}
+}
+
+// NewStepOnIndex creates new migration step with update and/or delete operation.
+// Migration will iterate on all elements selected by query and delete or update items
+// based on supplied callback functions.
+func NewStepOnIndex(query storage.Query, opts ...option) StepFn {
```
aloknerurkar, line 54: So the Option type needs to be exported right? Users need to provide these, so a private type cannot work.

pkg/storagev2/migration/index.go
```diff
@@ -0,0 +1,124 @@
...
+	o := newOptions()
+	o.applyAll(opts)
+
+	return func(s storage.Store) error {
+		itemsForDelete := make([]storage.Item, 0)
+		itemsForUpdate := make([]storage.Item, 0)
+
+		err := s.Iterate(query, func(r storage.Result) (bool, error) {
```
aloknerurkar, line 62: We will need to do this in batches. We can potentially have millions of entries which can cause out of memory here.

Check [this](https://github.com/ethersphere/bee/blob/localstorev2.0/pkg/localstorev2/internal/pinning/pinning.go#L248) for example.

pkg/storagev2/migration/index.go
aloknerurkar, line 54: One more problem here is as the StepFn is dependent on the Query, we will not be able to do multiple prefixes here right?

Either we need to have multiple StepFns for each step or we need to have a way where we can configure the helpers for different queries. I would prefer we have multiple StepFns.

aloknerurkar (changes requested):

pkg/storagev2/migration/index.go
vladopajic: `option` doesn't need to be exported, because it is not intended to use `option` type, rather user should create it's instances via factory methods like ` WithItemDeleteFn` or `WithItemUpdaterFn`.

pkg/storagev2/migration/index.go
vladopajic: I am not sure if this example could be used here... In this example iteration continues while there are elements selected by query, so it is possible to stop it at any time and start iteration again.

In case of code from PR, iteration should continue where previous has stopped. Is there a way to achieve such behaviour?

aloknerurkar: So in this case, I would suggest adding support to provide multiple step functions for a particular migration step. This way users can configure different functions for different queries.

aloknerurkar: But then each step is like a version. We dont need to bump the version for every change in a particular item. There might be multiple of these for a particular version bump.

aloknerurkar: > I am not sure if this example could be used here... In this example iteration continues while there are elements selected by query, so it is possible to stop it at any time and start iteration again.
> 
> In case of code from PR, iteration should continue where previous has stopped. Is there a way to achieve such behaviour?

So what we need to do is maintain a skip count. In the IterateFn we can skip entries till this count and then continue. Also, if you delete entries we need not increment the skip count as the next iteration will not show these.

There is no other way than to do this as we can run into out of memory here if the store is few GBs.

pkg/storagev2/migration/steps_chain_test.go
```diff
@@ -0,0 +1,65 @@
...
+	populateStore(t, store, populateItemsCount)
+
+	stepsFn := make([]migration.StepFn, 0)
+
+	// Create 10 step functions where each would remove single element, having value [0-10)
+	for i := 0; i < 10; i++ {
+		valForRemoval := i
+		var stepFn migration.StepFn
+
+		// We create two types of step functions, each should have eqvivalent
```
mrekucci, line 29: `eqvivalent` -> `equivalent`.

pkg/storagev2/migration/index.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+type (
+	// ItemDeleteFn is callback function called in migration step
+	// to check if Item should be removed in this step.
+	ItemDeleteFn func(storage.Item) (delete bool)
+
+	// ItemUpdateFn is callback function called in migration step
+	// to check if Item should be updated in this step.
+	ItemUpdateFn func(storage.Item) (updateditem storage.Item, hasChanged bool)
```
mrekucci, line 18: `updateditem` -> `updatedItem`.

pkg/storagev2/migration/index.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+type option func(*options)
+
+type options struct {
+	deleteFn ItemDeleteFn
+	updateFn ItemUpdateFn
+}
+
+func newOptions() *options {
```
mrekucci, line 42: The name is a little bit misleading, it should be something like `defaultOptions` or `noopOptions`. Also, the procedure that this function wraps is not complex or used in other places, so it can be nested. The same goes for the `applyAll` function.

pkg/storagev2/migration/index.go
```diff
@@ -0,0 +1,128 @@
...
+
+// NewStepOnIndex creates new migration step with update and/or delete operation.
+// Migration will iterate on all elements selected by query and delete or update items
+// based on supplied callback functions.
+func NewStepOnIndex(query storage.Query, opts ...option) StepFn {
+	o := newOptions()
+	o.applyAll(opts)
+
+	return func(s storage.Store) error {
+		itemsForDelete := make([]storage.Item, 0)
```
mrekucci, line 63: You can avoid the zero length by declaring nil slice instead, which is the [preferred method](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices).

pkg/storagev2/migration/index.go
```diff
@@ -0,0 +1,128 @@
...
+		itemsForUpdate := make([]storage.Item, 0)
+
+		err := s.Iterate(query, func(r storage.Result) (bool, error) {
+			item := r.Entry
+
+			if delete := o.deleteFn(item); delete {
```
mrekucci, line 69: `delete` is a reserved keyword.

pkg/storagev2/migration/index.go
vladopajic: More often then not, types created with `func NewType() Type` or just `func New() Type` methods have some default values set (depending on the use case). That is the reason for having these functions (to create type with default values), because creating zero value types is usually done via semantics `Type{}` or `new(Type)`. Some devs like having `New` funcs as constructor for zero values as well, so in that case `New` func creates type with zero value or default values, whichever suits more usage of particular type.

However, I have updated name to `defaultOptions` as suggested.

----

Could you explain other suggestion?

mrekucci: The other suggestions have become irrelevant because this [CL](https://github.com/ethersphere/bee/pull/3421/commits/cbd92d3cc30b36b09bc0a358e492272708e19613) made after this comment contains new functions that use those functions.

pkg/storagev2/migration/index_test.go
```diff
@@ -0,0 +1,387 @@
...
+	})
+}
+
+func TestStepIndex_BatchSize(t *testing.T) {
+	t.Parallel()
+
+	const populateItemsCount = 128
+	for i := 1; i <= 2*populateItemsCount; i <<= 1 {
+		i := i
+		t.Run(fmt.Sprintf("callback called once per item, with batch size: %d", i), func(t *testing.T) {
```
aloknerurkar, line 187: Nit: This `,` can be removed. The test will add `_` for space anyway.

aloknerurkar (approved):

PR #3424, opened 2022-10-12
feat: migration initiation

### LINKED ISSUE #3401: Migration function implementation
Implement core functionality for data migration (from [proposal](https://hackmd.io/@ethswarm/H1_sxXOMs)).

umerm-work:
### Checklist

### Description
Closes #3401 

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3424)

### DIFF
--- a/pkg/storagev2/migration/export_test.go
+++ b/pkg/storagev2/migration/export_test.go
@@ -1,0 +1,9 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package migration
+
+var (
+	ErrStorageVersionItemUnmarshalInvalidSize = errStorageVersionItemUnmarshalInvalidSize
+)
--- a/pkg/storagev2/migration/export_test.go
+++ b/pkg/storagev2/migration/export_test.go
@@ -1,0 +1,5 @@
+package migration
+
+type (
+	StorageVersionItem = storageVersionItem
+)
--- a/pkg/storagev2/migration/export_test.go
+++ b/pkg/storagev2/migration/export_test.go
@@ -1,5 +1,0 @@
-package migration
-
-type (
-	StorageVersionItem = storageVersionItem
-)
--- a/pkg/storagev2/migration/migration.go
+++ b/pkg/storagev2/migration/migration.go
@@ -4,10 +4,8 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"sort"
-	"unsafe"
-
 	storage "github.com/ethersphere/bee/pkg/storagev2"
+	"sort"
 )
 
 type Version = uint64
@@ -75,7 +73,7 @@ type storageVersionItem struct {
 	Version uint64
 }
 
-const storageVersionItemSize = unsafe.Sizeof(&storageVersionItem{})
+const storageVersionItemSize = 8
 
 func (s *storageVersionItem) ID() string {
 	return "storage_version"
--- a/pkg/storagev2/migration/migration.go
+++ b/pkg/storagev2/migration/migration.go
@@ -16,7 +16,7 @@ func Migrate(s storage.Store, sm StepsMap) error {
 		return err
 	}
 	for key, step := range sm {
-		if key == version+1 {
+		if key == version+1 || key > version {
 			err := step(s)
 			if err != nil {
 				return err
--- a/pkg/storagev2/migration/migration.go
+++ b/pkg/storagev2/migration/migration.go
@@ -3,6 +3,8 @@ package migration
 import (
 	"encoding/binary"
 	"errors"
+	"fmt"
+	"sort"
 	"unsafe"
 
 	storage "github.com/ethersphere/bee/pkg/storagev2"
@@ -33,20 +35,39 @@ func Migrate(s storage.Store, sm StepsMap) error {
 				return err
 			}
 		} else {
-			// there is no next version defiend
+			// there is no next version defined
 			// so we stop here
 			return nil
 		}
 	}
 }
 
+// ValidateVersions checks versions if they are in order n (where n min version value), n+1, n+2, n+3... (all values are increasing orders)
 func ValidateVersions(sm StepsMap) error {
-	// TODO
-	// all versions should be in order n (where n min version value), n+1, n+2, n+3... (all values are increasing orders)
-	// 5,6,7,8 ok
-	// 5,6,7,8,10 nok
+	keys := make([]int, len(sm))
+	i := 0
+	for k := range sm {
+		keys[i] = int(k)
+		i++
+	}
+	sort.Ints(keys)
+	missing := FindMissingNumbers(keys)
 
+	if len(missing) > 0 {
+		return fmt.Errorf("missing versions: %v", missing)
+	}
 	return nil
+}
+
+// FindMissingNumbers finds missing numbers in a slice of numbers
+func FindMissingNumbers(numbers []int) []int {
+	var missing []int
+	for i := 0; i < len(numbers)-1; i++ {
+		if numbers[i+1]-numbers[i] != 1 {
+			missing = append(missing, numbers[i]+1)
+		}
+	}
+	return missing
 }
 
 type storageVersionItem struct {
--- a/pkg/storagev2/migration/migration.go
+++ b/pkg/storagev2/migration/migration.go
@@ -36,7 +36,7 @@ type storageVersionItem struct {
 
 const storageVersionItemSize = unsafe.Sizeof(&storageVersionItem{})
 
-func (s storageVersionItem) ID() string {
+func (s *storageVersionItem) ID() string {
 	return "storage_version"
 }
 
@@ -44,23 +44,23 @@ func (s storageVersionItem) Namespace() string {
 	return "migration"
 }
 
-func (s storageVersionItem) Marshal() ([]byte, error) {
+func (s *storageVersionItem) Marshal() ([]byte, error) {
 	buf := make([]byte, storageVersionItemSize)
 	binary.LittleEndian.PutUint64(buf, s.Version)
 	return buf, nil
 }
 
-func (s storageVersionItem) Unmarshal(bytes []byte) error {
+func (s *storageVersionItem) Unmarshal(bytes []byte) error {
 	ni := storageVersionItem{}
 	ni.Version = binary.LittleEndian.Uint64(bytes)
-	s = ni
+	s = &ni
 	return nil
 }
 
 func GetVersion(s storage.Store) (Version, error) {
 
 	item := storageVersionItem{}
-	err := s.Get(item)
+	err := s.Get(&item)
 	if err != nil {
 		return 0, err
 	}
--- a/pkg/storagev2/migration/migration.go
+++ b/pkg/storagev2/migration/migration.go
@@ -13,22 +13,39 @@ type StepFn func(storage.Store) error
 type StepsMap = map[Version]StepFn
 
 func Migrate(s storage.Store, sm StepsMap) error {
-	version, err := GetVersion(s)
+	if err := ValidateVersions(sm); err != nil {
+		return err
+	}
+
+	currentVersion, err := GetVersion(s)
 	if err != nil {
 		return err
 	}
-	for key, step := range sm {
-		if key == version+1 || key > version {
-			err := step(s)
+
+	for nextVersion := currentVersion + 1; ; nextVersion++ {
+		if stepFn, ok := sm[nextVersion]; ok {
+			err := stepFn(s)
 			if err != nil {
 				return err
 			}
-			err = s.Put(&storageVersionItem{Version: key})
+			err = SetVersion(s, nextVersion)
 			if err != nil {
 				return err
 			}
+		} else {
+			// there is no next version defiend
+			// so we stop here
+			return nil
 		}
 	}
+}
+
+func ValidateVersions(sm StepsMap) error {
+	// TODO
+	// all versions should be in order n (where n min version value), n+1, n+2, n+3... (all values are increasing orders)
+	// 5,6,7,8 ok
+	// 5,6,7,8,10 nok
+
 	return nil
 }
 
--- a/pkg/storagev2/migration/migration.go
+++ b/pkg/storagev2/migration/migration.go
@@ -17,8 +17,11 @@ func Migrate(s storage.Store, sm StepsMap) error {
 	}
 	for key, step := range sm {
 		if key == version+1 {
-			step(s)
-			err := s.Put(&storageVersionItem{Version: key})
+			err := step(s)
+			if err != nil {
+				return err
+			}
+			err = s.Put(&storageVersionItem{Version: key})
 			if err != nil {
 				return err
 			}
--- a/pkg/storagev2/migration/migration.go
+++ b/pkg/storagev2/migration/migration.go
@@ -1,3 +1,7 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
 package migration
 
 import (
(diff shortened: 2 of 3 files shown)


pkg/storagev2/migration/migration_test.go
```diff
@@ -0,0 +1,76 @@
...
+		1: func(s storage.Store) error {
+			err := s.Put(&obj1{
+				Id:      "aaaaaaaaaaa",
+				SomeInt: 3,
+				Buf:     []byte("hello"),
+			})
+			if err != nil {
+				return err
+			}
+			return nil
```
vladopajic, line 26: just `return err`

pkg/storagev2/migration/migration.go
```diff
@@ -0,0 +1,68 @@
...
+			err := step(s)
+			if err != nil {
+				return err
+			}
+			err = s.Put(&storageVersionItem{Version: key})
+			if err != nil {
+				return err
+			}
+		}
+	}
```
vladopajic, line 29: this for cycle will execute only once, but it needs to execute as many as there are newer versions

pkg/storagev2/migration/migration_test.go
```diff
@@ -0,0 +1,76 @@
...
+		2: func(s storage.Store) error {
+			err := s.Put(&obj1{
+				Id:      "bbbbbbbbbbb",
+				SomeInt: 1,
+				Buf:     []byte("hey"),
+			})
+			if err != nil {
+				return err
+			}
+			return nil
```
vladopajic, line 37: just `return err`

pkg/storagev2/migration/migration_test.go
```diff
@@ -0,0 +1,76 @@
...
+		},
+	}
+
+	t.Run("migration test", func(t *testing.T) {
+		t.Parallel()
+		if err := Migrate(S, stepsMapTests); err != nil {
+			t.Errorf("Migrate() error = %v", err)
+		}
```
vladopajic, line 45: needs more assertions here:

- assert initial store version
- assert last store version (after migration)
- assert if store has items after migration

pkg/storagev2/migration/export_test.go
```diff
@@ -0,0 +1,5 @@
+package migration
+
+type (
+	StorageVersionItem = storageVersionItem
+)
```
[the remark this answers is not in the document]
umerm-work: Yes, I was going to use it in the test. Will remove this.

pkg/storagev2/migration/migration.go
```diff
@@ -33,22 +35,41 @@ func Migrate(s storage.Store, sm StepsMap) error {
...
-	// TODO
-	// all versions should be in order n (where n min version value), n+1, n+2, n+3... (all values are increasing orders)
-	// 5,6,7,8 ok
-	// 5,6,7,8,10 nok
+	keys := make([]int, len(sm))
+	i := 0
+	for k := range sm {
+		keys[i] = int(k)
+		i++
+	}
```
vladopajic, line 52: Can `i` variable be part of `for`-loop?

Also, `versions` might be more indicative name then `keys`.

pkg/storagev2/migration/migration.go
```diff
@@ -33,22 +35,41 @@ func Migrate(s storage.Store, sm StepsMap) error {
...
+	sort.Ints(keys)
```
vladopajic, line 53: `sort.Ints` is needed for `FindMissingNumbers` function, therefore it should be moved to this function. Because otherwise this function might not produce expected results.

pkg/storagev2/migration/migration_test.go
```diff
@@ -118,12 +194,22 @@ func TestMigrate(t *testing.T) {
...
 		}
 
-		// TODO Assert that migrated values are in store
+		ObjectExists(t, s, 3, objT1, objT2, objT3)
 	})
 }
+func ObjectExists(t *testing.T, s storage.Store, expectedKeys int, keys ...storage.Key) {
+	if len(keys) != expectedKeys {
+		t.Errorf("expected %v keys, got %v", expectedKeys, len(keys))
+	}
```
vladopajic, line 209: `expectedKeys` is passed in argument which checks passed in arguments? Should it be fine to trust passed in `keys`?

pkg/storagev2/migration/migration_test.go
```diff
@@ -51,25 +51,101 @@ func TestGetSetVersion(t *testing.T) {
...
+			}
+		})
+	}
 }
 
 func TestMigrate(t *testing.T) {
 	t.Parallel()
+	objT1 := &obj1{Id: "aaa", SomeInt: 1}
+	objT2 := &obj1{Id: "bbb", SomeInt: 2}
+	objT3 := &obj1{Id: "ccc", SomeInt: 3}
```
vladopajic, line 135: 👍

Perhaps have same for others tests?

pkg/storagev2/migration/migration_test.go
vladopajic: Usually test helpers functions like `ObjectExists` start with prefix `assert` eg.  `assertObjectExists`.

pkg/storagev2/migration/migration.go
```diff
@@ -0,0 +1,112 @@
...
+	"fmt"
+	"sort"
+	"unsafe"
+
+	storage "github.com/ethersphere/bee/pkg/storagev2"
+)
+
+type Version = uint64
+type StepFn func(storage.Store) error
+type StepsMap = map[Version]StepFn
```
notanatol, line 15: we usually group these in one
```suggestion
type (
    Version = uint64
    StepFn func(storage.Store) error
    StepsMap = map[Version]StepFn
)
```
in regards to the type aliases, perhaps we should avoid them without concrete necessity: https://yourbasic.org/golang/type-alias/

pkg/storagev2/migration/migration.go
umerm-work: `i` can't be part of for range loop as per loop syntax.

pkg/storagev2/migration/migration_test.go
umerm-work: Yes. Done

pkg/storagev2/migration/migration.go
vladopajic: right, this is range of maps

pkg/storagev2/migration/migration.go
```diff
@@ -0,0 +1,115 @@
...
+		i++
+	}
+
+	missing := FindMissingNumbers(versions)
+
+	if len(missing) > 0 {
+		return fmt.Errorf("missing versions: %v", missing)
+	}
+	return nil
+}
```
vladopajic, line 62: Just realized that this function could be simplified. All we need to do is find `min` and `max` version numbers, and then we can check if all version steps are defined by doing something like this:

```
if (maxVersion - minVersion) == len(sm) {
  // ok
} else {
 // nok
}
```

umerm-work: So lets say `[1 2 3]` is the version array

```
3-1 == 3 {
}

```
This is always false.

vladopajic: if we tweak formula a bit, like so:
```
(maxVersion - minVersion) == len(sm) - 1 
```

umerm-work: Thank worked. Thanks.

pkg/storagev2/migration/migration.go
```diff
@@ -0,0 +1,103 @@
...
+			// so we stop here
+			return nil
+		}
+	}
+}
+
+// ValidateVersions checks versions if they are in order n (where n min version value), n+1, n+2, n+3... (all values are increasing orders)
+func ValidateVersions(sm StepsMap) error {
+	versions := make([]int, len(sm))
+	i := 0
```
vladopajic, line 50: var `i` seems unused ?

pkg/storagev2/migration/migration_test.go
```diff
@@ -0,0 +1,243 @@
...
+	t.Parallel()
+	objT1 := &obj1{Id: "aaa", SomeInt: 1}
+	objT2 := &obj1{Id: "bbb", SomeInt: 2}
+	objT3 := &obj1{Id: "ccc", SomeInt: 3}
+
+	tests := []struct {
+		name    string
+		input   StepsMap
+		wantErr bool
+	}{
```
mrekucci, line 62: Testing empty input would be also nice.

mrekucci (commented):
It would be nice if the marshaling/unmarshaling of the `storageVersionItem` will be tested. There is a helper function `TestItemMarshalAndUnmarshal` for this located in the `storagetest` package, see how it is used in other places for comprehensive test coverage. There is also a missing header in each file.

pkg/storagev2/migration/migration.go
```diff
@@ -0,0 +1,111 @@
+package migration
```
mrekucci, line 1: Missing header.

pkg/storagev2/migration/migration_test.go
```diff
@@ -0,0 +1,279 @@
...
+		{
+			name: "zero values",
+			test: &storagetest.ItemMarshalAndUnmarshalTest{
+				Item:    &StorageVersionItem{},
+				Factory: func() storage.Item { return new(StorageVersionItem) },
+			},
+		},
+		{
+			name: "min value",
```
mrekucci, line 239: `1` is not a minimum value.
The test should cover the following cases:
- min value (already covered in "zero values")
- max value
- random value (within the min - max range)
- invalid size

testnet.yaml
```diff
@@ -0,0 +1,22 @@
+
```
mrekucci, line 1: This file shouldn't be included. Add its name to your `.git/info/exclude` so it will be ignored by git.

pkg/storagev2/migration/migration.go
```diff
@@ -0,0 +1,111 @@
...
+	"encoding/binary"
+	"errors"
+	"fmt"
+	storage "github.com/ethersphere/bee/pkg/storagev2"
+	"sort"
+)
+
+type (
+	// StepFn is a function that migrates the storage to the next version
+	StepFn func(storage.Store) error
```
mrekucci, line 13: I'd consider renaming `StepFn` to `Action` and `StepsMap` to `Steps`.
@aloknerurkar @vladopajic what do you think?

pkg/storagev2/migration/migration.go
```diff
@@ -0,0 +1,101 @@
...
+		if stepFn, ok := sm[nextVersion]; ok {
+			err := stepFn(s)
+			if err != nil {
+				return err
+			}
+			err = SetVersion(s, nextVersion)
+			if err != nil {
+				return err
+			}
+		} else {
```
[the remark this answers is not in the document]
mrekucci:
```golang
stepFn, ok := sm[nextVersion]
if !ok {
	return nil
}

err := stepFn(s)
if err != nil {
	return err
}
return SetVersion(s, nextVersion)
```

pkg/storagev2/migration/migration.go
```diff
@@ -0,0 +1,111 @@
...
+func (s *StorageVersionItem) Marshal() ([]byte, error) {
+	buf := make([]byte, storageVersionItemSize)
+	binary.LittleEndian.PutUint64(buf, s.Version)
+	return buf, nil
+}
+
+// Unmarshal implements the storage.Item interface.
+func (s *StorageVersionItem) Unmarshal(bytes []byte) error {
+	if len(bytes) != storageVersionItemSize {
+		return fmt.Errorf("invalid bytes size")
```
mrekucci, line 91: The error should be factored out so you can check it in your tests. By following the convention set out in other files, you get:

```golang
// errTagIDAddressItemUnmarshalInvalidSize is returned when trying
// to unmarshal buffer that is not of size tagIDAddressItemSize.
errStorageVersionItemUnmarshalInvalidSize = errors.New("unmarshal StorageVersionItem: invalid size")
```

pkg/storagev2/migration/migration.go
vladopajic: If we zoom out and look for the most generic name for `func(storage.Store) error` what would it be? 

---

In the time of writing proposal I haven't got any good generic name idea for such function, because signature is vary broad and can fit many concepts. 

Decision was to have have highly specific name in the context of migration. At that time options like `step`, `increment`, `stage` were floating in my mind. 

Name `action` seems to be going in generic name direction. I would rather avoid generic name unless abstraction for this type is occurring in other places.

---

One potential name in generic category was `mutator` because it implies that `store` is going to be mutated in this function (bonus it's also a verb).

pkg/storagev2/migration/migration.go
```diff
@@ -0,0 +1,119 @@
...
+			return 0, nil
+		}
+		return 0, err
+	}
+	return item.Version, nil
+}
+
+func SetVersion(s storage.Store, v uint64) error {
+	return s.Put(&StorageVersionItem{Version: v})
+}
```
vladopajic, line 119: Would it make sense to unexport `SetVersion` function?

pkg/storagev2/migration/migration.go
mrekucci: I wouldn't say it's generic, you have to read it in the context of the package, which is `migration.Action` and `migration.Steps`. I think compared to `migration.StepFn` and `migration.StepsMap` they are clearer, especially in the case of `StepsMap` where I consider the `Map` suffix to be an unnecessary detail.

vladopajic: `migration.Steps` is better then `migration.StepsMap` we agree on that. But for `migration.Action` I am not quite sure...

pkg/storagev2/migration/migration_test.go
```diff
@@ -0,0 +1,298 @@
...
+		},
+		{
+			name: "max value",
+			test: &storagetest.ItemMarshalAndUnmarshalTest{
+				Item:    &migration.StorageVersionItem{Version: 999999999999999999},
+				Factory: func() storage.Item { return new(migration.StorageVersionItem) },
+			},
+		},
+		{
+			name: "invalid size",
```
mrekucci, line 251: This is not a correct test for invalid size, the correct is:
```golang
{
	name: "invalid size",
	test: &storagetest.ItemMarshalAndUnmarshalTest{
		Item: &storagetest.ItemStub{
			MarshalBuf:   []byte{0xFF},
			UnmarshalBuf: []byte{0xFF},
		},
		Factory:      func() storage.Item { return new(migration.StorageVersionItem) },
		UnmarshalErr: migration.ErrStorageVersionItemUnmarshalInvalidSize,
	},
},
```
You have to create `export_test.go` file in order to export the `migration.ErrStorageVersionItemUnmarshalInvalidSize` error.

pkg/storagev2/migration/migration_test.go
```diff
@@ -0,0 +1,298 @@
...
+			test: &storagetest.ItemMarshalAndUnmarshalTest{
+				Item:    &migration.StorageVersionItem{Version: 0},
+				Factory: func() storage.Item { return new(migration.StorageVersionItem) },
+			},
+		},
+		{
+			name: "random value",
+			test: &storagetest.ItemMarshalAndUnmarshalTest{
+				Item:    &migration.StorageVersionItem{Version: 1234},
```
mrekucci, line 260: `1234` -> `rand.Uint64()`.

pkg/storagev2/migration/migration.go
mrekucci: We can go with `migration.StepFn`.

pkg/storagev2/migration/migration.go
vladopajic: In that case would it make sense to have `SetVersion` export only for tests files (similary to how `errStorageVersionItemUnmarshalInvalidSize` was exported)?

pkg/storagev2/migration/migration_test.go
```diff
@@ -0,0 +1,304 @@
...
+		newVersion, err := migration.Version(s)
+		if err != nil {
+			t.Errorf("Version() unexpected error: %v", err)
+		}
+		if newVersion != 8 {
+			t.Errorf("new version = %v must be 8", newVersion)
+		}
+
+		assertObjectExists(t, s, objT1, objT2, objT3)
+	})
```
aloknerurkar, line 220: One more test case to add here is failure in one of the step function. If that happens we need to ensure that the version was correctly set to the last successful migration step.

umerm-work: Done

aloknerurkar (approved):

PR #3425, opened 2022-10-12
fix: increase timeout for release action and bump base docker images

vandot:
Latest RC release failed in the end because of default timeout that goreleaser uses (30min). Good is that artifacts were uploaded but workflow was cut by the timeout and release is marked as failed.
Base docker images are upgraded as a part of a regular update of all infra components.


istae (approved):

vladopajic (approved):
LGTM

PR #3428, opened 2022-10-13
fix: retrieval rewrite

istae:
### Checklist

### Description
The PR simplifies the retrieval logic while maintaining preventive and on error retries.
Tests passed without any edits.

### Related Issue (Optional)
https://github.com/ethersphere/bee-backlog/issues/36

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3428)

### DIFF
--- a/pkg/retrieval/retrieval.go
+++ b/pkg/retrieval/retrieval.go
@@ -246,16 +246,13 @@ func (s *Service) retrieveChunk(ctx context.Context, done chan struct{}, result 
 	)
 
 	defer func() {
+		if err != nil {
+			s.metrics.TotalErrors.Inc()
+		}
 		select {
 		case result <- retrievalResult{err: err, chunk: chunk, retrieveAttempted: retrieveAttempted, peer: peer}:
 		case <-done:
 			return
-		}
-	}()
-
-	defer func() {
-		if err != nil {
-			s.metrics.TotalErrors.Inc()
 		}
 	}()
 
--- a/pkg/retrieval/retrieval.go
+++ b/pkg/retrieval/retrieval.go
@@ -162,12 +162,13 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr, sourcePeerAddr swarm.
 
 			select {
 			case <-ctx.Done():
-				loggerV1.Debug("failed to get chunk", "chunk_address", addr, "error", ctx.Err())
-				return nil, fmt.Errorf("retrieval: %w", ctx.Err())
+				return nil, ctx.Err()
 			case <-preemptiveTicker.C:
 				retry()
 			case <-retryC:
+
 				s.metrics.PeerRequestCounter.Inc()
+
 				inflight++
 
 				go func() {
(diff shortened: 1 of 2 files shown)


istae:
@metacertain writes:

so the definition of requestRound in current retrieval is actually continuing with a skiplist that has the overdraft entries removed by a skiplist reset, and the time difference between the beginning of 2 requestRounds is at least 600 milliseconds (with 1024 rounds this covers around 10 minutes at least)
this currently enables an ultralightnode trying to retrieve all the chunks of a 200 MB file to keep trying for up to 10 minutes to get a peer with free accounting balance for each of these chunks, so sooner or later they will be retrieved.
the effective filesize limit that you can get with an ultralight node comes from this 10 minute window and the / second free bandwidth, so 600 * 7 KB/sec from each peer ( lets say with 64 peers  this  means 600 * 448 KB ~ 270 MBs )

without this, we would try to retrieve too many chunks at the same time from our peers, and many of the chunks would fail because all accounting balances are saturated with pending requests after a short while (especially as an ultra light node) 

the current logic is, for up to 1024 times, we start with a reset skiplist that only contains nodes with real failures (either request failed to send or later error) 
each of these times, we select up to 32 next-closest addresses and try to get accounting enable the request etc

this means the maximum number of next closest peer selections is 1024 * 32, but we will give up much sooner in case we run out of peers with no failure / overdraft or if we have enough real failures after successful requests sent

we give up (as an originator) when 32 succesfully sent requests got an actual failure in any of the rounds
we give up (as a forwarder) when 1 successfully sent request got an actual failure (and there are no rounds, just 1 time 32 next-closest addresses are selected) 
we also give up if we run out of next-closest-peers and nobody on the skiplist is an overdraft entry

what i see in the rewrite attempt, is that it never does a skiplist reset, so overdraft entries are never removed, and never retried when we could later perhaps get free accounting balance
even if this exists, you still need the request rounds so that a number of peers are tried before the skiplist reset, so its not that only 1 peer is constantly waited to be retried in case of an accounting block

istae:
cont..

actually this 1024 limit is completely artificial and could be an infinite loop of request rounds anyway, so until the user explicitly cancels a request its still trying to continue no matter the time

that will give us infinite download size on ultra light node

its not that we would stay in the loop forever unless there are yet to be tried peers anyway (because of accounting blocks)

istae:
The lastest commit should address @metacertain's comments

aloknerurkar (approved):

pkg/retrieval/retrieval.go
```diff
@@ -126,156 +126,145 @@ func (s *Service) RetrieveChunk(ctx context.Context, addr, sourcePeerAddr swarm.
...
+
+	defer func() {
+		select {
+		case result <- retrievalResult{err: err, chunk: chunk, retrieveAttempted: retrieveAttempted, peer: peer}:
+		case <-done:
+			return
+		}
+	}()
+
+	defer func() {
```
mrekucci, line 256: No need for a second `defer` call, the error check can be put at the beginning of the `defer` above.

mrekucci (approved):
Aside from the comment, LGTM.

janos (approved):
LGTM, I am also questioning the buffered channel.

aloknerurkar (approved):

PR #3430, opened 2022-10-17
feat: stamp validity check upon sample insert

metacertain:
### Checklist

### Description
This PR intends to add stamp validity checks for items that make it into a reserve commitment sample

Reliability improvement

### Related Issue (Optional)
Storage incentives phase 2

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3430)

### DIFF
--- a/pkg/localstore/localstore.go
+++ b/pkg/localstore/localstore.go
@@ -188,6 +188,9 @@ type DB struct {
 	metrics metrics
 
 	logger log.Logger
+
+	//
+	validStamp postage.ValidStampFn
 }
 
 // Options struct holds optional parameters for configuring DB.
@@ -278,7 +281,7 @@ func safeInit(rootPath, sharkyBasePath string, db *DB) error {
 // New returns a new DB.  All fields and indexes are initialized
 // and possible conflicts with schema from existing database is checked.
 // One goroutine for writing batches is created.
-func New(path string, baseKey []byte, ss storage.StateStorer, o *Options, logger log.Logger) (db *DB, err error) {
+func New(path string, baseKey []byte, ss storage.StateStorer, o *Options, logger log.Logger, validStamp postage.ValidStampFn) (db *DB, err error) {
 	if o == nil {
 		// default options
 		o = &Options{
@@ -309,6 +312,7 @@ func New(path string, baseKey []byte, ss storage.StateStorer, o *Options, logger
 		reserveEvictionWorkerDone: make(chan struct{}),
 		metrics:                   newMetrics(),
 		logger:                    logger.WithName(loggerName).Register(),
+		validStamp:                validStamp,
 	}
 	if db.cacheCapacity == 0 {
 		db.cacheCapacity = defaultCacheCapacity
--- a/pkg/localstore/localstore.go
+++ b/pkg/localstore/localstore.go
@@ -189,7 +189,6 @@ type DB struct {
 
 	logger log.Logger
 
-	//
 	validStamp postage.ValidStampFn
 }
 
@@ -215,7 +214,8 @@ type Options struct {
 	// DisableSeeksCompaction toggles the seek driven compactions feature on leveldb
 	// and is passed on to shed.
 	DisableSeeksCompaction bool
-
+	// Stamp validator for reserve sampler
+	ValidStamp postage.ValidStampFn
 	// MetricsPrefix defines a prefix for metrics names.
 	MetricsPrefix string
 	Tags          *tags.Tags
@@ -281,7 +281,7 @@ func safeInit(rootPath, sharkyBasePath string, db *DB) error {
 // New returns a new DB.  All fields and indexes are initialized
 // and possible conflicts with schema from existing database is checked.
 // One goroutine for writing batches is created.
-func New(path string, baseKey []byte, ss storage.StateStorer, o *Options, logger log.Logger, validStamp postage.ValidStampFn) (db *DB, err error) {
+func New(path string, baseKey []byte, ss storage.StateStorer, o *Options, logger log.Logger) (db *DB, err error) {
 	if o == nil {
 		// default options
 		o = &Options{
@@ -312,7 +312,7 @@ func New(path string, baseKey []byte, ss storage.StateStorer, o *Options, logger
 		reserveEvictionWorkerDone: make(chan struct{}),
 		metrics:                   newMetrics(),
 		logger:                    logger.WithName(loggerName).Register(),
-		validStamp:                validStamp,
+		validStamp:                o.ValidStamp,
 	}
 	if db.cacheCapacity == 0 {
 		db.cacheCapacity = defaultCacheCapacity
--- a/pkg/localstore/sampler.go
+++ b/pkg/localstore/sampler.go
@@ -202,9 +202,9 @@ func (db *DB) ReserveSample(
 			_, err = db.validStamp(chunk, stampData)
 			if err == nil {
 				if !soc.Valid(chunk) && !cac.Valid(chunk) {
-					insert(item.transformedAddress)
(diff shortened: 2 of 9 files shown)


pkg/localstore/localstore.go
```diff
@@ -278,7 +281,7 @@ func safeInit(rootPath, sharkyBasePath string, db *DB) error {
 // New returns a new DB.  All fields and indexes are initialized
 // and possible conflicts with schema from existing database is checked.
 // One goroutine for writing batches is created.
-func New(path string, baseKey []byte, ss storage.StateStorer, o *Options, logger log.Logger) (db *DB, err error) {
+func New(path string, baseKey []byte, ss storage.StateStorer, o *Options, logger log.Logger, validStamp postage.ValidStampFn) (db *DB, err error) {
```
aloknerurkar, line 284: Can we add it to the Options and not an extra parameter? This way we dont need to add it everywhere?

Also, it can be initialized to nil and in the Sample function we can check if its not nil before using it. That way this doesnt have to be changed everywhere/

pkg/localstore/localstore.go
```diff
@@ -188,6 +188,9 @@ type DB struct {
 	metrics metrics
 
 	logger log.Logger
+
+	//
```
aloknerurkar, line 192: Remove the extra comment tag

aloknerurkar (changes requested):

aloknerurkar (approved):

pkg/localstore/sampler.go
```diff
@@ -180,8 +188,28 @@ func (db *DB) ReserveSample(
...
+			stamp := postage.NewStamp(item.chunkItem.BatchID, item.chunkItem.Index, item.chunkItem.Timestamp, item.chunkItem.Sig)
+
+			stampData, err := stamp.MarshalBinary()
+			if err != nil {
+				logger.Info("error marshaling stamp for chunk", chunk.Address(), err)
+				continue
+			}
+			_, err = db.validStamp(chunk, stampData)
+			if err == nil {
+				if !soc.Valid(chunk) && !cac.Valid(chunk) {
```
istae, line 204: in pullsync, we make these checks before we store the chunks.

pkg/localstore/sampler.go
```diff
@@ -180,8 +188,28 @@ func (db *DB) ReserveSample(
...
+			chunk := swarm.NewChunk(swarm.NewAddress(item.chunkItem.Address), item.chunkItem.Data)
+
+			stamp := postage.NewStamp(item.chunkItem.BatchID, item.chunkItem.Index, item.chunkItem.Timestamp, item.chunkItem.Sig)
+
+			stampData, err := stamp.MarshalBinary()
+			if err != nil {
+				logger.Info("error marshaling stamp for chunk", chunk.Address(), err)
+				continue
+			}
+			_, err = db.validStamp(chunk, stampData)
```
istae, line 202: Before merging this branch, I would suggest that we test the branch on mainnet, and see how many of these errors we actually see. Maybe it's not worth adding it.

ldeffenb: If a node has lots of pinned chunks with expired stamps, you'll see a LOT of them.  And that node would not be able to participate in storage incentives because of those expired-but-locally-pinned chunks.

notanatol (approved):

pkg/localstore/sampler.go
metacertain: this check intends to catch if the data would be arbitrarily changed (became invalid) later on

istae: can you clarify what you mean?

metacertain: any kind of storage layer unreliability that would be caught in every other use of retrieved data (either local or from the network) wouldn't be caught here without this check.
not doing this check here is explicitly trusting the storage layer

istae, line 204: Since most chunks are are cacs, checking `cac.Valid(..)` first in the if statement is a simple optimization to have here.

pkg/storage/testing/chunk.go
```diff
@@ -72,6 +72,27 @@ func GenerateTestRandomChunkAt(target swarm.Address, po int) swarm.Chunk {
 	return swarm.NewChunk(addr, data).WithStamp(stamp)
 }
 
+// GenerateTestRandomChunkAt generates an invalid (!) chunk with address of proximity order po wrt target.
+func GenerateValidRandomChunkAt(target swarm.Address, po int) swarm.Chunk {
```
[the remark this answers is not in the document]
metacertain: this is a helper function for a test?

istae (approved):

pkg/storage/testing/chunk.go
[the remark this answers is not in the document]
aloknerurkar: Yes so it has independent logic which is assumed to be correct while running the test. So it has to be unit-tested like any other code. We do this in a lot of places in the code. Eg:

https://github.com/ethersphere/bee/blob/master/pkg/storage/mock/storer_test.go
https://github.com/ethersphere/bee/blob/master/pkg/swarm/test/helper_test.go

The reason I am stressing about this is we had this GenerateRandomChunkAt which was incorrect all this time. This makes the tests using this function unreliable as we don't know if this logic does what it's supposed to.

pkg/localstore/sampler.go
aloknerurkar: We ran this on `dev-bee-storage` nodes. We do see some invalid stamp messages. But we dont see the `data invalid` log. So this means our assumptions are correct.

@ldeffenb With this change, it could be possible that your node could be able to participate if it can check and discard all the uninteresting chunks within the round time.

pkg/storage/testing/chunk.go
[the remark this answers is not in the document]
istae: Which GenerateRandomChunkAt was wrong all this time 👀

pkg/localstore/sampler.go
ldeffenb: > @ldeffenb With this change, it could be possible that your node could be able to participate if it can check and discard all the uninteresting chunks within the round time.

Once this is on the mainnet, I might do a hack to have it generate a sample even though it is not staked/not playing.   Then I can see the time it takes to generate the sample (and the system load as well), and compare my hash manually with what is eventually revealed to know if it is safe to stake.   We (@mfw78 and I) are working on a monitor similar to monBee that we'll hopefully release on Monday that makes the schelling game pretty easy to watch.

ldeffenb: @aloknerurkar @istae Do you know if the blocksPerRound will remain at 152 in the gnosis deployment resulting in ~3 times faster game rounds, or will it go to 364 to keep the rounds at approximately 30 minutes each?   Or maybe land somewhere in between at some other integer multiple of 4?  Just curious for planning purposes.  My testnet nodes at depth 2 are taking anywhere from 3-5 minutes to generate their samples as a point of reference.

pkg/localstore/localstore.go
ldeffenb: And just to note: I do have one testnet node with one copy of my OSM tiles pinned from zooms 0 through 12 (vs mainnet zooms 0 through 18).  It has been successfully playing and winning the storage incentives, but I also know that it was wiped recently has doesn't have any expired stamp chunks in it, at least not pinned in my dataset.

pkg/localstore/sampler.go
metacertain: we intend to keep 152 in gnosis, however this might still change. the actual time window to calculate the sample and apply is the claim + commit phase, so for example you should be fine with 3-5 minute calculation

PR #3437, opened 2022-10-18
fix: puller rewrite and bug fixes

istae:
### Checklist

### Description
Fixes bugs tagged in the `reserve-bugs` branch.
Some bugs are:
1. syncing with peers only within sync radius, not neighborhood depth
2. liveSync process wrongly exits on error (specially context cancellation)


pkg/puller/puller.go
```diff
@@ -406,13 +304,19 @@ func (p *Puller) liveSyncWorker(ctx context.Context, peer swarm.Address, bin uin
...
 		}
+
 		top, ruid, err := p.syncer.SyncInterval(ctx, peer, bin, from, pullsync.MaxCursor)
 		if err != nil {
-			loggerV2.Debug("liveSyncWorker exit on sync error", "peer_address", peer, "bin", bin, "from", from, "error", err)
-			if ruid == 0 {
-				p.metrics.LiveWorkerErrCounter.Inc()
+			p.metrics.LiveWorkerErrCounter.Inc()
+
+			if errors.Is(err, context.Canceled) {
```
aloknerurkar, line 312: This is risky to do if the peer is not reachable. We will stay in this retry loop which could cause a lot of CPU usage. Maybe we should have some retry counter.

Also, I would add a log message to say which attempt it is when we call SyncInterval again. This can help us identify this problem.

Also unit test needs to be added for this behaviour.

aloknerurkar (changes requested):

pkg/localstore/subscription_pull.go
```diff
@@ -44,18 +43,16 @@ func (db *DB) SubscribePull(ctx context.Context, bin uint8, since, until uint64)
 
 	chunkDescriptors := make(chan storage.Descriptor)
 
-	in, out, clean := flipflop.NewFallingEdge(flipFlopBufferDuration, flipFlopWorstCaseDuration)
+	trigger := make(chan struct{}, 1)
```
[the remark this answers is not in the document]
istae: see line 208, subscription_pull.go. 
It is a select with a default, so no blocks

pkg/localstore/subscription_pull.go
```diff
@@ -171,9 +167,9 @@ func (db *DB) SubscribePull(ctx context.Context, bin uint8, since, until uint64)
 		defer db.pullTriggersMu.Unlock()
 
 		for i, t := range db.pullTriggers[bin] {
-			if t == in {
+			if t == trigger {
 				db.pullTriggers[bin] = append(db.pullTriggers[bin][:i], db.pullTriggers[bin][i+1:]...)
```
mrekucci, line 171: The channel can be closed here.

mrekucci (approved):

pkg/puller/puller.go
```diff
@@ -104,235 +101,131 @@ func (p *Puller) manage(warmupTime time.Duration) {
...
-			syncRadius := p.reserveState.GetReserveState().StorageRadius
-
-			// we defer the actual start of syncing to get out of the iterator first
-			var (
-				peersToSync       []peer
-				peersToRecalc     []peer
-				peersDisconnected = make(map[string]peer)
-			)
 
 			p.syncPeersMtx.Lock()
```
vladopajic, line 111: Does this `Lock()` has corresponding `Unlock()`?

pkg/puller/puller.go
```diff
@@ -104,235 +101,131 @@ func (p *Puller) manage(warmupTime time.Duration) {
...
 				return false, false, nil
 			}, topology.Filter{Reachable: true})
 
-			p.syncPeersMtx.Unlock()
-
-			for _, v := range peersToSync {
-				p.syncPeer(ctx, v.addr, v.po, syncRadius)
+			for addr, peer := range peersDisconnected {
+				p.disconnectPeer(addr, peer.po)
 			}
```
vladopajic, line 141: Is this going to call `disconnectPeer` for the second time, where first time is on line 133 `p.disconnectPeer(addr.ByteString(), po)`?

pkg/puller/puller.go
```diff
@@ -104,235 +101,131 @@ func (p *Puller) manage(warmupTime time.Duration) {
...
 		}
 	}
 }
 
-func (p *Puller) disconnectPeer(peer swarm.Address, po uint8) {
-	loggerV2 := p.logger.V(2).Register()
+// disconnectPeer cancels all existing syncing and removes the peer entry from the syncing map.
+// Must be called under lock.
+func (p *Puller) disconnectPeer(peerAddr string, po uint8) {
+	loggerV2 := p.logger
```
mrekucci, line 154: This logger will log to `Debug`, not to `v=2`. Is this desirable? If so, just use `p.logger` directly. The same is in the methods below.

pkg/puller/puller.go
```diff
@@ -390,13 +283,30 @@ func (p *Puller) histSyncWorker(ctx context.Context, peer swarm.Address, bin uin
...
 		p.metrics.LiveWorkerIterCounter.Inc()
+
+		if sleep {
+			select {
+			case <-p.quit:
+				loggerV2.Debug("liveSyncWorker quit on shutdown", "peer_address", peer, "bin", bin, "cursor", cur)
+				return
+			case <-ctx.Done():
+				loggerV2.Debug("liveSyncWorker context cancelled", "peer_address", peer, "bin", bin, "cursor", cur)
+				return
```
vladopajic, line 304: Should `ctx.Done()` signal be enough, as `p.quity` seems to be canceling context? (There are multiple occurrences of this pattern)

Anyhow, it would be simpler to follow if there was only one end signal. Preferably create one `Context` and use that context as ending signal propagate where necessary).

pkg/puller/puller.go
mrekucci: Nice catch!

pkg/puller/puller.go
istae: this is temporary, thanks

pkg/puller/puller.go
```diff
@@ -434,7 +350,7 @@ func (p *Puller) liveSyncWorker(ctx context.Context, peer swarm.Address, bin uin
 			p.logger.Error(err, "liveSyncWorker exit on add peer interval", "peer_address", peer, "bin", bin, "from", from, "error", err)
 			return
 		}
-		loggerV2.Debug("liveSyncWorker pulled bin failed", "bin", bin, "from", from, "topmost", top, "peer_address", peer)
+		loggerV2.Debug("liveSyncWorker pulled bin", "bin", bin, "from", from, "topmost", top, "peer_address", peer)
 
 		from = top + 1
 	}
```
vladopajic, line 356: `(p *Puller) Close()` method seems to have flaw, what is usage of `cc` channel?

pkg/puller/puller.go
```diff
@@ -508,11 +424,14 @@ func peerIntervalKey(peer swarm.Address, bin uint8) string {
 type syncPeer struct {
 	address        swarm.Address
 	binCancelFuncs map[uint8]func() // slice of context cancel funcs for historical sync. index is bin
+	po             uint8
+
+	cursors []uint64
```
vladopajic, line 429: `cursors` property seems to be unused in `syncPeer` structure, consider removing it.

pkg/puller/puller.go
istae: yes, disconnect for out of depth peers, and disconnect call for peers that were in the syncPeers map but are no longer connected (eg missing in the iterator).

pkg/localstore/subscription_pull.go
istae: that will trigger the above select, it should be okay like this

pkg/puller/puller.go
istae: will look into this, thanks.

pkg/puller/puller.go
istae: no, we use it :)

pkg/puller/puller.go
istae: we have a global context that is created in manage, it's used to signal blocking processes to terminate on exit. 
We also need a perBin contexts in case we need to stop syncing a specific bin.

vladopajic: If perBin context is created using global context, then canceling global context would cancel perBin context. If context are created in this way then `p.quit` is not needed.

istae: i see what you mean now, we do not need to listen on p.quite in the syncing for loops. Thanks.

janos (commented):
I am not so sure on removing the falling edge detection on subscription. The reasoning in the PR description that it is related to the node bootup does not have to be true. That is a protection for any situation where a large number of chunks are triggering the subscription to call the pullIndex.Iterate too frequently causing an extensive I/O. I think that removing this type of protections should be backed up with measurements.

pkg/puller/puller.go
```diff
@@ -390,29 +279,46 @@ func (p *Puller) histSyncWorker(ctx context.Context, peer swarm.Address, bin uin
...
 	from := cur + 1
+
+	sleep := false
+
 	for {
 		p.metrics.LiveWorkerIterCounter.Inc()
+
+		if sleep {
+			select {
+			case <-ctx.Done():
```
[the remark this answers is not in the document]
istae: The removal of the falling edge detection is out of the scope of this PR, so I will revert. 

The manage loop disconnects out of the depth peers and gone peers (peers that are in syncPeer, but did not come up in the next run of the iterator. `disconnectPeer` calls `peer.gone()` which cancels contexts.

pkg/puller/puller.go
```diff
@@ -104,247 +100,154 @@ func (p *Puller) manage(warmupTime time.Duration) {
...
 
-	defer func() {
-		p.wg.Done()
-		p.metrics.HistWorkerDoneCounter.Inc()
-	}()
-	loggerV2.Debug("histSyncWorker starting", "peer_address", peer, "bin", bin, "cursor", cur)
 	for {
 		p.metrics.HistWorkerIterCounter.Inc()
+
+		if sleep {
```
janos, line 240: The value of `sleep` variable is never changed.

It appears that the retry functionality in histSyncWorker is not tested, according to the coverage report. It would be good to have this functionality tested as it seems fragile enough for accidental subtle bugs.

janos (commented):
As a measure of precaution, I would suggest to add regression tests for all issues that these changes are addressing, for both validation and also for protection of reintroducing same issues in the future.

Additionally, I get this test case failure:

```
=== CONT  TestDepthChange/move_peer_around
    /Users/janos/go/projects/ethswarm.org/bee/pkg/puller/puller_test.go:440: got unexpected interval: [], want [[1 1]] bin 3
```

which passes on the master branch.

pkg/puller/puller.go
```diff
@@ -104,247 +100,154 @@ func (p *Puller) manage(warmupTime time.Duration) {
...
 	}
-	p.cursorsMtx.Unlock()
+	delete(p.syncPeers[po], addr.ByteString())
 }
 
-func (p *Puller) recalcPeer(ctx context.Context, peer swarm.Address, po, d uint8) (dontSync bool) {
-	loggerV2 := p.logger.V(2).Register()
+// recalcPeers starts or stops syncing process for peers per bin depending on the current sync radius.
+// Must be called under lock.
+func (p *Puller) recalcPeers(ctx context.Context, syncRadius uint8) (dontSync bool) {
```
vladopajic, line 166: Is return value from this func used?

pkg/puller/puller.go
```diff
@@ -354,87 +257,79 @@ func (p *Puller) histSyncWorker(ctx context.Context, peer swarm.Address, bin uin
...
 			p.logger.Error(err, "liveSyncWorker exit on add peer interval", "peer_address", peer, "bin", bin, "from", from, "error", err)
 			return
 		}
-		loggerV2.Debug("liveSyncWorker pulled bin failed", "bin", bin, "from", from, "topmost", top, "peer_address", peer)
+
+		if syncErr != nil {
+			p.metrics.LiveWorkerErrCounter.Inc()
+			p.logger.Error(syncErr, "liveSyncWorker sync error", "peer_address", peer, "bin", bin, "from", from, "topmost", top)
+			sleep = true
+		}
```
vladopajic, line 330: `syncErr` should be handled right after `p.syncer.SyncInterval` method is called; also name should be follow same convention.

pkg/puller/puller.go
```diff
@@ -71,11 +68,8 @@ func New(stateStore storage.StateStorer, topology topology.Driver, reserveState
 		syncer:       pullSync,
 		metrics:      newMetrics(),
 		logger:       logger.WithName(loggerName).Register(),
-		cursors:      make(map[string]peerCursors),
-
-		syncPeers: make([]map[string]*syncPeer, bins),
-		quit:      make(chan struct{}),
-		wg:        sync.WaitGroup{},
+		syncPeers:    make([]map[string]*syncPeer, bins),
+		quit:         make(chan struct{}),
```
vladopajic, line 72: Could `quit` channel be remove in favour of global `context` (like ctx created in manage goroutine)?

pkg/node/node.go
```diff
@@ -931,7 +931,7 @@ func NewBee(interrupt chan struct{}, sysInterrupt chan os.Signal, addr string, p
 
 	var pullerService *puller.Puller
 	if o.FullNodeMode && !o.BootnodeMode {
-		pullerService = puller.New(stateStore, kad, batchStore, pullSyncProtocol, logger, puller.Options{}, warmupTime)
+		pullerService = puller.New(stateStore, kad, batchStore, pullSyncProtocol, logger, puller.Options{}, warmupTime, puller.DefaultSyncSleepDur)
```
aloknerurkar, line 934: Seems like the DefaultSyncSleepDur can be part of the puller.Options. Also, it can be default unless specified as it is defined in the puller package anyway.

pkg/puller/puller.go
```diff
@@ -354,76 +251,68 @@ func (p *Puller) histSyncWorker(ctx context.Context, peer swarm.Address, bin uin
...
 		}
+
 		if s > cur {
-			loggerV2.Debug("histSyncWorker syncing finished", "bin", bin, "cursor", cur)
+			p.logger.Debug("histSyncWorker syncing finished", "bin", bin, "cursor", cur)
 			return
 		}
-		top, ruid, err := p.syncer.SyncInterval(ctx, peer, bin, s, cur)
+		top, _, err := p.syncer.SyncInterval(ctx, peer, bin, s, cur)
 		if err != nil {
```
aloknerurkar, line 263: So if there is actual error because of network connection, how would the worker be stopped? This could be multiple workers for each peer. So potentially we have a lot of workers just wasting CPU no?

pkg/puller/puller.go
```diff
@@ -354,76 +251,68 @@ func (p *Puller) histSyncWorker(ctx context.Context, peer swarm.Address, bin uin
...
-			// to try to send a message with an expired context, which is
-			// bound to fail.
-			ctxC, cancelC := context.WithTimeout(context.Background(), 10*time.Second)
-			defer cancelC()
-			if err := p.syncer.CancelRuid(ctxC, peer, ruid); err != nil {
-				loggerV2.Debug("histSyncWorker cancel ruid failed", "error", err)
-			}
-			return
+		top, _, err := p.syncer.SyncInterval(ctx, peer, bin, from, pullsync.MaxCursor)
+		if err != nil {
```
aloknerurkar, line 310: Same as the comment above. We should ideally look for particular errors to restart, if we get terminal errors like stream reset etc we should quit early.

pkg/puller/puller_test.go
```diff
@@ -445,6 +447,44 @@ func TestDepthChange(t *testing.T) {
...
+	)
+
+	puller, _, kad, pullsync := newPuller(opts{
+		kad: []mockk.Option{
+			mockk.WithEachPeerRevCalls(mockk.AddrTuple{Addr: addr, PO: 0}),
+			mockk.WithDepth(0),
+		},
+		pullSync: []mockps.Option{
+			mockps.WithCursors([]uint64{1}),
+			mockps.WithSyncError(errors.New("sync error"))},
```
aloknerurkar, line 466: It would be better if this is a function which returns error for few times and then returns success. We should also test if this can restart correctly.

pkg/puller/puller_test.go
```diff
@@ -445,6 +447,44 @@ func TestDepthChange(t *testing.T) {
 	}
 }
 
+// TestContinueSyncing adds a single peer with PO 0 to hist and live sync only a peer
```
aloknerurkar, line 450: There is still a lack of test coverage here. The `syncPeer.gone` and the `cancelBins` functionality is not being tested according to the coverage report.

pkg/puller/puller.go
```diff
@@ -354,76 +251,68 @@ func (p *Puller) histSyncWorker(ctx context.Context, peer swarm.Address, bin uin
...
-				p.metrics.HistWorkerErrCounter.Inc()
-			}
-
-			// since we use bin context cancellation to cancel interval
-			// sync operations, the context can be expired here, causing us
-			// to try to send a message with an expired context, which is
-			// bound to fail.
-			ctxC, cancelC := context.WithTimeout(context.Background(), 10*time.Second)
-			defer cancelC()
-			if err := p.syncer.CancelRuid(ctxC, peer, ruid); err != nil {
```
aloknerurkar, line 377: Why is this being removed? This makes it a breaking change in the protocols.

@janos Can you add some brief description about this cancel ruid functionality? Maybe we are missing something?

aloknerurkar (changes requested):

pkg/puller/puller.go
istae: see comment above

pkg/puller/puller_test.go
istae: >> test if this can restart correctly.
what do you mean ?

aloknerurkar: This will always return error from SyncInterval. Instead, we should return error and then on the next call we should succeed.

pkg/puller/puller.go
aloknerurkar: We should be able to test this behaviour. I think this is not trivial to understand from the code, so also we can add some comments about this.

We could test this by calling `ResetPeers` on this mock kademlia. This should return empty for the eachPeerRev calls and trigger the cancellations.

metacertain (approved):

pkg/puller/puller.go
janos: The whole pullsync cancel protocol is a way to signal a call to a context's cancel function for a particular syncing request. To avoid unnecessary wait in pullsync handler if the "client" syncing peer has problems storing intervals. Since stream termination (reset) can be detected only on their Read or Write methods, it was needed to have a mechanism to terminate other functions that happen in between if there is an error on the other peer.

.golangci.yml
```diff
@@ -23,7 +23,6 @@ linters:
     - importas
     - ineffassign
     - misspell
-    - nakedret
```
janos, line 26: I have no preference on naked vs not-naked returns, but could not see a strong enough need in this pr to change the linter configuration, and consequently the coding policy. The changes with naked returns have no functional requirement, while naked returns could be a source of a subtle problems.

janos (commented):
I would add that if the CancelRuid functionality is removed, an alternative approach to possible goroutone leak is required and to cleanup the functionality completely as without CancelRuid function calls, the whole pullsync cancel protocol can be removed.

pkg/node/node.go
```diff
@@ -931,7 +931,7 @@ func NewBee(interrupt chan struct{}, sysInterrupt chan os.Signal, addr string, p
 
 	var pullerService *puller.Puller
 	if o.FullNodeMode && !o.BootnodeMode {
-		pullerService = puller.New(stateStore, kad, batchStore, pullSyncProtocol, logger, puller.Options{}, warmupTime)
+		pullerService = puller.New(stateStore, kad, batchStore, pullSyncProtocol, logger, puller.Options{SyncSleepDur: puller.DefaultSyncSleepDur}, warmupTime)
```
aloknerurkar, line 934: We can even move the warmup time to the options. Maybe you can do this in your follow up PR.

aloknerurkar (approved):

PR #3447, opened 2022-10-21
fix(localstore): remove pullIndex on caching

aloknerurkar:
### Checklist

### Description
Currently, the localstore indexes are not reliable to be used for logical components like the reserve, cache etc. This is a known issue and there is a storage rewrite in progress to handle that.

For the current release, in order to test storage incentives, we will change the behaviour to delete the pullIndex when a chunk is put into the gcIndex. This is because, the pullIndex is now used to create the reserve sample. The cache chunks should not be present here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3447)

### DIFF
--- a/pkg/localstore/gc_test.go
+++ b/pkg/localstore/gc_test.go
@@ -119,7 +119,7 @@ func testDBCollectGarbageWorker(t *testing.T) {
 		}
 	}
 
-	t.Run("pull index count", newItemsCountTest(db.pullIndex, int(gcTarget)))
+	t.Run("pull index count", newItemsCountTest(db.pullIndex, 0))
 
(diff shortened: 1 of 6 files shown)


notanatol (approved):

metacertain (approved):

mrekucci (approved):

ldeffenb:
This won't solve expired-stamp pinned chunks being present in the pull-index since they're not going into the gc because of the pin.

janos (approved):

aloknerurkar:
> This won't solve expired-stamp pinned chunks being present in the pull-index since they're not going into the gc because of the pin.

Yes unfortunately. As you know already the pinning is tightly tied to the other features at the moment. The storage rewrite will definitely address this problem. But for the short term, this will help others who are not using the pinning as much.

PR #3453, opened 2022-10-24
wip: feat: stream cache

notanatol:
### Checklist

### Description
libp2p version 0.18 has a heavier (memory wise) stream object, and the suggested approach is to re-use streams as much as possible. This PR introduces a cache for the streams.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3453)

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -44,6 +44,7 @@ import (
 	"github.com/ethersphere/bee/pkg/netstore"
 	"github.com/ethersphere/bee/pkg/p2p"
 	"github.com/ethersphere/bee/pkg/p2p/libp2p"
+	"github.com/ethersphere/bee/pkg/p2p/streamcache"
 	"github.com/ethersphere/bee/pkg/pingpong"
 	"github.com/ethersphere/bee/pkg/pinning"
 	"github.com/ethersphere/bee/pkg/postage"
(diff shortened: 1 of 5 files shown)


vladopajic:
Should there be some logic wrapping around Stream here? Because protocols that obtain Stream in this way will fully close Stream after message has been sent, thus when Stream is later obtained from cache it would be closed,

notanatol:
> Should there be some logic wrapping around Stream here? Because protocols that obtain Stream in this way will fully close Stream after message has been sent, thus when Stream is later obtained from cache it would be closed,

yes, we wrap the bare streams and we add some functionality on top
right now I'm looking into "hijacking" the close (and maybe reset) methods to turn them into no-ops

janos (commented):
It is a very nice and clean cache layer implementation. Unfortunately, sharing libp2p streams is a bit problematic as Read and Write methods have to be synchronous and it is possible to have them called concurrently which usually results with data loss. Basically, all sent data have to be read before new data has to be written to the stream.

notanatol:
This path has proven to be very tricky as libp2p does not work nicely with cached streams. Likely libp2p has a strict lifecycle constrains for streams that are not meant to be interfered by the user.

PR #3456, opened 2022-10-25
chore(pullsync): cancel ruid cleanup

istae:
### Checklist

### Description
PR removes the cancellation ruid and all related code after the latest changes to the puller and pullsync protocols.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3456)

### DIFF
--- a/pkg/pullsync/pullsync.go
+++ b/pkg/pullsync/pullsync.go
@@ -77,16 +77,11 @@ type Syncer struct {
 	validStamp postage.ValidStampFn
 	rate       *rate.Rate
 
-	// ctx    context.Context
-	// cancel func()
-
 	Interface
 	io.Closer
 }
 
 func New(streamer p2p.Streamer, storage pullstorage.Storer, unwrap func(swarm.Chunk), validStamp postage.ValidStampFn, logger log.Logger) *Syncer {
-
-	// ctx, cancel := context.WithCancel(context.Background())
 
 	return &Syncer{
 		streamer:   streamer,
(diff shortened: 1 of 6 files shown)


notanatol (approved):

pkg/pullsync/pullsync.go
```diff
@@ -304,46 +281,20 @@ func (s *Syncer) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) (er
...
 		}
 	}()
-	var ru pb.Ruid
-	if err := r.ReadMsgWithContext(ctx, &ru); err != nil {
-		return fmt.Errorf("send ruid: %w", err)
-	}
-	loggerV2.Debug("peer pulling", "peer_address", p.Address, "ruid", ru.Ruid)
-	ctx, cancel := context.WithCancel(ctx)
+	loggerV2.Debug("peer pulling", "peer_address", p.Address)
+
```
aloknerurkar, line 285: So if the peer goes away and we have active sync workers waiting to restart, but within this time, if the peer comes back and issues a fresh pull sync request, we should check here if we have the workers running, if yes we should clean them up.

istae: what does clean up imply here?

aloknerurkar: What happens to the sync workers that are in progress? How does the handler take those into consideration?

If the peers try to sync again, then kademlia topo signal after the worker wakes up doesn't remove the peer, what will happen?

This ruid handling would cancel those contexts and close them. We would need to do something similar?

istae: If I understood the question:

If there is some hanging makeOffer calls, let's say, because the requesting peer disconnected, the makeOffer, (in the worst case that there are not enough chunks to satisfy the request) will timeout in 15 seconds with the worseCase timer in the falling edge detector, and the next write will error out. If the worry is there could be a leak, this is not the case.

aloknerurkar (approved):

PR #3466, opened 2022-10-26
fix(api): direct upload

### LINKED ISSUE #3227: Direct upload gets stuck in dev mode
### Context
Bee 1.7, macOS

### Summary
I was working on an app that uses direct upload (== deferred: false) for uploading chunks. The app is working with a Bee node running on mainnet, however as soon as I try it in dev mode it gets stuck at the first time it tries to upload a chunk. If the same chunk was already uploaded in deferred mode the upload will succeed though.

### Expected behavior
Either the direct upload should work or it should return an error indicating that the direct upload is not supported in dev mode.

### Steps to reproduce
1. Start Bee in dev mode (e.g. `./bee dev`)
2. Buy a postage stamp (`swarm-cli stamp buy --depth 20 --amount 1000000`)
3. Store the postage batch id in an environment variable called `STAMP` (`export STAMP="your-postage-batch-id"`)
4. Generate empty chunk (`dd if=/dev/zero of=./empty count=8 bs=1`)
5. Upload the empty file with direct mode:
```
curl \
    -X POST \
    -H "Swarm-Deferred-Upload: false" \
    -H "Swarm-Postage-Batch-Id: $STAMP" \
    --data-binary @empty http://localhost:1633/chunks
```

The last step will get stuck forever with Bee printing nothing in the logs. If you use the same command but change the deferred upload to `true` then it will be succesful (and after that direct uploads will succeed as well):

```
 % curl \
    -X POST \
    -H "Swarm-Deferred-Upload: true" \ 
    -H "Swarm-Postage-Batch-Id: $STAMP" \
    --data-binary @empty http://localhost:1633/chunks

{"reference":"b34ca8c22b9e982354f9c7f50b470d66db428d880c8a904d5fe4ec9713171526"}

```

umerm-work:
### Checklist

### Description
Closes #3227

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3466)

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -105,14 +105,15 @@ const (
 )
 
 var (
-	errInvalidNameOrAddress = errors.New("invalid name or bzz address")
-	errNoResolver           = errors.New("no resolver connected")
-	errInvalidRequest       = errors.New("could not validate request")
-	errInvalidContentType   = errors.New("invalid content-type")
-	errDirectoryStore       = errors.New("could not store directory")
-	errFileStore            = errors.New("could not store file")
-	errInvalidPostageBatch  = errors.New("invalid postage batch id")
-	errBatchUnusable        = errors.New("batch not usable")
+	errInvalidNameOrAddress        = errors.New("invalid name or bzz address")
+	errNoResolver                  = errors.New("no resolver connected")
+	errInvalidRequest              = errors.New("could not validate request")
+	errInvalidContentType          = errors.New("invalid content-type")
+	errDirectoryStore              = errors.New("could not store directory")
+	errFileStore                   = errors.New("could not store file")
+	errInvalidPostageBatch         = errors.New("invalid postage batch id")
+	errBatchUnusable               = errors.New("batch not usable")
+	errUnsupportedDevNodeOperation = errors.New("operation not supported in dev mode")
 )
 
 type Service struct {
@@ -780,6 +781,9 @@ func (s *Service) newStamperPutter(r *http.Request) (storage.Storer, func() erro
 		return nil, noopWaitFn, fmt.Errorf("request deferred: %w", err)
 	}
 
+	if !deferred && s.beeMode == DevMode {
+		return nil, noopWaitFn, errUnsupportedDevNodeOperation
+	}
 	exists, err := s.batchStore.Exists(batch)
 	if err != nil {
 		return nil, noopWaitFn, fmt.Errorf("batch exists: %w", err)
--- a/pkg/api/chunk.go
+++ b/pkg/api/chunk.go
@@ -70,6 +70,8 @@ func (s *Service) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 			jsonhttp.NotFound(w, "batch with id not found")
 		case errors.Is(err, errInvalidPostageBatch):
 			jsonhttp.BadRequest(w, "invalid batch id")
+		case errors.Is(err, errUnsupportedDevNodeOperation):
+			jsonhttp.BadRequest(w, errUnsupportedDevNodeOperation)
 		default:
 			jsonhttp.BadRequest(w, nil)
 		}
@@ -161,8 +163,14 @@ func (s *Service) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 
 	if err = wait(); err != nil {
 		s.logger.Debug("chunk upload: sync chunk failed", "error", err)
-		s.logger.Error(nil, "chunk upload: sync chunk failed")
-		jsonhttp.InternalServerError(w, "sync failed")
+		switch {
+		case errors.Is(err, errUnsupportedDevNodeOperation):
+			s.logger.Error(err, "chunk upload: direct upload not supported in dev mode")
+			jsonhttp.BadRequest(w, "dev mode does not support this operation")
+		default:
+			s.logger.Error(err, "chunk upload: sync chunk failed")
+			jsonhttp.InternalServerError(w, "sync failed")
+		}
 		return
 	}
 
--- a/pkg/node/devnode.go
+++ b/pkg/node/devnode.go
@@ -410,7 +410,6 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 		WsPingPeriod:       60 * time.Second,
 		Restricted:         o.Restricted,
 	}, debugOpts, 1, erc20)
-
 	apiService.MountAPI()
 	apiService.SetProbe(probe)
 
@@ -443,7 +442,6 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 		Handler:           apiService,
 		ErrorLog:          stdlog.New(b.errorLogWriter, "", 0),
 	}
-
 	go func() {
 		logger.Info("starting api server", "address", apiListener.Addr())
 
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -124,6 +124,7 @@ type testServerOptions struct {
 	BackendOpts []backendmock.Option
 	Erc20Opts   []erc20mock.Option
 	ChainID     int64
+	beeMode     api.BeeNodeMode
 }
 
 func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.Conn, string, *chanStorer) {
@@ -193,8 +194,11 @@ func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.
 		SyncStatus:       o.SyncStatus,
 		Staking:          o.StakingContract,
 	}
-
-	s := api.New(o.PublicKey, o.PSSPublicKey, o.EthereumAddress, o.Logger, transaction, o.BatchStore, api.FullMode, true, true, backend, o.CORSAllowedOrigins)
+	// by default bee mode is set to full mode
+	if o.beeMode == api.LightMode {
+		o.beeMode = api.FullMode
+	}
+	s := api.New(o.PublicKey, o.PSSPublicKey, o.EthereumAddress, o.Logger, transaction, o.BatchStore, o.beeMode, true, true, backend, o.CORSAllowedOrigins)
 
 	s.SetP2P(o.P2P)
 	s.SetSwarmAddress(&o.Overlay)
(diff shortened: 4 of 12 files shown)


pkg/api/api.go
```diff
@@ -859,6 +860,9 @@ func (p *pushStamperPutter) Put(ctx context.Context, mode storage.ModePut, chs .
 						}
 						return nil
 					}
+					if errors.Is(err, ErrDevNodeNotSupported) {
```
aloknerurkar, line 863: This is not required. This error should be returned anyway.

pkg/api/chunk.go
```diff
@@ -70,6 +70,8 @@ func (s *Service) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 			jsonhttp.NotFound(w, "batch with id not found")
 		case errors.Is(err, errInvalidPostageBatch):
 			jsonhttp.BadRequest(w, "invalid batch id")
+		case errors.Is(err, ErrDevNodeNotSupported):
```
aloknerurkar, line 73: Why are we handling this error here? We only get the error once the putter is used. We should not get this error here.

pkg/node/devnode.go
```diff
@@ -405,12 +405,21 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
...
 
-	apiService.Configure(signer, authenticator, tracer, api.Options{
+	chunkC := apiService.Configure(signer, authenticator, tracer, api.Options{
 		CORSAllowedOrigins: o.CORSAllowedOrigins,
 		WsPingPeriod:       60 * time.Second,
 		Restricted:         o.Restricted,
 	}, debugOpts, 1, erc20)
-
+	go func() {
+		for {
```
aloknerurkar, line 414: This goroutine needs to be stopped on shutdown. Right now its leaking. We need to have select here to select chunk or shutdown signal.

aloknerurkar (changes requested):

pkg/api/chunk.go
umerm-work: Forgot to remove this.

aloknerurkar:
The commit message is not correct. The part in the `()` needs to be the package we are changing. `chunk` is not the correct name. It would be better to use nothing if you are not sure. But ideally this should be `api` package change.

pkg/api/api.go
umerm-work: It's not returning this error. That is why I have to write it explicitly.

pkg/node/devnode.go
umerm-work: @aloknerurkar, where can I get the `shutdown` signal from? Is there already something available in this file?

aloknerurkar: What you can do is use for loop over the channel. Then you can close this channel in the Shutdown function. Please try to go through the code and understand how all this works.

pkg/node/devnode.go
```diff
@@ -405,12 +405,23 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
...
+			select {
+			case ch := <-chunkC:
+				if ch.Direct {
+					ch.Err <- api.ErrDevNodeNotSupported
+					return
```
aloknerurkar, line 419: This is completely wrong. We cannot stop the routine the first time we get chunk on the channel. We need to send error for each chunk.

Also, what is the point of the double return? 

This routine needs to be running till shutdown is called. The shutdown should send some signal for this to stop.

umerm-work: You are right. Corrected this mistake yesterday but still trying to find the shutdown signal.

pkg/node/devnode.go
```diff
@@ -72,6 +72,7 @@ type DevBee struct {
 	errorLogWriter   io.Writer
 	apiServer        *http.Server
 	debugAPIServer   *http.Server
+	quitChunk        chan bool
```
aloknerurkar, line 75: `chan bool` is not required. You can instead use `chan struct{}`. This is more optimal.

pkg/node/devnode.go
```diff
@@ -462,6 +474,7 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 func (b *DevBee) Shutdown() error {
 	var mErr error
 
+	b.quitChunk <- true
```
aloknerurkar, line 477: You can `close(b.quitChunk)` instead of passing value here. It does the same thing.

pkg/node/devnode.go
```diff
@@ -405,12 +406,23 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
...
 	apiService := api.New(mockKey.PublicKey, mockKey.PublicKey, overlayEthAddress, logger, mockTransaction, batchStore, api.DevMode, true, true, chainBackend, o.CORSAllowedOrigins)
 
-	apiService.Configure(signer, authenticator, tracer, api.Options{
+	chunkC := apiService.Configure(signer, authenticator, tracer, api.Options{
 		CORSAllowedOrigins: o.CORSAllowedOrigins,
 		WsPingPeriod:       60 * time.Second,
 		Restricted:         o.Restricted,
 	}, debugOpts, 1, erc20)
-
+	go func() {
```
aloknerurkar, line 414: Add some comment here on why this new goroutine is needed.

pkg/node/devnode.go
aloknerurkar, line 75: This should be called `quitErrPusher` or something. As you are replicating a pusher which returns error.

aloknerurkar (changes requested):

pkg/node/devnode.go
umerm-work: But we have to send a signal to goroutine to close it. If we close it without passing a signal then the code will panic.

aloknerurkar: So you need to understand how the channels work. Even if you close the channel, you can read from it and it will tell you that it is closed.

```
c := make(chan struct{})
close(c)
// this next operation is okay and it will immediately return as the channel is closed. The second parameter
// tells you that it is closed. But we dont care about it as this channel is only used to signal. So no need to write
// anything in it.
<-c
```

This type of pattern is used all over the code. You can follow any closer method implemented by the packages and see.

umerm-work: Thanks for explanation.

pkg/node/devnode.go
```diff
@@ -405,12 +407,24 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
...
 	apiService := api.New(mockKey.PublicKey, mockKey.PublicKey, overlayEthAddress, logger, mockTransaction, batchStore, api.DevMode, true, true, chainBackend, o.CORSAllowedOrigins)
 
-	apiService.Configure(signer, authenticator, tracer, api.Options{
+	chunkC := apiService.Configure(signer, authenticator, tracer, api.Options{
 		CORSAllowedOrigins: o.CORSAllowedOrigins,
 		WsPingPeriod:       60 * time.Second,
 		Restricted:         o.Restricted,
 	}, debugOpts, 1, erc20)
-
+	// this sends signal to the pusher that direct upload is not supported in dev mode
```
aloknerurkar, line 415: This doesnt send signal to pusher. It sends signal to the putter in the API.

mrekucci (changes requested):
I'd suggest testing to see if the same problem also occurs on the `/bytes`, `/bzz`, and `/feed` endpoints. If this is the case, I'd suggest addressing it in the `newStamperPutter` method, where `s.beeMode` can be queried and `deferred` information is available.

umerm-work:
> 
So yes this problem occurs in `/bytes` and `/bzz` endpoints but not sure about `/feed` since it requires an Ethereum address as a param and in dev mode, we don't have any address.

agazso:
> > 
> 
> So yes this problem occurs in `/bytes` and `/bzz` endpoints but not sure about `/feed` since it requires an Ethereum address as a param and in dev mode, we don't have any ethereum address.

The `/feed` requires a feed address, that indeed is the same format as the ethereum address (because it uses the same elliptic curve cryptography to generate addresses), but it is independent of the node's own ethereum address. In fact any valid ethereum address can be used as a feed address.

umerm-work:
> > > 
> > 
> > 
> > So yes this problem occurs in `/bytes` and `/bzz` endpoints but not sure about `/feed` since it requires an Ethereum address as a param and in dev mode, we don't have any ethereum address.
> 
> The `/feed` requires a feed address, that indeed is the same format as the ethereum address (because it uses the same elliptic curve cryptography to generate addresses), but it is independent of the node's own ethereum address. In fact any valid ethereum address can be used as a feed address.

Thanks So `/bytes` , `/bzz` and `/feed` are giving `bad request`  in response and they don't hang the request like in `/chunk` endpoint. Soo just going to add message `dev node not supported`

pkg/api/api.go
```diff
@@ -113,6 +113,7 @@ var (
 	errFileStore            = errors.New("could not store file")
 	errInvalidPostageBatch  = errors.New("invalid postage batch id")
 	errBatchUnusable        = errors.New("batch not usable")
+	ErrDevNodeNotSupported  = errors.New("dev node not supported")
```
mrekucci, line 116: I'd change the error to `errUnsupportedDevNodeOperation  = errors.New("operation not supported in dev mode")`. Also, the error doesn't have to be exported, re-export can be done in the `export_test.go` file if needed in tests.

mrekucci (approved):

pkg/api/chunk.go
```diff
@@ -161,8 +163,14 @@ func (s *Service) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 
 	if err = wait(); err != nil {
 		s.logger.Debug("chunk upload: sync chunk failed", "error", err)
-		s.logger.Error(nil, "chunk upload: sync chunk failed")
-		jsonhttp.InternalServerError(w, "sync failed")
+		switch {
+		case errors.Is(err, errUnsupportedDevNodeOperation):
+			s.logger.Error(nil, "chunk upload: direct upload not supported in dev mode")
```
aloknerurkar, line 168: Would be better to send the err here and in the default case to the log.Error. So we can see it in logs also.

aloknerurkar (approved):

PR #3467, opened 2022-10-27
refactor: keyval pairs for logger in accounting

### LINKED ISSUE #3448: Bad logs from pseudosettle
### Context
bee 1.9.0

### Summary
Logs are showing strange "non-string key:" from accounting's pseudosettle.
```
"time"="2022-10-21 21:36:36.278223" "level"="error" "logger"="node/accounting" "msg"="pseudosettle peer" "error"="allowance expectation refused" "<non-string-key: \"4c300813842c766>"="accepted lower payment than expected"
```

### Expected behavior
New logs should be in the form "key:", "value".

### Actual behavior
See bogus log above.

### Steps to reproduce
You just have to trigger the log somehow!

### Possible solution
All of the logger.* invocations in (at least) pkg/accounting/accounting.go should be reviewed, but the following line is the one I actually saw from my node.
https://github.com/ethersphere/bee/blob/ff9b3749a97cece2d51c6c66089bb0553e026367/pkg/accounting/accounting.go#L1162

swarmHaseeb:
### Checklist

### Description

improved logging
### Related Issue (Optional)

closes #3448 

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3467)

### DIFF
--- a/pkg/accounting/accounting.go
+++ b/pkg/accounting/accounting.go
@@ -365,8 +365,7 @@ func (c *creditAction) Apply() error {
 	c.accounting.metrics.CreditEventsCount.Inc()
 
 	if c.price.Cmp(c.accountingPeer.reservedBalance) > 0 {
-		c.accounting.logger.Debug(ErrOverRelease.Error(), "peer", c.peer, "error", ErrOverRelease)
-		c.accounting.logger.Error(nil, ErrOverRelease.Error())
+		c.accounting.logger.Error(ErrOverRelease, ErrOverRelease.Error(), "peer", c.peer)
 		c.accountingPeer.reservedBalance.SetUint64(0)
 	} else {
 		c.accountingPeer.reservedBalance.Sub(c.accountingPeer.reservedBalance, c.price)
@@ -381,8 +380,7 @@ func (c *creditAction) Apply() error {
 		if increasedExpectedDebtReduced.Cmp(c.accountingPeer.earlyPayment) > 0 {
 			err = c.accounting.settle(c.peer, c.accountingPeer)
 			if err != nil {
-				c.accounting.logger.Debug("failed to settle with credited peer", "peer", c.peer, "error", err)
-				c.accounting.logger.Error(nil, "failed to settle with credited peer")
+				c.accounting.logger.Error(err, "failed to settle with credited peer", "peer", c.peer)
 			}
 		}
 
@@ -425,8 +423,7 @@ func (c *creditAction) Apply() error {
 	if increasedExpectedDebtReduced.Cmp(c.accountingPeer.earlyPayment) > 0 {
 		err = c.accounting.settle(c.peer, c.accountingPeer)
 		if err != nil {
-			c.accounting.logger.Debug("failed to settle with credited peer", "peer", c.peer, "error", err)
-			c.accounting.logger.Error(nil, "failed to settle with credited peer")
+			c.accounting.logger.Error(err, "failed to settle with credited peer", "peer", c.peer)
 		}
 	}
 
@@ -442,8 +439,7 @@ func (c *creditAction) Cleanup() {
 	defer c.accountingPeer.lock.Unlock()
 
 	if c.price.Cmp(c.accountingPeer.reservedBalance) > 0 {
-		c.accounting.logger.Debug("attempting to release more balance than was reserved for peer", "peer", c.peer)
-		c.accounting.logger.Error(nil, "attempting to release more balance than was reserved for peer")
+		c.accounting.logger.Error(ErrOverRelease, ErrOverRelease.Error(), "peer", c.peer)
 		c.accountingPeer.reservedBalance.SetUint64(0)
 	} else {
 		c.accountingPeer.reservedBalance.Sub(c.accountingPeer.reservedBalance, c.price)
@@ -674,8 +670,7 @@ func (a *Accounting) notifyPaymentThresholdUpgrade(peer swarm.Address, accountin
 	// announce new payment threshold to peer
 	err := a.pricing.AnnouncePaymentThreshold(context.Background(), peer, accountingPeer.paymentThresholdForPeer)
 	if err != nil {
-		a.logger.Debug("announcing increased payment threshold", "value", accountingPeer.paymentThresholdForPeer, "peer", peer, "error", err)
-		a.logger.Error(nil, "announcing increased payment threshold")
+		a.logger.Error(err, "announcing increased payment threshold", "value", accountingPeer.paymentThresholdForPeer, "peer", peer)
 	}
 }
 
@@ -992,8 +987,7 @@ func (a *Accounting) NotifyPaymentSent(peer swarm.Address, amount *big.Int, rece
 	currentBalance, err := a.Balance(peer)
 	if err != nil {
 		if !errors.Is(err, ErrPeerNoBalance) {
-			a.logger.Debug("notify payment sent; failed to persist balance", "error", err)
-			a.logger.Error(nil, "notify payment sent; failed to persist balance")
+			a.logger.Error(err, "notify payment sent; failed to persist balance")
 			return
 		}
 	}
@@ -1005,8 +999,7 @@ func (a *Accounting) NotifyPaymentSent(peer swarm.Address, amount *big.Int, rece
 
 	err = a.store.Put(peerBalanceKey(peer), nextBalance)
 	if err != nil {
-		a.logger.Debug("notify payment sent; failed to persist balance", "error", err)
-		a.logger.Error(nil, "notify payment sent; failed to persist balance")
+		a.logger.Error(err, "notify payment sent; failed to persist balance")
 		return
 	}
 
@@ -1140,8 +1133,7 @@ func (a *Accounting) NotifyRefreshmentSent(peer swarm.Address, attemptedAmount, 
 			a.metrics.AccountingDisconnectsEnforceRefreshCount.Inc()
 			_ = a.blocklist(peer, 1, "failed to refresh")
 		}
-		a.logger.Debug("notifyrefreshmentsent failed to refresh", "error", receivedError)
-		a.logger.Error(nil, "notifyrefreshmentsent failed to refresh")
+		a.logger.Error(receivedError, "notifyrefreshmentsent failed to refresh")
 		return
 	}
 
@@ -1167,8 +1159,7 @@ func (a *Accounting) NotifyRefreshmentSent(peer swarm.Address, attemptedAmount, 
 	// compare received refreshment amount to expectation
 	if expectedAllowance.Cmp(amount) > 0 {
 		// if expectation is not met, blocklist peer
-		a.logger.Debug("accepted lower payment than expected", "pseudosettle peer", peer.String(), "error", ErrEnforceRefresh)
-		a.logger.Error(nil, "accepted lower payment than expected")
+		a.logger.Error(ErrEnforceRefresh, "accepted lower payment than expected", "pseudosettle peer", peer.String())
 		a.metrics.ErrRefreshmentBelowExpected.Inc()
 		_ = a.blocklist(peer, 1, "failed to meet expectation for allowance")
 		return
@@ -1178,8 +1169,7 @@ func (a *Accounting) NotifyRefreshmentSent(peer swarm.Address, attemptedAmount, 
 	currentBalance, err := a.Balance(peer)
 	if err != nil {
 		if !errors.Is(err, ErrPeerNoBalance) {
-			a.logger.Debug("notifyrefreshmentsent failed to get balance", "error", err)
-			a.logger.Error(nil, "notifyrefreshmentsent failed to get balance")
+			a.logger.Error(err, "notifyrefreshmentsent failed to get balance")
 			return
 		}
 	}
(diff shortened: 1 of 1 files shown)


swarmHaseeb:
failed tests are unrelated

pkg/accounting/accounting.go
```diff
@@ -380,7 +381,8 @@ func (c *creditAction) Apply() error {
 		if increasedExpectedDebtReduced.Cmp(c.accountingPeer.earlyPayment) > 0 {
 			err = c.accounting.settle(c.peer, c.accountingPeer)
 			if err != nil {
-				c.accounting.logger.Error(err, "failed to settle with credited peer", c.peer)
+				c.accounting.logger.Debug("failed to settle with credited peer", "peer", c.peer, "error", err)
+				c.accounting.logger.Error(nil, "failed to settle with credited peer")
```
ldeffenb, line 385: If you've got an err and you're doing a logger.Error, why not provide err as the first argument instead of nil?     This applies to lots of similar updates throughout the file.   If you're doing the log anyway, why not provide the visibility to the underlying error that triggered the log?

pkg/accounting/accounting.go
```diff
@@ -987,7 +992,8 @@ func (a *Accounting) NotifyPaymentSent(peer swarm.Address, amount *big.Int, rece
 	currentBalance, err := a.Balance(peer)
 	if err != nil {
 		if !errors.Is(err, ErrPeerNoBalance) {
-			a.logger.Error(err, "notify payment sent; failed to load balance")
+			a.logger.Debug("notify payment sent; failed to persist balance", "error", err)
+			a.logger.Error(nil, "notify payment sent; failed to persist balance")
```
ldeffenb, line 996: If you put the err in the logger.Error invocation, then you don't need the redundant logger.Debug since it adds no additional information.

pkg/accounting/accounting.go
```diff
@@ -999,7 +1005,8 @@ func (a *Accounting) NotifyPaymentSent(peer swarm.Address, amount *big.Int, rece
 
 	err = a.store.Put(peerBalanceKey(peer), nextBalance)
 	if err != nil {
-		a.logger.Error(err, "notify payment sent; failed to persist balance")
+		a.logger.Debug("notify payment sent; failed to persist balance", "error", err)
+		a.logger.Error(nil, "notify payment sent; failed to persist balance")
```
ldeffenb, line 1009: Another redundant logger.Debug if you put the err in the logger.Error.

pkg/accounting/accounting.go
```diff
@@ -1133,7 +1140,8 @@ func (a *Accounting) NotifyRefreshmentSent(peer swarm.Address, attemptedAmount,
 			a.metrics.AccountingDisconnectsEnforceRefreshCount.Inc()
 			_ = a.blocklist(peer, 1, "failed to refresh")
 		}
-		a.logger.Error(receivedError, "notifyrefreshmentsent failed to refresh")
+		a.logger.Debug("notifyrefreshmentsent failed to refresh", "error", receivedError)
+		a.logger.Error(nil, "notifyrefreshmentsent failed to refresh")
```
ldeffenb, line 1144: Ditto again, but I'll quit mentioning it at this point as it all hinges on why the error (receivedError in this case) is not provided to the logger.Error invocation.

ldeffenb (changes requested):
The logger.Debug invocations look like they provide the proper pairing of key/value pairs, but I think we're not using logger.Error as well as it could be used, especially where it can eliminate the redundant logger.Debug that only provides the err as the key/value.

pkg/accounting/accounting.go
swarmHaseeb: so umm if you check out other packages you will see that this is kind of a standard where debug contains all of the information but err log contains a general error. I am not sure why we did it the way we did it (could be that we do not want to leak unnecessary info in error, @mrekucci would be better person to explain this) but changing the way its done within pr will no longer look like a standard anymore and maybe if it is required a better way should be another pr focused on refactoring all of those erros. 
Please have a look
https://github.com/ethersphere/bee/blob/master/pkg/api/balances.go#L77
https://github.com/ethersphere/bee/blob/master/pkg/api/stewardship.go#L56
https://github.com/ethersphere/bee/blob/master/pkg/api/postage.go#L101

pkg/accounting/accounting.go
swarmHaseeb: comment added below

pkg/accounting/accounting.go
swarmHaseeb: comment added below

pkg/accounting/accounting.go
ldeffenb: I've seen that tendency in the other packages and quite frankly don't really understand it.   I don't see how the err itself would leak info, but it certainly goes a long way to knowing what's going on without cranking the verbosity up to debug.

But I'm not one to fight city hall, and the PR definitely fixes the original issue, so I'll mark it approved.

ldeffenb (approved):
Keeping in mind my comment below, this looks like it fixes the issue.

https://github.com/ethersphere/bee/pull/3467#discussion_r1006551627

pkg/accounting/accounting.go
swarmHaseeb: ill discuss this with @mrekucci once he gets back from vacation. Ty :)

istae: I also would like to know why we are doing this :) ? The API has a lot of these.

pkg/accounting/accounting.go
```diff
@@ -365,7 +365,8 @@ func (c *creditAction) Apply() error {
 	c.accounting.metrics.CreditEventsCount.Inc()
 
 	if c.price.Cmp(c.accountingPeer.reservedBalance) > 0 {
-		c.accounting.logger.Error(ErrOverRelease, "for peer %v")
+		c.accounting.logger.Debug(ErrOverRelease.Error(), "peer", c.peer, "error", ErrOverRelease)
```
aloknerurkar, line 368: Accounting calls happen way to often. Adding duplicate debug and error logs here is not right. The logs would go out of control. API package has different requirements.

Remove all the debug logs here. You can add the extra information in the error logs itself.

ldeffenb: > Remove all the debug logs here. You can add the extra information in the error logs itself.

Given that, then any err can go to the first argument of the .Error log instead of being an additional key/value pair?

aloknerurkar: Yes, error should go into the first argument. If any additional key-value pairs need to be added we can always add them.

pkg/accounting/accounting.go
```diff
@@ -365,7 +365,7 @@ func (c *creditAction) Apply() error {
 	c.accounting.metrics.CreditEventsCount.Inc()
 
 	if c.price.Cmp(c.accountingPeer.reservedBalance) > 0 {
-		c.accounting.logger.Error(ErrOverRelease, "for peer %v")
+		c.accounting.logger.Error(ErrOverRelease, ErrOverRelease.Error(), "peer", c.peer)
```
istae, line 368: ErrOverRelease.Error() would cause the error string to be printed twice

pkg/accounting/accounting.go
```diff
@@ -439,7 +439,7 @@ func (c *creditAction) Cleanup() {
 	defer c.accountingPeer.lock.Unlock()
 
 	if c.price.Cmp(c.accountingPeer.reservedBalance) > 0 {
-		c.accounting.logger.Error(nil, "attempting to release more balance than was reserved for peer")
+		c.accounting.logger.Error(ErrOverRelease, ErrOverRelease.Error(), "peer", c.peer)
```
istae, line 442: same here, I would replace ErrOverRelease.Error() with some meaningful msg

istae (changes requested):

pkg/accounting/accounting.go
mrekucci: Correct, this `ErrOverRelease.Error()` should be some meaningful log message.

pkg/accounting/accounting.go
```diff
@@ -1159,7 +1159,7 @@ func (a *Accounting) NotifyRefreshmentSent(peer swarm.Address, attemptedAmount,
 	// compare received refreshment amount to expectation
 	if expectedAllowance.Cmp(amount) > 0 {
 		// if expectation is not met, blocklist peer
-		a.logger.Error(ErrEnforceRefresh, "pseudosettle peer", peer, "accepted lower payment than expected")
+		a.logger.Error(ErrEnforceRefresh, "accepted lower payment than expected", "pseudosettle peer", peer.String())
```
mrekucci, line 1162: Unnecessary to call `.String()`. The error should also be `nil` instead of `ErrEnforceRefresh` because the caller did not return any error.

mrekucci (commented):
I would suggest using `peer_address` instead of `peer` (or relevant to the context) as the key for the logger, see other log messages. Also, this PR doesn't fix the following log message: `a.logger.Warning("accounting: notifyrefreshmentsent failed to decrease originated balance: %v", err)`.

pkg/accounting/accounting.go
swarmHaseeb: it should be nil as Peter said because the error was not returned to us, so we can simply extract the string from error using .Error()

pkg/accounting/accounting.go
```diff
@@ -365,7 +365,7 @@ func (c *creditAction) Apply() error {
 	c.accounting.metrics.CreditEventsCount.Inc()
 
 	if c.price.Cmp(c.accountingPeer.reservedBalance) > 0 {
-		c.accounting.logger.Error(ErrOverRelease, "for peer %v")
+		c.accounting.logger.Error(nil, ErrOverRelease.Error(), "peer_address", c.peer)
```
istae, line 368: ErrOverRelease.Error() should be changed with a meaningful error message

pkg/accounting/accounting.go
```diff
@@ -439,7 +439,7 @@ func (c *creditAction) Cleanup() {
 	defer c.accountingPeer.lock.Unlock()
 
 	if c.price.Cmp(c.accountingPeer.reservedBalance) > 0 {
-		c.accounting.logger.Error(nil, "attempting to release more balance than was reserved for peer")
+		c.accounting.logger.Error(nil, ErrOverRelease.Error(), "peer_address", c.peer)
```
istae, line 442: ErrOverRelease.Error() should be changed with a meaningful error message

pkg/accounting/accounting.go
```diff
@@ -1159,7 +1159,7 @@ func (a *Accounting) NotifyRefreshmentSent(peer swarm.Address, attemptedAmount,
 	// compare received refreshment amount to expectation
 	if expectedAllowance.Cmp(amount) > 0 {
 		// if expectation is not met, blocklist peer
-		a.logger.Error(ErrEnforceRefresh, "pseudosettle peer", peer, "accepted lower payment than expected")
+		a.logger.Error(nil, ErrEnforceRefresh.Error(), "pseudosettle peer", peer)
```
istae, line 1162: ErrOverRelease.Error() should be changed with a meaningful error message

istae (changes requested):
please address the comment that a meaningful msg should be put in place of ErrOverRelease.Error()

pkg/accounting/accounting.go
istae: you have not resolved this :(

pkg/accounting/accounting.go
istae: @mrekucci why are we doing things this way?

pkg/accounting/accounting.go
```diff
@@ -439,7 +439,7 @@ func (c *creditAction) Cleanup() {
 	defer c.accountingPeer.lock.Unlock()
 
 	if c.price.Cmp(c.accountingPeer.reservedBalance) > 0 {
-		c.accounting.logger.Error(nil, "attempting to release more balance than was reserved for peer")
+		c.accounting.logger.Error(nil, "allowance expectation refused", "peer_address", c.peer)
```
istae, line 442: why can't we go with "attempting to release more balance than was reserved for peer"

pkg/accounting/accounting.go
```diff
@@ -1159,7 +1159,7 @@ func (a *Accounting) NotifyRefreshmentSent(peer swarm.Address, attemptedAmount,
 	// compare received refreshment amount to expectation
 	if expectedAllowance.Cmp(amount) > 0 {
 		// if expectation is not met, blocklist peer
-		a.logger.Error(ErrEnforceRefresh, "pseudosettle peer", peer, "accepted lower payment than expected")
+		a.logger.Error(nil, "allowance expectation refused", "pseudosettle peer", peer)
```
istae, line 1162: why can't we go with "accepted lower payment than expected"

swarmHaseeb: but if it is not an error, and if error field is not being used, then why use Error in first place

istae: go with 
`a.logger.Error(nil, "accepted lower payment than expected", "pseudosettle_peer", peer)`

swarmHaseeb: > go with `a.logger.Error(nil, "accepted lower payment than expected", "pseudosettle_peer", peer)`

sure but 2 lines below you are blocklisting a peer for a different message so just putting this here

istae: the intention here is that with the Error func it has a highest logging level,  so if a user turns off Debug or warning level logs, this error msg won't show.

And we go with a nil error because it was not an error we "caught" from a function call or something

swarmHaseeb: yes then it might not be an error. but anyways, ive made the changes

istae (approved):

PR #3471, opened 2022-10-28
chore: sampler logs fix and agent timer limit helper func

istae:
### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3471)

### DIFF
--- a/pkg/localstore/sampler.go
+++ b/pkg/localstore/sampler.go
@@ -196,18 +196,18 @@ func (db *DB) ReserveSample(
 
 			stampData, err := stamp.MarshalBinary()
 			if err != nil {
-				logger.Debug("error marshaling stamp for chunk", "address", chunk.Address(), "error", err)
+				logger.Debug("error marshaling stamp for chunk", "chunk_address", chunk.Address(), "error", err)
 				continue
 			}
 			_, err = db.validStamp(chunk, stampData)
 			if err == nil {
 				if !cac.Valid(chunk) && !soc.Valid(chunk) {
-					logger.Debug("data invalid for chunk address", "address", chunk.Address())
+					logger.Debug("data invalid for chunk address", "chunk_address", chunk.Address())
 				} else {
 					insert(item.transformedAddress)
 				}
 			} else {
-				logger.Info("invalid stamp for chunk", "address", chunk.Address(), "error", err)
+				logger.Info("invalid stamp for chunk", "chunk_address", chunk.Address(), "error", err)
 			}
 
 		}
@@ -232,7 +232,7 @@ func (db *DB) ReserveSample(
 		Items: sampleItems,
 		Hash:  swarm.NewAddress(hash),
 	}
-	logger.Info("sampler done", "stats", stat.String(), "sample", sample)
+	logger.Info("sampler done", "stats", stat, "sample", sample)
 
 	return sample, nil
 }
--- a/pkg/localstore/sampler.go
+++ b/pkg/localstore/sampler.go
@@ -196,18 +196,18 @@ func (db *DB) ReserveSample(
 
 			stampData, err := stamp.MarshalBinary()
 			if err != nil {
-				logger.Info("error marshaling stamp for chunk", chunk.Address(), err)
+				logger.Debug("error marshaling stamp for chunk", "address", chunk.Address(), "error", err)
 				continue
 			}
 			_, err = db.validStamp(chunk, stampData)
 			if err == nil {
 				if !cac.Valid(chunk) && !soc.Valid(chunk) {
-					logger.Info("data invalid for chunk address", chunk.Address())
+					logger.Debug("data invalid for chunk address", "address", chunk.Address())
 				} else {
 					insert(item.transformedAddress)
 				}
 			} else {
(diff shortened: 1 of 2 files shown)


pkg/storageincentives/agent.go
```diff
@@ -244,11 +244,11 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 		// compute the current phase
 		p := block % blocksPerRound
 		if p < blocksPerPhase {
-			currentPhase = commit // [0, 37]
-		} else if p >= blocksPerPhase && p < 2*blocksPerPhase { // [38, 76]
+			currentPhase = commit // [0, 38)
```
ldeffenb, line 247: These range comments are incorrect.  A given block cannot be in multiple phases.   commit is 0-37 (< blocks/4), reveal is 38-75 (< blocks/2) and claim is 76-151.  There is no 152 remainder when doing a mod on 152.    The code matches the newly deployed contract, but the comments don't match the code (unless you go with non-inclusive on the right).

From the contract at https://goerli.etherscan.io/address/0x2e4ccEb9a88bcfbC7779Fa693714783A345c1343#code
```
    function currentPhaseCommit() public view returns (bool){
        if ( block.number % roundLength < roundLength / 4 ) {
            return true;
        }
        return false;
    }

    function currentPhaseReveal() public view returns (bool){
        uint256 number = block.number % roundLength;
        if ( number >= roundLength / 4 && number < roundLength / 2 ) {
            return true;
        }
        return false;
    }

    function currentPhaseClaim() public view returns (bool){
        if ( block.number % roundLength >= roundLength / 2 ) {
            return true;
        }
        return false;
    }
```

ldeffenb (changes requested):

pkg/storageincentives/agent.go
istae: hi @ldeffenb, thanks for the comment. 

The reveal phase had 39 blocks in the previous iteration of the contract. 
We redeployed the contract with some fixes, which includes making every phase the same length. 

But yes, if we hadn't changed the contract, you are absolutely correct.

istae: The upcoming release will have the new contract addresses.

istae: commit is inclusive 0, exclusive 38 [0,38)
reveal is inclusive 38, exclusive 76 [38, 76)
claim is inclusive 76, exclusive 152 [76, 152)

in other words, [0, 37], [38, 75], [76, 151]

maybe I am missing something, but this looks correct and is aligned with the code and the new contract.

ldeffenb: The comments say the following:
commit: 0, 38
reveal: 38, 76
claim: 76, 152
Which is correct if they are going non-inclusive on the end value.   But to a first glance, it appears that blocks offset 38 and 76 are actually in two phases when they are not.  And it also implies that the % (mod) function can return 152, which it cannot.  
I'm suggesting that comments reflect the actual round/block offsets of:
commit: 0, 37 (38 blocks)
reveal: 38, 75 (38 blocks)
claim: 76, 151 (76 blocks)
Possibly even adding the parentheticals for 100% clarity to the reader.

notanatol (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ethersphere/bee/3471)* status: all files reviewed, 1 unresolved discussion (waiting on @aloknerurkar and @istae)

pkg/localstore/sampler.go
```diff
@@ -232,7 +232,7 @@ func (db *DB) ReserveSample(
 		Items: sampleItems,
 		Hash:  swarm.NewAddress(hash),
 	}
-	logger.Info("Sampler done", "Stats", stat.String(), "Sample", sample)
+	logger.Info("sampler done", "stats", stat.String(), "sample", sample)
```
mrekucci, line 235: The `.String()` call on `stat` is unnecessary.

mrekucci (commented):
I would suggest using `chunk_address` instead of `address` as the key in log messages because it's more contextual.

pkg/storageincentives/agent.go
```diff
@@ -331,31 +329,40 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
...
+	timeLimiterBlock, err := a.backend.HeaderByNumber(ctx, new(big.Int).SetUint64(previousRoundBlockNumber))
 	if err != nil {
-		return 0, nil, err
+		a.metrics.BackendErrors.Inc()
+		return 0, err
 	}
-	a.metrics.SampleDuration.Set(time.Since(t).Seconds())
 
-	return storageRadius, sample.Hash.Bytes(), nil
+	return time.Duration(timeLimiterBlock.Time) * time.Second / time.Nanosecond, nil
```
mrekucci, line 365: Not sure about this part `time.Second / time.Nanosecond`. Since `time.Nanosecond` is equal to `1`, this will always return time in seconds. Is that the intention? If so, then `time.Nanosecond` can be omitted.

mrekucci (approved):

PR #3476, opened 2022-10-31
feat: delimit based on remaining balance

metacertain:
### Checklist

### Description
If a batch of stamps expires while reserve sampling is done, its exclusion is probabilistic.
This PR attempts to create a consensus on minimum batch balance in order to be included in the RC sampling.

#### Motivation and Context 
Circumstances of content expiring in large volume during a round can be a source of discrepancy between sampling results of different nodes. This change of convention to exclude data that is expected to expire can also create a discrepancy if mixed with legacy behaviour, so it's best to add this with a release that changes the redistribution contract address.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3476)

### DIFF
--- a/pkg/postage/batchstore/reserve.go
+++ b/pkg/postage/batchstore/reserve.go
@@ -125,7 +125,7 @@ func (s *store) cleanup() error {
 	return nil
 }
 
-func (s *store) GetBatchIDsExpiringUntil(until *big.Int) (map[string]bool, error) {
+func (s *store) GetBatchIDsBelowValue(until *big.Int) (map[string]bool, error) {
 
 	s.mtx.RLock()
 	defer s.mtx.RUnlock()
--- a/pkg/postage/batchstore/reserve.go
+++ b/pkg/postage/batchstore/reserve.go
@@ -35,6 +35,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"math"
+	"math/big"
 
 	"github.com/ethersphere/bee/pkg/postage"
 )
@@ -122,6 +123,36 @@ func (s *store) cleanup() error {
 	}
 
 	return nil
+}
+
+func (s *store) GetBatchIDsExpiringUntil(until *big.Int) (map[string]bool, error) {
+
+	s.mtx.RLock()
+	defer s.mtx.RUnlock()
+
+	exclude := make(map[string]bool)
+
+	err := s.store.Iterate(valueKeyPrefix, func(key, value []byte) (stop bool, err error) {
+
+		b, err := s.get(valueKeyToID(key))
+		if err != nil {
+			return false, err
+		}
+
+		// batches whose balance is below the total cumulative payout
+		if b.Value.Cmp(until) < 0 {
+			exclude[string(b.ID)] = true
+		} else {
+			return true, nil // stop early as an optimization at first value above the total cumulative payout
+		}
+
+		return false, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return exclude, nil
 }
 
 // computeRadius calculates the radius by using the sum of all batch depths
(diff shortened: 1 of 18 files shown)


istae (changes requested):
Another approach is using the current price, calculate the minimum remaining balance a batch needs to have in order to not expire before the end of the claim phase.

We know how long in minutes the claim phase is, we know the price, so it's easy to calculate the minimum balance.

Now, the second change I'd like to see if, we should iterate the batchstore in ascending order and collect the list of problematic batches and pass this to the ReserveSample.

The list can be generated by batchstore pkg where we pass x time as an argument, and the batchstore pkg returns the list of batches that will exprire within x.

x in the case is the amount of seconds until the end of the claim phase. 

so the flow is agent sends x to batchstore,
batchstore returns the list of batches
the agent calls ReserveSample with the list

istae:
This way, batchstore does not need to be a dependency for localstore as well.

metacertain:
that would have been our first approach, however, the price potentially changes in the sampling period, as the claim just has that effect, and depending on whether the sampling starts before or after this change you might get different ideas of what the common minimum balance for the round is.
to resolve this, we save the last consensually known price in the reveal phase (when price is not supposed to change) and use that

zelig:
what is the status if ths PR? IS @istae 's alternative implemented?  

AFAIK the issue is that no matter how we calculate expiring batches and what time limit we agree on for their  expiry, the price might change during sampling but could potentially be set before sampling starts.

so how about defines sampled set to include chunks that would servive the next round with  the last consensual price taken from the previous claim         ?

istae:
I would hold on to this PR until @umerm-work 's PR is merged

pkg/postage/batchstore/reserve.go
```diff
@@ -124,6 +125,36 @@ func (s *store) cleanup() error {
 	return nil
 }
 
+func (s *store) GetBatchIDsExpiringUntil(until *big.Int) (map[string]bool, error) {
```
istae, line 128: better name is `GetBatchIDsBelowValue`

pkg/storageincentives/agent.go
```diff
@@ -196,13 +215,13 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 		}
 	})
 
-	phaseEvents.On(sample, func(ctx context.Context, _ PhaseType) {
+	phaseEvents.On(sample, func(ctx context.Context, block uint64, _ PhaseType) {
 
 		mtx.Lock()
```
istae, line 220: instead of passing the block to the events publisher, 
you can have a new `block` var at the top of the start function, similar to what we did with the `round` var

pkg/storageincentives/agent.go
```diff
@@ -387,7 +406,17 @@ func (a *Agent) play(ctx context.Context, round uint64) (uint8, []byte, error) {
...
+	nextRoundBeginningBlock := nextRoundNumber * a.blocksPerRound
+
+	a.latestPriceMtx.Lock()
+
+	difference := nextRoundBeginningBlock - a.latestPriceBlock
+	minimumBalance := new(big.Int).Add(a.latestTotalPayout, new(big.Int).Mul(a.latestPrice, big.NewInt(int64(difference))))
+
+	a.latestPriceMtx.Unlock()
+
+	sample, err := a.sampler.ReserveSample(ctx, salt, storageRadius, uint64(timeLimiter), minimumBalance)
```
istae, line 419: it'd be cleaner to just pass the list of batchIDs to ignore, instead of minimum balance

pkg/storageincentives/agent.go
```diff
@@ -387,7 +406,17 @@ func (a *Agent) play(ctx context.Context, round uint64) (uint8, []byte, error) {
 		return 0, nil, err
 	}
 
-	sample, err := a.sampler.ReserveSample(ctx, salt, storageRadius, uint64(timeLimiter))
+	nextRoundNumber := (block / a.blocksPerRound) + 1
```
istae, line 409: let's move these calculations and the producing the sample to a new function

vladopajic:
Constraint with minimum batch balance for reserve sample was implemented in #4055.

PR #3486, opened 2022-11-02
feat: stamps contract expire loop

swarmHaseeb:
### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3486)

### DIFF
--- a/pkg/postage/postagecontract/contract.go
+++ b/pkg/postage/postagecontract/contract.go
@@ -356,12 +356,12 @@ func (c *postageContract) hasExpired(ctx context.Context) (bool, error) {
 }
 
 func (c *postageContract) expireLimitedBatches(ctx context.Context) error {
-	callData, err := postageStampABI.Pack("expireLimited", big.NewInt(20))
+	callData, err := postageStampABI.Pack("expireLimited", big.NewInt(50))
 	if err != nil {
 		return err
 	}
 
-	_, err = c.sendTransaction(sctx.SetGasLimit(ctx, 500_000), callData, "expire limited batches")
+	_, err = c.sendTransaction(sctx.SetGasLimit(ctx, 1_100_000), callData, "expire limited batches")
 	if err != nil {
 		return fmt.Errorf("expire limited batches: %w", err)
 	}
--- a/pkg/postage/postagecontract/contract.go
+++ b/pkg/postage/postagecontract/contract.go
@@ -49,7 +49,10 @@ type Interface interface {
 	CreateBatch(ctx context.Context, initialBalance *big.Int, depth uint8, immutable bool, label string) ([]byte, error)
 	TopUpBatch(ctx context.Context, batchID []byte, topupBalance *big.Int) error
 	DiluteBatch(ctx context.Context, batchID []byte, newDepth uint8) error
-	ExpireBatches(ctx context.Context) error
+}
+
+type PostageBatchExpirer interface {
+	ExpireBatches(context.Context) error
 }
 
 type postageContract struct {
@@ -311,24 +314,25 @@ func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topUpA
 }
 
 func (c *postageContract) ExpireBatches(ctx context.Context) error {
-	count, err := c.countExpiredBatches(ctx)
+	hasExpired, err := c.hasExpired(ctx)
 	if err != nil {
 		return fmt.Errorf("count expired batches: %w", err)
 	}
 
-	for i := 0; i < int(count.Int64()); i += 100 {
+	if hasExpired {
 		err := c.expireLimitedBatches(ctx, big.NewInt(100))
 		if err != nil {
 			return fmt.Errorf("expire batches: %w", err)
 		}
 	}
+
 	return nil
 }
 
-func (c *postageContract) countExpiredBatches(ctx context.Context) (*big.Int, error) {
-	callData, err := postageStampABI.Pack("countExpiry")
+func (c *postageContract) hasExpired(ctx context.Context) (bool, error) {
+	callData, err := postageStampABI.Pack("hasExpired")
 	if err != nil {
-		return nil, err
+		return false, err
 	}
 
 	result, err := c.transactionService.Call(ctx, &transaction.TxRequest{
@@ -336,14 +340,14 @@ func (c *postageContract) countExpiredBatches(ctx context.Context) (*big.Int, er
 		Data: callData,
 	})
 	if err != nil {
-		return nil, err
+		return false, err
 	}
 
-	results, err := postageStampABI.Unpack("countExpiry", result)
+	results, err := postageStampABI.Unpack("hasExpired", result)
 	if err != nil {
-		return nil, err
+		return false, err
 	}
-	return abi.ConvertType(results[0], new(big.Int)).(*big.Int), nil
+	return results[0].(bool), nil
 }
 
 func (c *postageContract) expireLimitedBatches(ctx context.Context, count *big.Int) error {
@@ -443,7 +447,7 @@ func LookupERC20Address(ctx context.Context, transactionService transaction.Serv
 
 type noOpPostageContract struct{}
 
-func (m *noOpPostageContract) ExpireBatches(ctx context.Context) error {
+func (m *noOpPostageContract) ExpireBatches(context.Context) error {
 	return ErrChainDisabled
 }
 
--- a/pkg/postage/postagecontract/contract.go
+++ b/pkg/postage/postagecontract/contract.go
@@ -313,13 +313,13 @@ func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topUpA
 func (c *postageContract) ExpireBatches(ctx context.Context) error {
 	count, err := c.countExpiredBatches(ctx)
 	if err != nil {
-		return err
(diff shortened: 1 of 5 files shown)


pkg/postage/postagecontract/contract.go
```diff
@@ -305,7 +312,68 @@ func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topUpA
 	return ErrBatchTopUp
 }
 
+func (c *postageContract) ExpireBatches(ctx context.Context) error {
+	count, err := c.CountExpiredBatches(ctx)
+	if err != nil {
+		return err
+	}
+	for i := new(big.Int).Set(big.NewInt(1)); i.Cmp(count) <= 0; i.Add(i, big.NewInt(100)) {
```
istae, line 320: it would be easier to ready by first converting `count` to int

pkg/postage/postagecontract/contract.go
```diff
@@ -305,7 +312,68 @@ func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topUpA
...
+		err := c.ExpireLimitedBatches(ctx, big.NewInt(100))
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *postageContract) CountExpiredBatches(ctx context.Context) (*big.Int, error) {
```
istae, line 329: no need to be a public function

pkg/postage/postagecontract/contract.go
```diff
@@ -305,7 +312,68 @@ func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topUpA
...
+	}
+
+	results, err := postageStampABI.Unpack("countExpiry", result)
+	if err != nil {
+		return nil, err
+	}
+	return abi.ConvertType(results[0], new(big.Int)).(*big.Int), nil
+}
+
+func (c *postageContract) ExpireLimitedBatches(ctx context.Context, count *big.Int) error {
```
istae, line 350: this does not need to be a public function.

pkg/postage/postagecontract/contract.go
```diff
@@ -270,6 +273,10 @@ func (c *postageContract) CreateBatch(ctx context.Context, initialBalance *big.I
 }
 
 func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topUpAmount *big.Int) error {
+	err := c.ExpireBatches(ctx)
```
istae, line 276: top up does not need expire batches

istae (changes requested):
nice work!
the agent and the create batch calls are missing the expire batches call.

istae (changes requested):
the agent is missing the call the expire batches when it enter the claim phase

pkg/postage/postagecontract/contract.go
```diff
@@ -207,6 +208,10 @@ func (c *postageContract) getBalance(ctx context.Context) (*big.Int, error) {
 }
 
 func (c *postageContract) CreateBatch(ctx context.Context, initialBalance *big.Int, depth uint8, immutable bool, label string) ([]byte, error) {
+	err := c.ExpireBatches(ctx)
+	if err != nil {
+		return nil, err
```
istae, line 213: this error should be wrapper with a meaningful error msg, like 
`fmt.Errorf("expire batches: %w", err)`

istae: and every else this is used

swarmHaseeb: so if we wrap it within ExpireBatches fn we probably do not need to wrap it anywhere else since the wrapped message will be returned nonetheless if i am correct.

istae (approved):

notanatol (approved):

pkg/postage/postagecontract/contract.go
```diff
@@ -49,6 +49,7 @@ type Interface interface {
 	CreateBatch(ctx context.Context, initialBalance *big.Int, depth uint8, immutable bool, label string) ([]byte, error)
 	TopUpBatch(ctx context.Context, batchID []byte, topupBalance *big.Int) error
 	DiluteBatch(ctx context.Context, batchID []byte, newDepth uint8) error
+	ExpireBatches(ctx context.Context) error
```
aloknerurkar, line 52: If we only have to export this single functionality, we should ideally create a new interface
```
type PostageBatchExpirer interface {
        ExpireBatches(context.Context) error
}
```

The storage incentives agent only needs to know about this single interface and not the whole contract.

pkg/postage/postagecontract/contract.go
```diff
@@ -305,7 +310,69 @@ func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topUpA
 	return ErrBatchTopUp
 }
 
+func (c *postageContract) ExpireBatches(ctx context.Context) error {
```
aloknerurkar, line 313: These functions are added in the postage contract. So they need to be tested here. This code is not being tested in the storage incentives agent tests as you use mock logic there.

This is not the first time that you have missed code coverage completely.

pkg/storageincentives/agent_test.go
```diff
@@ -92,8 +94,8 @@ func TestAgent(t *testing.T) {
 				incrementBy: tc.incrementBy,
 				block:       tc.blocksPerRound}
 			contract := &mockContract{t: t, baseAddr: addr}
-
-			service := createService(addr, backend, contract, uint64(tc.blocksPerRound), uint64(tc.blocksPerPhase))
+			postageContract := contractMock.New()
+			service := createService(addr, backend, contract, postageContract, uint64(tc.blocksPerRound), uint64(tc.blocksPerPhase))
```
aloknerurkar, line 98: If we only expose the `PostageBatchExpirer` interface I mentioned. We dont need to instantiate the contract here.

aloknerurkar (changes requested):

pkg/postage/postagecontract/contract.go
```diff
@@ -305,7 +310,69 @@ func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topUpA
...
+	count, err := c.countExpiredBatches(ctx)
+	if err != nil {
+		return fmt.Errorf("count expired batches: %w", err)
+	}
+
+	for i := 0; i < int(count.Int64()); i += 100 {
```
mrekucci, line 319: Casting `int64` to `int` can cause a problem on 32-bit architectures if the value represented is above the `int32` range. But the main problem here is the call to `count.Int64()`. There are methods such as `IsInt64()` that can determine if a number stored in `big.Int` can be represented as `int64`, which should be respected. To avoid assumptions, and thus possible truncation of ranges, I would suggest writing the for loop in the following way:
```Golang
for i := big.NewInt(0); i.Cmp(count) <= 0; i.Add(i, big.NewInt(100)) {
// ...
}
```

pkg/postage/postagecontract/mock/contract.go
```diff
@@ -60,3 +65,9 @@ func WithDiluteBatchFunc(f func(ctx context.Context, batchID []byte, newDepth ui
 		m.diluteBatch = f
 	}
 }
+
+func WithExpiresBatchesFunc(f func(ctx context.Context) error) Option {
```
mrekucci, line 69: This function is unused.

pkg/postage/postagecontract/contract.go
swarmHaseeb: i didnt miss this, the postage contract wasnt complete so creating a test didnt make any sense, also i cover minimum of 85% when writing tests :)

pkg/postage/postagecontract/mock/contract.go
swarmHaseeb: the contract was completed and approved yesterday so this part is now a wip

pkg/postage/postagecontract/contract.go
swarmHaseeb: this is no longer necessary as instead of count we are taking a bool from contract

pkg/postage/postagecontract/contract.go
```diff
@@ -311,39 +314,40 @@ func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topUpA
...
 		}
 	}
+
 	return nil
 }
 
-func (c *postageContract) countExpiredBatches(ctx context.Context) (*big.Int, error) {
-	callData, err := postageStampABI.Pack("countExpiry")
+func (c *postageContract) hasExpired(ctx context.Context) (bool, error) {
+	callData, err := postageStampABI.Pack("hasExpired")
```
agazso, line 333: This has been changed in the contract to be called `expiredBatchesExist`

pkg/postage/postagecontract/contract.go
```diff
@@ -311,39 +314,40 @@ func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topUpA
...
 		To:   &c.postageContractAddress,
 		Data: callData,
 	})
 	if err != nil {
-		return nil, err
+		return false, err
 	}
 
-	results, err := postageStampABI.Unpack("countExpiry", result)
+	results, err := postageStampABI.Unpack("hasExpired", result)
```
agazso, line 346: This has been changed in the contract to be called `expiredBatchesExist`

swarmHaseeb:
@istae this pr can be closed

PR #3497, opened 2022-11-04
chore: boost all agent calls

istae:
### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3497)

### DIFF
--- a/pkg/storageincentives/redistribution/redistribution.go
+++ b/pkg/storageincentives/redistribution/redistribution.go
@@ -110,7 +110,7 @@ func (c *contract) Claim(ctx context.Context) error {
 		Value:       big.NewInt(0),
 		Description: "claim win transaction",
 	}
-	err = c.sendAndWait(ctx, request)
+	err = c.sendAndWaitWithBoost(ctx, request, 50)
 	if err != nil {
 		return fmt.Errorf("claim: %w", err)
 	}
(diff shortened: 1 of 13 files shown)


notanatol (approved):

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -154,7 +154,7 @@ func (c *contract) Reveal(ctx context.Context, storageDepth uint8, reserveCommit
 		Value:       big.NewInt(0),
 		Description: "reveal transaction",
 	}
-	err = c.sendAndWaitWithBoost(ctx, request, 50)
+	err = c.sendAndWait(ctx, request, 50)
```
ldeffenb, line 157: I've been taking a slightly less aggressive approach with Reveal only boosting 10-20%, commit boosting 50%, and claim boosting 25-50%.   If a node misses the reveal, nothing is really lost but the chance to play.   Missing the commit is really bad.   Missing the claim is not as bad, but can also cost a LOT more gas.   Until the EIP transactions are put in place, this boost is a guaranteed increase in the gas paid for the transactions.

istae: We have work in progress to move to the new EIP transactions.
We are also adding new mechanisms to prevent the overloaded postage contract expire() from causing transactions to run of gas, which should lead to reductions in gas costs everywhere.

PR #3500, opened 2022-11-05
refactor: sharky cleanup and simplification of code

istae:
### Checklist

### Description
Write operations now follow a very basic round robin strategy with circular distribution of requests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3500)

### DIFF
--- a/pkg/sharky/recovery.go
+++ b/pkg/sharky/recovery.go
@@ -16,7 +16,7 @@ import (
 
 // Recovery allows disaster recovery.
 type Recovery struct {
-	shards []*slots
+	slots []*slots
 }
 
 var ErrShardNotFound = errors.New("shard not found")
@@ -31,47 +31,28 @@ func NewRecovery(dir string, shardCnt int, datasize int) (*Recovery, error) {
 		if err != nil {
 			return nil, err
 		}
-		fi, err := file.Stat()
-		if err != nil {
-			return nil, err
-		}
 		if err = file.Close(); err != nil {
 			return nil, err
 		}
-		size := uint32(fi.Size() / int64(datasize))
 		ffile, err := os.OpenFile(path.Join(dir, fmt.Sprintf("free_%03d", i)), os.O_RDWR|os.O_CREATE, 0666)
 		if err != nil {
 			return nil, err
 		}
-		sl := newSlots(ffile, nil)
-		sl.data = make([]byte, size/8)
-		shards[i] = sl
+		shards[i] = newSlots(ffile)
 	}
 	return &Recovery{shards}, nil
 }
 
-// Add marks a location as used (not free).
-func (r *Recovery) Add(loc Location) error {
-	sh := r.shards[loc.Shard]
-	l := len(sh.data)
-	if diff := int(loc.Slot/8) - l; diff >= 0 {
-		sh.extend(diff + 1)
-		for i := 0; i <= diff; i++ {
-			sh.data[l+i] = 0x0
-		}
-	}
-	sh.push(loc.Slot)
-	return nil
+// Use marks a location as used (not free).
+func (r *Recovery) Use(loc Location) {
+	r.slots[loc.Shard].Use(loc.Slot)
 }
 
 // Save saves all free slots files of the recovery (without closing).
 func (r *Recovery) Save() error {
 	err := new(multierror.Error)
-	for _, sh := range r.shards {
-		for i := range sh.data {
-			sh.data[i] ^= 0xff
-		}
-		err = multierror.Append(err, sh.save())
+	for _, sh := range r.slots {
+		err = multierror.Append(err, sh.Save())
 	}
 	return err.ErrorOrNil()
 }
@@ -79,7 +60,7 @@ func (r *Recovery) Save() error {
 // Close closes data and free slots files of the recovery (without saving).
 func (r *Recovery) Close() error {
 	err := new(multierror.Error)
-	for _, sh := range r.shards {
+	for _, sh := range r.slots {
 		err = multierror.Append(err, sh.file.Close())
 	}
 	return err.ErrorOrNil()
--- a/pkg/sharky/recovery.go
+++ b/pkg/sharky/recovery.go
@@ -10,8 +10,6 @@ import (
 	"io/fs"
 	"os"
 	"path"
-
-	"github.com/hashicorp/go-multierror"
 )
 
 // Recovery allows disaster recovery.
@@ -22,7 +20,7 @@ type Recovery struct {
 var ErrShardNotFound = errors.New("shard not found")
 
 func NewRecovery(dir string, shardCnt int, datasize int) (*Recovery, error) {
-	shards := make([]*slots, shardCnt)
+	slots := make([]*slots, shardCnt)
 	for i := 0; i < shardCnt; i++ {
 		file, err := os.OpenFile(path.Join(dir, fmt.Sprintf("shard_%03d", i)), os.O_RDONLY, 0666)
 		if err != nil {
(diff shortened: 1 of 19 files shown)


pkg/sharky/store.go
```diff
@@ -188,17 +156,12 @@ func (s *Store) Write(ctx context.Context, data []byte) (loc Location, err error
 // Note that releasing is not safe for obfuscating earlier content, since
 // even after reuse, the slot may be used by a very short blob and leaves the
 // rest of the old blob bytes untouched
-func (s *Store) Release(ctx context.Context, loc Location) error {
-	sh := s.shards[loc.Shard]
-	err := sh.release(ctx, loc.Slot)
+func (s *Store) Release(loc Location) {
 	s.metrics.TotalReleaseCalls.Inc()
```
vladopajic, line 160: neat: move this line below close to other metrics

pkg/sharky/slots_test.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"bytes"
+	"testing"
+)
+
+func TestFreeSlotsPersist(t *testing.T) {
+	t.Parallel()
+
+	basedir := dirFS(t.TempDir())
+	var err error
```
vladopajic, line 16: this is variable `err` declaration needed?

pkg/sharky/slots_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	ffile, err := basedir.Open("free_000")
```
vladopajic, line 18: neat: `"free_000"` is good candidate for constant var

pkg/sharky/slots_test.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	slot := newSlots(ffile.(sharkyFile))
```
vladopajic, line 23: would `slots` name be more suitable instead of `slot`?

pkg/sharky/slots_test.go
```diff
@@ -0,0 +1,89 @@
...
+	if nextFree != 10 {
+		t.Fatalf("want slot 10, got %d", nextFree)
+	}
+
+	// save and close file
+	err = slot.Save()
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = ffile.Close()
```
vladopajic, line 60: if `Close()` is exposed from by `slots` struct, then it would be handy to use it here

pkg/sharky/store.go
```diff
@@ -84,64 +94,32 @@ func (s *Store) create(index uint8, maxDataSize int, basedir fs.FS) (*shard, err
...
 		slots:       sl,
 		quit:        s.quit,
 	}
-	terminated := make(chan struct{})
-	sh.slots.wg.Add(1)
+	s.wg.Add(1)
 	go func() {
-		defer sh.slots.wg.Done()
+		defer s.wg.Done()
 		sh.process()
```
vladopajic, line 113: Currently new goroutine is started for each `shard` instance, would it be possible to avoid these goroutines? Are they really necessary?

--- 

All that `sh.process()` function does is essentially two operations:

```
slot := sh.slots.Next()
sh.slots.Use(slot)
```

Would it be possible to just call these two functions in (store.go:136) instead of reading from channel? Probably one more utility function would be needed to get shard, but overall code might be more simpler.

pkg/sharky/recovery.go
```diff
@@ -31,55 +31,36 @@ func NewRecovery(dir string, shardCnt int, datasize int) (*Recovery, error) {
 		if err != nil {
```
vladopajic, line 31: nit:

```
if err != nil {
	if errors.Is(err, fs.ErrNotExist) {
		return nil, fmt.Errorf("index %d: %w", i, ErrShardNotFound)
	}
	return nil, err
}
```

pkg/sharky/slots.go
```diff
@@ -10,38 +10,33 @@ import (
...
-	sl.data, err = io.ReadAll(sl.file)
+func (sl *slots) load() error {
+	data, err := io.ReadAll(sl.file)
 	if err != nil {
 		return err
 	}
-	sl.size = uint32(len(sl.data) * 8)
-	sl.head = sl.next(0)
-	return err
+	sl.data = data
```
vladopajic, line 30: Should statement 

```sl.data = data```

be guarded with mutex?

File: pkg/sharky/store.go [hunk 9]
```diff
@@ -84,64 +94,32 @@ func (s *Store) create(index uint8, maxDataSize int, basedir fs.FS) (*shard, err
...
-func (s *Store) Read(ctx context.Context, loc Location, buf []byte) (err error) {
+func (s *Store) Read(loc Location, buf []byte) (err error) {
 	sh := s.shards[loc.Shard]
-	select {
-	case sh.reads <- read{ctx: ctx, buf: buf[:loc.Length], slot: loc.Slot}:
-		s.metrics.TotalReadCalls.Inc()
-	case <-ctx.Done():
-		return ctx.Err()
-	case <-sh.quit:
-		return ErrQuitting
```
vladopajic, line 127: should var `ErrQuitting` be removed since it is now unused?

pkg/sharky/store.go
```diff
@@ -44,12 +48,11 @@ type Store struct {
...
-		quit:        make(chan struct{}),
-		metrics:     newMetrics(),
+		maxDataSize:     maxDataSize,
+		shards:          make([]*shard, shardCnt),
+		quit:            make(chan struct{}),
+		metrics:         newMetrics(),
+		availableShards: make(chan availableShard),
 	}
 	for i := range store.shards {
 		s, err := store.create(uint8(i), maxDataSize, basedir)
```
vladopajic, line 58: nit: `maxDataSize` could be accessed in ` store.create(..)`  via property `s.maxDataSize`

pkg/sharky/store.go
```diff
@@ -150,34 +128,24 @@ func (s *Store) Write(ctx context.Context, data []byte) (loc Location, err error
 	if len(data) > s.maxDataSize {
 		return loc, ErrTooLong
 	}
+
 	s.wg.Add(1)
 	defer s.wg.Done()
```
vladopajic, line 133: are these statements still needed?

vladopajic:
This sharky implementation does not have guaranties which shard and location is returned next. As an outcome this could lead to 1) probably more cache invalidation (on hdd/ssd level); and 2) fragmented data. Would it make sense to have sharky to always return first available location in consistent order? This way disk IO will probably hit same sector(s) on disk (probably having cache hit).

pkg/sharky/store.go
```diff
@@ -44,12 +48,11 @@ type Store struct {
 // - maxDataSize - positive integer representing the maximum blob size to be stored
 func New(basedir fs.FS, shardCnt int, maxDataSize int) (*Store, error) {
```
vladopajic, line 49: Is store ever writing data down to disk (closing and saving shards)?

pkg/sharky/store.go
```diff
@@ -28,12 +27,17 @@ var (
 // - read prioritisation over writing
 // - free slots allow write
 type Store struct {
```
vladopajic, line 29: This file should be tested.

istae:
> 

Order of writes does not guarantee order of reads however. We cannot tell which chunk will be retrieved in the future, so it does not matter which shard we place them.
That being said, there is a some benefit in having shards store only a certain PO.
For example, when the node is calculating the reserve sample and is going through the entire reserve, reading the shards in order could be a great optimization.

pkg/sharky/slots.go
istae: no, this function is only used during bootup so no mutex is needed

pkg/sharky/store.go
istae: good idea actually :)

notanatol (approved):

nikipapadatou:
Consider this again after the major localstore releases.

PR #3506, opened 2022-11-07
fix: cancel and resend to use new gas price

istae:
### Checklist

### Description
As part of EIP 1559, cancel and resend trx should use new suggested price.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3506)

### DIFF
--- a/pkg/transaction/transaction.go
+++ b/pkg/transaction/transaction.go
@@ -58,7 +58,7 @@ type StoredTransaction struct {
 	To          *common.Address // recipient of the transaction
 	Data        []byte          // transaction data
 	GasPrice    *big.Int        // used gas price
-	GasTipBoost int             // percantage used to boost the priority fee (eg: tip)
+	GasTipBoost int             // percentage used to boost the priority fee (eg: tip)
 	GasLimit    uint64          // used gas limit
 	Value       *big.Int        // amount of wei to send
 	Nonce       uint64          // used nonce
--- a/pkg/transaction/transaction.go
+++ b/pkg/transaction/transaction.go
@@ -40,8 +40,6 @@ var (
 	ErrAlreadyImported     = errors.New("already imported")
 )
 
-// minGasPrice determines the minimum gas price
-// threshold (in wei) for the creation of a transaction.
 const DefaultTipBoostPercent = 20
 
 // TxRequest describes a request for a transaction that can be executed.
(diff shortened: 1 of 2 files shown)


pkg/transaction/transaction.go
```diff
@@ -38,7 +38,6 @@ var (
 	ErrTransactionReverted = errors.New("transaction reverted")
 	ErrUnknownTransaction  = errors.New("unknown transaction")
 	ErrAlreadyImported     = errors.New("already imported")
-	ErrGasPriceTooLow      = errors.New("gas price too low")
 )
 
 // minGasPrice determines the minimum gas price
```
ldeffenb, line 43: Obsolete comment block?

pkg/transaction/transaction.go
```diff
@@ -284,34 +285,47 @@ func (t *transactionService) prepareTransaction(ctx context.Context, request *Tx
...
-	gasPrice := request.GasPrice
-	if gasPrice == nil {
-		gasPrice, err = t.backend.SuggestGasPrice(ctx)
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	gasTipCap, err := t.backend.SuggestGasTipCap(ctx)
+	gasFeeCap, gasTipCap, err := t.suggestedFeeAndTip(ctx, request.GasPrice, tipBoostPercent)
```
ldeffenb, line 288: I like the new merged routine.  From my understanding of EIP-1559, you'll never need one without the other.

pkg/transaction/transaction.go
```diff
@@ -403,14 +417,19 @@ func (t *transactionService) ResendTransaction(ctx context.Context, txHash commo
...
 	tx := types.NewTx(&types.DynamicFeeTx{
 		Nonce:     storedTransaction.Nonce,
 		ChainID:   t.chainID,
 		To:        storedTransaction.To,
 		Value:     storedTransaction.Value,
 		Gas:       storedTransaction.GasLimit,
-		GasTipCap: storedTransaction.GasPrice,
-		GasFeeCap: storedTransaction.GasPrice,
+		GasTipCap: gasTipCap,
+		GasFeeCap: gasFeeCap,
```
ldeffenb, line 432: Definitely seems to make more sense to use updated numbers if we're resending the transaction.   But @mfw78 told me a while back that any change like this will also change the resulting transaction ID which might mess up the transaction watcher?

ldeffenb: I was wanting the resend to boost the price when he described the extended ramifications of doing such a thing to an existing transaction nonce.  But if the watcher is only watching for the nonce and not the specific transaction ID, then I guess this would work?

ldeffenb (changes requested):
I would make this an approve, but I'm concerned about the Resend changing the prices which may result in a new transaction id which may affect the transaction completion watcher.

But definitely it's worth removing the obsolete minGasPrice comment.

notanatol (approved):

mrekucci (approved):

PR #3509, opened 2022-11-08
fix(sampler, redist): varius adjustment for the storage incentives

istae:
### Checklist

### Description
Reduces the gas limit for agent contract calls
Fixes a sampler log

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3509)

### DIFF
--- a/pkg/puller/puller.go
+++ b/pkg/puller/puller.go
@@ -29,7 +29,7 @@ const loggerName = "puller"
 
 var errCursorsLength = errors.New("cursors length mismatch")
 
-const DefaultSyncErrorSleepDur = time.Second * 5
+const DefaultSyncSleepDur = time.Minute
 
 type Options struct {
 	Bins         uint8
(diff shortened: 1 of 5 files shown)


pkg/puller/puller.go
```diff
@@ -29,7 +29,7 @@ const loggerName = "puller"
 
 var errCursorsLength = errors.New("cursors length mismatch")
 
-const DefaultSyncSleepDur = time.Minute
+const DefaultSyncErrorSleepDur = time.Second * 5
```
aloknerurkar, line 32: So I thought this should be higher than the kademlia trigger timeout right?... If not this can just keep trying multiple times before it can stop no?

istae: kademlia subscription fires whenever there is any change in the depth or topology, so a peer connecting/disconnecting or even storage radius moving around triggers it, so it's unpredictable. 

the kademlia timer triggers the manage loop to attempt new connections.

the original was never to have a higher timeout, as they are not related.

aloknerurkar (approved):

mrekucci (approved):

PR #3516, opened 2022-11-09
feat: add go leak check

vladopajic:
### Checklist

This PR adds goroutine leak check in unit tests. 
Check was commented out in packages with leaks detected. This was done in order to be able to track packages with leaks. Future PRs should fix problems in these packages (issue  #3522 tracks this effort)

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3516)


mrekucci:
I would recommend removing the commented checks and creating a new issue for each failing package check where it will be investigated and solved separately.

notanatol (approved):

pkg/accounting/main_test.go
```diff
@@ -0,0 +1,13 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package accounting_test
```
istae, line 5: if these leak tests fall under the _test file, why not add them to the already existing test file?

vladopajic: Only function defined in `main_test.go` files is `TestMain` function which does not do any testing per say. 

`TestMain` is common function for all tests which in this case adds logic for checking goroutine leaks in all `_test` files of current package. Since it adds additional logic to all tests it's makes more sense to be in individual file because in case when package has multiple `_test` files it wouldn't be clear which file should host common logic.

vladopajic: @istae ping

istae (approved):

PR #3520, opened 2022-11-10
fix: pullsync make offer timeout 

istae:
### Checklist

### DIFF
--- a/pkg/puller/puller.go
+++ b/pkg/puller/puller.go
@@ -29,7 +29,7 @@ const loggerName = "puller"
 
 var errCursorsLength = errors.New("cursors length mismatch")
 
-const DefaultSyncErrorSleepDur = time.Second * 5
+const DefaultSyncErrorSleepDur = time.Second * 30
 
 type Options struct {
 	Bins         uint8
--- a/pkg/pullsync/pullsync.go
+++ b/pkg/pullsync/pullsync.go
@@ -50,6 +50,7 @@ var (
 
 const (
 	storagePutTimeout = 5 * time.Second
+	makeOfferTimeout  = 5 * time.Minute
 )
 
 // how many maximum chunks in a batch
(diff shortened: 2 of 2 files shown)


pkg/puller/puller.go
```diff
@@ -29,7 +29,7 @@ const loggerName = "puller"
 
 var errCursorsLength = errors.New("cursors length mismatch")
 
-const DefaultSyncErrorSleepDur = time.Second * 5
+const DefaultSyncErrorSleepDur = time.Second * 30
```
mrekucci, line 32: I'm just curious as to why the length of sleep needs to be increased.

mrekucci (approved):

pkg/puller/puller.go
istae: it's an arbitrary value, 5 seconds seemed a bit aggressive to me after some thinking

aloknerurkar (approved):

pkg/pullsync/pullsync.go
```diff
@@ -50,6 +50,7 @@ var (
 
 const (
 	storagePutTimeout = 5 * time.Second
+	makeOfferTimeout  = 5 * time.Minute
```
ldeffenb, line 53: My nodes are (well, were before the stampes expired) routinely taking longer than 5 minutes to construct the offer.    makeOffer's context is automatically cancelled at the end of the commit round by the agent, so why have a local timeout specified here?

ldeffenb (commented):
Just curious about the local timeout?  If anything, shouldn't the caller specify the desired maximum attempt duration?

pkg/pullsync/pullsync.go
istae: this is puller/pullsync stuff, not reserve sampler :)

ldeffenb: Duh. makeOffer, not makeSample! That's what I get for commenting before I'm fully awake. I'll blame it on Nicole, the hurricane that passed by us last night!

istae: jeez, a bad year for florida this year! so many hurricanes

ldeffenb: At least Nicole was barely a category 1, just above a tropical storm.   Amazingly, we didn't even lose power; a few blinks overnight, but no outright loss.

Issue #3522, opened 2022-11-10
goroutine leaks reported with `goleak` library


vladopajic:
After adding `goleak` check in PR #3516, we observed that codebase has many packages with leaks detected by this library. 
This issue tracks progress for fixing these issues in all packages were problems have been detected. 

---

Packages with issues:
- [x] accounting
- [x] feeds/sequence
- [x] file
- [x] hive
- [x] postage/batchstore
- [x] localstore
- [x] statestore/leveldb
- [x] postage/listener
- [x] intervalstore
- [x] pusher
- [x] retrieval
- [x] node
- [x] feeds
- [x] api
- [x] p2p/protobuf
- [x] p2p/libp2p
- [x] topology/kademlia

PR #3529, opened 2022-11-11
fix(sequence): goroutine leak

vladopajic:
### Checklist

As part of https://github.com/ethersphere/bee/issues/3522, this PR addresses goroutine leaks in sequence package.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3529)

### DIFF
--- a/pkg/feeds/sequence/sequence.go
+++ b/pkg/feeds/sequence/sequence.go
@@ -158,7 +158,7 @@ func (f *asyncFinder) At(ctx context.Context, at, after int64) (ch swarm.Chunk, 
 		return nil, nil, &index{uint64(after)}, nil
 	}
 	// if chunk exists construct an initial interval with base=0
-	c := make(chan *result, 32)
+	c := make(chan *result, 8)
 	i := newInterval(0)
 	i.found = &result{ch, nil, 0, 0}
 
--- a/pkg/feeds/sequence/sequence.go
+++ b/pkg/feeds/sequence/sequence.go
@@ -158,7 +158,7 @@ func (f *asyncFinder) At(ctx context.Context, at, after int64) (ch swarm.Chunk, 
 		return nil, nil, &index{uint64(after)}, nil
 	}
 	// if chunk exists construct an initial interval with base=0
-	c := make(chan *result)
+	c := make(chan *result, 32)
 	i := newInterval(0)
 	i.found = &result{ch, nil, 0, 0}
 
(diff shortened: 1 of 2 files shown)


pkg/feeds/sequence/sequence.go
```diff
@@ -233,12 +233,13 @@ func (f *asyncFinder) at(ctx context.Context, at int64, min int, i *interval, c
 			}
 		}(l)
 	}
+
+	wg.Wait()
 }
 
 func (f *asyncFinder) asyncGet(ctx context.Context, at int64, index uint64) <-chan swarm.Chunk {
-	c := make(chan swarm.Chunk)
+	c := make(chan swarm.Chunk, 1)
```
zelig, line 241: i think this changes the logic substantially and   is incorrect

vladopajic: Only change in behavior is that goroutine started at line 242 will finish execution without waiting for other goroutine to read from response channel. There is no need for this goroutine to wait because it's responsability was to provide result  in background. 

Please, explain if I am missing the bigger picture.

mrekucci: There is no need for a buffered channel, you can avoid the goroutine leak by using the already passed context:
```golang
select {
	case c <- ch:
	case <-ctx.Done():
}
```

pkg/feeds/sequence/sequence.go
```diff
@@ -158,7 +158,7 @@ func (f *asyncFinder) At(ctx context.Context, at, after int64) (ch swarm.Chunk,
 		return nil, nil, &index{uint64(after)}, nil
 	}
 	// if chunk exists construct an initial interval with base=0
-	c := make(chan *result)
+	c := make(chan *result, 32)
```
mrekucci, line 161: I'm not sure about this change, could you perhaps provide some reasoning as to how this buffered channel solves the goroutine leak and how the number 32 was derived?

pkg/feeds/sequence/sequence.go
vladopajic: Yes that could also do the trick. 

Currently implemented solution, with buffered channel, has these advantages:
- is more robust pattern which could be reused in similar situations even when context is not provided
- goroutine will end early thus freeing it's resources earlier; note that even without intention to fix leaks, pattern without buffered channel will cause goroutines to pile up even when their results it ready, but consumer reading those results is slow.

As a general rule of thumb, whenever there is async method which returns promise like result via channels, it is always best approach to use buffered channel.

pkg/feeds/sequence/sequence.go
vladopajic: The reason to use buffered channel here is similar as described below. 

In this case there many producer gorutines `f.at` which will wait to write to channel `c`.  If we wanted to solve this without buffered channel,  lines 229 and 231 would require to have select statements with context signal. 

Channel capacity is irrelevant, but should be bit larger because there could be multiple results writing simultaneously.

pkg/feeds/sequence/sequence.go
mrekucci: In general, every goroutine should have an explicit cancellation story, relying on the buffered channel is non-deterministic. In this case, asynchronicity is provided by spinning the goroutine and waiting for the result in select, not by making the channel buffered. Also, if the parent goroutine terminates prematurely due to a quit or context done signal, the child goroutine will not be terminated.

Buffered channels are useful when you want to limit the number of goroutines you will launch, or want to limit the amount of work that is queued up.

mrekucci (approved):

PR #3530, opened 2022-11-11
fix(file): goroutine leak

vladopajic:
### Checklist

As part of https://github.com/ethersphere/bee/issues/3522, this PR addresses goroutine leaks in file package.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3530)

### DIFF
--- a/pkg/file/buffer_test.go
+++ b/pkg/file/buffer_test.go
@@ -167,23 +167,22 @@ func TestCopyBuffer(t *testing.T) {
 			}()
 
 			// receive the writes
-			// err may or may not be EOF, depending on whether writes end on
-			// chunk boundary
-			expected := dataSize
-			readTotal := 0
+			// err may or may not be EOF, depending on whether writes end on chunk boundary
 			readData := []byte{}
 			for {
 				select {
-				case res := <-resultC:
-					readData = append(readData, res.data...)
-					readTotal += res.n
-					if readTotal == expected {
-						// check received content
+				case res, ok := <-resultC:
+					if !ok {
+						// when resultC is closed (there is no more data to read)
+						// assert if read data is same as source
 						if !bytes.Equal(srcBytes, readData) {
 							t.Fatal("invalid byte content received")
 						}
+
 						return
 					}
+
+					readData = append(readData, res.data...)
 				case err := <-errC:
 					if err != nil && !errors.Is(err, io.EOF) {
 						t.Fatal(err)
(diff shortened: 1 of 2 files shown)


pkg/file/buffer_test.go
```diff
@@ -158,31 +150,24 @@ func TestCopyBuffer(t *testing.T) {
...
+			errC := make(chan error, 16)
 			go func() {
 				src := bytes.NewReader(srcBytes)
 
 				buf := make([]byte, swarm.ChunkSize)
 				c, err := io.CopyBuffer(chunkPipe, src, buf)
-				if err != nil {
-					errC <- err
-				}
+				errC <- err
```
aloknerurkar, line 164: It would be better to return here if the err != nil right? If not we will be sending a duplicate error on line 167. The Close can be done in a defer statement.

pkg/file/buffer_test.go
```diff
@@ -194,23 +179,20 @@ func TestCopyBuffer(t *testing.T) {
 			for {
 				select {
 				case c := <-sizeC:
+					d := <-dataC
```
aloknerurkar, line 182: There is a case in the reader if it encounters an EOF on line 209, only size will be sent. The data channel will not be written and this naked read will block. Is this not a problem?

pkg/file/buffer_test.go
vladopajic: `errC` will collect all errors, and fail test whenever error happens, thus there will not be second error because first error would end test. 

---

> The Close can be done in a defer statement.

I am not sure what Close?

pkg/file/buffer_test.go
vladopajic: This wasn't the problem because we would never encounter EOF, because test would end as soon as expected data is received. If we attempted to read once more time after receiving expected data, we would encounter this issue. 

However, in latest commit data and size channels have been combined to single channel (there was no need for two channels really).

vladopajic: There is one more commit after this in which we will read until EOF and assert expected data only then. 

It make more sense to do it like this because it covers one more case - what if there was more data to read from chunk pipe but we ended earlier.

pkg/file/buffer_test.go
```diff
@@ -39,24 +39,19 @@ func TestChunkPipe(t *testing.T) {
 			t.Parallel()
 
 			buf := file.NewChunkPipe()
-			sizeC := make(chan int, 255)
-			errC := make(chan error, 1)
+			sizeC := make(chan int, 16)
```
mrekucci, line 42: Since the channels are now the same size and the `close` calls have been removed, I recommend using `result` type and having just one channel (similarly, as you have done below):

```Golang
// ...
type result struct {
	n   int
	err error
}
resultC := make(chan result, 16)
// ...
```

This will make the code easier to follow.

vladopajic: This test case was structured differently and errC and sizeC are used separately. They are the same size but without any particular reason, except to provide extra buffer.

mrekucci: What are the reasons for using them separately (both are the result of a single Read call)?

vladopajic: Not sure, developer might have intended to use similar error handling structure in both tests?

mrekucci (approved):

aloknerurkar (approved):

PR #3531, opened 2022-11-11
fix(hive): goroutine leak

vladopajic:
### Checklist

- [ ] I have read the [coding guide](https://github.com/ethersphere/bee/blob/master/CODING.md).
- [ ] My change requires a documentation update, and I have done it.
- [ ] I have added tests to cover my changes.
- [x] I have filled out the description and linked the related issues.

As part of https://github.com/ethersphere/bee/issues/3522, this PR addresses goroutine leaks in hive package.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3531)

PR #3533, opened 2022-11-11
fix(postage/listener): goroutine leak

vladopajic:
### Checklist

As part of https://github.com/ethersphere/bee/issues/3522, this PR addresses goroutine leaks in postage/listener package.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3533)

### DIFF
--- a/pkg/postage/listener/listener.go
+++ b/pkg/postage/listener/listener.go
@@ -233,7 +233,7 @@ func (l *listener) Listen(ctx context.Context, from uint64, updater postage.Even
 
 	l.logger.Debug("batch factor", "value", batchFactor)
 
-	synced := make(chan error, 1)
+	synced := make(chan error)
 	closeOnce := new(sync.Once)
 	paged := make(chan struct{}, 1)
 	paged <- struct{}{}
--- a/pkg/postage/listener/listener.go
+++ b/pkg/postage/listener/listener.go
@@ -233,7 +233,7 @@ func (l *listener) Listen(ctx context.Context, from uint64, updater postage.Even
 
 	l.logger.Debug("batch factor", "value", batchFactor)
 
-	synced := make(chan error)
+	synced := make(chan error, 1)
 	closeOnce := new(sync.Once)
 	paged := make(chan struct{}, 1)
 	paged <- struct{}{}
@@ -341,7 +341,9 @@ func (l *listener) Listen(ctx context.Context, from uint64, updater postage.Even
 			l.logger.Error(err, "failed syncing event listener; shutting down node error")
 		}
 		closeOnce.Do(func() { synced <- err })
-		l.syncingStopped.Signal() // trigger shutdown in start.go
+		if l.syncingStopped != nil {
+			l.syncingStopped.Signal() // trigger shutdown in start.go
+		}
 	}()
 
 	return synced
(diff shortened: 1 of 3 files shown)


pkg/postage/listener/listener.go
```diff
@@ -66,7 +66,8 @@ type listener struct {
 	blockTime time.Duration
 
 	postageStampAddress common.Address
-	quit                chan struct{}
+	ctx                 context.Context
```
mrekucci, line 69: I recommend following the [documentation](https://github.com/golang/go/blob/master/src/context/context.go#L31-L37) and not abusing the context:
> Programs that use Contexts should follow these rules to keep interfaces
consistent across packages and enable static analysis tools to check context
propagation:
>
>**Do not store Contexts inside a struct type**; instead, pass a Context
explicitly to each function that needs it.

pkg/postage/listener/listener.go
```diff
@@ -61,12 +61,11 @@ type BlockHeightContractFilterer interface {
...
-	logger    log.Logger
-	ev        BlockHeightContractFilterer
-	blockTime time.Duration
-
+	logger              log.Logger
+	ev                  BlockHeightContractFilterer
+	blockTime           time.Duration
 	postageStampAddress common.Address
-	quit                chan struct{}
+	ctxCancel           context.CancelFunc
```
aloknerurkar, line 68: I dont see why we need to change this. We have this type of quit handling at a bunch of places and I prefer this as it is simple and easier to understand than the context approach. I dont think this needs to be part of this change at all, so I would revert this change back.

This PR should only enable leak check and not need to change any logic around the codebase. Also if you are going through other packages I would recommend you to only focus on enabling the leak check.

aloknerurkar (changes requested):

pkg/postage/listener/listener.go
vladopajic: Initially when this PR was published, there was a problem because code needed to use context and quit channel to signal termination, but in code only quit channel was being used. 

First reason for that change was to use only one approach to signal termination, because it would be sufficient enough. Having to listen on two signals is error prone as users can mistakenly avoid listening on one (as manifested in this code)

Since then there was newer PR which fixed this #3671.

---

Another reasoning why quit channel was replaced with context was the same as described [here](https://github.com/ethersphere/bee/pull/3531#discussion_r1065714484).

pkg/postage/listener/listener_test.go
```diff
@@ -515,19 +531,19 @@ func TestListenerBatchState(t *testing.T) {
 	}
 }
 
-func newEventUpdaterMock() (*updater, chan interface{}) {
-	c := make(chan interface{})
+func newEventUpdaterMock() *updater {
+	c := make(chan interface{}, 1)
```
mrekucci, line 535: I recommend inlining the `make` call into the returned struct. The same is below.

pkg/postage/listener/listener.go
```diff
@@ -233,7 +233,7 @@ func (l *listener) Listen(ctx context.Context, from uint64, updater postage.Even
 
 	l.logger.Debug("batch factor", "value", batchFactor)
 
-	synced := make(chan error)
+	synced := make(chan error, 1)
```
mrekucci, line 236: Is the buffered channel here necessary?

vladopajic: No, it's not needed anymore, this was leftover from previous changes.

mrekucci (approved):

aloknerurkar (approved):

PR #3535, opened 2022-11-13
fix: add transaction hash

### LINKED ISSUE #3359: Return transaction hash for endpoints that creates transactions
To my knowledge, when a user calls some Bee's endpoint which creates a transaction on a blockchain, it then awaits several confirmations before incorporating the requested change it into its internal state. For the creation of postage stamps, this is solved using the `usable` flag which indicates that the stamp is not ready to be used yet. This solves it only for this endpoint, but actually, there are several endpoints that are not handled and this creates a problem for handling these actions in UIs.

Imagine, that as a user of Swarm Desktop you decide to, for example, dilute a postage batch. You click fill out the form with information and submit it. The spinning wheel will appear (the API request to Bee endpoint is sent), then it disappears without any error and it seems like it was a success (the API request was finished, we got Response from Bee), but then the user refreshes the page (or Swarm Desktop does it for him) but nothing has changed! 😱 Well that is because of this "void period" when Bee internally awaits the confirmations... 

For the creation of postage batches, we can regularly poll the batch detail to see if it is ready, but for other actions, this is not true. To my knowledge, there is no way how to handle this case with the API at the moment. There is the `GET /transaction` endpoint on Debug API, but if there are more transactions happening at the same time, then we can not map these to the proper API call.

There are two solutions I can see:
 1. Return response for the API call, only after Bee gets its confirmations. 
 2. For all transaction creation endpoints, return also the transaction hash.

I am more in favor of the second solution, as having an open request for even several minutes does not sound really like a great idea.

umerm-work:
### Checklist

### Description
Closes #3359 

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3535)

### DIFF
--- a/openapi/SwarmCommon.yaml
+++ b/openapi/SwarmCommon.yaml
@@ -385,6 +385,8 @@ components:
       properties:
         batchID:
           $ref: "#/components/schemas/BatchID"
+        txHash:
+          $ref: "#/components/schemas/TransactionHash"
 
     Response:
       type: object
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -68,6 +68,7 @@ func (b hexByte) MarshalJSON() ([]byte, error) {
 
 type postageCreateResponse struct {
 	BatchID hexByte `json:"batchID"`
+	TxHash  string  `json:"txHash"`
 }
 
 func (s *Service) postageCreateHandler(w http.ResponseWriter, r *http.Request) {
@@ -90,7 +91,7 @@ func (s *Service) postageCreateHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	batchID, err := s.postageContract.CreateBatch(
+	txHash, batchID, err := s.postageContract.CreateBatch(
 		r.Context(),
 		paths.Amount,
 		paths.Depth,
@@ -124,6 +125,7 @@ func (s *Service) postageCreateHandler(w http.ResponseWriter, r *http.Request) {
 
 	jsonhttp.Created(w, &postageCreateResponse{
 		BatchID: batchID,
+		TxHash:  txHash,
 	})
 }
 
@@ -461,7 +463,7 @@ func (s *Service) postageTopUpHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	hexBatchID := hex.EncodeToString(paths.BatchID)
 
-	err := s.postageContract.TopUpBatch(r.Context(), paths.BatchID, paths.Amount)
+	_, err := s.postageContract.TopUpBatch(r.Context(), paths.BatchID, paths.Amount)
 	if err != nil {
 		if errors.Is(err, postagecontract.ErrInsufficientFunds) {
 			logger.Debug("topup batch: out of funds", "batch_id", hexBatchID, "amount", paths.Amount, "error", err)
@@ -499,7 +501,7 @@ func (s *Service) postageDiluteHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	hexBatchID := hex.EncodeToString(paths.BatchID)
 
-	err := s.postageContract.DiluteBatch(r.Context(), paths.BatchID, paths.Depth)
+	_, err := s.postageContract.DiluteBatch(r.Context(), paths.BatchID, paths.Depth)
 	if err != nil {
 		if errors.Is(err, postagecontract.ErrInvalidDepth) {
 			logger.Debug("dilute batch: invalid depth", "error", err)
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -100,24 +100,24 @@ func (s *Service) postageCreateHandler(w http.ResponseWriter, r *http.Request) {
 	)
 	if err != nil {
 		if errors.Is(err, postagecontract.ErrChainDisabled) {
-			logger.Debug("create batch: no chain backend", "error", err)
+			logger.Debug("create batch: no chain backend", "error", err, "txHash", txHash)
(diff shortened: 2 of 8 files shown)


notanatol (approved):

pkg/postage/postagecontract/contract.go
```diff
@@ -263,49 +263,49 @@ func (c *postageContract) getBalance(ctx context.Context) (*big.Int, error) {
...
 	err = c.ExpireBatches(ctx)
 	if err != nil {
-		return nil, err
+		return "", nil, err
 	}
 
 	_, err = c.sendApproveTransaction(ctx, totalAmount)
 	if err != nil {
-		return nil, err
+		return "", nil, err
```
swarmHaseeb, line 289: We can also return transaction hash in case of a failed transaction for more information. I also propose to add them to debug logs for easier debugging. 
Also in case of a failed transaction it would be wise to know the Gas Price and other related information (Tip/boost percent etc) for debugging purposes.
If you look at pkg/transaction/transaction.go Line 164 ` err = t.backend.SendTransaction(ctx, signedTx)` it looks like what Abel planned was to return the tx hash in case of error and he probably may have left it in the middle (or had planned to complete it later in another PR) so you can use that to actually get the details of transaction more easily. @istae

swarmHaseeb (changes requested):

pkg/postage/postagecontract/contract.go
umerm-work: So we already sending an error here, which is being handled in the endpoint. I think the error is enough to describe the failure of the transaction.

openapi/SwarmCommon.yaml
```diff
@@ -385,6 +385,8 @@ components:
       properties:
         batchID:
           $ref: "#/components/schemas/BatchID"
+        txHash:
```
istae, line 388: please update version

pkg/api/postage.go
```diff
@@ -461,7 +463,7 @@ func (s *Service) postageTopUpHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	hexBatchID := hex.EncodeToString(paths.BatchID)
 
-	err := s.postageContract.TopUpBatch(r.Context(), paths.BatchID, paths.Amount)
+	_, err := s.postageContract.TopUpBatch(r.Context(), paths.BatchID, paths.Amount)
```
istae, line 466: missing transaction in the json response

pkg/api/postage.go
```diff
@@ -499,7 +501,7 @@ func (s *Service) postageDiluteHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	hexBatchID := hex.EncodeToString(paths.BatchID)
 
-	err := s.postageContract.DiluteBatch(r.Context(), paths.BatchID, paths.Depth)
+	_, err := s.postageContract.DiluteBatch(r.Context(), paths.BatchID, paths.Depth)
```
istae, line 504: missing transaction in the json response

pkg/postage/postagecontract/contract.go
```diff
@@ -263,49 +263,49 @@ func (c *postageContract) getBalance(ctx context.Context) (*big.Int, error) {
 	return abi.ConvertType(results[0], new(big.Int)).(*big.Int), nil
 }
 
-func (c *postageContract) CreateBatch(ctx context.Context, initialBalance *big.Int, depth uint8, immutable bool, label string) ([]byte, error) {
+func (c *postageContract) CreateBatch(ctx context.Context, initialBalance *big.Int, depth uint8, immutable bool, label string) (string, []byte, error) {
 
 	if depth <= BucketDepth {
-		return nil, ErrInvalidDepth
+		return "", nil, ErrInvalidDepth
```
istae, line 269: i would use naked returns or named returns values to make the change more "pretty" and get rid of all of the quotes

pkg/postage/postagecontract/contract.go
```diff
@@ -263,49 +263,49 @@ func (c *postageContract) getBalance(ctx context.Context) (*big.Int, error) {
 	return abi.ConvertType(results[0], new(big.Int)).(*big.Int), nil
 }
 
-func (c *postageContract) CreateBatch(ctx context.Context, initialBalance *big.Int, depth uint8, immutable bool, label string) ([]byte, error) {
+func (c *postageContract) CreateBatch(ctx context.Context, initialBalance *big.Int, depth uint8, immutable bool, label string) (string, []byte, error) {
```
istae, line 266: ideally, instead of a string, it should return `common.Hash` type

pkg/postage/postagecontract/contract.go
istae: we log the transaction actually, it's done by the transaction monitor service
./pkg/transaction/monitor.go L103

istae (changes requested):

pkg/postage/postagecontract/contract.go
umerm-work: So I initially set it to `common.Hash`  but thought we only going to use string. Any specific reason for `common.Hash`?

swarmHaseeb (approved):

pkg/postage/postagecontract/contract.go
istae: it's more specific, easier to read/understand, similar to how we return swarm.Address instead of []byte

openapi/SwarmCommon.yaml
umerm-work: @istae sorry, which version should it be?

istae (approved):

PR #3553, opened 2022-11-15
fix: cancel should have same gas as the stored trx

istae:
### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3553)

### DIFF
--- a/pkg/transaction/transaction.go
+++ b/pkg/transaction/transaction.go
@@ -478,7 +478,7 @@ func (t *transactionService) CancelTransaction(ctx context.Context, originalTxHa
 		ChainID:   t.chainID,
 		To:        &t.sender,
 		Value:     big.NewInt(0),
-		Gas:       storedTransaction.GasLimit,
+		Gas:       21000,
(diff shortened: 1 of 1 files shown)


notanatol (approved):

ldeffenb:
Doesn't the gas on a cancel need to be higher than the existing transaction with the same nonce?   The only time I've used cancel is when I have a languishing transaction that was submitted with too low of a gas price.  If you don't increase the gas price, then the replace won't happen and the cancel will never work either.  At least, that's my understanding and experience.

istae:
> Doesn't the gas on a cancel need to be higher than the existing transaction with the same nonce? The only time I've used cancel is when I have a languishing transaction that was submitted with too low of a gas price. If you don't increase the gas price, then the replace won't happen and the cancel will never work either. At least, that's my understanding and experience.

@ldeffenb  We got some errors complaining about cancellation price being different than the original trx price, we are in the process of testing this a bit more.

mrekucci (approved):

PR #3554, opened 2022-11-15
fix: use the correct hash

notanatol:
### Checklist

### Description
In case of error we are logging an incorrect value for the tx: `0x00000000...`

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3554)


pkg/api/transaction.go
```diff
@@ -183,8 +183,8 @@ func (s *Service) transactionCancelHandler(w http.ResponseWriter, r *http.Reques
 
 	txHash, err := s.transaction.CancelTransaction(ctx, paths.Hash)
 	if err != nil {
-		logger.Debug("cancel transaction failed", "tx_hash", txHash, "error", err)
```
istae, line 186: maybe also include the original and the new cancellation transaction in the logs

PR #3555, opened 2022-11-16
fix: pullsync will only return an error if the interval needs to be retried

istae:
### Checklist

### Description
Pullsync will only return an error if the interval needs to be retried, (eg: connection or localstore related errors), for other errors, we simply log and continue the operation.
This way the puller can easily determine when to retry the same interval.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3555)

### DIFF
--- a/pkg/pullsync/pullsync.go
+++ b/pkg/pullsync/pullsync.go
@@ -118,7 +118,7 @@ func (s *Syncer) Protocol() p2p.ProtocolSpec {
 // It returns the BinID of highest chunk that was synced from the given interval.
 // If the requested interval is too large, the downstream peer has the liberty to
 // provide less chunks than requested.
-func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8, from, to uint64) (topmost uint64, err error) {
+func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8, from, to uint64) (uint64, error) {
 	isLiveSync := to == MaxCursor
 	loggerV2 := s.logger.V(2).Register()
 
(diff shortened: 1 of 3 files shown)


notanatol (approved):

pkg/pullsync/pullsync.go
```diff
@@ -168,24 +168,21 @@ func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8
...
+			s.logger.Debug("syncer got a zero address hash on offer")
+			continue
 		}
 		s.metrics.Offered.Inc()
 		s.metrics.DbOps.Inc()
 		have, err = s.storage.Has(ctx, a)
 		if err != nil {
-			err = fmt.Errorf("storage has: %w", err)
-			return
+			s.logger.Debug("storage has", "error", err)
```
mrekucci, line 185: Shouldn't be there a `continue` statement after this log message?

istae: if the storage.Has fails, it's best to just ask the peer for it.

pkg/pullsync/pullsync.go
```diff
@@ -120,7 +120,7 @@ func (s *Syncer) Protocol() p2p.ProtocolSpec {
 // provide less chunks than requested.
 func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8, from, to uint64) (uint64, error) {
 	isLiveSync := to == MaxCursor
-	loggerV2 := s.logger.V(2).Register()
+	// loggerV2 := s.logger.V(2).Register()
```
mrekucci, line 123: This can be removed when it's not used.

mrekucci (approved):

PR #3556, opened 2022-11-16
fix: log nonce on cancellation 

notanatol:
### Checklist

### Description
Cancel transaction fails with "nonce too low", the logs do not show the nonce.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3556)

### DIFF
--- a/pkg/api/transaction.go
+++ b/pkg/api/transaction.go
@@ -183,7 +183,7 @@ func (s *Service) transactionCancelHandler(w http.ResponseWriter, r *http.Reques
 
 	txHash, err := s.transaction.CancelTransaction(ctx, paths.Hash)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/api/transaction.go
```diff
@@ -183,7 +183,7 @@ func (s *Service) transactionCancelHandler(w http.ResponseWriter, r *http.Reques
 
 	txHash, err := s.transaction.CancelTransaction(ctx, paths.Hash)
 	if err != nil {
-		logger.Debug("cancel transaction failed", "tx_hash", paths.Hash, "error", err, "canceled_tx_hash", txHash)
```
notanatol, line 186: this always prints zero value for `canceled_tx_hash`, so omitting.

mrekucci (approved):

PR #3566, opened 2022-11-21
fix: relax reserve eviction during GC

aloknerurkar:
### Checklist

### Description
The localstore reserve size calculation is not correct and for storage incentives, we manually calculate the syncing index to calculate the reserve size.

In the private testnet we observed that the reserve eviction was causing some inconsistencies in the samples generated. During GC, we evict chunks from the reserve till we reach 90% capacity and move them to the cache. As we no longer consider cache chunks in the sample calculation, we have decided to relax the eviction till we are at 100% capacity. So, only evict if we have more chunks.

Also, as the storage radius is now dynamic, we will need to calculate the reserve size using the current radius during eviction. This might lead to node having to store some more data, but eventually, all the expiring batches should be cleaned up.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3566)

### DIFF
--- a/pkg/localstore/gc.go
+++ b/pkg/localstore/gc.go
@@ -403,7 +403,7 @@ func (db *DB) evictReserve() (totalEvicted uint64, done bool, err error) {
 	if err != nil {
 		return 0, false, err
 	}
-	db.logger.Debug("gc: reserve size", reserveSizeStart, "target", target)
+	db.logger.Debug("gc: reserve eviction", "reserveSizeStart", reserveSizeStart, "target", target)
 	if reserveSizeStart <= target {
 		return 0, true, nil
 	}
--- a/pkg/localstore/gc.go
+++ b/pkg/localstore/gc.go
@@ -388,7 +388,7 @@ func (db *DB) evictReserve() (totalEvicted uint64, done bool, err error) {
 		totalTimeMetric(db.metrics.TotalTimeEvictReserve, start)
 	}(time.Now())
 
-	target = db.reserveCapacity
+	target = db.reserveEvictionTarget()
 
 	db.batchMu.Lock()
 	defer db.batchMu.Unlock()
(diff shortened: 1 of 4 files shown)


pkg/localstore/gc.go
```diff
@@ -387,13 +387,23 @@ func (db *DB) evictReserve() (totalEvicted uint64, done bool, err error) {
...
+	if db.radiusFunc != nil {
+		reserveSizeStart, err = db.ComputeReserveSize(db.radiusFunc())
+		target = db.reserveCapacity
+	} else {
+		reserveSizeStart, err = db.reserveSize.Get()
+	}
 	if err != nil {
 		return 0, false, err
 	}
+	db.logger.Debug("gc: reserve size", reserveSizeStart, "target", target)
```
mrekucci, line 406: You're missing the key for the `reserveSizeStart`.

pkg/localstore/gc.go
```diff
@@ -387,13 +387,23 @@ func (db *DB) evictReserve() (totalEvicted uint64, done bool, err error) {
...
+
 	target = db.reserveEvictionTarget()
+
 	db.batchMu.Lock()
 	defer db.batchMu.Unlock()
-	reserveSizeStart, err := db.reserveSize.Get()
+
+	var reserveSizeStart uint64
+	if db.radiusFunc != nil {
+		reserveSizeStart, err = db.ComputeReserveSize(db.radiusFunc())
```
janos, line 398: This code is not tested, but should be as it is executing during bee runtime.

janos (approved):

janos (approved):
👍

pkg/localstore/gc.go
```diff
@@ -387,13 +387,23 @@ func (db *DB) evictReserve() (totalEvicted uint64, done bool, err error) {
...
+		target = db.reserveCapacity
+	} else {
+		reserveSizeStart, err = db.reserveSize.Get()
+	}
 	if err != nil {
```
mrekucci, line 403: I'd recommend moving the error check also into the `else` block.

pkg/localstore/gc.go
```diff
@@ -387,13 +387,23 @@ func (db *DB) evictReserve() (totalEvicted uint64, done bool, err error) {
...
 		return 0, false, err
 	}
+	db.logger.Debug("gc: reserve eviction", "reserveSizeStart", reserveSizeStart, "target", target)
```
mrekucci, line 406: We prefer to use snake-case instead of camel-case for log keys.

istae (approved):

PR #3577, opened 2022-11-23
fix: save stamp issuer data after usage not service close

istae:
### Checklist

### Description
We have seen MANY chunks whose stamp indexes are colliding with others chunks belonging to the same batch.
This is possible if stamp issuer data is not stored to disk properly on `Close()` such that on the next restart of the node, the same bucket and indexes are assigned to a different uploaded chunk. 

This change makes sure that after a stamp is used, the bucket data is stored immediately with a callback.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3577)

### DIFF
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -287,7 +287,11 @@ func (s *Service) postageGetStampBucketsHandler(w http.ResponseWriter, r *http.R
 		}
 		return
 	}
-	defer func() { _ = save() }()
+	defer func() {
+		if err := save(); err != nil {
+			s.logger.Debug("stamp issuer save", "error", err)
+		}
+	}()
 
 	b := issuer.Buckets()
 	resp := postageStampBucketsResponse{
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -273,7 +273,7 @@ func (s *Service) postageGetStampBucketsHandler(w http.ResponseWriter, r *http.R
 	}
 	hexBatchID := hex.EncodeToString(paths.BatchID)
 
-	issuer, err := s.post.GetStampIssuer(paths.BatchID)
+	issuer, save, err := s.post.GetStampIssuer(paths.BatchID)
 	if err != nil {
 		logger.Debug("get stamp issuer: get issuer failed", "batch_id", hexBatchID, "error", err)
 		logger.Error(nil, "get stamp issuer: get issuer failed")
@@ -287,6 +287,7 @@ func (s *Service) postageGetStampBucketsHandler(w http.ResponseWriter, r *http.R
 		}
 		return
 	}
+	defer save()
 
 	b := issuer.Buckets()
 	resp := postageStampBucketsResponse{
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -287,7 +287,7 @@ func (s *Service) postageGetStampBucketsHandler(w http.ResponseWriter, r *http.R
 		}
 		return
 	}
-	defer save()
+	defer func() { _ = save() }()
 
 	b := issuer.Buckets()
 	resp := postageStampBucketsResponse{
--- a/pkg/api/pss.go
+++ b/pkg/api/pss.go
@@ -78,7 +78,7 @@ func (s *Service) pssPostHandler(w http.ResponseWriter, r *http.Request) {
 		jsonhttp.BadRequest(w, "invalid postage batch id")
 		return
 	}
-	i, err := s.post.GetStampIssuer(batch)
+	i, save, err := s.post.GetStampIssuer(batch)
 	if err != nil {
 		logger.Debug("get postage batch issuer failed", "batch_id", hex.EncodeToString(batch), "error", err)
 		logger.Error(nil, "get postage batch issuer failed")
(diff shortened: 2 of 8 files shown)


pkg/postage/service_test.go
```diff
@@ -99,7 +99,7 @@ func TestGetStampIssuer(t *testing.T) {
 	}
 	t.Run("found", func(t *testing.T) {
 		for _, id := range ids[1:4] {
-			st, err := ps.GetStampIssuer(id)
+			st, _, err := ps.GetStampIssuer(id)
```
aloknerurkar, line 102: We should add a test for this.

pkg/postage/service.go
```diff
@@ -162,22 +162,24 @@ func (ps *service) IssuerUsable(st *StampIssuer) bool {
...
 	ps.lock.Lock()
 	defer ps.lock.Unlock()
 	for _, st := range ps.issuers {
 		if bytes.Equal(batchID, st.data.BatchID) {
 			if !ps.IssuerUsable(st) {
-				return nil, ErrNotUsable
+				return nil, nil, ErrNotUsable
 			}
-			return st, nil
+			return st, func() { _ = ps.save() }, nil
```
notanatol, line 173: just for the context, why did we go from `func() error` -> `func()`?

istae: because the caller of the callback does not what to do with error, so we swallow the error here
in `service.Closer` we return the error thought

pkg/api/postage.go
```diff
@@ -287,6 +287,7 @@ func (s *Service) postageGetStampBucketsHandler(w http.ResponseWriter, r *http.R
 		}
 		return
 	}
+	defer func() { _ = save() }()
```
mrekucci, line 290: I would recommend at least logging the error so that there is a crumb if it fails to save.

pkg/api/pss.go
```diff
@@ -92,6 +92,8 @@ func (s *Service) pssPostHandler(w http.ResponseWriter, r *http.Request) {
 		}
 		return
 	}
+	defer func() { _ = save() }()
```
mrekucci, line 95: The same here.

pkg/api/soc.go
```diff
@@ -134,6 +134,8 @@ func (s *Service) socUploadHandler(w http.ResponseWriter, r *http.Request) {
 		}
 		return
 	}
+	defer func() { _ = save() }()
```
mrekucci, line 137: Also here.

pkg/postage/service_test.go
```diff
@@ -99,24 +102,37 @@ func TestGetStampIssuer(t *testing.T) {
...
 		}
+
+		// check if the save() call persisted the stamp issuers
+		for i, id := range ids[1:4] {
+			issuer := new(postage.StampIssuer)
+			err := store.Get(fmt.Sprintf("postage%d%d", chainID, i), issuer)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if !bytes.Equal(id, issuer.ID()) {
```
mrekucci, line 122: Since those byte slices are quite long and difficult to compare by eye in case of a mismatch, I recommend doing the following:
```Golang
if diff := cmp.Diff(id, issuer.ID()); diff != "" {
	t.Fatalf("issuer mismatch (-want +have):\n%s", diff)
}
```

pkg/postage/service.go
```diff
@@ -162,22 +162,24 @@ func (ps *service) IssuerUsable(st *StampIssuer) bool {
...
 	ps.lock.Lock()
 	defer ps.lock.Unlock()
 	for _, st := range ps.issuers {
 		if bytes.Equal(batchID, st.data.BatchID) {
 			if !ps.IssuerUsable(st) {
-				return nil, ErrNotUsable
+				return nil, nil, ErrNotUsable
 			}
-			return st, nil
+			return st, func() error { return ps.save() }, nil
```
vladopajic, line 173: neat: `return st, ps.save, nil` (no need to define function)

pkg/postage/service.go
```diff
@@ -33,7 +33,7 @@ var (
 type Service interface {
 	Add(*StampIssuer) error
 	StampIssuers() []*StampIssuer
-	GetStampIssuer([]byte) (*StampIssuer, error)
+	GetStampIssuer([]byte) (*StampIssuer, func() error, error)
```
vladopajic, line 36: Preferably `Service` should expose `Save` function and `GetStampIssuer` should retain old signature. Because returned `save` function of `GetStampIssuer` is not uniquely created on each call  (of `GetStampIssuer`) therfore if this function doesn't do anything specific to this call is should not be returned by this method.

istae: I prefer this pattern of returning a "cleanup" or "done" callback because it forces the user to use it. 
The exposed `Save` does not enforce this.

vladopajic: How is returning this function forcing user to use it? In this PR it appears that new function signature actually creates dogsled in some places.

istae: Outside of tests, I don't see it anywhere else.

mrekucci (approved):

notanatol (approved):

Issue #3039, opened 2022-06-23
localstore write-in-place inconsistency bug


acud:
There's a bug in the localstore which is related to the postage stamp indexing scheme which results in an `HTTP 500` when trying to upload a file through the API.

The error would usually look like:
```
level=debug msg="bytes upload: split write all: put upload: same slot remove: get value: leveldb: not found" traceID=....
```

This relates to the code in `pkg/localstore/mode_put.go`:
```go
func (db *DB) putUpload(
	batch *leveldb.Batch,
	loc *releaseLocations,
	binIDs map[uint8]uint64,
	item shed.Item,
) (exists bool, gcSizeChange int64, err error) {

	previous, err := db.postageIndexIndex.Get(item)
	if err != nil {
		if !errors.Is(err, leveldb.ErrNotFound) {
			return false, 0, fmt.Errorf("postage index get: %w", err)
		}
	} else {
		if item.Immutable {
			return false, 0, ErrOverwrite
		}
		// if a chunk is found with the same postage stamp index,
		// replace it with the new one only if timestamp is later
		if !later(previous, item) {
			return false, 0, nil
		}
		_, err = db.setRemove(batch, previous, true)
		if err != nil {
			return false, 0, fmt.Errorf("same slot remove: %w", err)  // <----- Error occurs here
		}

		previousIdx, err := db.retrievalDataIndex.Get(previous)
		if err != nil {
			return false, 0, fmt.Errorf("could not fetch previous item: %w", err)
		}

		l, err := sharky.LocationFromBinary(previousIdx.Location)
		if err != nil {
			return false, 0, err
		}

		loc.add(l)
}
...
```

The postage indexing scheme hands out slots within a postage batch. More exactly, there are specific indexes inside every collision bucket in the postage batch. Once a chunk gets stamped, the number of collisions seen within the bucket so far is given out as the postage batch index and is serialized as part of the postage stamp on the chunk while being stamped by the `Stamper`. The localstore is aware of this index. This is so that users can buy storage space that can be later reused to store other content (hence the timestamp check when storing), and so that postage stamps will become multi-use rather than a one-off event.

The bug at hand manifests in the following manner (that's at least, to the best of my understanding):
- buy a postage stamp `x`, use it normally
- `db nuke` the node (prior to #3011), stampissuers would get deleted
- the forgetful node would then resync the chain data, and would reinstate the stampissuer, but all collision buckets would be zeroed out
[...]

PR #3579, opened 2022-11-23
fix(localstore): remove reserve calculation from localstore

aloknerurkar:
### Checklist

### Description
During the recent storage incentives changes, we found that the reserve size calculation in the localstore was off. A bunch of efforts has already been made to understand and fix this, but due to the dynamic nature of storage depth, it seems like it is not possible to have a static count of reserve size stored in the localstore, which can represent the current reserve size.

The reserve size is now synonymous with the pullIndex as this is the index we use to generate the sample. This index will now be used to represent the reserve chunks in the localstore. So if chunks are put into GC index, which represents the cache, we should remove them from pullIndex as they are no longer part of the reserve. This change will make the addition of pullIndex more strict and only allow syncing related contexts to add the pullIndex.

As the reserve size calculation is now completely removed, there are a lot of simplifications in the mode_put file.

Reserve evictions will now be triggered if we are above reserve capacity as computed with the depth seen by the depthmonitor. This is periodic and will only happen if we are above capacity. Localstore will only trigger GC evictions from now.

All the unit tests are adjusted to now verify the new expected values in the indexes.

A bug related to postageIndex collisions is also fixed. During GC, we were not deleting the postageIndexIndex entry, which led to us searching for a chunk that is not present in the localstore. #3039 

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3579)

### DIFF
--- a/pkg/localstore/gc.go
+++ b/pkg/localstore/gc.go
@@ -19,6 +19,7 @@ package localstore
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/ethersphere/bee/pkg/sharky"
@@ -170,6 +171,7 @@ func (db *DB) collectGarbage() (evicted uint64, done bool, err error) {
 	// get rid of dirty entries
 	for _, item := range candidates {
 		if swarm.NewAddress(item.Address).MemberOf(db.dirtyAddresses) {
+			fmt.Println("skipping", swarm.NewAddress(item.Address))
 			continue
 		}
 
@@ -182,6 +184,7 @@ func (db *DB) collectGarbage() (evicted uint64, done bool, err error) {
 		// the target since the gc size always is bound to change even if to a minor degree in the time between
 		// candidate collection and the mutex acquisition.
 		if gcSize-totalChunksEvicted <= target {
+			fmt.Println("stopping", totalChunksEvicted, gcSize, target)
 			done = true
 			break
 		}
@@ -218,10 +221,6 @@ func (db *DB) collectGarbage() (evicted uint64, done bool, err error) {
 			return 0, false, err
 		}
 		err = db.postageChunksIndex.DeleteInBatch(batch, item)
-		if err != nil {
-			return 0, false, err
-		}
-		err = db.postageIndexIndex.DeleteInBatch(batch, item)
 		if err != nil {
 			return 0, false, err
 		}
@@ -311,6 +310,38 @@ func (db *DB) incGCSizeInBatch(batch *leveldb.Batch, change int64) (err error) {
 	return nil
 }
 
+func (db *DB) reserveEvictionWorker() {
+	defer close(db.reserveEvictionWorkerDone)
+	for {
+		select {
+		case <-db.reserveEvictionTrigger:
+			evictedCount, done, err := db.evictReserve()
+			if err != nil {
+				db.logger.Error(err, "evict reserve failed")
+			}
+
+			if !done {
+				db.triggerReserveEviction()
+			}
+
+			if testHookEviction != nil {
+				testHookEviction(evictedCount)
+			}
+		case <-db.close:
+			return
+		}
+	}
+}
+
+func (db *DB) triggerReserveEviction() {
+	select {
+	case db.reserveEvictionTrigger <- struct{}{}:
+	case <-db.close:
+	default:
+	}
+
+}
+
 func (db *DB) evictReserve() (totalEvicted uint64, done bool, err error) {
 	var target uint64
 	db.metrics.EvictReserveCounter.Inc()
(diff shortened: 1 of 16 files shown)


pkg/localstore/gc_test.go
```diff
@@ -1152,173 +1150,31 @@ func TestReserveEvictionWorker(t *testing.T) {
...
-			}
-		}
-		batchIDs = nil
-		return nil
-	}
+	t.Run("pull index count", newItemsCountTest(db.pullIndex, chunkCount-1))
 
-	checkReserveComputation := func(t *testing.T, count uint64) {
-		t.Helper()
+	// postage index will not be deleted
```
aloknerurkar, line 1157: comment needs to be updated

istae (approved):

pkg/localstore/mode_put.go
```diff
@@ -248,20 +239,51 @@ func (db *DB) put(ctx context.Context, mode storage.ModePut, chs ...swarm.Chunk)
...
+		return 0, ErrOverwriteImmutable
+	}
+	// if a chunk is found with the same postage stamp index,
+	// replace it with the new one only if timestamp is later
+	if !later(previous, item) {
+		return 0, ErrOverwrite
+	}
+
+	// remove older chunk
+	previousIdx, err := db.retrievalDataIndex.Get(previous)
```
istae, line 270: reminder to ignore error here

pkg/localstore/gc.go
```diff
@@ -187,13 +177,12 @@ func (db *DB) collectGarbage() (evicted uint64, done bool, err error) {
...
 		if err != nil {
 			return 0, false, err
 		}
-		item.Location = i.Location
 
-		db.metrics.GCStoreTimeStamps.Set(float64(item.StoreTimestamp))
+		db.metrics.GCStoreTimeStamps.Set(float64(storedItem.StoreTimestamp))
 		db.metrics.GCStoreAccessTimeStamps.Set(float64(item.AccessTimestamp))
 
 		// delete from retrieve, pull, gc
```
mrekucci, line 188: The repetitive error handling can be improved be employing the following:
```Golang
err = nil
deleteInBatch := func(fn func(*leveldb.Batch, shed.Item) error, batch *leveldb.Batch, keyFields shed.Item) {
	if err != nil {
		return
	}
	err = fn(batch, keyFields)
}

deleteInBatch(db.retrievalDataIndex.DeleteInBatch, batch, item)
deleteInBatch(db.retrievalAccessIndex.DeleteInBatch, batch, item)
deleteInBatch(db.pushIndex.DeleteInBatch, batch, storedItem)
deleteInBatch(db.pullIndex.DeleteInBatch, batch, item)
deleteInBatch(db.gcIndex.DeleteInBatch, batch, item)
deleteInBatch(db.postageIndexIndex.DeleteInBatch, batch, storedItem)
deleteInBatch(db.postageChunksIndex.DeleteInBatch, batch, item)
```

pkg/localstore/gc.go
```diff
@@ -388,22 +333,18 @@ func (db *DB) evictReserve() (totalEvicted uint64, done bool, err error) {
...
-	}
+	target = db.reserveCapacity
+
+	reserveSizeStart, err := db.reserveSize.Get()
 	if err != nil {
 		return 0, false, err
 	}
-	db.logger.Debug("gc: reserve eviction", "reserveSizeStart", reserveSizeStart, "target", target)
+
+	db.logger.Debug("gc: reserve eviction", "reserve-size-start", reserveSizeStart, "target", target)
```
mrekucci, line 346: `reserve-size-start` -> `reserve_size_start`

pkg/localstore/gc_test.go
```diff
@@ -1072,32 +1075,31 @@ func TestReserveEvictionWorker(t *testing.T) {
...
-			t.Fatal("eviction timeout")
-		}
-		reserveSize, err := db.reserveSize.Get()
-		if err != nil {
-			t.Fatal(err)
-		}
-		if reserveSize == evictTarget {
+	select {
+	case c := <-testHookEvictionChan:
+		if c == 1 {
```
mrekucci, line 1082: Shouldn't the test fail if `c != 1`?

pkg/localstore/reserve.go
```diff
@@ -180,6 +180,24 @@ func (db *DB) ComputeReserveSize(startPO uint8) (uint64, error) {
...
+	}
 
 	return count, err
 }
+
+// SetReserveSize will update the localstore reserve size as calculated by the
+// depthmonitor using the updated storage depth
+func (db *DB) setReserveSize(size uint64) error {
+	err := db.reserveSize.Put(size)
+	if err == nil {
```
mrekucci, line 197: Returning early is preferable:
```Golang
err := db.reserveSize.Put(size)
if err != nil {
	return err
}
	
if size > db.reserveCapacity {
	db.triggerReserveEviction()
}
return nil
```

notanatol (approved):

janos (approved):

pkg/localstore/gc.go
aloknerurkar: Interesting. This can be done in a lot of places actually. But I dont want to prolong this change as we need to start testing. Will do this in a follow-up PR later.

PR #3585, opened 2022-11-24
fix: gc to continue on retrieval error

istae:
### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3585)

### DIFF
--- a/pkg/localstore/gc.go
+++ b/pkg/localstore/gc.go
@@ -180,6 +180,9 @@ func (db *DB) collectGarbage() (evicted uint64, done bool, err error) {
 		storedItem, err := db.retrievalDataIndex.Get(item)
 		if err != nil {
 			if errors.Is(err, leveldb.ErrNotFound) {
+				if err = db.gcIndex.DeleteInBatch(batch, item); err != nil {
+					return 0, false, err
+				}
 				continue
 			}
 			return 0, false, err
--- a/pkg/localstore/gc.go
+++ b/pkg/localstore/gc.go
@@ -179,6 +179,9 @@ func (db *DB) collectGarbage() (evicted uint64, done bool, err error) {
 
 		storedItem, err := db.retrievalDataIndex.Get(item)
 		if err != nil {
+			if errors.Is(err, leveldb.ErrNotFound) {
+				continue
+			}
 			return 0, false, err
 		}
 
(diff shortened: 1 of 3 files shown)


pkg/localstore/mode_put.go
```diff
@@ -466,6 +466,11 @@ func (db *DB) addToCache(
 	batch *leveldb.Batch,
 	item shed.Item,
 ) (int64, error) {
+
+	err := db.pullIndex.DeleteInBatch(batch, item)
```
aloknerurkar, line 470: So moving these doesnt make any difference. The index is only updated when the Batch is written. So order doesnt matter.

pkg/localstore/mode_set.go
```diff
@@ -365,6 +365,10 @@ func (db *DB) setUnpin(batch *leveldb.Batch, addr swarm.Address) (gcSizeChange i
 	item.StoreTimestamp = i.StoreTimestamp
 	item.BinID = i.BinID
 	item.BatchID = i.BatchID
+	err = db.pullIndex.DeleteInBatch(batch, item)
```
aloknerurkar, line 368: Same

pkg/localstore/gc.go
```diff
@@ -179,6 +179,9 @@ func (db *DB) collectGarbage() (evicted uint64, done bool, err error) {
 
 		storedItem, err := db.retrievalDataIndex.Get(item)
 		if err != nil {
+			if errors.Is(err, leveldb.ErrNotFound) {
```
aloknerurkar, line 182: So I had thought about this. But ideally this shouldnt happen as gcIndex is also deleted once it is removed.

Either way, if retrievalDataIndex is not present, then we should atleast deleted the gcIndex so we dont go through this entry again.

pkg/localstore/mode_set.go
istae: so here if an item is in the pushIndex, it should not affect the reserve, this is why it's moved above the pushIndex check.

pkg/localstore/mode_put.go
istae: Here we have early returns from the function due to the pin check, this should not affect the chunk being removed from the reserve.

pkg/localstore/gc.go
istae: yes should not happen but we have thousands of "msg"="collect garbage failed" errors

pkg/localstore/mode_put.go
aloknerurkar: So on L481 we return if there is an existing pinIndex. In this case, we dont put the item to gcIndex and hence we should not delete the pullIndex.

aloknerurkar (approved):

PR #3614, opened 2022-12-02
fix: synchronize the reserve sampling process

### LINKED ISSUE #3611: disable reserve sampler during chunks evictions
The reserve sampler must wait for the eviction process to finish before producing a sample.

aloknerurkar:
### Checklist

### Description
In the localstore we usually guard parallel operations using the `batchMu` mutex. This is a huge bottleneck, but one that cannot be avoided at the moment. Due to the nature of the current localstore code, we need to have this exclusive lock to protect any changes to the underlying DB.

For the sampling, we never acquired this lock. But we should as there could be changes to the pullIndex during iteration which should not happen. Also, currently, we do not acquire this lock when we evict batches from the batchstore. Both these cases are now fixed.

With this change, if there is a sampling process ongoing and reserve evictions happen, we will cancel the sampler. This will cause the player to not participate in the lottery game for that round. This is considered better as if the player does participate and end up submitting a wrong sample, he might get frozen. Also, having the lock for the sampling process seems to be an overkill and would slow down the Put path causing bottlenecks.

Along with this, I have added metrics for the BatchEviction process which were missing.

Fixes #3611 hopefully.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3614)


istae (approved):

pkg/localstore/sampler.go
```diff
@@ -84,6 +87,11 @@ func (db *DB) ReserveSample(
 
 	t := time.Now()
 
+	// protect the DB from any updates till we finish creating the sample
+	db.batchMu.Lock()
```
istae, line 91: On second thought, because the sampler can take several minutes, locking the function at the top may not be a good idea as it will delay everything else.

I propose making the addrChan a buffered channel with a large size (several million) and unlocking `batchMu` when the iterator finishes. 

By moving the mutex unlock to when the iterator finishes, we risk `db.get` returning nil entries as gc could be running in the background. Here I advise we introduce a second mutex called `gcMutex` that only the reserve sampler and the gc uses.

istae (changes requested):

pkg/localstore/sampler.go
istae: If we go with a buffered channel of size 2^22 (the size of the reserve) it's a 134 mb overhead .

pkg/localstore/store_lock.go
```diff
@@ -0,0 +1,41 @@
+package localstore
```
istae, line 1: missing header

pkg/localstore/store_lock.go
```diff
@@ -0,0 +1,41 @@
...
+
+import "resenje.org/multex"
+
+const (
+	Upload   string = "upload"
```
janos, line 6: I do not thing that these constants need to be exported as they are only used as internal lock keys. Because of that, also, a more explanatory name would be good, like `lockKeyUpload` or something similar. When type parametrized multex version could be used, the custom type could be used for clarity instead naming. Just a suggestion.

janos (commented):
The approach looks pretty good with not so drastic changes. It should be good to go after linter is satisfied.

istae (approved):

mrekucci (approved):

Issue #3610, opened 2022-11-30
Bootnodes should have pushync disabled


istae:
Bootnodes are not storage nodes (puller is disabled) so a chunks that lands in a bootnode through pushsync cannot be reliably pulled by it's neighbors.

PR #3616, opened 2022-12-02
fix: pushsync disabled for bootnode

vladopajic:
### Checklist

Closes:  #3610

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3616)


pkg/node/node.go
```diff
@@ -921,7 +921,8 @@ func NewBee(interrupt chan struct{}, sysInterrupt chan os.Signal, addr string, p
 
 	pinningService := pinning.NewService(storer, stateStore, traversalService)
 
-	pushSyncProtocol := pushsync.New(swarmAddress, nonce, p2ps, storer, kad, tagService, o.FullNodeMode, pssService.TryUnwrap, validStamp, logger, acc, pricer, signer, tracer, warmupTime)
```
istae, line 924: I wonder if we can get away with not even registering the protocol for bootnodes 🤔

istae: @aloknerurkar @mrekucci @notanatol wdyt?

notanatol: > @aloknerurkar @mrekucci @notanatol wdyt?

we could do that but I feel that having a switch in place is a more "fine grained" solution

notanatol (approved):

pkg/node/node.go
aloknerurkar: Sorry for not getting back!

So @istae is right. This will only disable storing on pushsync. However we will still see pushsync protocol traffic. I think the intention of this issue is to not have that as well. I think it would be better to disable the protocol itself, similar to how we do for pullsync i.e. not register the protocol on the libp2p host.

@janos Could there be any issues if we dont register this protocol? We might start seeing more errors for pushsync? Or would the nodes understand that this peer doesnt have the protocol and not send messages?

istae: @vladopajic let's go with p2p.WithBlocklistStreams

vladopajic:
Since pushsync is now disabled for bootnodes, integration tests in beekeeper should be updated not to include bootnodes.

PR https://github.com/ethersphere/beekeeper/pull/297

pkg/node/node.go
```diff
@@ -944,7 +945,7 @@ func NewBee(interrupt chan struct{}, sysInterrupt chan os.Signal, addr string, p
 	pushSyncProtocolSpec := pushSyncProtocol.Protocol()
 	pullSyncProtocolSpec := pullSyncProtocol.Protocol()
 
-	if o.FullNodeMode {
+	if o.FullNodeMode && !o.BootnodeMode {
```
istae, line 948: this may be the only change needed, the  `includeSelf` changes may not be needed

istae (changes requested):

pkg/pushsync/pushsync.go
```diff
@@ -87,7 +87,7 @@ type PushSync struct {
 	tracer         *tracing.Tracer
 	validStamp     postage.ValidStampFn
 	signer         crypto.Signer
-	isFullNode     bool
+	includeSelf    bool
```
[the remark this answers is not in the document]
vladopajic: isn't it better to have consistent naming?

istae (approved):

PR #3641, opened 2022-12-10
fix: deadlock in localstore batchstore interaction

aloknerurkar:
### Checklist

### Description
The current implementation of localstore and batchstore is a little problematic. We assign the localstore.EvictBatch function to batchstore to perform evictions due to batch expirations and batchstore.Unreserve to localstore to free reserve when we go over capacity.

This causes a deadlock situation as reserveEvictionWorker could acquire the GC lock and then wait for batchstore lock which is held by the cleanup function of batchstore which calls evictFn (localstore.EvictBatch), which waits on the GC lock. Solution is to allow only 1 reserve eviction process to start at a time. This is done by adding a new lock.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3641)

### DIFF
--- a/pkg/localstore/reserve_test.go
+++ b/pkg/localstore/reserve_test.go
@@ -646,8 +646,6 @@ func TestReserveSize(t *testing.T) {
 }
 
 func TestComputeReserveSize(t *testing.T) {
-	t.Parallel()
-
 	const chunkCountPerPO = 10
 	const maxPO = 10
 	var chs []swarm.Chunk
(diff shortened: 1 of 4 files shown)


janos (approved):
Looks good.

istae (approved):

pkg/localstore/reserve_test.go
```diff
@@ -646,8 +646,6 @@ func TestReserveSize(t *testing.T) {
 }
 
 func TestComputeReserveSize(t *testing.T) {
-	t.Parallel()
```
istae, line 649: not sure if accidental

istae (approved):

pkg/localstore/reserve_test.go
aloknerurkar: so this package doesnt have parallel tests. It shouldnt also as they override global variables. I think this was added by mistake but there should not be any t.Parallel tests in localstore for now.

PR #3643, opened 2022-12-10
fix: recalc puller peers more periodically

istae:
### Checklist

### Description
It was observed on the private testnet that no pullsyncing was occurring because the topology change subscription never fired as there were no changes in the topology after the warm up period.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3643)

### DIFF
--- a/pkg/puller/puller.go
+++ b/pkg/puller/puller.go
@@ -29,7 +29,10 @@ const loggerName = "puller"
 
 var errCursorsLength = errors.New("cursors length mismatch")
 
-const DefaultSyncErrorSleepDur = time.Second * 30
+const (
+	DefaultSyncErrorSleepDur = time.Second * 30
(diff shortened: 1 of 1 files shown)


pkg/puller/puller.go
```diff
@@ -101,51 +101,59 @@ func (p *Puller) manage(ctx context.Context, warmupTime time.Duration) {
...
+	defer tick.Stop()
 
-			p.syncPeersMtx.Unlock()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-tick.C:
+			onChange()
+		case <-c:
```
aloknerurkar, line 155: So in order to prevent two simultaneous runs of this, would it be better to do a ticker reset here? this should prevent abuse of this I think.

aloknerurkar (approved):

PR #3644, opened 2022-12-12
fix: batchstore localstore deadlock

aloknerurkar:
### Checklist

### Description
The previous fix for synchronizing the eviction processes was not the right fix for the goroutine pileup seen on the testnet. The problem happens because of the batchstore.Mutex contention as PutChainState in batchstore can call EvictBatch under the batchstore.Lock which will be waiting for GC lock held by reserveEvictionWorker that could be running simultaneously waiting for the batchstore.Mutex through Unreserve.

Solution is to not have any locking when we call evictFn from batchstore. So now we will register the expiredBatches in localstore when called from the batchstore and trigger the reserveEviction worker. This will enable only 1 flow of eviction and hence prevent any locking problems.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3644)

### DIFF
--- a/pkg/localstore/gc.go
+++ b/pkg/localstore/gc.go
@@ -302,11 +302,24 @@ func (db *DB) reserveEvictionWorker() {
 	for {
 		select {
 		case <-db.reserveEvictionTrigger:
+			db.lock.Lock(lockKeyBatchExpiry)
+			expiredBatches := db.expiredBatches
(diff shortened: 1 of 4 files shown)


notanatol (approved):

janos (approved):
👍

istae (approved):
nice

PR #3650, opened 2022-12-14
fix: use eviction radius for new batches

aloknerurkar:
### Checklist

### Description
It could happen that we pushsync chunks of newer batches that are outside our radius of responsibility as we have not seen the batch before. So we consider the current radius of this batch is `0` in localstore and might end up storing chunks for it.

With this fix, we use the initial value of current storage radius for a batch while adding it to batchstore. This would disallow any chunks from lower bins to be replicated by pushsync in the reserve.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3650)


notanatol (approved):

istae:
fixed by https://github.com/ethersphere/bee/pull/3651

Issue #3660, opened 2022-12-16
batch syncing does not respond to interrupt signal


istae:
⚠️ Support requests in an issue-format will be closed immediately. For support, go to Swarm's [Discord](https://discord.gg/XGNBEKktkD).

### Summary
The node outputs the log `waiting to sync postage contract data, this may take a while... more info available in Debug loglevel` and this stage, the node will not respond to any interrupt signal, like ctrl+C

PR #3671, opened 2022-12-21
fix: gracefully suspend bee node on interrupt signal

vladopajic:
### Checklist

### Description
Closes: #3660

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3671)

### DIFF
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -83,9 +83,9 @@ func (c *command) initStartCmd() (err error) {
 			// ctx is global context of bee node; which is canceled after interrupt signal is received.
 			ctx, cancel := context.WithCancel(context.Background())
 			sysInterruptChannel := make(chan os.Signal, 1)
-			go func() {
-				signal.Notify(sysInterruptChannel, syscall.SIGINT, syscall.SIGTERM)
+			signal.Notify(sysInterruptChannel, syscall.SIGINT, syscall.SIGTERM)
 
+			go func() {
 				select {
 				case <-sysInterruptChannel:
 					logger.Info("received interrupt signal")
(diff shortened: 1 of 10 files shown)


cmd/bee/cmd/start.go
```diff
@@ -238,16 +242,24 @@ func (c *command) initStartCmd() (err error) {
...
 					case <-b.SyncingStopped():
+						logger.Debug("syncing has stopped")
 					}
 
-					logger.Info("shutting down")
+					logger.Info("shutting down...")
 				},
 				stop: func() {
-					// Shutdown
+					// Whenever program is being stoppend we need to close
```
notanatol, line 254: typo in `stoppend`

notanatol (approved):

mrekucci:
The interrupt signal still does not work when batch synchronization is in progress. Also when a node is waiting to be funded.
The interrupt signal probably has to be propagated to the postage listener and also to the init chequebook service.

vladopajic:
@mrekucci In latest change main bee node context has been propagated to all components. Please, could you check once more time?

cmd/bee/cmd/start.go
```diff
@@ -259,10 +269,12 @@ func (c *command) initStartCmd() (err error) {
 
 					// If shutdown function is blocking too long,
 					// allow process termination by receiving another signal.
+					shutdownTimeout := time.Second * 5
 					select {
-					case <-sysInterruptChannel:
-						logger.Debug("received interrupt signal")
+					case <-time.After(shutdownTimeout):
```
aloknerurkar, line 274: This is not correct. Adding a hard-coded timeout is not the right thing to do. There could be valid cases when the node takes more than 5 secs to stop.

As the comment said, the correct way to handle this is to check for a second interrupt signal to force stop if user does not have the patience. I am not sure if this can be done using context cancellation.

aloknerurkar (changes requested):

vladopajic:
@aloknerurkar I have manually testes all those scenarios. However there could be some issue if some logic is not responding to context cancelation signal. This would be hard to catch on this level; instead it should be covered/tested in respective package.

PR #3672, opened 2022-12-21
feat(storagev2): simplify ChunkStore Put API

aloknerurkar:
### Checklist

### Description
The current signature of the Put API of the `ChunkStore` interface returns an exists boolean along with the error. This was mainly used in the old implementation to take care of the tags functionality. In most other places this value is not checked/used.

With the new localstore, we plan to move the tag functionality into the upload store. So we can remove this and make things simpler from the implementation perspective.

Now the Put API will only return error. Any duplicate updates will be handled by the implementation. This could be done by simply overwriting contents or updating state etc.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3672)


notanatol (approved):

mrekucci (approved):
Except for the linting errors, LGTM.

PR #3683, opened 2023-01-03
feat: redistribution status

### LINKED ISSUE #3675: Redistribution state endpoint
It would be good to have an endpoint which returns the internals of the redistribution game from the perspective of a node so that the node operator does not have to use a blockchain explorer to reverse-engineer the state of the game from the blockchain data.

For example it could return the current state of the node in the game as:
```bash
curl http://localhost:1635/redistributionstate
{
  "state": "claim",
  "round": 168187,
  "block": 25564502,
  "reward": "0.0341",
  "fees": "0.0001213"
}
```

`state` could be one of `claim | commit | reveal | idle | frozen`
`round` is a number of the round of a game
`block` is the block number of the round of a game
`reward` is the cumulative reward of a node
`fees` is the cumulative fees of a node

Some of this information is already displayed in the [monSI](https://github.com/rndlabs/monSI) tool, however it shows all the information for all the nodes and with this endpoint only the relevant information regarding the given node would be returned.

umerm-work:
### Checklist

### Description
Closes #3675 

## Open API Spec Version Changes (if applicable)
4.0.0 to 4.1.0
Swarm.yaml

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3683)

### DIFF
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 
 info:
-  version: 4.0.0
+  version: 4.1.0
   title: Bee API
   description: "A list of the currently provided Interfaces to interact with the swarm, implementing file operations and sending messages"
 
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 
 info:
-  version: 4.1.0
+  version: 4.0.0
   title: Bee API
   description: "A list of the currently provided Interfaces to interact with the swarm, implementing file operations and sending messages"
 
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -121,7 +121,6 @@ type Service struct {
 	auth            auth.Authenticator
 	tags            *tags.Tags
 	storer          storage.Storer
-	stateStorer     storage.StateStorer
 	resolver        resolver.Interface
 	pss             pss.Interface
 	traversal       traversal.Traverser
@@ -235,7 +234,7 @@ type ExtraOptions struct {
 	IndexDebugger    StorageIndexDebugger
 }
 
-func New(publicKey, pssPublicKey ecdsa.PublicKey, ethereumAddress common.Address, logger log.Logger, transaction transaction.Service, batchStore postage.Storer, stateStorer storage.StateStorer, beeMode BeeNodeMode, chequebookEnabled bool, swapEnabled bool, chainBackend transaction.Backend, cors []string) *Service {
+func New(publicKey, pssPublicKey ecdsa.PublicKey, ethereumAddress common.Address, logger log.Logger, transaction transaction.Service, batchStore postage.Storer, beeMode BeeNodeMode, chequebookEnabled, swapEnabled bool, chainBackend transaction.Backend, cors []string) *Service {
 	s := new(Service)
 
 	s.CORSAllowedOrigins = cors
@@ -249,7 +248,6 @@ func New(publicKey, pssPublicKey ecdsa.PublicKey, ethereumAddress common.Address
 	s.ethereumAddress = ethereumAddress
 	s.transaction = transaction
 	s.batchStore = batchStore
-	s.stateStorer = stateStorer
 	s.chainBackend = chainBackend
 	s.metricsRegistry = newDebugMetrics()
 	s.preMapHooks = map[string]func(v string) (string, error){
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -120,6 +120,7 @@ type Service struct {
 	auth            auth.Authenticator
 	tags            *tags.Tags
 	storer          storage.Storer
+	stateStorer     storage.StateStorer
 	resolver        resolver.Interface
 	pss             pss.Interface
 	traversal       traversal.Traverser
@@ -225,7 +226,7 @@ type ExtraOptions struct {
 	IndexDebugger    StorageIndexDebugger
 }
 
-func New(publicKey, pssPublicKey ecdsa.PublicKey, ethereumAddress common.Address, logger log.Logger, transaction transaction.Service, batchStore postage.Storer, beeMode BeeNodeMode, chequebookEnabled bool, swapEnabled bool, chainBackend transaction.Backend, cors []string) *Service {
+func New(publicKey, pssPublicKey ecdsa.PublicKey, ethereumAddress common.Address, logger log.Logger, transaction transaction.Service, batchStore postage.Storer, stateStorer storage.StateStorer, beeMode BeeNodeMode, chequebookEnabled bool, swapEnabled bool, chainBackend transaction.Backend, cors []string) *Service {
 	s := new(Service)
 
 	s.CORSAllowedOrigins = cors
@@ -239,6 +240,7 @@ func New(publicKey, pssPublicKey ecdsa.PublicKey, ethereumAddress common.Address
 	s.ethereumAddress = ethereumAddress
 	s.transaction = transaction
 	s.batchStore = batchStore
+	s.stateStorer = stateStorer
 	s.chainBackend = chainBackend
 	s.metricsRegistry = newDebugMetrics()
 	s.preMapHooks = map[string]func(v string) (string, error){
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -14,6 +14,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/ethersphere/bee/pkg/storageincentives"
 	"io"
 	"math"
 	"math/big"
@@ -182,6 +183,8 @@ type Service struct {
 
 	preMapHooks map[string]func(v string) (string, error)
 	validate    *validator.Validate
+
+	redistributionAgent *storageincentives.Agent
 }
 
 func (s *Service) SetP2P(p2p p2p.DebugService) {
@@ -193,6 +196,12 @@ func (s *Service) SetP2P(p2p p2p.DebugService) {
 func (s *Service) SetSwarmAddress(addr *swarm.Address) {
 	if s != nil {
 		s.overlay = addr
+	}
+}
+
+func (s *Service) SetRedistributionAgent(redistributionAgent *storageincentives.Agent) {
+	if s != nil {
+		s.redistributionAgent = redistributionAgent
 	}
 }
 
@@ -261,7 +270,6 @@ func New(publicKey, pssPublicKey ecdsa.PublicKey, ethereumAddress common.Address
 		}
 		return name
 	})
-
 	return s
 }
 
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -177,6 +177,9 @@ func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.
 	storeRecipient := statestore.NewStateStore()
 	recipient := pseudosettle.New(nil, o.Logger, storeRecipient, nil, big.NewInt(10000), big.NewInt(10000), o.P2P)
 
+	if o.StateStorer == nil {
+		o.StateStorer = storeRecipient
+	}
 	erc20 := erc20mock.New(o.Erc20Opts...)
 	backend := backendmock.New(o.BackendOpts...)
 
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -16,7 +16,6 @@ import (
 	contractMock "github.com/ethersphere/bee/pkg/postage/postagecontract/mock"
 	"github.com/ethersphere/bee/pkg/settlement/swap/erc20"
 	"github.com/ethersphere/bee/pkg/storageincentives"
-	"github.com/ethersphere/bee/pkg/storageincentives/redistribution"
 	mock2 "github.com/ethersphere/bee/pkg/storageincentives/staking/mock"
 	"io"
 	"math/big"
@@ -212,7 +211,7 @@ func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.
 
 	s.SetP2P(o.P2P)
 
-	o.redistributionAgent = createRedistributionAgentService(o.Overlay, nil, o.StateStorer, erc20)
+	o.redistributionAgent = createRedistributionAgentService(o.Overlay, o.StateStorer, erc20)
 	s.SetSwarmAddress(&o.Overlay)
 	s.SetRedistributionAgent(o.redistributionAgent)
 	s.SetProbe(o.Probe)
@@ -717,9 +716,7 @@ func (c *chanStorer) Close() error {
 	panic("not implemented") // TODO: Implement
 }
 
-func createRedistributionAgentService(
-	addr swarm.Address,
-	contract redistribution.Contract, storer storage.StateStorer, erc20Service erc20.Service) *storageincentives.Agent {
+func createRedistributionAgentService(addr swarm.Address, storer storage.StateStorer, erc20Service erc20.Service) *storageincentives.Agent {
 	var blocksPerRound uint64 = 12
 	var blocksPerPhase uint64 = 4
 	postageContract := contractMock.New(contractMock.WithExpiresBatchesFunc(func(context.Context) error {
@@ -740,7 +737,7 @@ func createRedistributionAgentService(
 		},
 		incrementBy: 2,
 		block:       blocksPerRound}
-	contract = &mockContract{}
+	contract := &mockContract{}
 
 	return storageincentives.New(addr, backend, log.Noop, nil, contract, postageContract, stakingContract, mockbatchstore.New(mockbatchstore.WithReserveState(&postage.ReserveState{StorageRadius: 0})), nil, time.Millisecond*10, blocksPerRound, blocksPerPhase, storer, erc20Service)
 }
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -12,6 +12,10 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
+	contractMock "github.com/ethersphere/bee/pkg/postage/postagecontract/mock"
+	"github.com/ethersphere/bee/pkg/storageincentives"
+	"github.com/ethersphere/bee/pkg/storageincentives/redistribution"
+	mock2 "github.com/ethersphere/bee/pkg/storageincentives/staking/mock"
 	"io"
 	"math/big"
 	"net"
@@ -82,6 +86,7 @@ func init() {
 
 type testServerOptions struct {
 	Storer             storage.Storer
+	StateStorer        storage.StateStorer
 	Resolver           resolver.Interface
 	Pss                pss.Interface
 	Traversal          traversal.Traverser
@@ -122,9 +127,10 @@ type testServerOptions struct {
 	BatchStore postage.Storer
 	SyncStatus func() (bool, error)
 
-	BackendOpts []backendmock.Option
-	Erc20Opts   []erc20mock.Option
-	beeMode     api.BeeNodeMode
+	BackendOpts         []backendmock.Option
+	Erc20Opts           []erc20mock.Option
+	beeMode             api.BeeNodeMode
+	redistributionAgent *storageincentives.Agent
 }
 
 func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.Conn, string, *chanStorer) {
@@ -199,11 +205,13 @@ func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.
 	if o.beeMode == api.LightMode {
 		o.beeMode = api.FullMode
 	}
-	s := api.New(o.PublicKey, o.PSSPublicKey, o.EthereumAddress, o.Logger, transaction, o.BatchStore, statestore.NewStateStore(), o.beeMode, true, true, backend, o.CORSAllowedOrigins)
+	s := api.New(o.PublicKey, o.PSSPublicKey, o.EthereumAddress, o.Logger, transaction, o.BatchStore, o.beeMode, true, true, backend, o.CORSAllowedOrigins)
 
 	s.SetP2P(o.P2P)
 	s.SetSwarmAddress(&o.Overlay)
+	o.redistributionAgent = createRedistributionAgentService(o.Overlay, nil, nil, 0, 0, o.StateStorer)
 
+	s.SetRedistributionAgent(o.redistributionAgent)
 	s.SetProbe(o.Probe)
 
 	noOpTracer, tracerCloser, _ := tracing.NewTracer(&tracing.Options{
@@ -375,7 +383,7 @@ func TestParseName(t *testing.T) {
 		pk, _ := crypto.GenerateSecp256k1Key()
 		signer := crypto.NewDefaultSigner(pk)
 
-		s := api.New(pk.PublicKey, pk.PublicKey, common.Address{}, log, nil, nil, nil, 1, false, false, nil, []string{"*"})
+		s := api.New(pk.PublicKey, pk.PublicKey, common.Address{}, log, nil, nil, 1, false, false, nil, []string{"*"})
 		s.Configure(signer, nil, nil, api.Options{}, api.ExtraOptions{Resolver: tC.res}, 1, nil)
 		s.MountAPI()
 
@@ -704,4 +712,24 @@ func (c *chanStorer) SubscribePush(ctx context.Context, skipf func([]byte) bool)
 
 func (c *chanStorer) Close() error {
 	panic("not implemented") // TODO: Implement
+}
+
+func createRedistributionAgentService(
+	addr swarm.Address,
+	backend storageincentives.ChainBackend,
+	contract redistribution.Contract,
+	blocksPerRound uint64,
+	blocksPerPhase uint64, storer storage.StateStorer) *storageincentives.Agent {
+
+	postageContract := contractMock.New(contractMock.WithExpiresBatchesFunc(func(context.Context) error {
+		return nil
+	}),
+		contractMock.WithGetRewardFunc(func(context.Context, common.Address) (*big.Int, error) {
+			return nil, nil
+		}),
+	)
+	stakingContract := mock2.New(mock2.WithIsFrozen(func(context.Context) (bool, error) {
+		return true, nil
+	}))
+	return storageincentives.New(addr, backend, log.Noop, nil, contract, postageContract, postageContract, stakingContract, mockbatchstore.New(mockbatchstore.WithReserveState(&postage.ReserveState{StorageRadius: 0})), nil, time.Millisecond*10, blocksPerRound, blocksPerPhase, storer)
 }
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -806,7 +806,7 @@ type mockContract struct {
 	mtx       sync.Mutex
 }
 
-func (m *mockContract) GetFee() *big.Int {
+func (m *mockContract) Fee() *big.Int {
 	return nil
 }
 
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -806,8 +806,8 @@ type mockContract struct {
 	mtx       sync.Mutex
 }
 
-func (m *mockContract) Fee() *big.Int {
-	return nil
+func (m *mockContract) Fee(ctx context.Context, txHash common.Hash) *big.Int {
+	return big.NewInt(1000)
 }
 
 func (m *mockContract) ReserveSalt(context.Context) ([]byte, error) {
@@ -825,25 +825,25 @@ func (m *mockContract) IsWinner(context.Context) (bool, error) {
 	return false, nil
 }
 
-func (m *mockContract) Claim(context.Context) (*big.Int, error) {
+func (m *mockContract) Claim(context.Context) (common.Hash, error) {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
 	m.callsList = append(m.callsList, claimCall)
-	return nil, nil
+	return common.Hash{}, nil
 }
 
-func (m *mockContract) Commit(context.Context, []byte, *big.Int) (*big.Int, error) {
+func (m *mockContract) Commit(context.Context, []byte, *big.Int) (common.Hash, error) {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
 	m.callsList = append(m.callsList, commitCall)
-	return nil, nil
+	return common.Hash{}, nil
 }
 
-func (m *mockContract) Reveal(context.Context, uint8, []byte, []byte) (*big.Int, error) {
+func (m *mockContract) Reveal(context.Context, uint8, []byte, []byte) (common.Hash, error) {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
 	m.callsList = append(m.callsList, revealCall)
-	return nil, nil
+	return common.Hash{}, nil
 }
 
 type mockMonitor struct {
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -199,7 +199,7 @@ func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.
 	if o.beeMode == api.LightMode {
 		o.beeMode = api.FullMode
 	}
-	s := api.New(o.PublicKey, o.PSSPublicKey, o.EthereumAddress, o.Logger, transaction, o.BatchStore, o.beeMode, true, true, backend, o.CORSAllowedOrigins)
+	s := api.New(o.PublicKey, o.PSSPublicKey, o.EthereumAddress, o.Logger, transaction, o.BatchStore, statestore.NewStateStore(), o.beeMode, true, true, backend, o.CORSAllowedOrigins)
 
 	s.SetP2P(o.P2P)
 	s.SetSwarmAddress(&o.Overlay)
@@ -375,7 +375,7 @@ func TestParseName(t *testing.T) {
 		pk, _ := crypto.GenerateSecp256k1Key()
 		signer := crypto.NewDefaultSigner(pk)
 
-		s := api.New(pk.PublicKey, pk.PublicKey, common.Address{}, log, nil, nil, 1, false, false, nil, []string{"*"})
+		s := api.New(pk.PublicKey, pk.PublicKey, common.Address{}, log, nil, nil, nil, 1, false, false, nil, []string{"*"})
 		s.Configure(signer, nil, nil, api.Options{}, api.ExtraOptions{Resolver: tC.res}, 1, nil)
 		s.MountAPI()
 
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -825,25 +825,25 @@ func (m *mockContract) IsWinner(context.Context) (bool, error) {
 	return false, nil
 }
 
-func (m *mockContract) Claim(context.Context) error {
+func (m *mockContract) Claim(context.Context) (*big.Int, error) {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
 	m.callsList = append(m.callsList, claimCall)
-	return nil
+	return nil, nil
 }
 
-func (m *mockContract) Commit(context.Context, []byte, *big.Int) error {
+func (m *mockContract) Commit(context.Context, []byte, *big.Int) (*big.Int, error) {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
 	m.callsList = append(m.callsList, commitCall)
-	return nil
+	return nil, nil
 }
 
-func (m *mockContract) Reveal(context.Context, uint8, []byte, []byte) error {
+func (m *mockContract) Reveal(context.Context, uint8, []byte, []byte) (*big.Int, error) {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
 	m.callsList = append(m.callsList, revealCall)
-	return nil
+	return nil, nil
 }
 
 type mockMonitor struct {
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/ethersphere/bee/pkg/storageincentives"
 	mock2 "github.com/ethersphere/bee/pkg/storageincentives/staking/mock"
 	"github.com/ethersphere/bee/pkg/swarm/test"
+	"github.com/ethersphere/bee/pkg/transaction"
 	"io"
 	"math/big"
 	"net"
@@ -216,7 +217,7 @@ func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.
 	s.SetP2P(o.P2P)
 
 	if o.redistributionAgent == nil {
-		o.redistributionAgent = createRedistributionAgentService(o.Overlay, o.StateStorer, erc20)
+		o.redistributionAgent = createRedistributionAgentService(o.Overlay, o.StateStorer, erc20, transaction)
 		s.SetRedistributionAgent(o.redistributionAgent)
 	}
 	s.SetSwarmAddress(&o.Overlay)
@@ -722,7 +723,7 @@ func (c *chanStorer) Close() error {
 	panic("not implemented") // TODO: Implement
 }
 
-func createRedistributionAgentService(addr swarm.Address, storer storage.StateStorer, erc20Service erc20.Service) *storageincentives.Agent {
+func createRedistributionAgentService(addr swarm.Address, storer storage.StateStorer, erc20Service erc20.Service, tranService transaction.Service) *storageincentives.Agent {
 	const blocksPerRound uint64 = 12
 	const blocksPerPhase uint64 = 4
 	postageContract := contractMock.New(contractMock.WithExpiresBatchesFunc(func(context.Context) error {
@@ -745,7 +746,7 @@ func createRedistributionAgentService(addr swarm.Address, storer storage.StateSt
 		block:       blocksPerRound}
 	contract := &mockContract{}
 
-	return storageincentives.New(addr, backend, log.Noop, &mockMonitor{}, contract, postageContract, stakingContract, mockbatchstore.New(mockbatchstore.WithReserveState(&postage.ReserveState{StorageRadius: 0})), &mockSampler{}, time.Millisecond*10, blocksPerRound, blocksPerPhase, storer, erc20Service)
+	return storageincentives.New(addr, backend, log.Noop, &mockMonitor{}, contract, postageContract, stakingContract, mockbatchstore.New(mockbatchstore.WithReserveState(&postage.ReserveState{StorageRadius: 0})), &mockSampler{}, time.Millisecond*10, blocksPerRound, blocksPerPhase, storer, erc20Service, tranService)
 }
 
 type mockchainBackend struct {
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -777,13 +777,6 @@ func (m *mockchainBackend) HeaderByNumber(context.Context, *big.Int) (*types.Hea
 	}, nil
 }
 
-type mockMonitor struct {
-}
-
-func (m *mockMonitor) IsFullySynced() bool {
-	return true
-}
-
 type contractCall int
 
 func (c contractCall) String() string {
@@ -851,11 +844,3 @@ func (m *mockContract) Reveal(context.Context, uint8, []byte, []byte) error {
 	m.callsList = append(m.callsList, revealCall)
 	return nil
 }
-
-//type mockSampler struct{}
-//
-//func (m *mockSampler) ReserveSample(context.Context, []byte, uint8, uint64) (storage.Sample, error) {
-//	return storage.Sample{
-//		Hash: test.RandomAddress(),
-//	}, nil
-//}
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -215,9 +215,11 @@ func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.
 
 	s.SetP2P(o.P2P)
 
-	o.redistributionAgent = createRedistributionAgentService(o.Overlay, o.StateStorer, erc20)
+	if o.redistributionAgent == nil {
+		o.redistributionAgent = createRedistributionAgentService(o.Overlay, o.StateStorer, erc20)
+		s.SetRedistributionAgent(o.redistributionAgent)
+	}
 	s.SetSwarmAddress(&o.Overlay)
-	s.SetRedistributionAgent(o.redistributionAgent)
 	s.SetProbe(o.Probe)
 
 	noOpTracer, tracerCloser, _ := tracing.NewTracer(&tracing.Options{
@@ -721,8 +723,8 @@ func (c *chanStorer) Close() error {
 }
 
 func createRedistributionAgentService(addr swarm.Address, storer storage.StateStorer, erc20Service erc20.Service) *storageincentives.Agent {
-	var blocksPerRound uint64 = 12
-	var blocksPerPhase uint64 = 4
+	const blocksPerRound uint64 = 12
+	const blocksPerPhase uint64 = 4
 	postageContract := contractMock.New(contractMock.WithExpiresBatchesFunc(func(context.Context) error {
 		return nil
 	}),
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -12,10 +12,13 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
+	"github.com/ethereum/go-ethereum/core/types"
 	contractMock "github.com/ethersphere/bee/pkg/postage/postagecontract/mock"
+	"github.com/ethersphere/bee/pkg/settlement/swap/erc20"
 	"github.com/ethersphere/bee/pkg/storageincentives"
 	"github.com/ethersphere/bee/pkg/storageincentives/redistribution"
 	mock2 "github.com/ethersphere/bee/pkg/storageincentives/staking/mock"
+	"github.com/ethersphere/bee/pkg/swarm/test"
 	"io"
 	"math/big"
 	"net"
@@ -209,7 +212,7 @@ func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.
 
 	s.SetP2P(o.P2P)
 	s.SetSwarmAddress(&o.Overlay)
-	o.redistributionAgent = createRedistributionAgentService(o.Overlay, nil, nil, 0, 0, o.StateStorer)
+	o.redistributionAgent = createRedistributionAgentService(o.Overlay, backend, nil, 12, 4, o.StateStorer, erc20)
 
 	s.SetRedistributionAgent(o.redistributionAgent)
 	s.SetProbe(o.Probe)
@@ -719,17 +722,141 @@ func createRedistributionAgentService(
 	backend storageincentives.ChainBackend,
 	contract redistribution.Contract,
 	blocksPerRound uint64,
-	blocksPerPhase uint64, storer storage.StateStorer) *storageincentives.Agent {
+	blocksPerPhase uint64, storer storage.StateStorer, erc20Service erc20.Service) *storageincentives.Agent {
 
 	postageContract := contractMock.New(contractMock.WithExpiresBatchesFunc(func(context.Context) error {
 		return nil
 	}),
-		contractMock.WithGetRewardFunc(func(context.Context, common.Address) (*big.Int, error) {
-			return nil, nil
-		}),
 	)
 	stakingContract := mock2.New(mock2.WithIsFrozen(func(context.Context) (bool, error) {
 		return true, nil
 	}))
-	return storageincentives.New(addr, backend, log.Noop, nil, contract, postageContract, postageContract, stakingContract, mockbatchstore.New(mockbatchstore.WithReserveState(&postage.ReserveState{StorageRadius: 0})), nil, time.Millisecond*10, blocksPerRound, blocksPerPhase, storer)
+	wait := make(chan struct{})
+	backend = &mockchainBackend{
+		limit: 144,
+		limitCallback: func() {
+			select {
+			case wait <- struct{}{}:
+			default:
+			}
+		},
+		incrementBy: 2,
+		block:       blocksPerRound}
+	contract = &mockContract{}
+	return storageincentives.New(addr, backend, log.Noop, nil, contract, postageContract, stakingContract, mockbatchstore.New(mockbatchstore.WithReserveState(&postage.ReserveState{StorageRadius: 0})), nil, time.Millisecond*10, blocksPerRound, blocksPerPhase, storer, erc20Service)
+}
+
+type mockchainBackend struct {
+	mu            sync.Mutex
+	incrementBy   uint64
+	block         uint64
+	limit         uint64
+	limitCallback func()
+}
+
+func (m *mockchainBackend) BlockNumber(context.Context) (uint64, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	ret := m.block
+	lim := m.limit
+	inc := m.incrementBy
+
+	if lim == 0 || ret+inc < lim {
+		m.block += inc
+	} else if m.limitCallback != nil {
+		m.limitCallback()
+		return 0, errors.New("reached limit")
+	}
+
+	return ret, nil
+}
+
+func (m *mockchainBackend) HeaderByNumber(context.Context, *big.Int) (*types.Header, error) {
+	return &types.Header{
+		Time: uint64(time.Now().Unix()),
+	}, nil
+}
+
+type mockMonitor struct {
+}
+
+func (m *mockMonitor) IsFullySynced() bool {
+	return true
+}
+
+type contractCall int
+
+func (c contractCall) String() string {
+	switch c {
+	case isWinnerCall:
+		return "isWinnerCall"
+	case revealCall:
+		return "revealCall"
+	case commitCall:
+		return "commitCall"
+	case claimCall:
+		return "claimCall"
+	}
+	return "unknown"
+}
+
+const (
+	isWinnerCall contractCall = iota
+	revealCall
+	commitCall
+	claimCall
+)
+
+type mockContract struct {
+	callsList []contractCall
+	mtx       sync.Mutex
+}
+
+func (m *mockContract) GetFee() *big.Int {
+	return nil
+}
+
+func (m *mockContract) ReserveSalt(context.Context) ([]byte, error) {
+	return nil, nil
+}
+
+func (m *mockContract) IsPlaying(context.Context, uint8) (bool, error) {
+	return true, nil
+}
+
+func (m *mockContract) IsWinner(context.Context) (bool, error) {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	m.callsList = append(m.callsList, isWinnerCall)
+	return false, nil
+}
+
+func (m *mockContract) Claim(context.Context) error {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	m.callsList = append(m.callsList, claimCall)
+	return nil
+}
+
+func (m *mockContract) Commit(context.Context, []byte, *big.Int) error {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	m.callsList = append(m.callsList, commitCall)
+	return nil
+}
+
+func (m *mockContract) Reveal(context.Context, uint8, []byte, []byte) error {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	m.callsList = append(m.callsList, revealCall)
+	return nil
+}
+
+type mockSampler struct{}
+
+func (m *mockSampler) ReserveSample(context.Context, []byte, uint8, uint64) (storage.Sample, error) {
+	return storage.Sample{
+		Hash: test.RandomAddress(),
+	}, nil
 }
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -18,7 +18,6 @@ import (
 	"github.com/ethersphere/bee/pkg/storageincentives"
 	"github.com/ethersphere/bee/pkg/storageincentives/redistribution"
 	mock2 "github.com/ethersphere/bee/pkg/storageincentives/staking/mock"
-	"github.com/ethersphere/bee/pkg/swarm/test"
 	"io"
 	"math/big"
 	"net"
@@ -853,10 +852,10 @@ func (m *mockContract) Reveal(context.Context, uint8, []byte, []byte) error {
 	return nil
 }
 
-type mockSampler struct{}
-
-func (m *mockSampler) ReserveSample(context.Context, []byte, uint8, uint64) (storage.Sample, error) {
-	return storage.Sample{
-		Hash: test.RandomAddress(),
-	}, nil
-}
+//type mockSampler struct{}
+//
+//func (m *mockSampler) ReserveSample(context.Context, []byte, uint8, uint64) (storage.Sample, error) {
+//	return storage.Sample{
+//		Hash: test.RandomAddress(),
+//	}, nil
+//}
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ethersphere/bee/pkg/settlement/swap/erc20"
 	"github.com/ethersphere/bee/pkg/storageincentives"
 	mock2 "github.com/ethersphere/bee/pkg/storageincentives/staking/mock"
+	"github.com/ethersphere/bee/pkg/swarm/test"
 	"io"
 	"math/big"
 	"net"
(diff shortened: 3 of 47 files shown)


pkg/storageincentives/agent.go
```diff
@@ -413,13 +464,60 @@ func (a *Agent) Close() error {
...
+	err := a.nodeState.Put(fmt.Sprintf("%s%s", redistributionStatusKey, a.overlay.String()), a.nodeStatus)
+	if err != nil {
+		a.logger.Error(err, "error saving node status")
+		return
+	}
+}
+
+// GetStatus returns the node status
+func (a *Agent) GetStatus() (status NodeStatus, err error) {
+	return status, a.nodeState.Get(fmt.Sprintf("%s%s", redistributionStatusKey, a.overlay.String()), &status)
```
istae, line 512: you do not overlay as part of the key

pkg/storageincentives/agent.go
```diff
@@ -49,25 +55,30 @@ type Agent struct {
 	monitor        Monitor
 	contract       redistribution.Contract
 	batchExpirer   postagecontract.PostageBatchExpirer
+	stake          staking.RedistributionStatus
 	reserve        postage.Storer
+	nodeState      storage.StateStorer
```
istae, line 60: `statestore` as the name should be okay

pkg/storageincentives/events.go
```diff
@@ -17,6 +17,8 @@ const (
 	claim
 	sample
 	sampleEnd
+	winner
```
istae, line 20: these are the different phases, so these new values do not belong here

pkg/storageincentives/agent.go
```diff
@@ -298,8 +320,25 @@ func (a *Agent) claim(ctx context.Context) error {
 	}
 
 	if isWinner {
+		a.mtx.Lock()
+		a.nodeStatus.State = winner.String()
```
istae, line 324: these "state" field is a bit ambiguous, it stores the current phase and other things like if the node won. 
Ideally there should be a `currentPhase` field, `isFrozen`, and `lastWonRound`

pkg/storageincentives/agent.go
```diff
@@ -268,21 +281,30 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 		}
 
 		prevPhase = currentPhase
-
 		mtx.Unlock()
+		a.mtx.Lock()
```
istae, line 285: this should be in the if statement above

pkg/storageincentives/agent.go
```diff
@@ -268,21 +281,30 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
...
+	a.setNodeStatusFee(a.contract.GetFee())
 	if err != nil {
 		a.metrics.ErrReveal.Inc()
 	}
 	return err
 }
 
 func (a *Agent) claim(ctx context.Context) error {
+	defer func() {
+		a.saveStatus()
```
istae, line 306: I think the code can be cleaned up by moving the new "state" related things to it's own separate package under the `storageincentives`.

istae (changes requested):
Good initial commit!

pkg/api/redistribution.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/storageincentives"
+	"github.com/ethersphere/bee/pkg/tracing"
+	"net/http"
+)
+
+func (s *Service) redistributionStatusHandler(w http.ResponseWriter, r *http.Request) {
+	logger := tracing.NewLoggerWithTraceID(r.Context(), s.logger.WithName("redistribution_status").Build())
+	var status storageincentives.NodeStatus
+	err := s.stateStorer.Get(s.overlay.String(), status)
```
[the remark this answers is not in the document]
umerm-work: This is an old change. I have added an agent as dependency.

pkg/storageincentives/agent.go
umerm-work: According to the issue, the state should be either `claim | commit | reveal | idle | frozen`

We already have a `frozen` field in the state. So the `isFrozen` property might be redundant. Could you please  elaborate more on `currentPhase` and `lastWonRound` ?

pkg/storageincentives/events.go
umerm-work: Where should they blong?

pkg/api/redistribution.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/ethersphere/bee/pkg/jsonhttp"
+	"github.com/ethersphere/bee/pkg/tracing"
+	"net/http"
+)
+
+type NodeStatusResponse struct {
+	State  string `json:"state"`
+	Round  uint64 `json:"round"`
+	Block  uint64 `json:"block"`
+	Reward string `json:"reward"`
```
mrekucci, line 17: The `NodeStatusResponse` shouldn't be exported.  Also, why it is needed when it basically is the same as `storageincentives.NodeStatus` (since both have JSON tags)?

pkg/api/redistribution.go
```diff
@@ -0,0 +1,39 @@
...
+		logger.Error(nil, "get redistribution status")
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+
+	jsonhttp.OK(w, NodeStatusResponse{
+		State:  status.State,
+		Round:  status.Round,
+		Block:  status.Block,
+		Reward: status.Reward.String(),
```
mrekucci, line 36: The `Reward` and `Fees` fields should be of type `*big.Int` and then `bigint.Wrap()` can be used to handle the conversion as is done in other places.

pkg/api/redistribution.go
```diff
@@ -0,0 +1,39 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
```
mrekucci, line 1: 2021 -> 2023

pkg/api/redistribution_test.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
```
mrekucci, line 1: 2022 -> 2023

pkg/api/redistribution.go
```diff
@@ -0,0 +1,39 @@
...
+type NodeStatusResponse struct {
+	State  string `json:"state"`
+	Round  uint64 `json:"round"`
+	Block  uint64 `json:"block"`
+	Reward string `json:"reward"`
+	Fees   string `json:"fees"`
+}
+
+func (s *Service) redistributionStatusHandler(w http.ResponseWriter, r *http.Request) {
+	logger := tracing.NewLoggerWithTraceID(r.Context(), s.logger.WithName("redistribution_status").Build())
```
mrekucci, line 22: The pattern for naming logger handlers is **method_endpointname** so in this case it will be `get_redistributionstate`.

pkg/storageincentives/agent.go
```diff
@@ -25,7 +28,10 @@ import (
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
-const loggerName = "storageincentives"
+const (
+	loggerName              = "storageincentives"
```
mrekucci, line 32: Grouping constants and variables is recommended if they are somehow related, which is not the case here.

pkg/storageincentives/agent.go
```diff
@@ -49,25 +55,30 @@ type Agent struct {
...
 	batchExpirer   postagecontract.PostageBatchExpirer
+	stake          staking.RedistributionStatus
 	reserve        postage.Storer
+	nodeState      storage.StateStorer
 	sampler        storage.Sampler
 	overlay        swarm.Address
 	quit           chan struct{}
 	wg             sync.WaitGroup
+	nodeStatus     NodeStatus
+	mtx            sync.Mutex
```
mrekucci, line 66: What is this mutex guarding?

pkg/storageincentives/agent.go
```diff
@@ -49,25 +55,30 @@ type Agent struct {
...
+	erc20Service   erc20.Service
+	initialBalance *big.Int // current balance of the node before starting the round
+}
+
+// NodeStatus provide internal status of the nodes in the redistribution game
+type NodeStatus struct {
+	State  string   `json:"state"`
```
mrekucci, line 73: Why is this of type `string` and not `PhaseType`?

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -209,3 +212,11 @@ func (c *contract) callTx(ctx context.Context, callData []byte) ([]byte, error)
 	}
 	return result, nil
 }
+
+func (c *contract) setFee(in *big.Int) {
+	c.accumulativeFee = in
+}
+
+func (c *contract) GetFee() *big.Int {
```
mrekucci, line 220: The naming convention for Golang getters is that they do not contain the `Get` prefix.

mrekucci (changes requested):
Please update the

pkg/api/redistribution.go
umerm-work: API response required `reward` and `fee` to be in `string` that is why I have separated it.

pkg/storageincentives/agent.go
umerm-work: Good suggestion.

pkg/storageincentives/agent.go
umerm-work: In node status. The state is being set at multiple places where there are chances of race conditions.

pkg/storageincentives/agent.go
umerm-work: I am removing the overlay from this in new implementation.

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,121 @@
...
+	Block  uint64    `json:"block"`
+	Reward *big.Int  `json:"reward"`
+	Fees   *big.Int  `json:"fees"`
+}
+
+func NewNode(logger log.Logger, stateStore storage.StateStorer, erc20Service erc20.Service) NodeState {
+	return NodeState{
+		stateStore:     stateStore,
+		erc20Service:   erc20Service,
+		logger:         logger.WithName("nodestatus").Register(),
```
mrekucci, line 45: Follow the convention it the other .go files and declare the logger name as constant.

pkg/transaction/transaction.go
```diff
@@ -152,6 +155,19 @@ func (t *transactionService) waitForAllPendingTx() error {
 
 	return nil
 }
+func (t *transactionService) setAccumulativeFee(gasTipCap, gasFeeCap *big.Int, gasLimit uint64) {
```
istae, line 158: this is not an accurate way to compute the fee.
What needs to be done is check the transaction info after the transaction is mined in a block, and get the price and amount of gas used for the transaction.

pkg/storageincentives/agent.go
```diff
@@ -42,59 +44,52 @@ type Monitor interface {
...
+		logger:                 logger.WithName(loggerName).Register(),
+		contract:               contract,
+		batchExpirer:           batchExpirer,
+		reserve:                reserve,
+		monitor:                monitor,
+		blocksPerRound:         blocksPerRound,
+		sampler:                sampler,
+		quit:                   make(chan struct{}),
+		redistributionStatuser: redistributionStatuser,
+		nodeState:              NewNode(log.Noop, stateStore, erc20Service),
```
istae, line 77: replace `log.Noop` ?

pkg/storageincentives/events.go
istae: so instead of passing `PhaseType` to `SetPhase`, i would simply pass a string, or redefine these "phases" in the nodestatus.go under a different name.

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,120 @@
...
+		nodeStatus: NodeStatus{
+			Round:  0,
+			Block:  0,
+			Reward: nil,
+			Fees:   nil,
+		},
+	}
+}
+
+func (n *NodeState) SetPhase(p PhaseType) {
```
istae, line 55: it is problematic to include the "frozen" state in the function, because when the next phase happens, for example, if the phase is currently "frozen" and then we move on to the "claim" phase, it will override the "frozen" state.

this why SetPhase should only include the claim, commit, and reveal phases, and you should have a new function and field field for freeze and winner.

pkg/storageincentives/events.go
```diff
@@ -17,6 +17,9 @@ const (
...
+	frozen
+	idle
```
istae, line 22: is this used anywhere?

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (n *NodeState) SetBlock(b uint64) {
+	n.mtx.Lock()
+	defer n.mtx.Unlock()
+	n.nodeStatus.Block = b
+}
+
+// SetFee sets the internal node status
+func (n *NodeState) SetFee(fee *big.Int) {
```
istae, line 74: rename to AddFee

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (n *NodeState) SetBalance() error {
+	// get current balance
+	currentBalance, err := n.erc20Service.BalanceOf(context.Background(), common.HexToAddress(n.overlay.String()))
+	if err != nil {
+		n.logger.Debug("error getting balance", "error", err, "overly address", n.overlay.String())
+		return err
+	}
+	n.initialBalance.Set(currentBalance)
```
istae, line 118: `initialBalance` should be named `currentBalance`

pkg/transaction/transaction.go
istae: I believe the receipt has some info, if not, we need to figure out another way

pkg/storageincentives/agent.go
```diff
@@ -254,35 +248,43 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
...
-
 		mtx.Unlock()
+
 	}
 }
 
 func (a *Agent) reveal(ctx context.Context, storageRadius uint8, sample, obfuscationKey []byte) error {
 	a.metrics.RevealPhase.Inc()
 	err := a.contract.Reveal(ctx, storageRadius, sample, obfuscationKey)
+	a.nodeState.SetFee(a.contract.Fee())
```
istae, line 277: we could pass the contract as a dependency to the nodeState, so you don't have to call `a.contract.Fee()` here and SetFee can call it internally

pkg/storageincentives/agent.go
```diff
@@ -298,8 +300,23 @@ func (a *Agent) claim(ctx context.Context) error {
 	}
 
 	if isWinner {
+		a.nodeState.SetPhase(winner)
```
istae, line 303: againt, winner and frozen are not "phases" they are more like the current state,
claim, reveal, and commit are the real phases

istae (changes requested):

pkg/storageincentives/agent.go
umerm-work: Oops.

pkg/storageincentives/events.go
umerm-work: `idle` should be the initial state but let me add it. Haven't initiated anywhere.

pkg/storageincentives/agent.go
umerm-work: Alright. So going to add the state separately.

pkg/storageincentives/nodestatus.go
umerm-work: My understanding is that if the node is `Frozen`, it cannot participate in the game. Separating Phase and State would be confusing.

pkg/api/redistribution_test.go
```diff
@@ -0,0 +1,46 @@
...
+		})
+		if err != nil {
+			t.Errorf("redistribution put state: %v", err)
+		}
+		srv, _, _, _ := newTestServer(t, testServerOptions{
+			DebugAPI:    true,
+			StateStorer: store,
+		})
+		jsonhttptest.Request(t, srv, http.MethodGet, "/redistributionstate", http.StatusOK,
+			jsonhttptest.WithRequestHeader("Content-Type", "application/json; charset=utf-8"),
```
mrekucci, line 36: The JSON response should also be checked.

pkg/transaction/transaction.go
umerm-work:
```
type Receipt struct {
	// Consensus fields: These fields are defined by the Yellow Paper
	Type              uint8  `json:"type,omitempty"`
	PostState         []byte `json:"root"`
	Status            uint64 `json:"status"`
	CumulativeGasUsed uint64 `json:"cumulativeGasUsed" gencodec:"required"`
	Bloom             Bloom  `json:"logsBloom"         gencodec:"required"`
	Logs              []*Log `json:"logs"              gencodec:"required"`
	TxHash          common.Hash    `json:"transactionHash" gencodec:"required"`
	ContractAddress common.Address `json:"contractAddress"`
	GasUsed         uint64         `json:"gasUsed" gencodec:"required"`
	BlockHash        common.Hash `json:"blockHash,omitempty"`
	BlockNumber      *big.Int    `json:"blockNumber,omitempty"`
	TransactionIndex uint        `json:"transactionIndex"`
}
```
The receipt doesn't have any info.

umerm-work: I was not able to find fee data anywhere else. Maybe @swarmHaseeb can explain.

pkg/storageincentives/nodestatus.go
istae: it cannot participate, yes, but the agent will keep running, going through each phase, and updating this field

umerm-work: Okay, so I have separated them now.

pkg/storageincentives/staking/contract.go
```diff
@@ -40,6 +40,11 @@ type Contract interface {
 	DepositStake(ctx context.Context, stakedAmount *big.Int) (common.Hash, error)
 	GetStake(ctx context.Context) (*big.Int, error)
 	WithdrawAllStake(ctx context.Context) (common.Hash, error)
+	RedistributionStatUser
+}
+
+type RedistributionStatUser interface {
```
mrekucci, line 46: `RedistributionStatUser` -> `RedistributionStatuser`.

pkg/storageincentives/staking/contract.go
```diff
@@ -40,6 +40,11 @@ type Contract interface {
...
+	IsFrozen(ctx context.Context) (bool, error)
```
mrekucci, line 47: `IsFrozen` -> `IsOverlayFrozen`

mrekucci (commented):
Please also include the file name of the associated OpenAPI version change in the PR description.

pkg/storageincentives/agent.go
```diff
@@ -109,6 +104,7 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 		phaseEvents    = newEvents()
 	)
 
+	a.nodeState.SetState(idle)
```
istae, line 107: not sure if we need this

pkg/storageincentives/agent.go
```diff
@@ -254,35 +249,43 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
...
-
 		mtx.Unlock()
+
 	}
 }
 
 func (a *Agent) reveal(ctx context.Context, storageRadius uint8, sample, obfuscationKey []byte) error {
 	a.metrics.RevealPhase.Inc()
 	err := a.contract.Reveal(ctx, storageRadius, sample, obfuscationKey)
+	a.nodeState.AddFee()
```
istae, line 278: I see that the `SaveStatus` is missing here, and calling it everywhere pollutes the code imo. 
maybe it's best that you simply save to disk on every call to the nodeState and remove all the `SaveStatus`

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,152 @@
...
+
+// CalculateWinnerReward calculates the reward for the winner
+func (n *NodeState) CalculateWinnerReward() error {
+	currentBalance, err := n.erc20Service.BalanceOf(context.Background(), common.HexToAddress(n.overlay.String()))
+	if err != nil {
+		n.logger.Debug("error getting balance", "error", err, "overly address", n.overlay.String())
+		return err
+	}
+	if currentBalance != nil {
+		n.nodeStatus.Reward = currentBalance.Sub(currentBalance, n.currentBalance)
```
istae, line 122: so this calculates the most recent reward won,
do we also need to keep track of the total reward we have won so far for all the rounds we won?

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/ethersphere/bee/pkg/log"
+	"github.com/ethersphere/bee/pkg/settlement/swap/erc20"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/storageincentives/redistribution"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"math/big"
+	"sync"
+)
+
+const redistributionStatusKey = "redistribution_state_"
```
istae, line 19: the last underscore "_" can be removed

pkg/transaction/transaction.go
```diff
@@ -152,6 +155,19 @@ func (t *transactionService) waitForAllPendingTx() error {
...
+	gasTip := big.NewInt(0)
+	gasFee := big.NewInt(0)
+	if gasTipCap != nil {
+		*gasTip = *gasTipCap
+	}
+	*gasFee = *gasFeeCap
+	t.AccumulativeFee = gasTipCap.Add(gasTip, gasFee.Mul(gasFee, big.NewInt(int64(gasLimit))))
+}
+
+func (t *transactionService) GetAccumulativeFee() *big.Int {
```
istae, line 168: this returns the fee for ALL transaction which also includes transaction not related to the storage incentives, I don't think this is the right way to do it.

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -186,7 +189,7 @@ func (c *contract) sendAndWait(ctx context.Context, request *transaction.TxReque
 	if err != nil {
 		return err
 	}
-
+	c.setFee(c.txService.GetAccumulativeFee())
```
istae, line 192: like I wrote, this is not the right way to do it.

instead, we should have a new function in the transaction service that would return the transaction fee based on a transaction hash. 

and here you would pass the `txHash` returned from `txService.Send` to this new `TransactionFee` function.

how the transaction fee is acquired using the transaction hash requires a bit of research 
@swarmHaseeb @metacertain @vladopajic your input is welcome here

pkg/storageincentives/agent.go
umerm-work: It was mentioned in issue so included this.

pkg/storageincentives/agent.go
umerm-work: It would be redundant to call `saveStatus` on every call because sometimes we are setting 2-3 values at once, and `saveStatus` will then be called 2-3 times as well.

Ideally, we should call `saveStatus` when we lose or win, or fail at some point.

pkg/storageincentives/nodestatus.go
umerm-work: So this is a cumulative reward. We only require to store the total reward. Tracking is not required at the moment.

pkg/storageincentives/agent.go
```diff
@@ -413,13 +441,18 @@ func (a *Agent) Close() error {
...
-	data := append(s.overlay.Bytes(), storageRadiusByte...)
+	data := append(a.overlay.Bytes(), storageRadiusByte...)
 	data = append(data, sample...)
 	data = append(data, key...)
 
 	return crypto.LegacyKeccak256(data)
 }
+
+// Status returns the node status
+func (a *Agent) Status() (status NodeStatus, err error) {
```
mrekucci, line 456: The named return values are not needed here.

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,152 @@
...
+	defer n.mtx.Unlock()
+	err := n.stateStore.Put(redistributionStatusKey, n.nodeStatus)
+	if err != nil {
+		n.logger.Error(err, "error saving node status")
+		return
+	}
+}
+
+// Status returns the node status
+func (n *NodeState) Status() (status NodeStatus, err error) {
```
mrekucci, line 139: Make the returned `NodeStatus` pointer and avoid named return values, it makes it harder to follow the code logic. The following will do:
```golang
status := new(NodeStatus)
if err := n.stateStore.Get(redistributionStatusKey, status); err != nil {
	return nil, err
}
return status, nil
```

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,152 @@
...
+		return err
+	}
+	if currentBalance != nil {
+		n.nodeStatus.Reward = currentBalance.Sub(currentBalance, n.currentBalance)
+	}
+	return nil
+}
+
+// SaveStatus saves the node status to the database
+func (n *NodeState) SaveStatus() {
```
mrekucci, line 128: Not sure why the `nodeStatus` needs to be stored in the database?

pkg/storageincentives/agent.go
umerm-work: Correct.. Forgot to remove it

pkg/storageincentives/nodestatus.go
umerm-work: So that we can fetch it whenever we want using `/redistributionstate` endpoint.

umerm-work:
@istae, I have updated the fee using the transaction by the hash method.

pkg/storageincentives/staking/contract.go
```diff
@@ -313,3 +318,33 @@ func (c *contract) paused(ctx context.Context) (bool, error) {
...
+
+	if len(results) == 0 {
+		return false, errors.New("unexpected empty results")
+	}
+	// true if overlay is not frozen
+	if results[0].(bool) {
+		return false, nil
+	}
+	// false if overlay is frozen
+	return true, nil
```
vladopajic, line 349: this logic here seems inverted?

`if results[0].(bool)` then it should return true, but in code it returns false?

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -209,3 +211,15 @@ func (c *contract) callTx(ctx context.Context, callData []byte) ([]byte, error)
 	}
 	return result, nil
 }
+
+func (c *contract) setFee(ctx context.Context, txHash common.Hash) {
+	fee, err := c.txService.TransactionFee(ctx, txHash)
```
vladopajic, line 216: does this method needs to be guarded with mutex when value of `accumulativeFee` is set?

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,157 @@
...
+	fee := n.contract.Fee()
+	if fee != nil {
+		n.nodeStatus.Fees.Add(n.nodeStatus.Fees, fee)
+		n.SaveStatus()
+	}
+}
+
+// CalculateWinnerReward calculates the reward for the winner
+func (n *NodeState) CalculateWinnerReward() error {
+	currentBalance, err := n.erc20Service.BalanceOf(context.Background(), common.HexToAddress(n.overlay.String()))
```
vladopajic, line 117: would it make more sense if context is passed to this function instead of being created here?

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,157 @@
...
+	status := new(NodeStatus)
+	if err := n.stateStore.Get(redistributionStatusKey, status); err != nil {
+		return nil, err
+	}
+	return status, nil
+}
+
+func (n *NodeState) SetBalance() error {
+	// get current balance
+	currentBalance, err := n.erc20Service.BalanceOf(context.Background(), common.HexToAddress(n.overlay.String()))
```
vladopajic, line 150: would it make more sense if context is passed to this function instead of being created here?

pkg/storageincentives/staking/contract.go
umerm-work: Yes, this is inverted.
https://github.com/ethersphere/storage-incentives/blob/04be7ba384d77e60b2dc99e7c679bf98d529f052/src/Staking.sol#L76

Contract Logic
true  -> Node not frozen
false -> Node is frozen

`IsOverlayFrozen` is the opposite of a contract.

Therefore `if results[0].(bool)` true would be false. Means the overlay is not frozen.

pkg/storageincentives/nodestatus.go
umerm-work: Yes, thanks.

pkg/storageincentives/nodestatus.go
umerm-work: Done

pkg/storageincentives/redistribution/redistribution.go
umerm-work: https://github.com/ethersphere/bee/blob/a670af9fb998e88c57dfac43ad7e6b6f01ddfd78/pkg/storageincentives/nodestatus.go#L105

pkg/storageincentives/agent.go
```diff
@@ -254,28 +249,33 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
...
 
 			a.metrics.CurrentPhase.Set(float64(currentPhase))
 
 			a.logger.Info("entering phase", "phase", currentPhase.String(), "round", round, "block", block)
 
+			a.nodeState.SetPhase(currentPhase)
+			a.nodeState.SetRound(round)
+			a.nodeState.SetBlock(block)
+			a.nodeState.SaveStatus()
+
```
istae, line 262: Since you only use the SetPhase, SetRound, and SetBlock here, 
you can have a new single func that takes the three values and saves the status internally

pkg/storageincentives/agent.go
```diff
@@ -254,28 +249,33 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
...
 		mtx.Unlock()
+
 	}
 }
 
 func (a *Agent) reveal(ctx context.Context, storageRadius uint8, sample, obfuscationKey []byte) error {
 	a.metrics.RevealPhase.Inc()
 	err := a.contract.Reveal(ctx, storageRadius, sample, obfuscationKey)
+	a.nodeState.AddFee()
 	if err != nil {
```
istae, line 279: move error check above

pkg/storageincentives/agent.go
```diff
@@ -298,8 +298,23 @@ func (a *Agent) claim(ctx context.Context) error {
 	}
 
 	if isWinner {
+		a.nodeState.SetState(winner)
```
istae, line 301: these states do not show much info. for example, if we win one round, do we want to forever show that we won in some round without other info?
maybe you want something like `lastWonRound` instead of permanently having a state of 'winner'.

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,157 @@
...
+
+// CalculateWinnerReward calculates the reward for the winner
+func (n *NodeState) CalculateWinnerReward(ctx context.Context) error {
+	currentBalance, err := n.erc20Service.BalanceOf(ctx, common.HexToAddress(n.overlay.String()))
+	if err != nil {
+		n.logger.Debug("error getting balance", "error", err, "overly address", n.overlay.String())
+		return err
+	}
+	if currentBalance != nil {
+		n.nodeStatus.Reward.Add(n.nodeStatus.Reward, currentBalance.Sub(currentBalance, n.currentBalance))
```
istae, line 123: good

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -209,3 +211,15 @@ func (c *contract) callTx(ctx context.Context, callData []byte) ([]byte, error)
...
+	if err != nil {
+		c.logger.Info("transaction fee error:", err)
+	}
+	c.accumulativeFee = fee
```
istae, line 220: this needs to be similar to the total reward where we increment the accumulativeFee eg `accumulativeFee += trx_fee`

pkg/transaction/transaction.go
```diff
@@ -578,3 +581,11 @@ func (t *transactionService) Close() error {
 	t.wg.Wait()
 	return nil
 }
+
+func (t *transactionService) TransactionFee(ctx context.Context, txHash common.Hash) (*big.Int, error) {
+	hash, _, err := t.backend.TransactionByHash(ctx, txHash)
```
istae, line 586: rename to `trx`

pkg/transaction/transaction.go
```diff
@@ -578,3 +581,11 @@ func (t *transactionService) Close() error {
...
+	if err != nil {
+		return nil, err
+	}
+	return hash.GasPrice().Mul(hash.GasPrice(), big.NewInt(int64(hash.Gas()))), nil
```
istae, line 590: as it turns out, the transaction type returned by `TransactionByHash` has a function called `Cost()` which calculates what you do on line 590 for you :)

istae (changes requested):

pkg/storageincentives/staking/contract.go
vladopajic: would it be more readable if this part is written like so:

```
isNotFrozen := results[0].(bool)
return !isNotFrozen, nil
```

pkg/storageincentives/redistribution/redistribution.go
umerm-work: This lower layer should only return the fee. The business logic of adding fees should be done in node status, which I am already doing.

https://github.com/ethersphere/bee/blob/a670af9fb998e88c57dfac43ad7e6b6f01ddfd78/pkg/storageincentives/nodestatus.go#L110

pkg/transaction/transaction.go
umerm-work: Thanks God.

pkg/storageincentives/staking/contract.go
umerm-work: This would be nice.

pkg/storageincentives/agent.go
umerm-work: How about we set `a.nodeState.SetState(idle)` somewhere here https://github.com/ethersphere/bee/blob/776a000954c03261c8f6a4a07ef131398bb0524c/pkg/storageincentives/agent.go#L260

Then it won't show winner forever.

istae: it's not just about showing it forever, it's about giving users meaningful, useful info.
I propose adding a new `lastWonRound` field

pkg/storageincentives/redistribution/redistribution.go
istae: I propose that we can simply return the trx fee directly from the call so 
`a.contract.Reveal`, `a.contract.Commit`, and `a.contract.Claim` would also return the associated fees that you can increment in `nodeState`

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,167 @@
...
+	logger         log.Logger
+	overlay        swarm.Address
+	mtx            sync.Mutex
+	nodeStatus     NodeStatus
+	currentBalance *big.Int
+	contract       redistribution.Contract
+}
+
+// NodeStatus provide internal status of the nodes in the redistribution game
+type NodeStatus struct {
```
istae, line 34: the name should ideally be `incentivesStatus` or just `status`, nodeStatus is too general imo

istae (changes requested):

pkg/storageincentives/nodestatus.go
umerm-work: Done

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+}
+
+func (n *NodeState) SetCurrentEvent(p PhaseType, r uint64, b uint64) {
+	n.mtx.Lock()
+	defer n.mtx.Unlock()
+	n.status.Phase = p
+	n.status.Round = r
+	n.status.Block = b
+	n.status.State = idle
```
istae, line 87: if the node was frozen, this would remove the frozen state...

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,162 @@
...
+
+const (
+	winner State = iota + 1
+	frozen
+	idle
+)
+
+func (s State) String() string {
+	switch s {
+	case winner:
```
istae, line 54: we can now remove the winner state as we have the lastWonRound

pkg/storageincentives/agent.go
```diff
@@ -298,8 +295,24 @@ func (a *Agent) claim(ctx context.Context) error {
...
+
 		err = a.contract.Claim(ctx)
+		if err != nil {
+			a.logger.Info("calculate winner reward", "err", err)
+		}
+		err = a.nodeState.CalculateWinnerReward(ctx)
+		if err != nil {
+			a.logger.Info("calculate winner reward", "err", err)
+		}
+		a.nodeState.AddFee()
```
istae, line 314: you are missing SaveStatus calls in many places...
i think you only need to save status when you win, get frozen, or you play in a round

pkg/storageincentives/agent.go
```diff
@@ -323,6 +336,16 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
...
+	// true if frozen
+	isFrozen, err := a.redistributionStatuser.IsOverlayFrozen(ctx)
+	if err != nil {
+		a.logger.Info("error checking if stake is frozen", "err", err)
+	}
+	if isFrozen {
+		a.nodeState.SetState(frozen)
+		a.nodeState.SaveStatus()
+	}
+
```
istae, line 348: it would be good to have a lastPlayedRound field as well

istae (changes requested):

pkg/storageincentives/agent.go
umerm-work: @istae so `AddFee()` already has `saveStatus()`

https://github.com/ethersphere/bee/blob/a715eed2f81eed6d49164bae2a9606004f599261/pkg/storageincentives/nodestatus.go#L116

umerm-work: If you look into node status package. You will see SaveStatus is also being called in some functions.

pkg/storageincentives/nodestatus.go
umerm-work: Removed

pkg/storageincentives/agent.go
umerm-work: Done

pkg/storageincentives/nodestatus.go
umerm-work: Done

pkg/storageincentives/agent.go
```diff
@@ -323,12 +333,23 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
...
 	storageRadius := a.radius.StorageRadius()
 
+	// true if frozen
+	isFrozen, err := a.redistributionStatuser.IsOverlayFrozen(ctx)
+	if err != nil {
+		a.logger.Info("error checking if stake is frozen", "err", err)
+	}
+	a.nodeState.isFrozen(false)
+	if isFrozen {
+		a.nodeState.setLastFrozenRound(round)
```
istae, line 343: set should be capital S for consistency :)

pkg/storageincentives/agent.go
```diff
@@ -323,12 +333,23 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
 
 	storageRadius := a.radius.StorageRadius()
 
+	// true if frozen
+	isFrozen, err := a.redistributionStatuser.IsOverlayFrozen(ctx)
+	if err != nil {
+		a.logger.Info("error checking if stake is frozen", "err", err)
+	}
+	a.nodeState.isFrozen(false)
```
istae, line 341: this should be in an else statement after line 342

pkg/storageincentives/agent.go
```diff
@@ -323,12 +333,23 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
...
+	if isFrozen {
+		a.nodeState.setLastFrozenRound(round)
+		a.nodeState.isFrozen(true)
+	}
+
 	isPlaying, err := a.contract.IsPlaying(ctx, storageRadius)
 	if !isPlaying || err != nil {
```
istae, line 348: we should only increment the metric on error, NOT when isPlaying is false, I know you didnt write this but can you fix it in your PR

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,148 @@
...
+	Round           uint64
+	LastWonRound    uint64
+	LastPlayedRound uint64
+	LastFrozenRound uint64
+	Block           uint64
+	Reward          *big.Int
+	Fees            *big.Int
+}
+
+func NewRedistributionState(logger log.Logger, stateStore storage.StateStorer, erc20Service erc20.Service) RedistributionState {
```
istae, line 44: file name should be changed too

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,148 @@
...
+	defer n.mtx.Unlock()
+	n.status.LastPlayedRound = p
+	n.saveStatus()
+}
+
+// AddFee sets the internal node status
+func (n *RedistributionState) AddFee(fee *big.Int) {
+	n.mtx.Lock()
+	defer n.mtx.Unlock()
+	if fee != nil {
```
istae, line 100: why would fee ever be nil ?

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// CalculateWinnerReward calculates the reward for the winner
+func (n *RedistributionState) CalculateWinnerReward(ctx context.Context) error {
+	currentBalance, err := n.erc20Service.BalanceOf(ctx, common.HexToAddress(n.overlay.String()))
+	if err != nil {
+		n.logger.Debug("error getting balance", "error", err, "overly address", n.overlay.String())
+		return err
+	}
+	if currentBalance != nil {
```
istae, line 113: currentBalance would never be nil I think

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,148 @@
...
+		n.status.Reward.Add(n.status.Reward, currentBalance.Sub(currentBalance, n.currentBalance))
+	}
+	return nil
+}
+
+// saveStatus saves the node status to the database
+func (n *RedistributionState) saveStatus() {
+	n.mtx.Lock()
```
istae, line 121: we cannot lock here as the outer function already locks it

pkg/storageincentives/nodestatus.go
```diff
@@ -0,0 +1,148 @@
...
+func (n *RedistributionState) Status() (*Status, error) {
+	status := new(Status)
+	if err := n.stateStore.Get(redistributionStatusKey, status); err != nil {
+		return nil, err
+	}
+	return status, nil
+}
+
+func (n *RedistributionState) setBalance(ctx context.Context) error {
+	// get current balance
```
istae, line 140: might as well lock here

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -153,12 +154,12 @@ func (c *contract) Reveal(ctx context.Context, storageDepth uint8, reserveCommit
 		Value:                big.NewInt(0),
 		Description:          "reveal transaction",
 	}
-	err = c.sendAndWait(ctx, request, 50)
+	err, fee := c.sendAndWait(ctx, request, 50)
```
istae, line 157: err should be the last variable

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -181,21 +182,20 @@ func (c *contract) ReserveSalt(ctx context.Context) ([]byte, error) {
 	return salt[:], nil
 }
 
-func (c *contract) sendAndWait(ctx context.Context, request *transaction.TxRequest, boostPercent int) error {
+func (c *contract) sendAndWait(ctx context.Context, request *transaction.TxRequest, boostPercent int) (error, *big.Int) {
 	txHash, err := c.txService.Send(ctx, request, boostPercent)
 	if err != nil {
-		return err
+		return err, nil
```
istae, line 188: why is err not last :(

pkg/transaction/transaction.go
```diff
@@ -100,13 +102,14 @@ type transactionService struct {
...
-	chainID *big.Int
-	monitor Monitor
+	logger          log.Logger
+	backend         Backend
+	signer          crypto.Signer
+	sender          common.Address
+	store           storage.StateStorer
+	chainID         *big.Int
+	monitor         Monitor
+	AccumulativeFee *big.Int
```
istae, line 112: A should be lowercase

pkg/storageincentives/nodestatus.go
istae: please a comment to the function like `Must be called under lock.`

istae (changes requested):

pkg/storageincentives/agent.go
istae: also a.nodeState.isFrozen should only be called when err is nil

umerm-work: So by default, it should be `false`. `Else` statement is just an extra.

pkg/storageincentives/redistribution/redistribution.go
umerm-work: This is fixed.

pkg/storageincentives/nodestatus.go
umerm-work: This is just for safety purposes if there is a 1% chance.

pkg/storageincentives/agent.go
umerm-work: If it's `true` then the old value will be overwritten.

pkg/transaction/transaction.go
umerm-work: This variable is no more required.

pkg/storageincentives/agent.go
umerm-work: Let me make Initial capital for functions which are being used in agent.go.

pkg/storageincentives/nodestatus.go
istae: dont think we need it

umerm-work: ok

pkg/storageincentives/agent.go
```diff
@@ -323,12 +333,23 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
...
 	storageRadius := a.radius.StorageRadius()
 
-	isPlaying, err := a.contract.IsPlaying(ctx, storageRadius)
-	if !isPlaying || err != nil {
+	// true if frozen
+	isFrozen, err := a.redistributionStatuser.IsOverlayFrozen(ctx)
+	if err != nil {
+		a.logger.Info("error checking if stake is frozen", "err", err)
+	}
+	a.nodeState.IsFrozen(false)
```
istae, line 341: this should be 
```
if err == nil {
  if isFrozen {
      a.nodeState.SetLastFrozenRound(round)
    }
    a.nodeState.IsFrozen(isFrozen)
}

File: pkg/api/api.go [hunk 66]
```diff
@@ -195,6 +198,12 @@ func (s *Service) SetSwarmAddress(addr *swarm.Address) {
 	}
 }
 
+func (s *Service) SetRedistributionAgent(redistributionAgent *storageincentives.Agent) {
```
istae, line 201: this is not used anywhere

pkg/api/redistribution.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (s *Service) redistributionStatusHandler(w http.ResponseWriter, r *http.Request) {
+	logger := tracing.NewLoggerWithTraceID(r.Context(), s.logger.WithName("get_redistributionstate").Build())
+
+	status, err := s.redistributionAgent.Status()
+	if err != nil {
+		logger.Debug("get redistribution status", "overlay address", s.overlay.String(), "error", err)
+		logger.Error(nil, "get redistribution status")
+		w.WriteHeader(http.StatusNotFound)
```
istae, line 33: should be internal server error

pkg/storageincentives/redistributionstate.go
```diff
@@ -0,0 +1,146 @@
...
+func (n *RedistributionState) AddFee(fee *big.Int) {
+	n.mtx.Lock()
+	defer n.mtx.Unlock()
+	n.status.Fees.Add(n.status.Fees, fee)
+	n.saveStatus()
+}
+
+// CalculateWinnerReward calculates the reward for the winner
+func (n *RedistributionState) CalculateWinnerReward(ctx context.Context) error {
+	currentBalance, err := n.erc20Service.BalanceOf(ctx, common.HexToAddress(n.overlay.String()))
```
istae, line 106: should be locked

istae (changes requested):

pkg/api/api.go
umerm-work: I missed it in node.go file. Let me add it.

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -23,9 +23,9 @@ type Contract interface {
 	ReserveSalt(context.Context) ([]byte, error)
 	IsPlaying(context.Context, uint8) (bool, error)
 	IsWinner(context.Context) (bool, error)
-	Claim(context.Context) error
-	Commit(context.Context, []byte, *big.Int) error
-	Reveal(context.Context, uint8, []byte, []byte) error
+	Claim(context.Context) (*big.Int, error)
+	Commit(context.Context, []byte, *big.Int) (*big.Int, error)
+	Reveal(context.Context, uint8, []byte, []byte) (*big.Int, error)
```
vladopajic, line 28: Would it make sense if `Claim`, `Commit` and `Reveal` method return transaction hash `common.Hash` to the caller, instead of fee? 

If so, these would be advantages:
- This way caller would be able to have more information (should it need) and not just fee.
- Caller would be able to collect all transaction hashes and calculate fees in separate goroutne. This would reduce total execution time of  of `Claim`, `Commit` and `Reveal` methods, because with current change they have one more operation to execute (fetch transaction).
- Error handling for case when transaction could not be fetched could be better handled (instead of swallowing it here)

pkg/storageincentives/agent.go
```diff
@@ -254,35 +247,37 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
...
 }
 
 func (a *Agent) reveal(ctx context.Context, storageRadius uint8, sample, obfuscationKey []byte) error {
 	a.metrics.RevealPhase.Inc()
-	err := a.contract.Reveal(ctx, storageRadius, sample, obfuscationKey)
+	fee, err := a.contract.Reveal(ctx, storageRadius, sample, obfuscationKey)
 	if err != nil {
 		a.metrics.ErrReveal.Inc()
 	}
+	a.nodeState.AddFee(fee)
```
istae, line 276: if err is not nil, we should return the error, if err is nil, only then we add the fee

pkg/storageincentives/agent.go
```diff
@@ -298,8 +293,23 @@ func (a *Agent) claim(ctx context.Context) error {
...
 		a.metrics.Winner.Inc()
-		err = a.contract.Claim(ctx)
+		err := a.nodeState.SetBalance(ctx)
+		if err != nil {
+			a.logger.Info("could not set balance", "err", err)
+		}
+
+		fee, err := a.contract.Claim(ctx)
+		if err != nil {
+			a.logger.Info("calculate winner reward", "err", err)
```
istae, line 305: return err if claim fails

pkg/storageincentives/agent.go
```diff
@@ -323,12 +333,24 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
...
 	storageRadius := a.radius.StorageRadius()
 
-	isPlaying, err := a.contract.IsPlaying(ctx, storageRadius)
-	if !isPlaying || err != nil {
+	// true if frozen
+	isFrozen, err := a.redistributionStatuser.IsOverlayFrozen(ctx)
+	if err != nil {
+		a.logger.Info("error checking if stake is frozen", "err", err)
+	}
+	if err == nil {
```
istae, line 341: simple else works

pkg/storageincentives/agent.go
```diff
@@ -298,8 +293,23 @@ func (a *Agent) claim(ctx context.Context) error {
...
+		err := a.nodeState.SetBalance(ctx)
+		if err != nil {
+			a.logger.Info("could not set balance", "err", err)
+		}
+
+		fee, err := a.contract.Claim(ctx)
+		if err != nil {
+			a.logger.Info("calculate winner reward", "err", err)
+		}
+		err = a.nodeState.CalculateWinnerReward(ctx)
```
istae, line 307: we should not call this if a.nodeState.SetBalance(ctx) fails

istae (changes requested):

pkg/storageincentives/agent.go
```diff
@@ -298,8 +294,26 @@ func (a *Agent) claim(ctx context.Context) error {
...
+		}
+		if errBalance == nil {
+			err = a.nodeState.CalculateWinnerReward(ctx)
+			if err != nil {
+				a.logger.Info("calculate winner reward", "err", err)
+			}
+		}
+		a.nodeState.AddFee(fee)
+
 		if err != nil {
```
istae, line 317: bad left over code

istae (changes requested):

pkg/storageincentives/agent.go
```diff
@@ -298,8 +294,26 @@ func (a *Agent) claim(ctx context.Context) error {
...
-		err = a.contract.Claim(ctx)
+		errBalance := a.nodeState.SetBalance(ctx)
+		if errBalance != nil {
+			a.logger.Info("could not set balance", "err", err)
+		}
+
+		fee, err := a.contract.Claim(ctx)
+		if err != nil {
+			a.logger.Info("calculate winner reward", "err", err)
+			return err
```
istae, line 307: return fmt.Errorf("error claiming win: %w", err)

pkg/storageincentives/redistribution/redistribution.go
istae: this is also a good idea, probably fits the nature of the change better, as fee and reward calculations should not affect the participation in the game

pkg/storageincentives/agent.go
umerm-work: I don't get it ?

pkg/storageincentives/redistribution/redistribution.go
umerm-work: Sounds good.

pkg/storageincentives/agent.go
istae: which error is this checking ?

umerm-work: Its for  `err := a.contract.Claim(ctx)` 
Anyway, I have changed calculate error name to remove confusion.

istae: you are already checking it on line 305 😭

istae: I'd recommend reviewing the code yourself once more

umerm-work: I haven't added this check. I thought it might be for metrics. Anyway updated it.

istae (approved):

pkg/storageincentives/agent.go
```diff
@@ -323,12 +333,24 @@ func (a *Agent) play(ctx context.Context) (uint8, []byte, error) {
...
+		a.logger.Info("error checking if stake is frozen", "err", err)
+	}
+	if err == nil {
+		if isFrozen {
+			a.nodeState.SetLastFrozenRound(round)
+		}
+		a.nodeState.IsFrozen(isFrozen)
+	}
+
+	_, err = a.contract.IsPlaying(ctx, storageRadius)
```
istae, line 348: why is `isPlaying` boolen check missing?

istae, line 348: why is `isPlaying` boolean removed here ??
the node cannot play if isPlaying false ?

istae (changes requested):

pkg/storageincentives/agent.go
umerm-work: https://github.com/ethersphere/bee/pull/3683#discussion_r1085827359

You requested this change. :)

istae: "we should only increment the metric on error, NOT when isPlaying is false, I know you didnt write this but can you fix it in your PR"

I don't understand what you mean. 
The current code does not even check if we are playing, I did not request this. What I requested was the metric was being incremented, that's it :)

istae (approved):

pkg/storageincentives/redistributionstate.go
```diff
@@ -0,0 +1,151 @@
...
+	defer n.mtx.Unlock()
+	n.status.Fees.Add(n.status.Fees, n.contract.Fee(ctx, txHash))
+	n.saveStatus()
+}
+
+// CalculateWinnerReward calculates the reward for the winner
+func (n *RedistributionState) CalculateWinnerReward(ctx context.Context) error {
+	n.mtx.Lock()
+	defer n.mtx.Unlock()
+	currentBalance, err := n.erc20Service.BalanceOf(ctx, common.HexToAddress(n.overlay.String()))
```
vladopajic, line 111: Mutex in `CalculateWinnerReward` seem to be overlocking. 

Lock is acquired on entire method which would include call to external service `n.erc20Service.BalanceOf` in same lock. If this method call takes too much time then all call to all other methods of `RedistributionState` would block further propagating blocking to other logic.

Could we reduce locking of `CalculateWinnerReward` logic down to the smallest as possible?

openapi/Swarm.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 
 info:
-  version: 4.0.0
+  version: 4.1.0
```
mrekucci, line 4: The version should not be changed here as it is already indicated in the PR description.

pkg/api/api_test.go
```diff
@@ -705,3 +721,144 @@ func (c *chanStorer) SubscribePush(ctx context.Context, skipf func([]byte) bool)
 func (c *chanStorer) Close() error {
 	panic("not implemented") // TODO: Implement
 }
+
+func createRedistributionAgentService(addr swarm.Address, storer storage.StateStorer, erc20Service erc20.Service) *storageincentives.Agent {
+	var blocksPerRound uint64 = 12
```
mrekucci, line 726: These two variables should be constants.

pkg/storageincentives/redistributionstate.go
```diff
@@ -0,0 +1,151 @@
...
+	Round           uint64
+	LastWonRound    uint64
+	LastPlayedRound uint64
+	LastFrozenRound uint64
+	Block           uint64
+	Reward          *big.Int
+	Fees            *big.Int
+}
+
+func NewRedistributionState(logger log.Logger, stateStore storage.StateStorer, erc20Service erc20.Service, contract redistribution.Contract) RedistributionState {
```
vladopajic, line 46: `RedistributionState` methods like `SetCurrentEvent`, `IsFrozen`, `SetLastFrozenRound` and others all seem to do `saveStatus()` at the end. This operation will potentially access disk therefore could take significant time in some situations. 

Would it be possible to decouple disk access from logic of these methods?

-----

As easiest approach would it make sense to call `saveStatus` in goroutine? Let's assume `IsFronzen` method (but this would apply to other methods as well)

```
func (n *RedistributionState) IsFrozen(f bool) {
	n.mtx.Lock()
	defer n.mtx.Unlock()
	n.status.IsFrozen = f
	go saveStatus(n.stateStore, n.status) // 1*
}
```
1* `saveStatus` should changed to accept copy of `status` variable and store it in background; thus `IsFrozen` would decouple from disk io

I am not able to estimate how often is `saveStatus` called but would it make sense to reduce this method calls? For example would it be possible to `saveStatus` only once every second if there are some changes to the state already made?

mrekucci (approved):
Apart from a few minor things mentioned, it looks good.

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -209,3 +208,12 @@ func (c *contract) callTx(ctx context.Context, callData []byte) ([]byte, error)
 	}
 	return result, nil
 }
+
+func (c *contract) Fee(ctx context.Context, txHash common.Hash) *big.Int {
```
vladopajic, line 212: method `Fee` seems to not be responsability of this structure

pkg/storageincentives/redistributionstate.go
umerm-work: So we could only apply lock to this statement instead of locking all method

`n.status.Reward.Add(n.status.Reward, currentBalance.Sub(currentBalance, n.currentBalance))`

pkg/storageincentives/redistributionstate.go
```diff
@@ -0,0 +1,151 @@
...
+	n.status.Block = b
+	n.saveStatus()
+}
+
+func (n *RedistributionState) IsFrozen(f bool) {
+	n.mtx.Lock()
+	defer n.mtx.Unlock()
+	n.status.IsFrozen = f
+	n.saveStatus()
+}
```
vladopajic, line 76: would it make sense for `IsFrozen` to accept additional parameter `round` so this method could also set `LastFrozenRound` if frozen? if so then `SetLastFrozenRound` method would not be needed.

would name `SetFrozen` be more convenient for this method (following same convention)?

pkg/storageincentives/redistribution/redistribution.go
umerm-work: The fee is calculated when we perform transactions on `claim`, `reveal`, and `commit` functions. So we require a cumulative fee for these functions. That is why I have placed it here.

pkg/storageincentives/redistributionstate.go
```diff
@@ -0,0 +1,151 @@
...
+	status := new(Status)
+	if err := n.stateStore.Get(redistributionStatusKey, status); err != nil {
+		return nil, err
+	}
+	return status, nil
+}
+
+func (n *RedistributionState) SetBalance(ctx context.Context) error {
+	n.mtx.Lock()
+	defer n.mtx.Unlock()
```
vladopajic, line 142: method `SetBalance` seems to have same overlocking problem like `CalculateWinnerReward`

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -23,9 +23,10 @@ type Contract interface {
 	ReserveSalt(context.Context) ([]byte, error)
 	IsPlaying(context.Context, uint8) (bool, error)
 	IsWinner(context.Context) (bool, error)
-	Claim(context.Context) error
-	Commit(context.Context, []byte, *big.Int) error
-	Reveal(context.Context, uint8, []byte, []byte) error
+	Claim(context.Context) (common.Hash, error)
+	Commit(context.Context, []byte, *big.Int) (common.Hash, error)
+	Reveal(context.Context, uint8, []byte, []byte) (common.Hash, error)
+	Fee(ctx context.Context, txHash common.Hash) *big.Int
```
vladopajic, line 29: method `Fee` seems to not be responsability of this inteface

pkg/storageincentives/redistributionstate.go
umerm-work: Good suggestion.

pkg/storageincentives/redistribution/redistribution.go
vladopajic: sure, but this method is not depended on `contract` therefore it shouldn't be defined here. would it make sense to move this method to `RedistributionState` structure?

umerm-work: This method is already being called in `RedistributionState` structure `Fee` method.

vladopajic: If no one other then `RedistributionState` is calling this `Fee` method, then it doesn't make much sense to have this method defined here instead of having it defined in `RedistributionState`. This would reduce (remove it entirely) dependency between `RedistributionState` and `contract`.

pkg/storageincentives/agent_test.go
```diff
@@ -234,6 +233,10 @@ type mockContract struct {
 	mtx       sync.Mutex
 }
 
+func (m *mockContract) Fee(ctx context.Context, txhash common.Hash) *big.Int {
+	return big.NewInt(1000)
+}
+
```
vladopajic, line 239: Fee is not needed anymore

pkg/storageincentives/redistributionstate.go
```diff
@@ -0,0 +1,151 @@
...
+
+type RedistributionState struct {
+	stateStore     storage.StateStorer
+	erc20Service   erc20.Service
+	logger         log.Logger
+	overlay        swarm.Address
+	mtx            sync.Mutex
+	status         Status
+	currentBalance *big.Int
+	contract       transaction.Service
```
vladopajic, line 30: would name `trxService` be more suitable instead of  `contract` ?

vladopajic (approved):
LGTM; please make sure that CI is green before this PR is merged (in order to ensure that data race is fixed)

Issue #3254, opened 2022-09-07
clean-up new overlay mining logic


notanatol:
# Task
Because new joiners to the network can use the zero nonce for generating their overlay (version 1.8 and later) it means the overlay mining logic can be safely removed in 1.9 version.

## Acceptance criterea
No overlay mining or mining interruption code is present in the codebase.

PR #3684, opened 2023-01-03
chore: remove overlay mining

vladopajic:
### Checklist

### Description
Closes: #3254

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3684)

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -10,13 +10,10 @@ package node
 import (
 	"context"
 	"crypto/ecdsa"
-	"encoding/binary"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
 	stdlog "log"
-	"math"
 	"math/big"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


pkg/node/node.go
```diff
@@ -491,45 +487,6 @@ func NewBee(ctx context.Context, addr string, publicKey *ecdsa.PublicKey, signer
 	if err != nil {
 		return nil, fmt.Errorf("check presence of nonce: %w", err)
 	}
-	if !nonceExists {
```
istae, line 494: a fresh node may not have a nonce, so this is needed I think

istae: overlayNonceExists can be refactored to return an 32 byte zero slice in the case that an old nonce does not exist

istae (changes requested):

pkg/node/node.go
vladopajic: isn't this already the case?

istae: aha yes

istae (approved):

notanatol (approved):

PR #3690, opened 2023-01-11
refactor: upload store with tags

aloknerurkar:
### Checklist

### Description
After discussions within the team, it was agreed that the current tags-related functionality is not ideal. We need to pass the tag around on the chunk and across components using context key-value pairs. This has become pretty ugly. With the new localstore, we have a separate component called upload store which is used to consume uploads. It is possible to clean up all the ugly tag handling now by moving the tag-related logic into the upload store.

This PR refactors the originally envisioned upload store by also adding the tag functionality. Each putter can be assigned a new tag ID and hence serves as a session to upload data and keep track of the information related to this particular session. The Tag IDs can be thought of as file descriptors. They can be monotonically increasing integers that can be reset when they overflow. The new tag feature is compatible with the existing implementation. So this would be transparent to the users mostly.

Along with the tag, we will also maintain chunk information. So we can show the users which chunk was uploaded with which stamp and also the timestamp when it was synced to the network. A separate API can be provided to show this information if required. This information should be cleaned up when the stamps expire as there is no other way to clean up the entries right now.

We will also support uploading chunks with different stamps. These will be separate push requests. Also users will be able to see all the stamps with which a chunk was uploaded.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3690)

### DIFF
--- a/pkg/localstorev2/internal/cache/cache_test.go
+++ b/pkg/localstorev2/internal/cache/cache_test.go
@@ -15,8 +15,6 @@ import (
 	"github.com/ethersphere/bee/pkg/localstorev2/internal/cache"
 	chunktest "github.com/ethersphere/bee/pkg/storage/testing"
 	storage "github.com/ethersphere/bee/pkg/storagev2"
-	"github.com/ethersphere/bee/pkg/storagev2/inmemchunkstore"
-	inmem "github.com/ethersphere/bee/pkg/storagev2/inmemstore"
 	"github.com/ethersphere/bee/pkg/storagev2/storagetest"
 	"github.com/ethersphere/bee/pkg/swarm"
 	"github.com/google/go-cmp/cmp"
@@ -136,27 +134,12 @@ func TestCacheEntryItem(t *testing.T) {
 }
 
 type testStorage struct {
-	store   storage.Store
-	chStore *chunkStore
-	putFn   func(storage.Item) error
+	internal.Storage
+	putFn func(storage.Item) error
 }
-
-type chunkStore struct {
-	storage.ChunkStore
-}
-
-func (c *chunkStore) GetWithStamp(ctx context.Context, address swarm.Address, _ []byte) (swarm.Chunk, error) {
-	return c.Get(ctx, address)
-}
-
-func (testStorage) Ctx() context.Context { return context.TODO() }
 
 func (t *testStorage) Store() storage.Store {
-	return &wrappedStore{Store: t.store, putFn: t.putFn}
-}
-
-func (t *testStorage) ChunkStore() internal.ChunkStore {
-	return t.chStore
+	return &wrappedStore{Store: t.Storage.Store(), putFn: t.putFn}
 }
 
 type wrappedStore struct {
@@ -171,16 +154,27 @@ func (w *wrappedStore) Put(i storage.Item) error {
 	return w.Store.Put(i)
 }
 
+func newTestStorage(t *testing.T) *testStorage {
+	t.Helper()
+
+	storg, closer := internal.NewInmemStorage()
+	t.Cleanup(func() {
+		err := closer()
+		if err != nil {
+			t.Errorf("failed closing storage: %v", err)
+		}
+	})
+
+	return &testStorage{Storage: storg}
+}
+
 func TestCache(t *testing.T) {
 	t.Parallel()
 
 	t.Run("fresh new cache", func(t *testing.T) {
 		t.Parallel()
 
-		st := &testStorage{
-			store:   inmem.New(),
-			chStore: &chunkStore{inmemchunkstore.New()},
-		}
+		st := newTestStorage(t)
 		c, err := cache.New(st, 10)
 		if err != nil {
 			t.Fatal(err)
@@ -191,10 +185,7 @@ func TestCache(t *testing.T) {
 	t.Run("putter", func(t *testing.T) {
 		t.Parallel()
 
-		st := &testStorage{
-			store:   inmem.New(),
-			chStore: &chunkStore{inmemchunkstore.New()},
-		}
+		st := newTestStorage(t)
 		c, err := cache.New(st, 10)
 		if err != nil {
 			t.Fatal(err)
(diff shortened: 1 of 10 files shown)


pkg/localstorev2/internal/cache/cache_test.go
```diff
@@ -136,17 +137,29 @@ func TestCacheEntryItem(t *testing.T) {
 
 type testStorage struct {
 	store   storage.Store
-	chStore storage.ChunkStore
+	chStore *chunkStore
```
mrekucci, line 140: The name distinction is not so clear, so we could refer to them as `indexStore` and `chunkStore`, at least internally.

pkg/storagev2/chunkstore.go
```diff
@@ -17,21 +17,42 @@ type Getter interface {
...
+	// If the chunk has multiple stamps, then the first stamp is returned in this
+	// query. In order to deterministically get the stamp, use the GetterWithStamp
+	// interface.
 	Get(context.Context, swarm.Address) (swarm.Chunk, error)
 }
 
+// GetterWithStamp is the interface that provides a way to specify a postage
+// stamp to filter the Get query. This is required as we will support multiple stamps
+// on a chunk. As a result some components would need to query chunk and provide
+// the stamp to return by using the argument.
```
mrekucci, line 29: It should also be documented if it returns `storage.ErrNotFound` if the chunk is not found.

pkg/localstorev2/internal/upload/uploadstore.go
```diff
@@ -147,104 +96,320 @@ func (i pushItem) String() string {
...
-		switch exists, err := s.Store().Has(tai); {
-		case err != nil:
-			return fmt.Errorf("store has item %q call failed: %w", tai, err)
-		case exists:
-			return nil
-		}
-		if err := s.Store().Put(tai); err != nil {
-			return fmt.Errorf("store put item %q call failed: %w", tai, err)
-		}
+func NewPushReporter(s internal.Storage) storage.PushReporter {
```
mrekucci, line 346: A description would be nice to have. The same applies to the `Report` method below.

pkg/localstorev2/internal/upload/uploadstore_test.go
```diff
@@ -297,59 +462,132 @@ func TestChunkPutter(t *testing.T) {
...
+			t.Fatalf("Close(...): unexpected error %v", err)
+		}
+
+		ti := &upload.TagItem{TagID: tagID}
+		err = ts.Store().Get(ti)
+		if err != nil {
+			t.Fatalf("Get(...): unexpected error %v", err)
+		}
+
+		exp := &upload.TagItem{
```
mrekucci, line 579: For consistency, I'd recommend naming it `want`.

pkg/localstorev2/internal/upload/uploadstore_test.go
```diff
@@ -297,59 +462,132 @@ func TestChunkPutter(t *testing.T) {
...
+			TagID:     tagID,
+			Split:     10,
+			Seen:      0,
+			Stored:    10,
+			Sent:      10,
+			Synced:    10,
+			StartedAt: now().Unix(),
+			Address:   addr,
+		}
+		if diff := cmp.Diff(ti, exp); diff != "" {
```
mrekucci, line 589: The first argument should be `exp` (a.k.a `want`) and the second should be `ti` (a.k.a `have`).
Also, check this in other places where you use `cmp.Diff`.

mrekucci (commented):
Nice!

pkg/storagev2/chunkstore.go
```diff
@@ -17,21 +17,42 @@ type Getter interface {
...
+type GetterWithStamp interface {
+	GetWithStamp(context.Context, swarm.Address, []byte) (swarm.Chunk, error)
```
istae, line 31: the name should technically be `GetWithBatch` as the ID is batchID, not stamp

istae (approved):
very cool!

pkg/localstorev2/internal/cache/cache_test.go
aloknerurkar: In the next PR I will make a common function for this for all the tests. Will make this change there.

pkg/storagev2/chunkstore.go
aloknerurkar: So the idea behind the name is that it returns swarm.Chunk with the Stamp accessor filled using the batch ID. Hence the name WithStamp. Also, I think batch is a more generic term so it would be better to call it stamp as the overall concept. Wdyt?

pkg/storagev2/chunkstore.go
```diff
@@ -17,21 +17,42 @@ type Getter interface {
...
-// Deleter is the interface that wraps the basic Delete method.
+// Deleter is the interface that wraps the basic Delete method. This delete will
+// remove all the stamps associated with the chunk. To delete based on stamps,
+// use the DeleterWithStamp interface.
 type Deleter interface {
 	// Delete a chunk by the given swarm.Address.
 	Delete(context.Context, swarm.Address) error
 }
 
+// DeleterWithStamp provides a way to Delete a chunk by specifying the stamp
```
aloknerurkar: @istae @mrekucci 

So I thought about this, but it will be difficult to do the DeleteWithStamp function. The problem is we have a ref counted chunkstore. This means a chunk can be put with the same stamp by more than 1 component. So deleting the stamp when requested by 1 component is not a good idea.

Instead, the ref counting will handle the deletion properly. RefCnt is added everytime a Put is done. So a delete will only delete the Chunk if refCnt goes to 0. This means all the stamps will be kept in chunkstore till the chunk reaches refCnt 0.

We only need the GetterWithStamp interface. Only the upload store and reserve store will need to use this when we are doing syncing. So these components need to track the BatchIDs they are interested in and then use GetterWithStamp to get the right stamp before interacting with the syncing protocols. So I am removing this interface for now. If we need this functionality in future we can think about doing this.

mrekucci (approved):

PR #3693, opened 2023-01-12
refactor: pin store changes

aloknerurkar:
### Checklist

### Description
Minor changes in the pin store putter implementation. The problem was that the Putter creation initially required a root reference. However, this is available at the end in some cases where we haven't finished splitting the file yet. So now the upload putter will be closed with the reference and the root collection item will only be created at the end.

Also, created a function in the internal package for tests to reduce common code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3693)

### DIFF
--- a/pkg/localstorev2/internal/internal.go
+++ b/pkg/localstorev2/internal/internal.go
@@ -62,7 +62,7 @@ func NewInmemStorage() (Storage, func() error) {
 
 	ctx, cancel := context.WithCancel(context.Background())
 
-	ts := &testStorage{
+	ts := &inmemRepository{
 		ctx:        ctx,
 		indexStore: inmemstore.New(),
 		chunkStore: &chunkStore{inmemchunkstore.New()},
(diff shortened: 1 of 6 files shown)


pkg/localstorev2/internal/internal.go
```diff
@@ -33,16 +36,59 @@ type PutterCloserWithReference interface {
...
+		chunkStore: &chunkStore{inmemchunkstore.New()},
+	}
+
+	return ts, func() error {
+		cancel()
+		return multierror.Append(ts.indexStore.Close(), ts.chunkStore.Close()).ErrorOrNil()
+	}
+}
+
+type testStorage struct {
```
mrekucci, line 77: Maybe `inmemRepository` would be a better name here.

pkg/localstorev2/internal/pinning/pinning.go
```diff
@@ -201,11 +178,12 @@ func (c *collectionPutter) Put(ctx context.Context, ch swarm.Chunk) error {
 	return nil
 }
 
-func (c *collectionPutter) Close() error {
+func (c *collectionPutter) Close(root swarm.Address) error {
```
mrekucci, line 181: It should be an error to call `Close` with a zero address.

mrekucci (approved):

PR #3698, opened 2023-01-12
feat: private key handling interface

notanatol:
### Checklist

### Description
Instead of passing the r1/k1 private key cryptography functions as arguments this PR encapsulates them and puts behind an interface.


pkg/crypto/edg.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2023 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package crypto
```
vladopajic, line 5: struct names `Secp256k1EDG` and `Secp256r1EDG` look identical...

are names `EDGSecp256K` and `EDGSecp256R` bit better? how about `EDGSecp256_K` and `EDGSecp256_R`?

pkg/crypto/edg.go
```diff
@@ -0,0 +1,33 @@
...
+
+import "crypto/ecdsa"
+
```
vladopajic, line 8: if we wanted to have explicit check if these structures implements `keystore.EDG` interface here via construct:

```
var _ keystore.EDG = (*Secp256k1EDG)(nil)
var _ keystore.EDG = (*Secp256r1EDG)(nil)
```
it would not be possible because this would lead to circular imports, right? 

If this is the case would it make more sense to have these structs specified in `keystore` package? Maybe have `edg.go` file which will hold interface together with these structures?

Or maybe `EDG` interface is more suitable in crypto package?

---

Ether way we decide it would be nice to have this explicit check because it further engraves design.

pkg/crypto/edg.go
```diff
@@ -0,0 +1,33 @@
...
+// Secp256k1EDG aggregates private key cryptography functions that employ secp256k1
+type Secp256k1EDG struct{}
```
janos, line 10: Since both Secp256k1EDG and Secp256r1EDG are static, they can be constructed as singletons for example similarly as base64 is constructing different encoding variants 
https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/encoding/base64/base64.go;l=99. This would remove the need to construct them every time when they are used. They do not occupy any memory (as they are struct{} type), but I think it would be more semantically correct.

Also, repeating tests TestEncodeSecp256r1PrivateKey TestGenerateSecp256r1Key and similars for Secp256k1EDG and Secp256r1EDG would ensure the correctness in time, however that seems trivial.

janos (approved):
Thanks @notanatol.

pkg/crypto/edg.go
notanatol: fixed

pkg/crypto/edg.go
notanatol: added types

vladopajic: I am not sure if `crypto` package should have dependency to `keystore` package. `crypto` package should be at higher level of abstraction then `keystore` therefore is shouldn't depend on it.

pkg/crypto/edg.go
```diff
@@ -0,0 +1,41 @@
...
+package crypto
+
+import (
+	"crypto/ecdsa"
+
+	"github.com/ethersphere/bee/pkg/keystore"
+)
+
+// EDGSecp256_k1 aggregates private key cryptography functions that employ secp256k1
+type EDGSecp256_k1 struct{}
```
vladopajic, line 14: How about if we have types unexported `edgSecp256_K1`, but singleton values exported `EDGSecp256_K1`?

vladopajic (approved):
LGTM 

Added minor comments, overall good to go.

PR #3702, opened 2023-01-14
feat: tx chunkstore and multiple stamp support

aloknerurkar:
### Checklist

### Description
This PR adds the TxChunkStore implementation for the localstorev2.ChunkStore. The chunkstore uses the storage.Store as well as Sharky. So implementing a transaction functionality is a little tricky.

The way it is done here is to have a simple transaction functionality over sharky, where released locations are only released on commit, so that they can be rolled back and committed locations are released on rollback to give back the slots to sharky.

Along with this, we use the TxStore internally in the transactions to revert the state of the Store items to their original state.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3702)

### DIFF
--- a/pkg/localstorev2/internal/chunkstore/chunkstore.go
+++ b/pkg/localstorev2/internal/chunkstore/chunkstore.go
@@ -181,6 +181,9 @@ func (c *chunkStampItem) Clone() storage.Item {
 	return clone
 }
 
+// Sharky provides an abstraction for the sharky.Store operations used in the
+// chunkstore. This allows us to be more flexible in passing in the sharky instance
+// to chunkstore. For eg, check the TxChunkStore implementation in this pkg.
 type Sharky interface {
 	Read(context.Context, sharky.Location, []byte) error
 	Write(context.Context, []byte) (sharky.Location, error)
--- a/pkg/localstorev2/internal/chunkstore/chunkstore.go
+++ b/pkg/localstorev2/internal/chunkstore/chunkstore.go
@@ -292,10 +292,6 @@ func (c *chunkStoreWrapper) GetWithStamp(
 	return c.get(ctx, addr, batchID)
 }
 
-func (c *chunkStoreWrapper) GetWithStamp(ctx context.Context, addr swarm.Address, batchID []byte) (swarm.Chunk, error) {
-	return nil, fmt.Errorf("not implemented")
-}
-
 func (c *chunkStoreWrapper) Has(_ context.Context, addr swarm.Address) (bool, error) {
 	return c.store.Has(&retrievalIndexItem{Address: addr})
 }
--- a/pkg/localstorev2/internal/chunkstore/chunkstore.go
+++ b/pkg/localstorev2/internal/chunkstore/chunkstore.go
@@ -5,6 +5,7 @@
 package chunkstore
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
 	"errors"
@@ -180,9 +181,15 @@ func (c *chunkStampItem) Clone() storage.Item {
 	return clone
 }
 
+type Sharky interface {
+	Read(context.Context, sharky.Location, []byte) error
+	Write(context.Context, []byte) (sharky.Location, error)
+	Release(context.Context, sharky.Location) error
+}
+
 type chunkStoreWrapper struct {
 	store  storage.Store
-	sharky *sharky.Store
+	sharky Sharky
 }
 
 // New returns a chunkStoreWrapper which implements the storage.ChunkStore interface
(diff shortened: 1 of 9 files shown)


pkg/localstorev2/internal/chunkstore/chunkstore.go
```diff
@@ -180,9 +181,15 @@ func (c *chunkStampItem) Clone() storage.Item {
 	return clone
 }
 
+type Sharky interface {
```
mrekucci, line 184: Some doc comments would be useful here.

pkg/localstorev2/internal/chunkstore/txchunkstore_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	chunkStore := chunkstore.NewTxChunkStore(inmemstore.NewTxStore(inmemstore.New()), sharky)
+
+	chunk := chunktest.GenerateTestRandomChunk()
+	stamps := []swarm.Stamp{chunk.Stamp()}
+	for i := 0; i < 2; i++ {
+		stamps = append(stamps, postagetesting.MustNewStamp())
+	}
+
+	verify := func(t *testing.T) {
```
vladopajic, line 51: nit-picks:
- could `verify` function have more descriptive name, e.g. `assertStoreContainsStamps`?
- could this function be defined outside `TestMultipleStampsRefCnt` so it can be potentially reused?

pkg/localstorev2/internal/chunkstore/txchunkstore_test.go
```diff
@@ -0,0 +1,140 @@
...
+		t.Helper()
+
+		rIdx := chunkstore.RetrievalIndexItem{
+			Address: chunk.Address(),
+		}
+
+		has, err := chunkStore.(*chunkstore.TxChunkStoreWrapper).Store().Has(&rIdx)
```
vladopajic, line 58: would it be possible to avoid type casting? also in similar case few lines bellow

vladopajic (approved):
LGTM; minor comments

pkg/localstorev2/internal/chunkstore/txchunkstore.go
```diff
@@ -0,0 +1,128 @@
...
+func (t *txChunkStoreWrapper) Put(ctx context.Context, chunk swarm.Chunk) error {
+	return t.TxChunkStoreBase.Put(ctx, chunk)
+}
+
+func (t *txChunkStoreWrapper) Delete(ctx context.Context, address swarm.Address) error {
+	return t.TxChunkStoreBase.Delete(ctx, address)
+}
+
+func (t *txChunkStoreWrapper) Commit() error {
+	t.doneMu.Lock()
```
mrekucci, line 65: Not quite sure why this lock is needed here. The same goes for the Rollback method.

pkg/localstorev2/internal/chunkstore/txchunkstore_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+		err := cs.Commit()
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		verify(t)
+	})
+
+	t.Run("rollback delete operations", func(t *testing.T) {
```
mrekucci, line 102: There should be tests for the rollback of the put operation when a new value is created and when an existing value is updated (probably as part fo the `storagetest.TestTxChunkStore` test suite).

pkg/storagev2/inmemstore/transaction.go
```diff
@@ -9,7 +9,7 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/ethersphere/bee/pkg/storagev2"
+	storage "github.com/ethersphere/bee/pkg/storagev2"
```
mrekucci, line 12: This change is unnecessary.

pkg/storagev2/storagetest/transaction.go
```diff
@@ -336,7 +336,7 @@ func checkTxChunkStoreFinishedTxInvariants(t *testing.T, store storage.TxChunkSt
 
 	ctx := context.Background()
 	want := storage.ErrTxDone
-	o007 := swarm.NewChunk(swarm.NewAddress([]byte("007")), []byte("Hello, World!"))
+	o007 := chunktest.GenerateTestRandomChunk()
```
mrekucci, line 339: I'd suggest not using these `GenerateTestRandomChunk' calls here, as it makes it harder to debug the values when tests fail. The same goes for the whole file.

pkg/storagev2/transaction.go
```diff
@@ -109,6 +109,10 @@ func (tx *TxState) Done() {
 	})
 }
 
+func NewChildTxState(tx *TxState) *TxState {
```
mrekucci, line 112: Some description would be nice. Also, I'd recommend moving it below the `NewTxState` constructor.

pkg/localstorev2/internal/chunkstore/txchunkstore.go
aloknerurkar: So the problem I see is that the IsDone check and Done are not done atomically. This is why another thread can call Commit around the same instant when the first one has already started executing the ops. I think this could be a problem in the current inmem txchunkstore as well. Wdyt?

pkg/localstorev2/internal/chunkstore/txchunkstore_test.go
aloknerurkar: So in the chunkstore updates would generally not happen as these are content-addressed values. So if the key is the same the data will also be the same. The only thing updated could be the stamps.

The reason I have these tests here as they test the refCnt functionality which is specific to this chunkstore. But I guess we could add 1 more test in txchunkstore to Put with different stamps and then rollback keeps the right stamps.

pkg/storagev2/inmemstore/transaction.go
aloknerurkar: So my vet plugin adds this on every file save. Not really sure how to turn it off, but I guess once we change the package name at the end, it should remove it.

mrekucci (approved):

PR #3703, opened 2023-01-14
fix: puller and pushync should use storage radius only for syncing

### LINKED ISSUE #3700: puller should only sync with peers of PO >= storage radius
@ldeffenb writes

> Ok, let's see if I can explain my issue with the current puller implementation.   It uses the kademlia depth to select peers to sync with.  So, if my node's overlay starts with 0x000 and the kademlia depth is 4, it will sync with peers from 0x0000 through 0x0fffff (leading 4 bits match).   Now, if the storage depth is 8, then it will sync bins 8 and higher from those nodes.  But my node really only wants chunks that start with 0x00... which is 8 bits matching my overlay.   But if I ask for bin 8 from one of my peers that starts with 0x0f, it will send me only chunks starting with 0x0f.. because that's what it has in its bin 8.   But I don't want those chunks in my reserve, so they go straight to the cache and get gc'd later.
And I do that with ALL of my peers that happen to start with 0x01 through 0x0f which I'm syncing because of my kademlia depth of 4.   And I ONLY ask those peers for bins greater than or equal to my storage radius, which is 8, which guarantees that 7/8 of the chunks that the puller pulls go into the cache and get gc'd.   And I won't get ANY 0x00 chunks from the 0x01 thorugh 0x0f peers because they have those chunks in lower numbered bins than my storage radius.
Did I explain that well enough?   And does it make sense?   It certainly fits my observations when the kademlia depth is less than the storage radius.   Chunks get pulled, cached, and gc'd.   And if you're doing enough of them, the gc happens before the cursors get recycled so it starts all over again from the very same peers.  And this keeps happening until the kademlia depth gets back to where it needs to be.
Add some trace statements to the puller if you don't believe this is what is happening.  A while back, I actually had a hacked puller that a) didn't send chunks that weren't within the bin relative to the pull requester, not myself, and b) didn't store locally pulled chunks that were outside of my radius.  This kept the disk churn to a minimum on my end.   But then I realized that (a) was actually benefiting the other nodes which would be good if it is generally deployed, but certainly didn't help on my end.
And the puller really doesn't want to pull high numbered bins from nodes if those chunks are guaranteed to be outside the pulling node's storage radius.  But in fact, it needs to pull from peer bins that are LESS than the pullers storage radius, because that's where the reserve-interested chunks reside over in the peer.
Bottom line:  My bin 8 isn't the same as my peer's bin 8 unless we are in the same depth 8 neighborhood (matching the first 8 bits of the overlay).

We should only be syncing with peers with PO >= storage radius. Any lower, the chunks from these peers do not land in our reserve.

istae:
### Checklist

### Description
Closes #3700 

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3703)

### DIFF
--- a/pkg/node/devnode.go
+++ b/pkg/node/devnode.go
@@ -6,6 +6,7 @@ package node
 
 import (
 	"context"
+	"crypto/rand"
 	"errors"
 	"fmt"
 	"io"
@@ -48,7 +49,6 @@ import (
 	"github.com/ethersphere/bee/pkg/storageincentives/staking"
 	stakingContractMock "github.com/ethersphere/bee/pkg/storageincentives/staking/mock"
 	"github.com/ethersphere/bee/pkg/swarm"
-	"github.com/ethersphere/bee/pkg/swarm/test"
 	"github.com/ethersphere/bee/pkg/tags"
 	"github.com/ethersphere/bee/pkg/topology/lightnode"
 	mockTopology "github.com/ethersphere/bee/pkg/topology/mock"
@@ -116,7 +116,10 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 	}
 	b.stateStoreCloser = stateStore
 
-	swarmAddress := test.RandomAddress()
+	swarmAddress, err := randomAddress()
+	if err != nil {
+		return nil, err
+	}
 
 	batchStore, err := batchstore.New(stateStore, func(b []byte) error { return nil }, swarmAddress, logger)
 	if err != nil {
@@ -512,4 +515,17 @@ func (b *DevBee) Shutdown() error {
 
 func pong(ctx context.Context, address swarm.Address, msgs ...string) (rtt time.Duration, err error) {
 	return time.Millisecond, nil
+}
+
+func randomAddress() (swarm.Address, error) {
+
+	b := make([]byte, 32)
+
+	_, err := rand.Read(b)
+	if err != nil {
+		return swarm.ZeroAddress, err
+	}
+
+	return swarm.NewAddress(b), nil
+
 }
--- a/pkg/node/devnode.go
+++ b/pkg/node/devnode.go
@@ -48,6 +48,7 @@ import (
 	"github.com/ethersphere/bee/pkg/storageincentives/staking"
 	stakingContractMock "github.com/ethersphere/bee/pkg/storageincentives/staking/mock"
 	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/ethersphere/bee/pkg/swarm/test"
 	"github.com/ethersphere/bee/pkg/tags"
 	"github.com/ethersphere/bee/pkg/topology/lightnode"
 	mockTopology "github.com/ethersphere/bee/pkg/topology/mock"
@@ -115,7 +116,9 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 	}
 	b.stateStoreCloser = stateStore
 
-	batchStore, err := batchstore.New(stateStore, func(b []byte) error { return nil }, logger)
+	swarmAddress := test.RandomAddress()
+
+	batchStore, err := batchstore.New(stateStore, func(b []byte) error { return nil }, swarmAddress, logger)
 	if err != nil {
 		return nil, fmt.Errorf("batchstore: %w", err)
 	}
@@ -235,7 +238,6 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 		},
 	}
 
-	var swarmAddress swarm.Address
 	storer, err := localstore.New("", swarmAddress.Bytes(), stateStore, lo, logger)
 	if err != nil {
 		return nil, fmt.Errorf("localstore: %w", err)
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -254,7 +254,7 @@ func NewBee(ctx context.Context, addr string, publicKey *ecdsa.PublicKey, signer
 
 	addressbook := addressbook.New(stateStore)
 
-	pubKey, _ := signer.PublicKey()
+	pubKey, err := signer.PublicKey()
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -894,12 +894,12 @@ func NewBee(ctx context.Context, addr string, publicKey *ecdsa.PublicKey, signer
 
 	pinningService := pinning.NewService(storer, stateStore, traversalService)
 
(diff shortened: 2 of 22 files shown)


pkg/node/devnode.go
```diff
@@ -115,7 +116,9 @@ func NewDevBee(logger log.Logger, o *DevOptions) (b *DevBee, err error) {
 	}
 	b.stateStoreCloser = stateStore
 
-	batchStore, err := batchstore.New(stateStore, func(b []byte) error { return nil }, logger)
+	swarmAddress := test.RandomAddress()
```
mrekucci, line 119: I would suggest not to include the `test` package here and instead create an address by calling one of the `swarm` create address functions.

pkg/pushsync/pushsync.go
```diff
@@ -78,6 +78,7 @@ type PushSync struct {
 	streamer       p2p.StreamerDisconnecter
 	storer         storage.Putter
 	topologyDriver topology.Driver
+	rs             postage.IsWithinRadius
```
mrekucci, line 81: `rs` -> `radiusChecker`, the is for the constructor param.

pkg/postage/interface.go
```diff
@@ -46,10 +47,15 @@ type ReserveStateGetter interface {
 	GetReserveState() *ReserveState
 }
 
+type IsWithinRadius interface {
```
mrekucci, line 50: Maybe `IsWithinRadius` -> `RadiusChecker`.

pkg/postage/batchstore/mock/store.go
```diff
@@ -13,21 +13,23 @@ import (
...
-	id                []byte
-	batch             *postage.Batch
-	getErr            error
-	getErrDelayCnt    int
-	updateErr         error
-	updateErrDelayCnt int
-	resetCallCount    int
+	rs                    *postage.ReserveState
+	cs                    *postage.ChainState
+	isWithinStorageRadius bool
```
mrekucci, line 25: Instead of a specific value, I would recommend passing the radius checker interface. This will provide more flexibility.

pkg/pusher/pusher.go
```diff
@@ -44,7 +44,7 @@ type Service struct {
 	storer            storage.Storer
 	pushSyncer        pushsync.PushSyncer
 	validStamp        postage.ValidStampFn
-	depther           topology.NeighborhoodDepther
+	rs                postage.ReserveStateGetter
```
mrekucci, line 47: Maybe `rs` -> `reserveGetter`. The same is for the constructor param.

pkg/node/node.go
```diff
@@ -254,6 +254,41 @@ func NewBee(ctx context.Context, addr string, publicKey *ecdsa.PublicKey, signer
 
 	addressbook := addressbook.New(stateStore)
 
+	pubKey, _ := signer.PublicKey()
+	if err != nil {
```
notanatol, line 258: I see that it was like this from before but maybe we can fix this now by either adding the `err` on line 257 or getting rid of the check altogether

notanatol (approved):

mrekucci (approved):

pkg/postage/batchstore/store.go
```diff
@@ -97,6 +100,15 @@ func (s *store) GetReserveState() *postage.ReserveState {
 	}
 }
 
+func (s *store) IsWithinStorageRadius(addr swarm.Address) bool {
+
+	s.mtx.RLock()
+	defer s.mtx.RUnlock()
+
+	po := swarm.Proximity(addr.Bytes(), s.base.Bytes())
+	return po >= s.GetReserveState().StorageRadius
```
aloknerurkar, line 109: Maybe its better to create a new interface `StorageRadiusGetter` and implement it in the batchstore. Its being used at a lot of places and maybe having a separate interface would be more flexible for eg in tests etc.

Also it would be better not to expose unnecessary reserve info to other packages.

pkg/puller/puller.go
```diff
@@ -134,7 +133,7 @@ func (p *Puller) manage(ctx context.Context, warmupTime time.Duration) {
 		prevRadius = syncRadius
 
 		_ = p.topology.EachPeerRev(func(addr swarm.Address, po uint8) (stop, jumpToNext bool, err error) {
-			if po >= neighborhoodDepth {
+			if po >= syncRadius {
```
aloknerurkar, line 136: So there are 2 cases in the issue mentioned. Does this handle both the cases?

pkg/postage/batchstore/store.go
```diff
@@ -97,6 +100,15 @@ func (s *store) GetReserveState() *postage.ReserveState {
 	}
 }
 
+func (s *store) IsWithinStorageRadius(addr swarm.Address) bool {
```
aloknerurkar, line 103: There is no test which covers this function.

pkg/topology/topology.go
```diff
@@ -59,7 +59,7 @@ type EachNeighbor interface {
 	EachNeighbor(EachPeerFunc) error
 	// EachNeighborRev iterates from farthest bin to closest within the neighborhood.
 	EachNeighborRev(EachPeerFunc) error
-	// IsWithinDepth checks if an address is the within neighborhood.
+	// IsWithinDepth checks if an address is within the neighborhood.
 	IsWithinDepth(swarm.Address) bool
```
aloknerurkar, line 63: Do we use this anymore?

aloknerurkar (changes requested):

pkg/puller/puller.go
istae: the other case i wanted to tackle in a separate PR, I can do it in this one, but the PR will be bigger.

istae:
unit tests are broken due the puller changes :(

pkg/topology/depthmonitor/depthmonitor.go
```diff
@@ -125,9 +125,9 @@ func (s *Service) manage(warmupTime, wakeupInterval time.Duration) {
 		case <-time.After(wakeupInterval):
 		}
 
-		reserveState := s.bs.GetReserveState()
+		radius := s.bs.GetReserveState().StorageRadius
```
aloknerurkar, line 128: Lets use the new function here.

pkg/puller/puller_test.go
```diff
@@ -405,6 +403,7 @@ func TestBinReset(t *testing.T) {
 // the tested unit.
 func TestDepthChange(t *testing.T) {
 	t.Parallel()
+	t.Skip()
```
aloknerurkar, line 406: Is this still required?

pkg/puller/puller_test.go
```diff
@@ -373,7 +371,7 @@ func TestBinReset(t *testing.T) {
 	waitCursorsCalled(t, pullsync, addr, false)
 	waitSyncCalled(t, pullsync, addr, false)
 
-	rs.setRadius(0)
+	// rs.setRadius(0)
```
aloknerurkar, line 374: Dont forget to remove this before check-in if not required.

aloknerurkar (approved):
Minor nit picks. Apart from that LGTM!

Issue #3695, opened 2023-01-12
Support multiple stamps per chunk in storage v2


aloknerurkar:
Chunk can be Put with different stamps so:

- Keep track of stamps for chunk in inmemchunkstore
- Add new GetterWithStamp interface to inmemchunkstore
- Add new testsuite

PR #3707, opened 2023-01-16
chore: multiple stamps per chunk

vladopajic:
Closes: #3695

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3707)

### DIFF
--- a/pkg/localstorev2/internal/upload/uploadstore_test.go
+++ b/pkg/localstorev2/internal/upload/uploadstore_test.go
@@ -35,19 +35,6 @@ func TestMain(m *testing.M) {
 	os.Exit(code)
 }
 
-var _ internal.Storage = (*testStorage)(nil)
-
-// testStorage is an implementation of internal.Storage for test purposes.
-type testStorage struct {
-	ctx        context.Context
-	indexStore storage.Store
-	chunkStore internal.ChunkStore
-}
-
-func (t *testStorage) Ctx() context.Context            { return t.ctx }
-func (t *testStorage) Store() storage.Store            { return t.indexStore }
-func (t *testStorage) ChunkStore() internal.ChunkStore { return t.chunkStore }
-
 func TestPushItem_MarshalAndUnmarshal(t *testing.T) {
 	t.Parallel()
 
--- a/pkg/localstorev2/internal/upload/uploadstore_test.go
+++ b/pkg/localstorev2/internal/upload/uploadstore_test.go
@@ -41,15 +41,7 @@ var _ internal.Storage = (*testStorage)(nil)
 type testStorage struct {
 	ctx        context.Context
 	indexStore storage.Store
-	chunkStore *chunkStore
-}
-
-type chunkStore struct {
-	storage.ChunkStore
-}
-
-func (c *chunkStore) GetWithStamp(ctx context.Context, address swarm.Address, _ []byte) (swarm.Chunk, error) {
-	return c.Get(ctx, address)
+	chunkStore internal.ChunkStore
 }
 
 func (t *testStorage) Ctx() context.Context            { return t.ctx }
--- a/pkg/storagev2/inmemchunkstore/inmemchunkstore.go
+++ b/pkg/storagev2/inmemchunkstore/inmemchunkstore.go
@@ -14,13 +14,18 @@ import (
 )
 
 type ChunkStore struct {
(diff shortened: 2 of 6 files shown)


pkg/storagev2/inmemchunkstore/inmemchunkstore.go
```diff
@@ -13,49 +14,110 @@ import (
 )
 
 type ChunkStore struct {
-	chunks sync.Map
+	chunks map[string][]swarm.Chunk
```
aloknerurkar, line 17: So we dont need to store the Chunk every time. We only need the Stamp field as this is the only thing that changes. This would save a lot of memory as well.

We should have a separate map for stamps. We can use the same map for the chunks and then also add an entry to this additional map for stamp. The key would be the chunk address itself. Check the implementation in localstorev2/internal/chunkstore pkg for some inspiration.

aloknerurkar (changes requested):

pkg/storagev2/inmemchunkstore/inmemchunkstore.go
```diff
@@ -13,49 +14,119 @@ import (
...
-	c.chunks.Range(func(_, val interface{}) bool {
-		stop, err := fn(val.(swarm.Chunk))
-		if err != nil {
-			retErr = err
-			return false
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	for _, data := range c.chunks {
+		for _, stamp := range data.stamps {
```
aloknerurkar, line 102: So this is not required. We should not treat chunks with multiple stamps as separate chunks for iteration. If a particular stamp is desired, the user has to use the GetWithStamp function. The Iterate function will be provided through some component using the GetWithStamp, so stamp-related information is not required here. 

Depending on how things look, we might not even have Iterate as a core API on ChunkStore, but a separate interface implemented by some component that uses the ChunkStore. For this PR, we should simply return the first stamp we have for the chunk.

pkg/storagev2/inmemchunkstore/inmemchunkstore.go
```diff
@@ -13,49 +14,119 @@ import (
...
+
+	data, ok := c.chunks[addr.ByteString()]
+	if !ok {
 		return nil, storage.ErrNotFound
 	}
-	return val.(swarm.Chunk), nil
+
+	// when batchID is not specified, chunk with first stamp is returned
+	if batchID == nil {
+		return makeChunk(data.chunk, data.stamps[0]), nil
```
aloknerurkar, line 47: Before dereferencing the stamps array we should check if there is any stamp present.

I think it is better to have 2 errors in storage package:
ErrNoStampsForChunk => if no stamp is present
ErrStampNotFound => returned by GetWithStamp if stamp is not found

As both these cases are different. So more information is better.

aloknerurkar (commented):
Also, I think you need to rebase on latest localstorev2. I think the uploadstore_test file is outdated.

vladopajic:
@aloknerurkar fyi, this PR is currently using up to date version of [localstorev2.0](https://github.com/ethersphere/bee/tree/localstorev2.0)

pkg/localstorev2/internal/upload/uploadstore_test.go
```diff
@@ -41,15 +41,7 @@ var _ internal.Storage = (*testStorage)(nil)
 type testStorage struct {
 	ctx        context.Context
 	indexStore storage.Store
-	chunkStore *chunkStore
```
aloknerurkar, line 44: Actually, I got confused because this testStorage is no longer used I think. All the tests now use the common function in the `internal.go` file. So this can be removed completely.

aloknerurkar (commented):
In order to pass CI you can add a function in the chunkstore package which returns a NotImplemented error to satisfy the new interface. My PR will add this later.

pkg/localstorev2/internal/upload/uploadstore_test.go
vladopajic: I have added GetWithStamp method to localstore/ChunkStore. I am not sure what do about other ting you are suggesting here?

aloknerurkar: So you can remove the testStorage object completely I think. Its not being used. So remove L41-L49

aloknerurkar (approved):

PR #3710, opened 2023-01-19
feat: add retry tx logic

swarmHaseeb:
### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3710)

### DIFF
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -344,6 +344,9 @@ func buildBeeNode(ctx context.Context, c *command, cmd *cobra.Command, logger lo
 		AdminPasswordHash:             c.config.GetString(optionNameAdminPasswordHash),
 		UsePostageSnapshot:            c.config.GetBool(optionNameUsePostageSnapshot),
 		EnableStorageIncentives:       c.config.GetBool(optionNameStorageIncentivesEnable),
+		BoostGasTip:                   c.config.GetIntSlice(optionNameBoostGasTip),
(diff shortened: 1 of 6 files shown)


pkg/transaction/transaction.go
```diff
@@ -107,10 +107,15 @@ type transactionService struct {
 	store   storage.StateStorer
 	chainID *big.Int
 	monitor Monitor
+
+	boostTipRange   []int
+	boostPriceRange []int
+	defaultBoost    bool
```
vladopajic, line 113: Would it make sense to have new type (say `BoostConfig`) which would hold these tree variables? 

It would make code little bit more cleaner because:
- passing one value would be easier then passing three values
- it would group similar configs under singe type

cmd/bee/cmd/start.go
```diff
@@ -344,6 +344,9 @@ func buildBeeNode(ctx context.Context, c *command, cmd *cobra.Command, logger lo
 		AdminPasswordHash:             c.config.GetString(optionNameAdminPasswordHash),
 		UsePostageSnapshot:            c.config.GetBool(optionNameUsePostageSnapshot),
 		EnableStorageIncentives:       c.config.GetBool(optionNameStorageIncentivesEnable),
+		BoostGasTip:                   c.config.GetIntSlice(optionNameBoostGasTip),
+		BoostGasPrice:                 c.config.GetIntSlice(optionNameBoostGasPrice),
```
vladopajic, line 348: what would happen if user specifies slice with only one value? should this case be validated before node is started or?

vladopajic (commented):
added some early comments

PR #3715, opened 2023-01-20
fix: continue migration on error

aloknerurkar:
### Checklist

### Description
Ideally, we should not see this error. But if we do, we should ignore it and continue as a best-effort cleanup.

A proper fix would be to identify why we dont have retrievalDataIndex. But this would be a much bigger investigation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3715)

### DIFF
--- a/pkg/localstore/migration_residue.go
+++ b/pkg/localstore/migration_residue.go
@@ -132,7 +132,8 @@ func migrateResidue(db *DB) error {
 		sItem, err := retrievalDataIndex.Get(item)
 		switch {
 		case errors.Is(err, leveldb.ErrNotFound):
-			return true, nil
+			// continue iteration on error
+			return false, nil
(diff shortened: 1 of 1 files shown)


mrekucci (approved):

pkg/localstore/migration_residue.go
```diff
@@ -132,7 +132,8 @@ func migrateResidue(db *DB) error {
 		sItem, err := retrievalDataIndex.Get(item)
 		switch {
 		case errors.Is(err, leveldb.ErrNotFound):
-			return true, nil
+			// continue iteration on error
```
mrekucci, line 135: Maybe would be good to log this as a warning.

istae (approved):

notanatol (approved):

PR #3724, opened 2023-01-25
fix: pull synce storage radius

umerm-work:
### Checklist

### DIFF
--- a/pkg/pullsync/pullsync.go
+++ b/pkg/pullsync/pullsync.go
@@ -73,12 +73,13 @@ type Syncer struct {
 	wg         sync.WaitGroup
 	unwrap     func(swarm.Chunk)
 	validStamp postage.ValidStampFn
+	radius     postage.RadiusChecker
 
 	Interface
 	io.Closer
 }
 
-func New(streamer p2p.Streamer, storage pullstorage.Storer, unwrap func(swarm.Chunk), validStamp postage.ValidStampFn, logger log.Logger) *Syncer {
+func New(streamer p2p.Streamer, storage pullstorage.Storer, unwrap func(swarm.Chunk), validStamp postage.ValidStampFn, logger log.Logger, radius postage.RadiusChecker) *Syncer {
 
 	return &Syncer{
 		streamer:   streamer,
@@ -89,6 +90,7 @@ func New(streamer p2p.Streamer, storage pullstorage.Storer, unwrap func(swarm.Ch
 		logger:     logger.WithName(loggerName).Register(),
 		wg:         sync.WaitGroup{},
 		quit:       make(chan struct{}),
+		radius:     radius,
 	}
 }
 
@@ -110,9 +112,9 @@ func (s *Syncer) Protocol() p2p.ProtocolSpec {
 }
 
 // SyncInterval syncs a requested interval from the given peer.
-// It returns the BinID of highest chunk that was synced from the given interval.
+// It returns the BinID of the highest chunk that was synced from the given interval.
 // If the requested interval is too large, the downstream peer has the liberty to
-// provide less chunks than requested.
+// provide fewer chunks than requested.
 func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8, from, to uint64) (uint64, error) {
 	loggerV2 := s.logger.V(2).Register()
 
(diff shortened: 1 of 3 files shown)


pkg/pullsync/pullsync.go
```diff
@@ -179,7 +181,7 @@ func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8
 			s.logger.Debug("storage has", "error", err)
 			continue
 		}
-		if !have {
+		if !have && bin >= s.radius.StorageRadius() {
```
istae, line 184: so what we want is the proximity order of the address >= storage radius (the address at line 171)

istae (changes requested):

pkg/pullsync/pullsync.go
umerm-work: Done

pkg/pullsync/pullsync.go
```diff
@@ -179,7 +181,8 @@ func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8
 			s.logger.Debug("storage has", "error", err)
 			continue
 		}
-		if !have {
+		po := swarm.Proximity(a.Bytes(), peer.Bytes())
```
istae, line 184: as an optimization, let's move the po check before the storage.Has to save on disk usage

istae (approved):
great work!

pkg/pullsync/pullsync.go
```diff
@@ -174,16 +176,20 @@ func (s *Syncer) SyncInterval(ctx context.Context, peer swarm.Address, bin uint8
...
-		if err != nil {
-			s.logger.Debug("storage has", "error", err)
-			continue
-		}
-		if !have {
-			wantChunks[a.ByteString()] = struct{}{}
-			ctr++
-			s.metrics.Wanted.Inc()
-			bv.Set(i / swarm.HashSize)
+		po := swarm.Proximity(a.Bytes(), peer.Bytes())
```
istae, line 179: the proximity calculation should be based on the chunk address and OUR own address, not peer, so you need to pass our overlay address to the pullsync :)

istae (changes requested):

pkg/pullsync/pullsync.go
istae: the goal of this change is that the chunks we are pulling in must belong in our reserve, then the reserve is chunks with po >= storage radius relative to our address

pkg/pullsync/pullsync.go
```diff
@@ -65,30 +66,34 @@ type Interface interface {
...
+	validStamp     postage.ValidStampFn
+	radius         postage.RadiusChecker
+	overlayAddress common.Address
 
 	Interface
 	io.Closer
 }
 
-func New(streamer p2p.Streamer, storage pullstorage.Storer, unwrap func(swarm.Chunk), validStamp postage.ValidStampFn, logger log.Logger) *Syncer {
+func New(streamer p2p.Streamer, storage pullstorage.Storer, unwrap func(swarm.Chunk), validStamp postage.ValidStampFn, logger log.Logger, radius postage.RadiusChecker, overlayAddress common.Address) *Syncer {
```
istae, line 84: should be swarm.Address

istae (changes requested):

pkg/pullsync/pullsync.go
umerm-work: https://github.com/ethersphere/bee/blob/e0171d2588c8a1d4c824b4f496874c966bbdc2d7/pkg/node/node.go#L294

All of the initialization functions are using common.Address. There is little difference. One way is to declare it again in `node.go` but that would be redundant.

notanatol (approved):

pkg/pullsync/pullsync_test.go
```diff
@@ -231,5 +232,5 @@ func newPullSync(s p2p.Streamer, o ...mock.Option) (*pullsync.Syncer, *mock.Pull
 	logger := log.Noop
 	unwrap := func(swarm.Chunk) {}
 	validStamp := func(ch swarm.Chunk, _ []byte) (swarm.Chunk, error) { return ch, nil }
-	return pullsync.New(s, storage, unwrap, validStamp, logger), storage
+	return pullsync.New(s, storage, unwrap, validStamp, logger, mockbatchstore.New(), swarm.MustParseHexAddress("ca1e9f3938cc1425c6061b96ad9eb93e134dfe8734ad490164ef20af9d1cf59c")), storage
```
mrekucci, line 235: It's getting longer, maybe reformat it like this:
```Golang
return pullsync.New(
	s,
	storage,
	unwrap,
	validStamp,
	logger,
	mockbatchstore.New(),
	swarm.MustParseHexAddress("ca1e9f3938cc1425c6061b96ad9eb93e134dfe8734ad490164ef20af9d1cf59c"),
), storage
```

mrekucci (approved):

PR #3725, opened 2023-01-25
feat: add iteration by prefix and skip first item

mrekucci:
### Checklist

### Description
Add iteration by prefix and skip first item option for index store query.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3725)

### DIFF
--- a/pkg/storagev2/storage.go
+++ b/pkg/storagev2/storage.go
@@ -32,13 +32,13 @@ type IterateFn func(Result) (bool, error)
 // directly in byte format or partially or fully unmarshal the data and check.
 type Filter func(string, []byte) bool
 
-// ItemAttribute tells the Query which Item
-// attribute should be loaded from the store.
-type ItemAttribute int
+// QueryItemProperty tells the Query which Item
+// property should be loaded from the store to the result.
+type QueryItemProperty int
 
 const (
 	// QueryItem indicates interest in the whole Item.
-	QueryItem ItemAttribute = iota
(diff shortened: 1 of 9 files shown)


pkg/storagev2/storage.go
```diff
@@ -71,8 +71,11 @@ type Query struct {
 	// to construct new object for the result.
 	Factory func() Item
 
-	// ItemAttribute indicates a specific interest of an Item attribute.
-	ItemAttribute ItemAttribute
+	// Prefix indicates interest in Item which hast this prefix.
+	Prefix string
```
aloknerurkar, line 75: It should be specified here that this Prefix is for the ID and not Namespace.

Also, like in shed, we could potentially need a boolean `includeFirst` or something. This is to tell the iteration if we should start from the Prefix match item or the next one. Mainly, I am thinking of the pagination issue. So if you process some entries and then call Iterate a second time, usually this prefix would be the key of the last item you looked at. So this boolean would tell us that we should start from the next one. As opposed to esad's case where we need to include the first one as well in some cases.

aloknerurkar (approved):
Apart from the minor comment, it looks good!

istae (approved):

pkg/storagev2/storage.go
mrekucci: Done.

istae (approved):

aloknerurkar (approved):

PR #3741, opened 2023-01-30
feat(localstorev2): localstore container implementation

aloknerurkar:
### Checklist

### Description
Initial skeleton for the localstore container class. This PR includes the high-level interfaces that localstore will expose for Upload and Pinning related functionality. Also tests to cover all the changes. These interfaces will eventually be used in the API and Pinning service implementations.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3741)

### DIFF
--- a/pkg/localstorev2/internal/pinning/pinning_test.go
+++ b/pkg/localstorev2/internal/pinning/pinning_test.go
@@ -17,6 +17,7 @@ import (
 	storage "github.com/ethersphere/bee/pkg/storagev2"
 	storagetest "github.com/ethersphere/bee/pkg/storagev2/storagetest"
 	"github.com/ethersphere/bee/pkg/swarm"
+	swarmtesting "github.com/ethersphere/bee/pkg/swarm/test"
 )
 
 type pinningCollection struct {
@@ -340,4 +341,12 @@ func TestPinCollectionItem(t *testing.T) {
 			})
 		})
 	}
+}
+
+func TestPinChunkItem(t *testing.T) {
+	t.Parallel()
+
+	storagetest.TestItemClone(t, &storagetest.ItemCloneTest{
+		Item: &pinstore.PinChunkItem{UUID: pinstore.NewUUID(), Addr: swarmtesting.RandomAddress()},
+	})
 }
--- a/pkg/localstorev2/internal/pinning/pinning_test.go
+++ b/pkg/localstorev2/internal/pinning/pinning_test.go
@@ -182,7 +182,7 @@ func TestPinStore(t *testing.T) {
 	})
 
 	t.Run("delete collection", func(t *testing.T) {
-		err := pinstore.DeletePin(st, tests[0].root.Address())
+		err := pinstore.DeletePin(context.TODO(), st, tests[0].root.Address())
 		if err != nil {
 			t.Fatal(err)
 		}
--- a/pkg/localstorev2/internal/pinning/pinning_test.go
+++ b/pkg/localstorev2/internal/pinning/pinning_test.go
@@ -347,6 +347,9 @@ func TestPinChunkItem(t *testing.T) {
 	t.Parallel()
 
 	storagetest.TestItemClone(t, &storagetest.ItemCloneTest{
-		Item: &pinstore.PinChunkItem{UUID: pinstore.NewUUID(), Addr: swarmtesting.RandomAddress()},
+		Item: &pinstore.PinChunkItem{
+			UUID: pinstore.NewUUID(),
+			Addr: swarmtesting.RandomAddress(),
+		},
 	})
 }
--- a/pkg/localstorev2/internal/pinning/pinning_test.go
+++ b/pkg/localstorev2/internal/pinning/pinning_test.go
@@ -122,7 +122,7 @@ func TestPinStore(t *testing.T) {
 	})
 
 	t.Run("verify root pins", func(t *testing.T) {
-		pins, err := pinstore.Pins(st.Store())
+		pins, err := pinstore.Pins(st.IndexStore())
 		if err != nil {
 			t.Fatal(err)
 		}
@@ -145,7 +145,7 @@ func TestPinStore(t *testing.T) {
 
 	t.Run("has pin", func(t *testing.T) {
 		for _, tc := range tests {
-			found, err := pinstore.HasPin(st.Store(), tc.root.Address())
+			found, err := pinstore.HasPin(st.IndexStore(), tc.root.Address())
 			if err != nil {
 				t.Fatal(err)
 			}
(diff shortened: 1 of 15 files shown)


notanatol (approved):

pkg/localstorev2/internal/upload/uploadstore.go
```diff
@@ -487,16 +492,77 @@ func (p *pushReporter) Report(
...
+
+func (nextTagID) ID() string { return "nextTagID" }
+
+func (n nextTagID) Marshal() ([]byte, error) {
+	buf := make([]byte, 8)
+	binary.LittleEndian.PutUint64(buf, uint64(n))
+	return buf, nil
+}
+
+func (n *nextTagID) Unmarshal(buf []byte) error {
```
janos, line 526: If renamed to UnmarshalBinary it will implement binary.BinaryUnmarshaler.

pkg/localstorev2/internal/upload/uploadstore.go
```diff
@@ -487,16 +492,77 @@ func (p *pushReporter) Report(
...
+)
+
+// nextTagID is a storage.Item which stores a uint64 value in the store.
+type nextTagID uint64
+
+func (nextTagID) Namespace() string { return "upload" }
+
+func (nextTagID) ID() string { return "nextTagID" }
+
+func (n nextTagID) Marshal() ([]byte, error) {
```
janos, line 520: If renamed to MarshalBinary it will implement binary.BinaryMarshaler.

pkg/localstorev2/localstore.go
```diff
@@ -1,25 +1,351 @@
...
+		DisableSeeksCompaction: opts.LdbDisableSeeksCompaction,
+	})
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed creating levelDB index store: %w", err)
+	}
+
+	sharkyBasePath := path.Join(basePath, "sharky")
+
+	if _, err := os.Stat(sharkyBasePath); os.IsNotExist(err) {
+		err := os.Mkdir(sharkyBasePath, 0775)
```
janos, line 169: The permission is before mask, which means that it would be impossible for user to set permission 0777, as it is limited to 0775 regardless of the mask. I suggest to make directories with 0777.

janos (approved):
Very nice.

pkg/localstorev2/localstore_test.go
```diff
@@ -0,0 +1,508 @@
...
+	t.Helper()
+
+	for _, ch := range chunks {
+		hasFound, err := repo.ChunkStore().Has(context.TODO(), ch.Address())
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if hasFound != has {
+			t.Fatalf("unexpected chunk state, exp has chunk %t got %t", has, hasFound)
```
mrekucci, line 38: The recommended format for test messages is: `"msg: want %v, have %v"`, so in this case: `"unexpected chunk state: want %t, have %t"`. The same applies below.

pkg/localstorev2/localstore_test.go
```diff
@@ -0,0 +1,508 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			if id != uint64(i) {
+				t.Fatalf("incorrect id generated, exp %d found %d", i, id)
+			}
+		}
+	})
+
+	t.Run("error on no tag", func(t *testing.T) {
```
mrekucci, line 111: "error on no tag" -> "no tag"

pkg/localstorev2/localstore_test.go
```diff
@@ -0,0 +1,508 @@
...
+		_, err = lstore.Upload(context.TODO(), false, 0)
+		if err == nil {
+			t.Fatal("expected error")
+		}
+	})
+
+	for _, tc := range []struct {
+		name   string
+		root   swarm.Chunk
+		chunks []swarm.Chunk
```
mrekucci, line 128: I would recommend having only the `chunks` field and making the first generated chunk always a root chunk, then you can remove the test `name` and `root` fields. The same goes for the `testPinStore`.

pkg/localstorev2/localstore_test.go
```diff
@@ -0,0 +1,508 @@
...
+			if info.Seen != seen {
+				t.Fatalf("unexpected seen count in session, exp %d found %d", seen, info.Seen)
+			}
+
+			if !info.Address.Equal(addr) {
+				t.Fatalf("unexpected swarm reference, exp %s, found %s", addr, info.Address)
+			}
+		}
+
+		t.Run("commit", func(t *testing.T) {
```
mrekucci, line 244: "commit" -> "done"

pkg/localstorev2/localstore_test.go
```diff
@@ -0,0 +1,508 @@
...
+
+			sessionInfo, err = lstore.GetSessionInfo(id)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			verify(t, sessionInfo, id, 20, 10, chunks[0].Address())
+		})
+
+		t.Run("rollback", func(t *testing.T) {
```
mrekucci, line 286: "rollback" -> "cleanup"

pkg/localstorev2/localstore_test.go
```diff
@@ -0,0 +1,508 @@
...
+
+func diskLocalstore(t *testing.T) func() (*localstore.DB, error) {
+	t.Helper()
+
+	return func() (*localstore.DB, error) {
+		dir, err := ioutil.TempDir(".", "testrepo*")
+		if err != nil {
+			t.Fatal(err)
+		}
+		t.Cleanup(func() { os.RemoveAll(dir) })
```
mrekucci, line 342: The returned error should be checked and printed in case it's not `nil`.

pkg/localstorev2/internal/pinning/pinning_test.go
```diff
@@ -341,3 +342,11 @@ func TestPinCollectionItem(t *testing.T) {
 		})
 	}
 }
+
+func TestPinChunkItem(t *testing.T) {
+	t.Parallel()
+
+	storagetest.TestItemClone(t, &storagetest.ItemCloneTest{
```
mrekucci, line 349: Seems to be too long, maybe consider:
```Golang
storagetest.TestItemClone(t, &storagetest.ItemCloneTest{
	Item: &pinstore.PinChunkItem{
		UUID: pinstore.NewUUID(),
		Addr: swarmtesting.RandomAddress(),
	},
})
```

mrekucci (commented):

pkg/localstorev2/internal/upload/uploadstore.go
aloknerurkar: @mrekucci @janos 

So currently the serialization interface is defined in the storagev2 package. This is a custom interface that is minimal in order to allow other types of serialization formats. For eg. json, protobuf etc. Going ahead we plan to use this same store as the statestore. There we will have other encodings.

I think, in future when we start using generics there will be a better solution to this. We can have a generic type that could implement different encodings and this way we can reuse the serialization. Wdyt?

Also, in order to change this we will have to change a lot of files. So either way, it would be better to do this in a separate PR.

mrekucci (approved):

PR #3743, opened 2023-01-30
feat: redistribution state unit test

umerm-work:
### Checklist

### Description
Add unit test remaining in #3683 

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3743)

### DIFF
--- a/pkg/storageincentives/redistributionstate_test.go
+++ b/pkg/storageincentives/redistributionstate_test.go
@@ -11,6 +11,7 @@ import (
 	erc20mock "github.com/ethersphere/bee/pkg/settlement/swap/erc20/mock"
 	"github.com/ethersphere/bee/pkg/statestore/mock"
 	transactionmock "github.com/ethersphere/bee/pkg/transaction/mock"
+	"github.com/google/go-cmp/cmp"
 	"math/big"
 	"testing"
 )
@@ -33,7 +34,7 @@ func createRedistribution(t *testing.T, erc20Opts []erc20mock.Option, txOpts []t
 	}
 	state, err := NewRedistributionState(log.Noop, common.Address{}, mock.NewStateStore(), erc20mock.New(erc20Opts...), transactionmock.New(txOpts...))
 	if err != nil {
-		t.Fatal("failed to connet")
+		t.Fatal("failed to connect")
 	}
 	return state
 }
@@ -47,10 +48,8 @@ func TestState(t *testing.T) {
 		LastPlayedRound: 2,
 		LastFrozenRound: 2,
 		Block:           2,
-		Fees:            big.NewInt(0),
-		Reward:          big.NewInt(0),
 	}
-	expected := Status{
+	want := Status{
 		Phase:           commit,
 		IsFrozen:        false,
 		Round:           2,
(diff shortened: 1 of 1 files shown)


pkg/storageincentives/redistributionstate_test.go
```diff
@@ -0,0 +1,183 @@
...
+		LastPlayedRound: 2,
+		LastFrozenRound: 0,
+		Block:           2,
+		Fees:            big.NewInt(0),
+		Reward:          big.NewInt(0),
+	}
+	t.Run("all state success", func(t *testing.T) {
+		t.Parallel()
+		state := createRedistribution(t, nil, nil)
+		if len(input.Phase.String()) > 0 && input.Round > 0 && input.Block > 0 {
```
istae, line 67: what is the purpose of this check, we can just insert the input diretly

pkg/storageincentives/redistributionstate_test.go
```diff
@@ -0,0 +1,183 @@
...
+		IsFrozen:        false,
+		Round:           2,
+		LastWonRound:    2,
+		LastPlayedRound: 2,
+		LastFrozenRound: 2,
+		Block:           2,
+		Fees:            big.NewInt(0),
+		Reward:          big.NewInt(0),
+	}
+	expected := Status{
```
mrekucci, line 53: The convention is to name it `want`.

pkg/storageincentives/redistributionstate_test.go
```diff
@@ -0,0 +1,183 @@
...
+			state.SetCurrentEvent(input.Phase, input.Round, input.Block)
+		}
+		state.SetLastWonRound(input.LastWonRound)
+		state.SetFrozen(input.IsFrozen, input.LastFrozenRound)
+		state.SetLastPlayedRound(input.LastPlayedRound)
+		got, err := state.Status()
+		if err != nil {
+			t.Fatal("failed to get state")
+		}
+		if got != nil && (got.IsFrozen != expected.IsFrozen || got.LastFrozenRound != expected.LastFrozenRound ||
```
istae, line 77: got should not be nil

pkg/storageincentives/redistributionstate_test.go
umerm-work: SetCurrentEvent requires three inputs. Anyway, yes we can directly add them.

pkg/storageincentives/redistributionstate_test.go
mrekucci, line 77: You can use the `cmp.Diff` function for comparison (as in other tests):
```Golang
if diff := cmp.Diff(want, have); diff != "" {
	t.Errorf("Clone(...): result mismatch (-want +have):\n%s", diff)
}
```

mrekucci (changes requested):

pkg/storageincentives/redistributionstate_test.go
umerm-work: This doesn't work as we have two `big.Int` properties. it would always fail. 

I have also used these. It won't work.

`reflect.DeepEqual(got, have)`
`cmp.Equal(got, have)`

istae: `firstWinResult.Reward.Cmp(expectedReward) != 0` works, right ?

pkg/storageincentives/redistributionstate_test.go
```diff
@@ -0,0 +1,181 @@
...
+		IsFrozen:        false,
+		Round:           2,
+		LastWonRound:    2,
+		LastPlayedRound: 2,
+		LastFrozenRound: 0,
+		Block:           2,
+		Fees:            big.NewInt(0),
+		Reward:          big.NewInt(0),
+	}
+	t.Run("all state success", func(t *testing.T) {
```
istae, line 64: we do not need a separate test case as you only have one input and expected output

pkg/storageincentives/redistributionstate_test.go
mrekucci: > This doesn't work as we have two big.Int properties. it would always fail.

You don't specify why it would fail, but I assume it's because of the unexported fields. If this is the case, you can pass the following option to the `cmp.Diff` call: `cmp.Option{cmp.AllowUnexported(new(big.Int))}`.

pkg/storageincentives/redistributionstate_test.go
umerm-work: Right

pkg/storageincentives/redistributionstate_test.go
umerm-work: Done thanks guys.

istae (approved):

PR #3755, opened 2023-02-01
feat: localstore container cache interfaces

aloknerurkar:
### Checklist

### Description
Cache-related interfaces in the localstore container. The difference between Cache and other components is that the cache is instantiated once and its size can be configured. Each Get and Put operation needs to be guarded by a transaction as these may include a few cache-related updates.

Also added tests to cover some error cases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3755)

### DIFF
--- a/pkg/localstorev2/internal/cache/cache.go
+++ b/pkg/localstorev2/internal/cache/cache.go
@@ -143,9 +143,9 @@ type Cache struct {
 // New creates a new Cache component with the specified capacity. The store is used
 // here only to read the initial state of the cache before shutdown if there was
 // any.
-func New(ctx context.Context, storg internal.Storage, capacity uint64) (*Cache, error) {
+func New(ctx context.Context, store internal.Storage, capacity uint64) (*Cache, error) {
 	state := &cacheState{}
-	err := storg.IndexStore().Get(state)
+	err := store.IndexStore().Get(state)
 	if err != nil && !errors.Is(err, storage.ErrNotFound) {
 		return nil, fmt.Errorf("failed reading cache state: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


pkg/localstorev2/internal/cache/cache.go
```diff
@@ -264,12 +264,12 @@ func (c *Cache) pushBack(
 
 // Putter returns a Storage.Putter instance which adds the chunk to the underlying
 // chunkstore and also adds a Cache entry for the chunk.
-func (c *Cache) Putter(store storage.Store, chStore storage.ChunkStore) storage.Putter {
+func (c *Cache) Putter(storg internal.Storage) storage.Putter {
```
mrekucci, line 267: `storg` -> `store` or https://www.abbreviations.com/serp.php?st=Storage&category=IT&qtype=2
The same is below.

pkg/localstorev2/localstore.go
```diff
@@ -348,3 +389,34 @@ func (db *DB) Pins() ([]swarm.Address, error) {
 func (db *DB) HasPin(root swarm.Address) (bool, error) {
 	return pinstore.HasPin(db.repo.IndexStore(), root)
 }
+
+// Lookup is the implementation of the CacheStore.Lookup function.
```
mrekucci, line 393: Change `function` to `method`. The same is above and below.

pkg/storagev2/repository.go
```diff
@@ -12,31 +12,38 @@ import (
 
 // Repository is a collection of stores that provides a unified interface
 // to access them. Access to all stores can be guarded by a transaction.
-type Repository struct {
+type Repository interface {
```
mrekucci, line 15: What is the purpose of having a `Repository` interface?

aloknerurkar: So in the localstore I felt like we should have some tests related to errors. In order to achieve that I had to override the repository in the DB pointer. As the repository is a pointer it is not possible to extend it and change behavior.

Check the new tests in localstore_test.go related to rollbacks.

mrekucci (approved):

notanatol (approved):

PR #3757, opened 2023-02-02
feat: depthmonitor sets init storage radius only for new nodes

istae:
### Checklist

### DIFF
--- a/pkg/topology/depthmonitor/depthmonitor.go
+++ b/pkg/topology/depthmonitor/depthmonitor.go
@@ -70,6 +70,7 @@ func New(
 	logger log.Logger,
 	warmupTime time.Duration,
 	wakeupInterval time.Duration,
+	freshNode bool,
 ) *Service {
 
 	s := &Service{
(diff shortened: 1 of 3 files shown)


notanatol (approved):

pkg/topology/depthmonitor/depthmonitor.go
```diff
@@ -156,6 +157,16 @@ func (s *Service) manage(warmupTime, wakeupInterval time.Duration) {
 			}
 		}
 	}
+	next()
+
+	for {
+		select {
+		case <-s.quit:
+			return
+		case <-time.After(wakeupInterval):
```
mrekucci, line 166: I'd recommend using `NewTimer` for efficiency. The documentation says:
> The underlying Timer is not recovered by the garbage collector until the timer fires. If efficiency is a concern, use NewTimer instead and call Timer.Stop if the timer is no longer needed.

istae: efficiency is not a concern though, at least in this context :)

mrekucci (approved):

PR #3760, opened 2023-02-04
fix(pusher): move stamp validation just before pushing

istae:
### Checklist

### Description
Because pusher does stamp validation before the semaphore acquire, under extreme heavy loads (as seen in the testnet), the pusher could be overwhelmed and batches that originally passed the validation may become expired while waiting in the queue to be pushed.

The fix is to move the validation to just before pushing the chunk.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3760)

### DIFF
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -113,13 +113,9 @@ func (s *Service) chunksWorker(warmupTime time.Duration, tracer *tracing.Tracer)
 	// inflight.set handles the backpressure for the maximum amount of inflight chunks
 	// and duplicate handling.
 	chunks, repeat, unsubscribe := s.storer.SubscribePush(ctx, s.inflight.set)
-	go func() {
-		<-s.quit
+	defer func() {
 		unsubscribe()
 		cancel()
-		if !timer.Stop() {
-			<-timer.C
-		}
 	}()
 
 	ctxLogger := func() (context.Context, log.Logger) {
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -127,6 +127,8 @@ func (s *Service) chunksWorker(warmupTime time.Duration, tracer *tracing.Tracer)
 	}
 
 	push := func(op *Op) {
+		defer wg.Done()
+
 		s.metrics.TotalToPush.Inc()
 		ctx, logger := ctxLogger()
 		startTime := time.Now()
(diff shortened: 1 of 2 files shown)


mrekucci (approved):

pkg/pusher/pusher.go
```diff
@@ -129,35 +125,44 @@ func (s *Service) chunksWorker(warmupTime time.Duration, tracer *tracing.Tracer)
...
-				s.metrics.TotalErrors.Inc()
-				s.metrics.ErrorTime.Observe(time.Since(startTime).Seconds())
-				loggerV1.Debug("cannot push chunk", "chunk_address", op.Chunk.Address(), "error", err)
-				return
+
+		if err := s.valid(op.Chunk); err != nil {
+			logger.Warning("stamp with is no longer valid, skipping syncing for chunk", "batch_id", hex.EncodeToString(op.Chunk.Stamp().BatchID()), "chunk_address", op.Chunk.Address(), "error", err)
+			ctx, cancel := context.WithTimeout(ctx, chunkStoreTimeout)
+			defer cancel()
+			if err = s.storer.Set(ctx, storage.ModeSetSync, op.Chunk.Address()); err != nil {
```
aloknerurkar, line 141: This only needs to be done if it is deferred push as this updates the pushIndex and other localstore entries. These are not present for direct push chunks.

For direct push we need to return error on the op.Err channel like below.

aloknerurkar (changes requested):

pkg/pusher/pusher.go
```diff
@@ -176,59 +183,52 @@ func (s *Service) chunksWorker(warmupTime time.Duration, tracer *tracing.Tracer)
...
+					chunks = nil
+					continue
+				}
+				select {
+				case cc <- &Op{Chunk: ch, Direct: false}:
+				case <-s.quit:
+					return
 				}
-				cancel()
+			case apiC := <-s.smugler:
```
zelig, line 199: smugler -> smuggler

pkg/pusher/pusher.go
```diff
@@ -129,35 +125,46 @@ func (s *Service) chunksWorker(warmupTime time.Duration, tracer *tracing.Tracer)
...
-			if err := s.pushChunk(ctx, op.Chunk, logger, op.Direct); err != nil {
-				// warning: ugly flow control
-				// if errc is set it means we are in a direct push,
-				// we therefore communicate the error into the channel
-				// otherwise we assume this is a buffered upload and
-				// therefore we repeat().
-				if op.Err != nil {
-					op.Err <- err
+
+		if !op.Direct {
```
aloknerurkar, line 137: So we are not returning error for direct upload chunks?

I was thinking of having this check inside the below valid if loop. If the op.Direct is set, you can return the validChunk error on op.Err back instead of going ahead to push. No?

aloknerurkar (approved):

PR #3766, opened 2023-02-07
feat(localstorev2): netstore interfaces

aloknerurkar:
### Checklist

### Description
This PR adds the netstore interfaces as specified in the new localstore design. The Netstore is a special storage.Getter and storage.Putter which deals with the network. It will be used for direct upload and download. On direct upload, it will straight away push the chunks to the pusher. On download, it will retrieve the chunk from the network if not found locally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3766)

### DIFF
--- a/pkg/localstorev2/netstore.go
+++ b/pkg/localstorev2/netstore.go
@@ -7,33 +7,80 @@ import (
 	"github.com/ethersphere/bee/pkg/pusher"
 	storage "github.com/ethersphere/bee/pkg/storagev2"
 	"github.com/ethersphere/bee/pkg/swarm"
+	"golang.org/x/sync/errgroup"
 )
 
-func (db *DB) DirectUpload() storage.Putter {
-	return storage.PutterFunc(func(ctx context.Context, ch swarm.Chunk) error {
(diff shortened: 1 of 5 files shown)


pkg/localstorev2/netstore.go
```diff
@@ -0,0 +1,102 @@
...
+
+// Download is the implementation of the NetStore.Download method.
+func (db *DB) Download(cache bool) storage.Getter {
+	return storage.GetterFunc(func(ctx context.Context, address swarm.Address) (swarm.Chunk, error) {
+		ch, err := db.Lookup().Get(ctx, address)
+		switch {
+		case err == nil:
+			return ch, nil
+		case errors.Is(err, storage.ErrNotFound):
+			if db.retrieval != nil {
```
istae, line 72: what happens if this is nil? can this ever be nil?

aloknerurkar: If the retrieval service is not set it will just return the ErrNotFound. Right now the way to provide it is in the Options. We can have some sort of validate function on the Options which checks for necessary fields and returns error if they are not set. But this could be a valid case for eg in devmode where we dont have network retrieval.

istae (approved):

PR #3769, opened 2023-02-08
fix: go1.20 with linting

notanatol:
### Checklist

### Description
separate PR for go1.20 and linter fixes for the deprecated functionality.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3769)


go.mod
```diff
@@ -23,11 +23,11 @@ require (
 	github.com/hashicorp/golang-lru v0.5.5-0.20210104140557-80c98217689d
 	github.com/ipfs/go-cid v0.3.2
 	github.com/kardianos/service v1.2.0
-	github.com/libp2p/go-libp2p v0.24.2
+	github.com/libp2p/go-libp2p v0.24.3-0.20230207035812-313b080ea4e2
```
notanatol, line 26: this will have to be updated to the release version once it's out.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -134,6 +133,8 @@ type Options struct {
 	Registry         *prometheus.Registry
 }
 
+var registerMetricsOnce sync.Once
```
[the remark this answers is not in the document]
notanatol: fixed both

janos (approved):
Thanks.

vladopajic:
Would it make sense to split this PR in two separate? 
First PR could be go version bump which could be merged to libp2p branch and second PR with lint version bump and fixes could be merged to master

notanatol:
> Would it make sense to split this PR in two separate? First PR could be go version bump which could be merged to libp2p branch and second PR with lint version bump and fixes could be merged to master

yes, but for a few files, because most of the fixes are based on the deprecations introduced in `go1.20` like `math/rand` being swapped for `crypto/rand`
anyway, the files would be:
- pkg/api/chunk_test.go
- pkg/file/pipeline/builder/builder_test.go
- pkg/p2p/libp2p/internal/breaker/breaker_test.go

vladopajic:
@notanatol It looks like `crypto/rand` is also be available in [go v1.17](https://pkg.go.dev/crypto/rand@go1.17.13). If this is the case then bee could have lint changes fixed in master branch?

go.mod
vladopajic: would it make sense to have this comment on this line as well? just to serve as reminder to developers in the future, since this comment will fade once this PR is closed

PR #3775, opened 2023-02-09
feat: localstorev2 reserve

istae:
### Checklist

### Description
The new reserve will maintain the storage radius, not the batchstore, so the depthmonitor is no longer.
The callback mechanism between localstore and batchstore is no longer

During migrations, BinIDs must be reset.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3775)


pkg/localstorev2/internal/reserve/reserve.go
```diff
@@ -0,0 +1,273 @@
...
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			return nil, err
+		}
+	}
+
+	var size int
+	err = store.Iterate(storagev2.Query{
```
aloknerurkar, line 75: So ideally if everything works as expected we should only have batchRadiusItems for the chunks we care about. So we should be able to just use `Count` for this and get the size right?

The reason I am saying this is tomorrow if we write a more optimum Count mechanism or use some other DB which provides one, we would do this much faster.

pkg/localstorev2/internal/reserve/sampler.go
```diff
@@ -0,0 +1,220 @@
...
+	chunkStore := store.ChunkStore()
+
+	t := time.Now()
+
+	// Phase 1: Iterate chunk addresses
+	g.Go(func() error {
+		defer close(addrChan)
+		iterationStart := time.Now()
+
+		err := indexStore.Iterate(storage.Query{
```
aloknerurkar, line 81: So this is the only thing we need really for the sampler from the internal/reserve package. I think it would be better to have the sampler code in the main localstorev2 package by creating a function for this in internal/reserve.

Mainly because I dont think of this as being internal code. Also, in the next phase, we will have inclusion proofs which are in the `bmt` package. So there might be some interaction there. Also, there were ideas about using a different hmacr.

pkg/localstorev2/internal/reserve/reserve.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	rs.size = size
+
+	return rs, nil
+}
+
+func (r *Reserve) Putter(store internal.Storage) storagev2.Putter {
+
+	return storagev2.PutterFunc(func(ctx context.Context, chunk swarm.Chunk) error {
```
aloknerurkar, line 96: For the putter we will need to do 2 more things:
1. First is to check for index collisions. You can refer to the internal/upload package for example.
2. We need to use the new package @mrekucci is adding to store stamps. We have found some issue with having stamps in chunkstore, so the reserve will need to manage its own stamps basically.

pkg/localstorev2/storer.go
```diff
@@ -306,17 +341,35 @@ func New(dirPath string, opts *Options) (*DB, error) {
...
-		logger:         opts.Logger.WithName(loggerName).Register(),
-		quit:           make(chan struct{}),
-		bgCacheWorkers: make(chan struct{}, 16),
-		dbCloser:       dbCloser,
-	}, nil
+	logger := opts.Logger.WithName(loggerName).Register()
+
+	reserveRadius := opts.Batchstore.GetReserveState().Radius
+
+	reserve, err := reserve.New(opts.Address, repo.IndexStore(), opts.ReserveCapacity, reserveRadius, opts.StateStore, opts.RadiusSetter, logger)
```
mrekucci, line 348: Variable 'reserve' collides with the imported package name.

mrekucci (commented):
Would be nice to have some meaningful comments for the struct/method/functions and also add/increase test coverage.

pkg/localstorev2/internal/reserve/reserve.go
istae: We need to sync about this.

pkg/localstorev2/internal/reserve/items.go
```diff
@@ -0,0 +1,187 @@
...
+	return ""
+}
+
+const batchRadiusItemSize = 1 + swarm.HashSize + swarm.HashSize + 8
+
+func (b *batchRadiusItem) Marshal() ([]byte, error) {
+
+	buf := make([]byte, batchRadiusItemSize)
+
+	i := 0
```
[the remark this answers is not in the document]
aloknerurkar: So I dont understand how this is more error prone?

The way you are copying is actually error-prone as you are giving access to the whole buffer
`buf[i:]` instead of the exact buffer which is preallocated `buf[start:end]`. Also knowing the start and end indexes from the code is better than following value of `i`.

Also, this makes it inconsistent with the other packages as well. Not sure if there is a strong enough reason for this. Different ways introduce different problems.

pkg/localstorev2/internal/reserve/items.go
```diff
@@ -0,0 +1,263 @@
...
+	return string(bin) + string(batchID)
+}
+
+func (b *batchRadiusItem) Clone() storage.Item {
+	if b == nil {
+		return nil
+	}
+	return &batchRadiusItem{
+		Bin:     b.Bin,
+		BatchID: copyBytes(b.BatchID, swarm.HashSize),
```
mrekucci, line 45: Using the `copyBytes` primitive does not create an exact clone, because a slice of size `swarm.HashSize` is allocated, even if the `BatchID` is nil. You should also add this edge case to the tests.

pkg/localstorev2/internal/reserve/reserve.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/ethersphere/bee/pkg/log"
+	storagev2 "github.com/ethersphere/bee/pkg/storagev2"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/ethersphere/bee/pkg/topology"
+)
+
+// loggerName is the tree path name of the logger for this package.
+const loggerName = "reserve"
+
+type Reserve struct {
```
mrekucci, line 22: Would be nice to have tests for the `Reserve` methods.

pkg/localstorev2/storer.go
```diff
@@ -108,6 +113,17 @@ type NetStore interface {
 	PusherFeed() <-chan *pusher.Op
 }
 
+type BinC struct {
```
mrekucci, line 116: Some description would be great to have. The same goes for the `ReverseStore` below.

pkg/localstorev2/internal/reserve/reserve.go
istae: these are extensively tested through the localstorev2/reserve_test.go

pkg/localstorev2/internal/upload/uploadstore.go
```diff
@@ -593,7 +593,7 @@ func Iterate(ctx context.Context, s internal.Storage, startFrom swarm.Chunk, con
 
 	return s.IndexStore().Iterate(q, func(r storage.Result) (bool, error) {
 		pi := r.Entry.(*pushItem)
-		chunk, err := s.ChunkStore().GetWithStamp(ctx, pi.Address, pi.BatchID)
+		chunk, err := s.ChunkStore().Get(ctx, pi.Address)
```
istae, line 596: this function `GetWithStamp`, does not exist

pkg/localstorev2/internal/chunkstamp/chunkstamp.go
```diff
@@ -47,7 +46,7 @@ type item struct {
 
 // ID implements the storage.Item interface.
 func (i *item) ID() string {
-	return path.Join(string(i.stamp.BatchID()), string(i.stamp.Index()))
+	return path.Join(string(i.stamp.BatchID()), string(i.stamp.BatchID()))
```
aloknerurkar, line 49: Was this changed by mistake? It should be Index.

pkg/localstorev2/internal/chunkstamp/chunkstamp.go
```diff
@@ -229,3 +228,15 @@ func DeleteAll(s internal.Storage, namespace string, addr swarm.Address) error {
 	}
 	return errs.ErrorOrNil()
 }
+
```
aloknerurkar, line 231: Would be good to have some comments for doc.

pkg/localstorev2/internal/reserve/items.go
```diff
@@ -0,0 +1,264 @@
...
+	Bin   uint8
+	BinID uint64
+}
+
+func (b *binItem) Namespace() string {
+	return "binID"
+}
+
+func (b *binItem) ID() string {
+	return string(b.Bin)
```
aloknerurkar, line 188: For this.... we need to format the string like we used to in the old code right?
`sync|%03d`

This can lead to incorrect prefix matches. This is for all the binIDs in the keys.

pkg/localstorev2/internal/reserve/items.go
```diff
@@ -0,0 +1,264 @@
...
+	"path"
+
+	storage "github.com/ethersphere/bee/pkg/storagev2"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+var errMarshalInvalidAddress = errors.New("marshal: invalid address")
+var errUnmarshalInvalidSize = errors.New("unmarshal: invalid size")
+
+type batchRadiusItem struct {
```
aloknerurkar, line 19: Would be good to have some comments about these items and how they should be used.

pkg/localstorev2/internal/reserve/reserve.go
```diff
@@ -0,0 +1,358 @@
...
+	radius   uint8
+}
+
+/*
+	pull by 	bin - binID
+	evict by 	bin - batchID
+	sample by 	bin
+*/
+
+func New(baseAddr swarm.Address, store storagev2.Store, capacity int, reserveRadius uint8, radiusSetter topology.SetStorageRadiuser, logger log.Logger) (*Reserve, error) {
```
aloknerurkar, line 57: I would prefer this function definition to be multiline.

```
func New(
    baseAddr swarm.Address,
    store storagev2.Store,
    ......
) (*Reserve, error) {
}
```

pkg/localstorev2/internal/reserve/reserve.go
```diff
@@ -0,0 +1,358 @@
...
+		case loaded && !item.ChunkIsImmutable:
+			prev := binary.BigEndian.Uint64(item.BatchTimestamp)
+			curr := binary.BigEndian.Uint64(chunk.Stamp().Timestamp())
+			if prev >= curr {
+				return errOverwriteOfNewerBatch
+			}
+			err = stampindex.Store(indexStore, reserveNamespace, chunk)
+			if err != nil {
+				return fmt.Errorf("failed updating stamp index: %w", err)
+			}
```
aloknerurkar, line 125: Here we also need to delete the old chunk reference from the ChunkStore.

pkg/localstorev2/internal/upload/uploadstore.go
aloknerurkar: Yes this will be fixed with @notanatol s changes for PushSub

pkg/localstorev2/internal/reserve/reserve.go
```diff
@@ -0,0 +1,358 @@
...
+		Factory:       func() storagev2.Item { return &chunkBinItem{} },
+		Prefix:        binIDToString(bin, startBinID),
+		PrefixAtStart: true,
+	}, func(res storagev2.Result) (bool, error) {
+		item := res.Entry.(*chunkBinItem)
+		if item.Bin > bin {
+			return true, nil
+		}
+
+		stop, err := cb(item.Address, item.BinID)
```
aloknerurkar, line 197: I dont think this will work. We will need to store the BatchID in the chunkBinItem. Both these functions need to read the right stamp for the chunk. Also this needs to be in the key as chunks with multiple stamps need to be pullsynced separately so that the peers get all the stamps associated with the chunk.

This is again the same for Sampling. In future if we delimit the sampling based on batch value, we need to consider all different stamps for the chunk.

pkg/localstorev2/reserve.go
```diff
@@ -0,0 +1,484 @@
...
+	}
+}
+
+// ReserveLastBinIDs returns all of the highest binIDs from all the bins in the reserve.
+func (db *DB) ReserveLastBinIDs() ([]uint64, error) {
+	return db.reserve.LastBinIDs(db.repo.IndexStore())
+}
+
+// BinC is the result returned from the SubscribeBin channel that contains the chunk address and the binID
+type BinC struct {
```
aloknerurkar, line 202: I think its better to just return the constructed chunk. This way pullsync can just check for stamp validity and directly send it.

I dont think we should expose this BinID outside the package. This will limit us if we come up with different design later. The user of this `SubscribeBin` function only cares about the chunks and not the internal data structures in the implementation.

aloknerurkar (changes requested):

pkg/localstorev2/internal/chunkstamp/chunkstamp.go
istae: you may be confusing this with the stampindex?

pkg/localstorev2/reserve.go
istae: the steps of the pullsync are:

peer1 requests chunks with a certain binID interval from peer2
peer2 prepares an offer of addresses and sends it
peer1 picks the addresses he wants from that offer and sends back the list of addresses
peer2 gets the chunks from the reserve and sends them to peer1

basically peer2 wants to iterate on the list of addresses and NOT chunks

pkg/localstorev2/internal/reserve/reserve.go
istae: 👍

pkg/localstorev2/internal/reserve/items.go
istae: i think because the bins are single byte, uint8, this is okay

istae: I had to make use of `sync|%03d` because we were using fmt.Sprintf and there was no clear formatting flag for me to use.

pkg/localstorev2/reserve.go
istae: with the multi stamp per chunk changes, we may need to rewrite the pullsync

pkg/localstorev2/internal/reserve/reserve.go
aloknerurkar: We need to have separate tests for this file. The localstorev2/reserve tests should only test the interaction with transactions, but the internal code needs to be tested separately.

```
PASS
	github.com/ethersphere/bee/pkg/localstorev2/internal/reserve	coverage: 27.6% of statements
ok  	github.com/ethersphere/bee/pkg/localstorev2/internal/reserve	0.125s	coverage: 27.6% of statements
```

There is no coverage reported here as some other package is testing this package. I think this is fundamentally wrong. If we make changes in the internal package, we will have to see which test to modify in the main one. I think this is a recipe for disaster. A seemingly harmless piece code in the old localstore caused leaks for years. If you dont want to do this in the current PR we can create a separate issue for this.

This is actually happening right now as there is a bug in the `EvictBatch` function as we are not deleting the batchRadiusItem. But this is not being checked anywhere.

pkg/localstorev2/reserve_test.go
```diff
@@ -0,0 +1,825 @@
...
+		opts.Batchstore = bs
+		st, err = storer.New(context.Background(), dir, opts)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		waitForRadius(t, st.Reserve(), 3)
+	})
+
+	t.Run("radius decrease due to under utilization", func(t *testing.T) {
```
aloknerurkar, line 403: This test is flaking for me locally.

pkg/localstorev2/internal/reserve/reserve.go
```diff
@@ -0,0 +1,403 @@
...
+			err = chunkstamp.Delete(indexStore, reserveNamespace, batchRadius.Address, batchRadius.BatchID)
+			if err != nil {
+				return false, err
+			}
+
+			err = chunkStore.Delete(context.Background(), batchRadius.Address)
+			if err != nil {
+				return false, err
+			}
+
```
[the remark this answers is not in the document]
istae: We already delete it, line 288. But it's in the iterater

pkg/localstorev2/internal/reserve/reserve.go
istae: We are deleting it but ok

pkg/localstorev2/reserve.go
```diff
@@ -0,0 +1,496 @@
...
+// ReservePutter returns a PutterSession for inserting chunks into the reserve.
+func (db *DB) ReservePutter(ctx context.Context) PutterSession {
+
+	trx, commit, rollback := db.repo.NewTx(ctx)
+
+	triggerBins := make(map[uint8]bool)
+	count := 0
+
+	// lock to avoid Putting a chunk that expires during the session.
+	db.lock.Lock(reserveLock)
```
aloknerurkar, line 100: This could be okay I think, but would it be better to have the Lock/Unlock in the individual functions of the putterSession? This way we will have lesser lock contention. Multiple calls to ReservePutter can also go through. For eg lot of routines would be using this function simultaneously in pullsync I think. Wdyt?

pkg/localstorev2/reserve.go
```diff
@@ -0,0 +1,496 @@
...
+}
+
+// BinC is the result returned from the SubscribeBin channel that contains the chunk address and the binID
+type BinC struct {
+	Address swarm.Address
+	BinID   uint64
+}
+
+// SubscribeBin returns a channel that feeds all the chunks in the reserve from a certain bin between a start and end binIDs.
+func (db *DB) SubscribeBin(ctx context.Context, bin uint8, start, end uint64) (<-chan *BinC, func(), <-chan error) {
```
aloknerurkar, line 222: Would it be better to have the `errc` inside the `BinC`?

I think this would make the UX easier for the client as he would potentially need just 1 for loop.

pkg/localstorev2/reserve.go
istae: do you mean lock inside `PutterFunc` ?

PR #3794, opened 2023-02-15
feat(localstorev2): sharky recovery

vladopajic:
Sharky recovery part for #3750 issue.

Closes: #3750

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3794)

### DIFF
--- a/pkg/localstorev2/internal/chunkstore/chunkstore.go
+++ b/pkg/localstorev2/internal/chunkstore/chunkstore.go
@@ -53,6 +53,9 @@ type retrievalIndexItem struct {
 	RefCnt    uint8
 }
 
+// TODO original type should be exported
+type RetrievalIndexItemTemp = retrievalIndexItem
+
 func (r *retrievalIndexItem) ID() string { return r.Address.ByteString() }
 
 func (retrievalIndexItem) Namespace() string { return "retrievalIdx" }
--- a/pkg/localstorev2/internal/chunkstore/chunkstore.go
+++ b/pkg/localstorev2/internal/chunkstore/chunkstore.go
@@ -53,9 +53,6 @@ type retrievalIndexItem struct {
 	RefCnt    uint8
 }
 
-// TODO original type should be exported
-type RetrievalIndexItemTemp = retrievalIndexItem
-
 func (r *retrievalIndexItem) ID() string { return r.Address.ByteString() }
 
 func (retrievalIndexItem) Namespace() string { return "retrievalIdx" }
--- a/pkg/localstorev2/internal/chunkstore/helpers.go
+++ b/pkg/localstorev2/internal/chunkstore/helpers.go
@@ -1,0 +1,47 @@
+// Copyright 2023 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package chunkstore
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/ethersphere/bee/pkg/sharky"
+	storage "github.com/ethersphere/bee/pkg/storagev2"
+)
+
+type LocationResult struct {
+	Err      error
+	Location sharky.Location
+}
+
+// IterateLocations iterates over entire retrieval index and plucks only sharky location.
+func IterateLocations(ctx context.Context, st storage.Store) <-chan LocationResult {
+	locationResultC := make(chan LocationResult, 128)
+
+	go func() {
+		defer close(locationResultC)
+
+		err := st.Iterate(storage.Query{
+			Factory: func() storage.Item { return new(retrievalIndexItem) },
+		}, func(r storage.Result) (bool, error) {
+			select {
+			case <-ctx.Done():
+				return true, ctx.Err()
+			default:
+			}
+
+			entry := r.Entry.(*retrievalIndexItem)
+			locationResultC <- LocationResult{Location: entry.Location}
+
+			return false, nil
+		})
+		if err != nil {
+			locationResultC <- LocationResult{Err: fmt.Errorf("iterate retrieval index error: %w", err)}
+		}
+	}()
+
+	return locationResultC
+}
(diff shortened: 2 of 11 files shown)


pkg/localstorev2/internal/chunkstore/chunkstore.go
```diff
@@ -53,6 +53,9 @@ type retrievalIndexItem struct {
 	RefCnt    uint8
 }
 
+// TODO original type should be exported
+type RetrievalIndexItemTemp = retrievalIndexItem
```
aloknerurkar, line 57: Instead of exporting this type, what we can do is write a package level function here:
```go
func IterateLocations(ctx context.Context, st storage.Store) <-chan Result {
    // do the query here
}
```

pkg/localstorev2/recover.go
```diff
@@ -0,0 +1,98 @@
...
+
+	return closer, nil
+}
+
+type locationResult struct {
+	err error
+	loc sharky.Location
+}
+
+func readSharkyLocations(locationResultC chan<- locationResult, store storage.Store) {
```
aloknerurkar, line 67: This function can be moved to internal/chunkstore

pkg/localstorev2/recover.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/ethersphere/bee/pkg/sharky"
+	storage "github.com/ethersphere/bee/pkg/storagev2"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+const (
+	sharkyDirtyFileName = ".DIRTY"
+)
+
+func sharkyRecovery(sharkyBasePath string, store storage.Store, opts *Options) (closerFn, error) {
```
aloknerurkar, line 25: Since this could be a potentially long-running task, we should be taking in a context in the New function and handling shutdown.

pkg/localstorev2/storer.go
```diff
@@ -212,9 +212,15 @@ func initDiskRepository(basePath string, opts *Options) (storage.Repository, io.
 	}
 
 	txStore := leveldbstore.NewTxStore(store)
+
+	recoveryCloser, err := sharkyRecovery(sharkyBasePath, txStore, opts)
```
aloknerurkar, line 216: This should be done before we start the Sharky instance as they will be operating on the same set of files underneath. This will be wrong.

aloknerurkar (changes requested):

vladopajic:
There is separate PR #3803 which adds new utilities to swarm package. Let's wait for that PR to be merged so those changes don't have to be included in this PR.

pkg/localstorev2/internal/chunkstore/helpers.go
```diff
@@ -0,0 +1,47 @@
...
+			Factory: func() storage.Item { return new(retrievalIndexItem) },
+		}, func(r storage.Result) (bool, error) {
+			select {
+			case <-ctx.Done():
+				return true, ctx.Err()
+			default:
+			}
+
+			entry := r.Entry.(*retrievalIndexItem)
+			locationResultC <- LocationResult{Location: entry.Location}
```
aloknerurkar, line 39: This is naked write. Should be prevented at all costs.

We could get a context done signal while we try to return the locationResult. So your select case should have a second case for this operation and not a default case. The entry can be initialized before the select.

pkg/localstorev2/internal/chunkstore/helpers.go
```diff
@@ -0,0 +1,47 @@
...
+
+			return false, nil
+		})
+		if err != nil {
+			locationResultC <- LocationResult{Err: fmt.Errorf("iterate retrieval index error: %w", err)}
```
aloknerurkar, line 44: Again you should have a select here to wait on Context cancellation. If there is a context cancellation on the caller side we should not expect this to be handled implicitly.

pkg/localstorev2/recover.go
```diff
@@ -0,0 +1,80 @@
...
+		return closer, err
+	}
+
+	defer func() {
+		if err := sharkyRecover.Close(); err != nil {
+			logger.Error(err, "failed closing sharky recovery")
+		}
+	}()
+
+	locationResultC := make(chan chunkstore.LocationResult, 128)
```
aloknerurkar, line 54: I dont see why we need this buffered channel. There is only 1 reader, so operations will only happen sequentially. See the other comments about naked writes. This entire operation will be controlled by context cancellation. So we can handle it immediately without using the buffered channel.

pkg/localstorev2/storer.go
```diff
@@ -202,6 +202,13 @@ func initDiskRepository(basePath string, opts *Options) (storage.Repository, io.
 		}
 	}
 
+	txStore := leveldbstore.NewTxStore(store)
```
aloknerurkar, line 205: We dont need this as this is read only operation. We can use the store directly here.

aloknerurkar (changes requested):

pkg/localstorev2/recover.go
vladopajic: Buffer was added in order to decouple goroutine in `chunkstore.IterateLocations` from goroutine running `addLocations`. 

It doesn't matter how many readers are there, but if reader is slower it will unnecessarily block writer which will impair performance.

aloknerurkar: So buffer of 128 does not guarantee unblocking anything. If there are millions of locations things could still block. So these type of magic numbers need to be avoided unless we know exactly why and how many we need.

vladopajic: Correct, for every number that is picked as capacity for buffer there could be case were this capacity is not going to be sufficient. 

However, adding small capacity for channel can still significantly improve perfs because reader and writer are going to decouple it's executions.

aloknerurkar: But this is not required. There is no evidence this will improve anything. Having a buffer can lead to producer overwhelming the consumer and taking CPU cycles from it. This might lead to an overall degradation in performance. When there are two routines interacting it is better to ensure they peform according to their speed and block the other. This is the most simple design and to change this there has to be some data to make any claims.

vladopajic: I forgot to mention that buffer has been removed from this channel.

We can continue this discussion in dedicated call on this topic.

aloknerurkar (approved):

PR #3815, opened 2023-02-20
fix(api): bad test case

notanatol:
### Checklist

### Description
Method `ResolveNameOrAddress` returns zero address for empty string but the test case was expecting an error.
The test wasn't failing simply because it itself was broken.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3815)

### DIFF
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -341,6 +341,8 @@ func TestParseName(t *testing.T) {
 	const bzzHash = "89c17d0d8018a19057314aa035e61c9d23c47581a61dd3a79a7839692c617e4d"
 	log := log.Noop
 
+	var errInvalidNameOrAddress = errors.New("invalid name or bzz address")
+
 	testCases := []struct {
 		desc       string
 		name       string
@@ -352,7 +354,7 @@ func TestParseName(t *testing.T) {
 		{
 			desc:    "empty name",
 			name:    "",
-			wantErr: api.ErrInvalidNameOrAddress,
+			wantAdr: swarm.ZeroAddress,
 		},
 		{
 			desc:    "bzz hash",
@@ -376,10 +378,10 @@ func TestParseName(t *testing.T) {
 			name: "not.good",
 			res: resolverMock.NewResolver(
 				resolverMock.WithResolveFunc(func(string) (swarm.Address, error) {
-					return swarm.ZeroAddress, api.ErrInvalidNameOrAddress
+					return swarm.ZeroAddress, errInvalidNameOrAddress
 				}),
 			),
-			wantErr: api.ErrInvalidNameOrAddress,
+			wantErr: errInvalidNameOrAddress,
 		},
 		{
 			desc:    "name resolved",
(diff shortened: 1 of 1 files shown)


istae (approved):

pkg/manifest/mantaray/marshal.go
```diff
@@ -256,13 +256,14 @@ func (n *Node) UnmarshalBinary(data []byte) error {
...
-				err := fmt.Errorf("not enough bytes for node fork: %d (%d)", (len(data) - offset), (nodeForkPreReferenceSize + refBytesSize))
-				return fmt.Errorf("%s on byte '%x': %w", err, []byte{b}, ErrInvalidManifest)
+				cause := fmt.Sprintf("not enough bytes for node fork: %d (%d)", (len(data) - offset), (nodeForkPreReferenceSize + refBytesSize))
+				return fmt.Errorf("%s on byte '%x': %w", cause, []byte{b}, ErrInvalidManifest)
 			}
 
 			err := f.fromBytes(data[offset : offset+nodeForkPreReferenceSize+refBytesSize])
 			if err != nil {
-				return fmt.Errorf("%s on byte '%x': %w", err, []byte{b}, ErrInvalidManifest)
+				cause := err.Error()
```
mrekucci, line 265: Not sure why the `cause` is factored out. The same is below.

pkg/p2p/streamtest/streamtest_test.go
```diff
@@ -375,7 +375,8 @@ func TestRecorder_resetAfterPartialWrite(t *testing.T) {
 
 		// stream should be closed and read should return streamtest.ErrStreamClosed
 		if _, err := rw.ReadString('\n'); !errors.Is(err, streamtest.ErrStreamClosed) {
-			return fmt.Errorf("got error %v, want %w", err, streamtest.ErrStreamClosed)
```
mrekucci, line 378: You can just change `%w` to `%v` if you don't want to wrap the error.

mrekucci (commented):
I would suggest postpone this PR until we move to Golang 1.20, where multi-errors are supported by the standard library: https://github.com/golang/go/issues/53435

pkg/manifest/mantaray/marshal.go
```diff
@@ -256,13 +256,14 @@ func (n *Node) UnmarshalBinary(data []byte) error {
...
+				return fmt.Errorf("%s on byte '%x': %w", cause, []byte{b}, ErrInvalidManifest)
```
vladopajic, line 266: neat: would it make sense to inline `err.Error()` in `fmt.Errorf` arguments list (to make code a bit more compact)?
If you decide to this, please update similar pattern on other occurrences in PR.

vladopajic (approved):
LGTM; left minor nit-picking comment

---

Nit-picking again, but PR title with `(scope)` looks better eg. `feat(lint): bump version to 1.51`

pkg/manifest/mantaray/marshal.go
notanatol: refactored

mrekucci (approved):

PR #3820, opened 2023-02-21
refactor: move out chunk stamp index from chunk store

### LINKED ISSUE #3783: Refactor disk txChunkStore
With the current design of the disk txChunkStore, we have some issues handling transactions that could be operating on the same indexes in the chunkstore. This is because we revert the state of the TxStore in the TxChunkStore and we can have multiple Txns happening at the same time that could be operating on the same indexes. Reverting them directly could cause inconsistencies.

The Txns in storage pkg don't guarantee ACID semantics strictly. The reason this is okay is in most components we will be operating on separate indexes. The only common point would be the ChunkStore. So we need to handle the reverse operations through the ChunkStore API and not look into how it is implemented right now with the Store as part of the transaction.

mrekucci:
### Checklist

### Description
Move the chunk stamp index from the chunk store into a separate package.

Closes #3783

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3820)

### DIFF
--- a/pkg/localstorev2/internal/chunkstamp/chunkstamp.go
+++ b/pkg/localstorev2/internal/chunkstamp/chunkstamp.go
@@ -1,0 +1,231 @@
+// Copyright 2023 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package chunkstamp
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"path"
+
+	"github.com/ethersphere/bee/pkg/localstorev2/internal"
+	"github.com/ethersphere/bee/pkg/postage"
+	"github.com/ethersphere/bee/pkg/storagev2"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/hashicorp/go-multierror"
+)
+
+var (
+	// errMarshalInvalidChunkStampItemNamespace is returned during marshaling if the namespace is not set.
+	errMarshalInvalidChunkStampItemNamespace = errors.New("marshal chunkstamp.Item: invalid namespace")
+	// errMarshalInvalidChunkStampAddress is returned during marshaling if the address is zero.
+	errMarshalInvalidChunkStampItemAddress = errors.New("marshal chunkstamp.item: invalid address")
+	// errUnmarshalInvalidChunkStampAddress is returned during unmarshaling if the address is not set.
+	errUnmarshalInvalidChunkStampItemAddress = errors.New("unmarshal chunkstamp.item: invalid address")
+	// errMarshalInvalidChunkStamp is returned if the stamp is invalid during marshaling.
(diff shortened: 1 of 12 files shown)


istae (approved):

pkg/localstorev2/internal/chunkstamp/chunkstamp.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// ID implements the storage.Item interface.
+func (i *item) ID() string {
+	return path.Join(string(i.stamp.BatchID()), string(i.stamp.Index()))
+}
+
+// Namespace implements the storage.Item interface.
+func (i *item) Namespace() string {
+	return path.Join("chunkStamp", i.address.ByteString())
```
aloknerurkar, line 58: I think you forgot to add the namespace here?

This should be prefixed by the namespace to avoid colliding between different namespaces.

pkg/localstorev2/internal/chunkstamp/chunkstamp.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// String implements the storage.Item interface.
+func (i item) String() string {
+	return path.Join(i.Namespace(), i.ID())
+}
+
+// Load returns swarm.Stamp related to the given address and batchID.
+// The storage.ErrNoStampsForChunk is returned if no record is found.
+func Load(s internal.Storage, namespace string, addr swarm.Address, batchID []byte) (swarm.Stamp, error) {
```
aloknerurkar, line 148: Now that we have a separate package for this would be it be better to have two functions?

Load with batch ID and LoadAny for getting any stamp.

pkg/localstorev2/internal/chunkstamp/chunkstamp.go
```diff
@@ -0,0 +1,224 @@
...
+		stamp:     chunk.Stamp(),
+	}
+	if err := s.IndexStore().Put(item); err != nil {
+		return fmt.Errorf("unable to put chunkstamp.item %s: %w", item, err)
+	}
+	return nil
+}
+
+// Delete removes the related stamps from the storage.
+func Delete(s internal.Storage, namespace string, addr swarm.Address) error {
```
aloknerurkar, line 196: I think we should call this DeleteAll.

pkg/localstorev2/internal/chunkstamp/chunkstamp.go
mrekucci: Done.

pkg/localstorev2/internal/chunkstamp/chunkstamp.go
mrekucci: Done.

pkg/localstorev2/internal/chunkstamp/chunkstamp.go
mrekucci: Done.

aloknerurkar (approved):

PR #3828, opened 2023-02-23
fix(comment-lint): disabled body rules

vladopajic:
Currently `bee`'s comment linting rules are vary strict about git commit body. This is causing CI lint job to fail for commits made by dependency bot because it is adding vary descriptive body of git commits. 

This PR:
- relaxes rules of commit body 
- changes `Get Commit Message` task to log short commit format 

Example of dependency bot PR (https://github.com/ethersphere/bee/pull/3827)

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3828)

### DIFF
--- a/.github/workflows/beekeeper.yml
+++ b/.github/workflows/beekeeper.yml
@@ -59,7 +59,7 @@ jobs:
       - name: Get Commit Message
         id: commit
         run: |
-          MSG=$(git log --format=%B -n 1 ${{github.event.after}})
+          MSG=$(git log --format=%s -n 1 ${{github.event.after}})
           echo "msg=${MSG}" >> $GITHUB_OUTPUT
       - name: Build - 0
         run: |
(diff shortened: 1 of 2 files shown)


notanatol (approved):

mrekucci (changes requested):
We follow the standard described [here](https://github.com/ethersphere/bee/blob/master/CODING.md#commit-messages), except that we do not require bodies because the PR numbers are referenced from the subject, thus I would not recommend relaxing the rules and changing this pattern.

vladopajic:
@mrekucci this PR just relaxes body of commits while keeping the same requirements for titles. Is there any specific need to restrict body of commit to only 72 characters? Wouldn't it be beneficial to have relaxed descriptions (bodies) of commits in order to better explain / summarise commits?

Issue #3742, opened 2023-01-30
nodes should have a minimum balance to cover transaction fees when entering a new storage incentives round


istae:
Before playing in a new round, the storage incentives agent should check that the wallet of the node has a minimum balance to cover the upcoming transactions.

One approach:
The minimum balance can be calculated as the current gasPrice * avgTransactionGasUsed * minimumTransactionsToCover where

avgTransactionGasUsed is the avg of the gas of claim, commit, and reveal transactions

and minimumTransactionsToCover is the number of transactions the wallet must be able to cover - the default could be 25.

ldeffenb:
And while we're at it, can we have a configurable "maximum gas price" where a node would opt not to play if the gas price at the start of the round (just before committing) is (arbitrarily according to the node operator) too high?

swarmHaseeb:
@ldeffenb Yes we already have a PR in pipeline for that :)

PR #3829, opened 2023-02-23
feat(storageincentives): game skip on insufficent funds

vladopajic:
### Checklist

### Description

Closes: #3742

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3829)


istae:
Please do not override the default checkboxes in the body :)

pkg/storageincentives/agent.go
```diff
@@ -472,3 +487,15 @@ func (a *Agent) wrapCommit(storageRadius uint8, sample []byte, key []byte) ([]by
...
 }
+
+func (a *Agent) canCommit(ctx context.Context) (bool, error) {
+	balance, err := a.backend.BalanceAt(ctx, a.state.ethAddress, nil)
+	if err != nil {
+		return false, err
+	}
+
+	minBalance := big.NewInt(minTxCountToCover)
+	minBalance.Mul(minBalance, a.state.AvgFee())
```
istae, line 498: the avgFee is zero at the beginning, maybe there should be some minimum default value instead of zero.

istae: also the avgFee may lose it's accuracy because these transactions only occur every few days, ideally what we want is to fetch the current gas price from the backend and multiply with some avg gas limit (this number is stable between some range)

pkg/storageincentives/agent.go
```diff
@@ -127,6 +130,18 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 		mtx.Unlock()
 
 		if round-1 == sampleRound { // the sample has to come from previous round to be able to commit it
+			canCommit, err := a.canCommit(ctx)
```
istae, line 133: this check should ideally be in play()

pkg/storageincentives/agent.go
istae: see gas limits https://gnosisscan.io/address/0x8c26b7ca61a6608b011cba43d8ca4476b6d8da17

pkg/storageincentives/agent.go
```diff
@@ -35,11 +35,19 @@ const loggerName = "storageincentives"
 const (
 	DefaultBlocksPerRound = 152
 	DefaultBlocksPerPhase = DefaultBlocksPerRound / 4
+
+	// min transactions required for each game
+	minTxCountToCover = 25
```
istae, line 40: the comment should be // min # of transactions our wallet should be able to cover
also on second thought 25 is too much, it should be 5.
if a node wins a round, then there are a total of 3 trxs. We leave a buffer of 2 trxs.

pkg/storageincentives/agent.go
```diff
@@ -361,6 +369,18 @@ func (a *Agent) play(ctx context.Context, round uint64) (uint8, []byte, error) {
...
 	}
 
+	hasEnough, err := a.hasEnoughFundsToPlay(ctx)
+	if err != nil {
+		a.logger.Error(err, "agent hasEnoughFundsToPlay failed")
+		return 0, nil, nil
+	}
+
+	if !hasEnough {
+		a.logger.Debug("agent has insufficient f to participate in next round")
```
istae, line 379: typo

pkg/storageincentives/agent.go
```diff
@@ -361,6 +369,18 @@ func (a *Agent) play(ctx context.Context, round uint64) (uint8, []byte, error) {
 		return 0, nil, nil
 	}
 
+	hasEnough, err := a.hasEnoughFundsToPlay(ctx)
```
istae, line 372: let's move this to after the `isPlaying` check.

If isPlaying is true, then check the funds before continuing

istae: BUT, if you want to warn the user through the logs that they don't have sufficient funds before even playing, then this is also good.

pkg/api/redistribution.go
```diff
@@ -13,6 +13,7 @@ import (
 )
 
 type nodeStatusResponse struct {
+	HasFunds        bool           `json:"hasFunds"`
```
istae, line 16: maybe "hasSufficientFunds" ?

istae (approved):

pkg/storageincentives/agent.go
```diff
@@ -472,3 +493,23 @@ func (a *Agent) wrapCommit(storageRadius uint8, sample []byte, key []byte) ([]by
...
+		return false, err
+	}
+
+	price, err := a.backend.SuggestGasPrice(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	avgGasPrice := big.NewInt(avgTxGas)
+	avgGasPrice = avgGasPrice.Mul(avgGasPrice, price)
```
aloknerurkar, line 509: So I am not clear on this. Why do we multiply `avgGasPrice * price`? I was thinking this should just be the `price * minTxCountToCover`

istae: the code here does need some rewording, but the calculation is correct.
the miminum balance  = currentPrice * avgGas * nTransactionToCover

```
avgTxFee := new(big.Int).Mul(big.NewInt(avgTxGas), price)
minBalance :=  avgTxFee.Mul(avgTxFee, big.NewInt(minTxCountToCover))
```

pkg/api/redistribution.go
```diff
@@ -41,16 +42,25 @@ func (s *Service) redistributionStatusHandler(w http.ResponseWriter, r *http.Req
 		return
 	}
 
+	hasSufficientFunds, err := s.redistributionAgent.HasEnoughFundsToPlay(r.Context())
+	if err != nil {
+		logger.Debug("has enough funds to play", "overlay address", s.overlay.String(), "error", err)
```
mrekucci, line 47: The keys should be in snake-case style: `overlay address` -> `overlay_address`.

pkg/storageincentives/agent.go
```diff
@@ -35,11 +35,19 @@ const loggerName = "storageincentives"
 const (
 	DefaultBlocksPerRound = 152
 	DefaultBlocksPerPhase = DefaultBlocksPerRound / 4
+
+	// min # of transactions our wallet should be able to cover
+	minTxCountToCover = 5
+
+	// average tx gas used by transactions issued from agent
```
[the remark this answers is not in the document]
istae: the unit is gas ?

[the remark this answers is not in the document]
mrekucci: Isn't the "gas" prices denoted in gwei?

mrekucci (approved):

PR #3833, opened 2023-02-24
feat(api): use new storagev2 interfaces

aloknerurkar:
### Checklist

### Description
- API package to use the new storer interface.
- Mock storer implementation for the tests
- Tests for Mock storer and adjusting original tests
- Change in openapi spec

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3833)


pkg/localstorev2/mock/mockstorer.go
```diff
@@ -0,0 +1,211 @@
+// Copyright 2023 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mockstorer
```
istae, line 5: why do we need a mock storer?

aloknerurkar: Usually, for unit tests we mock the objects that the package uses and are not defined in the package. The storer should be tested separately. Having API pkg unit tests depend on the actual storer implementation correctness is not right. There could be bugs in the implementation etc. The goal of the unit tests here is just to ensure the code in the api pkg files is correct.

The type of tests you are mentioning are integration tests. We can have these separately. Or we can develop beekeeper tests which will do the integration tests. We will probably have to change a bunch of these anyway, so we can even add some. Along with this we need to provide debug APIs for showing storer state to verify in the tests.

istae: > Having API pkg unit tests depend on the actual storer implementation correctness is not right. There could be bugs in the implementation etc

I don't understand this. We are not mocking here something complex like kademlia that's needs external connections.
We are mocking something that can be created as easily as `storer.New(ctx context.Context, "", opts)`

But I get it, we don't really want to spend the energy rewriting already existing tests.

pkg/localstorev2/storer.go
```diff
@@ -58,9 +58,11 @@ type UploadStore interface {
 	// users requests to pin the data, a new pinning collection is created.
 	Upload(ctx context.Context, pin bool, tagID uint64) (PutterSession, error)
 	// NewSession can be used to obtain a tag ID to use for a new Upload session.
-	NewSession() (uint64, error)
+	NewSession() (SessionInfo, error)
 	// GetSessionInfo will show the information about the session.
 	GetSessionInfo(tagID uint64) (SessionInfo, error)
```
mrekucci, line 63: `GetSessionInfo` -> `Session`.
`DeleteSessionInfo` -> `DeleteSession`.
Also, some comments on those methods would be nice.

pkg/storagev2/chunkstore.go
```diff
@@ -71,13 +77,17 @@ type ChunkStore interface {
...
+	Hasser
 
-	// Has checks whether a chunk exists in the store.
-	Has(context.Context, swarm.Address) (bool, error)
 	// Iterate over chunks in no particular order.
 	Iterate(context.Context, IterateChunkFn) error
 }
 
+type ReadOnlyChunkStore interface {
+	Getter
```
mrekucci, line 87: What about `io.Closer`?

pkg/api/api.go
```diff
@@ -353,25 +348,18 @@ func (s *Service) Close() error {
 
 // getOrCreateTag attempts to get the tag if an id is supplied, and returns an error if it does not exist.
```
mrekucci, line 349: The method name doesn't match the comment.

pkg/api/api.go
```diff
@@ -778,183 +728,98 @@ func equalASCIIFold(s, t string) bool {
...
-		return nil, noopWaitFn, fmt.Errorf("request deferred: %w", err)
+		return err
 	}
+	return p.PutterSession.Put(ctx, chunk.WithStamp(stamp))
+}
 
-	if !deferred && s.beeMode == DevMode {
-		return nil, noopWaitFn, errUnsupportedDevNodeOperation
+func (p *putterSessionWrapper) Done(ref swarm.Address) error {
+	return multierror.Append(p.PutterSession.Done(ref), p.save()).ErrorOrNil()
```
mrekucci, line 753: Since we are using Golang 1.20, I recommend using the `errors.Join` instead of the `multierror` package.
The same applies to the other places where the `multierror` is used.

pkg/api/api.go
```diff
@@ -353,25 +348,18 @@ func (s *Service) Close() error {
 
 // getOrCreateTag attempts to get the tag if an id is supplied, and returns an error if it does not exist.
 // If no id is supplied, it will attempt to create a new tag with a generated name and return it.
-func (s *Service) getOrCreateTag(tagUid string) (*tags.Tag, bool, error) {
+func (s *Service) getOrCreateSessionID(tagUid uint64) (uint64, error) {
```
mrekucci, line 351: You can register this as a hook in `preMapHooks` similar to the `resolveNameOrAddress` method. That way you'll just use the tag on the struct.

pkg/api/api.go
```diff
@@ -778,183 +728,98 @@ func equalASCIIFold(s, t string) bool {
...
-		if err != nil {
-			return nil, err
-		}
-		if has || swarm.ContainsChunkWithAddress(chs[:i], c.Address()) {
-			exists[i] = true
-			continue
-		}
-		stamp, err := p.stamper.Stamp(c.Address())
+func (r *cleanupOnErrWriter) WriteHeader(statusCode int) {
+	if statusCode > 400 {
```
mrekucci, line 816: `400` -> `http.StatusBadRequest`

pkg/api/pin.go
```diff
@@ -38,14 +38,45 @@ func (s *Service) pinRootHash(w http.ResponseWriter, r *http.Request) {
...
+			err = putter.Put(r.Context(), chunk)
+			if err != nil {
+				return err
+			}
+			return nil
+		},
+	)
+	if err != nil {
+		err = multierror.Append(err, putter.Cleanup())
+		logger.Debug("pin collection failed", "error", multierror.Append(err, putter.Cleanup()))
```
mrekucci, line 69: Not sure why is the `putter.Cleanup()` called the second time here? Use `errors.Join` instead.

pkg/api/router.go
```diff
@@ -210,9 +210,9 @@ func (s *Service) mountAPI() {
 	))
 
 	handle("/chunks/{address}", jsonhttp.MethodHandler{
-		"GET":    http.HandlerFunc(s.chunkGetHandler),
-		"HEAD":   http.HandlerFunc(s.hasChunkHandler),
-		"DELETE": http.HandlerFunc(s.removeChunk),
+		"GET":  http.HandlerFunc(s.chunkGetHandler),
+		"HEAD": http.HandlerFunc(s.hasChunkHandler),
+		// "DELETE": http.HandlerFunc(s.removeChunk),
```
mrekucci, line 215: The dead code should be removed, the same is below.

pkg/api/soc.go
```diff
@@ -42,133 +42,122 @@ func (s *Service) socUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+			jsonhttp.NotFound(w, "batch with id not found")
+		case errors.Is(err, errInvalidPostageBatch):
+			jsonhttp.BadRequest(w, "invalid batch id")
+		default:
+			jsonhttp.BadRequest(w, nil)
+		}
+		return
+	}
+
+	ow := &cleanupOnErrWriter{ResponseWriter: w, onErr: putter.Cleanup}
```
mrekucci, line 89: The logger is not injected. The same applies wherever `cleanupOnErrWriter` is used.

pkg/steward/steward.go
```diff
@@ -55,37 +52,31 @@ func New(getter storage.Getter, t traversal.Traverser, r retrieval.Interface, p
...
+	getter := s.netStore.Download(false)
+
 	fn := func(addr swarm.Address) error {
-		c, err := s.getter.Get(ctx, storage.ModeGetSync, addr)
+		c, err := getter.Get(ctx, addr)
+		if err != nil {
+			return err
+		}
+
+		err = uploaderSession.Put(ctx, c)
```
mrekucci, line 64: No need to check the error, the following will do: `return uploaderSession.Put(ctx, c)`.

pkg/api/api.go
aloknerurkar: So the problem is the usage is different in different APIs. For eg we create a tag only when deferred upload/pin flag is set in header. Can this be handled in premaphook?

pkg/api/api.go
aloknerurkar: So for the validation package there is some formatting used. I dont want to add more changes here, so maybe we can do this in a separate change.

mrekucci (approved):

pkg/api/api.go
```diff
@@ -778,183 +728,99 @@ func equalASCIIFold(s, t string) bool {
...
-			return nil, err
-		}
-		if has || swarm.ContainsChunkWithAddress(chs[:i], c.Address()) {
-			exists[i] = true
-			continue
-		}
-		stamp, err := p.stamper.Stamp(c.Address())
+func (r *cleanupOnErrWriter) WriteHeader(statusCode int) {
+	// if there is an error status returned, cleanup.
+	if statusCode > http.StatusBadRequest {
```
janos, line 817: Shouldn't the check include 400 with `>=`?

pkg/api/api_test.go
```diff
@@ -143,6 +137,7 @@ func newTestServer(t *testing.T, o testServerOptions) (*http.Client, *websocket.
 	signer := crypto.NewDefaultSigner(pk)
 
 	if o.Logger == nil {
+		// o.Logger = log.NewLogger("test", log.WithSink(os.Stdout))
```
janos, line 140: Remove commented code.

pkg/api/bytes.go
```diff
@@ -103,39 +106,23 @@ func (s *Service) bytesUploadHandler(w http.ResponseWriter, r *http.Request) {
...
-			logger.Debug("done split failed", "error", err)
-			logger.Error(nil, "done split failed")
-			jsonhttp.InternalServerError(w, "done split filed")
-			return
-		}
+	err = putter.Done(address)
+	if err != nil {
+		logger.Debug("done split failed", "error", err)
+		logger.Error(nil, "done split failed")
+		jsonhttp.InternalServerError(ow, "done split filed")
```
janos, line 120: Missing return after internal server error call.

janos (approved):

PR #3854, opened 2023-03-01
fix: tx chunk store put and delete operation correctness

### LINKED ISSUE #3783: Refactor disk txChunkStore

mrekucci:
### Checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3854)

Closes #3783


pkg/localstorev2/internal/chunkstore/transaction.go
```diff
@@ -0,0 +1,163 @@
...
+
+	var err *multierror.Error
+	for _, loc := range cs.txSharky.writtenLocs {
+		err = multierror.Append(err, cs.txSharky.Sharky.Release(context.Background(), loc))
+	}
+	return err.ErrorOrNil()
+}
+
+func (cs *txChunkStoreWrapper) NewTx(state *storage.TxState) storage.TxChunkStore {
+	txStore := cs.txStore.NewTx(state)
```
aloknerurkar, line 128: I thought we decided to remove this transaction?

pkg/localstorev2/internal/chunkstore/transaction.go
```diff
@@ -0,0 +1,163 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (cs *txChunkStoreWrapper) Rollback() error {
+	if err := cs.txStore.Rollback(); err != nil {
```
aloknerurkar, line 116: What is happening to the `revOps`?

aloknerurkar (changes requested):

pkg/localstorev2/internal/chunkstore/transaction.go
mrekucci: Fixed.

pkg/localstorev2/internal/chunkstore/transaction.go
mrekucci: Fixed.

pkg/localstorev2/internal/chunkstore/transaction.go
```diff
@@ -0,0 +1,193 @@
...
+	var (
+		sum = sha256.Sum256(buf)
+		loc sharky.Location
+		err error
+	)
+
+	t.opsMu.Lock()
+	defer t.opsMu.Unlock()
+
+	loc, ok := t.toReleaseLocs[sum]
```
aloknerurkar, line 41: So I cannot think of a better way at the moment. But lets add some comments here on why this is being done.

aloknerurkar (approved):

pkg/localstorev2/internal/chunkstore/transaction.go
mrekucci: Done.

PR #3856, opened 2023-03-01
feat: push subscriber

notanatol:
### Checklist

### Description
Push subscriber

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3856)


pkg/localstorev2/subscribe_push.go
```diff
@@ -0,0 +1,104 @@
...
+const subscribePushEventKey = "subscribe-push"
+
+func (db *DB) SubscribePush(ctx context.Context) (chunks chan swarm.Chunk, stop func()) {
+	chunks = make(chan swarm.Chunk)
+
+	trigger, unsub := db.events.Subscribe(subscribePushEventKey)
+	defer unsub()
+
+	// send signal for the initial iteration
+	db.events.Trigger(subscribePushEventKey)
```
istae, line 24: no longer needed

istae (approved):

pkg/localstorev2/uploadstore.go
```diff
@@ -50,7 +50,7 @@ func (db *DB) Upload(ctx context.Context, pin bool, tagID uint64) (PutterSession
 		}),
 		done: func(address swarm.Address) error {
 			defer db.clearDirty(tagID)
-
+			defer db.events.Trigger(subscribePushEventKey)
```
istae, line 53: there is a nuance here and that this defer should be above the clearDirty()

pkg/localstorev2/subscribe_push.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+const subscribePushEventKey = "subscribe-push"
+
+func (db *DB) SubscribePush(ctx context.Context) (chunks chan swarm.Chunk, stop func()) {
+	chunks = make(chan swarm.Chunk)
+
+	trigger, unsub := db.events.Subscribe(subscribePushEventKey)
+	defer unsub()
```
istae, line 21: the unsub should be at the start of the goroutine above

pkg/localstorev2/subscribe_push.go
```diff
@@ -0,0 +1,104 @@
...
+	"context"
+	"sync"
+
+	"github.com/ethersphere/bee/pkg/localstorev2/internal/upload"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+const subscribePushEventKey = "subscribe-push"
+
+func (db *DB) SubscribePush(ctx context.Context) (chunks chan swarm.Chunk, stop func()) {
```
aloknerurkar, line 17: We should keep the signature of the function the same. Pusher currently maintains a skipList for chunks. This prevents us from repeating same chunks over and over again. So we need to check if the chunk returned from the iteration is in the skipList and if yes we should move on to next chunk.

Also, we will need a repeat-type function here. This is because if we fail to push a chunk we need to start over again. Maybe this can be avoided, but then this would need changes in Pusher. It would be better if we can handle it here.

pkg/localstorev2/uploadstore.go
aloknerurkar: I think its better to clear first as if this trigger somehow starts an iteration before the clearDirty happens the iteration will stop due to the dirtyTag and no new trigger would come.

pkg/pusher/pusher.go
```diff
@@ -112,7 +112,7 @@ func (s *Service) chunksWorker(warmupTime time.Duration, tracer *tracing.Tracer)
 
 	// inflight.set handles the backpressure for the maximum amount of inflight chunks
 	// and duplicate handling.
-	chunks, repeat, unsubscribe := s.storer.SubscribePush(ctx, s.inflight.set)
+	chunks, _, unsubscribe := s.storer.SubscribePush(ctx, s.inflight.set)
```
aloknerurkar, line 115: Is this change required?

pkg/localstorev2/subscribe_push.go
istae: we decided to remove this repeat function to reduce the avoidable complexity. 

instead, on error, we feed chunk back into the pusher worker as such:
https://github.com/ethersphere/bee/pull/3856/files#diff-a01b070dce6187e6c1adf4a7aa78408ad5bbd8e16d3fe741edde8a190a128f03R163

pkg/pusher/pusher.go
notanatol: maybe we can discuss this further, but we considered a retry mechanism where instead of resetting the iterator we push the chunk again into the "feeding" channel `line:162`

pkg/localstorev2/subscribe_push.go
notanatol: removed

pkg/localstorev2/subscribe_push.go
notanatol: moved

PR #3877, opened 2023-03-10
feat: add metrics for the new storage

mrekucci:
### Checklist

### Description
Add metrics for the new storage.

Part of #3749

### DIFF
--- a/pkg/storagev2/metrics.go
+++ b/pkg/storagev2/metrics.go
@@ -1,0 +1,335 @@
+// Copyright 2023 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package storage
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	m "github.com/ethersphere/bee/pkg/metrics"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// metrics groups repository related prometheus counters.
+type metrics struct {
+	TxTotalDuration prometheus.Histogram
+
+	IndexStoreCalls         prometheus.CounterVec
+	IndexStoreCallsDuration prometheus.HistogramVec
+
+	ChunkStoreCalls         prometheus.CounterVec
+	ChunkStoreCallsDuration prometheus.HistogramVec
+}
+
+// newMetrics is a convenient constructor for creating new metrics.
+func newMetrics() metrics {
+	const subsystem = "storage"
+
+	return metrics{
+		TxTotalDuration: prometheus.NewHistogram(
+			prometheus.HistogramOpts{
+				Namespace: m.Namespace,
+				Subsystem: subsystem,
+				Name:      "tx_total_duration",
+				Help:      "Total duration of transaction.",
+			},
+		),
+		IndexStoreCalls: *prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: m.Namespace,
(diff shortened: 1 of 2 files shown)


istae (approved):

pkg/storagev2/metrics.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+	return size, err
+}
+
+// Iterate implements the TxStore interface.
+func (m txIndexStoreWithMetrics) Iterate(query Query, fn IterateFn) error {
+	now := time.Now()
+	err := m.TxStore.Iterate(query, fn)
+	m.metrics.IndexStoreCallsDuration.WithLabelValues("Iterate").Observe(float64(time.Since(now).Nanoseconds()))
+	if err == nil || errors.Is(err, ErrNotFound) {
```
aloknerurkar, line 162: I think if we keep adding cases here for errors we will complicate this. I would prefer if we check for errors blindly at this level. We dont really know the context of the error so we should not infer stuff from it.

Same comment for all the methods.

aloknerurkar (changes requested):
I think these are nice to have, but not sure how useful they will be in debugging issues. They would help to find issues on the storage level I think. The problem is we could potentially have multiple indexstore and chunkstore calls as a single operation and so these metrics might not give the whole picture. So, we should have metrics on storer side when we create the sessions/putters etc. This way we can get metrics of how much time it takes to put to storage from the API side etc. We could have these as separate metrics in storer package.

I feel we can still have these metrics and put them behind some config flag from options. If the flag is set we construct the TxStores using the metrics wrapper. WDYT?

pkg/storagev2/metrics.go
```diff
@@ -0,0 +1,320 @@
...
+// txChunkStoreWithMetrics wraps TxChunkStore and adds metrics.
+type txChunkStoreWithMetrics struct {
+	TxChunkStore
+
+	metrics metrics
+}
+
+// Commit implements the TxChunkStore interface.
+func (m txChunkStoreWithMetrics) Commit() error {
+	now := time.Now()
```
aloknerurkar, line 220: We can also have a metric that will calculate the total time since the inception of this transaction. We can have a New method that creates this initial timestamp and then check it in Commit and Rollback to report the TTL of the transaction as a whole. This could be useful information.

mrekucci: Done.

pkg/storagev2/metrics.go
mrekucci: The `ErrNotFound` is a sentinel, so instead of returning another parameter, a boolean flag indicating that the record was/wasn't found we use this sentinel error. So counting it as an error would be a false positive.

mrekucci:
> I think these are nice to have, but not sure how useful they will be in debugging issues. They would help to find issues on the storage level I think. The problem is we could potentially have multiple indexstore and chunkstore calls as a single operation and so these metrics might not give the whole picture. So, we should have metrics on storer side when we create the sessions/putters etc. This way we can get metrics of how much time it takes to put to storage from the API side etc. We could have these as separate metrics in storer package.
> 
> I feel we can still have these metrics and put them behind some config flag from options. If the flag is set we construct the TxStores using the metrics wrapper. WDYT?

Yes, there will be an additional set of metrics for each component in the new local store package (when completed). These metrics focus on the performance/monitoring of the storage abstraction, this can be useful when we decide to change the underlying storage. I don't see any advantage in having them behind a flag, so they can be enabled/disabled. Perhaps you could elaborate on this?

aloknerurkar (approved):

PR #3882, opened 2023-03-14
feat: add localstore components metrics

### LINKED ISSUE #3749: Metrics for the new localstore

mrekucci:
### Checklist

### Description
Add metrics for the new localstore components.

Closes #3749

### DIFF
--- a/pkg/localstorev2/cachestore.go
+++ b/pkg/localstorev2/cachestore.go
@@ -15,31 +15,39 @@ import (
 
 // Lookup is the implementation of the CacheStore.Lookup method.
 func (db *DB) Lookup() storage.Getter {
-	return storage.GetterFunc(func(ctx context.Context, address swarm.Address) (swarm.Chunk, error) {
-		txnRepo, commit, rollback := db.repo.NewTx(ctx)
-		ch, err := db.cacheObj.Getter(txnRepo).Get(ctx, address)
-		switch {
-		case err == nil:
-			return ch, commit()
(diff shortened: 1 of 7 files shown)


notanatol (approved):

pkg/localstorev2/internal/reserve/reserve.go
```diff
@@ -93,7 +99,15 @@ func New(
 }
 
 // Put stores a new chunk in the reserve and returns if the reserve size should increase.
-func (r *Reserve) Put(ctx context.Context, store internal.Storage, chunk swarm.Chunk) (bool, error) {
+func (r *Reserve) Put(ctx context.Context, store internal.Storage, chunk swarm.Chunk) (inc bool, err error) {
+	defer func(start time.Time) {
+		r.metrics.MethodCallsDuration.WithLabelValues("Put").Observe(float64(time.Since(start).Nanoseconds()))
```
istae, line 104: nitpick: I would recommend using `time.Since(start).Seconds()` as it outputs float64 too, it's easier to  work with when building dashboards

aloknerurkar (commented):
I think there is a better way. We dont have to change every method in all the internal functions. Having the metrics defined in internal is also not correct as they are exported right.

Instead we should have wrapper Putter/Getter/PutterSessions which will add to the correct metric based on user configurations. Lets have a call on this before we do anything here.

aloknerurkar (approved):

PR #3894, opened 2023-03-16
fix: multiple issues in TimeoutReader

vladopajic:
### Checklist

- [x] I have read the [coding guide](https://github.com/ethersphere/bee/blob/master/CODING.md).
- [ ] My change requires a documentation update, and I have done it.
- [ ] I have added tests to cover my changes.
- [ ] I have filled out the description and linked the related issues.

PR #3895, opened 2023-03-16
fix: timeout reader cancel func call on normal read

mrekucci:
### Checklist

### Description
Fixes the case where the timeout reader `cancel` function is called after the normal completion of a read.

### DIFF
--- a/pkg/util/ioutil/ioutil.go
+++ b/pkg/util/ioutil/ioutil.go
@@ -6,6 +6,7 @@ package ioutil
 
 import (
 	"context"
+	"errors"
 	"io"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


notanatol (approved):

vladopajic:
there is PR https://github.com/ethersphere/bee/pull/3894 which also fixes this issue

aloknerurkar (approved):

mrekucci:
> there is PR #3894 which also fixes this issue

PR #3894 removes context, which is undesirable.

vladopajic:
> > there is PR #3894 which also fixes this issue
> 
> PR #3894 removes context, which is undesirable.

Could you point to conclusive case were implementation without context is going to cause any issues?

PR #3902, opened 2023-03-20
feat(retrieval, node): use new storagev2 interfaces

aloknerurkar:
### Checklist

### Description
This PR fixes the remaining things required for build and test to run on the integration branch.

1. Fix retrieval pkg to use the new localstore. By the current design, retrieval no longer cares about stamps. This will cause a protocol bump.
2. Fix node pkg init for node and devnode and bootstrap functionality.
3. Fix storageincentives agent related code.
4. Move `pkg/localstore` to `pkg/_localstore` to prevent tests from running this directory. Once the import logic is implemented only the necessary amount of code should be copied from this folder to some internal pkg and then this should be deleted.
5. Disable the db related cmds for now. These need to be reimplemented/re-thought.


pkg/localstorev2/reserve.go
```diff
@@ -92,6 +92,8 @@ func (db *DB) IsWithinStorageRadius(addr swarm.Address) bool {
 	return swarm.Proximity(addr.Bytes(), db.baseAddr.Bytes()) >= db.reserve.Radius()
 }
 
+func (db *DB) IsFullySynced() bool { return false }
```
istae, line 95: was this left out temporarily on purpose?
this function should check that the reserve size is above the threshold and the chunk rate is zero

aloknerurkar: Yes I havent implemented it. I just added it to satisfy the interfaces.

istae: What I had in mind for this was to have a simple function created in node.go that would be passed to the storage incentives agent that would check that the puller rate is zero and the reserve size is above the threshold, this way, the localstore does not have to implement the `IsFullySynced`

pkg/api/postage.go
```diff
@@ -233,15 +233,15 @@ func (s *Service) postageGetAllStampsHandler(w http.ResponseWriter, _ *http.Requ
...
-			StorageRadius: b.StorageRadius,
-			BatchTTL:      batchTTL,
+			BatchID:     b.ID,
+			Value:       bigint.Wrap(b.Value),
+			Start:       b.Start,
+			Owner:       b.Owner,
+			Depth:       b.Depth,
+			BucketDepth: b.BucketDepth,
+			Immutable:   b.Immutable,
+			// StorageRadius: b.StorageRadius,
```
aloknerurkar, line 243: @istae Should we remove this field from the API response? I think this is no longer tracked in the batchstore right?

istae: yes, I believe so, we use a common storage radius everywhere now

pkg/localstorev2/reserve.go
aloknerurkar: Do you want me to do this in the current PR or should we do it in a separate one?

pkg/localstorev2/storer.go
```diff
@@ -449,6 +461,34 @@ func (db *DB) Close() error {
 	return err
 }
 
+func (db *DB) SetRetrievalService(r retrieval.Interface) {
+	db.retrieval = r
```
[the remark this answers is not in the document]
aloknerurkar: This is used in the Download.Getter. This is basically the netstore functionality.

pkg/localstorev2/storer.go
```diff
@@ -449,6 +461,34 @@ func (db *DB) Close() error {
...
+}
+
+func (db *DB) SetSyncer(s SyncReporter) {
+	db.lock.Lock(lockKeySetSyncer)
```
istae, line 469: why do we need a look for this? this should be a one time operation.

cmd/bee/cmd/db.go
```diff
@@ -2,6 +2,9 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+//go:build oldlocalstore
```
mrekucci, line 5: Some comment would be nice to have about this build tag.

pkg/localstorev2/storer.go
mrekucci: Maybe use `sync.Once`.

pkg/node/node.go
```diff
@@ -1082,7 +1113,7 @@ func NewBee(ctx context.Context, addr string, publicKey *ecdsa.PublicKey, signer
 		debugService.MustRegisterMetrics(p2ps.Metrics()...)
 		debugService.MustRegisterMetrics(pingPong.Metrics()...)
 		debugService.MustRegisterMetrics(acc.Metrics()...)
-		debugService.MustRegisterMetrics(storer.Metrics()...)
+		// debugService.MustRegisterMetrics(localStore.Metrics()...)
```
mrekucci, line 1116: Why is this commented out? The new localstore metrics should also be registered.

cmd/bee/cmd/cmd.go
```diff
@@ -153,7 +153,7 @@ func newCommand(opts ...option) (c *command, err error) {
 	}
 
 	c.initVersionCmd()
-	c.initDBCmd()
+	// c.initDBCmd()
```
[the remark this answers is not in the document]
aloknerurkar: So I dont think we should completely remove this command. Problem is right now it uses the old localstore code. In order for the binary to build I have added the build tag and commented this line.

We need to see what commands to support going ahead. Also we need to see how the export/import should be done for this command going ahead. This is why I have commented it as it should be done before we merge. WDYT?

cmd/bee/cmd/db.go
aloknerurkar: As I mentioned, there is no information here as this is temporary, just so that we can build the binary.

pkg/localstorev2/reserve.go
aloknerurkar: So for now I have kept this in the storer as we already have the syncer and size values here. So this function is a one-liner either way. We can move it later if you still feel strongly about this.

cmd/bee/cmd/cmd.go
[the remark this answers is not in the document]
mrekucci: Maybe add some TODO description there.

mrekucci (approved):

PR #3903, opened 2023-03-20
refactor: improve underlying goroutine termination of timeout reader

mrekucci:
### Checklist

### Description
Improve timeout reader:
- don't terminate the underlying goroutine when zero bytes are read
- terminate underlying goroutine on any error encounter

### DIFF
--- a/pkg/util/ioutil/ioutil.go
+++ b/pkg/util/ioutil/ioutil.go
@@ -6,7 +6,6 @@ package ioutil
 
 import (
 	"context"
-	"errors"
 	"io"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


mrekucci:
> * goroutine in `TimeoutReader` function should be updated accordingly

Can you be more specific?

> * switch statement in `Read` seems complex, why have you decided to go along that way, compared to previous approach of just writing data back to timeout goroutine?

The `Stop` call has been moved to the main goroutine to avoid possible data-race. According to the [documentation](https://pkg.go.dev/time#Timer.Stop), it is also recommended to empty the timer after calling the `Stop` method.

pkg/util/ioutil/ioutil.go
```diff
@@ -25,7 +24,7 @@ func (tr *timeoutReader) Read(p []byte) (int, error) {
 	n, err := tr.r.Read(p)
 
 	switch {
-	case (n == 0 || errors.Is(err, io.EOF)) && tr.t.Stop():
+	case err != nil && tr.t.Stop():
```
istae, line 27: can we write this function without the `fallthrough`

mrekucci:
> > Can you be more specific?
> 
> goroutine created in `TimeoutReader` should also be updated with same fix, namely, switch and cases should be removed.

Not sure I follow, can you provide some code?

> > The `Stop` call has been moved to the main goroutine to stop the timer as soon as possible and to avoid possible data-race. According to the [documentation](https://pkg.go.dev/time#Timer.Stop), it is also recommended to empty the timer after calling the `Stop` method.
> 
> Have a look at this PR, it seem to have simpler fix for all problems in this utility #3894

PR #3894 looks overly complicated, without context the timeout reader can be simplified and just simply leverage `time.AfterFunc`.

pkg/util/ioutil/ioutil.go
mrekucci: Done.

istae (approved):

vladopajic (approved):

Issue #3876, opened 2023-03-10
improve retrieval metrics


istae:

istae:
the retrieval metrics can be improved by:
1. displaying error rate for retrievals
2. displaying time duration for retrieving chunks (both successful and failed attempts)
3. Avg # number of retries before the chunk is retrieved

istae:
new dashboards are needed as well in grafana
this is a requirement for closing this issue

PR #3904, opened 2023-03-20
feat: more retrieval metrics

vladopajic:
### Checklist

### Description
Adding metrics as part of  #3876.


istae:
there some existing metrics already, can we cleanup and use only the ones we need.

istae:
> > there some existing metrics already, can we cleanup and use only the ones we need.
> 
> how should we know which metrics are not needed?

When you are also fixing the dashboards, the decision is yours to keep and discard the metrics you need :)

istae (approved):

PR #3905, opened 2023-03-20
docs: added missing fields to openapi definition

NoahMaizels:
Added missing fields for debug api "phase" and "hasSufficientFunds" to openapi SwarmCommon.yaml file.


NoahMaizels:
> our linter doesn't like capitals in the commit message 🤣

I just changed the commit message so it's all lowercase, lemme know if that works, thanks!

vladopajic (approved):

istae (approved):

PR #3919, opened 2023-03-22
refactor: timeout reader

mrekucci:
### Checklist

### DIFF
--- a/pkg/util/ioutil/ioutil.go
+++ b/pkg/util/ioutil/ioutil.go
@@ -7,74 +7,78 @@ package ioutil
 import (
 	"context"
 	"io"
+	"sync/atomic"
 	"time"
 )
 
-// timeoutReader monitors the progress of the io.Reader Read method
-// and calls the cancel function when reading does not progress for
-// the specified time.
-type timeoutReader struct {
-	t *time.Timer
-	r io.Reader
-	n chan int
+// idleReader monitors the progress of the io.Reader Read method
+// and calls the callbackFn function when reading does not progress
+// for the specified idle duration.
+type idleReader struct {
+	total  atomic.Uint64
+	reader io.Reader
+
+	idle  time.Duration
+	timer *time.Timer
+
+	done   atomic.Bool
+	cancel context.CancelFunc
 }
 
 // Read implements the io.Reader Read interface.
-func (tr *timeoutReader) Read(p []byte) (int, error) {
-	n, err := tr.r.Read(p)
-
-	v := n
-	if err != nil && tr.t.Stop() {
+func (ir *idleReader) Read(p []byte) (int, error) {
+	if !ir.done.Load() && ir.timer.Stop() {
 		select {
-		case <-tr.t.C:
+		case <-ir.timer.C:
 		default:
 		}
-
-		// Negative value signals that error has
-		// occurred and the goroutine should terminate.
-		v = -1
 	}
 
-	select {
-	case tr.n <- v:
-	default:
-	}
+	n, err := ir.reader.Read(p)
 
+	if !ir.done.Load() {
+		ir.timer.Reset(ir.idle)
+		ir.total.Add(uint64(n))
+		if err != nil {
+			ir.cancel()
+			ir.done.Store(true)
+		}
+	}
 	return n, err
 }
 
-// TimeoutReader creates a new timeoutReader instance and starts
-// a goroutine that monitors the progress of reading from the given reader.
-// If no progress is made for the duration of the given timeout, then the
-// reader executes the given cancel function and terminates the underlying
-// goroutine. This goroutine will also terminate when the given context
-// is canceled or if an error is returned by the given reader.
-func TimeoutReader(ctx context.Context, r io.Reader, timeout time.Duration, cancel func(uint64)) io.Reader {
-	tr := &timeoutReader{t: time.NewTimer(timeout), r: r, n: make(chan int)}
+// IdleReader creates a new io.Reader instance and starts a goroutine that
+// monitors the progress of reading from the given reader. If no progress is
+// made for the duration of the given idle, then the reader executes the given
+// cancelFn function and terminates the underlying goroutine. This goroutine
+// will also terminate when the given context is canceled or if an error is
+// returned by the given reader. The duration of the Read method is not taken
+// into account when monitoring the progress of reading.
+func IdleReader(ctx context.Context, reader io.Reader, idle time.Duration, callbackFn func(uint64)) io.Reader {
+	ctx, cancel := context.WithCancel(ctx)
+	ir := &idleReader{
+		reader: reader,
+		timer:  time.NewTimer(idle),
+		idle:   idle,
+		cancel: cancel,
+	}
 
 	go func() {
-		var total uint64 = 0
+		select {
+		case <-ctx.Done():
+		case <-ir.timer.C:
+			callbackFn(ir.total.Load())
+		}
 
-		for {
-			select {
-			case n := <-tr.n:
-				switch {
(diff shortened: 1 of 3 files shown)


pkg/util/ioutil/ioutil.go
```diff
@@ -62,11 +72,13 @@ func TimeoutReader(ctx context.Context, r io.Reader, timeout time.Duration, canc
 				case n < 0:
```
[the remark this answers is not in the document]
mrekucci: The channel was removed.

notanatol (approved):

pkg/util/ioutil/ioutil.go
```diff
@@ -7,69 +7,70 @@ package ioutil
...
+			tr.done.Store(true)
+			tr.timer.Stop()
+			select {
+			case <-tr.timer.C:
+			default:
+			}
+		}()
 
-		for {
+		for !tr.done.Load() {
```
vladopajic, line 67: is condition in for loop nessesery? is for loop also nessesery? 

goroutine  will block on select statement and return once data is received on ether channel (there is no next iteration)

pkg/util/ioutil/ioutil.go
```diff
@@ -7,69 +7,70 @@ package ioutil
...
+	switch {
+	case err != nil && !tr.done.Load():
+		tr.cancel()
+		tr.done.Store(true)
+	case !tr.done.Load():
+		tr.total.Add(uint64(n))
+		tr.timer.Reset(tr.timeout)
 	}
-
 	return n, err
```
vladopajic, line 39: would it make sense to return non nil error when reader has timed out?

vladopajic: this case would be following:

1. read some data (but there are more data to be read)
2. reader is now blocked
3. reader timeouts
4 if at this point reading from reader is attempted should it return non nil err without actually attempting to read from underlying reader?

pkg/util/ioutil/ioutil.go
```diff
@@ -7,69 +7,70 @@ package ioutil
...
-	select {
-	case tr.n <- v:
-	default:
+	n, err := tr.reader.Read(p)
+	switch {
+	case err != nil && !tr.done.Load():
+		tr.cancel()
+		tr.done.Store(true)
+	case !tr.done.Load():
+		tr.total.Add(uint64(n))
```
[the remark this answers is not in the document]
mrekucci: So the `total` is relevant when the reader is timed-out since it passes the value to the callback.

pkg/util/ioutil/ioutil.go
vladopajic: Fair enough.

pkg/util/ioutil/ioutil.go
```diff
@@ -7,70 +7,68 @@ package ioutil
...
-		select {
-		case <-tr.t.C:
-		default:
+	n, err := tr.reader.Read(p)
+	if !tr.done.Load() && tr.timer.Reset(tr.timeout) {
+		tr.total.Add(uint64(n))
+		if err != nil {
+			tr.cancel()
+			tr.done.Store(true)
 		}
```
vladopajic, line 36: Would it make sense to avoid `done` variable? eg:

```
if tr.timer.Reset(tr.timeout) {
		tr.total.Add(uint64(n))
		if err != nil {
			tr.cancel()
			r.timer.Stop()
		}
}
// note: `timer.Stop()` would need to be called in goroutine as well in order to stop timer when context is cancled
```

pkg/util/ioutil/ioutil.go
```diff
@@ -7,70 +7,68 @@ package ioutil
...
-					total += uint64(n)
-				}
-			case <-tr.t.C:
-				cancel(total)
-				return
-			case <-ctx.Done():
-				return
+			case <-tr.timer.C:
+			default:
 			}
```
vladopajic, line 63: since we are leaving this goroutine at this point, could we just abandon any value that may be present in this channel without draining it?

mrekucci: Done.

pkg/util/ioutil/ioutil.go
```diff
@@ -7,70 +7,63 @@ package ioutil
...
-				cancel(total)
-				return
-			case <-ctx.Done():
-				return
-			}
+		select {
+		case <-ctx.Done():
+			return
+		case <-tr.timer.C:
+			cancelFn(tr.total.Load())
```
vladopajic, line 65: `cancelFn` might take some time to execute (hypothetically speaking), this could cause following error:

1. timer is fired
2. `cancelFn` is being now executed (this is goroutine A)
3.  reader continues to read data (this is goroutine B; assuming B is faster then A), it resets timer because there are more data to read
=> timer will continue ticking once again

This might not be big problem, but if would make sense to avoid it if possible. Could stopping/cleanup be done before calling cancel?

mrekucci: Point 3 is invalid because the `select` statement is not in the loop. In other words, if the timer is fired and the value is read, then reset, there will be no more goroutine waiting to read from it.

pkg/util/ioutil/ioutil.go
```diff
@@ -7,71 +7,61 @@ package ioutil
...
+			return
 		}
+		tr.done.Store(true)
```
istae, line 64: unreachable code

pkg/util/ioutil/ioutil.go
vladopajic: > then reset, there will be no more goroutine waiting to read from it.

Agree. What will happen, in this case, is that timer goroutine will keep alive for duration of timeout (because timer got restarted), but nobody will ever going to read from it (goroutine has returned). In other words, some resources associated to this timer will be unnecessarily kept.

pkg/util/ioutil/ioutil.go
mrekucci: Done.

pkg/util/ioutil/ioutil.go
```diff
@@ -7,71 +7,59 @@ package ioutil
...
-				return
-			case <-ctx.Done():
-				return
-			}
+		select {
+		case <-ctx.Done():
+		case <-tr.timer.C:
+			cancelFn(tr.total.Load())
 		}
+		tr.done.Store(true)
```
vladopajic, line 62: would it also make sense to stop timer here? because for normal reads, when reading is executed before timeout, it would make sense to stop timer in order to gracefully clean up resources.

mrekucci: Done.

istae (approved):

vladopajic (approved):

vladopajic (changes requested):

pkg/util/ioutil/ioutil.go
```diff
@@ -7,70 +7,70 @@ package ioutil
...
+	cancel context.CancelFunc
 }
 
 // Read implements the io.Reader Read interface.
 func (tr *timeoutReader) Read(p []byte) (int, error) {
-	n, err := tr.r.Read(p)
-
-	v := n
-	if err != nil && tr.t.Stop() {
+	if !tr.done.Load() && tr.timer.Stop() {
```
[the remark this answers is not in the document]
mrekucci: The purpose is not to block the reader but to fire the `cancelFn`.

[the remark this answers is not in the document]
vladopajic: @mrekucci ☝️, should this be an issue?

mrekucci:
Closed in favor of #3930

PR #3921, opened 2023-03-22
feat: retrieval sticky skiplist

### LINKED ISSUE #3910: sticky skiplists for retrieval

istae:
### Checklist

### Description
Similar to the pushsync protocol, the PR adds a new global sticky skiplist to the retrieval where, when an retrieval attempt fails, 
the chunk-peer pair is added to the skiplist so that on the next retry, disjoint paths to the storer node can be found.

The PR also extracts the closest peer selection out of the goroutine so that no two or more requests can try the same peer
(this is a strange edge case as seen in dev-bee-gateway pod 3 https://monitoring.mainnet.ethswarm.org/d/qRpzaWKnz/retrieval?orgId=1&from=1679479309812&to=1679487741989&var-namespace=dev-bee-gateway&var-instance=bee-3-0 )

overdraft skiplist is also removed in both the retrieval and pushsync.

Turned off preemptive retries for forwarders.
 
Closes #3910


pkg/pushsync/pushsync.go
```diff
@@ -42,24 +41,22 @@ const (
...
-	maxPeers          = 64
+	maxPushErrors     = 32
 )
 
 var (
 	ErrNoPush            = errors.New("could not push chunk")
 	ErrOutOfDepthStoring = errors.New("storing outside of the neighborhood")
 	ErrWarmup            = errors.New("node warmup time not complete")
-	errNotAttempted      = errors.New("peer not attempted")
+	errPrepareCredit     = errors.New("peer not attempted")
```
mrekucci, line 59: The error name is not inline with the error description.

pkg/pushsync/pushsync.go
```diff
@@ -332,118 +329,101 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
+	defer close(done)
 
-	timer := time.NewTimer(0)
-	defer timer.Stop()
+	retryC := make(chan struct{})
+
+	retry := func() {
+		go func() {
+			select {
+			case retryC <- struct{}{}:
```
mrekucci, line 359: Also, consider checking here the passed context.

pkg/skippeers/skippeers.go
```diff
@@ -0,0 +1,73 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
mrekucci, line 1: 2020 -> 2023

pkg/pushsync/pushsync.go
```diff
@@ -332,118 +329,101 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
 				}
 
-				if skipPeers.OverdraftListEmpty() { // no peers in skip list means we can be confident that we are the closest peer
-					// we don't act on ErrWantSelf unless there are no overdraft peers
-					if !ps.radiusChecker.IsWithinStorageRadius(ch.Address()) {
-						return swarm.ZeroAddress, false, ErrOutOfDepthStoring
-					}
-					ps.pushToNeighbourhood(ctx, fullSkipList, ch, origin, originAddr)
-					return swarm.ZeroAddress, false, err
+				if !ps.radiusChecker.IsWithinStorageRadius(ch.Address()) {
```
aloknerurkar, line 379: This changes the logic quite a bit. We have removed the whole overdraft logic in order to make this skiplist generic. The point of the overdraft peers was that you can retry again as they could potentially be usable. But now we will no longer consider them.

pkg/pushsync/pushsync.go
```diff
@@ -332,118 +329,101 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
+			inflight--
 
-			if errors.Is(result.err, errNotAttempted) {
-				logger.Debug("not attempted: adding overdraft peer to skiplist", "peer_address", result.peer)
-				skipPeers.AddOverdraft(result.peer)
-			}
+			ps.measurePushPeer(result.pushTime, result.err, origin)
 
-			if ps.warmedUp() && !errors.Is(result.err, errNotAttempted) {
+			if ps.warmedUp() && !errors.Is(result.err, errPrepareCredit) {
```
aloknerurkar, line 426: If we dont add these peers (errPrepareCredit) to the skiplist, the closest peer will keep returning the same peer first. This would cause us to retry prematurely. Previously we would have skipped and then cleared the overdraft list and only then consider this peer.

aloknerurkar (commented):
I am a little concerned about the pushsync changes. Not sure if we should make changes to the logic of a protocol in a patch release, also after RC.

pkg/skippeers/skippeers_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	skipList.Add(addr1, addr2, time.Millisecond*10)
+
+	if !skipList.ChunkPeers(addr1)[0].Equal(addr2) {
+		t.Fatal("peer should be skipped")
+	}
+
+	skipList.Add(addr1, addr3, time.Millisecond*10)
+
+	time.Sleep(time.Millisecond * 11)
```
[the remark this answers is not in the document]
istae: this time it's okay because the sleep here is not waiting for some worker or operation to terminate.
The Add function records the time at which the item should expire. The sleep is there to make sure the current time is past that marker, the test will not flake whatever sleep amount is put there.

pkg/pushsync/pushsync.go
istae: It was discussed a little more than a week back that after not finding a single case in mainnet of having to delay retries in the event of having overdrafted peers, supporting this extra logic is not needed, the retry mechanism is sufficient.

pkg/pushsync/pushsync.go
istae: these peers get added to the other skiplist already that is discarded after the request terminates.
The `ps.skiplist` is the global one.

vladopajic:
@istae there is [logs-local.txt](https://github.com/ethersphere/bee/pull/3921/files#diff-f3ab2bc767acf1c8948670a0d645843bd83b403276338ebea3d256ce74c4bd8e) file committed, it looks like unintentional change?

vladopajic (approved):

pkg/pushsync/pushsync.go
```diff
@@ -332,118 +329,99 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
+	retry()
 
 	// nextPeer attempts to lookup the next peer to push the chunk to, if there are overdrafted peers the boolean would signal a re-attempt
-	nextPeer := func() (peer swarm.Address, retry bool, err error) {
+	nextPeer := func() (peer swarm.Address, err error) {
 
-		fullSkipList := append(ps.skipList.ChunkSkipPeers(ch.Address()), skipPeers.All()...)
+		fullSkipList := append(skip, ps.skipList.ChunkPeers(ch.Address())...)
 
 		peer, err = ps.topologyDriver.ClosestPeer(ch.Address(), includeSelf, topology.Filter{Reachable: true}, fullSkipList...)
```
aloknerurkar, line 371: Should we remove this filter? Wdyt?

aloknerurkar (approved):

PR #3932, opened 2023-03-24
feat: add testing logger

vladopajic:
### Checklist

### Description
Tests in `bee` codebase are using `log.Noop` quite often. Since these logs could be potentially beneficial for debugging failed test new logger/builder, `NewTestLogger`, is added  specially design for drop in replacement for `log.Noop`. This logger will use `t.Log` as sink for all logging outputs.


mrekucci (commented):
I'd suggest replacing it with the following and putting it into the `testutil` package:
```Golang
// NewTestLogger returns a new log.Logger that uses t.Log method
// as the log sink. It is particularly useful for debugging tests.
func NewTestLogger(t testing.T) log.Logger {
	return log.NewLogger(t.Name(), log.WithSink(ioutil.WriterFunc(func(p []byte) (int, error) {
		t.Log(string(p))
		return len(p), nil
	})))
}
```

vladopajic:
> I'd suggest replacing it with the following and putting it into the `testutil` package:
> 
> ```go
> // NewTestLogger returns a new log.Logger that uses t.Log method
> // as the log sink. It is particularly useful for debugging tests.
> func NewTestLogger(t testing.T) log.Logger {
> 	return log.NewLogger(t.Name(), log.WithSink(ioutil.WriterFunc(func(p []byte) (int, error) {
> 		t.Log(string(p))
> 		return len(p), nil
> 	})))
> }
> ```

updated.

as a side note, more idiomatic approach to hosting utilities would be in packages for which utilities are designed for. 

`testutil` is a necessary evil because these utilities don't have better package were they could to be hosted. imo, it would be better if `bee` had `bytes` package were different types and utilities for working with bytes are hosted. in this case usage of `testutil.RandBytes(...)` would be `bytes.Rand(...)` which appears more neater. 

having this said I don't see problem with hosting `NewTestLogger` in `testutil`. it's better to have utilities then not. when time comes any of these utilities might find better place; and this kind of refactoring would be trivial.

notanatol (approved):

mrekucci (approved):

PR #3942, opened 2023-03-28
perf: delay out of depth syncing for some time

istae:
### Checklist

### Description
There are two groups of pullsyncing: neighbors and non-neighbors. For fresh nodes especially, each time we lower the storage radius, all previous intervals below the radius are reset, meaning syncing starts from scratch for those bins. For peers that is within the radius, no resyncing occurs because the reset radiuses are below the storage radius, but for the out of depth peers, syncing has to be done from scratch. 

To prevent too much resyscing, we delay the the start of the out of depth syncing until after 24 hours of running, which should be enough time for the storage radius to stabilize.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -958,7 +958,7 @@ func NewBee(ctx context.Context, addr string, publicKey *ecdsa.PublicKey, signer
 	)
 
 	if o.FullNodeMode && !o.BootnodeMode {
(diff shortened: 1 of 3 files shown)


mrekucci (commented):
It would be good if you could add a description and indicate in which situations this PR is useful.

aloknerurkar (approved):

vladopajic (approved):

PR #3943, opened 2023-03-28
feat(storageincentives): persisting agent data between rounds

vladopajic:
### Checklist

- [x] I have read the [coding guide](https://github.com/ethersphere/bee/blob/master/CODING.md).
- [ ] My change requires a documentation update, and I have done it.
- [x] I have added tests to cover my changes.
- [x] I have filled out the description and linked the related issues.

### Description
In this change storage incentives agent will persist data between round.

PR #3948, opened 2023-03-29
feat(storageincentives): purge agent data

vladopajic:
### Checklist

### Description
Follow up for PR https://github.com/ethersphere/bee/pull/3943

### DIFF
--- a/pkg/storageincentives/agent.go
+++ b/pkg/storageincentives/agent.go
@@ -15,19 +15,19 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/ethersphere/bee/pkg/settlement/swap/erc20"
-	"github.com/ethersphere/bee/pkg/storageincentives/staking"
-	"github.com/ethersphere/bee/pkg/transaction"
-
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethersphere/bee/pkg/crypto"
 	"github.com/ethersphere/bee/pkg/log"
 	"github.com/ethersphere/bee/pkg/postage"
 	"github.com/ethersphere/bee/pkg/postage/postagecontract"
+	"github.com/ethersphere/bee/pkg/settlement/swap/erc20"
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/storageincentives/redistribution"
+	"github.com/ethersphere/bee/pkg/storageincentives/staking"
 	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/ethersphere/bee/pkg/transaction"
+	"github.com/hashicorp/go-multierror"
 )
 
 const loggerName = "storageincentives"
@@ -41,6 +41,8 @@ const (
 
 	// average tx gas used by transactions issued from agent
 	avgTxGas = 250_000
+
+	purgeDataOlderThenXRounds = 10
 )
 
 type ChainBackend interface {
@@ -233,7 +235,10 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 		prevPhase    PhaseType = -1
 		currentPhase PhaseType
 		checkEvery   uint64 = 1
+		newRoundC           = make(chan uint64, 1)
 	)
+
+	a.startOutdatedDataPurger(newRoundC)
 
 	// optimization, we do not need to check the phase change at every new block
 	if blocksPerPhase > 10 {
@@ -253,7 +258,11 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 		}
 
 		round := block / blocksPerRound
-		currentRound.Store(round)
+		prevRound := currentRound.Swap(round)
+
+		if prevRound != round {
+			newRoundC <- round
+		}
 
 		a.metrics.Round.Set(float64(round))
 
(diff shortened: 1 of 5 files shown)


pkg/storageincentives/agent.go
```diff
@@ -302,6 +311,61 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 	}
 }
 
+func (a *Agent) startOutdatedDataPurger(onNewRoundC <-chan uint64) {
```
aloknerurkar, line 314: If we do this on every round, then we will mostly be deleting 1 entry per round. Having a separate goroutine for deleting single entry seems to be too much. Instead we can trigger this on the end of the Claim phase and not have this routine.

pkg/storageincentives/agent.go
```diff
@@ -41,6 +41,8 @@ const (
 
 	// average tx gas used by transactions issued from agent
 	avgTxGas = 250_000
+
+	purgeDataOlderThenXRounds = 10
```
mrekucci, line 45: Maybe `purgeStaleDataThreshold`?

pkg/storageincentives/agent.go
```diff
@@ -200,6 +202,9 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 
 		round := currentRound.Load()
 
+		// Start purging outdated data
```
mrekucci, line 205: This comment doesn't add much value.

pkg/storageincentives/agent.go
```diff
@@ -302,6 +307,48 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 	}
 }
 
+func purgeDataHandler(logger log.Logger, store storage.StateStorer, currentRound uint64) {
```
mrekucci, line 310: Maybe `purgeStaleDataHandler`?

pkg/storageincentives/agent.go
```diff
@@ -302,6 +307,48 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
...
+	if currentRound <= purgeDataOlderThenXRounds {
+		return
+	}
+
+	purgeRound := func(round uint64) error {
+		mErr := &multierror.Error{}
```
mrekucci, line 316: Use `errors.Join` instead of the `multierror` package.

aloknerurkar (approved):

pkg/storageincentives/agent.go
```diff
@@ -41,6 +40,8 @@ const (
 
 	// average tx gas used by transactions issued from agent
 	avgTxGas = 250_000
+
+	purgeStaleDataThreshold = 10
```
istae, line 44: why does this start at 10?

vladopajic: I was anxious going with threshold value which would clear all stale data asap, (value 1) because there was issue reported that API returns some values, related to round, with off by one error. Therefore value 10 was set because it makes enough room to accommodate any potential off by one error.

In future, when we are more confident, this could be easily tweaked to clear all stale data asap.

pkg/storageincentives/agent.go
```diff
@@ -302,6 +305,41 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
...
+
+	from, err := getLastPurgedRound(store)
+	if err != nil {
+		logger.Error(err, "failed getting last purged round")
+		return
+	}
+
+	to := currentRound - purgeStaleDataThreshold
+
+	for i := from; i < to; i++ {
```
istae, line 329: this `from` is problematic because it will start from zero and iterate until a large number

What I recommend is to remove the `getLastPurgedRound` and `saveLastPurgedRound` and tackle the problem this way:

for each commit, reveal, and sample entries, store the round as well, and make the keys sortable by round number (this may be the case already because you append the round number to the key)
and when it's time to purge, iterate on the each prefix (commit, reveal, and sample) and delete the entry if the round is within the threshold.

istae (changes requested):

pkg/storageincentives/agent.go
vladopajic: agree, that's more neater approach :+1: 

PR was updated, please review again

pkg/storageincentives/agent.go
```diff
@@ -302,6 +307,60 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
...
 
+func purgeStaleDataHandler(logger log.Logger, store storage.StateStorer, currentRound uint64) {
+	if currentRound <= purgeStaleDataThreshold {
+		return
+	}
+
+	thresholdRound := currentRound - purgeStaleDataThreshold
+	keysForRemoval := make([]string, 0)
+
+	stripRoundValue := func(key string) (uint64, error) {
```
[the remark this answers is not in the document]
vladopajic: I have realized the same when I was refactoring agent in later PRs. 

Since this logic for storing round data was already merged, and potentially used by some nodes I think we should merge this PR anyway - just to have complete round data logic in place. Later we could improve this, by storing round data via state, and then remove mechanism for storing round data manually (current approach).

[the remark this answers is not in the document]
istae: > potentially used by some nodes
 
 the release is not out yet so we are still in the green to make the fixes 👍

vladopajic:
Closing this PR based on above discussion.

PR #3950, opened 2023-03-29
fix: pullsync ignore overwrite error

istae:
### Checklist


aloknerurkar:
@mrekucci so in some places in pushsync/pusher code we only add 1 chunk at a time... So we did not need to handle session related stuff. Secondly we cannot include storer package in pusher/pushsync code as the pusher.Op is included in storer so there is a cyclic dependency. To fix this we would have to move Putter session into the storage interfaces. We can do that later if required.

mrekucci (approved):

PR #3953, opened 2023-03-30
feat: add service status protocol

mrekucci:
### Checklist

### Description

Closes ethersphere/bee-backlog#54


notanatol (approved):

openapi/SwarmDebug.yaml
```diff
@@ -1150,3 +1150,20 @@ paths:
           $ref: "SwarmCommon.yaml#/components/responses/400"
         default:
           description: Default response
+
+  "/status":
+    get:
+      summary: Get all available loggers that match the specified expression.
```
janos, line 1156: It looks to me that this description is not correct.

janos (approved):

openapi/SwarmDebug.yaml
mrekucci: Nice catch. Fixed!

istae:
This PR could not come any sooner 🙏

pkg/rate/rate.go
```diff
@@ -46,7 +46,7 @@ func (r *Rate) Add(count int) {
 // the rate is computed by first calculating how far along the current time is in the current window
 // as a ratio between 0 and 1.0. Then, the sum of currentWindowCount + (1 - ratio) * previousWindowCounter
 // is returned as the interpolated counter between the two windows.
-func (r *Rate) Rate() float64 {
+func (r *Rate) SyncRate() float64 {
```
istae, line 49: Why rename this? This rate pkg is not specific to chunks per seconds or sync rate.

istae (changes requested):

pkg/api/status.go
```diff
@@ -0,0 +1,35 @@
...
+func (s *Service) statusGetHandler(w http.ResponseWriter, r *http.Request) {
+	logger := s.logger.WithName("get_status").Build()
+
+	if s.beeMode == DevMode {
+		logger.Warning("status endpoint is disabled in dev mode")
+		jsonhttp.BadRequest(w, errUnsupportedDevNodeOperation)
+		return
+	}
+
+	snapshot := s.statusService.Snapshot()
```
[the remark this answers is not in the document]
aloknerurkar: What we can do is use Topology.EachNeighbour interface and send msgs to nodes connected in the neighbourhood and print the status for all of them along with itself.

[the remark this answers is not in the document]
istae: We need the status of all peers, not just neighbors I think. This would give us a bigger picture of what's going on with the network

pkg/rate/rate.go
mrekucci: This was accidentally included by the rename action when I did renaming of the `depthmonitor.SyncReporter`. Fixed!

pkg/api/status.go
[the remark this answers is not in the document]
mrekucci: Done.

pkg/rate/rate.go
istae: 🙏

pkg/status/status.go
```diff
@@ -0,0 +1,219 @@
...
+			defer cancel()
+
+			ss, err := svc.requestStatusSnapshot(ctx, address)
+			if err != nil {
+				svc.logger.Debug("cannot get status snapshot for peer", "peer_address", address, "error", err)
+				return false, true, nil
+			}
+
+			svc.mu.Lock()
+			svc.connectedPeersSnapshots = append(
```
istae, line 183: this forever appends the snapshots to the slice

istae (changes requested):
The snapshots are acquired by a worker that runs every 10 min, and get appends to a slice and I do not think this will work.

what we need is, on-demand, when the endpoint is hit, acquire the current status of all of the connected peers, sort the list by proximity order, and return the list.

pkg/status/status.go
mrekucci: This is irrelevant since the logic was changed to on-demand.

openapi/Swarm.yaml
```diff
@@ -487,7 +487,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "SwarmCommon.yaml#/components/schemas/Status"
+                $ref: "SwarmCommon.yaml#/components/schemas/HealthStatus"
```
aloknerurkar, line 490: This might cause changes on the infra side if they are using this endpoint. Instead I would call the new one BeeStatus or NodeStatus or something.

pkg/api/api.go
```diff
@@ -233,6 +243,7 @@ type ExtraOptions struct {
 	Steward          steward.Interface
 	SyncStatus       func() (bool, error)
 	IndexDebugger    StorageIndexDebugger
+	StatusService    *status.Service
```
aloknerurkar, line 240: StatusService is too generic I think. I would prefer we call it something like NodeStatus.

pkg/status/status.go
```diff
@@ -0,0 +1,189 @@
...
+		default:
+		}
+
+		ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+		defer cancel()
+
+		ss, err := s.requestStatusSnapshot(ctx, address)
+		if err != nil {
+			s.logger.Debug("cannot get status snapshot for peer", "peer_address", address, "error", err)
+			return false, true, nil
```
aloknerurkar, line 78: I am not sure if this is correct usage of `jumpToNext`. This is used to jump to next bin I think. Can you check? I think this should be false here.

istae: this would skip to the next bin, so we cannot return true here. if we cannot get the status, we can have a new error field in the response that would be filled on error

openapi/Swarm.yaml
mrekucci: The name is not returned by the endpoint so shouldn't be an issue.

pkg/status/status.go
mrekucci: Done.

pkg/api/api.go
mrekucci: Done.

PR #3954, opened 2023-03-30
fix(retrieval): dont retrieve zero value chunks

aloknerurkar:
### Checklist

### Description
We see a lot of retrieval requests with zero address which is causing a lot of failed activity on the mainnet. This PR checks for this and prevents unnecessary retrieval requests.

### DIFF
--- a/pkg/retrieval/retrieval.go
+++ b/pkg/retrieval/retrieval.go
@@ -388,7 +388,7 @@ func (s *Service) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) (e
 		return fmt.Errorf("zero address queried by peer %s", p.Address.String())
 	}
 
-	span, _, ctx := s.tracer.StartSpanFromContext(ctx, "handle-retrieve-chunk", s.logger, opentracing.Tag{Key: "address", Value: swarm.NewAddress(req.Addr).String()})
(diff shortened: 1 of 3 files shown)


pkg/retrieval/retrieval.go
```diff
@@ -443,3 +450,15 @@ func (s *Service) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) (e
 	}
 	return nil
 }
+
+func isZeroAddress(addr swarm.Address) bool {
+	if addr.IsZero() {
+		return true
+	}
+
+	if addr.Equal(swarm.NewAddress(make([]byte, swarm.HashSize))) {
```
mrekucci, line 459: So this will be allocation heavy. I'd recommend maybe attaching `IsEmpty` method to the `swarm.Address` similar to the `IsZero` method. Then you'll just do the following check: `address.IsZero || address.IsEmpty`.

pkg/swarm/swarm.go
```diff
@@ -35,6 +35,13 @@ var (
 	ErrInvalidChunk = errors.New("invalid chunk")
 )
 
+var (
+	// EmptyAddr is the address that is all zeroes.
+	EmptyAddr = NewAddress(make([]byte, HashSize))
```
mrekucci, line 40: `EmptyAddr` -> `EmptyAddress`.

mrekucci (approved):

PR #3958, opened 2023-03-31
chore(storageincentives): removed sample end phase/event

vladopajic:
### Checklist

### Description
Removing sample end phase/event because it appears to be unnecessary and introduces confusion because it is not valid phase.
Logic of old sample end event was added to sample handler where it suits more. 

Additional, minor change, was cleaning up commit handler. It now follows same handler convention of previous PR. Also previous phase type event check was removed because handler will be able to know if phase could be played or not based on stored information.

### DIFF
--- a/pkg/storageincentives/agent.go
+++ b/pkg/storageincentives/agent.go
@@ -132,25 +132,18 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 		}
 	}
 
-	// when the sample finishes, if we are in the commit phase, run commit
-	phaseEvents.On(sampleEnd, func(ctx context.Context, previous PhaseType) {
-		if previous == commit {
-			phaseEvents.Cancel(claim)
-			a.handleCommit(ctx, currentRound.Load())
-		}
-	})
-
 	// when we enter the commit phase, if the sample is already finished, run commit
-	phaseEvents.On(commit, func(ctx context.Context, previous PhaseType) {
-		if previous == sampleEnd {
-			phaseEvents.Cancel(claim)
-			a.handleCommit(ctx, currentRound.Load())
-		}
+	phaseEvents.On(commit, func(ctx context.Context, _ PhaseType) {
+		phaseEvents.Cancel(claim)
+
+		round := currentRound.Load()
+		isPhasePlayed, err := a.handleCommit(ctx, round)
+		printPhaseResult(commit, round, err, isPhasePlayed)
 	})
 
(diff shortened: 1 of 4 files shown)


pkg/storageincentives/agent.go
```diff
@@ -170,7 +163,19 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 		isPhasePlayed, err := a.handleSample(ctx, round)
 		printPhaseResult(sample, round, err, isPhasePlayed)
 
-		phaseEvents.Publish(sampleEnd)
```
istae, line 173: Participation in the round is decided in the claim phase, so naturally the sampler starts running in the claim phase, and is allowed to run until the end of the commit phase. 
So there are two cases we have to cover:
1. sampler finishes in the claim phase
2. sampler finished after the start of the commit phase

so if you remove this sampleEnd trigger, for cases that the sample finished after the start of the commit phase, the sample won't get submitted.

istae (changes requested):

pkg/storageincentives/agent.go
istae: if there are missing comments about this, we can utilize this PR to fix that

vladopajic: > so if you remove this sampleEnd trigger, for cases that the sample finished after the start of the commit phase, the sample won't get submitted.

please read few lines below from highlighted line, that code cover this concern?

istae: aha, I see.

istae: aha, I see now. What is this change required though?

vladopajic: it's easier to follow through logic

also it will make possible other simplifications in following PRs

pkg/storageincentives/agent.go
```diff
@@ -132,25 +132,18 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
...
+		printPhaseResult(commit, round, err, isPhasePlayed)
 	})
 
 	phaseEvents.On(reveal, func(ctx context.Context, _ PhaseType) {
 		// cancel previous executions of the commit and sample phases
-		phaseEvents.Cancel(commit, sample, sampleEnd)
+		phaseEvents.Cancel(commit, sample)
 
 		round := currentRound.Load()
 		isPhasePlayed, err := a.handleReveal(ctx, round)
```
istae, line 149: should be "isRoundPlayed"

istae (approved):

PR #3972, opened 2023-04-05
fix(retrieve, pushsync): wait for overdraft peers to refresh

istae:
### Checklist

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -403,7 +403,7 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
 
 			// no peers left
 			if errors.Is(err, topology.ErrNotFound) {
-				if skip.PruneExpiresAfter(overDraftRefresh) == 0 { //no peers would be removed, we have depleted ALL peers
+				if skip.PruneExpiresAfter(overDraftRefresh) == 0 { //no overdraft peers, we have depleted ALL peers
 					if inflight == 0 {
 						ps.logger.Debug("no peers left", "chunk_address", ch.Address(), "error", err)
 						return nil, err
@@ -427,7 +427,7 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
 			}
 
 			if err != nil {
-				if inflight == 0 {
+				if inflight > 0 {
 					continue
 				} else {
 					return nil, fmt.Errorf("get closest for address %s, allow upstream %v: %w", ch, origin, err)
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -41,10 +41,12 @@ const (
 )
 
 const (
-	defaultTTL         = 30 * time.Second // request time to live
-	preemptiveInterval = 5 * time.Second  // P90 request time to live
-	sanctionWait       = 5 * time.Minute
-	replicationTTL     = 5 * time.Second // time to live for neighborhood replication
+	defaultTTL                       = 30 * time.Second // request time to live
+	preemptiveInterval               = 5 * time.Second  // P90 request time to live
+	sanctionWait                     = 5 * time.Minute
+	replicationTTL                   = 5 * time.Second // time to live for neighborhood replication
+	overDraftRefresh                 = time.Millisecond * 600
+	maxDuration        time.Duration = 1<<63 - 1
 )
 
 const (
@@ -56,7 +58,6 @@ var (
 	ErrNoPush            = errors.New("could not push chunk")
 	ErrOutOfDepthStoring = errors.New("storing outside of the neighborhood")
 	ErrWarmup            = errors.New("node warmup time not complete")
-	errPrepareCredit     = errors.New("preparing credit failed")
 )
 
 type PushSyncer interface {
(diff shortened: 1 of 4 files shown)


pkg/skippeers/skippeers_test.go
```diff
@@ -28,9 +28,12 @@ func TestPeerSkipList(t *testing.T) {
 
 	skipList.Add(addr1, addr3, time.Millisecond*10)
 
-	time.Sleep(time.Millisecond * 11)
+	skipList.PruneExpiresAfter(time.Millisecond)
```
aloknerurkar, line 31: Can we also verify the return value here?

pkg/retrieval/retrieval.go
```diff
@@ -275,10 +304,13 @@ func (s *Service) retrieveChunk(ctx context.Context, peer swarm.Address, done ch
 	// Reserve to see whether we can request the chunk
 	creditAction, err := s.accounting.PrepareCredit(creditCtx, peer, chunkPrice, isOrigin)
 	if err != nil {
+		skip.Add(addr, peer, overDraftRefresh)
 		return
 	}
 	defer creditAction.Cleanup()
 
+	skip.Add(addr, peer, maxDuration)
```
aloknerurkar, line 312: We add the overdraft peers separately. If there is an error with this request, we will add the peer to the errSkip list in retrieveChunk. If the request was successful we will have the chunk. So I am not sure if there is any point in adding the peer to the local skiplist here. We might end up wasting some CPU to check a list with duplicate peers in the closestPeer function.

istae: if we add the peer to the skiplist at the end of the request, a preemtive attempt could try the same peer before the first request returns.

aloknerurkar (approved):

pkg/skippeers/skippeers.go
```diff
@@ -70,4 +72,15 @@ func (l *List) PruneExpired() {
 			delete(l.skip, k)
 		}
 	}
+
+	return count
+}
+
+func (l *List) Reset() {
```
istae, line 79: @aloknerurkar @vladopajic @mrekucci just to confirm, do we need this Reset to prevent memory leak?

aloknerurkar: I dont think we need it as this skip list is local to the function and once the function returns it should be GCd.

istae: i was not sure since it's a map within a map

pkg/pushsync/pushsync.go
```diff
@@ -41,10 +41,12 @@ const (
...
-	defaultTTL         = 30 * time.Second // request time to live
-	preemptiveInterval = 5 * time.Second  // P90 request time to live
-	sanctionWait       = 5 * time.Minute
-	replicationTTL     = 5 * time.Second // time to live for neighborhood replication
+	defaultTTL                       = 30 * time.Second // request time to live
+	preemptiveInterval               = 5 * time.Second  // P90 request time to live
+	sanctionWait                     = 5 * time.Minute
+	replicationTTL                   = 5 * time.Second // time to live for neighborhood replication
+	overDraftRefresh                 = time.Second
+	maxDuration        time.Duration = 1<<63 - 1
```
mrekucci, line 49: Use `math.MaxInt64` instead.

pkg/pushsync/pushsync.go
```diff
@@ -397,13 +400,40 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
 
 			peer, err := nextPeer()
+
+			// no peers left
+			if errors.Is(err, topology.ErrNotFound) {
+				if skip.PruneExpiresAfter(overDraftRefresh) == 0 { //no overdraft peers, we have depleted ALL peers
+					if inflight == 0 {
+						ps.logger.Debug("no peers left", "chunk_address", ch.Address(), "error", err)
+						return nil, err
+					} else {
```
mrekucci, line 410: Because of the `return` the `else` is unnecessary here.

pkg/pushsync/pushsync.go
```diff
@@ -397,13 +400,40 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
+
+			if errors.Is(err, topology.ErrWantSelf) {
+				return nil, err
+			}
+
 			if err != nil {
-				if inflight == 0 {
+				if inflight > 0 {
+					continue
+				} else {
```
mrekucci, line 433: Because of the `continue` above `else` is unnecessary here.

pkg/retrieval/retrieval.go
```diff
@@ -106,17 +106,20 @@ func (s *Service) Protocol() p2p.ProtocolSpec {
...
 const (
-	retrieveChunkTimeout = 10 * time.Second
+	retrieveChunkTimeout = time.Second * 10
 	preemptiveInterval   = time.Second
+	overDraftRefresh     = time.Second
 	skiplistDur          = time.Minute
 	maxRetrievedErrors   = 32
 	originSuffix         = "_origin"
+
+	maxDuration time.Duration = 1<<63 - 1
```
mrekucci, line 116: The same as above, use `math.MaxInt64` instead.

pkg/retrieval/retrieval.go
```diff
@@ -220,16 +213,43 @@ func (s *Service) RetrieveChunk(ctx context.Context, chunkAddr, sourcePeerAddr s
...
+					continue
+				}
+
+				// no peers left
+				if errors.Is(res.err, topology.ErrNotFound) {
+					if skip.PruneExpiresAfter(overDraftRefresh) == 0 { //no overdraft peers, we have depleted ALL peers
+						if inflight == 0 {
+							loggerV1.Debug("no peers left", "chunk_address", chunkAddr, "error", res.err)
+							return nil, res.err
+						} else {
```
mrekucci, line 228: As above, unnecessary `else` statement.

mrekucci (approved):

PR #4000, opened 2023-04-11
perf(puller): reduce historical sync workload

istae:
### Checklist

### Description

When the neighborhood size is large, node have to sync with many peers that ultimately respond with the same chunks because reserves are identical across neighborhoods. It is wasteful to have to sync with ALL of the neighbors at the SAME time. 

So, we can limit the amount of parallel hist sync jobs and that should reduce the intensity of the initial syncing.

Since reserves are identical across neighborhoods (we get the same chunk from different peers), the time to find a stable storage radius does not get affected. 

This change also propagates through the network because peers now have to respond to fewer sync requests. 

At the end, we still complete historical syncing with every neighbor. 

The depthmonitor wake up interval is also increased.

### DIFF
--- a/pkg/puller/puller.go
+++ b/pkg/puller/puller.go
@@ -41,7 +41,7 @@ const (
 	histSyncTimeout          = time.Minute * 15
 	histSyncTimeoutBlockList = time.Hour * 24
 
-	maxHistJobs = swarm.MaxBins
+	maxHistSyncsPerBin = 2
 )
 
 type Options struct {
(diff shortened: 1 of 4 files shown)


pkg/puller/puller.go
```diff
@@ -324,24 +308,42 @@ func (p *Puller) histSyncWorker(ctx context.Context, peer swarm.Address, bin uin
...
+	}
+
+	for {
+		p.metrics.HistWorkerIterCounter.Inc()
+
+		select {
+		case <-ctx.Done():
+			loggerV2.Debug("histSyncWorker context cancelled", "peer_address", peer, "bin", bin, "cursor", cur)
+			return
+		case p.activeHistoricalJobs <- struct{}{}:
```
mrekucci, line 339: Not sure what is the main purpose of the `activeHistoricalJobs` channel?

istae: it's to limit the number of sync jobs that are inflight

mrekucci (approved):

aloknerurkar (approved):

PR #4001, opened 2023-04-12
perf(pushsync, retrieval): skiplist prunes items internally on expiration 

istae:
### Checklist

### Description
It was identified in a testing session that during uploads, most CPU cycles are wasted on pruning the skiplist.
With this change, the prune call is no longer needed before every push/retrieval.
The clean up call of items is done once at the time of expiration.
Internally, the skiplist now uses a minimum heap to wake up at the time of the closest expiration. 

### Before
localhost-profile

### After
localhost-heap

### DIFF
--- a/pkg/node/bootstrap.go
+++ b/pkg/node/bootstrap.go
@@ -187,10 +187,11 @@ func bootstrapNode(
 
 	storer := inmemstore.New()
 
-	retrieve := retrieval.New(swarmAddress, storer, p2ps, kad, logger, acc, pricer, tracer, o.RetrievalCaching, noopValidStamp)
+	retrieve, cleanup := retrieval.New(swarmAddress, storer, p2ps, kad, logger, acc, pricer, tracer, o.RetrievalCaching, noopValidStamp)
 	if err = p2ps.AddProtocol(retrieve.Protocol()); err != nil {
 		return nil, fmt.Errorf("retrieval service: %w", err)
 	}
+	defer cleanup()
 
 	ns := netstore.New(storer, noopValidStamp, retrieve, logger)
 
--- a/pkg/node/bootstrap.go
+++ b/pkg/node/bootstrap.go
@@ -187,11 +187,11 @@ func bootstrapNode(
 
 	storer := inmemstore.New()
 
-	retrieve, cleanup := retrieval.New(swarmAddress, storer, p2ps, kad, logger, acc, pricer, tracer, o.RetrievalCaching, noopValidStamp)
+	retrieve := retrieval.New(swarmAddress, storer, p2ps, kad, logger, acc, pricer, tracer, o.RetrievalCaching, noopValidStamp)
 	if err = p2ps.AddProtocol(retrieve.Protocol()); err != nil {
 		return nil, fmt.Errorf("retrieval service: %w", err)
 	}
-	defer cleanup()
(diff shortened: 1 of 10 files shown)


pkg/node/bootstrap.go
```diff
@@ -187,10 +187,11 @@ func bootstrapNode(
 
 	storer := inmemstore.New()
 
-	retrieve := retrieval.New(swarmAddress, storer, p2ps, kad, logger, acc, pricer, tracer, o.RetrievalCaching, noopValidStamp)
+	retrieve, cleanup := retrieval.New(swarmAddress, storer, p2ps, kad, logger, acc, pricer, tracer, o.RetrievalCaching, noopValidStamp)
```
aloknerurkar, line 190: Instead of adding a separate return argument for the closer,  we can just make the retrieval service implement the io.Closer and then add that to the list of closers. This doesnt change the API and I guess would reduce some of the test changes as well.

Same for the pushsync protocol. It seems like both should be implementing the Close to cleanup their internal state from now on. I think this is better and exposing the skiplist.Close method directly. If we need to add more stuff to cleanup later, implementing the closer is a better option.

pkg/skippeers/skippeers.go
```diff
@@ -76,11 +158,40 @@ func (l *List) PruneExpiresAfter(d time.Duration) int {
...
 
+	l.mtx.Lock()
 	for k := range l.skip {
 		delete(l.skip, k)
 	}
+	l.mtx.Unlock()
+}
+
+// An IntHeap is a min-heap of ints.
+type timeHeap []int64
```
aloknerurkar, line 173: This is a good candidate for introducing generics :)

pkg/skippeers/skippeers.go
```diff
@@ -5,41 +5,119 @@
...
+	}
+
+	l.skip[chunk.ByteString()][peer.ByteString()] = t
+}
+
+func (l *List) AddForever(chunk, peer swarm.Address) {
+	l.mtx.Lock()
+	defer l.mtx.Unlock()
+
+	heap.Push(l.minHeap, maxDuration)
```
aloknerurkar, line 107: Do we need to add this to the heap? I dont think these entries will be used to trigger the pruning, so it would be better if we just dont add them to the heap. This would even reduce some unnecessary ops while adding to heap etc.

pkg/skippeers/skippeers.go
```diff
@@ -5,41 +5,119 @@
...
+	}
+
+	if _, ok := l.skip[chunk.ByteString()]; !ok {
+		l.skip[chunk.ByteString()] = make(map[string]int64)
+	}
+
+	l.skip[chunk.ByteString()][peer.ByteString()] = t
+}
+
+func (l *List) AddForever(chunk, peer swarm.Address) {
```
aloknerurkar, line 103: It would be good if we can have tests for this function as well.

pkg/skippeers/main_test.go
```diff
@@ -0,0 +1,15 @@
+// Copyright 2022 The Swarm Authors. All rights reserved.
```
mrekucci, line 1: `2022` -> `2023`

mrekucci (approved):

aloknerurkar (approved):

PR #4002, opened 2023-04-12
feat(status): include bee mode in response

### LINKED ISSUE #3977: status api refactor

mrekucci:
### Checklist

### Description
include bee mode in the snapshot response; refactoring and cleanup

Closes #3977


pkg/status/internal/pb/status.proto
```diff
@@ -15,9 +15,10 @@ message Get {}
 // the appropriate values that are a snapshot of the current state
 // of the running node.
 message Snapshot {
-  uint64 ReserveSize = 1;
-  double PullsyncRate = 2;
-  uint32 StorageRadius = 3;
-  uint64 ConnectedPeers = 4;
-  uint64 NeighborhoodSize = 5;
+  string BeeMode = 1;
```
aloknerurkar, line 18: If we want to maintain backward compatibility, we should be adding this field at the end I think. This allows us to handle messages from older protocol as well.

mrekucci: So the minor version of the protocol is bumped to cover that.

istae (commented):
Can we added the same fields from `/status/peer` to `/status` as well pls?

mrekucci:
> Can we add the same fields from `/status/peer` to `/status` as well pls?

It is already done in this PR.

aloknerurkar (approved):

istae (approved):

PR #4008, opened 2023-04-13
feat(pushync): forwarder multiplexer

istae:
### Checklist

### Description
See the linked issue.

Major changes are:
the forwarder node that can reach the neighborhood fires multiples requests to different peers.
as soon as the chunk lands in the neighborhood, it's stored and replicated. 
the pushsync request now has a new 'replicate bool' field which tells the next peer to only store the chunk and not forward it.


mrekucci (approved):

pkg/pushsync/pushsync.go
```diff
@@ -183,78 +186,55 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
-			}
+		span, _, ctxd := ps.tracer.StartSpanFromContext(ctx, "pushsync-replication-storage", ps.logger, opentracing.Tag{Key: "address", Value: chunkAddress.String()})
+		defer span.Finish()
 
-			debit, err := ps.accounting.PrepareDebit(ctx, p.Address, price)
-			if err != nil {
-				return fmt.Errorf("prepare debit to peer %s before writeback: %w", p.Address.String(), err)
-			}
-			defer debit.Cleanup()
+		chunk, err = ps.validStamp(chunk, ch.Stamp)
```
vladopajic, line 203: Lines [203-232] and [239-266] seems to be doing same thing. Would it be possible to avoid duplication?

pkg/pushsync/pushsync.go
```diff
@@ -183,78 +186,55 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
-			defer span.Finish()
-
-			chunk, err = ps.validStamp(chunk, ch.Stamp)
+		var err error
+		defer func() {
 			if err != nil {
-				return fmt.Errorf("pushsync replication invalid stamp: %w", err)
+				ps.metrics.HandlerReplicationErrors.Inc()
 			}
+		}()
```
vladopajic, line 198: Final error handling on line 193 seems to be shadowing error handling 148. 
If `ch.Replicate` is true, would error handling on line 148 get swallowed in case of error? Is this by intention?

pkg/pushsync/pushsync.go
```diff
@@ -361,33 +336,25 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
+			return swarm.ZeroAddress, topology.ErrNotFound
 		}
-		return peer, nil
+
+		return peer, err
 	}
 
+	retry()
+
 	for sentErrorsLeft > 0 {
```
vladopajic, line 358: does `sentErrorsLeft` needs to be increased to accommodate errors for all forwarding request in case of multiplexing to entire neighborhood?

pkg/pushsync/pushsync.go
```diff
@@ -432,14 +401,27 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
+				retry()
+				continue
+			}
+
+			ps.metrics.TotalSendAttempts.Inc()
 			inflight++
 
 			go func() {
 				ctxd, cancel := context.WithTimeout(ctx, defaultTTL)
 				defer cancel()
```
vladopajic, line 423: Would it make sense to move context with timeout creation to `ps.pushChunk` func ?

pkg/pushsync/pushsync.go
```diff
@@ -85,7 +84,7 @@ type PushSync struct {
 	tracer         *tracing.Tracer
 	validStamp     postage.ValidStampFn
 	signer         crypto.Signer
-	includeSelf    bool
+	fullNode       bool
 	warmupPeriod   time.Time
```
vladopajic, line 88: is `warmupPeriod` still used?

pkg/pushsync/pushsync.go
```diff
@@ -448,20 +430,14 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
 
 			ps.metrics.TotalFailedSendAttempts.Inc()
 			logger.Debug("could not push to peer", "peer_address", result.peer, "error", result.err)
 
-			if result.sent {
-				sentErrorsLeft--
-			}
+			sentErrorsLeft--
 
 			retry()
```
vladopajic, line 442: what is reason for doing `retry()` here? it seems that after result is handled it shouldn't retry to push to closes peer?

pkg/pushsync/pushsync.go
```diff
@@ -480,37 +456,41 @@ func (ps *PushSync) measurePushPeer(t time.Time, err error, origin bool) {
...
 }
 
-func (ps *PushSync) pushPeer(ctx context.Context, skip *skippeers.List, resultChan chan<- receiptResult, doneChan <-chan struct{}, peer swarm.Address, ch swarm.Chunk, origin bool) {
+func (ps *PushSync) prepareCredit(ctx context.Context, peer swarm.Address, ch swarm.Chunk, origin bool) (accounting.Action, error) {
+
+	// compute the price we pay for this receipt and reserve it for the rest of this function
+	receiptPrice := ps.pricer.PeerPrice(peer, ch.Address())
+
+	creditCtx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
```
vladopajic, line 465: Should `creditCtx` use parent context provided via arguments?

pkg/pushsync/pushsync.go
```diff
@@ -608,7 +581,7 @@ func (ps *PushSync) pushToNeighbour(ctx context.Context, peer swarm.Address, ch
 	// cancelled, then glue the stuff on the new context
 	span := tracing.FromContext(ctx)
 
-	ctx, cancel := context.WithTimeout(context.Background(), replicationTTL)
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTTL)
```
vladopajic, line 584: Should ctx use parent context provided via arguments?

pkg/pushsync/pushsync.go
```diff
@@ -563,41 +541,36 @@ func (ps *PushSync) pushPeer(ctx context.Context, skip *skippeers.List, resultCh
...
-		go ps.pushToNeighbour(ctx, peer, ch, origin)
+		go ps.replicateWithPeer(ctx, peer, ch, origin)
 		return false, false, nil
 	}, topology.Filter{Reachable: true})
 }
 
-// pushToNeighbour handles in-neighborhood replication for a single peer.
-func (ps *PushSync) pushToNeighbour(ctx context.Context, peer swarm.Address, ch swarm.Chunk, origin bool) {
+// replicateWithPeer handles in-neighborhood replication for a single peer.
+func (ps *PushSync) replicateWithPeer(ctx context.Context, peer swarm.Address, ch swarm.Chunk, origin bool) {
```
vladopajic, line 573: `replicateWithPeer` seems to be doing very similar to `pushChunk`, would it be possible to avoid duplication?

pkg/pss/pss.go
```diff
@@ -108,6 +109,9 @@ func (p *pss) Send(ctx context.Context, topic Topic, payload []byte, stamper pos
 
 	// push the chunk using push sync so that it reaches it destination in network
 	if _, err = p.pusher.PushChunkToClosest(ctx, tc); err != nil {
+		if errors.Is(err, topology.ErrWantSelf) {
```
aloknerurkar, line 112: Will this be also replicated?

I think for pss to work, we will need to unwrap the trojan chunk. The handler in the pushsync does the unwrap. So I guess if this is replicated, it should be unwrapped.

pkg/pushsync/pushsync.go
```diff
@@ -183,78 +186,55 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
-			ps.metrics.HandlerReplication.Inc()
+	if ch.Replicate {
 
-			var err error
-			defer func() {
-				if err != nil {
-					ps.metrics.HandlerReplicationErrors.Inc()
-				}
-			}()
+		ps.metrics.HandlerReplication.Inc()
```
aloknerurkar, line 191: Before actually storing this we should also check if this chunk is actually within our storage radius. This could be used by some malicious node or some node with incorrect state to falsely make a peer store a chunk.

pkg/pushsync/pushsync.go
```diff
@@ -361,33 +336,25 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
 
-				if !ps.radiusChecker.IsWithinStorageRadius(ch.Address()) {
-					return swarm.ZeroAddress, ErrOutOfDepthStoring
-				}
-
-				ps.pushToNeighbourhood(ctx, fullSkipList, ch, origin, originAddr)
-				return swarm.ZeroAddress, err
-			}
+		// we are a full node and in the neighborhood of the chunk
+		if ps.fullNode && ps.radiusChecker.IsWithinStorageRadius(ch.Address()) {
```
aloknerurkar, line 342: Before checking the radius and returning ErrWantSelf, we should also ensure we are warmed up?

I guess now the radius starts at higher number and then comes down so this is okay?

pkg/pushsync/pushsync.go
```diff
@@ -432,14 +401,27 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
-			ps.metrics.TotalSendAttempts.Inc()
+			// since we can reach into the neighborhood of the chunk
+			// act as the multiplexer and push the chunk in parallel to multiple peers
+			if ps.fullNode && swarm.Proximity(peer.Bytes(), ch.Address().Bytes()) >= ps.radiusChecker.StorageRadius() {
+				for ; parallelForwards > 0; parallelForwards-- {
+					retry()
+				}
+			}
 
+			action, err := ps.prepareCredit(ctx, peer, ch, origin)
```
aloknerurkar, line 412: We should also think about the economics of this. Right now the multiplexer is bearing the cost of the parallel ops and he also doesnt store the chunk. Should he be getting reimbursed somehow? Maybe debiting the requestor peer for the multiplexing.

As he can reach the neighborhood, there is a chance that he will be reimbursed through retrievals in future. But it would be good to write all this down.

pkg/pushsync/pushsync.go
istae: it was originally done like this one purpose I believe

pkg/pushsync/pushsync.go
istae: the neighborhood replication is has to be done independent of everything

pkg/pushsync/pushsync.go
istae: naturally every peer has a likely chance of becoming a forwarder so it's balance out

pkg/pushsync/pushsync.go
```diff
@@ -153,141 +154,99 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
 	if err != nil {
 		return fmt.Errorf("pushsync stamp unmarshall: %w", err)
 	}
 	chunk.WithStamp(stamp)
 
 	if cac.Valid(chunk) {
-		if ps.unwrap != nil {
-			go ps.unwrap(chunk)
-		}
+		go ps.unwrap(chunk)
```
aloknerurkar, line 179: is the ps.unwrap != nil check not required?

aloknerurkar (approved):

pkg/pushsync/pushsync.go
```diff
@@ -650,53 +571,59 @@ func (ps *PushSync) pushToNeighbour(ctx context.Context, peer swarm.Address, ch
...
 	if err = r.ReadMsgWithContext(ctx, &receipt); err != nil {
-		return
+		return nil, err
 	}
 
 	if !ch.Address().Equal(swarm.NewAddress(receipt.Address)) {
-		// if the receipt is invalid, give up
-		return
+		err = fmt.Errorf("invalid receipt. chunk %s, peer %s", ch.Address(), peer)
+		return nil, err
```
vladopajic: could error be returned here without assignment?

pkg/pushsync/pushsync.go
```diff
@@ -633,12 +544,22 @@ func (ps *PushSync) pushToNeighbour(ctx context.Context, peer swarm.Address, ch
...
 	if err != nil {
-		err = fmt.Errorf("new stream for peer %s: %w", peer.String(), err)
 		return
 	}
 
+	err = creditAction.Apply()
+}
+
+// replicateWithPeer handles in-neighborhood replication for a single peer.
+func (ps *PushSync) pushChunkToPeer(ctx context.Context, peer swarm.Address, ch swarm.Chunk, replicate bool) (*pb.Receipt, error) {
```
vladopajic, line 554: comment should start with method name eg. `// pushChunkToPeer ...`

pkg/pushsync/pushsync.go
```diff
@@ -633,12 +544,22 @@ func (ps *PushSync) pushToNeighbour(ctx context.Context, peer swarm.Address, ch
 	}
 	defer creditAction.Cleanup()
```
vladopajic, line 543: could `replicateWithPeer` reuse `prepareCredit` function to prepare credit action?

pkg/pushsync/pushsync.go
```diff
@@ -153,141 +154,99 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
 	} else if !soc.Valid(chunk) {
 		return swarm.ErrInvalidChunk
 	}
 
 	price := ps.pricer.Price(chunkAddress)
 
-	// if the peer is closer to the chunk, AND it's a full node, we were selected for replication. Return early.
-	if p.FullNode {
-		if closer, _ := p.Address.Closer(chunkAddress, ps.address); closer {
+	store := func(ctx context.Context) error {
```
vladopajic, line 186: could this method be moved to individual function, out of `handler`? `handler` looks quite big atm, so every reduction in it's size would simplify reading.

pkg/pushsync/pushsync.go
```diff
@@ -153,141 +154,99 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
...
-			}
-
-			return debit.Apply()
+			return store(ctx)
 		}
 
 		ps.metrics.Forwarder.Inc()
-
 		return fmt.Errorf("handler: push to closest: %w", err)
 	}
```
vladopajic, line 251: is this `s.accounting.PrepareDebit` still needed here (few lines below)? it looks like `store` function is going to do debiting, if this is true then this will debit once again?

pkg/pushsync/pushsync.go
```diff
@@ -432,14 +383,26 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
+
+			ps.metrics.TotalSendAttempts.Inc()
 			inflight++
 
 			go func() {
-				ctxd, cancel := context.WithTimeout(ctx, defaultTTL)
-				defer cancel()
-				ps.pushPeer(ctxd, skip, resultChan, done, peer, ch, origin)
+				ps.push(ctx, resultChan, done, peer, ch, action)
 			}()
```
vladopajic, line 406: nit: unnecessary anonymous function

pkg/pushsync/pushsync.go
istae: that debit call is for forwarder nodes only

pkg/pushsync/pushsync.go
```diff
@@ -480,138 +437,90 @@ func (ps *PushSync) measurePushPeer(t time.Time, err error, origin bool) {
...
-		if err != nil {
-			err = fmt.Errorf("tag %d increment: %w", ch.TagID(), err)
-			return
+	defer func() {
+		select {
+		case resultChan <- receiptResult{pushTime: now, peer: peer, err: err, receipt: receipt}:
+		case <-doneChan:
 		}
-	}
+	}()
```
vladopajic: could code withing this deferred anonymous function be moved at the end of function? 
this would 1) avoid defer func construct; and 2) var block from 462-465 could be removed?

pkg/pushsync/pushsync.go
```diff
@@ -623,20 +503,25 @@ func (ps *PushSync) pushToNeighbour(ctx context.Context, peer swarm.Address, ch
 		}
 	}()
 
-	creditCtx, cancel := context.WithTimeout(context.Background(), time.Second)
-	defer cancel()
+	action, err := ps.prepareCredit(ctx, peer, ch, origin)
```
vladopajic, line 506: `err` is re declared here from line 487, if error happens here will it be handled on line 499?

pkg/pushsync/pushsync.go
```diff
@@ -480,138 +437,90 @@ func (ps *PushSync) measurePushPeer(t time.Time, err error, origin bool) {
...
-		Address: ch.Address().Bytes(),
-		Data:    ch.Data(),
-		Stamp:   stamp,
-	})
-	if err != nil {
-		_ = streamer.Reset()
-		err = fmt.Errorf("chunk %s deliver to peer %s: %w", ch.Address(), peer, err)
-		return
-	}
+func (ps *PushSync) push(ctx context.Context, resultChan chan<- receiptResult, doneChan <-chan struct{}, peer swarm.Address, ch swarm.Chunk, action accounting.Action) {
```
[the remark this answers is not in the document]
istae: context is not canceled ever

pkg/pushsync/pushsync.go
```diff
@@ -420,7 +366,12 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
 				}
 			}
 
+			// we store the chunk
 			if errors.Is(err, topology.ErrWantSelf) {
+				ps.replicateInNeighborhood(ctx, ps.skipList.ChunkPeers(ch.Address()), ch, origin)
```
vladopajic, line 371: could `ps.replicateInNeighborhood` be executed in individual goroutine?

pkg/pushsync/pushsync.go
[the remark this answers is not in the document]
vladopajic: from the perspective of this function, context should signal end of execution. having context and doneChan simultaneously is signaling that something is not right (in this case this context should be canceled, instead of closing doneChan). this probably has implications on upper levels, but nothing that couldn't be solved.

pkg/pushsync/pushsync.go
```diff
@@ -432,36 +383,42 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
 
 			ps.metrics.TotalFailedSendAttempts.Inc()
 			logger.Debug("could not push to peer", "peer_address", result.peer, "error", result.err)
 
-			if result.sent {
-				sentErrorsLeft--
-			}
+			sentErrorsLeft--
 
 			retry()
```
vladopajic, line 423: this case will process results of push operations and when first successful push has happens this function will return (line 415). when parallel forwards are also happening, it appears that there could be these issues:
1) logic from this case statement will not execute; which could skew metrics and telemetry a bit
2) race condition could prevent chunks to be forwarded to other peers if `pushToClosest` returns early after getting first successful push. this can happen because `context` and/or `doneChan` (whichever is relevant) would get closed therefore closing push goroutine.

istae: You have a valid point with the early context cancellation.

istae: I removed the done channel and the push result return now waits on the parent ctx's done channel instead.
The push func creates it's own context to allow each push attempt to run independently of the parent context

vladopajic: for 1), wouldn't it be issue for `TotalSendAttempts` counter is incremented 5 times (1 default forward  + 4 multiplexing) and it's counterpart counter measuring errors `TotalFailedSendAttempts` never even when last 4 push operations have failed? It looks like there will be no logging and no telemetry for multiplexed chunks?

PR #4020, opened 2023-04-19
Misc localstore

aloknerurkar:
### Checklist

### Description
Some misc fixes from the first testing session:

- Memory leak due to the time.After call in reserve worker
- Using reserve putter session in pullsync. This is due to the high memory usage of ReservePut calls observed in the profiles.
- Reporter uses tag based locking to avoid races. This leads to incorrect tags reported.

### DIFF
--- a/pkg/storer/storer.go
+++ b/pkg/storer/storer.go
@@ -311,7 +311,7 @@ func defaultOptions() *Options {
 		LdbDisableSeeksCompaction: defaultDisableSeeksCompaction,
 		CacheCapacity:             defaultCacheCapacity,
 		Logger:                    log.Noop,
-		ReserveCapacity:           2 ^ 22,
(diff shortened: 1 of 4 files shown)


pkg/storer/storer.go
```diff
@@ -311,7 +311,7 @@ func defaultOptions() *Options {
 		LdbDisableSeeksCompaction: defaultDisableSeeksCompaction,
 		CacheCapacity:             defaultCacheCapacity,
 		Logger:                    log.Noop,
-		ReserveCapacity:           2 ^ 22,
+		ReserveCapacity:           4194304, // 2^22 chunks
```
mrekucci, line 314: I'd suggest changing `4194304` to `4_194_304` since it would be more readable.

mrekucci (approved):

PR #4024, opened 2023-04-21
fix(pullsync, reserve): misc fixes and memory improvements

aloknerurkar:
### Checklist

### Description
- Fix incorrect overwrite errors seen on the test cluster.
- Also some improvements related to memory allocations.
- Fix broken test and add new ones

### DIFF
--- a/pkg/pullsync/pullsync.go
+++ b/pkg/pullsync/pullsync.go
@@ -166,7 +166,7 @@ func (s *Syncer) Sync(ctx context.Context, peer swarm.Address, bin uint8, start 
 
 	var (
 		bvLen      = len(offer.Chunks)
-		wantChunks = make(map[string]struct{})
+		wantChunks = make(map[string]struct{}, bvLen)
 		ctr        = 0
 		have       bool
 	)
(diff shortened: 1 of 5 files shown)


pkg/pullsync/pullsync.go
```diff
@@ -545,7 +548,13 @@ func (s *Syncer) Close() error {
 	cc := make(chan struct{})
 	go func() {
 		defer close(cc)
-		s.wg.Wait()
+		for {
```
istae, line 551: What is the benefit of replacing this with the wait group?

istae:
It'd would be great to see the heap profiles of before and after this change.

pkg/pullsync/pullsync.go
```diff
@@ -200,7 +199,7 @@ func (s *Syncer) Sync(ctx context.Context, peer swarm.Address, bin uint8, start
 			}
 
 			if !have {
-				wantChunks[a.ByteString()] = struct{}{}
+				wantChunks[a.ByteString()+string(batchID)] = struct{}{}
```
istae, line 202: 👍

pkg/pullsync/pullsync.go
aloknerurkar: The wait group is racy. The handler Add and this Wait causes data race. This made the tests flaky as I added the Close to the tests.
The only way to fix this would be to add another mutex. But since this waitgroup is mainly used to signal ongoing sync handlers to close, I think this is an easier approach.

istae: hmm, it's strange that new handle calls make it through the s.quit check

PR #4030, opened 2023-04-24
fix(cac): data and span size validation

vladopajic:
### Checklist

### Description
Content addressed chunk validation had bug when validating data size (in `New` function) which prevented chunk with data of zero length to be created. 
This PR fixed this issue (unifying data length validation). Additionally PR improves tests and cosmetics of `cac` package.


pkg/cac/cac.go
```diff
@@ -8,85 +8,87 @@ import (
...
-		return nil, errTooShortChunkData
+	if err := validateDataLength(dataLength - swarm.SpanSize); err != nil {
+		return nil, fmt.Errorf("invalid CAC data length (%d): %w", dataLength, err)
 	}
+
 	return newWithSpan(data[swarm.SpanSize:], data[:swarm.SpanSize])
 }
 
+func validateDataLength(dataLength int) error {
+	if dataLength <= 0 {
```
ldeffenb: I don't know a lot about spans, but does the change from < swarm.SpanSize to <= 0 invalidate a CAC of a zero-length file?  swarm.SpanSize has already been removed in the call to this method.

pkg/cac/cac.go
```diff
@@ -8,85 +8,88 @@ import (
...
+	if err := validateDataLength(dataLength - swarm.SpanSize); err != nil {
+		return nil, fmt.Errorf("invalid CAC data length (%d): %w", dataLength, err)
 	}
+
 	return newWithSpan(data[swarm.SpanSize:], data[:swarm.SpanSize])
 }
 
+// validateDataLength validates if data length (without span) is correct.
+func validateDataLength(dataLength int) error {
+	if dataLength <= 0 { // dataLength could be negative when span size is subtracted
```
ldeffenb, line 49: In particular, consider the following chunk from mainnet which is used by my OSM dataset.
b34ca8c22b9e982354f9c7f50b470d66db428d880c8a904d5fe4ec9713171526

ldeffenb (changes requested):
Please comment on the length validation change which seems to invalidate a currently valid chunk in the swarm.

pkg/cac/cac.go
ldeffenb: The comment doesn't match the condition check.  It's not invalidating negatives, but also invalidating zero.

pkg/cac/cac.go
vladopajic: Currently, `cac` has varying validation for edge case when data == 0.

When `cac` is created with `New` function, error will be returned if `len(data) == 0`
https://github.com/ethersphere/bee/blob/41d6eefa62838830c934f912cd7e87e150ccd68f/pkg/cac/cac.go#L28

When `cac` is created with `NewWithDataSpan` function, it would be possible to create `cac` with zero length data (avoiding condition shown above) if `len(data) == swarm.SpanSize` https://github.com/ethersphere/bee/blob/41d6eefa62838830c934f912cd7e87e150ccd68f/pkg/cac/cac.go#L44

This PR unifies validation to and follows through decision to forbid case when `data == 0`. If this needs to be corrected (to allow this case), validation condition will be updated.

pkg/cac/cac.go
vladopajic: The reason why its invalidating zero data lengths was explained above.

ldeffenb: > In particular, consider the following chunk from mainnet which is used by my OSM dataset. b34ca8c22b9e982354f9c7f50b470d66db428d880c8a904d5fe4ec9713171526

Just to re-verifiy, I uploaded a zero length file to the gateway that resulting in the following reference.   If you expand the manifest recursively on that reference, you will find that it resolves to the previous zero-length reference for the file's (zero length) content.

4974f1792f764351cb64b72412f536f6721855f9535e7c31c7b308b20e32f43d

pkg/cac/cac.go
ldeffenb: I suspect this new decision forbidding data==0 will fail the upload of a zero length file.  Or zero length content with the /bytes endpoint, which is effectively the same thing.  I can forsee many data structures that may require a reference to zero-length content, which this change will no longer allow storing within the swarm.

pkg/cac/cac.go
```diff
@@ -46,7 +46,7 @@ func NewWithDataSpan(data []byte) (swarm.Chunk, error) {
 
 // validateDataLength validates if data length (without span) is correct.
 func validateDataLength(dataLength int) error {
-	if dataLength <= 0 { // dataLength could be negative when span size is subtracted
+	if dataLength < 0 { // dataLength could be negative when span size is subtracted
```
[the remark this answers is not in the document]
vladopajic: Thanks for helping; I wasn't sure which case should actually be valid.

pkg/cac/cac.go
```diff
@@ -8,85 +8,88 @@ import (
...
+	"fmt"
 
 	"github.com/ethersphere/bee/pkg/bmtpool"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
 var (
-	errTooShortChunkData = errors.New("short chunk data")
-	errTooLargeChunkData = errors.New("data too large")
+	ErrChunkDataShort = errors.New("chunk data length is not sufficient")
```
mrekucci, line 18: Instead of two separate errors, I would suggest creating a new error type that will carry the invalid length:

```
type InvalidChunkDataLengthError uint

func (e InvalidChunkDataLengthError) Error() string {
	return fmt.Sprintf("invalid CAC data length: %d", uint(e))
}
```

and then just below return `InvalidChunkDataLengthError(dataLength)`.

vladopajic: It looks that this case is more suitable for two errors, one error should indicate when **data** exceeds max allowed length, and second should indicate case when **span** size is not exactly `swarm.SpanSize`. There was problem with error text which has been corrected in new commits.

--- 

Just on the side note... Consider following case which uses `InvalidChunkDataLengthError` type as suggested:


```
func foo() error { // foo return error from `cac.New` and some other error
	err := foo1() 
	if err != nil {
		return err
	}
	
	return cac.New(...)
}

{ // some other logic call `foo`, and want's to know what has happend
	...
	err := foo()
	if error.Is(err, InvalidChunkDataLengthError(0)) { // but what if value could be 1, 2, 3, or 141 ...?
		// problem with cac
	} else {
		// some other problem
	}
	...
}

```

 Errors specified in suggested way implies that other logic which uses `cac` wouldn't be able to use `errors.Is(...)` when processing error result because it would need to provide exact value of data length which could be different in different cases. Therefore this construct should probably be avoided if code wants to be future-proof.

mrekucci: The `errors.Is(...)` shouldn't be an issue, you can use `errors.As(err, new(InvalidChunkDataLengthError))` instead.

aloknerurkar (approved):

pkg/cac/cac.go
```diff
@@ -7,86 +7,89 @@ package cac
...
 
 	"github.com/ethersphere/bee/pkg/bmtpool"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
 var (
-	errTooShortChunkData = errors.New("short chunk data")
-	errTooLargeChunkData = errors.New("data too large")
+	ErrChunkSpanShort = fmt.Errorf("chunk span must have exactly length of %d", swarm.SpanSize)
+	ErrChunkDataLarge = fmt.Errorf("chunk data exceeds maximum allowed length")
```
notanatol, line 18: this could be left out as `errors.New` since no formatting is happening, but since it's compile time might as well leave it like this.

notanatol (approved):

PR #4053, opened 2023-05-06
fix(storer): misc fixes from testing

aloknerurkar:
### Checklist

### Description
Latest round of testing after rebase

Issues:
1. Initialize inmem store for stampissuer on restart. Required till peter's changes are in.
2. Direct upload gets stuck if we see lot of failures. This is because we dont give up the sem till we are able to repeat.
3. Reserve.Put lock. The Put method needs lock as simultaneous puller routines might try to add the same chunks which causes new binIDs to be assigned to same chunk as we dont have sync.
4. In the private testnet we dont have any outside connections, which leads to all nodes having reachability unknown. So this is a temporary fix for the private testnet.

Results:
1. Uploads, direct uploads, pinning all working well
2. Syncing also working well
3. Retrieval also works well

### DIFF
--- a/pkg/storer/internal/reserve/reserve.go
+++ b/pkg/storer/internal/reserve/reserve.go
@@ -27,7 +27,8 @@ const loggerName = "reserve"
 const reserveNamespace = "reserve"
 
 type Reserve struct {
-	mtx sync.Mutex
+	mtx      sync.Mutex
+	putterMu sync.Mutex
 
 	baseAddr     swarm.Address
 	radiusSetter topology.SetStorageRadiuser
@@ -85,6 +86,8 @@ func New(
 
 // Put stores a new chunk in the reserve and returns if the reserve size should increase.
 func (r *Reserve) Put(ctx context.Context, store internal.Storage, chunk swarm.Chunk) (bool, error) {
+	r.putterMu.Lock()
+	defer r.putterMu.Unlock()
 
 	indexStore := store.IndexStore()
 	chunkStore := store.ChunkStore()
@@ -431,10 +434,8 @@ func (r *Reserve) SetRadius(store storage.Store, rad uint8) error {
 	return store.Put(&radiusItem{Radius: rad})
 }
 
+// should be called under lock
 func (r *Reserve) incBinID(store storage.Store, bin uint8) (uint64, error) {
-	r.mtx.Lock()
-	defer r.mtx.Unlock()
-
 	item := &binItem{Bin: bin}
 	err := store.Get(item)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/topology/kademlia/kademlia.go
```diff
@@ -1260,7 +1260,7 @@ func nClosePeerInSlice(peers []swarm.Address, addr swarm.Address, spf sanctioned
 }
 
 func (k *Kad) IsReachable() bool {
-	return k.reachability == p2p.ReachabilityStatusPublic
+	return k.reachability != p2p.ReachabilityStatusPrivate
```
istae, line 1263: this should be explicitly Public, if there is middle "unknown" state, we don't want these nodes to store chunks.

pkg/storer/internal/reserve/reserve.go
```diff
@@ -84,6 +86,8 @@ func New(
 
 // Put stores a new chunk in the reserve and returns if the reserve size should increase.
 func (r *Reserve) Put(ctx context.Context, store internal.Storage, chunk swarm.Chunk) (bool, error) {
+	r.putterMu.Lock()
```
istae, line 89: func that returns a new binID has a lock already

istae: if syncing started working "better" after this lock, then there is another hidden issue

aloknerurkar: So we need the lock for the whole process of checking the chunkBinIndex. Parallel puller routines could be trying to add the same chunk and as only the binID increment is logged, it will end up adding duplicate items as we dont wait for the Put to finish before we check in Has.

pkg/topology/kademlia/kademlia.go
vladopajic, line 1263: it would be nice to know why this logic was changed, event better to have this documented here. 
someone reading this logic may assume that there is error in logic, because node will be reachable when `reachability` is `p2p.ReachabilityStatusUnknown`.

vladopajic (approved):
LGTM; minor comments

mrekucci (approved):

pkg/topology/kademlia/kademlia.go
aloknerurkar: So this is a temporary fix as I mentioned in the description. The reason is, in the private network, nodes only connect to other nodes in the private network. As a result, they are never "public". This is sort of a hack for this branch. I will revert it before checkin.

vladopajic: okay got it

how do you keep track of things that will need to be changed before/after this branch is merged to master?

PR #4054, opened 2023-05-08
fix: add reachability field to status response

notanatol:
### Checklist

- [X] I have added tests to cover my changes.

### Description
Closes: #4048


mrekucci (changes requested):
Issue #4048 is not clear about this but the field should be added on the status protocol level so we can retrieve the reachability status of the other peers. Look at the following PR as an example: https://github.com/ethersphere/bee/pull/4010/files

pkg/status/status.go
```diff
@@ -74,6 +79,7 @@ func (s *Service) LocalSnapshot() (*Snapshot, error) {
 		ConnectedPeers:   connectedPeers,
 		NeighborhoodSize: neighborhoodSize,
 		BatchTotalAmount: s.chainstate.GetChainState().TotalAmount.String(),
+		IsReachable:      s.topologyDriver.IsReachable(),
```
istae, line 82: instead of getting this field from the peer, we can actually get it directly from our kademlia, 

here we update a peer's rechability https://github.com/ethersphere/bee/blob/master/pkg/topology/kademlia/kademlia.go#L1338

we can add a new function to kademlia that returns of a peer is reachable or not.

istae: next we should remove the `topology.Filter{Reachable: true}` filter in status.go so that we try to get the status of ALL of our connected peers.

istae (changes requested):

mrekucci (changes requested):
A new `IsPeerReachable` field should be returned as a result of the protocol message returned by the requested peer(s).

notanatol:
> A new `IsPeerReachable` field should be returned as a result of the protocol message returned by the requested peer(s).

okay but I just removed it as requested by @istae 

should I put it back Esad?

mrekucci:
> > A new `IsPeerReachable` field should be returned as a result of the protocol message returned by the requested peer(s).
> 
> okay but I just removed it as requested by @istae
> 
> should I put it back Esad?

We send the request to the peer(s) anyway, so taking it out of the kademlia just complicates the situation a bit and is not uniform. I would suggest going with the previous route, then you won't need additional method in the interface and a new structure to bake in the new field.

istae:
> > > A new `IsPeerReachable` field should be returned as a result of the protocol message returned by the requested peer(s).
> > 
> > 
> > okay but I just removed it as requested by @istae
> > should I put it back Esad?
> 
> We send the request to the peer(s) anyway, so taking it out of the kademlia just complicates the situation a bit and is not uniform. I would suggest going with the previous route, then you won't need additional method in the interface and a new structure to bake in the new field.

It's best that the endpoint returns the reachability of a peer from the perspective of our node (from our kademlia).

istae (approved):

vladopajic (approved):

mrekucci (approved):

PR #4055, opened 2023-05-08
feat(sample): exclude batches with low balance

vladopajic:
### Checklist

### Description
Exclude batches from sample with low balance.

### DIFF
--- a/pkg/storer/reserve.go
+++ b/pkg/storer/reserve.go
@@ -11,6 +11,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"math/big"
 	"sort"
 	"sync"
 	"testing"
@@ -253,6 +254,23 @@ func (db *DB) evictBatch(ctx context.Context, batchID []byte, upToBin uint8) (er
 	return nil
 }
 
+func (db *DB) batchesBelowValue(until *big.Int) (map[string]struct{}, error) {
+	res := make(map[string]struct{})
+
+	if until == nil {
+		return res, nil
+	}
+
+	err := db.batchstore.Iterate(func(b *postage.Batch) (bool, error) {
+		if b.Value.Cmp(until) < 0 {
+			res[string(b.ID)] = struct{}{}
+		}
+		return false, nil
+	})
+
+	return res, err
+}
+
 func (db *DB) unreserve(ctx context.Context) (err error) {
 	dur := captureDuration(time.Now())
 	defer func() {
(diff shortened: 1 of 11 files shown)


pkg/storer/reserve.go
```diff
@@ -621,25 +651,27 @@ func transformedAddressSOC(hasher *bmt.Hasher, chunk swarm.Chunk) (swarm.Address
...
-	NotFound           atomic.Int64
-	NewIgnored         atomic.Int64
-	IterationDuration  atomic.Int64
-	GetDuration        atomic.Int64
-	HmacrDuration      atomic.Int64
-	ValidStampDuration atomic.Int64
+	TotalIterated       atomic.Int64
+	NotFound            atomic.Int64
+	NewIgnored          atomic.Int64
+	BelowBalanceIgnored atomic.Int64
```
ldeffenb, line 657: Pardon my ignorance, but what's the difference between "BelowBalanceIgnored" and "ValidStampDuration"?  I thought the stamp duration (TTL) was based on the balance of the stamp vs the current storage price?

vladopajic: `BelowBalanceIgnored` is count of chunks that are ignored form sample calculation logic because it's batch balance value is not enough to cover storage cost until next round of shelling game.
`ValidStampDuration` is total time spent for stamp validation (in sample calculation logic). actually it should be this, but on new storage implementation this piece of logic is missing. [here](https://github.com/ethersphere/bee/blob/exclude-batches-close-to-expiry/pkg/_localstore/sampler.go#L234) is link to old sampling logic which has this metrics.

ldeffenb: Or am I completely mistaken that there ever was a minimum stamp lifetime in the determination of what should be hashed for the reserve calculations?   Maybe it was that new chunks were excluded because they may not be fully propagated into the swarm?

aloknerurkar: > Or am I completely mistaken that there ever was a minimum stamp remaining lifetime in the determination of what should be hashed for the reserve calculations? Maybe it was that new chunks were excluded because they may not be fully propagated into the swarm?

Yes so till now we only excluded chunks based on timestamp. This was not fully correct but this was to mainly protect against new live syncing chunks which may not be propagated fully during the sampling. With this change, we will tighten the screws even more. The goal is to not be affected by stamp expirations during the sampling. So we will ignore batches that may not last on all nodes while they are sampling.

aloknerurkar (approved):

PR #4061, opened 2023-05-11
chore: switch to mainnet abi in unit tests

vladopajic:
### Checklist

### Description
Changing ABIs used in unit tests from testnet to mainnet. 
From the standpoint of tests both should work, but using mainnet ABIs adds more value because those are more important to testnet's.

### DIFF
--- a/pkg/postage/listener/listener_test.go
+++ b/pkg/postage/listener/listener_test.go
@@ -32,7 +32,7 @@ var (
 
 var (
 	postageStampContractAddress = common.HexToAddress("eeee")
-	postageStampContractABI     = abiutil.MustParseABI(chaincfg.Testnet.PostageStampABI)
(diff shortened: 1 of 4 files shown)


notanatol (approved):

mrekucci (changes requested):
The main reason why tests use testnet and not mainnet is for feature development. When new features are developed, they are tested against contracts deployed in the testnet, and the contracts may change frequently during the development cycle before they are finalized and released to the mainnet. Using the mainnet will make the development cycle very complicated.

mrekucci:
> @mrekucci would it make sense to use mainnet ABIs by default for all unit tests, but when particular feature, that requires different ABIs, is developed only it's unit tests could switch to testnet ABI temporally, until feature goes live?

It's certainly possible, but it may bring less clarity to the codebase as to why some tests use the mainnet and some use the testnet, it will also bring more bookkeeping PRs once contracts are released to the mainnet. I'd suggest rather avoiding such additional complexity.

PR #4072, opened 2023-05-17
feat: example use of new atomics data structs

notanatol:
### Checklist

### Description
Going forward we should adopt the new semantics of dealing with atomics.

### DIFF
--- a/pkg/api/probe.go
+++ b/pkg/api/probe.go
@@ -8,21 +8,11 @@ import "sync/atomic"
 
 // ProbeStatus is the status of a probe.
 // ProbeStatus is treated as a sync/atomic int32.
-type ProbeStatus int32
-
-// get returns the value of the ProbeStatus.
-func (ps *ProbeStatus) get() ProbeStatus {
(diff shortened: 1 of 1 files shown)


pkg/api/probe.go
```diff
@@ -7,22 +7,12 @@ package api
...
-func (ps *ProbeStatus) set(v ProbeStatus) {
-	atomic.StoreInt32((*int32)(ps), int32(v))
-}
+// ProbeStatus is treated as a sync/atomic bool.
+type ProbeStatus bool
 
 // String implements the fmt.Stringer interface.
 func (ps ProbeStatus) String() string {
-	switch ps.get() {
+	switch ps {
```
vladopajic, line 15: since `ProbeStatus` type is `bool` then `uknown` string value is not needed anymore (this is unreachable code), right?

vladopajic (approved):
LGTM

istae:
so it's like the uber atomic library, nice.

PR #4077, opened 2023-05-18
fix(pusher, retrieval, salud):  panic and leak fix

istae:
### Checklist

### Description
includes a panic fix in status protocol.
fix for retrieval goroutine leak

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -6,9 +6,7 @@ package status
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"io"
 
 	"github.com/ethersphere/bee/pkg/log"
 	"github.com/ethersphere/bee/pkg/p2p"
@@ -136,7 +134,7 @@ func (s *Service) handler(ctx context.Context, _ p2p.Peer, stream p2p.Stream) er
 	}()
 
 	var msgGet pb.Get
-	if err := r.ReadMsgWithContext(ctx, &msgGet); err != nil && !errors.Is(err, io.EOF) {
+	if err := r.ReadMsgWithContext(ctx, &msgGet); err != nil {
 		loggerV2.Debug("read message failed", "error", err)
 		return fmt.Errorf("read message: %w", err)
 	}
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -106,9 +106,6 @@ func (s *Service) PeerSnapshot(ctx context.Context, peer swarm.Address) (*Snapsh
 
 	ss := new(pb.Snapshot)
 	if err := r.ReadMsgWithContext(ctx, ss); err != nil {
-		if errors.Is(err, io.EOF) {
-			return nil, nil
-		}
 		return nil, fmt.Errorf("read message failed: %w", err)
 	}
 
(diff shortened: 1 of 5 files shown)


ldeffenb (commented):
This change to status.go appears to fix a panic that I just opened in #4078

ldeffenb:
Lost 2 more nodes overnight.   This fix really needs to get released!

vladopajic (approved):

pkg/status/status.go
```diff
@@ -106,9 +106,6 @@ func (s *Service) PeerSnapshot(ctx context.Context, peer swarm.Address) (*Snapsh
 
 	ss := new(pb.Snapshot)
 	if err := r.ReadMsgWithContext(ctx, ss); err != nil {
-		if errors.Is(err, io.EOF) {
```
mrekucci, line 109: Why is this removed?

istae: why are we catch EOF error here ?

mrekucci: It should signals the end of the message, so it should be:

```Golang
if !errors.Is(err, io.EOF) {
	return nil, fmt.Errorf("read message failed: %w", err)
}
```

istae: it's an error if ReadMsgWithContext returns io.EOF as we did not receive a proper msg

mrekucci: The protobuf package is using the [`io.Read`](https://github.com/gogo/protobuf/blob/master/io/full.go#L90) underneath, that has the following behavior:
> // When Read encounters an error or end-of-file condition after
// successfully reading n > 0 bytes, it returns the number of
// bytes read. It may return the (non-nil) error from the same call
// or return the error (and n == 0) from a subsequent call.
// An instance of this general case is that a Reader returning
// a non-zero number of bytes at the end of the input stream may
// return either err == EOF or err == nil. The next Read should
// return 0, EOF.
//
// Callers should always process the n > 0 bytes returned before
// considering the error err. Doing so correctly handles I/O errors
// that happen after reading some bytes and also both of the
// allowed EOF behaviors.
//
// Implementations of Read are discouraged from returning a
// zero byte count with a nil error, except when len(p) == 0.
// Callers should treat a return of 0 and nil as indicating that
// nothing happened; in particular it does not indicate EOF.

ldeffenb: So if protobuf considers it an error, and if the entire protobuf (presumably) wasn't read before the EOF was indicated, then it didn't work and the error should be passed back up to the caller, not nil, nil.

istae: I agree with @ldeffenb here, EOF err is an early termination error so the read call did not fetch the entire msg.

What I got from this
> // When Read encounters an error or end-of-file condition after
// successfully reading n > 0 bytes, it returns the number of
// bytes read. It may return the (non-nil) error from the same call
// or return the error (and n == 0) from a subsequent call.
// An instance of this general case is that a Reader returning
// a non-zero number of bytes at the end of the input stream may
// return either err == EOF or err == nil. The next Read should
// return 0, EOF.

if read encounters an error when it received some n > 0 bytes, it may or not send the error back.
BUT, on the next read call, n will be 0, and the error is returned.

aloknerurkar (approved):

Issue #3992, opened 2023-04-10
introduce a layer of memory caching for the statestore


istae:
Any time a chunk moves around in the network through retrieval, pushsync, and pullsync protocols, the associated batch has to be fetches from the batchstore for validation.
On occasion, this process can be a bottleneck, and may be improved with adding a layer of memory caching to skip on reading the batch from disk.

An lru cache with a large max size should be used (similar to what we have in hive pkg)

PR #4089, opened 2023-05-22
feat: lru cache for batchstore

quinton11:
Addresses #3992


istae:
Hey, nice PR, it's a good step in the right direction.
It would be good to cache all statestore items, but it's not the right time to tackle this as there is large rewrite of the localstore and statestore. 
Once the rewrite is out, we can come back to this. So for now, closing this.

istae:
related #4177

Issue #3979, opened 2023-04-07
For "redistributionstate" endpoint, change "lastPlayedRound" to actually show last round played, and add "lastSelectedRound" so show if the node's neighborhood was selected


NoahMaizels:
For "redistributionstate" endpoint, change "lastPlayedRound" to actually show last round played, and add a new return value - "lastSelectedRound" so show if the node's neighborhood was selected.

The name of "lastPlayedRound" is confusing, since if the sampler failed, it could show that the lastPlayedRound was a round which the node did not actually play in. 

And adding lastSelectedRound in addition to a lastPlayedRound which actually shows the round played would be helpful for debugging purposes. It could for example help to identify nodes who frequently fail during the sampler process, such as nodes with CPUs that aren't fast enough.

Issue #3850, opened 2023-02-27
/redistributionstate lastPlayedRound off-by-one


ldeffenb:
### Context
mainnet 1.12.0+

### Summary
My node won the last round that it played, but the /redistributionstate's lastPlayedRound is one less than the (correct) lastWonRound.

```
  "isFrozen": false,
  "isFullySynced": true,
  "phase": "reveal",
  "round": 175611,
  "lastWonRound": 175369,
  "lastPlayedRound": 175368,
  "lastFrozenRound": 0,
  "block": 26692911,
  "reward": "10246290407424000",
  "fees": "20343529360500000"
```

### Expected behavior
I would expect the lastPlayedRound to always be >= lastWonRound.

### Actual behavior
```
  "isFrozen": false,
  "isFullySynced": true,
  "phase": "reveal",
  "round": 175611,
  "lastWonRound": 175369,
  "lastPlayedRound": 175368,
  "lastFrozenRound": 0,
  "block": 26692911,
  "reward": "10246290407424000",
  "fees": "20343529360500000"
```

### Steps to reproduce
Check the /redistributionstate API output on a node that won the last round played.   Or even notice that the lastPlayedRound is actually one less than the round that the node committed and revealed.

### Possible solution
I'm guessing that the lastPlayedRound is being picked up during the claim phase of the previous round when the node begins calculating the reserve hash (aka "starts playing"), but the actual round number that should be captured is the next round which increments at the end of claim and beginning of commit which is the actual round being played.

vladopajic:
@ldeffenb could you confirm if this is still happening in v1.15.0?

ldeffenb:
Yes, this is still happening on 1.15.   The output just now from one of my staked nodes:
{
  "minimumFunds": "18750000000000000",
  "hasSufficientFunds": true,
  "isFrozen": false,
  "isFullySynced": false,
  "phase": "claim",
  "round": 183317,
  "lastWonRound": 183291,
  "lastPlayedRound": 183290,
  "lastFrozenRound": 182326,
  "block": 27864261,
  "reward": "378811176714240000",
  "fees": "115269532552000000"
}

PR #4090, opened 2023-05-22
refactor: Added information for redistributionState endpoint

quinton11:
See #3979

Changes include a "lastSelectedRound" value indicating the last round an agent's neightbourhood was selected as well as the "lastPlayedRound"  now referring to the most recently played round


pkg/storageincentives/agent.go
```diff
@@ -369,6 +369,9 @@ func (a *Agent) handleClaim(ctx context.Context, round uint64) (bool, error) {
 }
 
 func (a *Agent) handleSample(ctx context.Context, round uint64) (bool, error) {
+	//last played round
```
mrekucci, line 372: This comment doesn't add much value.

pkg/api/redistribution.go
```diff
@@ -22,6 +22,7 @@ type redistributionStatusResponse struct {
 	LastWonRound       uint64         `json:"lastWonRound"`
 	LastPlayedRound    uint64         `json:"lastPlayedRound"`
 	LastFrozenRound    uint64         `json:"lastFrozenRound"`
+	LastSelectedRound  uint64         `json:"lastSelectedRound"`
```
mrekucci, line 25: This should be also reflected in the OpenAPI spec.

mrekucci (changes requested):

pkg/storageincentives/agent.go
```diff
@@ -369,6 +369,9 @@ func (a *Agent) handleClaim(ctx context.Context, round uint64) (bool, error) {
 }
 
 func (a *Agent) handleSample(ctx context.Context, round uint64) (bool, error) {
+
+	a.state.SetLastPlayedRound(round)
+
```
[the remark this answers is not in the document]
quinton11: I see, correct me if I'm wrong but if its at the end of handleSample func doesn't that mean it will be updated only when the node's neighbourhood is selected?( `a.contract.isPlaying`) From #3979 i believe the "lastPlayedRound" referred to the previous round not necessarily one that the node's neighbourhood participated in hence the distinction with the "lastSelectedRound".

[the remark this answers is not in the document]
ldeffenb: I thought lastSelectedRound is when the neighborhood was selected.  lastPlayedRound would be one where the node actually submitted (or possibly calculated) a sample hash meaning that the node actually played the round.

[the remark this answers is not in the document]
quinton11: You're right, will address

vladopajic (approved):
LGMT

istae:
@vandot @mrekucci
there is some credentials errors under the OpenAPI check, do you know what's it about?

vandot:
@istae credentials are not exposed in this PR.
PRs from forks don't have access to Github Secrets and that behaviour cannot be changed.

quinton11:
What do you suggest? I revert the changes made to the openapi spec?

vandot:
@istae Outside Collaborators can clone and create direct PRs

vladopajic:
Why should commits to the PR upload new openapi spec? That step should only be done if commit is happening in main branch, right?

pkg/storageincentives/agent.go
```diff
@@ -316,7 +316,7 @@ func (a *Agent) handleClaim(ctx context.Context, round uint64) (bool, error) {
 		return false, nil
 	}
 
-	a.state.SetLastPlayedRound(round)
+	a.state.SetLastSelectedRound(round)
```
ldeffenb, line 319: Personally, "LastSelectedRound" sounds like it should be set when the node's neighborhood is selected, like right at the start of "handleSample"?  Even before the checks to see if we will be playing that round.

pkg/storageincentives/agent.go
```diff
@@ -415,6 +416,7 @@ func (a *Agent) handleSample(ctx context.Context, round uint64) (bool, error) {
 		return false, err
 	}
 
+	a.state.SetLastPlayedRound(round)
```
ldeffenb, line 419: I haven't chased the new code the whole way through, but what is the definition of "Played"?   With this Set here, it means that a sample was generated.  But to me, a node hasn't "Played" until it has successfully submitted the Commit transaction?   Or am I picking nits here?

pkg/storageincentives/agent.go
quinton11: Well, considering the fact that  'neighbourhood chosen' together with the round is logged to the user after the checks, just before making the sample, it made sense to set the `LastSelectedRound` there. What do you think?

quinton11: But yeah, should be swapped with `LastPlayedRound`

ldeffenb: > But yeah, should be swapped with `LastPlayedRound`

Only if SetSelectedRound is moved up to the beginning before the checks.   The neighborhood is selected for a round even if the node chooses not to play for some reason.   I'd figure that's the primary difference between "Selected" and "Played".

pkg/storageincentives/agent.go
quinton11: No you're not, that would be more accurate

pkg/storageincentives/agent.go
quinton11: right, will amend, but wouldn't that make the positioning of the log message `a.logger.Info("neighbourhood chosen", "round", round)` inaccurate? Since all the checks have to pass before its logged?

pkg/storageincentives/agent.go
```diff
@@ -370,6 +371,8 @@ func (a *Agent) handleClaim(ctx context.Context, round uint64) (bool, error) {
 
 func (a *Agent) handleSample(ctx context.Context, round uint64) (bool, error) {
 
+	a.state.SetLastSelectedRound(round)
```
ldeffenb, line 374: Because handleSample is invoked in the claim phase, the round is likely to be the current (claim) round and not the upcoming round for which the node is selected and the sample is about to be calculated.   This is what triggered the whole shebang with #3850

ldeffenb (changes requested):

pkg/storageincentives/agent.go
quinton11: Placing it at the top of `handleClaim` then will be safer?

ldeffenb: Nope, both handleSample and handleClaim have the same round number and it's the round that is just ending.  During the claim phase, the NEXT round's neighborhood is selected and the sample for that round will begin to be calculated.

I'm leaving it to the powers that be to decide on the proper fix, but simply SetLastSelectedRound(round+1) or handleSample(ctx,round+1) would work.  It's no worse than the (round - 1) check at line 269. which would probably be busted if handleSample(ctx,round+1) is done.

The comment at line 268 about the sample being from the "previous round" is technically incorrect because the sample is actually for THIS round, the calculation was just started in the previous round's claim phase.

So IMNSHO, the fix should be for handleSample(ctx,round+1) and remove the (round-1) in the check at line 269 since the sample would actually be calculated for the proper (e.g. next) round.

pkg/storageincentives/agent.go
```diff
@@ -400,6 +398,8 @@ func (a *Agent) handleSample(ctx context.Context, round uint64) (bool, error) {
 		return false, nil
 	}
 
+	a.state.SetLastSelectedRound(round + 1)
```
ldeffenb, line 401: I agree with the +1, but why move it from the start until after the validation?   IMHO the node/neighborhood is selected even if it is not fully synced.  If someone is watching this value to determine the frequency of neighborhood selection, it is no longer correct.

On the other hand, if we are not fully synced and therefore have the wrong idea of the storage radius, we may think we are selected when in fact we are not.   I can see it both ways.

pkg/storageincentives/agent.go
```diff
@@ -277,6 +277,8 @@ func (a *Agent) handleCommit(ctx context.Context, round uint64) (bool, error) {
 		return false, err
 	}
 
+	a.state.SetLastPlayedRound(round)
+
```
ldeffenb, line 281: Nice, we've played once we've successfully committed.

pkg/storageincentives/agent.go
```diff
@@ -369,6 +370,9 @@ func (a *Agent) handleClaim(ctx context.Context, round uint64) (bool, error) {
 }
 
 func (a *Agent) handleSample(ctx context.Context, round uint64) (bool, error) {
+
+	a.state.SetLastSelectedRound(round + 1)
```
ldeffenb, line 374: And as soon as we start handling a sample, we know our neighborhood has been selected, regardless of what other conditions may apply.  If a node sees Selected without Played, they know they missed out for some reason.

ldeffenb: And line 174 covers the sample that finishes calculating after the commit phase has started by re-triggering the commit event which gets us marked as Playing.

ldeffenb (approved):
LGTM now!

pkg/storageincentives/agent.go
vladopajic: @ldeffenb `handleSample` is called on begging on every `claim` phase. At this point agent still doesn't know if node is participating in this round until it calls `isParticipatingInUpcomingRound` in redistribution smart contract, right? If this is right, then logic in this line is not correct because `LastSelectedRound` is always going to be set. 

Imo, this should be corrected by rearranging early return conditions:

- check if `isParticipatingInUpcomingRound` first, early return on error or false
- set last selected round (at this point neighborhood is selected)
- check other conditions, and early return if those are meet 
- finally make sample

does this make sense ?

pkg/storageincentives/agent.go
vladopajic: +1, this is the best place to set last played round

pkg/api/redistribution.go
vladopajic: @quinton11 hey 🌞

ldeffenb (changes requested):
Changes proposed by @vladopajic seem to be necessary.

pkg/storageincentives/agent.go
quinton11: Right, so the initial placement was correct? Just after the neighbourhood check?

vladopajic: > Right, so the initial placement was correct? Just after the neighbourhood check?

Not quite exactly as initial. I am thinking that this should be way to go:

```
func (a *Agent) handleSample(ctx context.Context, round uint64) (bool, error) {
	storageRadius := a.store.StorageRadius()

	isPlaying, err := a.contract.IsPlaying(ctx, storageRadius)
	if err != nil {
		a.metrics.ErrCheckIsPlaying.Inc()
		return false, err
	}
	if !isPlaying {
		a.logger.Info("not playing in this round")
		return false, nil
	}

	a.metrics.NeighborhoodSelected.Inc()
	a.state.SetLastSelectedRound(round + 1)

	if !a.state.IsFullySynced() {
		a.logger.Info("skipping round because node is not fully synced")
		return false, nil
	}

	if a.state.IsFrozen() {
		a.logger.Info("skipping round because node is frozen")
		return false, nil
	}

	if !a.health.IsHealthy() {
		a.logger.Info("skipping round because node is unhealhy", "round", round)
		return false, nil
	}

	_, hasFunds, err := a.HasEnoughFundsToPlay(ctx)
	if err != nil {
		return false, fmt.Errorf("has enough funds to play: %w", err)
	}

	if !hasFunds {
		a.logger.Info("insufficient funds to participate in next round", "round", round)
		a.metrics.InsufficientFundsToPlay.Inc()
		return false, nil
	}

	sample, err := a.makeSample(ctx, storageRadius)
	if err != nil {
		return false, err
	}

	a.state.SetSampleData(round, sample)

	return true, nil
}
```

vladopajic:
@quinton11 please make sure that lint check is passing. you can run `make lint` command locally to run lint check

pkg/storageincentives/agent.go
quinton11: Okay, so the neighbourhood check before the node state checks, right

pkg/api/redistribution.go
quinton11: @vladopajic  I reverted the changes made to the openapi spec to prevent triggering a new upload. Would open another pr from a clone with the changes and link to this pr, what do you think?

vladopajic: Thanks for being cautious. It might have more sense to add that change in this PR.

quinton11: If I add the change to this PR the OpenApi job will fail at the credential check. Or I'm not getting your meaning?

vladopajic: It will, but just ignore it. Bee team members can merge PRs avoiding CI checks.

quinton11:
@vladopajic all checks bar OpenApi have passed

vladopajic (approved):
LGTM

istae (approved):

PR #4091, opened 2023-05-22
fix(storer): fixes from load testing

aloknerurkar:
### Checklist

### Description
1. The eviction removal and put should be synchronized.
2. Dont use ReservePutter in puller as parallel goroutines cause memory to pile up as they are waiting to acquire the Put lock. Single instance Put cleans up the txn on every PUT. This coupled with a lock in storer prevents multiple goroutines from allocating the Putter.
3. Reduce in-mem footprint of levelDB


pkg/storer/internal/reserve/reserve.go
```diff
@@ -344,11 +350,18 @@ func (r *Reserve) EvictBatchBin(ctx context.Context, store internal.Storage, bin
 	return len(evicted), nil
 }
 
-func removeChunk(store internal.Storage, item *batchRadiusItem) error {
+func removeChunk(ctx context.Context, store internal.Storage, item *batchRadiusItem) error {
```
mrekucci, line 353: You can use the context in the `chunkStore.Delete` call below.

mrekucci (approved):
Apart from the comment and lint issues, LGTM!

Issue #3960, opened 2023-04-03
mechanism for pausing/resuming storage incentives game


vladopajic:
**Problem**
Node operators can potentially experience penalties in cases when their node is currently participating in storage incentives game and they are doing maintenance on their infra (bee node upgrades, or anything else causing downtime of bee node).

In order for them to gracefully perform maintenance bee node needs to have mechanism for pausing storage incentives game. Additionally node operator would need to be able to know when game has paused.

**Solution 1** (mechanism just for storage incentives)
- Add API endpoint for pausing and resuming storage incentives
- Extend `GET redistributionstate` endpoint with information about current running status of storage incentives

**Solution 2** (scaled solution for any component of bee)
Assuming that node operators need to gracefully terminate their nodes, it might be worthwhile to implement mechanism which will not only stop storage incentives, but also other components/features of node. Compared to approach in solution 1, this solution should be generalized. 
 
- Add API endpoint for gracefully stop node, this endpoint should stop all components of node
  - endpoint example `/POST /api/node/stop`
  - for beginning this endpoint should only pause storage incentives agent
  - later, this endpoint could be used to gracefully end other components of node:
     - since node is about to be terminated node could be marked as scheduled for termination so other nodes could take up it's spot in the neighborhood?
     - stop serving any new requests for push sync protocol ?
  - possibly terminate node process after everything has stopped? have this feature controlled via request?

nikipapadatou:
@mrekucci 
* Make sure there is sufficient logging about rounds. To be improved if needed.
* Document that upgrading nodes should be outside the reveal phase.

istae:
waiting on noah to finish up the docs before we can close this

NoahMaizels:
Added warnings to docs explaining that nodes should not be shut down during a round the node is playing or else they can get frozen:

https://docs.ethswarm.org/docs/bee/working-with-bee/upgrading-bee
https://docs.ethswarm.org/docs/bee/working-with-bee/staking#check-redistribution-status

PR #4098, opened 2023-05-23
feat: graceful node shutdown 

vladopajic:
### Checklist

### Description

As part of https://github.com/ethersphere/bee/issues/3960, this PR adds mechanism for gracefully halting bee node.

### DIFF
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -120,8 +120,8 @@ func (c *command) initStartCmd() (err error) {
 					// block main goroutine until it is interrupted or stopped
 					select {
 					case <-ctx.Done():
-					case <-beeNode.Load().(*node.Bee).SyncingStopped():
-						logger.Debug("syncing has stopped")
+					case <-beeNode.Load().(*node.Bee).ShutdownSigC():
+						logger.Info("received shutdown signal from node")
 					}
 
 					logger.Info("shutting down...")
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -880,6 +880,19 @@ paths:
         default:
           description: Default response
 
+  "/shutdown":
+    post:
+      summary: Schedules graceful shutdown of the node.
+      description: "This is idempotent endpoint which schedules graceful shutdown of the node.\n\n
+        Graceful shutdown procedure will terminate node process after all internal components of the node have gracefully terminated. This procedure will mindfully stop 
+        participating in storage incentives whereby avoiding freezing or slashing staked funds."
+      tags:
+        - Shutdown
+      responses:
+        "200":
+          description: Indicates that node graceful shutdown procedure was initiated.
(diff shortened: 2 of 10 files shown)


mrekucci:
What is the benefit that outweighs the security risk of an adversary shutting down a node remotely via the HTTP API? Why can't this be done by just sending an OS signal to the running process?

vladopajic:
> What is the benefit that outweighs the security risk of an adversary shutting down a node remotely via the HTTP API? Why can't this be done by just sending an OS signal to the running process?

If a malicious individual gains access to a Bee node's API, they would probably not be concerned about the `/shutdown` endpoint. Assuming this individual is malicious, their goal would be to maximize harm to the node, including its funds and data. They have better options than shutting down the node in a graceful manner. These options may include:

  - Purchasing stamps and attaching gibberish data to them.
  - Launching a DDoS attack on the node, leading to high memory usage and potential data corruption upon forceful termination by the operating system.
   - Unstaking funds from the node, causing financial losses and removing the node from participation in the Schelling game. The attacker may prefer to keep the node alive to deceive the node operator into thinking everything is normal.

Based on the above, it appears that the `/shutdown` endpoint does not introduce any new risks. Please correct me if I have missed anything.

Shutting down a node on `SIGTERM` has not been considered an ideal approach, in my biased opinion, as it can potentially cause harm. Assuming a node has this feature, the node operator may attempt to terminate the node. However, if the node needs to wait for the Schelling game to finish, the operator might forcefully terminate (kill) the node by repeating the termination command until it shuts down. The usage of `SIGTERM` can vary depending on the user/ programmer, but typically a program should aim to end execution as soon as possible, allowing enough time for graceful closure of resources.

mrekucci:
> Based on the above, it appears that the `/shutdown` endpoint does not introduce any new risks. Please correct me if I have missed anything.

The potential attack vector is quite significant, if node operators do not properly secure their HTTP endpoints, then a simple script can potentially take down a large portion of the network.

> Shutting down a node on `SIGTERM` has not been considered an ideal approach, in my biased opinion, as it can potentially cause harm. Assuming a node has this feature, the node operator may attempt to terminate the node. However, if the node needs to wait for the Schelling game to finish, the operator might forcefully terminate (kill) the node by repeating the termination command until it shuts down. The usage of `SIGTERM` can vary depending on the user/ programmer, but typically a program should aim to end execution as soon as possible, allowing enough time for graceful closure of resources.

It can be mapped to another signal.

istae:
I agree with @mrekucci here.
On top of that, the general recommendation to node operators is to not shutdown their nodes during reveal or claims phases, which is provided in the redistribution status endpoint already.

aloknerurkar:
On top of this, even UX-wise this doesnt make sense. After doing a shutdown, the API is gone. So in order to start the node again the operator has to anyway access the CLI. Also, the risk of someone shutting down majority of the network is too big of a risk for no reward.

vladopajic:
Got it. I will open new PR with implementation using OS signal.

PR #4107, opened 2023-05-24
feat: export and import pinning db

acha-bill:
### Checklist

### Description

Add subcommands to export and import pinning db.

### Related Issue (Optional)
#3907

### DIFF
--- a/pkg/storer/internal/pinning/pinning.go
+++ b/pkg/storer/internal/pinning/pinning.go
@@ -174,13 +174,9 @@ func (p pinChunkItem) String() string {
 // that are part of this collection. The root pin is only updated on successful close of this
 // Putter.
 // If no UUID is specified in the options, a new UUID will be generated for this collection.
-func NewCollection(st internal.Storage, opts *swarm.CollectionOptions) internal.PutterCloserWithReference {
-	collectionUUID := newUUID()
-	if opts != nil && len(opts.UUID) > 0 {
-		collectionUUID = opts.UUID
-	}
+func NewCollection(st internal.Storage) internal.PutterCloserWithReference {
 	return &collectionPutter{
-		collection: &pinCollectionItem{UUID: collectionUUID},
+		collection: &pinCollectionItem{UUID: newUUID()},
 		st:         st,
 	}
 }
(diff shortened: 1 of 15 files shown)


pkg/storer/internal/pinning/pinning.go
```diff
@@ -158,12 +158,16 @@ func (p pinChunkItem) String() string {
 
 // NewCollection returns a putter wrapped around the passed storage.
 // The putter will add the chunk to Chunk store if it doesnt exists within this collection.
-// It will create a new UUID for the collection which can be used to iterate on all the chunks
 // that are part of this collection. The root pin is only updated on successful close of this
 // Putter.
-func NewCollection(st internal.Storage) internal.PutterCloserWithReference {
+// If no UUID is specified in the options, a new UUID will be generated for this collection.
+func NewCollection(st internal.Storage, opts *swarm.CollectionOptions) internal.PutterCloserWithReference {
```
aloknerurkar, line 164: The UUID is completely internal to the pinning store. We dont expose it anywhere. In case of export, then import, it is not necessary to reuse the UUID. The UUID is just a unique identifier that ties the root collection Addr to all the chunks it has. 

So remove all these changes related to adding these options. Anyone outside the pinning store doesnt know about this UUID.

pkg/storer/internal/pinning/pinning.go
```diff
@@ -314,6 +318,20 @@ func DeletePin(ctx context.Context, st internal.Storage, root swarm.Address) err
 	return nil
 }
 
+func IterateCollectionChunks(st storage.Store, root swarm.Address, iterateFn func(swarm.Address) (bool, error)) error {
+	err := st.Iterate(storage.Query{
+		Factory:      func() storage.Item { return &pinChunkItem{UUID: root.Bytes()} },
```
aloknerurkar, line 323: Not sure how this is working. The root is not the UUID. You need to read the UUID first.

Check export.go file in this pkg. It already has IterateCollection function. You can just move it here. Previously no one outside the pkg was using it. But now we do.

aloknerurkar (changes requested):

aloknerurkar (approved):

PR #4113, opened 2023-05-27
fix(storer): tags bug and encrypted ref handling

aloknerurkar:
### Checklist

### Description
1. Tags were not being cleared from dirty tag list on failure. Also, allows subscribe push worker to go ahead if it encounters uncommitted tags.
2. Handle encrypted references in upload. For pinning we will store the root encrypted reference. In case of upload we store the swarm hash and not the encryption key. This is not required right now.


mrekucci (approved):

mrekucci (commented):
`TestPinStore` tests are failing.

PR #4115, opened 2023-05-29
fix(puller): added new mtx locking to interval state storing

istae:
### Checklist

### Description
There is a potential for a data race in the way the synced intervals are stored between historical and live syncing so we added a mtx lock.

### DIFF
--- a/pkg/puller/puller.go
+++ b/pkg/puller/puller.go
@@ -153,6 +153,9 @@ func (p *Puller) manage(ctx context.Context, warmupDur time.Duration) {
 		p.recalcPeers(ctx, newRadius)
 	}
 
+	tick := time.NewTicker(recalcPeersDur)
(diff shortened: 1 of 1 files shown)


pkg/puller/puller.go
```diff
@@ -149,22 +151,17 @@ func (p *Puller) manage(ctx context.Context, warmupDur time.Duration) {
...
 	for {
+
+		onChange()
+
 		select {
 		case <-ctx.Done():
 			return
-		case <-tick.C:
-			onChange()
+		case <-time.After(recalcPeersDur):
```
aloknerurkar, line 163: ticker is much more efficient than time.After. Although this is just 1 routine, it might do new allocation on every run of the loop. This can be easily avoided.

Any reason why this was changed?

vladopajic (approved):

PR #4118, opened 2023-05-29
feat: salud to record peer latency

istae:
### Checklist

### Description
Since salud acquires the status msg from every peer (reachable or not), we can use this mechanism to record peer latency instead of kademlia's ping.

### DIFF
--- a/pkg/node/bootstrap.go
+++ b/pkg/node/bootstrap.go
@@ -126,8 +126,18 @@ func bootstrapNode(
 	}
 	b.hiveCloser = hive
 
+<<<<<<< HEAD
 	kad, err := kademlia.New(swarmAddress, addressbook, hive, p2ps, &noopPinger{}, logger,
 		kademlia.Options{Bootnodes: bootnodes, BootnodeMode: o.BootnodeMode, StaticNodes: o.StaticNodes, DataDir: o.DataDir})
+=======
+	metricsDB, err := shed.NewDBWrap(stateStore.DB())
+	if err != nil {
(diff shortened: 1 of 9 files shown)


aloknerurkar (commented):
I think the chainsyncer is more difficult to beat than looking at batch commitment. By beat, I mean if there is a malicious actor, he can easily get the batch-commitment value from other nodes and send that across to the nodes connected to him. Also he can keep using it till any new batch is bought. I think with chainsync, it is more of challenge-response, so peer can ask for a single value from a set of them, which is more difficult.

I dont mind this change, but we should do it when we have slashing of stakes. This would be far greater punishment than blocklisting.

0xCandy:
Is the 'Check' from time to time？ Because my VPN or VPS will change my server IP occasionally. Sometimes a bad server can cause my network to disconnect, and then a well-performed server can re-establish my network.

Besides， I find that when every time I start my bee full node， it can connect a lot of peers. however, the number of my peers will decrease over time and finally decays to just a few dozen nodes. I am wondering whether it is caused by the potential 'Checking' issues.

istae:
@0xCandy the check does not disconnected from peers. 
a few dozen nodes signals that your node is having serious network issues. on average, the expected peer count is around 150.

0xCandy:
> @0xCandy the check does not disconnect from peers. a few dozen peers signals that your node is having serious network issues. on average, the expected peer count is around 150.

I mean, if my network is bad and peers disconnected, then periodic re-check may help recover my peers.

PR #4135, opened 2023-06-06
fix(reserve): best attempt to remove all related items

istae:
### Checklist

### Description
It was observed during testing that a dirty shutdown during a running eviction can cause unwanted entries to be left behind and cause future evictions to fail. 
so we change the `removeChunk` to do a best attempt to remove all items and not return an error early.

### DIFF
--- a/pkg/postage/stampissuer.go
+++ b/pkg/postage/stampissuer.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"math/big"
 	"path"
+	"sync"
 	"time"
 
 	storage "github.com/ethersphere/bee/pkg/storage"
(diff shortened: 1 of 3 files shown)


aloknerurkar:
So I dont mind this change, but lets not merge it now. The reason being we would like to see if this is happening in the testing. The last time we found these errors, we found a synchronization issue which was major. In the ideal case, this should not happen. Also, @mrekucci is working on the recovery mechanism, which also needs to be tested. We should make sure that after recovery we dont see this anymore. This could cause incorrect reserveSize calculation on restart everytime if its not fixed.

aloknerurkar (approved):

PR #4141, opened 2023-06-08
chore(api): add rchash

vladopajic:
### Checklist

### Description
Adding `rchash` endpint which will be used for testing.

### DIFF
--- a/pkg/api/router.go
+++ b/pkg/api/router.go
@@ -336,6 +336,12 @@ func (s *Service) mountAPI() {
 		web.FinalHandlerFunc(s.healthHandler),
 	))
 
+	handle("/rchash/{depth}/{anchor1}/{anchor2}", web.ChainHandlers(
+		web.FinalHandler(jsonhttp.MethodHandler{
+			"GET": http.HandlerFunc(s.rchash),
+		}),
+	))
+
 	if s.Restricted {
 		handle("/auth", jsonhttp.MethodHandler{
 			"POST": web.ChainHandlers(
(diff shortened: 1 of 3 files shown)


notanatol:
my impression was that stuff that we use for testing should go into `/debug`?

vladopajic:
@notanatol  yeah, but sooner or later all endpoints should be under same API, because `debug` API should be merged with main, right? Having this in mind this endpoint was added to main API.

notanatol (approved):

pkg/api/router.go
```diff
@@ -336,6 +336,12 @@ func (s *Service) mountAPI() {
 		web.FinalHandlerFunc(s.healthHandler),
 	))
 
+	handle("/rchash/{depth}/{anchor1}/{anchor2}", web.ChainHandlers(
```
notanatol, line 339: don't forget to add this endpoint to the list in `auth.go`

vladopajic: @notanatol what should be right policy for this endpoint?

pkg/storageincentives/agent.go
```diff
@@ -542,6 +542,42 @@ func (a *Agent) Status() (*Status, error) {
...
+type SampleWithProofs struct {
+	Items    []storer.SampleItem
+	Hash     swarm.Address
+	Duration time.Duration
+}
+
+func (a *Agent) SampleWithProofs(
+	ctx context.Context,
+	anchor1,
+	anchor2 []byte,
```
aloknerurkar, line 554: Seems like we are not using this?

aloknerurkar (approved):

PR #4149, opened 2023-06-11
fix(direct-upload): use same concurrency as deferred upload

istae:
### Checklist

### Description
PR makes direct upload concurrency level the same as differed uploads.
multiple parallel uploads are limited by the same worker so the node remains protected against goroutine pileup.

### DIFF
--- a/pkg/storer/netstore.go
+++ b/pkg/storer/netstore.go
@@ -74,12 +74,12 @@ func (db *DB) Download(cache bool) storage.Getter {
 						select {
 						case <-ctx.Done():
 						case <-db.quit:
-						case db.bgCacheWorkers <- struct{}{}:
-							db.bgCacheWorkersWg.Add(1)
+						case db.bgCacheLimiter <- struct{}{}:
+							db.bgCacheLimiterWg.Add(1)
 							go func() {
 								defer func() {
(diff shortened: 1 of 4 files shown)


pkg/storer/netstore.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	directUploadWorkers = 16
```
aloknerurkar, line 18: This is not the same. We can have multiple calls to DirectUpload. It creates 16 workers for each upload.

aloknerurkar (changes requested):

pkg/storer/netstore.go
istae: Does 16 workers mean 16 chunk push attempts in parallel?

aloknerurkar: Yes 16 push attempts in parallel per upload. So the call to DirectUpload returns a putter which can use upto 16 routines to push in parallel.

istae: in that case, we do not actually need this limiter. the pusher is already limited.

istae:
imo it'd would been ok to simply require a batch ID starting with the new release

aloknerurkar (approved):

Issue #4129, opened 2023-06-03
Move stamp issuer data to stamperstore


aloknerurkar:
Move the stamp issuer data to stamper store. Ensure the data is only saved when the session is completed successfully.

PR #4150, opened 2023-06-12
feat: move stamp issuer to stamperstore

acha-bill:
### Checklist

### Description

Move stamperIssuer data from statestore to stamperstore

### Related Issue (Optional)
#4129

### DIFF
--- a/pkg/api/chunk_stream.go
+++ b/pkg/api/chunk_stream.go
@@ -74,7 +74,7 @@ func (s *Service) chunkUploadStreamHandler(w http.ResponseWriter, r *http.Reques
 		case errors.Is(err, errUnsupportedDevNodeOperation):
 			jsonhttp.BadRequest(w, errUnsupportedDevNodeOperation)
 		case errors.Is(err, postage.ErrBatchInUse):
-			jsonhttp.Conflict(w, "batch is in use")
+			jsonhttp.BadRequest(w, "batch is in use by another upload process")
 		default:
 			jsonhttp.BadRequest(w, nil)
 		}
--- a/pkg/api/chunk_stream.go
+++ b/pkg/api/chunk_stream.go
@@ -74,7 +74,7 @@ func (s *Service) chunkUploadStreamHandler(w http.ResponseWriter, r *http.Reques
 		case errors.Is(err, errUnsupportedDevNodeOperation):
 			jsonhttp.BadRequest(w, errUnsupportedDevNodeOperation)
 		case errors.Is(err, postage.ErrBatchInUse):
-			jsonhttp.BadRequest(w, "batch is in use by another upload process")
+			jsonhttp.BadRequest(w, postage.ErrBatchInUse)
 		default:
 			jsonhttp.BadRequest(w, nil)
 		}
--- a/pkg/api/chunk_stream.go
+++ b/pkg/api/chunk_stream.go
@@ -73,6 +73,8 @@ func (s *Service) chunkUploadStreamHandler(w http.ResponseWriter, r *http.Reques
 			jsonhttp.BadRequest(w, "invalid batch id")
 		case errors.Is(err, errUnsupportedDevNodeOperation):
 			jsonhttp.BadRequest(w, errUnsupportedDevNodeOperation)
+		case errors.Is(err, postage.ErrBatchInUse):
+			jsonhttp.Conflict(w, "batch is in use")
 		default:
 			jsonhttp.BadRequest(w, nil)
 		}
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -184,8 +184,8 @@ func (s *Service) postageGetStampsHandler(w http.ResponseWriter, r *http.Request
 	resp := postageStampsResponse{}
 	stampIssuers, err := s.post.StampIssuers()
 	if err != nil {
-		logger.Debug("get stamps: get issuers failed", "error", err)
-		logger.Error(nil, "get stamps: get issuers failed")
+		logger.Debug("get issuers failed", "error", err)
+		logger.Error(nil, "get issuers failed")
 		jsonhttp.InternalServerError(w, "cannot get issuers")
 		return
 	}
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -182,8 +182,15 @@ func (s *Service) postageGetStampsHandler(w http.ResponseWriter, r *http.Request
 	}
 
 	resp := postageStampsResponse{}
-	resp.Stamps = make([]postageStampResponse, 0, len(s.post.StampIssuers()))
-	for _, v := range s.post.StampIssuers() {
+	stampIssuers, err := s.post.StampIssuers()
+	if err != nil {
+		logger.Debug("get stamps: get issuers failed", "error", err)
+		logger.Error(nil, "get stamps: get issuers failed")
+		jsonhttp.InternalServerError(w, "cannot get issuers")
+		return
+	}
+	resp.Stamps = make([]postageStampResponse, 0, len(stampIssuers))
+	for _, v := range stampIssuers {
 		exists, err := s.batchStore.Exists(v.ID())
 		if err != nil {
 			logger.Debug("get stamp issuer: check batch failed", "batch_id", hex.EncodeToString(v.ID()), "error", err)
@@ -318,7 +325,14 @@ func (s *Service) postageGetStampHandler(w http.ResponseWriter, r *http.Request)
 	hexBatchID := hex.EncodeToString(paths.BatchID)
 
 	var issuer *postage.StampIssuer
-	for _, stampIssuer := range s.post.StampIssuers() {
+	stampIssuers, err := s.post.StampIssuers()
+	if err != nil {
+		logger.Debug("get stamp issuer: get issuers failed", "batch_id", hexBatchID, "error", err)
+		logger.Error(nil, "get stamp issuer: get issuers failed")
+		jsonhttp.InternalServerError(w, "get issuers failed")
+		return
+	}
+	for _, stampIssuer := range stampIssuers {
 		if bytes.Equal(paths.BatchID, stampIssuer.ID()) {
 			issuer = stampIssuer
 			break
--- a/pkg/node/statestore.go
+++ b/pkg/node/statestore.go
@@ -98,9 +98,9 @@ func setOverlayNonce(s storage.StateStorer, overlayNonce []byte) error {
 }
 
 func migrateStamperData(stateStore storage.StateStorer, stamperStore storage.Store) error {
-	var keys [][]byte
+	var keys []string
 	err := stateStore.Iterate("postage", func(key, value []byte) (bool, error) {
-		keys = append(keys, key)
+		keys = append(keys, string(key))
 		st := &postage.StampIssuer{}
 		if err := st.UnmarshalBinary(value); err != nil {
 			return false, err
@@ -117,7 +117,7 @@ func migrateStamperData(stateStore storage.StateStorer, stamperStore storage.Sto
 	}
 
 	for _, key := range keys {
-		if err = stateStore.Delete(string(key)); err != nil {
+		if err = stateStore.Delete(key); err != nil {
 			return err
 		}
 	}
--- a/pkg/node/statestore.go
+++ b/pkg/node/statestore.go
@@ -9,11 +9,11 @@ import (
 	"fmt"
 	"path/filepath"
 
-	"github.com/ethersphere/bee/pkg/statestore/storeadapter"
-	"github.com/ethersphere/bee/pkg/storage/leveldbstore"
-
 	"github.com/ethersphere/bee/pkg/log"
+	"github.com/ethersphere/bee/pkg/postage"
+	"github.com/ethersphere/bee/pkg/statestore/storeadapter"
 	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/storage/leveldbstore"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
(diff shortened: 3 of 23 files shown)


pkg/postage/service.go
```diff
@@ -44,27 +42,29 @@ type Service interface {
 // stores the active batches.
 type service struct {
 	lock         sync.Mutex
-	store        storage.StateStorer
+	store        storage.Store
 	postageStore Storer
 	chainID      int64
 	issuers      []*StampIssuer
```
aloknerurkar, line 48: We should remove this in-mem state of StampIssuer completely. The problem is due to the nature of the operations in API, if we dont commit the transaction to storer, we should not be updating the stamp indexes into stamper store. The `save` function is only called on successful completion of the upload.

In order to have proper synchronization, every time we do a GetStampIssuer, we should read the state from the stamper store. Also, all other updates like expiry etc should also be done on the store directly. Also, we should only allow access to a particular stampIssuer to 1 routine as multiple routines can try to use the same indexes if this is no longer in memory. This hit in performance is okay as we only call GetStampIssuer once during upload.

pkg/postage/stampissuer.go
```diff
@@ -262,6 +262,57 @@ func (si *StampIssuer) SetExpired(e bool) {
...
+	return &stampIssuerItem{
+		data: stampIssuerData{
+			BatchID: ID,
+		},
+	}
+}
+
+// ID is the batch ID.
+func (s *stampIssuerItem) ID() string {
+	return string(s.data.BatchID)
```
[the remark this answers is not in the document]
acha-bill: `s.data` is not a pointer

pkg/postage/service.go
```diff
@@ -44,58 +41,26 @@ type Service interface {
...
-
-	for _, v := range ps.issuers {
-		if bytes.Equal(st.data.BatchID, v.data.BatchID) {
-			return false
-		}
-	}
-	ps.issuers = append(ps.issuers, st)
-
-	return true
+	return ps.save(st)
```
aloknerurkar, line 63: Here we need to check if the stamp issuer is already saved. Otherwise this will overwrite the stampIssuer.Data

aloknerurkar (commented):
Changes look good. There is just one additional thing left to do. We need to write a small migration for the old stamper data which is in statestore. This will need to be a highly specific migration and I think it might be in the node pkg itself where you can call a function after creating the stamper store, pass it the statestore and then do the migration.

pkg/postage/service.go
```diff
@@ -165,24 +143,35 @@ func (ps *service) IssuerUsable(st *StampIssuer) bool {
...
-			}
-			return st, func() error {
-				return ps.save(st)
-			}, nil
-		}
+
+	if _, ok := ps.issuerLocks[string(batchID)]; !ok {
+		ps.issuerLocks[string(batchID)] = &sync.Mutex{}
+	}
+	ps.issuerLocks[string(batchID)].Lock()
```
aloknerurkar, line 150: So the problem with using sync.Mutex here is that it is a blocking call. This will block the uploader thread. If we want to make this a blocking call we need to use the context as an argument and then use some sort of context-based locking for eg x/sync.semaphore and pass the http.Request context so that on cancellation from user it exits cleanly. This will complicate things as you will be waiting while holding the service.Lock which can cause other problems.

I think its better to have an empty struct in the map just to signal that the batch ID is in use and if it exists we return error, something is ErrBatchInUse and we can say in docs that currently only 1 upload is allowed per batch.

pkg/postage/service.go
```diff
@@ -165,24 +143,35 @@ func (ps *service) IssuerUsable(st *StampIssuer) bool {
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if !ps.IssuerUsable(item.Issuer) {
+		return nil, nil, ErrNotUsable
 	}
-	return nil, nil, ErrNotFound
+	return item.Issuer, func() error {
+		defer ps.issuerLocks[string(batchID)].Unlock()
```
aloknerurkar, line 165: If you use the empty struct, I think you will need the lock here as save doesnt acquire the lock.

pkg/postage/service.go
aloknerurkar: This still needs to be resolved right?

Add can be called from 2 places, so if we already have the stamp issuer created we should not update it. For example, the first time Add is called is when the node creates a batch. Then, it can receive a HandleCreate event as well which may be delayed or could have got on restart or resync of the node. If we had already issued some indexes then the update will overwrite them.

pkg/postage/service.go
```diff
@@ -165,24 +145,36 @@ func (ps *service) IssuerUsable(st *StampIssuer) bool {
...
-			if !ps.IssuerUsable(st) {
-				return nil, nil, ErrNotUsable
-			}
-			return st, func() error {
-				return ps.save(st)
-			}, nil
-		}
+
+	if _, ok := ps.issuersInUse[string(batchID)]; ok {
+		return nil, nil, ErrBatchInUse
```
aloknerurkar, line 150: We need to also handle this error on the API. In API there are multiple instances of newStamperPutter calls which then handle different types of error. This error needs to be handled separately by sending a detailed error message back to the user that the request failed because another process is using this batch.

aloknerurkar (approved):

pkg/node/statestore.go
```diff
@@ -87,3 +96,30 @@ func overlayNonceExists(s storage.StateStorer) ([]byte, bool, error) {
 func setOverlayNonce(s storage.StateStorer, overlayNonce []byte) error {
 	return s.Put(OverlayNonce, overlayNonce)
 }
+
+func migrateStamperData(stateStore storage.StateStorer, stamperStore storage.Store) error {
```
istae, line 100: this will run everytime the node starts up, right?
Maybe this should be part of some migration logic that runs only once after upgrading to the new release?

pkg/postage/service.go
```diff
@@ -191,49 +191,43 @@ func (ps *service) save(st *StampIssuer) error {
...
-	}
-	return nil
+	return ps.store.Iterate(
+		storage.Query{
+			Factory: func() storage.Item { return new(StampIssuerItem) },
+		}, func(result storage.Result) (bool, error) {
+			issuer := result.Entry.(*StampIssuerItem).Issuer
+			exists, err := ps.postageStore.Exists(issuer.ID())
+			if err != nil {
+				return true, err
```
istae, line 224: instead of returning, maybe log the error and continue with the iterator

pkg/postage/service.go
```diff
@@ -44,58 +43,36 @@ type Service interface {
...
 
 // NewService constructs a new Service.
-func NewService(store storage.StateStorer, postageStore Storer, chainID int64) (Service, error) {
-	s := &service{
+func NewService(store storage.Store, postageStore Storer, chainID int64) Service {
+	return &service{
 		store:        store,
 		postageStore: postageStore,
 		chainID:      chainID,
+		issuersInUse: make(map[string]any),
```
istae, line 58: what is the purpose of the map? I think we can get rid of it.

pkg/postage/service.go
```diff
@@ -165,24 +153,36 @@ func (ps *service) IssuerUsable(st *StampIssuer) bool {
...
+	}, nil
 }
 
-// save persists the specified stamp issuer to the statestore.
+// save persists the specified stamp issuer to the stamperstore.
 func (ps *service) save(st *StampIssuer) error {
-	st.bucketMu.Lock()
-	defer st.bucketMu.Unlock()
-	if err := ps.store.Put(ps.keyForIndex(st.data.BatchID), st); err != nil {
+	if err := ps.store.Put(&StampIssuerItem{
```
istae, line 183: can multiple parallel uploads using the same stamp issuer cause a data race when saving the issuer?

istae (commented):
nice PR, some comments.

pkg/postage/service.go
acha-bill: We use it to track which issuers are being used

pkg/postage/service.go
acha-bill: That's what `issuersInUse` map solves. It prevents multiple routines from using the same issuer.

pkg/node/statestore.go
acha-bill: The idea was that after the migration, future runs would be noop since the stamps would have been migrated to stamperstore and deleted from statestore. Then we can remove the code later on.

How can we tie this migration to the new release?

pkg/postage/service.go
istae: @aloknerurkar is this ok?

pkg/node/statestore.go
istae: https://github.com/ethersphere/bee/pull/4154 takes care of this, right @aloknerurkar ?

aloknerurkar: @istae So right now the problem is the stampissuer data is in statestore. But going ahead it will be in stamper store. This is why the statestore migration doesnt remove these entries. Instead this separate migration is required. This migration will happen after the statestore migration. So the entries will be ported to new statestore then moved to stamper store and deleted.

Only the first time it starts there will be postage related entries in the statestore. So the migration will be a no-op after that.

pkg/postage/service.go
aloknerurkar: @istae So the problem is right now we have the transactions in the storer. But when we use the stampIssuer the data is in a different storage layer. As a result its very difficult to maintain consistency. Even if we don't call the Save function after using the issuer, the indexes that were assigned in-memory cannot be recovered if the transaction or the upload fails for eg. Also, on top of that, if we share it across upload sessions, it is impossible to recover indexes.

So we thought of keeping it simple and only allowing one upload per stamp at a time. I think the risk to reward is not that great here as if we have indexes that are not being used then the stamp is being wasted.

aloknerurkar: Because we will no longer have in in-mem if the thing fails, the next upload will read the old state, so it wont lose anything.

pkg/api/chunk_stream.go
```diff
@@ -67,6 +67,8 @@ func (s *Service) chunkUploadStreamHandler(w http.ResponseWriter, r *http.Reques
 			jsonhttp.BadRequest(w, "invalid batch id")
 		case errors.Is(err, errUnsupportedDevNodeOperation):
 			jsonhttp.BadRequest(w, errUnsupportedDevNodeOperation)
+		case errors.Is(err, postage.ErrBatchInUse):
+			jsonhttp.BadRequest(w, "batch is in use by another upload process")
```
mrekucci, line 71: This message is repeated, make it an error, similar to `errUnsupportedDevNodeOperation`.

pkg/api/postage.go
```diff
@@ -182,8 +182,15 @@ func (s *Service) postageGetStampsHandler(w http.ResponseWriter, r *http.Request
 	}
 
 	resp := postageStampsResponse{}
-	resp.Stamps = make([]postageStampResponse, 0, len(s.post.StampIssuers()))
-	for _, v := range s.post.StampIssuers() {
+	stampIssuers, err := s.post.StampIssuers()
+	if err != nil {
+		logger.Debug("get stamps: get issuers failed", "error", err)
```
mrekucci, line 187: The `get stamps:` prefix is not needed because the logger already inserts it in the log message. The same is below.

pkg/node/statestore.go
mrekucci: Agree with @istae Is there a way this can be done as part of the migration logic?

pkg/node/statestore.go
```diff
@@ -87,3 +96,30 @@ func overlayNonceExists(s storage.StateStorer) ([]byte, bool, error) {
...
+	var keys [][]byte
```
mrekucci, line 101: Since the key below is cast to type `string`, I recommend using `var keys []string` instead and casting the key to type `string` on append.

pkg/node/statestore_test.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		si := postage.NewStampIssuer("", "", bID, big.NewInt(3), 11, 10, 1000, true)
+		err = stateStore.Put(fmt.Sprintf("postage%s", string(si.ID())), si)
+		if err != nil {
+			t.Fatal(err)
+		}
+		ids[string(si.ID())] = 0
+	}
+
+	// init stamper store should migrate 10 stamps from state store
```
mrekucci, line 43: Instead of this comment, I would recommend making a proper description in `t.Fatal(err)`.

pkg/node/statestore_test.go
```diff
@@ -0,0 +1,77 @@
...
+	stamperStore, err := InitStamperStore(log.Noop, dataDir, stateStore)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// check stamper store has 10 stamps
```
mrekucci, line 49: The same here, make a meaningful error message instead.

pkg/node/statestore_test.go
```diff
@@ -0,0 +1,77 @@
...
+		if v != 1 {
+			nMissing++
+		}
+	}
+	if nMissing > 0 {
+		t.Fatalf("missing %d stamps", nMissing)
+	}
+
+	t.Cleanup(func() {
+		_ = stateStore.Close()
```
mrekucci, line 73: Why are these errors ignored? Also why there is need for a sleep?

pkg/postage/batchservice/batchservice_test.go
```diff
@@ -54,12 +54,14 @@ func (m *mockBatchListener) HandleCreate(b *postage.Batch, topUpAmount *big.Int)
 	return nil
 }
 
-func (m *mockBatchListener) HandleTopUp(_ []byte, _ *big.Int) {
+func (m *mockBatchListener) HandleTopUp(id []byte, newBalance *big.Int) error {
```
mrekucci, line 57: If the method params are not used inside its body they should be renamed to `_`.

pkg/node/statestore.go
aloknerurkar: I dont think so, the migration logic is localized to the store itself. Here we need to take the entries from a completely different store. Which is why this has to be specific I think.

pkg/node/statestore_test.go
acha-bill: The sleep gives levelDB some time to shut down it's goroutines.
For some reason, levelDB takes longer to close the channels and this results in gourtine leak. 

```
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 51 in state select, with github.com/syndtr/goleveldb/leveldb.(*DB).mpoolDrain on top of the stack:
goroutine 51 [select]:
github.com/syndtr/goleveldb/leveldb.(*DB).mpoolDrain(0x14000532c40)
	/Users/macbookpro/go/pkg/mod/github.com/syndtr/goleveldb@v1.0.1-0.20210819022825-2ae1ddf74ef7/leveldb/db_state.go:110 +0xe4
created by github.com/syndtr/goleveldb/leveldb.openDB
	/Users/macbookpro/go/pkg/mod/github.com/syndtr/goleveldb@v1.0.1-0.20210819022825-2ae1ddf74ef7/leveldb/db.go:149 +0x424
```

The simpliest fix was to sleep

aloknerurkar: So this is not the right fix actually. This might work on your machine but it may not always work on CI.

I think you will have to add some ignore calls for these leaks. Look at the storage pkg for examples.

mrekucci (approved):

PR #4153, opened 2023-06-12
fix: stewardship uses new or old batch

### LINKED ISSUE #4128: Stewardship PUT API requires stamp
With the new storage layer, the download of chunks no longer retrieves the stamps for the chunks. This is because when downloading content, we dont really care about stamps. The stamps could be expired but the chunks could be part of cache of some nodes.

The only place where we use this is in `PUT /stewardship`. The stewardship PUT is used to reupload content on the network. This is mainly useful if you have pinned content that may not be retrievable for some reason. So now, in order to reupload, user will have to provide a stamp to use. The node remembers stamps now, so if user uses the same stamp and it is valid, all the chunks should be stamped with the same indexes and reuploaded.

The user can also use a different stamp, in which case new indexes will be used for the chunks. This allows more flexibility for users as well.

aloknerurkar:
### Checklist

### Description
Fixes #4128

### DIFF
--- a/pkg/api/stewardship_test.go
+++ b/pkg/api/stewardship_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/ethersphere/bee/pkg/jsonhttp"
 	"github.com/ethersphere/bee/pkg/jsonhttp/jsonhttptest"
 	"github.com/ethersphere/bee/pkg/log"
+	mockpost "github.com/ethersphere/bee/pkg/postage/mock"
 	"github.com/ethersphere/bee/pkg/steward/mock"
 	mockstorer "github.com/ethersphere/bee/pkg/storer/mock"
 	"github.com/ethersphere/bee/pkg/swarm"
@@ -30,6 +31,7 @@ func TestStewardship(t *testing.T) {
 		Storer:  storer,
 		Logger:  logger,
 		Steward: stewardMock,
+		Post:    mockpost.New(mockpost.WithAcceptAll()),
 	})
 
 	t.Run("re-upload", func(t *testing.T) {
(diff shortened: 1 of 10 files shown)


istae (approved):
imo it'd would been ok to simply require a batch ID starting with the new release

pkg/api/stewardship_test.go
```diff
@@ -92,19 +96,28 @@ func Test_stewardshipHandlers_invalidInputs(t *testing.T) {
...
 	}}
 
 	for _, method := range []string{http.MethodGet, http.MethodPut} {
 		method := method
-		for _, tc := range tests {
+		for idx, tc := range tests {
 			tc := tc
-			t.Run(method+" "+tc.name, func(t *testing.T) {
-				t.Parallel()
+			if idx < 2 || method == http.MethodPut {
```
vladopajic, line 112: what is the reason for having this `if` block on line 112?

pkg/steward/steward.go
```diff
@@ -58,7 +59,12 @@ func (s *steward) Reupload(ctx context.Context, root swarm.Address) error {
 			return err
 		}
 
-		return uploaderSession.Put(ctx, c)
+		stamp, err := stamper.Stamp(c.Address())
+		if err != nil {
+			return fmt.Errorf("stamping chunk %s: %w", c.Address().String(), err)
```
mrekucci, line 64: No need for `.String()`.

mrekucci (commented):
These changes should also be reflected in the OpenAPI specification.

pkg/api/stewardship_test.go
aloknerurkar: The test no. 3 is only applicable to MethodPut.

vladopajic: would it be possible to add additional field to test cases structs with headers that each test case supports?

aloknerurkar: I have moved it to a separate test. Dont think there is any advantage of making these changes.

aloknerurkar:
> These changes should also be reflected in the OpenAPI specification.

Will do a separate PR with all the API changes.

vladopajic (approved):

aloknerurkar:
> imo it'd would been ok to simply require a batch ID starting with the new release

I agree with you. This is what I initially planned, but if you see the comment thread on the issue, it was requested and not hard to do.

PR #4161, opened 2023-06-16
feat: tx crash recovery mechanism

### LINKED ISSUE #4085: Add recovery mechanism for txn

mrekucci:
### Checklist

### Description
In-progress transaction recovery mechanism on node crash. 

Closes #4085

### DIFF
--- a/pkg/storage/inmemstore/transaction.go
+++ b/pkg/storage/inmemstore/transaction.go
@@ -106,7 +106,8 @@ func (s *TxStore) NewTx(state *storage.TxState) storage.TxStore {
 			TxState: state,
 			Store:   s.Store,
 		},
-		revOps: storage.NewTxRevertStack(
+		revOps: storage.NewTxRevertStack[storage.Key, storage.Item](
+			new(storage.InMemTxRevertOpStore[storage.Key, storage.Item]),
 			map[storage.TxOpCode]storage.TxRevertFn[storage.Key, storage.Item]{
 				storage.PutCreateOp: func(key storage.Key, item storage.Item) error {
 					return s.Store.Delete(item)
--- a/pkg/storage/inmemstore/transaction.go
+++ b/pkg/storage/inmemstore/transaction.go
@@ -63,13 +63,23 @@ func (s *TxStore) Put(item storage.Item) error {
 
 // Delete implements the Store interface.
 func (s *TxStore) Delete(item storage.Item) error {
+	if err := s.IsDone(); err != nil {
+		return err
+	}
+
+	prev := item.Clone()
+	var reverseOp *storage.TxRevertOp[storage.Key, storage.Item]
+	if err := s.TxStoreBase.Get(prev); err == nil {
+		reverseOp = &storage.TxRevertOp[storage.Key, storage.Item]{
+			Origin:   storage.DeleteOp,
+			ObjectID: prev.String(),
+			Val:      prev,
+		}
+	}
+
 	err := s.TxStoreBase.Delete(item)
 	if err == nil {
-		err = s.revOps.Append(&storage.TxRevertOp[storage.Key, storage.Item]{
-			Origin:   storage.DeleteOp,
-			ObjectID: item.String(),
-			Val:      item,
-		})
+		err = s.revOps.Append(reverseOp)
 	}
 	return err
 }
@@ -78,7 +88,13 @@ func (s *TxStore) Delete(item storage.Item) error {
 func (s *TxStore) Commit() error {
 	defer s.release()
 
-	return s.TxState.Done()
+	if err := s.TxState.Done(); err != nil {
+		return err
+	}
+	if err := s.revOps.Clean(); err != nil {
+		return fmt.Errorf("inmemstore: unable to clean revert operations: %w", err)
+	}
+	return nil
 }
 
 // Rollback implements the Tx interface.
@@ -86,11 +102,11 @@ func (s *TxStore) Rollback() error {
 	defer s.release()
 
 	if err := s.TxStoreBase.Rollback(); err != nil {
-		return err
+		return fmt.Errorf("inmemstore: unable to rollback: %w", err)
 	}
 
 	if err := s.revOps.Revert(); err != nil {
-		return fmt.Errorf("inmemstore: unable to rollback: %w", err)
+		return fmt.Errorf("inmemstore: unable to revert operations: %w", err)
 	}
 	return nil
 }
--- a/pkg/storage/inmemstore/transaction.go
+++ b/pkg/storage/inmemstore/transaction.go
@@ -20,7 +20,7 @@ type TxStore struct {
 
 	// Bookkeeping of invasive operations executed
 	// on the Store to support rollback functionality.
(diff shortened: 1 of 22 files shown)


pkg/storer/internal/chunkstore/transaction.go
```diff
@@ -165,16 +173,27 @@ func (cs *txChunkStoreWrapper) NewTx(state *storage.TxState) storage.TxChunkStor
...
+				storage.PutOp: func(key swarm.Address, val swarm.Chunk) error {
+					return chunkStore.Delete(context.Background(), key)
+				},
+				storage.DeleteOp: func(key swarm.Address, val swarm.Chunk) error {
+					return chunkStore.Put(context.Background(), val)
+				},
+			},
+		),
 	}
 }
```
aloknerurkar, line 199: Can we remove the txSharky allocation in NewTxChunkStore? These are never used.

pkg/storer/internal/reserve/reserve.go
```diff
@@ -345,7 +345,7 @@ func removeChunk(ctx context.Context, store internal.Storage, item *batchRadiusI
 		return err
 	}
 
-	err := indexStore.Delete(&chunkBinItem{Bin: item.Bin, BinID: item.BinID})
+	err := indexStore.Delete(&chunkBinItem{Bin: item.Bin, BinID: item.BinID, Address: item.Address})
```
aloknerurkar, line 348: So I saw this error on the test cluster. The problem is we are not doing a Get in the TxStore.Delete. This will need to be fixed. Basically, the current assumption in the code is that for Delete we only need to populate the Key interface for the item. So the TxStore should do a get first and fill all the fields. Otherwise the rollback operations will populate incorrect items. For eg. in this case, it will give error for missing address, but batch ID will be nil in case of rollback.Put.

pkg/storage/leveldbstore/transaction.go
```diff
@@ -79,17 +190,15 @@ func (s *TxStore) Put(item storage.Item) error {
 func (s *TxStore) Delete(item storage.Item) error {
 	err := s.TxStoreBase.Delete(item)
 	if err == nil {
-		s.revOps.Append(&storage.TxRevertOp{
+		val, merr := item.Marshal()
```
aloknerurkar, line 193: We should do a Get before we call Marshal on the item. This follows the comment I had in the internal/reserve.go file. The item is not populated in most cases for Delete. Also, the user would have to do a Get anyway if he has to send it filled.

pkg/storage/leveldbstore/transaction.go
```diff
@@ -7,12 +7,130 @@ package leveldbstore
...
+			op.Origin,
+			op.ObjectID,
+			err,
+		))
+	}
+	if errs != nil {
+		return errs
+	}
+
+	return s.db.Put(s.id, s.batch.Dump(), &opt.WriteOptions{Sync: true})
```
aloknerurkar, line 108: We should see if we can get away with the sync write here. If we have to do it on every append, it will be slow as shit. Or we can have some background goroutine which keeps doing from time to time. For the revert, it should be fine.

pkg/storer/internal/chunkstore/transaction.go
```diff
@@ -165,16 +173,27 @@ func (cs *txChunkStoreWrapper) NewTx(state *storage.TxState) storage.TxChunkStor
...
-			ChunkStore: &chunkStoreWrapper{
-				store:  cs.store,
-				sharky: txSharky,
-			},
+			TxState:    state,
+			ChunkStore: chunkStore,
 		},
 		store:    cs.store,
 		txSharky: txSharky,
+		revOps: storage.NewInMemTxRevertOpStore(
```
aloknerurkar, line 187: The inmem revert op store might not be enough here right? I guess we need some sort of disk recovery mechanism here.

pkg/storer/internal/chunkstore/transaction.go
mrekucci: Done.

pkg/storer/internal/reserve/reserve.go
mrekucci: Done.

pkg/storage/leveldbstore/transaction.go
mrekucci: Done.

pkg/storage/leveldbstore/transaction.go
mrekucci: It would be better to bench this first.

mrekucci:
Okay, but we'll trade speed for potential inconsistency. If a node crashes and LevelDB doesn't have the data stored on disk at that time, then there will be no proper recovery.

aloknerurkar:
> Okay, but we'll trade speed for potential inconsistency. If a node crashes and LevelDB doesn't have the data stored on disk at that time, then there will be no proper recovery.

I think even levelDB has some repair mechanism on start. But not sure what that does. But as I said we could do it in background for the Append.

pkg/storage/inmemstore/transaction.go
```diff
@@ -36,46 +36,50 @@ func (s *TxStore) Put(item storage.Item) error {
...
 	err := s.TxStoreBase.Delete(item)
 	if err == nil {
-		s.revOps.Append(&storage.TxRevertOp{
-			Origin:   storage.DeleteOp,
-			ObjectID: item.String(),
-			Revert: func() error {
-				return s.TxStoreBase.Store.Put(item)
-			},
-		})
+		err = s.revOps.Append(reverseOp)
```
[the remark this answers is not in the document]
mrekucci: Done.

pkg/storage/leveldbstore/transaction.go
```diff
@@ -40,57 +110,60 @@ func (s *TxStore) Put(item storage.Item) error {
...
-			},
-		})
+			Key:      key(item),
+			Val:      val,
+		}
+	}
+
+	err := s.TxStoreBase.Delete(item)
+	if err == nil {
+		err = s.revOps.Append(reverseOp)
```
aloknerurkar, line 166: For ErrNotFound in L151, the reverseOp will be nil right? Is this handled?

aloknerurkar (approved):
Tests are still failing. LGTM otherwise. Mainly we should add some logs on recovery and test it in the 30 node cluster to ensure everything is happening smoothly.

pkg/storage/leveldbstore/transaction.go
```diff
@@ -7,13 +7,85 @@ package leveldbstore
...
+func (s *txRevertOpStore) Revert() error {
+	if s == nil {
+		return nil
+	}
+
+	s.batchMu.Lock()
+	defer s.batchMu.Unlock()
+	defer s.batch.Reset()
+
+	s.batch.Delete(s.id)
```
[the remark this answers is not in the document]
mrekucci: This basically adds the deletion of the in-progress transaction state into the same batch as the reverse ops, so it'll be deleted along with the reverse ops.

istae (approved):

PR #4164, opened 2023-06-16
feat(retrieval): use healthy peers for requests

istae:
### Checklist

### Description
Should boost retrieval success rate and performance.

### DIFF
--- a/pkg/retrieval/retrieval.go
+++ b/pkg/retrieval/retrieval.go
@@ -364,7 +364,7 @@ func (s *Service) prepareCredit(ctx context.Context, peer, chunk swarm.Address, 
 // retrieve request.
 func (s *Service) closestPeer(addr swarm.Address, skipPeers []swarm.Address, allowUpstream bool) (swarm.Address, error) {
 
-	closest, err := s.peerSuggester.ClosestPeer(addr, false, topology.Select{Reachable: true}, skipPeers...)
+	closest, err := s.peerSuggester.ClosestPeer(addr, false, topology.Select{Reachable: true, Healthy: true}, skipPeers...)
 	if err != nil {
 		return swarm.Address{}, err
 	}


notanatol (approved):

pkg/retrieval/retrieval.go
```diff
@@ -364,7 +364,7 @@ func (s *Service) prepareCredit(ctx context.Context, peer, chunk swarm.Address,
 // retrieve request.
 func (s *Service) closestPeer(addr swarm.Address, skipPeers []swarm.Address, allowUpstream bool) (swarm.Address, error) {
 
-	closest, err := s.peerSuggester.ClosestPeer(addr, false, topology.Select{Reachable: true}, skipPeers...)
+	closest, err := s.peerSuggester.ClosestPeer(addr, false, topology.Select{Reachable: true, Healthy: true}, skipPeers...)
```
aloknerurkar, line 367: Is this required? Are we seeing failed retrieval attempts?

I am just afraid of restricting to only healthy peers here. We could have reachable peers who could potentially respond but wont be considered.

istae: On occasion, yes, there are failed downloads, but still a very small rate, like two or three failed downloads per day, out of like 500 smoke test downloads. the other issue is download speed is lagging behind uploads, which is counterintuitive. 

adding healthy peer checks to pushsync improved retrieval as well but adding this check to retrieval is the next step.

using only healthy peers would boost performance and reliability.

istae: we can put this on hold until after the localstore release if you'd like

aloknerurkar (approved):
I am inclined to try this out as retrievals are slow and pushsync is really fast!

mrekucci (approved):

PR #4167, opened 2023-06-19
fix: retry on shallow receipt for direct upload

aloknerurkar:
### Checklist

### Description
In the pusher we return shallow receipt error on pushing if we receive a receipt with invalid depth. This is to allow the pusher to retry again to get a better receipt. This was missing in the direct upload. In the old version, direct upload keeps retrying on error, so this change will revert to old behavior. After the allowed retry limit, this error will be swallowed by the pusher and the operation will be termed success.

### DIFF
--- a/pkg/storer/netstore.go
+++ b/pkg/storer/netstore.go
@@ -46,7 +46,7 @@ func (db *DB) DirectUpload() PutterSession {
 							return ErrDBQuit
 						case err := <-op.Err:
 							// if we get a shallow receipt error, we retry the upload, the pusher will
-							// have an allowed no. of retries before after which a shallow receipt will
+							// have an allowed no. of retries after which a shallow receipt will
 							// no longer be returned as error.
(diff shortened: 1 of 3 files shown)


notanatol (approved):

pkg/storer/netstore.go
```diff
@@ -44,6 +45,13 @@ func (db *DB) DirectUpload() PutterSession {
 						case <-db.quit:
 							return ErrDBQuit
 						case err := <-op.Err:
+							// if we get a shallow receipt error, we retry the upload, the pusher will
+							// have an allowed no. of retries after which a shallow receipt will
+							// no longer be returned as error.
+							if errors.Is(err, pusher.ErrShallowReceipt) {
+								db.logger.Debug("direct upload: shallow receipt received, retrying", "chunk", ch.Address())
+								goto RETRY
```
mrekucci, line 53: Why not use a `for` loop?

PR #4187, opened 2023-06-28
perf: reduce locking context during evictions

aloknerurkar:
### Checklist

### Description
Currently, the eviction function holds the lock for the entire duration of the call. This includes holding the lock during iteration of the bin items and also Put to cache. These are not required. We only need the lock while we remove the chunk related indexes.

### DIFF
--- a/pkg/storer/internal/reserve/reserve.go
+++ b/pkg/storer/internal/reserve/reserve.go
@@ -13,6 +13,7 @@ import (
 	"sync/atomic"
 
 	"github.com/ethersphere/bee/pkg/log"
+	"github.com/ethersphere/bee/pkg/postage"
 	storage "github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/storer/internal"
 	"github.com/ethersphere/bee/pkg/storer/internal/chunkstamp"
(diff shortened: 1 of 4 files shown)


pkg/storer/internal/reserve/reserve_test.go
```diff
@@ -218,17 +218,26 @@ func TestEvict(t *testing.T) {
 		}
 	}
 
-	totalEvicted := 0
+	chsToDelete := make([]swarm.Chunk, 0)
```
mrekucci, line 221: The slice length is missing.

mrekucci (approved):

pkg/storer/internal/reserve/reserve_test.go
aloknerurkar: This is not required right? There is a check based on length, so I dont want to preallocate.

pkg/storer/reserve.go
```diff
@@ -209,33 +209,48 @@ func (db *DB) evictBatch(ctx context.Context, batchID []byte, upToBin uint8) (er
...
+				return false, err
 			}
+			evicted++
+			return false, nil
+		})
 
-			db.logger.Info("reserve eviction", "bin", b, "evicted", evicted, "batchID", hex.EncodeToString(batchID), "size", db.reserve.Size())
+		// if there was an error, we still need to update the chunks that have already
+		// been evicted from the reserve
+		db.lock.Lock(reserveUpdateLockKey)
```
istae, line 239: why lock here?

aloknerurkar: Ohh... I missed that they were changed to atomics... I thought there was a lock there which was removed. Fixed

istae (approved):

pkg/storer/internal/reserve/reserve.go
```diff
@@ -280,57 +281,58 @@ func (r *Reserve) IterateChunksItems(store internal.Storage, startBin uint8, cb
...
-					return err
-				}
-
-				cb(c)
-
-				err = removeChunk(ctx, store, item)
-				if err != nil {
-					return err
-				}
+		stamp, err := chunkstamp.LoadWithBatchID(store.IndexStore(), reserveNamespace, batchRadius.Address, batchRadius.BatchID)
```
istae, line 308: the stamp does not have have to be loaded actually
the batchID is already known, 
the callback can simply return address and batchID

istae (changes requested):

istae (approved):

PR #4192, opened 2023-06-28
refactor(listener): use atomic pointer for chain state and add extra log line

mrekucci:
### Checklist

### Description
Replace chain state guarding mutex for atomic pointer and add extra logging to see if the listener was silently shutdown.

### DIFF
--- a/pkg/postage/batchstore/store.go
+++ b/pkg/postage/batchstore/store.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"math"
 	"math/big"
-	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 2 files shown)


istae (approved):

ldeffenb:
I can confirm that this branch resolves a node that was previously stuck on an earlier "put chain state" block number after multiple restarts.   When restarting the node with this version, it successfully caught up and followed the current chain block.

Also, a single Control-C (in a shell prompt on Ubuntu) causes the new "shutting down event listener" to log, although the node still doesn't shut down cleanly due to lots of pullsync handler messages.

aloknerurkar (approved):

PR #4197, opened 2023-06-29
fix(reserve): optimizing the cleanup worker

istae:
### Checklist

### DIFF
--- a/pkg/storer/reserve.go
+++ b/pkg/storer/reserve.go
@@ -23,6 +23,8 @@ const (
 	reserveOverCapacity  = "reserveOverCapacity"
 	reserveUnreserved    = "reserveUnreserved"
 	reserveUpdateLockKey = "reserveUpdateLockKey"
+
+	cleanupDur = time.Hour * 6
 )
 
 var errMaxRadius = errors.New("max radius reached")
@@ -76,7 +78,11 @@ func (db *DB) reserveWorker(ctx context.Context, warmupDur, wakeUpDur time.Durat
 	// syncing can now begin now that the reserver worker is running
 	db.syncer.Start(ctx)
 
-	wakeUpTicker := time.NewTicker(wakeUpDur)
+	radiusWakeUpTicker := time.NewTicker(wakeUpDur)
+	defer radiusWakeUpTicker.Stop()
+
+	cleanUpTicker := time.NewTicker(cleanupDur)
+	defer cleanUpTicker.Stop()
 
 	for {
 		select {
@@ -86,7 +92,7 @@ func (db *DB) reserveWorker(ctx context.Context, warmupDur, wakeUpDur time.Durat
 				db.logger.Error(err, "reserve unreserve")
 			}
 			db.metrics.OverCapTriggerCount.Inc()
-		case <-wakeUpTicker.C:
+		case <-radiusWakeUpTicker.C:
 			radius := db.reserve.Radius()
 			if db.reserve.Size() < threshold(db.reserve.Capacity()) && db.syncer.SyncRate() == 0 && radius > 0 {
 				radius--
@@ -97,11 +103,12 @@ func (db *DB) reserveWorker(ctx context.Context, warmupDur, wakeUpDur time.Durat
 				db.logger.Info("reserve radius decrease", "radius", radius)
 			}
 			db.metrics.StorageRadius.Set(float64(radius))
-
-			if err := db.reserveCleanup(ctx); err != nil {
-				db.logger.Error(err, "cleanup")
-			}
-
+		case <-cleanUpTicker.C:
+			go func() {
+				if err := db.reserveCleanup(ctx); err != nil {
+					db.logger.Error(err, "cleanup")
+				}
+			}()
 		case <-db.quit:
 			return
 		}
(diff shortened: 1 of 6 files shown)


pkg/storer/reserve.go
```diff
@@ -276,21 +281,29 @@ func (db *DB) removeChunk(ctx context.Context, address swarm.Address, batchID []
...
 	dur := captureDuration(time.Now())
+	count := 0
 	defer func() {
 		db.metrics.MethodCallsDuration.WithLabelValues("reserve", "cleanup").Observe(dur())
+		db.metrics.ReserveSize.Set(float64(db.reserve.Size()))
+		db.logger.Info("cleanup finished", "total", count, "duration", dur())
 	}()
 
-	return db.reserve.IterateChunksItems(db.repo, 0, func(ci reserve.ChunkItem) (bool, error) {
+	ids := map[string]struct{}{}
```
ldeffenb, line 291: Anxious to see if this improves (reduces) the elapsed time of reserveCleanup!

pkg/storer/reserve.go
```diff
@@ -276,21 +281,29 @@ func (db *DB) removeChunk(ctx context.Context, address swarm.Address, batchID []
 
 func (db *DB) reserveCleanup(ctx context.Context) error {
 	dur := captureDuration(time.Now())
+	count := 0
 	defer func() {
 		db.metrics.MethodCallsDuration.WithLabelValues("reserve", "cleanup").Observe(dur())
+		db.metrics.ReserveSize.Set(float64(db.reserve.Size()))
+		db.logger.Info("cleanup finished", "total", count, "duration", dur())
```
ldeffenb, line 288: Would "removed" be better than "total" the latter of which might imply that it's the total size of the reserve?

ldeffenb (changes requested):

pkg/storer/reserve.go
```diff
@@ -276,21 +281,29 @@ func (db *DB) removeChunk(ctx context.Context, address swarm.Address, batchID []
 
 func (db *DB) reserveCleanup(ctx context.Context) error {
 	dur := captureDuration(time.Now())
+	removed := 0
 	defer func() {
 		db.metrics.MethodCallsDuration.WithLabelValues("reserve", "cleanup").Observe(dur())
+		db.metrics.ReserveSize.Set(float64(db.reserve.Size()))
```
aloknerurkar, line 287: what about the Reserve.Size ?
We set the metric but we are not adjusting the size of the reserve. Also there is no easy way to do it as we are iterating with bin 0. So there could be chunks which are not accounted for in the reserve size. We will need to do a complete recalculation after the cleanup is done I think.

pkg/storer/reserve.go
```diff
@@ -276,21 +281,29 @@ func (db *DB) removeChunk(ctx context.Context, address swarm.Address, batchID []
...
+		db.logger.Info("cleanup finished", "removed", removed, "duration", dur())
```
aloknerurkar, line 288: We should also add a metric for this. If this happens a lot in 6 hrs we have a problem.

pkg/storer/reserve.go
aloknerurkar: This is happening right now. After restart the reserve values drop.

Screenshot 2023-06-28 at 11 54 01 PM

ldeffenb: > what about the Reserve.Size ?

He's right.  I missed the calls to db.reserve.AddSize(1) and db.reserve.AddSize(-evicted) in ReservePutter and evictBatch above.

istae: nice catch

ldeffenb (approved):
It all looks correct to me.   I've got an idea on lazy-population of the ids[] which can go into a different PR at some point in the future.

pkg/storer/reserve.go
```diff
@@ -276,21 +281,39 @@ func (db *DB) removeChunk(ctx context.Context, address swarm.Address, batchID []
...
 	dur := captureDuration(time.Now())
+	removed := 0
 	defer func() {
 		db.metrics.MethodCallsDuration.WithLabelValues("reserve", "cleanup").Observe(dur())
-	}()
+		db.metrics.ReserveCleanup.Add(float64(removed))
+		db.logger.Info("cleanup finished", "removed", removed, "duration", dur())
 
-	return db.reserve.IterateChunksItems(db.repo, 0, func(ci reserve.ChunkItem) (bool, error) {
+		db.lock.Lock(reserveUpdateLockKey)
```
aloknerurkar, line 290: We should only do this `if removed > 0`

pkg/storer/reserve.go
```diff
@@ -276,21 +281,39 @@ func (db *DB) removeChunk(ctx context.Context, address swarm.Address, batchID []
...
+		return false, nil
+	})
+	if err != nil {
+		return err
+	}
+
+	return db.reserve.IterateChunksItems(db.repo, 0, func(ci reserve.ChunkItem) (bool, error) {
+		if _, ok := ids[string(ci.BatchID)]; !ok {
+			removed++
+			db.logger.Debug("cleanup expired batch", "batch_id", hex.EncodeToString(ci.BatchID))
```
mrekucci, line 314: Logging might slow the iteration.

mrekucci (approved):
Apart from the comments, LGTM.

pkg/storer/reserve.go
istae: we should only catch a few chunks here :)

aloknerurkar (approved):

Issue #4087, opened 2023-05-22
Postage batch expiration logic cleanup


aloknerurkar:

acha-bill:
Noting to be done. @aloknerurkar

PR #4203, opened 2023-07-03
feat: save expired batches in db

acha-bill:
### Checklist

### Description
Expired batches were being stored in-mem. This PR saves them in DB and deletes them when their associated data has been successfully cleaned up.

### Related Issue (Optional)
#4087

### DIFF
--- a/pkg/storer/reserve.go
+++ b/pkg/storer/reserve.go
@@ -24,7 +24,6 @@ const (
 	reserveUnreserved    = "reserveUnreserved"
 	reserveUpdateLockKey = "reserveUpdateLockKey"
 	batchExpiry          = "batchExpiry"
-	expiredBatchAccess   = "expiredBatchAccess"
 
 	cleanupDur = time.Hour * 6
 )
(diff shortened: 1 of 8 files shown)


pkg/storer/reserve.go
```diff
@@ -253,18 +255,18 @@ func (db *DB) ReservePutter() storage.Putter {
...
+func (db *DB) EvictBatch(ctx context.Context, batchID []byte) error {
 	if db.reserve == nil {
 		// if reserve is not configured, do nothing
 		return nil
 	}
 
-	db.lock.Lock(expiredBatchAccess)
-	db.expiredBatches = append(db.expiredBatches, batchID)
-	db.lock.Unlock(expiredBatchAccess)
+	err := db.batchstore.SaveExpired(batchID)
```
aloknerurkar, line 264: So I dont think this makes much sense. It seems to be better to store the expired batches in storer itself. Saving it in batchstore context on callback from batchstore seems like unnecessary work.

You can use the store in storer itself to store it when you get the call and then trigger the eviction. When you are evicting you iterate on the store and get the batches instead of getting them from in-mem. Also, on initialization you will have to check if there are entries left to be processed and trigger the eviction again.

aloknerurkar (changes requested):

aloknerurkar (approved):

mrekucci (approved):

PR #4220, opened 2023-07-12
fix(cache): include namespace in the cache key

mrekucci:
### Checklist

### Description
Includes a namespace in the cache key, which addresses the case where two entries from different namespaces may have the same ID.

### DIFF
--- a/pkg/storage/cache/cache.go
+++ b/pkg/storage/cache/cache.go
@@ -6,9 +6,16 @@ package cache
 
 import (
 	"errors"
+
 	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/storage/storageutil"
 	"github.com/hashicorp/golang-lru/v2"
 )
+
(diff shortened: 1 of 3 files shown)


pkg/storage/cache/cache.go
```diff
@@ -21,20 +28,29 @@ type Cache struct {
...
 
-	return &Cache{store, lru, newMetrics()}
+	return &Cache{store, lru, newMetrics()}, nil
+}
+
+// MustWrap is like Wrap but panics on error.
+func MustWrap(store storage.Store, capacity int) *Cache {
+	c, err := Wrap(store, capacity)
+	if err != nil {
+		panic(err)
```
notanatol, line 51: is it ok if we panic while we init the state store?

mrekucci: Good catch, fixed!

aloknerurkar (approved):

PR #4230, opened 2023-07-21
fix: reserve wg caused data race in tests

notanatol:
### Checklist

### Description
This PR introduces new concurrent data structure `waitingCounter` and uses it to get around the race in tests cause by the use of wait groups.


aloknerurkar (approved):
There is some test artifact added to the PR. Make sure to remove it before checkin.

PR #4234, opened 2023-07-29
perf: improve eviction worker cpu use

aloknerurkar:
### Checklist

### Description
Debugging on resource-constrained machines, we found some unnecessary goroutine creations and also excessive expiry checks. Thanks to @ldeffenb for helping out with this! These changes should improve the CPU consumption of the nodes.

Also, a bug was found with the move to cache where the reserve size was not being changed correctly. This is fixed by moving the cache op into the same transaction, so it can be reverted together.

### DIFF
--- a/pkg/storer/reserve.go
+++ b/pkg/storer/reserve.go
@@ -32,8 +32,8 @@ const (
 	cleanupDur = time.Hour * 6
 )
 
-func reserveUpdateBatchLockKey(batchID []byte, bin uint8) string {
-	return fmt.Sprintf("%s%s%d", reserveUpdateLockKey, string(batchID), bin)
+func reserveUpdateBatchLockKey(batchID []byte) string {
+	return fmt.Sprintf("%s%s", reserveUpdateLockKey, string(batchID))
 }
 
 var errMaxRadius = errors.New("max radius reached")
@@ -352,10 +352,7 @@ func (db *DB) ReservePutter() storage.Putter {
 				var (
 					newIndex bool
 				)
-				lockKey := reserveUpdateBatchLockKey(
-					chunk.Stamp().BatchID(),
-					db.po(chunk.Address()),
-				)
+				lockKey := reserveUpdateBatchLockKey(chunk.Stamp().BatchID())
 				db.lock.Lock(lockKey)
 				err = db.Execute(ctx, func(tx internal.Storage) error {
 					newIndex, err = db.reserve.Put(ctx, tx, chunk)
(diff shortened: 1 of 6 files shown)


pkg/storer/reserve.go
```diff
@@ -172,69 +172,71 @@ func (db *DB) evictionWorker(ctx context.Context) {
...
-				defer expiryWorkers.Release(1)
 
-				err := db.removeExpiredBatch(ctx, b)
+			// we ensure unreserve is not running and if it is we cancel it and wait
+			// for it to finish, this is to prevent unreserve and expirations running
+			// at the same time. The expiration will free up space so the unreserve
+			// target might change. This is to prevent unreserve from running with
+			// the old target.
+			if !unreserveSem.TryAcquire(1) {
+				cancelUnreserve()
```
istae, line 201: instead of cancelling, would it make sense to wait for the unreserve to finish first with `unreserveSem.Acquire` ?

pkg/storer/reserve.go
```diff
@@ -172,69 +172,71 @@ func (db *DB) evictionWorker(ctx context.Context) {
...
+				defer db.events.Trigger(reserveOverCapacity)
+			}
+			defer unreserveSem.Release(1)
 
-		if err := expiryWorkers.Acquire(ctx, 4); err != nil {
-			db.logger.Error(err, "wait for expiry workers")
-			return
-		}
-		expiryWorkers.Release(4)
+			defer db.events.Trigger(reserveUnreserved)
```
istae, line 212: why is this emitted?

pkg/storer/reserve.go
```diff
@@ -172,69 +172,71 @@ func (db *DB) evictionWorker(ctx context.Context) {
...
+					err := db.removeExpiredBatch(ctx, b)
+					if err != nil {
+						db.logger.Error(err, "remove expired batch", "batch_id", hex.EncodeToString(b))
+					} else {
+						db.metrics.ExpiredBatchCount.Inc()
+					}
+				}()
+			}
+
+			if err := expiryWorkers.Acquire(ctx, 4); err != nil {
```
istae, line 234: what is the purpose of this?

aloknerurkar: This is to ensure all expiry workers finished before we give up the sem. At the most, we will start 4 workers in parallel to expire. There are more workers here as for expired batches each of them will be working on different entries, so they can all do it in parallel.

pkg/storer/reserve.go
aloknerurkar: This was actually added for the test mainly. No one is listening for it. I thought its easier to use the same one than to have a separate key. Wdyt?

pkg/storer/reserve.go
aloknerurkar: So it could happen that the unreserve target is substantial. Which is why this cancellation was added to mainly prevent removing valid chunks till we need to.

pkg/storer/reserve.go
istae: right, works for me, maybe a small comment is needed.

pkg/storer/reserve.go
istae: i see, we can have at most 4 `removeExpiredBatch` funcs running in parallel. the `Acquire` here waits for the above calls to finish.

pkg/storer/reserve.go
istae: as long as canceling is safe, im ok with this.
another small issue maybe this cancel func needs to be initially assigned a noOp func to prevent a panic

aloknerurkar: This was actually tested. We even found that there was actually an issue with the counts. This is fixed in the next PR where we keep track of the counts per batch. That [fix](https://github.com/ethersphere/bee/pull/4235/files#diff-e37eed2ab5dfe1533be0257473f5b76738361f02d2be71da1b7bf4d5216a00c7R321) was tested with @ldeffenb s help, where the counts were verified after restarts.

istae (approved):

PR #4235, opened 2023-07-29
perf: batch id prefix for batchradius

aloknerurkar:
### Checklist

### Description
So this optimization was reverted mainly because we did not find any major improvement. But seems like this becomes more important as the radius goes higher. On the mainnet we are at radius 8, so we would have 100s of empty iterations which will waste a lot of time. So we will introduce this change again.

### DIFF
--- a/pkg/storer/internal/reserve/items.go
+++ b/pkg/storer/internal/reserve/items.go
@@ -33,13 +33,13 @@ func (b *batchRadiusItem) Namespace() string {
 	return "batchRadius"
 }
 
-// bin/batchID/ChunkAddr
+// batchID/bin/ChunkAddr
 func (b *batchRadiusItem) ID() string {
-	return batchBinToString(b.Bin, b.BatchID) + b.Address.ByteString()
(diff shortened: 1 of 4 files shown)


pkg/storer/internal/reserve/items.go
```diff
@@ -33,13 +33,13 @@ func (b *batchRadiusItem) Namespace() string {
...
-// bin/batchID/ChunkAddr
+// batchID/bin/ChunkAddr
 func (b *batchRadiusItem) ID() string {
-	return batchBinToString(b.Bin, b.BatchID) + b.Address.ByteString()
+	return batchBinToString(b.BatchID, b.Bin) + b.Address.ByteString()
 }
 
-func batchBinToString(bin uint8, batchID []byte) string {
-	return string(bin) + string(batchID)
+func batchBinToString(batchID []byte, bin uint8) string {
```
istae, line 41: we can remove this func now

istae (approved):

notanatol (approved):

PR #4237, opened 2023-07-31
fix: swallowed batchExpiry trigger

aloknerurkar:
### Checklist

### Description
Missed expiry triggers. If expiry triggers are swallowed after we already selected batches to expire and no new signals for expiry are received, we will keep a hold of the expired chunks till cleanup comes and helps us. This change ensures we trigger a run if we have selected entries to ensure we havent swallowed any triggers. If there are no more batches, then it wont re-trigger it.

### DIFF
--- a/pkg/storer/reserve.go
+++ b/pkg/storer/reserve.go
@@ -179,7 +179,13 @@ func (db *DB) evictionWorker(ctx context.Context) {
 		}
 
 		go func() {
-			defer expirySem.Release(1)
+			reTrigger := false
+			defer func() {
+				expirySem.Release(1)
+				if reTrigger {
+					db.events.Trigger(batchExpiry)
+				}
+			}()
 
 			batchesToEvict, err := db.getExpiredBatches()
 			if err != nil {
@@ -190,6 +196,10 @@ func (db *DB) evictionWorker(ctx context.Context) {
 			if len(batchesToEvict) == 0 {
 				return
 			}
+
+			// After this point we start swallowing signals, so ensure we do 1 more
+			// trigger at the end.
+			reTrigger = true
 
 			// we ensure unreserve is not running and if it is we cancel it and wait
 			// for it to finish, this is to prevent unreserve and expirations running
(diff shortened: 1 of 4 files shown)


notanatol (approved):

pkg/storer/reserve.go
```diff
@@ -213,6 +213,15 @@ func (db *DB) evictionWorker(ctx context.Context) {
 			// it uses the same one as reserve eviction.
 			defer db.events.Trigger(reserveUnreserved)
 
+			// this trigger ensures that if the expiry triggers were swallowed
+			// after we checked for expired batches, we re-check the batches
+			// again. It should not be fired if there are no more batches to evict.
+			defer func() {
+				_ = time.AfterFunc(1*time.Second, func() {
```
istae, line 220: this is probably due to the early exit if the semaphore cannot be acquired here https://github.com/ethersphere/bee/pull/4237/files#diff-acdf952e80a8279dfab4acea0ff533558f2857121497d2a2a343d74058ddef7dR175

by changing the interactions with the semaphore, this after func won't be needed.

istae: since the internals of the cleanupExpired func is inside a goroutine anyways, waiting for the semaphore could work, eg calling `Acquire`.

istae: if the above is not ideal, you can also stick a `db.events.Trigger(batchExpiry)` here https://github.com/ethersphere/bee/pull/4237/files/53efc7eff387949ca0b290249807abb4f2b0b12d#diff-acdf952e80a8279dfab4acea0ff533558f2857121497d2a2a343d74058ddef7dR176

istae: if the above is not ideal, you can add a forever loop inside the goroutine of `cleanupExpired`
```go
for {
 batchesToEvict, err := db.getExpiredBatches()
 if len(batchesToEvict) == 0 {
  break
 }
}
```

aloknerurkar: If the main worker waits to acquire the semaphore, then it cannot process unreserve/cleanup-related signals till the expiration finishes. So the idea was the main worker should be unblocked all the time. Secondly, if we start goroutines for each trigger waiting to acquire, then we might overwhelm the CPU. Also if we trigger before returning, it will cause excessive checking and creation of goroutines.

istae: what about the loop idea?

aloknerurkar: Okay this is better

ldeffenb: > what about the loop idea?

The loop still has a race condition because the zero is detected before the semaphore is released, so if a trigger fires and is ignored between the loop end and the defer semaphore release, that trigger will still be ignored.

aloknerurkar: > > what about the loop idea?
> 
> The loop still has a race condition because the zero is detected before the semaphore is released, so if a trigger fires and is ignored between the loop end and the defer semaphore release, that trigger will still be ignored.

So I feel this is better than the time.After idea. We would proactively remove instead of waiting for the signal to fire it again. If we miss out on entries due to the race, we always have either cleanup or new expirations that can trigger us again.

istae (approved):

PR #4242, opened 2023-08-07
fix: simplify cache handling

aloknerurkar:
### Checklist

### Description
Simplifying cache component. Having a simpler timestamp index to iterate makes things much simpler and less error prone. The linked list is very error prone and leaves the cache useless if we hit any issues.

Verified on mainnet dev cluster:
Screenshot 2023-08-08 at 4 31 09 PM

### DIFF
--- a/pkg/storer/internal/cache/cache.go
+++ b/pkg/storer/internal/cache/cache.go
@@ -84,8 +84,6 @@ func (c cacheEntry) String() string {
 	)
 }
 
-const cacheOrderIndexSize = 2*swarm.HashSize + 8
-
 var _ storage.Item = (*cacheOrderIndex)(nil)
 
 type cacheOrderIndex struct {
@@ -398,8 +396,8 @@ func (c *Cache) MoveFromReserve(
 		return nil
 	}
 
-	if len(entriesToAdd) > int(c.capacity) {
-		entriesToAdd = entriesToAdd[len(entriesToAdd)-int(c.capacity):]
+	if len(entriesToAdd) > c.capacity {
+		entriesToAdd = entriesToAdd[len(entriesToAdd)-c.capacity:]
 	}
 
 	var entriesToRemove int
--- a/pkg/storer/internal/cache/cache.go
+++ b/pkg/storer/internal/cache/cache.go
@@ -9,27 +9,30 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"sync"
+	"strconv"
+	"sync/atomic"
+	"time"
 
 	storage "github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/storer/internal"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
-const cacheEntrySize = 3 * swarm.HashSize
+var now = time.Now
+
+const cacheEntrySize = swarm.HashSize + 8
 
 var _ storage.Item = (*cacheEntry)(nil)
 
 var (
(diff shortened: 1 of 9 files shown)


notanatol (approved):

pkg/storer/internal/cache/cache.go
```diff
@@ -53,8 +61,7 @@ func (c *cacheEntry) Unmarshal(buf []byte) error {
 	}
 	newEntry := new(cacheEntry)
 	newEntry.Address = swarm.NewAddress(append(make([]byte, 0, swarm.HashSize), buf[:swarm.HashSize]...))
-	newEntry.Prev = internal.AddressOrZero(buf[swarm.HashSize : 2*swarm.HashSize])
-	newEntry.Next = internal.AddressOrZero(buf[2*swarm.HashSize:])
+	newEntry.AccessTimestamp = int64(binary.LittleEndian.Uint64(buf[swarm.HashSize:]))
```
istae, line 64: why is LittleEndian used here?
for the most part BigEndian is used (see internal/reserve) for having sorted.

aloknerurkar: So the key creation is separate. The `ID()` function provides a string timestamp to enable the sorting. The way the value is stored doesnt really matter for the sorting.

istae:
it's not part of the changes but is `db.lock.Lock(cacheAccessLockKey)` is needed still ?

pkg/storer/internal/cache/cache.go
```diff
@@ -295,69 +265,54 @@ func (c *Cache) Putter(store internal.Storage) storage.Putter {
...
+			addSize = false
+			err := removeOldest(
 				ctx,
-				state,
-				newEntry,
-				chunk,
 				store.IndexStore(),
 				batch,
 				store.ChunkStore(),
+				1,
```
istae, line 290: here maybe we can remove at least 1,000~50,000 items so we do not have to remove oldest entries every time.

pkg/storer/internal/cache/cache.go
```diff
@@ -295,69 +265,54 @@ func (c *Cache) Putter(store internal.Storage) storage.Putter {
...
+		err = batch.Put(&cacheOrderIndex{
+			Address:         newEntry.Address,
+			AccessTimestamp: newEntry.AccessTimestamp,
+		})
+		if err != nil {
+			return fmt.Errorf("failed adding cache order index: %w", err)
+		}
+
+		addSize := true
+		if c.size.Load() == int64(c.capacity) {
```
istae, line 283: should be `>=`

aloknerurkar:
> it's not part of the changes but is `db.lock.Lock(cacheAccessLockKey)` needed still ?

So right now it is needed, but it can be removed if we move the cache size handling into storer pkg. Right now as it's in the internal cache, we can remove the lock. It can be done as a follow up change.

pkg/storer/migration/step_02.go
```diff
@@ -0,0 +1,46 @@
...
+			}
+			entries = append(entries, entry)
+			return false, nil
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	for _, entry := range entries {
```
istae, line 38: maybe use a batch here?

istae (approved):

pkg/storer/migration/step_02.go
aloknerurkar: So I did not do this now as this would involve changing the interfaces in the migration pkg in storage. Right now it doesnt use BatchedStore. In any case the migration runs only once in the start and does not use transactions. Which is why I thought its okay for now.

PR #4245, opened 2023-08-07
fix(reserve): simplify batch expiration and unreserve interaction

istae:
### Checklist

### Description
Batch expiration and unreserve worker has been greatly simplified.

1. Batch expiration still has higher priority over unreserve. The unreserve now listens on batch expiration events and stops early if the expiration event is fired.
2. Expiration and unreserve still do not overlap (same as before).
3. Cleanup worker is removed (mainly for simplification, can be readded if it's discovered it would add value, has not so far)

### DIFF
--- a/pkg/storer/reserve.go
+++ b/pkg/storer/reserve.go
@@ -56,7 +56,7 @@ func (db *DB) startReserveWorkers(
 	// start eviction worker first as there could be batch expirations because of
 	// initial contract sync
 	db.inFlight.Add(1)
-	go db.worker(ctx)
+	go db.evictionWorker(ctx)
 
 	select {
 	case <-time.After(warmupDur):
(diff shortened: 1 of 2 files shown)


notanatol (approved):

acha-bill (approved):

pkg/storer/reserve.go
```diff
@@ -144,7 +140,8 @@ func (db *DB) removeExpiredBatch(ctx context.Context, batchID []byte) error {
 	})
 }
 
-func (db *DB) evictionWorker(ctx context.Context) {
+func (db *DB) worker(ctx context.Context) {
```
aloknerurkar, line 143: I think evictionWorker is a better name for this.

aloknerurkar (approved):
Nice! Seems like this was waayyy overengineered!

PR #4247, opened 2023-08-08
fix(migration): reset puller intervals

istae:
### Checklist

### DIFF
--- a/pkg/statestore/storeadapter/migration.go
+++ b/pkg/statestore/storeadapter/migration.go
@@ -14,19 +14,35 @@ import (
 func allSteps() migration.Steps {
 	return map[uint64]migration.StepFn{
 		1: epochMigration,
+		2: deletePrefix("sync_interval"),
 	}
 }
 
-var deleteEntries = []string{
-	"statestore_schema",
-	"tags",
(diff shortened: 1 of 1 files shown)


pkg/statestore/storeadapter/migration.go
```diff
@@ -14,19 +14,35 @@ import (
...
 
-var deleteEntries = []string{
-	"statestore_schema",
-	"tags",
-	"sync_interval",
-	"kademlia-counters",
-	"addressbook",
-	"batch",
+func deletePrefix(prefix string) migration.StepFn {
+	return func(s storage.Store) error {
```
aloknerurkar, line 22: Instead, you can just create the storeadapter here using the store and use the statestore Iterate function. That will give you the keys and then you can use the storeadapter.Delete to delete them.

If not you can add ItemProperty in the Query to QueryItemID and only get the keys. Then you can use that to delete like `store.Delete(newProxyItem(res.ID, []byte(nil)))`. The problem with this is that the Unmarshal wont populate the ID field of the newProxyItem. Hence it wont be able to delete it.

aloknerurkar (approved):

PR #4251, opened 2023-08-08
fix(reserve): reconstruct binIDs

istae:
### Checklist

### DIFF
--- a/pkg/storer/migration/step_03.go
+++ b/pkg/storer/migration/step_03.go
@@ -19,9 +19,9 @@ import (
 // ChunkBinItem and BatchRadiusItem entries to use a new BinID field.
 func step_03(st storage.BatchedStore) error {
 	/*
-		STEP 1, remove all of the BinItem entired
-		STEP 2, iterate ChunkBinItem, delete old ChunkBinItem and BatchRadiusItem
-		STEP 3, get new binID using IncBinID, create new ChunkBinItem and BatchRadiusItem
+		STEP 1, remove all of the BinItem entires
+		STEP 2, remove all of the ChunkBinItem entries
+		STEP 3, iterate BatchRadiusItem, get new binID using IncBinID, create new ChunkBinItem and BatchRadiusItem
 	*/
 
 	rs := reserve.Reserve{}
@@ -37,6 +37,7 @@ func step_03(st storage.BatchedStore) error {
 	}
 	logger.Info("removed all bin index entries")
 
+	// STEP 2
 	var chunkBinItems []*reserve.ChunkBinItem
 	err := st.Iterate(
 		storage.Query{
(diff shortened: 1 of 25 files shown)


pkg/storer/migration/step_03.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/ethersphere/bee/pkg/storer/internal/reserve"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// step_03 is a migration step that removes all BinItem entries and migrates
+// ChunkBinItem and BatchRadiusItem entries to use a new BinID field.
+func step_03(st storage.BatchedStore) error {
+	/*
+		STEP 1, remove all of the BinItem entired
+		STEP 2, iterate ChunkBinItem, delete old ChunkBinItem and BatchRadiusItem
```
istae, line 23: let's either remove the comment, or let's correct it

pkg/storer/migration/step_03.go
```diff
@@ -0,0 +1,158 @@
...
+				if err != nil {
+					return err
+				}
+
+				err = b.Put(&reserve.ChunkBinItem{
+					BatchID:   item.BatchID,
+					Bin:       item.Bin,
+					Address:   item.Address,
+					BinID:     newBinID,
+					ChunkType: swarm.ChunkTypeContentAddressed,
```
istae, line 143: this has to come from the already existing item

aloknerurkar: > this has to come from the already existing item

Okay, this is a problem. We will have to write some special function to fill this field in this case. The problem is the chunkBinItem may or may not be present. This is because the problematic entries would have overwritten the chunkBinItems. So even if we try to fill from the existing chunkBinItem and if its not found we will have issues.

Instead, this field is not used right now. So what we can do is update with the default value which is unspecified. And then have a separate function called in the storer.New function which looks at the chunkstore and populates this field properly for the ones that are unspecified.

istae: the problem with leaving it as default is the sampler has this check
```
				// Skip chunks if they are not SOC or CAC
				if chItem.Type != swarm.ChunkTypeSingleOwner &&
					chItem.Type != swarm.ChunkTypeContentAddressed {
					wstat.RogueChunk++
					continue
				}
```

istae: we can remove the field altogether and have the type computed on the fly wherever needed

aloknerurkar: > 

So I figured out how to do it... check the latest diff

istae (commented):
Nice!

aloknerurkar (approved):

acha-bill (approved):

PR #4290, opened 2023-09-05
fix(puller): reset intervals when peer reserve epoch changes

### LINKED ISSUE #3821: nuked nodes could cause intervals to be misaligned for peers attempting to pullsync with the nuked peer
⚠️ Support requests in an issue-format will be closed immediately. For support, go to Swarm's [Discord](https://discord.gg/XGNBEKktkD).

### Summary
When a node is nuked, it's binIds are also nuked, and after the node reboots from a nuke, the binIDs start from zero.
Peers that restart syncing with the nuked nodes will miss the certain intervals of chunks because the locally stored sync intervals will be ahead of the nuked binIDs

### Expected behavior
Syncing peers should reset their own intervals when they detect that the peers has been nuked.

### Possible solution
The nodes can store a timestamp of when they first starting running. 
When ever there is a change in that timestamp when peers exchange cursors, the peer can reset the sync intervals of that peer.

### LINKED ISSUE #4211: pullsync breaks on nuked peers
### Context
2.0.0 rc9 in testnet swarm

### Summary
Most of the testnet swarm nodes were nuked to clear out bogus data.  My own nodes were nuked later than most.  When I started re-loading the OSM data into my loader node, its reserve grew, and my local peers reserve grew, but the bulk of the swarm didn't pull the chunks into their reserves.

Also, one day later and the early nuked nodes still have reserve sizes all over the place.   Consider the following output from /status/peers:

```
      "reserveSize": 0,
      "reserveSize": 0,
      "reserveSize": 1048856,
      "reserveSize": 1317940,
      "reserveSize": 1352126,
      "reserveSize": 1352868,
      "reserveSize": 1368953,
      "reserveSize": 1371643,
      "reserveSize": 1376327,
      "reserveSize": 1377377,
      "reserveSize": 1381349,
      "reserveSize": 1382755,
      "reserveSize": 1386114,
      "reserveSize": 1391279,
      "reserveSize": 1392433,
      "reserveSize": 1442512,
      "reserveSize": 1461856,
      "reserveSize": 1711720,
      "reserveSize": 1711720,
      "reserveSize": 1711720,
      "reserveSize": 1711721,
      "reserveSize": 224233,
      "reserveSize": 329048,
      "reserveSize": 356116,
      "reserveSize": 412148,
      "reserveSize": 628406,
      "reserveSize": 725655,
```

### Expected behavior
I would have expected all of the nodes to pull the newly uploaded chunks from the uploader node.  At least the nodes that didn't directly receive the pushsync that the uploader was doing.

### Actual behavior
The swarm has not picked up the chunks introduced into the swarm.

### Steps to reproduce
Have an existing swarm (storageRadius=0 is easiest to see this issue) with lots of data in the reserve.   This swarm would have all of the puller's Intervals set to numbers substantially greater than 1 since data has been pulled from all of it's peers' bins.

Nuke most of the nodes setting the reserve of all to zero, but NOT clearing the puller's peerIntervals.   Keep a few nodes not-nuked, especially one that has a stamp with a lot of uploaded data that the other nodes just nuked.

Observe that the nuked nodes do no pull the existing data from the non-nuked node.  It's reserve will still be large, while the other's don't pull that existing data.

Now nuke the remaining nodes.  The entire swarm's reserves will be at or near zero at this point.

Load lots of data into the recently nuked node that previously held the data.   It's reserve will increase (remember, we're talking about a storageRadius=0 swarm for simplicity).

All of the second-pass nuked nodes will receive the data, but the originally nuked nodes will not get all of it.   The reason is that their puller's peerIntervals were never reset and they won't be asking to pull indexes starting at 1, but will be trying to pull bin indexes that are as large as before the nuke.

If you nuke one or a few of the early nuked nodes, they will pull the data, but even then, the other early nuked nodes won't pull it because they still have high Indexes in their peerInterval bins.

Theory: Starting a brand new node into this swarm may allow the reserve to completely heal, but that new node will end up being a conduit for every single chunk as it pulls from the nodes that have the chunks and has those chunks subsequently pulled from it by those nodes with the stuck-high puller Indexes for the existing nodes, but starting-at-one indexes for all bins from the new node.

### Possible solution
The obvious solution is that peers need to reset their Intervals for a peer that has been nuked.   Currently, puller peerIndexes are only reset if the storageRadius reduces (an impossibility with a storageRadius=0 swarm, and an unlikely event in a large active swarm).
https://github.com/ethersphere/bee/blob/6fb07f335dd02e1118cab5cac40282b5aa5551b4/pkg/puller/puller.go#L147

The only method of detecting that a peer has been nuked that I have come up with is to detect that his cursors have reduced.   These cursors are pulled at the top of syncPeer but are not currently persisted.
https://github.com/ethersphere/bee/blob/6fb07f335dd02e1118cab5cac40282b5aa5551b4/pkg/puller/puller.go#L215

I propose that the cursors for a peer be recorded, not for recalling for use, but to detect that the cursor for a given peer's bin has reduced.  If a reduction in any cursor is detected, clear the Intervals for that peer so that pulling will start at index 0 for all bins allowing a full pull from that nuke-detected peer.   This would be a peer-specific version of resetIntervals because you really don't want to start all over on everyone!

Unfortunately, this proposal is beyond my golang capabilities, or I might attempt a PR.  Storing the cursors is the part that I'm not 100% positive of how to accomplish.

istae:
### Checklist

### Description

See issues below.

The reserve now has a epoch timestamp, and this timestamp is exchanged with peers when the pullsync cursors are requested.
in the case of a nuke or db reset, this epoch will be reset, and when this happens, the syncing peer will reset it's own synced intervals with that peer.

closes #4211 
closes #3821


pkg/pullsync/pb/pullsync.proto
```diff
@@ -11,7 +11,8 @@ option go_package = "pb";
 message Syn {}
 
 message Ack {
-  repeated uint64 Cursors = 1;
+  uint64 Epoch = 1;
```
mrekucci, line 14: If you mark `Epoch` as `2`, protobuf takes care of backwards compatibility and you don't need to change the `protocolVersion` below (which is more for backwards incompatible logical changes).

pkg/storer/internal/reserve/reserve.go
```diff
@@ -70,6 +71,19 @@ func New(
...
 
+	epochItem := &EpochItem{}
+	err = store.Get(epochItem)
+	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			err := store.Put(&EpochItem{Timestamp: uint64(time.Now().Unix())})
+			if err != nil {
+				return nil, err
+			}
+		} else {
```
mrekucci, line 82: This `else` branch is not needed, replace it with `return nil, err`.

istae: it's needed because there is more code that needs to run below

acha-bill (approved):

mrekucci (approved):

PR #4297, opened 2023-09-07
fix: include the reason when the postage contract call fails

### LINKED ISSUE #3320: Not much information why postage stamp buying fails
### Context
Bee 1.8.1

### Summary
I was trying to buy a postage stamp and it failed. I tried to find out what is the problem and neither the error messages neither the logs are helpful.

The error I get from the endpoint is: `Error: Internal Server Error: cannot create batch`, status 500.

In the logs I only found this logline (with INFO verbosity):
```
"time"="2022-09-21 16:42:36.763231" "level"="error" "logger"="node/api" "msg"="create batch: create failed"
```

With DEBUG verbosity an additional line is printed:
```
"time"="2022-09-21 16:42:36.763211" "level"="debug" "logger"="node/api" "msg"="create batch: create failed" "error"="create batch: depth 17 bucketDepth 16 immutable false: transaction reverted"
```

The problem seems to be that the gas limit is too low for the transaction, but that can be deduced only by looking at the transactions (in blockscout).

### Expected behavior
It would be good if either the error message or at least the logs would give some information to why there was an error.

mrekucci:
…tract

### Checklist

### Description

Indicate the reason for the unsuccessful call of the postage contract.

Closes #3320

### DIFF
--- a/pkg/postage/postagecontract/contract.go
+++ b/pkg/postage/postagecontract/contract.go
@@ -199,7 +199,13 @@ func (c *postageContract) sendTransaction(ctx context.Context, callData []byte, 
 	}
 
 	if receipt.Status == 0 {
-		return nil, transaction.ErrTransactionReverted
+		err := transaction.ErrTransactionReverted
+		if res, cErr := c.transactionService.Call(ctx, request); cErr == nil {
+			if reason, uErr := abi.UnpackRevert(res); uErr == nil {
+				err = fmt.Errorf("%w: reason: %s", err, reason)
+			}
+		}
+		return nil, err
 	}
 
 	return receipt, nil


pkg/postage/postagecontract/contract.go
```diff
@@ -199,7 +199,13 @@ func (c *postageContract) sendTransaction(ctx context.Context, callData []byte,
 	}
 
 	if receipt.Status == 0 {
-		return nil, transaction.ErrTransactionReverted
+		err := transaction.ErrTransactionReverted
+		if res, cErr := c.transactionService.Call(ctx, request); cErr == nil {
+			if reason, uErr := abi.UnpackRevert(res); uErr == nil {
+				err = fmt.Errorf("%w; reason: %s", err, reason)
```
istae, line 205: is the semicolon on purpose?

istae (approved):

pkg/postage/postagecontract/contract.go
mrekucci: > is the semicolon on purpose?

yes

istae: Don't we usually use a colon?

notanatol: I think this approach should suffice:
https://github.com/ethersphere/bee/blob/08bbf64b9513826e72f9f949c6cf0530fc4fdd26/pkg/node/node.go#L1287

notanatol (approved):

PR #4299, opened 2023-09-08
fix(chunkstore): increase refCnt capacity

### LINKED ISSUE #4284: Limited to 256 pins of a single chunk via multiple references?
### Context
bee 1.17.3

### Summary
It hasn't happened yet (actually, it has, see comment below), but I individually pin lots of different references, some of which share very common chunks (see https://discord.com/channels/799027393297514537/1142109674477473883).  I just noticed that the chunkstore reference counter is only a uint8.  This means that any given chunk can only be pinned a maximum of 256 times.   And after that, I suspect the reference count will go to zero which can cause serious issues by marking the sharky entry free when in fact it is still referenced by the remainder of the localstore.

https://github.com/ethersphere/bee/blob/44846bba2287e2ce117baa253454befd6313161f/pkg/storer/internal/chunkstore/chunkstore.go#L41

In point of fact, with my millions of pinned PNG references, many of which may share the IEND chunks, I wouldn't be surprised if my OSM upload/pinner node has already wrapped the refCnt on some of my chunks, exposing my node to severe (and potentially non-recoverable) chunkstore/sharky corruption...  (It has, see comment and attached file below).

### Expected behavior
I would not expect such a small limit on the allowed reference counts of a single chunk.

### Actual behavior
Theoretically, if a given chunk is included in multiple parent pins, the reference count can be overflowed.

### Steps to reproduce
Create a file, then put that unencrypted file reference into 257 manifests with other unique additional files.  Then pin all 257 parent manifest references.  This should create 257 parent pins, each of which will increment the shared child reference chunk(s) resulting in a refCnt overflow in the chunkstore.

### Possible solution
The chunkstore refCnt should be a much larger size such that it cannot (reasonably) happen, say maybe a Uint32?

istae:
### Checklist

### Description
closes #4284

### DIFF
--- a/pkg/storer/internal/chunkstore/chunkstore.go
+++ b/pkg/storer/internal/chunkstore/chunkstore.go
@@ -28,7 +28,7 @@ var (
 	errUnmarshalInvalidRetrievalIndexLocationBytes = errors.New("unmarshal RetrievalIndexItem: invalid location bytes")
 )
 
-const RetrievalIndexItemSize = swarm.HashSize + 8 + sharky.LocationSize + 8
+const RetrievalIndexItemSize = swarm.HashSize + 8 + sharky.LocationSize + 4
 
 var _ storage.Item = (*RetrievalIndexItem)(nil)
 
@@ -67,7 +67,7 @@ func (r *RetrievalIndexItem) Marshal() ([]byte, error) {
 	i += 8
 
 	copy(buf[i:i+7], locBuf)
-	i += 7
+	i += sharky.LocationSize
 
 	binary.LittleEndian.PutUint32(buf[i:], r.RefCnt)
 
@@ -93,7 +93,7 @@ func (r *RetrievalIndexItem) Unmarshal(buf []byte) error {
 		return errUnmarshalInvalidRetrievalIndexLocationBytes
 	}
 	ni.Location = *loc
-	i += 7
+	i += sharky.LocationSize
 
 	ni.RefCnt = binary.LittleEndian.Uint32(buf[i:])
 
--- a/pkg/storer/internal/chunkstore/chunkstore.go
+++ b/pkg/storer/internal/chunkstore/chunkstore.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/storage/storageutil"
 	"github.com/ethersphere/bee/pkg/swarm"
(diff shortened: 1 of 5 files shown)


pkg/storer/internal/chunkstore/chunkstore.go
```diff
@@ -70,16 +79,24 @@ func (r *RetrievalIndexItem) Unmarshal(buf []byte) error {
 		return errUnmarshalInvalidRetrievalIndexSize
 	}
 
+	i := 0
+	ni := new(RetrievalIndexItem)
+
+	ni.Address = swarm.NewAddress(append(make([]byte, 0, swarm.HashSize), buf[i:i+swarm.HashSize]...))
```
mrekucci, line 85: You can use `swarm.NewAddress(slices.Clone(buf[i:i+swarm.HashSize]))` instead.

pkg/storer/migration/step_05.go
```diff
@@ -0,0 +1,160 @@
...
+
+import (
+	"context"
+	"encoding/binary"
+	"errors"
+	"os"
+
+	"github.com/ethersphere/bee/pkg/log"
+	"github.com/ethersphere/bee/pkg/sharky"
+	storage "github.com/ethersphere/bee/pkg/storage"
```
mrekucci, line 15: This "storage" import rename is not needed.

pkg/storer/migration/step_05.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/ethersphere/bee/pkg/storage/storageutil"
+	"github.com/ethersphere/bee/pkg/storer/internal/chunkstore"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+const oldretrievalIndexItemSize = swarm.HashSize + 8 + sharky.LocationSize + 1
```
mrekucci, line 21: `oldretrievalIndexItemSize` -> `oldRetrievalIndexItemSize`

pkg/storer/migration/step_05_test.go
```diff
@@ -0,0 +1,52 @@
...
+// license that can be found in the LICENSE file.
+
+package migration_test
+
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
```
mrekucci, line 12: Unnecessary newline.

mrekucci (approved):

mrekucci (approved):

PR #4309, opened 2023-09-12
fix: pre and post localstore initiation migration steps

istae:
### Checklist

### Description
Because the retrievalIndexItem's field is changing, we need to make sure all old items are migrated before any other recovery and migration takes place due to unmarshalling errors.

As a result, the migration is now in two parts, pre and post migration. The pre migration will run before any other recovery and migration process of the localstore.

The PR also contains a new field in /debugstore


pkg/storage/migration/migration.go
```diff
@@ -28,12 +28,12 @@ var (
 )
 
 // Migrate migrates the storage to the latest version
-func Migrate(s storage.BatchedStore, sm Steps) error {
+func Migrate(s storage.BatchedStore, group string, sm Steps) error {
```
mrekucci, line 31: This `group` parameter should be properly documented, e.g.: what effect it has.

pkg/storer/migration/all_steps.go
```diff
@@ -10,8 +10,8 @@ import (
 	"github.com/ethersphere/bee/pkg/storer/internal/reserve"
 )
 
-// AllSteps lists all migration steps for localstore IndexStore.
-func AllSteps(
+// PostSteps lists all migration steps for localstore IndexStore after the localstore is intiated.
```
mrekucci, line 13: I'd suggest naming it `before` and `after`.

pkg/storer/debug.go
```diff
@@ -30,10 +31,11 @@ type CacheStat struct {
 }
 
 type ReserveStat struct {
```
mrekucci, line 33: It would be good if this is reflected in the OpenAPI spec.

mrekucci (approved):

pkg/storer/debug.go
istae: we dont have an entry in openAPI for /debugstore

mrekucci: If it is not a temporary API, it should be documented in the OpenAPI specification.

mrekucci (approved):

pkg/storer/migration/all_steps.go
```diff
@@ -21,6 +21,12 @@ func AllSteps(
 		2: step_02,
 		3: step_03(chunkStore, reserve.ChunkType),
 		4: step_04(sharkyPath, sharkyNoOfShards),
-		5: step_05,
+	}
+}
+
+// BeforeIinitSteps lists all migration steps for localstore IndexStore before the localstore is intiated.
+func BeforeIinitSteps() migration.Steps {
```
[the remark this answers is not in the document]
istae: The problem with this approach is that it gets messy very quickly. For example, if sharky recovery runs when running the new release, which item do you use?

PR #4310, opened 2023-09-12
chore: remove retrieval ctx scope leak

janos:
### Checklist

### Description
This PR removes the need for topCtx variable that needs to creep into the goroutine that is created in the retrieval singleflight to preserve the tracing span. This is possible by upgrading the singleflight to the latest v0.4.0 release which preserves values between function calls, while keeping the same cancelation behaviour.

It is optional to wait for bee to support building with bee 1.21 and use context.WithoutCancel in the retrieval goroutine.

The side effect of the singleflight upgrade is the usage of type parameters to avoid runtime type casts.

General code improvement.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -51,7 +51,7 @@ require (
 	golang.org/x/time v0.3.0
 	gopkg.in/yaml.v2 v2.4.0
 	resenje.org/multex v0.1.0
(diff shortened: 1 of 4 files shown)


mrekucci (approved):
LGTM!

istae (approved):

PR #4321, opened 2023-09-14
feat: propagate stream error to origin node for pushsync and retrieval protocols

istae:
### Checklist

### Description
Introduces a simple mechanism to return error messages for pushsync and retrieval protocols so the origin node can log the error messages.
Bumps protocol versions.

To be included in the upcoming breaking release (ph4).


pkg/p2p/p2p.go
```diff
@@ -222,3 +223,30 @@ const (
 func NewSwarmStreamName(protocol, version, stream string) string {
 	return "/swarm/" + protocol + "/" + version + "/" + stream
 }
+
+var errDelivery = errors.New("received delivery error msg")
```
mrekucci, line 227: This error is not necessary, I'd recommend doing the following:
```Golang
// DeliveryError contains information about a failed chunk delivery to a peer.
type DeliveryError struct {
	chunk swarm.Chunk
	peer  swarm.Address
	msg   string
}

// Error implements the error interface.
func (e *DeliveryError) Error() string {
	return fmt.Sprintf("delivery of %s for %s: %s", e.chunk.Address().String(), e.peer.String(), e.msg)
}

// NewDeliveryError is a convenience constructor for DeliveryError.
func NewDeliveryError(chunk swarm.Chunk, peer swarm.Address, msg string) error {
	return &DeliveryError{
		chunk: chunk,
		peer:  peer,
		msg:   msg,
	}
}
```

The `Unwrap` method is not needed in this case.

istae: thanks for the suggestion, I made the change but without peer or chunk address because the existing logs take care of this

mrekucci (approved):

PR #4327, opened 2023-09-20
fix: save stamps for all upload types regardless of any errors 

istae:
### Checklist

### Description
The stamp indexes are saved to disk after every upload session regardless of upload result.

### DIFF
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -276,7 +276,7 @@ func (s *Service) postageGetStampBucketsHandler(w http.ResponseWriter, r *http.R
 	}
 	hexBatchID := hex.EncodeToString(paths.BatchID)
 
-	issuer, save, err := s.post.GetStampIssuer(paths.BatchID)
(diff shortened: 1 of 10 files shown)


pkg/api/postage.go
```diff
@@ -276,7 +276,7 @@ func (s *Service) postageGetStampBucketsHandler(w http.ResponseWriter, r *http.R
 	}
 	hexBatchID := hex.EncodeToString(paths.BatchID)
 
-	issuer, save, err := s.post.GetStampIssuer(paths.BatchID)
+	issuer, _, err := s.post.GetStampIssuer(paths.BatchID)
```
mrekucci, line 279: Not sure about this. Why is the `save` func ignored?

istae: because the stamp issuer is not updated, it's a read only operation

mrekucci (approved):

mrekucci (approved):

PR #4358, opened 2023-09-29
fix: include the reason when the redistribution contract call fails

mrekucci:
### Checklist

### Description
Indicate the reason for the unsuccessful call of the redistribution contract.

### DIFF
--- a/pkg/postage/postagecontract/contract.go
+++ b/pkg/postage/postagecontract/contract.go
@@ -178,7 +178,7 @@ func (c *postageContract) sendApproveTransaction(ctx context.Context, amount *bi
 	return receipt, nil
 }
 
-func (c *postageContract) sendTransaction(ctx context.Context, callData []byte, desc string) (*types.Receipt, error) {
+func (c *postageContract) sendTransaction(ctx context.Context, callData []byte, desc string) (receipt *types.Receipt, err error) {
 	request := &transaction.TxRequest{
 		To:          &c.postageStampContractAddress,
 		Data:        callData,
@@ -187,25 +187,22 @@ func (c *postageContract) sendTransaction(ctx context.Context, callData []byte, 
 		Value:       big.NewInt(0),
 		Description: desc,
 	}
+	defer func() {
+		err = c.transactionService.UnwrapRevertReason(ctx, request, err)
+	}()
 
 	txHash, err := c.transactionService.Send(ctx, request, transaction.DefaultTipBoostPercent)
 	if err != nil {
 		return nil, err
 	}
 
-	receipt, err := c.transactionService.WaitForReceipt(ctx, txHash)
+	receipt, err = c.transactionService.WaitForReceipt(ctx, txHash)
 	if err != nil {
 		return nil, err
 	}
 
 	if receipt.Status == 0 {
-		err := transaction.ErrTransactionReverted
-		if res, cErr := c.transactionService.Call(ctx, request); cErr == nil {
-			if reason, uErr := abi.UnpackRevert(res); uErr == nil {
-				err = fmt.Errorf("%w: reason: %s", err, reason)
-			}
-		}
-		return nil, err
+		return nil, transaction.ErrTransactionReverted
 	}
 
 	return receipt, nil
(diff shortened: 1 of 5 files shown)


pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -191,7 +191,7 @@ func (c *contract) sendAndWait(ctx context.Context, request *transaction.TxReque
 	}
 
 	if receipt.Status == 0 {
-		return txHash, transaction.ErrTransactionReverted
+		return txHash, c.txService.UnwrapRevertReason(ctx, request, transaction.ErrTransactionReverted)
```
nugaon, line 194: The process seems to fail before it would get a receipt. Supposedly the `EstimateGas` call (invoked by `c.txService.Send` above) returns with `exection reverted`.

mrekucci: Fixed.

istae (approved):

pkg/postage/postagecontract/contract.go
```diff
@@ -199,13 +199,7 @@ func (c *postageContract) sendTransaction(ctx context.Context, callData []byte,
...
 
 	if receipt.Status == 0 {
-		err := transaction.ErrTransactionReverted
-		if res, cErr := c.transactionService.Call(ctx, request); cErr == nil {
-			if reason, uErr := abi.UnpackRevert(res); uErr == nil {
-				err = fmt.Errorf("%w: reason: %s", err, reason)
-			}
-		}
-		return nil, err
+		return nil, c.transactionService.UnwrapRevertReason(ctx, request, transaction.ErrTransactionReverted)
```
nugaon, line 202: could be deferred at the starting of the function such as in redistribution.go

nugaon (approved):
seems good for our current problem

pkg/transaction/transaction.go
```diff
@@ -584,3 +588,17 @@ func (t *transactionService) TransactionFee(ctx context.Context, txHash common.H
...
 	return trx.Cost(), nil
 }
+
+func (t *transactionService) UnwrapRevertReason(ctx context.Context, req *TxRequest, err error) error {
+	if err == nil {
+		return nil
+	}
+
+	if res, cErr := t.Call(ctx, req); cErr == nil {
+		if reason, uErr := abi.UnpackRevert(res); uErr == nil {
```
janos, line 598: UnapckRevert will unpack abi encoded "Error(string)", but I do not think that it will unpack custom errors that are for example used in the redistribution contract. To unpack custom errors, contract abi is required, get the error with abi.ABI.ErrorByID using the first 4 bytes of the data and use the retuned *abi.Error to Unpack the error.

I did not try it, in this PR, but used this approach to get custom errors from the call traces.

pkg/postage/postagecontract/contract.go
mrekucci: Done.

pkg/transaction/transaction.go
janos: This function may benefit from unit tests with different types of revert errors to ensure that they are not silently ignored.

mrekucci: Yes, this is for revert reason as specified in: https://solidity.readthedocs.io/en/latest/control-structures.html#revert

janos: Exactly, this is the one type of revert, there may be custom errors packed, and custom errors are used in our contracts. They supposed to be packed differently as they have arguments and custom name.

mrekucci: Can the custom errors be rewritten into this generic form so the error extraction is unified?

janos: They can be just strings, but sometimes additional values are useful to be passed. In any case, solidity allows custom revert errors and I do not think that they are unpacked here.

mrekucci: Ok, merging this. The custom errors need to be reviewed and handled in different PR.

n00b21337: > Can the custom errors be rewritten into this generic form so the error extraction is unified?

actually no, this generic form wasted a lot of space so was turned into a less wasteful form of generating errors, which in turn saves gas and in the end money for users. Its not a big deal to use them just needs to be unpacked properly will show examples of what is the difference, should be easy enough.

PR #4403, opened 2023-10-16
chore: lower min depth

istae:
### Checklist


mrekucci (commented):
The `postage_api.patch` should also be deleted and removed from the `beekeeper.yml`.

istae:
@mrekucci we can keep the min depth at 17 which is the lowest allowed depth anyways by the contract so no need to remove the patch.

mrekucci (approved):

PR #4416, opened 2023-10-20
fix: delete expired stamp issuers

istae:
### Checklist

### DIFF
--- a/pkg/node/statestore.go
+++ b/pkg/node/statestore.go
@@ -55,7 +55,7 @@ func InitStamperStore(logger log.Logger, dataDir string, stateStore storage.Stat
 	if err != nil {
 		return nil, err
 	}
-	// TODO: remove migration
+	// TODO: remove migration after it has been a few months after the localstoreV2 release
 	err = migrateStamperData(stateStore, stamperStore)
 	if err != nil {
 		stamperStore.Close()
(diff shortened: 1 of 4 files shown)


pkg/node/statestore.go
```diff
@@ -55,6 +55,7 @@ func InitStamperStore(logger log.Logger, dataDir string, stateStore storage.Stat
 	if err != nil {
 		return nil, err
 	}
+	// TODO: remove migration
```
mrekucci, line 58: Add conditions to the comment when the migration should be removed.

pkg/postage/service.go
```diff
@@ -202,22 +191,48 @@ func (ps *service) HandleStampExpiry(id []byte) {
...
 		exists, err := ps.postageStore.Exists(issuer.ID())
 		if err != nil {
 			ps.logger.Error(err, "set expired: checking if issuer exists", "id", issuer.ID())
 			return err
 		}
-		issuer.SetExpired(!exists)
+		if !exists {
+			err := ps.store.Delete(&StampIssuerItem{Issuer: issuer})
+			if err != nil {
+				ps.logger.Error(err, "set expired: delete stamp data", "id", issuer.ID())
```
mrekucci, line 208: If the node can continue its operation then this should be on the `Debug`/`Waring` level not `Error`.

pkg/postage/service_test.go
```diff
@@ -201,3 +202,40 @@ func TestGetStampIssuer(t *testing.T) {
 		}
 	})
 }
+
+func TestSetExpired(t *testing.T) {
+	store := inmemstore.New()
+	defer store.Close()
```
mrekucci, line 208: Use `t.Cleanup(store.Close())` instead.

mrekucci (approved):

PR #4428, opened 2023-10-26
perf(pin_api): fetch chunks in parallel

### LINKED ISSUE #4427: POST /pins/ref should fetch and store chunks in parallel

istae:
### Checklist

### Description
closes #4427

### DIFF
--- a/pkg/api/pin.go
+++ b/pkg/api/pin.go
@@ -55,6 +55,7 @@ func (s *Service) pinRootHash(w http.ResponseWriter, r *http.Request) {
 	sem := semaphore.NewWeighted(100)
 	var errTraverse error
 	var mtxErr sync.Mutex
+	var wg sync.WaitGroup
 
 	err = traverser.Traverse(
 		r.Context(),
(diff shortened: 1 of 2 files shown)


pkg/api/pin.go
```diff
@@ -49,28 +52,60 @@ func (s *Service) pinRootHash(w http.ResponseWriter, r *http.Request) {
 	getter := s.storer.Download(true)
 	traverser := traversal.New(getter)
 
+	sem := semaphore.NewWeighted(100)
+	var errTraverse error
+	var mtxErr sync.Mutex
+	var wg sync.WaitGroup
```
mrekucci, line 58: You can use "golang.org/x/sync/errgroup" instead and get rid of the mutex and error.

istae: i thought about using that but I still need access to the internal error inside to goroutine to stop the traversal.

istae:
it's capped at 100 parallel retrieval requests

ldeffenb:
Ah, I see the weighted semaphore now!   That's what I get for commenting before actually studying, rather than skimming, the code!

PR #4433, opened 2023-10-27
fix: cache and stamper store

istae:
### Checklist

### Description
Includes various improvements to how the cache works.
The storer now has a cache worker that keeps track of the cache size and evicts as necessary.
This means that other cache ops do not need to remove any items before inserting entries.

In the postage service, the set expire func now removes all expired stamp issuers and stamp data.

Added new metrics to cachestore.

### DIFF
--- a/pkg/postage/service.go
+++ b/pkg/postage/service.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/ethersphere/bee/pkg/log"
 	"github.com/ethersphere/bee/pkg/storage"
+	"golang.org/x/sync/errgroup"
 )
 
 // loggerName is the tree path name of the logger for this package.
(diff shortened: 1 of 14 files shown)


pkg/postage/service.go
```diff
@@ -193,7 +194,57 @@ func (ps *service) HandleStampExpiry(id []byte) {
...
+				if !exists {
+					count++
+
+					select {
+					case deleteItemC <- item:
+					case <-ctx.Done():
+						return false, ctx.Err()
+					}
+
+					if count%100_000 == 0 {
```
acha-bill, line 235: Where does 100k come from?

acha-bill (approved):

pkg/postage/service.go
istae: a random number used ask a checkpoint to show the user that the process is on going

Issue #4341, opened 2023-09-23
replace storer transactions with leveldb batches


istae:
the chunkstore operations do not currently accept the leveldb batches for writes.
this can cause issues especially in areas of the localstore where the chunkstore is used in tandem with other stores that does indeed utilizes batches for write (eg reserve.Put).

the chunkstore transaction implementation is too complex and bug prone.
transactions for sharky especially is not needed, if a db operation fails, the written data and used slot can simply remain as is and do not need to be recovered.
we have `compact` cmd now which can give the used sharky space back to the node.

also the `reverseOP` transaction idea for the indexstore should ideally be completely dropped and all writes must be done using the 'leveldb.batch`.

istae:
The localstore should ideally looks like this.

Parent calls are exported, core functions of the storer pkg, like Reserve.Put.
The parent func initiates a Transaction. 
Any internal storage calls the parent function makes must utilize the batch, and cannot independently make any Writes to the storage. 

The transaction is internally a leveldb.Batch and exposes the storage calls like `Put, Delete, Get, Commit`.
The transaction also exposes the Chunkstore.

Any write ops are accumulated in the batch of the transcation.

Any sharky writes are immediately written to disk.
Sharky releases are kept in memory until the batch is commited.

sharky_write -> write to disk, keep write locations in memory
sharky_release -> keep release location in memory
batch_write -> write to batch
batch_commit -> commit batch to disk, release location in disk
if batch_commit fails, release all sharky_write location, no ops sharky_release

istae:
wrapSync should be removed with a simple mutex in chunkstore

nikipapadatou:
more discussions are needed between bee team engineers to proceed with this.

PR #4449, opened 2023-11-08
fix: replace chunk locker with mutext

notanatol:
### Checklist

### Description
replace chunk locker mechanism with an stdlib's mutex

Related: #4341

### DIFF
--- a/pkg/storage/repository.go
+++ b/pkg/storage/repository.go
@@ -7,6 +7,7 @@ package storage
 import (
 	"context"
 	"errors"
+	"sync"
 	"time"
 
 	m "github.com/ethersphere/bee/pkg/metrics"
@@ -29,7 +30,6 @@ type repository struct {
 
 	txIndexStore TxStore
 	txChunkStore TxChunkStore
-	locker       ChunkLocker
 }
 
 // IndexStore returns Store.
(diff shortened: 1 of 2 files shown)


ldeffenb:
Are you planning any performance impact tests on this change?   Seems to me that changing from a (presumably) chunk-specific lock to a global local on the syncChunkStore is going to seriously throttle (slow down) parallel operations in the node.  But then again, something needs to better protect against the random issues we see in the relational integrity of the chunkStore.

notanatol:
Yes, according to these two benchmarks (that both use `wrapSync`) the performance is comparable, most likely to the previous implementation of locker not doing any meaningful guarding.
current branch:
```
BenchmarkReservePutter-8  8481	    152460 ns/op	   27886 B/op	     396 allocs/op
BenchmarkCachePutter-8  6631	    180415 ns/op	   20683 B/op	     320 allocs/op
```
master:
```
BenchmarkReservePutter-8  7214	    154002 ns/op	   27914 B/op	     397 allocs/op
BenchmarkCachePutter-8   6877	    184118 ns/op	   20587 B/op	     322 allocs/op
```
however this task has been de-prioritized and it will be continued sometime next year, I'll be moving it into draft state now.

istae:
this is being handled in this PR https://github.com/ethersphere/bee/pull/4561

PR #4491, opened 2023-12-04
feat: non-local redundancy

nugaon:
Along with the neighborhood reserve replication and node caching, Reed-Solomon erasure coding and dispersed replicas make the redundancy in Bee.

From these new redundancies, the client can recover requested data from different neighborhoods of the kademlia network.

Details in the PR descriptions
- [x] #4429
- [x] #4448 
- [x] #4453
- [x] #4492
- [x] #4507
- [x] #4522 
- [x] #4529 

I think we need to change the version after #4529


pkg/encryption/encryption.go
```diff
@@ -185,3 +185,10 @@ func GenerateRandomKey(l int) Key {
 	}
 	return key
 }
+
+func min(a, b int) int {
```
notanatol, line 189: this function should already be built into go1.21

pkg/file/redundancy/redundancy.go
```diff
@@ -0,0 +1,204 @@
...
+	parities := p.Parities(shards)
+
+	n := shards + parities
+	// realloc for parity chunks if it does not override the prev one
+	// caculate parity chunks
+	enc, err := erasureEncoderFunc(shards, parities)
+	if err != nil {
+		return err
+	}
+	// make parity data
```
mrekucci, line 156: This comment doesn't add much value. The same for most of the comments below. Comment the why, not the what, the idea being that the code should be self-documenting and the comments should only be a last resort for explaining stuff. Comments can get out of sync, but the code is always the code, so there's no need for unnecessary redundancy.

pkg/file/redundancy/redundancy_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"sync"
+	"testing"
+
+	"github.com/ethersphere/bee/pkg/file/pipeline"
+	"github.com/ethersphere/bee/pkg/file/pipeline/bmt"
+	"github.com/ethersphere/bee/pkg/file/redundancy"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// MOCK ENCODER
```
mrekucci, line 20: Unnecessary comment.

pkg/file/redundancy/redundancy_test.go
```diff
@@ -0,0 +1,158 @@
...
+	for i := m.shards; i < m.shards+m.parities; i++ {
+		data := make([]byte, 32)
+		data[swarm.SpanSize-1], data[swarm.SpanSize] = uint8(indicatedValue), uint8(indicatedValue)
+		buffer[i] = data
+		indicatedValue++
+	}
+	return nil
+}
+
+// PARITY CHAIN WRITER
```
mrekucci, line 46: Unnecessary comment.

pkg/file/redundancy/redundancy_test.go
```diff
@@ -0,0 +1,158 @@
...
+	chainWriteCalls int
+	sumCalls        int
+	validCalls      []bool
+}
+
+func NewParityChainWriter() *ParityChainWriter {
+	return &ParityChainWriter{}
+}
+
+// ACCESSORS
```
mrekucci, line 59: Unnecessary comment.

pkg/file/redundancy/redundancy_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (c *ParityChainWriter) ChainWriteCalls() int {
+	c.Lock()
+	defer c.Unlock()
+	return c.chainWriteCalls
+}
+func (c *ParityChainWriter) SumCalls() int { c.Lock(); defer c.Unlock(); return c.sumCalls }
+
+// METHODS
```
mrekucci, line 68: Unnecessary comment.

pkg/file/redundancy/redundancy_test.go
```diff
@@ -0,0 +1,158 @@
...
+					parityCallback := func(level int, span, address []byte) error {
+						parityCount++
+						return nil
+					}
+
+					for i := 0; i < shardCount; i++ {
+						buffer := make([]byte, 32)
+						_, err := io.ReadFull(rand.Reader, buffer)
+						if err != nil {
+							t.Error(err)
```
mrekucci, line 120: Why is it feasible to continue to the `params.ChunkWrite` when an error is encountered?

pkg/file/redundancy/redundancy_test.go
```diff
@@ -0,0 +1,158 @@
...
+						err = params.ChunkWrite(0, buffer, parityCallback)
+						if err != nil {
+							t.Error(err)
+						}
+					}
+					if shardCount != maxShards {
+						// encode should be called automatically when reaching maxshards
+						err := params.Encode(0, parityCallback)
+						if err != nil {
+							t.Error(err)
```
mrekucci, line 131: Why continue to the checks when an error is encountered?

pkg/file/redundancy/table.go
```diff
@@ -0,0 +1,64 @@
...
+	for k, s := range et.shards {
+		if maxShards >= s {
+			return et.parities[k]
+		}
+	}
+	return 0
+}
+
+// getMinShards returns back the minimum shard number respect to the given parity number
+func (et *erasureTable) GetMinShards(parities int) (int, error) {
```
mrekucci, line 57: This function is not used. Also, the comment should start with the function name.

pkg/replicas/getter.go
```diff
@@ -0,0 +1,163 @@
+// Copyright 2023 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// the code below implements the integration of dispersed replicas in chunk fetching.
```
mrekucci, line 5: Package comment should be of the form 'Package replicas ...'

pkg/replicas/getter.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// ErrSwarmageddon is returned in case of a vis mayor called Swarmageddon.
+// Swarmageddon is the situation when none of the replicas can be retrieved.
+// If 2^{depth} replicas were uploaded and they all have valid postage stamps
+// then the probability of Swarmageddon is less than 0.000001
+// assuming the error rate of chunk retrievals stays below the level expressed
+// as depth by the publisher.
+type ErrSwarmageddon struct {
```
mrekucci, line 27: This error type doesn't add much clarity since the `errors.Join` can be returned directly and checked with `errors.Is` without wrapping it to an extra layer. Also, the error is only used in the tests.

pkg/replicas/getter.go
```diff
@@ -0,0 +1,163 @@
...
+func (g *getter) Get(ctx context.Context, addr swarm.Address) (ch swarm.Chunk, err error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// channel that the results (retrieved chunks) are gathered to from concurrent
+	// workers each fetching a replica
+	resultC := make(chan swarm.Chunk)
+	// errc collects the errors
+	errc := make(chan error, 17)
+	var errs []error
```
mrekucci, line 72: Not need to be a slice, the `errors.Join` will handle the wrapping. Also, it should be declared as close to the for loop as possible.

pkg/replicas/getter.go
```diff
@@ -0,0 +1,163 @@
...
+	defer ticker.Stop()
+	for level := uint8(0); level <= uint8(g.level); {
+		select {
+		// at least one chunk is retrieved, cancel the rest and return early
+		case chunk := <-resultC:
+			cancel()
+			return chunk, nil
+
+		case err = <-errc:
+			errs = append(errs, err)
```
mrekucci, line 111: `errs = errors.Join(errs, err)`

pkg/replicas/putter.go
```diff
@@ -0,0 +1,60 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
mrekucci, line 1: `2020` -> `2023`

pkg/replicas/putter.go
```diff
@@ -0,0 +1,60 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// the code below implements the integration of dispersed replicas in chunk upload.
```
mrekucci, line 5: Package comment should be of the form 'Package replicas ...'

pkg/replicas/putter.go
```diff
@@ -0,0 +1,60 @@
...
+
+// NewPutter is the putter constructor
+func NewPutter(p storage.Putter) storage.Putter {
+	return &putter{p}
+}
+
+// Put makes the getter satisfy the storage.Getter interface
+func (p *putter) Put(ctx context.Context, ch swarm.Chunk) (err error) {
+	rlevel := getLevelFromContext(ctx)
+	errs := []error{p.putter.Put(ctx, ch)}
```
mrekucci, line 33: The error doesn't need to be a slice: `errs := p.putter.Put(ctx, ch)`.

pkg/replicas/putter.go
```diff
@@ -0,0 +1,60 @@
...
+				err = p.putter.Put(ctx, sch)
+			}
+			errc <- err
+		}()
+	}
+
+	wg.Wait()
+	close(errc)
+	for err := range errc {
+		errs = append(errs, err)
```
mrekucci, line 57: `errs = append(errs, err)` - > `errs = errors.Join(errs, err)`

pkg/replicas/putter.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	return errors.Join(errs...)
```
mrekucci, line 59: `return errors.Join(errs...)` -> `return errs`

pkg/replicas/replicas.go
```diff
@@ -0,0 +1,145 @@
+// Copyright 2020 The Swarm Authors. All rights reserved.
```
mrekucci, line 1: `2020` -> `2023`

pkg/replicas/replicas.go
```diff
@@ -0,0 +1,145 @@
...
+	//
+	// counts of replicas used for levels of increasing security
+	// the actual number of replicas needed to keep the error rate below 1/10^6
+	// for the five levels of redundancy are 0, 2, 4, 5, 19
+	// we use an approximation as the successive powers of 2
+	counts     = [5]int{0, 2, 4, 8, 16}
+	sums       = [5]int{0, 2, 6, 14, 30}
+	privKey, _ = crypto.DecodeSecp256k1PrivateKey(append([]byte{1}, make([]byte, 31)...))
+	signer     = crypto.NewDefaultSigner(privKey)
+	owner, _   = hex.DecodeString("dc5b20847f43d67928f49cd4f85d696b5a7617b5")
```
mrekucci, line 39: Where is this hex string coming from?

pkg/replicas/replicas.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+// SetLevel sets the redundancy level in the context
+func SetLevel(ctx context.Context, level redundancy.Level) context.Context {
```
mrekucci, line 43: Doesn't need to be exported since it is used only in the tests (use the `export_test.go` file to re-export only for tests).

pkg/replicas/replicas.go
```diff
@@ -0,0 +1,145 @@
...
+
+// replicate returns a replica params strucure seeded with a byte of entropy as argument
+func (rr *replicator) replicate(i uint8) (sp *replica) {
+	// change the last byte of the address to create SOC ID
+	id := make([]byte, 32)
+	copy(id, rr.addr)
+	id[0] = i
+	// calculate SOC address for potential replica
+	h := swarm.NewHasher()
+	_, _ = h.Write(id)
```
mrekucci, line 91: It should be stated why are the returned errors ignored.

pkg/encryption/encryption.go
nugaon: ah yeah, you are right since then I updated :)

pkg/file/joiner/joiner_test.go
```diff
@@ -951,3 +953,121 @@ func TestJoinerIterateChunkAddresses_Encrypted(t *testing.T) {
...
+			readCheck()
+
+			// remove data chunks in order to trigger recovery
+			maxShards := tc.rLevel.GetMaxShards()
+			maxParities := tc.rLevel.GetParities(maxShards)
+			if tc.encryptChunk {
+				maxShards = tc.rLevel.GetMaxEncShards()
+				maxParities = tc.rLevel.GetEncParities(maxShards)
+			}
+			removeCount := maxParities
```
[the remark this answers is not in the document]
nugaon: min func should be used instead of that, but I understand why are you saying. 
fixed.

pkg/file/pipeline/hashtrie/hashtrie.go
```diff
@@ -20,57 +21,85 @@ var (
...
+	if h.chunkCounters[level] == h.maxChildrenChunks {
+		// at this point the erasure coded chunks have been written
+		err := h.wrapFullLevel(level)
+		return err
 	}
 	return nil
 }
 
+func (h *hashTrieWriter) writeToDataLevel(span, ref, key, data []byte) error {
+	// write dataChunks to the level above
```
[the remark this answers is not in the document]
nugaon: actually it regards to that one single function call right after.
I'm going to write a description to this method tho. :)

pkg/file/redundancy/getter/getter.go
```diff
@@ -0,0 +1,333 @@
...
+
+	"github.com/ethersphere/bee/pkg/encryption/store"
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/swarm"
+	"github.com/klauspost/reedsolomon"
+)
+
+/// ERRORS
+
+type cannotRecoverError struct {
```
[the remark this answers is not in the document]
nugaon: then I cannot export error as you suggest that below

pkg/file/redundancy/getter/getter.go
```diff
@@ -0,0 +1,333 @@
...
+// getter retrieves children of an intermediate chunk
+// it caches sibling chunks if erasure decoding was called on the level already
+type getter struct {
+	storage.Getter
+	storage.Putter
+	mu          sync.Mutex
+	sAddresses  []swarm.Address          // shard addresses
+	pAddresses  []swarm.Address          // parity addresses
+	cache       map[string]inflightChunk // map from chunk address to cached shard chunk data
+	erasureData [][]byte                 // data + parity shards for erasure decoding; TODO mutex
```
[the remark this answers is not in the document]
nugaon: not anymore

pkg/file/redundancy/getter/getter.go
```diff
@@ -0,0 +1,333 @@
...
+	shards := len(sAddresses)
+	parities := len(pAddresses)
+	n := shards + parities
+	erasureData := make([][]byte, n)
+	cache := make(map[string]inflightChunk, n)
+	// init cache
+	for i, addr := range sAddresses {
+		cache[addr.String()] = inflightChunk{
+			pos: i,
+			// wait channel initialization is needed when recovery starts
```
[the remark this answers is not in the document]
nugaon: yes, that's why I commented on its place where/when the initialization happens

pkg/file/redundancy/level.go
```diff
@@ -0,0 +1,137 @@
...
+
+const (
+	NONE Level = iota
+	MEDIUM
+	STRONG
+	INSANE
+	PARANOID
+)
+
+const maxLevel = 8
```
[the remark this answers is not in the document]
nugaon: it wasn't related to that, but I removed since it is not used

pkg/file/redundancy/table.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// newErasureTable initializes a shards<->parities table
+//
+//	 the value order must be strictly descending in both arrays
+//		example usage:
+//			shards := []int{94, 68, 46, 28, 14, 5, 1}
+//			parities := []int{9, 8, 7, 6, 5, 4, 3}
+//			var et = newErasureTable(shards, parities)
+func newErasureTable(shards, parities []int) *erasureTable {
```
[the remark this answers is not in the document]
nugaon: nothing : D still learning go

pkg/file/redundancy/redundancy_test.go
nugaon: I don't understand what you mean by that, the `t.Error(err)` will stop the test with error, won't it?

nugaon: aa, I see, so shall I use Fatal?

pkg/replicas/replicas.go
nugaon: private key `[]byte{1}` should result in this address

pkg/replicas/replicas.go
nugaon: used in #4492

pkg/replicas/replicas.go
mrekucci: Please add this as a comment.

pkg/api/dirs.go
```diff
@@ -63,6 +64,16 @@ func (s *Service) dirUploadHandler(
 	}
 	defer r.Body.Close()
 
+	rLevelNum, err := strconv.ParseUint(r.Header.Get(SwarmRedunancyLevel), 10, 1)
```
ldeffenb, line 67: What is the ,1?   According to the docs:

```The bitSize argument specifies the integer type that the result must fit into. Bit sizes 0, 8, 16, 32, and 64 correspond to int, int8, int16, int32, and int64. If bitSize is below 0 or above 64, an error is returned.```

File: pkg/api/api.go [hunk 31]
```diff
@@ -79,6 +80,7 @@ const (
 	SwarmCollectionHeader     = "Swarm-Collection"
 	SwarmPostageBatchIdHeader = "Swarm-Postage-Batch-Id"
 	SwarmDeferredUploadHeader = "Swarm-Deferred-Upload"
+	SwarmRedunancyLevel       = "Swarm-Redundancy-Level"
```
ldeffenb, line 83: `SwarmRedunancyLevel` is missing a "d".

ldeffenb (changes requested):
Just noting the points that make no sense as I try to understand the upcoming redundancy improvements.

pkg/api/dirs.go
nugaon: great catch ;)

pkg/replicas/replicas.go
nugaon: resolved in #4522

pkg/api/dirs.go
```diff
@@ -63,6 +64,16 @@ func (s *Service) dirUploadHandler(
 	}
 	defer r.Body.Close()
 
+	rLevelNum, err := strconv.ParseUint(r.Header.Get(SwarmRedundancyLevelHeader), 10, 8)
```
istae, line 67: see how we read headers here https://github.com/ethersphere/bee/blob/master/pkg/api/bzz.go#L39

pkg/api/bzz.go
```diff
@@ -452,7 +456,10 @@ func (s *Service) serveManifestEntry(
 // downloadHandler contains common logic for dowloading Swarm file from API
 func (s *Service) downloadHandler(logger log.Logger, w http.ResponseWriter, r *http.Request, reference swarm.Address, additionalHeaders http.Header, etag bool) {
 	headers := struct {
-		Cache *bool `map:"Swarm-Cache"`
+		Cache                 *bool           `map:"Swarm-Cache"`
+		Strategy              getter.Strategy `map:"Swarm-Redundancy-Strategy"`
+		FallbackMode          bool            `map:"Swarm-Redundancy-Fallback-Mode"`
+		ChunkRetrievalTimeout time.Duration   `map:"Swarm-Chunk-Retrieval-Timeout"`
```
istae, line 462: what is the purpose of this?

pkg/api/bzz.go
```diff
@@ -462,7 +469,12 @@ func (s *Service) downloadHandler(logger log.Logger, w http.ResponseWriter, r *h
 	if headers.Cache != nil {
 		cache = *headers.Cache
 	}
-	reader, l, err := joiner.New(r.Context(), s.storer.Download(cache), reference)
+
+	ctx := r.Context()
+	ctx = getter.SetStrategy(ctx, headers.Strategy)
+	ctx = getter.SetStrict(ctx, headers.FallbackMode)
+	ctx = getter.SetFetchTimeout(ctx, headers.ChunkRetrievalTimeout)
+	reader, l, err := joiner.New(ctx, s.storer.Download(cache), s.storer.Cache(), reference)
```
istae, line 477: storer.Cache() is not a reliable source for storing chunks for any operation, what is the purpose of using the cache here?

istae (changes requested):

pkg/file/redundancy/level.go
```diff
@@ -0,0 +1,176 @@
...
+const (
+	// no redundancy will be added
+	NONE Level = iota
+	// expected 1% chunk retrieval error rate
+	MEDIUM
+	// expected 5% chunk retrieval error rate
+	STRONG
+	// expected 10% chunk retrieval error rate
+	INSANE
+	// expected 50% chunk retrieval error rate
```
mrekucci, line 28: This number seems to be incorrect, we had a much higher number when performing the tests. The same goes for the above.

pkg/file/redundancy/level.go
```diff
@@ -0,0 +1,176 @@
...
+	case MEDIUM:
+		return mediumEt, nil
+	case STRONG:
+		return strongEt, nil
+	case INSANE:
+		return insaneEt, nil
+	case PARANOID:
+		return paranoidEt, nil
+	default:
+		return erasureTable{}, errors.New("redundancy: level NONE does not have erasure table")
```
mrekucci, line 76: Since `Level` is an opaque type, any `unit8` can be cast into the `Level`, thus the default case won't always translate into the `NONE`. The `NONE` case should be handled separately from the default (unknown) case. The same is below.

pkg/file/redundancy/level.go
```diff
@@ -0,0 +1,176 @@
...
+	[]int{
+		88, 85, 82, 79, 76, 72, 69, 66, 62, 59,
+		55, 51, 48, 44, 39, 35, 30, 24,
+	},
+)
+
+// DISPERSED REPLICAS INIT
+
+// GetReplicaCounts returns back the ascending dispersed replica counts for all redundancy levels
+func GetReplicaCounts() [5]int {
```
mrekucci, line 168: This getter is unnecessary, just export the  `replicaCounts`.

pkg/file/redundancy/level.go
```diff
@@ -0,0 +1,176 @@
...
+		18, 17, 16, 15, 14, 13, 12, 11, 10, 9,
+		8, 7, 6, 5, 4, 3, 2, 1,
+	},
+	[]int{
+		88, 85, 82, 79, 76, 72, 69, 66, 62, 59,
+		55, 51, 48, 44, 39, 35, 30, 24,
+	},
+)
+
+// DISPERSED REPLICAS INIT
```
mrekucci, line 165: Comment without added value.

pkg/replicas/getter.go
mrekucci: Still unresolved.

pkg/replicas/getter.go
mrekucci: Still unresolved.

pkg/replicas/getter.go
mrekucci: Still unresolved.

pkg/replicas/getter.go
mrekucci: Still unresolved.

pkg/replicas/putter.go
mrekucci: Still unresolved.

pkg/replicas/putter.go
mrekucci: Still unresolved.

pkg/replicas/putter.go
mrekucci: Still unresolved.

pkg/replicas/putter.go
mrekucci: Still unresolved.

pkg/replicas/putter.go
mrekucci: Still unresolved.

pkg/replicas/replicas.go
mrekucci: Still unresolved.

pkg/replicas/replicas.go
mrekucci: Still unresolved.

pkg/file/redundancy/level.go
```diff
@@ -0,0 +1,176 @@
...
+	STRONG
+	// expected 10% chunk retrieval error rate
+	INSANE
+	// expected 50% chunk retrieval error rate
+	PARANOID
+)
+
+// NewLevel returns a Level coresponding to the passed number parameter
+// throws an error if there is no level for the passed number
+func NewLevel(n uint8) (Level, error) {
```
mrekucci, line 34: This should be called `ParseLevel` and should take a `string`.

pkg/api/bzz.go
nugaon: what should I put here? there is no other putter available under `s.storer`

pkg/file/redundancy/level.go
nugaon: what do you mean by "much higher number"? you got errors only with higher retrieval error rate?

pkg/api/dirs.go
nugaon: removed and took value from caller after header parse.

pkg/file/redundancy/level.go
istae: @mrekucci i think this 50% is the error rate for not being able to find chunks, not the data growth percentage

pkg/api/bzz.go
istae: what is the purpose of using the cache here?

zelig: so cache must be used to store things locally that have no postage stamp. There are three ways to get there:
- evicted from reserve (either too distant, too cheap or expired)
- landed with us through retrieval and has no postage stamp (yet)
- and now also created with the help of parities,  no postage stamp available (yet).

pkg/file/redundancy/level.go
zelig: an array is an lvalue it is always copied (as opposed to passed by reference like slices)

pkg/api/bzz.go
```diff
@@ -462,7 +469,12 @@ func (s *Service) downloadHandler(logger log.Logger, w http.ResponseWriter, r *h
 	if headers.Cache != nil {
 		cache = *headers.Cache
 	}
-	reader, l, err := joiner.New(r.Context(), s.storer.Download(cache), reference)
+
+	ctx := r.Context()
+	ctx = getter.SetStrategy(ctx, headers.Strategy)
+	ctx = getter.SetStrict(ctx, headers.FallbackMode)
+	ctx = getter.SetFetchTimeout(ctx, headers.ChunkRetrievalTimeout)
```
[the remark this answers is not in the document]
zelig: within the getter itself

pkg/api/bzz.go
zelig: @istae Configuring this counts as an advanced dev option. 
By setting chunk retrieval timeout to a suffienctly short interval one can effectively render some slow chunks not retrievable and thus 'simulate' chunk loss even in environments of perfect availability. 
This is especially useful since it allows for client side trigger of redundancy-based recovery of data implemented in this PR.

pkg/api/bzz.go
zelig: Lets get to the bottom of this. 

Reconstructed chunks have no postage stamp therefore they cannot be in the reserve. Putting them in a pinstore is wasteful and unclear in terms of expiry. 
The only possible place to put is the localstore cache.

The status of reconstructed chunks is similar to chunks obtained through retrieval which is also cached. 
The only thing we absolutely need to make sure of is that these cached chunks can and will be put in the reserve once they are offered by peers as part of pullsyncing as a chunk with a  valid postage stamp.

There is one caveat about using the cache is the intricate scenario described and resolved in this hackmdQq https://hackmd.io/@zelig/Bkqol64dp

ldeffenb: An even worse consideration is that the cache size is user-specified and can therefore be set to an arbitrarily small value, possibly ensuring that the reconstructed chunks have already been purged before they are needed again.

pkg/api/bzz.go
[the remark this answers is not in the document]
zelig: in the new PR, you see it set via `SetParamsInContext` function in `api/bzz.go` 
https://github.com/ethersphere/bee/pull/4529/files#diff-271c890949913a595e694b46ef1980ab35c3e76734f1c469d5e1308d4750b645R293

and the default is coming from the `getParamsFromContext` in `getter/strategy.go
https://github.com/ethersphere/bee/pull/4529/files#diff-c862612c9989df4553c26eff52fa807f29796c148561c900dc6fac1b82197029R35

pkg/file/pipeline/hashtrie/hashtrie_test.go
```diff
@@ -34,13 +46,47 @@ func init() {
 	binary.LittleEndian.PutUint64(span, 1)
 }
 
+// NewErasureHashTrieWriter returns back an redundancy param and a HastTrieWriter pipeline
```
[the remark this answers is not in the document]
zelig: resolves in https://github.com/ethersphere/bee/pull/4529

pkg/file/redundancy/getter/getter.go
```diff
@@ -0,0 +1,226 @@
...
+	cancel     func()          // cancel function for RS decoding
+	remove     func()          // callback to remove decoder from decoders cache
+}
+
+type Getter interface {
+	storage.Getter
+	io.Closer
+}
+
+// New returns a decoder object used tos retrieve children of an intermediate chunk
```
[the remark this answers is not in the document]
zelig: resolves in https://github.com/ethersphere/bee/pull/4529

pkg/file/redundancy/getter/strategies.go
```diff
@@ -0,0 +1,128 @@
...
+	defer g.remove()
+	var cancels []func()
+	cancelAll := func() {
+		for _, cancel := range cancels {
+			cancel()
+		}
+	}
+	defer cancelAll()
+	run := func(s Strategy) error {
+		if s == PROX { // NOT IMPLEMENTED
```
istae, line 68: what happens if strategy is DATA? because the strategy is incremented below, then for cases where the strategy is DATA, no other strategy is tried because PROX is not implemented ?

pkg/file/redundancy/getter/strategies.go
```diff
@@ -0,0 +1,128 @@
...
+			return fmt.Errorf("strategy %d not implemented", s)
+		}
+
+		var stop <-chan time.Time
+		if s < RACE {
+			timer := time.NewTimer(strategyTimeout)
```
istae, line 74: max time allowed to fetch a chunk from the network is 500 ms?

pkg/file/redundancy/getter/strategies.go
```diff
@@ -0,0 +1,128 @@
...
+	return context.WithValue(ctx, strategyKey{}, s)
+}
+
+// SetStrict sets the strict mode for the retrieval
+func SetStrict(ctx context.Context, strict bool) context.Context {
+	return context.WithValue(ctx, modeKey{}, strict)
+}
+
+func (g *decoder) prefetch(ctx context.Context, strategy int, strict bool, strategyTimeout, fetchTimeout time.Duration) {
+	if strict && strategy == NONE {
```
istae, line 56:
```suggestion
	if (strict && strategy == NONE) || strategy == PROX {
```

pkg/file/redundancy/getter/getter.go
```diff
@@ -0,0 +1,226 @@
...
+func New(addrs []swarm.Address, shardCnt int, g storage.Getter, p storage.Putter, strategy Strategy, strict bool, fetchTimeout time.Duration, remove func()) Getter {
+	ctx, cancel := context.WithCancel(context.Background())
+	size := len(addrs)
+	if fetchTimeout == 0 {
+		fetchTimeout = 30 * time.Second
```
istae, line 50: defaults should not defined in the constructor, import it from the retrieval pkg

pkg/file/redundancy/getter/getter.go
```diff
@@ -0,0 +1,226 @@
...
+// missing gathers missing data shards not yet retrieved
+// it sets the chunk as inflight and returns the index of the missing data shards
+func (g *decoder) missing() (m []int) {
+	for i := 0; i < g.shardCnt; i++ {
+		select {
+		case <-g.waits[i]: // if chunk is retrieved, ignore
+			continue
+		default:
+		}
+		_ = g.fly(i, true) // commit (RS) or will commit to retrieve the chunk
```
istae, line 163: why does missing call set the chunks in flight?

pkg/file/redundancy/getter/getter.go
```diff
@@ -0,0 +1,226 @@
...
+		rsg.prefetch(ctx, strategy, strict, strategyTimeout, fetchTimeout)
+		rsg.wg.Done()
+	}()
+	return rsg
+}
+
+// Get will call parities and other sibling chunks if the chunk address cannot be retrieved
+// assumes it is called for data shards only
+func (g *decoder) Get(ctx context.Context, addr swarm.Address) (swarm.Chunk, error) {
+	i, ok := g.cache[addr.ByteString()]
```
istae, line 87: is this cache even necessary?
would the `GET` call receive an addr not part of this decoder??

pkg/file/redundancy/getter/getter.go
```diff
@@ -0,0 +1,226 @@
...
+	if !ok {
+		return nil, storage.ErrNotFound
+	}
+	if g.fly(i, true) {
```
istae, line 91: the `prefetch` calls essentially sets all of these chunks to inflight, no?
why is this fly check necessary, why can't we jump to the select below?

pkg/file/redundancy/getter/getter.go
```diff
@@ -0,0 +1,226 @@
...
+	//  write chunk to rsbuf and signal waiters
+	g.rsbuf[i] = ch.Data() // save the chunk in the RS buffer
+	if i < len(g.waits) {
+		close(g.waits[i]) // signal that the chunk is retrieved
+	}
+
+	// if all chunks are retrieved, signal ready
+	n := g.fetchedCnt.Add(1)
+	if n == int32(g.shardCnt) {
+		close(g.ready) // signal that just enough chunks are retrieved for decoding
```
istae, line 150: are we allowed to close ready if the shardCnt includes parity chunks?

pkg/file/redundancy/getter/getter.go
```diff
@@ -0,0 +1,226 @@
...
+	go func() {
+		rsg.prefetch(ctx, strategy, strict, strategyTimeout, fetchTimeout)
+		rsg.wg.Done()
+	}()
+	return rsg
+}
+
+// Get will call parities and other sibling chunks if the chunk address cannot be retrieved
+// assumes it is called for data shards only
+func (g *decoder) Get(ctx context.Context, addr swarm.Address) (swarm.Chunk, error) {
```
istae, line 86: what this `Get` should ideally do is to wait for either the chunk channel `g.waits[i]` to finish OR wait for a signal for the recovery to finish.
there should also be an error channel that returns when a chunk is not fetchable AND unreconvery which would translate into a `storage.ErrNotFound`.

as it stands, this `GET` nevers properly returns an error, and simply waits for a context timeout in this case that recovery and/or fetch failed.

pkg/file/redundancy/getter/getter.go
zelig: surely

pkg/file/redundancy/getter/getter.go
zelig: because we want singleflight behaviour on fetching, ie if prefetching fetches a chunk,, then its queried with joiner Get --> decoder Get (or the other way round), then we should just wait on the inflight fetch.
Similarly, when prefetch fetched shardCnt chunks the other chunks can be put to inflight, so that if they are `Get-ed' by the joiner they are not fetched but wait till revovered.

zelig: And of course there is not always a prefetch on every data chunk (NONE and PROX will not prefetch some chunks)

pkg/replicas/replicas.go
zelig: resolved in #4529

pkg/api/bzz.go
zelig: I am introducing a recoder accordingly 
resolved in #4529

openapi/SwarmCommon.yaml
```diff
@@ -934,6 +934,52 @@ components:
...
+    SwarmRedundancyStrategyParameter:
+      in: header
+      name: swarm-redundancy-strategy
+      schema:
+        type: integer
+        enum: [0, 1, 2, 3]
+      required: false
+      description: >
+        Specify the retrieve strategy on redundant data.
+        The mumbers stand for NONE, DATA, PROX and RACE, respectively.
```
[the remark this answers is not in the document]
zelig: resolved in #4529

pkg/api/bzz.go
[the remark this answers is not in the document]
zelig: resolved in #4529

notanatol (approved):

acha-bill (dismissed):

istae (approved):

acha-bill (approved):

PR #4492, opened 2023-12-06
feat: replicas integration

nugaon:
Integration of #4453 

### Checklist

### DIFF
--- a/pkg/file/pipeline/hashtrie/hashtrie.go
+++ b/pkg/file/pipeline/hashtrie/hashtrie.go
@@ -56,7 +56,7 @@ func NewHashTrieWriter(
 		chunkCounters:          make([]uint8, 9),
 		effectiveChunkCounters: make([]uint8, 9),
 		maxChildrenChunks:      uint8(rParams.MaxShards() + rParams.Parities(rParams.MaxShards())),
-		replicaPutter:          replicaPutter,
+		replicaPutter:          replicas.NewPutter(replicaPutter),
 	}
 	h.parityChunkFn = func(level int, span, address []byte) error {
 		return h.writeToIntermediateLevel(level, true, span, address, []byte{})
@@ -268,8 +268,7 @@ func (h *hashTrieWriter) Sum() ([]byte, error) {
 		if err != nil {
 			return nil, err
 		}
-		putter := replicas.NewPutter(h.replicaPutter)
-		err = putter.Put(h.ctx, swarm.NewChunk(swarm.NewAddress(rootHash), rootData))
+		err = h.replicaPutter.Put(h.ctx, swarm.NewChunk(swarm.NewAddress(rootHash), rootData))
 		if err != nil {
 			return nil, fmt.Errorf("hashtrie: cannot put dispersed replica %s", err.Error())
 		}
--- a/pkg/file/pipeline/hashtrie/hashtrie.go
+++ b/pkg/file/pipeline/hashtrie/hashtrie.go
@@ -5,11 +5,15 @@
 package hashtrie
 
 import (
+	"context"
 	"encoding/binary"
 	"errors"
+	"fmt"
 
 	"github.com/ethersphere/bee/pkg/file/pipeline"
 	"github.com/ethersphere/bee/pkg/file/redundancy"
+	"github.com/ethersphere/bee/pkg/replicas"
+	"github.com/ethersphere/bee/pkg/storage"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
@@ -21,6 +25,7 @@ var (
 const maxLevel = 8
 
 type hashTrieWriter struct {
+	ctx                    context.Context // context for put function of dispersed replica chunks
 	refSize                int
 	cursors                []int  // level cursors, key is level. level 0 is data level holds how many chunks were processed. Intermediate higher levels will always have LOWER cursor values.
 	buffer                 []byte // keeps intermediate level data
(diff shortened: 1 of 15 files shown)


pkg/replicas/replicas.go
```diff
@@ -125,21 +112,30 @@ func (rr *replicator) replicas() {
 }
 
 // add inserts the soc replica into a replicator so that addresses are balanced
-func (rr *replicator) add(r *replica, d uint8) (depth int, rank int) {
-	if d == 0 {
+func (rr *replicator) add(r *replica, rLevel redundancy.Level) (depth int, rank int) {
```
zelig, line 115: these changes do not improve the code,

pkg/file/redundancy/level.go
```diff
@@ -135,3 +150,19 @@ var encParanoidEt = newErasureTable(
 		55, 51, 48, 44, 39, 35, 30, 24,
 	},
 )
+
```
zelig, line 153: These changes expose a lot of convoluted internals to public which is a sign that they are not at the right place. 
I think this should remain as per the original, honestly.

pkg/file/redundancy/level.go
```diff
@@ -82,6 +82,21 @@ func (l Level) GetMaxEncShards() int {
 	return (swarm.Branches - p) / 2
 }
 
+// GetReplicaCount returns back the dispersed replica number
```
zelig, line 85: These changes expose a lot of convoluted internals to public which is a sign that they are not at the right place.
I think this should remain as per the original, honestly.

pkg/file/pipeline/hashtrie/hashtrie.go
```diff
@@ -21,20 +25,29 @@ var (
 const maxLevel = 8
 
 type hashTrieWriter struct {
+	ctx                    context.Context // context for put function of dispersed replica chunks
```
zelig, line 28: context should not be a field if possible

pkg/file/pipeline/hashtrie/hashtrie.go
```diff
@@ -21,20 +25,29 @@ var (
...
 	pipelineFn             pipeline.PipelineFunc
 	rParams                redundancy.IParams
 	parityChunkFn          redundancy.ParityChunkCallback
-	chunkCounters          []uint8 // counts the chunk references in intermediate chunks. key is the chunk level.
-	effectiveChunkCounters []uint8 // counts the effective  chunk references in intermediate chunks. key is the chunk level.
-	maxChildrenChunks      uint8   // maximum number of chunk references in intermediate chunks.
+	chunkCounters          []uint8        // counts the chunk references in intermediate chunks. key is the chunk level.
+	effectiveChunkCounters []uint8        // counts the effective  chunk references in intermediate chunks. key is the chunk level.
+	maxChildrenChunks      uint8          // maximum number of chunk references in intermediate chunks.
+	replicaPutter          storage.Putter // putter to save dispersed replicas of the root chunk
```
zelig, line 39: if only used for replicas then this should be right away initialised  with `replicas.NewPutter`

pkg/file/pipeline/hashtrie/hashtrie.go
```diff
@@ -245,5 +259,19 @@ func (h *hashTrieWriter) Sum() ([]byte, error) {
...
-	return data[swarm.SpanSize:], nil
+	rootHash := data[swarm.SpanSize:]
+
+	// save disperse replicas of the root chunk
+	if h.rParams.Level() != redundancy.NONE {
+		rootData, err := h.rParams.GetRootData()
+		if err != nil {
+			return nil, err
+		}
+		putter := replicas.NewPutter(h.replicaPutter)
```
zelig, line 270: this should already be part of the `hashTrieWriter` inititialisation

pkg/file/pipeline/hashtrie/hashtrie.go
```diff
@@ -245,5 +259,19 @@ func (h *hashTrieWriter) Sum() ([]byte, error) {
...
+		err = putter.Put(h.ctx, swarm.NewChunk(swarm.NewAddress(rootHash), rootData))
```
zelig, line 271: here we should not use context from the `hashTrieWriter`. Either `Sum` should be extended with a context or in fact the whole integration is possibly at the wrong place. I think it should happen after and where Sum is called. There the putter/getter used underlyingly is known so its easy to then just wrap the putter in the replicas.Putter and use that to Put the chunk gotten from the getter.

zelig (changes requested):
I have reservations as to the content of this PR. 
It should really be super simple, the fact that it is not indicates design problem.

Also I am missing the replicas integration into the joiner. which should be trivial.

pkg/file/pipeline/hashtrie/hashtrie.go
nugaon: hashtrie's Sum() method should satisfy pipeline interface that does not define context.
context is needed to put SOC replica.

pkg/file/pipeline/hashtrie/hashtrie.go
nugaon: the redundancy cache is available in the hashtrie package from which the root chunk data is retrieved, so I would argue whether the integration is a great place.

nugaon:
[data race error](https://github.com/ethersphere/bee/actions/runs/7214662399/job/19657491688?pr=4492) will be fixed by the new refactored test cases in #4507

pkg/file/redundancy/level.go
nugaon: `GetReplicaCount` is called in hashtrie_test.go and IMO it is like erasure tables.
I moved indexBases back to replicas packages as we discussed offline.

PR #4507, opened 2023-12-13
feat(redundancy/getter): RS decoder rewrite

zelig:
rewrite of the `file/redunancy/getter` package with extensive tests
the getter now solicits initial strategies [NONE, DATA, PROX, RACE] with strict or fallback modes.
 A third parameter sets the timeout for individual chunk requests that when set below the maximum latency can efffectively simulate erasures. 
The API needs to support these as headers in a coming PR.


nugaon (commented):
CI cannot build the code. 
please reassign review when the test pass

Issue #3408, opened 2022-10-09
Add/create a linter for logger key-value pairs


mrekucci:
The current logger requires a key-value pair after the message string. This is often forgotten and only one value is added. Explore a way to create a custom liner that will check and enforce this pattern.

See: https://blog.sylver.dev/build-a-custom-go-linter-in-5-minutes

PR #4509, opened 2023-12-15
refactor: logger modifications

rolandlor:
### Checklist

- [X] I have filled out the description and linked the related issues.

### Description
The logger was modified to enforce the handling of key vale pairs after message string in logger.
Related issue: https://github.com/ethersphere/bee/issues/3408

### DIFF
--- a/cmd/bee/cmd/db.go
+++ b/cmd/bee/cmd/db.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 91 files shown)


mrekucci:
Thanks for the PR. There are two main issues I see with this approach (using a tuple struct):
1) it makes the logger API more chatty and less ergonomic
2) it could lead to an increase in the number of allocations and thus more pressure on the GC (this should be benchmarked)

PR #4529, opened 2023-12-21
feat: redundancy retrieve api

zelig:
the final push of non-local redundancy.
contains extensive joiner and bzz api tests to demonstrate the correctness of the various decoder prefetching strategies and to test header-based setting of paramater values.

The PR also contains many fixes:
- ~~remove langos and not use prebuffering on endoints~~ 
-  introduce fetchtimeout on decoder get calls
- decoder to handle short final chunks on all levels
- decoder defaults handling

UPDATE: the final push contains the following improvements:
- put back Langos lookaheadbuffer and use it if redundancy level is 0 or when strategy is DATA for continuity of expectations (to be calibrated separately)
- correct context and error handling in Get
- decoding defaults defined within package but assigned by the API in case header undefined.
- pseudorand moved ioutils->testutils


pkg/file/redundancy/getter/strategies.go
```diff
@@ -30,15 +30,39 @@ const (
...
 
 // GetParamsFromContext extracts the strategy and strict mode from the context
-func GetParamsFromContext(ctx context.Context) (s Strategy, strict bool, fetcherTimeout time.Duration) {
-	s, _ = ctx.Value(strategyKey{}).(Strategy)
-	strict, _ = ctx.Value(modeKey{}).(bool)
-	fetcherTimeout, _ = ctx.Value(fetcherTimeoutKey{}).(time.Duration)
-	return s, strict, fetcherTimeout
+func GetParamsFromContext(ctx context.Context) (s Strategy, strict bool, fetcherTimeout time.Duration, err error) {
+	var ok bool
+	s, strict, fetcherTimeoutVal := NONE, true, "30s"
```
istae, line 35: these defaults should be defined in the api package.
After parsing the headers, and before calling `SetParamsInContext`, the defaults should be assigned if the values were not passed in the header.

zelig: even though it is not the API's defaults but generic defaults for the functionality?

istae: the pkg can export the defaults and whichever pkg is setting the context params, they can use those exported defaults if not values are supplied.

pkg/file/redundancy/getter/getter.go
```diff
@@ -97,10 +98,39 @@ func (g *decoder) Get(ctx context.Context, addr swarm.Address) (swarm.Chunk, err
 	}
 	select {
 	case <-g.waits[i]:
-		return swarm.NewChunk(addr, g.rsbuf[i]), nil
 	case <-ctx.Done():
-		return nil, ctx.Err()
+		select {
```
istae, line 102: context deadline must be respected.
move the cases inside the inner select to the outer select.

pkg/file/joiner/joiner.go
```diff
@@ -118,20 +118,25 @@ func New(ctx context.Context, g storage.Getter, putter storage.Putter, address s
 	spanFn := func(data []byte) (redundancy.Level, int64) {
 		return 0, int64(bmt.LengthFromSpan(data[:swarm.SpanSize]))
 	}
-	var strategy getter.Strategy
-	var strict bool
-	var fetcherTimeout time.Duration
+	strategy, strict, fetcherTimeout, err := getter.GetParamsFromContext(ctx)
```
istae, line 121: it appears that `fetcherTimeout` 's only purpose is to allow for the testing of the feature, which imo, does not really belong in production code, so we have to test the feature in other way(s)

pkg/api/bzz_test.go
```diff
@@ -16,20 +16,219 @@ import (
...
+//
+//  1. upload a file with a given redundancy level and encryption
+//
+//  2. [positive test] download the file by the reference returned by the upload API response
+//     This uses range queries to target specific (number of) chunks of the file structure
+//     During path traversal in the swarm hash tree, the underlying mocksore (forgetting)
+//     is in 'recording' mode, flagging all the retrieved chunks as chunks to forget.
+//     This is to simulate the scenario where some of the chunks are not available/lost
+//     NOTE: For this to work one needs to switch off lookaheadbuffer functionality
+//     (see langos pkg)
```
acha-bill, line 57: Is this langos comment still relevant?

pkg/file/joiner/joiner.go
```diff
@@ -98,7 +98,7 @@ func (g *decoderCache) GetOrCreate(addrs []swarm.Address, shardCnt int) storage.
 // New creates a new Joiner. A Joiner provides Read, Seek and Size functionalities.
 func New(ctx context.Context, g storage.Getter, putter storage.Putter, address swarm.Address) (file.Joiner, int64, error) {
 	// retrieve the root chunk to read the total data length the be retrieved
-	rLevel := replicas.GetLevelFromContext(ctx)
+	rLevel := redundancy.GetLevelFromContext(ctx)
```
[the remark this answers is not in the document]
acha-bill: Also, the API doc says default is supposed to be NONE. Not PARANOID. Either the API doc needs updating or the default value needs to be updated.

pkg/storer/mock/forgetting.go
significance, line 1: 🔮

pkg/util/ioutil/pseudorand/reader.go
```diff
@@ -0,0 +1,182 @@
+// Copyright 2023 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// this is a pseudorandom reader that generates a deterministic
+// sequence of bytes based on the seed. It is used in tests to
+// enable large volumes of pseudorandom data to be generated
+// and compared without having to store the data in memory.
+package pseudorand
```
istae, line 9: this pkg probably belongs to the pkg/util/testutil directory

istae:
t.Parallel() should be used wherever possible

pkg/util/ioutil/pseudorand/reader.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// fill fills the buffer with the hash of the current segment.
+func (r *Reader) fill() {
+	if r.cur >= r.len {
+		return
+	}
+	bufSegments := bufSize / 32
+	start := r.cur / bufSegments
+	rem := (r.cur % bufSize) / 32
```
nugaon, line 174: first it is devided by 32 then you multiply it by that. it is not necessary

nugaon (commented):
all in all seems good for me though the tests fails.

pkg/api/bzz.go
```diff
@@ -497,7 +500,9 @@ func (s *Service) downloadHandler(logger log.Logger, w http.ResponseWriter, r *h
 	}
 	w.Header().Set(ContentLengthHeader, strconv.FormatInt(l, 10))
 	w.Header().Set("Access-Control-Expose-Headers", ContentDispositionHeader)
-	http.ServeContent(w, r, "", time.Now(), langos.NewBufferedLangos(reader, lookaheadBufferSize(l)))
+	http.ServeContent(w, r, "", time.Now(), reader)
+	// NOTE: temporary workaround for testing, watch this...
+	// http.ServeContent(w, r, "", time.Now(), langos.NewBufferedLangos(reader, lookaheadBufferSize(l)))
```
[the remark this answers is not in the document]
zelig: I will put it back

pkg/api/bzz_test.go
```diff
@@ -16,20 +16,219 @@ import (
...
+// download strategies, but only correct parameter passing and correct recovery functionality
+//
+// The test cases have the following structure:
+//
+//  1. upload a file with a given redundancy level and encryption
+//
+//  2. [positive test] download the file by the reference returned by the upload API response
+//     This uses range queries to target specific (number of) chunks of the file structure
+//     During path traversal in the swarm hash tree, the underlying mocksore (forgetting)
+//     is in 'recording' mode, flagging all the retrieved chunks as chunks to forget.
```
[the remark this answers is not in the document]
zelig: yes, in recording mode we can specify a set of chunks to forget in a meaningful way, i.e if we random access a range within a file, the chunks on the traversal path in the Swarm hash tree representing the file will be recorded as forgotten

pkg/file/joiner/joiner.go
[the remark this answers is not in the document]
zelig: First of all the default in the API doc talks about NONE being the default in the context of the encoding only.
As for the erasure coding, level is encoded in the span, so the level of redundancy is known by the decoder at download.

But in the case of replicas, the level is not known by the downloader. Therefore we choose to fall back to trying all replicas as if the uploader had been PARANOID.
The way replicas work for retrieval is that initially only the root chunk is retrieved, then we fall back to extending it to replicas which are present assuming the next level of security for encoding. Indeed in the definition the replicas for level 3 are a superset of level 3 and a subset of level 4.

Note that even if the encoding is 'known', it would mean we know that the publisher intended to upload say 4 replicas (level 2). However, by design replicas can be added by third parties, therefore you might as well try just in case.

Also, note that attempting replicas belonging only to higher levels of redundancy are attempted only when desparate, ie., if lower levels are not found.

A user configurablee header value should probably be introduced and set on the context so that the user can control the swap traffic.

pkg/file/joiner/joiner.go
zelig: well, i consider user-side testability of non-local redundancy in the production environment an advanced feature offering.

pkg/file/redundancy/getter/getter.go
zelig: yes, I did feel unconfortable about how this is handled. Unfortunately we cannot move from inside select to outer, but i will think about a better and correct way to handle this.

pkg/file/redundancy/getter/getter.go
```diff
@@ -97,10 +100,39 @@ func (g *decoder) Get(ctx context.Context, addr swarm.Address) (swarm.Chunk, err
...
+		copy(data, chdata)
 	}
+	g.rsbuf[i] = data
+}
+
+// getData returns the data shard from the RS buffer
+func (g *decoder) getData(i int) []byte {
+	if i == g.shardCnt-1 && g.lastLen > 0 {
+		return g.rsbuf[i][:g.lastLen] // cut padding
+	}
```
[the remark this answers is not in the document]
zelig: why calculate it when it is known from the retrieved chunk. 
we need to remember the length after. retrieval so that when it is padded to full for RS decoding (assumes uniform shard length)

pkg/file/redundancy/level.go
```diff
@@ -167,3 +168,19 @@ func GetReplicaCounts() [5]int {
...
+type levelKey struct{}
+
+// SetLevelInContext sets the redundancy level in the context
+func SetLevelInContext(ctx context.Context, level Level) context.Context {
+	return context.WithValue(ctx, levelKey{}, level)
+}
+
+// GetLevelFromContext is a helper function to extract the redundancy level from the context
+func GetLevelFromContext(ctx context.Context) Level {
+	rlevel := PARANOID
```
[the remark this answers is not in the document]
zelig: see also comment thread above

pkg/file/joiner/joiner.go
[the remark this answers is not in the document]
ldeffenb: > A user configurable header value should probably be introduced and set on the context so that the user can control the swap traffic.

My vote would go for that!  Especially when each node of a /bzz manifest retrieval would be using replicas as many of them fit easily within a single 4K chunk.

nugaon:
Overall, I think we should keep some pre-fetching mechanism for case when the data is not erasure coded. Without that, data retrieval may be really slow.

pkg/file/redundancy/getter/getter.go
zelig: it was not correct, fixed context logic

notanatol (approved):

acha-bill (approved):

istae (approved):

PR #3797, opened 2023-02-15
feat: modular key handling

notanatol:
### Checklist

- [X] I have read the [coding guide](https://github.com/ethersphere/bee/blob/master/CODING.md).
- [X] I have added tests to cover my changes.

### Description
This change will make switching between `secp` revisions very easy, and would make reverting of the libp2p functionality less painful.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ethersphere/bee/3797)

PR #4542, opened 2024-01-12
chore(linter): using binary staticcheck

notanatol:
### Checklist

### Description
recently I noticed that staticcheck messages in my IDE do not get picked by the `golangci-lint` here's why:

```
swarm:bee/ (fix-add-static-check) $ golangci-lint -v linters | grep staticcheck                                                                                                                                                                                                  [18:45:58]
INFO [config_reader] Config search paths: [./ /home/swarm/swarm/bee /home/swarm/swarm /home/swarm /home /]
INFO [config_reader] Used config file .golangci.yml
staticcheck (megacheck): It's a set of rules from staticcheck. It's not the same thing as the staticcheck binary. The author of staticcheck doesn't support or approve the use of staticcheck as a library inside golangci-lint. [fast: false, auto-fix: false]
```
The author of staticcheck insists the linter to be used as a separate binary.

This PR adds the check before the `golangci-lint` step and updates the `//nolint:` directives using the original syntax.

### DIFF
--- a/pkg/keystore/file/key.go
+++ b/pkg/keystore/file/key.go
@@ -83,6 +83,7 @@ func encryptKey(k *ecdsa.PrivateKey, password string, edg keystore.EDG) ([]byte,
 		}
 		addr = a
 	case elliptic.P256():
+		//lint:ignore SA1019 to be addressed soon
 		addr = elliptic.Marshal(elliptic.P256(), k.PublicKey.X, k.PublicKey.Y)
 	default:
 		return nil, fmt.Errorf("unsupported curve: %v", k.PublicKey.Curve)
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -28,7 +28,6 @@ linters:
     - prealloc
     - predeclared
     - promlinter
-    - staticcheck
     - thelper
     - typecheck
     - unconvert
--- a/Makefile
+++ b/Makefile
@@ -86,6 +86,7 @@ format:
 
 .PHONY: lint
 lint: linter
+	$(GOBIN)/staticcheck ./...
 	$(GOLANGCI_LINT) run ./...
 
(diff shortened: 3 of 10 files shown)


pkg/keystore/file/key.go
```diff
@@ -83,6 +83,7 @@ func encryptKey(k *ecdsa.PrivateKey, password string, edg keystore.EDG) ([]byte,
 		}
 		addr = a
 	case elliptic.P256():
+		//lint:ignore SA1019 to be addressed soon
 		addr = elliptic.Marshal(elliptic.P256(), k.PublicKey.X, k.PublicKey.Y)
```
notanatol, line 87: @janos this `elliptic.Marshal` has been deprecated, should it be replaced with someting like
```golang
pubKey, _ := k.ECDH()
addr = pubKey.Bytes()
```

janos: Hi @notanatol. Given that elliptic.P256 is used only for the libp2p and will not be used for the ethereum address that uses btcec.S256, it should be safe to use basically anything and the proposed solution should be ok. Current marshalling is not relevant to the libp2p key files as the Address field is not used for libp2p.

pkg/keystore/file/key.go
```diff
@@ -83,6 +83,7 @@ func encryptKey(k *ecdsa.PrivateKey, password string, edg keystore.EDG) ([]byte,
 		}
 		addr = a
 	case elliptic.P256():
+		//lint:ignore SA1019 to be addressed soon
```
janos, line 86: Actually, I believe that the most correct is to set the ethereum address to the empty string if the curve is not btcec.S256(), as that makes sense only for that curve. The switch was added on #3797 and elliptic.Marshal does not provide a valid ethereum address.

PR #4549, opened 2024-01-17
fix(reserve): evict just enough chunks to reach the capacity

### LINKED ISSUE #4538: evict just enough chunks of a batch to fall below the reserve capacity
To minimize chunks loss, we should only evict enough chunks fall below the capacity.
This can be done by passing a "max evict counter" calculated in the unreserve function and passed to the evict batch function.

The other required change is that the collected items should be sorted by binID first so that the older chunks are evicted first. 

Also, we can temporarily disable caching of evicted chunks until a performant and stable solution is found: see https://github.com/ethersphere/bee/issues/4530

istae:
### Checklist

closes #4538

### DIFF
--- a/pkg/storer/internal/reserve/reserve.go
+++ b/pkg/storer/internal/reserve/reserve.go
@@ -24,15 +24,7 @@ import (
 	"resenje.org/multex"
 )
 
-// loggerName is the tree path name of the logger for this package.
-const loggerName = "reserve"
 const reserveNamespace = "reserve"
-
-/*
-	pull by 	bin - binID
-	evict by 	bin - batchID
-	sample by 	bin
-*/
 
 type Reserve struct {
 	baseAddr     swarm.Address
@@ -59,7 +51,7 @@ func New(
 		baseAddr:     baseAddr,
 		capacity:     capacity,
 		radiusSetter: radiusSetter,
-		logger:       logger.WithName(loggerName).Register(),
+		logger:       logger.WithName(reserveNamespace).Register(),
 		mutx:         multex.New(),
 	}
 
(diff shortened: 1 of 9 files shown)


pkg/storer/internal/reserve/reserve.go
```diff
@@ -335,6 +360,14 @@ func (r *Reserve) EvictBatchBin(
 		return 0, err
 	}
 
+	// evict oldest chunks first
+	sort.Slice(evicted, func(i, j int) bool {
+		return evicted[i].BinID < evicted[j].BinID
```
ldeffenb, line 365: Color me not fully grokking this, but how is sorting by BinID getting us the "oldest chunks first" as the comment states?

istae: binID is an unique, incremented value that is assigned to each chunk in the reserve. So by removing the chunks with the lowest binID, we can at least bring some order (and not just random) as to which chunks should be removed first.

ldeffenb: So binID has nothing to do with the pullsync bins, nor anything else within bee that is called bin?   And since it is unique, I would gather that each "bin" contains but a single chunk?  Bizarre.

One of my personal software development mantras is "Naming can make or break a project".

istae: the reserve has "bins", which correspond to the proximity orders of the chunks (bin 0 = po 0, etc). so a "binID" is the unique ID of a chunk in the bin.

pullsync gets chunks from a peer using the peer's binIDs (so called bin intervals) for each bin/po.

notanatol (approved):

acha-bill (approved):

PR #4557, opened 2024-01-24
feat: periodically save stamp issuer data during upload

### LINKED ISSUE #4544: stamp issuer bucket data should be persisted more often during uploads
as part of the upload process, before chunks are sent out to the network, each chunk is stamped using the stamp issuer.
each stamp contains a specific bucket index that is incremented for every chunk that lands in that particular bucket.

Side note: There are by default 2^16 buckets per batch. The max hits per bucket is defined by the batch's depth.

The bucket and the data are kept in memory and are only persisted when the upload finishes. If for whatever reason the node shutdown in the middle of an upload, the newly assigned bucket indexes are not persisted to disk. This can cause uploads in the future to reuse those indices. In the case of mutable batches, chunks that were previously assigned to those indices will be replaced with the new chunks, causing data loss, and in the case of immutable batches, the new chunks will be rejected.

In short, it's important that the bucket data is persisted more often. How often have to be decided with caution as the bucket data as it is today is a few hundred kilobytes.

acha-bill:
### Checklist

### Description

Instead of saving the stamp issuer data only when a putter session is done, periodically save the stamp issuer data while the session is active.

### Related Issue (Optional)
Closes #4544

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -800,6 +800,53 @@ func (p *putterSessionWrapper) Cleanup() error {
 	return errors.Join(p.PutterSession.Cleanup(), p.save())
 }
 
+// putterSessionWrapperWithSaver is a wrapper around the putter session that saves the stamp issuer periodically while the session is active.
+type putterSessionWrapperWithSaver struct {
+	*putterSessionWrapper
+	shutdownC chan any
+	once      *sync.Once
+	logger    log.Logger
+}
+
+func (p *putterSessionWrapperWithSaver) Put(ctx context.Context, chunk swarm.Chunk) error {
+	err := p.putterSessionWrapper.Put(ctx, chunk)
+	if err != nil {
+		return err
+	}
+
+	// when upload starts and the first put is called, start the saver goroutine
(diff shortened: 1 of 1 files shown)


notanatol (approved):

pkg/api/api.go
```diff
@@ -800,6 +800,53 @@ func (p *putterSessionWrapper) Cleanup() error {
...
+	close(p.shutdownC)
+	return p.putterSessionWrapper.Done(ref)
+}
+
+func (p *putterSessionWrapperWithSaver) Cleanup() error {
+	close(p.shutdownC)
+	return p.putterSessionWrapper.Cleanup()
+}
+
+func (p *putterSessionWrapperWithSaver) saver() {
```
istae, line 834: let's add some unit tests please

pkg/api/api.go
```diff
@@ -800,6 +800,53 @@ func (p *putterSessionWrapper) Cleanup() error {
...
+	ticker := time.NewTicker(1 * time.Second)
```
istae, line 835: we need to measure/profile this change

Issue #4527, opened 2023-12-20
Upload w/Pinning duplicates RefCnt


ldeffenb:
### Context
1.18.2

### Summary
If a given content is uploaded multiple times with `swarm-pin: true` specified, the resulting refCnt for the involved chunks is left with extra increments based on how many times the content was uploaded with pins.  This will cause the content to never leave the localstore/sharky even after the single actual pin is removed.

### Expected behavior
If data is uploaded and pushed, the resulting refCnt should reflect just the actual single pin (and potential reserve status), not the redundant orphaned pins from the extra uploads.  Then if the pin is deleted, the content would be removed.

### Actual behavior
A new pinning collection is created and completed for each upload with suitable ChunkStore.Put() calls for each involved chunk.
https://github.com/ethersphere/bee/blob/ad2602c31c2b5a479ab8e76e793f3cf2ea548c71/pkg/storer/internal/pinning/pinning.go#L251
But when the collection is closed, the single root reference pinCollectionItem is updated to the new UUID orphaning the original UUID's pinChunkItems such that they will never be deleted causing their refCnt increments to persist forever preventing those chunks from ever leaving the localstore or sharky. 
https://github.com/ethersphere/bee/blob/ad2602c31c2b5a479ab8e76e793f3cf2ea548c71/pkg/storer/internal/pinning/pinning.go#L268
https://github.com/ethersphere/bee/blob/ad2602c31c2b5a479ab8e76e793f3cf2ea548c71/pkg/storer/internal/pinning/pinning.go#L269

Here are the extensive logs that I added to my bee node source to track down this behavior:
```
Original upload creates a new collection AND increments the stamp:

"time"="2023-12-20 13:43:13.786854" "level"="debug" "logger"="node/storer" "msg"="pinning.NewCollection" "UUID"="47ff794c-2b3c-4bf8-b7bd-67e9293f56b3"
"time"="2023-12-20 13:43:13.787038" "level"="debug" "logger"="node/postage" "msg"="StampIssuer.increment" "chunk"="ee77388fe89615fb7e07b5690a64a3ed8e43d8f206d51e56e4bebde4de9387f7" "batch"="b9ca120bc0adc895ab7c9ebaf1288838d5fbe956b3f935812917617869ee39f2" "bucket"=61047 "bCnt"=31 "maxused"=46 "upperbound"=256
[...]

PR #4558, opened 2024-01-25
fix(pinning): check before writing root address

notanatol:
### Checklist

### Description
Closes: #4527

### DIFF
--- a/pkg/storer/internal/pinning/pinning.go
+++ b/pkg/storer/internal/pinning/pinning.go
@@ -243,6 +243,16 @@ func (c *collectionPutter) Put(ctx context.Context, st internal.Storage, writer 
 		return nil
 	}
 
+	rootCollection := &pinCollectionItem{Addr: ch.Address()}
+	has, err := st.IndexStore().Has(rootCollection)
+	if err != nil {
+		return fmt.Errorf("pin store: check previous root: %w", err)
+	}
+	if has {
+		// duplicate PUT of the same root address
+		return nil
+	}
+
 	err = writer.Put(collectionChunk)
 	if err != nil {
 		return fmt.Errorf("pin store: failed putting collection chunk: %w", err)
@@ -264,11 +274,19 @@ func (c *collectionPutter) Close(st internal.Storage, writer storage.Writer, roo
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 
-	// Save the root pin reference.
-	c.collection.Addr = root
-	err := writer.Put(c.collection)
+	collection := &pinCollectionItem{Addr: root}
+	has, err := st.IndexStore().Has(collection)
 	if err != nil {
-		return fmt.Errorf("pin store: failed updating collection: %w", err)
+		return fmt.Errorf("pin store: check previous root: %w", err)
+	}
+
+	if !has {
+		// Save the root pin reference.
+		c.collection.Addr = root
+		err := writer.Put(c.collection)
+		if err != nil {
+			return fmt.Errorf("pin store: failed updating collection: %w", err)
+		}
 	}
 
 	err = writer.Delete(&dirtyCollection{UUID: c.collection.UUID})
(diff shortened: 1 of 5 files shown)


pkg/storer/internal/pinning/pinning.go
```diff
@@ -243,6 +243,16 @@ func (c *collectionPutter) Put(ctx context.Context, st internal.Storage, writer
 		return nil
 	}
 
+	rootCollection := &pinCollectionItem{Addr: ch.Address()}
+	has, err := st.IndexStore().Has(rootCollection)
+	if err != nil {
+		return fmt.Errorf("pin store: check previous root: %w", err)
+	}
+	if has {
```
ldeffenb, line 251: This doesn't solve all of the problems, but only Put's that happen AFTER the initial pin is actually closed.   Consider the following:

1) Two or more simultaneous operations start pinning the exact same content.
2) They all do the individual Puts, and the root collection doesn't yet exist, so they all actually store their collectionChunks.
3) One of the pinners actually executes the Close causing the pinCollectionItem to exist.  That one "wins" and actually completes the pin.
4) The other pinners will now quit Putting their collectionChunks, but their originally stored (raced) records still exist.   And worse, their chunkstore.Put() invocation have increased the chunks' RefCnts.
5) When the other pinners finally invoke Close, they'll correctly not replace the original pinCollectionItem, but all of those incorrect RefCnt increments are still there causing those chunks to never be able to be removed from the chunkstore.

pkg/storer/internal/pinning/pinning.go
```diff
@@ -264,11 +274,19 @@ func (c *collectionPutter) Close(st internal.Storage, writer storage.Writer, roo
...
+	}
+
+	if !has {
+		// Save the root pin reference.
+		c.collection.Addr = root
+		err := writer.Put(c.collection)
+		if err != nil {
+			return fmt.Errorf("pin store: failed updating collection: %w", err)
+		}
 	}
```
ldeffenb, line 290: This doesn't solve the problem of residual collectionChunks with their corresponding forever-remaining increments of the chunkstore RefCnt.   Consider:

1) Two or more simultaneous operations start pinning the exact same content.
2) They all do the individual Puts, and the root collection doesn't yet exist, so they all actually store their collectionChunks and invoke chunkstore.Put incrementing RefCnt on those chunks.
3) One of the pinners actually executes the Close causing the pinCollectionItem to exist.  That one "wins" and actually completes the pin.
4) When the other pinners finally invoke Close, they'll correctly not replace the original pinCollectionItem, but all of those incorrect RefCnt increments are still there causing those chunks to never be able to be removed from the chunkstore.

ldeffenb (commented):
I corrected my original review by moving that comment to Close() where it belongs, but added a new concern with the Has() check that was added to Put().

pkg/storer/internal/pinning/pinning.go
notanatol: it's been removed now

pkg/storer/internal/pinning/pinning.go
notanatol: Basically returning an error on line :276 will make `cleanupOnErrWriter` wrapper object to perform a cleanup (by calling the `Cleanup` method) whenever the error is detected
https://github.com/ethersphere/bee/blob/bf56d82e1aff5d5e9fde395044a410e9b8baf03f/pkg/api/api.go#L870-L874
notice the `onErr` pointing to the `.Cleanup` method
https://github.com/ethersphere/bee/blob/bf56d82e1aff5d5e9fde395044a410e9b8baf03f/pkg/api/bzz.go#L106-L110
which in turn will decrement the ref count(s) to the correct value. See the test below.

ldeffenb: Thank you for that description.   I should have followed the error path back out from Close() myself.  I'm setting up a testnet environment to verify this on a full OSM tileset load whenever it is merged.

ldeffenb (approved):
Been tracking this one as it developed and the current version looks like it should solve the pinning refCnt issue that it intends to solve.   Looking forward to testing it once it merges.

acha-bill (approved):

PR #4561, opened 2024-01-30
fix(localstore): leveldb-batch based transactions for both index and chunkstore for every localstore op

### LINKED ISSUE #4341: replace storer transactions with leveldb batches

istae:
### Checklist

### Description
An extensive cleanup of the db transaction.
Every localstore write operation now utilizes a transaction. Calls to the chunkstore are also part of the transaction.

also tackled: evict just enough chunks of a batch to fall below the reserve capacity

### Related Issue (Optional)
closes #4341 #4538


pkg/node/node.go
```diff
@@ -184,8 +184,8 @@ const (
 	minPaymentThreshold           = 2 * refreshRate           // minimal accepted payment threshold of full nodes
 	maxPaymentThreshold           = 24 * refreshRate          // maximal accepted payment threshold of full nodes
 	mainnetNetworkID              = uint64(1)                 //
-	ReserveCapacity               = 4_194_304                 // 2^22 chunks
-	reserveWakeUpDuration         = 30 * time.Minute          // time to wait before waking up reserveWorker
+	ReserveCapacity               = 16384                     // 2^14 chunks
```
istae, line 187: TODO: remove before merge to master

PR #4562, opened 2024-01-31
fix: upload store fixes

### LINKED ISSUE #4541: uploadstore fixes
* stampindex check can be removed in the `uploadPutter.Put` as the chunks passed to the putter are refreshly uploaded and stamped chunks.
* `uploadPutter.Close` should delete all associated uploadItems and chunk data, see `remove()` (confirm this is currently leaking).
* `Report` should not fetch `uploadItem` for the tagID but can use the tagID directly from the chunk (it must assured that the callers of `Report` has a chunk with a properly set tagID)
* see https://hackmd.io/D3lK13p6S8egANbvkrIYSw#Deferred-Uploads for more description

acha-bill:
### Checklist

### Description

1. Remove stamp index check from `uploadPutter.Put`
2. Delete all associated uploadItems after close

### Related Issue (Optional)
Closes #4541

### DIFF
--- a/pkg/storer/internal/upload/uploadstore.go
+++ b/pkg/storer/internal/upload/uploadstore.go
@@ -861,6 +861,20 @@ func IterateAll(st storage.Store, iterateFn func(addr swarm.Address, isSynced bo
 	)
 }
 
+func IterateAllTagItems(st storage.Store, cb func(ti *TagItem) (bool, error)) error {
+	return st.Iterate(
+		storage.Query{
+			Factory: func() storage.Item {
+				return new(TagItem)
+			},
+		},
+		func(result storage.Result) (bool, error) {
+			ti := result.Entry.(*TagItem)
+			return cb(ti)
+		},
+	)
+}
+
 // BatchIDForChunk returns the first known batchID for the given chunk address.
 func BatchIDForChunk(st storage.Store, addr swarm.Address) ([]byte, error) {
 	var batchID []byte
--- a/pkg/storer/internal/upload/uploadstore.go
+++ b/pkg/storer/internal/upload/uploadstore.go
@@ -17,7 +17,6 @@ import (
 	"github.com/ethersphere/bee/pkg/storage/storageutil"
 	"github.com/ethersphere/bee/pkg/storer/internal"
 	"github.com/ethersphere/bee/pkg/storer/internal/chunkstamp"
-	"github.com/ethersphere/bee/pkg/storer/internal/stampindex"
 	"github.com/ethersphere/bee/pkg/swarm"
 )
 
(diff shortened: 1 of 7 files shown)


istae:
we should also add a new migration step that removes all leftover items from previous sessions.
in the next release, we can add a line in the release notes that the new release will cleanup all upload items so that before deploying the new release, all uploads must first finish.

pkg/storer/migration/step_05.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/ethersphere/bee/pkg/storage"
+	"github.com/ethersphere/bee/pkg/storer/internal/upload"
+)
+
+// step_05 is a migration step that removes all upload items from the store.
+func step_05(st storage.BatchedStore) error {
+	logger := log.NewLogger("migration-step-05", log.WithSink(os.Stdout))
+	logger.Info("start removing upload items")
+
+	var itemsToDelete []storage.Item
```
istae, line 21: there is a chance you might blow up the memory.
i would use a channel here so that one worker feeds the items, the other picks them up and deletes from storage

istae (changes requested):

pkg/storer/internal/upload/uploadstore.go
```diff
@@ -848,15 +827,31 @@ func DeleteTag(st storage.Store, tagID uint64) error {
...
 	return st.Iterate(
 		storage.Query{
 			Factory: func() storage.Item { return new(uploadItem) },
 		},
 		func(r storage.Result) (bool, error) {
-			address := swarm.NewAddress([]byte(r.ID[:32]))
-			synced := r.Entry.(*uploadItem).Synced != 0
-			return iterateFn(address, synced)
+			ui := r.Entry.(*uploadItem)
+			ui.IdFunc = func() string {
```
istae, line 837: what is the purpose of this?

istae (changes requested):

pkg/storer/internal/upload/uploadstore.go
acha-bill: when uploadItems are marshalled, the address and batchID (which make up the ID) are not stored. 
It was assumed that the user would always know these properties when trying to get the item from the store. 
```
ui := &uploadItem {
   Address: addr,
   BatchID: id,
}
store.Get(ui)
```
So there was no need to save the address and batch to storage.

But this is a problem when iterating the store. 
Since the address and batchID were not stored, the resulting upload items don't have a unique ID.
```
iterate(func (r Result) {
  ui := r.Entry.(*uploadItem)
  println(ui.ID() == "/") // true (since address and batchID are empty, ID() == "/" for all items
  
  // Luckily, the ID is stored in the `Result`. I introduced IdFunc here as a workaround
  ui.IdFunc = func() { return r.ID }
}
```

istae: got it, thanks.

notanatol (approved):

istae (approved):

PR #4563, opened 2024-01-31
fix: remove epoch migration code

### LINKED ISSUE #4540: remove any old localstore rewrite epoch migration logic
* remove any epoch migration code
* remove any old localstore migration steps and code
* remove any old statestore migration steps and code
* any and other remaining old migration code

notanatol:
### Checklist

### Description
closes #4540


pkg/statestore/test/store.go
```diff
@@ -225,5 +225,5 @@ func testStoreIterator(t *testing.T, store storage.StateStorer, prefix string, s
 func testEmpty(t *testing.T, store storage.StateStorer) {
 	t.Helper()
 
-	testStoreIterator(t, store, "", 1) // 1 because of the schema entry.
+	testStoreIterator(t, store, "", 0) // 1 because of the schema entry.
```
notanatol, line 228: @istae is this comment still relevant?

do we still have 'schema entry' or was it part of the previous localstore?

istae: this schema entry probably has to do with the empty prefix and the way leveldb is implemented?
maybe @janos can help us here?

istae: which begs the question, how does changing it to `0` continue to work?

istae: looks like we manually insert a `schema` entry to keep a track of the migrations https://github.com/ethersphere/bee/blob/master/pkg/statestore/leveldb/leveldb.go#L176

istae (approved):

pkg/statestore/test/store.go
janos: Yes, that is correct, schema name was inserted under a specific key, but now since putSchemaName is not used, 0 is the correct value. Comment should be removed.

janos (approved):

acha-bill (approved):

Issue #4566, opened 2024-02-05
API for localstore scanning


nikipapadatou:
A new tool has been implemented which scans a localstore folder for corrupted and invalid chunks versus total amount of chunks per file belonging to pinned content.

The tool is here: https://github.com/ethersphere/bee/tree/feat-integrity-cmd

We need an API to serve this tool to the users, along with its relevant documentation:

-----
* go run $(pwd)/cmd/bee db validate-pin --data-dir /path/to/localstore
This will produce a csv file with all the pins and their stats.

To select the ones that are problematic run:
* awk '$1 >0 || $2>0' address.csv > invalid.csv
The file will contain the addresses of pins that the user might want to unpin/re-upload. The original files can be found by their hash.
-----

ldeffenb:
I simply expanded the /pins/{reference} to include detailed information on the chunks involved in the pin set.
```
curl http://192.168.10.36:11633/pins/0000001bf5eff96586a0ed6fe34fc8d859c69b0316a29912f14c4edcbbe732dc | jq
{
  "reference": "0000001bf5eff96586a0ed6fe34fc8d859c69b0316a29912f14c4edcbbe732dc",
  "chunkCount": 5,
  "chunks": [
    {
      "address": "0000001bf5eff96586a0ed6fe34fc8d859c69b0316a29912f14c4edcbbe732dc",
      "err": false,
      "refCnt": 100,
      "local": true,
      "cached": false,
      "reserve": 0,
      "upload": 0,
      "repaired": false
    },
    {
      "address": "4bd95ce93becc5e9fc1bf3f2304c7b2863bbbc9479fe248f319a809faf166a60",
      "err": false,
      "refCnt": 100,
      "local": true,
      "cached": false,
      "reserve": 0,
      "upload": 0,
      "repaired": false
    },
    {
      "address": "57b2c67a935478cbd5b72332a942751a9536931e51e73a00a90e009b9ea959c1",
      "err": false,
      "refCnt": 100,
      "local": true,
      "cached": false,
      "reserve": 0,
      "upload": 0,
      "repaired": false
    },
    {
      "address": "9b7f230858aabf6cf3e7b24effe56f821fcbe4989d3b12652865a83e5ad65d85",
      "err": false,
      "refCnt": 100,
      "local": true,
      "cached": false,
      "reserve": 0,
      "upload": 0,
      "repaired": false
    },
    {
[...]

PR #4573, opened 2024-02-07
feat: pinned reference integrity check API

notanatol:
### Checklist

### Description
This is a new debug API endpoint that performs the same validation as the pin integrity check command line tool: https://hackmd.io/@F-mGf2lBSoGylLGQE6rRkA/ByVp6Ac9p

To trigger chunk validation for all pins: send an http get request to `/check/pin`, if you want to limit to a specific reference add a query parameter:

```sh
curl -v -XGET "localhost:1635/check/pin?ref=0874f94e9009628a70247c7b77f9034c8729c38d153a9fa42058bf5c7b341495"
```

The data will be flushed periodically and is `.jq` friendly: `jq 'select(.invalid>0) | .ref' out.json`

The background job is responsive to interruptions if you want to stop the pin validation process.

Because the validation process is disk bound it's not desirable to trigger it on a busy instance (while downloading/uploading).

Closes: #4566

### DIFF
--- a/pkg/api/integritycheck.go
+++ b/pkg/api/integritycheck.go
@@ -31,7 +31,8 @@ func (s *Service) pinIntegrityHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	out := make(chan storer.PinStat)
-	go s.pinIntegrity.Check(logger, querie.Ref.String(), out)
+
+	go s.pinIntegrity.Check(r.Context(), logger, querie.Ref.String(), out)
 
 	flusher, ok := w.(http.Flusher)
 	if !ok {
--- a/pkg/api/integritycheck.go
+++ b/pkg/api/integritycheck.go
@@ -41,6 +41,7 @@ func (s *Service) pinIntegrityHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	w.Header().Set("Transfer-Encoding", "chunked")
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
 	w.WriteHeader(http.StatusOK)
 	flusher.Flush()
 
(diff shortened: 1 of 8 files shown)


ldeffenb:
> ### Description
> This is a new debug API endpoint that performs the same checks as the pin integrity check command line tool: https://hackmd.io/@ethswarm/rkmT1Ud9p

image

notanatol:
@ldeffenb updated the description: https://hackmd.io/@F-mGf2lBSoGylLGQE6rRkA/ByVp6Ac9p

ldeffenb:
> @ldeffenb updated the description: https://hackmd.io/@F-mGf2lBSoGylLGQE6rRkA/ByVp6Ac9p

Thank you.   That one is readable by the general public!

pkg/storer/storer.go
```diff
@@ -278,15 +278,15 @@ func initDiskRepository(
 	basePath string,
 	locker storage.ChunkLocker,
 	opts *Options,
-) (storage.Repository, io.Closer, error) {
+) (storage.Repository, *PinIntegrity, io.Closer, error) {
```
istae, line 281: instead of returning so many things from the constructor, the PinIntegrity can returned with a get function?

pkg/api/integritycheck.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+type PinIntegrityResponse struct {
+	Ref     swarm.Address `json:"ref"`
+	Total   int           `json:"total"`
+	Missing int           `json:"missing"`
+	Invalid int           `json:"invalid"`
+}
+
+func (s *Service) pinIntegrityHandler(w http.ResponseWriter, r *http.Request) {
```
istae, line 21: would it be possible to unit test this?

notanatol: added

pkg/storer/storer.go
notanatol: so I tried adding a getter to the `storage.Repository` but the compiler says that now there are `183` places that need changing (to comply with the new interface) so it seems it's no feasible. I'll look into other ways but for now I don't have a better approach.

istae (approved):

acha-bill (approved):

PR #4585, opened 2024-02-19
feat: add reserveSizeWithinRadius to status protocol

### LINKED ISSUE #4576: add reserveSizeWithinRadius to the status protocol
the status protocol returns the current reserveSize
we should also return the reserveSizeWithinRadius in the status protocol as well

acha-bill:
### Checklist

closes #4576


notanatol (approved):

istae (changes requested):
instead of trigger a recount each time status is called (keep in mind that peers can request your status), it would be better to return the previously recorded count that was computed by the reserve worker.

PR #4587, opened 2024-02-20
feat: cleanup and repair commands

notanatol:
### Checklist

### Description
Porting two of @ldeffenb functionalities for fixing incorrect reference count and removing invalid chunks from sharky.

### DIFF
--- a/pkg/storer/fix_refcnt.go
+++ b/pkg/storer/fix_refcnt.go
@@ -1,0 +1,283 @@
+// Copyright 2023 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package storer
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sort"
+	"time"
+
+	"github.com/ethersphere/bee/pkg/storer/internal/cache"
+	"github.com/ethersphere/bee/pkg/storer/internal/chunkstore"
+	pinstore "github.com/ethersphere/bee/pkg/storer/internal/pinning"
+	"github.com/ethersphere/bee/pkg/storer/internal/reserve"
+	"github.com/ethersphere/bee/pkg/storer/internal/upload"
+	"github.com/ethersphere/bee/pkg/swarm"
+)
+
+// FixRefCnt attempts to correct the RefCnt in all RetrievalIndexItems by scanning the actual chunk referers.
+func FixRefCnt(ctx context.Context, basePath string, opts *Options, repair bool) error {
+	logger := opts.Logger
+
+	store, err := initStore(basePath, opts)
+	if err != nil {
+		return fmt.Errorf("failed creating levelDB index store: %w", err)
+	}
+	defer func() {
+		if err := store.Close(); err != nil {
+			logger.Error(err, "failed closing store")
+		}
+	}()
+
+	statusInterval := 15 * time.Second
+	ticked := false
+	ticker := time.NewTicker(statusInterval)
+	go func() {
+		for range ticker.C {
+			ticked = true
+		}
+	}()
+
+	startChunkCount := time.Now()
+	total := 0
+	_ = chunkstore.Iterate(store, func(item *chunkstore.RetrievalIndexItem) error {
+		total++
+		if ticked {
+			ticked = false
+			logger.Info("..still counting chunks", "total", total, "elapsed", time.Since(startChunkCount).Round(time.Second))
+		}
+		return nil
+	})
+	ticker.Stop()
+	logger.Info("chunk count finished", "total", total, "elapsed", time.Since(startChunkCount).Round(time.Second))
+
+	pinTotal := -1
+	if false {
+		ticked = false
+		ticker.Reset(statusInterval)
+		startPinCount := time.Now()
+		_ = pinstore.IteratePinnedChunks(store, func(addr swarm.Address) (bool, error) {
+			pinTotal++
+			if ticked {
+				ticked = false
+				logger.Info("..still counting pins", "total", pinTotal, "elapsed", time.Since(startPinCount).Round(time.Second))
+			}
+			return false, nil
+		})
+		logger.Info("pin count finished", "total", pinTotal, "elapsed", time.Since(startPinCount).Round(time.Second))
+	}
+
+	type ItemInfo struct {
+		item       *chunkstore.RetrievalIndexItem
+		pinCnt     uint32
+		reserveCnt uint32
+		cacheCnt   uint32
+		uploadCnt  uint32
+		batches    [][]byte
+	}
+
+	start := time.Now()
+	repaired := 0
+	processed := 0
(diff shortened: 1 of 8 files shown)


pkg/storer/fix_refcnt.go
```diff
@@ -0,0 +1,283 @@
...
+		if err != nil {
+			logger.Error(err, "Failed to create batch")
+		}
+		batchHits := 0
+
+		logger.Info("Summarizing refCnts")
+		refs := make(map[uint32]uint32)
+		for _, item := range items {
+			refs[item.item.RefCnt]++
+			newRefCnt := item.pinCnt*100 + item.reserveCnt + item.cacheCnt + item.uploadCnt
```
notanatol, line 229: @ldeffenb  what is the reason for this multiplication? `item.pinCnt*100`

ldeffenb: That is one of my hacks that you don't need.  In my case, if a chunk is pinned, I want it to STAY in the sharky regardless of the bugs that might inappropriately decrement the refCnt by 1 here or there.   You'll just want pinCnt, and drop the *100.

pkg/storer/fix_refcnt.go
```diff
@@ -0,0 +1,283 @@
...
+			ticked = false
+			logger.Info("..still counting chunks", "total", total, "elapsed", time.Since(startChunkCount).Round(time.Second))
+		}
+		return nil
+	})
+	ticker.Stop()
+	logger.Info("chunk count finished", "total", total, "elapsed", time.Since(startChunkCount).Round(time.Second))
+
+	pinTotal := -1
+	if false {
```
nugaon, line 68: It can be removed, iterated below and overwritten by `pinScanCount`

pkg/storer/fix_refcnt.go
```diff
@@ -0,0 +1,283 @@
...
+		if ticked {
+			ticked = false
+			logger.Info("..still counting chunks", "total", total, "elapsed", time.Since(startChunkCount).Round(time.Second))
+		}
+		return nil
+	})
+	ticker.Stop()
+	logger.Info("chunk count finished", "total", total, "elapsed", time.Since(startChunkCount).Round(time.Second))
+
+	pinTotal := -1
```
nugaon, line 67: why not 0?

pkg/storer/validate.go
```diff
@@ -305,3 +305,112 @@ func validatePins(logger log.Logger, store storage.Store, pin, location string,
...
+
+	validChunk := func(item *chunkstore.RetrievalIndexItem, buf []byte) (bool, error) {
+		err := readFn(context.Background(), item.Location, buf)
+		if err != nil {
+			logger.Warning("invalid chunk", "address", item.Address, "timestamp", time.Unix(int64(item.Timestamp), 0), "location", item.Location, "error", err)
+			return false, err
+		}
+
+		ch := swarm.NewChunk(item.Address, buf)
+		if !cac.Valid(ch) {
```
nugaon, line 334: it should call the proposed `soc.Valid` (and continue with its block) and this line should be an `else if`

pkg/storer/validate.go
```diff
@@ -305,3 +305,112 @@ func validatePins(logger log.Logger, store storage.Store, pin, location string,
...
+			if soc.Valid(ch) {
```
nugaon, line 335: this should precede the cac.Valid because it does not only check the validity of the chunk but also counts the SOC chunks.

pkg/storer/validate.go
```diff
@@ -305,3 +305,112 @@ func validatePins(logger log.Logger, store storage.Store, pin, location string,
...
+				}
+
+				computedAddr := swarm.NewAddress(h)
+
+				if !cac.Valid(swarm.NewChunk(computedAddr, buf)) {
+					logger.Warning("computed chunk is also an invalid cac", "err", err)
+					return false, err
+				}
+
+				sharedEntry := chunkstore.RetrievalIndexItem{Address: computedAddr}
```
nugaon, line 356: I'm just guessing what is happening here but it assumes that the invalid chunk was content addressed and the address is not valid. It recalculates the address and tries to retrieve that from the same set from which it came from. but for that the computedAddress should be checked to be different than item.Address at least.
In any way, I don't see why this section is required. I would just return false and errInvalidChunk at the start of this block.

pkg/storer/validate.go
```diff
@@ -305,3 +305,112 @@ func validatePins(logger log.Logger, store storage.Store, pin, location string,
...
+	var wg sync.WaitGroup
+
+	for i := 0; i < 8; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			buf := make([]byte, swarm.SocMaxChunkSize)
+			for item := range iteratateItemsC {
+				valid, err := validChunk(item, buf[:item.Location.Length])
+				if !valid && err == nil {
```
nugaon, line 391: it only removes the chunk if there is a `sharedEntry` but the chunk is invalid anyways.

pkg/storer/fix_refcnt.go
```diff
@@ -0,0 +1,283 @@
...
+		// we deliberately choose to iterate the whole store again for each shard
+		// so that we do not store all the items in memory (for operators with huge localstores)
+		_ = chunkstore.Iterate(store, func(item *chunkstore.RetrievalIndexItem) error {
+			if ticked {
+				ticked = false
+				logger.Info("..still searching chunks", "skip", skip, "found", len(items), "total", total, "elapsed", time.Since(n).Round(time.Second))
+			}
+			if skip > 0 {
+				skip--
+				return nil
```
nugaon, line 106: would be better to continue with the cursor of the iteration.

pkg/storer/fix_refcnt.go
```diff
@@ -0,0 +1,283 @@
...
+		if err != nil {
+			logger.Error(err, "Failed to create batch")
+		}
+		batchHits := 0
+
+		logger.Info("Summarizing refCnts")
+		refs := make(map[uint32]uint32)
+		for _, item := range items {
+			refs[item.item.RefCnt]++
+			newRefCnt := item.pinCnt + item.reserveCnt + item.cacheCnt + item.uploadCnt
```
nugaon, line 229: someone please verify that reference count comes from these databases only.

pkg/storer/internal/reserve/reserve.go
```diff
@@ -510,3 +510,19 @@ func (r *Reserve) IncBinID(store storage.Store, bin uint8) (uint64, error) {
...
+	err := store.Iterate(storage.Query{
+		Factory: func() storage.Item { return &BatchRadiusItem{} },
+	}, func(res storage.Result) (bool, error) {
+		item := res.Entry.(*BatchRadiusItem)
+
+		stop, err := cb(item)
+		if stop || err != nil {
+			return true, err
+		}
+		return false, nil
```
nugaon, line 524: just return cb

PR #4604, opened 2024-03-04
fix(seg65)

zelig:


notanatol (approved):

istae (approved):

istae (changes requested):
newSpanEncryption and newDataEncryption should be exported functions so that all the pkgs can reuse the same implemention

istae (approved):

zelig (commented):
its ok

PR #4610, opened 2024-03-08
fix(getter): redundancy getter cleanup

istae:
### Checklist

### Description
1. removed close call
2. remove ctx in the struct antipattern
3. inflight requests are canceled when recovery is initialized.
4. the joiner does not construct a redundancy getter when fallback mode is false and the strategy is none.

### DIFF
--- a/pkg/file/redundancy/getter/getter.go
+++ b/pkg/file/redundancy/getter/getter.go
@@ -193,7 +193,6 @@ func (g *decoder) prefetch() {
 		if err == nil || g.config.Strict {
 			break
 		}
-
 	}
 
 	if err != nil {
--- a/pkg/file/redundancy/getter/getter.go
+++ b/pkg/file/redundancy/getter/getter.go
@@ -25,26 +25,25 @@ var (
 // if retrieves children of an intermediate chunk potentially using erasure decoding
 // it caches sibling chunks if erasure decoding started already
 type decoder struct {
-	fetcher           storage.Getter  // network retrieval interface to fetch chunks
-	putter            storage.Putter  // interface to local storage to save reconstructed chunks
-	addrs             []swarm.Address // all addresses of the intermediate chunk
-	inflight          []atomic.Bool   // locks to protect wait channels and RS buffer
-	cache             map[string]int  // map from chunk address shard position index
-	waits             []chan error    // wait channels for each chunk
-	rsbuf             [][]byte        // RS buffer of data + parity shards for erasure decoding
-	goodRecovery      chan struct{}   // signal channel for successful retrieval of shardCnt chunks
-	badRecovery       chan struct{}   // signals that either the recovery has failed or not allowed to run
-	lastLen           int             // length of the last data chunk in the RS buffer
-	shardCnt          int             // number of data shards
-	parityCnt         int             // number of parity shards
-	wg                sync.WaitGroup  // wait group to wait for all goroutines to finish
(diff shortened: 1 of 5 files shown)


notanatol (approved):

acha-bill (approved):

pkg/file/redundancy/getter/getter.go
```diff
@@ -138,21 +118,30 @@ func (g *decoder) fetch(ctx context.Context, i int, waitForRecovery bool) (err e
...
-			select {
-			case <-fctx.Done(): // local context
-			case <-g.ctx.Done(): // global context
-			}
-			cancel()
-			g.wg.Done()
-		}()
+		// when the recovery is triggered, we can terminate any inflight requests.
+		// we do the extra bool check to not fire an unnecessary goroutine
+		if waitForRecovery {
```
zelig, line 136: there is no need to run this routine for every fetcgh.
creating context and cancel hould be done onky once

pkg/file/redundancy/getter/getter.go
```diff
@@ -256,34 +243,32 @@ func (g *decoder) runStrategy(ctx context.Context, s Strategy) error {
...
-			return ctx.Err()
-		case <-c:
-			if g.fetchedCnt.Load() >= int32(g.shardCnt) {
-				return nil
-			}
-			if g.failedCnt.Load() > int32(allowedErrs) {
-				return errStrategyFailed
-			}
+	for range c {
+		if g.fetchedCnt.Load() >= int32(g.shardCnt) {
```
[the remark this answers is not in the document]
istae: every time c returns, it's guaranteed that either fetchedCnt or failedCnt is already incremented (inside fetch) so eventually one of the conditions are met. 

if m is less than shard + parity, then a previous run of fetches has already finished and incremented the counters.

pkg/file/redundancy/getter/getter.go
istae: context is canceled is a separate goroutine when the recovery is initialized...this is how we achieve early termination.
it's safe to call cancel twice in separate goroutines.

pkg/file/redundancy/getter/getter.go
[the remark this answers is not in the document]
zelig: think about this once more. Because of the single flight pattern and Get calls interleaving prefetch fetches, there is a decent probability that the loop will not terminate. All you need to imagine is that the shardCnt-th successful fetch is.
I think we should trigger closing the initRecovery channel when we  increment the count (in one atomic call)

pkg/file/redundancy/getter/getter.go
zelig: it would make sense to cancel on

NoahMaizels:
Last week:
* Published guide to node state and health endoints and reorganized troubleshooting section https://docs.ethswarm.org/docs/bee/working-with-bee/troubleshooting
* Published new content in erasure coding section including cost calculator: https://docs.ethswarm.org/docs/develop/access-the-swarm/erasure-coding/
* Submitted backlog of issues to bee repo related to api problems / improvements
* Various 2.0 related comms (tweets, discord announcements)
* Working with Tamas on subtitles and editing for health and status api endpoints video guide

This week:
* Present new docs content and highlight video guides at SCC
* Adapt and add doubling and neighborhoods guide to docs
* Add devops guide to docs
* 2.0 comms

PR #4612, opened 2024-03-13
fix: use neighborhood suggester only on mainnet

acha-bill:
## Description

Use neighborhood suggester option only on mainnet


istae (approved):
thanks!

PR #4615, opened 2024-03-18
feat: v2

acha-bill:
## Description

Major version update to v2


janos (approved):
Thank you for this change. SwarmScan can import bee v2 correctly, now.

istae (approved):

PR #4627, opened 2024-03-27
fix: small api fixes

### LINKED ISSUE #4617: Multiple Small API Fixes
#### Summary

After reviewing the health and status related endpoints in order to write a guide for node operators I came across a variety of bugs / difficult to understand responses from multiple endpoints. Most are very small fixes, for those which are more in depth I will submit separate issues. In this issue, I will only include those which have very simple fixes.

#### Details

1.  /status endpoint returns the value of "0" for the proximity of the node with itself, the proximity should actually be 256 
2.  For the object returned from the /status endpoint, if the object will be used both for /status and /status/peers it would be better to not use "peer" for the key name, and rather use "overlay" in both, since it is confusing when it used for the node which is making the request itself, since a node cannot be a peer of itself. 
3.  For /redistributionstate, change lastSampleDuration value from nanoseconds to seconds
4.  For /redistributionstate, change lastSampleDuration key name to lastSampleDurationSeconds
5. for /health, in the results the version number for the Debug API is zero. This should probably be updated to reflect whatever the current version number is.

> {
>   "status": "ok",
>   "version": "1.18.1-ed24b890",
>   "apiVersion": "5.1.0",
>   "debugApiVersion": "0.0.0"
> }
> 

6. For the return object from /status and /status/peers, there is an entry with the key name "neighborhoodSize". The typical meaning of "neighborhood size" would include both the node itself and all its neighbors, however the returned value is not inclusive of the node itself.  So either the value should include the node itself, or the name of the entry should be changed to "neighborsCount". I think keeping the entry name the same and changing the value to include the node itself is probably the less ambiguous method and so is better.

acha-bill:
## Description

Small api fixes

closes #4617

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -19,7 +19,7 @@ builds:
       - -X github.com/ethersphere/bee/v2.commitHash={{ .ShortCommit }}
       - -X github.com/ethersphere/bee/v2.commitTime={{ .CommitTimestamp }}
       - -X github.com/ethersphere/bee/v2/pkg/api.Version={{ .Env.BEE_API_VERSION }}
-      - -X github.com/ethersphere/bee/v2/pkg/debugapi.Version={{ .Env.BEE_DEBUG_API_VERSION }}
+      - -X github.com/ethersphere/bee/v2/pkg/api.DebugVersion={{ .Env.BEE_DEBUG_API_VERSION }}
     env:
       - CGO_ENABLED=0
     goos:
@@ -43,7 +43,7 @@ builds:
       - -X github.com/ethersphere/bee/v2.commitHash={{ .ShortCommit }}
       - -X github.com/ethersphere/bee/v2.commitTime={{ .CommitTimestamp }}
       - -X github.com/ethersphere/bee/v2/pkg/api.Version={{ .Env.BEE_API_VERSION }}
-      - -X github.com/ethersphere/bee/v2/pkg/debugapi.Version={{ .Env.BEE_DEBUG_API_VERSION }}
+      - -X github.com/ethersphere/bee/v2/pkg/api.DebugVersion={{ .Env.BEE_DEBUG_API_VERSION }}
     env:
       - CGO_ENABLED=0
     goos:
@@ -69,7 +69,7 @@ builds:
       - -X github.com/ethersphere/bee/v2.commitHash={{ .ShortCommit }}
       - -X github.com/ethersphere/bee/v2.commitTime={{ .CommitTimestamp }}
       - -X github.com/ethersphere/bee/v2/pkg/api.Version={{ .Env.BEE_API_VERSION }}
-      - -X github.com/ethersphere/bee/v2/pkg/debugapi.Version={{ .Env.BEE_DEBUG_API_VERSION }}
+      - -X github.com/ethersphere/bee/v2/pkg/api.DebugVersion={{ .Env.BEE_DEBUG_API_VERSION }}
     env:
       - CGO_ENABLED=0
     goos:
@@ -89,7 +89,7 @@ builds:
       - -X github.com/ethersphere/bee/v2.commitHash={{ .ShortCommit }}
       - -X github.com/ethersphere/bee/v2.commitTime={{ .CommitTimestamp }}
       - -X github.com/ethersphere/bee/v2/pkg/api.Version={{ .Env.BEE_API_VERSION }}
-      - -X github.com/ethersphere/bee/v2/pkg/debugapi.Version={{ .Env.BEE_DEBUG_API_VERSION }}
+      - -X github.com/ethersphere/bee/v2/pkg/api.DebugVersion={{ .Env.BEE_DEBUG_API_VERSION }}
     env:
       - CGO_ENABLED=0
     goos:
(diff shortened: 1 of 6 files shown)


.goreleaser.yml
```diff
@@ -19,7 +19,7 @@ builds:
       - -X github.com/ethersphere/bee/v2.commitHash={{ .ShortCommit }}
       - -X github.com/ethersphere/bee/v2.commitTime={{ .CommitTimestamp }}
       - -X github.com/ethersphere/bee/v2/pkg/api.Version={{ .Env.BEE_API_VERSION }}
-      - -X github.com/ethersphere/bee/v2/pkg/debugapi.Version={{ .Env.BEE_DEBUG_API_VERSION }}
```
NoahMaizels, line 22: This is to fix point 5 from https://github.com/ethersphere/bee/issues/4617?

pkg/api/redistribution.go
```diff
@@ -56,20 +55,20 @@ func (s *Service) redistributionStatusHandler(w http.ResponseWriter, r *http.Req
...
+		MinimumGasFunds:           bigint.Wrap(minGasFunds),
+		HasSufficientFunds:        hasSufficientFunds,
+		IsFrozen:                  status.IsFrozen,
+		IsFullySynced:             status.IsFullySynced,
+		Phase:                     status.Phase.String(),
+		LastWonRound:              status.LastWonRound,
+		LastPlayedRound:           status.LastPlayedRound,
+		LastFrozenRound:           status.LastFrozenRound,
+		LastSelectedRound:         status.LastSelectedRound,
+		LastSampleDurationSeconds: status.SampleDuration.Seconds(),
```
NoahMaizels, line 67: Point 4 from https://github.com/ethersphere/bee/issues/4617?

pkg/api/status.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 type statusSnapshotResponse struct {
-	Peer                    string  `json:"peer"`
+	Overlay                 string  `json:"overlay"`
```
NoahMaizels, line 20: Should we bring this one up for discussion? I haven't had a chance to talk it over with anyone else.

pkg/api/status_test.go
```diff
@@ -32,7 +32,7 @@ func TestGetStatus(t *testing.T) {
 			PullsyncRate:            64,
 			StorageRadius:           8,
 			ConnectedPeers:          0,
-			NeighborhoodSize:        0,
+			NeighborhoodSize:        1,
```
NoahMaizels, line 35: What's this change for?

pkg/status/status.go
```diff
@@ -124,7 +124,7 @@ func (s *Service) LocalSnapshot() (*Snapshot, error) {
 		PullsyncRate:            syncRate,
 		StorageRadius:           uint32(storageRadius),
 		ConnectedPeers:          connectedPeers,
-		NeighborhoodSize:        neighborhoodSize,
+		NeighborhoodSize:        neighborhoodSize + 1, // include self
```
NoahMaizels, line 127: This is for point 6 from https://github.com/ethersphere/bee/issues/461?

pkg/status/status_test.go
```diff
@@ -96,12 +97,13 @@ func TestStatusLightNode(t *testing.T) {
...
-		StorageRadius:   0,
-		BatchCommitment: 1024,
-		IsReachable:     true,
+		BeeMode:          api.LightMode.String(),
+		ReserveSize:      0,
+		PullsyncRate:     0,
+		StorageRadius:    0,
+		BatchCommitment:  1024,
+		IsReachable:      true,
+		NeighborhoodSize: 1,
```
NoahMaizels, line 106: What does adding `NeighborhoodSize: 1,` do here?

NoahMaizels (commented):
In reference to https://github.com/ethersphere/bee/issues/4617:

- Was point 1 addressed? I don't think I see some change here which would address it
- Seems that all the other points (2 through 6) have been addressed
- Has there been any discussion already about points 2 and 6? I don't know if changing those might have some sort of unintended side effects?
  - Point 2: Has there been any discussion about renaming `peer`? We had some uncertainty on that one.
  - Point 6: Has there been any discussion about changing the neighborhoodSize to be inclusive of all nodes in the neighborhood?

.goreleaser.yml
acha-bill: yes

pkg/api/redistribution.go
acha-bill: yes

pkg/status/status.go
acha-bill: yes

NoahMaizels (approved):

openapi/SwarmCommon.yaml
```diff
@@ -1,6 +1,6 @@
 openapi: 3.0.3
 info:
-  version: 3.2.7
+  version: 3.2.8
```
janos, line 3: The API has breaking changes (renaming two fields) and the version is only updating the bugfix number. To be semantically correct, API number should be raised to 4, or to keep old fields for backward compatibility. It depends on the expectations for the API stability, but since the major version is not 0, the API supposed to be stable.

janos (commented):
The changes are ok, it is just a question are backward compatibility changes are needed or the API can be incompatible at these two fields.

acha-bill:
Let's leave this breaking change and update the API version then.
BTW, we are going to be adding more API breaking changes in the next release with the merge of debug and bee APIs. So this can be seen as a first step. 

wdyt @NoahMaizels ?

NoahMaizels:
> Let's leave this breaking change and update the API version then. BTW, we are going to be adding more API breaking changes in the next release with the merge of debug and bee APIs. So this can be seen as a first step.
> 
> wdyt @NoahMaizels ?

Actually after considering it a bit, I think maybe it would be best to include these changes together with which release merges the debug api, since that's also a breaking change and I know we have plans to communicate about that, so I guess we may as well bundle all these breaking changes into a single release and communicate them all together in advance.

NoahMaizels:
> Let's leave this breaking change and update the API version then. BTW, we are going to be adding more API breaking changes in the next release with the merge of debug and bee APIs. So this can be seen as a first step.
> 
> wdyt @NoahMaizels ?

Yeah I agree with you, I think it can be included in the next release since like you said it will be including breaking changes as well and we already have plans to communicate about those in advance

NoahMaizels (approved):

PR #4655, opened 2024-04-23
chore: bump libp2p to v0.33.2

acha-bill:


istae (approved):

janos (approved):
The upgrade is ok from the libp2p package api, but it should be tested against nodes that run libp2p v0.30.0 for possible regressions. I suggest to have some canary nodes updates that currently have memory issues visible on the metrics dashboards.

The upgrade for the whole libp2p stack is quite extensive in order to fix a timer leak with a one line. If problems are detected, I suggest to use a patched version of libp2p v0.30.0 and to deal with the complexities of extensive testing in time, without rush, as this memory leak is an urgent issue.

PR #4677, opened 2024-05-15
feat: feed wrapping

nugaon:
This feature drops the JSON structure with CAC reference and timestamp in feed payload in order to be capable of wrapping a whole content addressed chunk instead. The feed payload resolution is backward compatible by checking the `span + ts + swarmHash` size in content and trying to retrieve the content under swarmHash. 

Because the SOC content needs to be resolved in one request and there is no `ts` attribute of a feed anymore, the Feed API must be updated and it has breaking change on its GET endpoint. Additionally, `at` parameter is no longer in use at sequential feeds.

Since now feed endpoint can return content bigger than one chunk and that can be erasure coded, erasure-code related request headers have been added.
Some application may use this endpoint to simply get the latest index of a sequential feed, for that new header is introduced `swarm-only-root-chunk` which is a boolean type and if it is true then Bee does not retrieve the whole chunk tree.

`GET /soc/{owner}/{id}` method is introduced with required path parameters `owner` and `id` which are hex encoded strings.
This new method handles root chunk as a Single Owner Chunk by default and resolves and returns content similarly as `GET /feed`. Both endpoints return newly introduced header `swarm-soc-signature` which encodes the signature of the Single Owner Chunk in question.

- [x] init joiner by root chunk
- [x] update bootstrap
- [x] update bzz and feeds api

### Checklist


pkg/file/joiner/joiner.go
```diff
@@ -116,6 +116,12 @@ func New(ctx context.Context, g storage.Getter, putter storage.Putter, address s
 		return nil, 0, err
 	}
 
+	return NewJoiner(ctx, g, putter, address, rootChunk)
+}
+
+// NewWithRoot creates a new Joiner with the already fetched root chunk.
```
zelig, line 122: wrong name of func

pkg/file/loadsave/loadsave.go
```diff
@@ -48,14 +49,33 @@ func NewReadonly(getter storage.Getter) file.LoadSaver {
 	}
 }
 
+// NewReadonly returns a new read-only load-saver
```
zelig, line 52: wrong name of func

zelig (commented):
This must go through a SWIP
is direct resolutiion into content really needed? wouldnt it be much simpler if this was done in 2 steps on cllient side?

extensive documentation, examples should accompany release of this feature
thorough test of backward compatibility

nugaon:
> This must go through a SWIP

Wrote SWIP and created [PR](https://github.com/ethersphere/SWIPs/pull/57) for that

> is direct resolutiion into content really needed? wouldnt it be much simpler if this was done in 2 steps on cllient side?

This is desired to eliminate; the content can be resolved by only 1 request. 
Moreover, with bigger content there is no API functionality to pass a whole chunk to resolve (and that way would be more complex)

pkg/api/feed.go
```diff
@@ -64,6 +64,14 @@ func (s *Service) feedGetHandler(w http.ResponseWriter, r *http.Request) {
 		queries.At = time.Now().Unix()
 	}
 
+	headers := struct {
```
martinconic, line 67: Should we keep the queries above with `At` and `After` ? Also bellow there is a lookup using this queries.

zelig (approved):

Cafe137:
Please add the new headers to `corsHandler` `allowedHeaders`

pkg/api/feed.go
nugaon: it may be needed for epoch based feeds.
I created an issue about that #4830

martinconic (approved):

janos (approved):

PR #4685, opened 2024-05-20
feat: mine new overlay

nugaon:
Allowing the overlay address mining even when nonce exists. That case, `localstore` and `kademlia-metrics` folders will be removed and `statestore` will be cleared to start a clean state in the new neighborhood. Before the wipe, 10 seconds wait time allows the user to change their mind.

### Checklist

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -321,21 +321,12 @@ func NewBee(
 			if err != nil {
 				return nil, fmt.Errorf("statestore: save new overlay: %w", err)
 			}
-		} else if !nonceExists {
-			err = setOverlay(stateStore, swarmAddress, nonce)
-			if err != nil {
-				return nil, fmt.Errorf("statestore: save new overlay: %w", err)
-			}
-		}
-	} else if targetNeighborhood == "" && !nonceExists {
-		err = setOverlay(stateStore, swarmAddress, nonce)
-		if err != nil {
-			return nil, fmt.Errorf("statestore: save new overlay: %w", err)
 		}
 	}
 
 	logger.Info("using overlay address", "address", swarmAddress)
 
+	// this will set overlay if it was not set before
 	if err = checkOverlay(stateStore, swarmAddress); err != nil {
 		return nil, fmt.Errorf("check overlay address: %w", err)
 	}
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -284,8 +284,9 @@ func NewBee(
 			const (
 				localstore = "localstore"
 				kademlia   = "kademlia-metrics"
+				statestore = "statestore"
 			)
-			dirsToNuke := []string{localstore, kademlia}
+			dirsToNuke := []string{localstore, kademlia, statestore}
 			for _, dir := range dirsToNuke {
 				err = removeContent(filepath.Join(o.DataDir, dir))
 				if err != nil {
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -283,7 +283,6 @@ func NewBee(
 		}
 	}
 
-	// set overlay if nonce is not saved or mine
 	if targetNeighborhood != "" {
 		neighborhood, err := swarm.ParseBitStrAddress(targetNeighborhood)
 		if err != nil {
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -17,6 +17,7 @@ import (
 	"math/big"
 	"net"
 	"net/http"
+	"os"
 	"path/filepath"
 	"runtime"
 	"sync"
@@ -274,11 +275,24 @@ func NewBee(
 		return nil, fmt.Errorf("compute overlay address: %w", err)
 	}
 
-	if nonceExists && o.TargetNeighborhood != "" {
-		logger.Warning("an overlay has already been created before, skipping targeting the selected neighborhood")
-	}
+	if o.TargetNeighborhood != "" {
+		if nonceExists {
+			logger.Info("Override nonce %d and clean state for neighborhood %s", nonce, o.TargetNeighborhood)
+			logger.Warning("you have another 10 seconds to change your mind and kill this process with CTRL-C...")
+			time.Sleep(10 * time.Second)
 
-	if !nonceExists {
+			const (
+				localstore = "localstore"
+				kademlia   = "kademlia-metrics"
+			)
+			dirsToNuke := []string{localstore, kademlia}
+			for _, dir := range dirsToNuke {
+				err = removeContent(filepath.Join(o.DataDir, dir))
+				if err != nil {
+					return nil, fmt.Errorf("delete %s: %w", dir, err)
+				}
+			}
+		}
 		// mine the overlay
 		targetNeighborhood := o.TargetNeighborhood
 		if o.TargetNeighborhood == "" && o.NeighborhoodSuggester != "" {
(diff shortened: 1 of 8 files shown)


pkg/node/node.go
```diff
@@ -1359,3 +1374,28 @@ func isChainEnabled(o *Options, swapEndpoint string, logger log.Logger) bool {
 	logger.Info("starting with an enabled chain backend")
 	return true // all other modes operate require chain enabled
 }
+
+// removeContent removes all files in path. Copied function from cmd/db.go
+func removeContent(path string) error {
```
istae, line 1379: best to move this func to pkg/util/ioutil and import it from there

pkg/node/node.go
```diff
@@ -278,11 +279,25 @@ func NewBee(
 		return nil, fmt.Errorf("compute overlay address: %w", err)
 	}
 
-	if nonceExists && o.TargetNeighborhood != "" {
-		logger.Warning("an overlay has already been created before, skipping targeting the selected neighborhood")
-	}
+	if o.TargetNeighborhood != "" {
+		if nonceExists {
```
istae, line 283: if the target neighborhood is specified, on each restart, wouldn't the directories would be nuked?

pkg/node/node.go
```diff
@@ -278,11 +279,25 @@ func NewBee(
...
+			for _, dir := range dirsToNuke {
+				err = removeContent(filepath.Join(o.DataDir, dir))
+				if err != nil {
+					return nil, fmt.Errorf("delete %s: %w", dir, err)
+				}
+			}
+		}
 		// mine the overlay
 		targetNeighborhood := o.TargetNeighborhood
 		if o.TargetNeighborhood == "" && o.NeighborhoodSuggester != "" {
```
istae, line 303: for the suggester to run. the target neighborhood has to be "",  but this check is inside the `if o.TargetNeighborhood != "" {` check on line 283, no?

istae (changes requested):
in order to achieve overlay change during runtime, i think you have to save the previous target-neighborhood in the statestore and only nuke the dirs (and change the overlay) if the target-neighborhood of the config option is different from the statestore one.

ldeffenb:
> save the previous target-neighborhood in the statestore

Couldn't you just check the saved overlay (if there is one) to see if it is already in the specified target neighborhood?

pkg/node/node.go
nugaon: correct, good catch! : ) move outside of the block

nugaon:
Thanks for the comments regarding the missed use-case when user reloads the client with the mining param.

I corrected that and it also sets `TargetNeighborhood` with `NeighborhoodSuggester` only if nonce was not defined before.

istae (commented):
some comments around the code changes would help
also let's post screenshots of the different test cases

pkg/node/node.go
```diff
@@ -278,37 +278,58 @@ func NewBee(
...
+				time.Sleep(10 * time.Second)
+
+				const (
+					localstore = "localstore"
+					kademlia   = "kademlia-metrics"
+					statestore = "statestore"
+				)
+				dirsToNuke := []string{localstore, kademlia, statestore}
+				for _, dir := range dirsToNuke {
+					err = ioutil.RemoveContent(filepath.Join(o.DataDir, dir))
```
janos, line 309: It would be better to scope this `err` value in the loop instead to the outer variable which is currently two ifs above.

pkg/node/node.go
```diff
@@ -278,37 +278,58 @@ func NewBee(
...
-			}
+	if targetNeighborhood != "" {
+		neighborhood, err := swarm.ParseBitStrAddress(targetNeighborhood)
+		if err != nil {
+			return nil, fmt.Errorf("invalid neighborhood. %s", targetNeighborhood)
 		}
 
-		if targetNeighborhood != "" {
+		if swarm.Proximity(swarmAddress.Bytes(), neighborhood.Bytes()) < uint8(len(targetNeighborhood)) {
+			if nonceExists {
```
janos, line 297: Removing directory before the `nbhdutil.MineOverlay`, which can fail, may result with only removed data.

pkg/node/node.go
```diff
@@ -278,37 +278,58 @@ func NewBee(
...
+				logger.Info("Override nonce %d and clean state for neighborhood %s", nonce, targetNeighborhood)
+				logger.Warning("you have another 10 seconds to change your mind and kill this process with CTRL-C...")
+				time.Sleep(10 * time.Second)
+
+				const (
+					localstore = "localstore"
+					kademlia   = "kademlia-metrics"
+					statestore = "statestore"
```
janos, line 305: TLDR The cleanest way is instead to remove directories, it would be better for these services to implement its own internal nuke methods (that handle file descriptors and internal storage reinitalizations) and call them here, instead to remove their files.

Statestore directory is removed, but a few lines bellow statestore is passed to the setOverlay function. I would assume that the statestore is no longer reliable after its files are removed. Special attention should be made to ensure when interfering with storage internals outside of the services that manage their state. In this case, statestore must be reinitialised (and passed again to all already initialized services that use statestore) or if possible initilizad after the data on disk is removed. Handling nuke on node startup is different than in a separate nuke cli command as the node continues to run.

Nuking buy removing data directories is a questionable operation as these paths are constructed locally by assumption that they will not change in other places of the code where services that store data in them are initialized. I know that this is a mistake taken from the nuke cli command, but it would be good to make new code better.

nugaon: good point, I reinited statestore after the wipe as it is needed to fetch `nonce` before wipe.

Regarding the nuking, no service uses these databases before this point, all affected DB initializations happen in this function. It should have a separate `clearDb` method with options for sure if some of the data was required to be kept (please let me know if there is any). 
so basically, this is the only place where databases should managed as files, shouldn't it?

janos: Thank you for fixing it.

The statestore as a general key/value store, is a tricky to clean. I believe that this is a good time to plan something like a consolidation of all keys (or key prefixes) that are stored. It would be good to remove only the required keys.

I would say that it would even better to have a Nuke function on the databases to cleanup internal storage, in this case leveldb and not to worry about the file management. This would require a few more changes, but I believe that would result in more reliable code, not to worry about the files, just databases as abstraction. This can be an improvement later.

istae: the statestore actually has a nuke function but it preserves keys related to maintaing the overlay addresses, it may be improved to remove those entries when called upon.

janos: It is much better to adapt StateStoreAdapter.Nuke for cleaning. I think that it would be the safer way, and maybe to have the list of preserved prefixes a bit more transparent.

pkg/node/node.go
```diff
@@ -278,37 +268,71 @@ func NewBee(
...
+				)
+				dirsToNuke := []string{localstore, kademlia}
+				for _, dir := range dirsToNuke {
+					err := ioutil.RemoveContent(filepath.Join(o.DataDir, dir))
+					if err != nil {
+						return nil, fmt.Errorf("delete %s: %w", dir, err)
+					}
+				}
+
+				stateStore.ClearForHopping()
```
istae, line 311: error checking?

pkg/node/node.go
```diff
@@ -278,37 +278,58 @@ func NewBee(
...
 
-		if targetNeighborhood != "" {
+		if swarm.Proximity(swarmAddress.Bytes(), neighborhood.Bytes()) < uint8(len(targetNeighborhood)) {
+			if nonceExists {
+				logger.Info("Override nonce %d and clean state for neighborhood %s", nonce, targetNeighborhood)
+				logger.Warning("you have another 10 seconds to change your mind and kill this process with CTRL-C...")
+				time.Sleep(10 * time.Second)
+
+				const (
+					localstore = "localstore"
```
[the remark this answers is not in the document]
nugaon: yes, reasonable to make constants for that. I moved localstorage and kademlia path constants to ioutil within utils package.

pkg/util/ioutil/ioutil.go
```diff
@@ -4,6 +4,18 @@
...
 
+import (
+	"errors"
+	"os"
+	"path/filepath"
+)
+
+// DB folders paths from bee datadir
+const (
+	DataPathLocalstore = "localstore"
```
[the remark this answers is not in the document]
nugaon: maybe `swarm.go`?

janos (approved):
LGTM. Thank you for addressing the comments @nugaon.

istae (approved):

martinconic (approved):

acha-bill (approved):

PR #4710, opened 2024-06-13
feat: add last synced block to status endpoint result

martinconic:
### Checklist

### Description
A new field, called lastSyncedBlock should be added to the status report that shows the last synced block of the postage worker.
It is the same value as the block field under the /chainstate endpoint.

This will make it easier to detect if the node or it's peer are lagging behind in syncing batches from the contract.


istae (approved):
Let's bump the minor version of the protocol.

acha-bill (approved):

Issue #4704, opened 2024-06-03
Pullsync ignores stamp timestamp changes


ldeffenb:
### Context
Bee 2.1.0 (and earlier)

### Summary
Related to #4703 
Pullsync only pulls chunks that are not already in the reserve based solely on address and batch.  If content was re-uploaded, the chunk/batch/stamp timestamp is updated and the newly stamped chunk is pushed.   Even after #4703 is fixed, pullsync will still not propagate the current timestamp for the chunk/stamp.

### Expected behavior
Stamp timestamps should match across all nodes and reserves.

### Actual behavior
Given that pullsync only requests chunks that are not already in the reserve based on address and batch, updated timestamps will not be pulled and therefore will not be updated throughout the storing neighborhood.   Only those nodes that were directly pushed (presuming #4703 is fixed) will have the current stamp timestamp.

https://github.com/ethersphere/bee/blob/8c61408db7cd63227cfe161967e3b70f90b9fde3/pkg/pullsync/pullsync.go#L185-L192

### Steps to reproduce
I'll leave this to your imagination.

### Possible solution
The pullsync protocol needs to be updated to communicate not only the batch, but the entire stamp so that the pulling node can determine if the local stamp timestamp is current and/or update to the latest timestamp.

ldeffenb:
It's worse than just the timestamp.   Pullsync will also not pull a re-stamped chunk that happens to have been re-stamped into a different INDEX of the batch's bucket.   This is a critical flaw in mutable stamps.

And worse, since many chunks exist in the swarm whose stamps were forgotten in the great batch purge, those chunks CANNOT be reliably re-stamped with the original stamp because they may end up at different indices which will not be properly propagated to nodes still holding the original stamped chunks.

zelig:
@ldeffenb you are spot on, i indicated this issue in this writeup https://hackmd.io/@zelig/SJVY-ACKa
it also explain how it can lead to reserve mismatches to this day since the 'great batch purge'.
instead of the batchID indeed the pullsync offers should use the hash of the serialised stamp. 
Nodes should store this hash inlocalstore alongside the stamp and can be used to detect  restampings on differring index.

istae:
https://github.com/ethersphere/bee/pull/4717

Issue #4703, opened 2024-06-03
Reserve stampindex timestamps not updated


ldeffenb:
### Context
2.1.0 (and possibly earlier)

### Summary
I am re-uploading existing data on an existing stamp to the swarm.   The pushing stamper updates the existing stamp's timestamp in this case.  But once the chunk is pushed, the storing node determines that the chunk/batch already exists in the reserve and does NOT update the timestamp in the stampindex.  This causes old timestamps to continue to exist in the reserve.   Also these obsolete (read: earlier) timestamps are subsequently pullsynced to other nodes.

Note: The stamp timestamp is critical for batch bucket/index replacements especially for mutable batches.   But even for re-uploaded (and newly timestamped) chunks, it is important that all reserves have the SAME timestamp for a stamped chunk.

https://github.com/ethersphere/bee/blob/8c61408db7cd63227cfe161967e3b70f90b9fde3/pkg/postage/stamper.go#L49-L51

https://github.com/ethersphere/bee/blob/8c61408db7cd63227cfe161967e3b70f90b9fde3/pkg/storer/internal/reserve/reserve.go#L100-L106

### Expected behavior
If the original stamp timestamp is updated, I would expect all reserve timestamps to be updated when the stamped chunk is pushed.

### Actual behavior
Reserve timestamps are not updated if the stamped chunk is already in the reserve.   See line numbers quoted in the possible solution.

### Steps to reproduce
Since stamp information is well hidden in the swarm implementation, you really can't see this happening without adding additional stamp and reserve timestamp logging.  You can find my implementation of this logging in the following files of my hacks branch.  Of course, other files were also modified to enable the logs.
pkg/postage/stamper.go
pkg/storer/internal/reserve/reserve.go
https://github.com/ldeffenb/bee/tree/2.1.0-hacks

### Possible solution
It's going to kill performance, but even if the reserve already Has the chunk/batch, the stampindex timestamp should still be verified and/or updated as necessary.
https://github.com/ethersphere/bee/blob/8c61408db7cd63227cfe161967e3b70f90b9fde3/pkg/storer/internal/reserve/reserve.go#L100
https://github.com/ethersphere/bee/blob/8c61408db7cd63227cfe161967e3b70f90b9fde3/pkg/storer/internal/reserve/reserve.go#L116
But please do NOT update the store if the new timestamp matches the original!   And also don't log it as an error either!
[...]

PR #4717, opened 2024-07-02
feat: store stamp hash

acha-bill:
### Checklist

### Description
Fix pullsyc ignoring chunks having batch with changed timestamp or index because reserve only checks for `addr/batchId`

1. Migrate `BatchRadiusItem` and `ChunkBinItem` to include the stamps hash.
2. Pullsync offer now includes the stamp hash. Then the reserve can check for any stamp changes using  the `hash` alongside the `batchID`.

### Related Issues
https://github.com/ethersphere/bee/issues/4703
https://github.com/ethersphere/bee/issues/4704


pkg/pullsync/pullsync.go
```diff
@@ -170,6 +170,7 @@ func (s *Syncer) Sync(ctx context.Context, peer swarm.Address, bin uint8, start
 
 		addr := offer.Chunks[i].Address
 		batchID := offer.Chunks[i].BatchID
+		batchHash := offer.Chunks[i].BatchHash
```
ldeffenb, line 173: Isn't it more accurately a `StampHash` than a `BatchHash`?   The latter fairly accurately describes the `BatchID` from where I read things.   Trying to keep the distinction between `Batch` and `Stamp`.

This, of course, will apply to everywhere that this appears.

pkg/storer/internal/reserve/reserve.go
```diff
@@ -97,7 +97,12 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 	r.multx.Lock(string(chunk.Stamp().BatchID()))
 	defer r.multx.Unlock(string(chunk.Stamp().BatchID()))
 
-	has, err := r.Has(chunk.Address(), chunk.Stamp().BatchID())
+	batchHash, err := chunk.Stamp().Hash()
+	if err != nil {
+		return err
+	}
+
+	has, err := r.Has(chunk.Address(), chunk.Stamp().BatchID(), batchHash)
```
ldeffenb, line 105: Will this result in chunks that were re-stamped within the same batch as before appearing in the reserve twice?   An instance per stamp which includes the timestamp, since the updated timestamp changes the stamp's hash?

So instead of actually replacing the old stamp (or timestamp) with the new stamp (or timestamp), the new stamp is actually added to the reserves because the timestamp is updated?

ldeffenb: This concern should be verifiable in a test that inserts a stamped chunk into the reserve, updates the timestamp in the stamp (like a re-upload of a chunk would do), and then insert that updated timestamp chunk.   I suspect they'll both end up in the reserve rather than updating the timestamp.

ldeffenb: Here's the line in the original stamper where an existing stamp on a chunk has it's timestamp updated if it is re-stamped (re-uploaded) from the same batch.
https://github.com/ethersphere/bee/blob/58b62d6819fc40e2bb1e272c56a04b53ee06d0d6/pkg/postage/stamper.go#L51

acha-bill: There wound't be duplicate chunks upon restamping. The test is covered by https://github.com/ethersphere/bee/blob/feat-add-batch-hash/pkg/storer/internal/reserve/reserve_test.go#L138

Note that `stampindex.LoadOrStore` does not check the stamp hash. So it'll return the old stampIndex (including the old hash) for the `batchID/stampIndex`.
We then remove the old chunk using this old stamp hash.
https://github.com/ethersphere/bee/blob/feat-add-batch-hash/pkg/storer/internal/reserve/reserve.go#L138

pkg/pullsync/pullsync.go
acha-bill: fixed

pkg/storer/internal/reserve/reserve.go
ldeffenb: > We then remove the old chunk using this old stamp hash. https://github.com/ethersphere/bee/blob/feat-add-batch-hash/pkg/storer/internal/reserve/reserve.go#L138

I have some (unproven) suspicions that `removeChunk` may introduce storage corruptions in the sharky.   I did a naive hack when trying to work around the non-updating timestamps and ended up with a totally ravaged localstore.  This was duplicated on several nodes and the only thing they had in common is that they were attempting to remove/re-insert chunks whose timestamp had changed.   I didn't call "foul" at that time because I always blame my own code first and it was more expeditious to just remove my hack and live with non-updating timestamps.   But this may need some good testing in the field, with database consistency checks run on nodes that have logged "`replacing chunk stamp index`".

Maybe even consider elevating this log to `Warning` since it should be rare enough and we'd really like to know it's happening even without `--verbosity 5`?

istae (changes requested):
reserveRepair cmd needs to be changed to reflect the addition of a new field

pkg/storer/internal/stampindex/stampindex.go
```diff
@@ -98,17 +109,17 @@ func (i *Item) Unmarshal(bytes []byte) error {
 		return errStampItemUnmarshalInvalidSize
 	}
 	nsLen := int(binary.LittleEndian.Uint64(bytes))
-	if len(bytes) != 8+nsLen+swarm.HashSize+swarm.StampIndexSize+swarm.StampTimestampSize+swarm.HashSize+1 {
+	if len(bytes) != 8+nsLen+swarm.HashSize+swarm.StampIndexSize+swarm.StampTimestampSize+swarm.HashSize+swarm.HashSize+1 {
```
martinconic, line 112: Is the same size as above? If yes , can we add this into a constant ?

pkg/storer/internal/reserve/reserve.go
acha-bill: Elevated to warning

pkg/storer/internal/stampindex/stampindex.go
```diff
@@ -51,35 +52,43 @@ type Item struct {
...
-	case len(i.batchID) != swarm.HashSize:
+	case len(i.BatchID) != swarm.HashSize:
 		return nil, errStampItemMarshalBatchIDInvalid
-	case len(i.stampIndex) != swarm.StampIndexSize:
+	case len(i.StampIndex) != swarm.StampIndexSize:
 		return nil, errStampItemMarshalBatchIndexInvalid
 	}
 
-	buf := make([]byte, 8+len(i.namespace)+swarm.HashSize+swarm.StampIndexSize+swarm.StampTimestampSize+swarm.HashSize+1)
+	buf := make([]byte, 8+len(i.namespace)+swarm.HashSize+swarm.StampIndexSize+swarm.StampTimestampSize+swarm.HashSize+1+swarm.HashSize)
```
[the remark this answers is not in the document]
acha-bill: I think its fine

pkg/storer/internal/stampindex/stampindex.go
acha-bill: Not necessarily and it cannot be constant because it depends on the namespace (variable length)

istae:
Protocol version must be bumped.

istae (commented):
1. Protocol version must be bumped.

istae (approved):

martinconic (approved):
LGTM

janos (approved):

pkg/storer/internal/reserve/reserve.go
acha-bill: Hi @ldeffenb , bringing this back :)

Can you elaborate on your suspicion?
I'm seeing an issue where after updating a chunk timestamp, the old chunk is not removed from sharky (same address).

```
│ bee chunkstore: deleting chunk from sharky address 2a86d275607ef7cf284b15145c6539b1936128e7aa991b00100bec6fbcca86a8                                                                                                                                                                                                                                                                                           │
│ bee chunkstore: verifying delete on address 2a86d275607ef7cf284b15145c6539b1936128e7aa991b00100bec6fbcca86a8                                                                                                                                                                                                                                                                                                  │
│ bee chunkstore: failed to delete chunk from sharky address 2a86d275607ef7cf284b15145c6539b1936128e7aa991b00100bec6fbcca86a8                                                                                                                                                                                                                                                                                   │
│ bee "time"="2024-08-01 18:01:24.720744" "level"="warning" "logger"="node/storer/reserve" "msg"="replacing chunk stamp index" "old_chunk"="2a86d275607ef7cf284b15145c6539b1936128e7aa991b00100bec6fbcca86a8" "new_chunk"="2a86d275607ef7cf284b15145c6539b1936128e7aa991b00100bec6fbcca86a8"
│ bee "time"="2024-08-01 18:01:24.720864" "level"="info" "logger"="node/storer/reserve" "msg"="putting new chunk" "address"="2a86d275607ef7cf284b15145c6539b1936128e7aa991b00100bec6fbcca86a8" 
│ bee chunkstore: put chunk in sharky address 2a86d275607ef7cf284b15145c6539b1936128e7aa991b00100bec6fbcca86a8                                                                                                                                                                                                                                                                                                  │
│ bee chunkstore: chunk already exists in sharky. skipping address 2a86d275607ef7cf284b15145c6539b1936128e7aa991b00100bec6fbcca86a8 count 2  
```

https://github.com/ethersphere/bee/blob/bill/playground/pkg/storer/internal/chunkstore/chunkstore.go#L131-L137

ldeffenb: @acha-bill  My suspicion was caused by the observation that every node that I had that ended up with sharky corruption (detected by `db validate`) showed logs that indicated that a timestamp was updated causing `removeChunk` to be invoked during a Put operation.   I don't have direct evidence of what causes the corruption, but I suspect that somehow either an index is not removed properly or possibly not inserted properly.  But somehow, the sharky ended up with hashes that didn't match the associated index.

ldeffenb: https://github.com/ethersphere/bee/blob/bill/playground/pkg/storer/internal/chunkstore/chunkstore.go#L131-L137

Looking at this code block, remember that chunks don't necessarily leave the sharky on removal from just one index.   If a chunk is pinned in some reference, it will remain in because of a residual refCnt in the index.   But looking further back in the code, I see that was taken into account.

There may also be something in the actual sharky implementation that causes a retrieval to still work until something is committed thereby finalizing the delete.   It's been a long time since I looked at that code in detail.

PR #4719, opened 2024-07-06
feat: pre-signed chunks

### LINKED ISSUE #2905: Add API for receive already signed chunks by client
#### Summary
We need a way for upload to a node chunks that was already signed by the client with postage stamps.
This primarily permits to publish contents signed with the user's address instead of the node's one.

#### Motivation
This is useful in all cases where a node is used by more than one user. Currently every published content is signed by node, and so the node owner is the first responsible of uploaded contents, where instead the final user should be.

Moreover, this permits to discharge node from computing load of sign every chunk, distributing it to clients.

#### Implementation
New API and a chunk validity check have to be implemented.
Old APIs can be kept untouched, eventually node's chunk signature can be disabled by options.

#### Drawbacks
Don't see any.

nugaon:
Introducing `Swarm-Postage-Stamp` header on `/chunk` and `/soc` POST endpoints.

It allows to push chunk to the Swarm network with pre-signed postage stamp on the attached chunk by the Bee client.

The header value must be a valid Postage Stamp otherwise error 400 has been thrown.
It means that the stamp is signed against the attached chunk and the signature is created by its on-chain owner.
A correct Postage Stamp value is 113 bytes encoded as hex string and consists of:
- :32 - ID of the batch
- 32:40 - index
- 40:48 - timestamp
- 48: - signature
 

Closes #2905

### Checklist

### DIFF
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -149,7 +149,11 @@ paths:
         - Chunk
       parameters:
         - $ref: "SwarmCommon.yaml#/components/parameters/SwarmTagParameter"
-        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmPostageBatchId"
+        - in: header
+          schema:
+            $ref: "SwarmCommon.yaml#/components/parameters/SwarmPostageBatchId"
+          required: false
+        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmPostageStamp"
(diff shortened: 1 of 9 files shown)


istae (approved):
Nice!

janos (approved):
👍

martinconic (approved):

PR #4727, opened 2024-07-17
feat(gsoc): subscribe

### LINKED ISSUE #4333: Subscribe to specific SOC address
#### Summary

Similarly to the Postal Service for Swarm (PSS), event-based messaging feature could be provided by leveraging Single Owner Chunks (SOC) which would offer lesser resource intensive messaging between two peers. There should be a new API endpoint that provides a websocket for relaying incoming SOC payload. 

#### Motivation

PSS messaging mines Content Addressed Chunk (CAC) to the recipient's neighborhood. 
The default mine requirement is `bucketDepth` (16 bits matching in the recipient's overlay address). On each trial, the Bee client incrementing a nonce and building up the BMT on the chunk to check this criteria. 
Therefore, this leads to many hash operations and many trials for the successful mining for the sender, moreover, it puts the burden on the recipient to try to "open" all incoming chunks as a PSS message. This feature is usually blocked on gateways as well because of potential DOS attack.

One could send messages to the recipient in the same SOC several times with the same mutable postage batch where the recipient would mine the identifier based on the sender's Ethereum address used for their SOC and start listening on new messages. This mined identifier could be shared in a specific consensual feed of the recipient that the sender could read.

By this, sender does not have to mine message and recipient does not have to scan all chunks for messages. 
The mining of the identifier is much faster than mining CAC.

#### Implementation

GET gsoc/subscribe/{socAddress}

where the `socAddress` is the calculated address from the mined identifier and the sender's Ethereum address and it should fall within the recipient's neighborhood.

nugaon:
This PR is partial implementation of the GSOC SWIP: https://github.com/ethersphere/SWIPs/pull/41
It addresses only to have GSOC subscription feature which neglects GSOC retrieval related functionalities.
This can substitute the PSS feature with lower resource requirements.

Introducing `/gsoc/subscribe/{address}` endpoint which opens a websocket channel for reading incoming payloads of the Single Owner Chunk under the defined address.

Modifications were required in the core packages so that different payloads of a SOC can be synced.
Identity Address of a chunk is used as a key to distinguish different payloads of a SOC which is the hash of the SOC address and its wrapped address in case of SOC and it remains the CAC address in the case of CAC.

Closes #4333

### Checklist


acha-bill (approved):

zelig (changes requested):
for this functionality too work, 
- pullsync needs to process  each instance of the soc on the same address.
  - if a soc has the same postage sstamp as an earlier one, the node it is pushsynced to might not even save it and therefore will not offer it via pullsync
  -  with different postage stamp but same soc address the soc should be saved and offered via pullsync and the downstrream party should request and save it
  
  -pushsync should never skip pushing the same soc addresss with different payload
  
  these assumptions need to be doublechecked to be sure the functionality makes sense. 
it shooould also be properly going through a SWIP

pkg/gsoc/gsoc.go
```diff
@@ -0,0 +1,100 @@
...
+	close(l.quit)
+	l.handlersMu.Lock()
+	defer l.handlersMu.Unlock()
+
+	l.handlers = make(map[[32]byte][]*handler) //unset handlers on shutdown
+
+	return nil
+}
+
+// handler defines code to be executed upon reception of a GSOC sub message.
```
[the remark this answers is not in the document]
nugaon: only gsoc api calls this without error handling. it should just pass the incoming chunk data to the websocket it does not have any business logic behind so I did not take it necessary

pkg/api/gsoc_test.go
```diff
@@ -0,0 +1,171 @@
...
+		t.Fatal(err)
+	}
+	cl.SetReadLimit(swarm.ChunkSize)
+
+	ch, _ := cac.New(payload)
+	socCh := soc.New(id, ch)
+	ch, _ = socCh.Sign(signer)
+	socCh, _ = soc.FromChunk(ch)
+
+	// close the websocket before calling pss with the message
```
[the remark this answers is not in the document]
nugaon: ofc I always copy from great sources : D

pkg/pullsync/pullsync.go
```diff
@@ -244,7 +247,9 @@ func (s *Syncer) Sync(ctx context.Context, peer swarm.Address, bin uint8, start
 
 		if cac.Valid(chunk) {
 			go s.unwrap(chunk)
-		} else if !soc.Valid(chunk) {
+		} else if chunk, err := soc.FromChunk(chunk); err == nil {
+			go s.gsocHandler(*chunk)
```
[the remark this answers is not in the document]
nugaon: you may be right as well as below, I just remove at both places

zelig:
this has failing CI checks @nugaon and my global concern is not responded to.

nugaon:
@zelig I believe that the push/pull syncs behave normally since it was tested out on mainnet already.

the CI checks are passing, parallel GSOC writes will be resolved presumably by #4866

pkg/storer/internal/reserve/reserve.go
```diff
@@ -24,11 +25,27 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/swarm"
 	"github.com/ethersphere/bee/v2/pkg/topology"
 	"golang.org/x/sync/errgroup"
-	"resenje.org/multex"
 )
 
 const reserveScope = "reserve"
 
+type multexLock struct {
```
acha-bill, line 32: Why not use "resenje.org/multex" as before?

pkg/storer/internal/reserve/reserve.go
```diff
@@ -145,95 +164,77 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
-			curr := binary.BigEndian.Uint64(chunk.Stamp().Timestamp())
-			if prev >= curr {
-				return fmt.Errorf("overwrite same chunk. prev %d cur %d batch %s: %w", prev, curr, hex.EncodeToString(chunk.Stamp().BatchID()), storage.ErrOverwriteNewerChunk)
-			}
 
-			// index collision with another chunk
-			if loadedStampIndex {
-				prev := binary.BigEndian.Uint64(oldStampIndex.StampTimestamp)
+			// same index
+			if bytes.Equal(sameAddressOldStamp.Index(), chunk.Stamp().Index()) {
```
acha-bill, line 169: Even in this same address condition, you also need to check if the new stamp is overwriting a different unrelated stamp index.

pkg/storer/internal/reserve/reserve.go
martinconic, line 32: Definitely this is not a good approach for the multexLock could lead to deadlocks when trying to lock multiple items. Why the change from resenje multex? cc. @janos

acha-bill (changes requested):

pkg/storer/internal/reserve/reserve.go
acha-bill: https://github.com/ethersphere/bee/blob/master/pkg/storer/internal/reserve/reserve.go#L155C7-L155C23

pkg/api/gsoc.go
```diff
@@ -0,0 +1,119 @@
...
+			// client gone
+			return
+		case <-ticker.C:
+			err = conn.SetWriteDeadline(time.Now().Add(writeDeadline))
+			if err != nil {
+				s.logger.Debug("gsoc ws: set write deadline failed", "error", err)
+				return
+			}
+			if err = conn.WriteMessage(websocket.PingMessage, nil); err != nil {
+				// error encountered while pinging client. client probably gone
```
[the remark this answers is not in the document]
nugaon: copied from PSS test

pkg/postage/stamper.go
```diff
@@ -39,13 +40,13 @@ func NewStamper(store storage.Store, issuer *StampIssuer, signer crypto.Signer)
 
 // Stamp takes chunk, see if the chunk can be included in the batch and
 // signs it with the owner of the batch of this Stamp issuer.
-func (st *stamper) Stamp(addr swarm.Address) (*Stamp, error) {
+func (st *stamper) Stamp(addr, idAddr swarm.Address) (*Stamp, error) {
```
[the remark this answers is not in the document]
nugaon: changed the interface

pkg/storer/internal/reserve/reserve.go
nugaon: that should be handled below, shouldn't it? (it does not return if there is no index collision)
anyway the tests are passing it should be fine.

pkg/postage/stamper.go
janos, line 43: It would be good to explain the purpose of arguments as there are two addresses now.

pkg/storer/internal/reserve/reserve_test.go
```diff
@@ -435,16 +446,17 @@ func TestSameChunkAddress(t *testing.T) {
...
+		checkStore(t, ts.IndexStore(), &reserve.ChunkBinItem{Bin: bin2, BinID: ch2BinID, StampHash: ch2StampHash}, false)
 		checkStore(t, ts.IndexStore(), &reserve.ChunkBinItem{Bin: bin2, BinID: ch3BinID, StampHash: ch3StampHash}, false)
 
 		size2 := r.Size()
 
-		// (ch1 + ch2) == 2 and then ch3 reduces reserve size by 1
-		if size2-size1 != 1 {
+		// (ch1 + ch2) == 2
+		if size2-size1 != 2 {
 			t.Fatalf("expected reserve size to increase by 1, got %d", size2-size1)
```
acha-bill, line 460:
```suggestion
			t.Fatalf("expected reserve size to increase by 2, got %d", size2-size1)
```

pkg/storer/internal/reserve/reserve.go
```diff
@@ -145,95 +148,77 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
-						"old_chunk", oldStampIndex.ChunkAddress,
-						"new_chunk", chunk.Address(),
-						"batch_id", hex.EncodeToString(chunk.Stamp().BatchID()),
-					)
-					// remove index items and chunk data
-					err = r.removeChunk(ctx, s, oldStampIndex.ChunkAddress, oldStampIndex.BatchID, oldStampIndex.StampHash)
-					if err != nil {
-						return fmt.Errorf("failed removing older chunk %s: %w", oldStampIndex.ChunkAddress, err)
-					}
-					shouldDecrReserveSize = true
```
acha-bill, line 172: `shouldDecreaseReserveSize` is now always false

acha-bill (changes requested):

pkg/gsoc/gsoc.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"sync"
+
+	"github.com/ethersphere/bee/v2/pkg/log"
+	"github.com/ethersphere/bee/v2/pkg/soc"
+	"github.com/ethersphere/bee/v2/pkg/swarm"
+)
+
+type Listener interface {
+	Subscribe(address [32]byte, handler handler) (cleanup func())
```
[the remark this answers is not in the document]
nugaon: I just wanted a define a package level struct for handling. why is it not okay? I don't want to export it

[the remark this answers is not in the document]
janos: Interface is defined in order to define a behaviour that can have multiple implementations. I see no reason to have an interface exposed or even created if there is only one implementation and that only the containing packages can implement more. What is the purpose of the interface here if only one implementation exists and no other package is able to create another implementation?

pkg/postage/stamper.go
nugaon: added comment to the interface, in the close future request address may be not defined.

pkg/gsoc/gsoc.go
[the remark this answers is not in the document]
nugaon: `handler` is used at multiple places and the code looks better with that than using `func([]byte)` definitions everywhere IMHO.

pkg/postage/stamper.go
janos: Thanks. It would be very nice to have full sentences with correct capitalization and dots at the end so that it is nicely readable on https://pkg.go.dev/github.com/ethersphere/bee. This is more of a general comment as well.

nugaon: got it, latest changes have this.

pkg/gsoc/gsoc.go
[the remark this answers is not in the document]
nugaon: I wouldn't pollute the code with making a type public when it is not used outside of the package.
`Subscribe` is called in the code and the handler function is not needed to be defined outside of the function call.

if it happens in the future, it may worth to make it public.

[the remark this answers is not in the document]
janos: It would be good to have a comment why the interface is deliberately left unimplementable outside of the package, if you insist in making it as it is.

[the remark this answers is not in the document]
nugaon: I added a comment line about it.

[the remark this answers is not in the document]
janos: This is not a comment about the interface. Would you, please, add a comment why the interface is made unimplementable by design.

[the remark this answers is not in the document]
nugaon: because of that I wrote in the comment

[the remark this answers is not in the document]
acha-bill: Janos is pointing out that since the `handler` type is unexported, the `Listener` interface cannot be implemented outside the package. Interfaces are generally intended for multiple implementations. If there will only be a single implementation of Listener (currently within the gsoc package), then it would be more appropriate to use a struct instead of an interface.

The comment is about the `Listener` interface. Not the `handler` type.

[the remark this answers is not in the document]
janos: @nugaon the comment you added is on the handler type https://github.com/ethersphere/bee/pull/4727/commits/691a43d51d2e2ea591543c6fc0620f534bdcb189 does not address any issues related to the interface in this discussion.

And I support to use a struct as per @acha-bill explanation.

pkg/postage/stamper.go
```diff
@@ -21,7 +21,9 @@ var (
 
 // Stamper can issue stamps from the given address of chunk.
 type Stamper interface {
-	Stamp(swarm.Address) (*Stamp, error)
+	// addr is the request address of the chunk and idAddr is the identity address of the chunk.
+	Stamp(addr, idAddr swarm.Address) (*Stamp, error)
```
istae, line 25: Let's have a new Stamp function, maybe called StampSOC that's specific to this case.

pkg/pullsync/pullsync.go
```diff
@@ -87,21 +88,23 @@ func New(
 	streamer p2p.Streamer,
 	store storer.Reserve,
 	unwrap func(swarm.Chunk),
+	gsocHandler func(soc.SOC),
```
istae, line 91: This handler function should take a pointer to SOC. Let's change this everywhere

pkg/storer/internal/reserve/reserve.go
```diff
@@ -100,9 +101,12 @@ func New(
 //     if the new chunk has a higher stamp timestamp (regardless of batch type and chunk type, eg CAC & SOC).
 func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
```
istae, line 103: There are too many changes and it's seems that we are no longer handling cases we used to before. It also feels like we've changed the tests until they started passing. What I would like to see is we revert all these changes. 

Maybe we can handle this by, in the case of a same chunk address collision, we first check by chunk type. If it's a soc then we compare the identity addresses by computing them. If some checks pass, we store the new chunk and return early.

istae (changes requested):

pkg/storer/internal/reserve/reserve.go
```diff
@@ -100,9 +101,12 @@ func New(
...
 
+	chunkType := storage.ChunkType(chunk)
+
 	// batchID lock, Put vs Eviction
-	r.multx.Lock(string(chunk.Stamp().BatchID()))
```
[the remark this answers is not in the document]
istae: Why do you need to lock by stamp?

pkg/postage/stamper.go
istae: So we can have the original function definition untouched

pkg/storer/internal/reserve/reserve.go
istae: Ping me for help on deep diving these changes.

pkg/api/soc.go
```diff
@@ -124,6 +124,8 @@ func (s *Service) socUploadHandler(w http.ResponseWriter, r *http.Request) {
 			jsonhttp.NotFound(w, "batch with id not found")
 		case errors.Is(err, errInvalidPostageBatch):
 			jsonhttp.BadRequest(w, "invalid batch id")
+		case errors.Is(err, errUnsupportedDevNodeOperation):
+			jsonhttp.NotImplemented(w, "operation is not supported in dev mode")
```
istae, line 128: In this soc upload handler. If you allow for deferred uploads, then you also need to make the identity adress changes in the uploadstore. If you allow for pinning, again the same problem. The pinstore works by chunk adress being unique so lots of changes are needed there.

istae:
The cachestore is also problematic because it expect that the chunk adress is unique so if you cache one soc then try to fetch the same soc from the network expecting a different payloads, the cache will return the first soc over and over again.

istae:
What is the plan with retrieving these chunks through the regular Api and not the gsoc handler? As things stands, if you try to retrieve a soc chunk, it's mostly random which version or payload you will get. If the only mechanism for accessing these chunks is through the gsoc handler, then do we actually need to store them?

pkg/storer/internal/reserve/reserve.go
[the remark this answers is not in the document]
nugaon: original idea was that if somehow a chunk with mutable batch stamped twice that have the same SOC address then it will do not let to put it into the reserve in the same time.
anyway you are right, I reverted this change since the batchID locking is higher level locking.

pkg/storer/internal/reserve/reserve.go
nugaon: there was some misunderstanding by what sort of logic the chunks should be stored in the reserve.
It was checked by batch ID when only the valid stamps matter. if there is a valid stamp which was not stored before you must store it.

pkg/storer/internal/reserve/reserve.go
```diff
@@ -145,95 +148,77 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
-			curr := binary.BigEndian.Uint64(chunk.Stamp().Timestamp())
-			if prev >= curr {
-				return fmt.Errorf("overwrite same chunk. prev %d cur %d batch %s: %w", prev, curr, hex.EncodeToString(chunk.Stamp().BatchID()), storage.ErrOverwriteNewerChunk)
-			}
 
-			// index collision with another chunk
-			if loadedStampIndex {
-				prev := binary.BigEndian.Uint64(oldStampIndex.StampTimestamp)
+			// same index
+			if bytes.Equal(sameAddressOldStamp.Index(), chunk.Stamp().Index()) {
```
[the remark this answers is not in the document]
nugaon: handled below

pkg/postage/stamper.go
nugaon: later it may change to have only one param but with idAddress.
I do not see the reason to complicate this further there should be only one stamp funtion

pkg/storer/internal/reserve/reserve.go
```diff
@@ -145,95 +148,77 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 			if err != nil {
 				return err
 			}
-			prev := binary.BigEndian.Uint64(sameAddressOldStampIndex.StampTimestamp)
```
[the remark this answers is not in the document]
nugaon: the overwrite should be stamp based and not batch based.

postage timestamp is attached to the stamp so you need to check it when there is a stamp collision with the already stored chunk.

pkg/pullsync/pullsync.go
nugaon: why is it need a pointer? I thought a struct value is already a reference.

in any way, I took the hook function of PSS as a sample above this line.

pkg/api/soc.go
nugaon: this is the gsoc subscription PR, it does not care about the retrievals including the pinning.

can you elaborate where should I change things in the uploadstore please?

nugaon:
This PR is one part of the SWIP https://github.com/ethersphere/SWIPs/pull/41 and only solves the GSOC subscription feature.
In other words, all retrieval related things are neglected in the changes.

Ultimately (in the next GSOC PR), all payloads must be stored and bee should be able to retrieve them.

pkg/soc/utils.go
```diff
@@ -0,0 +1,34 @@
...
+
+package soc
+
+import "github.com/ethersphere/bee/v2/pkg/swarm"
+
+// IdentityAddress returns the internally used address for the chunk
+// since the single owner chunk address is not a unique identifier for the chunk,
+// but hashing the soc address and the wrapped chunk address is.
+// it is used in the reserve sampling and other places where a key is needed to represent a chunk.
+func IdentityAddress(chunk swarm.Chunk) (swarm.Address, error) {
```
istae, line 13: a unit test would be nice

istae:
Summarizing my comments, for me to approve the PR:

1. the reserve changes must be reverted (and handled in a future PR that fixes the chunkstore to be able to store multiple payloads of a SOC).
2. the SOC upload API must only be usable with direct and non-pinned uploads unless the pinstore and uploadstore is fixed to correctly store multiples payloads of a SOC

nugaon:
> 1. the reserve changes must be reverted (and handled in a future PR that fixes the chunkstore to be able to store multiple payloads of a SOC).

different payloads of a SOC need to be synced and it is likely that one uploads different payloads with one postage batch so this change is needed.
I will see what I can do with the chunkstore problem in this short time but it is better in this way than deciding what to store in the reserve based on batch ID.

> 2\. the SOC upload API must only be usable with direct and non-pinned uploads unless the pinstore and uploadstore is fixed to correctly store multiples payloads of a SOC.

pinning GSOC is futile, it is used for information signaling. The way how you can request different payloads of a SOC will be handled in a different PR.
It is [still unclear for me](https://github.com/ethersphere/bee/pull/4727#discussion_r1828328081) where and what I should tackle at uploadstore.

> 3\. my other comments addressed that are not related to the above

which ones? : )

nugaon:
> 1. the reserve changes must be reverted (and handled in a future PR that fixes the chunkstore to be able to store multiple payloads of a SOC).

chunkstore overwrites the previous GSOC payload since this PR does not address GSOC store.
pullSync will pull the newer payload of SOC. It calls ReserveHas(pkg/pullsync/pullsync.go:295) on offered items where StampHash is part of the ID of BatchRadiusItem -> all different SOC payloads will be stamped differently.

> 2\. the SOC upload API must only be usable with direct and non-pinned uploads unless the pinstore and uploadstore is fixed to correctly store multiples payloads of a SOC.

As mentioned, this PR does not handle multiple GSOC store. that will be introduced in a different PR which will introduce postage stamping breaking change because that is required for multiple GSOC store in order to avoid from infinite store (detailed in the SWIP).  
deferred upload is not supported in this phase because multiple GSOC upload in the same time will overwrite for the latest version within one pullsync time and the information signal will be lost.

I hope I addressed all other concerns.

nugaon:
@istae please do not force push to the PR and do not commit directly (create another PR based on this).

istae:
closing because this PR which took over this one was merged https://github.com/ethersphere/bee/pull/4901

Issue #4696, opened 2024-05-28
/stewardship doesn't check final leaves


ldeffenb:
### Context
2.1.0 (and all earlier, I suspect)

### Summary
I have a file reference on the swarm mainnet that is missing a chunk, but yet the /stewardship API insists that the reference is retrievable.   I have the reference locally pinned, but due to the earlier bugs in the swarm, one of the BMT chunks is no longer local and in fact, no longer in the swarm.

mainet Reference: `a3ad3e5ea830b930c62ddf11d198c6492261bbdb612fe17fabe7585b5aedd52d`
Composed of BMT chunks:
```
a3ad3e5ea830b930c62ddf11d198c6492261bbdb612fe17fabe7585b5aedd52d
9204ce19779429a0c37f0f2eb23d539dd9226d84c4bd287a2aede3a2940177a1
cf973b6794a7d1dd5c8df9c2085f31643f4da935df791d0bb077800641600dfb
d71d0a8405546fe15228edecc4f6efc0c56f336f81ec48f6377fc89ad84c0f5c
e1cc5a6c1e9b270bde443f392cfa9fa8b0cef168d147fe23828790a8521b4584
ec9bc0f22a84e2e4af4c3ddcd5213b3114ea60441a94ca14486e7db8c91c93f1 <- Missing
fc7f913696b3d06756cfaee1d84db4b0e1a7c041f43bd9456174b7bd0816c136
```

If you try to retrieve the missing chunk with the /chunks API, you will see that it is truly missing, even though /stewardship says the entire main reference is retrievable.  Pulling the file contents with /bytes will give you an incomplete file.

```
curl http://localhost:1633/bytes/a3ad3e5ea830b930c62ddf11d198c6492261bbdb612fe17fabe7585b5aedd52d
curl: (18) transfer closed with 22592 bytes remaining to read
```

### Expected behavior
If /stewardship says a reference is retrievable, I'd expect to be able to retrieve it.

### Actual behavior
/stewardship says it is retrievable
/bytes is short
/chunks of the missing chunk fails.

### Steps to reproduce
Just try the references shown above on the current mainnet swarm.

### Possible solution
The stewardship API invokes `s.steward.IsRetrievable`
https://github.com/ethersphere/bee/blob/97e7ee699be3b4325a233b1ca2dc177cd88f17e1/pkg/api/stewardship.go#L94

[...]

PR #4735, opened 2024-07-26
fix: stewardship api doesn't check final leaves

martinconic:
### Checklist

### Description
[Issue](https://github.com/ethersphere/bee/issues/4696)

### DIFF
--- a/pkg/steward/steward.go
+++ b/pkg/steward/steward.go
@@ -34,6 +34,7 @@ type steward struct {
 	netStore     storer.NetStore
 	traverser    traversal.Traverser
 	netTraverser traversal.Traverser
+	netGetter    retrieval.Interface
 }
 
 func New(ns storer.NetStore, r retrieval.Interface, joinerPutter storage.Putter) Interface {
@@ -41,6 +42,7 @@ func New(ns storer.NetStore, r retrieval.Interface, joinerPutter storage.Putter)
 		netStore:     ns,
 		traverser:    traversal.New(ns.Download(true), joinerPutter),
 		netTraverser: traversal.New(&netGetter{r}, joinerPutter),
+		netGetter:    r,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/steward/steward.go
```diff
@@ -79,8 +81,11 @@ func (s *steward) Reupload(ctx context.Context, root swarm.Address, stamper post
 
 // IsRetrievable implements Interface.IsRetrievable method.
 func (s *steward) IsRetrievable(ctx context.Context, root swarm.Address) (bool, error) {
-	noop := func(leaf swarm.Address) error { return nil }
-	switch err := s.netTraverser.Traverse(ctx, root, noop); {
+	fn := func(a swarm.Address) error {
+		_, err := s.netGetter.RetrieveChunk(ctx, a, swarm.ZeroAddress)
```
ldeffenb, line 85: I considered this naive approach, but what it ends up doing is retrieving all non-leaf chunks TWICE from the swarm doubling the bandwidth cost of a /stewardship check.   The first netGetter call happens while expanding the BMT for the non-leaf chunks, but ALL chunks end up doing this callback resulting in the netGetter (which bypasses the local cache) redundantly pulling most of the chunks a second time.

ldeffenb (changes requested):
I'm just not sure how to do this efficiently without having the BMT traverser somehow informing the callback if the netGetter was used or not yet for the address being iterated.   Otherwise I would have proposed a solution...

istae (approved):

pkg/steward/steward.go
janos: Good catch. By my opinion, this can be a comment in the code for the improvement, as the functionality is here with minimal changes.

janos (approved):

pkg/steward/steward.go
ldeffenb: If that's the case, then we might as well allow the traverser to use the normal getter method that uses the local cache and do ALL of the netgetter checks here in the callback?   Not many more changes, and eliminates the redundancy.

janos: The net getter is to be able to get the chunk even if it is not locally available. It would be good to cache the chunk upon retrieval in localstore, though.

ldeffenb: As I understand the netGetter, it ONLY retrieves from the swarm, and doesn't even try the localstore (try it on a locally pinned, but stamp-expired reference, it will fail, while the /bytes and /bzz will still work).   My point is that since we're going to netGet every single traversed reference in the callback, then don't make a netTraverser using the netGetter, but use a normal traverser that will use localstore and only go to the swarm if necessary.  That will reduce the redundancy of hitting the swarm at least.

janos: Yes, as long as it is certain that the getter based on localstore will fallback to retrieve the chunk from the network.

acha-bill (approved):

pkg/steward/steward.go
martinconic: The retrieval protocol has a singleflight mechanism so this improves efficiency of requests that are fired in succession.

Issue #4739, opened 2024-07-26
SOC pullsync with different payloads


nugaon:
### Context

most recent master branch
fdp-play 5 nodes

### Summary

When multiple payloads being uploaded under one Single Owner Chunk address, the pullsync will not transfer the chunk with newer payload to the downstream.
Nevertheless, the SOC gets marked for syncing each time when a node gets a new payload.

This can lead to Shelling point mismatches.

### Expected behavior

All valid versions of a Single Owner Chunk are synchronized in the neighborhood. 

### Actual behavior

It always synchronizes the first payload that reached upstream even when a new (and different) payload should be transferred.  

### Steps to reproduce

My tests based on the #4727 gsoc-subscribe branch and I tried to upload different payloads from one node meanwhile another node was listening on the `/gsoc/subscribe/{address}` websocket endpoint.

In order to carry out this manual testing easily, I used the [GSOC JS library](https://github.com/anythread/gsoc).

receiver node:
```js
gsoc = require('./dist/index.js')
is = new gsoc.InformationSignal('http://localhost:11633')
cancelSub = is.subscribe({onMessage: msg => console.log('gsoc-pull-sync', msg), onError: console.log})
```

sender node:
```js
gsoc = require('./dist/index.js')
is = new gsoc.InformationSignal('http://localhost:1633', {postageBatchId: '69650e45a19fe3a625b73ef69cf586f6209ff3588e012a05c8e50072a7e39263'})
is.write('Hello there!')
is.write('This is a really important and uplifting message. I hope you will receive it!')
```

### Possible solution
Reserve [chunk-get](https://github.com/ethersphere/bee/blob/master/pkg/storer/internal/reserve/reserve.go#L211-L225) should be able to retrieve chunk satisfying optional postageBatchHash parameter.

nikipapadatou:
included in the GSOC solution in 2.3

PR #4749, opened 2024-08-06
fix: override same chunk

acha-bill:
### Checklist

### Description
Override chunk with the same address.

When we attempt to delete the chunk, the delete operation is appended to the current batch. However, since this [Get](https://github.com/ethersphere/bee/blob/master/pkg/storer/internal/chunkstore/chunkstore.go#L77) reads from disk, it returns the old `refCnt`. So it ignores the new chunk.

To fix this, we introduce a new `replace` method that does not relly on `refCnt`.

### Related Issue (Optional)

#4739

### DIFF
--- a/pkg/storer/internal/reserve/reserve.go
+++ b/pkg/storer/internal/reserve/reserve.go
@@ -153,6 +153,9 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 						"batch_id", hex.EncodeToString(chunk.Stamp().BatchID()),
 					)
 					err = r.removeChunk(ctx, s, oldItem.ChunkAddress, oldItem.BatchID, oldItem.StampHash)
+					if err != nil {
+						return fmt.Errorf("failed removing older chunk %s: %w", oldItem.ChunkAddress, err)
+					}
 				}
 			}
 
--- a/pkg/storer/internal/reserve/reserve.go
+++ b/pkg/storer/internal/reserve/reserve.go
@@ -116,22 +116,6 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 	r.multx.Lock(strconv.Itoa(int(bin)))
 	defer r.multx.Unlock(strconv.Itoa(int(bin)))
 
-	// If this chunk already exists in the chunkstore, we delete it in a separate transaction.
-	// The oldChunk will be restored if the next tx fails.
-	oldChunk, err := r.st.ChunkStore().Get(ctx, chunk.Address())
-	if err != nil && !errors.Is(err, storage.ErrNotFound) {
-		return err
-	}
-	if oldChunk != nil {
-		err = r.st.Run(ctx, func(s transaction.Store) error {
-			r.logger.Warning("deleting chunk from chunkstore", "address", chunk.Address().String())
-			return s.ChunkStore().Delete(ctx, chunk.Address())
-		})
-		if err != nil {
-			return err
-		}
-	}
-
 	err = r.st.Run(ctx, func(s transaction.Store) error {
 
 		oldItem, loadedStamp, err := stampindex.LoadOrStore(s.IndexStore(), reserveNamespace, chunk)
@@ -202,7 +186,17 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 			return err
 		}
 
-		err = s.ChunkStore().Put(ctx, chunk)
+		chunkType := storage.ChunkType(chunk)
+		has, err := s.ChunkStore().Has(ctx, chunk.Address())
+		if err != nil {
+			return err
+		}
+		if !has {
+			err = s.ChunkStore().Put(ctx, chunk)
+		} else if chunkType == swarm.ChunkTypeSingleOwner {
+			r.logger.Warning("replacing SOC in chunkstore", "chunk", chunk.Address())
+			err = s.ChunkStore().Replace(ctx, chunk)
+		}
 		if err != nil {
 			return err
 		}
@@ -214,11 +208,6 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 		return nil
 	})
 	if err != nil {
-		if oldChunk != nil {
-			err = errors.Join(err, r.st.Run(ctx, func(s transaction.Store) error {
-				return s.ChunkStore().Put(ctx, oldChunk)
-			}))
-		}
 		return err
 	}
 	return nil
--- a/pkg/storer/internal/reserve/reserve.go
+++ b/pkg/storer/internal/reserve/reserve.go
@@ -123,7 +123,7 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 			return fmt.Errorf("load or store stamp index for chunk %v has fail: %w", chunk, err)
 		}
 
-		sameAddressOldChunkStamp, err := chunkstamp.LoadWithBatchID(s.IndexStore(), reserveNamespace, chunk.Address(), chunk.Stamp().BatchID())
+		sameAddressOldChunkStamp, err := chunkstamp.Load(s.IndexStore(), reserveNamespace, chunk.Address())
 		if err != nil && !errors.Is(err, storage.ErrNotFound) {
 			return err
 		}
--- a/pkg/storer/internal/reserve/reserve.go
+++ b/pkg/storer/internal/reserve/reserve.go
@@ -117,11 +117,98 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 	defer r.multx.Unlock(strconv.Itoa(int(bin)))
 
 	return r.st.Run(ctx, func(s transaction.Store) error {
-		// index collision (same or different chunk)
+
 		oldItem, loadedStamp, err := stampindex.LoadOrStore(s.IndexStore(), reserveNamespace, chunk)
 		if err != nil {
 			return fmt.Errorf("load or store stamp index for chunk %v has fail: %w", chunk, err)
 		}
+
+		sameAddressOldChunkStamp, err := chunkstamp.LoadWithBatchID(s.IndexStore(), reserveNamespace, chunk.Address(), chunk.Stamp().BatchID())
+		if err != nil && !errors.Is(err, storage.ErrNotFound) {
+			return err
+		}
+
+		// same address
+		if sameAddressOldChunkStamp != nil {
+			sameAddressOldStampIndex, err := stampindex.LoadWithStamp(s.IndexStore(), reserveNamespace, sameAddressOldChunkStamp)
+			if err != nil && !errors.Is(err, storage.ErrNotFound) {
+				return err
+			}
+			prev := binary.BigEndian.Uint64(sameAddressOldStampIndex.StampTimestamp)
+			curr := binary.BigEndian.Uint64(chunk.Stamp().Timestamp())
+			if prev >= curr {
+				return fmt.Errorf("overwrite prev %d cur %d batch %s: %w", prev, curr, hex.EncodeToString(chunk.Stamp().BatchID()), storage.ErrOverwriteNewerChunk)
+			}
+			if loadedStamp {
+				prev := binary.BigEndian.Uint64(oldItem.StampTimestamp)
+				if prev >= curr {
+					return fmt.Errorf("overwrite prev %d cur %d batch %s: %w", prev, curr, hex.EncodeToString(chunk.Stamp().BatchID()), storage.ErrOverwriteNewerChunk)
+				}
+			}
+
+			// load item to get the binID
+			oldBatchRadiusItem := &BatchRadiusItem{
+				Bin:       bin,
+				Address:   chunk.Address(),
+				BatchID:   sameAddressOldStampIndex.BatchID,
+				StampHash: sameAddressOldStampIndex.StampHash,
+			}
+			err = s.IndexStore().Get(oldBatchRadiusItem)
+			if err != nil {
+				return err
+			}
+
+			err = errors.Join(
+				s.IndexStore().Delete(oldBatchRadiusItem),
+				s.IndexStore().Delete(&ChunkBinItem{Bin: bin, BinID: oldBatchRadiusItem.BinID}),
+				stampindex.Delete(s.IndexStore(), reserveNamespace, swarm.NewChunk(chunk.Address(), nil).WithStamp(sameAddressOldChunkStamp)),
+				chunkstamp.DeleteWithStamp(s.IndexStore(), reserveNamespace, chunk.Address(), sameAddressOldChunkStamp),
+			)
+			if err != nil {
+				return err
+			}
+
+			err = stampindex.Store(s.IndexStore(), reserveNamespace, chunk)
+			if err != nil {
+				return err
+			}
+
+			err = chunkstamp.Store(s.IndexStore(), reserveNamespace, chunk)
+			if err != nil {
+				return err
+			}
+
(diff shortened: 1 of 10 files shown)


pkg/storer/internal/reserve/reserve.go
```diff
@@ -197,6 +213,15 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 
 		return nil
 	})
+	if err != nil {
+		if oldChunk != nil {
```
ldeffenb, line 217: And if the process terminates after the delete transaction and before this recovery, the original chunk is completely gone, right?

acha-bill: fixed

pkg/storer/internal/transaction/transaction.go
```diff
@@ -242,6 +242,13 @@ func (c *chunkStoreTrx) Iterate(ctx context.Context, fn storage.IterateChunkFn)
 	return chunkstore.Iterate(ctx, c.indexStore, c.sharkyTrx, fn)
 }
 
+func (c *chunkStoreTrx) Replace(ctx context.Context, ch swarm.Chunk) (err error) {
+	defer handleMetric("chunkstore_replace", c.metrics)(&err)
```
ldeffenb, line 246: See if I can remember to add this metric to one of my Grafana dashboards!

ldeffenb (approved):
I like the look of this better, simply replace it within the transaction rather than the possibility of dangling/orphan chunks.

martinconic (approved):

nugaon (approved):
While it resolves #4739 , I hoped we can handle two problems at the same time.

This fix does not cover the case when one uses postage stamp for the replaced single owner chunk; that stamp is used up for a non-living chunk on the uploader side when storage providers do not benefit from this pull syncing.
Former is not that significant problem especially if the batch is mutable while the last could be a huge improvement IMO since this is a supported feature in storage incentives with which operators can grow they reserve.

Nevertheless, there is no chunk retrieval mechanism to download different payloads of SOC at the moment, (also may require sharky changes) so we should merge this in order to have a more solid shelling point.

pkg/storer/internal/reserve/reserve.go
```diff
@@ -152,6 +152,28 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 			if err != nil {
 				return fmt.Errorf("failed updating stamp index: %w", err)
 			}
+		} else {
```
istae, line 155: this separate if and else does not work.
it could be that the "new" chunk is replacing a chunk on a stamp collision, but the "new" chunk may have already been stored in the reserve with an older timestamp (and different stamp index), which means you will end up with two copies of the same chunk in the reserve. not good.

please do not deviate from the pseudo code here  https://github.com/ethersphere/bee/compare/master...pseudocode-reserve-fix

pkg/storer/internal/reserve/reserve.go
```diff
@@ -186,7 +208,16 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 			return err
 		}
 
-		err = s.ChunkStore().Put(ctx, chunk)
+		has, err := s.ChunkStore().Has(ctx, chunk.Address())
```
istae, line 211: checking against the chunkstore wont work because the cache, pin, and the uploadstore also shares the chunkstore with the reserve, so you may end up with false positives. 

please do not deviate from the pseudo code here https://github.com/ethersphere/bee/compare/master...pseudocode-reserve-fix

istae (changes requested):

pkg/storer/internal/reserve/reserve.go
```diff
@@ -122,6 +122,93 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
+			}
+			err = s.IndexStore().Get(oldBatchRadiusItem)
+			if err != nil {
+				return err
+			}
+
+			err = errors.Join(
+				s.IndexStore().Delete(oldBatchRadiusItem),
+				s.IndexStore().Delete(&ChunkBinItem{Bin: bin, BinID: oldBatchRadiusItem.BinID}),
+				stampindex.Delete(s.IndexStore(), reserveNamespace, swarm.NewChunk(chunk.Address(), nil).WithStamp(sameAddressOldChunkStamp)),
```
istae, line 164: maybe add a new `DeleteWithStamp`

pkg/storer/internal/reserve/reserve.go
```diff
@@ -122,6 +122,93 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
+			}
+
+			err = chunkstamp.Store(s.IndexStore(), reserveNamespace, chunk)
+			if err != nil {
+				return err
+			}
+
+			err = s.IndexStore().Put(&BatchRadiusItem{
+				Bin:       bin,
+				BinID:     oldBatchRadiusItem.BinID,
```
istae, line 183: a new binID must be generated (so that the new chunks with a new stamp can be synced with neighbors)

pkg/storer/internal/reserve/reserve.go
```diff
@@ -122,6 +122,93 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
+				BatchID:   chunk.Stamp().BatchID(),
+				StampHash: stampHash,
+			})
+			if err != nil {
+				return err
+			}
+
+			err = s.IndexStore().Put(&ChunkBinItem{
+				Bin:       bin,
+				BinID:     oldBatchRadiusItem.BinID,
```
istae, line 194: same issue with binID

pkg/storer/internal/reserve/reserve.go
```diff
@@ -122,6 +122,93 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
+				s.IndexStore().Delete(oldBatchRadiusItem),
+				s.IndexStore().Delete(&ChunkBinItem{Bin: bin, BinID: oldBatchRadiusItem.BinID}),
+				stampindex.Delete(s.IndexStore(), reserveNamespace, swarm.NewChunk(chunk.Address(), nil).WithStamp(sameAddressOldChunkStamp)),
+				chunkstamp.DeleteWithStamp(s.IndexStore(), reserveNamespace, chunk.Address(), sameAddressOldChunkStamp),
+			)
+			if err != nil {
+				return err
+			}
+
+			err = stampindex.Store(s.IndexStore(), reserveNamespace, chunk)
```
istae, line 171: lines 171-202, and lines 249-274, can be a single function

pkg/storer/internal/reserve/reserve.go
```diff
@@ -122,6 +122,93 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
+			sameAddressOldStampIndex, err := stampindex.LoadWithStamp(s.IndexStore(), reserveNamespace, sameAddressOldChunkStamp)
+			if err != nil && !errors.Is(err, storage.ErrNotFound) {
+				return err
+			}
+			prev := binary.BigEndian.Uint64(sameAddressOldStampIndex.StampTimestamp)
+			curr := binary.BigEndian.Uint64(chunk.Stamp().Timestamp())
+			if prev >= curr {
+				return fmt.Errorf("overwrite prev %d cur %d batch %s: %w", prev, curr, hex.EncodeToString(chunk.Stamp().BatchID()), storage.ErrOverwriteNewerChunk)
+			}
+			if loadedStamp {
```
istae, line 142: if this is true, you must also remove the old chunk if the loadedStamp's address is not the same as the new chunk.
```suggestion
			if loadedStamp {
			  if !chunk.Address().Equal(oldItem.ChunkAddress) {
			    r.removeChunk(ctx, s, oldItem.ChunkAddress, oldItem.BatchID, oldItem.StampHash)
		          } 
```

istae (changes requested):

pkg/storer/internal/reserve/reserve.go
```diff
@@ -122,6 +122,93 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
+
+		sameAddressOldChunkStamp, err := chunkstamp.LoadWithBatchID(s.IndexStore(), reserveNamespace, chunk.Address(), chunk.Stamp().BatchID())
+		if err != nil && !errors.Is(err, storage.ErrNotFound) {
+			return err
+		}
+
+		// same address
+		if sameAddressOldChunkStamp != nil {
+			sameAddressOldStampIndex, err := stampindex.LoadWithStamp(s.IndexStore(), reserveNamespace, sameAddressOldChunkStamp)
+			if err != nil && !errors.Is(err, storage.ErrNotFound) {
```
istae, line 134: if the Load returns any error, lines 141 will panic because `sameAddressOldStampIndex` will be nil

pkg/storer/internal/reserve/reserve.go
```diff
@@ -122,6 +122,93 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
+				return err
+			}
+			prev := binary.BigEndian.Uint64(sameAddressOldStampIndex.StampTimestamp)
+			curr := binary.BigEndian.Uint64(chunk.Stamp().Timestamp())
+			if prev >= curr {
+				return fmt.Errorf("overwrite prev %d cur %d batch %s: %w", prev, curr, hex.EncodeToString(chunk.Stamp().BatchID()), storage.ErrOverwriteNewerChunk)
```
istae, line 140: this error msg and the one on line 145 can be improved by indicating with check, eg same/different chunk address

nugaon (commented):
hmm, I was not aware of these kind of different stores used for storing chunk, but I have some questions/comments regarding that below.

pkg/storer/internal/reserve/reserve.go
istae: the new chunk may at the same time overwrite an older chunk with the same address AND an older chunk with a different address but same stamp index :)

pkg/storer/internal/reserve/reserve.go
```diff
@@ -122,6 +122,93 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
 		}
+
+		sameAddressOldChunkStamp, err := chunkstamp.LoadWithBatchID(s.IndexStore(), reserveNamespace, chunk.Address(), chunk.Stamp().BatchID())
+		if err != nil && !errors.Is(err, storage.ErrNotFound) {
+			return err
+		}
+
+		// same address
+		if sameAddressOldChunkStamp != nil {
+			sameAddressOldStampIndex, err := stampindex.LoadWithStamp(s.IndexStore(), reserveNamespace, sameAddressOldChunkStamp)
```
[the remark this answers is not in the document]
acha-bill: `sameAddressOldStampIndex` was loaded so that we can read the `hash` immediately instead of hashing the stamp again.

pkg/storer/internal/reserve/reserve.go
```diff
@@ -122,6 +122,93 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
+			}
+
+			// load item to get the binID
+			oldBatchRadiusItem := &BatchRadiusItem{
+				Bin:       bin,
+				Address:   chunk.Address(),
+				BatchID:   sameAddressOldStampIndex.BatchID,
+				StampHash: sameAddressOldStampIndex.StampHash,
+			}
+			err = s.IndexStore().Get(oldBatchRadiusItem)
```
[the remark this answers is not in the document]
acha-bill: It's a different item type. We can't load different kinds of items at once.

istae (changes requested):
nice!
one final change request https://github.com/ethersphere/bee/pull/4749/files#diff-e37eed2ab5dfe1533be0257473f5b76738361f02d2be71da1b7bf4d5216a00c7R294: the size change here now also has to consider the same chunk address check.

should be:
```
if !loadedStamp && sameAddressOldChunkStamp == nil {
   r.size.Add(1)
}

**Review by istae — APPROVED** [review]

File: pkg/storer/internal/reserve/reserve.go (hunk 12 above)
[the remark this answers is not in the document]
Comment (nugaon) [reply pkg/storer/internal/reserve/reserve.go:133]: hm, I've a the feeling the hashing would be faster than a database read.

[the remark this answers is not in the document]
Comment (nugaon) [reply pkg/storer/internal/reserve/reserve.go:133]: and is it not the same as `oldItem`?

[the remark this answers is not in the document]
Comment (nugaon) [reply pkg/storer/internal/reserve/reserve.go:133]: `sameAddressOldChunkStamp` does not ensure that the chunk retrieved has also the same stamp (not only batchId), or does that? (I don't see clear thing regarding that in the `LoadWithBatchID`'s query)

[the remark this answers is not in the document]
Comment (acha-bill) [reply pkg/storer/internal/reserve/reserve.go:133]: > hm, I've a the feeling the hashing would be faster than a database read.

Maybe. However, reading the stored hash is better than computing it again because reading is resistant to changes in the hash function.

> sameAddressOldChunkStamp does not ensure that the chunk retrieved has also the same stamp (not only batchId), or does that? (I don't see clear thing regarding that in the LoadWithBatchID's query)

I'm not sure what you mean but `LoadWithBatchID` just finds the `stamp` associated with the chunk address and batchID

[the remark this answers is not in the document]
Comment (nugaon) [reply pkg/storer/internal/reserve/reserve.go:133]: > However, reading the stored hash is better than computing it again because reading is resistant to changes in the hash function.

If something changes in the hash function all db records regarding that should be updated anyway, shouldn't it?

> I'm not sure what you mean but `LoadWithBatchID` just finds the `stamp` associated with the chunk address and batchID

I mean that the chunk which stored and you want to override is retrieved by only the chunk address and batchID but this chunk should be replaced only if their stamp indices are the same as well (since if they are on different index they can be seem as legit payed stores).
My argue here is `sameAddressOldChunkStamp` returns a stamp which may not satisfy the stamp index match with the one that the inflight chunk has <-> it returns a stored chunk with a legit stamp and it wants to replace with the new chunk with a different/unrelated legit stamp.

It does not cover the situation when one uploads different payloads with the same batchID on the same SOC address used up multiple valid slots.

Before the same address condition there should be a search for retrieving all chunks with same chunk address and batch id and continue the replacement logic if stamp indices match.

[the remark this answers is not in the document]
Comment (acha-bill) [reply pkg/storer/internal/reserve/reserve.go:133]: > It does not cover the situation when one uploads different payloads with the same batchID on the same SOC address used up multiple valid slots.

We don't store multiple payloads for the same chunk address. This PR is to override the old one with the new one.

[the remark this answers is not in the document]
Comment (nugaon) [reply pkg/storer/internal/reserve/reserve.go:133]: but then why is the same batchId condition at chunk retrieval on line 126?

File: pkg/storer/internal/reserve/reserve.go [hunk 14]
```diff
@@ -199,6 +293,15 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 	})
 }
 
+func (r *Reserve) deleteWithStamp(s transaction.Store, oldBatchRadiusItem *BatchRadiusItem, sameAddressOldChunkStamp swarm.Stamp) error {
+	return errors.Join(
+		s.IndexStore().Delete(oldBatchRadiusItem),
+		s.IndexStore().Delete(&ChunkBinItem{Bin: oldBatchRadiusItem.Bin, BinID: oldBatchRadiusItem.BinID}),
```
istae, line 299: where is this `BinID` coming from? The item you are passing does not have this field.
please improve upon the unit tests to catch this bug.

istae (changes requested):

istae (approved):

PR #4791, opened 2024-08-29
Fix overwrite same chunk only with same batchid

acha-bill:
### Checklist

### Description
Overwrite same chunk only with same batchID

### DIFF
--- a/pkg/storer/internal/reserve/reserve_test.go
+++ b/pkg/storer/internal/reserve/reserve_test.go
@@ -155,13 +155,10 @@ func TestSameChunkAddress(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	privKey, err := crypto.GenerateSecp256k1Key()
-	if err != nil {
-		t.Fatal(err)
-	}
-	signer := crypto.NewDefaultSigner(privKey)
+	binBinIDs := make(map[uint8]uint64)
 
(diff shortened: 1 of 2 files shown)


pkg/storer/internal/reserve/reserve_test.go
```diff
@@ -271,6 +285,53 @@ func TestSameChunkAddress(t *testing.T) {
...
+	t.Run("chunk with different batchID remains untouched", func(t *testing.T) {
+		signer := getSigner(t)
+		batch := postagetesting.MustNewBatch()
+		s1 := soctesting.GenerateMockSocWithSigner(t, []byte("data"), signer)
+		ch1 := s1.Chunk().WithStamp(postagetesting.MustNewFields(batch.ID, 0, 3))
+		batch = postagetesting.MustNewBatch()
+		s2 := soctesting.GenerateMockSocWithSigner(t, []byte("update"), signer)
+		ch2 := s2.Chunk().WithStamp(postagetesting.MustNewFields(batch.ID, 0, 4))
+
+		err = r.Put(ctx, ch1)
```
istae, line 298: i would add a check here that the two chunks have the same address
also we should have the same unit test for content addressed chunks as well

acha-bill: updated

istae (approved):

martinconic (approved):

PR #4798, opened 2024-09-02
fix: reduce reserve cnt on same chunk and index collision

acha-bill:
### Checklist

### Description
Reduce the reserve count when putting the same address chunk that has an index collision with a different chunk


pkg/storer/internal/reserve/reserve.go
```diff
@@ -165,6 +165,7 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
 					if err != nil {
 						return fmt.Errorf("failed removing older chunk %s: %w", oldStampIndex.ChunkAddress, err)
 					}
+					r.size.Add(-1)
```
istae, line 168: this cannot be here, it's has to go outside the transaction.
the `r.size.Add(1)` is wrongly placed in the transaction. pls fix both.

istae (changes requested):

istae (approved):

martinconic (approved):

PR #4810, opened 2024-09-10
Add missing symbol in .golangci.yml

damuzhi0810:
### Checklist

### Description

Add missing symbol in .golangci.yml

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -62,7 +62,7 @@ issues:
     - linters:
         - forbidigo
       path: cmd/bee/cmd
-      text: "use of `fmt.Print`"   ## allow fmt.Print in cmd directory
+      text: "use of fmt.Print"   ## allow fmt.Print in cmd directory
     - linters:
         - dogsled
       path: pkg/api/(.+)_test\.go # temporally disable dogsled in api test files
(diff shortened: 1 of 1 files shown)


martinconic (approved):

martinconic:
Please use for commits the pattern from [here](https://github.com/ethersphere/bee/blob/master/commitlint.config.js). And you can add back the backticks

gacevicljubisa:
If we use changes made on [commit](https://github.com/ethersphere/bee/pull/4810/commits/f6ad42a4e86bc5d75d417ce48ff3457914b59d5c), it will only exclude `fmt.Print` from specified package path, then linter would report error and will not allow cases like this:
```
fmt.Printf("starting import process with data-dir at %s\n", dataDir)
```
or
```
fmt.Println(beeASCII)
```
This PR should be closed. Can you confirm this @istae ?

gacevicljubisa:
The current version of `golangci-lint` being used is v1.61.0. The idea behind the `cmd` package is to allow the entire family of `fmt.Print` functions. The linter currently uses an error expression with backticks as specified in the master branch. If we change this to "use of `fmt.Print`," it will trigger errors for every occurrence of `fmt.Print`, which may not be the desired behavior.

Issue #4823, opened 2024-09-13
Invalid (obsolete) comment in pullsync


ldeffenb:
### Location
https://github.com/ethersphere/bee/blob/06a0aca71fb3c941f2ec85a6cd4937c4e62b30de/pkg/pullsync/pullsync.go#L53-L54

### Description
Code says `handleMaxChunksPerSecond` which is 250, but the comment says 100.

ldeffenb:
I know it's not the formal documentation, but it's not exactly a "bug" either.

PR #4828, opened 2024-09-17
fix: adapt comment to corresponding used value

martinconic:
### Checklist

### Description
Fixes [4823](https://github.com/ethersphere/bee/issues/4823)

### DIFF
--- a/pkg/pullsync/pullsync.go
+++ b/pkg/pullsync/pullsync.go
@@ -51,7 +51,7 @@ const (
 	pageTimeout                     = time.Second
 	makeOfferTimeout                = 15 * time.Minute
 	handleMaxChunksPerSecond        = 250
-	handleRequestsLimitRate         = time.Second / handleMaxChunksPerSecond // handle max 100 chunks per second per peer
+	handleRequestsLimitRate         = time.Second / handleMaxChunksPerSecond // handle max 250 chunks per second per peer
 )
 
 // Interface is the PullSync interface.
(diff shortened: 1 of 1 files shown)


janos (approved):

pkg/pullsync/pullsync.go
```diff
@@ -51,7 +51,7 @@ const (
 	pageTimeout                     = time.Second
 	makeOfferTimeout                = 15 * time.Minute
 	handleMaxChunksPerSecond        = 250
-	handleRequestsLimitRate         = time.Second / handleMaxChunksPerSecond // handle max 100 chunks per second per peer
+	handleRequestsLimitRate         = time.Second / handleMaxChunksPerSecond // handle max 250 chunks per second per peer
```
ldeffenb, line 54: Personally, I would have avoided redundancy with something like:
`handle MaxChunksPerSecond per second per peer`

ldeffenb (changes requested):
Why mention the 250 twice?

pkg/pullsync/pullsync.go
martinconic: good point

acha-bill (approved):

pkg/pullsync/pullsync.go
```diff
@@ -51,7 +51,7 @@ const (
 	pageTimeout                     = time.Second
 	makeOfferTimeout                = 15 * time.Minute
 	handleMaxChunksPerSecond        = 250
-	handleRequestsLimitRate         = time.Second / handleMaxChunksPerSecond // handle max 100 chunks per second per peer
+	handleRequestsLimitRate         = time.Second / handleMaxChunksPerSecond // handle max `handleMaxChunksPerSecond` chunks per second per peer
```
ldeffenb, line 54: I like that much better as it clarifies that the rate is per second per peer (the latter being important) while not being redundant about the actual value.

ldeffenb (commented):
Let 'er rip!

PR #4839, opened 2024-09-28
fix: Eliminate local transaction nonce cache.

attila-lendvai:
**WARNING:** i'm not experienced in golang, and i don't know the bee internals either. this commit requires review by someone who understands the codebase.

### Checklist

### Description
Stop keeping the next nonce in our local state. Instead we always calculate it from the blockchain nonce and the local list of pending transactions.

Fixes issue [#4826](https://github.com/ethersphere/bee/issues/4826) where the local nonce cache was out of sync with the blockchain for an unknown reason. Because of that no transactions got incorporated into the blockchain. This made it impossible to unstake or to participation in the reserve incentives, probably forever (until fixed).

It only became apparent when i tried to unstake my nodes at the v2.2.0 upgrade, and some of them got stuck. It also explains why i stopped seeing rewards on my nodes.

With this commit on top of v2.1.0 i've successfully unstaked a previously hung node.

And an eternal guiding principle: do not optimize any program until it works as specified. And the only exception is when it's so slow that it substantially hinders development.

#### A potential unaddressed issue
If a third party creates a transaction, then the bee's transactions may get stuck. There should be a check that the blockchain nonce is lower than the lowest nonce in our pending tx list. If not, then the nonce of the pending tx'es should be updated.

### DIFF
--- a/pkg/transaction/transaction.go
+++ b/pkg/transaction/transaction.go
@@ -193,11 +193,6 @@ func (t *transactionService) Send(ctx context.Context, request *TxRequest, boost
 		return common.Hash{}, err
 	}
 
-	err = t.putNonce(nonce + 1)
-	if err != nil {
-		return common.Hash{}, err
-	}
-
 	txHash = signedTx.Hash()
 
 	err = t.store.Put(storedTransactionKey(txHash), StoredTransaction{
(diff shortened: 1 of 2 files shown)


janos:
Hi Attila, thanks. This is a nice improvement, but it may need even more changes. The transaction monitor main loop may be problematic and blocking while it does not need to be. This has to be debugged more, but certainly this change is a good step.

TestTransactionSend needs to be adjusted as it fails currently.

martinconic (approved):

janos (approved):
👍

Issue #4842, opened 2024-10-01
reserve size doubling tasks


istae:
[Swip-21](https://github.com/ethersphere/SWIPs/pull/56/files) can actually be tackled by a relatively small number of changes.

- [x] 1) user provides a whole-number radius delta, _d_ through config options (default being 0)
- [x] 2) reserve capacity is now a dynamic value and each increase in _d_ doubles the reserve capacity, eg 2^(22+_d_). When the new capacity is calculated initially, the reserve worker will automatically reduce the current storage radius until the reserve is full.
- [x] 3) in the `ReserveSampler` iterator, the chunks that are considered need to be filtered by checking that their proximity relative to the anchor is greater than equal to the current storage radius +  _d_.
- [x] 4) salud package should no longer punish peers for having "wrong radiuses" relative to the network.
- [x] 5) consider what do to with the db cmds that rely on the reserve size.
- [x] 6) reserve should return the stat (neighborhood adress and size) of each neighborhood.
- [x] 7) new API for the stats above
- [x] 8) ABI changes. If node has been staked before and if there is a delta change then new bzz has to be staked. The node should err out if there is not enough bzz.
- [x] 9) commited depth changes in the redis contract and the agent

zelig:
> 3\. in the `ReserveSampler` iterator, the chunks that are considered need to be filtered by checking that their proximity relative to the anchor is greater than equal to the current storage radius +  _d_.

should be ` - _h_` where _h_ stands for height, ie. the number of doublings

zelig:
In the light of being cautions about supporting anything more than double (due to problems of resource overload in case of height difference in a neighbourhood as well as sampling issues due to lack of proper indexing beyond height 1).
So I recommend 
1. the contract API as well as beewconfig to allow only the operation of 'increase height by one' as a permanent feature and 
2. restrict height to 1 as a temporary feature in bee only.
3. full support for any height in both bee and the contract in the operational logic to keep implementation simple and lasting

ldeffenb:
> 1. the contract API as well as beewconfig to allow only the operation of 'increase height by one' as a permanent feature and

Will a node be allowed to DECREASE the height?   Like go from doubled back down to a single neighborhood?

zelig:
> > 1. the contract API as well as beewconfig to allow only the operation of 'increase height by one' as a permanent feature and
> 
> Will a node be allowed to DECREASE the height? Like go from doubled back down to a single neighborhood?

YES

[...]

PR #4853, opened 2024-10-08
feat: add status neighborhoods endpoint

martinconic:
### Checklist

### Description
Part 7 of [#4842](https://github.com/ethersphere/bee/issues/4842)

### DIFF
--- a/openapi/SwarmCommon.yaml
+++ b/openapi/SwarmCommon.yaml
@@ -936,10 +936,17 @@ components:
     StatusNeighborhoodResponse:
       type: object
       properties:
-        address:
-          type: string
+        neighborhood:
+          $ref: "#/components/schemas/Neighborhood"
         reserveSizeWithinRadius:
           type: integer
+        proximity:
+          type: integer
+
+    Neighborhood:
+      type: string
+      description: Swarm address of a neighborhood in string binary format, usually limited to as many bits as the current storage radius.
+      example: "011010111"
 
     StatusNeighborhoodsResponse:
       type: object
--- a/openapi/SwarmCommon.yaml
+++ b/openapi/SwarmCommon.yaml
@@ -933,6 +933,23 @@ components:
           items:
             $ref: "#/components/schemas/StatusSnapshotResponse"
 
+    StatusNeighborhoodResponse:
+      type: object
+      properties:
+        address:
+          type: string
+        chunkCount:
+          type: integer
+
+    StatusNeighborhoodsResponse:
+      type: object
+      properties:
+        stamps:
+          type: array
+          nullable: false
+          items:
+            $ref: "#/components/schemas/StatusNeighborhoodResponse"
+
     ApiChunkInclusionProof:
       type: object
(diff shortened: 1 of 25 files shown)


openapi/SwarmCommon.yaml
```diff
@@ -933,6 +933,23 @@ components:
           items:
             $ref: "#/components/schemas/StatusSnapshotResponse"
 
+    StatusNeighborhoodResponse:
+      type: object
+      properties:
+        address:
+          type: string
+        chunkCount:
```
ldeffenb, line 941: Is this `reserveSize` or `reserveSizeWithinRadius` or something completely different?

istae: This field should be called `reserveSizeWithinRadius`. Good call.

martinconic: renamed

openapi/SwarmCommon.yaml
```diff
@@ -933,6 +933,23 @@ components:
           items:
             $ref: "#/components/schemas/StatusSnapshotResponse"
 
+    StatusNeighborhoodResponse:
+      type: object
+      properties:
+        address:
+          type: string
```
janos, line 940: Maybe to be more specific as it is a string of a number in a binary representation?

pkg/storer/reserve.go
```diff
@@ -498,8 +498,8 @@ func (db *DB) SubscribeBin(ctx context.Context, bin uint8, start uint64) (<-chan
 }
 
 type NeighborhoodStat struct {
-	Address    swarm.Address
-	ChunkCount int
+	Address                 swarm.Address
```
janos, line 501: This field suggests that it is a swarm address, but it looks to me that it represents a neighborhood just using the swarm.Address type. I am questioning if both the field name and its type are correct. Maybe the field name can stay the same if the type would be of the specific neighborhood type. It looks to me that neighborhood would benefit from its own type, for example swarm.Neighborhood instead to use swarm.Address as a container that may mislead the code reader. There are technical similarities between two types but the semantics of them is different. Another option, field name could be Neighborhood for clarity.

pkg/api/status.go
```diff
@@ -159,3 +169,42 @@ func (s *Service) statusGetPeersHandler(w http.ResponseWriter, r *http.Request)
...
+
+	if len(nhoods) == 0 {
+		jsonhttp.NotFound(w, "neighborhoods not found")
+		return
+	}
+
+	for _, n := range nhoods {
+		binaryAddr := ""
+		for _, b := range n.Address.Bytes() {
+			binaryAddr += fmt.Sprintf("%08b ", b)
```
janos, line 201: I think that the binary representation could a method of a potentially new type swarm.Neighborhood. Just a suggestion.

martinconic: Searched for such a thing but could not find it, `bitStr(` not the best naming

istae (approved):

PR #4856, opened 2024-10-10
feat(staking): change height on restart

istae:
### Checklist

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -1032,15 +1032,16 @@ func NewBee(
 				logger.Warning("staked amount does not sufficiently cover the additional reserve capacity. Stake should be at least 2^h * 10 BZZ, where h is the number of doublings.")
 			}
 
-			tx, changed, err := stakingContract.ChangeHeight(ctx, uint8(o.ReserveCapacityDoubling))
-			if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/node/node.go
```diff
@@ -1004,18 +1004,44 @@ func NewBee(
...
+		}
+		if updated {
+			logger.Info("updated new reserve capacity doubling height in the staking contract", "transaction", tx, "new_height", o.ReserveCapacityDoubling)
 		}
-		if stake.Cmp(big.NewInt(0)) > 0 {
-			logger.Debug("changing overlay address in staking contract")
-			tx, err := stakingContract.ChangeStakeOverlay(ctx, common.BytesToHash(nonce))
+
+		if o.ReserveCapacityDoubling > 0 {
+			stake, err := stakingContract.GetPotentialStake(ctx)
```
martinconic, line 1034: Can we call GetPotentialStake only once , before the two if statements?

martinconic (approved):

pkg/node/node.go
istae: In a real life scenario, it's very unlikely that we will call get stake twice here

Issue #4838, opened 2024-09-27
convert `/rchash` response `duration` to seconds


rampall:
convert `/rchash` response `duration` to seconds to maintain consistency with `lastSampleDurationSeconds` from `/redistributionstate`

PR #4862, opened 2024-10-11
fix: show rchash duration response in seconds

martinconic:
### Checklist

### Description
Fixes [#4838](https://github.com/ethersphere/bee/issues/4838)

### DIFF
--- a/openapi/SwarmCommon.yaml
+++ b/openapi/SwarmCommon.yaml
@@ -1,6 +1,6 @@
 openapi: 3.0.3
 info:
-  version: 4.1.0
+  version: 4.2.0
   title: Common Data Types
   description: |
     \*****bzzz*****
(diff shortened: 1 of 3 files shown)


openapi/SwarmCommon.yaml
```diff
@@ -1016,7 +1016,7 @@ components:
       type: object
       properties:
         duration:
-          type: integer
```
janos, line 1019: This is a some sort of a breaking change as previously integer parsing will fail on floating point number.

I am not sure how significant this is.

janos (approved):

openapi/SwarmCommon.yaml
martinconic: since `.Seconds()` returns float64 it was needed to have a `number` here instead of `integer`

istae: maybe it's appropriate here to define a new "seconds" type with a nice description and example

martinconic: seconds type added

istae (approved):

PR #4867, opened 2024-10-15
fix: check all transactions in monitor check pending with timeout

martinconic:
### Checklist

### Description

Fixes [#4826](https://github.com/ethersphere/bee/issues/4826).
This is part two of the solution, provided by @janos . The issue investigation and solution are described in the above issue https://github.com/ethersphere/bee/issues/4826#issuecomment-2391045378.

### DIFF
--- a/pkg/transaction/monitor.go
+++ b/pkg/transaction/monitor.go
@@ -178,17 +178,17 @@ func (tm *transactionMonitor) hasWatches() bool {
 }
 
 func watchStart(watches []transactionWatch) time.Time {
-	var start time.Time
-	for _, w := range watches {
+	if len(watches) == 0 {
+		return time.Time{}
+	}
+	start := watches[0].start
+	for _, w := range watches[1:] {
 		if w.start.Before(start) {
 			start = w.start
 		}
 	}
 	return start
 }
-
-// Time for transaction is checked if it is accepted for mining.
-const transactionWatchNotFoundTimeout = 20 * time.Second
 
 // check pending checks the given block (number) for confirmed or cancelled transactions
 func (tm *transactionMonitor) checkPending(block uint64) error {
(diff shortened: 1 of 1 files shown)


pkg/transaction/monitor.go
```diff
@@ -169,44 +171,34 @@ func (tm *transactionMonitor) watchPending() {
...
+	return start
 }
 
-func (tm *transactionMonitor) hasWatches() bool {
-	tm.lock.Lock()
-	defer tm.lock.Unlock()
-	return len(tm.watchesByNonce) > 0
-}
+// Time for transaction is checked if it is accepted for mining.
+const transactionWatchNotFoundTimeout = 20 * time.Second
```
istae, line 191: how is this defined at 20 sec?
maybe this should be a multiple of the block time of the blockchain?

pkg/transaction/monitor.go
```diff
@@ -169,44 +171,34 @@ func (tm *transactionMonitor) watchPending() {
...
 
-	potentiallyConfirmedTxWatches := make(map[uint64]map[common.Hash][]transactionWatch)
-	for n, watches := range tm.watchesByNonce {
-		if n < nonce {
-			potentiallyConfirmedTxWatches[n] = watches
+func watchStart(watches []transactionWatch) time.Time {
+	var start time.Time
+	for _, w := range watches {
+		if w.start.Before(start) {
+			start = w.start
```
istae, line 184: is it possible to have a `w.start` that is 0? because when we append these watches, the start is initialized with `time.Now()`

pkg/transaction/monitor.go
```diff
@@ -169,44 +171,34 @@ func (tm *transactionMonitor) watchPending() {
...
 	defer tm.lock.Unlock()
+	return len(tm.watchesByNonce) > 0
+}
 
-	potentiallyConfirmedTxWatches := make(map[uint64]map[common.Hash][]transactionWatch)
-	for n, watches := range tm.watchesByNonce {
-		if n < nonce {
-			potentiallyConfirmedTxWatches[n] = watches
+func watchStart(watches []transactionWatch) time.Time {
+	var start time.Time
```
acha-bill, line 181: Should be initialized to the max time

acha-bill (changes requested):

pkg/transaction/monitor.go
janos: In the scope of a function, zero value is possible. In the scope of the package, currently it is not.

pkg/transaction/monitor.go
janos: This is a completely rough estimate. A block time multiple is a very good approach.

pkg/transaction/monitor.go
janos: Yes, that is right. Good point.

gacevicljubisa (approved):

acha-bill (approved):

istae (approved):

Issue #4734, opened 2024-07-24
Add notice that API is not accessible during initialization


NoahMaizels:
#### Summary

Currently when a new node is initialized, while syncing batch data, the Bee API is not accessible. However, there is no message returned from the API stating that the node is syncing batch data, it simply appears that the API endpoint is not working properly. It would be good if a descriptive message is returned instead that states the node is currently syncing batch data, so that it's clear there is no malfunction. The message could be something like "Currently syncing batch data, API is not available while syncing. Try again in a few minutes."

#### Motivation

Make the user experience easier.

nugaon:
would be better to give back another response for requests than 404. it is more like a 503.

NoahMaizels:
> would be better to give back another response for requests than 404. it is more like a 503.

I think that whatever the response is, it needs to indicate whether the API is not available since the node is in the process of starting up, or whether it is not available due to some other reason / error

PR #4869, opened 2024-10-19
fix(api): add api initialization notice

gacevicljubisa:
### Checklist

### Description

This PR addresses issue #4734 by improving the behavior of the Bee API during the node initialization process. Previously, when the node was syncing batch data, the API appeared to be inaccessible, providing no indication of the ongoing sync process. Now, all routes are mounted immediately, but some may be temporarily disabled during initialization, returning appropriate status codes.

### Changes

- **503 Service Unavailable**: Routes that are disabled during the syncing phase now return a `503 Service Unavailable` status with the message:  
  *"Node is syncing. This endpoint is unavailable. Try again later."*  
  Once the syncing is complete, the routes will be enabled and respond normally.

- **501 Not Implemented**: If the `swapEnabled` or `chequebookEnabled` flags are set to `false`, routes that depend on these features will now return a `501 Not Implemented` status.

- **404 Not Found**: Any routes that do not exist will return a `404 Not Found` status.

This change improves the user experience by providing clear feedback during node initialization, preventing confusion when the API appears to be inaccessible. It ensures that users are informed about the node's state and can understand why certain API routes may not be available temporarily.

### How it works

- All routes are mounted immediately but may not be active depending on the node's state.
- Disabled routes return `503` until they are enabled once the sync is complete.
- Feature-dependent routes (like those for Swap and Chequebook) return `501` if the corresponding feature is disabled.
- Invalid or non-existent routes return a `404` as expected.

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/4734

### DIFF
--- a/pkg/api/router.go
+++ b/pkg/api/router.go
@@ -528,9 +528,11 @@ func (s *Service) mountBusinessDebug() {
 	handle("/wallet/withdraw/{coin}", web.ChainHandlers(
 		s.checkChequebookAvailability,
 		s.checkSwapAvailability,
-		s.gasConfigMiddleware("wallet withdraw"),
 		web.FinalHandler(jsonhttp.MethodHandler{
-			"GET": http.HandlerFunc(s.walletWithdrawHandler),
+			"POST": web.ChainHandlers(
+				s.gasConfigMiddleware("wallet withdraw"),
+				web.FinalHandlerFunc(s.walletWithdrawHandler),
+			),
 		}),
 	))
 
--- a/pkg/api/router.go
+++ b/pkg/api/router.go
@@ -25,9 +25,7 @@ const (
 	rootPath   = "/" + apiVersion
 )
 
-type MountOption func(*Service)
-
-func (s *Service) Mount(opts ...MountOption) {
+func (s *Service) Mount() {
 	router := mux.NewRouter()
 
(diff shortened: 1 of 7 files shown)


pkg/api/router.go
```diff
@@ -51,15 +45,26 @@ func (s *Service) MountDebug() {
...
+		o(s)
+	}
 }
 
-func (s *Service) MountAPI() {
-	if s.router == nil {
-		s.router = mux.NewRouter()
-		s.router.NotFoundHandler = http.HandlerFunc(jsonhttp.NotFoundHandler)
+// WithFullAPI will enable all available endpoints, because some endpoints are not available during syncing.
+func WithFullAPI() MountOption {
```
istae, line 55: why have an option if we can explicitly enable the the full API using `EnableFullAPI()`

gacevicljubisa: Initial idea to provide options is for unit testing purposes. Was considering to add `WithFullAPIDisabled` option as well. And then expand `testServerOptions` in api_test.go. 
Also it can be used in one line to mount and enable `apiService.Mount(api.WithFullAPI())`, which is the case in devnode.go

istae (approved):
Great work!

janos (approved):
👏

martinconic (approved):

PR #4874, opened 2024-10-22
fix(reserve): trigger eviction in the worker if the size is above cap…

istae:
…acity

### Checklist


gacevicljubisa (approved):

istae:
> it turns out that we check if a node is frozen by comparing the lastUpdatedBlockNumberOfAddress with the most recent block height [here ](https://github.com/ethersphere/bee/blob/master/pkg/storageincentives/staking/contract.go#L243) 
which gets called by the agent [here](https://github.com/ethersphere/bee/blob/master/pkg/storageincentives/agent.go#L228) 
as soon as the node start up, it updates the height in the staking contract which updates the lastUpdatedBlockNumberOfAddress with the current block, next this agent gets initialized, and since we are at the same block height, the IsOverlayFrozen check returns true bumping the block in [this ](https://github.com/ethersphere/bee/blob/master/pkg/storageincentives/agent.go#L228) call by one should fix this issue

martinconic (approved):

PR #4893, opened 2024-11-04
chore: add docker-build target to Makefile with support for configurable build flags

gacevicljubisa:
### Checklist

### Description
This PR introduces a new target, docker-build, in the Makefile, allowing for configurable build flags when creating the Docker image. Specifically, users can now set the REACHABILITY_OVERRIDE_PUBLIC and IMAGE flags, along with other flags as needed, when building the image.

Example of usage:
```
make docker-build REACHABILITY_OVERRIDE_PUBLIC=true IMAGE=ethersphere/bee:latest
```


gacevicljubisa:
Way to use build flag when creating docker image:
```
docker build --build-arg REACHABILITY_OVERRIDE_PUBLIC=true -t bee-image-name -f Dockerfile .
```

acha-bill:
If we needed to support other flags defined in the makefile here as well, we would end up maintaining variables in 2 places.
Would it work if we exposed the variable in the makefile to the environment so that the value can be passed to make?

makefile
```
export REACHABILITY_OVERRIDE_PUBLIC ?= false
```

Then use it with the build arg as you suggested

gacevicljubisa:
> If we needed to support other flags defined in the makefile here as well, we would end up maintaining variables in 2 places. Would it work if we exposed the variable in the makefile to the environment so that the value can be passed to make?
> 
> makefile
> 
> ```
> export REACHABILITY_OVERRIDE_PUBLIC ?= false
> ```
> 
> Then use it with the build arg as you suggested

we could add this to Makefile:

```
# Load .env if it exists
ifneq (,$(wildcard ./.env))
	include .env
	export
endif
```
And if we locally have .env file with the specific flag set to the value. Then no other changes or maintinance is needed. 

Another way would be that we have a new target in Makefile that will be responsibile for building the image itself:
```
docker-build:
	docker build -t $(IMAGE) . --no-cache
```
And this is example how it can be used:
```
make docker-build REACHABILITY_OVERRIDE_PUBLIC=true IMAGE=ethersphere/bee:latest
```

nugaon:
> Another way would be that we have a new target in Makefile that will be responsibile for building the image itself:

I would vote on this option.

having .env variable is error prone between builds.

acha-bill (approved):
lgtm. 
Please rename the PR title

gacevicljubisa:
> > Another way would be that we have a new target in Makefile that will be responsibile for building the image itself:
> 
> I would vote on this option.
> 
> having .env variable is error prone between builds.

@acha-bill @nugaon 
Proposed solution will not work without `.env` file. 

The only idea I have now is back to build arguments. 

@darkobas2 do you have any other better idea?

darkobas2:
As binding an app inside a docker container to 127.0.0.1 has very limited use or use cases i would suggest we always build docker with :1633 with the existing way to override the api port host

gacevicljubisa:
Default Values:

- REACHABILITY_OVERRIDE_PUBLIC: false (default)
- BATCHFACTOR_OVERRIDE_PUBLIC: 5 (default)
- IMAGE: ethersphere/bee:latest (default Docker image tag)

How to Use:

1. To build the Docker image using default values: Simply run:

    ```sh
    make docker-build
    ```

2. To override the default values:

    ```sh
    make docker-build REACHABILITY_OVERRIDE_PUBLIC=true BATCHFACTOR_OVERRIDE_PUBLIC=2 IMAGE=ethersphere/bee:v1.0
    ```

acha-bill (approved):
LGTM

istae (approved):

nugaon (approved):
LGTM!

Issue #3831, opened 2023-02-24
enable paralleltest lint check in `pkg/postage`


vladopajic:
Currently `pkg/postage` has disabled `paralleltest` lint check (see `golangci.yaml` file). This check should be enabled for this package and tests should be corrected to use `t.Parallel()`.

Acceptance criteria:
- enable `paralleltest` lint check in  `pkg/postage`
  - all tests should use `t.Parallel()`; if tests for some reason could not use `t.Parallel()` then it disabling lint check per tests case should be alright
- all tests should pass 
- there should be no lint errors

emekaokoli19:
@vladopajic From what I understand, the issue here is to add t.parallel() to the test functions in pkg/postage right?
Also, how do I enable paralleltest for pkg/postage? Do I make changes to the golangci.yml file?

vladopajic:
You are right on everything :)

PR #4913, opened 2024-11-25
fix: enable paralleltest lint check in package postage

martinconic:
### Checklist

### Description
Fixes [3831](https://github.com/ethersphere/bee/issues/3831)

### DIFF
--- a/pkg/postage/postagecontract/contract_test.go
+++ b/pkg/postage/postagecontract/contract_test.go
@@ -30,10 +30,6 @@ var postageStampContractABI = abiutil.MustParseABI(chaincfg.Testnet.PostageStamp
 func TestCreateBatch(t *testing.T) {
 	t.Parallel()
 
-	defer func(b uint8) {
-		postagecontract.BucketDepth = b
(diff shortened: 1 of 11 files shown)


pkg/postage/postagecontract/contract_test.go
```diff
@@ -28,10 +28,8 @@ import (
 var postageStampContractABI = abiutil.MustParseABI(chaincfg.Testnet.PostageStampABI)
 
 func TestCreateBatch(t *testing.T) {
-	defer func(b uint8) {
```
janos, line 31: By making this test run in paralel, other tests that use `postagecontract.BucketDepth` may have read an incorrect value (in this case 9) or a data race if they also want to set value of that global variable.

As this test changes `postagecontract.BucketDepth` it should not run in paralel. Also, it should put back the original value when it finishes, maybe even better to use t.Cleanup than defer and make subtests here parallel.

gacevicljubisa (approved):

istae (approved):

PR #4920, opened 2024-12-04
fix: save tags when using chunk/stream

### LINKED ISSUE #4858: Chunks uploaded with websocket are not counted into tag stats
### Context
Bee 2.2.0

### Summary
Uploading chunks with http requests increments the tag `split` value, as it is also incremented uploading with bzz protocol.
Instead uploading using websocket doesn't update it.
Because of this, I suspect that upload with websocket doesn't actual use the tag at all.

### Expected behavior
I should have the same behavior on tag, session and storage uploading with http or uploading with websocket.

### Actual behavior
It seams that uploading with websocket doesn't use tag's session.

### Steps to reproduce
Create a tag, and use it to upload with websocket.
After the upload, try to look at tag status. It has all values to 0, even if chunks have been uploaded.

acha-bill:
### Checklist

### Description

Do not use HTTP request context so that tags will be saved when `putter.Done` is called after the HTTP connection is upgraded to websocket.

### Related Issue (Optional)
closes #4858

### DIFF
--- a/pkg/api/chunk_stream.go
+++ b/pkg/api/chunk_stream.go
@@ -14,8 +14,8 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/jsonhttp"
 	"github.com/ethersphere/bee/v2/pkg/log"
 	"github.com/ethersphere/bee/v2/pkg/postage"
-	storage "github.com/ethersphere/bee/v2/pkg/storage"
-	storer "github.com/ethersphere/bee/v2/pkg/storer"
+	"github.com/ethersphere/bee/v2/pkg/storage"
+	"github.com/ethersphere/bee/v2/pkg/storer"
 	"github.com/ethersphere/bee/v2/pkg/swarm"
 	"github.com/gorilla/websocket"
 )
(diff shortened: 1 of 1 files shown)


pkg/api/chunk_stream.go
```diff
@@ -56,7 +56,7 @@ func (s *Service) chunkUploadStreamHandler(w http.ResponseWriter, r *http.Reques
 	}
 
 	// if tag not specified use direct upload
-	putter, err := s.newStamperPutter(r.Context(), putterOptions{
+	putter, err := s.newStamperPutter(context.Background(), putterOptions{
```
martinconic, line 59: why is context.Background() better here?

istae: probably because the go routine below handles the websocket connection but since this function returns, the context is canceled.

acha-bill: yep

martinconic (approved):

istae (approved):
pls add a comment here why this background context is necessary as it's not immediately obvious

PR #4955, opened 2025-01-23
fix: stewardship with erasure encoding

### LINKED ISSUE #4948: erasure coding stewardship endpoint returns false
for erasure encoded data during upload (we tried level 3), the stewardship endpoint returns isRetrievable=false.

The endpoint returns true, however, for the same upload without erasure coding.

acha-bill:
### Checklist

### Description

When iterating over chunks, parity chunks were incorrectly attempted to be fetched from the redundancy getter, which only supports shards.

### Related Issue (Optional)
closes #4948

### DIFF
--- a/pkg/file/joiner/joiner.go
+++ b/pkg/file/joiner/joiner.go
@@ -19,7 +19,7 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/file/redundancy"
 	"github.com/ethersphere/bee/v2/pkg/file/redundancy/getter"
 	"github.com/ethersphere/bee/v2/pkg/replicas"
-	storage "github.com/ethersphere/bee/v2/pkg/storage"
+	"github.com/ethersphere/bee/v2/pkg/storage"
 	"github.com/ethersphere/bee/v2/pkg/swarm"
 	"golang.org/x/sync/errgroup"
 )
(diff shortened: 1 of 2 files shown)


pkg/file/joiner/joiner.go
```diff
@@ -406,6 +406,12 @@ func (j *joiner) processChunkAddresses(ctx context.Context, fn swarm.AddressIter
 			if j.refLength == encryption.ReferenceSize && i < shardCnt {
 				addr = swarm.NewAddress(data[cursor : cursor+swarm.HashSize*2])
 			}
+
+			// not a shard
+			if i >= shardCnt {
```
istae, line 411: maybe it should be `i >shardCnt` 🤔 
you are also reading `i` from a goroutine, there could be some data race issues here. Pls also check if we are accessing variables here that we should not be without some synchronization mechanism.

istae (changes requested):

pkg/file/joiner/joiner.go
acha-bill: `i` is 0-indexed. So it should be `i >= shardCnt`
The gourtine that reads `i` has to return before the next iteration. See the waitgroup used below. So there can be no race here.

istae:
Merging this now, a fix for https://github.com/ethersphere/bee/issues/4977 will be included in a different PR

PR #4958, opened 2025-01-23
perf: various improvements in pusher, pushsync, salud, reacher

istae:
### Checklist

### Description
1. Minor clean up in reacher.go and increased go worker count
2. Light nodes error out if a reserve capacity doubling is set in the config
3. in the pusher, as an optimization, instead of a full stamp validation which is expense, we instead only check if the batch is still alive. Also increased max parallel push count from 100 to 128.
4. in the pushsync protocol, forward nodes no long perform receipt validation. Also, for the shallow receipts check, we apply some tolerance to the proximity of the store node and chunk based on the maximum allowed reserve doubling. Tag reporting was also being made for direct uploads, this is has been changed to only happen for deferred uploads by checking the chunk's tag ID.
5. In the salud service, we sort the peer list in desc order by protocol response duration so that the faster peers are given priority if a bin's population drops below the salud threshold.
6. increased max allowed cache worker count from 16 to 128.
7.  as an optimization, the kademlia closest peer terminates early if the iterator has already a closest peer and the current bin count is greater than the chunk's proximity with the node's address.


martinconic (approved):

pkg/pushsync/pushsync.go
```diff
@@ -285,8 +289,6 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 	case errors.Is(err, topology.ErrWantSelf):
 		stored, reason = true, "want self"
 		return store(ctx)
-	case errors.Is(err, ErrShallowReceipt):
-		fallthrough
```
gacevicljubisa, line 289: Now it will go to default? 
If this is ok, should we have a unit test for this case? both `err==nil` and `default` case are not covered by unit tests...

pkg/pusher/pusher.go
```diff
@@ -311,10 +307,10 @@ func (s *Service) pushDirect(ctx context.Context, logger log.Logger, op *Op) err
 		}
 	}()
 
-	_, err = s.validStamp(op.Chunk)
-	if err != nil {
-		logger.Warning(
-			"stamp with is no longer valid, skipping direct upload for chunk",
+	ok, err := s.batchExist.Exists(op.Chunk.Stamp().BatchID())
```
gacevicljubisa, line 310: If batch not found it will return nil? Is this also ok?

istae: yes, for direct uploads, the chunk resync is retried if this error is not nil. because batch is no longer valid at this point, we should stop syncing, hence the nil return

pkg/pushsync/pushsync.go
istae: forwarder nodes no longer perform receipt checks, this is why the shallow receipt error check is removed.
see https://github.com/ethersphere/bee/pull/4958/files#diff-522faf538629254cbcb2fed372ba3f9c216497f868f45f6bd82cd6d5d6143c69R471

Issue #4959, opened 2025-01-24
[2.4.0] Access-Control-Expose-Headers is overridden, breaking browser Swarm apps


Cafe137:
### Context
2.4.0

### Summary
Access-Control-Expose-Headers is overridden, impairing browser functionality

### Actual behavior
This is the line that causes the problem.

https://github.com/ethersphere/bee/blob/master/pkg/api/bzz.go#L307

Take the `/feeds/:owner/:topic` endpoint for example. `additionalHeaders` is set there to allow browsers to read `swarm-feed-index`, `swarm-feed-index-next`. It seems that the line in bzz.go discards the previously set headers.

Edit: It seems there is a second issue. The headers are joined with `;`, but it should be `,`.

Cafe137:
These are my local fixes which I can confirm work, but obviously bad quality code 🙂 

```
diff --git a/pkg/api/bzz.go b/pkg/api/bzz.go
index 14a236c1..704cf636 100644
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -616,13 +616,18 @@ func (s *Service) downloadHandler(logger log.Logger, w http.ResponseWriter, r *h
 
        // include additional headers
        for name, values := range additionalHeaders {
-               w.Header().Set(name, strings.Join(values, "; "))
+               w.Header().Set(name, strings.Join(values, ", "))
        }
        if etag {
                w.Header().Set(ETagHeader, fmt.Sprintf("%q", reference))
        }
        w.Header().Set(ContentLengthHeader, strconv.FormatInt(l, 10))
-       w.Header().Set("Access-Control-Expose-Headers", ContentDispositionHeader)
+       if w.Header().Get("Access-Control-Expose-Headers") != "" {
+               w.Header().Set("Access-Control-Expose-Headers", w.Header().Get("Access-Control-Expose-Headers") + ", " + ContentDispositionHeader)
+       } else {
+               w.Header().Set("Access-Control-Expose-Headers", ContentDispositionHeader)
+       }
+
 
        if headersOnly {
                w.WriteHeader(http.StatusOK)
```

gacevicljubisa:
After reviewing the problem and discussing it, here are some thoughts

### 1. Regarding the Delimiter (`;` vs `,`)

[...]

PR #4960, opened 2025-01-24
fix(api): resolve header overwriting and delimiter issue in Access-Control-Expose-Headers

gacevicljubisa:
…response header

### Checklist

### Description

- Replaced w.Header().Set with w.Header().Add to append headers rather than overwrite them.
- Corrected delimiter for Access-Control-Expose-Headers from `;` to `,` as per MDN documentation.

- Browser behavior: The semicolon delimiter (;) was causing issues, particularly in Chromium-based browsers. This was corrected by switching to a comma (,) delimiter for separating multiple header values.
- Header overwriting: The use of w.Header().Set was causing overwriting of Access-Control-Expose-Headers. Replacing it with w.Header().Add ensures headers are appended, resolving inconsistencies in the response headers.

### Related Issue (Optional)

[[2.4.0] Access-Control-Expose-Headers is overridden, breaking browser Swarm apps #4959](https://github.com/ethersphere/bee/issues/4959)

### DIFF
--- a/pkg/api/bzz.go
+++ b/pkg/api/bzz.go
@@ -304,7 +304,7 @@ func (s *Service) fileUploadHandler(
 		span.SetTag("tagID", tagID)
 	}
 	w.Header().Set(ETagHeader, fmt.Sprintf("%q", reference.String()))
-	w.Header().Set("Access-Control-Expose-Headers", SwarmTagHeader)
+	w.Header().Set(AccessControlExposeHeaders, SwarmTagHeader)
 
 	jsonhttp.Created(w, bzzUploadResponse{
 		Reference: reference,
@@ -451,7 +451,7 @@ FETCH:
 			// we should implement an append functionality for this specific header,
 			// since different parts of handlers might be overriding others' values
 			// resulting in inconsistent headers in the response.
-			w.Header().Set("Access-Control-Expose-Headers", SwarmFeedIndexHeader)
+			w.Header().Set(AccessControlExposeHeaders, SwarmFeedIndexHeader)
 			goto FETCH
 		}
 	}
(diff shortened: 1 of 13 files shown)


Cafe137:
Would it be possible to add tests that check that headers are added instead of being replaced, to avoid a future regression?

gacevicljubisa:
> Would it be possible to add tests that check that headers are added instead of being replaced, to avoid a future regression?

Yes, sure.

pkg/api/bzz.go
```diff
@@ -551,8 +551,7 @@ func (s *Service) serveManifestEntry(
 	mtdt := manifestEntry.Metadata()
 	if fname, ok := mtdt[manifest.EntryMetadataFilenameKey]; ok {
 		fname = filepath.Base(fname) // only keep the file name
-		additionalHeaders[ContentDispositionHeader] =
-			[]string{fmt.Sprintf("inline; filename=\"%s\"", fname)}
+		additionalHeaders[ContentDispositionHeader] = []string{fmt.Sprintf("inline; filename=\"%s\"", fname)}
```
janos, line 554: I would suggest to use this opportunity to harden the filename handling in this header. It is tricky to deal with a user provided input and even more so in the complex string as this one. If the username has quotes in it it will break the header value parser resulting in a partial filename or no filename at all (depending on browsers and their versions). I would suggest to use the stdlib's approach and add the function escapeQuotes from here https://github.com/golang/go/blob/go1.23.5/src/mime/multipart/writer.go#L154. Going through different RFCs may be overwhelming, and escapeQuotes should be just fine, as long as most browsers, applications and libraries handle the escaping as such.

pkg/api/bzz_test.go
```diff
@@ -1090,3 +1090,25 @@ func TestDirectUploadBzz(t *testing.T) {
 		}),
 	)
 }
+
```
janos, line 1093: I suppose that this test will be made functional. Maybe to add also the quotes escaping test in it?

janos (approved):
LGTM. Thanks for addressing my comments.

istae (approved):

Issue #4982, opened 2025-02-04
Panic when retrieving chunk after attempting to disable cache


Cafe137:
### Context
Bee 2.4.0, light node, mainnet

### Summary
Bee crashed with a panic after disabling the cache(maybe improperly?), in about ~20-30 GET requests (not reproducible 100%). I am attempting to fetch chunks (SOCs, but may be irrelevant).

### Expected behavior
No crash

### Actual behavior

```
panic: runtime error: makeslice: cap out of range

goroutine 336 [running]:
github.com/ethersphere/bee/v2/pkg/storer/internal/cache.(*Cache).RemoveOldest(0x140025803d8, {0x101fa2fc0, 0x140001cc780}, {0x101fa6060, 0x1400253c2a0}, 0xffffffffffffffff)
	github.com/ethersphere/bee/v2/pkg/storer/internal/cache/cache.go:190 +0xbc
github.com/ethersphere/bee/v2/pkg/storer.(*DB).cacheWorker(0x140023ac008, {0x101fa2fc0, 0x140001cc780})
	github.com/ethersphere/bee/v2/pkg/storer/cachestore.go:48 +0x180
created by github.com/ethersphere/bee/v2/pkg/storer.New in goroutine 38
	github.com/ethersphere/bee/v2/pkg/storer/storer.go:589 +0xbcc
```

### Steps to reproduce
I added these entries to my config yaml:

```yaml
cache-retrieval: false
cache-capacity: 0
```

### Possible solution
TBD

istae:
so it crashes again after a restart?

Cafe137:
@istae 

Not right after restart; not automatically.

When I start requesting data from Swarm, some (1-30) requests work properly until it crashes again.

PR #4986, opened 2025-02-05
fix: prevent cache size underflow

acha-bill:
### Checklist

### Description
Prevent cache size from underflowing
 
### Related Issue (Optional)
#4982

### DIFF
--- a/pkg/storer/internal/cache/cache.go
+++ b/pkg/storer/internal/cache/cache.go
@@ -14,7 +14,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	storage "github.com/ethersphere/bee/v2/pkg/storage"
+	"github.com/ethersphere/bee/v2/pkg/storage"
 	"github.com/ethersphere/bee/v2/pkg/storer/internal/transaction"
 	"github.com/ethersphere/bee/v2/pkg/swarm"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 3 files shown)


istae:
Has the underflow happened in production?

gacevicljubisa:
Size filed is of type Int, so negative number are allowed.

```go
type Cache struct {
	size     atomic.Int64
	capacity int
	glock    *multex.Multex // blocks Get and Put ops while shallow copy is running.
}
```

acha-bill:
>  Has the underflow happened in production?

Only mentioned in #4982

> Size filed is of type Int64, so negative number are allowed.

But when returning, it's cast to uint64. This causes the underflow to MaxUint64 if size was < 0

pkg/storer/internal/cache/cache.go
```diff
@@ -232,7 +232,9 @@ func (c *Cache) RemoveOldest(ctx context.Context, st transaction.Storage, count
 				if err != nil {
 					return err
 				}
-				c.size.Add(-1)
+				if c.size.Load() > 0 {
```
istae, line 235: instead of doing this here: I would make the check here https://github.com/ethersphere/bee/blob/master/pkg/storer/internal/cache/cache.go#L64

but we need to understand why the size went below zero

istae: or maybe it did not go down below zero, and the cache size was too large to begin with?

acha-bill: Looking at the panic, 
```
github.com/ethersphere/bee/v2/pkg/storer/internal/cache.(*Cache).RemoveOldest(0x140025803d8, {0x101fa2fc0, 0x140001cc780}, {0x101fa6060, 0x1400253c2a0}, 0xffffffffffffffff)
```

`RemoveOldest` was called with a `count = 0xffffffffffffffff`. That's max uint64. 
This indicates that the size went to -1.


When `RemoveOldest ` is called, `len(evictItems)` is not always equal to the cache size at that instant because of the `put` happening in another goroutine.

```
fmt.Println("len of evict items", len(evictItems), "cache size", c.size.Load())

> len of items to evict 10, cache size 10, size after 0
> len of items to evict 1, cache size 6, size after 5
> len of items to evict 6, cache size 5, size after -1 
```

istae (approved):

gacevicljubisa (approved):

PR #4993, opened 2025-02-06
perf: remove unncessary goroutine pileup of skippeer workers

istae:
### Checklist

### Description
Before, for each pushsync and retrieval, a separate skippeers list was created, along side it a worker that would prune entries based on a timer. Now, the worker is created once only for the errSkip peers list, and not for each request.

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -144,7 +144,7 @@ func New(
 		metrics:                 newMetrics(),
 		tracer:                  tracer,
 		signer:                  signer,
-		errSkip:                 skippeers.NewList(),
+		errSkip:                 skippeers.NewList(time.Minute),
 		warmupPeriod:            time.Now().Add(warmupTime),
 		shallowReceiptTolerance: shallowReceiptTolerance,
 	}
@@ -387,7 +387,7 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
 		return nil, fmt.Errorf("pushsync: storage radius: %w", err)
 	}
 
-	skip := skippeers.NewList()
+	skip := skippeers.NewList(0)
 	defer skip.Close()
 
 	for sentErrorsLeft > 0 {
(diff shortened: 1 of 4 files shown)


pkg/retrieval/retrieval.go
```diff
@@ -158,7 +158,7 @@ func (s *Service) RetrieveChunk(ctx context.Context, chunkAddr, sourcePeerAddr s
 
 	v, _, err := s.singleflight.Do(ctx, flightRoute, func(ctx context.Context) (swarm.Chunk, error) {
 
-		skip := skippeers.NewList()
+		skip := skippeers.NewList(0)
```
martinconic, line 161: Is `0` ok here? You do not want any worker to start here?

pkg/pushsync/pushsync.go
```diff
@@ -387,7 +387,7 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
 		return nil, fmt.Errorf("pushsync: storage radius: %w", err)
 	}
 
-	skip := skippeers.NewList()
+	skip := skippeers.NewList(0)
```
martinconic, line 390: Is 0 ok here? You do not want any worker to start here?

pkg/retrieval/retrieval.go
istae: Yep that's the point :) the per request skiplist does not need a worker

pkg/skippeers/skippeers.go
```diff
@@ -25,24 +25,26 @@ type List struct {
...
 	return l
 }
 
-func (l *List) worker() {
+func (l *List) worker(d time.Duration) {
 
 	defer l.wg.Done()
 
-	ticker := time.NewTicker(time.Minute)
+	ticker := time.NewTicker(d)
```
martinconic, line 47: Should we have a fallback check before just in case we send d=0 , or smaller?

istae: The check in the constructor should be enough

acha-bill (approved):

PR #5025, opened 2025-02-27
feat: push metrics

istae:
### Checklist

### Description
TODO: a unique id that gets created on startup needs to be added as a label before the metrics are pushed so that the metrics from one particular node do not mix with another node's metrics


pkg/metrics/registry/registry.go
```diff
@@ -0,0 +1,130 @@
...
+				// Collect CPU metrics
+				percentages, err := cpu.Percent(0, false)
+				if err == nil && len(percentages) > 0 {
+					r.cpuGauge.Set(percentages[0])
+				} else if err != nil {
+					logger.Debug("failed to collect CPU metrics", "error", err)
+				}
+
+				// Collect memory metrics
+				vm, err := mem.VirtualMemory()
```
istae, line 103: i think when the metrics are gathered for pushing, the mem and cpu metrics are already collected, I dont think you need to add them manually like this

PR #5029, opened 2025-02-28
feat: feed legacy payload header

### LINKED ISSUE #5027: `GET /feeds/{owner}/{topic}` incorrectly resolves 40/72 bytes payloads as `timestamp|reference`
### Context
Latest

### Summary
Since the wrapped chunk feed improvements, it is possible to have root chunks directly in feed updates, instead of the legacy format, which is `timestamp|reference` (40 bytes) or `timestamp|reference|encryption_key` (72 bytes).

There is an issue, in which payloads which are exactly 40 or 72 bytes long, fail with a 404 as Bee attempts to resolve them in the legacy way. This is problematic, especially in feed-based messaging applications, where one feed update may very likely hit 40 or 72 bytes exactly.

**For example, assume that the root chunk has this payload: `This string is exactly 40 bytes in utf-8`**

### Expected behavior
`GET /feeds/{owner}/{topic}` responds with `This string is exactly 40 bytes in utf-8`

### Actual behavior
`GET /feeds/{owner}/{topic}` responds with 404, because it resolves `This str` as timestamp and `ing is exactly 40 bytes in utf-8` as reference, and tries to follow the reference.

### Steps to reproduce
The string `This string is exactly 40 bytes in utf-8` can be tested to reproduce.

### Possible solution

1. Turn off legacy resolving by default, enable it with a param/header
2. Add a param/header that forces wrapped chunk resolving

nugaon:
### Checklist

### Description
resolves #5027


.github/workflows/beekeeper.yml
```diff
@@ -14,7 +14,7 @@ env:
   SETUP_CONTRACT_IMAGE: "ethersphere/bee-localchain"
   SETUP_CONTRACT_IMAGE_TAG: "0.9.2"
   BEELOCAL_BRANCH: "main"
-  BEEKEEPER_BRANCH: "master"
+  BEEKEEPER_BRANCH: "feat/feed-wrapping"
```
gacevicljubisa, line 17: Beekeeper [PR](https://github.com/ethersphere/beekeeper/pull/469) should be merged together with these changes. Also, `BEEKEEPER_BRANCH` should be `master` when merged.

gacevicljubisa (approved):

acha-bill (approved):

PR #5032, opened 2025-03-04
perf(uploadstore): use deletewithstamp instead and tag deletion case …

istae:
…cleanup

### Checklist

### Description
internal.Uploadstore now uses `DeleteWithStamp` as the regular Delete does an internal load of the stamp, which is unnecessary in this case.
Also, removed the assignment of zero as tagID to the uploadItem in the case that the tag is missing. The item eventually get removed when the chunk is synced, so there is no need to the extra Put operation.

### DIFF
--- a/pkg/storer/internal/upload/uploadstore.go
+++ b/pkg/storer/internal/upload/uploadstore.go
@@ -565,9 +565,18 @@ func CleanupDirty(st transaction.Storage) error {
 // Report is the implementation of the PushReporter interface.
 func Report(ctx context.Context, st transaction.Store, chunk swarm.Chunk, state storage.ChunkState) error {
 
-	ui := &uploadItem{Address: chunk.Address(), BatchID: chunk.Stamp().BatchID()}
-
 	indexStore := st.IndexStore()
+
+	ui := &uploadItem{Address: chunk.Address(), BatchID: chunk.Stamp().BatchID()}
+	err := indexStore.Get(ui)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


acha-bill (approved):

gacevicljubisa (approved):

martinconic:
When testing deferred uploads and deleting the tag use( in the stage where the BMT is constructed), I then do NOT see any logs that show pushsync is pushing chunks. There my by some weird cases when testing deferred uploads and deleting in the process the tag that is used.

martinconic:
Even if previous change seems to fix the issue, I get another one when testing:
When testing deferred uploads and deleting the tag used(after BMT was created and when pushsync is starded), I get this this error messages a lot:
```
"time"="2025-03-06 18:44:47.394951" "level"="debug" "logger"="node/pushsync" "msg"="could not push to peer" "chunk_address"="aed9ed3c4862c213cfe5d3bc721543c321402d7b1bed6cec0d01a50ba90e37b4" "id_address"="aed9ed3c4862c213cfe5d3bc721543c321402d7b1bed6cec0d01a50ba90e37b4" "peer_address"="97d97526816a0bd1e88bf4bad949c408ee4fd6476ff12ca60eb2b36f0360490f" "error"="new stream for peer 97d97526816a0bd1e88bf4bad949c408ee4fd6476ff12ca60eb2b36f0360490f: send headers: read message: stream reset"
``` 

The file seems to be uploaded in the end, I can download it, but the messages show until each chunk is sent

PR #5037, opened 2025-03-06
feat(pushsync): forward chunk after storing

acha-bill:
### Checklist

### Description
Forward a chunk to the closest peer after storing it

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -287,7 +287,20 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) 
 
 	if ps.topologyDriver.IsReachable() && swarm.Proximity(ps.address.Bytes(), chunkAddress.Bytes()) >= rad {
 		stored, reason = true, "is within AOR"
-		return store(ctx)
+		err = store(ctx)
+		if err != nil {
+			return err
+		}
+
+		_, err = ps.pushToClosest(ctx, chunk, false)
+		if err != nil {
+			if !errors.Is(err, topology.ErrNotFound) && !errors.Is(err, topology.ErrWantSelf) {
+				// Do not error out pushsync if we cannot forward to the closest peer after storing.
+				// The peer will get it via pullsync.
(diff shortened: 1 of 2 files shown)


pkg/pushsync/pushsync.go
```diff
@@ -287,7 +287,16 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 
 	if ps.topologyDriver.IsReachable() && swarm.Proximity(ps.address.Bytes(), chunkAddress.Bytes()) >= rad {
 		stored, reason = true, "is within AOR"
-		return store(ctx)
+		err = store(ctx)
+		if err != nil {
+			return err
+		}
+
+		err = ps.forwardToClosest(ctx, chunk, ps.address)
```
istae, line 295: The reason why we return the receipt as soon as chunk lands in the neighborhood is to optimize the request and reduce the number of hops as the chunk will be synced in the neighborhood via pullsync anyways.

What I recommend doing is to keep the logic of returning the receipt early, but continue to forward it either in a way that does not stagnate the request, here's how you can achieve that:

instead of adding a new function, what you can do instead is inside the if statement, store the chunk, then set "stored" to true, then enter the switch statement after the `pushToClosest` call. if the error is ErrWantSelf AND you already stored, just return nil. Also inside the err==nil case you would need to check again if stored is true before running the logic inside the block.

istae (changes requested):

pkg/pushsync/pushsync.go
acha-bill: `ps.forwardToClosest` only runs when the chunk is stored and there's already an early return of the receipt in the store function. `forwardToClosest` only continues to forward.

istae: i see. 
another point is the `pushToClosest` already does what `forwardToClosest` is supposed to do, so no need to repeat the functionality and maintain two functions for sending chunks. imo it's better if you refactor your changes to accommodate only calling `pushToClosest`.

istae (approved):
great!

gacevicljubisa (approved):

PR #5047, opened 2025-03-12
feat: add isWarmingUp to status response

acha-bill:
### Checklist

### Description

Add `isWarmingUp` to `/status` response.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -461,7 +461,7 @@ func NewBee(
 
 		apiService.Mount()
 		apiService.SetProbe(probe)
-		apiService.SetWarmupTime(time.Now().Add(o.WarmupTime))
+		apiService.SetIsWarmingUp(true)
 		apiService.SetSwarmAddress(&swarmAddress)
 
 		apiServer := &http.Server{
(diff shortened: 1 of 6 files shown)


martinconic (approved):

istae (changes requested):
@acha-bill the issue is about exposing the warmup period of the node you are running in the /status API. We do not need to expose if ours peers are in warmup via the status _protocol_ .

pkg/node/node.go
```diff
@@ -461,7 +461,7 @@ func NewBee(
 
 		apiService.Mount()
 		apiService.SetProbe(probe)
-
+		apiService.SetWarmupTime(time.Now().Add(o.WarmupTime))
```
istae, line 464: here 's the tricky thing, if the postage batches are syncing, starting here, then setting this up here becomes inaccurate as we have not run this line yet https://github.com/ethersphere/bee/pull/5047/files#diff-515b2d0691202c4ea5e578ec896258ae9c440c381b608ed1eb7116a27847057dR782

maybe it would help if instead of passing a time to the api pkg, it can be done by toggling a bool externally?

istae (changes requested):

pkg/node/node.go
acha-bill: Now that I think of it again, the setter is better

acha-bill: In the next PR we'll improve how we detect and set the warmup

istae (approved):
very nice, much cleaner this time around

PR #5053, opened 2025-03-14
chore: golang 1.24

gacevicljubisa:
### Checklist


martinconic:
My recommendation would be that we should update most of the libraries separate one from each other and separate from upgrading to the new Go version. Doing this we can avoid struggling to find possible issues/incompatibilities that might come with the new updates. the `libp2p` library should be in this category and any update could be sensitive and testes well.

PR #5057, opened 2025-03-20
feat: soc dispersed replica

nugaon:
Dispersed replica implementation of the SOC

### Checklist

### DIFF
--- a/pkg/api/soc.go
+++ b/pkg/api/soc.go
@@ -14,8 +14,10 @@ import (
 
 	"github.com/ethersphere/bee/v2/pkg/accesscontrol"
 	"github.com/ethersphere/bee/v2/pkg/cac"
+	"github.com/ethersphere/bee/v2/pkg/file/redundancy"
 	"github.com/ethersphere/bee/v2/pkg/jsonhttp"
 	"github.com/ethersphere/bee/v2/pkg/postage"
+	"github.com/ethersphere/bee/v2/pkg/replicas"
 	"github.com/ethersphere/bee/v2/pkg/soc"
 	"github.com/ethersphere/bee/v2/pkg/storer"
 	"github.com/ethersphere/bee/v2/pkg/swarm"
@@ -47,10 +49,11 @@ func (s *Service) socUploadHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	headers := struct {
-		BatchID        []byte        `map:"Swarm-Postage-Batch-Id"`
-		StampSig       []byte        `map:"Swarm-Postage-Stamp"`
-		Act            bool          `map:"Swarm-Act"`
-		HistoryAddress swarm.Address `map:"Swarm-Act-History-Address"`
+		BatchID        []byte            `map:"Swarm-Postage-Batch-Id"`
+		StampSig       []byte            `map:"Swarm-Postage-Stamp"`
+		Act            bool              `map:"Swarm-Act"`
+		HistoryAddress swarm.Address     `map:"Swarm-Act-History-Address"`
+		RLevel         *redundancy.Level `map:"Swarm-Redundancy-Level"`
 	}{}
 	if response := s.mapStructure(r.Header, &headers); response != nil {
 		response("invalid header params", logger, w)
@@ -64,11 +67,23 @@ func (s *Service) socUploadHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	var (
-		putter storer.PutterSession
-		err    error
+		basePutter storer.PutterSession
+		putter     storer.PutterSession
+		err        error
 	)
 
+	rLevel := redundancy.DefaultLevel
+	if headers.RLevel != nil {
+		rLevel = *headers.RLevel
+	}
+
 	if len(headers.StampSig) != 0 {
+		if headers.RLevel != nil {
+			logger.Error(nil, "redundancy level is not supported with stamp signature")
+			jsonhttp.BadRequest(w, "redundancy level is not supported with stamp signature")
+			return
+		}
+		rLevel = redundancy.NONE
 		stamp := postage.Stamp{}
 		if err := stamp.UnmarshalBinary(headers.StampSig); err != nil {
 			errorMsg := "Stamp deserialization failure"
@@ -91,6 +106,10 @@ func (s *Service) socUploadHandler(w http.ResponseWriter, r *http.Request) {
 			Pin:      false,
 			Deferred: false,
 		})
+		basePutter = putter
+		if rLevel != redundancy.NONE {
+			putter = replicas.NewSocPutter(putter, rLevel)
+		}
 	}
 	if err != nil {
 		logger.Debug("get putter failed", "error", err)
@@ -183,7 +202,7 @@ func (s *Service) socUploadHandler(w http.ResponseWriter, r *http.Request) {
 	reference := sch.Address()
 	historyReference := swarm.ZeroAddress
 	if headers.Act {
-		reference, historyReference, err = s.actEncryptionHandler(r.Context(), putter, reference, headers.HistoryAddress)
+		reference, historyReference, err = s.actEncryptionHandler(r.Context(), basePutter, reference, headers.HistoryAddress)
 		if err != nil {
 			logger.Debug("access control upload failed", "error", err)
 			logger.Error(nil, "access control upload failed")
@@ -229,11 +248,16 @@ func (s *Service) socGetHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	headers := struct {
-		OnlyRootChunk bool `map:"Swarm-Only-Root-Chunk"`
+		OnlyRootChunk bool              `map:"Swarm-Only-Root-Chunk"`
+		RLevel        *redundancy.Level `map:"Swarm-Redundancy-Level"`
 	}{}
 	if response := s.mapStructure(r.Header, &headers); response != nil {
 		response("invalid header params", logger, w)
 		return
+	}
+	rLevel := redundancy.DefaultLevel
+	if headers.RLevel != nil {
+		rLevel = *headers.RLevel
 	}
 
 	address, err := soc.CreateAddress(paths.ID, paths.Owner)
@@ -244,6 +268,9 @@ func (s *Service) socGetHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	getter := s.storer.Download(true)
+	if rLevel != 0 {
+		getter = replicas.NewSocGetter(getter, rLevel)
+	}
 	sch, err := getter.Get(r.Context(), address)
 	if err != nil {
 		logger.Error(err, "soc retrieval has been failed")
--- a/pkg/replicas/getter_soc.go
+++ b/pkg/replicas/getter_soc.go
@@ -1,0 +1,135 @@
+// Copyright 2023 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// the code below implements the integration of dispersed replicas in chunk fetching.
+// using storage.Getter interface.
+package replicas
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/ethersphere/bee/v2/pkg/file/redundancy"
+	"github.com/ethersphere/bee/v2/pkg/storage"
+	"github.com/ethersphere/bee/v2/pkg/swarm"
+)
+
+// getter is the private implementation of storage.Getter, an interface for
+// retrieving chunks. This getter embeds the original simple chunk getter and extends it
+// to a multiplexed variant that fetches chunks with replicas for SOC.
+//
+// the strategy to retrieve a chunk that has replicas can be configured with a few parameters:
+//   - RetryInterval: the delay before a new batch of replicas is fetched.
+//   - depth: 2^{depth} is the total number of additional replicas that have been uploaded
+//     (by default, it is assumed to be 4, ie. total of 16)
+//   - (not implemented) pivot: replicas with address in the proximity of pivot will be tried first
+type socGetter struct {
+	wg sync.WaitGroup
+	storage.Getter
+	level redundancy.Level
+}
+
+// NewSocGetter is the getter constructor
+func NewSocGetter(g storage.Getter, level redundancy.Level) storage.Getter {
+	return &socGetter{Getter: g, level: level}
+}
+
+// Get makes the socGetter satisfy the storage.Getter interface
+func (g *socGetter) Get(ctx context.Context, addr swarm.Address) (ch swarm.Chunk, err error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// channel that the results (retrieved chunks) are gathered to from concurrent
+	// workers each fetching a replica
+	resultC := make(chan swarm.Chunk)
+	// errc collects the errors
+	errc := make(chan error, 17)
+	var errs error
+	errcnt := 0
+
+	// concurrently call to retrieve chunk using original SOC address
+	g.wg.Add(1)
+	go func() {
+		defer g.wg.Done()
+		ch, err := g.Getter.Get(ctx, addr)
+		if err != nil {
+			errc <- err
+			return
+		}
+
+		select {
+		case resultC <- ch:
+		case <-ctx.Done():
+		}
+	}()
+	// counters
+	n := 0      // counts the replica addresses tried
+	target := 2 // the number of replicas attempted to download in this batch
+	total := g.level.GetReplicaCount()
+
+	//
+	rr := newSocReplicator(addr, g.level)
+	next := rr.c
+	var wait <-chan time.Time // nil channel to disable case
+	// addresses used are doubling each period of search expansion
+	// (at intervals of RetryInterval)
+	ticker := time.NewTicker(RetryInterval)
+	defer ticker.Stop()
+	for level := uint8(0); level <= uint8(g.level); {
+		select {
+		// at least one chunk is retrieved, cancel the rest and return early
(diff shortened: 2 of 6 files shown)


pkg/api/feed.go
```diff
@@ -230,7 +238,9 @@ func (s *Service) feedPostHandler(w http.ResponseWriter, r *http.Request) {
 		logger:         logger,
 	}
 
-	l := loadsave.New(s.storer.ChunkStore(), s.storer.Cache(), requestPipelineFactory(r.Context(), putter, false, 0), redundancy.DefaultLevel)
+	rLevel := headers.RedundancyLevel
+
+	l := loadsave.New(s.storer.ChunkStore(), s.storer.Cache(), requestPipelineFactory(r.Context(), putter, false, 0), rLevel)
```
martinconic, line 244: Why do we have 2 redundancy levels: one inside `requestPipelineFactory` , and the other `rLevel` that is different different ?

pkg/feeds/factory/factory.go
```diff
@@ -19,12 +19,17 @@ func New(getter storage.Getter) feeds.Factory {
 	return &factory{getter}
 }
 
-func (f *factory) NewLookup(t feeds.Type, feed *feeds.Feed) (feeds.Lookup, error) {
+func (f *factory) NewLookup(t feeds.Type, feed *feeds.Feed, specialGetter storage.Getter) (feeds.Lookup, error) {
```
martinconic: Can we have a more sugestive naming instead of `specialGetter`, we definitely will not know in the future what this is for ?

pkg/api/feed.go
nugaon: could you elaborate this please? I see only one rLevel reference.

pkg/feeds/factory/factory.go
nugaon: this getter overwrites the default getter of the feed factory.
I can call it simple `getter`. would it be better or do you have some other name in mind?

pkg/replicas/getter_soc.go
```diff
@@ -0,0 +1,137 @@
...
+// retrieving chunks. This getter embeds the original simple chunk getter and extends it
+// to a multiplexed variant that fetches chunks with replicas for SOC.
+//
+// the strategy to retrieve a chunk that has replicas can be configured with a few parameters:
+//   - SOCRetryInterval: the delay before a new batch of replicas is fetched.
+//   - depth: 2^{depth} is the total number of additional replicas that have been uploaded
+//     (by default, it is assumed to be 4, ie. total of 16)
+//   - (not implemented) pivot: replicas with address in the proximity of pivot will be tried first
+type socGetter struct {
+	wg sync.WaitGroup
```
gacevicljubisa: Correct me if I am wrong, but I do not see if WaitGroup is awaited?
Is it maybe required in Get:
```golang
    defer func() {
        g.wg.Wait()
    }()
```

pkg/replicas/getter_soc.go
```diff
@@ -0,0 +1,137 @@
...
+// Get makes the socGetter satisfy the storage.Getter interface
+func (g *socGetter) Get(ctx context.Context, addr swarm.Address) (ch swarm.Chunk, err error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// channel that the results (retrieved chunks) are gathered to from concurrent
+	// workers each fetching a replica
+	resultC := make(chan swarm.Chunk)
+	// errc collects the errors
+	errc := make(chan error, 17)
```
gacevicljubisa: Why the buffer size is 17?

gacevicljubisa: Can we calculate proper buffer size based on redundancy level

pkg/replicas/getter_soc.go
```diff
@@ -0,0 +1,137 @@
...
+				case resultC <- ch:
+				case <-ctx.Done():
+				}
+			}()
+			n++
+			if n < target {
+				continue
+			}
+			next = nil
+			wait = ticker.C
```
gacevicljubisa: Does it have more sense here to have:
```
wait = time.After(SOCRetryInterval)
```
This can replace ticker... And here you reset it every time?

pkg/replicas/putter_soc.go
```diff
@@ -0,0 +1,90 @@
...
+	for r := range rr.c {
+		wg.Add(1)
+		go func(r *socReplica) {
+			defer wg.Done()
+			// create a new chunk with the replica address
+			sch := swarm.NewChunk(swarm.NewAddress(r.addr), ch.Data())
+			err := p.putter.Put(ctx, sch)
+			errc <- err
+		}(r)
+	}
```
gacevicljubisa, line 60: Context is not checked here, so if canceled, all gorotuines will continue.

pkg/replicas/getter_soc.go
nugaon: really nice catch, it is a problem in the cac replica getter as well.

pkg/replicas/putter_soc.go
nugaon: it is only a type init before the Put function call that handles ctx.
is it really needed?

pkg/replicas/putter_soc.go
```diff
@@ -0,0 +1,90 @@
...
+	errc := make(chan error, p.rLevel.GetReplicaCount())
+	wg := sync.WaitGroup{}
+	for r := range rr.c {
+		wg.Add(1)
+		go func(r *socReplica) {
+			defer wg.Done()
+			// create a new chunk with the replica address
+			sch := swarm.NewChunk(swarm.NewAddress(r.addr), ch.Data())
+			err := p.putter.Put(ctx, sch)
+			errc <- err
```
gacevicljubisa: Good practice is to check and send only if `err != nil`.

pkg/api/soc.go
```diff
@@ -47,10 +50,11 @@ func (s *Service) socUploadHandler(w http.ResponseWriter, r *http.Request) {
...
 	headers := struct {
-		BatchID        []byte        `map:"Swarm-Postage-Batch-Id"`
-		StampSig       []byte        `map:"Swarm-Postage-Stamp"`
-		Act            bool          `map:"Swarm-Act"`
-		HistoryAddress swarm.Address `map:"Swarm-Act-History-Address"`
+		BatchID        []byte            `map:"Swarm-Postage-Batch-Id"`
+		StampSig       []byte            `map:"Swarm-Postage-Stamp"`
+		Act            bool              `map:"Swarm-Act"`
+		HistoryAddress swarm.Address     `map:"Swarm-Act-History-Address"`
+		RLevel         *redundancy.Level `map:"Swarm-Redundancy-Level"`
```
gacevicljubisa, line 57: I would suggest to use for redundancy level header (`RLevel`), value type instead of a pointer, and all nil checks can be removed. Code will default to `redundancy.NONE` of the header is missing.

pkg/api/soc.go
```diff
@@ -229,12 +254,17 @@ func (s *Service) socGetHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	headers := struct {
-		OnlyRootChunk bool `map:"Swarm-Only-Root-Chunk"`
+		OnlyRootChunk bool              `map:"Swarm-Only-Root-Chunk"`
+		RLevel        *redundancy.Level `map:"Swarm-Redundancy-Level"`
```
gacevicljubisa: I would suggest to use for redundancy level header (`RLevel`), value type instead of a pointer, and all nil checks can be removed. Code will default to `redundancy.NONE` of the header is missing.

pkg/replicas/putter_soc.go
```diff
@@ -0,0 +1,90 @@
...
+	return errors.Join(errs...)
+}
+
+// socPutterSession extends the original socPutter
+type socPutterSession struct {
+	socPutter
+	ps storer.PutterSession
+}
+
+// NewSocPutter is the putterSession constructor
```
gacevicljubisa: comment method name should be NewSocPutterSession

pkg/replicas/replicas_soc.go
```diff
@@ -0,0 +1,100 @@
...
+
+import (
+	"github.com/ethersphere/bee/v2/pkg/file/redundancy"
+	"github.com/ethersphere/bee/v2/pkg/swarm"
+)
+
+// socReplicator running the find for replicas
+type socReplicator struct {
+	addr   []byte // chunk address
+	sizes  [5]int // number of distinct neighbourhoods redcorded for each depth
```
gacevicljubisa: typo in comment, should be `recorded`

pkg/replicas/replicas_soc.go
```diff
@@ -0,0 +1,100 @@
...
+	c      chan *socReplica
+	rLevel redundancy.Level
+}
+
+// newSocReplicator socReplicator constructor
+func newSocReplicator(addr swarm.Address, rLevel redundancy.Level) *socReplicator {
+	rr := &socReplicator{
+		addr:   addr.Bytes(),
+		sizes:  redundancy.GetReplicaCounts(),
+		c:      make(chan *socReplica, 16),
```
gacevicljubisa: Does it have sense to use buffer size that will match replica count?

pkg/replicas/getter.go
```diff
@@ -36,7 +36,6 @@ var ErrSwarmageddon = errors.New("swarmageddon has begun")
 //     (by default, it is assumed to be 4, ie. total of 16)
 //   - (not implemented) pivot: replicas with address in the proximity of pivot will be tried first
 type getter struct {
-	wg sync.WaitGroup
```
martinconic, line 39: Why did you moved this WaitGroup in the Get function as a variable?

pkg/replicas/getter_soc.go
```diff
@@ -0,0 +1,136 @@
...
+func NewSocGetter(g storage.Getter, level redundancy.Level) storage.Getter {
+	return &socGetter{Getter: g, level: level}
+}
+
+// Get makes the socGetter satisfy the storage.Getter interface
+func (g *socGetter) Get(ctx context.Context, addr swarm.Address) (ch swarm.Chunk, err error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	var wg sync.WaitGroup
```
martinconic, line 46: wg should be a variable in the socGetter struct.

gacevicljubisa: Why, it doesn't need to be shared across different methods, or exported? It is only internal to Get?

pkg/replicas/replicas_soc.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// socReplica of the mined SOC chunk (address) that serve as replicas
+type socReplica struct {
+	addr  []byte // byte slice of SOC address
+	nonce uint8  // byte of the mined nonce
+}
+
+// replicate returns a replica params structure seeded with a byte of entropy as argument
+func (rr *socReplicator) replicate(i uint8, bitsRequired uint8) (sp *socReplica) {
```
martinconic, line 43: How did you get to this logic? How can I validate it? It is documented somewhere so I can validate it?

pkg/replicas/replicas_soc.go
nugaon: hmm, as I see it is not even used anymore, I will just remove it

pkg/replicas/replicas_soc.go
nugaon: yes, it has a sense definitely : )

pkg/replicas/getter.go
nugaon: because it is only used in the Get function and it may mix up with another getter call's `Wait.Add(1)` which seems wrong.

pkg/replicas/replicas_soc.go
nugaon: I just figured it out, it seemed a right and lightweight solution.
It is validated by the same `dispersed` function that is used in CAC replication test.
I wrote an explanation for it in the PR's description.

pkg/api/feed.go
```diff
@@ -231,7 +239,9 @@ func (s *Service) feedPostHandler(w http.ResponseWriter, r *http.Request) {
 		logger:         logger,
 	}
 
-	l := loadsave.New(s.storer.ChunkStore(), s.storer.Cache(), requestPipelineFactory(r.Context(), putter, false, 0), redundancy.DefaultLevel)
+	rLevel := headers.RedundancyLevel
```
gacevicljubisa, line 242: For Get, should the default be None? or Paranoid?

pkg/api/soc.go
```diff
@@ -64,11 +68,22 @@ func (s *Service) socUploadHandler(w http.ResponseWriter, r *http.Request) {
...
-		err    error
+		basePutter storer.PutterSession // the putter used to store regular chunks
+		putter     storer.PutterSession // the putter used to store SOC replica chunks
+		err        error
 	)
 
+	var rLevel redundancy.Level
+	if headers.RLevel != nil {
+		rLevel = *headers.RLevel
+	}
```
gacevicljubisa, line 79: What value should the default one?

pkg/replicas/putter.go
```diff
@@ -40,22 +39,17 @@ func (p *putter) Put(ctx context.Context, ch swarm.Chunk) (err error) {
...
-			}
-			errc <- err
-		})
+		sch, err := soc.New(r.id, ch).Sign(signer)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+
+		if err = p.putter.Put(ctx, sch); err != nil {
```
gacevicljubisa: Here I removed concurrent PUT, because this should be IO write operation.

pkg/api/bzz.go
```diff
@@ -400,13 +401,14 @@ func (s *Service) serveReference(logger log.Logger, address swarm.Address, pathV
 		cache = *headers.Cache
 	}
 
-	rLevel := redundancy.DefaultLevel
+	rLevel := redundancy.PARANOID
```
martinconic: Actually DefaultLevel is PARANOID. We should use that one, or remove the logic of having a default redundancy level. Same in all other places where it is changed from DefaultLevel to PARANOID

pkg/replicas/getter_soc.go
```diff
@@ -0,0 +1,132 @@
...
+			n++
+			if n < target {
+				continue
+			}
+			next = nil
+			wait = time.After(SOCRetryInterval)
+		}
+	}
+
+	return nil, nil
```
martinconic, line 135: Here maybe we should return some error, as replicas were not found, right?

pkg/replicas/getter_soc.go
```diff
@@ -0,0 +1,132 @@
...
+			// ticker switches on the address channel
+		case <-wait:
+			next = rr.c
+			level++
+			target = 1 << level
+			n = 0
+			continue
+
+			// getting the addresses in order
+		case so := <-next:
```
martinconic, line 106: I thin this is better:
```
case so, ok := <-next:
    if !ok {
        next = nil
        continue
    }
```

pkg/replicas/getter_soc.go
```diff
@@ -0,0 +1,132 @@
...
+		select {
+		// at least one chunk is retrieved, cancel the rest and return early
+		case chunk := <-resultC:
+			cancel()
+			return chunk, nil
+
+		case err = <-errc:
+			errs = errors.Join(errs, err)
+			errcnt++
+			if errcnt > total {
```
martinconic, line 93: `total` is `g.level.GetReplicaCount()`. this does not include the original address right? if so, than maybe we should have here `total+1` ?

pkg/replicas/getter_soc.go
```diff
@@ -0,0 +1,132 @@
...
+				return nil, errors.Join(ErrSwarmageddon, errs)
```
martinconic, line 94: is `ErrSwarmageddon` the right error here?

pkg/replicas/putter_soc.go
martinconic: I think the context check should be inside the for loop

pkg/replicas/replicas_soc.go
```diff
@@ -0,0 +1,98 @@
...
+	addr := make([]byte, 32)
+	copy(addr, rr.addr)
+	mirroredBits := mirrorBitsToMSB(i, bitsRequired)
+	// zero out the first leading bitsRequired bits of addr[0] and set mirroredBits of `i`
+	addr[0] &= 0xFF >> bitsRequired
+	addr[0] |= mirroredBits
+	if addr[0] == rr.addr[0] {
+		// xor MSB after the mirrored bits because the iteration found the original address
+		addr[0] ^= 1 << (bitsRequired - 1)
```
martinconic, line 52: I am not confident about this code, hard to understand things from this file. I think we need some tests here.

pkg/replicas/putter_soc.go
```diff
@@ -0,0 +1,79 @@
...
+		rLevel: rLevel,
+	}
+}
+
+// Put makes the putter satisfy the storage.Putter interface
+func (p *socPutter) Put(ctx context.Context, ch swarm.Chunk) error {
+	errs := []error{}
+	// Put base chunk first
+	if err := p.putter.Put(ctx, ch); err != nil {
+		return err
```
martinconic, line 40: We would need here to return the error with more context, or detail

pkg/api/bzz.go
gacevicljubisa: Maybe to remove DefaultLevel, and use PARANOID explcitly on every place. 
What do you think @janos @martinconic ?

janos: I agree to be more explicit internally in this case.

pkg/replicas/getter_soc.go
gacevicljubisa: I think yes. @nugaon explained that is what should be returned.

Issue #5005, opened 2025-02-13
Nodes metrics for speeds


nikipapadatou:
As part of our ongoing effort to improve upload and download speeds in the network, we need better visibility into network performance across different regions. Currently, our data comes mainly from nodes running in Europe, which gives us a limited perspective. To optimize performance globally, we need input from node operators across different locations (e.g., China, Americas, etc.).

We are asking Swarm node operators to voluntarily share key metrics related to upload/download speeds. This will help us:
* Identify regional performance bottlenecks
* Understand the impact of node distribution on retrieval times
* Improve Swarm’s global data availability

First, we need to define what are the key values that we want to get from the nodes. Operators who choose to participate should provide information around the following areas:
* Geographical region (continent or country)
* Upload speed (average, max, min)
* Download speed (average, max, min)
* Latency to other nodes (if measurable)
* Time taken to retrieve a specific file (X MB in size)
These values should be gathered over a period of at least a few days to account for variability in network conditions.

Node operators can opt-in to report the data and can opt-out anytime if they do not wish to share data.

istae:
To achieve the collection of metrics, we need the following:

1. Currently, metrics from by the Bee client is accumulated using a Bee endpoint, meaning the metrics are pulled from the client. We need a separate mechanism that will push the metrics to a metrics collector endpoint, similar to how metrics are pushed by Beekeepers runtime to a central collector.
2. The push mechanism can be enabled by a feature flag using the configuration options. The endpoint to which metrics will be pushed to should also be behind a config option, with the default pointing to a collector the Swarm org is running.
3. The Prometheus collector system needs to be set up by the infra team, and exposed via a single endpoint. 

For specific metrics, we need an investigation to see if the required metrics are already captured by the Bee client or we need to introduce new ones.

We also need to investigate how to capture and transmit geographical region data as parts of the metrics.

janos:
As an opt-in feature, I support it. The downside of the opt-in approach is that the data that is received may be biased by the sources that in general wishes to participate that may share some other common aspects. The opt-out approach gives more diverse data, but I would not advocate for it as it violates privacy in a non-transparent way. Many projects choose the approach based on their core values, and for Swarm I think that opt-in is better, but the best would be not to have it at all.

## Metrics

[...]

PR #5058, opened 2025-03-24
feat: add structured histogram to status proto for up/down speeds

janos:
### Checklist

### Description
This PR adds metrics to the status protocol for analysing upload and download speeds on different nodes.

#### Upload and download speed metrics

Upload and download metrics for speed are new ones so that they are measuring only the /bzz uploads and downloads, not /bytes or any other POST methods, making them more specific than content api duration metric which aggregates measurements from all endpoints into one metric.

##### Structured approach (Histogram type)

This approach uses the structured histogram. This comes with a few drawbacks. One is the complexity and more important one is the inability to add vectors in such way because of the restrictions in the prometheus client for directly getting all labels form a vector metric.

##### Alternative (string field with all metrics)

To overcome both drawbacks, approach where all metrics are sent in a form of a string field in prometheus format. This is more flexible (easier to expose metrics) and convenient, but less transparent and less pretty by my opinion.

It is up for a discussion on both approaches.

TODO: add new fields to the status response.

### Related Issue (Optional)
#5005

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -139,7 +139,7 @@ require (
 	github.com/pelletier/go-toml v1.8.0 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	github.com/prometheus/client_model v0.6.0 // indirect
+	github.com/prometheus/client_model v0.6.0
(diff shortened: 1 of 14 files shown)


janos:
Closing in favour to #5059.

PR #5081, opened 2025-04-23
fix: cascading pushsync requests

### LINKED ISSUE #5067: Out of control request cascading in pushsync in-neighborhood phase
### Context & Summary
The recently merged PR https://github.com/ethersphere/bee/pull/5037 introduced a serious issue in pushsync.

Before, the early store and return following "is in AOR" condition met prohibited in neighborhood nodes from doing splitting behaviour. However, the new "pushToClosest" call instead of returning immediately created an exponential cascading effect where every in neighborhood node creates 2x requests for each request received, leading to never terminating request cascades, ultimately leading to temporary disconnects within the neighborhood (caused by the continously timing out - hanging requests creating accounting issues as well)

### Expected behavior & suggestions
The splitting behaviour should only be excercised by a non-AOR node who pushes towards an in-AOR node. When the downstream peer is in-AOR, the node should check that itself is not in-AOR before doing splitting.
Alternatively, when doing pushToClosest after in-AOR is determined and store() is done, an extra argument could make sure that splitting doesnt happen in this case (so only 1x request is done towards the actual closest node in-neighborhood). Furthermore in this case it is important to make sure that the actual closest node doesn't keep pushing to the 2nd-closest etc, creating a non-exponential request cascade, by making sure errwantself is not suppressed in this case at any point in the function call chain.

acha-bill:
### Checklist

### Description
Multiplex only if node is non in AOR

### Related Issue (Optional)
fixes #5067

### DIFF
--- a/pkg/pushsync/pushsync.go
+++ b/pkg/pushsync/pushsync.go
@@ -459,7 +459,8 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
 
 			// since we can reach into the neighborhood of the chunk
 			// act as the multiplexer and push the chunk in parallel to multiple peers
-			if swarm.Proximity(peer.Bytes(), ch.Address().Bytes()) >= rad {
+			if swarm.Proximity(peer.Bytes(), ch.Address().Bytes()) >= rad &&
+				swarm.Proximity(ps.address.Bytes(), ch.Address().Bytes()) < rad {
 				for ; parallelForwards > 0; parallelForwards-- {
 					retry()
 					sentErrorsLeft++


pkg/pushsync/pushsync.go
```diff
@@ -459,7 +459,8 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
 
 			// since we can reach into the neighborhood of the chunk
 			// act as the multiplexer and push the chunk in parallel to multiple peers
-			if swarm.Proximity(peer.Bytes(), ch.Address().Bytes()) >= rad {
+			if swarm.Proximity(peer.Bytes(), ch.Address().Bytes()) >= rad &&
+				swarm.Proximity(ps.address.Bytes(), ch.Address().Bytes()) < rad {
```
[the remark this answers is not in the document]
acha-bill: Enhanced test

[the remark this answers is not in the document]
nikipapadatou: @zelig @significance your input about these changes please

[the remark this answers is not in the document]
significance: @acha-bill we discussed this a little in research sync, could you give us a little extra commentary on what's going on here and amend the comment please dude? just so we are sure we're all on the same page

pkg/pushsync/pushsync_test.go
```diff
@@ -860,15 +863,24 @@ func TestMultiplePushesAsForwarder(t *testing.T) {
...
+
+	// rad=0
+	// should not multiplex because nodes are in AOR. Expects 1 send to closestPeer only.
+	// should store chunk as nodes are in AOR
+	psPeer1, storerPeer1, _ := createPushSyncNode(t, peer1, defaultPrices, recorder0, nil, defaultSigner(chunk), mock.WithPeers(closestPeer, peer4))
+	psPeer2, storerPeer2, _ := createPushSyncNode(t, peer2, defaultPrices, recorder0, nil, defaultSigner(chunk), mock.WithPeers(closestPeer, peer4))
+	psPeer3, storerPeer3, _ := createPushSyncNode(t, peer3, defaultPrices, recorder0, nil, defaultSigner(chunk), mock.WithPeers(closestPeer, peer4))
+
+	// should not store chunk as no one forwards to it
+	psPeer4, storerPeer4, _ := createPushSyncNode(t, peer3, defaultPrices, nil, nil, defaultSigner(chunk))
```
nugaon: it should be peer4 instead of peer3

pkg/pushsync/pushsync_test.go
```diff
@@ -881,10 +893,13 @@ func TestMultiplePushesAsForwarder(t *testing.T) {
 		streamtest.WithBaseAddr(pivotNode),
 	)
 
-	psPivot, _, _ := createPushSyncNode(t, pivotNode, defaultPrices, recorder, nil, defaultSigner(chunk), mock.WithPeers(peer1, peer2, peer3))
+	// rad=2, pivot-peer1 po=3, pivot-chunk po=1
```
nugaon: po=1 with peer1 and peer2
po=2 with peer3

pkg/pushsync/pushsync_test.go
```diff
@@ -899,18 +914,78 @@ func TestMultiplePushesAsForwarder(t *testing.T) {
...
+		{
+			name:          "peer1",
+			ps:            psPeer1,
+			storer:        storerPeer1,
+			wantChunk:     true,
+			wantSentCount: 1,
+		},
+		{
+			name:          "peer2",
+			ps:            psPeer1,
```
nugaon: psPeer2

nugaon (commented):
a test case may be needed where peers pass the chunk to node with addr 71 and that is in connection with node 70 and in the same neighborhood. node 71 should forward the chunk to node 70 since that is the closest to the chunk address.

acha-bill:
> a test case may be needed where peers pass the chunk to node with addr 71 and that is in connection with node 70 and in the same neighborhood. node 71 should forward the chunk to node 70 since that is the closest to the chunk address.

It was already tested in the other PR.
I've brought those changes here now.

pkg/pushsync/pushsync.go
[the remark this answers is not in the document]
acha-bill: amended the descr

pkg/pushsync/pushsync_test.go
```diff
@@ -308,6 +310,123 @@ func TestShallowReceiptTolerance(t *testing.T) {
...
+	t.Parallel()
+	// chunk data to upload
+	chunk := testingc.FixtureChunk("7000")
+
+	triggerPeer := swarm.MustParseHexAddress("00")
+	pivotPeer := swarm.MustParseHexAddress("71")
+	closestPeer := swarm.MustParseHexAddress("70")
+	closestPeer1 := swarm.MustParseHexAddress("72")
+
+	psClosestPeer, closestPeerStorer, closestAccounting := createPushSyncNode(t, closestPeer, defaultPrices, nil, nil, defaultSigner(chunk), mock.WithClosestPeerErr(topology.ErrWantSelf))
```
nugaon, line 325: this ErrWantSelf mocking is strange, it should work without that by setting neighbor peers that may return this error.

pkg/pushsync/pushsync_test.go
```diff
@@ -308,6 +310,123 @@ func TestShallowReceiptTolerance(t *testing.T) {
...
+	_, closestPeer1Storer, _ := createPushSyncNode(t, closestPeer1, defaultPrices, nil, nil, defaultSigner(chunk))
+
+	recorder1 := streamtest.New(streamtest.WithProtocols(psClosestPeer.Protocol()), streamtest.WithBaseAddr(pivotPeer))
+
+	// pivot connects to closestpeer and closestpeer1 but should only forward once to closestpeer after storing
+	// closestpeer should not forward the chunk to closestpeer1 as it is already the closest peer
```
nugaon, line 331: anyway it cannot, because it does not have peers defined.

pkg/pushsync/pushsync.go
```diff
@@ -550,6 +564,7 @@ func (ps *PushSync) push(parentCtx context.Context, resultChan chan<- receiptRes
 		return
 	}
 
+	ps.logger.Debug("chunk sent", "chunk_address", ch.Address().String(), "peer_address", peer)
```
nugaon, line 567: this can be handy on manual test, also shallow receipt logs should be checked.

martinconic:
LGTM , but needs more testing

PR #5084, opened 2025-04-26
wasm

talentlessguy:
### Checklist


cmd/bee/cmd/start_unix.go
```diff
@@ -7,15 +7,131 @@
...
-	"github.com/ethersphere/bee/v2/pkg/log"
+	"github.com/ethersphere/bee/v2"
+	"github.com/ethersphere/bee/v2/pkg/node"
+	"github.com/spf13/cobra"
 )
 
-func isWindowsService() (bool, error) {
-	return false, nil
-}
+func (c *command) initStartCmd() (err error) {
```
sbackend123, line 24: In `cmd/start_windows.go` and `cmd/start_unix.go` most of the logic is the same, with the only differences being the runtime setup (Windows Service + Event Log vs. Unix signal handling).
Maybe we could extract the shared logic into a common file and add platform-specific adapters?

pkg/accounting/accounting_shared.go
```diff
@@ -0,0 +1,1106 @@
...
+// PayFunc is the function used for async monetary settlement
+type PayFunc func(context.Context, swarm.Address, *big.Int)
+
+// RefreshFunc is the function used for sync time-based settlement
+type RefreshFunc func(context.Context, swarm.Address, *big.Int)
+
+// Mutex is a drop in replacement for the sync.Mutex
+// it will not lock if the context is expired
+type Mutex struct {
+	mu chan struct{}
```
sbackend123, line 96: Why can't we use sync.Mutex here instead of channel? 
Looks like chan was added only for ctx expiration case, but I think it is possible to handle context expiration without creating our own version of mutex.
Or I miss smth?

pkg/node/chain.go
```diff
@@ -351,7 +351,7 @@ func (m noOpChainBackend) Metrics() []prometheus.Collector {
 }
 
 func (m noOpChainBackend) CodeAt(context.Context, common.Address, *big.Int) ([]byte, error) {
-	return common.FromHex(sw3abi.SimpleSwapFactoryDeployedBinv0_6_9), nil
+	return common.FromHex(sw3abi.SimpleSwapFactoryDeployedBinv0_6_5), nil
```
sbackend123, line 354: Why lower version here?

pkg/fs/fs_js.go
```diff
@@ -0,0 +1,572 @@
...
+	switch whence {
+	case io.SeekStart:
+		f.currOffset = offset
+		return f.currOffset, nil
+	case io.SeekCurrent:
+		f.currOffset += offset
+		return f.currOffset, nil
+	case io.SeekEnd:
+		f.currOffset = -offset
+		return f.currOffset, nil
```
sbackend123, line 169: Looks like potential problem to me. We return negative offset and no error... So method which calls this one has to keep in mind, that offset might be negative.

sbackend123 (commented):
Great job, but I would suggest:
1. Split this PR to a few separate.
2. PR has a lot of refactoring like split existing logic between *_js.go and *_shared.go
But it doesn't look like renaming, it looks like 2 new files, so it also makes review harder. In future I would suggest to avoid it. 
3. For me it is not quite clear, is the Bee browser build, is this actually a dial-only node? How is participation in overlay/DHT ensured without incoming: relay/signal/bootstrap policy? I understand, it should work somehow, but it would be great to have some doc
4. What is planned to replace Prometheus in the browser build?
5. how will the node behave if the peer does not support ws?

pkg/bzz/address.go
```diff
@@ -113,11 +113,14 @@ func (a *Address) Equal(b *Address) bool {
...
 func multiaddrEqual(a, b ma.Multiaddr) bool {
-	if a == nil || b == nil {
-		return a == b
+	switch {
+	case a == nil && b == nil:
+		return true
+	case a == nil || b == nil:
+		return false
+	default:
+		return a.Equal(b)
```
sbackend123, line 122: May be instead? Logic is same, but easier to read, I think
`if a == nil || b == nil {
    return a == nil && b == nil
}
return a.Equal(b)
`

pkg/node/chain.go
v1rtl: the branch is very out of date from upstream so this is just an older version

pkg/accounting/accounting_shared.go
v1rtl: this is copied from original bee source

cmd/bee/cmd/start_unix.go
v1rtl: you can't do platform-specific compilation by putting everything in one file. go build tags apply for the whole file

v1rtl:
> Great job, but have a few questions/suggestions:
> 
>     1. Split this PR to a few separate.
> 
>     2. PR has a lot of refactoring like split existing logic between *_js.go and *_shared.go
>        But it doesn't look like renaming, it looks like 2 new files, so it also makes review harder. In future I would suggest to avoid it.
> 
>     3. For me it is not quite clear, is the Bee browser build, is this actually a dial-only node? How is participation in overlay/DHT ensured without incoming: relay/signal/bootstrap policy? I understand, it should work somehow, but it would be great to have some doc
> 
>     4. What is planned to replace Prometheus in the browser build?
> 
>     5. how will the node behave if the peer does not support ws?

1. this branch is an experiment and isn't meant to be merged. there are some significant changes required for Bee in-browser to work properly, such as support for non-TCP addresses in overlays, until then there's no point in developing it further before that. Once all the requirements are met, I will create a new PR (or set of PRs) for proper wasm support.

2. I did it as a temporary hack to tell go to avoid compiling certain parts of code. In later PRs I will instead use `<file_native>.go` / `<file.go>` / `<file_js.go>`. This would also produce less merge conflicts

3. It's a dial only node yes. Can't listen on a TCP socket in a browser. Not sure what bootstrap policy is, the way it currently works is that it connects to a bootnode to obtain a list of peers, and runs a DHT in-browser to retrieve chunks. Other peers see browser node as `localhost` as if it would be running locally. When deployed on the web, those will instead be public IP addresses or website names.

4. Nothing. Prometheus is stripped off to decrease bundle size. A lot of metrics for a browser node are irrelevant, and you wouldn't gather metrics in-browser anyway.

5. It skips nodes with no WS available

v1rtl:
superseeded by #5300

PR #5094, opened 2025-05-08
feat: import batches from exported postage contract events log file

martinconic:
### Checklist


pkg/node/snapshot.go
```diff
@@ -0,0 +1,163 @@
...
+	"fmt"
+	"io"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/core/types"
+	archive "github.com/ethersphere/batch-archive"
+	"github.com/ethersphere/bee/v2/pkg/log"
+)
+
+type SnapshotBlockHeightContractFilterer struct {
```
gacevicljubisa, line 22: Maybe you could use more simple name, like `ArchivedLogFilterer` or `SnapshotLogFilterer`

pkg/node/snapshot.go
```diff
@@ -0,0 +1,163 @@
...
+			addressMatch := false
+			for _, addr := range query.Addresses {
+				if log.Address == addr {
+					addressMatch = true
+					break
+				}
+			}
+			if !addressMatch {
+				continue
+			}
```
gacevicljubisa, line 119: You can use `slices.Contains` to simpfily the code. Something like this:
```golang
if !slices.Contains(query.Addresses, log.Address) {
continue
}
```

go.mod
```diff
@@ -55,6 +57,8 @@ require (
 	resenje.org/web v0.4.3
 )
 
+require github.com/ethersphere/batch-archive v0.0.2
```
janos, line 60: https://github.com/ethersphere/batch-archive/tree/main/archive should have only the gzip file. Both files are not needed and uncompressed one is quite large.

pkg/node/snapshot.go
```diff
@@ -0,0 +1,153 @@
...
+
+	"slices"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/core/types"
+	archive "github.com/ethersphere/batch-archive"
+	"github.com/ethersphere/bee/v2/pkg/log"
+)
+
+type SnapshotLogFilterer struct {
```
janos, line 24: It would be good to add a line to check that SnapshotLogFilterer implements listener.BlockHeightContractFilterer interface, mainly for clarity.

pkg/node/snapshot.go
```diff
@@ -0,0 +1,153 @@
...
+	if err := f.loadAndProcessSnapshot(); err != nil {
+		return nil, fmt.Errorf("failed to load and process snapshot during initialization: %w", err)
+	}
+
+	return f, nil
+}
+
+func (f *SnapshotLogFilterer) loadAndProcessSnapshot() error {
+	f.logger.Info("loading batch snapshot during construction")
+	data := archive.GetBatchSnapshot(true)
```
janos, line 45: archive.GetBatchSnapshot should always provide gzip data to reduce executable binary size.

go.mod
gacevicljubisa, line 60: you can move this to in require block above

pkg/node/node.go
```diff
@@ -802,6 +802,32 @@ func NewBee(
 		}
 	)
 
+	if !o.Resync && !batchStoreExists && (networkID == mainnetNetworkID) {
```
gacevicljubisa, line 805: To be able to test, we should have or different build, or we could use `use-postage-snapshot` flag (set default to be true), but in this case we should remove the `bootstrapNode` logic.

go.mod
martinconic: Sure, well spotted, this was done automatically with `go mod tidy`, I did not wrote it by hand

pkg/node/node.go
```diff
@@ -802,6 +802,32 @@ func NewBee(
...
+		chainBackend, err := NewSnapshotLogFilterer(logger)
+		if err != nil {
+			logger.Debug("failed to initialize batch snapshot chain backend", "error", err)
+		} else {
+			eventListener := listener.New(b.syncingStopped, logger, chainBackend, postageStampContractAddress, postageStampContractABI, o.BlockTime, postageSyncingStallingTimeout, postageSyncingBackoffTimeout)
+
+			batchSvc, err := batchservice.New(stateStore, batchStore, logger, eventListener, overlayEthAddress.Bytes(), post, sha3.New256, o.Resync)
```
gacevicljubisa, line 812: maybe to rename `batchSvc` here to `snapshotBatchSvc`?

pkg/node/snapshot.go
```diff
@@ -0,0 +1,153 @@
...
+	maxBlockHeight uint64
+	isLoaded       bool
+}
+
+func NewSnapshotLogFilterer(logger log.Logger) (*SnapshotLogFilterer, error) {
+	f := &SnapshotLogFilterer{
+		logger: logger,
+	}
+
+	if err := f.loadAndProcessSnapshot(); err != nil {
```
gacevicljubisa, line 36: Not sure how much time needs for this, but it can be long running operation, and it is not best practice to have it in constructor. Maybe `sync.Once` can be used in FilterLogs to load all data using Lazy Loading principle.

pkg/node/snapshot.go
```diff
@@ -0,0 +1,153 @@
...
+		return fmt.Errorf("error scanning batch import data: %w", err)
+	}
+
+	f.loadedLogs = parsedLogs
+	f.maxBlockHeight = currentMaxBlockHeight
+	f.isLoaded = true
+	return nil
+}
+
+func (f *SnapshotLogFilterer) FilterLogs(ctx context.Context, query ethereum.FilterQuery) ([]types.Log, error) {
```
gacevicljubisa, line 98: FilterLogs is iterating in each step trhough complete list of parsedLogs. 
- Maybe it would be better that we return all logs in one step (we could modify the interface to have additional return value)
- Also, our exported events are stored in order, we could break the loop when BlockNumber goes out of the query range (export tool is ours)
- Additionaly, if we leave the query steps, we could sequentially parse in `FilterLogs`, not to store all data locally,

pkg/node/snapshot.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		filtered = append(filtered, log)
+	}
+
+	f.logger.Debug("filtered logs", "input_count", len(f.loadedLogs), "output_count", len(filtered))
+	return filtered, nil
+}
+
+func (f *SnapshotLogFilterer) BlockNumber(_ context.Context) (uint64, error) {
```
gacevicljubisa, line 151: underscore can be removed `BlockNumber(context.Context) (uint64, error)`

pkg/node/snapshot.go
martinconic: This was my first implementation, but it does not work due to how the listener is implemented using pagination and blocknumber calculation. For the current importer implementation I would not change how the listener is implemented

janos: This looks like a nice candidate for https://pkg.go.dev/iter#Seq2, if the time is not too short for such change. Why Seq2, because the second argument should be the error type.

pkg/node/snapshot.go
martinconic: I choose this way because it would be easier to test, to validate that the batches JSON file has valid batches logs. I would not want to have in FilterLogs at the same time, parsing the batches output file, validating the logs, and then filter them.

pkg/node/snapshot.go
gacevicljubisa: New version tagged https://github.com/ethersphere/batch-archive/releases/tag/v0.0.3 which only provides gzip data

pkg/node/node.go
```diff
@@ -802,6 +802,30 @@ func NewBee(
 		}
 	)
 
+	if !o.Resync && !batchStoreExists && (networkID == mainnetNetworkID) {
```
acha-bill, line 805: Does it make sense to have another option to bypass the snapshot?

pkg/node/snapshot.go
```diff
@@ -0,0 +1,168 @@
...
+	return f.initErr
+}
+
+func (f *SnapshotLogFilterer) FilterLogs(ctx context.Context, query ethereum.FilterQuery) ([]types.Log, error) {
+	if err := f.ensureLoaded(); err != nil {
+		return nil, fmt.Errorf("failed to ensure snapshot was loaded for FilterLogs: %w", err)
+	}
+	if !f.isLoaded {
+		return nil, fmt.Errorf("snapshot not loaded, cannot filter logs (initialization might have failed without an error)")
+	}
```
acha-bill, line 105: I think it'll be easier to read if these were bundled in 1 function `f.loadSnapshot() error`. You can use the `sync.Once` directly in loadsnapshot. No need for `isLoaded`

pkg/node/snapshot.go
```diff
@@ -0,0 +1,168 @@
...
+	return filtered, nil
+}
+
+func (f *SnapshotLogFilterer) BlockNumber(_ context.Context) (uint64, error) {
+	if err := f.ensureLoaded(); err != nil {
+		return 0, fmt.Errorf("failed to ensure snapshot was loaded for BlockNumber: %w", err)
+	}
+	if !f.isLoaded {
+		return 0, fmt.Errorf("snapshot not loaded, cannot get block number")
+	}
```
acha-bill, line 166: same here

pkg/node/snapshot.go
acha-bill: +1

pkg/node/node.go
martinconic: There is an option `UsePostageSnapshot` that was used trying to implement a similar solution a few years ago. It seems that either that flag and the previous solutions are not fully implemented and not working.  So I would not add a new option. Instead, in a separate PR(for better clarity) I would remove the old functionality and use the existing option `UsePostageSnapshot`

janos: I am also not in favour of reusing the existing `UsePostageSnapshot` option as it has historical context and it already is not functional given that the feed address for the postage snapshot on the swarm network is a 0 address, currently in the code.

I would suggest that the importing of embedded snapshot should be a default behaviour, not something that has to be enabled, but that an option to disable the import would be quite useful for "debugging" case, where someone wants to avoid importing for whatever reason (problem or investigation), to have a fallback to the current node behaviour related to the batches fetching. That option as a fallback should not be another complexity that users should care about in regular node usage.

gacevicljubisa (approved):

janos (approved):

PR #5097, opened 2025-05-13
fix: call recovery after cache eviction

nugaon:
### Checklist

### Description

During long-lived joiner operations, reconstructed chunks can be evicted from the local cache due to memory pressure. When these chunks are needed again, the current implementation doesn't attempt to recover them a second time. Instead, it falls back to direct network fetching, which fails with ErrNotFound if the chunk isn't available in the network.

This creates a reliability issue where successfully recovered chunks become inaccessible once evicted from cache, potentially breaking long-running operations like downloads or uploads of large files.

This PR implements a erasure redecoder functionality with the following components:

- ReDecoder: A new wrapper that attempts network fetch first, and only falls back to erasure recovery if the network fetch fails with ErrNotFound.
- Lazy Decoder Instantiation: Recovery decoders are only created on-demand when network fetch fails, saving resources.
- Memory-Efficient Caching: Maintains the existing memory optimization of nulling decoders after successful recovery while keeping track of successful recoveries.

### DIFF
--- a/pkg/file/joiner/joiner.go
+++ b/pkg/file/joiner/joiner.go
@@ -65,6 +65,22 @@ func fingerprint(addrs []swarm.Address) string {
 	return string(h.Sum(nil))
 }
 
+// createRemoveCallback returns a function that handles the cleanup after a recovery attempt
+func (g *decoderCache) createRemoveCallback(key string) func(error) {
+	return func(err error) {
+		g.mu.Lock()
+		defer g.mu.Unlock()
+		if err != nil {
+			// signals that a new getter is needed to reattempt to recover the data
+			delete(g.cache, key)
+		} else {
+			// signals that the chunks were fetched/recovered/cached so a future getter is not needed
+			// The nil value indicates a successful recovery
+			g.cache[key] = nil
+		}
+	}
+}
+
 // GetOrCreate returns a decoder for the given chunk address
 func (g *decoderCache) GetOrCreate(addrs []swarm.Address, shardCnt int) storage.Getter {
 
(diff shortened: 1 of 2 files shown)


pkg/file/joiner/joiner.go
```diff
@@ -83,22 +98,31 @@ func (g *decoderCache) GetOrCreate(addrs []swarm.Address, shardCnt int) storage.
...
+			// Create a factory function that will instantiate the decoder only when needed
+			recovery := func() storage.Getter {
+				g.config.Logger.Debug("lazy-creating recovery decoder after fetch failed", "key", key)
+				g.mu.Lock()
+				defer g.mu.Unlock()
+				d, ok := g.cache[key]
+				if ok && d != nil {
+					return d
+				}
+				d = getter.New(addrs, shardCnt, g.fetcher, g.putter, decoderCallback, g.config)
```
martinconic, line 114: I think here it is possible to have a deadlock. At this point, the same goroutine is holding the lock from line 96 and trying to acquire it again here in decoderCallback.

nugaon: decoderCallback is called in prefetch function only that is executed by a different go routine

martinconic (approved):
Is there a way to manually test this?

pkg/file/redundancy/getter/redecoder.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
```
martinconic: Please change the year

nugaon:
> Is there a way to manually test this?

The erasure data recovery should be triggered, for example, by leaving out some chunks for upload of an erasure coded chunk tree similarly as [beekeeper does](https://github.com/ethersphere/beekeeper/blob/master/pkg/check/redundancy/redundancy.go#L171-L181). Then, attempt to retrieve the data and see what happens in case of a Bee which does not include the changes and the one which has it. The bee instances should be started with lowered size cache parameters e.g. `--cache-capacity=1`.

gacevicljubisa (approved):

PR #5107, opened 2025-06-02
fix: remove sleep and refactor code

gacevicljubisa:
### Checklist

### Description
Remove time sleep on node start and improve the code in node shutdown.

### DIFF
--- a/cmd/bee/cmd/start.go
+++ b/cmd/bee/cmd/start.go
@@ -68,7 +68,6 @@ func (c *command) initStartCmd() (err error) {
 			}
 
 			fmt.Print(beeWelcomeMessage)
-			time.Sleep(5 * time.Second)
 			fmt.Printf("\n\nversion: %v - planned to be supported until %v, please follow https://ethswarm.org/\n\n", bee.Version, endSupportDate())
 			logger.Info("bee version", "version", bee.Version)
 
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -106,7 +106,7 @@ type Bee struct {
 	pullSyncCloser           io.Closer
 	pssCloser                io.Closer
 	gsocCloser               io.Closer
-	ethClientCloser          func()
+	ethClientCloser          io.Closer
 	transactionMonitorCloser io.Closer
 	transactionCloser        io.Closer
 	listenerCloser           io.Closer
@@ -394,7 +394,7 @@ func NewBee(
 	if err != nil {
 		return nil, fmt.Errorf("init chain: %w", err)
 	}
-	b.ethClientCloser = chainBackend.Close
+	b.ethClientCloser = chainBackend
 
 	logger.Info("using chain with network network", "chain_id", chainID, "network_id", networkID)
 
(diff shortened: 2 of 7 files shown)


pkg/node/node.go
```diff
@@ -1335,42 +1341,27 @@ func (b *Bee) Shutdown() error {
...
+		{b.pusherCloser, "pusher"},
+		{b.pullerCloser, "puller"},
+		{b.accountingCloser, "accounting"},
+		{b.pullSyncCloser, "pull sync"},
+		{b.hiveCloser, "hive"},
+		{b.saludCloser, "salud"},
+	}
+
+	for _, nc := range closers {
+		wg.Add(1)
```
acha-bill: Add to wg before the loop

acha-bill (commented):
Nice. Once small comment

cmd/bee/cmd/start.go
```diff
@@ -68,7 +68,6 @@ func (c *command) initStartCmd() (err error) {
 			}
 
 			fmt.Print(beeWelcomeMessage)
-			time.Sleep(5 * time.Second)
```
janos, line 71: I was thinking, if the purpose for this sleep is to have beeWelcomeMessage more visible to the users, it would make sense to have this sleep only if the bee is in the terminal. This is just a suggestion, to use `term.IsTerminal(int(os.Stdout.Fd()))` instead. But only if there is an actual need to have such wait.

pkg/node/node.go
gacevicljubisa: > Add to wg before the loop

You mean:
```go
wg.Add(len(closers))
```
?

acha-bill: yes

cmd/bee/cmd/start.go
acha-bill: In the spirit of enhancing the node spinup time, I think we can remove it. - even in terminal.

gacevicljubisa: I vote to remove `time.Sleep`..

@niki ?

nikipapadatou: it's not pretty clear to me what the sleep time has been used for. if its presence affects the node spinup, then sure, remove it. But we need to consider all possible cases it might have an effect on.

martinconic: The sleep time was introduced as part of the tribute to `ldeffenb` , so we can have some seconds to read  the quote. I think we can remove now the sleep.

pkg/node/node.go
```diff
@@ -1393,10 +1384,7 @@ func (b *Bee) Shutdown() error {
 
 	wg.Wait()
 
-	if c := b.ethClientCloser; c != nil {
-		c()
-	}
-
+	tryClose(b.ethClientCloser, "eth client")
 	tryClose(b.accesscontrolCloser, "accesscontrol")
```
martinconic, line 1388: Why did you not include all Closers into namedCloser ?

gacevicljubisa: With this refactoring, goal was to have the same execution order and reduce the amount of code in same time.

martinconic (approved):

PR #5109, opened 2025-06-02
fix(salud): use exponential backoff for wake up of salud check

gacevicljubisa:
### Checklist

### Description
Idea is to reduce the time that salud spends to update the peer status. This will help the light node spin time and readiness.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/ethersphere/bee/v2
 
 go 1.24.0
 
-toolchain go1.24.2
-
 require (
 	contrib.go.opencensus.io/exporter/prometheus v0.4.2
 	github.com/armon/go-radix v1.0.0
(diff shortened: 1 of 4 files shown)


janos (approved):

martinconic:
Please can you explain more on how this helps light nodes spin time and readiness and if there are other benefits from these changes? In the end after exponential backoff reaches 5 minutes, it will stay at 5 minutes as it is now.

gacevicljubisa:
> Please can you explain more on how this helps light nodes spin time and readiness and if there are other benefits from these changes? In the end after exponential backoff reaches 5 minutes, it will stay at 5 minutes as it is now.

Salud has own wakup time and it is hardcoded to 5 minutes. In the first run it can not find any peers. Beacuse of that, node needs to wait next run to be responsive. Idea for the expontial Backoff is to reduce the wait time. And later continue working on every 5 minutes.

Issue #5142, opened 2025-07-02
Gas fee limit for storage incentive transactions


gacevicljubisa:
#### Summary
Add a new setting in the Bee client to let node operators set a maximum gas fee for storage incentive transactions in the redistribution game (commit, reveal, claim phases). It can be an array with a limit for each phase, with default values that make sense.

#### Motivation
Node operators in the Swarm network join the redistribution game to earn rewards, but gas fees can get high and make it too expensive. Setting a gas fee limit for each phase helps operators control costs and avoid surprises, especially when gas prices spike.

#### Implementation
Add a new setting called `max-gas-fee-per-phase` to the Bee config file.
Example:
```
max-gas-fee-per-phase:
  - commit: 20000000000  # 20 Gwei for commit
  - reveal: 20000000000  # 20 Gwei for reveal
  - claim: 30000000000   # 30 Gwei for claim
```

#### Drawbacks
If gas fees go above the limit, operators might miss out on rewards because transactions won’t go through.

nugaon:
I think it would be useful on other blockchain transactions as well such as purchasing postage batches.
I'm unsure about how important is to have different limits on phases, I would just use one value for all.
additionally, retry mechanism could be added when the transaction fails because of hitting the limitation.

gacevicljubisa:
For the purchase of postage batches, there is a `Gas-Price` header, which allows user to specify the value. But it doeesn't represent limit. So, I agree, we can use it there as well, or maybe on all transactions together?

gacevicljubisa:
Regarding retry mechanism, for redistribution game, we could retry until phase ends?

0xCardiE:
I am ok with multiple values, lets say users could go more heavy on Claim, or at least until it makes sense in $.
As long as there is sane defaults so users can have both, simple and advanced if they want to thinker.

Definitely for having at least one default limit so users never pay above value that they could gain from claim.

gacevicljubisa:
@0xCardiE how can we calculate the possible reward before the claim phase? Maybe we shouldn't play the game if the possibile transaction expanses can be higher then reward?

0xCardiE:
3 ways you can get this from postageStamp contract

<img width="675" height="196" alt="Image" src="https://github.com/user-attachments/assets/675e50a6-3e0f-4d10-94eb-0fc5d6b3d427" />

Last one would be most precise. Some code here to get it with JS

```

class AccuratePotCalculator {
[...]

PR #5144, opened 2025-07-03
fix: replace SuggestGasPrice with custom SuggestedFeeAndTip

gacevicljubisa:
### Checklist

### Description
- Removes deprecated `SuggestGasPrice` from codebase and replaces it with logic used in `suggestedFeeAndTip` method.
- Adds `minimum-gas-tip-cap` flag, minimum gas tip cap in wei for transactions, default 0 means use suggested gas tip cap.
- Adds unit tests for `SuggestedFeeAndTip` in wrapped package.
- Ensures when user use `Gas-Price` header, that that value is used for the gas fee. 
- Removes unused functions from `wrapped` package

### DIFF
--- a/pkg/transaction/wrapped/fee.go
+++ b/pkg/transaction/wrapped/fee.go
@@ -1,4 +1,4 @@
-// Copyright 2021 The Swarm Authors. All rights reserved.
+// Copyright 2025 The Swarm Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 


pkg/transaction/wrapped/wrapped.go
```diff
@@ -16,7 +17,16 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/transaction"
 )
 
-var _ transaction.Backend = (*wrappedBackend)(nil)
+const (
+	MinimumGasTipCap  = 1_500_000_000 // 1.5 Gwei
+	PercentageDivisor = 100
+	BaseFeeMultiplier = 2
```
janos: I would keep these constants in the `transaction` package as they are general. That would avoid importing a specific implementation of `wrapped` package, like in the test file, which is now dependent on the `wrapped` package.

janos (commented):
I have only one comment related to inner-dependencies.

pkg/transaction/wrapped/wrapped.go
gacevicljubisa: `MinimumGasTipCap` is moved to `transaction` pkg. `PercentageDivisor` and  `BaseFeeMultiplier`, are changed to unexported.

janos: Great, thanks.

janos (approved):

pkg/transaction/wrapped/wrapped.go
```diff
@@ -219,3 +217,50 @@ func (b *wrappedBackend) Close() error {
...
+	}
+
+	minimumTip := big.NewInt(transaction.MinimumGasTipCap)
+	if gasTipCap.Cmp(minimumTip) < 0 {
+		gasTipCap.Set(minimumTip)
+	}
+
+	var gasFeeCap *big.Int
+
+	if gasPrice == nil {
```
martinconic: Can `gasPrice == 0` ?

gacevicljubisa: Yes, when the `Gas-Price` header is used and explicitly set to 0.
This would effectively create a transaction with zero gas fees, which would likely fail when submitted to the network since miners require some compensation for processing transactions.

martinconic (approved):

pkg/transaction/wrapped/wrapped.go
```diff
@@ -219,3 +217,50 @@ func (b *wrappedBackend) Close() error {
...
+		gasFeeCap = new(big.Int).Mul(latestBlockHeader.BaseFee, big.NewInt(int64(baseFeeMultiplier)))
+		gasFeeCap.Add(gasFeeCap, gasTipCap)
+
+	} else {
+		gasFeeCap = new(big.Int).Set(gasPrice)
+	}
+
+	// gasTipCap cannot exceed gasFeeCap
+	if gasTipCap.Cmp(gasFeeCap) > 0 {
+		gasTipCap.Set(gasFeeCap)
```
nugaon: I would just throw an error here because gasFeeCap (`maxFeePerGas`) includes the gasTipCap, so you either set it (maxFeePerGas - baseFeePerGas) and risk you'll have a stucked transaction or you raise gasFeeCap and you overwrote the passed gassPrice that was set by the user. Neither of them are good.

pkg/transaction/wrapped/wrapped.go
```diff
@@ -16,7 +17,15 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/transaction"
 )
 
-var _ transaction.Backend = (*wrappedBackend)(nil)
+const (
+	percentageDivisor = 100
+	baseFeeMultiplier = 2
```
nugaon: so this is basically the "max gas base fee volatility ratio allowed from the previous block".

IMHO this does not really put a real fee cap on the transactions, it more like prevents sending transactions on spikes. the operator can spend a lot if the network has already congestion.

A simpler solution like a max fee cap from the user side would be better (with default relying completely on the fee suggestions).

pkg/transaction/transaction.go
```diff
@@ -41,14 +41,12 @@ var (
...
 	ErrAlreadyImported     = errors.New("already imported")
-	ErrEIP1559NotSupported = errors.New("network does not appear to support EIP-1559 (no baseFee)")
 )
 
 const (
-	DefaultGasLimit               = 1_000_000
-	DefaultTipBoostPercent        = 25
-	MinimumGasTipCap              = 1_500_000_000 // 1.5 Gwei
-	RedistributionTipBoostPercent = 50
+	MinimumGasTipCap       = 1_500_000_000 // 1.5 Gwei
```
nugaon: this is really arbitrary,I would make it configurable too with defaulting only to the suggestions.

martinconic:
Is there a way we can manually test this? How did you tested it?

pkg/transaction/wrapped/wrapped.go
gacevicljubisa: I agree that we should have max fee cap, this is suggested in this [issue](https://github.com/ethersphere/bee/issues/5142) and it could be nice improvement for the current logic.

gacevicljubisa: `max gas base fee volatility ratio allowed from the previous block` should be `12.5%`, and this [post](https://github.com/ethereum/pm/issues/328#issuecomment-853234014) is the reason why I am using `baseFeeMultiplier = 2`

pkg/transaction/transaction.go
gacevicljubisa: To have it as a additiional config flag, but default should be suggested gas tip from `eth_maxPriorityFeePerGas`?

pkg/transaction/wrapped/wrapped.go
nugaon: nice source, but I referred to something else, namely it does not solve the problem for spending too much on transactions. let's say there is a price at block X and during 50 consecutive blocks (250 sec) all maxing out the 12.5% raising in each round there will be 361 times more fee compare to X.

gacevicljubisa:
> Is there a way we can manually test this? How did you tested it?

You can execute any transaction manually, for example create a postage batch. 
Also, I have added unit tests.

pkg/transaction/backend/backend.go
```diff
@@ -0,0 +1,34 @@
...
+	"context"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+// Backend is the interface that an ethclient.Client satisfies.
+type Backend interface {
```
janos: It is a good idea to have the interface backend.Backend as restricted to ethclient.Client methods, but there may be already an interface that does that but in a rather questionable package https://github.com/ethereum/go-ethereum/blob/v1.16.1/ethclient/simulated/backend.go#L39 that is meant for simulations. Fortunately, the linked interface is referencing other interfaces. Maybe to reference here the same interfaces as in the *simulated* package. Just a suggestion.

On the naming, maybe to have different type names for backend.Backend and transaction.Backend as the later is not ethclient.Client compatible, just for clarity, even thou, the package name is different.

pkg/transaction/wrapped/wrapped.go
```diff
@@ -215,7 +215,46 @@ func (b *wrappedBackend) BlockByNumber(ctx context.Context, number *big.Int) (*t
 	return block, nil
 }
 
-func (b *wrappedBackend) Close() error {
+func (b *wrappedBackend) Close() {
```
nugaon, line 186: just my 2 cents,
changing this does not seem necessary - even though ethclient.Client implements the `Close()` method like this. 
I think it is better to use Bee common interfaces for similar functionalities such as `Close()` and if it does not return error (atm) then just return nil as it was before.
by that the code seems more consistent.

pkg/transaction/transaction_test.go
```diff
@@ -75,6 +75,39 @@ func signerMockForTransaction(t *testing.T, signedTx *types.Transaction, sender
 	)
 }
 
+func checkStoredTransaction(t *testing.T, transactionService transaction.Service, txHash common.Hash, request *transaction.TxRequest, recipient common.Address, gasLimit uint64, gasPrice *big.Int, nonce uint64) {
```
nugaon, line 78: nice refactor!

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -17,7 +17,10 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/transaction"
 )
 
-const loggerName = "redistributionContract"
+const (
+	loggerName      = "redistributionContract"
+	BoostTipPercent = 50
```
nugaon, line 22: in a future PR, this could be dynamic based on how far the node is from the end of the round phase. wdyt?

nugaon (approved):

acha-bill (approved):

pkg/storageincentives/redistribution/redistribution.go
gacevicljubisa: That is nice idea

pkg/transaction/backend/backend.go
gacevicljubisa: Maybe to rename `backend.Backend` to `backend.Geth`?

gacevicljubisa: And regarding the ethereum interfaces, if we add those, we need to implement 11 additional methods in wrapped package:

```go
BlockByHash(ctx context.Context, hash common.Hash) (*types.Block, error)
HeaderByHash(ctx context.Context, hash common.Hash) (*types.Header, error)
PendingBalanceAt(ctx context.Context, account common.Address) (*big.Int, error)
PendingCodeAt(ctx context.Context, account common.Address) ([]byte, error)
PendingStorageAt(ctx context.Context, account common.Address, key common.Hash) ([]byte, error)
PendingTransactionCount(ctx context.Context) (uint, error)
StorageAt(ctx context.Context, account common.Address, key common.Hash, blockNumber *big.Int) ([]byte, error)
SubscribeFilterLogs(ctx context.Context, q ethereum.FilterQuery, ch chan<- types.Log) (ethereum.Subscription, error)
SubscribeNewHead(ctx context.Context, ch chan<- *types.Header) (ethereum.Subscription, error)
TransactionCount(ctx context.Context, blockHash common.Hash) (uint, error)
TransactionInBlock(ctx context.Context, blockHash common.Hash, index uint) (*types.Transaction, error)
```

`backend.Get` interface would be like this:

```go
type Geth interface {
	ethereum.BlockNumberReader
	ethereum.ChainReader
	ethereum.ChainStateReader
	ethereum.ContractCaller
	ethereum.GasEstimator
	ethereum.GasPricer1559
	ethereum.LogFilterer
	ethereum.PendingStateReader
	ethereum.TransactionReader
	ethereum.TransactionSender
	ethereum.ChainIDReader
	Close()
}
```

Maybe to add `panic("unimplemented")` to each one of those wer are not using?

pkg/transaction/wrapped/wrapped.go
gacevicljubisa: Constructor `wrapped.NewBackend` needs to accept `ethclient.Client`, which implements `Close()`, so it was easier to remove error.

pkg/transaction/backend/backend.go
janos: Eleven unused methods seems to extensive just to use existing interfaces. It is a sign that separation of methods in upstream interfaces do not fit our usage so well. Thank you for looking into this. Feel free to leave it as it is in this pr.

janos: > Maybe to rename `backend.Backend` to `backend.Geth`?

Yes, that sounds ok.

Issue #5137, opened 2025-07-01
Investigate health check in salud and fix possiible issues


gacevicljubisa:
It seems that `if` [statement ](https://github.com/ethersphere/bee/blob/d996ebe5b79e57166d0c4bf1c01df2f3bafb1694/pkg/salud/salud.go#L231-L235) doesn't properly updates the health status of the node peer.

Health in this place is if the peer is in a bin that has good population. Health of the peer cannot be determined. Naming in this part of the code could be a bit better.

nugaon:
in the code that I found is [kademlia's pruneoversaturatedBins](https://github.com/ethersphere/bee/blob/master/pkg/topology/kademlia/kademlia.go#L674) uses the health value of the peer.

It only disconnects the minimum number of peers needed to get the bin back to an acceptable size -> so unhealthy nodes can still remain there, but those nodes will be prioritized to be removed instead of a random one. It means that removing this part can even improve on the topology by keeping healthy nodes and correct metric data.

gacevicljubisa:
@janos any opinion on this? what can be improved regarding naming?

janos:
While looking deeper at the usage of the Healthy field, the mentioned `if` statement is "correct" **indirectly**, but the usage of the Health field is *abused* in a sense that [pruneoversaturatedBins](https://github.com/ethersphere/bee/blob/master/pkg/topology/kademlia/kademlia.go#L674), that @nugaon nicely pointed out, is disconnecting peers that have Healthy bool false, so if the number of peers in the bin where a particular peer is is less than minPeersPerbin, we are setting the Health to true in order not to disconnect it. 

My opinion is that this approach (if I did not miss something) is confusing as the Health of the peer should reflect the Health status, not only to be the flag that marks if the peer should be disconnected at some point. At least the name Healthy is not the best one in that context.

But, I think that this if statement should be just removed completely from the salud() function and that it is up to the pruneoversaturatedBins to decide which peers should be disconnected also regarding the bin population, which is already done by binPruneCount function when it is set as PruneCountFunc in kadOptions.

nugaon:
> if the number of peers in the bin where a particular peer is is less than minPeersPerbin, we are setting the Health to true in order not to disconnect it.

but it does not disconnect from all nodes with false healthy status, just until `pruneCount` (as you also referred later in your response), so it does not have any sense to mark it true and potentially remove a normal healthy node from the connections.

[...]

PR #5145, opened 2025-07-04
fix: remove setting peer healthy true at undersaturated bins

nugaon:
related issue: #5137 

### Checklist

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -975,7 +975,7 @@ func NewBee(
 		return nil, fmt.Errorf("status service: %w", err)
 	}
 
-	saludService := salud.New(nodeStatus, kad, localStore, logger, detector, api.FullMode.String(), salud.DefaultMinPeersPerBin, salud.DefaultDurPercentile, salud.DefaultConnsPercentile)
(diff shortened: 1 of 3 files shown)


janos (approved):
I think that this change is logically and technically correct. But, I would wait for merging after the bee 2.6.0 release.

gacevicljubisa:
> I think that this change is logically and technically correct. But, I would wait for merging after the bee 2.6.0 release.

Also, lets run this on light-testnet first.

gacevicljubisa:
Should the log msg be different here:
https://github.com/ethersphere/bee/blob/15120e83cb0ef905d69b40f88f6a2a445e6fa3c7/pkg/salud/salud.go#L208C1-L209C1

Maybe: `response duration above threshold`?

gacevicljubisa:
Here are the [metrics ](https://grafana.proxy.ethswarm.org/d/ddvvzigc3591ce/health-service?var-interval=$__auto&orgId=1&from=now-24h&to=now&timezone=browser&var-datasource=default&var-cluster=swarm1&var-namespace=bee-light-testnet&var-pod=$__all) from `bee-light-testnet`. Also, when NetworkRadius=1 HealthRate was 0.

image

PR #5148, opened 2025-07-08
chore: investigate libp2p.shareTCPListener

acha-bill:


martinconic:
As I debugged Bee is that [here in go-libp2p library](https://github.com/libp2p/go-libp2p/blob/7b7c3ed4ce048c99a865d4349b5f7484a6a93c04/config/config.go#L297) `ShareTCPListener` is `false`

the issue caused by the fact that we are introducing this ShareTCPListener() in here:
libp2p.go - file
```
    opts := []libp2p.Option{
        libp2p.ShareTCPListener(),
        libp2p.ListenAddrStrings(listenAddrs...),
        security,
        // Use dedicated peerstore instead the global DefaultPeerstore
        libp2p.Peerstore(libp2pPeerstore),
        libp2p.UserAgent(userAgent()),
        libp2p.ResourceManager(rm),
    }
```
if we introduce it  in the transports:
```
    transports := []libp2p.Option{
        libp2p.Transport(tcp.NewTCPTransport),
        libp2p.ShareTCPListener(),
    }
```
seems we are not getting a `false` value in the go-libp2p library

janos:
Hi, after investigations that @martinconic did with the AutoTLS that is required for WSS support for lbp2p transports, libp2p.ShareTCPListener() does not help us a lot as WSS requires a dedicated port for its own transport. I am suggesting not to use ShareTCP listener because of this requirement as it is not the solution for WSS and also as we encountered some issues while using it (reachability with AutoNAT was always private).

We should instead support multiple underlay addresses (with different ports) in bee handshake and hive protocols as well as in addressbook and kademlia.

martinconic:
> Hi, after investigations that @martinconic did with the AutoTLS that is required for WSS support for lbp2p transports, libp2p.ShareTCPListener() does not help us a lot as WSS requires a dedicated port for its own transport. I am suggesting not to use ShareTCP listener because of this requirement as it is not the solution for WSS and also as we encountered some issues while using it (reachability with AutoNAT was always private).
> 
> We should instead support multiple underlay addresses (with different ports) in bee handshake and hive protocols as well as in addressbook and kademlia.

I agree, and also think there is no need and use for ShareTCP listener in Bee right now

martinconic:
Closing based on above comments

Issue #5143, opened 2025-07-02
Redistribution handler panics when called with storage incentives disabled


Cafe137:
### Context

```
2.6.0-rc7-027599b3a
```

```
storage-incentives-enable: false
```

```
curl "http://localhost:1633/redistributionstate"
```

### Expected behavior
Would prefer a `HTTP 4xx` stating that this feature is disabled

### Actual behavior
```
"time"="2025-07-02 14:44:29.243619" "level"="error" "logger"="node" "msg"="http: panic serving 127.0.0.1:56707: runtime error: invalid memory address or nil pointer dereference
    goroutine 1351823 [running]:
    net/http.(*conn).serve.func1()
    \tnet/http/server.go:1947 +0xb0
    panic({0x101e12e60?, 0x1030826d0?})
    \truntime/panic.go:792 +0x124
    github.com/ethersphere/bee/v2/pkg/storageincentives.(*Agent).Status(...)
    \tgithub.com/ethersphere/bee/v2/pkg/storageincentives/agent.go:556
    github.com/ethersphere/bee/v2/pkg/api.(*Service).redistributionStatusHandler(0x140005e0388, {0x10203dc50, 0x1400476fd10}, 0x14000c64140)
    \tgithub.com/ethersphere/bee/v2/pkg/api/redistribution.go:41 +0xbc
    net/http.HandlerFunc.ServeHTTP(0x1400daf33b8?, {0x10203dc50?, 0x1400476fd10?}, 0x12a0ed078?)
    \tnet/http/server.go:2294 +0x38
    github.com/ethersphere/bee/v2/pkg/jsonhttp.HandleMethods(0x14001eb2ed0, {0x101848f3b, 0x2b}, {0x10182fe47, 0x1f}, {0x10203dc50, 0x1400476fd10}, 0x14000c64140)
    \tgithub.com/ethersphere/bee/v2/pkg/jsonhttp/handlers.go:25 +0xe4
    github.com/ethersphere/bee/v2/pkg/jsonhttp.MethodHandler.ServeHTTP(0x102040d58?, {0x10203dc50?, 0x1400476fd10?}, 0x101c81548?)
    \tgithub.com/ethersphere/bee/v2/pkg/jsonhttp/handlers.go:17 +0x50
[...]

PR #5150, opened 2025-07-09
fix(api): resolve panic when storage incentives disabled

gacevicljubisa:
…point

### Checklist

### Description
Adds handler that checks if redistributionAgent is initialized.

### Related Issue (Optional)
Redistribution handler panics when called with storage incentives disabled [#5143](https://github.com/ethersphere/bee/issues/5143)


acha-bill (approved):

janos (approved):

martinconic (approved):

pkg/api/router.go
```diff
@@ -208,6 +208,16 @@ func (s *Service) checkChequebookAvailability(handler http.Handler) http.Handler
 	})
 }
 
+func (s *Service) checkStorageIncentivesAvailability(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if s.redistributionAgent == nil {
+			jsonhttp.NotImplemented(w, "Storage incentives are disabled. This endpoint is unavailable.")
```
nugaon: well, it is implemented, it is more like a `404 Not Found` or a `403 Forbidden`. I would simply go with 404

nugaon (commented):
seems good just my 2 cents

pkg/api/router.go
gacevicljubisa: We have used this status code when `Swap is disabled` and `Chequebook is disabled`, which was done in this [PR](https://github.com/ethersphere/bee/pull/4869). But we can change it to `403 Forbidden`, because route does exit, but it is disabled on server side. In that case we should change this for all 3 endpoints to have the same behaviour. 
@janos, @nugaon any opion on this?

nugaon: 403 is also fine from my side on all similar places

gacevicljubisa: done

nugaon: tests still need to be adjusted: https://github.com/ethersphere/bee/actions/runs/16437874357/job/46453779979?pr=5150

PR #5159, opened 2025-07-16
fix: cancel tx if missing from blockchain after some timeout

### LINKED ISSUE #5133: When transaction is not existing (rpc issue), transaction watch doesn't delete it on time

acha-bill:
### Checklist

### Description
Cancel a tx if it's missing from the blockchain after a certain timeout—currently, 5 block times.

closes #5133

### DIFF
--- a/pkg/transaction/monitor.go
+++ b/pkg/transaction/monitor.go
@@ -21,8 +21,8 @@ import (
 var ErrTransactionCancelled = errors.New("transaction cancelled")
 var ErrMonitorClosed = errors.New("monitor closed")
 
-// MaxBlocksToWait is the maximum number of blocks to wait before considering a missing transaction as cancelled.
-const MaxBlocksToWait = 5
+// maxBlocksToWait is the maximum number of blocks to wait before considering a missing transaction as cancelled.
+const maxBlocksToWait = 5
 
 // Monitor is a nonce-based watcher for transaction confirmations.
 // Instead of watching transactions individually, the senders nonce is monitored and transactions are checked based on this.
@@ -202,7 +202,7 @@ func (tm *transactionMonitor) checkPending(block uint64) error {
 			receipt, err := tm.backend.TransactionReceipt(tm.ctx, txHash)
 			if err != nil {
 				// wait for a few blocks to be mined before considering a transaction not existing
-				transactionWatchNotFoundTimeout := MaxBlocksToWait * tm.pollingInterval
+				transactionWatchNotFoundTimeout := maxBlocksToWait * tm.pollingInterval
 				if errors.Is(err, ethereum.NotFound) {
 					if watchStart(watches).Add(transactionWatchNotFoundTimeout).Before(time.Now()) {
 						cancelledNonces = append(cancelledNonces, nonceGroup)
--- a/pkg/transaction/export_test.go
+++ b/pkg/transaction/export_test.go
@@ -6,4 +6,5 @@ package transaction
 
 var (
 	StoredTransactionKey = storedTransactionKey
+	MaxBlocksToWait      = maxBlocksToWait
 )


martinconic:
Is there a way we can test this?

martinconic:
Maybe we should add one more test for when the transaction appears later, similar to this:
```
	t.Run("missing transaction appears later", func(t *testing.T) {
		t.Parallel()

		monitor := transaction.NewMonitor(
			logger,
			backendsimulation.New(
				backendsimulation.WithBlocks(
					backendsimulation.Block{
						Number: 0,
					},
					backendsimulation.Block{
						Number: 1,
					},
					backendsimulation.Block{
						Number: 2,
					},
					backendsimulation.Block{
						Number: 3,
						Receipts: map[common.Hash]*types.Receipt{
							txHash: {TxHash: txHash},
						},
					},
				),
			),
			sender,
			pollingInterval,
			cancellationDepth,
		)

		receiptC, errC, err := monitor.WatchTransaction(txHash, nonce)
		if err != nil {
			t.Fatal(err)
		}

		select {
		case receipt := <-receiptC:
			if receipt.TxHash != txHash {
				t.Fatal("got wrong receipt")
			}
		case err := <-errC:
			t.Fatal(err)
		case <-time.After(testTimeout):
			t.Fatal("timed out")
		}

		err = monitor.Close()
		if err != nil {
			t.Fatal(err)
		}
	})

```

pkg/transaction/monitor_test.go
```diff
@@ -388,4 +388,41 @@ func TestMonitorWatchTransaction(t *testing.T) {
...
+		case <-time.After(testTimeout):
+			t.Fatal("timed out")
+		}
+
+		err = monitor.Close()
+		if err != nil {
+			t.Fatal(err)
+		}
+	})
 }
```
martinconic, line 428: Maybe we should one more test similar with this one, when transaction appears later:
```
t.Run("missing transaction appears later", func(t *testing.T) {
		t.Parallel()

		monitor := transaction.NewMonitor(
			logger,
			backendsimulation.New(
				backendsimulation.WithBlocks(
					backendsimulation.Block{
						Number: 0,
					},
					backendsimulation.Block{
						Number: 1,
					},
					backendsimulation.Block{
						Number: 2,
					},
					backendsimulation.Block{
						Number: 3,
						Receipts: map[common.Hash]*types.Receipt{
							txHash: {TxHash: txHash},
						},
					},
				),
			),
			sender,
			pollingInterval,
			cancellationDepth,
		)

		receiptC, errC, err := monitor.WatchTransaction(txHash, nonce)
		if err != nil {
			t.Fatal(err)
		}

		select {
		case receipt := <-receiptC:
			if receipt.TxHash != txHash {
				t.Fatal("got wrong receipt")
			}
		case err := <-errC:
			t.Fatal(err)
		case <-time.After(testTimeout):
			t.Fatal("timed out")
		}

		err = monitor.Close()
		if err != nil {
			t.Fatal(err)
		}
	})
```

nugaon: the original test is when the transaction does not appear on chain at all

pkg/transaction/monitor.go
```diff
@@ -21,6 +21,9 @@ import (
 var ErrTransactionCancelled = errors.New("transaction cancelled")
 var ErrMonitorClosed = errors.New("monitor closed")
 
+// MaxBlocksToWait is the maximum number of blocks to wait before considering a missing transaction as cancelled.
+const MaxBlocksToWait = 5
```
janos: This variable does not need to be exported. It can be exported for tests only in export_test.go.

gacevicljubisa:
> Is there a way we can test this?

Maybe we can test it mannualy with insufficient gas, invalid signature, low nonce, or insufficient funds.

mfw78:
In general, I'm not sure about this mechanism. If there has been a transaction that has been signed, and broadcast to some RPC, while that RPC may have some failures of some sort, the `bee` node has no way to intrinsically verify this.

As such, transactions should never be deemed "cancelled" unless in the circumstance that another transaction has been verified on chain that corresponds to the cancelled transaction's `nonce`.

For example, if the transaction monitor sent to the RPC a transaction, and after some time, when checking the account's `nonce` from the RPC it hadn't incremented (ie. the transaction hadn't been confirmed), then it seems that the "failed transaction" could instead be determined to be "stale" and the transaction monitor could simply broadcast another transaction using the same `nonce` (and if this subsequent transaction is verified, then the stale transaction would change state to 'dropped' / 'cancelled').

Issue #5157, opened 2025-07-15
Support Swarm-Feed-Legacy-Resolve query param in `GET /bzz/<hash>/` endpoint


Cafe137:
#### Summary
Support `Swarm-Feed-Legacy-Resolve` query param in `GET /bzz/<hash>/` endpoint

#### Motivation
This is needed to be able to open legacy feed manifests.

#### Implementation
N/A

#### Drawbacks
N/A

Cafe137:
Test vector which opens in 2.5 but not 2.6:

- http://localhost:1633/bzz/637036f7f8e7d9a52b8710bf54d1dd0030ced72a6bd26fea959d4bdf48609b8a/
- http://localhost:1633/bzz/2fdac7bd34a7e27e46e3fd2b6c67f13e3f04231ef92226b8d64a5f693ec80cae/

Issue #5156, opened 2025-07-15
Make Swarm-Feed-Legacy-Resolve query param instead of header


Cafe137:
#### Summary
Make Swarm-Feed-Legacy-Resolve query param instead of header

#### Motivation
I should be able to share a feed URL such that when somebody opens it in their browser, it just works.
Right now this is not possible because it requires setting a header which requires technical knowledge in the browsers.

#### Implementation
N/A

#### Drawbacks
N/A

acud:
@Cafe137 this is already done no? I think this could be closed

PR #5163, opened 2025-07-21
feat: feed legacy payload header

### LINKED ISSUE #5155: OpenAPI yaml -- Wrong: `legacy-feed-resolution`; Correct: `Swarm-Feed-Legacy-Resolve`
### Location
`GET /feeds/{owner}/{topic}`

### Description
A new header flag has been added to resolve feeds in the legacy way. The OpenAPI yaml wrongly calls this header `legacy-feed-resolution`, but in the source code it is `Swarm-Feed-Legacy-Resolve`

nugaon:
Resolves #5155 #5156 #5157

### Checklist

### DIFF
--- a/pkg/api/bzz_test.go
+++ b/pkg/api/bzz_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/manifest"
 	mockbatchstore "github.com/ethersphere/bee/v2/pkg/postage/batchstore/mock"
 	mockpost "github.com/ethersphere/bee/v2/pkg/postage/mock"
+	testingsoc "github.com/ethersphere/bee/v2/pkg/soc/testing"
 	"github.com/ethersphere/bee/v2/pkg/storage/inmemchunkstore"
 	mockstorer "github.com/ethersphere/bee/v2/pkg/storer/mock"
 	"github.com/ethersphere/bee/v2/pkg/swarm"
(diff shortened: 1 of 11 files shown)


pkg/api/bzz_test.go
```diff
@@ -1144,3 +1188,14 @@ func TestBzzDownloadHeaders(t *testing.T) {
 		jsonhttptest.WithExpectedResponseHeader(api.ContentTypeHeader, "text/html; charset=utf-8"),
 	)
 }
+
+// func toLegacyChunk(t *testing.T, at uint64, payload []byte) swarm.Chunk {
```
acha-bill: Remove unused

acha-bill (approved):
lgtm

janos (commented):
As this is a user facing feature, what would happen when the current feed structure becomes legacy. Since the structure can be changed and it is something that users interfere with over the api, I would suggest to have versioning on the feed, at least for api options (header in this case).

nugaon:
the current feed usage wraps the root chunk of the referred content so there is no specific structure in the feed payload.

there is no proposal now to change the arbitrary legacy feed structure but if that's the case in the future, API could be versioned with that change as you mentioned - except the header part since it has problematic use-case #5157. maybe instead of a boolean, a version number can be used. but at the moment, I don't see why it would be necessary since there is only one version of structure of legacy feeds and most probably there won't be new one.

janos:
> maybe instead of a boolean, a version number can be used. but at the moment, I don't see why it would be necessary since there is only one version of structure of legacy feeds and most probably there won't be new one.

Exactly, instead of boolean, a version would make changes in the future much easier. With the boolean, a transition to some new header would be needed. In general, we do not know how things will change in the future, but we can prepare to handle changes. Since the change already happened creating legacy feeds, it is possible to have another change in the future.

nugaon:
IMHO starting to version a legacy feature is futile and makes the code and the documentation more complex.
The whole point is to transition out of payload structures in feeds.

janos:
> IMHO starting to version a legacy feature is futile and makes the code and the documentation more complex. The whole point is to transition out of payload structures in feeds.

OK, as you wish, my opinion is different as described. We will handle the backward incompatible changes later, if they appear for this cli flag.

gacevicljubisa:
If we use something like `?swarm-feed-format=legacy` instead, it would be much easier to add other formats in the future if we need to. It's a small change now that could save us from a headache later on.

Other than that, the rest of the changes in the PR look great to me

nugaon:
@gacevicljubisa the problem is the same from my perspective. also, just imagine you need to remember of the query name AND the "legacy" as well. if I needed to choose from any one them I would go with the simple versioning since that is simpler from this aspect.

gacevicljubisa:
> @gacevicljubisa the problem is the same from my perspective. also, just imagine you need to remember of the query name AND the "legacy" as well. if I needed to choose from any one them I would go with the simple versioning since that is simpler from this aspect.

I agree it adds a little more complexity, but either way, it's another custom parameter that users will have to look up in the documentation.

Since they'll be checking the docs regardless, I think the small extra step of typing `=legacy` (or something similar that will allow versioning) is a good trade-off. It makes the API more flexible for the future.

But if you insist, we can go with the current approach.

sbackend123 (approved):

PR #5167, opened 2025-07-23
refactor: sampling optimalizations

nugaon:
This PR optimizes the `ReserveSample` function to improve performance during sampling operations. The key optimizations include adding `ChunkType ` to the `SampleItem` struct to avoid redundant parsing, reusing hashers in worker goroutines instead of creating new ones for each SOC chunk and using new optimized [hasher function](https://github.com/ethersphere/crypto) for transformed address calculations. These changes reduce memory allocations, GC pressure, and CPU usage without altering the functional behavior of the sampling process.

### Checklist

### DIFF
--- a/pkg/storer/sample.go
+++ b/pkg/storer/sample.go
@@ -66,7 +66,6 @@ func (db *DB) ReserveSample(
 	consensusTime uint64,
 	minBatchBalance *big.Int,
 ) (Sample, error) {
-
 	g, ctx := errgroup.WithContext(ctx)
 
 	allStats := &SampleStats{}
@@ -86,7 +85,8 @@ func (db *DB) ReserveSample(
 
 	allStats.BatchesBelowValueDuration = time.Since(t)
 
-	chunkC := make(chan *reserve.ChunkBinItem)
+	workers := max(4, runtime.NumCPU())
+	chunkC := make(chan *reserve.ChunkBinItem, 3*workers)
 
 	// Phase 1: Iterate chunk addresses
 	g.Go(func() error {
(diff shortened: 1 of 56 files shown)


pkg/storer/sample.go
```diff
@@ -184,7 +189,7 @@ func (db *DB) ReserveSample(
 		close(sampleItemChan)
 	}()
 
-	sampleItems := make([]SampleItem, 0, SampleSize)
+	sampleItems := make([]SampleItem, 0, SampleSize+1)
```
martinconic, line 191: why +1 ?

nugaon: lower there is a condition when `if len(sampleItems) > SampleSize` it cuts from the end. so sampleItems used length is SampleSize+1.

pkg/swarm/hasher.go
```diff
@@ -15,24 +15,33 @@ func NewHasher() hash.Hash {
 	return sha3.NewLegacyKeccak256()
 }
 
+// NewHasher returns new Keccak-256 hasher.
```
martinconic: Please edit comment accordingly.

pkg/swarm/hasher.go
```diff
@@ -15,24 +15,33 @@ func NewHasher() hash.Hash {
...
 type PrefixHasher struct {
-	hash.Hash
-	prefix []byte
+	sha3.StatefulHash
+	prefix       []byte
+	initialState []byte
 }
 
 // NewPrefixHasher returns new hasher which is Keccak-256 hasher
 // with prefix value added as initial data.
```
martinconic: Please edit comment accordingly.

pkg/swarm/hasher.go
```diff
@@ -15,24 +15,33 @@ func NewHasher() hash.Hash {
...
 func NewPrefixHasher(prefix []byte) hash.Hash {
```
martinconic, line 31: This also called in proof.go. Do we want that one to use the stateful hasher?

pkg/swarm/hasher.go
```diff
@@ -15,24 +15,33 @@ func NewHasher() hash.Hash {
 	return sha3.NewLegacyKeccak256()
 }
 
+// NewHasher returns new Keccak-256 hasher.
+func NewStatefulHasher(state []byte) (sha3.StatefulHash, error) {
```
martinconic, line 19: Should new tests be added in hasher_test for this new stateful hasher?

pkg/swarm/hasher.go
nugaon: yeah, it is used in the sample creation as well and both should do the same calculation.

pkg/swarm/hasher.go
nugaon: I made [some test in its repo](https://github.com/ethersphere/crypto/blob/master/sha3/state_test.go)

acud:
closing this as we are cleaning up the open pull requests, and other changes are inbound for this topic. if there are any improvements here, we will implement them in subsequent PRs.

PR #5175, opened 2025-07-29
test: sampling chunk fetch time

nugaon:
### Checklist

### DIFF
--- a/pkg/storer/sample.go
+++ b/pkg/storer/sample.go
@@ -132,8 +132,7 @@ func (db *DB) ReserveSample(
 
 	db.logger.Debug("reserve sampler workers", "count", workers)
 
-	for i := 0; i < workers; i++ {
-		workerID := i
+	for range workers {
 		g.Go(func() error {
 			wstat := SampleStats{}
 			hasher := bmt.NewHasher(prefixHasherFactory)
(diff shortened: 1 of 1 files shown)


pkg/storer/sample.go
```diff
@@ -121,15 +132,37 @@ func (db *DB) ReserveSample(
 
 	db.logger.Debug("reserve sampler workers", "count", workers)
 
-	for range workers {
+	for i := 0; i < workers; i++ {
```
martinconic: Or you could just write `for i := range workers {...}`

pkg/storer/sample.go
```diff
@@ -121,15 +132,37 @@ func (db *DB) ReserveSample(
...
 				addStats(wstat)
+				addWaitTime(waitTime)
+
+				db.logger.Debug("sampling worker statistics",
+					"worker_id", workerID,
+					"wait_time", waitTime,
+				)
 			}()
 
 			for chItem := range chunkC {
```
martinconic, line 151: In this loops we want to measure only the waiting for a chunk or the total worker execution time? It seems that it measures the second one.

pkg/storer/sample.go
nugaon: it was a leftover from previous subprocess metrics (which are not included anymore), so I removed together with its logging

pkg/storer/sample.go
nugaon: correct. we need to lower this overall time.

martinconic (approved):

PR #5177, opened 2025-08-01
feat(storer): add comprehensive metrics for ReserveSample performance

gacevicljubisa:
### Checklist

### DIFF
--- a/pkg/storer/metrics.go
+++ b/pkg/storer/metrics.go
@@ -33,6 +33,17 @@ type metrics struct {
 	ExpiryRunsCount         prometheus.Counter
 
 	ReserveMissingBatch prometheus.Gauge
+
+	// ReserveSample metrics
+	ReserveSampleDuration           *prometheus.HistogramVec
+	ReserveSampleChunksIterated     prometheus.Counter
+	ReserveSampleChunksLoaded       prometheus.Counter
+	ReserveSampleChunksLoadFailed   prometheus.Counter
+	ReserveSampleTaddrDuration      *prometheus.HistogramVec
+	ReserveSampleStampValidations   prometheus.Counter
+	ReserveSampleStampValidDuration *prometheus.HistogramVec
+	ReserveSampleSize               prometheus.Gauge
+	ReserveSampleWorkers            prometheus.Gauge
 }
 
 // newMetrics is a convenient constructor for creating new metrics.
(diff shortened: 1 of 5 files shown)


janos (approved):

pkg/storer/sample.go
```diff
@@ -241,13 +254,18 @@ func (db *DB) ReserveSample(
...
+				db.metrics.ReserveSampleStampValidDuration.WithLabelValues("invalid").Observe(time.Since(stampValidStart).Seconds())
 				db.logger.Debug("invalid stamp for chunk", "chunk_address", ch.Address(), "error", err)
 				continue
 			}
 
-			stats.ValidStampDuration += time.Since(start)
+			stampValidDuration := time.Since(stampValidStart)
+			stats.ValidStampDuration += stampValidDuration
+			db.metrics.ReserveSampleStampValidations.Inc()
+			db.metrics.ReserveSampleStampValidDuration.WithLabelValues("valid").Observe(stampValidDuration.Seconds())
```
nugaon: seconds is too low resolution

pkg/storer/sample.go
```diff
@@ -267,6 +285,9 @@ func (db *DB) ReserveSample(
 
 	db.logger.Info("reserve sampler finished", "duration", time.Since(t), "storage_radius", committedDepth, "consensus_time_ns", consensusTime, "stats", fmt.Sprintf("%+v", allStats))
 
+	// Record final sample size
+	db.metrics.ReserveSampleSize.Set(float64(len(sampleItems)))
```
nugaon: it does not seem a valuable metric. this should be always 16 if stats.TotalIterated >= 16

pkg/storer/sample.go
```diff
@@ -104,6 +115,7 @@ func (db *DB) ReserveSample(
 			select {
 			case chunkC <- ch:
 				stats.TotalIterated++
+				db.metrics.ReserveSampleChunksIterated.Inc()
```
nugaon: it could be more performant if stats.totaliterated was just sent in the end.

pkg/storer/sample.go
```diff
@@ -149,18 +160,22 @@ func (db *DB) ReserveSample(
 				chunk, err := db.ChunkStore().Get(ctx, chItem.Address)
 				if err != nil {
 					wstat.ChunkLoadFailed++
+					db.metrics.ReserveSampleChunksLoadFailed.Inc()
```
nugaon: same as at TotalIterated

pkg/storer/sample.go
```diff
@@ -149,18 +160,22 @@ func (db *DB) ReserveSample(
...
 					db.logger.Debug("failed loading chunk", "chunk_address", chItem.Address, "error", err)
 					continue
 				}
 
 				wstat.ChunkLoadDuration += time.Since(chunkLoadStart)
+				db.metrics.ReserveSampleChunksLoaded.Inc()
```
nugaon: this is TotalIterated - (BelowBalanceIgnored + RogueChunk + ChunkLoadFailed). There are metrics for all of them.

pkg/storer/sample.go
```diff
@@ -149,18 +160,22 @@ func (db *DB) ReserveSample(
...
 
 				taddrStart := time.Now()
 				taddr, err := transformedAddress(hasher, chunk, chItem.ChunkType)
 				if err != nil {
 					return err
 				}
-				wstat.TaddrDuration += time.Since(taddrStart)
+				taddrDuration := time.Since(taddrStart)
+				wstat.TaddrDuration += taddrDuration
+				db.metrics.ReserveSampleTaddrDuration.WithLabelValues(chItem.ChunkType.String()).Observe(taddrDuration.Seconds())
```
nugaon: `Seconds` resolution is too low. Miliseconds or nanoseconds would be better

pkg/storer/sample.go
```diff
@@ -241,13 +254,18 @@ func (db *DB) ReserveSample(
...
 				stats.InvalidStamp++
+				db.metrics.ReserveSampleStampValidDuration.WithLabelValues("invalid").Observe(time.Since(stampValidStart).Seconds())
 				db.logger.Debug("invalid stamp for chunk", "chunk_address", ch.Address(), "error", err)
 				continue
 			}
 
-			stats.ValidStampDuration += time.Since(start)
+			stampValidDuration := time.Since(stampValidStart)
+			stats.ValidStampDuration += stampValidDuration
+			db.metrics.ReserveSampleStampValidations.Inc()
```
nugaon: it is `stats.SampleInserts`

nugaon (commented):
IMO that would be better if it sent metrics in batches wherever it is possible and avoid from redundant metric data.

pkg/storer/sample.go
gacevicljubisa: A better approach would be to remove TotalIterated since it uses a mutex. Using Inc() on a counter with atomic operations is much faster than a mutex. The overhead of writing to an atomic variable is negligible compared to the total amount of current work.

nugaon: `TotalIterated` is incremented without mutex, the mutex is used only at the end of the goroutine when it is added to the common, aggregating. Not sure which one is really the faster (atomic operations have hardware level contention still and there are many increments) but one of them should be kept only since they are the same.

pkg/storer/sample.go
```diff
@@ -437,3 +445,30 @@ func getChunkType(chunk swarm.Chunk) swarm.ChunkType {
 	}
 	return swarm.ChunkTypeUnspecified
 }
+
+func (db *DB) recordReserveSampleMetrics(duration time.Duration, stats *SampleStats, workers int, err error) {
+	status := "success"
+	if err != nil {
+		status = "failure"
+	}
+	db.metrics.ReserveSampleDuration.WithLabelValues(status, fmt.Sprintf("%d", workers)).Observe(duration.Seconds())
```
nugaon: below you send workers as a separate metric.

pkg/storer/sample.go
```diff
@@ -147,20 +152,23 @@ func (db *DB) ReserveSample(
...
-				wstat.ChunkLoadDuration += time.Since(chunkLoadStart)
+				wstat.ChunkLoadDuration += chunkLoadDuration
 
 				taddrStart := time.Now()
 				taddr, err := transformedAddress(hasher, chunk, chItem.ChunkType)
 				if err != nil {
 					return err
 				}
-				wstat.TaddrDuration += time.Since(taddrStart)
+				taddrDuration := time.Since(taddrStart)
```
nugaon: you created a new variable for it but you don't use that anywhere else than in the below line.

pkg/storer/metrics.go
```diff
@@ -163,6 +165,33 @@ func newMetrics() metrics {
...
+		ReserveSampleRunSummary: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Namespace: m.Namespace,
+				Subsystem: subsystem,
+				Name:      "reserve_sample_run_summary",
+				Help:      "Summary metrics for the last ReserveSample run.",
+			},
+			[]string{"metric"},
+		),
+		ReserveSampleLastRunTimestamp: prometheus.NewGauge(
```
nugaon, line 187: why is it important? we have metrics for reserve sampling duration and there is a approximate time for it when crawling happens of stat metrics.

pkg/storer/metrics.go
```diff
@@ -163,6 +165,33 @@ func newMetrics() metrics {
 				Help:      "Number of times the expiry worker was fired.",
 			},
 		),
+		ReserveSampleDuration: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Namespace: m.Namespace,
+				Subsystem: subsystem,
+				Name:      "reserve_sample_duration_seconds",
+				Help:      "Duration of ReserveSample operations in seconds.",
+				Buckets:   []float64{30, 60, 120, 300, 600, 900, 1200, 1500, 1800, 2400, 3000, 3600, 4800},
```
nugaon: the range is really large here, 30 seconds to 80 minutes... none of them seems realistic.

gacevicljubisa: From 3min to 30min as upper limit?

pkg/storer/metrics.go
gacevicljubisa: With the exact finish timestamp, we can calculate the precise start time and detect delays in the game timing. This enables monitoring whether sampling starts on schedule within the expected phase. It also simplifies time-based calculations without needing to parse logs.

PR #5187, opened 2025-08-11
feat(p2p): add AutoTLS support for secure WebSocket connections

martinconic:
### Checklist

- [ ] I have read the [coding guide](https://github.com/ethersphere/bee/blob/master/CODING.md).
- [ ] My change requires a documentation update, and I have done it.
- [ ] I have added tests to cover my changes.
- [ ] I have filled out the description and linked the related issues.

### Description
This pull request introduces an AutoTLS implementation for the Bee node. The primary motivation for this feature is to support the In-Browser project, which requires secure communication channels for clients.

Modern web browsers strictly enforce secure contexts, meaning they can only connect to endpoints using secure protocols like HTTPS or Secure WebSockets (wss://). AutoTLS automates the process of obtaining, managing, and renewing valid TLS certificates from a Certificate Authority (like Let's Encrypt).

By implementing AutoTLS, Bee nodes can now accept wss:// connections directly, enabling browsers to interact with the Bee network securely and seamlessly.

**Implementation Details**
The implementation is based on the approach outlined in the official [libp2p blog post on AutoTLS](https://blog.libp2p.io/autotls/) and follows the patterns from the [go-libp2p AutoTLS example](https://github.com/libp2p/go-libp2p/blob/master/examples/autotls/main.go).

**Testing & Validation**
The functionality was manually tested by running a Bee node (built with the flag above) and attempting to connect to it using wscat over a secure websocket.

**Test Command:**
`wscat --no-check -c wss://<node-public-ip>.<peer-id>.libp2p.direct:5500`

**Result:**
A successful connection was established, and the node responded with the multistream header, confirming the wss listener is active and secured with a valid TLS certificate:

```
Connected (press CTRL+C to quit)
< /multistream/1.0.0
```

### Related Issue (Optional)
[#5171](https://github.com/ethersphere/bee/issues/5171)

PR #5190, opened 2025-08-17
Feat: gossip extra webrtc underlays

lat-murmeldjur:
### Checklist

### Description

This PR attempts to implement SWIP-24 with the difference of using WebRTC instead of Websockets to enable browser based nodes connecting to the swarm network.

It includes an addition to the handshake protocol with added webrtc underlay fields in protobuf, saving the received webrtc underlays for each peer, as well as added webrtc underlay fields per bzzaddress in gossip.  For development purposes right now all ipv4 webrtc underlays are attached to handshake / communicated in gossip, however further refinement can be done to filter out private webrtc addresses from the communication, or to restrict the number of communicated webrtc underlays to 1 per node.

The reason WebRTC seems to be more feasible for browser connections is the fact that secure websocket connections are required when the browser based node is loaded in HTTPS (which is a requirement for service workers that enable displaying swarm hosted websites in a browser based node), however most browsers do not accept untrusted self signed certificates from swarm nodes for tls/ws or wss. WebRTC seems to work better from across different platforms and browsers, not requiring domain based trusted security certificates from each bee node. This enables connecting to webrtc underlays for example from android chrome.


v1rtl:
why not have both ws and wrtc?

lat-murmeldjur:
> why not have both ws and wrtc?

We can, just that from the browser-based-node perspective websocket seems less utilizable because of a number of factors:
- A browser based node requires a service worker to be able to load assets on relative paths (for example a swarm hosted website)
- A service worker requires https
- Most browsers only allow websocket connections from a https site to wss or tls/ws
- A bee node that wants to provide a proper trusted wss or tls/ws address would require a domain name, because ip based certificates are untrusted, and don't seem to work well with the browser websocket API that does not provide a way to accept these certificates, it seems

The exception to this seemed to be shared web workers in chrome, which also give a warning about how connecting to unsecured ws endpoints is deprecated, but seemed not to enforce this so far, which could be a bug.

v1rtl:
the restriction regarding secure contexts for websocket will soon go away with #5187

lat-murmeldjur:
> the restriction regarding secure contexts for websocket will soon go away with #5187

that would be great, but like how is the bee node going to have a domain name?
the problem I see is that in hive protocol we gossip ip based multiunderlays, not domain name based ones.
and when the browser API tries to connect to this ip based websocket address, there's no way to issue a trustable certificate for that ip address, and also no way to accept lots of untrusted certificates in the browser (one for each connection) like you would in server side go-libp2p for example with InsecureSkipVerify

mfw78:
In general, I'm not sure why there is a need for a specific WebRTC underlays field, or any specific underlays field and not just have the the existing underlay field store the multiaddr. The "repurposing" of this is suitable as the version of the handshake protocol would have to be bumped nonetheless, and all version bumps are breaking - so doing it in the manner proposed in this PR would not preserve any backwards compatibility.

lat-murmeldjur:
> In general, I'm not sure why there is a need for a specific WebRTC underlays field, or any specific underlays field and not just have the the existing underlay field store the multiaddr. The "repurposing" of this is suitable as the version of the handshake protocol would have to be bumped nonetheless, and all version bumps are breaking - so doing it in the manner proposed in this PR would not preserve any backwards compatibility.

im not sure i understand your point correctly, and i dont mean to argue, but in general, protobuf3 is backwards compatible as long as you dont change existing fields and their numbers, you can add new fields with higher numbers, as all fields are optional

of course this can be solved in many ways, i simply added a new field for simplicity, so the bee node does not need to cherry pick which protocol to use from a mixed bunch, and the browser based node can find what it needs in the new field

lat-murmeldjur:
> In general, I'm not sure why there is a need for a specific WebRTC underlays field, or any specific underlays field and not just have the the existing underlay field store the multiaddr. The "repurposing" of this is suitable as the version of the handshake protocol would have to be bumped nonetheless, and all version bumps are breaking - so doing it in the manner proposed in this PR would not preserve any backwards compatibility.

another thing i experienced that might make this approach more feasible, is that when a bee node connects to another bee node on a webrtc address, it uses a different random port than it uses for listening, and this random port is not dialable, so if we start saving these new connectout-only underlays with the advertised connectable overlays, the gossip will get polluted by these unreachable addresses. so for my PR we only save/gossip the webrtc underlays of a peer communicated by it in the handshake, which are dialable (however they might be private addresses in this PR)

v1rtl:
> > the restriction regarding secure contexts for websocket will soon go away with #5187
> 
> that would be great, but like how is the bee node going to have a domain name?

autotls gives you one as long as you're publicly accessible

> the problem I see is that in hive protocol we gossip ip based multiunderlays, not domain name based ones.

that's no longer relevant soon. bee will support multiple underlays.

> and when the browser API tries to connect to this ip based websocket address, there's no way to issue a trustable certificate for that ip address

there is, see AutoTLS

> and also no way to accept lots of untrusted certificates in the browser (one for each connection) like you would in server side go-libp2p for example with InsecureSkipVerify

that's not needed. AutoTLS handles everything

lat-murmeldjur:
> there is, see AutoTLS
> 
> >
will you need to buy a domain for each bee node, or does autoTLS hand out free domain names?

v1rtl:
> > there is, see AutoTLS
> > >
> 
> will you need to buy a domain for each bee node, or does autoTLS hand out free domain names?

no, you don't buy anything. you're issued a [peerid].libp2p.direct domain with https

ref: https://blog.libp2p.io/autotls/

PR #5194, opened 2025-08-19
feat(swip25): pull syncing optimization

nugaon:
This SWIP describes a more efficient way to synchronise content between peers in the same neighbourhood by syncing each chunk only once.
It removed the maxPOdelta exception for syncing bins from peers below storage depth. Accordingly, bee should not allow shallow receipts on pushing chunks. That should be tackled in a different PR. 

[Forwarding chunks to the closest address on the network](https://github.com/ethersphere/bee/pull/5081) is also required since chunks will be pulled from its closest full nodes. 

This is the implementation of [SWIP 25](https://github.com/ethersphere/SWIPs/pull/66) for details check the proposal.

### Checklist

### DIFF
--- a/pkg/puller/puller.go
+++ b/pkg/puller/puller.go
@@ -132,6 +132,7 @@ func (p *Puller) manage(ctx context.Context) {
 
 	var prevRadius uint8
 
+	// change in radius or in neighborhood peerset
 	changeCheck := func() {
 		p.syncPeersMtx.Lock()
 		defer p.syncPeersMtx.Unlock()


pkg/puller/puller.go
```diff
@@ -152,32 +150,52 @@ func (p *Puller) manage(ctx context.Context) {
 			}
 			p.logger.Debug("radius decrease", "old_radius", prevRadius, "new_radius", newRadius)
 		}
+		changed := newRadius != prevRadius
```
martinconic, line 154: can you rename please this to something like `radiusChanged` or `rChanged`. It would be easier to figure out in the code what is this boolean all about.

pkg/puller/tree.go
```diff
@@ -0,0 +1,177 @@
...
+	C [2]*TreeNode[T]
+}
+
+// Put will override the value if the key is already present
+func (t *TreeNode[T]) Put(key []byte, p *T) *TreeNode[T] {
+	bitIndex, err := bitOfBytes(key, t.L)
+	if err != nil {
+		// cannot go further in the binary representation, override leaf
+		if t.V != nil && !bytes.Equal(t.K, key) {
+			panic(errShouldNeverHappen)
```
martinconic, line 30: Do we want panics or should we use errors? Same bellow

pkg/puller/tree.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"github.com/ethersphere/bee/v2/pkg/swarm"
+)
+
+var errShouldNeverHappen = errors.New("should never happen in case of keys have the same length, bits of those are matching until the starting level and method was called from root")
+
+// TreeNode is a leaf compacted binary tree
+// representing the address space of the neighborhood
+type TreeNode[T any] struct {
+	K []byte
```
martinconic, line 18: Are there cases were we would need concurrency suport via a mutex?

pkg/puller/tree.go
```diff
@@ -0,0 +1,177 @@
...
+// peerTreeNode is a specialized TreeNode for managing sync peers
+type peerTreeNode struct {
+	*TreeNode[peerTreeNodeValue]
+}
+
+// BinAssignment assigns bins to each peer for syncing by traversing the tree
+func (t *peerTreeNode) BinAssignment() (peers []*peerTreeNodeValue) {
+	if t.isLeaf() {
+		bl := uint8(len(t.V.SyncBins))
+		for i := t.L; i < bl; i++ {
```
martinconic, line 130: What if t.L > bl ?

pkg/puller/puller.go
```diff
@@ -152,32 +150,52 @@ func (p *Puller) manage(ctx context.Context) {
...
+				syncPeer = newSyncPeer(addr, p.bins, po)
+				p.syncPeers[addr.ByteString()] = syncPeer
+				if po >= newRadius {
+					changed = true
+				}
+			} else {
+				syncPeer.syncBins = make([]bool, p.bins)
+			}
+			if po >= newRadius {
+				_ = bt.Put(addr.Bytes(), &peerTreeNodeValue{SyncBins: syncPeer.syncBins})
```
martinconic, line 174: Why we do not have also here `changed = true` ?

pkg/puller/puller.go
nugaon: it is related to that the parameters for the pulling have changed that partially includes the radius change but also when a new peer is conencted to the node and that is a neighbor or a neighbor node is disconnected.

pkg/puller/puller.go
nugaon: because at this point the pullsync parameters may have not changed to reinit syncings but you want to build up the tree with the neighbor peers in case there was a change in the neighborhood peerset eventually.

pkg/puller/tree.go
nugaon: the tree must be used with the prerequisites defined in the `errShouldNeverHappen`. If there is an error, no recovery should or can happen on these special cases indicating a significant bug in the logic. in such case, it seems better to panic.

pkg/puller/tree.go
nugaon: there is no concurrent use of TreeNode at the moment so it does not require any concurrency handling for now

PR #5199, opened 2025-08-27
feat(bzz): underlay list serialization functions

janos:
### Checklist

### Description
This PR implements bzz.SerializeUnderlay and bzz.DeserializeUnderlay to be used in places where serialization of multiple underlay addresses is needed, like in Handshake and Hive protocols.

Both functions are backward compatible, but multiaddr.NewMultiaddrBytes will fail deliberately and consistently if it receives a serialized list instead a single multiaddr address serialized bytes.

This PR does not change bzz.Address.Underlay to be a slice of addresses as that involves a larger scope of changes, but only provides the serialization tools for handling multiple underlay addresses.

Support for multiple underlay addresses for different node types (in-browser, bee server) to choose from. Issue [#5201](https://github.com/ethersphere/bee/issues/5201)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -133,7 +133,7 @@ require (
 	github.com/multiformats/go-multiaddr-fmt v0.1.0 // indirect
 	github.com/multiformats/go-multibase v0.2.0 // indirect
 	github.com/multiformats/go-multicodec v0.9.0 // indirect
-	github.com/multiformats/go-varint v0.0.7 // indirect
+	github.com/multiformats/go-varint v0.0.7
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/onsi/ginkgo/v2 v2.15.0 // indirect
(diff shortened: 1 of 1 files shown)


janos:
> Looks reasonable, but have a few questions.

Thanks.

> 
> 1. As I understand, each underlay-overlay addresses pair must have signature. How will it be handled with multiple underlay addresses? Will we have multiple signatures or just one for whole batch of underlay addresses?

Actually, the signing is done on the serialized underlay address (byte slice), so serialized multiple underlay addresses does not change anything related to signatures.

> 2. If we have we have multiple signatures, than we must specify order of signatures and underlay addresses to make verification easier. So we take underlay_address[0] + overlay address and signatures[0] and verify, than doing same for address[1] and signature[1] and so on.

We just need one signature for one node. For multiple underlay addresses, all of then can be serialized for single signing (handhake). Function bzz.generateSignData is using underlay as []byte.

janos:
This PR is included in the PR that integrates this functions https://github.com/ethersphere/bee/pull/5204. I am closing this one.

Issue #5201, opened 2025-08-28
Advertizing multiple underlay addresses


janos:
#### Summary

Bee nodes should be able to send multiple underlay addresses from peers in p2p and other protocols, store multiple underlay addresses internally and handle multiple underlay addresses in connectivity related services such as Kademlia.

#### Motivation

So far, bee node used only one underlay address to identify a connectivity to its peers. With the progress on *in-browser node* implementation, an additional transport WSS is needed to be supported and nodes need to advetize their public (with exposed IPs) underlay addresses to other peers through the handshake and hive protocols. Bee node should be able to store multiple underlay addresses in its addressbook. Kademlia and libp2p should be able to determine which underlay address should be used to connect to peers based on specific criteriums. To prioritize TCP on regular nodes and to use only WSS for in-browser nodes. In the future the logic can be extended to prioritize quic transport, for example.

#### Implementation

It was discussed with the research team about the approach to achieve this feature and the decision was made not to change any p2p protocol messages, but to use Underlay []byte array to serialize multiple underlay addresses. That way, backward compatibility is preserved and there is no need to change anything related to the Bzz address signature.

The proposed serialization is in the PR #5199.

Technically bzz.Address go type should have its Underlay field changed to a slice and from there many other changes would be needed.

Existing tests should be adjusted for support of multiple underlay addresses and if needed new tests should be written.

#### Drawbacks

If the complexity is too high, consider making multiple PR requests, if possible.

PR #5204, opened 2025-09-01
Multiple underlay addresses

sbackend123:
### Checklist

### Description
Second part of PR: [#5199](https://github.com/ethersphere/bee/pull/5199)

Part of: Support for multiple underlay addresses for different node types (in-browser, bee server) to choose from. Issue [#5201](https://github.com/ethersphere/bee/issues/5201)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/ethersphere/bee/v2
 
 go 1.24.0
 
-toolchain go1.24.2
-
 require (
 	contrib.go.opencensus.io/exporter/prometheus v0.4.2
 	github.com/armon/go-radix v1.0.0
--- a/pkg/api/util.go
+++ b/pkg/api/util.go
@@ -211,13 +211,54 @@ func mapStructure(input, output interface{}, hooks map[string]func(v string) (st
 			field.SetString(value)
 		case reflect.Slice:
 			if value == "" {
-				return nil // Nil slice.
+				return nil
 			}
-			val, err := hex.DecodeString(value)
-			if err != nil {
-				return err
+
+			elemT := field.Type().Elem()
+
+			// 1) []byte — (hex → []byte)
+			if elemT.Kind() == reflect.Uint8 {
+				val, err := hex.DecodeString(value)
+				if err != nil {
+					return err
+				}
+				field.SetBytes(val)
+				return nil
 			}
-			field.SetBytes(val)
+
+			// 2) []byte -> []multiaddr.Multiaddr
+			if elemT.Kind() == reflect.Interface && // TODO add test case
+				elemT == reflect.TypeOf((*multiaddr.Multiaddr)(nil)).Elem() {
+
+				parts := strings.Split(value, ",")
+				out := reflect.MakeSlice(field.Type(), 0, len(parts))
+				seen := make(map[string]struct{}, len(parts))
+
+				for _, p := range parts {
+					p = strings.TrimSpace(p)
+					if p == "" {
+						continue
+					}
+					// guaranty  leading '/'
+					p = "/" + strings.TrimLeft(p, "/")
+
+					ma, err := multiaddr.NewMultiaddr(p)
+					if err != nil {
+						return err
+					}
+					s := ma.String()
+					if _, ok := seen[s]; ok {
+						continue // deduplication
+					}
+					seen[s] = struct{}{}
+					out = reflect.Append(out, reflect.ValueOf(ma))
+				}
+
+				field.Set(out)
+				return nil
+			}
+			return fmt.Errorf("unsupported slice type %v", field.Type())
+
 		case reflect.Array:
 			switch field.Interface().(type) {
 			case common.Hash:
--- a/pkg/api/util.go
+++ b/pkg/api/util.go
@@ -227,7 +227,7 @@ func mapStructure(input, output interface{}, hooks map[string]func(v string) (st
 			}
 
 			// 2) []byte -> []multiaddr.Multiaddr
-			if elemT.Kind() == reflect.Interface && // TODO add test case
+			if elemT.Kind() == reflect.Interface &&
 				elemT == reflect.TypeOf((*multiaddr.Multiaddr)(nil)).Elem() {
 
 				parts := strings.Split(value, ",")
--- a/pkg/api/util_test.go
+++ b/pkg/api/util_test.go
@@ -7,6 +7,7 @@ package api_test
 import (
 	"errors"
 	"fmt"
+	ma "github.com/multiformats/go-multiaddr"
 	"math"
 	"math/big"
 	"reflect"
@@ -94,6 +95,9 @@ type (
 
 	mapSwarmAddressTest struct {
 		SwarmAddressVal swarm.Address `map:"swarmAddressVal"`
+	}
+	mapMultiaddressesTest struct {
+		MultiAddressesVal []ma.Multiaddr `map:"multiAddressesVal"`
 	}
 )
 
@@ -497,7 +501,23 @@ func TestMapStructure(t *testing.T) {
 		name: "swarm.Address value",
 		src:  map[string]string{"swarmAddressVal": "1234567890abcdef"},
 		want: &mapSwarmAddressTest{SwarmAddressVal: swarm.MustParseHexAddress("1234567890abcdef")},
-	}}
+	}, {
+		name: "ma.MultiAddress single not nil value",
+		src:  map[string]string{"multiAddressesVal": "/ip4/127.0.0.1/tcp/1634/p2p/16Uiu2HAkx8ULY8cTXhdVAcMmLcH9AsTKz6uBQ7DPLKRjMLgBVYkA"},
+		want: &mapMultiaddressesTest{MultiAddressesVal: []ma.Multiaddr{mustParseMultiaddr("/ip4/127.0.0.1/tcp/1634/p2p/16Uiu2HAkx8ULY8cTXhdVAcMmLcH9AsTKz6uBQ7DPLKRjMLgBVYkA")}},
+	}, {
+		name: "ma.MultiAddress single empty value",
+		src:  map[string]string{"multiAddressesVal": ""},
+		want: &mapMultiaddressesTest{MultiAddressesVal: nil},
+	},
+		{
+			name: "ma.MultiAddress multiple values",
+			src:  map[string]string{"multiAddressesVal": "/ip4/127.0.0.1/tcp/1634/p2p/16Uiu2HAkx8ULY8cTXhdVAcMmLcH9AsTKz6uBQ7DPLKRjMLgBVYkA,/ip6/127.0.0.1/tcp/1634/p2p/16Uiu2HAkx8ULY8cTXhdVAcMmLcH9AsTKz6uBQ7DPLKRjMLgBVYkA"},
+			want: &mapMultiaddressesTest{MultiAddressesVal: []ma.Multiaddr{
+				mustParseMultiaddr("/ip4/127.0.0.1/tcp/1634/p2p/16Uiu2HAkx8ULY8cTXhdVAcMmLcH9AsTKz6uBQ7DPLKRjMLgBVYkA"),
+				mustParseMultiaddr("/ip6/127.0.0.1/tcp/1634/p2p/16Uiu2HAkx8ULY8cTXhdVAcMmLcH9AsTKz6uBQ7DPLKRjMLgBVYkA")}},
+		},
+	}
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
@@ -592,4 +612,12 @@ func TestMapStructure_InputOutputSanityCheck(t *testing.T) {
 			t.Fatalf("expected error; have none")
 		}
 	})
+}
+
+func mustParseMultiaddr(a string) ma.Multiaddr {
+	addr, err := ma.NewMultiaddr(a)
+	if err != nil {
+		panic(err)
+	}
+	return addr
 }
--- a/pkg/bzz/address.go
+++ b/pkg/bzz/address.go
@@ -110,7 +110,7 @@ func (a *Address) Equal(b *Address) bool {
 		return a == b
 	}
 
-	return a.Overlay.Equal(b.Overlay) && IsUnderlayEqual(a.Underlay, b.Underlay) && bytes.Equal(a.Signature, b.Signature) && bytes.Equal(a.Nonce, b.Nonce)
+	return a.Overlay.Equal(b.Overlay) && AreUnderlaysEqual(a.Underlay, b.Underlay) && bytes.Equal(a.Signature, b.Signature) && bytes.Equal(a.Nonce, b.Nonce)
 }
 
 func IsAtLeastOneUnderlayEqual(a, b []ma.Multiaddr) bool {
@@ -127,7 +127,7 @@ func IsAtLeastOneUnderlayEqual(a, b []ma.Multiaddr) bool {
 	return false
 }
 
-func IsUnderlayEqual(a, b []ma.Multiaddr) bool {
+func AreUnderlaysEqual(a, b []ma.Multiaddr) bool {
 	if len(a) != len(b) {
 		return false
 	}
--- a/pkg/bzz/address.go
+++ b/pkg/bzz/address.go
@@ -14,7 +14,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethersphere/bee/v2/pkg/crypto"
 	"github.com/ethersphere/bee/v2/pkg/swarm"
@@ -28,7 +27,7 @@ var ErrInvalidAddress = errors.New("invalid address")
 // It consists of a peers underlay (physical) address, overlay (topology) address and signature.
 // Signature is used to verify the `Overlay/Underlay` pair, as it is based on `underlay|networkID`, signed with the public key of Overlay address
 type Address struct {
-	Underlay        ma.Multiaddr
+	Underlay        []ma.Multiaddr
 	Overlay         swarm.Address
 	Signature       []byte
 	Nonce           []byte
@@ -36,14 +35,14 @@ type Address struct {
 }
 
 type addressJSON struct {
-	Overlay   string `json:"overlay"`
-	Underlay  string `json:"underlay"`
-	Signature string `json:"signature"`
-	Nonce     string `json:"transaction"`
+	Overlay   string   `json:"overlay"`
+	Underlay  []string `json:"underlay"`
+	Signature string   `json:"signature"`
+	Nonce     string   `json:"transaction"`
 }
 
-func NewAddress(signer crypto.Signer, underlay ma.Multiaddr, overlay swarm.Address, networkID uint64, nonce []byte) (*Address, error) {
-	underlaysBinary := SerializeUnderlays([]ma.Multiaddr{underlay}) // todo: support multiple underlay addresses
+func NewAddress(signer crypto.Signer, underlays []ma.Multiaddr, overlay swarm.Address, networkID uint64, nonce []byte) (*Address, error) {
+	underlaysBinary := SerializeUnderlays(underlays)
 
 	signature, err := signer.Sign(generateSignData(underlaysBinary, overlay.Bytes(), networkID))
 	if err != nil {
@@ -51,7 +50,7 @@ func NewAddress(signer crypto.Signer, underlay ma.Multiaddr, overlay swarm.Addre
 	}
 
 	return &Address{
-		Underlay:  underlay,
+		Underlay:  underlays,
 		Overlay:   overlay,
 		Signature: signature,
 		Nonce:     nonce,
@@ -84,16 +83,13 @@ func ParseAddress(underlay, overlay, signature, nonce []byte, validateOverlay bo
 		return nil, ErrInvalidAddress
 	}
 
-	// handle only the first underlay, for now
-	multiUnderlay := multiUnderlays[0]
-
 	ethAddress, err := crypto.NewEthereumAddress(*recoveredPK)
 	if err != nil {
 		return nil, fmt.Errorf("extract blockchain address: %w: %w", err, ErrInvalidAddress)
 	}
 
 	return &Address{
-		Underlay:        multiUnderlay,
+		Underlay:        multiUnderlays,
 		Overlay:         swarm.NewAddress(overlay),
 		Signature:       signature,
 		Nonce:           nonce,
@@ -114,21 +110,46 @@ func (a *Address) Equal(b *Address) bool {
 		return a == b
 	}
 
-	return a.Overlay.Equal(b.Overlay) && multiaddrEqual(a.Underlay, b.Underlay) && bytes.Equal(a.Signature, b.Signature) && bytes.Equal(a.Nonce, b.Nonce)
+	return a.Overlay.Equal(b.Overlay) && IsUnderlayEqual(a.Underlay, b.Underlay) && bytes.Equal(a.Signature, b.Signature) && bytes.Equal(a.Nonce, b.Nonce)
 }
 
-func multiaddrEqual(a, b ma.Multiaddr) bool {
-	if a == nil || b == nil {
-		return a == b
+func IsUnderlayEqual(a, b []ma.Multiaddr) bool {
+	if len(a) != len(b) {
+		return false
 	}
 
-	return a.Equal(b)
+	toMap := func(addrs []ma.Multiaddr) map[string]int {
+		m := make(map[string]int, len(addrs))
+		for _, addr := range addrs {
+			m[addr.String()]++
+		}
+		return m
+	}
+
+	ma := toMap(a)
+	mb := toMap(b)
+
+	if len(ma) != len(mb) {
+		return false
+	}
+
+	for k, v := range ma {
+		if mb[k] != v {
+			return false
+		}
+	}
+	return true
 }
 
 func (a *Address) MarshalJSON() ([]byte, error) {
+	underlays := make([]string, len(a.Underlay))
+	for i, underlay := range a.Underlay {
+		underlays[i] = underlay.String()
+	}
+
 	return json.Marshal(&addressJSON{
 		Overlay:   a.Overlay.String(),
-		Underlay:  a.Underlay.String(),
+		Underlay:  underlays,
 		Signature: base64.StdEncoding.EncodeToString(a.Signature),
 		Nonce:     common.Bytes2Hex(a.Nonce),
 	})
@@ -148,23 +169,43 @@ func (a *Address) UnmarshalJSON(b []byte) error {
 
 	a.Overlay = addr
 
-	m, err := ma.NewMultiaddr(v.Underlay)
+	multiaddrs, err := stringsToMultiAddr(v.Underlay)
 	if err != nil {
 		return err
 	}
 
-	a.Underlay = m
+	a.Underlay = multiaddrs
 	a.Signature, err = base64.StdEncoding.DecodeString(v.Signature)
 	a.Nonce = common.Hex2Bytes(v.Nonce)
 	return err
 }
 
 func (a *Address) String() string {
-	return fmt.Sprintf("[Underlay: %v, Overlay %v, Signature %x, Transaction %x]", a.Underlay, a.Overlay, a.Signature, a.Nonce)
+	return fmt.Sprintf("[Underlay: %v, Overlay %v, Signature %x, Transaction %x]", a.underlaysToString(), a.Overlay, a.Signature, a.Nonce)
 }
 
 // ShortString returns shortened versions of bzz address in a format: [Overlay, Underlay]
 // It can be used for logging
 func (a *Address) ShortString() string {
-	return fmt.Sprintf("[Overlay: %s, Underlay: %s]", a.Overlay.String(), a.Underlay.String())
+	return fmt.Sprintf("[Overlay: %s, Underlays: %v]", a.Overlay.String(), a.underlaysToString())
+}
+
+func (a *Address) underlaysToString() []string {
+	underlays := make([]string, len(a.Underlay))
+	for i, underlay := range a.Underlay {
+		underlays[i] = underlay.String()
+	}
+	return underlays
+}
+
+func stringsToMultiAddr(addrs []string) ([]ma.Multiaddr, error) {
+	multiAddrs := make([]ma.Multiaddr, len(addrs))
+	for i, addr := range addrs {
+		multiAddr, err := ma.NewMultiaddr(addr)
+		if err != nil {
+			return nil, err
+		}
+		multiAddrs[i] = multiAddr
+	}
+	return multiAddrs, nil
 }
--- a/pkg/bzz/address.go
+++ b/pkg/bzz/address.go
@@ -36,7 +36,7 @@ type Address struct {
 
 type addressJSON struct {
 	Overlay   string   `json:"overlay"`
-	Underlay  []string `json:"underlay"`
+	Underlays []string `json:"underlays"`
 	Signature string   `json:"signature"`
 	Nonce     string   `json:"transaction"`
 }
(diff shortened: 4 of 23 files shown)


pkg/bzz/address.go
```diff
@@ -28,30 +27,30 @@ var ErrInvalidAddress = errors.New("invalid address")
...
 	EthereumAddress []byte
 }
 
 type addressJSON struct {
-	Overlay   string `json:"overlay"`
-	Underlay  string `json:"underlay"`
-	Signature string `json:"signature"`
-	Nonce     string `json:"transaction"`
+	Overlay   string   `json:"overlay"`
+	Underlay  []string `json:"underlay"`
```
janos: Maybe plural in Field name and json tag name?

pkg/bzz/address.go
```diff
@@ -114,21 +110,46 @@ func (a *Address) Equal(b *Address) bool {
...
 	}
 
-	return a.Overlay.Equal(b.Overlay) && multiaddrEqual(a.Underlay, b.Underlay) && bytes.Equal(a.Signature, b.Signature) && bytes.Equal(a.Nonce, b.Nonce)
+	return a.Overlay.Equal(b.Overlay) && IsUnderlayEqual(a.Underlay, b.Underlay) && bytes.Equal(a.Signature, b.Signature) && bytes.Equal(a.Nonce, b.Nonce)
 }
 
-func multiaddrEqual(a, b ma.Multiaddr) bool {
-	if a == nil || b == nil {
-		return a == b
+func IsUnderlayEqual(a, b []ma.Multiaddr) bool {
```
janos: Would a two nested for loops with the Equal method be simpler? Maybe even more performant?

pkg/hive/hive.go
```diff
@@ -186,13 +186,21 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
 			return err
 		}
 
-		if !s.allowPrivateCIDRs && manet.IsPrivateAddr(addr.Underlay) {
+		var isPrivate bool
+		for _, underlay := range addr.Underlay {
+			if !s.allowPrivateCIDRs && manet.IsPrivateAddr(underlay) {
+				isPrivate = true
+				break
```
janos: Actually, here, we should filter out private CIDRs and advetize only the public ones. Here we will not advertize any node that will have at leas one private mutliaddr. If there is at least one public multiaddr, the node should be advertized, if all underlays are private, than it should be skipped.

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -250,20 +252,22 @@ func (s *Service) Handle(ctx context.Context, stream p2p.Stream, remoteMultiaddr
...
 
-	bzzAddress, err := bzz.NewAddress(s.signer, advertisableUnderlay, s.overlay, s.networkID, s.nonce)
+	bzzAddress, err := bzz.NewAddress(s.signer, advertisableUnderlays, s.overlay, s.networkID, s.nonce)
 	if err != nil {
 		return nil, err
 	}
 
-	advertisableUnderlayBytes, err := bzzAddress.Underlay.MarshalBinary()
+	// TODO support multiple underlays
+	advertisableUnderlayBytes, err := bzzAddress.Underlay[0].MarshalBinary()
```
janos: This should contain all underlay addresses, not just the first one. It should use the SerializeUnderlays function to serialzie them all and in the handler it should use DeserializeUnderlays.

pkg/p2p/p2p.go
```diff
@@ -61,7 +61,7 @@ var ErrNetworkUnavailable = errors.New("network unavailable")
 type Service interface {
 	AddProtocol(ProtocolSpec) error
 	// Connect to a peer but do not notify topology about the established connection.
-	Connect(ctx context.Context, addr ma.Multiaddr) (address *bzz.Address, err error)
+	Connect(ctx context.Context, addrs []ma.Multiaddr) (address *bzz.Address, err error)
```
janos, line 64: It is a question is the responsibility of the p2p Service implementation to choose which underlay address to use to connect to. One way to look at this is to have the same logic in one place transparent to the caller of the Connect function. This could be fine completely. I am just thinking would keeping the Connect function predictable to the caller with only one underlay address be more clear. In that case, the caller would know if the connection succeeded with the exactly the underlay that it passed to the Connect function. This may have a consequence that the logic of selecting the "desired" underlay address would need to be used always before Connect, and if it is always the same logic, than it makes sense to be hidden within the Connect function, but this can remove some flexibility that may be needed in some parts of the code. Again, it is nice to have libp2ppeer.AddrInfo mutliple underlay addresses and let it deal with the connectivity internally. I am just thinking, I still do not have the answer for this change.

pkg/hive/hive.go
sbackend123: Sorry, didn't get you

pkg/p2p/p2p.go
sbackend123: I think it depends on how we choose the transport for a node that supports both TCP and WSS. If we make that choice explicitly, then it makes sense to keep `Connect` unchanged. Otherwise, the functionality provided by libp2p frees us from having to perform sequential dials across different underlays (which can be quite slow) or implement parallel dials (which increases code complexity).

pkg/hive/hive.go
janos: This change will set as any node as "private" (isPrivate = true) that has at least one underlay address with a private IP address. But it may be and is very possible that that node has other underlay addresses that are not private, and the whole node will be skipped in advertizement thought he hive protocol (as per current state of the changeset).

Actually, the node should be considered as "private" only if all underlay addresses have a private IP address. Or node should not be considered as private if it has at least one underlay address with a public IP address.

In short, if allowPrivateCIDRs is false than filter out all underlay addresses that have private IP address and send only them through the hive message. If the length of such filtered underlay addresses is 0, do not advertize the node at all. If allowPrivateCIDRs is true, no filtering is needed and all nodeses are advetized.

pkg/p2p/p2p.go
janos: Yes, the case of libp2p functionality that is already provided is a strong one.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -708,35 +708,27 @@ func buildUnderlayAddress(addr ma.Multiaddr, peerID libp2ppeer.ID) (ma.Multiaddr
...
+	for _, r := range remotes {
+		if overlay, found := s.peers.isConnected(id, r); found {
+			return &bzz.Address{Overlay: overlay, Underlay: addrs}, p2p.ErrAlreadyConnected
 		}
-		return address, p2p.ErrAlreadyConnected
 	}
 
-	if err := s.connectionBreaker.Execute(func() error { return s.host.Connect(ctx, *info) }); err != nil {
+	// libp2p will chose first ready address (possible multi dial).
+	info := libp2ppeer.AddrInfo{ID: id, Addrs: remotes}
```
janos: `remotes` that are returned from the normalizeAddrsSamePeer have "decapsulated" hostAddr. Are such multiadds ok (complete) to be set as AddrInfo.Addrs?

pkg/hive/hive.go
```diff
@@ -299,17 +310,31 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
 			ctx, cancel := context.WithTimeout(ctx, pingTimeout)
 			defer cancel()
 
-			start := time.Now()
-
 			// check if the underlay is usable by doing a raw ping using libp2p
-			if _, err := s.streamer.Ping(ctx, multiUnderlay); err != nil {
+			// check if node is running in browser, and if it is, ping wss only
```
janos: The comment refers to a particular situation, but the code is actually more general, trying to find any reachable underlay, which would work just fine in wss case as well. I think that the comment should be updated.

pkg/bzz/address.go
```diff
@@ -114,21 +110,52 @@ func (a *Address) Equal(b *Address) bool {
...
+			if a[i] != nil && b[j] != nil && a[i].Equal(b[j]) {
+				return true
+			}
+		}
 	}
+	return false
+}
 
-	return a.Equal(b)
+func IsUnderlayEqual(a, b []ma.Multiaddr) bool {
```
janos: Would name `AreUnderlaysEqual` (or something similar with plurals) make more sense as it compares lists?

pkg/p2p/libp2p/libp2p.go
sbackend123: as far as I understand - yes, if we still have peer id

pkg/api/util.go
```diff
@@ -211,13 +211,54 @@ func mapStructure(input, output any, hooks map[string]func(v string) (string, er
...
+				return nil
+			}
+
+			// 2) []byte -> []multiaddr.Multiaddr
+			if elemT.Kind() == reflect.Interface &&
+				elemT == reflect.TypeOf((*multiaddr.Multiaddr)(nil)).Elem() {
+
+				parts := strings.Split(value, ",")
+				out := reflect.MakeSlice(field.Type(), 0, len(parts))
+				seen := make(map[string]struct{}, len(parts))
```
gacevicljubisa: Why do we have any changes in mapStructure? It is only used in api handlers and I didn't found any handler that is using as an input slice of Multiaddr?

pkg/bzz/address.go
```diff
@@ -109,21 +110,52 @@ func (a *Address) Equal(b *Address) bool {
...
+		for j := 0; j < len(b); j++ {
+			if a[i] != nil && b[j] != nil && a[i].Equal(b[j]) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func AreUnderlaysEqual(a, b []ma.Multiaddr) bool {
```
gacevicljubisa, line 121: We should have unit tests for this func

pkg/bzz/address.go
```diff
@@ -109,21 +110,52 @@ func (a *Address) Equal(b *Address) bool {
...
 	}
 
-	return a.Overlay.Equal(b.Overlay) && multiaddrEqual(a.Underlay, b.Underlay) && bytes.Equal(a.Signature, b.Signature) && bytes.Equal(a.Nonce, b.Nonce)
+	return a.Overlay.Equal(b.Overlay) && AreUnderlaysEqual(a.Underlay, b.Underlay) && bytes.Equal(a.Signature, b.Signature) && bytes.Equal(a.Nonce, b.Nonce)
 }
 
-func multiaddrEqual(a, b ma.Multiaddr) bool {
-	if a == nil || b == nil {
-		return a == b
+func IsAtLeastOneUnderlayEqual(a, b []ma.Multiaddr) bool {
```
gacevicljubisa: this can be removed, it is not used

pkg/bzz/address.go
```diff
@@ -109,21 +110,52 @@ func (a *Address) Equal(b *Address) bool {
...
+}
+
+func AreUnderlaysEqual(a, b []ma.Multiaddr) bool {
+	if len(a) != len(b) {
+		return false
 	}
 
-	return a.Equal(b)
+	used := make([]bool, len(b))
+	for i := 0; i < len(a); i++ {
```
martinconic: Maybe some small go modernisation of these loops

pkg/bzz/address.go
```diff
@@ -143,23 +175,43 @@ func (a *Address) UnmarshalJSON(b []byte) error {
...
 }
 
 // ShortString returns shortened versions of bzz address in a format: [Overlay, Underlay]
 // It can be used for logging
 func (a *Address) ShortString() string {
-	return fmt.Sprintf("[Overlay: %s, Underlay: %s]", a.Overlay.String(), a.Underlay.String())
+	return fmt.Sprintf("[Overlay: %s, Underlays: %v]", a.Overlay.String(), a.underlaysToString())
+}
+
+func (a *Address) underlaysToString() []string {
```
martinconic: rename to `underlaysAsStrings`

pkg/bzz/address.go
```diff
@@ -143,23 +175,43 @@ func (a *Address) UnmarshalJSON(b []byte) error {
...
+	underlays := make([]string, len(a.Underlay))
+	for i, underlay := range a.Underlay {
+		underlays[i] = underlay.String()
+	}
+	return underlays
+}
+
+func stringsToMultiAddr(addrs []string) ([]ma.Multiaddr, error) {
```
martinconic: small refactor this one to maybe `parseMultiaddrs` or `multiaddrsFromStrings`

pkg/p2p/libp2p/libp2p.go
```diff
@@ -823,17 +846,47 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (address *bzz.
...
 }
 
+func normalizeAddrsSamePeer(addrs []ma.Multiaddr) (libp2ppeer.ID, []ma.Multiaddr, error) {
+	var id libp2ppeer.ID
+	seen := make(map[string]struct{})
+	remotes := make([]ma.Multiaddr, 0)
+
+	for _, a := range addrs {
+		info, err := libp2ppeer.AddrInfoFromP2pAddr(a)
+		if err != nil {
```
martinconic: In this for loop if there is an error with one of the addresses we return, maybe we should continue with the other addresses before returning

pkg/hive/hive.go
```diff
@@ -183,16 +183,19 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
 				s.logger.Debug("broadcast peers; peer not found in the addressbook, skipping...", "peer_address", p)
 				continue
 			}
+			s.logger.Debug("send peers", "peer", p.String(), "err", err)
```
gacevicljubisa: delete this log

pkg/hive/hive.go
```diff
@@ -293,17 +296,31 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
...
+				pingSuccessful bool
+				start          time.Time
+			)
+			for _, underlay := range multiUnderlay {
+				s.logger.Debug("check and add peers, try ping underlay", "overlay", newPeer.Overlay, "underlay", underlay.String())
+
+				// ping each underlay address, pick first available
+				start = time.Now()
+				if _, err := s.streamer.Ping(ctx, underlay); err != nil {
+					s.logger.Debug("unreachable peer underlay", "peer_address", hex.EncodeToString(newPeer.Overlay), "underlay", underlay)
```
gacevicljubisa: Print error in logs

pkg/hive/hive.go
```diff
@@ -293,17 +296,31 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
 			ctx, cancel := context.WithTimeout(ctx, pingTimeout)
 			defer cancel()
 
-			start := time.Now()
+			var (
+				pingSuccessful bool
+				start          time.Time
+			)
+			for _, underlay := range multiUnderlay {
+				s.logger.Debug("check and add peers, try ping underlay", "overlay", newPeer.Overlay, "underlay", underlay.String())
```
gacevicljubisa: delete log

pkg/topology/kademlia/kademlia.go
```diff
@@ -105,7 +105,7 @@ type Options struct {
 // kadOptions are made from Options with default values set
 type kadOptions struct {
 	SaturationFunc binSaturationFunc
-	Bootnodes      []ma.Multiaddr
+	Bootnodes      []ma.Multiaddr // TODO each of bootnodes will have multiple addresses
```
martinconic: remove the comment

pkg/topology/kademlia/kademlia.go
```diff
@@ -820,7 +828,7 @@ func (k *Kad) connectBootNodes(ctx context.Context) {
 			if attempts >= maxBootNodeAttempts {
 				return true, nil
 			}
-			bzzAddress, err := k.p2p.Connect(ctx, addr)
+			bzzAddress, err := k.p2p.Connect(ctx, []ma.Multiaddr{addr}) // TODO use multiple underlay addresses, after bootnodes would support it
```
martinconic: remove the comment

martinconic (changes requested):

pkg/p2p/libp2p/libp2p.go
```diff
@@ -409,6 +409,14 @@ func (s *Service) handleIncoming(stream network.Stream) {
...
 	peerID := stream.Conn().RemotePeer()
 	handshakeStream := newStream(stream, s.metrics)
+
+	// peerAddrs := s.host.Peerstore().Addrs(peerID)
+	// connectionAddr := stream.Conn().RemoteMultiaddr()
+	// handshakeAddr := SelectBestAdvertisedAddress(peerAddrs, connectionAddr)
+
+	// s.logger.Debug("selected handshake address", "peer_id", peerID, "selected_addr", handshakeAddr.String(), "connection_addr", connectionAddr.String(), "advertised_count", len(peerAddrs))
+
+	// TODO: use handshakeAddr from the code above instead of the connection RemoteMultiaddr
```
martinconic: address  TODO  or remove comment

pkg/p2p/libp2p/libp2p.go
```diff
@@ -1177,3 +1230,42 @@ func isNetworkOrHostUnreachableError(err error) bool {
 	}
 	return false
 }
+
+func buildFullMA(addr ma.Multiaddr, peerID libp2ppeer.ID) (ma.Multiaddr, error) {
```
gacevicljubisa, line 1331: We have buildFullMAs and buildFullMA, one in libp2p, and one in handshake.

martinconic: yeah saw that also, we should unify them into one place only

pkg/api/util.go
janos: This is to handle multiple underlays in path for /connect endpoint, but it is not needed to pass multiple underlays for that endpoint. We can remove these changes an leave /connect endpoint as it was.

pkg/api/util_test.go
```diff
@@ -593,3 +613,11 @@ func TestMapStructure_InputOutputSanityCheck(t *testing.T) {
 		}
 	})
 }
+
+func mustParseMultiaddr(a string) ma.Multiaddr {
```
janos: This should be prettier as test Helper function, than to panic.

pkg/bzz/address.go
```diff
@@ -28,33 +27,30 @@ var ErrInvalidAddress = errors.New("invalid address")
...
+	Underlay        []ma.Multiaddr
 	Overlay         swarm.Address
 	Signature       []byte
 	Nonce           []byte
 	EthereumAddress []byte
 }
 
 type addressJSON struct {
-	Overlay   string `json:"overlay"`
-	Underlay  string `json:"underlay"`
```
janos, line 40: As we discussed at some point, we should keep the Underlay field and have the underlay of p2p connection there, for example, as that is the case with the current code and it makes the api backward compatible. We can deprecate the field gradually.

pkg/bzz/address.go
janos: I suppose that it was needed for type []string in addressJSON struct, but it should be checked if addressJSON can have []ma.Multiaddr for Underlays and the marshalling will produce list of strings in resulted json. In that case, this function could be removed.

pkg/hive/hive_test.go
```diff
@@ -321,10 +367,6 @@ func expectOverlaysEventually(t *testing.T, exporter ab.Interface, wantOverlays
 			t.Errorf("overlay %s expected but not found", v.String())
 		}
 	}
-
-	if t.Failed() {
```
janos, line 325: I am not sure why is this removed.

pkg/hive/hive_test.go
```diff
@@ -380,3 +422,43 @@ func readAndAssertPeersMsgs(in []byte, expectedLen int) ([]pb.Peers, error) {
 
 	return peers, nil
 }
+
+func ComparePeerMsgs(got, want []*pb.BzzAddress) error {
```
janos: This should not be exported and can be a helper function.

pkg/hive/hive_test.go
```diff
@@ -380,3 +422,43 @@ func readAndAssertPeersMsgs(in []byte, expectedLen int) ([]pb.Peers, error) {
...
+				ovlHex, shortHex(g.Nonce), shortHex(w.Nonce))
+		}
+	}
+	return nil
+}
+
+func flattenByOverlay(batches []*pb.BzzAddress) map[string]*pb.BzzAddress {
+	m := make(map[string]*pb.BzzAddress)
+	for _, batch := range batches {
+		m[hex.EncodeToString(batch.Overlay)] = batch
```
janos: This will replace any previously assigned value for this key. Is this the desired behaviour?

pkg/hive/hive_test.go
```diff
@@ -380,3 +422,43 @@ func readAndAssertPeersMsgs(in []byte, expectedLen int) ([]pb.Peers, error) {
...
+		}
+		if !bytes.Equal(g.Nonce, w.Nonce) {
+			return fmt.Errorf("peer %s: expected nonce (got=%s want=%s)",
+				ovlHex, shortHex(g.Nonce), shortHex(w.Nonce))
+		}
+	}
+	return nil
+}
+
+func flattenByOverlay(batches []*pb.BzzAddress) map[string]*pb.BzzAddress {
```
janos: Improve naming.

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -146,32 +141,36 @@ func (s *Service) Handshake(ctx context.Context, stream p2p.Stream, peerMultiadd
...
+			return nil, fmt.Errorf("extract addr from P2P: %w", err)
+		}
 
-	advertisableUnderlay, err := s.advertisableAddresser.Resolve(observedUnderlay)
-	if err != nil {
-		return nil, err
-	}
+		if s.libp2pID != observedUnderlayAddrInfo.ID {
+			// NOTE eventually we will return error here, but for now we want to gather some statistics
+			s.logger.Warning("received peer ID does not match ours", "their", observedUnderlayAddrInfo.ID, "ours", s.libp2pID)
```
janos: Address this according to the comment.

pkg/p2p/libp2p/libp2p.go
janos: We should address the handshake underlay value in a separate pr, as it is an issue from way before. I would suggest to remove the comment and create an issue to improve the handshake remote address handling in a separate issue.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -547,10 +553,28 @@ func (s *Service) handleIncoming(stream network.Stream) {
 	s.logger.Debug("stream handler: successfully connected to peer (inbound)", "address", i.BzzAddress.Overlay, "light", i.LightString(), "user_agent", peerUserAgent)
 }
 
+func (s *Service) notifyReacherConnected(stream network.Stream, overlay swarm.Address, peerID libp2ppeer.ID) {
+	if s.reacher != nil {
```
janos: return early if reacher is nil, instead of nesting the code.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -750,7 +767,9 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (address *bzz.
 	}
 
 	handshakeStream := newStream(stream, s.metrics)
-	i, err := s.handshakeService.Handshake(ctx, handshakeStream, stream.Conn().RemoteMultiaddr(), stream.Conn().RemotePeer())
+	s.logger.Info("Connect: handshake peer", "stream.Conn().RemotePeer()", stream.Conn().RemotePeer())
```
janos: Remove this log line.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -707,35 +731,28 @@ func buildUnderlayAddress(addr ma.Multiaddr, peerID libp2ppeer.ID) (ma.Multiaddr
...
+	for _, r := range remotes {
+		if overlay, found := s.peers.isConnected(peerID, r); found {
+			return &bzz.Address{Overlay: overlay, Underlay: addrs}, p2p.ErrAlreadyConnected
 		}
-		return address, p2p.ErrAlreadyConnected
 	}
 
-	if err := s.connectionBreaker.Execute(func() error { return s.host.Connect(ctx, *info) }); err != nil {
+	// libp2p will chose first ready address (possible multi dial).
+	s.logger.Info("Connect: debug full connectivity test", "id", peerID.String(), "remotes", remotes)
```
janos: Remove this log line.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -781,7 +800,10 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (address *bzz.
 		return nil, p2p.ErrPeerBlocklisted
 	}
 
+	s.logger.Info("addIfNotExists peer", "stream.Conn()", stream.Conn(), "overlay", overlay, "full_node", i.FullNode)
```
janos: Remove this log line.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -781,7 +800,10 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (address *bzz.
...
+
 	if exists := s.peers.addIfNotExists(stream.Conn(), overlay, i.FullNode); exists {
+		s.logger.Info("addIfNotExists peer EXIST", "stream.Conn()", stream.Conn(), "overlay", overlay, "full_node", i.FullNode)
```
janos: Remove this log line.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -796,6 +818,7 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (address *bzz.
 	}
 
 	if i.FullNode {
+		s.logger.Info("Connect: full node: put address to addr book", "overlay", overlay, "bzz_add", i.BzzAddress.String())
```
janos: Remove this log line.

pkg/p2p/libp2p/libp2p.go
janos: Yes, I agree, I've added this one here for quick fix while debugging.

pkg/topology/kademlia/kademlia.go
```diff
@@ -421,6 +421,8 @@ func (k *Kad) connectionAttemptsHandler(ctx context.Context, wg *sync.WaitGroup,
 			return
 		}
 
+		k.logger.Info("debug: kademlia: got addr grom addr book", "addr", bzzAddr.String())
```
janos: Remove this log line.

pkg/topology/kademlia/kademlia.go
```diff
@@ -429,6 +431,8 @@ func (k *Kad) connectionAttemptsHandler(ctx context.Context, wg *sync.WaitGroup,
 			}
 		}
 
+		k.logger.Info("debug: kademlia: try to connect", "addr", bzzAddr.String())
```
janos: Remove this log line.

pkg/topology/kademlia/kademlia.go
```diff
@@ -477,6 +481,8 @@ func (k *Kad) connectionAttemptsHandler(ctx context.Context, wg *sync.WaitGroup,
 			case peer := <-peerConnChan:
 				addr := peer.addr.String()
 
+				k.logger.Debug("debug kademlia: connectionAttemptsHandler, got peer from chan", "peer_address", addr)
```
janos: Remove this log line.

pkg/topology/kademlia/kademlia.go
```diff
@@ -745,6 +751,8 @@ func (k *Kad) Start(ctx context.Context) error {
 	k.wg.Add(1)
 	go k.manage()
 
+	k.logger.Debug("kademlia: started", "previously_connected", k.previouslyConnected())
```
janos: Remove this log line.

pkg/topology/kademlia/kademlia.go
```diff
@@ -1076,6 +1084,7 @@ outer:
 				cCtx, cCancel := context.WithTimeout(k.bgBroadcastCtx, time.Minute)
 				defer cCancel()
 
+				k.logger.Debug("Announce", "connectedPeer", connectedPeer.String(), "peer", peer.String())
```
janos: Remove this log line.

pkg/topology/kademlia/kademlia.go
```diff
@@ -1093,6 +1102,8 @@ outer:
 	default:
 	}
 
+	k.logger.Debug("Announce", "peer", peer.String(), "addrs", addrs)
```
janos: Remove this log line.

pkg/topology/kademlia/kademlia.go
```diff
@@ -1108,6 +1119,7 @@ func (k *Kad) AnnounceTo(ctx context.Context, addressee, peer swarm.Address, ful
 		return errAnnounceLightNode
 	}
 
+	k.logger.Debug("AnnounceTo", "addressee", addressee.String(), "peer", peer.String())
```
janos: Remove this log line.

pkg/p2p/libp2p/libp2p.go
janos: I have no clear idea where to put this function. It should belong to the libp2p package, but an import cycle is created with the handshake function. We can deduplicate this function when the handshake issue is addressed in a separate pr.

pkg/hive/hive.go
```diff
@@ -327,27 +341,50 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
...
-		multiUnderlay, err := ma.NewMultiaddrBytes(p.Underlay)
+		multiUnderlays, err := bzz.DeserializeUnderlays(p.Underlay)
 		if err != nil {
 			s.metrics.PeerUnderlayErr.Inc()
 			s.logger.Debug("multi address underlay", "error", err)
 			continue
 		}
 
+		if len(multiUnderlays) == 0 {
+			s.logger.Debug("check and add peers, NO UNDERLAYS", "overlay", swarm.NewAddress(p.Overlay).String())
```
gacevicljubisa: Remove capitalization

go.mod
```diff
@@ -132,7 +132,7 @@ require (
 	github.com/multiformats/go-multiaddr-fmt v0.1.0 // indirect
 	github.com/multiformats/go-multibase v0.2.0 // indirect
 	github.com/multiformats/go-multicodec v0.9.0 // indirect
-	github.com/multiformats/go-varint v0.0.7 // indirect
+	github.com/multiformats/go-varint v0.0.7
```
gacevicljubisa: move this to direct imports

pkg/bzz/address.go
```diff
@@ -175,36 +177,41 @@ func (a *Address) UnmarshalJSON(b []byte) error {
 
 	a.Overlay = addr
 
-	multiaddrs, err := stringsToMultiAddr(v.Underlays)
+	// append the underlay for backward compatibility
+	if !slices.Contains(v.Underlays, v.Underlay) {
```
martinconic, line 181: Good catch, nice

pkg/p2p/libp2p/libp2p.go
martinconic: The only solution that I can think is to have an `multiaddrs.go` file(or utils.go) in internals package and put the code there

pkg/hive/hive.go
```diff
@@ -327,27 +342,51 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
...
+
 		// if peer exists already in the addressBook
 		// and if the underlays match, skip
 		addr, err := s.addressBook.Get(swarm.NewAddress(p.Overlay))
-		if err == nil && addr.Underlay.Equal(multiUnderlay) {
+		if err == nil && bzz.AreUnderlaysEqual(addr.Underlays, multiUnderlays) {
 			continue
 		}
 
+		s.logger.Debug("check and add peers, adding peer", "overlay", swarm.NewAddress(p.Overlay).String())
```
gacevicljubisa: maybe remove this log

gacevicljubisa (approved):

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -146,32 +165,57 @@ func (s *Service) Handshake(ctx context.Context, stream p2p.Stream, peerMultiadd
...
+			return nil, fmt.Errorf("get host advertizable addresses: %w", err)
+		}
+
+		advertisableUnderlays = append(advertisableUnderlays, hostAddrs...)
 	}
 
-	advertisableUnderlayBytes, err := bzzAddress.Underlay.MarshalBinary()
+	// sort to remove potential duplicates
+	slices.SortFunc(advertisableUnderlays, func(a, b ma.Multiaddr) int {
+		if a.Equal(b) {
```
martinconic: can we use here return cmp.Compare(a.String(), b.String()) from [ the `cmp` package](https://pkg.go.dev/cmp). ?

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -239,35 +279,62 @@ func (s *Service) Handle(ctx context.Context, stream p2p.Stream, remoteMultiaddr
...
```
martinconic: can we use here return cmp.Compare(a.String(), b.String()) from the cmp package ?

pkg/p2p/libp2p/libp2p.go
```diff
@@ -750,7 +828,31 @@ func (s *Service) Connect(ctx context.Context, addr ma.Multiaddr) (address *bzz.
 	}
 
 	handshakeStream := newStream(stream, s.metrics)
-	i, err := s.handshakeService.Handshake(ctx, handshakeStream, stream.Conn().RemoteMultiaddr(), stream.Conn().RemotePeer())
+
+	waitPeersCtx, cancel := context.WithTimeout(ctx, peerstoreWaitAddrsTimeout)
```
martinconic: Can we extract this block in a function, seems to be repeated ?

Issue #4822, opened 2024-09-13
ENS resolution


nugaon:
⚠️ Support requests in an issue-format will be closed immediately. For support, go to Swarm's [Discord](https://discord.gg/XGNBEKktkD).

### Context

Bee 2.2.0

### Summary

When I try to open my registered personal page on `http://vps:1633/bzz/nugaon.eth` I get 400 error.

### Expected behavior

I can resolve the nugaon.eth address as so far.

### Actual behavior

```
"time"="2024-09-13 15:37:30.887535" "level"="debug" "logger"="node/api" "msg"="resolve name: attempting to resolve string to address" "string"="nugaon.eth"
"time"="2024-09-13 15:37:30.887822" "level"="debug" "logger"="node/api/act_decryption_handler" "msg"="invalid path params" "error"="1 error(s) occurred: `address=nugaon.eth`: invalid name or bzz address: 1 error occurred:\n\t* failed parsing CID nugaon.eth err invalid cid: selected encoding not supported: could not parse\n\n"
"time"="2024-09-13 15:37:30.887974" "level"="error" "logger"="node/api/act_decryption_handler" "msg"="invalid path params" "error"="1 error(s) occurred: `address=nugaon.eth`: invalid name or bzz address: 1 error occurred:\n\t* failed parsing CID nugaon.eth err invalid cid: selected encoding not supported: could not parse\n\n"
"time"="2024-09-13 15:37:30.888239" "level"="debug" "logger"="node/api" "msg"="content length header not found"
"time"="2024-09-13 15:37:30.888468" "level"="info" "logger"="node/api" "msg"="api access" "ip"="X.X.X.X" "method"="GET" "host"="vps:1633" "uri"="/bzz/nugaon.eth/" "proto"="HTTP/1.1" "status"=400 "size"=240 "duration"="1.412272ms" "user-agent"="browser"
```

### Steps to reproduce

open http://{hostName}:1633/bzz/nugaon.eth in the browser.

### Possible solution
seemingly the problem caused by the newly introduced ACT feature

PR #5213, opened 2025-09-11
fix: ens error handling

### LINKED ISSUE #4450: `invalid path params` ens name resolution errors
⚠️ Support requests in an issue-format will be closed immediately. For support, go to Swarm's [Discord](https://discord.gg/XGNBEKktkD).

### Context
bee: 1.17.5

### Summary
1. when the following command is run
`
curl localhost:1633/bzz/fairdatasociety.eth/ 
`
the following error is thrown
`
{"code":400,"message":"invalid path params","reasons":[{"field":"address","error":"invalid hex byte: U+0069 'i'"}]}
`
2. when the following is run
`
curl localhost:1633/bzz/octalmage.eth/ 
`
the following error is thrown
`
{"code":400,"message":"invalid path params","reasons":[{"field":"address","error":"invalid name or bzz address: 1 error occurred:\n\t* failed parsing CID octalmage.eth err invalid cid: selected encoding not supported: could not parse\n\n"}]}
`

### Expected behavior
error is thrown instead of the ens name getting resolved. likely due to the "fa"

nugaon:
Fixes #4450 #4822

### Checklist

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -633,17 +633,6 @@ func (e *validationError) Error() string {
 	return fmt.Sprintf("`%s=%v`: %v", e.Entry, e.Value, e.Cause)
 }
 
-// serviceUnavailableError is a custom error type for service unavailable errors.
-type serviceUnavailableError struct {
-	Entry string
-	Cause error
-}
-
-// Error implements the error interface.
-func (e *serviceUnavailableError) Error() string {
-	return fmt.Sprintf("`%s`: %v", e.Entry, e.Cause)
-}
-
 // mapStructure maps the input into output struct and validates the output.
 // It's a helper method for the handlers, which reduces the chattiness
 // of the code.


gacevicljubisa (approved):

gacevicljubisa:
I have approved. The only comment that I have is that the `mapStructure` method should focus purely on data transformation (mapping input to struct) and return errors, while HTTP response handling should be completely separate. But this was already there before this PR. 

There is no need for this status to be checked for each endpoint where we are calling the `mapStructure`. In future this can grow and lead to unexpected responses and it definitely complicates the logic of the `mapStructure`.

nugaon:
from the practical perspective, I somewhat agree with the response handling as it is now because otherwise the error responses should be handled the same way on each endpoint which is redundant. Nevertheless, you are also right about these specific errors are not necessary to be checked in every case. From my side, I would go with this as it is a fix PR and did not intend to break the original structure.

PR #5222, opened 2025-09-22
fix: pin eviction guard

### LINKED ISSUE #5215: Bug: Pinned chunks can be evicted from storage despite pin protection
## Summary

Pinned chunks are not protected from eviction processes, causing permanent data loss when batches expire or storage pressure triggers evictions. This violates the fundamental guarantee that pinned data should persist until explicitly unpinned.

## Expected Behavior

Pinned chunks should be excluded from all eviction processes:
  - Batch expiry evictions
  - Storage pressure evictions
  - Manual batch evictions via `EvictBatch()` API

## Current Behavior

Pinned chunks are evicted along with other chunks when:
  1. Their postage batch expires
  2. Storage capacity triggers reserve evictions
  3. Manual batch eviction is called

After eviction, pin collections reference non-existent chunks, causing download failures with `storage.ErrNotFound`.

## Root Cause

The eviction logic in `pkg/storer/internal/reserve/reserve.go:330 (EvictBatchBin)` never checks pin status before removing chunks:

```
  func (r *Reserve) EvictBatchBin(ctx context.Context, batchID []byte, count int, bin uint8) (int, error) {
      // Iterates through chunks and removes them without pin protection
      err := RemoveChunkWithItem(ctx, s, item)  // No HasPin() check
  }
```

## Steps to Reproduce

  1. Upload and pin content to a bee node
  2. Wait for the postage batch to expire, OR trigger storage pressure
  3. Attempt to download the pinned content
  4. Observe `storage.ErrNotFound errors` despite pin status showing as active

## Proposed Solution

Add pin protection checks to eviction paths:

  1. In `EvictBatchBin()`: Before adding chunks to `evicteditems`, check if chunk address exists in any pin collection
  2. In `RemoveChunkWithItem()`: Add safety check using `HasPin()` before deletion
  3. Performance optimization: Consider caching pin status or maintaining a pin-protected chunk index

## Required Changes

  - Modify `pkg/storer/internal/reserve/reserve.go` eviction logic
  - Add pin status checks in `pkg/storer/internal/pinning/pinning.go`
  - Update eviction metrics to track pin-protected chunks
  - Add integration tests covering pin-eviction scenarios

## Impact

  - High: Data loss for pinned content
  - Affects: All nodes using pinning functionality
  - Regression risk: Low (adding protection, not removing functionality)

## Additional Context

This issue may be exacerbated during node upgrades when migration processes trigger eviction cleanup, explaining reports of pinned data loss after version upgrades.

nugaon:
The changes implemented comprehensive pin protection for chunk eviction by addressing a critical bug in the `EvictBatchBin` method.

The fix modified the `EvictBatchBin` method to properly check each chunk against all pin collections using the newly created `pinstore.IsChunkPinnedInCollection` function. When a pinned chunk is found, the eviction only removes reserve related records of that chunk. This ensures that non-pinned chunks can still be evicted while maintaining the integrity of pinned content.

Reserve capacity eviction is directly protected through the `EvictBatchBin` method. 
Batch expiry eviction is automatically protected because the entire flow from `EvictBatch` ultimately calls `EvictBatchBin`. 
Cache eviction was determined to not require pin protection because pinned chunks are stored in the main ChunkStore layer. 

Closes #5215

### Checklist

### DIFF
--- a/pkg/storer/internal/reserve/reserve.go
+++ b/pkg/storer/internal/reserve/reserve.go
@@ -338,7 +338,10 @@ func (r *Reserve) EvictBatchBin(
 	r.multx.Lock(string(batchID))
 	defer r.multx.Unlock(string(batchID))
 
-	var evicteditems []*BatchRadiusItem
+	var (
+		evictedItems       []*BatchRadiusItem
+		pinnedEvictedItems []*BatchRadiusItem
+	)
 
 	if count <= 0 {
 		return 0, nil
@@ -359,17 +362,22 @@ func (r *Reserve) EvictBatchBin(
 		}
 
 		// Check if the chunk is pinned in any collection
+		pinned := false
 		for _, uuid := range pinUuids {
 			has, err := pinstore.IsChunkPinnedInCollection(r.st.IndexStore(), batchRadius.Address, uuid)
 			if err != nil {
 				return true, err
 			}
 			if has {
-				return false, nil
+				pinned = true
+				pinnedEvictedItems = append(pinnedEvictedItems, batchRadius)
+				break
 			}
 		}
 
-		evicteditems = append(evicteditems, batchRadius)
+		if !pinned {
+			evictedItems = append(evictedItems, batchRadius)
+		}
 		count--
 		if count == 0 {
 			return true, nil
@@ -385,11 +393,26 @@ func (r *Reserve) EvictBatchBin(
 
 	var evicted atomic.Int64
 
-	for _, item := range evicteditems {
+	for _, item := range evictedItems {
 		func(item *BatchRadiusItem) {
 			eg.Go(func() error {
 				err := r.st.Run(ctx, func(s transaction.Store) error {
 					return RemoveChunkWithItem(ctx, s, item)
+				})
+				if err != nil {
+					return err
+				}
+				evicted.Add(1)
+				return nil
+			})
+		}(item)
+	}
+
+	for _, item := range pinnedEvictedItems {
+		func(item *BatchRadiusItem) {
+			eg.Go(func() error {
+				err := r.st.Run(ctx, func(s transaction.Store) error {
+					return RemoveChunkMetaData(ctx, s, item)
 				})
 				if err != nil {
 					return err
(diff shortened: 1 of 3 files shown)


nugaon:
This change may affect other parts of the system and the followings were considered:

if pinned chunks remain in the reserve when they should have been evicted then:
1) the reserve sample may include chunks that shouldn't be there, leading to different commitment hashes. 
2) storage radius calculation may deviate from the expected value

because of these points, during the eviction, pinned chunks are protected by retaining the chunk in the chunkstore but it removes all reserve related records.
This should be taken into account when designing migration processes.

pkg/storer/internal/reserve/reserve.go
```diff
@@ -381,6 +408,21 @@ func (r *Reserve) EvictBatchBin(
 		}(item)
 	}
 
+	for _, item := range pinnedEvictedItems {
+		func(item *BatchRadiusItem) {
+			eg.Go(func() error {
+				err := r.st.Run(ctx, func(s transaction.Store) error {
+					return RemoveChunkMetaData(ctx, s, item)
```
martinconic, line 415: Why do we want to remove the Pinned Chunks metadata, would this affect retrieval of those pinned chunks if storage pressure eviction or manual eviction?

gacevicljubisa:
> This should be taken into account when designing migration processes.

Can we add this as a note in migration?

pkg/storer/internal/reserve/reserve.go
```diff
@@ -336,21 +338,46 @@ func (r *Reserve) EvictBatchBin(
 	r.multx.Lock(string(batchID))
 	defer r.multx.Unlock(string(batchID))
 
-	var evicteditems []*BatchRadiusItem
+	var (
+		evictedItems       []*BatchRadiusItem
+		pinnedEvictedItems []*BatchRadiusItem
```
gacevicljubisa, line 343: What do you think to change names of this 2 collections and add some comment, something like:

```go
    nonPinnedItems []*BatchRadiusItem  // Fully evicted from reserve + chunkstore
    pinnedItems    []*BatchRadiusItem  // Only evicted from reserve, kept in chunkstore
```

gacevicljubisa:
I have tested it with expired batch and seems good. @nugaon if you can, please address other comments. I will approve it now

pkg/storer/internal/reserve/reserve.go
```diff
@@ -336,21 +338,46 @@ func (r *Reserve) EvictBatchBin(
...
 	}, func(res storage.Result) (bool, error) {
 		batchRadius := res.Entry.(*BatchRadiusItem)
 		if batchRadius.Bin >= bin {
 			return true, nil
 		}
-		evicteditems = append(evicteditems, batchRadius)
+
+		// Check if the chunk is pinned in any collection
+		pinned := false
+		for _, uuid := range pinUuids {
```
janos, line 366: This will slow down batch eviction linearly with the number of collections. Is there a possibility to have collection uuid as a field in BatchRadiusItem so that all of these lookups (Has method calls) are not done? From what I see, it would need some complexity to manage the synchronisations between BatchRadiusItem and pinChunkItem keys, but if the performance degradation is significant, it may be necessary. It is ok to try with this simpler solution.

nugaon: that is information redundant approach but faster one for sure.
IMO this is still a good enough approach because if you don't use pinning feature you won't notice any performance difference. It is questionable how much this feature is used to worth modifying the underlying DB structure to be redundant.

PR #5225, opened 2025-09-24
fix(transaction): improve legacy gasPrice to EIP-1559 dynamic fee

gacevicljubisa:
### Checklist

### Description
Previously, when a legacy gasPrice was provided, both gasFeeCap and 
gasTipCap were set to gasPrice, causing excessive priority fees 
(gasPrice - baseFee overpayment). 

Now correctly maps:
- gasFeeCap = gasPrice (preserves user's total ceiling)
- gasTipCap = gasPrice - baseFee (actual realizable tip)

Also adds improved error handling for edge cases.

### DIFF
--- a/pkg/transaction/wrapped/fee.go
+++ b/pkg/transaction/wrapped/fee.go
@@ -20,20 +20,44 @@ var (
 	ErrEIP1559NotSupported = errors.New("network does not appear to support EIP-1559 (no baseFee)")
 )
 
-// SuggestedFeeAndTip calculates the recommended gasFeeCap and gasTipCap for a transaction.
+// SuggestedFeeAndTip calculates the recommended gasFeeCap (maxFeePerGas) and gasTipCap (maxPriorityFeePerGas) for a transaction.
+// If gasPrice is provided (legacy mode):
+//   - On EIP-1559 networks: gasFeeCap = gasPrice; gasTipCap = max(gasPrice - baseFee, minimumTip) to respect the total cap while enforcing a tip floor where possible.
+//   - On pre-EIP-1559 networks: returns (gasPrice, gasPrice) for legacy transaction compatibility.
+//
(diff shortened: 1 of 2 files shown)


gacevicljubisa:
Example how to test:
```
curl -XPOST -H "Gas-Price:6000000" localhost:1633/stamps/870000000/17
```

pkg/transaction/wrapped/fee.go
```diff
@@ -20,20 +20,44 @@ var (
...
+			return new(big.Int).Set(gasPrice), new(big.Int).Set(gasPrice), nil
+		}
+
+		baseFee := latestBlockHeader.BaseFee
+		if gasPrice.Cmp(baseFee) < 0 {
+			return nil, nil, fmt.Errorf("specified gas price %s is below current base fee %s", gasPrice, baseFee)
+		}
+
+		// nominal tip = gasPrice - baseFee
+		gasTipCap := new(big.Int).Sub(gasPrice, baseFee)
```
martinconic, line 45: what if gasPrice - baseFee < baseFee ? Should we have max(gasPrice - baseFee, baseFee) ?

gacevicljubisa: When a user specifies a `gasPrice`, they're explicitly setting their maximum willingness to pay. The code validates that this price is at least the `baseFee`, and if it's not, returns an error.  In EIP-1559, you cannot pay less than the base fee. And the tip is the remainder after accounting for the mandatory base fee.

martinconic: Ok, but if gasFee is high and close to gasPrice, than the resulting gasTip is close to 0 and has no effect.

gacevicljubisa: Yes. But it is up to user to set GasPrice he will spend, and no more then this

martinconic (approved):

PR #5227, opened 2025-09-24
fix: add chain availability check

gacevicljubisa:
### Checklist

### Description
- Move noOpChainBackend from node package to transaction/backendnoop package
- Add type checking in isChainAvailable middleware to detect disabled chain backend
- Return 403 Forbidden instead of 503 Service Unavailable when chain is disabled
- Resolve circular dependency between node and api packages
- Add tests for backendnoop.Backend

### DIFF
--- a/pkg/node/chain.go
+++ b/pkg/node/chain.go
@@ -57,7 +57,7 @@ func InitChain(
 	if chainEnabled {
 		rpcClient, err := rpc.DialContext(ctx, endpoint)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/api/router.go
```diff
@@ -218,6 +219,17 @@ func (s *Service) checkStorageIncentivesAvailability(handler http.Handler) http.
 	})
 }
 
+func (s *Service) isChainAvailable(handler http.Handler) http.Handler {
```
martinconic: I would change the name to something else since this suggests a `bool` return. Maybe: requireChainAvailable, chainAvailabilityMiddleware, checkChainAvailability

PR #5229, opened 2025-09-25
fix: strategy channels of redundancy getter

nugaon:
the loop of channels in `runStrategy`  doesn't properly drain all the channels, causing a potential deadlock

### Checklist


nugaon:
This change was presumably based on a wrong assumption, since the drained channel is buffered with a capacity equal to the number of goroutines writing to it, and the resource leak is handled by the cancellable context in the fetch calls.

Issue #5228, opened 2025-09-24
Corruption of pinned erasure-coded content post migration


rampall:
**Description**  
Pinned content on my Bee node `v2.5.0` that was erasure coded became inaccessible after upgrading to `v2.6.0`.  
- `swarm-cli pinning list` still shows the pinned files.  
- However, I cannot download them via `curl` or `swarm-cli`.  
- If I **unpin** a file, I am able to download it (likely fetched fresh from the network), but then I would lose any pinned data that is no longer available on the network (e.g., due to postage stamp expiry).

**Steps - Not likely Reproducable**  
1. Pin erasure-coded content on Bee v2.5.0  
2. Upgrade Bee node to v2.6.0  
3. Run `swarm-cli pinning list` → files still appear pinned  
4. Try downloading pinned file with `curl` or `swarm-cli` → fails  
5. Unpin the file  
6. Retry download → succeeds (fetched fresh), but previously pinned-only data is permanently lost

**Expected behavior**  
- Pinned files (including erasure coded) should remain accessible after upgrade, regardless of postage stamp expiry or network availability.
- Data migration after any bee update should have some fallback mechanism to prevent data corruption (?)

**Environment (please complete the following information):**  
- Bee Node: Upgraded from v2.5.0 → v2.6.0  
- Pinning CLI: `swarm-cli`  
- OS: Debian
- Running Bee as a service: Yes  

**Additional context**  
- A similar issue occurred about a year ago, a few months after erasure coding was introduced.  
- This suggests it may not be version-specific.  
- Possibility: the **first run after updating Bee** may have encountered an RPC endpoint failure, interrupting data migration (Bee as a service keeps retrying on failure).

Cafe137:
`ffe6d7fc805ede4581af1a01d9fcedf21332cfe3353d74a4961e53c360d06342`

This hash can be downloaded by any node, but cannot be pinned.

This hash has been uploaded with erasure coding.

nugaon:
Maybe the problem was the root chunk was missing from the network and it got replaced by a dispersed replica that the pinning service does not take into account <-> retrieval does not save recovered root chunk to cache, and the [pin does not retrieve the dispersed chunk from the network](https://github.com/ethersphere/bee/blob/23adc87da33784a34659464da3ff207ae3e70822/pkg/traversal/traversal.go#L60-L63)

PR #5230, opened 2025-09-26
fix(api): resolve pinning failures for erasure-coded chunk

gacevicljubisa:
### Checklist

### DIFF
--- a/pkg/api/chequebook.go
+++ b/pkg/api/chequebook.go
@@ -22,7 +22,6 @@ import (
 
 const (
 	errChequebookBalance           = "cannot get chequebook balance"
-	errChequebookNoAmount          = "did not specify amount"
 	errChequebookNoWithdraw        = "cannot withdraw"
 	errChequebookNoDeposit         = "cannot deposit"
 	errChequebookInsufficientFunds = "insufficient funds"
--- a/pkg/api/pin.go
+++ b/pkg/api/pin.go
@@ -10,8 +10,10 @@ import (
 	"net/http"
 	"sync"
 
+	"github.com/ethersphere/bee/v2/pkg/encryption/store"
 	"github.com/ethersphere/bee/v2/pkg/file/redundancy"
 	"github.com/ethersphere/bee/v2/pkg/jsonhttp"
+	"github.com/ethersphere/bee/v2/pkg/replicas"
 	"github.com/ethersphere/bee/v2/pkg/storage"
 	"github.com/ethersphere/bee/v2/pkg/storer"
 	"github.com/ethersphere/bee/v2/pkg/swarm"
@@ -52,7 +54,7 @@ func (s *Service) pinRootHash(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	getter := s.storer.Download(true)
+	getter := store.New(replicas.NewGetter(s.storer.Download(true), redundancy.DefaultLevel))
 	traverser := traversal.New(getter, s.storer.Cache(), redundancy.DefaultLevel)
 
 	sem := semaphore.NewWeighted(100)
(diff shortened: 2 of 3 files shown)


gacevicljubisa:
## Testing

### Steps to Reproduce the Issue:

1. Upload a file using erasure coding, resulting in a specific hash (e.g., `ffe6d7fc805ede4581af1a01d9fcedf21332cfe3353d74a4961e53c360d06342`)
2. Attempt to pin the file using the Bee API
3. Observe that the pinning operation fails, despite the file being downloadable

### Example Download Command:

```
curl localhost:1633/bzz/ffe6d7fc805ede4581af1a01d9fcedf21332cfe3353d74a4961e53c360d06342/ -o "wiki copy.png"
```

### Verification:

After applying this fix, the same hash should be pinable via:

```
curl -X POST localhost:1633/pins/ffe6d7fc805ede4581af1a01d9fcedf21332cfe3353d74a4961e53c360d06342
```

martinconic (approved):

pkg/api/chequebook.go
```diff
@@ -22,7 +22,6 @@ import (
 
 const (
 	errChequebookBalance           = "cannot get chequebook balance"
-	errChequebookNoAmount          = "did not specify amount"
```
zelig, line 25: how is this change relevant to the PR?
plese apply discipline

pkg/api/pss.go
```diff
@@ -24,8 +24,7 @@ import (
 )
 
 const (
-	writeDeadline   = 4 * time.Second // write deadline. should be smaller than the shutdown timeout on api close
-	targetMaxLength = 3               // max target length in bytes, in order to prevent grieving by excess computation
```
zelig, line 28: how is this change relavant to this PR?

pkg/api/pin.go
```diff
@@ -52,7 +53,7 @@ func (s *Service) pinRootHash(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	getter := s.storer.Download(true)
```
zelig, line 55: I do not think this does what we need. 
What caused the problem and how did you fix it?

Honestly, I dont this replicas (as opposed to parities) should be pinned at all, cos they are the same as the root chunk. When it comes to repair it should just upload the replicas too to swarm .

Also it is very suspicious that redundancy.DefaultLevel is used here. You know the level of a pinned file we can assume, no?

zelig (changes requested):
I dont think this should go in . we need:
- to explain what was the issue
- to explain why the issue does not appear now
- why redundancy.DefaultLevel is used

pkg/api/pss.go
gacevicljubisa: reverted

pkg/api/chequebook.go
gacevicljubisa: reverted

pkg/api/pin.go
gacevicljubisa: Idea behind this PR  was to addresses just specific [comment](https://github.com/ethersphere/bee/issues/5228#issuecomment-3333384881) in issue [#5228 ](https://github.com/ethersphere/bee/issues/5228). 

Thanks for clarifying that replicas shouldn't be pinned. 
I used redundancy.DefaultLevel as an assumption, and it is much used all over the codebase. Recently I created new [issue](https://github.com/ethersphere/bee/issues/5282) related to the use of redundancy.DefaultLevel and inconsistent `Swarm-Redundancy-Level` header.

I wasn't aware we can extract the redundancy level from the pinned file.

Can you suggest me the flow, how pinning should be handled if route chunk can not be found?

Also, how we can repair by uploading the replicas too to swarm?

PR #5244, opened 2025-10-11
fix: extract advertised addrs to set public reachability for multiple underlay addresses connectivity

gacevicljubisa:
### Checklist

### Description
This PR improves peer connectivity and reachability testing when nodes have multiple underlay addresses by using advertised addresses from the peerstore instead of connection addresses.
Fixes full-connectivity issue on `multiple-underlay-addresses` branch.

### DIFF
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -390,7 +390,7 @@ func (s *Service) filterAdvertisableUnderlays(underlays []ma.Multiaddr) []ma.Mul
 
 	var publicUnderlays []ma.Multiaddr
 	for _, u := range underlays {
-		if manet.IsPublicAddr(u) {
+		if !manet.IsPrivateAddr(u) {
 			publicUnderlays = append(publicUnderlays, u)
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/hive/hive.go
```diff
@@ -393,3 +381,21 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
...
+// based on the allowPrivateCIDRs setting. If allowPrivateCIDRs is false,
+// only public addresses are returned.
+func (s *Service) filterAdvertisableUnderlays(underlays []ma.Multiaddr) []ma.Multiaddr {
+	if s.allowPrivateCIDRs {
+		return underlays
+	}
+
+	var publicUnderlays []ma.Multiaddr
+	for _, u := range underlays {
+		if manet.IsPublicAddr(u) {
```
gacevicljubisa: @janos is this ok to check here if `IsPublicAddr`?

gacevicljubisa: Previously, it was `!manet.IsPrivateAddr(u)`

janos: I am not sure what to say exactly as these two functions are not exactly negation of eachother, given their implementations. It should be fine, but maybe to keep `!manet.IsPrivateAddr(u)`.

gacevicljubisa: So in that case `!manet.IsPrivateAddr()` would return `true` for loopback addresses?

PR #5256, opened 2025-10-27
feat: per-chunk stamped putter for WebSocket chunk uploads

agazso:
### Checklist

### Description
This PR adds support for the WebSocket chunk stream endpoint (`/chunks/stream`) to accept pre-signed (stamped) chunks, enabling client-side ECDSA stamp signing.

**Changes:**
- When `Swarm-Postage-Batch-Id` header is omitted, each WebSocket message must include a 113-byte stamp prefix before the chunk data
- Extract `newStampedPutterWithBatch` to avoid repeated batch DB lookups when batch info is already cached
- Cache batch metadata per WebSocket connection for performance
- Add `swarm-tag` query parameter fallback for browser WebSocket compatibility

#### Motivation and Context
The WebSocket chunk stream endpoint exists as a performance optimization for uploading chunks with less overhead than the HTTP endpoint. However, it currently does not support pre-signed chunks — the server always re-stamps using the batch ID from the connection header.

This prevents clients from doing client-side stamp signing, which is needed for:
- **Browser-based uploads** where the private key lives in the browser
- **Parallel stamp signing** using Web Workers for performance
- **Per-chunk batch selection** (different chunks can use different batches)

This change adds an opt-in mode: if the client omits the `Swarm-Postage-Batch-Id` header, the server expects each chunk message to arrive pre-stamped (113-byte stamp prefix + chunk data). The existing header-based mode continues to work unchanged.

#### Message format (per-chunk stamping mode)
```
[113 bytes: postage stamp] [8 bytes: span] [up to 4096 bytes: payload]
```

#### Decoder abstraction
A `chunkDecoder` function type selects the appropriate parsing strategy at connection time:
- `decodeChunkWithoutStamp` — used when `BatchID` header is present (existing behavior)
- `decodeChunkWithStamp` — used when `BatchID` header is omitted (new per-chunk mode)

#### Browser WebSocket compatibility
Browsers cannot set custom HTTP headers on WebSocket connections. The `swarm-tag` value can now be passed as a query parameter: `/chunks/stream?swarm-tag=123`.

### DIFF
--- a/pkg/api/chunk_stream.go
+++ b/pkg/api/chunk_stream.go
@@ -167,7 +167,7 @@ func (s *Service) handleUploadStream(
 	)
 
 	// Cache for batch validation to avoid database lookups for every chunk
-	// Key: batch ID hex string, Value: stored batch info
+	// Key: batch ID, Value: stored batch info
 	// This avoids the expensive batchStore.Get() call for each chunk
 	batchCache := make(map[string]*postage.Batch)
 


pkg/api/chunk_stream.go
```diff
@@ -55,29 +55,34 @@ func (s *Service) chunkUploadStreamHandler(w http.ResponseWriter, r *http.Reques
...
-		case errors.Is(err, postage.ErrNotFound):
-			jsonhttp.NotFound(w, "batch with id not found")
-		case errors.Is(err, errInvalidPostageBatch):
-			jsonhttp.BadRequest(w, "invalid batch id")
-		case errors.Is(err, errUnsupportedDevNodeOperation):
-			jsonhttp.BadRequest(w, errUnsupportedDevNodeOperation)
-		default:
-			jsonhttp.BadRequest(w, nil)
+	// Create connection-level putter only if BatchID is provided
+	// If BatchID is not provided, per-chunk stamps must be used
```
acud: I'm not sure I fully understand what `per-chunk stamps` are. Do you mean these are presigned stamps? If so, consider changing the comment to something like `if ...., the API caller is expected to provide pre-signed stamps (and is also expected to keep track over stamp state over time)`

pkg/api/chunk_stream.go
```diff
@@ -190,14 +208,109 @@ func (s *Service) handleUploadStream(
 			return
 		}
 
-		chunk, err := cac.NewWithDataSpan(msg)
+		logger.Debug("chunk upload stream",
```
acud: do we really need this message?

pkg/api/chunk_stream.go
```diff
@@ -190,14 +208,109 @@ func (s *Service) handleUploadStream(
...
+		// Check if this message contains a per-chunk stamp prepended to the chunk data
+		// Format: stamp[113 bytes] + chunk[data]
+		var (
+			chunk       swarm.Chunk
+			chunkPutter = putter // default to connection-level putter
+			chunkData   = msg
+		)
+
+		// If message is large enough to contain a stamp + chunk data, try to extract the stamp
+		if len(msg) >= postage.StampSize+swarm.SpanSize {
```
acud: Suggestion: reduce nesting here.
Compare:
```go
if ... {
 .....
 if ....
}
```
vs.
```go
if condition {
 return
}
....
if ...
```
Personally I find the latter easier to read. In the context here: check once that the size constraints match, return if not, then continue execution, no need to branch out.

pkg/api/chunk_stream.go
```diff
@@ -190,14 +208,109 @@ func (s *Service) handleUploadStream(
...
+			potentialStamp := msg[:postage.StampSize]
+			potentialChunkData := msg[postage.StampSize:]
+
+			logger.Debug("chunk upload stream: attempting to extract stamp",
```
acud: nit: probably can remove this once this is inching closer to where we want it to go

pkg/api/chunk_stream.go
```diff
@@ -190,14 +208,109 @@ func (s *Service) handleUploadStream(
...
+				"message_size", len(msg),
+				"stamp_size", postage.StampSize,
+				"potential_stamp_len", len(potentialStamp),
+				"potential_chunk_len", len(potentialChunkData),
+				"first_8_bytes", msg[:8])
+
+			// Try to unmarshal as a stamp
+			stamp := postage.Stamp{}
+			if err := stamp.UnmarshalBinary(potentialStamp); err == nil {
```
acud: So here I have some doubts about the approach, I think doing guesswork here is probably not necessary.
Let's break down the possibilities:
1. batch ID was provided - we will use 1 batch across the entire session and stamp all stamps. Chunks arrive without a stamp.
2. batch ID was not provided - stamps can arrive in theory with any batch ID. Stamps are prepended to chunk.

My issue with this is:
1. You have to guess whether the chunk is supposed to have a stamp or not (you might try to unmarshal the chunk as a stamp just because it falls, size-wise, as maybe a chunk with a stamp.
2. Then the rest of the conditional logic happens if it's not a stamp+chunk.

**But**, you already know. In `L61` you already know whether you should expect a stamp or not. This can be passed downstream either as a bool or not. Then have two functions that decode a message that have the same signature, e.g.:
```go
func decodeWithStamp(msg []byte) (stamp, chunk, err) {...}
func decodeNoStamp(msg []byte) (stamp, chunk, err) {...}
```
The latter will always return a nil stamp.
Then you still need to figure out which batch ID to use but that's not difficult.
I think that if you do a small changes here (including injecting the correct message decoder, the original code almost doesn't need to change, apart from the batch caching that you added. Hopefully this helps.

agazso:
> Thanks @agazso :) I have some suggestions on how this can be simplified

Hi @acud, I addressed your comments, please review again.

pkg/api/chunk_stream.go
```diff
@@ -28,14 +29,23 @@ func (s *Service) chunkUploadStreamHandler(w http.ResponseWriter, r *http.Reques
...
+		BatchID  []byte `map:"Swarm-Postage-Batch-Id"` // Optional: omit if caller provides pre-signed stamps per chunk
 		SwarmTag uint64 `map:"Swarm-Tag"`
 	}{}
 	if response := s.mapStructure(r.Header, &headers); response != nil {
 		response("invalid header params", logger, w)
 		return
 	}
 
+	// Fallback: read tag from query parameter (browser WebSocket can't set headers)
+	if headers.SwarmTag == 0 {
```
acud, line 41: i think this should be reflected in the OpenAPI spec. also, can't we send the headers in with the initial request before the connection gets upgraded somehow?

agazso: The browser [WebSocket API](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket) cannot send headers. Extra information can be sent either in the `url` or in `protocols`, to me it seems that sending the tag in the `url` is better.

https://stackoverflow.com/questions/4361173/http-headers-in-websockets-client-api

agazso: > i think this should be reflected in the OpenAPI spec. also, can't we send the headers in with the initial request before the connection gets upgraded somehow?

I updated the OpenAPI spec

pkg/api/chunk_stream.go
```diff
@@ -28,14 +29,23 @@ func (s *Service) chunkUploadStreamHandler(w http.ResponseWriter, r *http.Reques
...
+		if qTag := r.URL.Query().Get("swarm-tag"); qTag != "" {
+			if parsed, err := strconv.ParseUint(qTag, 10, 64); err == nil {
+				headers.SwarmTag = parsed
```
akrem-chabchoub: Why error is ignored here ?

akrem-chabchoub (commented):
Thanks @agazso 💯

akrem-chabchoub:
OpenAPI is failing in CI: https://github.com/ethersphere/bee/actions/runs/23790139097/job/69467764814?pr=5256

agazso:
> OpenAPI is failing in CI: https://github.com/ethersphere/bee/actions/runs/23790139097/job/69467764814?pr=5256

Seems to be a credential issue. @acud ?

pkg/api/chunk_stream.go
```diff
@@ -190,14 +252,71 @@ func (s *Service) handleUploadStream(
...
+		// Determine the putter to use
+		var (
+			chunk       swarm.Chunk
+			chunkPutter = putter
+		)
+
+		// If stamp was extracted, create a per-chunk putter
+		if stamp != nil {
+			batchID := stamp.BatchID()
+			batchIDHex := string(batchID)
```
gacevicljubisa: The variable name `batchIDHex` is misleading, `stamp.BatchID()` returns raw bytes, maybe rename to `batchIDKey` for clarity.

pkg/api/chunk_stream.go
agazso: Good catch, I added a fix!

gacevicljubisa (commented):
I would suggest adding at least some minim reasonable unit tests:
- Happy path for stamped mode
- Invalid stamp data

pkg/api/chunk_stream.go
agazso: Fixed

agazso:
> I would suggest adding at least some minim reasonable unit tests:
> 
>     * Happy path for stamped mode
> 
>     * Invalid stamp data

Thanks for the suggestion, I added two unit tests.

acud:
@agazso I think the openapi build is really due to the fork issue. [This](https://github.com/ethersphere/bee/pull/5421) PR rendered fine, so I think it should be fine to merge to `master`

pkg/api/chunk_stream.go
```diff
@@ -190,14 +256,71 @@ func (s *Service) handleUploadStream(
...
+		if err != nil {
+			logger.Debug("chunk upload stream: create chunk failed", "error", err, "chunk_size", len(chunkData))
 			logger.Error(nil, "chunk upload stream: create chunk failed")
+			sendErrorClose(websocket.CloseInternalServerErr, "invalid chunk data")
 			return
 		}
 
-		err = putter.Put(ctx, chunk)
+		err = chunkPutter.Put(ctx, chunk)
 		if err != nil {
```
akrem-chabchoub, line 327: Not sure if we can use `chunkPutter.Cleanup()` here ?

pkg/api/chunk_stream.go
```diff
@@ -190,14 +256,71 @@ func (s *Service) handleUploadStream(
...
+					sendErrorClose(websocket.CloseInternalServerErr, "batch not found")
+				default:
+					sendErrorClose(websocket.CloseInternalServerErr, "stamped putter creation failed")
+				}
+				return
+			}
+		}
+
+		chunk, err = cac.NewWithDataSpan(chunkData)
+		if err != nil {
```
akrem-chabchoub, line 316: Not sure if we can use `chunkPutter.Cleanup()` here ?

agazso:
> I'm not 100% sure, just I want to raise my concerns

I guess your concern may be that when there is an error the `chunkPutter` is not cleaned up properly and that may leak the session.

I think then the `cleanup` should be called when the `chunkPutter` is not the same as the `putter`.

Is that correct @akrem-chabchoub ?

akrem-chabchoub:
> > I'm not 100% sure, just I want to raise my concerns
> 
> I guess your concern may be that when there is an error the `chunkPutter` is not cleaned up properly and that may leak the session.
> 
> I think then the `cleanup` should be called when the `chunkPutter` is not the same as the `putter`.
> 
> Is that correct @akrem-chabchoub ?

@agazso Yes this is correct, If the putter session is the same we should not cleanup for ONE failed chunk I think.

pkg/api/chunk_stream.go
```diff
@@ -111,11 +165,23 @@ func (s *Service) handleUploadStream(
 		gone = make(chan struct{})
 		err  error
 	)
+
+	// Cache for batch validation to avoid database lookups for every chunk
+	// Key: batch ID hex string, Value: stored batch info
```
akrem-chabchoub: Small nit on the comment: it says the map key is a “batch ID hex string,” but the code uses `batchIDKey := string(batchID)`

PR #5257, opened 2025-10-28
Add validation for nat-addr

martinconic:
### Checklist

### Description
This PR adds validation for the `nat-addr` configuration option to ensure it is a valid and publicly routable address.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -1476,29 +1476,29 @@ func validatePublicAddress(addr string) error {
 
 	host, port, err := net.SplitHostPort(addr)
 	if err != nil {
-		return fmt.Errorf("invalid address: %w", err)
+		return fmt.Errorf("%w", err)
 	}
 	if host == "" {
-		return errors.New("invalid address: host is empty")
+		return errors.New("host is empty")
 	}
 	if port == "" {
-		return errors.New("invalid address: port is empty")
+		return errors.New("port is empty")
 	}
 	if _, err := strconv.ParseUint(port, 10, 16); err != nil {
-		return fmt.Errorf("invalid address: port is not a valid number: %w", err)
+		return fmt.Errorf("port is not a valid number: %w", err)
 	}
 	if host == "localhost" {
-		return errors.New("invalid address: localhost is not a valid address")
+		return errors.New("localhost is not a valid address")
 	}
 	ip := net.ParseIP(host)
 	if ip == nil {
-		return errors.New("invalid address: not a valid IP address")
+		return errors.New("not a valid IP address")
 	}
 	if ip.IsLoopback() {
-		return errors.New("invalid address: loopback address is not a valid address")
+		return errors.New("loopback address is not a valid address")
 	}
 	if ip.IsPrivate() {
-		return errors.New("invalid address: private address is not a valid address")
+		return errors.New("private address is not a valid address")
 	}
 
 	return nil
--- a/pkg/node/node_test.go
+++ b/pkg/node/node_test.go
@@ -87,8 +87,6 @@ func TestValidatePublicAddress(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			t.Parallel()
-
 			err := node.ValidatePublicAddress(tc.addr)
 			if tc.expErr && err == nil {
 				t.Fatal("expected error, but got none")


pkg/node/node_test.go
```diff
@@ -0,0 +1,56 @@
...
+			name:    "empty host",
+			natAddr: ":1635",
+		},
+		{
+			name:    "localhost",
+			natAddr: "localhost:1635",
+		},
+		{
+			name:    "loopback",
+			natAddr: "127.0.0.1:1635",
```
janos: Maybe IPv6 case as well.

pkg/node/node.go
```diff
@@ -222,6 +222,23 @@ func NewBee(
 		return nil, fmt.Errorf("tracer: %w", err)
 	}
 
+	if o.NATAddr != "" {
+		host, _, err := net.SplitHostPort(o.NATAddr)
```
gacevicljubisa: Should we check if the port is not empty and if it is able to parse it as a port number

janos: The port should be checked if it is valid, too.

pkg/node/node.go
```diff
@@ -222,6 +222,23 @@ func NewBee(
...
+		}
+		if host == "localhost" {
+			return nil, errors.New("invalid NAT address: localhost is not a valid NAT address")
+		}
+		ip := net.ParseIP(host)
+		if ip != nil && ip.IsLoopback() {
+			return nil, errors.New("invalid NAT address: loopback address is not a valid NAT address")
+		}
+	}
+
```
gacevicljubisa: Maybe extract this to a function

gacevicljubisa: And maybe we should move this to `libp2p` pacakge, because in the constructor, we already have the incomplete check? 

```
	host, port, err := net.SplitHostPort(addr)
	if err != nil {
		return nil, fmt.Errorf("address: %w", err)
	}
```

gacevicljubisa: @janos ?

martinconic: I would leave the function here, I think we will need to use it to check also for some other addresses, like API

pkg/node/node.go
```diff
@@ -1463,3 +1468,32 @@ func isChainEnabled(o *Options, swapEndpoint string, logger log.Logger) bool {
 	logger.Info("starting with an enabled chain backend")
 	return true // all other modes operate require chain enabled
 }
+
+func ValidatePublicAddress(addr string) error {
```
janos: This function does not need to be exported.

pkg/node/node.go
```diff
@@ -1463,3 +1468,32 @@ func isChainEnabled(o *Options, swapEndpoint string, logger log.Logger) bool {
...
+	if port == "" {
+		return errors.New("invalid address: port is empty")
+	}
+	if _, err := strconv.ParseUint(port, 10, 16); err != nil {
+		return fmt.Errorf("invalid address: port is not a valid number: %w", err)
+	}
+	if host == "localhost" {
+		return errors.New("invalid address: localhost is not a valid address")
+	}
+	ip := net.ParseIP(host)
```
gacevicljubisa, line 1493: If host is invalid, then `net.ParseIP(host)` will return nil. In that case we need to return error.

pkg/node/node_test.go
```diff
@@ -0,0 +1,86 @@
+// Copyright 2023 The Swarm Authors. All rights reserved.
```
gacevicljubisa: Copyright 2025

pkg/node/node_test.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"testing"
+
+	"github.com/ethersphere/bee/v2/pkg/node"
+)
+
+func TestValidatePublicAddress(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
```
gacevicljubisa, line 16: invalid host is missing.

pkg/node/node.go
gacevicljubisa: Should we use it again in libp2p constructor?

pkg/node/node.go
```diff
@@ -222,6 +223,10 @@ func NewBee(
 		return nil, fmt.Errorf("tracer: %w", err)
 	}
 
+	if err := validatePublicAddress(o.NATAddr); err != nil {
+		return nil, err
```
gacevicljubisa: format err that it should print the o.NATAddr, that will enable to double check what is the actual 
o.NATAddr used, to be more obvious. 
This can happen when you assign flag value with some script.

janos (approved):

gacevicljubisa (approved):

Issue #5232, opened 2025-10-01
Use the new testing/synctest package from golang 1.25 in tests


martinconic:
#### Summary
Use the `New testing/synctest package` that should help us with improving tests and fixing Flaky ones. [testing/synctest package](https://go.dev/doc/go1.25#new-testingsynctest-package)

PR #5262, opened 2025-10-29
Add synctest in storage incentives

akrem-chabchoub:
### Checklist

### Description
Add synctest in storage incentives.
Remove parallel execution from tests (synctest).
### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5232

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,17 +121,17 @@ endif
 .PHONY: test-ci
 test-ci:
 ifdef cover
-	$(GO) test -run "[^FLAKY]$$" -coverprofile=cover.out ./pkg/storageincentives
+	$(GO) test -run "[^FLAKY]$$" -coverprofile=cover.out ./...
 else
-	$(GO) test -run "[^FLAKY]$$" ./pkg/storageincentives
+	$(GO) test -run "[^FLAKY]$$" ./...
 endif
 
 .PHONY: test-ci-race
 test-ci-race:
 ifdef cover
-	$(GO) test -race -run "[^FLAKY]$$" -coverprofile=cover.out ./pkg/storageincentives
+	$(GO) test -race -run "[^FLAKY]$$" -coverprofile=cover.out ./...
 else
-	$(GO) test -race -run "[^FLAKY]$$" ./pkg/storageincentives
+	$(GO) test -race -run "[^FLAKY]$$" ./...
 endif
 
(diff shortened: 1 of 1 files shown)


acud (approved):
I don't think there's any problem with having both `t.Parallel()` and `synctest.Test` together. It is also evident that it is done in other PRs. Not sure if there's anything preventing us from having both.

janos:
> I don't think there's any problem with having both `t.Parallel()` and `synctest.Test` together. It is also evident that it is done in other PRs. Not sure if there's anything preventing us from having both.

There is also no benefit from having t.Parallel() either in most cases. When running tests, the total execution time is the same with or without parallel tests as tests are no longer waiting with synctest.

I would say that t.Parallel() should be added only in places where the improvement exist, and even to review such extensive usage of t.Parallel() in bee tests.

acud:
> There is also no benefit from having t.Parallel() either in most cases. When running tests, the total execution time is the same with or without parallel tests as tests are no longer waiting with synctest.
> 
> I would say that t.Parallel() should be added only in places where the improvement exist, and even to review such extensive usage of t.Parallel() in bee tests.

Thanks @janos for the clarification. However still, even if the particular tests "don't wait", shouldn't we get faster test-suite execution for unit tests just because the runners can crunch more tests per time unit due to multi-threading? I understand the consideration regarding particular tests, however I'm talking about the test-suite time efficiency in general.

janos:
> > There is also no benefit from having t.Parallel() either in most cases. When running tests, the total execution time is the same with or without parallel tests as tests are no longer waiting with synctest.
> > I would say that t.Parallel() should be added only in places where the improvement exist, and even to review such extensive usage of t.Parallel() in bee tests.
> 
> Thanks @janos for the clarification. However still, even if the particular tests "don't wait", shouldn't we get faster test-suite execution for unit tests just because the runners can crunch more tests per time unit due to multi-threading? I understand the consideration regarding particular tests, however I'm talking about the test-suite time efficiency in general.

As always, it depends, if the scheduler is taking the same time to schedule tests and goroutines as to run tests, total execution may even take longer. This should always be measured for particular cases. We did some measurement of the test in some packages and saw no difference in speed even on local executions with many cpu cores.

acud:
Cool. Thanks @janos. Scratch my comment then :) :pray:

PR #5264, opened 2025-10-29
test(pingpong): add synctest to ping test

akrem-chabchoub:
### Checklist

### Description
Add synctest in ping test.
### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5232

### DIFF
--- a/pkg/pingpong/pingpong_test.go
+++ b/pkg/pingpong/pingpong_test.go
@@ -48,7 +48,7 @@ func TestPing(t *testing.T) {
 		}
 
 		// check that RTT is the sum of all message latencies
-		if rtt != 6*messageLatency {
+		if rtt != time.Duration(2*len(greetings))*messageLatency {
 			t.Errorf("invalid RTT value %v", rtt)
 		}
 


pkg/pingpong/pingpong_test.go
```diff
@@ -8,106 +8,95 @@ import (
...
-	if err != nil {
-		t.Fatal(err)
-	}
+		// check that RTT is a sane value (rtt can be 0 in synctest virtual time)
+		if rtt < 0 {
+			t.Errorf("invalid RTT value %v", rtt)
+		}
 
-	// check that RTT is a sane value
-	if rtt <= 0 {
```
janos, line 59: This test must evaluate that the rtt is not 0 as well, as zero value can also signal that the protocol does not work.

It could be now made that the protocol waits when responding to messages with the middleware for any os (to remove the goos check), that and to check if the rtt value is measured correctly.

akrem-chabchoub: I tried to check the `rtt` is not zero as well, but the tests are failing. I think as I already mentioned since time is virtual in `synctest` so the the `rtt` can be zero.
https://github.com/ethersphere/bee/actions/runs/19616730484/job/56172634154

janos: Correct, we have to make the time to advance in order to measure rtt in the bubble, by blocking something. I thought thought that it is good to use the middleware that you removed that sleeps, but for any os. Test is working with that approach, but it actually does not make the test correct as it measures the middleware, and we want to measure messages reading and writing latencies. So, I have added the option to recorder to introduce the message latency so that we can measure rtt, now even exactly, which is much better than before.

akrem-chabchoub: Ok, good.
 [Here](https://github.com/ethersphere/bee/pull/5264/commits/cc7e9407b5536c659ff1beb4285298101bd27545#diff-25c2552e844cd58595936f53944451fd5be87daac81493aceeb95ed9dcfdaaceR51) it might be better to compare `rtt` as `2 * len(greetings) * messageLatency` ?

janos: Good point. Updated.

PR #5279, opened 2025-11-06
feat: soc dispersed replica v2

gacevicljubisa:
### Checklist

### Description

This PR is a continuation of the original PR https://github.com/ethersphere/bee/pull/5057.

#### Original description

Add support for dispersed replicas on Single Owner Chunks (SOCs) to improve data availability and retrieval reliability in the Swarm network.

SOC replicas are generated by allowing additional addresses to represent the same SOC. It is achieved by lighten the validation of SOCs which ignores the first byte of the address. This makes possible to saturate dispersed replicas evenly across the whole network since nodes arrange into neighborhoods based on address prefix. The addresses created in a way to iterate over all variations in the given depth of the redundancy level + 1 (e.g. level is 2, and the original address starts with 101, then it uploads SOCs with addresses same after the 3 first bits where the first 3 bit variations are 001, 011, (101 is not because the original address has it), 111 and 100 (flipping the last bit).

#### New changes

Fixes SOC dispersed replica functionality by setting explicit default redundancy levels and fixing implementation issues.

- Set default redundancy level to `PARANOID` for handlers
- Use `wg.Go()` instead of manual goroutine management
- Change redundancy level header to pointer type for proper optional handling
- Refactor feeds factory to use functional options pattern
- Add comprehensive tests for SOC replicas
- Introduce the iterator for iterating over replica addreses one by one for better concurrency management and as the original implementation had bugs

feed and soc replica PUT endpoints `swarm-redundancy-level` header: create and push dispersed replicas according to the passed level: MEDIUM 2, STRONG 4, INSANE 8, PARANOID 16.

feed and soc replica GET endpoints `swarm-redundancy-level` header: for calibrating how deeply dispersed replicas should be checked. By default it is zero.

### DIFF
--- a/pkg/api/pin.go
+++ b/pkg/api/pin.go
@@ -53,7 +53,7 @@ func (s *Service) pinRootHash(w http.ResponseWriter, r *http.Request) {
 	}
 
 	getter := s.storer.Download(true)
-	traverser := traversal.New(getter, s.storer.Cache(), redundancy.PARANOID)
+	traverser := traversal.New(getter, s.storer.Cache(), redundancy.DefaultLevel)
 
 	sem := semaphore.NewWeighted(100)
 	var errTraverse error
--- a/pkg/hive/hive_test.go
+++ b/pkg/hive/hive_test.go
@@ -147,7 +147,7 @@ func TestBroadcastPeers(t *testing.T) {
 			underlays = []ma.Multiaddr{u, u2}
 		} else {
 			n := (i % 3) + 1
-			for j := range n {
+			for j := 0; j < n; j++ {
 				port := i + j*10000
 				u, err := ma.NewMultiaddr("/ip4/127.0.0.1/udp/" + strconv.Itoa(port))
 				if err != nil {
--- a/pkg/replicas/getter_soc.go
+++ b/pkg/replicas/getter_soc.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/ethersphere/bee/v2/pkg/file/redundancy"
 	"github.com/ethersphere/bee/v2/pkg/replicas/combinator"
+	"github.com/ethersphere/bee/v2/pkg/soc"
 	"github.com/ethersphere/bee/v2/pkg/storage"
 	"github.com/ethersphere/bee/v2/pkg/swarm"
 	"golang.org/x/sync/semaphore"
@@ -57,32 +58,44 @@ func (g *socGetter) Get(ctx context.Context, addr swarm.Address) (ch swarm.Chunk
 	}
 
 	// Try to retrieve replicas.
+	// Context for cancellation of replica fetching.
+	// Once a replica is found, this context is cancelled to stop further replica requests.
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	// sem is used to limit the number of concurrent replica fetch operations.
 	sem := semaphore.NewWeighted(socGetterConcurrency)
 	replicaIter := combinator.IterateReplicaAddresses(addr, int(g.level))
 
+	// resultChan is used to send the first successfully fetched chunk back to the main goroutine.
 	resultChan := make(chan swarm.Chunk, 1)
+	// doneChan signals when all replica iteration and fetching attempts have concluded.
 	doneChan := make(chan struct{})
 
+	// This goroutine iterates through potential replica addresses and dispatches
+	// concurrent fetch operations, respecting the concurrency limit.
 	go func() {
-		defer close(doneChan)
+		defer close(doneChan) // Ensure doneChan is closed when all replica attempts are finished.
 		for replicaAddr := range replicaIter {
 			select {
 			case <-ctx.Done():
+				// If the context is cancelled (e.g., a replica was found or parent context cancelled),
+				// stop dispatching new replica requests.
 				return
 			default:
 			}
 
+			// Acquire a semaphore slot to limit concurrency.
 			if err := sem.Acquire(ctx, 1); err != nil {
+				// If context is cancelled while acquiring, stop.
 				return
 			}
 
 			wg.Add(1)
+			// Each replica fetch is performed in its own goroutine.
 			go func(replicaAddr swarm.Address) {
-				defer sem.Release(1)
-				defer wg.Done()
+				defer sem.Release(1) // Release the semaphore slot when done.
+				defer wg.Done()      // Decrement the WaitGroup counter.
 
 				ch, err := g.Getter.Get(ctx, replicaAddr)
 				if err != nil {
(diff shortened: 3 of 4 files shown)


pkg/api/feed.go
```diff
@@ -66,15 +67,21 @@ func (s *Service) feedGetHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	headers := struct {
-		OnlyRootChunk bool `map:"Swarm-Only-Root-Chunk"`
+		OnlyRootChunk   bool             `map:"Swarm-Only-Root-Chunk"`
+		RedundancyLevel redundancy.Level `map:"Swarm-Redundancy-Level"`
```
gacevicljubisa, line 71: Should we use here Paranoid as default as on other endpoints?

gacevicljubisa: What about all other places where `Swarm-Redundancy-Level` is used as header?

janos: This is a good question. In all other handlers getRedundancyLevel is used to get the redundancy from header and if it is not set, PARANOID is used. Here in feeds, replicas getter is not used if the header is not set. I do not know it that is the expected behaviour. 

Maybe Viktor should know when he reviews this PR.

I am not sure if we need `headers.RedundancyLevel > redundancy.NONE` at all, as replicas getter should not make any replicas requests for NONE redundancy.

gacevicljubisa: Regarding `redundancy.NONE`, yes, that make sense, maybe we can directly use:
```go
	getter = replicas.NewSocGetter(s.storer.Download(true), rLevel)
```

janos: I agree.

janos:
Hi @zelig. I would kindly ask you to review this PR for correctness and in general. It is a combined effort from multiple developers and we have tried to validate the behaviour through unit tests, but we would like to have your validation because of potential risks. Would you check also what should be default redundancy levels in api handlers if Swarm-Redundancy-Level is not set in the request?

pkg/replicas/getter_soc.go
```diff
@@ -0,0 +1,116 @@
...
+				defer wg.Done()
+
+				ch, err := g.Getter.Get(ctx, replicaAddr)
+				if err != nil {
+					mu.Lock()
+					errs = errors.Join(errs, fmt.Errorf("get chunk replica address %v: %w", replicaAddr, err))
+					mu.Unlock()
+					return
+				}
+
```
janos, line 107: @zelig Would it make sense to check if the chunk data content validates with the requested address `addr`? Like this:

```cac.Valid(swarm.NewChunk(addr, ch.Data()))```

As the replica's data hash should be equal with the chunk original address.

File: .github/workflows/beekeeper.yml [hunk 3]
```diff
@@ -14,7 +14,7 @@ env:
   SETUP_CONTRACT_IMAGE: "ethersphere/bee-localchain"
   SETUP_CONTRACT_IMAGE_TAG: "0.9.4"
   BEELOCAL_BRANCH: "main"
-  BEEKEEPER_BRANCH: "master"
+  BEEKEEPER_BRANCH: "feat/soc-dispersed"
```
zelig, line 17: Why?

pkg/api/pin.go
```diff
@@ -53,7 +53,7 @@ func (s *Service) pinRootHash(w http.ResponseWriter, r *http.Request) {
 	}
 
 	getter := s.storer.Download(true)
-	traverser := traversal.New(getter, s.storer.Cache(), redundancy.DefaultLevel)
+	traverser := traversal.New(getter, s.storer.Cache(), redundancy.PARANOID)
```
zelig: why is this change?

pkg/hive/hive_test.go
```diff
@@ -147,7 +147,7 @@ func TestBroadcastPeers(t *testing.T) {
 			underlays = []ma.Multiaddr{u, u2}
 		} else {
 			n := (i % 3) + 1
-			for j := 0; j < n; j++ {
+			for j := range n {
```
zelig: what does this change have anyhting to do with the PR?

pkg/replicas/getter.go
```diff
@@ -51,16 +50,19 @@ func (g *getter) Get(ctx context.Context, addr swarm.Address) (ch swarm.Chunk, e
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	var wg sync.WaitGroup
+	defer wg.Wait()
```
zelig, line 54: why would you want to wait for all routines to terminate? unless you cancel them each when you get one response

pkg/replicas/getter_soc.go
zelig: yes absolutely

pkg/topology/kademlia/kademlia_test.go
```diff
@@ -2244,7 +2244,7 @@ func generateMultipleUnderlays(t *testing.T, n int, baseUnderlay string) []ma.Mu
 	t.Helper()
 	underlays := make([]ma.Multiaddr, n)
 
-	for i := 0; i < n; i++ {
+	for i := range n {
```
zelig: why here, why now?

.github/workflows/beekeeper.yml
janos: This is for CI to use the updated beekeeper checks. Before this PR is merged, beekeeper main will be up to date. If beekeeper main has the changes for this PR, other PRs CI checks would fail.

pkg/api/pin.go
janos: Reverted.

pkg/hive/hive_test.go
janos: Reverted.

pkg/replicas/getter.go
janos: This is actually needed as a test hack (to use [replicas.Wait](https://github.com/ethersphere/bee/pull/5279/files#diff-9359d2306bcf6654d55d233801f6066e238f04ecdbf8d02b7259fe25652ad612L174) only avalilable in tests) to avoid data races is removed by nugaon. The current code is race free in production, as well.

Canceling is done when when the chunk is found by calling cancel() on line 92.

If we do not care about other goroutines, we can revert changes in getter.go and getter_test.go and export_test.go so that tests do not report data race. Should we do that?

pkg/replicas/getter_soc.go
janos: Thanks, it is added.

pkg/topology/kademlia/kademlia_test.go
janos: Reverted.

.github/workflows/beekeeper.yml
gacevicljubisa: This is the beekeeper branch ([PR](https://github.com/ethersphere/beekeeper/pull/546)) where are modifications that are able to test soc disparsed replica on CI. This needs to be reverted to master before merge. Also, beekeeper PR should be merged as well.

acud:
Closing this due to lack of activity for the past 4 months. Please reopen "if and when"

PR #5281, opened 2025-11-13
fix(joiner): resolve race condition

gacevicljubisa:
### Checklist

### Description
Fixes a race condition in pkg/file/joiner that caused CI test timeouts. Refactored decoderCache.GetOrCreate to prevent deadlocks by correctly managing mutex locks around getter.New calls.

Makefile: Filter linker warnings and disable test caching for test-ci-race

### DIFF
--- a/pkg/file/joiner/joiner.go
+++ b/pkg/file/joiner/joiner.go
@@ -230,8 +230,8 @@ func (j *joiner) ReadAt(buffer []byte, off int64) (read int, err error) {
 
 	readLen := min(int64(cap(buffer)), j.span-off)
 	var bytesRead int64
-	var eg errgroup.Group
-	j.readAtOffset(buffer, j.rootData, 0, j.span, off, 0, readLen, &bytesRead, j.rootParity, &eg)
+	eg, ectx := errgroup.WithContext(j.ctx)
+	j.readAtOffset(buffer, j.rootData, 0, j.span, off, 0, readLen, &bytesRead, j.rootParity, eg, ectx)
 
 	err = eg.Wait()
 	if err != nil {
@@ -249,6 +249,7 @@ func (j *joiner) readAtOffset(
 	bytesRead *int64,
 	parity int,
 	eg *errgroup.Group,
+	ectx context.Context,
 ) {
 	// we are at a leaf data chunk
 	if subTrieSize <= int64(len(data)) {
@@ -299,7 +300,12 @@ func (j *joiner) readAtOffset(
 
 		func(address swarm.Address, b []byte, cur, subTrieSize, off, bufferOffset, bytesToRead, subtrieSpanLimit int64) {
 			eg.Go(func() error {
-				ch, err := g.Get(j.ctx, addr)
+				select {
+				case <-ectx.Done():
+					return ectx.Err()
+				default:
+				}
+				ch, err := g.Get(ectx, addr)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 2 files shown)


martinconic (approved):

pkg/file/joiner/joiner.go
```diff
@@ -211,8 +230,8 @@ func (j *joiner) ReadAt(buffer []byte, off int64) (read int, err error) {
 
 	readLen := min(int64(cap(buffer)), j.span-off)
 	var bytesRead int64
-	var eg errgroup.Group
-	j.readAtOffset(buffer, j.rootData, 0, j.span, off, 0, readLen, &bytesRead, j.rootParity, &eg)
+	eg, ectx := errgroup.WithContext(j.ctx)
+	j.readAtOffset(buffer, j.rootData, 0, j.span, off, 0, readLen, &bytesRead, j.rootParity, eg, ectx)
```
acud, line 234: `ectx==ctx` and so therefore it is not needed to add it to the signature, since `readAtOffset` can just select on the struct field

pkg/file/joiner/joiner.go
```diff
@@ -280,7 +300,12 @@ func (j *joiner) readAtOffset(
 
 		func(address swarm.Address, b []byte, cur, subTrieSize, off, bufferOffset, bytesToRead, subtrieSpanLimit int64) {
 			eg.Go(func() error {
-				ch, err := g.Get(j.ctx, addr)
+				select {
+				case <-ectx.Done():
+					return ectx.Err()
+				default:
+				}
+				ch, err := g.Get(ectx, addr)
```
acud, line 308: If the `Get` method already gets passed the context, shouldn't/wouldn't it better that it just checks whether the context expired before+after doing anything? this way you get better coverage of the whole codebase instead of having to dot every usage of the method elsewhere.

pkg/file/joiner/joiner.go
```diff
@@ -105,15 +112,20 @@ func (g *decoderCache) GetOrCreate(addrs []swarm.Address, shardCnt int) storage.
 			// Create a factory function that will instantiate the decoder only when needed
 			recovery := func() storage.Getter {
 				g.config.Logger.Debug("lazy-creating recovery decoder after fetch failed", "key", key)
+
+				if d, ok := g.getFromCache(key); ok && d != nil {
```
acud, line 116: since `getter.New` should return almost immediately, wouldn't it be better to just lock once and do both the get and the set, instead of locking twice, checking twice and potentially calling (and allocating) in vain to `getter.New`?
Bear in mind also that `getter.New` has side-effects as it starts a whole set of goroutines with `go d.prefetch()`. It would be good to avoid it if it's not necessary.

pkg/file/joiner/joiner.go
```diff
@@ -93,9 +101,8 @@ func (g *decoderCache) GetOrCreate(addrs []swarm.Address, shardCnt int) storage.
 	}
 
 	key := fingerprint(addrs)
-	g.mu.Lock()
```
acud, line 96: Just my 2-cents: since one mutex lock-unlock cycle turns into multiple, the chances of having race conditions increases. Since multiple callers may have now access to the same critical section at the same time, this might cause other problems elsewhere down the line (once some of the interleaving logic changes).

gacevicljubisa:
Thanks for the review! Upon testing on the current master branch, the race condition that this PR was intended to address no longer appears to be present. Closing this PR as it's no longer needed.

Issue #5247, opened 2025-10-13
Deprecation notice about dev-bee


bcsorvasi:
Add warnings in Bee v2.7 when running dev-bee command.

PR #5283, opened 2025-11-18
docs(bee-dev): update dev command with deprecation warning and doc link

akrem-chabchoub:
### Checklist

### Description
Add warning in Bee when running `bee dev`, `bee help` and `bee dev --help` commands.

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5247
image

image

image

### DIFF
--- a/cmd/bee/cmd/start_dev.go
+++ b/cmd/bee/cmd/start_dev.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/node"
 	"github.com/kardianos/service"
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
 )
 
 func (c *command) initStartDevCmd() (err error) {
@@ -68,8 +69,11 @@ func (c *command) initStartDevCmd() (err error) {
 ║                                                                              ║
 ╚══════════════════════════════════════════════════════════════════════════════╝
 `
-			// Apply yellow color to the warning box (works on modern terminals)
-			fmt.Print("\u001b[33m" + warningBox + "\u001b[0m")
+			if term.IsTerminal(int(os.Stdout.Fd())) {
+				fmt.Print("\u001b[33m" + warningBox + "\u001b[0m")
+			} else {
+				fmt.Print(warningBox)
+			}
 			fmt.Println()
 
 			fmt.Println("Starting in development mode")


gacevicljubisa:
This warning box looks very nice. 

I would mark command as DEPRECATED in short cmd. In long cmd description, I would add the mentioned warning and note that it will be removed in future versions. 

@bcsorvasi what do you think?

cmd/bee/cmd/start_dev.go
```diff
@@ -54,6 +55,23 @@ func (c *command) initStartDevCmd() (err error) {
...
+║                                                                              ║
+║  For more information, please refer to the official documentation:           ║
+║  https://docs.ethswarm.org/docs/develop/tools-and-features/bee-dev-mode      ║
+║                                                                              ║
+║  Please check back regularly for updates.                                    ║
+║                                                                              ║
+╚══════════════════════════════════════════════════════════════════════════════╝
+`
+			// Apply yellow color to the warning box (works on modern terminals)
+			fmt.Print("\u001b[33m" + warningBox + "\u001b[0m")
```
janos: It should be checked if the output is a terminal and only then to print colors. If it is redirected to the log file it should not print colors, or maybe not to use colors at all?

cmd/bee/cmd/start_dev.go
```diff
@@ -54,6 +54,23 @@ func (c *command) initStartDevCmd() (err error) {
 
 			fmt.Println(beeASCII)
 			fmt.Println()
+
+			warningBox := `
+╔══════════════════════════════════════════════════════════════════════════════╗
```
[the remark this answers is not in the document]
janos: I am not sure if this makes the output prettier.

bcsorvasi:
> This warning box looks very nice.
> 
> I would mark command as DEPRECATED in short cmd. In long cmd description, I would add the mentioned warning and note that it will be removed in future versions.
> 
> @bcsorvasi what do you think?

@gacevicljubisa I would use the original "will be deprecated" phrasing in the short since it is not yet deprecated, it is rather a grace period. 

And yes, the long version and the warning box looks pretty nice!

gacevicljubisa:
> > This warning box looks very nice.
> > I would mark command as DEPRECATED in short cmd. In long cmd description, I would add the mentioned warning and note that it will be removed in future versions.
> > @bcsorvasi what do you think?
> 
> @gacevicljubisa I would use the original "will be deprecated" phrasing in the short since it is not yet deprecated, it is rather a grace period.

@bcsorvasi when we mark it `deprecated` we do not need to remove functionality. Functionality can be removed in the future, which can be noted in the long description. 

But if it is already agreed how to approach this, I do not mind.

cmd/bee/cmd/start_dev.go
akrem-chabchoub: Good point Janos, i fixed it by checking if its a terminal before using colors

janos: Thank you, Akrem.

PR #5290, opened 2025-11-23
Add WebRTC addresses

lat-murmeldjur:
This PR has two parts: 
- WebRTC addresses added in libp2p node setup
- go-libp2p related dependencies' versions bumped to newer versions that make WebRTC connections work

WebRTC advantages:
- Zero configuration needed for security, works from secure browser contexts even on android chrome
- Functioning transport alternative for browser based nodes

WebRTC disadvantages:
- 6-7 seconds before connections become usable (need to wait this much in browser node after dial but before initiating handshake towards the bee node)
- Random limitations such as only available from browser tab but not from browser shared web worker

### DIFF
--- a/pkg/p2p/libp2p/internal/handshake/handshake.go
+++ b/pkg/p2p/libp2p/internal/handshake/handshake.go
@@ -153,7 +153,7 @@ func (s *Service) Handshake(ctx context.Context, stream p2p.Stream, peerMultiadd
 
 	w, r := protobuf.NewWriterAndReader(stream)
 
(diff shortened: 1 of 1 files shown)


acud:
@lat-murmeldjur iirc webrtc previously caused all sorts of problems with bee nodes. I think that integrating this would currently be a significant effort for us to pursue (there's quite a bit of testing involved here). I would suggest to try and live with the existing WSS support

acud:
closing due to lack of activity

PR #5309, opened 2025-12-26
fix(transaction): prevent concurrent map access in monitor

mfw78:
### Checklist

### Description

Fixes a race condition in `pkg/transaction/monitor.go` that causes `fatal error: concurrent map iteration and map write` panics.

The `checkPending()` function was iterating over `watchesByNonce` without holding the lock, while `WatchTransaction()` could concurrently modify the map. Classic Go map concurrency footgun. 🔫

**Solution:** Snapshot-based approach that minimizes lock contention:
1. **Phase 1** (lock held): Snapshot transaction data needed for RPC calls
2. **Phase 2** (lock released): Make slow RPC calls (`TransactionReceipt`, `NonceAt`)
3. **Phase 3** (lock held): Notify subscribers and cleanup

This avoids blocking `WatchTransaction()` callers during potentially slow (100-500ms) RPC operations.

N/A

#### Motivation and Context

Discovered while running Bee nodes in a Kurtosis test environment. The panic occurred during chequebook deployment when transaction monitoring was active.

```
fatal error: concurrent map iteration and map write

goroutine 182 [running]:
github.com/ethersphere/bee/v2/pkg/transaction.(*transactionMonitor).checkPending(0x4000ab8d00, 0xd46)
        /app/pkg/transaction/monitor.go:198 +0x334
```

### Related Issue (Optional)

N/A

N/A

### DIFF
--- a/pkg/transaction/monitor.go
+++ b/pkg/transaction/monitor.go
@@ -190,46 +190,52 @@ func watchStart(watches []transactionWatch) time.Time {
 	return start
 }
 
-// check pending checks the given block (number) for confirmed or cancelled transactions
+// checkPending checks the given block for confirmed or cancelled transactions.
 func (tm *transactionMonitor) checkPending(block uint64) error {
-	confirmedNonces := make(map[uint64]*types.Receipt)
-	var cancelledNonces []uint64
-	for nonceGroup, watchMap := range tm.watchesByNonce {
+	// Snapshot nonces and tx hashes to check (releases lock during slow RPC calls).
+	tm.lock.Lock()
+	snapshot := make(map[uint64]map[common.Hash]time.Time, len(tm.watchesByNonce))
+	for nonce, watchMap := range tm.watchesByNonce {
+		snapshot[nonce] = make(map[common.Hash]time.Time, len(watchMap))
 		for txHash, watches := range watchMap {
+			snapshot[nonce][txHash] = watchStart(watches)
+		}
+	}
+	tm.lock.Unlock()
+
+	// Check receipts without holding lock (RPC calls can be slow).
+	confirmedNonces := make(map[uint64]*types.Receipt)
+	for nonce, txMap := range snapshot {
+		for txHash, start := range txMap {
 			receipt, err := tm.backend.TransactionReceipt(tm.ctx, txHash)
 			if err != nil {
-				// wait for a few blocks to be mined before considering a transaction not existing
-				transactionWatchNotFoundTimeout := 5 * tm.pollingInterval
-				if errors.Is(err, ethereum.NotFound) && watchStart(watches).Before(time.Now().Add(transactionWatchNotFoundTimeout)) {
-					// if both err and receipt are nil, there is no receipt
-					// the reason why we consider this only potentially cancelled is to catch cases where after a reorg the original transaction wins
+				if errors.Is(err, ethereum.NotFound) && start.Before(time.Now().Add(5*tm.pollingInterval)) {
 					continue
 				}
 				return err
 			}
 			if receipt != nil {
-				// if we have a receipt we have a confirmation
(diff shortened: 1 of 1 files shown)


pkg/transaction/monitor.go
```diff
@@ -190,46 +190,52 @@ func watchStart(watches []transactionWatch) time.Time {
...
-					// the reason why we consider this only potentially cancelled is to catch cases where after a reorg the original transaction wins
+				if errors.Is(err, ethereum.NotFound) && start.Before(time.Now().Add(5*tm.pollingInterval)) {
 					continue
 				}
 				return err
 			}
 			if receipt != nil {
-				// if we have a receipt we have a confirmation
-				confirmedNonces[nonceGroup] = receipt
+				confirmedNonces[nonce] = receipt
```
gacevicljubisa, line 218: I think we could stop checking other txHashes here, for this nonce, because we found the receipt?

pkg/transaction/monitor.go
```diff
@@ -190,46 +190,52 @@ func watchStart(watches []transactionWatch) time.Time {
...
-	for nonceGroup := range tm.watchesByNonce {
-		if _, ok := confirmedNonces[nonceGroup]; ok {
+	// Check for cancellations.
+	var cancelledNonces []uint64
+	for nonce := range snapshot {
+		if _, ok := confirmedNonces[nonce]; ok {
 			continue
 		}
-
 		oldNonce, err := tm.backend.NonceAt(tm.ctx, tm.sender, new(big.Int).SetUint64(block-tm.cancellationDepth))
```
gacevicljubisa, line 229: I do not see the reason to call `tm.backend.NonceAt` multiple times in the loop, this should be lazy loaded (we should only fetch oldNonce once and only if there are unconfirmed nonces to check). Maybe we can use this PR to improve this?

akrem-chabchoub (approved):

pkg/transaction/monitor.go
```diff
@@ -190,46 +190,59 @@ func watchStart(watches []transactionWatch) time.Time {
...
+				if errors.Is(err, ethereum.NotFound) && start.Before(time.Now().Add(5*tm.pollingInterval)) {
 					continue
 				}
 				return err
 			}
 			if receipt != nil {
-				// if we have a receipt we have a confirmation
-				confirmedNonces[nonceGroup] = receipt
+				confirmedNonces[nonce] = receipt
+				break // found receipt for this nonce, no need to check other tx hashes
```
acud: @mfw78 we discussed this change and figured that this change, being orthogonal to the original PR should be reverted. Ideally let's just stick to the concurrent map access related changes. Thanks!

acud:
I would also suggest that the other changes should be reverted. This component is critical and I'm not sure we have unit test coverage to see that these extra changes are inconsequential. I'd suggest to stick to the original changeset.

mfw78:
> I would also suggest that the other changes should be reverted. This component is critical and I'm not sure we have unit test coverage to see that these extra changes are inconsequential. I'd suggest to stick to the original changeset.

Can you kindly confirm if by this you mean to just simply revert to [a17e3a9](https://github.com/ethersphere/bee/pull/5309/commits/a17e3a9ca24510f27a74d8ac11db51d53a91f073)?

acud:
> > I would also suggest that the other changes should be reverted. This component is critical and I'm not sure we have unit test coverage to see that these extra changes are inconsequential. I'd suggest to stick to the original changeset.
> 
> Can you kindly confirm if by this you mean to just simply revert to [a17e3a9](https://github.com/ethersphere/bee/pull/5309/commits/a17e3a9ca24510f27a74d8ac11db51d53a91f073)?

Affirmative :+1:

mfw78:
Done

PR #5312, opened 2026-01-05
fix(node): allow hostnames in public address validation

akrem-chabchoub:
### Checklist

### Description
Allow hostname to pass the validation for public address.

Test:
`./dist/bee start --nat-addr="example.com:1234" --config=config.yml`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,6 +47,7 @@ require (
 	go.uber.org/atomic v1.11.0
 	go.uber.org/goleak v1.3.0
 	golang.org/x/crypto v0.45.0
+	golang.org/x/net v0.47.0
 	golang.org/x/sync v0.18.0
 	golang.org/x/sys v0.38.0
 	golang.org/x/term v0.37.0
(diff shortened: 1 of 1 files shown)


martinconic:
Is there a benefit or it is ok to allow any valid hostname? I think we should only have hostnames that point to the IP where Bee is installed, right? Bee should not start if we provide  `nat-addr: example.com:1634`. The `example.com` is not a valid valid hostname, any other hostname that does not point to the Bee instance from where Bee started should not be allowed, right?

akrem-chabchoub:
> Is there a benefit or it is ok to allow any valid hostname? I think we should only have hostnames that point to the IP where Bee is installed, right? Bee should not start if we provide `nat-addr: example.com:1634`. The `example.com` is not a valid valid hostname, any other hostname that does not point to the Bee instance from where Bee started should not be allowed, right?

The current validation only checks hostname format, not DNS resolution or whether the resolved IP is valid/public. This allows example.com:1634 to pass.
For hostname lookup, its already happening [here](https://github.com/ethersphere/bee/blob/master/pkg/p2p/libp2p/static_resolver.go#L91).
We may keep the current validation as it is to not repeat the same lookup in 2 different places and add validation for public/private IP in [static resolver](https://github.com/ethersphere/bee/blob/master/pkg/p2p/libp2p/static_resolver.go#L91).
What do you think ?

cc @janos

go.mod
```diff
@@ -169,7 +169,7 @@ require (
 	go.uber.org/zap v1.27.0 // indirect
 	golang.org/x/exp v0.0.0-20240213143201-ec583247a57a // indirect
 	golang.org/x/mod v0.29.0 // indirect
-	golang.org/x/net v0.47.0 // indirect
+	golang.org/x/net v0.47.0
```
gacevicljubisa: move this line to direct references

gacevicljubisa (approved):

PR #5314, opened 2026-01-07
fix(p2p): add tcp share when ws enabled

acud:
### Checklist

### Description
libp2p needs a specific option added to the constructor when trying to reuse the existing TCP port listener both for normal TCP connections _and_ for incoming WS connections. Since this is exactly what is done in bee (there is no separate port listener for WS connections), we must include this option when the WS transport is enabled.

### DIFF
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -333,7 +333,7 @@ func New(ctx context.Context, signer beecrypto.Signer, networkID uint64, overlay
 			// of different registers.
 			storagePath := filepath.Join(o.AutoTLSStorageDir, o.AutoTLSDomain)
 
-			if err := os.MkdirAll(storagePath, 0700); err != nil {
+			if err := os.MkdirAll(storagePath, 0o700); err != nil {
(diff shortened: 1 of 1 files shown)


janos:
Hi Elad, we have experimented with ShareTCPListener option and had some connectivity issues with it. As we need WSS, we have left ShareTCPListener for WS to address after WSS is implemented as that is actually needed for in-browser node. Plain text WS is not possible to work from browser, so it is deprioritized. It requires investigations on testnet in order to debug.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -428,6 +428,7 @@ func New(ctx context.Context, signer beecrypto.Signer, networkID uint64, overlay
 		transports = append(transports, libp2p.Transport(ws.New, wsOpt))
 	} else if o.EnableWS {
 		transports = append(transports, libp2p.Transport(ws.New))
+		opts = append(opts, libp2p.ShareTCPListener())
```
martinconic, line 431: We extensively tested this setting a few months ago and ran into several issues. While our recent `libp2p` updates might have resolved those problems, I’d recommend holding off on merging this into `autotls-v2` for now.

PR #5315, opened 2026-01-09
fix: soft-fail on insufficient stake & auto-update height via API

martinconic:
### Checklist

### Description
1.  On startup, if the node has insufficient stake for the configured `ReserveCapacityDoubling`, it now logs a warning and **skips** the on-chain [UpdateHeight](cci:1://file:///Users/cmartinconi/Work/ethswarm/code/bee/pkg/storageincentives/staking/contract.go:141:0-158:1) transaction instead of crashing or reverting. The node starts with the increased local capacity but without redistribution rewards.
2.  When a user deposits stake via the `POST /stake/deposit` endpoint, the handler now automatically triggers the [UpdateHeight](cci:1://file:///Users/cmartinconi/Work/ethswarm/code/bee/pkg/storageincentives/staking/contract.go:141:0-158:1) transaction upon success. This allows a "soft-failed" node to seamlessly upgrade its on-chain status without a restart.

### DIFF
--- a/pkg/api/staking.go
+++ b/pkg/api/staking.go
@@ -83,7 +83,8 @@ func (s *Service) stakingDepositHandler(w http.ResponseWriter, r *http.Request) 
 	// if the node has started with insufficient stake
 	tx, updated, err := s.stakingContract.UpdateHeight(r.Context())
 	if err != nil {
-		logger.Error(err, "update height failed")
+		logger.Debug("update height failed", "error", err)
+		logger.Error(nil, "update height failed")
 	} else if updated {
 		logger.Warning("reserve capacity doubling updated after stake deposit. Node will be FROZEN for ~2 rounds.", "transaction", tx)
 	}


gacevicljubisa:
Instead of polling every 10 minutes, we should trigger UpdateHeight directly from stakingDepositHandler after a successful DepositStake. Staking can only happen via the API `POST /stake/deposit/{amount}`, so we already know exactly when the stake changes — there's no need to poll the chain for something we initiated ourselves.

The only case polling would cover is if someone stakes directly on the smart contract bypassing the bee API, but that's not a supported workflow.

pkg/api/staking.go
```diff
@@ -77,6 +78,21 @@ func (s *Service) stakingDepositHandler(w http.ResponseWriter, r *http.Request)
 		jsonhttp.InternalServerError(w, "cannot stake")
 		return
 	}
+
+	// if the deposit is successful, we should update the height of the node in the staking contract
+	// this is done to make sure that the node is participating in the redistribution game with the correct height
+	// if the node has started with insufficient stake
+	go func() {
```
gacevicljubisa: I think this call should be synchronous.  The `stakingAccessHandler` middleware acquires `stakingSem`, which is released after the return of `stakingDepositHandler`, but goroutine is still runnning.

Also, synchronously, you can return an error,  or at least a warning in the response

pkg/api/staking.go
```diff
@@ -77,6 +77,17 @@ func (s *Service) stakingDepositHandler(w http.ResponseWriter, r *http.Request)
 		jsonhttp.InternalServerError(w, "cannot stake")
 		return
 	}
+
+	// if the deposit is successful, we should update the height of the node in the staking contract
+	// this is done to make sure that the node is participating in the redistribution game with the correct height
+	// if the node has started with insufficient stake
+	tx, updated, err := s.stakingContract.UpdateHeight(r.Context())
+	if err != nil {
+		logger.Error(err, "update height failed")
```
gacevicljubisa: If we do not retrun here an error, I would suggest to follow the existing pattern:

```go
  logger.Debug("update height failed", "error", err)
  logger.Error(nil, "update height failed")
```

PR #5326, opened 2026-01-22
feat(p2p): support inbound-only peers without underlay addresses

gacevicljubisa:
### Checklist

### Description
Allow peers without dialable underlay addresses (e.g., browsers, WebRTC
clients, NAT-restricted nodes) to establish connections and use protocols
over existing streams.

Changes:
- Remove empty underlay validation in bzz.ParseAddress and MarshalJSON
- Skip addressbook persistence for peers with empty underlays
- Add nil check in notifyReacherConnected to prevent reacher crashes
- Rename peerRegistry.underlays to overlayToPeerID for clarity

Inbound-only peers can participate in protocols (pricing, retrieval, pss)
but are excluded from Kademlia topology and hive gossip since they cannot
be dialed back.

### DIFF
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -50,7 +50,7 @@ var (
 )
 
 type Service struct {
-	streamer          p2p.Streamer
+	streamer          p2p.Bee260CompatibilityStreamer
 	addressBook       addressbook.GetPutter
 	addPeersHandler   func(...swarm.Address)
 	networkID         uint64
@@ -67,7 +67,7 @@ type Service struct {
 	overlay           swarm.Address
 }
 
-func New(streamer p2p.Streamer, addressbook addressbook.GetPutter, networkID uint64, bootnode bool, allowPrivateCIDRs bool, overlay swarm.Address, logger log.Logger) *Service {
+func New(streamer p2p.Bee260CompatibilityStreamer, addressbook addressbook.GetPutter, networkID uint64, bootnode bool, allowPrivateCIDRs bool, overlay swarm.Address, logger log.Logger) *Service {
 	svc := &Service{
 		streamer:          streamer,
 		logger:            logger.WithName(loggerName).Register(),
@@ -195,6 +195,8 @@ func (s *Service) sendPeers(ctx context.Context, peer swarm.Address, peers []swa
 			s.logger.Debug("skipping peers: no advertisable underlays", "peer_address", p)
 			continue
 		}
+
+		advertisableUnderlays = p2p.FilterBee260CompatibleUnderlays(s.streamer.IsBee260(peer), advertisableUnderlays)
 
 		peersRequest.Peers = append(peersRequest.Peers, &pb.BzzAddress{
 			Overlay:   addr.Overlay.Bytes(),
--- a/pkg/p2p/libp2p/internal/handshake/handshake.go
+++ b/pkg/p2p/libp2p/internal/handshake/handshake.go
@@ -153,7 +153,7 @@ func (s *Service) Handshake(ctx context.Context, stream p2p.Stream, peerMultiadd
 
 	w, r := protobuf.NewWriterAndReader(stream)
 
-	peerMultiaddrs = filterBee260CompatibleUnderlays(o.bee260compatibility, peerMultiaddrs)
+	peerMultiaddrs = p2p.FilterBee260CompatibleUnderlays(o.bee260compatibility, peerMultiaddrs)
 
 	if err := w.WriteMsgWithContext(ctx, &pb.Syn{
 		ObservedUnderlay: bzz.SerializeUnderlays(peerMultiaddrs),
@@ -208,7 +208,7 @@ func (s *Service) Handshake(ctx context.Context, stream p2p.Stream, peerMultiadd
 		return a.Equal(b)
 	})
 
-	advertisableUnderlays = filterBee260CompatibleUnderlays(o.bee260compatibility, advertisableUnderlays)
+	advertisableUnderlays = p2p.FilterBee260CompatibleUnderlays(o.bee260compatibility, advertisableUnderlays)
 
 	bzzAddress, err := bzz.NewAddress(s.signer, advertisableUnderlays, s.overlay, s.networkID, s.nonce)
 	if err != nil {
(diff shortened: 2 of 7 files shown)


janos:
Also, do not add peer without underlays to the addressbook in hive here https://github.com/ethersphere/bee/blob/master/pkg/hive/hive.go#L314.

gacevicljubisa:
> Also, do not add peer without underlays to the addressbook in hive here https://github.com/ethersphere/bee/blob/master/pkg/hive/hive.go#L314.

They are skipped here
https://github.com/ethersphere/bee/blob/706fe4c692c10f415583846429be6f3c5bb88ab0/pkg/hive/hive.go#L288-L291

pkg/p2p/libp2p/libp2p_test.go
```diff
@@ -82,6 +85,19 @@ func newService(t *testing.T, networkID uint64, o libp2pServiceOpts) (s *libp2p.
 	opts := o.libp2pOpts
 	opts.Nonce = nonce
 
+	// If emptyUnderlays is set, use a custom host factory that creates a host with no listen addresses
+	// This simulates an inbound-only peer (browser, WebRTC connection, strict NAT)
+	if o.emptyUnderlays {
+		opts = libp2p.WithHostFactory(
```
akrem-chabchoub: This overwrites `opts`, losing any previously set fields like `FullNode` from `o.libp2pOpts`.
I think we should append to previous `opts` ?

pkg/p2p/libp2p/connections_test.go
```diff
@@ -1477,6 +1492,135 @@ type (
...
+
+		_, err := s2.Connect(ctx, s1Addrs)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		expectPeers(t, s2, overlay1)
+		expectPeersEventually(t, s1, overlay2)
+
+		// Verify s1 (has underlays) is persisted in s2's addressbook
```
akrem-chabchoub, line 1615: I think we should add this test case: 
s2 should not be saved in s1's addressbook

pkg/p2p/libp2p/libp2p_test.go
janos: Yes, this makes emptyUnderlays option coupled with libp2pOpts in a not transparent way.

It is a problem with WithHostFactory function that, just returns the new libp2p.Options instead just to set the unexported field internally. This is just not practical, or composable.

I suggest to add in export_test.go:

```go
func SetHostFactory(o *Options, factory func(...libp2pm.Option) (host.Host, error)) {
	o.hostFactory = factory
}
```

as more flexible alternative. WithHostFactory is not even named clearly, its name does not suggest that it is (one of) a constructor of the Options type.

lat-murmeldjur:
I'm not sure this is necessary. From browser connecting to testnet wss addresses is already possible. The browser node also has an (ephemeral) underlay.

v1rtl:
Without this Bee in browser fails to maintain connections. It is really necessary. I dont see any drawback to not support inbound-only peers

gacevicljubisa:
> I'm not sure this is necessary. From browser connecting to testnet wss addresses is already possible. The browser node also has an (ephemeral) underlay.

@lat-murmeldjur Can you provide more details about ephemeral underlay?

lat-murmeldjur:
> > I'm not sure this is necessary. From browser connecting to testnet wss addresses is already possible. The browser node also has an (ephemeral) underlay.
> 
> @lat-murmeldjur Can you provide more details about ephemeral underlay?

Yes.
If you go to the weeb-3 client here: [https://lat-murmeldjur.github.io/weeb-3/](https://lat-murmeldjur.github.io/weeb-3/)
In the logs at the bottom of the page you can see lines like 
Observed address /ip4/37.76.0.186/tcp/28943/wss for peer QmfSx1ujzboapD5h2CiqTJqUy46FeTDwXBszB3XUCfKEEj stored/overwritten

This is our ephemeral underlay in the browser, observed by the bee node, and sent back to the browser in the libp2p-identify protocol.

The reason why getting it through this libp2p-protocol was necessary, is because connecting to a tls/ws address broke 1 thing:
In the handshake, we no longer get back our observed underlay field filled out by the bee node. I suspect that could have been the reason for [v1rtl](https://github.com/v1rtl) requiring this PR as well. So without observed underlay being signed after handshake i got disconnected immediately. Now that i read it from the libp2p-identify, wait for it (after dial, before handshake) and then use it for the  handshake it completes and I can upload and download after that.

gacevicljubisa:
> Without this Bee in browser fails to maintain connections. It is really necessary. I dont see any drawback to not support inbound-only peers

As I understand, with this PR, the browser client doesn't need the workaround of reading the observed underlay from libp2p-identify before the handshake. The handshake now accepts empty underlays, so the browser can complete the handshake immediately without waiting for the observed address.

lat-murmeldjur:
> > Without this Bee in browser fails to maintain connections. It is really necessary. I dont see any drawback to not support inbound-only peers
> 
> As I understand, with this PR, the browser client doesn't need the workaround of reading the observed underlay from libp2p-identify before the handshake. The handshake now accepts empty underlays, so the browser can complete the handshake immediately without waiting for the observed address.

It's okay I'm not objecting to this change just wanted to make it clear that strictly speaking you don't need this to make connections work.

gacevicljubisa:
With introducing the [RemoteMultiaddr](https://github.com/ethersphere/bee/blob/6ddf9b4585f5bce700a5a31c581a044ad1ebeb3b/pkg/p2p/libp2p/libp2p.go#L649) this PR is no longer required.

mfw78:
@gacevicljubisa I'll comment here, so that we don't pollute the other PR.

The problem IMO that needs to be resolved is that a solution should _not_ require some multiaddr. Sure, it's nice to send an observed multiaddr from the connectee back to the connector, however for NAT-bound peers connecting out, it has no relevance (it does let them know that they are connected to a globally routable network, albeit via NAT, but the success of a connection is enough in and of itself to satisfy this). Subsequently, the address _was_ being added to the peer store, which is suitable for the connectee, but pointless from a `hive` perspective.

It seems odd to be signing a multiaddr, simply just trusting it, in order to satisfy the handshake, and actual fact, why can't the ack have a bzz address with no signed multiaddr in it (ie. the signature is over an empty Vec<Multiaddr>). Examples of where this is particularly painful is that with the introduction of the peer store waiting, this essentially causes all light clients behind NAT to get blocked for 10s on connection.

bcsorvasi:
Hi @mfw78,

Thanks for taking the time to lay this out and explain your concerns.

To make sure this gets the proper visibility and structured discussion it deserves, could you please open a dedicated issue in the repository outlining:

- The core problem as you see it
- The specific pain points (e.g. NAT-bound peers, light client blocking, etc.)
- The proposed alternative approach

We’d like our research team to review this in more depth and evaluate whether this requirement should exist in its current form, as well as the protocol-level implications of changing it.

Opening an issue will help us track the discussion properly without losing context in the PR thread.

Appreciate the thoughtful feedback!

PR #5329, opened 2026-01-23
fix: make hive bee 2.6.0 backward compatible

janos:
### Checklist

### Description
Send only one undelray address in hive to bee 2.6.0 or earlier nodes.

This is a backward compatibility support similar to the one already added to the handshake protocol.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -35,7 +35,7 @@ jobs:
     - name: Increase UDP buffer sizes (macOS)
       if: matrix.os == 'macos-latest'
       run: |
-        sudo sysctl -w kern.ipc.maxsockbuf=7500000
+        sudo sysctl -w kern.ipc.maxsockbuf=6291456
     - name: Build
       run: make build
     - name: Test with race detector (Ubuntu and MacOS)


pkg/p2p/p2p.go
```diff
@@ -237,3 +243,18 @@ func (e *ChunkDeliveryError) Error() string {
...
+
+// FilterBee260CompatibleUnderlays select a single underlay to pass if
+// bee260compatibility is true. Otherwise it passes the unmodified underlays
+// slice. This function can be safely removed when bee version 2.6.0 is
+// deprecated.
+func FilterBee260CompatibleUnderlays(bee260compatibility bool, underlays []ma.Multiaddr) []ma.Multiaddr {
+	if !bee260compatibility {
+		return underlays
+	}
+	underlay := bzz.SelectBestAdvertisedAddress(underlays, nil)
```
gacevicljubisa, line 255: Here we could only return TCP. It is prefered, but 2.6 doesn't support other transports, so we can send TCP only.

acud: makes sense

PR #5330, opened 2026-01-24
refactor(libp2p): use precise connection cleanup in handshake error handling

gacevicljubisa:
### Checklist

### Description

**Problem**

In the handshake error paths, we called `s.host.Network().ClosePeer(peerID)`, which terminates **all** libp2p connections to that peer, not just the one that failed. With NAT'd peers and simultaneous dials, duplicate connection attempts are common: a failed handshake on a *new, redundant* connection would tear down an existing healthy session, notify kademlia, depeer the neighbor, and force a full reconnect cycle. Worst case was the duplicate-peer path in `Connect()`: a mere `FullClose()` failure on a redundant handshake stream called `s.Disconnect(overlay, ...)`, actively removing a healthy peer from the topology.

**Fix**

Close only the connection that carried the failed handshake (`stream.Conn().Close()`) and leave established sessions untouched:

- Use `stream.Conn().Close()` instead of `ClosePeer()` in the handshake error paths of `handleIncoming` and `Connect`
- In the duplicate-peer path of `Connect`, close only the new redundant connection instead of disconnecting the peer
- Also apply the same cleanup to the `buildFullMAs` fallback error path in `Connect` (was missed in the first pass)

**Why this is safe**

- The peer registry is already designed for multiple connections per peer: `addIfNotExists` tracks every connection, and the `Disconnected` notifiee only removes the peer when its **last** connection drops. Closing a single connection is handled correctly end to end.
- The inbound duplicate-peer path (`handleIncoming`) already used `stream.Conn().Close()` on master, this PR makes the outbound path and remaining error paths consistent with it.
- Blocklisted peers intentionally keep `ClosePeer()`: for those we do want to sever everything.

**Expected impact**

Lower peer churn and a more stable topology: fewer spurious `bee_kademlia_total_inbound_disconnections`, fewer reconnect attempts, longer peer session lifetimes. Being an error-path change, the effect is most visible in NAT-heavy environments; we'll verify with a canary comparing disconnection rate and session lifetime split by bee version.

N/A

Prevents disrupting healthy concurrent connections when rejecting redundant or failed connection attempts, reducing unnecessary kademlia topology churn.


janos (commented):
👍

janos (approved):

martinconic (approved):

akrem-chabchoub (approved):

janos (approved):

PR #5331, opened 2026-01-24
fix(puller): prevent massive log lines from joined errors

gacevicljubisa:
### Checklist

### Description
Previously, when many errors were joined together during sync
operations, all errors were logged in a single line, creating
log entries up to 30KB (e.g., 245 identical "batch not found" 
errors).

This commit adds countErrors() to count total errors and logs
only the count plus one example error, keeping entries concise
while maintaining diagnostic value.

Changes:
- Add countErrors() helper to count errors in joined error chains
- Update syncWorker logging to show error_count and example_error
- Add test coverage for error counting

Example output:
- Before: "error"="batch not found\nbatch not found\n..." (30KB)
- After: "error_count"=245 "example_error"="batch not found"

### DIFF
--- a/pkg/puller/error_summary_test.go
+++ b/pkg/puller/error_summary_test.go
@@ -1,4 +1,4 @@
-// Copyright 2025 The Swarm Authors. All rights reserved.
+// Copyright 2026 The Swarm Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
--- a/pkg/puller/puller.go
+++ b/pkg/puller/puller.go
@@ -382,11 +382,7 @@ func (p *Puller) syncPeerBin(parentCtx context.Context, peer *syncPeer, bin uint
 					return
 				}
 				errCount := countErrors(err)
-				if errCount > 1 {
-					p.logger.Debug("syncWorker interval failed", "error_count", errCount, "example_error", errors.Unwrap(err), "peer_address", address, "bin", bin, "cursor", cursor, "start", start, "topmost", top)
-				} else {
-					p.logger.Debug("syncWorker interval failed", "error", err, "peer_address", address, "bin", bin, "cursor", cursor, "start", start, "topmost", top)
-				}
+				p.logger.Debug("syncWorker interval failed", "error_count", errCount, "example_error", errors.Unwrap(err), "peer_address", address, "bin", bin, "cursor", cursor, "start", start, "topmost", top)
 			}
 
 			_ = p.limiter.WaitN(ctx, count)


pkg/puller/error_summary_test.go
```diff
@@ -0,0 +1,72 @@
+// Copyright 2025 The Swarm Authors. All rights reserved.
```
janos: 2025 -> 2026

gacevicljubisa:
> The approach is maybe too high level. It looks to me that errors.Join is used in a way that is not ideal, maybe even misused. This is solving the problem with long log lines, but does not solve the root cause of using errors as lists of errors.
> 
> I would say that this is ok solution but ideally it would be better in the future to address the usage of errors.Join.

I agree.

pkg/puller/puller.go
```diff
@@ -348,7 +381,12 @@ func (p *Puller) syncPeerBin(parentCtx context.Context, peer *syncPeer, bin uint
 					p.logger.Debug("syncWorker interval failed, quitting", "error", err, "peer_address", address, "bin", bin, "cursor", cursor, "start", start, "topmost", top)
 					return
 				}
-				p.logger.Debug("syncWorker interval failed", "error", err, "peer_address", address, "bin", bin, "cursor", cursor, "start", start, "topmost", top)
+				errCount := countErrors(err)
+				if errCount > 1 {
+					p.logger.Debug("syncWorker interval failed", "error_count", errCount, "example_error", errors.Unwrap(err), "peer_address", address, "bin", bin, "cursor", cursor, "start", start, "topmost", top)
+				} else {
```
akrem-chabchoub: `errors.Unwrap()` looks for `Unwrap() error`, but `errors.Join()` implements `Unwrap() []error`. 
Would this return nil here?

acud: I'm not sure I understand the question?

By the way @gacevicljubisa, maybe it is just better to remove the `if` statement and log both ways with the `error_count` field? Not sure if such granularity is needed

akrem-chabchoub: So I didn't find a way to test and see errors locally in puller logs, so I created this unit test in [error_summary_test.go](https://github.com/ethersphere/bee/pull/5331/changes#diff-9bd17db7ac8cafdfce6699cacbc104def53c9fe7063c3ba847e739a921a6ced8) to validate it but i got the issue that i meantioned.
```
t.Run("errors.Unwrap returns nil for joined errors", func(t *testing.T) {
		err1 := errors.New("storage: not found")
		err2 := errors.New("storage: not found")
		joined := errors.Join(err1, err2)
		
		unwrapped := errors.Unwrap(joined)
		if unwrapped == nil {
			t.Errorf("expected not nil, got %v", unwrapped)
		}
	})
```

akrem-chabchoub: `errors.Join()` creates an error with `Unwrap() []error` but `errors.Unwrap()` looks for `Unwrap() error`.
Since the signatures don't match, `errors.Unwrap()` returns nil.

PR #5332, opened 2026-01-25
fix(postage): prevent nil pointer panic in TopUpBatch on transaction …

akrem-chabchoub:
### Checklist

### Description
Before when running this command you will get panic error, now with the fix:
`curl -s -X PATCH "http://localhost:1640/stamps/topup/{BATCH_ID}/2" \
  -H "Gas-Price: 10"`
  ```
  {"code":500,"message":"cannot topup batch"}
  ```
Config:
```
minimum-gas-tip-cap: 2
full-node: false
swap-enable: false
```
Logs:
```
"time"="2026-01-29 13:39:34.554380" "level"="debug" "logger"="node/transaction" "msg"="estimate gas failed" "sender_address"="0xb25D7c21Fd221719826028D371bEabf74eB061EE" "error"="Transaction execution fails"
"time"="2026-01-29 13:39:34.923401" "level"="debug" "logger"="node/api/patch_stamp_topup" "msg"="topup batch: topup failed" "batch_id"="fe5db63cc90c61509744f952c6fae608c7488ce194227d8d31ca9d7fe502b371" "amount"="10" "error"="topup batch: amount 10: Invalid, intrinsic gas too low"
"time"="2026-01-29 13:39:34.923453" "level"="error" "logger"="node/api/patch_stamp_topup" "msg"="topup batch: topup failed"
```

### DIFF
--- a/pkg/postage/postagecontract/contract.go
+++ b/pkg/postage/postagecontract/contract.go
@@ -390,7 +390,7 @@ func (c *postageContract) CreateBatch(ctx context.Context, initialBalance *big.I
 	if err != nil {
 		return
 	}
-	txHash = receipt.TxHash
+
 	for _, ev := range receipt.Logs {
 		if ev.Address == c.postageStampContractAddress && len(ev.Topics) > 0 && ev.Topics[0] == c.batchCreatedTopic {
 			var createdEvent batchCreatedEvent
@@ -415,6 +415,7 @@ func (c *postageContract) CreateBatch(ctx context.Context, initialBalance *big.I
 			if err != nil {
 				return
 			}
+			txHash = receipt.TxHash
 			return
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/postage/postagecontract/contract.go
```diff
@@ -447,10 +447,11 @@ func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topupB
 
 	receipt, err := c.sendTopUpBatchTransaction(ctx, batch.ID, topupBalance)
 	if err != nil {
-		txHash = receipt.TxHash
 		return
 	}
 
+	txHash = receipt.TxHash
```
martinconic: I think you can remove this completely since this assignment is done only in bellow conditions.

pkg/postage/postagecontract/contract.go
```diff
@@ -447,7 +447,6 @@ func (c *postageContract) TopUpBatch(ctx context.Context, batchID []byte, topupB
 
 	receipt, err := c.sendTopUpBatchTransaction(ctx, batch.ID, topupBalance)
 	if err != nil {
-		txHash = receipt.TxHash
```
gacevicljubisa, line 450: Yes, this is obvious panic, if error happens, receipt is nil

gacevicljubisa:
Also, we should return txHash only in case that there is no error:
```
txHash = receipt.TxHash
```
so we should adapt for CreateBatch and DiluteBatch

Also, it would be good if this is covered in unit tests as well.

PR #5336, opened 2026-01-29
fix(p2p): use connection remote addr as fallback when peerstore is empty

gacevicljubisa:
### Checklist

### Description
Fixes a backwards-compatibility bug affecting local nodes running v2.6.0 behind NAT without port forwarding.

### DIFF
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -1466,17 +1466,17 @@ func (s *Service) determineCurrentNetworkStatus(err error) error {
 }
 
 // peerMultiaddrs builds full multiaddresses for a peer given information from
-// libp2p host peerstore and falling back to the remote address from the
-// connection.
+// the libp2p host peerstore. If the peerstore doesn't have addresses yet,
+// it falls back to using the remote address from the active connection.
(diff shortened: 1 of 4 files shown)


pkg/p2p/libp2p/libp2p.go
```diff
@@ -646,7 +646,7 @@ func (s *Service) handleIncoming(stream network.Stream) {
 	peerID := stream.Conn().RemotePeer()
 	handshakeStream := newStream(stream, s.metrics)
 
-	peerMultiaddrs, err := s.peerMultiaddrs(s.ctx, peerID)
+	peerMultiaddrs, err := s.peerMultiaddrs(s.ctx, stream.Conn().RemoteMultiaddr(), peerID)
```
akrem-chabchoub, line 649: Is it always guaranteed that the `stream.Conn()` is not `nil` here ?

janos (approved):
👍

pkg/p2p/libp2p/libp2p.go
gacevicljubisa: stream is always associated with a valid network.Conn, so this shouldn't happen

akrem-chabchoub (approved):

PR #5341, opened 2026-01-30
fix(chain): add rpc retry logic

gacevicljubisa:
### Checklist

### Description
The bee node was experiencing intermittent connection failures when communicating with blockchain RPC endpoints (especially HAProxy-backed services), resulting in errors like:

```
"could not get block number" "error"="Post \"http://rpc-sepolia-haproxy...\": EOF"
```

These EOF errors occur due to short timeout configurations on HAProxy which can prematurely close connections during RPC communication. Without retry logic, these transient network errors (EOF, connection resets) would cause operations to fail immediately, impacting node stability and reliability.

Implemented a custom retryRoundTripper that wraps the HTTP transport with retry logic

Key Features:

- Automatic retries: Up to 3 retry attempts for transient errors
- Exponential backoff: 100ms → 200ms → 400ms between retries (capped at 5s)
- Error detection: Retries on EOF, connection reset, connection refused, and broken pipe errors
- Request body handling: Rewinds request bodies between retries using GetBody() to avoid content-length mismatches
- Uses http.DefaultTransport wrapped with retry logic

### DIFF
--- a/pkg/node/chain.go
+++ b/pkg/node/chain.go
@@ -60,6 +60,15 @@ func (r *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error)
 		if attempt > 0 {
 			backoffDuration := min(r.backoff*time.Duration(1<<uint(attempt-1)), 5*time.Second)
 			time.Sleep(backoffDuration)
+
+			// Reset request body for retry if possible
+			if req.Body != nil && req.GetBody != nil {
+				var err error
+				req.Body, err = req.GetBody()
+				if err != nil {
+					return nil, fmt.Errorf("failed to reset request body for retry: %w", err)
+				}
+			}
 		}
 
 		resp, err := r.transport.RoundTrip(req)
@@ -71,6 +80,10 @@ func (r *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error)
 
 		if !isRetryableError(err) {
 			return nil, err
+		}
+
+		if req.Body != nil && req.GetBody == nil {
+			return nil, fmt.Errorf("cannot retry request with non-rewindable body: %w", err)
 		}
 	}
 


pkg/node/chain.go
```diff
@@ -37,8 +40,92 @@ const (
...
+
+// RoundTrip executes a single HTTP transaction, returning a Response for the Request.
+// It retries on transient errors like EOF, connection reset, and connection refused.
+func (r *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	var lastErr error
+
+	for attempt := 0; attempt <= r.maxRetries; attempt++ {
+		if attempt > 0 {
+			backoffDuration := min(r.backoff*time.Duration(1<<uint(attempt-1)), 5*time.Second)
+			time.Sleep(backoffDuration)
```
janos, line 62: Instead of sleep, use time.After and req.Context in select to return as soon as possible if context is done.

pkg/node/chain.go
```diff
@@ -37,8 +40,92 @@ const (
...
+
+			// Reset request body for retry if possible
+			if req.Body != nil && req.GetBody != nil {
+				var err error
+				req.Body, err = req.GetBody()
```
janos, line 67: This may be problematic if the Body is already read or closed. There is a check below for non-rewritable body, so it should be ok, I suppose.

In general, only idempotent requests should be retried, but in case of rpc, identifying idempotent requests are not easy. Getting information is ok, but posting information may present a problem by duplicating them.

gacevicljubisa: Maybe it is the simpliest to increase the values for the timeouts on the RPC side? 
These align with HAProxy timeouts to prevent errors, but they look low?

```go
func newHTTPClient() *http.Client {
    return &http.Client{
        Transport: &http.Transport{
            DialContext: (&net.Dialer{
                Timeout:   9 * time.Second, // < HAProxy timeout connect (10s)
                KeepAlive: 5 * time.Second, // Keep connections alive but close before idle timeout
            }).DialContext,
            IdleConnTimeout:       8 * time.Second, // < HAProxy timeout client (10s)
            TLSHandshakeTimeout:   9 * time.Second, // < HAProxy timeout connect (10s)
            ResponseHeaderTimeout: 9 * time.Second, // < HAProxy timeout server (10s)
        },
    }
}
```

janos: I agree, even reties would make the rpc endpoint do more work in this case. Increasing the timeout would be better in this case.

pkg/node/chain.go
```diff
@@ -37,8 +40,92 @@ const (
...
+		if req.Body != nil && req.GetBody == nil {
+			return nil, fmt.Errorf("cannot retry request with non-rewindable body: %w", err)
+		}
+	}
+
+	return nil, fmt.Errorf("request failed after %d retries: %w", r.maxRetries, lastErr)
+}
+
+// isRetryableError determines if an error should trigger a retry.
+func isRetryableError(err error) bool {
```
akrem-chabchoub, line 94: Are these checks for error codes compatible with windows ?
AFAIK that windows have different error codes, maybe some unit tests can reveal it ?

pkg/node/chain.go
```diff
@@ -37,8 +40,92 @@ const (
...
+		if !isRetryableError(err) {
+			return nil, err
+		}
+
+		if req.Body != nil && req.GetBody == nil {
+			return nil, fmt.Errorf("cannot retry request with non-rewindable body: %w", err)
+		}
+	}
+
+	return nil, fmt.Errorf("request failed after %d retries: %w", r.maxRetries, lastErr)
```
akrem-chabchoub, line 90: The loop above runs for attempt in 0 to maxRetries, for example with maxRetries = 3, there are 4 attempts, but the error says 3 retries.

akrem-chabchoub:
Take a look on this package, it may help simplifying the implementation:
https://github.com/hashicorp/go-retryablehttp

gacevicljubisa:
Closing this PR in favor of [#5404](https://github.com/ethersphere/bee/issues/5404). Based on the review feedback and here, I have created new issue that proposes adding configurable HTTP client settings for the RPC connection as part of the bee configuration.

Thanks everyone for the review and discussion

PR #5344, opened 2026-02-03
fix(transaction): improve gas estimation reliability and error handling

gacevicljubisa:
### Checklist

### Description
This PR improves gas estimation reliability and introduces a configurable
fallback gas limit for contract transactions.

**Gas estimation fix:**
Fixes "intrinsic gas too low" errors when `MinEstimatedGasLimit` was 0 and
gas estimation failed. The fallback ensures transactions can proceed even
when RPC balance checks fail, since actual transaction cost is typically
much lower than the maximum theoretical cost used in estimation.

**Configurable `--gas-limit-fallback` flag:**
Adds a new `--gas-limit-fallback` CLI flag (default: 500,000) to make the
fallback gas limit configurable.

**Contract gas limit API refactor:**
Refactors contract constructors (`postagecontract`, `staking`,
`redistribution`) to accept an explicit `gasLimit uint64` parameter instead
of a `setGasLimit bool`. This removes the hidden coupling to
`transaction.DefaultGasLimit` and gives callers direct control over the gas
limit value. The node computes `contractGasLimit` once based on
`TrxDebugMode` and passes it to all contract constructors.

The existing `Gas-Limit` HTTP header already provides per-request override
for API-triggered transactions via `sctx.GetGasLimit(ctx)`.


cmd/bee/cmd/cmd.go
```diff
@@ -83,6 +83,7 @@ const (
 	optionReserveCapacityDoubling          = "reserve-capacity-doubling"
 	optionSkipPostageSnapshot              = "skip-postage-snapshot"
 	optionNameMinimumGasTipCap             = "minimum-gas-tip-cap"
+	optionNameGasLimitFallback             = "gas-limit-fallback"
```
akrem-chabchoub, line 85: This flag should be added to `docker-compose` and `env` files

PR #5345, opened 2026-02-03
fix: fixes for topology oversaturated and block list streams tests

martinconic:
### Checklist

### Description
This PR addresses two distinct race conditions in the libp2p package that were causing intermittent failures in 
TestTopologyOverSaturated and TestWithBlocklistStreams


pkg/p2p/libp2p/peer.go
gacevicljubisa, line 1: No changes are needed here in this file

gacevicljubisa:
Now `b.blocklist.Exists(peer.Address)` is called twice, in `handleIncoming` and in handshake `Handle`.. Seems that now that call in `handleIncoming` is redundant?

martinconic:
> Now `b.blocklist.Exists(peer.Address)` is called twice, in `handleIncoming` and in handshake `Handle`.. Seems that now that call in `handleIncoming` is redundant?

Very good catch

acud (approved):

PR #5347, opened 2026-02-04
fix(p2p): improve reacher address selection and transport filtering

gacevicljubisa:
### Checklist

### Description
This PR improves the reliability of peer reachability testing by using multiaddresses directly from handshake, and adds proper transport filtering and prioritization.

### DIFF
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -1057,7 +1057,11 @@ func (s *Service) Connect(ctx context.Context, addrs []ma.Multiaddr) (address *b
 			return address, p2p.ErrAlreadyConnected
 		}
 
-		if err := s.connectionBreaker.Execute(func() error { return s.host.Connect(ctx, *info) }); err != nil {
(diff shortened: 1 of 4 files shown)


janos (approved):

pkg/p2p/libp2p/libp2p.go
```diff
@@ -124,6 +124,9 @@ type Service struct {
 	enableWS           bool
 	autoTLSCertManager autoTLSCertManager
 	zapLogger          *zap.Logger
+	hasTCPTransport    bool
+	hasWSTransport     bool
+	hasWSSTransport    bool
```
janos: Would it be nicer to have a `enabledTransports map[bzz.Transport]bool`, just for easier grouping of flags? Maybe isTransportSupported would be even simpler in that case.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -124,6 +124,7 @@ type Service struct {
 	enableWS           bool
```
akrem-chabchoub: This is not called anywhere, should be deleted ?

PR #5348, opened 2026-02-05
fix(p2p): update peer address on reconnect and optimize reacher

gacevicljubisa:
### Checklist

### Description
This PR improves the reacher component in two ways:

**1. Optimize peer scheduling with a min-heap**

Replace the `map[string]*peer` with a min-heap ordered by `retryAfter` time,
backed by an index map for O(1) lookups. The previous implementation iterated
over all peers on every scheduling cycle to find the one with the earliest
retry time. The heap reduces this to O(1) peek / O(log n) updates.

**2. Update peer underlay address on reconnect**

When a peer reconnects (same overlay, different underlay), the reacher now
updates the stored multiaddr and resets the retry timer to trigger an
immediate re-ping. Previously, duplicate `Connected` calls for the same
overlay were silently ignored, causing the reacher to keep pinging the
stale address. This is particularly important for peers that support
multiple underlays or change IP addresses between connections.

After the immediate re-ping, the next scheduled ping follows the normal
`RetryAfterDuration` cadence.

### DIFF
--- a/pkg/p2p/libp2p/internal/reacher/heap.go
+++ b/pkg/p2p/libp2p/internal/reacher/heap.go
@@ -2,8 +2,6 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// Package reacher runs a background worker that will ping peers
-// from an internal queue and report back the reachability to the notifier.
 package reacher
 
 // peerHeap is a min-heap of peers ordered by retryAfter time.


akrem-chabchoub (approved):

pkg/p2p/libp2p/internal/reacher/reacher.go
```diff
@@ -27,11 +28,41 @@ type peer struct {
 	overlay    swarm.Address
 	addr       ma.Multiaddr
 	retryAfter time.Time
+	index      int // index in the heap
+}
+
+// peerHeap is a min-heap of peers ordered by retryAfter time.
+type peerHeap []*peer
```
martinconic: I would move the heap into a separate file

pkg/p2p/libp2p/internal/reacher/reacher.go
```diff
@@ -80,7 +112,7 @@ func (r *reacher) manage() {
 
 	defer r.wg.Done()
 
-	c := make(chan *peer)
+	c := make(chan peer)
```
martinconic, line 87: Here and also bellow, why not using a pointer to a peer anymore?

gacevicljubisa: By sending a value copy through the channel, each ping goroutine gets its own independent snapshot of the peer data, which is safe to read concurrently

PR #5356, opened 2026-02-10
feat: tunning leveldb configs

martinconic:
### Checklist

### Description
This PR tunes the embedded goleveldb configuration to better handle the heavy write throughput and concurrent read patterns observed in production nodes. The default settings were causing significant write stalls and suboptimal read performance under load.

### Changes:

**Increase Block Cache (32MB -> 256MB):**
* Significantly improves read performance for the metadata index (TagItems, PushItems).
* Reduces disk I/O by keeping a much larger portion of the frequently accessed index "hot" in RAM.
* Rationale: 32MB was too small for the millions of items in the index, causing constant cache thrashing.

**Increase Write Buffer (32MB -> 128MB):**
* Allows larger write batches to accumulate in memory before flushing to disk.
* Reduces the frequency of Level 0 file creation, which directly reduces write stalls during bursty upload traffic.

**Increase Compaction Trigger (8 -> 16 files):**
* Allows more Level 0 files to accumulate before triggering a compaction.
* Prevents the database from "locking up" writes to force a compaction during heavy load spikes.

**Increase Table Size (2MB -> 8MB):**
* Reduces the total number of .ldb files the system has to manage.
* Lowers file system overhead and helps avoid "too many open files" errors.

**Optimize Bloom Filter (64 bits -> 10 bits):**
* Reduces the memory overhead of bloom filters per key. 10 bits is the industry standard for ~1% false positive rate.
* Rationale: 64 bits was excessive and wasted RAM that is better used for the Block Cache.

**Disable Seeks Compaction:**
* Prevents read operations from triggering background compactions.
* Rationale: In our read-heavy metadata workload, this feature often caused unnecessary write amplification, fighting against actual data writes for I/O bandwidth.

### Hardware Impact: 
These changes increase the memory footprint of the storer component by approximately 400-500MB.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -152,7 +152,7 @@ require (
 	github.com/pion/webrtc/v4 v4.1.2 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
-	github.com/prometheus/client_model v0.6.2 // indirect
+	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/common v0.64.0
 	github.com/prometheus/procfs v0.16.1 // indirect
 	github.com/prometheus/statsd_exporter v0.26.1 // indirect


martinconic:
Closing this for the moment since the possible benefits are not so rewarding and they came with a draw back in RAM memory increase

PR #5361, opened 2026-02-12
feat(upload): auto-detect file content-type if not sent

### LINKED ISSUE #5346: Bzz upload: auto-detect content type instead of trusting user header
### Context
- Bee version: 2.6.0
- Component: `/bzz` upload endpoint (`pkg/api/bzz.go`)

### Summary

The `/bzz` endpoint requires users to provide a `Content-Type` header when uploading files, but there's no validation that this header matches the actual file content. This leads to incorrect content types being stored in the manifest metadata.

### Expected behavior

When a user uploads `README.md` with `Content-Type: image/png`, the system should either:
1. Reject the request with a validation error, or
2. Detect the actual content type and use that instead

### Actual behavior

The incorrect `Content-Type` header is blindly trusted and stored in the manifest. When the file is downloaded later, the browser receives the wrong MIME type and fails to render the content correctly:
- A text file marked as `image/png` displays as broken image
- Users see corrupted/unreadable content in their browser

### Steps to reproduce

1. Upload a markdown file with wrong content type:
```bash
curl -XPOST \
  -H "Swarm-Postage-Batch-Id: <batch-id>" \
  -H "Content-Type: image/png" \
  --data-binary @README.md \
  localhost:1633/bzz
```

2. See the file in the browser
```bash
http://localhost:1633/bzz/<reference>
```
3. Browser tries to render markdown as PNG and fails

### Possible solution

Remove the `Content-Type` header requirement and auto-detect the content type from the file bytes using Go's built-in `http.DetectContentType()`.

Benefits:
- Users don't need to specify the correct MIME type
- Eliminates this class of bugs entirely
- Files are always served with correct content types
- Simpler API (one less required header)

### Note

Check other endpoints that accept content type from users for the same issue.

akrem-chabchoub:
### Checklist

### Description
For single-file POST `/bzz`, Content-Type is optional: if the client sends it, we keep it exactly, if not we infer it from the first bytes of the body.

### Related Issue (Optional)
#5346


acud:
what is the status of this PR?

akrem-chabchoub:
Still not fully tested yet.
Currently on hold due to other priorities.
Work is expected to resume asap.

pkg/api/bzz.go
```diff
@@ -137,8 +142,18 @@ func (s *Service) bzzUploadHandler(w http.ResponseWriter, r *http.Request) {
...
 
-	if headers.IsDir || headers.ContentType == multiPartFormData {
-		s.dirUploadHandler(ctx, logger, span, ow, r, putter, r.Header.Get(ContentTypeHeader), headers.Encrypt, tag, headers.RLevel, headers.Act, headers.HistoryAddress)
+	contentTypeHdr := strings.TrimSpace(headers.ContentType)
+	r.Header.Set(ContentTypeHeader, contentTypeHdr)
+	mt, _, errParseCT := mime.ParseMediaType(contentTypeHdr)
+	isMultipart := errParseCT == nil && mt == multiPartFormData
+	if headers.IsDir || isMultipart {
+		if contentTypeHdr == "" {
+			logger.Debug("content-type required for directory upload")
```
acud: if both messages are showing the same, maybe `Error` is enough?

pkg/api/bzz.go
```diff
@@ -137,8 +142,18 @@ func (s *Service) bzzUploadHandler(w http.ResponseWriter, r *http.Request) {
 		logger:         logger,
 	}
 
-	if headers.IsDir || headers.ContentType == multiPartFormData {
-		s.dirUploadHandler(ctx, logger, span, ow, r, putter, r.Header.Get(ContentTypeHeader), headers.Encrypt, tag, headers.RLevel, headers.Act, headers.HistoryAddress)
+	contentTypeHdr := strings.TrimSpace(headers.ContentType)
+	r.Header.Set(ContentTypeHeader, contentTypeHdr)
+	mt, _, errParseCT := mime.ParseMediaType(contentTypeHdr)
+	isMultipart := errParseCT == nil && mt == multiPartFormData
+	if headers.IsDir || isMultipart {
```
acud: `if (headers.IsDir || isMultipart) && contentTypeHdr == "" {...}`

pkg/api/bzz.go
akrem-chabchoub: Yes, my bad, It should be removed

pkg/api/bzz.go
akrem-chabchoub: Maybe this is more readable, wdyt ?
```go
isDirUpload := headers.IsDir || isMultipart
if !isDirUpload {
	s.fileUploadHandler(ctx, logger, span, ow, r, putter, headers.Encrypt, tag, headers.RLevel, headers.Act, headers.HistoryAddress)
	return
}

if contentTypeHdr == "" {
	logger.Error(nil, "content-type required for directory upload")
	jsonhttp.BadRequest(w, errInvalidContentType)
	return
}

s.dirUploadHandler(ctx, logger, span, ow, r, putter, headers.Encrypt, tag, headers.RLevel, headers.Act, headers.HistoryAddress)
```

acud: both are fine :+1: the point was to just reduce nesting

pkg/api/bzz.go
```diff
@@ -137,11 +142,24 @@ func (s *Service) bzzUploadHandler(w http.ResponseWriter, r *http.Request) {
 		logger:         logger,
 	}
 
-	if headers.IsDir || headers.ContentType == multiPartFormData {
-		s.dirUploadHandler(ctx, logger, span, ow, r, putter, r.Header.Get(ContentTypeHeader), headers.Encrypt, tag, headers.RLevel, headers.Act, headers.HistoryAddress)
+	contentTypeHdr := strings.TrimSpace(headers.ContentType)
+	r.Header.Set(ContentTypeHeader, contentTypeHdr)
```
martinconic, line 146: Is this necessary?

pkg/api/bzz.go
```diff
@@ -137,11 +142,24 @@ func (s *Service) bzzUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+	mt, _, errParseCT := mime.ParseMediaType(contentTypeHdr)
+	isMultipart := errParseCT == nil && mt == multiPartFormData
+
+	isDirUpload := headers.IsDir || isMultipart
+	if !isDirUpload {
+		s.fileUploadHandler(ctx, logger, span, ow, r, putter, headers.Encrypt, tag, headers.RLevel, headers.Act, headers.HistoryAddress)
+		return
+	}
+
+	if contentTypeHdr == "" {
```
martinconic, line 156: Maybe a test for empty content type if there is none? Also if no content type and no body?

pkg/api/dirs.go
```diff
@@ -58,7 +57,7 @@ func (s *Service) dirUploadHandler(
 	}
 
 	// The error is ignored because the header was already validated by the caller.
```
martinconic, line 59: The comment is not relevant anymore, right? the header is not validated by the caller. Maybe something like this: 
```
// Parse error is ignored; unsupported media types are caught by the default case below.
```

, or do we want to validate here?

sbackend123 (approved):
How would it work on edge cases like:
- empty body

pkg/api/bzz.go
akrem-chabchoub: Before my changes the `dirUploadHandler` was accepting `contentTypeString` (although it is present request).
So I changed that [here](https://github.com/ethersphere/bee/pull/5361/changes#diff-7d6cda1d4cc1d145ce2a7bda675bc253d442467dd1cc6d47eba487ff5fa06161L47) and did trim and set the new value in main handler.

acud: I think that I agree with @martinconic that this is not really clear why this is needed. If you're already reading, cleaning, parsing and checking the content type header - it does not make any more sense to update the value in the `Header` type.

akrem-chabchoub: @acud @martinconic 
Good point, i agree this can look a bit strange at first.

i set the header on purpose, so we clean `Content-Type` one time, and then all next steps read the same value.

without this, one part can use the trimmed value, but another part (`fileUploadHandler`) can still read the original raw header. this can cause small inconsistent behavior later, especially for multipart handling and metadata.

if changing the header feels too implicit, i can refactor and pass the cleaned value explicitly to the next functions. i am fine with both styles, but i think we should keep one canonical content-type value in one place.

pkg/api/dirs.go
akrem-chabchoub: Good catch, agreed.
I’d keep validation behavior as-is here, since parse failures and unsupported types is handled in the switch case.

pkg/api/bzz.go
acud: Thanks @akrem-chabchoub.
The problem from my perspective is that you're already doing cleaning and checking of a value, mutating it, but then you're telling downstream function calls to potentially do the same, again, because you've removed the argument from the signature. This is from my perspective not necessary and potentially duplicates code.

This may work in middleware style calls where the middlewares are decoupled (and you want to be setting a header on one middleware that processes for example a user-login or something along those lines, providing you with extra details down the call stack), but I can't see how this is needed here (you're already at the last middleware in the API). 

I won't block the PR on this, please choose how you'd like to proceed with this.

acud (approved):

PR #5370, opened 2026-02-17
fix(p2p): skip addressbook and reacher for unreachable addresses

gacevicljubisa:
### Checklist

### Description
When the peerstore has no addresses for a peer, peerMultiaddrs was falling back to RemoteMultiaddr (NAT IP + ephemeral port). This caused the reacher to ping unreachable ports and hive to gossip unreachable addresses.

Remove the fallback from peerMultiaddrs and handle it explicitly at call sites — RemoteMultiaddr is now only used as ObservedUnderlay for the handshake. Addressbook storage and reacher notifications are skipped when no real peerstore addresses are available.

### DIFF
--- a/pkg/p2p/libp2p/connections_test.go
+++ b/pkg/p2p/libp2p/connections_test.go
@@ -13,6 +13,7 @@ import (
 	"reflect"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -27,6 +28,7 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/statestore/mock"
 	"github.com/ethersphere/bee/v2/pkg/swarm"
 	"github.com/ethersphere/bee/v2/pkg/topology/lightnode"
+	libp2pcrypto "github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/libp2p/go-libp2p/p2p/host/eventbus"
 
 	libp2pm "github.com/libp2p/go-libp2p"
@@ -34,6 +36,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
 	libp2ppeer "github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/peerstore"
 	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
 	swarmt "github.com/libp2p/go-libp2p/p2p/net/swarm/testing"
 	ma "github.com/multiformats/go-multiaddr"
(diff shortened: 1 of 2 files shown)


mfw78:
Can we please have some consideration in regards to this. Recent introductions / changes related to the observed overlay and fallback solutions has created undocumented friction introducing arbitrary 10s delays that weren't documented, in addition to assumptions about the use of the libp2p `identify` service. Can these changes actually be _documented_ somewhere?

pkg/p2p/libp2p/libp2p.go
akrem-chabchoub, line 1124: Should persist only if there is `peerAddrs` also?

gacevicljubisa:
> Can we please have some consideration in regards to this. Recent introductions / changes related to the observed overlay and fallback solutions has created undocumented friction introducing arbitrary 10s delays that weren't documented, in addition to assumptions about the use of the libp2p `identify` service. Can these changes actually be _documented_ somewhere?

The PR itself is a fix for the addressbook/reacher pollution caused by the fallback RemoteMultiaddr. 
And, yes, the new 10s peerstoreWaitAddrsTimeout is not documented anywhere. It only waits if peerstore Addrs doesn't return any addresses for a given peer. I'll add a clear godoc block for those changes. And discuss with team if we can document elsewhere.

Do you have any objections to the actual changes in this PR (the logic, skipping addressbook/reacher when we only have RemoteMultiaddr, etc.)? Or is the comment purely about the missing documentation?

mfw78:
> It only waits if peerstore Addrs doesn't return any addresses for a given peer.

The comment is mostly that when this is being touched, I want to make it clear that changes with arbitrary timeouts are affecting client developers. By waiting on `peerstoreWaitAddrsTimeout`, this creates an implicit requirement for using the libp2p `identify` protocol (undocumented), and for example, in `rust-libp2p`, the `identify` protocol will automatically filter all private addresses, and when the client doesn't have a publicly accessible addresses, there is nothing to communicate here, meaning that _all_ light clients will be subject to a 10s delay.

I effectively had to fork the `rust-libp2p` `identify` protocol and then during the middle of the handshake push the observed multiaddr via `identify` targeted to that specific peer (otherwise by just adding the external address, it would push the NAT-locked port to ALL connected clients, therefore polluting the peer store on those clients with inaccessible ports).

~~In regards to the actual PR, I'm not sure about skipping the address book, as light clients can still effectively participate in kademlia routing so they are not totally useless~~. Actually, sorry, I think in regards to this that there should be be support for empty observed multiaddrs, and if it is empty, don't store it (inaccessible light node is irrelevant).

gacevicljubisa:
@mfw78 Here is the [PR](https://github.com/ethersphere/bee/pull/5326) where we tried to make support for empty underlays. Please check the discussion there. It was decidede to not merge it beacuse we have added RemoteMultiaddress as fallback. But we can reconsider this...

gacevicljubisa:
@mfw78 what is you opinion on the support of empty underlays, in this [PR](https://github.com/ethersphere/bee/pull/5326)

pkg/p2p/libp2p/libp2p.go
gacevicljubisa: Here are the underlays the responder computed about themselves inside Handle(), and we should save them.

pkg/p2p/libp2p/libp2p.go
```diff
@@ -644,7 +660,8 @@ func (s *Service) handleIncoming(stream network.Stream) {
 		return
 	}
 
-	if i.FullNode {
+	// Only persist in addressbook when we have real peerstore addresses.
+	if i.FullNode && len(peerAddrs) > 0 {
```
martinconic, line 664: Shouldn't we have `observedAddrs` here?

pkg/p2p/libp2p/libp2p.go
```diff
@@ -727,7 +744,9 @@ func (s *Service) handleIncoming(stream network.Stream) {
 		return
 	}
 
-	s.notifyReacherConnected(overlay, peerMultiaddrs)
+	if len(peerAddrs) > 0 {
```
martinconic, line 747: Same here?

pkg/p2p/libp2p/libp2p.go
martinconic: If the peer is behind NAT and doesn't know its public IP (or reports a private IP/port that is unreachable from outside), and we save that to our addressbook, we are poisoning our own storage with unreachable addresses?

pkg/p2p/libp2p/libp2p.go
gacevicljubisa: No, this is actual intention of the PR, not to store `RemoteMultiaddr()`.

pkg/p2p/libp2p/libp2p.go
gacevicljubisa: Here as well, we take `peerAddrs`, which do not have `RemoteMultiaddr`. 
if there are no `peerAddrs`, `RemoteMultiaddr` is reassigned to observedAddrs, and that is why it is not sent to reacher in this situation.

pkg/p2p/libp2p/libp2p.go
gacevicljubisa: The intention of PR is to reduce that situation.

Issue #5166, opened 2025-07-23
Running only one sampling process


nugaon:
#### Summary
Implement synchronization mechanism to ensure only one sampling process runs at a time in agent.go, and prevent concurrent calls to rchash endpoint which can cause race conditions and inconsistent state.

#### Motivation
Currently, multiple sampling processes can be triggered concurrently, leading to:
1. Race conditions in the agent.go module
2. Inconsistent state when multiple processes try to access and modify the same resources
3. Performance degradation and increased error rates when rchash endpoint is called simultaneously
4. Potential deadlocks or system instability under high load

#### Implementation
Potential implementation approaches:
Add a mutex lock around the sampling process in agent.go

acud:
already fixed

PR #5373, opened 2026-02-18
fix: deduplicate concurrent ReserveSample calls with singleflight

gacevicljubisa:
### Checklist

### Description
fixes https://github.com/ethersphere/bee/issues/5166

**The problem**: makeSample (storage incentives agent) and SampleWithProofs (`/rchash` API endpoint) both can independently call ReserveSample, meaning concurrent calls with the same parameters would trigger duplicate expensive sampling operations.

**The fix**: Introduces a singleflight.Group on the Agent struct to deduplicate concurrent calls. Both callers now go through a shared reserveSampleAndHash method keyed by anchor + depth. If multiple goroutines call it simultaneously with the same key, only one actual ReserveSample + sampleHash executes — the rest wait and share the result.

### DIFF
--- a/pkg/storageincentives/agent.go
+++ b/pkg/storageincentives/agent.go
@@ -27,6 +27,7 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/storer"
 	"github.com/ethersphere/bee/v2/pkg/swarm"
 	"github.com/ethersphere/bee/v2/pkg/transaction"
+	"resenje.org/singleflight"
 )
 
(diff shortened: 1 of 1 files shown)


acud:
@gacevicljubisa can you explain how are the concurrent calls possible?

gacevicljubisa:
> @gacevicljubisa can you explain how are the concurrent calls possible?

@acud I have updated the description with explanation.

akrem-chabchoub (approved):
Maybe some tests to cover these changes ?

PR #5374, opened 2026-02-18
test: reserve sample deterministic test + benchmarks

acud:
### Description
Adds a non-random test to the reserve sample and some benchmarks as a baseline.

Needed for verifying that ReserveSample reaches the same result deterministically (and covers an e2e over all the hashing operations involved).

Benchmarks are needed for ReserveSample optimization.

### DIFF
--- a/pkg/accesscontrol/access_test.go
+++ b/pkg/accesscontrol/access_test.go
@@ -71,7 +71,6 @@ func getPrivKey(keyNumber int) *ecdsa.PrivateKey {
 }
 
 func TestDecryptRef_Publisher(t *testing.T) {
-	t.Parallel()
 	ctx := context.Background()
 	id1 := getPrivKey(1)
 	s := kvsmock.New()


pkg/storer/sample_test.go
```diff
@@ -325,3 +382,99 @@ func assertSampleNoErrors(t *testing.T, sample storer.Sample) {
 		t.Fatalf("got unexpected invalid stamps")
 	}
 }
+
+// Benchmark results:
```
akrem-chabchoub, line 386: Why it is intentional to keep benchmark results here ?

pkg/accesscontrol/access_test.go
```diff
@@ -71,7 +71,6 @@ func getPrivKey(keyNumber int) *ecdsa.PrivateKey {
 }
 
 func TestDecryptRef_Publisher(t *testing.T) {
-	t.Parallel()
```
akrem-chabchoub, line 74: This is for flaky test ? or something else ?

martinconic (approved):

pkg/storer/sample_test.go
acud: so that when we improve on them later on, that there's a baseline to compare it against

pkg/accesscontrol/access_test.go
acud: yes

PR #5378, opened 2026-02-19
chore: close as soon as possible handshake stream

janos:
### Checklist

### Description
@gacevicljubisa noticed that handshake.FullClose() can be performed earlier. This PR optimizes and cleans the code related to the handshake stream.

### DIFF
--- a/pkg/p2p/libp2p/libp2p.go
+++ b/pkg/p2p/libp2p/libp2p.go
@@ -627,36 +627,29 @@ func (s *Service) handleIncoming(stream network.Stream) {
 
 	overlay := i.BzzAddress.Overlay
 
+	if err = handshakeStream.FullClose(); err != nil {
+		s.logger.Debug("stream handler: could not close stream", "peer_address", overlay, "error", err)
+		s.logger.Error(nil, "stream handler: unable to handshake with peer", "peer_address", overlay)
+		_ = s.Disconnect(overlay, "could not fully close stream on handshake")
+		return
+	}
+
 	blocked, err := s.blocklist.Exists(overlay)
 	if err != nil {
 		s.logger.Debug("stream handler: blocklisting: exists failed", "peer_address", overlay, "error", err)
 		s.logger.Error(nil, "stream handler: internal error while connecting with peer", "peer_address", overlay)
-		_ = handshakeStream.Reset()
 		_ = stream.Conn().Close()
 		return
 	}
 
 	if blocked {
 		s.logger.Error(nil, "stream handler: blocked connection from blocklisted peer", "peer_address", overlay)
-		_ = handshakeStream.Reset()
 		_ = s.host.Network().ClosePeer(peerID)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/p2p/libp2p/libp2p.go
```diff
@@ -627,36 +627,29 @@ func (s *Service) handleIncoming(stream network.Stream) {
...
 	}
 
 	if blocked {
 		s.logger.Error(nil, "stream handler: blocked connection from blocklisted peer", "peer_address", overlay)
-		_ = handshakeStream.Reset()
 		_ = s.host.Network().ClosePeer(peerID)
 		return
 	}
 
 	if exists := s.peers.addIfNotExists(stream.Conn(), overlay, i.FullNode); exists {
```
acud, line 651: consider (not sure if this is problematic):
if you close the stream earlier before the peer has been added to this peers struct, the other peer will progress with the handshake protocol (continues execution) and may fire a request before the peer was added (the call on this line). i'm not sure if this can cause downstream problems (maybe a protocol already tries to send a message, and some check on the libp2p abstraction that tries to check if the peer exists on this list happens before the peer was added

janos: Yes, interesting observation. The question is, is handshake stream keeping this synchronization. I am not sure that I can see that by keeping the handshake stream active, other protocols are blocked from sending messages. Am interested to see if I am missing something. Could you elaborate more on your assumptions, or try to see if there could be problems?

I even cannot see why handshake stream is even created earlier than it is needed. This can also be changed.

acud: iirc the `FullClose` usage was meant as a means of synchronizing both peers and making sure both peers are at the same stage of the handshake. Previously they would be reading on the stream, waiting for something, and when we would close the stream they could then continue execution.

For example on the connecting side, the protocols start firing [here](https://github.com/ethersphere/bee/blob/master/pkg/p2p/libp2p/libp2p.go#L1132).
The receiver [here](https://github.com/ethersphere/bee/blob/master/pkg/p2p/libp2p/libp2p.go#L676).

The question is whether under some lock contention on that peer registry or under circumstances will we see one peer finishing the handshake on the other before the other one making it to the line where that peer gets onto the peer registry. If yes then there might be some downstream errors or some weird halfway state (peer is "halfway" connected). Whether the current handshake protocol prevents that fully - can't say, need to have a further look.

martinconic:
I have also tried these and the issues from connections_test.go seems to get worse. Investigating this I am getting to a similar opinion as Elad. I am not sure about Closing early.

janos:
I am closing the PR as it appears that the stream is kept active in order to synchronize node internals. There is a possibility to improve the code to encapsulate the required operations on both sides, to be more obvious why the stream is kept open, but it is just code maintenance, not a functionality improvement. There is other PR that deals with the improvements of the tests that may be related to this part of the code as well.

Issue #5174, opened 2025-07-29
Optimize ReserveSample Function


nugaon:
# Task
Optimize the `ReserveSample` function in the Bee node to improve performance and reduce resource consumption. This function is critical for node operators participating in storage incentives, but currently runs too slowly to complete within required lottery round time constraints (issue #5132). Performance bottlenecks in sample fetching negatively impact node operators and contribute to higher staking requirements (issue #4849).

## Acceptance criteria

- Sample generation completes within lottery round time constraints
- Reduced resource (CPU, memory etc.) usage during sampling process
- No regressions in sample correctness
- Performance metrics showing improvement compared to baseline

acud:
I am working on this ticket. There's already a few directions for optimization here.
One this is missing and that's at least some basic benchmarks on the current performance, this is added with #5374.

Also, optimizing stuff here on the level of the actual `ReserveSample` function is like throwing stones at a tank. The major problems here are: 1. the BMT hasher. 2. the memory allocations generated by iterating over the chunks that need to be sampled.

Regarding the BMT hasher, there are several inefficiencies:
1. it spins up massive amount of goroutines for every hashing operation (every node = goroutine = 255 goroutines per chunk). This means 256 allocations for a 4k chunk
2. this pressures the GC and the scheduler

I already have a solution for this. Currently for Linux amd64 only. I've compiled the XKCP AVX implementations of keccak (SIMD) into assembler code and found a way to reference them from the go code without using cgo (direct assembler code). There's a significant improvement there already, but it needs further testing.

Regarding the memory allocations:
1. The fact that there's an iterator on the reserve sample method causes massive amounts of allocations needed in order to move chunk data to that method from the storage. 
2. This is opposed to it fitting on the stack (so it does not escape). 
3. We could move the majority of those hashing operations down the call stack and "closer" to the data, allowing the hashing to happen with the data on the stack, returning only the transformed addresses together with the chunk addresses. The chunks that are needed for the proofs (very little) could then be individually fetched from the storage and put through the prover, minimizing the memory stress on the application.

acud:
The SIMD hasher has been merged. There's one more improvement I would try to get merged and that is to move the `ReserveSample` iterators into the reserve, and change with it also the `Getter` interfaces (or extend them) with an interface that reads directly into a buffer (that can be reused).
[...]

PR #5381, opened 2026-02-25
perf: bmt SIMD hasher

acud:
### Description
This PR tries to improve the current BMT hasher to use SIMD when available. 

The current BMT hasher is highly inefficient - it uses massive goroutine spawning to compute a single chunk address. For a full chunk, this means 255 goroutines are spawned per chunk, creating GC stress and significant scheduler stress. For the ReserveSample function used in the redistribution game, this means excessive memory and CPU stress in order to calculate the reserve sample.

The idea here was to use an existing keccak implementation, compile it to assembler and try to call it directly from our go code without having to use `cgo` which comes with its own set of side-effects.

I used (I==me+Claude) the XKCP project (from the keccak authors) and built a build script that builds, extracts and wraps the compiled code correctly. Currently only linux amd64 is supported. Windows and mac should fall back to the go legacy sha3 hasher.

So far the results are promising:
- `x1.6` faster BMT hashing on my local machine (laptop)
- `x2.5` faster on AVX2 supported data-center CPUs (Hetzner)
- `x5` faster on newer AVX512 architectures

There's a few more things to iron out and test:
- [x] is the scalar hashing fallback (to the go crypto legacy keccak hashing) reasonably fast? (I am imagining this only for dev use anyway - mac devs etc) or do we need to fallback to the previous implementation of BMT that spawned all those goroutines?
- [x] as of such - do we need a BMT factory? (to spin up the right type of hasher) do we want in this case to maintain both implementations? (copies?)
- since SIMD instructions can cause CPU throttling, we must test on different configurations and providers and see that the possible throttling doesn't nullify the performance improvements that SIMD gives us. in other words we have to compare workload benchmarks against the current implementation on trunk.
- [ ] figure out whether the excessive stack frame allocation can somehow be avoided
- [ ] see if using `bmtpool` would improve anything at all
- [x] fix the linker error: `/usr/bin/ld: warning: /tmp/go-link-1425710637/000001.o: missing .note.GNU-stack section implies executable stack
/usr/bin/ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker`

Worth noting:
- I've also tried to port the existing go crypto library legacy sha3 implementation (written in go) to use the new go 1.26 SIMD primitives. Still needs a bit of research.
- windows not supported cuz `shadow space` (don't ask)

### Test plan
- run on selected production nodes for a month and see if all is well, no panics, better performance. Only linux amd64, various configurations.

### Related Issue (Optional)
#5174 

### References:
- https://github.com/XKCP/XKCP
- https://github.com/acud/XKCP <- my fork with the necessary build setup to create the necessary .syso files.
- https://pkg.go.dev/simd/archsimd


acud:
Just as an update - these changes are being actively tested on a few mainnet nodes to verify everything works well and we don't get any memory corruption/panics

martinconic:
Peer reviewing with Claude found the following, maybe are suitable for addressing:

1. Catastrophic regression on the default linux/amd64 configuration (severity: high)

When SIMDOptIn=false (the default), the code still takes the new SIMD code path but calls keccak.Sum256x4Scalar — which spawns goroutines internally:


// scalar fallback spawns 4 goroutines per batch
for i := 0; i < 4; i++ { go func() { ... }() }
wg.Wait()
For a 128-branch BMT (64 leaf nodes, batchWidth=4): the scalar path spawns ~124 goroutines vs the original path's 64. Combined with WaitGroup + Mutex overhead on every batch, benchmarks show ~14× slower than the original goroutine path. This affects every linux/amd64 node that doesn't explicitly pass --use-simd-hashing — i.e., all standard deployments.

2. bmtpool singleton won't benefit from SIMD even when opted in (severity: high)

bmtpool.init() runs before start.go sets SIMDOptIn=true. So [pkg/bmtpool/bmtpool.go](vscode-webview://0dun2l9tc71qscrnlu51b5m5c6ph0rjldj4320s5atmgvv90v31r/pkg/bmtpool/bmtpool.go)'s global pool — used by cac.go, the upload pipeline, pss/trojan.go, and most proof generation — is always created with SIMD disabled. Only code that calls bmt.NewPrefixHasher() directly (i.e., ReserveSample) gets the benefit.

pkg/keccak/CHECKSUM
```diff
@@ -0,0 +1,2 @@
+398953aa6d446ef64210a94ddf1f807ad6b9137d913b66adf83bbd6f665d5fbd  keccak_times4_linux_amd64.syso
+9e53e1b590f657b2c817b816126111dedc167b5b7543e50106dd4435fee48b3c  keccak_times8_linux_amd64.syso
```
akrem-chabchoub, line 2: Should we mention from which commit we build this ? or not needed ?

pkg/bmt/hasher_goroutine.go
```diff
@@ -0,0 +1,338 @@
+// Copyright 2021 The Swarm Authors. All rights reserved.
```
akrem-chabchoub: 2021 -> 2026 (Its there for 5 files)

cmd/bee/cmd/cmd.go
```diff
@@ -88,6 +88,7 @@ const (
 	optionAutoTLSDomain                    = "autotls-domain"
 	optionAutoTLSRegistrationEndpoint      = "autotls-registration-endpoint"
 	optionAutoTLSCAEndpoint                = "autotls-ca-endpoint"
+	optionUseSIMD                          = "use-simd-hashing"
```
akrem-chabchoub, line 91: Could we add it to `packaging` folder ?

akrem-chabchoub (changes requested):
Thanks @acud !
I'm trying to understand/test more the PR, but this is a small round of review/clarifications...

pkg/keccak/CHECKSUM
acud: added a new file that references that current `HEAD` on the XKCP remote :+1:

cmd/bee/cmd/cmd.go
akrem-chabchoub: @acud Also I think this should be in `docker-compose.yml`

akrem-chabchoub (approved):

martinconic (approved):

pkg/bmt/pool_simd.go
```diff
@@ -0,0 +1,170 @@
...
+	concat     [8][]byte
+	leafConcat [8][]byte
+}
+
+// simdNode is a reusable segment hasher node in the SIMD BMT.
+type simdNode struct {
+	isLeft      bool
+	parent      *simdNode
+	left, right []byte
+	hasher      hash.Hash
```
janos: It looks to me that hasher from simdNode can be moved to simdTree struct in order to avoid creating hashers for every node (255 hashers per chunk) on line pool_simd.go:137, as hashers are all the same. By moving the hasher to the simdTree, that hasher can be created only once in newSIMDTree and used in hasher_simd.go:142 and hasher_simd.go:99 as h.bmt.hasher instead to get it from the node. That would avoid creating a lot of hashers when hashing which should reduce the gc pressure. The newSIMDNode could accept the hasher size instead of the hasher. But that just a smaller optimization, the main one would be to avoid calling hashfunc() in double nested loop on line 137. It looks to me that this change would be logically correct, if I did not miss something else.

pkg/bmt/pool_simd.go
```diff
@@ -0,0 +1,170 @@
...
+	prefixLen := len(prefix)
+	n := newSIMDNode(0, nil, hashfunc())
+	prevlevel := []*simdNode{n}
+	allLevels := [][]*simdNode{prevlevel}
+	count := 2
+	for level := depth - 2; level >= 0; level-- {
+		nodes := make([]*simdNode, count)
+		for i := 0; i < count; i++ {
+			parent := prevlevel[i/2]
+			nodes[i] = newSIMDNode(i, parent, hashfunc())
```
janos: Related to the previous comment on line 114. Hashers are created for every node, but it looks to me that only one hasher is needed, by the usage of the hasher. Since the BMT Pool guarantees that a single hasher instance used by one worker at a time, and the SIMD hasher itself never spawns internal goroutines, it is guaranteed that the single bmt hasher is only ever accessed sequentially.

janos (approved):
Nice one. I have comments that may improve the performance by reducing the gc pressure, otherwise it is solid.

PR #5384, opened 2026-02-27
test: fix flaky reserve worker startup

acud:
### Description
A data race in the reserve worker's startup was created in a way that it may get scheduled after the dependent tests already started. I've added a channel to synchronize this behavior. In `node.go` I am not using this synchronization as it appears to affect the node startup time in the integration tests.

This fixes the flakes in the TestReserveEvict and the TestReplaceOldIndex tests.

### DIFF
--- a/pkg/storer/reserve.go
+++ b/pkg/storer/reserve.go
@@ -136,7 +136,9 @@ func (db *DB) reserveWorker(ctx context.Context, ready chan<- struct{}) {
 		db.events.Trigger(reserveOverCapacity)
 	}
 
-	close(ready)
+	if ready != nil {
+		close(ready)
+	}
 
 	for {
 		select {
@@ -470,7 +472,6 @@ func (db *DB) SubscribeBin(ctx context.Context, bin uint8, start uint64) (<-chan
 	errC := make(chan error, 1)
 
 	db.inFlight.Go(func() {
-
 		trigger, unsub := db.reserveBinEvents.Subscribe(string(bin))
 		defer unsub()
 		defer close(out)
--- a/pkg/storer/reserve_test.go
+++ b/pkg/storer/reserve_test.go
@@ -77,7 +77,11 @@ func TestIndexCollision(t *testing.T) {
 		}
 		readyC := make(chan struct{})
 		storer.StartReserveWorker(context.Background(), pullerMock.NewMockRateReporter(0), networkRadiusFunc(0), readyC)
-		<-readyC
+		select {
+		case <-readyC:
+		case <-t.Context().Done():
+			t.Fatal("start reserve worker timeout")
+		}
 		testF(t, baseAddr, storer)
 	})
 	t.Run("mem", func(t *testing.T) {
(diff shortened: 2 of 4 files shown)


martinconic:
I have run the CI/CD tests on this PR multiple times, and one time on tests failed that I did not saw before:
https://github.com/ethersphere/bee/actions/runs/22489298892/job/65364049599?pr=5384

Could be because of these changes? can you please check it out?

acud:
Checked it out. Can't see how this is related. The goroutine dump in the CI suggest some problem in the CI runner. Note the IO waits etc.

pkg/storer/reserve_test.go
```diff
@@ -374,7 +387,9 @@ func TestNetworkRadius(t *testing.T) {
 		if err != nil {
 			t.Fatal(err)
 		}
-		storer.StartReserveWorker(context.Background(), pullerMock.NewMockRateReporter(0), networkRadiusFunc(1))
+		readyC := make(chan struct{})
+		storer.StartReserveWorker(context.Background(), pullerMock.NewMockRateReporter(0), networkRadiusFunc(1), readyC)
+		<-readyC
 		time.Sleep(time.Second)
```
akrem-chabchoub, line 425: Will `time.Sleep` still be flaky?
Not sure whether [synctest](https://pkg.go.dev/testing/synctest) can help here ?

acud: so far i have tried to reproduce any flakiness of these tests locally and even without synctest it passes without a problem

pkg/storer/reserve.go
```diff
@@ -135,6 +136,8 @@ func (db *DB) reserveWorker(ctx context.Context) {
 		db.events.Trigger(reserveOverCapacity)
 	}
 
+	close(ready)
```
gacevicljubisa: Even if it not read, and we are sure it is created, maybe it is better to put:
```go
if ready != nil {
    close(ready)
}
```

pkg/storer/storer.go
```diff
@@ -674,10 +674,10 @@ func (db *DB) SetRetrievalService(r retrieval.Interface) {
 	db.retrieval = r
 }
 
-func (db *DB) StartReserveWorker(ctx context.Context, s Syncer, radius func() (uint8, error)) {
+func (db *DB) StartReserveWorker(ctx context.Context, s Syncer, radius func() (uint8, error), ready chan<- struct{}) {
```
janos, line 679: Document the purpose of the ready chan so that it is clear that it will only be closed once and nothing will be written to it. There are a few indirections in the code to make sure that this is the case. Also as @gacevicljubisa nicely noticed, the channel can be nil and in that case, should not be closed.

pkg/storer/reserve_test.go
```diff
@@ -75,7 +75,9 @@ func TestIndexCollision(t *testing.T) {
 		if err != nil {
 			t.Fatal(err)
 		}
-		storer.StartReserveWorker(context.Background(), pullerMock.NewMockRateReporter(0), networkRadiusFunc(0))
+		readyC := make(chan struct{})
+		storer.StartReserveWorker(context.Background(), pullerMock.NewMockRateReporter(0), networkRadiusFunc(0), readyC)
+		<-readyC
```
janos: It would be good to avoid naked reads from the channel here. For example to have select on on t.Context().Done()?

This goes for all instances in this pr.

acud: why? it is a test. i can't see why you'd like to have a graceful context cancellation here. i can't really clearly see when a test context would be cancelled (ctrl-c? 600s timeout?), but in any case, if we are stuck on the read from the channel i'd actually like to see an ugly test failure with a goroutine dump (all goroutines blocked message) rather than having the test exit gracefully.

janos: Test failure on timeout would be much easier to read than to investigate the panic goroutine dump.

PR #5390, opened 2026-03-09
fix(pushsync): prevent silent chunk loss on shallow receipts

gacevicljubisa:
### Checklist

### Description
Chunks could silently disappear from the network after a successful upload due to three bugs in the pushsync/pusher pipeline.

### 1. Storer stored chunks it was about to evict (root cause)

- When a forwarder had no closer peer (`ErrWantSelf`) but the chunk was outside its AOR, it stored the chunk and sent a receipt anyway
- The chunk landed in a low proximity bin and was evicted almost immediately by `unreserve()`, leaving the network with no copy
- **Fix:** return the pre-existing but unused `ErrOutOfDepthStoring` when `ErrWantSelf` fires but `proximity(chunk, self) < radius`; the origin treats it as a hard failure and retries with the next peer

### 2. Shallow receipt abandoned inflight parallel pushes

- On any shallow receipt, `pushToClosest` returned immediately, discarding results from other parallel multiplex pushes that could have been valid
- Only 1 of the 32-attempt error budget was used per call
- **Fix:** treat shallow receipt as a regular peer failure — burn the budget, skip the peer, retry; surface `ErrShallowReceipt` only after the full budget is exhausted

### 3. False `ChunkSynced` after retry exhaustion

- After 6 pusher-level retries, chunks were marked `ChunkSynced` even though no node in the correct neighbourhood had stored them
- **Fix:** report `ChunkCouldNotSync` instead

### DIFF
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -276,13 +276,13 @@ func (s *Service) pushDeferred(ctx context.Context, logger log.Logger, op *Op) (
 			loggerV1.Error(err, "pusher: failed reporting chunk")
 			return true, err
 		}
+		s.attempts.delete(op.identityAddress)
 	case errors.Is(err, pushsync.ErrShallowReceipt):
 		if s.shallowReceipt(op.identityAddress) {
 			return true, err
 		}
-		// Retry budget exhausted: no peer in the correct neighborhood stored the
-		// chunk. Report CouldNotSync rather than Synced to avoid falsely marking
-		// the chunk as delivered.
+		// budget exhausted; report CouldNotSync so the chunk isn't marked delivered
+		s.metrics.TotalCouldNotSync.Inc()
 		if err := s.storer.Report(ctx, op.Chunk, storage.ChunkCouldNotSync); err != nil {
 			loggerV1.Error(err, "pusher: failed to report sync status")
 			return true, err
@@ -292,6 +292,7 @@ func (s *Service) pushDeferred(ctx context.Context, logger log.Logger, op *Op) (
 			loggerV1.Error(err, "pusher: failed to report sync status")
 			return true, err
 		}
+		s.attempts.delete(op.identityAddress)
 	default:
 		loggerV1.Error(err, "pusher: failed PushChunkToClosest")
 		return true, err
@@ -333,13 +334,16 @@ func (s *Service) pushDirect(ctx context.Context, logger log.Logger, op *Op) err
 		if err != nil {
 			loggerV1.Error(err, "pusher: failed to store chunk")
 		}
+		s.attempts.delete(op.identityAddress)
 	case errors.Is(err, pushsync.ErrShallowReceipt):
 		if s.shallowReceipt(op.identityAddress) {
 			return err
 		}
-		// out of attempts for retry, swallow error
-		err = nil
-	case err != nil:
+		// budget exhausted; propagate err instead of falsely reporting success
+		s.metrics.TotalCouldNotSync.Inc()
+	case err == nil:
+		s.attempts.delete(op.identityAddress)
+	default:
 		loggerV1.Error(err, "pusher: failed PushChunkToClosest")
 	}
 
--- a/pkg/pusher/pusher.go
+++ b/pkg/pusher/pusher.go
@@ -280,7 +280,10 @@ func (s *Service) pushDeferred(ctx context.Context, logger log.Logger, op *Op) (
 		if s.shallowReceipt(op.identityAddress) {
 			return true, err
 		}
-		if err := s.storer.Report(ctx, op.Chunk, storage.ChunkSynced); err != nil {
+		// Retry budget exhausted: no peer in the correct neighborhood stored the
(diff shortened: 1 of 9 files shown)


pkg/pushsync/pushsync.go
```diff
@@ -486,12 +503,17 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
 					return result.receipt, nil
-				case errors.Is(err, ErrShallowReceipt):
-					ps.errSkip.Add(idAddress, result.peer, skiplistDur)
-					return result.receipt, err
+				} else if errors.Is(err, ErrShallowReceipt) {
+					// Treat shallow receipt like any other failure: exhaust the full
+					// error budget and wait for any other inflight parallel pushes
+					// (e.g. multiplex forwards) before giving up. Only return
+					// ErrShallowReceipt once the entire budget is spent.
+					shallowReceiptResult = result.receipt
```
acud: you've removed the skiplist addition line that was on the left side of the diff (L493). any particular reason?

pkg/pushsync/pushsync.go
```diff
@@ -419,11 +428,19 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
 				if skip.PruneExpiresAfter(idAddress, overDraftRefresh) == 0 { // no overdraft peers, we have depleted ALL peers
 					if inflight == 0 {
 						if ps.fullNode {
+							// If a peer already has the chunk (even at wrong depth), don't
+							// store locally — the chunk is closer to its neighbourhood than us.
+							if shallowReceiptResult != nil {
```
acud: this prevents the pss code from executing. i'm not sure if its 100% correct. maybe for the sake of eventual correctness you could put this below the pss unwrap block (before the returning of the error)

pkg/pushsync/pushsync.go
gacevicljubisa: actually it is not removed, it is on the common failure path at [pushsync.go:524](https://github.com/ethersphere/bee/blob/68b9c47199cf3a40cd27bf265e38f4c64912628b/pkg/pushsync/pushsync.go#L524)

pkg/pushsync/pushsync.go
```diff
@@ -486,12 +503,17 @@ func (ps *PushSync) pushToClosest(ctx context.Context, ch swarm.Chunk, origin bo
...
-				case errors.Is(err, ErrShallowReceipt):
-					ps.errSkip.Add(idAddress, result.peer, skiplistDur)
-					return result.receipt, err
+				} else if errors.Is(err, ErrShallowReceipt) {
+					// Treat shallow receipt like any other failure: exhaust the full
+					// error budget and wait for any other inflight parallel pushes
+					// (e.g. multiplex forwards) before giving up. Only return
+					// ErrShallowReceipt once the entire budget is spent.
+					shallowReceiptResult = result.receipt
+					result.err = err
```
acud: not sure if I understand this fully, but this statement and the next one in the `else` block seem like `assigned but never used` where is `result.err` used?

pkg/pusher/pusher.go
```diff
@@ -280,7 +280,10 @@ func (s *Service) pushDeferred(ctx context.Context, logger log.Logger, op *Op) (
 		if s.shallowReceipt(op.identityAddress) {
 			return true, err
 		}
-		if err := s.storer.Report(ctx, op.Chunk, storage.ChunkSynced); err != nil {
+		// Retry budget exhausted: no peer in the correct neighborhood stored the
+		// chunk. Report CouldNotSync rather than Synced to avoid falsely marking
+		// the chunk as delivered.
+		if err := s.storer.Report(ctx, op.Chunk, storage.ChunkCouldNotSync); err != nil {
```
martinconic, line 286: Changing to ChunkCouldNotSync, will cause some false reporting because the ChunkCouldNotSync state is not checked and updated in uploadstore.go ->Report function. Maybe you would need to alsi include in that switch statement also ChunkCouldNotSync.

pkg/pushsync/pushsync.go
```diff
@@ -299,6 +299,14 @@ func (ps *PushSync) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream)
 
 	switch receipt, err := ps.pushToClosest(ctx, chunk, false); {
 	case errors.Is(err, topology.ErrWantSelf):
+		// Only store if we are actually within our neighborhood. If the chunk
+		// is outside our AOR we would store it in a low bin and unreserve()
+		// would evict it almost immediately, leaving the chunk nowhere on the
+		// network while the origin believes it was delivered.
+		if swarm.Proximity(ps.address.Bytes(), chunkAddress.Bytes()) < rad {
```
martinconic, line 300: What happens if you have doubling set on, and the chunk is saved into  a sister neighborhood? Should we have here a check based on the committed depth?

lat-murmeldjur:
This looks good to me still.
see https://github.com/ethersphere/SWIPs/pull/97 for a follow up logic extension

Issue #4884, opened 2024-10-30
Stamperstore ldb writes produce high disk write


Cafe137:
### Context
Windows 10 AMD64, Bee 2.2.0

### Summary
I am writing 1,000,000 chunks at the rate of 200,000 per hour (mantaray with 600,000 forks). This use case produces high disk write that may have bad side effects such as degraded performance or an impact on SSD lifetime. Based on the screenshot number 2 below, the culprit seems to be the stamperstore writes. Network bandwidth is negligible compared to the disk write bandwidth.

![Capture](https://github.com/user-attachments/assets/256e42c0-321d-4312-a335-485edfecc349)

![Capture2](https://github.com/user-attachments/assets/9008a554-0aef-49b8-aa54-3306852e34fb)

### Expected behavior
I have no knowledge on what exactly happens under the hood. My naive assumption is that for each stamping, the corresponding ~2MB ldb gets overwritten, but I may be completely off here.

In case of write-heavy operations like this, in-memory caching with periodic persists to the disk (or WAL, to not lose stamp counters in case of a crash) would likely improve the situation. (stamperstore ≤ 500MB on my machine™)

### Actual behavior
Please see screenshots above. Disk write frequently surpasses 100MB/s.

### Steps to reproduce
On demand I can create a small JS project that creates and writes many chunks in an endless loop.

### Possible solution
Please see expected behaviour.

istae:
Can you describe what the upload sizes are? and which api endpoint are you using?

Cafe137:
> Can you describe what the upload sizes are? and which api endpoint are you using?

I was uploading the wikipedia, and at this point I was writing the mantaray forks. Many small `/bytes` endpoint uploads. Probably minimal time spent splitting/chunking, and most time spent stamping, which turned out to have an I/O bottleneck.

Sorry, I know this isn't much, I will eventually get back to larger data uploads where I can investigate this better.

Cafe137:
@istae I have an update on this

I used the default Activity Monitor of MacOS this time, not Windows.

```
Version: 2.4.0-390a402e
Mode: light
```

I restarted my Bee node to clear Disk IO statistics.

I used 10MB for input `10mb.bin`:

```
dd if=/dev/urandom of=10mb.bin bs=10000000 count=1
```

This is roughly ~2463 chunks that I uploaded using `POST /chunks`.

| Chunks uploaded | Bytes Written |
| --- | --- |
| 0 | 8,5 MB |
[...]

PR #5392, opened 2026-03-09
perf: stamp issuer write coalescing

martinconic:
### Checklist

### Description
Resolves [#4884](https://github.com/ethersphere/bee/issues/4884)

This PR addresses extreme LevelDB I/O bottlenecks and CPU serialization overhead encountered during massive high-frequency chunk uploads. 

### The Problem
Under a load of hundreds of thousands of concurrent chunks, the node suffered from performance degradation across two distinct vectors:
1. **LevelDB Put Lock Saturation**: Writing a stamp item for every chunk requires locking LevelDB. Under 1M concurrent pushsyncs, this sequentially locked the disk queue, choking disk writes up to >100MB/s and stalling the networking routines for **>140 seconds** cumulatively on CPU profiles.
2. **JSON Deserialization Overhead**: Updating peer and surplus balances required fetching extensive math objects from the statestore. The unoptimized branch utilized JSON unmarshal to decode these values from ASCII text. Under heavy load, the CPU burned **>20 seconds** of pure flat execution time just parsing strings.
### The Solution
1. **StamperStoreCache (Debounced Async Writes)**
   - Wrapped the Stamper Store in a highly concurrent memory cache.
   - Network routines now instantly write payloads to a thread-safe map in RAM.
   - A background flush loop sweeps the map every 5 seconds, writing thousands of items into LevelDB in a single, atomic batch. Disk churn drops massively, and the master process is completely freed from I/O wait times.
2. **Binary BigInt Encoding**
   - Implemented binary marshaling natively onto the central BigInt struct.
   - Refactored all accounting database getters/setters to utilize raw binary memory allocations instead of strings.
   - Included a silent fallback mechanism: the binary unmarshaler seamlessly parses older JSON-encoded database values so upgrades require **no explicit database migration**.

### DIFF
--- a/pkg/postage/service.go
+++ b/pkg/postage/service.go
@@ -116,6 +116,8 @@ func (s *service) recoverBuckets() error {
 				if bytes.Equal(issuer.data.BatchID, item.BatchID) {
 					if err := issuer.recover(item.BatchIndex); err != nil {
 						s.logger.Error(err, "postage recovery of bucket count failed")
+					} else {
+						issuer.dirty = true
 					}
 					break
 				}
--- a/pkg/postage/service_test.go
+++ b/pkg/postage/service_test.go
@@ -372,7 +372,6 @@ func TestCrashRecovery(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer ps2.Close()
 
 	issuers := ps2.StampIssuers()
 	if len(issuers) != 1 {
@@ -385,5 +384,31 @@ func TestCrashRecovery(t *testing.T) {
 	}
 	if buckets[bIdx1] != 1 {
 		t.Errorf("bucket %d: want 1, got %d", bIdx1, buckets[bIdx1])
+	}
+
+	// Clean shutdown — recovered counts must be flushed to disk.
+	if err := ps2.Close(); err != nil {
+		t.Fatal(err)
+	}
+
+	// Restart with wasClean=true — recovery is skipped, but counts must still
+	// be correct because ps2.Close() persisted the recovered state.
+	ps3, err := postage.NewService(log.Noop, store, pstore, 1, true)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer ps3.Close()
+
+	issuers3 := ps3.StampIssuers()
+	if len(issuers3) != 1 {
+		t.Fatalf("post-recovery clean restart: expected 1 issuer, got %d", len(issuers3))
+	}
+
+	buckets3 := issuers3[0].Buckets()
+	if buckets3[bIdx0] != 3 {
+		t.Errorf("post-recovery clean restart: bucket %d: want 3, got %d", bIdx0, buckets3[bIdx0])
+	}
+	if buckets3[bIdx1] != 1 {
+		t.Errorf("post-recovery clean restart: bucket %d: want 1, got %d", bIdx1, buckets3[bIdx1])
 	}
 }


acud:
@martinconic thanks for this. I have some comments:
1. this PR states it is about stamperstore disk usage, but modifies also the accounting package. those are different problems and it would be good to be a bit more meticulous about the scope of PRs
2. there is no need to reinvent serialization for the `big.Int` package. it already supports multiple formats out of the box, please see the [standard library documentation](https://pkg.go.dev/math/big@go1.26.1#Int)
3. there is a shutdown logic added to the stamp persistence store. i am not sure whether this is needed on a package level or if we could have this on the application level. if we have components implementing this themselves we already are starting to duplicate logic. it is something that the application should provide. see the `.DIRTY` file as it touches this topic too. perhaps something similar is needed (or reusing of the same thing).
4. bucket recovery logs in `Debug` (why not `Error`?) and swallows the error, the actual meaning of this is the future uploads of the user will override previously uploaded data. this is pretty severe.
5. persisting the stamp buckets (in the loop with the ticket) emits warning, not error
6. closing of the postage service reads off the channel it is closing, which should return immediately, not sure why it is needed

pkg/node/node.go
```diff
@@ -660,7 +661,21 @@ func NewBee(
 	b.p2pService = p2ps
 	b.p2pHalter = p2ps
 
-	post, err := postage.NewService(logger, stamperStore, batchStore, chainID)
+	dirtyItem := &stamperDirtyItem{}
```
acud: i'm not sure why this should be here. why can all this code live inside the stamper store init and close methods?

pkg/node/node.go
```diff
@@ -1430,6 +1445,11 @@ func (b *Bee) Shutdown() error {
 	tryClose(b.topologyCloser, "topology driver")
 	tryClose(b.storageIncetivesCloser, "storage incentives agent")
 	tryClose(b.stateStoreCloser, "statestore")
+	if b.stamperCleanShutdown != nil {
```
acud: ditto. `Close` is called in `L1453` anyway, why cant this be executed as part of that method as the last thing done? and in any case, if you really wanna keep it outside, this block should execute _after_ `L1453` no? since you still didn't close it, how do you know to say it was shutdown cleanly? also, even if it did you wouldn't know, because the `err` returned from `Close` is not available in this scope

pkg/postage/service.go
```diff
@@ -75,6 +87,69 @@ func NewService(logger log.Logger, store storage.Store, postageStore Storer, cha
...
+		storage.Query{
+			Factory: func() storage.Item { return new(StampItem) },
+		}, func(result storage.Result) (bool, error) {
+			item := result.Entry.(*StampItem)
+			for _, issuer := range s.issuers {
+				if bytes.Equal(issuer.data.BatchID, item.BatchID) {
+					if err := issuer.recover(item.BatchIndex); err != nil {
+						s.logger.Error(err, "postage recovery of bucket count failed")
+					}
+					issuer.setDirty(true)
```
acud: why is it dirty all of a sudden? it was just restored from disk, so it is not dirty per-se (since dirty means there are writes which aren't flushed yet)

acud (changes requested):

pkg/node/statestore.go
```diff
@@ -42,21 +42,59 @@ func InitStateStore(logger log.Logger, dataDir string, cacheCapacity uint64) (st
 	return stateStore, caching, err
 }
 
+// stamperDirtyItem is a storage.Item used to mark that the stamper store was
+// not cleanly shut down. It is written on startup and deleted on clean shutdown,
+// following the same pattern as the sharky .DIRTY file.
+type stamperDirtyItem struct{}
```
acud: i meant more that this can go straight into the place where you open the leveldb statestore. it can then be handed over to all dependencies, and you don't need to go through the hassle of namespacing the key or generating all of encapsulating type overhead needed in order to write and remove a single simple key.
in `pkg/storage/leveldbstore/store.go`:
```go
// New returns a new store the backed by leveldb.
// If path == "", the leveldb will run with in memory backend storage.
func New(path string, opts *opt.Options) (*Store, error) {
 ...
}
```

becomes something like:
```go

const dirtyKey = []byte(".store-dirty-shutdown")

// New returns a new store the backed by leveldb.
// If path == "", the leveldb will run with in memory backend storage.
func New(path string, opts *opt.Options) (*Store, bool, error) {
	var (
		err error
		db  *leveldb.DB
	)

	if path == "" {
		db, err = leveldb.Open(ldbStorage.NewMemStorage(), opts)
	} else {
		db, err = leveldb.OpenFile(path, opts)
		if ldbErrors.IsCorrupted(err) && !opts.GetReadOnly() {
			db, err = leveldb.RecoverFile(path, opts)
		}
	}
	if err != nil {
		return nil, false, err
	}

	dirty, err := db.Has(dirtyKey, &opt.ReadOptions{})
	if err != nil {
		return nil, false, fmt.Errof("has dirty record: %w", err)
	}

	err := db.Put(dirtyKey, []byte{}, &opt.WriteOptions{})
	if err != nil {
		return nil, false, fmt.Errof("put dirty record: %w", err)
	}

	return &Store{
		db:   db,
		path: path,
	}, dirty, nil
}
```
and the `Close` method gets amended accordingly:
```go
// Close implements the storage.Store interface.
func (s *Store) Close() (err error) {
	return errors.Join(
		s.db.Delete(dirtyKey),
		s.db.Close(),
	)
}
```

akrem-chabchoub (approved):

pkg/postage/service_test.go
```diff
@@ -298,3 +301,68 @@ func TestSetExpired(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := ps.Add(issuer); err != nil {
+		t.Fatal(err)
+	}
+	if err := ps.Close(); err != nil {
+		t.Fatal(err)
+	}
+
```
acud, line 348: can we add some checks here to see that the bucket counts are still not recovered? apart from the comment below stating that recovery `should be` triggered - there's no programmatic check that it hasn't happened before.

acud (approved):
Nice to have: missing a few assertions in the test

Issue #5388, opened 2026-03-09
RPC calls optimisation


sbackend123:
## Problem

Depending on the chain, the block time may differ.

We currently have several issues related to RPC calls:

* **Timeout is too short**, which causes the connection to drop and forces us to retry the RPC call.
* **Redundant RPC calls**: the same RPC requests are executed multiple times for the same transactions.

There may be additional issues as well, and this is currently being investigated.

## Notes

Since block time varies across chains, the current RPC behavior may be especially problematic on slower chains.

sbackend123:
Update: according to [dashboard](https://grafana.proxy.ethswarm.org/d/sb4522j/rpc-calls?folderUid=dffio3swnaxogb&orgId=1&from=now-1h&to=now&timezone=browser) most our RPC calls are **block_number** and **eth_logs** which are used in several components

PR #5394, opened 2026-03-12
feat: optimize rpc calls

sbackend123:
### Checklist

### Description
Removed  RPC patterns in the transaction flow. Added cache layer for BlockNumber rpc call.
Fixed lint issues in the new cache package and cleaned up minor lint findings in adjacent test/helper code.

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5388

Screenshot from 2026-03-16 13-29-22
 
 Highly likely hit ratio is low because of high load errors (EOF), which is fix by [PR](https://github.com/ethersphere/bee/pull/5341)


cmd/bee/cmd/deploy.go
```diff
@@ -60,6 +60,7 @@ func (c *command) initDeployCmd() error {
 				blocktime,
 				true,
 				c.config.GetUint64(optionNameMinimumGasTipCap),
+				blocktime-2,
```
gacevicljubisa: Just to note that current blocktime on our CI is 1s. So this way caching is not used there... How we could achive to use it there?

pkg/bmt/proof_test.go
```diff
@@ -26,7 +26,7 @@ func TestProofCorrectness(t *testing.T) {
 	verifySegments := func(t *testing.T, exp []string, found [][]byte) {
 		t.Helper()
 
-		var expSegments [][]byte
+		expSegments := make([][]byte, 0, len(exp))
```
gacevicljubisa, line 29: These are fine but feel like they belong in a separate PR.

pkg/node/chain.go
```diff
@@ -51,6 +51,7 @@ func InitChain(
 	pollingInterval time.Duration,
 	chainEnabled bool,
 	minimumGasTipCap uint64,
+	blockCacheTTl time.Duration,
```
gacevicljubisa: Typo blockCacheTTl -> blockCacheTTL

pkg/transaction/wrapped/cache/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+		Invalidates: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: m.Namespace,
+			Subsystem: subsystem,
+			Name:      prefix + "_invalidations",
+			Help:      prefix + " cache invalidations",
+		}),
+	}
+}
+
+func NewMetrics() Metrics {
```
gacevicljubisa: Seems not used (dead code).

pkg/transaction/wrapped/cache/cache.go
```diff
@@ -0,0 +1,107 @@
...
+	c.metrics.Misses.Inc()
+
+	result, err, shared := c.group.Do(string(c.key), func() (any, error) {
+		c.metrics.Loads.Inc()
+		val, err := loader()
+		if err != nil {
+			c.metrics.LoadErrors.Inc()
+			return val, err
+		}
+		c.Set(val, time.Now())
```
gacevicljubisa: Should we use here `now` from argument?

gacevicljubisa: What could happen if loader takes a while?

gacevicljubisa:
**Suggestion:** Instead of a fixed TTL (`blocktime*85/100`, `blocktime-2`) + `BlockNumber` RPC, use a single `HeaderByNumber(nil)` call periodically to get **both** the block number and timestamp, then extrapolate between syncs with zero RPC calls.

**How it works:**
- Every ~20 blocks, call `HeaderByNumber(nil)` (latest) — returns block number + timestamp in one RPC call. Store as anchor point.
- Between syncs, estimate the current block from pure math: `currentBlock = anchorBlock + (now - anchorTimestamp) / blockTime`. No RPC calls at all.
- Cache TTL: `anchorTimestamp + (currentBlock - anchorBlock + 1) * blockTime - now`.
?

pkg/transaction/wrapped/cache/cache.go
sbackend123: I do not think so because there is difference between time when we request value and time, when we set new value if we had to load it.

pkg/bmt/proof_test.go
sbackend123: Yes, but linter was failing so I decided to fix everything. Next time will create separate PR

cmd/bee/cmd/deploy.go
sbackend123: In deploy.go blockTime is hardcoded by itself. But if we are talking about CI, normally  cacheTTL is about 85% of block time, so even if it is 1 second, than we will get 850 ms

pkg/transaction/wrapped/cache/cache.go
```diff
@@ -0,0 +1,107 @@
+// Copyright 2025 The Swarm Authors. All rights reserved.
```
janos: 2026

pkg/transaction/wrapped/cache/cache_test.go
```diff
@@ -0,0 +1,189 @@
+// Copyright 2025 The Swarm Authors. All rights reserved.
```
janos: 2026

pkg/transaction/wrapped/cache/keys.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2025 The Swarm Authors. All rights reserved.
```
janos: 2026

pkg/transaction/wrapped/cache/metrics.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2025 The Swarm Authors. All rights reserved.
```
janos: 2026

pkg/transaction/wrapped/cache/cache.go
```diff
@@ -0,0 +1,107 @@
...
+type ExpiringSingleFlightCache[T any] struct {
+	ttl time.Duration
+
+	mu        sync.RWMutex
+	value     T
+	valid     bool
+	expiresAt time.Time
+
+	group   singleflight.Group
+	key     Key
```
janos: The key here is only used in metrics and in c.group.Do as the key, but that key is always the same as the singleflight group is the same and the key is the same for the same instance of the ExpiringSingleFlightCache type. I owuld suggest to remove the Key type and just to provide the metrics prefix string in the constructor NewExpiringSingleFlightCache.

pkg/transaction/wrapped/cache/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+	Invalidates prometheus.Counter
+}
+
+type Metrics struct {
+	BlockNumber metricSet
+	Unknown     metricSet
+}
+
+func newMetricSet(prefix string) metricSet {
+	subsystem := "eth_backend_cache"
```
janos, line 21: Maybe to append the prefix to subsystem once instead to prepend to every metric name. The end result (final metric name) should be the same.

pkg/transaction/wrapped/cache/cache_test.go
```diff
@@ -0,0 +1,189 @@
...
+			defer wg.Done()
+			results[idx], errs[idx] = c.GetOrLoad(time.Now(), func() (uint64, error) {
+				loadCount.Add(1)
+				<-gate
+				return 77, nil
+			})
+		}(i)
+	}
+
+	time.Sleep(50 * time.Millisecond)
```
janos: Could synctest be used here to avoid sleeping and potentially have a flaky test as synchronization not happen in 50ms?

pkg/transaction/wrapped/cache/cache.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (c *ExpiringSingleFlightCache[T]) GetOrLoad(now time.Time, loader Loader[T]) (T, error) {
+	if v, ok := c.Get(now); ok {
+		c.metrics.Hits.Inc()
+		return v, nil
+	}
+
+	c.metrics.Misses.Inc()
+
+	result, err, shared := c.group.Do(string(c.key), func() (any, error) {
```
janos: This group is always calling the same key, so it can be even the static string.

pkg/transaction/wrapped/wrapped.go
```diff
@@ -61,14 +69,17 @@ func (b *wrappedBackend) TransactionByHash(ctx context.Context, hash common.Hash
...
-	if err != nil {
-		b.metrics.TotalRPCErrors.Inc()
-		return 0, err
-	}
-	return blockNumber, nil
+	return b.blockNumberCache.GetOrLoad(time.Now(), func() (uint64, error) {
+		b.metrics.TotalRPCCalls.Inc()
+		b.metrics.BlockNumberCalls.Inc()
+
+		blockNumber, err := b.backend.BlockNumber(ctx)
```
janos: The problem with the golang.org/x/sync/singleflight is that it is not context.Context aware. The context that is passed from the first caller will influence the end result of all other callers of this function. Meaning, if the first caller cancels or times out on the context, all other callers will receive that error even if they did not cancel or the timeout for their call did not happen. This is why I've created the resenje.org/signleflight, that is context aware and will not terminate the execution of the function until all callers terminate their contexts. There are a few places where resenje.org/singleflight is used in bee, if you like, you can look at it and consider using if you see important for this case. Basically GetOrLoad would require to accept context, pass to signleflight.Do and the context from the callback would be used in b.backend.BlockNumber.

cmd/bee/cmd/deploy.go
janos: This is very strange, as block time is defined as `const blocktime = 15` on line 16, which is actually a nanosecond. As that is passed in poolingInterval, it looks to me that it is a very short time, I suppose that block time should be in minutes, not nanoseconds. Luckily, this is only for the deploy command. Bee has its own block time passed from options. I believe that line 16 in this file should look like this `const blocktime = 15 * time.Minute`, which is not related to this PR.

Also what happens if blocktime value is 1 (if someone changes the constant) and we have a negative duration? If the constant is changed to the much longer value like 15*time.Minute, literal 2 in here would reduce only two nanoseconds from that large number.

pkg/transaction/wrapped/cache/cache.go
```diff
@@ -0,0 +1,107 @@
...
+	return zero, false
+}
+
+func (c *ExpiringSingleFlightCache[T]) Set(value T, now time.Time) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.value = value
+	c.valid = true
+	c.expiresAt = now.Add(c.ttl)
```
janos: Ideally, for the block number caching, the expiration time should be the time when the next block number is expected and that is around the block time interval. This is to have as least as possible time when the cache is reporting the older block number comapred what is the current block number.

In this generalized caching implementation, it is hard to generalize such requirement as it is very specific to how the block number is increased (in some ~ period, that is configurable). So the ideal situation would be to calculate what would be the time of the next block number and to set expiresAt to that time, or a bit later. And to watch if the block number is actually increased.

In the current implementation, the worst case would be that the cache will keep the block number cache up to the block time (15 minutes, for example on some networks) when the actual block number is increased, if the value is set on the cache a moment before the new block number. On average, statistically in a very large sample of caching, that time would be half of the block time (7 minutes).

I am not sure if such precision is needed, and if the delay of the new block number of up to the whole block time is acceptable. This is just my observation.

gacevicljubisa:
I don't see how this cache meaningfully reduces RPC calls. The BlockNumber callers (postage listener, storage incentives, tx confirmation, API) run on independent staggered timers and rarely overlap within the same block window. With the TTL at 85% of block time, each caller almost always finds the cache expired by its next poll, resulting in a fresh RPC call anyway. The singleflight deduplication only helps when callers hit BlockNumber at the exact same instant, which is uncommon given the staggered schedules. On Gnosis mainnet the block time is only 5s, so the cache TTL would be ~4.25s — hardly enough to span multiple independent caller cycles.

If we do want to reduce BlockNumber RPC calls, a more effective approach would be to use a single HeaderByNumber(nil) call periodically (e.g. every ~20 blocks, or more) to get both the block number and timestamp, then extrapolate between syncs: currentBlock = anchorBlock + (now - anchorTimestamp) / blockTime (mentioned [here](https://github.com/ethersphere/bee/pull/5394#issuecomment-4067744700)). 

The filterPendingTransactions refactor (eliminating redundant TransactionByHash calls in nextNonce) is a genuine improvement and worth keeping.

@janos wdyt?

janos:
> I don't see how this cache meaningfully reduces RPC calls. The BlockNumber callers (postage listener, storage incentives, tx confirmation, API) run on independent staggered timers and rarely overlap within the same block window. With the TTL at 85% of block time, each caller almost always finds the cache expired by its next poll, resulting in a fresh RPC call anyway. The singleflight deduplication only helps when callers hit BlockNumber at the exact same instant, which is uncommon given the staggered schedules. On Gnosis mainnet the block time is only 5s, so the cache TTL would be ~4.25s — hardly enough to span multiple independent caller cycles.
> 
> If we do want to reduce BlockNumber RPC calls, a more effective approach would be to use a single HeaderByNumber(nil) call periodically (e.g. every ~20 blocks, or more) to get both the block number and timestamp, then extrapolate between syncs: currentBlock = anchorBlock + (now - anchorTimestamp) / blockTime (mentioned [here](https://github.com/ethersphere/bee/pull/5394#issuecomment-4067744700)).
> 
> The filterPendingTransactions refactor (eliminating redundant TransactionByHash calls in nextNonce) is a genuine improvement and worth keeping.
> 
> @janos wdyt?

Yes, given that block numbers are changing very frequently on 5 to 12s depending on the network (gnosis and sepolia), but not exactly the same period every time. It is a very good suggestion to calculate block number as you described to reduce even more the frequency of rpc call to get the block number and estimate the block time by HeaderByNumber. In that case even signleflight is not needed, as internally, block numbers will always be returned by its specific cache.

I would even go further and use the block time value that is calculated from the HeaderByNumber instead specifying it using options statically.

The consequence could be that it is required to get the block number as the node startup as both block number and block time are needed as known values.  Maybe that is even good to do, and to exit the application in case that there are problems with the rpc endpoint when getting the block number on start. Just thinking.

sbackend123:
> **Suggestion:** Instead of a fixed TTL (`blocktime*85/100`, `blocktime-2`) + `BlockNumber` RPC, use a single `HeaderByNumber(nil)` call periodically to get **both** the block number and timestamp, then extrapolate between syncs with zero RPC calls.
> 
> **How it works:**
> 
>     * Every ~20 blocks, call `HeaderByNumber(nil)` (latest) — returns block number + timestamp in one RPC call. Store as anchor point.
> 
>     * Between syncs, estimate the current block from pure math: `currentBlock = anchorBlock + (now - anchorTimestamp) / blockTime`. No RPC calls at all.
> 
>     * Cache TTL: `anchorTimestamp + (currentBlock - anchorBlock + 1) * blockTime - now`.
>       ?

For me it sounds like overengineering a little bit from one side, and not so flexible implementations from another side (may be I miss smth?). I thought about whatif would like to cache another type of value? Than 

> > I don't see how this cache meaningfully reduces RPC calls. The BlockNumber callers (postage listener, storage incentives, tx confirmation, API) run on independent staggered timers and rarely overlap within the same block window. With the TTL at 85% of block time, each caller almost always finds the cache expired by its next poll, resulting in a fresh RPC call anyway. The singleflight deduplication only helps when callers hit BlockNumber at the exact same instant, which is uncommon given the staggered schedules. On Gnosis mainnet the block time is only 5s, so the cache TTL would be ~4.25s — hardly enough to span multiple independent caller cycles.
> > If we do want to reduce BlockNumber RPC calls, a more effective approach would be to use a single HeaderByNumber(nil) call periodically (e.g. every ~20 blocks, or more) to get both the block number and timestamp, then extrapolate between syncs: currentBlock = anchorBlock + (now - anchorTimestamp) / blockTime (mentioned [here](https://github.com/ethersphere/bee/pull/5394#issuecomment-4067744700)).
> > The filterPendingTransactions refactor (eliminating redundant TransactionByHash calls in nextNonce) is a genuine improvement and worth keeping.
> > @janos wdyt?
> 
> Yes, given that block numbers are changing very frequently on 5 to 12s depending on the network (gnosis and sepolia), but not exactly the same period every time. It is a very good suggestion to calculate block number as you described to reduce even more the frequency of rpc call to get the block number and estimate the block time by HeaderByNumber. In that case even signleflight is not needed, as internally, block numbers will always be returned by its specific cache.
> 
> I would even go further and use the block time value that is calculated from the HeaderByNumber instead specifying it using options statically.
> 
> The consequence could be that it is required to get the block number as the node startup as both block number and block time are needed as known values. Maybe that is even good to do, and to exit the application in case that there are problems with the rpc endpoint when getting the block number on start. Just thinking.

I have couple of concerns:
1. Drift over time: block production is not perfectly uniform, so we can accumulate error between syncs, especially if the resync interval is relatively large.

2. Risk of overshooting: extrapolation may produce a block number that has not actually been produced yet
 
3. With this approach we resolve problem with block_number call only. If in future we would need to reduce some other rpc calls, we might need to add smth else. 

I think, there can be other corner cases, which would be more difficult to catch, since  debugging will become more complex.

gacevicljubisa:
1. Drift over time
 - Block time variance on Gnosis is small (typically <1s around the 5s average). With a resync every ~20 blocks (~100s), the maximum accumulated error is ~1 block. Using floor() in the extrapolation ensures you never jump ahead. 
 - You can also calculate average block time dynamically on each resync
2. Risk of overshooting
 - `floor((now - anchorTimestamp) / blockTime)` naturally rounds down, so you'll be at most 1 block behind
3. Only solves block_number 
 - true, but also, using only caching layer, I am not sure how much can we achive with other rpc calls

pkg/storer/reserve_test.go
```diff
@@ -543,6 +543,7 @@ func TestSubscribeBin(t *testing.T) {
 			chunks             = make([]swarm.Chunk, 0, int(chunksPerPO)*2)
 			putter             = storer.ReservePutter()
 		)
+		chunks = make([]swarm.Chunk, 0, chunksPerPO*2)
```
gacevicljubisa: it is a duplicate,there is already allocation on line 543

pkg/transaction/wrapped/cache/metrics.go
```diff
@@ -0,0 +1,75 @@
...
+
+type metricSet struct {
+	Hits        prometheus.Counter
+	Misses      prometheus.Counter
+	Loads       prometheus.Counter
+	SharedLoads prometheus.Counter
+	LoadErrors  prometheus.Counter
+}
+
+type Metrics struct {
```
gacevicljubisa: Dead code, both Metrics struct and Collectors() method remain in the code but are never referenced.

pkg/transaction/wrapped/wrapped.go
```diff
@@ -61,14 +84,66 @@ func (b *wrappedBackend) TransactionByHash(ctx context.Context, hash common.Hash
...
+	anchor, err := b.blockNumberCache.PeekOrLoad(
+		ctx,
+		now,
+		func(anchor blockNumberAnchor, expiresAt, now time.Time) (bool, time.Time) {
+			if now.Before(expiresAt) {
+				return true, expiresAt
+			}
+
+			_, elapsedBlocks := b.estimatedBlockNumberWithElapsed(anchor, now)
+			if elapsedBlocks <= b.blockSyncInterval {
```
gacevicljubisa: Should this be < instead?

go.mod
```diff
@@ -58,6 +58,8 @@ require (
 	resenje.org/web v0.4.3
 )
 
+require github.com/kylelemons/godebug v1.1.0 // indirect
```
gacevicljubisa: Why is this added?

pkg/transaction/transaction.go
```diff
@@ -392,19 +392,15 @@ func (t *transactionService) nextNonce(ctx context.Context) (uint64, error) {
...
 
 	// PendingNonceAt returns the nonce we should use, but we will
 	// compare this to our pending tx list, therefore the -1.
 	maxNonce := onchainNonce - 1
-	for _, txHash := range pendingTxs {
-		trx, _, err := t.backend.TransactionByHash(ctx, txHash)
-		if err != nil {
-			t.logger.Error(err, "pending transaction not found", "tx", txHash)
-			return 0, err
+	for _, trx := range pending {
```
gacevicljubisa: Maybe add a warning here:

```suggestion
	for txHash, trx := range pending {
		if trx == nil {
			t.logger.Warning("pending transaction data unavailable, relying on onchain nonce", "tx", txHash)
			continue
		}
		maxNonce = max(maxNonce, trx.Nonce())
	}
```

pkg/storer/reserve_test.go
acud: that's also why the linter is complaining

pkg/transaction/wrapped/wrapped.go
```diff
@@ -8,29 +8,52 @@ import (
...
 
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethersphere/bee/v2/pkg/transaction"
 	"github.com/ethersphere/bee/v2/pkg/transaction/backend"
+	"github.com/ethersphere/bee/v2/pkg/transaction/wrapped/cache"
 )
 
+const defaultBlockSyncInterval = 10
```
acud: if the value is always this one, i think it makes more sense to have the CLI flag default to be 10 and get rid of this const. it just complicates the configuration and adjustment (and gets rid of the necessity of using a sentinel value)

pkg/transaction/wrapped/cache/cache.go
```diff
@@ -0,0 +1,103 @@
...
+		c.metrics.SharedLoads,
+		c.metrics.LoadErrors,
+	}
+}
+
+func (c *ExpiringSingleFlightCache[T]) Peek() (T, time.Time, bool) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	if c.valid {
```
acud: if this generalized cache object is already aware of the `expiresAt` time, i think that it can already be "smart enough" to know something has expired already. you can already determine here whether `valid` is still valid, and in fact you can do away with the field on the type altogether and just check whether `now() >= expiresAt`, since any subsequent usage of this value has to also compare `now` to the `expiresAt` time (otherwise you're blindly using that value).

also a note that generics increases the complexity here with no obvious place where it can be reused.

pkg/transaction/wrapped/cache/cache.go
```diff
@@ -0,0 +1,101 @@
...
+	return []prometheus.Collector{
+		c.metrics.Hits,
+		c.metrics.Misses,
+		c.metrics.Loads,
+		c.metrics.SharedLoads,
+		c.metrics.LoadErrors,
+	}
+}
+
+func (c *ExpiringSingleFlightCache[T]) Peek() (T, time.Time, bool) {
```
acud: this method does not need to be exported apart from for unit tests. can we not have it tested by just checking the callbacks?

pkg/transaction/wrapped/cache/cache.go
```diff
@@ -0,0 +1,101 @@
...
+func (c *ExpiringSingleFlightCache[T]) Set(value T, expiresAt time.Time) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.value = value
+	c.expiresAt = expiresAt
+}
+
+func (c *ExpiringSingleFlightCache[T]) PeekOrLoad(ctx context.Context, now time.Time, canReuse ReuseEvaluator[T], loader Loader[T]) (T, error) {
+	if value, expiresAt, ok := c.Peek(); ok {
```
acud: i am not 100% sure that `Peek` is needed. since the cache is expiring, and it already has a notion of expiration time, then why do you need to use both peer _and_ pass `now` (passed by the caller) back _to_ the caller using the data they have already provided just in order to check whether `now() >= expirationTime`?

would it not be easier to inline the mutex locking and the `now>expirationTime` check?

pkg/transaction/wrapped/cache/cache.go
sbackend123: Yes, we can

pkg/transaction/wrapped/cache/cache.go
sbackend123: Maybe I’m going too much into the details, but PeekOrLoad is already not the most trivial function, and I would keep Peek at least to avoid expanding PeekOrLoad(). Especially since there is also Set, which works with the mutex as well, so we end up with three methods, each responsible for a specific task. Does it makes sense for you?

acud: Yes I agree it is not trivial and there's already a bit too much indirection for what it is supposed to do. I just find it odd that the caller passes `now` to `PeekOrLoad`... just to end up getting it back in the callback (plus shadowing) and do the relevance check in the callback with:
```go
			if now.Before(expiresAt) {
				return true, expiresAt
			}
```
So why pass `now` at all? you can have it cleaner with just passing `c.value` and `c.expiresAt` into the callback and tell the callback to do the check and call it a day. In both cases `Peek` is not needed.

sbackend123: Ok, simplified

pkg/transaction/wrapped/wrapped.go
```diff
@@ -61,14 +79,65 @@ func (b *wrappedBackend) TransactionByHash(ctx context.Context, hash common.Hash
...
+			if header == nil || header.Number == nil {
+				b.metrics.TotalRPCErrors.Inc()
+				return blockNumberAnchor{}, time.Time{}, errors.New("latest block header unavailable")
+			}
+
+			anchor := blockNumberAnchor{
+				number:    header.Number.Uint64(),
+				timestamp: time.Unix(int64(header.Time), 0).UTC(),
+			}
+			return anchor, b.nextExpectedBlockTime(anchor, 0), nil
```
acud: i'm not sure i understand this very well. intuitively i'm interpreting this as `this fetched block value will expire in the next upcoming block tick`, but the whole change is about "caching" (or trusting the guessing of block numbers) in chunks (i.e. expiration date is more like a `next sync` date).
also, the passing of `0` here as a time value isn't clear - why?

pkg/transaction/wrapped/wrapped.go
```diff
@@ -8,29 +8,47 @@ import (
...
+) transaction.Backend {
 	return &wrappedBackend{
-		backend:          backend,
-		minimumGasTipCap: int64(minimumGasTipCap),
-		metrics:          newMetrics(),
+		backend:           backend,
+		minimumGasTipCap:  int64(minimumGasTipCap),
+		blockTime:         blockTime,
+		metrics:           newMetrics(),
+		blockSyncInterval: blockSyncInterval,
```
gacevicljubisa, line 54: We should prevent `blockSyncInterval` to be 0. Maybe to set it to 1 in that case?
In pkg/transaction/wrapped/wrapped.go#L135 division by zero can happen

sbackend123: Fixed

pkg/transaction/wrapped/wrapped.go
sbackend123: After brief discussion decided to remove expiresAt, so nextExpectedBlockTime is not needed anymore and we can simplify implementation

cmd/bee/cmd/cmd.go
```diff
@@ -61,6 +61,7 @@ const (
 	optionNameRedistributionAddress        = "redistribution-address"
 	optionNameStakingAddress               = "staking-address"
 	optionNameBlockTime                    = "block-time"
+	optionNameBlockSyncInterval            = "block-sync-interval"
```
akrem-chabchoub, line 64: Please update the `packaging` folder with the new flag (in `docker-compose`,...)

PR #5406, opened 2026-03-20
chore(p2p): prefer tcp when discovering bootnodes

acud:
### Checklist

### Description
This PR introduces a prioritization of TCP underlays over other types of underlays when connecting to bootnodes.

### AI Disclosure

### DIFF
--- a/pkg/topology/kademlia/kademlia.go
+++ b/pkg/topology/kademlia/kademlia.go
@@ -829,9 +829,13 @@ func (k *Kad) connectBootNodes(ctx context.Context) {
 			k.metrics.TotalBootNodesConnectionAttempts.Inc()
 
 			if err != nil {
+				if errors.Is(err, p2p.ErrUnsupportedAddresses) {
+					k.logger.Debug("bootnode address transport not supported, skipping", "bootnode_address", addr)
+					return false, nil
+				}
 				if !errors.Is(err, p2p.ErrAlreadyConnected) {
 					k.logger.Error(err, "connect to bootnode failed", "bootnode_address", addr)
-					return false, err
+					return false, nil
 				}
 				k.logger.Debug("bootnode already connected", "bootnode_address", addr)
 				return false, nil


pkg/p2p/discover.go
```diff
@@ -9,11 +9,31 @@ import (
...
+		case ma.P_TLS:
+			return false
+		}
+	}
+	return hasTCP
+}
+
+func sortAddrsByTCPPreference(addrs []ma.Multiaddr) {
+	sort.SliceStable(addrs, func(i, j int) bool {
+		return hasTCPProtocol(addrs[i]) && !hasTCPProtocol(addrs[j])
```
gacevicljubisa: Suggestion to reuse existing function `ClassifyTransport` here:

```go
func sortAddrsByTCPPreference(addrs []ma.Multiaddr) {
	sort.SliceStable(addrs, func(i, j int) bool {
		return bzz.ClassifyTransport(addrs[i]).Priority() < bzz.ClassifyTransport(addrs[j]).Priority()
	})
}
```

gacevicljubisa:
If I got it correctly, if bootnode has multiple underlays with TCP and WSS, and TCP for some reason is not connectable, currently it will never try other WSS address (because of the priority). This is beacuse in kademlia.go:[832-835](https://github.com/ethersphere/bee/blob/148f520a730598894e5fd983836bd67520e675e4/pkg/topology/kademlia/kademlia.go#L832-L835) - the callback returns (false, err) on any Connect failure (except ErrAlreadyConnected), and Discover propagates that error, aborting the loop. Would it be better that we skip and continue with next underlay?

acud:
Thanks @gacevicljubisa. Not sure what do you mean with this assumption. My understanding is that the bool in the callback return signature is `stop`, not `continue`, so returning `false` is fine and would continue to iterate over addresses. Please double check me on this. My understanding is that only returning `true` stops the loop, which happens in two places (either too many attempts or 3 bootnodes connected):
```go
		if _, err := p2p.Discover(ctx, addr, func(addr ma.Multiaddr) (stop bool, err error) {
			loggerV1.Debug("connecting to bootnode", "bootnode_address", addr)
			if attempts >= maxBootNodeAttempts {
				return true, nil
			}
.....


			// connect to max 3 bootnodes
			return connected >= 3, nil
		});
```

gacevicljubisa:
> Thanks @gacevicljubisa. Not sure what do you mean with this assumption. My understanding is that the bool in the callback return signature is `stop`, not `continue`, so returning `false` is fine and would continue to iterate over addresses. Please double check me on this. My understanding is that only returning `true` stops the loop, which happens in two places (either too many attempts or 3 bootnodes connected):
> 
> ```go
> 		if _, err := p2p.Discover(ctx, addr, func(addr ma.Multiaddr) (stop bool, err error) {
> 			loggerV1.Debug("connecting to bootnode", "bootnode_address", addr)
> 			if attempts >= maxBootNodeAttempts {
> 				return true, nil
> 			}
> .....
> 
> 
> 			// connect to max 3 bootnodes
> 			return connected >= 3, nil
> 		});
> ```

When the callback returns (false, err) with a non-nil error, Discover propagates that error immediately and aborts the loop. The remaining addresses are never tried. Look at discover.go:[57-59](https://github.com/ethersphere/bee/blob/148f520a730598894e5fd983836bd67520e675e4/pkg/p2p/discover.go#L57-L59)

I would suggest after Connect in kademlia.go:[826](https://github.com/ethersphere/bee/blob/148f520a730598894e5fd983836bd67520e675e4/pkg/topology/kademlia/kademlia.go#L826) we check for `p2p.ErrUnsupportedAddresses`:

```go
	if errors.Is(err, p2p.ErrUnsupportedAddresses) {
		k.logger.Debug("bootnode address transport not supported, skipping", "bootnode_address", addr)
		return false, nil
	}
```

And also, not to propage error from Connect at all in kademlia.go:[831-838](https://github.com/ethersphere/bee/blob/148f520a730598894e5fd983836bd67520e675e4/pkg/topology/kademlia/kademlia.go#L831-L838)

```go
	if err != nil {
		if errors.Is(err, p2p.ErrAlreadyConnected) {
			k.logger.Debug("bootnode already connected", "bootnode_address", addr)
			return false, nil
		}
		k.logger.Error(err, "connect to bootnode failed", "bootnode_address", addr)
		return false, nil
	}
```

acud:
nice thanks @gacevicljubisa I didn't catch that. pushed a commit to fix if you can have a look.

janos (approved):

Issue #5397, opened 2026-03-13
Node shutdown cleanups


acud:
Currently, the bee node logs a lot of information while shutting down. This is apparently pullsync related, although there are other packages involved. Also, the shutdown takes relatively a lot of time in some cases. The main problem with this is that in the shutdown mechanism there are code sections with individual timeouts which can cause dirty shutdowns on some components which is not ideal.

The scope of the task is to sweep through the shutdown sequence in the `node.go` file and make sure it is sound:
- verify all of the inter-dependencies and make sure that the shutdown happens in the right order
- verify the context cancellation is correct
- find out whether internal timeouts are triggered and if so - why?
- handle here is the excessive logging that happens on shutdown - in some areas we might want to catch specifically context cancellation errors and not log them. Instead, we could use prometheus metrics to keep track of context cancellations if that is needed (or specifically check whether we are shutting down actively in order to know whether to log things or not).

AC:
- node shuts down quickly without waiting for more messages
- no internal timeouts occur in the shutdown process
- shutdown occurs within a graceful kubernetes pod shutdown time-window
- test this on a node connected to the testnet and mainnet

PR #5408, opened 2026-03-23
fix: speed up node shutdown

sbackend123:
Fixed order of closing components. Removed unnecessary logging. Made Close() methods safer. Fixed context in pull sync component.

### Checklist

### Description
**Context:** see the linked issue

**What was done:**

* Fixed the shutdown order of storages. Previously, `StateStore` was closed before `LocalStore`. `LocalStore` uses `BatchStore`, which depends on `StateStore`. During shutdown or in-flight work, `BatchStore` could access `StateStore`, causing failed `Put/Get` operations, with a risk of incomplete flush and inconsistent local state.

* Fixed a context handling issue in the pull sync component. Upon receiving a shutdown signal, the component waited 10 seconds to allow its sub-goroutines to terminate gracefully. However, the sub-goroutines did not receive the context cancellation signal.

* Added logs to track how long each component takes to shut down.

* Removed warning logs in `libp2p` for context cancellation when the node is shutting down, to avoid log noise.

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5397

### AI Disclosure

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -1360,7 +1360,9 @@ func (b *Bee) Shutdown() error {
 
 		start := time.Now()
 		b.logger.Debug("starting shutdown", "component", component)
-		defer b.logger.Debug("finished shutdown", "component", component, "elapsed", time.Since(start))
+		defer func() {
+			b.logger.Debug("finished shutdown", "component", component, "elapsed", time.Since(start))
+		}()
 		if err := c.Close(); err != nil {
 			mErr = multierror.Append(mErr, fmt.Errorf("%s: %w", component, err))
 		}


pkg/api/api.go
```diff
@@ -174,8 +174,9 @@ type Service struct {
 
 	metrics metrics
 
-	wsWg sync.WaitGroup // wait for all websockets to close on exit
-	quit chan struct{}
+	wsWg      sync.WaitGroup // wait for all websockets to close on exit
+	closeOnce sync.Once
```
acud: why is this needed? how come that a quit signal can happen twice?

pkg/p2p/libp2p/internal/reacher/reacher.go
```diff
@@ -43,8 +43,9 @@ type reacher struct {
 	peerHeap  peerHeap         // min-heap ordered by retryAfter
 	peerIndex map[string]*peer // lookup by overlay for O(1) access
 
-	newPeer chan struct{}
-	quit    chan struct{}
+	newPeer   chan struct{}
+	quit      chan struct{}
+	closeOnce sync.Once
```
acud: ditto

pkg/api/api.go
janos: I believe that this is the protection of the double closed channel, which is panicing.

acud: yeah that is clear. the question is why would the channel be closed twice in the first place? it suggests a problem in the calling code no?

janos: I suppose that it is just a common protection, as there is no way to restrict how many times an exported method is called. I did not notice that the Close method is called twice anywhere, but I suppose that a layer of safety is a good thing here.

sbackend123:
> Good hardening of the close functions. I did not like to write so many similar code for service shutdown or closing and wanted to have even the blocking mechanism, to wait until all the goroutines are closed before the Close() function returns (for consistency and reliability), and I have wrote something like this https://pkg.go.dev/resenje.org/x/shutdown#Graceful. If you think that something could be stolen from there, or to make something better and unified for bee, feel free to investigate, or use.



> Good hardening of the close functions. I did not like to write so many similar code for service shutdown or closing and wanted to have even the blocking mechanism, to wait until all the goroutines are closed before the Close() function returns (for consistency and reliability), and I have wrote something like this https://pkg.go.dev/resenje.org/x/shutdown#Graceful. If you think that something could be stolen from there, or to make something better and unified for bee, feel free to investigate, or use.

After discussion with @acud  decided to remove sync.Once()

janos:
> After discussion with @acud decided to remove sync.Once()

Cool, it is good that there has been a discussion.

pkg/node/node.go
```diff
@@ -1350,12 +1352,15 @@ func (b *Bee) Shutdown() error {
...
 	// tryClose is a convenient closure which decrease
 	// repetitive io.Closer tryClose procedure.
-	tryClose := func(c io.Closer, errMsg string) {
+	tryClose := func(c io.Closer, component string) {
 		if c == nil {
 			return
 		}
+
+		b.logger.Debug("starting shutdown", "component", component)
+		defer b.logger.Debug("finished shutdown", "component", component)
```
acud: maybe it is worth to add to it the time-span it took to shut down the component

acud:
@sbackend123 there are merge conflicts preventing the CI from running

PR #5409, opened 2026-03-24
fix: sharky corruption

### LINKED ISSUE #4737: sharky corruption after host crash
### Context
v2.1.0 on sepolia testnet swarm

### Summary
I had 13 nodes running on a single host, 2 on an internal NVME SSD and the other 11 on a USB-3-connected SATA SSD.  Due to reasons outside of swarm, this host crashed without warning.   On powering up, I discovered that all 11 nodes on the external SSD had sharky corruptions as indicated by `db validate`.   The 2 nodes on the internal NVME SSD validated cleanly.

It is also worth noting that this host crashed while it was very active pushing a new OSM hybrid-redundancy dataset into the testnet swarm.   The nodes were definitely NOT idle when the crash happened.

### Expected behavior
I would hope that recovery from a system crash would not cause a corrupted sharky.  But I can almost understand it with a USB-connected drive.

### Actual behavior
```
"time"="2024-07-26 13:22:49.541998" "level"="info" "logger"="node" "msg"="validation finished" "duration"="40m1.635440786s" "invalid"=2 "soc"=1319448 "total"=2137804
"time"="2024-07-26 13:23:33.389734" "level"="info" "logger"="node" "msg"="validation finished" "duration"="40m29.69367306s" "invalid"=41 "soc"=1321454 "total"=2140554
"time"="2024-07-26 13:30:12.261303" "level"="info" "logger"="node" "msg"="validation finished" "duration"="49m38.373076047s" "invalid"=228 "soc"=4198475 "total"=6213034
"time"="2024-07-26 13:24:33.398450" "level"="info" "logger"="node" "msg"="validation finished" "duration"="41m12.410722607s" "invalid"=148 "soc"=1336141 "total"=2159790
"time"="2024-07-26 13:24:23.690402" "level"="info" "logger"="node" "msg"="validation finished" "duration"="40m45.640637335s" "invalid"=22 "soc"=1329395 "total"=2150760
"time"="2024-07-26 13:24:28.904050" "level"="info" "logger"="node" "msg"="validation finished" "duration"="40m36.74342581s" "invalid"=47 "soc"=1319130 "total"=2137422
"time"="2024-07-26 13:24:36.198084" "level"="info" "logger"="node" "msg"="validation finished" "duration"="40m24.78211206s" "invalid"=31 "soc"=1319296 "total"=2137618
"time"="2024-07-26 13:24:15.713920" "level"="info" "logger"="node" "msg"="validation finished" "duration"="39m40.56859582s" "invalid"=8 "soc"=1319289 "total"=2137613
"time"="2024-07-26 13:20:43.081423" "level"="info" "logger"="node" "msg"="validation finished" "duration"="35m53.235082068s" "invalid"=88 "soc"=1325899 "total"=2146762
"time"="2024-07-26 13:24:59.632308" "level"="info" "logger"="node" "msg"="validation finished" "duration"="40m5.682027361s" "invalid"=10 "soc"=1329492 "total"=2151330
"time"="2024-07-26 13:25:11.043347" "level"="info" "logger"="node" "msg"="validation finished" "duration"="40m12.737092877s" "invalid"=5 "soc"=1366998 "total"=2201914
```
### Steps to reproduce
I guess if you have a very active node and crash the host with the data-dir on an external USB drive, you may be able to duplicate the corruption.

### Possible solution
Don't run with data-dir on an external drive?

At least it may be worth noting somewhere in the documentation that running on external drives may be at a risk for data corruption.

martinconic:
### Checklist

### Description
**Fixes #4737**

## Problem

When a node crashes or loses power while actively storing chunks (e.g. during a file upload via pushsync), two corruption scenarios can occur:

**Scenario 1 — slots bitmap wiped on crash.**
`slots.save()` previously called `Truncate(0)` before seeking back to position 0 and rewriting the bitmap. If the node crashed between the truncate and the rewrite, the slots file was left empty. On next startup, Sharky would treat all slots as free and begin overwriting chunks that were still referenced by the LevelDB index.

**Scenario 2 — shard data lost from OS page cache.**
Sharky writes chunk data to a shard file via the OS page cache. LevelDB, which has its own write-ahead log, commits the chunk index entry durably. If the node crashes before the OS flushes the page cache to disk, the LevelDB index points to Sharky slots that contain stale or zeroed data. The node then serves corrupted chunks.

This was reproducible specifically when nodes were actively uploading under load, consistent with the reports in the issue. Idle nodes were unaffected because no new writes were in-flight.

## Solution

**Fix 1 — remove `Truncate(0)` from `slots.save()`.**
Slots only ever grow (the only mutation is `extend`), so `sl.data` is always >= the previous file size. Seeking to 0 and overwriting is always safe — no stale tail bytes can survive. The truncate before write was unnecessary and introduced a crash window. Removing it closes that window with zero performance impact.

**Fix 2 — validate and prune corrupted chunks on recovery.**
The existing `.DIRTY` file mechanism already detects unclean shutdowns. On recovery, instead of simply rebuilding the slots bitmap from the index, each chunk's data is now read from Sharky and its content hash is validated (CAC and SOC). Valid chunks are registered in the recovery bitmap as before. Corrupted entries — those whose data is unreadable or whose hash does not match the indexed address — are removed from the LevelDB index so the node starts clean and does not serve invalid data.

Full cleanup per corrupted chunk includes:
- `RetrievalIndexItem` — the Sharky location index entry
- `BatchRadiusItem` — reserve bin/batch mapping
- `ChunkBinItem` — reserve bin iteration index
- `stampindex` entry — duplicate stamp prevention index
- `chunkstamp` entry — stamp lookup by address

Without removing all of these, the reserve size accounting would be overcounted and iteration over the reserve would hit `ErrNotFound` for pruned addresses.

If a corrupted index entry cannot be deleted, an error is returned and node startup is aborted to prevent operating on corrupt state.

### AI Disclosure

### DIFF
--- a/pkg/storer/recover.go
+++ b/pkg/storer/recover.go
@@ -110,11 +110,11 @@ func validateAndAddLocations(ctx context.Context, store storage.Store, sharkyRec
 	}
 
 	for _, item := range corrupted {
-		if err := errors.Join(
-			store.Delete(item),
-			reserve.DeleteCorruptedChunkMetadata(store, baseAddr, item.Address),
-		); err != nil {
-			return 0, fmt.Errorf("recovery: failed deleting corrupted chunk %s: %w", item.Address, err)
+		if err := reserve.DeleteCorruptedChunkMetadata(store, baseAddr, item.Address); err != nil {
+			return 0, fmt.Errorf("recovery: failed deleting corrupted chunk metadata %s: %w", item.Address, err)
+		}
+		if err := store.Delete(item); err != nil {
+			return 0, fmt.Errorf("recovery: failed deleting corrupted chunk index %s: %w", item.Address, err)
 		}
 	}
 


pkg/storer/recover.go
```diff
@@ -51,29 +54,61 @@ func sharkyRecovery(ctx context.Context, sharkyBasePath string, store storage.St
...
+		return err
 	}
 
 	if err := sharkyRecover.Save(); err != nil {
 		return err
 	}
 
+	for _, item := range corrupted {
+		if err := store.Delete(item); err != nil {
+			logger.Error(err, "recovery: failed deleting corrupted chunk index entry", "address", item.Address)
```
acud: this is a pretty severe error. i'm not sure if we shouldn't even panic after something like this because the possibility of data corruption is already very high. would be happy for other input, @janos wdyt?

pkg/storer/recover.go
```diff
@@ -51,29 +54,61 @@ func sharkyRecovery(ctx context.Context, sharkyBasePath string, store storage.St
...
+		}
+	}
+
+	if len(corrupted) > 0 {
+		logger.Info("recovery: removed corrupted chunk index entries", "count", len(corrupted))
```
acud: this should be a warning

pkg/storer/recover.go
janos: Yes, this is a pretty serious error and I think that it is ok to return error here. This function is indirectly called during the node startup and it is ok to nicely communicate the error with the storage to the user. It would be nicer then panic, as it will run in the main goroutine, so it will make the node stop. Since it is based on its usage, this behaviour should be part of the function comment.

pkg/storer/recover.go
```diff
@@ -51,29 +55,64 @@ func sharkyRecovery(ctx context.Context, sharkyBasePath string, store storage.St
...
+	var corrupted []*chunkstore.RetrievalIndexItem
+
+	err := chunkstore.IterateItems(store, func(item *chunkstore.RetrievalIndexItem) error {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		buf := make([]byte, item.Location.Length)
```
acud: it is not needed to allocate a new buffer for every chunk. you can preallocate once outside of the loop with an extra size (there's specific const we use for that, i think some type of soc with header size const)

pkg/storer/recover.go
```diff
@@ -51,29 +55,64 @@ func sharkyRecovery(ctx context.Context, sharkyBasePath string, store storage.St
...
-		if res.Err != nil {
-			return res.Err
+// validateAndAddLocations iterates every chunk index entry, reads its data from
+// Sharky, and validates the content hash. Valid chunks are registered with the
+// recovery so their slots are preserved. Corrupted entries (unreadable data or
+// hash mismatch) are logged, excluded from the recovery bitmap, and deleted from
+// the index store so the node starts clean without serving invalid data.
+// If a corrupted index entry cannot be deleted, an error is returned and the
+// node startup is aborted to prevent serving or operating on corrupt state.
+func validateAndAddLocations(ctx context.Context, store storage.Store, sharkyRecover *sharky.Recovery, logger log.Logger) error {
```
acud: how long does this take comparing to the previous one which did not validate? it seems that the recovery will not take significantly more resources and time. would be good to have an idea about this

martinconic: The additional cost is in the same order of magnitude as db repair and db validate, which already do the same work. Recovery only runs on unclean shutdown, so this is not a concern for normal operation.

pkg/storer/recover_test.go
```diff
@@ -0,0 +1,122 @@
+// Copyright 2024 The Swarm Authors. All rights reserved.
```
gacevicljubisa: update year

pkg/storer/recover.go
```diff
@@ -51,29 +55,65 @@ func sharkyRecovery(ctx context.Context, sharkyBasePath string, store storage.St
...
+	buf := make([]byte, swarm.SocMaxChunkSize)
+
+	err := chunkstore.IterateItems(store, func(item *chunkstore.RetrievalIndexItem) error {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		if err := sharkyRecover.Read(ctx, item.Location, buf[:item.Location.Length]); err != nil {
```
gacevicljubisa, line 89: Would be nice to have unit test for Read error path

pkg/storer/recover.go
```diff
@@ -51,29 +55,65 @@ func sharkyRecovery(ctx context.Context, sharkyBasePath string, store storage.St
...
 	}
 
+	for _, item := range corrupted {
+		if err := store.Delete(item); err != nil {
+			logger.Error(err, "recovery: failed deleting corrupted chunk index entry", "address", item.Address)
+			return fmt.Errorf("recovery: failed deleting corrupted chunk index entry %s: %w", item.Address, err)
+		}
+	}
+
+	if len(corrupted) > 0 {
```
gacevicljubisa, line 121: Should we also remove the orphaned reserve index entries for pruned chunks? After deleting the RetrievalIndexItem, these are left behind:

- BatchRadiusItem
- ChunkBinItem
- stampindex entry
- chunkstamp entry

Without cleanup, the reserve size accounting will be off and iteration will hit `ErrNotFound` for those addresses.

gacevicljubisa: Also, maybe emit a metric (gauge or counter) for the number of pruned entries?

acud: 👍 cleanup of different indexes and deduct from reserve size (i think that's a separate key). ideally we should somehow leverage the existing delete functionality inside the codebase for this, not to have to patch this up later on when something changes.

pkg/storer/recover.go
```diff
@@ -51,29 +56,72 @@ func sharkyRecovery(ctx context.Context, sharkyBasePath string, store storage.St
...
+		if !cac.Valid(ch) && !soc.Valid(ch) {
+			logger.Warning("recovery: invalid chunk hash, marking corrupted", "address", item.Address)
+			corrupted = append(corrupted, item)
+			return nil
 		}
+
+		return sharkyRecover.Add(item.Location)
+	})
+	if err != nil {
+		return 0, err
```
acud: missing error instrumentation here

pkg/storer/recover.go
```diff
@@ -51,29 +56,72 @@ func sharkyRecovery(ctx context.Context, sharkyBasePath string, store storage.St
...
 	}
 
 	if err := sharkyRecover.Save(); err != nil {
-		return err
+		return 0, err
```
acud: missing error instrumentation here

pkg/storer/recover.go
```diff
@@ -51,29 +56,72 @@ func sharkyRecovery(ctx context.Context, sharkyBasePath string, store storage.St
...
+	}
+
+	for _, item := range corrupted {
+		if err := errors.Join(
+			store.Delete(item),
+			reserve.DeleteCorruptedChunkMetadata(store, baseAddr, item.Address),
+		); err != nil {
+			logger.Error(err, "recovery: failed deleting corrupted chunk", "address", item.Address)
```
acud: you're logging and returning the error here, which is an antipattern, since the calling context will also probably either log or return. finally, if the error is swallowed (non-fatal) then it should be logged anyway

pkg/storer/recover.go
```diff
@@ -51,29 +56,71 @@ func sharkyRecovery(ctx context.Context, sharkyBasePath string, store storage.St
...
+		return 0, fmt.Errorf("recovery: failed iterating chunk index: %w", err)
 	}
 
 	if err := sharkyRecover.Save(); err != nil {
-		return err
+		return 0, fmt.Errorf("recovery: failed saving sharky recovery state: %w", err)
+	}
+
+	for _, item := range corrupted {
+		if err := errors.Join(
```
gacevicljubisa: From my point of view, sequential approach is cleaner for error reporting:

```suggestion
for _, item := range corrupted {
    if err := reserve.DeleteCorruptedChunkMetadata(store, baseAddr, item.Address); err != nil {
        return 0, fmt.Errorf("recovery: failed deleting corrupted chunk metadata %s: %w", item.Address, err)
    }
    if err := store.Delete(item); err != nil {
        return 0, fmt.Errorf("recovery: failed deleting corrupted chunk index %s: %w", item.Address, err)
    }
}
```

martinconic: nice suggestion

PR #5415, opened 2026-03-30
chore: remove backwards compatibility with 2.6.0

### LINKED ISSUE #5340: Remove backwards compatibility code to clean up codebase
# Remove backwards compatibility code to clean up codebase

## Description
The codebase contains backwards compatibility code for Bee version 2.6.0 and potentially other older versions. As the network moves forward, this compatibility code is no longer needed and should be removed to improve code quality, performance, and maintainability.

## Bee 2.6.0 Compatibility Code

The following Bee 2.6.0-specific compatibility code has been identified:

**Affected files:**
- [pkg/p2p/libp2p/internal/handshake/handshake.go](pkg/p2p/libp2p/internal/handshake/handshake.go#L70-L81) - `WithBee260Compatibility` option and field
- [pkg/p2p/libp2p/libp2p.go](pkg/p2p/libp2p/libp2p.go#L1483-L1520) - `IsBee260()` and `bee260BackwardCompatibility()` methods
- [pkg/p2p/libp2p/peer.go](pkg/p2p/libp2p/peer.go#L24) - `bee260Compatibility` tracking map and related methods
- [pkg/p2p/p2p.go](pkg/p2p/p2p.go#L247-L260) - `FilterBee260CompatibleUnderlays()` function
- [pkg/hive/hive.go](pkg/hive/hive.go#L53) - `Bee260CompatibilityStreamer` interface usage
- [pkg/p2p/libp2p/version_test.go](pkg/p2p/libp2p/version_test.go) - Tests for 2.6.0 compatibility

**Interfaces to remove:**
- `p2p.Bee260CompatibilityStreamer` interface

## Other Legacy Compatibility Code

Additional backwards compatibility code exists that may also be outdated:

**Legacy address format compatibility:**
- [pkg/bzz/address.go](pkg/bzz/address.go#L43) - `Underlay` field in `addressJSON` for backward compatibility
- [pkg/bzz/underlay.go](pkg/bzz/underlay.go#L23-L60) - Serialization/deserialization logic for single vs. multiple underlay addresses
- [pkg/bzz/underlay_test.go](pkg/bzz/underlay_test.go#L208) - `TestLegacyCompatibility` test

**Other legacy code:**
- [pkg/crypto/crypto.go](pkg/crypto/crypto.go#L124) - `LegacyKeccak256()` function

## Benefits

Removing this backwards compatibility code will:
- **Simplify the codebase** - Less conditional logic and special case handling
- **Improve performance** - Remove unnecessary version checks and filtering operations
- **Reduce maintenance burden** - Fewer edge cases to test and maintain
- **Clean up interfaces** - Remove compatibility-specific interfaces and types
- **Improve code readability** - Clearer intent without compatibility workarounds

## Recommendation

Before removal, ensure that:
1. The network has sufficiently upgraded beyond version 2.6.0
2. A deprecation notice has been communicated to node operators
3. Any remaining nodes on old versions are identified and can be notified

## Tasks

- [ ] Audit backwards compatibility for Bee 2.6.0 and confirm it can be safely removed
- [ ] Identify any other legacy compatibility code beyond what's listed above
- [ ] Remove Bee 2.6.0 compatibility option and related filtering logic
- [ ] Remove `bee260Compatibility` tracking from peer registry
- [ ] Remove `Bee260CompatibilityStreamer` interface
- [ ] Simplify underlay address serialization (remove legacy single-address format support if safe)
- [ ] Review and remove any other deprecated compatibility code
- [ ] Update tests to reflect the removal
- [ ] Document any breaking changes in release notes

sbackend123:
### Checklist

### Description

**Breaking changes**

Removes Bee 2.6.0 backward-compatibility code and the legacy single-multiaddr underlay wire encoding. Minimum supported Bee version on the network is 2.7.0; nodes on 2.6.0 or earlier will fail to complete a handshake with this build.

Underlay wire format
The 0x99 list prefix (underlayListPrefix) is kept. The only supported encoding is now always the prefixed list format, including for a single address or an empty list:

[0x99][varint_len_1][addr_1_bytes][varint_len_2][addr_2_bytes]...

Changes in detail:

- Removed the legacy path where a single underlay was serialized as raw multiaddr bytes (without the 0x99 prefix).
- Even a peer with one underlay is encoded as a one-entry prefixed list.
- An empty underlay list serializes to [0x99] and deserializes to an empty slice.
- DeserializeUnderlays rejects payloads that do not start with 0x99 (including legacy raw single-multiaddr bytes from Bee 2.6.0).
- ParseAddress still rejects addresses with zero underlays after deserialization.

Removed 2.6.0 compatibility shims

- WithBee260Compatibility handshake option
- FilterBee260CompatibleUnderlays / Bee260CompatibilityStreamer / IsBee260
- User-agent–based Bee 2.6.0 detection in libp2p peer registry

Database migrations

Legacy migration step implementations are removed; version indices are retained as NOOPs so existing data directories can still migrate safely (ValidateVersions requires contiguous version numbers). Comments document why these slots must not be removed or renumbered.

Also removes unused legacy helpers (olditems, oldstampindex, obsolete storer migration step files, etc.).

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5340

### AI Disclosure


acud:
@sbackend123 I think that removing all migration code might be a bit too eager - can you make sure that the migration code covers the currently active bee node version still showing up in swarmscan?

janos:
> @sbackend123 I think that removing all migration code might be a bit too eager - can you make sure that the migration code covers the currently active bee node version still showing up in swarmscan?

I agree, as there are still 2.6 nodes (not many, but still ~6%), it is early to release bee without compatibility.

This code is intended to be temporary. It is a good effort, but I would suggest to hold on merging, for some time.

sbackend123:
> @sbackend123 I think that removing all migration code might be a bit too eager - can you make sure that the migration code covers the currently active bee node version still showing up in swarmscan?

I ran the previous bee version (v2.6.0) locally, let it sync to current block height on maiinet and build a real data directory. Then I started this build on the same data directory without clearing it. The node started successfully with no migration errors.

I also ran earlier versions trying to sync with mainnet, but get error `connect new stream: incompatible stream: failed to negotiate protocol: protocols not supported`.

acud:
> > @sbackend123 I think that removing all migration code might be a bit too eager - can you make sure that the migration code covers the currently active bee node version still showing up in swarmscan?
> 
> I ran the previous bee version (v2.6.0) locally, let it sync to current block height on maiinet and build a real data directory. Then I started this build on the same data directory without clearing it. The node started successfully with no migration errors.
> 
> I also ran earlier versions trying to sync with mainnet, but get error `connect new stream: incompatible stream: failed to negotiate protocol: protocols not supported`.

I suppose it does not error the migrations because there's nothing to do in the migrations in the current build. I am not 100% sure about the reasoning behind the necessity of those migrations. But we should definitely leave in place the migration code to support the migration path from the node version still running on the network (2.6.0 definitely). The older ones we can mark as nops.

mfw78:
> I agree, as there are still 2.6 nodes (not many, but still ~6%), it is early to release bee without compatibility.

Those 2.6.0 nodes are on some `dirty` tag mostly, so it's on them. They are running custom code - we shouldn't be persisting with cruft in the protocol because someone doesn't update their modifications.

acud:
@sbackend123 this needs a rebase

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -398,8 +373,27 @@ func (s *Service) GetWelcomeMessage() string {
...
+		s.logger.Debug("handshake invalid ack address payload", "underlay_len", len(ack.Address.Underlay), "underlay_first_byte", firstByteString(ack.Address.Underlay), "underlay_prefix", payloadPrefix(ack.Address.Underlay), "overlay_len", len(ack.Address.Overlay), "error", err)
 		return nil, ErrInvalidAck
 	}
 
 	return bzzAddress, nil
 }
+
+func firstByteString(data []byte) string {
+	if len(data) == 0 {
+		return "none"
```
acud, line 385: maybe just `""`?

pkg/storer/migration/step_02.go
```diff
@@ -4,53 +4,13 @@
...
-// step_02 migrates the cache to the new format.
-// the old cacheEntry item has the same key, but the value is different. So only
-// a Put is needed.
-func step_02(st transaction.Storage) func() error {
+import "github.com/ethersphere/bee/v2/pkg/storer/internal/transaction"
 
+// step_02 was a migration step that migrated the cache to a new format.
+// It is now a NOOP since all nodes have already run this migration,
+// and new nodes start with an empty database.
+func step_02(_ transaction.Storage) func() error {
```
acud, line 12: if this is not used anywhere - why do we need to keep the file?

acud (changes requested):
there's a whole chunk of changes that are in `pkg/storer/migration/all_steps.go` that is completely not touched but the code can probably also be removed. as for the individual migration steps which are not in use already - there's no need of keeping the files with empty functions that do nothing - those can be removed.

sbackend123:
> there's a whole chunk of changes that are in `pkg/storer/migration/all_steps.go` that is completely not touched but the code can probably also be removed. as for the individual migration steps which are not in use already - there's no need of keeping the files with empty functions that do nothing - those can be removed.

If I understand everything correctly, we can't remove it completely because of our migration mechanism. 
Existing upgraded nodes may appear fine, but new installations would be broken, because ValidateVersions  checks versions are without gaps (added unit test for that) and when we have currentVersion=0 and then next one 8 -> fail.

pkg/bzz/address.go
akrem-chabchoub, line 222: Nice work @sbackend123.

I just have a concern here, what will happen for the `2.6.0` nodes here ? because I see you removed `Underlay` field from `addressJSON`, I know this is not needed anymore, but in the Unmarshal the node will have the `Underlays` filed empty.

Maybe we can keep it in `addressJSON` (just to for the Unmarshal) and we get that `Underlay` addr and set it to `Underlays` slice
 Right or maybe I'm missing some parts ?

sbackend123: This is main point, we remove compatibility with 2.6.0 so minimal supported version would be 2.7.0. So we force everybody update their versions

pkg/storer/migration/all_steps.go
```diff
@@ -19,19 +19,23 @@ func AfterInitSteps(
...
-		5: step_05(st, logger),
-		6: step_06(st, logger),
-		7: resetReserveEpochTimestamp(st),
+		1: legacyNoopStep,
+		2: legacyNoopStep,
+		3: legacyNoopStep,
+		4: legacyNoopStep,
+		5: legacyNoopStep,
+		6: legacyNoopStep,
+		7: legacyNoopStep,
```
acud, line 33: please add a comment here that noop steps should never be removed since it would cause the new migrations to not execute on old nodes that have a higher migration version. the same on the `legacyNoopStep` - it should explicitly be commented that any migration step marked with this function should never be modified or touched and that the noop function should not be removed. i can easily see an LLM agent "refactoring" this out in the future and breaking stuff because it might not see the bigger context of this.

pkg/statestore/storeadapter/migration.go
```diff
@@ -8,24 +8,26 @@ import (
...
-		7: deletePrefix(st, puller.IntervalPrefix),
-		8: deletePrefix(st, puller.IntervalPrefix),
+		1: noop,
+		2: noop,
+		3: noop,
+		4: noop,
+		5: noop,
+		6: noop,
+		7: noop,
+		8: noop,
```
acud: like in the other migrations comment - the sequence and indexes of these migrations should be commented such that they aren't removed in the future

pkg/bzz/address.go
akrem-chabchoub: Yes, I know.
I’m just thinking out loud, if I were a node operator on `2.6.0` or earlier, I might suddenly get disconnected without really understanding why. That could be a bit confusing. 
But as @janos mentioned, those nodes are only around 6%, so the impact seems pretty small.

akrem-chabchoub (approved):

acud (approved):
LGTM but will need a squash + rebase after the upcoming release

acud:
@sbackend123 linter failing

gacevicljubisa:
@sbackend123 can you update the PR description?

pkg/bzz/underlay.go
```diff
@@ -39,27 +39,12 @@ const maxUnderlaysPerPeer = 20
...
-		b := addrs[0].Bytes()
-		if len(b) > maxUnderlayBytes {
-			return nil, fmt.Errorf("underlay data size %d exceeds maximum of %d bytes: %w", len(b), maxUnderlayBytes, ErrUnderlayByteSizeExceeded)
-		}
-		return b, nil
-	}
-
-	// For 0 or 2+ addresses, the custom list format with the prefix is used.
-	// The format is: [prefix_byte][varint_len_1][addr_1_bytes]...
+	// The format is: [varint_len_1][addr_1_bytes]...
```
acud: missing the prefix byte here. see comment on the left

pkg/bzz/underlay.go
```diff
@@ -88,22 +71,14 @@ func DeserializeUnderlays(data []byte) ([]multiaddr.Multiaddr, error) {
...
 	}
 
-	// Otherwise, the data is handled as a single, backward-compatible multiaddr.
-	addr, err := multiaddr.NewMultiaddrBytes(data)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse as single multiaddr: %w", err)
-	}
-	// The result is returned as a single-element slice for a consistent return type.
-	return []multiaddr.Multiaddr{addr}, nil
+	return deserializeList(data)
```
acud: this doesn't seem right. with this change, the underlay should always start with the prefix byte. the `if` block above should check whether the first byte is _not_ a prefix byte and throw if that is the case, then just deserialize like in the current if statement below (happy path not nested).
also some input length validation is needed since the deserialize call can panic on a short buffer (pass just `0x99` as the underlay)

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -219,6 +199,7 @@ func (s *Service) Handshake(ctx context.Context, stream p2p.Stream, peerMultiadd
 
 	observedUnderlays, err := bzz.DeserializeUnderlays(resp.Syn.ObservedUnderlay)
 	if err != nil {
+		s.logger.Debug("handshake invalid synack observed underlay payload", "payload_len", len(resp.Syn.ObservedUnderlay), "first_byte", firstByteString(resp.Syn.ObservedUnderlay), "payload_prefix", payloadPrefix(resp.Syn.ObservedUnderlay), "error", err)
```
acud: lots of stuff in this log line... not sure if it will be just swallowed in the rest of the logs. maybe just better to instrument the error with a tiny bit of information of what went wrong (not all the fields here). the same for the other logline

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -529,3 +500,21 @@ func (s *Service) parseCheckAck(ctx context.Context, ack *pb.Ack) (*bzz.Address,
...
 }
+
+func firstByteString(data []byte) string {
+	if len(data) == 0 {
+		return ""
+	}
+	return fmt.Sprintf("0x%02x", data[0])
+}
+
+func payloadPrefix(data []byte) string {
```
acud: not sure i understand why this is needed... maybe some info about why and how to use this... also why is `n=16`?

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -529,3 +500,21 @@ func (s *Service) parseCheckAck(ctx context.Context, ack *pb.Ack) (*bzz.Address,
...
+	if len(data) == 0 {
+		return ""
+	}
+	n := 16
+	if len(data) < n {
+		n = len(data)
+	}
+	return fmt.Sprintf("%x", data[:n])
```
acud: you [can](https://pkg.go.dev/builtin@go1.26.4#min) just `fmt.Sprintf("%x",data[:min(16,len(data))])`

acud (changes requested):

pkg/bzz/underlay.go
sbackend123: Fixed: in case if array of underlays contains only 0x99 prefix, decerialize will not panic, it just returns prefix (behavior unchanged)

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -529,3 +498,10 @@ func (s *Service) parseCheckAck(ctx context.Context, ack *pb.Ack) (*bzz.Address,
 
 	return bzzAddress, nil
 }
+
+func invalidObservedUnderlayErr(payload []byte, err error) error {
+	if len(payload) == 0 {
+		return fmt.Errorf("%w: observed underlay (len=0): %w", ErrInvalidSyn, err)
+	}
+	return fmt.Errorf("%w: observed underlay (len=%d, first=0x%02x): %w", ErrInvalidSyn, len(payload), payload[0], err)
```
acud: why do we need conditional error formatting? this appears to be needed just so you have a sneak peek into the first byte (not sure what is the added value of this in production). i would suggest to get rid of this helper and just stringify the `payload` in the error formatting (one format). the `0x99` ascii character is the trademark sign so you would be able to see it in the logs correctly (if ever needed)

acud (changes requested):
last one...

acud (approved):

PR #5423, opened 2026-04-05
fix(puller): non-blocking peer disconnect and sync error backoff

misaakidis:
### Checklist

### Description

Two liveness bugs in the puller are fixed.

#### Bug 1 — topology change holds `syncPeersMtx` for seconds

`onChange` holds `syncPeersMtx` while calling `disconnectPeer`, which called `peer.stop()`:

```
onChange()
  lock(syncPeersMtx)
  → disconnectPeer(peer)
    → peer.stop()
      → cancel per-bin contexts
      → peer.wg.Wait()          ← blocks until all Sync() calls return
```

`Sync()` calls `ReadMsgWithContext`, which only unblocks after the stream times out (`pageTimeout = 1s` per bin). For a node with *B* active bins syncing to a peer that is being disconnected during a radius decrease, `disconnectPeer` held the lock for up to *B* seconds. Any subsequent topology-change notification queued behind the same `manage()` loop was blocked for that entire duration, causing the sync map to diverge from the live topology.

**Fix:** split `peer.stop()` into two methods:

| Method | Behaviour | Used by |
|---|---|---|
| `cancelBins()` | cancel all per-bin contexts, clear the map, **no wait** | `disconnectPeer` |
| `stop()` | `cancelBins()` + `peer.wg.Wait()` | `Close()` (shutdown) |

`disconnectPeer` now calls `cancelBins()`: the peer is removed from the sync map immediately and its goroutines drain in the background. `Close()` still calls `p.wg.Wait()` across all peers, so shutdown correctness is unchanged.

#### Bug 2 — tight CPU spin on non-fatal sync errors

When `Sync()` returns a non-fatal error (stream reset, protocol error, timeout), the goroutine logged the error and fell through to `limiter.WaitN(ctx, count)` with `count=0`. `WaitN(ctx, 0)` returns immediately, so the goroutine looped back and retried with no delay. Any persistent non-fatal error caused a continuous CPU spin until the peer disconnected or the context was cancelled.

**Fix:** add a `syncRetryBackoff = 1s` sleep with a `ctx.Done()` escape after any non-fatal sync error:

```go
select {
case <-time.After(syncRetryBackoff):
case <-ctx.Done():
    return
}
```

This bounds the retry rate to ≤ 1 call/s per goroutine under persistent errors.

### Related Issue

Both bugs were surfaced during pull-sync optimisation work. The topology-freeze bug is most visible at depth transitions where many peers disconnect in rapid succession. See also: `fix/pullsync-interval-advancement`.

### DIFF
--- a/pkg/puller/puller_test.go
+++ b/pkg/puller/puller_test.go
@@ -508,6 +508,55 @@ func TestContinueSyncing(t *testing.T) {
 	}
 }
 
+// TestSyncErrorBackoff verifies that a non-fatal sync error is followed by a
+// backoff before the next retry. Under synctest the worker blocks after one
+// failed call; without the backoff it would spin and consume both replies, so
+// the "exactly one call" check fails without the fix.
+func TestSyncErrorBackoff(t *testing.T) {
+	addr := swarm.RandAddress(t)
+
+	synctest.Test(t, func(t *testing.T) {
+		// Topmost=0 keeps top < start so the interval never advances and the loop
+		// retries with the same start value. Two replies cover the failed retry.
+		ps := mockps.NewPullSync(
+			mockps.WithCursors([]uint64{100}, 0),
+			mockps.WithSyncError(errors.New("stream error")),
+			mockps.WithReplies(
+				mockps.SyncReply{Bin: 0, Start: 1, Topmost: 0, Peer: addr},
+				mockps.SyncReply{Bin: 0, Start: 1, Topmost: 0, Peer: addr},
+			),
+		)
+		kad := kadMock.NewMockKademlia(
+			kadMock.WithEachPeerRevCalls(kadMock.AddrTuple{Addr: addr, PO: 0}),
+		)
+
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		p := puller.New(swarm.RandAddress(t), mock.NewStateStore(), kad, resMock.NewReserve(resMock.WithRadius(0)), ps, nil, log.Noop, puller.Options{Bins: 1})
+		p.Start(ctx)
+
+		kad.Trigger()
+
+		// One failed call, then the worker blocks in the backoff.
+		synctest.Wait()
+		if got := len(ps.SyncCalls(addr)); got != 1 {
+			t.Fatalf("expected worker to back off after 1 failed call, got %d calls", got)
+		}
+
+		// Sleeping out the backoff releases exactly one retry.
+		time.Sleep(puller.SyncRetryBackoff)
+		synctest.Wait()
+		if got := len(ps.SyncCalls(addr)); got != 2 {
+			t.Fatalf("expected exactly one retry after the backoff, got %d calls", got)
+		}
+
+		if err := p.Close(); err != nil {
+			t.Errorf("close puller: %v", err)
+		}
+	})
+}
+
 // TestRadiusDecreaseLiveness verifies that after a radius decrease a new sync
 // worker starts without the manager blocking while the disconnected peer's
 // in-flight worker drains. Under synctest, a blocking disconnect holds up the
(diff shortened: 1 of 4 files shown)


acud:
I think bug 2 is probably going to be easier to reason about. it would be beneficial to split changes into small PRs rather than cramming them together - it makes reviews more difficult, especially for core protocol. the same goes for the other PR.

misaakidis:
I've noted the preference for atomic PRs for the future. For this specific case, would it be okay to leave this PR as is? These liveness bugs are somewhat coupled and validating them together ensures the fix is cohesive.

Btw, I have closed the other PR #5424 as the pullsync interval advancement behavior is correct in the setting of pulling from all peers (as per v2.7.1.)

acud:
> I've noted the preference for atomic PRs for the future. For this specific case, would it be okay to leave this PR as is? These liveness bugs are somewhat coupled and validating them together ensures the fix is cohesive.
> 
> Btw, I have closed the other PR #5424 as the pullsync interval advancement behavior is correct in the setting of pulling from all peers (as per v2.7.1.)

Yeah all good leave it as it is, however it needs a rebase as there are merge conflicts. 
Thanks for closing the other PR. I'll try to get to this one tomorrow.

pkg/puller/puller_test.go
```diff
@@ -507,6 +507,59 @@ func TestContinueSyncing(t *testing.T) {
 	}
 }
 
+// TestSyncErrorBackoff verifies that a non-fatal sync error is followed by a
+// backoff before the next retry, bounding the retry rate to roughly 1/s.
+func TestSyncErrorBackoff(t *testing.T) {
+	t.Parallel()
```
acud: i wonder if we could use `synctest` here instead of the sleeps, spinlock etc.

go.mod
acud, line 14: how is this related to this PR?

misaakidis: this dependency bump is not related to the puller/backoff changes in this PR.
It came from another commit: 61fab37 (chore: update postage snapshot to v0.0.6 (#5401)), which is the commit tagged v2.7.1.

.github/workflows/beekeeper.yml
```diff
@@ -220,6 +228,56 @@
...
+      - name: Test radius-decrease liveness
+        id: radius-decrease
+        run: timeout ${TIMEOUT} beekeeper check --cluster-name local-dns --checks ci-radius-decrease
+      - name: Collect debug artifacts
+        if: failure()
+        run: |
+          bash .github/bin/beekeeper_artifacts.sh local-dns
+          export FAILED='no-test'
+          if ${{ steps.radius-decrease.outcome=='failure' }}; then FAILED=radius-decrease; fi
+          curl -sSf -X POST -H "Content-Type: application/json" -d "{\"text\": \"**${RUN_TYPE}** Radius-Decrease Test Error\nBranch: \`${{ github.head_ref }}\`\nUser: @${{ github.event.pull_request.user.login }}\nDebugging artifacts: [click](https://$BUCKET_NAME.$AWS_ENDPOINT/artifacts_$VERTAG.tar.gz)\nStep failed: \`${FAILED}\`\"}" https://beehive.ethswarm.org/hooks/${{ secrets.TUNSHELL_KEY }}
```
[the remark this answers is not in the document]
misaakidis: Fixed and amended commit that used this vulnerable approach.

misaakidis:
Created a separate validation branch for the CI experiment: [#ci/radius-decrease-validation](https://github.com/ethersphere/bee/tree/ci/radius-decrease-validation) . This PR now stays focused on the puller fix and its direct test coverage only.

go.mod
acud: i see yes. not sure why this commit is on this branch. it would be a good habit to work straight off `master` so that you can then rebase easily.

misaakidis:
Rebased on latest master, ported tests to use synctest, and split into two atomic commits:
- fix(puller): non-blocking peer disconnect
- fix(puller): back off after non-fatal sync errors

brainiac-five:
@gacevicljubisa please report on test progress in week 27.

PR #5429, opened 2026-04-08
chore: stabilize flaky tests and remove flaky runner

sbackend123:
Fix CI

### Checklist

### Description
Goal:  stabilize flaky tests and remove ci-runner for "flaky" tests.
Cross-cutting change:  tests use `rand` with an explicit source so failures are easier to reproduce and debug.

Per test
- **TestBzzUploadDownloadWithRedundancy** —  It was consistently failing on the master branch. The reason: it wasn’t reading `api.SwarmRedundancyLevelHeader`, so it was using the default (PARANOID) level.

- **TestFinder** — Stable on Linux on `master`; only renamed.

- **TestDBNuke** — Fails on Windows. **Cause:** LevelDB sits under a cache layer, which sits under the store, etc. On `Close()`, the cache layer should close the underlying DB but did not, so the file stayed open and the test could not remove its temp directory. **Fix:** close the DB from the cache’s `Close()`, and adjust the cache test so LevelDB is not closed twice.

- **TestGetterRACE** — Stable on Linux on `master`; renamed and given `rand` + source for reproducibility.

- **TestPushChunkToNextClosest** — With `origin=true`, `pushToClosest` may deliver to several nearest peers in parallel; order and outcomes are nondeterministic. Old assertions assumed a fixed successful peer and strict pivot vs peer bookkeeping, and hard-coded which peer “fails” and which “succeeds”. **Fix:** any of the nearest candidates may succeed; assert push stream activity toward **both** nearest candidates and that **exactly one** ends up with a positive balance vs the pivot.

- **TestMakeInclusionProofs** — Stable on Linux on `master`; only renamed.

- **TestAddressBookQuickPrune** — With `storageRadius = 2`, the “good” and “bad” peers can share bin 1. `connectNeighbours` skips that bin (`PO < depth`), while `connectBalanced` can treat the slot as satisfied by the already-connected “good” peer and never dial the “bad” one — **flaky** behavior depending on random overlay geometry. **Fix:** avoid that collision (e.g. don’t put a connected peer in the same balanced bin as the bad peer) and align assertions with the implicit dial from `AddPeers` plus explicit `Trigger`s (e.g. wait for at least `MaxConnAttempts` failed connects, then assert address book prune).

- **TestAnnounceBgBroadcast** — Assertions relied on timing instead of the background goroutine actually running. `cancel()` could run right after `Announce` before `BroadcastPeers` blocked on `<-ctx.Done()`. After `Close()`, a `select` with a fixed `100ms` timeout could fail on slow CI. **Fix:** `bgStarted` closed on first real entry into the background `BroadcastPeers`, and a more generous wait for shutdown on slow CI.

- **TestSnapshot** — Expected an almost immediate snapshot while Kademlia updates asynchronously. **Fix:** poll / wait up to a timeout for the snapshot.

- **TestStart** (*non-empty addressbook* subtest) — Besides the three address-book peers, bootstrap dialing adds more `Connect` calls, so the total is **> 3**, not exactly 3.

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5418

### AI Disclosure


cmd/bee/cmd/db_test.go
```diff
@@ -203,7 +203,17 @@ func TestDBNuke_FLAKY(t *testing.T) {
 
 	db.Close()
 
-	err = newCommand(t, cmd.WithArgs("db", "nuke", "--data-dir", dataDir)).Execute()
+	// Retrying avoids a short OS-level race after db.Close(), where file handles
+	// may still be getting released and early removal can fail on some platforms.
+	backoff := 50 * time.Millisecond
```
acud: maybe just sleep once then (`1s`) and try once?

cmd/bee/cmd/db_test.go
```diff
@@ -203,7 +203,17 @@ func TestDBNuke_FLAKY(t *testing.T) {
...
+	for range 3 {
+		err = newCommand(t, cmd.WithArgs("db", "nuke", "--data-dir", dataDir)).Execute()
```
acud: here, if the first command returns an error (application level), this loop will just swallow it and try again which might just get rid of the application level error, obfuscating the error.

pkg/storage/cache/cache_test.go
```diff
@@ -20,7 +19,6 @@ func TestCache(t *testing.T) {
 	if err != nil {
 		t.Fatalf("create store failed: %v", err)
 	}
-	testutil.CleanupCloser(t, store)
```
acud, line 23: why?

pkg/topology/kademlia/kademlia_test.go
```diff
@@ -2220,6 +2226,60 @@ func waitBcast(t *testing.T, d *mock.Discovery, pivot swarm.Address, addrs ...sw
 	}
 }
 
+func waitChanClosed(t *testing.T, ch <-chan struct{}) {
+	t.Helper()
+
+	err := spinlock.Wait(spinLockWaitTime, func() bool {
```
acud: i would really be happy to get rid of this `spinlock` functionality at some point. it adds a functionality that is just shorter to express using normal golang idioms:
```go
func waitChanClosed(t *testing.T, ch <-chan struct{}) {
	t.Helper()
    select {
      case <-ch:
        return
      case <-time.After(timeWait):
        t.Fatal('timed out')
    }
```
done.

acud:
@sbackend123, thanks for this. what about removing the flaky test runner from the github workflows & makefile?

pkg/storage/cache/cache_test.go
sbackend123: because we close store on cache level, so trying to close it second times you get error

sbackend123:
> @sbackend123, thanks for this. what about removing the flaky test runner from the github workflows & makefile?

Already removed, we also need remove it from required check (I do not have rights for that)

akrem-chabchoub (commented):
The `Test (flaky)` workflow is in a freeze mode.
Although it is removed from `go.yml`, not sure why its left here in CI

Image

cmd/bee/cmd/db_test.go
```diff
@@ -203,6 +203,9 @@ func TestDBNuke_FLAKY(t *testing.T) {
 
 	db.Close()
 
+	// Waiting avoids a short OS-level race after db.Close(), where file handles
+	// may still be getting released and early removal can fail on some platforms.
+	time.Sleep(2 * time.Second)
```
akrem-chabchoub, line 208: Would it be better to replace this with a condition based wait/retry to ensure more deterministic behavior across slower ci environments rather than setting hardcoded time?

acud: @akrem-chabchoub there was something else here before that was a retry... see my previous comment here. if you have any suggestions - welcome (but code example would also be good)

sbackend123:
> The `Test (flaky)` workflow is in a freeze mode. Although it is removed from `go.yml`, not sure why its left here in CI
> Image

It is highly  likely  related to https://github.com/ethersphere/bee/pull/5429#issuecomment-4243624220

acud:
@sbackend123 the job requirement has been removed

pkg/api/bzz.go
```diff
@@ -499,7 +499,7 @@ func (s *Service) serveReference(logger log.Logger, address swarm.Address, pathV
 	}
 
 	ctx := r.Context()
-	ls := loadsave.NewReadonly(s.storer.Download(cache), s.storer.Cache(), redundancy.DefaultLevel)
+	ls := loadsave.NewReadonly(s.storer.Download(cache), s.storer.Cache(), rLevel)
```
martinconic, line 502: Good catch

martinconic:
I think all CI/CD tests should run 10-20 times at least to make sure flaky tests do not occur

sbackend123:
> I think all CI/CD tests should run 10-20 times at least to make sure flaky tests do not occur

I tried on Linux about 50 times, but would be cool if somebody with mac OS would try the same.

acud (approved):

akrem-chabchoub (approved):

PR #5431, opened 2026-04-09
refactor: replace full-node flag with explicit node-mode config

### LINKED ISSUE #5172: Investigate and refactor Bee node modes in node.go
Based on configuration option provided we have to ensure this scenarios are well implemented in Bee:

- blockchain-rpc-endpoint + swap-enable = light node with chequebook
- blockchain-rpc-endpoint = light node without chequebook
- swap-enable = error (disabled chain backend)
- neither = ultra-light node

martinconic:
### Checklist

### Description
Fixes #5172. The previous approach to node mode was implicit and
inconsistent: the mode was inferred from a combination of `--full-node`
(bool) and whether `blockchain-rpc-endpoint` was non-empty, which
diverged from the documented definition of modes and allowed silent
misconfigurations.

This PR introduces a single explicit `--node-mode` option with three
valid values: `full`, `light`, and `ultra-light`.

**Key changes:**

- `NodeMode` string type added to `pkg/node` with constants `FullMode`,
  `LightMode`, `UltraLightMode`; replaces `FullNodeMode bool` in `Options`
- `--node-mode` flag added; `--full-node` deprecated with a warning
  pointing to `--node-mode=full` (see Backward compatibility below)
- `resolveNodeMode()` in `start.go` enforces mode requirements at startup:
  | Mode | Requires |
  |---|---|
  | `full` | `blockchain-rpc-endpoint` + `swap-enable: true` + `chequebook-enable: true` + `storage-incentives-enable: true` |
  | `light` | `blockchain-rpc-endpoint` (`swap-enable` optional; `chequebook-enable` requires `swap-enable`) |
  | `ultra-light` | `swap-enable` must be false (`blockchain-rpc-endpoint` ignored if set) |
- `isChainEnabled` rewritten: no longer infers chain state from the RPC
  endpoint string — now reads directly from `NodeMode`
- `storage-incentives-enable` default corrected to `false` (full node only)
- `chequebook-enable` default corrected to `false` (requires swap-enable)
- All packaging yamls (`bee.yaml`, scoop, homebrew-amd64, homebrew-arm64)
  restructured into labelled mode-based sections

## Backward compatibility

`chequebook-enable` and `storage-incentives-enable` previously defaulted to
`true`, and `--full-node` implied them. With the corrected `false` defaults, an
existing `--full-node` config that relied on those implicit values would now
fail strict full-mode validation and the node **would not start** after upgrade.

To keep upgrades smooth, a legacy `--full-node: true` config (when `node-mode`
is not set) **still starts**: `swap-enable`, `chequebook-enable` and
`storage-incentives-enable` are auto-enabled for any the operator did not set
explicitly, a deprecation warning is logged for each, and the node comes up as a
fully-functional full node. This is a restoration of the previous implied
behavior — not silent degradation: the node runs full with chequebook and
incentives on, exactly as before.

The compatibility default only fills in values left unset. It does **not** mask
genuine misconfigurations:
- explicitly setting any of those options to `false` while requesting a full
  node still fails validation;
- a full node still requires `blockchain-rpc-endpoint` — that cannot be
  inferred;
- the explicit `--node-mode=full` path is unaffected and remains strict (no
  auto-enable), since it is the new, intentional model.

## Test plan

- [x] Start with `node-mode: ultra-light` and no RPC — node starts cleanly
- [x] Start with `node-mode: light`, no RPC — rejected with clear error
- [x] Start with `node-mode: light` + RPC — node starts in light mode
- [x] Start with `node-mode: full`, no RPC — rejected with clear error
- [x] Start with `node-mode: full` + RPC + `swap-enable: false` — rejected
- [x] Start with `node-mode: full` + RPC + `swap-enable: true` + cb/incentives — full node
- [x] Start with legacy `full-node: true` — deprecation warning, behaves as `node-mode: full`
- [x] Start with legacy `full-node: true` + only `blockchain-rpc-endpoint` (no swap/chequebook/incentives) — swap/chequebook/incentives auto-enabled with warnings, starts as full node
- [x] Start with legacy `full-node: true` + `chequebook-enable: false` — still rejected (explicit disable not masked)
- [x] Start with legacy `full-node: true` + no `blockchain-rpc-endpoint` — still rejected
- [x] Start with `node-mode: ultra-light` + `swap-enable: true` — rejected
- [x] `--node-mode` takes precedence over legacy `--full-node` with warning
- [x] `node-mode: ultra-light` with RPC endpoint ignores RPC and starts cleanly
- [x] `node-mode: light` with RPC + swap + chequebook starts cleanly
- [x] Invalid string formats (`""`, uppercase, whitespace) rejected

### AI Disclosure


martinconic:
Beekeeper changes are also needed

acud (approved):

.github/workflows/beekeeper.yml
```diff
@@ -19,7 +19,7 @@ env:
   SETUP_CONTRACT_IMAGE: "ethersphere/bee-localchain"
   SETUP_CONTRACT_IMAGE_TAG: "0.9.4"
   BEELOCAL_BRANCH: "main"
-  BEEKEEPER_BRANCH: "master"
+  BEEKEEPER_BRANCH: "refactor/node-mode-config"
```
sbackend123, line 22: is it for testing only or?

martinconic: yes, will revert before merge

acud:
needs a rebase

cmd/bee/cmd/cmd.go
```diff
@@ -304,9 +305,13 @@ func (c *command) setAllFlags(cmd *cobra.Command) {
 	cmd.Flags().Duration(optionNameBlockchainRpcKeepalive, 30*time.Second, "blockchain rpc TCP keepalive interval")
 	cmd.Flags().String(optionNameSwapFactoryAddress, "", "swap factory addresses")
 	cmd.Flags().String(optionNameSwapInitialDeposit, "0", "initial deposit if deploying a new chequebook")
+	cmd.Flags().String(optionNameNodeMode, string(node.UltraLightMode), "node operational mode: full, light, or ultra-light")
 	cmd.Flags().Bool(optionNameSwapEnable, false, "enable swap")
-	cmd.Flags().Bool(optionNameChequebookEnable, true, "enable chequebook")
-	cmd.Flags().Bool(optionNameFullNode, false, "cause the node to start in full mode")
+	cmd.Flags().Bool(optionNameChequebookEnable, false, "enable chequebook (requires swap-enable)")
```
akrem-chabchoub, line 310: Not sure how this change will affect old users ?

akrem-chabchoub (commented):
Thanks @martinconic, Glad to see this change, It will simplify and enhance the UX 🚀

cmd/bee/cmd/cmd.go
akrem-chabchoub: Maybe some tests could help for those changes ?

Cafe137:
What's the actual difference between `swap-enable` and `chequebook-enable`? Do we really need both?

martinconic:
> What's the actual difference between `swap-enable` and `chequebook-enable`? Do we really need both?

The actual semantics are:

  - swap-enable = "participate in the swap settlement protocol at all" (peer cheque protocol + factory + cashout).
  - chequebook-enable = "I have funded my own chequebook and want to pay peers with cheques."

Functionally yes, we need them both, because swap=on, chequebook=off (receive-only / cash-out-only) is a genuine state. It's useful for: a node booting before its chequebook is funded, an operator who wants to earn from serving content without depositing collateral yet, and any deferred-funding workflow.

akrem-chabchoub (commented):
Just there are some git conflicts

martinconic:
> Just there are some git conflicts

New things were merged until my last commit, so yes, will address the merge conflicts.

cmd/bee/cmd/start.go
```diff
@@ -369,6 +373,109 @@ func buildBeeNode(ctx context.Context, c *command, cmd *cobra.Command, logger lo
...
+		case node.FullMode:
+			if err := validateFullMode(); err != nil {
+				return "", err
+			}
+		case node.LightMode:
+			if rpcEndpoint == "" {
+				return "", errors.New("light node requires blockchain-rpc-endpoint to be set")
+			}
+		case node.UltraLightMode:
+			if swapEnable {
```
aloknerurkar, line 458: nit: If we are returning error for incorrectly setting swap, we can also return error for incorrectly setting storage incentives for light and ultra light mode?

honestly I feel like it will be cleaner to have separate functions for the 3 different modes. There are so many if-else cases in the node.go file, it is easy to do something wrong here. But that is a much bigger change.

aloknerurkar (approved):

cmd/bee/cmd/start.go
```diff
@@ -369,6 +373,109 @@ func buildBeeNode(ctx context.Context, c *command, cmd *cobra.Command, logger lo
...
+	if c.config.GetBool(optionNameFullNode) && !c.config.IsSet(optionNameNodeMode) {
+		logger.Warning("--full-node is deprecated, use --node-mode=full instead")
+		for _, key := range []string{
+			optionNameSwapEnable,
+			optionNameChequebookEnable,
+			optionNameStorageIncentivesEnable,
+		} {
+			if !c.config.IsSet(key) {
+				logger.Warning("enabling option implied by legacy --full-node; set it explicitly or use --node-mode=full", "option", key)
+				c.config.Set(key, true)
```
akrem-chabchoub, line 405: For old users who ran a full node with `swap-enable` left at its default (`false`, never explicitly set), upgrading while still using the deprecated `--full-node` flag  causes `swap-enable`,`chequebook-enable` and `storage-incentives-enable` to be **silently forced to change** on startup. This is only surfaced via a log, which is easy to miss maybe ?

Does this not risk misbehavior or misunderstanding for node operators ?

akrem-chabchoub (approved):

cmd/bee/cmd/start.go
martinconic: Prior to this PR, chequebook-enable and storage-incentives-enable actually defaulted to true. So an existing full node was already running with chequebook and incentives active.

When we corrected those defaults to false in this PR to make the new configuration explicit, an existing --full-node config that relied on the old defaults would suddenly fail startup validation after an upgrade.

The auto-enable logic in the legacy fallback doesn't introduce unexpected behavior—it restores the previous implicit full-node stack so existing nodes don't crash on upgrade, while logging clear warnings advising operators to migrate to --node-mode=full. Crucially, it only applies to options left unset; an explicit false is never overridden and still fails validation as expected.

gacevicljubisa (changes requested):
Requesting changes on the upgrade path. The legacy branch currently gets some of the new semantics (strict validation, swap forced on) but not all (no defaults restored for light nodes), so existing configs break in both directions:

- light node with swap-enable: true, no chequebook-enable → chequebook-enable default flipped (cmd.go:363), nothing restores it, SetPayFunc never installed (node.go:1115) — starts clean, silently loses cheque settlement
- any config still carrying chequebook-enable: true without swap (the old shipped default, still in packaging/docker/env) → exits at start.go:417; master silently gated this under swap (node.go:595)
- bootnode with storage-incentives-enable: false → exits at start.go:472, though node.go:1295 never starts the agent for bootnodes anyway
- BEE_NODE_MODE missing from the compose passthrough (docker-compose.yml:33)

None of these are in the test plan, and the tests go through a bare viper.New() so the pflag default of ultra-light (which must not count as set) is never exercised.

Proposal: two regimes, one switch.

- node-mode unset → master's behavior verbatim: old defaults, old inference, no new validation. Log a deprecation warning with the equivalent node-mode: line and the release it will be removed in.
- node-mode set → the mode owns the config, no sub-flags: ultra-light = no chain/swap/incentives, light = chain + swap, full = chain + swap + incentives. full-node, swap-enable, chequebook-enable, storage-incentives-enable become deprecated and warn; explicit contradictions error. The cost is a chequebook deploy on light nodes, which only needs gas (swap-initial-deposit is 0) — and a light node that buys stamps needs gas anyway.

That removes the restore loop and validateFullMode entirely (only "chain required for light/full" remains), gives a clean cutover date, and makes node-mode: full sufficient on its own — today it's rejected without three more flags.

PR #5435, opened 2026-04-14
feat: pubsub

nugaon:
# pubsub

A brokered publish/subscribe protocol over Bee's p2p layer, exposed to end-users via WebSocket.

## Overview

One node acts as a **broker**: it accepts p2p streams, validates incoming messages, and fans them out to all connected receivers. Other nodes connect as either a **publisher** (publish + receive) or a **subscriber** (receive only).

## Connection flow

```
WebSocket client
      │  HTTP GET /pubsub/{topic}  (broker peer multiaddr in HTTP request headers)
      ▼
api/pubsub.go  ──►  pubsub.Service.Connect  ──►  mode.Connect (opens p2p stream)
                                                         │
                                               Broker node — brokerHandler
                                                     ├─ readwrite=1  ──►  handlePublisher
                                                     └─ readwrite=0  ──►  handleSubscriber
```

The connecting node receives a `SubscriberConn` which is bridged to the WebSocket: inbound WS frames are forwarded to the p2p stream (publishers only); p2p frames from the broker are decoded and forwarded to the WS client (all connections).

## Mode system

Protocol-specific behaviour is isolated behind a `Mode` interface. The `Service` is mode-agnostic: it delegates header construction, message reading/validation, and broadcast formatting to the active mode. New protocol variants can be added by implementing `Mode` and registering a mode ID in `newMode`.

## GSOC Ephemeral mode (mode 1)

Messages are Single Owner Chunks (SOC) signed by the holder of the topic's private key. The topic address is derived from the SOC owner public key and an arbitrary topic ID, so only the key holder(s) can publish and the messages carry full SOC wrapping.

### Publisher → Broker (p2p)

```
[ sig: 65 B ][ span: 8 B LE ][ payload: up to 4 KB ]
```

The broker verifies the ECDSA signature on every message before broadcasting.

### Broker → Subscriber (p2p)

Every broker frame begins with a 1-byte message type. `0x01` is reserved at the **service level** and is valid across all modes; all other type bytes are mode-specific and decoded by the active mode implementation.

| First byte | Level | Meaning | Payload |
|---|---|---|---|
| `0x01` | service | Ping | _(no fields — keepalive sent every 30 s)_ |
| `0x02` | GSOC mode | Handshake + data | `[SOC ID: 32 B][owner: 20 B][sig: 65 B][span: 8 B][payload: N B]` |
| `0x03` | GSOC mode | Data only | `[sig: 65 B][span: 8 B][payload: N B]` |

The handshake frame (SOC identity metadata) is sent once per topic on the first broadcast. Subsequent messages are data-only. Ping frames are consumed by `readServiceMessage` before mode dispatch and never surfaced to the WebSocket client.

### WebSocket (both directions)

Both the publisher (inbound) and subscriber (outbound) see the raw SOC payload:

```
[ sig: 65 B ][ span: 8 B ][ payload: N B ]
```

The node handles all p2p framing and signature verification transparently.

## Multi-WebSocket multiplexer

Multiple WebSocket sessions to the same topic on the same node share a **single p2p stream** to the broker via `SubscriberConn`:

- `CreateSubscriberConn` increments a ref count instead of opening a new stream when one already exists
- A single `runMux` goroutine fans out incoming broker messages to all per-session channels; on stream error it immediately clears the shared conn so new sessions get a fresh stream
- `RemoveSubscriberConn` decrements refs; `FullClose` is called exactly once when refs reach zero

## API

| Method | Path | Description |
|---|---|---|
| `GET` (WS upgrade) | `/pubsub/{topic}` | Connect to a topic as subscriber or publisher |
| `GET` | `/pubsub/` | List active topics with role and connection count |

**Query params for `/pubsub/{topic}`:**
- `peer` (required): multiaddr of the broker peer
- `gsoc-eth-address` + `gsoc-topic` (optional): upgrade to publisher role

**Header:**
- `swarm-keep-alive` (optional): WebSocket ping period in seconds (default: 60)

## Configuration

| Flag | Default | Description |
|---|---|---|
| `--pubsub-broker-mode` | `false` | Enable broker role on this node |

Subscriber nodes use `ConnectAllowLight` when dialing the broker, so broker nodes can run in light-node mode.

## Extensibility

- **New modes** — implement `Mode` for a different message format and register a mode ID.
- **Access control** — `validatePublisher` can enforce allow-lists, or stake checks.
- **Replay** — `handleSubscriber` can replay on connect.
- **WASM / event-driven clients** — `broadcast` can forward to an event emitter, making the broker embeddable in a WASM-compiled client without a separate p2p layer.
- **Per-subscriber transforms** — `formatBroadcast` receives the individual `brokerSubscriber`, enabling selective filtering or transformation.

## Tests

**`pkg/pubsub` (unit)**
- `mode_1_test.go` — `GSOCEphemeralMode`: message formatting (handshake vs data), `ReadBrokerMessage` for ping/handshake/data/invalid-sig, `setGsocParams` address mismatch and valid params, `SubscriberConn` fanout, unsubscribe, and full-channel overflow.
- `pubsub_test.go` — broker handler error paths (disabled, missing/unknown headers) and `Service.Topics()` (empty list, broker-role registration via spinlock).

**`pkg/api` (integration)**
- `pubsub_test.go` — nil service → 404, empty list → 200, missing peer/invalid multiaddr/bad GSOC params → 400, subscriber WS connect/teardown, and a two-topic scenario: opens two concurrent WS sessions, writes broker pings through the mock p2p streams, queries `GET /pubsub/` asserting both topics are listed, then closes each connection sequentially verifying the topic count decrements to 0.

### Checklist

### Related Issue (Optional)
[SWIPs #93 — PubSub protocol](https://github.com/ethersphere/SWIPs/pull/93) — this PR implements Milestone 1 (Direct messaging) of that SWIP.

### AI Disclosure


pkg/pubsub/mode.go
```diff
@@ -0,0 +1,210 @@
...
+const (
+	// P2P headers
+	HeaderGsocOwner = "pubsub-gsoc-owner"
+	HeaderGsocID    = "pubsub-gsoc-id"
+)
+
+// Mode defines mode-specific behavior for the pubsub protocol.
+// Each mode determines its own roles, wire format, and message handling.
+type Mode interface {
+	ID() uint8
```
zelig: make an 'enum type' here please, it will still remain extensible

pkg/pubsub/mode.go
```diff
@@ -0,0 +1,210 @@
...
+	if len(opts.GsocOwner) > 0 {
+		headers[HeaderGsocOwner] = opts.GsocOwner
+	}
+	if len(opts.GsocID) > 0 {
+		headers[HeaderGsocID] = opts.GsocID
+	}
+	return p.NewStream(ctx, overlay, headers, protocolName, protocolVersion, streamName)
+}
+
+// ValidateParticipant sets SOC parameters on the broker side so it can validate the messages.
```
zelig: I kindly request the renaming 'participant' to 'publisher', since it is exactly what they are. participant is neutral and should subsume subscriber as well as publisher as well as broker roles.

pkg/pubsub/pubsub.go
```diff
@@ -0,0 +1,399 @@
...
+	}
+}
+
+// broadcastToSubscribers sends a message to all subscribers of a topic.
+func (s *Service) broadcastToSubscribers(bc *brokerConn, rawMsg []byte) {
+	bc.mu.Lock()
+	defer bc.mu.Unlock()
+
+	for _, sub := range bc.subscribers {
+		msg := bc.mode.FormatBroadcast(bc, sub, rawMsg)
```
zelig, line 357: I dont think this is needed inside the fanout loop. Why would you reserialise for every subscriber?

nugaon: the message serialization varies between subscribers e.g. one of them did not get the SOC ID and address for message verification, another subscriber already got it. Then, these metadata will be either attached or not in the message payload.

pkg/pubsub/mode.go
nugaon: renamed, though a 'publisher' receives messages as well.

pkg/pubsub/pubsub.go
zelig: why would it vary between subscribers? In fact it must be the same, otherwise same msg can be repeated.

pkg/api/router.go
```diff
@@ -44,7 +44,15 @@ func (s *Service) Mount() {
 
 	s.Handler = web.ChainHandlers(
 		httpaccess.NewHTTPAccessLogHandler(s.logger, s.tracer, "api access"),
-		handlers.CompressHandler,
```
acud, line 47: how is it that this is needed? we've had websockets running in the API for years now. by what you're saying, they couldn't have worked so far. it would be good to know why this is needed

pkg/log/httpaccess/http_access.go
```diff
@@ -118,13 +125,27 @@ func (rr *responseRecorder) CloseNotify() <-chan bool {
...
 }
 
-// Hijack implements http.Hijacker.
+// Hijack implements http.Hijacker so that WebSocket upgrades pass through
+// correctly. Without this, the underlying connection is hijacked by the
+// upgrader but the recorder still holds a reference to the (now-hijacked)
+// ResponseWriter, causing "response.Write on hijacked connection" errors.
 func (rr *responseRecorder) Hijack() (net.Conn, *bufio.ReadWriter, error) {
-	return rr.ResponseWriter.(http.Hijacker).Hijack()
+	h, ok := rr.ResponseWriter.(http.Hijacker)
```
acud, line 133: not sure why this is needed. can you please explain why the existing websocket machinery that is used to stream chunks cannot be reused? also you are affecting that machinery with this change.

pkg/api/pubsub.go
```diff
@@ -0,0 +1,137 @@
...
+		pingPeriod = time.Minute
+	}
+
+	isPublisher := connectOpts.ReadWrite
+
+	s.wsWg.Add(1)
+	go func() {
+		pubsub.ListeningWs(ctx, conn, pubsub.WsOptions{PingPeriod: pingPeriod, Cancel: cancel}, logger, mode, isPublisher)
+		cancel()
+		_ = conn.Close()
```
acud, line 120: please use defer statements inside the closure before the `ListeningWs` call for both cleanup commands. when that call panics, the runtime will do a stack unwind and will execute the defer statements. in your version here, since the calls are made after the potential panic, they never get cleaned up

pkg/api/pubsub.go
```diff
@@ -0,0 +1,137 @@
...
+	paths := struct {
+		Topic string `map:"topic" validate:"required"`
+	}{}
+	if response := s.mapStructure(mux.Vars(r), &paths); response != nil {
+		response("invalid path params", logger, w)
+		return
+	}
+
+	var topicAddr [32]byte
+	if decoded, err := hex.DecodeString(paths.Topic); err == nil && len(decoded) == swarm.HashSize {
```
acud, line 33: can you please add some comments on what `topic` could be? there's zero comments on the handler and zero inline comments, which is not so nice to read (i'd rather not have to reach out to the openapi spec or BOS to check what a parameter might mean). ideally you can add a comment in the anonymous type on that field, explaining what it is.
also just to be a little bit more specific. if a parameter is a hex address OR an arbitrary length string, then it is ONLY an arbitrary length string. whether someone, by convention, chooses to use one over the other, it is a bad practice to disambiguate them. i would suggest to have just one _type_ and use it however using conventions. i.e. _always hash the topic_.

pkg/api/pubsub.go
```diff
@@ -0,0 +1,137 @@
...
+		copy(topicAddr[:], decoded)
+	} else {
+		h := swarm.NewHasher()
+		_, _ = h.Write([]byte(paths.Topic))
+		copy(topicAddr[:], h.Sum(nil))
+	}
+
+	peerMultiaddr := r.URL.Query().Get("peer")
+	if peerMultiaddr == "" {
+		jsonhttp.BadRequest(w, "missing peer query param")
```
acud, line 43: query params are usually optional, and rest path parameters are mandatory. if the peer multiaddr is mandatory, it should be in the path params, and you avoid these sort of checks then

pkg/api/pubsub.go
```diff
@@ -0,0 +1,137 @@
...
+		if err != nil {
+			jsonhttp.BadRequest(w, "invalid gsoc-topic query param")
+			return
+		}
+		connectOpts.GsocOwner = gsocOwner
+		connectOpts.GsocID = gsocID
+		connectOpts.ReadWrite = true
+	}
+
+	headers := struct {
```
acud, line 73: nit: i would suggest to put this part next to the other path parsing part at the top

pkg/api/pubsub.go
```diff
@@ -0,0 +1,137 @@
...
+		KeepAlive int `map:"Swarm-Keep-Alive"`
+	}{}
+	if response := s.mapStructure(r.Header, &headers); response != nil {
+		response("invalid header params", logger, w)
+		return
+	}
+
+	// Connect to broker peer
+	ctx, cancel := context.WithCancel(context.Background())
```
acud, line 82: you lost the main api context here. so if the api shuts down, you don't know about it. you want to add another layer of cancellation over the existing parent context

pkg/api/pubsub.go
```diff
@@ -0,0 +1,137 @@
...
+	mode, err := s.pubsubSvc.Connect(ctx, underlay, topicAddr, pubsub.ModeGSOCEphemeral, connectOpts)
+	if err != nil {
+		cancel()
+		logger.Info("pubsub connect failed", "error", err)
+		jsonhttp.InternalServerError(w, "pubsub connect failed")
+		return
+	}
+	// Upgrade to WebSocket
+	upgrader := websocket.Upgrader{
+		ReadBufferSize:  swarm.ChunkWithSpanSize,
```
acud, line 92: can't this be bigger? see `swarm.SocMaxChunkSize`

pkg/api/pubsub.go
```diff
@@ -0,0 +1,137 @@
...
+		WriteBufferSize: swarm.ChunkWithSpanSize,
+		CheckOrigin:     s.checkOrigin,
+	}
+
+	logger.Info("upgrading to websocket")
```
acud, line 97: this really gives 0 context (which api endpoint? which client?). suggesting to remove

pkg/api/pubsub.go
```diff
@@ -0,0 +1,137 @@
...
+	conn, err := upgrader.Upgrade(w, r, nil)
+	if err != nil {
+		// Upgrade() hijacks the connection before returning an error,
+		// so do NOT write an HTTP response here.
+		cancel()
+		logger.Info("websocket upgrade failed", "error", err)
+		logger.Error(nil, "websocket upgrade failed")
+		return
+	}
+	logger.Info("websocket upgrade successful")
```
acud, line 107: ditto

pkg/api/pubsub.go
```diff
@@ -0,0 +1,137 @@
...
+
+	pingPeriod := time.Duration(headers.KeepAlive) * time.Second
+	if pingPeriod == 0 {
+		pingPeriod = time.Minute
+	}
+
+	isPublisher := connectOpts.ReadWrite
```
acud, line 114: er?

pkg/p2p/libp2p/libp2p.go
```diff
@@ -621,7 +620,7 @@ func (s *Service) handleIncoming(stream network.Stream) {
 		observedAddrs, err = buildFullMAs([]ma.Multiaddr{stream.Conn().RemoteMultiaddr()}, peerID)
 		if err != nil {
 			s.logger.Debug("stream handler: handshake: build remote multiaddrs fallback", "peer_id", peerID, "error", err)
-			s.logger.Error(nil, "stream handler: handshake: build remote multiaddrs fallback", "peer_id", peerID)
```
acud, line 623: please revert these changes, they are not needed.

pkg/pubsub/mode_1_test.go
```diff
@@ -0,0 +1,381 @@
+// Copyright 2026 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Testing out Mode 1 (GSoC Ephemeral) implementation
```
acud, line 5: in go, the file naming scheme is `<file>.go` and its corresponding `<file>_test.go`. let's stick to that please

pkg/pubsub/ws.go
```diff
@@ -0,0 +1,126 @@
...
+	PingPeriod time.Duration
+	Cancel     context.CancelFunc
+}
+
+// ListeningWs bridges a subscriber's p2p stream to a WebSocket connection.
+// The Mode handles all wire-format details: reading broker messages,
+// verifying them, and returning the payload to forward to the WebSocket.
+// If the subscriber is a Publisher, it also reads from the WebSocket
+// and writes raw messages to the p2p stream.
+func ListeningWs(ctx context.Context, conn *websocket.Conn, options WsOptions, logger log.Logger, mode Mode, isPublisher bool) {
```
acud, line 25: i can't see why anything apart from the `api` package should leak `websocket` types. this does not make much sense and i would request that the abstractions handle this requirement properly. just in the same way that we don't leak `cobra`/`viper` types outside of the `cmd` package, in the same way no HTTP types should leak out of the `api` package. thanks

pkg/pubsub/pubsub.go
```diff
@@ -0,0 +1,374 @@
...
+// Connect establishes a subscriber connection to a broker peer.
+func (s *Service) Connect(ctx context.Context, underlay ma.Multiaddr, topicAddr [32]byte, modeID ModeID, opts ConnectOptions) (Mode, error) {
+	key := TopicModeKey{TopicAddr: topicAddr, ModeID: modeID}
+	m, err := s.getOrCreateMode(key)
+	if err != nil {
+		return nil, err
+	}
+
+	s.logger.Info("connecting to broker peer", "underlay", underlay)
+	bzzAddr, err := s.p2p.ConnectAllowLight(ctx, []ma.Multiaddr{underlay})
```
acud, line 177: let's please not mix these things. a protocol should not be able to dial to another peer. this is not good news. there is already existing functionality in the `api` package to dial to peers. any sort of connection orchestration should be done from there by a different client that would know to tell bee to connect to another bee. for me having a protocol that can coerce a node to create connections via other messages (be they websocket or p2p messages) is a serious attack surface. a protocol should be "read message, write message". look at pushsync/pullsync and their relevant orchestration layers pusher/puller respectively. we are doing things with this architecture for a good reason. if this deviation is required we should at least be shown some justification as to _why_.

pkg/pubsub/pubsub.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+}
+
+// fanOut broadcasts a message to all registered WS session channels.
+func (sc *SubscriberConn) fanOut(msg []byte) {
+	sc.subsMu.Lock()
+	defer sc.subsMu.Unlock()
+	for _, ch := range sc.subs {
+		select {
+		case ch <- msg:
```
acud, line 353: one reader stalling delays the rest. and you're holding the mutex all that time

pkg/pubsub/pubsub.go
```diff
@@ -0,0 +1,374 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	s.modes[key] = m
+	return m, nil
+}
+
+// writeRaw writes raw bytes to the stream.
+func writeRaw(stream p2p.Stream, data []byte) error {
```
acud, line 291: why can't you just `Write` once? not sure why the helper is needed. libp2p streams can handle enough data unless you can prove otherwise?

pkg/pubsub/pubsub.go
```diff
@@ -0,0 +1,374 @@
...
+	logger  log.Logger
+}
+
+// Subscribe registers a new WS session and returns its per-session message channel.
+func (sc *SubscriberConn) Subscribe() (uint64, <-chan []byte) {
+	sc.subsMu.Lock()
+	defer sc.subsMu.Unlock()
+	id := sc.nextID
+	sc.nextID++
+	ch := make(chan []byte, 16)
```
acud, line 332: channel buffer size is either 0 or 1 please. this is a recurring pattern i see that in PRs this rule is starting to be broken. this is against go styleguides and against _our adopted styleguide_.

acud (changes requested):
Thanks @nugaon , once these changes are addressed I will give it another whirl. It seems that for what we are trying to achieve this is too complex. It can be slimmed down significantly

pkg/pubsub/pubsub.go
nugaon: this is an API endpoint that connects to another bee.

pkg/pubsub/pubsub.go
nugaon: because in this case I'm not interested in `n`, only the error.

pkg/pubsub/ws.go
nugaon: I thought it may be a good idea for a webassembly port for in-browser client, because it just needs to replace this file with some eventEmitter stuff.

pkg/log/httpaccess/http_access.go
nugaon: there was some security error, unfortunately I don't remember now. :c

pkg/api/router.go
nugaon: it worked just maybe it wasn't the best. AI said this part would be good to check it with other experts as well.

pkg/api/pubsub.go
nugaon: your connection has a publisher role, if you have read and write operation.

pkg/api/pubsub.go
nugaon: logger := s.logger.WithName("pubsub").Build()

pkg/api/pubsub.go
nugaon: hmm, I forgot but maybe we just give back span with soc payload on the websocket. 
anyway, it should be maybe even bigger because in the future there should be similar like #5497

pkg/api/pubsub.go
nugaon: in the future it will be removed (milestone 3)

pkg/api/pubsub.go
nugaon: there is a SWIP about it that details, but I copy here:

> The topic address is `soc.CreateAddress(socID, ownerAddr)`, so only the holder of the topic private key can publish.

it regards now to the only channelMode so it can be anything in the future.

pkg/api/pubsub.go
nugaon: or optional

pkg/api/pubsub.go
acud: still zero context... which peer? parameters used? any possible information to help debug why the upgrade failed?

pkg/pubsub/pubsub.go
acud: just ignore the returned value then, there's no need for this otherwise. a stream shouldn't limit you in how much data you can write, and should take care of doing what you're doing here

nugaon:
codeQL error seems false.

acud:
@nugaon please revert the api package changes. the static analyzer is complaining and it could be that you're just tripping something off with your changes. since you can't remember why you did them, then please undo them (and if any other error arises then - we will deal with it in due time)

pkg/pubsub/pubsub.go
acud: why do you need it? there's already an existing one in the api package. why don't you use it? and why do you need to duplicate the functionality here? we have already clear tested paths for establishing connections. why do you need to do an extra 2 hops through abstractions in order to perform something that is already clearly available in the api package? also, why do you need to make any user facing functionality perform these two things at once? why don't you split the connection and pubsub subscription establishing into two api calls?

zelig:
should this be closed: @nugaon @bcsorvasi

nugaon:
Closing this PR as [SWIP 60](https://github.com/ethersphere/SWIPs/pull/104) details further this POC and being implemented again based on that.

Issue #5444, opened 2026-04-23
Missing tag and deferred headers for SOC


agazso:
### Context
Bee 2.7.1

### Summary
The `swarm-tag` and `swarm-deferred-upload` header is missing on the `/soc` endpoint when uploading a Single-owner Chunk. That means that sometimes uploading an SOC may take several seconds (>15 seconds in some cases) because the API waits until chunk is pushed to the correct neighborhood.

### Expected behavior
I would expect to have the `swarm-tag` and `swarm-deferred-upload` headers similarly to the `/chunks` endpoint with the same semantics.

### AI Disclosure
- [x] This issue does not contain suggestions and text generated by an LLM.

PR #5445, opened 2026-04-25
fix: reverse SOC/CAC detection order in POST /chunks handler

significance:
### Checklist

### Description

The `POST /chunks` handler tries CAC (Content Addressed Chunk) parsing before SOC (Single Owner Chunk) parsing. Since CAC parsing accepts any data between 8-4104 bytes, valid SOC data is always misclassified as CAC and stored at the wrong content-addressed hash instead of the correct SOC address (`Hash(Identifier || Owner)`).

This breaks client-side SOC construction workflows where pre-stamped SOCs are uploaded via `POST /chunks` — the chunk is stored but can never be retrieved at its expected SOC address.

The fix reverses the detection order: try SOC parsing first, fall back to CAC.

N/A — no API contract change, only fixes incorrect internal routing of chunk types.

#### Motivation and Context

Client-side SOC construction (e.g. using `bee-js` `makeContentAddressedChunk().toSingleOwnerChunk()` then `uploadChunk()`) relies on `POST /chunks` correctly identifying and storing SOCs at their SOC address. Without this fix, all SOCs uploaded via this endpoint are silently stored as CACs at the wrong address.

### Related Issue

N/A

N/A

### AI Disclosure

🕊️

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -94,6 +94,7 @@ const (
 	SwarmActTimestampHeader           = "Swarm-Act-Timestamp"
 	SwarmActPublisherHeader           = "Swarm-Act-Publisher"
 	SwarmActHistoryAddressHeader      = "Swarm-Act-History-Address"
+	SwarmChunkTypeHeader              = "Swarm-Chunk-Type"
 
 	ImmutableHeader = "Immutable"
 	GasPriceHeader  = "Gas-Price"
--- a/pkg/api/chunk.go
+++ b/pkg/api/chunk.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/ethersphere/bee/v2/pkg/accesscontrol"
 	"github.com/ethersphere/bee/v2/pkg/cac"
@@ -37,6 +38,7 @@ func (s *Service) chunkUploadHandler(w http.ResponseWriter, r *http.Request) {
 		SwarmTag       uint64        `map:"Swarm-Tag"`
 		Act            bool          `map:"Swarm-Act"`
 		HistoryAddress swarm.Address `map:"Swarm-Act-History-Address"`
+		ChunkType      string        `map:"Swarm-Chunk-Type"`
 	}{}
 	if response := s.mapStructure(r.Header, &headers); response != nil {
 		response("invalid header params", logger, w)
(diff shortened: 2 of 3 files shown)


acud:
@significance thanks for this. a few comments:
- the `/chunks` api endpoint was never made to upload SOCs
- i am in favor of this api simplification, but in which case we should also get rid of the `soc` api path - it makes no sense to keep both and just increases confusion in the api
- having a unified api would be better. but it would also require just having a unified type of upload requests. right now those are asymmetric - the soc endpoint requires both the owner and the id and verifies them. i am not entirely sure this is needed to be done in this way
- i would probably just make a simple validation on both without having to leak too much about the internals of the chunk (like is done in the retrieval protocol for example) and leave the task of crafting a valid chunk to external tooling altogher

agazso:
> @significance thanks for this. a few comments:
> 
>     * the `/chunks` api endpoint was never made to upload SOCs
> 
>     * i am in favor of this api simplification, but in which case we should also get rid of the `soc` api path - it makes no sense to keep both and just increases confusion in the api
> 
>     * having a unified api would be better. but it would also require just having a unified type of upload requests. right now those are asymmetric - the soc endpoint requires both the owner and the id and verifies them. i am not entirely sure this is needed to be done in this way
> 
>     * i would probably just make a simple validation on both without having to leak too much about the internals of the chunk (like is done in the retrieval protocol for example) and leave the task of crafting a valid chunk to external tooling altogher

I think this is highly needed and it would be good to have a generic chunk upload endpoint. In my opinion what would make the most sense is to have a `/chunks` endpoint to upload any type of chunk applying the detection logic from the PR. 

This PR introduces a potential performance penalty when using only content-addressed chunks, because the signature would need to be validated even if the chunk is content-addressed. This could be mitigated by having a `/cac` endpoint (semantically similarly to the `/soc` endpoint) which could only accept content-addressed chunks. Alternatively the unified `/chunks` endpoint could have options, either with query parameters or headers to indicate which chunk type is being uploaded explicitly.

See related issues/mentions:
- https://github.com/snaha/swarm-id/issues/234
- https://github.com/ethersphere/bee/issues/5444

acud:
> This PR introduces a potential performance penalty when using only content-addressed chunks, because the signature would need to be validated even if the chunk is content-addressed. This could be mitigated by having a /cac endpoint (semantically similarly to the /soc endpoint) which could only accept content-addressed chunks. Alternatively the unified /chunks endpoint could have options, either with query parameters or headers to indicate which chunk type is being uploaded explicitly. 

I think that having a header would make sense, then we can have a unified interface and get rid of the extra endpoints. The question is whether we want to have opinionated soc validation like today in the current soc upload endpoint (for me this is a bit too much. not sure it needs to go that deep).

significance:
yes, must admit i did not consider the performance implication there, as i was in app dev mode and just scratching an itch/removing barriers : )

from bee pov header sgtm, thank you both 🙇

significance:
https://github.com/ethersphere/SWIPs/pull/67

btw @agazso @acud this is the longer term solution to this problem 

this is great for a one chunk solution but for efficiencies sake i would like to  pursue the chunk typing system proposed above and streaming presigned chunks en masse

significance:
and immediately running into the problem again because CAC ⊄ SOC, SOC ⊂ CAC

> When Swarm-Chunk-Type: cac is set but the payload is a valid SOC, the handler silently stores it at the wrong address.

there is no way to mitigate this without the chunk types

acud:
> [ethersphere/SWIPs#67](https://github.com/ethersphere/SWIPs/pull/67)
> 
> btw @agazso @acud this is the longer term solution to this problem
> 
> this is great for a one chunk solution but for efficiencies sake i would like to pursue the chunk typing system proposed above and streaming presigned chunks en masse

that SWIP is more about how we pass chunks around in the protocols. while it doesn't necessarily touch upon the point of API contracts (but rather wire protocol contracts), at least in theory - there's nothing preventing us from doing it in this way. the only thing to keep in mind is that once it leaves the wire protocol domain and enters the API, both upload and download must use the new _protobuf_ type.

significance:
> > [ethersphere/SWIPs#67](https://github.com/ethersphere/SWIPs/pull/67)
> > btw @agazso @acud this is the longer term solution to this problem
> > this is great for a one chunk solution but for efficiencies sake i would like to pursue the chunk typing system proposed above and streaming presigned chunks en masse
> 
> that SWIP is more about how we pass chunks around in the protocols. while it doesn't necessarily touch upon the point of API contracts (but rather wire protocol contracts), at least in theory - there's nothing preventing us from doing it in this way. the only thing to keep in mind is that once it leaves the wire protocol domain and enters the API, both upload and download must use the new _protobuf_ type.

so i guess i am putting the cart before the horse a bit there: in order for us to stream arbitrary chunk types we would need to send headers with the chunk, and this formalises them. perhaps we should add the stamp(s) into this too

i think this PR is gtg from my perspective. i have removed the ai canary and tests seem to pass. shall we proceed to review? 

very keen to get this one in the next release if i can so that i can release swapchat 3.0 😃

acud:
the chunk and soc apis already have the possibility to provide the stamp as a header. is that what you mean @significance?

significance:
@acud no, not exactly - the suggestion would be that a unified streaming chunk api could include both chunks and headers (with stamps and potentially chunk types) hence averting the current requirement for every individual presigned chunk to be posted separately

agazso:
> @acud no, not exactly - the suggestion would be that a unified streaming chunk api could include both chunks and headers (with stamps and potentially chunk types) hence averting the current requirement for every individual presigned chunk to be posted separately

This partially exist (for content-addressed chunks) with the websocket chunk endpoint, see this PR: https://github.com/ethersphere/bee/pull/5256

What is missing is the chunk type so that single-owner chunks can be posted as well.

significance:
just a note i've capitulated and ported swapchat 3.0 over to the soc api, so the urgency is gone from that perspective 🧦

acud:
since this appears to be at a halt, i'm closing this for now.

PR #5446, opened 2026-04-26
fix: make stamps removal sync and data race in push chunk tests

sbackend123:
### Checklist

### Description
1. Collect all matching *StampItem in a slice during the iteration, then Delete only after the iterator returns.

2. Separate problem: fix data race in TestPushChunkToNextClosest, TestMultiplePushesAsForwarder TestPushChunkToClosestErrorAttemptRetry tests. 

The record.bytes() method in streamtest returned a direct reference to the internal slice without holding the mutex, causing a data race when tests read stream records while background goroutines were still writing to them. Fixed by acquiring the lock and returning a copy of the slice.

3. Fix gsoc test timeout (2s too small for CI) and fix TestSubscribe 
**Context:**
`removeStampItems` deletes all `StampItem` rows for a given batch by prefix-iterating the stamper store and calling store.Delete for each entry. For a time this used a helper goroutine and an **unbuffered** channel: the iterator sent each `*StampItem` to the goroutine, which called Delete outside the callback.

That pattern was carried over from an earlier fire-and-forget SetExpired implementation (node startup) where the goal was not to block; it was not a good fit for the later synchronous API, `HandleStampExpiry`, where callers and tests expect all matching rows to be gone when the function returns.

**Bonus problem:**
Previous implementation may lead to deadlock. 

Many stores protect `Iterate` and `Delete` with a lock, often an RWMutex. 
With the goroutine + unbuffered channel pattern and more than one StampItem for the same batch prefix we may have scenario:

- The iterator’s callback does the first send on the channel. The consumer receives and calls Delete. Delete blocks: it must wait for the write lock, but the iterator still holds the read lock for the walk.
- The callback returns; the walk continues to the next key. The second send blocks: the unbuffered channel is empty but the consumer is stuck inside the first Delete (not reading the next value).
- The walk cannot complete while the callback is blocked on send. The read lock is not released until the walk finishes, so the first Delete never acquires the write lock.

### AI Disclosure

### DIFF
--- a/pkg/postage/service.go
+++ b/pkg/postage/service.go
@@ -290,37 +290,33 @@ func (ps *service) HandleStampExpiry(ctx context.Context, id []byte) error {
 	return nil
 }
 
-// removeStampItems
+// removeStampItems removes all stamp items belonging to the given batch.
 func (ps *service) removeStampItems(ctx context.Context, batchID []byte) error {
 	ps.logger.Debug("removing expired stamp items", "batchID", hex.EncodeToString(batchID))
 
-	deleteItemC := make(chan *StampItem)
-	go func() {
-		for item := range deleteItemC {
-			_ = ps.store.Delete(item)
-		}
-	}()
+	var toDelete []*StampItem
 
-	count := 0
-
-	defer func() {
-		close(deleteItemC)
-		ps.logger.Debug("removed expired stamps", "batchID", hex.EncodeToString(batchID), "count", count)
-	}()
-
-	return ps.store.Iterate(
+	err := ps.store.Iterate(
 		storage.Query{
 			Factory: func() storage.Item { return new(StampItem) },
 			Prefix:  string(batchID),
 		}, func(result storage.Result) (bool, error) {
-			select {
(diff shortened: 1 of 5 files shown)


gacevicljubisa (approved):

pkg/postage/service.go
```diff
@@ -290,37 +290,33 @@ func (ps *service) HandleStampExpiry(ctx context.Context, id []byte) error {
...
-			count++
+			toDelete = append(toDelete, result.Entry.(*StampItem))
 			return false, nil
 		})
+	if err != nil {
+		return err
+	}
+
+	for _, item := range toDelete {
+		_ = ps.store.Delete(item)
```
acud: here the error is swallowed, while other branches return an error. i would at least check whether an error occurred or not, and return an error (i'm not a big fan of `errors.Join` as a wall of errors is usually not very helpful - having perhaps just the first error returned is fine while continuing the execute the loop)

martinconic (approved):

akrem-chabchoub (approved):

PR #5447, opened 2026-04-26
fix: read exactly M values from channel

sbackend123:
### Checklist

### Description
Read exactly M values from channel instead of reading "forever", which leads to deadlock in really seldom cases, which happen in mostly in CI

`runStrategy` assumes that while consuming results from `c`, one of the two exit conditions will always become true before reads are exhausted. Because of that assumption, the previous loop used `for range c` (which only terminates when c is closed).

**Why this can deadlock**
In the DATA -> RACE fallback path, some DATA-owned shard fetches may still be finishing while RACE starts.
RACE can include data shard indices whose waits[i] is not closed yet; for those indices, fetch takes the fly=false path and waits on waits[i].
For a successful DATA fetch, the operation order is:

```
close(g.waits[i])       // 1) unblocks waiting fly=false RACE goroutine
g.fetchedCnt.Add(1)     // 2) increments success counter
```
There is a small window between (1) and (2):
- the waiting RACE goroutine can unblock and push its result into `c`,
- `runStrategy` can consume that last value from `c`,
- but `fetchedCnt` may still be stale for that check.

If this happens on the **last** available message in `c`, neither exit condition may trigger in that iteration, and the next read blocks forever (no more writers, channel not closed).

### AI Disclosure

### DIFF
--- a/pkg/file/redundancy/getter/getter.go
+++ b/pkg/file/redundancy/getter/getter.go
@@ -248,7 +248,8 @@ func (g *decoder) runStrategy(s Strategy) error {
 		}(i)
 	}
 
-	for range c {
+	for range len(m) {
+		<-c
 		if g.fetchedCnt.Load() >= int32(g.shardCnt) {
 			return nil
 		}
@@ -257,7 +258,7 @@ func (g *decoder) runStrategy(s Strategy) error {
 		}
 	}
 
-	return nil
+	return errStrategyFailed
 }
 
 // recover wraps the stages of data shard recovery:


pkg/file/redundancy/getter/getter.go
acud, line 247: this is a goroutine leak. when context is cancelled, `g.fetch` returns and always tries to write to `c`. when the `for range` loop exits prematurely below, it won't drain the channel, causing the goroutines waiting to write to leak. this needs to be a `select` case with a case for context expired.

acud (changes requested):

pkg/file/redundancy/getter/getter.go
sbackend123: May be I miss smth, but from my perspective there is no leak. 
There is complicated logic, but to simplify:
- `c` is buffered channel, with size `m`
- we create `m` goroutines and each  will return exactly 1 value. If context is cancelled, it will be just ctx.Err() value. So even if no goroutine is reading, writing goroutines wouldn't block.

acud (approved):

martinconic (approved):

akrem-chabchoub (approved):

gacevicljubisa (approved):

nugaon:
I fixed this problem accidentally as well with #5449 but that solves more, because on DATA strategy early return, the subsequent RACE strategy will not deal with stale counters. the `c` should be drained fully.

_Edit: though it makes sense to give a dedicated error return at the end of the block, even if the assumption should be correct now. that return never should be reached anyway and a named error could indicate it._

acud:
@nugaon I don't fully understand why the buffered channel is needed in the first place. This is an anti-pattern in go and is against our coding guidelines: [channel size should be one or none](https://github.com/ethersphere/bee/blob/master/CODINGSTYLE.md#channel-size-is-one-or-none). If we would respect context cancellation and not have these sort of returns that forever block on a channel (aka: have a `select` statement) we would not need this channel draining logic which I honestly find confusing.

nugaon:
on fan-in cases like this where we have exact number of go processes, creating the channel with the size of the routines could be considerable because it does not block at all.

we could make it with waitgroups and wait for all processes to return. 
using that is just an additional primitive to achieve this, making a bit more complex.

acud:
EDIT: i previously claimed there's a goroutine leak possible in the other PR and now i see this is not possible. i would still support the solution from @janos since it is more idiomatic go

janos (approved):
This PR successfully fixes the deadlock that it describes with the minimal changes to the current implementation.

Since the channel is buffered to exactly the number of goroutines, no goroutine leak would happen as every write to the channel would not block.

There is no need to drain the channel. The channel c and the results sitting in its buffer will be cleaned up by the garbage collector, when all goroutines finish.

But, the whole implementation here is questionable since the allocation of the whole buffered channel can be avoided. The approach that was taken here makes the getter allocate the channel, but also to spawn the m number of goroutines. For maximal speed for finding the replica, a large number of goroutines is prefered, but that does not scale linearly. I would assume that there can be a concurrency level that would be optimal for both retrieval speed and resources utilization.

Just as an example, a slightly different code is avoiding buffered channel and avoids the deadlock:

```go
	c := make(chan error)

	defer cancel()

	for _, i := range m {
		go func(i int) {
			err := g.fetch(ctx, i, false)
			select {
			case c <- err:
			case <-ctx.Done():
			}
		}(i)
	}

	for range len(m) {
		select {
		case <-c:
			if g.fetchedCnt.Load() >= int32(g.shardCnt) {
				return nil
			}
			if g.failedCnt.Load() > int32(allowedErrs) {
				return errStrategyFailed
			}
		case <-ctx.Done():
			return nil
		}
	}
```

And for my preference, I would protect the goroutines with a semaphore.

But as I said, this change is enough for fixing the problem.

nugaon:
If we use buffered channels, those should be drained not to release resources, but to settle the state (fetchedCnt, errorCnt).

The problem that #5448 describes is what if two fetches happen on the same resource (DATA to RACE strategy fallback) and this wrong accounting affects the another process. 
For me, it seems like this code change still do not prevent that.
on DATA strategy, one chunk missing, return with error `errStrategyFailed`, cancel context and early return -> go processes still can modify the state regarding those chunks and the other RACE strategy can wait for these chunks but they will get failedCnt bumped on those since DATA strategy context cancellation.

zelig:
I believe already #5505 is sufficient to fix the problem.

PR #5456, opened 2026-05-11
feat(tracing): migrate from OpenTracing/Jaeger to OpenTelemetry

### LINKED ISSUE #5010: docs: use of opentracing spans and jaeger with bee
### Location
N/A

 _Originally posted by @martinconic in [#5005](https://github.com/ethersphere/bee/issues/5005#issuecomment-2663930944)_

From my understanding, there is support within `bee` to collect tracing spans and that this was used in early development in conjunction with [Jaeger](https://www.jaegertracing.io/) for assisting with diagnosing / debugging node interactions. This is one of the main reasons why there are "headers" within the `libp2p` setup within `bee`. Having this capability - but no demonstration / documentation on how to use it - represents tangible limitations on new client developers (and perhaps some present client developers).

martinconic:
### Checklist

### Description
Migrates bee's tracing stack from the archived OpenTracing API and
  deprecated `jaeger-client-go` to the OpenTelemetry Go SDK with an
  OTLP exporter. The wrapper at `pkg/tracing` keeps a familiar API
  shape so the call-site sweep is bounded to ~8 production files plus
  the test suite.

  Closes #5010.

  ### ⚠️ Breaking change for operators using tracing

  **The Jaeger UDP agent on port `:6831` is no longer the receiver.** Operators
  must run an OTLP-capable collector (Jaeger v2, OpenTelemetry Collector, or any
  backend with an OTLP receiver) and point bee at it.

  The old `tracing-endpoint` / `tracing-host` / `tracing-port` keys are
  **deprecated, not rejected**: a node that still carries them starts normally,
  the values are ignored, and a warning is logged pointing operators at the new
  `tracing-otlp-*` options. This keeps existing configs bootable across the
  upgrade. The old values are not translated — operators relying on tracing must
  still migrate to the `tracing-otlp-*` options. The deprecated keys are hidden
  from `--help` and will be removed in a future release.

  #### Configuration flags

  Deprecated (accepted but ignored):
  - `tracing-endpoint` (replaced by `tracing-otlp-endpoint`)
  - `tracing-host` / `tracing-port` (folded into `tracing-otlp-endpoint`)

  Added:

  | Flag | Default | Meaning |
  |---|---|---|
  | `tracing-otlp-endpoint` | `127.0.0.1:4318` | OTLP collector address `host:port` (default port is `4318` for `http`, `4317` for `grpc`) |
  | `tracing-otlp-insecure` | `false` | Disable TLS for the OTLP exporter (suitable for a local collector). When `false`, set `tracing-otlp-ca-file` to verify the collector certificate against a private CA (otherwise the system root CAs are used) |
  | `tracing-otlp-ca-file` | `""` | Path to a PEM-encoded CA bundle used to verify the OTLP collector certificate; ignored when `tracing-otlp-insecure=true` |
  | `tracing-otlp-protocol` | `http` | OTLP transport: `http` or `grpc` |
  | `tracing-sampling-ratio` | `1.0` | Head-based sampling ratio in `[0, 1]`; `0` samples nothing, `1` samples everything |

  Unchanged: `tracing-enable`, `tracing-service-name`.

  #### Nested YAML form

  Tracing keys now also accept the nested form (mirroring `blockchain-rpc.*`):

  ```yaml
  tracing:
    enable: true
    otlp-endpoint: 127.0.0.1:4318
    otlp-insecure: true
    otlp-protocol: http
    sampling-ratio: 1.0
    service-name: bee
```
  Flat tracing-* keys still work; if both are present in the same config file,
  the nested form wins and a warning is logged.

####  Wire-format compatibility

  Spans propagated across libp2p streams use a new versioned binary carrier, and
  the standard W3C traceparent header is used over HTTP. Mixed-version peers
  will start fresh traces at the boundary rather than continuing the old Jaeger
  binary format.

 #### Coordination

  Requires a paired beekeeper update — the workflow at
  .github/workflows/beekeeper.yml temporarily points at
  BEEKEEPER_BRANCH: feat/bee-otlp-tracing-config and must be flipped back to
  master before merging this PR (and after the beekeeper PR lands).

  OpenTracing is archived and jaeger-client-go is deprecated; OpenTelemetry is
  the supported successor. Migrating now lets bee plug into any OTLP-capable
  backend (Jaeger v2, OTel Collector, vendor agents) and unlocks adopting
  semconv and tail-sampling at the collector layer in follow-up work.

### AI Disclosure

### CODE REVIEW

File: .github/workflows/beekeeper.yml [hunk 1]
```diff
@@ -19,7 +19,7 @@ env:
   SETUP_CONTRACT_IMAGE: "ethersphere/bee-localchain"
   SETUP_CONTRACT_IMAGE_TAG: "0.9.4"
   BEELOCAL_BRANCH: "main"
-  BEEKEEPER_BRANCH: "master"
+  BEEKEEPER_BRANCH: "feat/bee-otlp-tracing-config"
```
sbackend123: Before merge do not forget to switch beekeeper on master again

sbackend123 (commented):
I would explicitly call out in the PR description and in the docs which configuration flags were added / removed and that the tracing endpoint semantics changed. Without that, anyone still on the old configuration may be caught off guard.

pkg/tracing/tracing.go
```diff
@@ -2,278 +2,319 @@
...
-// StartSpanFromContext starts a new tracing span that is either a root one or a
-// child of existing one from the provided Context. If logger is provided, a new
-// log Entry will be returned with "traceID" log field.
-func (t *Tracer) StartSpanFromContext(ctx context.Context, operationName string, l log.Logger, opts ...opentracing.StartSpanOption) (opentracing.Span, log.Logger, context.Context) {
-	if t == nil {
-		t = noopTracer
+	ratio := o.SamplingRatio
+	if ratio <= 0 {
+		ratio = 1
+	}
```
gacevicljubisa: Why not allow 0? It can serve as a switch off... Flag is already setting it by default to 1.

cmd/bee/cmd/cmd.go
```diff
@@ -40,9 +40,9 @@ const (
 	optionWelcomeMessage                   = "welcome-message"
 	optionCORSAllowedOrigins               = "cors-allowed-origins"
 	optionNameTracingEnabled               = "tracing-enable"
-	optionNameTracingEndpoint              = "tracing-endpoint"
-	optionNameTracingHost                  = "tracing-host"
-	optionNameTracingPort                  = "tracing-port"
+	optionNameTracingOTLPEndpoint          = "tracing-otlp-endpoint"
+	optionNameTracingOTLPInsecure          = "tracing-otlp-insecure"
+	optionNameTracingSamplingRatio         = "tracing-sampling-ratio"
 	optionNameTracingServiceName           = "tracing-service-name"
```
gacevicljubisa, line 48: If we are changing this configuration, we can support nested flags as well for tracing:
```yaml
tracing:
  enable: true
  otlp-endpoint: ...
```

gacevicljubisa (commented):
> I would explicitly call out in the PR description and in the docs which configuration flags were added / removed and that the tracing endpoint semantics changed. Without that, anyone still on the old configuration may be caught off guard.

Yes, I agree. Updating the node with old config will result with "unknown flag" error.
Also, this is breaking change for the tracing for the node operators who are using tracing, they will need to switch to OTLP receiver, instead of jaeger UDP agent.

Additionally, OTel has a second exporter, `otlptracegrpc`. It might be worth to expose a `tracing-otlp-protocol: http|grpc` flag (defaulting to http), so operators can choose.

cmd/bee/cmd/cmd.go
```diff
@@ -281,9 +302,10 @@ func (c *command) setAllFlags(cmd *cobra.Command) {
 	cmd.Flags().Uint64(optionNameNetworkID, chaincfg.Mainnet.NetworkID, "ID of the Swarm network")
 	cmd.Flags().StringSlice(optionCORSAllowedOrigins, []string{}, "origins with CORS headers enabled")
 	cmd.Flags().Bool(optionNameTracingEnabled, false, "enable tracing")
-	cmd.Flags().String(optionNameTracingEndpoint, "127.0.0.1:6831", "endpoint to send tracing data")
-	cmd.Flags().String(optionNameTracingHost, "", "host to send tracing data")
-	cmd.Flags().String(optionNameTracingPort, "", "port to send tracing data")
+	cmd.Flags().String(optionNameTracingOTLPEndpoint, "127.0.0.1:4318", "OTLP endpoint to send tracing data (host:port); default port is 4318 for http, 4317 for grpc")
+	cmd.Flags().Bool(optionNameTracingOTLPInsecure, true, "disable TLS for the OTLP exporter (useful for a local collector)")
```
gacevicljubisa: We do have this `insecure` flag, and default is `false`, but what happens with default set to `true`? There are no flags that would allow specifying the CA...

pkg/tracing/tracing.go
```diff
@@ -2,278 +2,350 @@
...
-	var span opentracing.Span
-	if parentContext := FromContext(ctx); parentContext != nil {
-		opts = append(opts, opentracing.ChildOf(parentContext))
-		span = t.tracer.StartSpan(operationName, opts...)
-	} else {
-		span = t.tracer.StartSpan(operationName, opts...)
+	// If no OTLP endpoint is configured, omit the exporter entirely. Spans are
+	// still created with valid contexts (useful for local development and unit
+	// tests) but nothing is shipped over the network.
+	if o.Endpoint != "" {
```
gacevicljubisa: Endpoint should be required when Enabled, if Endpoint="" probably it will fiail silently.

cmd/bee/cmd/cmd.go
gacevicljubisa: Also, I would suggest that by default, this should be set to `false`, and if CA is not provided, log the warning where it would be described that CA is reqired, or to set to insecure, or to disable tracing.

gacevicljubisa (changes requested):

pkg/tracing/tracing.go
```diff
@@ -2,278 +2,389 @@
...
-	ErrContextNotFound = errors.New("tracing context not found")
+// ErrContextNotFound is returned when tracing context is not present in
+// p2p Headers, HTTP headers, or the go context.
+var ErrContextNotFound = errors.New("tracing context not found")
 
-	// noopTracer is the tracer that does nothing to handle a nil Tracer usage.
-	noopTracer = &Tracer{tracer: new(opentracing.NoopTracer)}
-)
+// LogField is the key in log message field that holds the tracing id value.
+const LogField = "traceID"
```
janos: This constant can be unexported and exported only for tests in export_test.go like this:

```go
const (
	LogField = logField
)
```

pkg/tracing/tracing.go
```diff
@@ -2,278 +2,389 @@
...
-// StartSpanFromContext starts a new tracing span that is either a root one or a
-// child of existing one from the provided Context. If logger is provided, a new
-// log Entry will be returned with "traceID" log field.
-func (t *Tracer) StartSpanFromContext(ctx context.Context, operationName string, l log.Logger, opts ...opentracing.StartSpanOption) (opentracing.Span, log.Logger, context.Context) {
-	if t == nil {
-		t = noopTracer
+// Supported OTLP transport values for Options.Protocol.
+const (
+	ProtocolHTTP = "http"
+	ProtocolGRPC = "grpc"
```
janos: ProtocolHTTP and ProtocolGRPC can be unexported.

janos (approved):

akrem-chabchoub (approved):

pkg/tracing/tracing.go
```diff
@@ -2,278 +2,389 @@
...
-// AddContextHTTPHeader adds a tracing span context to provided HTTP headers
-// from the go context. If the tracing span context is not present in
-// go context, ErrContextNotFound is returned.
-func (t *Tracer) AddContextHTTPHeader(ctx context.Context, headers http.Header) error {
-	if t == nil {
-		t = noopTracer
-	}
+// httpPropagator carries trace context across HTTP via the W3C TraceContext
+// standard headers (traceparent, tracestate).
+var httpPropagator propagation.TextMapPropagator = propagation.TraceContext{}
```
gacevicljubisa: maybe move to the top with the other package-level vars

pkg/tracing/tracing.go
```diff
@@ -2,278 +2,389 @@
...
+
+	res, err := resource.New(context.Background(),
+		resource.WithAttributes(semconv.ServiceName(o.ServiceName)),
+	)
+	if err != nil {
+		return nil, nil, fmt.Errorf("otel resource: %w", err)
+	}
+
+	ratio := o.SamplingRatio
+	if ratio < 0 {
```
gacevicljubisa, line 149: mybe to add: `if ratio > 1 { ratio = 1 }`

pkg/tracing/tracing.go
```diff
@@ -2,278 +2,389 @@
...
 }
 
-// StartSpanFromContext starts a new tracing span that is either a root one or a
-// child of existing one from the provided Context. If logger is provided, a new
-// log Entry will be returned with "traceID" log field.
-func (t *Tracer) StartSpanFromContext(ctx context.Context, operationName string, l log.Logger, opts ...opentracing.StartSpanOption) (opentracing.Span, log.Logger, context.Context) {
-	if t == nil {
-		t = noopTracer
+// Supported OTLP transport values for Options.Protocol.
+const (
```
gacevicljubisa: also, move this to the top

pkg/tracing/tracing.go
```diff
@@ -2,278 +2,389 @@
...
-	if err := w.Flush(); err != nil {
-		return err
-	}
+	ctx, span := t.tracer.Start(ctx, operationName, opts...)
+	return span, loggerWithTraceID(span.SpanContext(), l), ctx
+}
 
-	headers[p2p.HeaderNameTracingSpanContext] = b.Bytes()
+// AddContextHeader serialises the active span context into the bee p2p header.
+// It is safe to call on a nil receiver.
```
gacevicljubisa, line 288: Delete this claim from all methods, or add the gurd:

```
if t == nil {
        t = noopTracer
}
```
because Tracer is exposed and it can be used without constructor.

pkg/tracing/tracing.go
```diff
@@ -2,278 +2,389 @@
...
 
-	var span opentracing.Span
-	if parentContext := FromContext(ctx); parentContext != nil {
-		opts = append(opts, opentracing.FollowsFrom(parentContext))
-		span = t.tracer.StartSpan(operationName, opts...)
-	} else {
-		span = t.tracer.StartSpan(operationName, opts...)
+	if parent := FromContext(ctx); parent.IsValid() {
+		ctx = trace.ContextWithSpanContext(ctx, parent)
 	}
```
gacevicljubisa: FromContext reads using trace.SpanContextFromContext(ctx). Setting it with trace.ContextWithSpanContext(ctx, parent) writes to the same context key value pair. 

I think this can be removed?

gacevicljubisa (approved):
I have suggested some change, comments are mostly suggestions, not blockers, but overal looks good. It would be great if certificate loading logic is tested and validated. 

Also, please edit the PR desc, to match the implementation for `tracing-otlp-insecure` flag, and add `tracing-otlp-ca-file` to table.

pkg/tracing/tracing.go
```diff
@@ -2,278 +2,389 @@
...
+	// SamplingRatio is the head-based sampling ratio for the parent-based
+	// sampler in the range [0, 1]. 0 disables sampling for non-parented spans;
+	// 1 samples everything. Values outside [0, 1] are clamped to the nearest bound.
+	SamplingRatio float64
+	// Protocol selects the OTLP exporter transport: "http" or "grpc". Empty
+	// defaults to "http".
+	Protocol string
+	// Logger, when non-nil, receives startup warnings (e.g. missing CA bundle
+	// when TLS is enabled). NewTracer does not log otherwise.
+	Logger log.Logger
```
gacevicljubisa: Logger can be removed, because its only use is in the constructor to give a warning (which is not best practice), maybe warning can be moved to start.go before we call the NewBee constructor, where you have access to all of the flags, and this kind of validation can be done there.

gacevicljubisa (changes requested):
Change overall look good. Configuration backwards compatibility is there.

But I think we should change the `p2p.HeaderNameTracingSpanContext = "tracing-span-context-v2"`, because it can happen that "new nodes" will communicate with "old ones", and we should have different key becuase of the difference in the payload.

gacevicljubisa (approved):
Before merging, it would be good to test the changes with OTEL collector (HTTP or GRPC), to check if traces are as we expect them. But overall it looks good.

acud:
@martinconic please let us know the results of manual testing here before merging

darkobas2:
> 🤖 **AI-assisted review** (Claude Code, with a Gemini second opinion). Suggestions for the visibility work — not blockers; the migration itself looks clean.

**Trace visibility / observability — prioritized:**

**High — enrich the OTel resource** (`pkg/tracing/tracing.go`, `NewTracer`)
The resource currently carries only `service.name`, so in any multi-node deployment every node's spans look identical in an OTLP backend — per-node trace search and the service graph collapse together. Worth adding:
- `service.instance.id` = the node's overlay address — highest value; identifies *which* node a trace came from. Note `NewTracer` is constructed before the overlay is known in `NewBee`, so it'd need threading in (or constructing the tracer after the overlay is derived).
- `service.version` (version + commit) — correlate trace behavior with releases.
- `deployment.environment` — derivable from the network id.
- `resource.WithFromEnv()` (+ `WithTelemetrySDK`/`WithHost`) — lets operators enrich via `OTEL_RESOURCE_ATTRIBUTES` (e.g. cluster/region) without a rebuild.

**High/Med — batch span processor uses all defaults** (`WithBatcher(exporter)`)
Default queue (2048)/batch/timeout can silently drop spans under load on hot paths (pushsync/retrieval), especially with the default sampling ratio of `1.0` — and a dropped span gives no signal. Consider exposing max-queue/timeout knobs, or documenting that `OTEL_BSP_*` env vars are honored.

**Med — harden the OTLP exporter** (`newOTLPClient`)
Add `WithCompression(gzip)` (cheap bandwidth win at scale); optionally `WithTimeout`/`WithRetry` for resilience to collector blips.

**Med — a few bare spans**
`pkg/storer/netstore.go` get/put spans carry only `success` — add the chunk `address`. `pingpong` spans lack a peer-address attribute. (Error recording via `RecordError`/`SetStatus` across pushsync/retrieval/pusher/api is consistent — nicely done.)

**Low / follow-up** — no spans yet on postage/kademlia/salud (postage would be the most useful next add for an end-to-end "where did this upload go"); baggage isn't propagated (only W3C TraceContext) — adding it would let a key like `batch_id` follow an upload across hops.

**Nice touches:** the versioned binary p2p span carrier (clean rolling-upgrade story), the configurable parent-based sampler with correct cross-node sampled-flag propagation, and a genuinely zero-cost no-op path when tracing is disabled.

Overall solid and ready — the one I'd suggest landing *with* this PR is the resource-attribute set (`service.instance.id` / `service.version` / `WithFromEnv`), since it's what makes traces distinguishable per-node in the backend.

gacevicljubisa:
@martinconic maybe it would help if we have a log of info level that confirms tracing is actually enabled and wired up

gacevicljubisa:
@darkobas2 any suggestion regarding the flags for this setup? would it be better to leave the same name `tracing-endpoint` as it was till now, because currently in this PR, we are deprecating `tracing-endpoint` and introducing the new set of flags. Currently in this PR:

- "tracing-otlp-endpoint"
- "tracing-otlp-insecure"
- "tracing-otlp-ca-file"
- "tracing-otlp-protocol"
- "tracing-sampling-ratio"

darkobas2:
@gacevicljubisa — I'd **drop the `otlp` infix** and go with a generic `tracing-*` family instead of `tracing-otlp-*`:

`tracing-endpoint`, `tracing-insecure`, `tracing-ca-file`, `tracing-protocol`, `tracing-sampling-ratio`

Why:
- OTLP is a transport detail; the flag concept is just "where traces go". Generic names are future-proof if the transport ever changes again, and avoid the awkward mix of `tracing-endpoint` (no prefix) sitting next to `tracing-otlp-protocol` (prefixed).
- Keeps the familiar `tracing-endpoint` name — less doc/muscle-memory churn.

The one catch to handle: `tracing-endpoint` changes **meaning** — it used to be a Jaeger agent address (UDP, e.g. `:6831`), now it's an OTLP collector endpoint (gRPC `:4317` / HTTP `:4318`). Same name, different required value = silent-breakage risk (a node keeps starting but points OTLP at a dead Jaeger port and quietly drops traces). So if we reuse the name, please pair it with:
1. A startup **info log** confirming tracing is wired up — `tracing enabled → endpoint=… protocol=… sampling=…`, plus a clear error if the exporter can't reach the endpoint. (This also covers your other comment about confirming tracing is actually enabled.)
2. A migration note in the changelog/release notes (Jaeger UDP → OTLP; ports change).

That keeps the naming clean *and* makes the semantic change loud instead of silent. No strong opinion on the `tracing-protocol` default (grpc vs http) — either's fine; http/4318 is marginally easier to front with standard proxies.

acud:
@martinconic can we please address PR reviews inside the PR and not open subsequent PRs? can you please squash the other PR into this once so we could follow up on the requested changes? thanks

PR #5463, opened 2026-05-13
fix(storer): unstick radius decrement on live networks

### LINKED ISSUE #5396: storageRadius decrease logic too restrictive on active networks
## Summary

The reserve worker only decreases `storageRadius` when `SyncRate() == 0`, which is unrealistic on active networks where pullsync never fully stops due to continuous uploads. Combined with the 15-minute check interval and single-step decrements, this causes nodes to report a stale (too high) storageRadius even when their reserve already contains enough chunks to justify a lower radius.

## Code Reference

`pkg/storer/reserve.go`, `reserveWorker()`:

```go
case <-thresholdTicker.C:
    radius := db.reserve.Radius()
    count, err := db.countWithinRadius(ctx)
    // ...
    if count < threshold(db.reserve.Capacity()) && db.syncer.SyncRate() == 0 && radius > db.reserveOptions.minimumRadius {
        radius--
        // ...
    }
```

Three constraints compound the problem:

1. **`SyncRate() == 0` gate** — On an active network, pullsync never fully stops. New data is continuously being uploaded, so there's always some sync activity. This condition blocks radius adjustment indefinitely.

2. **15-minute ticker** (`reserveWakeUpDuration = 15 * time.Minute`) — Even when conditions are met, the radius can only decrease by 1 every 15 minutes.

3. **Single-step decrement** (`radius--`) — If the radius needs to drop by several levels (e.g., after a restart with batchstore reset), adaptation takes `N * 15 minutes`.

## Observable Impact

After a node restart with batchstore resync, the radius starts high and needs to decrease to match the node's actual reserve capacity. Example:

- Node has ~115M chunks in reserve (near full capacity for its configuration)
- `storageRadius` is stuck at 5 instead of the correct value of 4
- `pullsyncRate` is ~8 chunks/sec (normal background sync activity)
- Radius never decreases because `SyncRate() != 0`

This causes:
- **Incorrect `committedDepth` reported to peers** via the status protocol, which feeds into salud's network radius consensus
- **Node marked unhealthy by salud** due to `committedDepth` mismatch (`self_radius` vs `network_radius`)
- **Potential redistribution game issues** — node plays with wrong radius, risking incorrect samples

## Nodes with `reserve-capacity-doubling` Are More Affected

Doubled nodes cover more neighborhoods and receive proportionally more chunk offers via pullsync. Their `SyncRate()` is structurally higher than non-doubled nodes, making the `== 0` condition even harder to satisfy.

## Suggestion

The radius adjustment could be based on the relationship between `reserveSize` and `capacity` rather than requiring zero sync activity. For example:

- **Remove or relax the `SyncRate() == 0` gate** — perhaps use a threshold relative to the node's capacity or check that sync rate is *stable* rather than zero
- **Allow multi-step radius adjustment** — if the reserve size clearly justifies a lower radius, skip intermediate steps
- **Reduce the check interval** — 15 minutes is very coarse for a value that affects network consensus

The radius *increase* path (in `unreserve()`) already operates without a sync rate gate, so there's precedent for the radius responding to actual reserve state rather than sync activity.

martinconic:
### Checklist

### Description
  Nodes whose local `storageRadius` drifts above the neighborhood consensus stay stuck there indefinitely. Salud (`pkg/salud/salud.go:243`) compares `CommittedDepth = capacityDoubling + Radius` against the network mode with zero tolerance, flips the node unhealthy on any mismatch, and redistribution stops playing rounds. The reporter of #5428 observed this on bee 2.7.1-rc2 after an RPC instability event; the reporter of #5396 reproduced the same condition with code-level analysis.
  
  The reserve worker's radius-decrement gate `db.syncer.SyncRate() == 0` is structurally unreachable on a live mainnet. SyncRate() measures historical sync only, but historical sync never goes quiet for the 15-minute rate window because (1) peer churn keeps starting new cursor-based historical workers and (2) every successful decrement calls puller.onChange → resetIntervals(prevRadius) which discards non-neighbor sync history and forces re-historical-sync, bumping SyncRate above zero for the next ~15-30 min. The gate locks closed permanently. Nodes whose local Radius drifts above the neighborhood consensus stay there indefinitely — salud (zero-tolerance comparison of CommittedDepth = capacityDoubling + Radius) flips the node unhealthy and redistribution stops playing rounds. Restart doesn't help because waitNetworkRFunc re-loads the wrong local value.

  Capacity-doubled nodes are structurally worse off: they pull more data, their SyncRate floor is higher, and the +1 from doubling amplifies any Radius mismatch into CommittedDepth mismatch.

  Three changes, applied together:

  1. Remove the SyncRate gate. The threshold check (count < threshold(capacity)) is already a correctness signal — if less than half of capacity sits within radius, the radius is wrong, regardless of sync activity. The unreserve path raises radius without consulting SyncRate; the asymmetry was unjustified.

  2. Allow a two-step decrement when count is far below threshold (count*4 <= threshold). Under approximately uniform CAC bin distribution, each step doubles count-within-radius, so two steps land at or below threshold rather than overshooting. Capped at two to keep behavior bounded; each radius change triggers one resetIntervals so multi-step in one tick is strictly cheaper than the equivalent single-step ticks 5 minutes apart.

  3. Reduce reserveWakeUpDuration from 15 to 5 minutes. With the gate gone, faster ticks bound worst-case recovery; tick cost is one countWithinRadius iteration.

  The existing test "radius doesn't change due to non-zero pull rate encoded the buggy behavior as an invariant. Renamed and inverted to assert the new contract: with an empty reserve, the worker decrements toward minimumRadius regardless of the syncer's reported rate.

  No protocol, storage layout, persistence format, or interface changes. The Syncer.SyncRate() interface stays — it's still used for "fully synced" determination at startup and for the status protocol's pullsyncRate field, both correct uses.

  Fixes #5396
  Fixes #5428

### AI Disclosure

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -205,7 +205,7 @@ const (
 	minPaymentThreshold           = 2 * refreshRate           // minimal accepted payment threshold of full nodes
 	maxPaymentThreshold           = 24 * refreshRate          // maximal accepted payment threshold of full nodes
 	mainnetNetworkID              = uint64(1)                 //
-	reserveWakeUpDuration         = 15 * time.Minute          // time to wait before waking up reserveWorker
+	reserveWakeUpDuration         = 5 * time.Minute           // time to wait before waking up reserveWorker
 	reserveMinEvictCount          = 1_000
 	cacheMinEvictCount            = 10_000
 	maxAllowedDoubling            = 1
--- a/pkg/storer/reserve.go
+++ b/pkg/storer/reserve.go
@@ -167,20 +167,43 @@ func (db *DB) reserveWorker(ctx context.Context, ready chan<- struct{}) {
 		case <-thresholdTicker.C:
 
 			radius := db.reserve.Radius()
+			if radius <= db.reserveOptions.minimumRadius {
+				continue
+			}
 			count, err := db.countWithinRadius(ctx)
 			if err != nil {
 				db.logger.Warning("reserve worker count within radius", "error", err)
 				continue
 			}
 
-			if count < threshold(db.reserve.Capacity()) && db.syncer.SyncRate() == 0 && radius > db.reserveOptions.minimumRadius {
-				radius--
-				if err := db.reserve.SetRadius(radius); err != nil {
-					db.logger.Error(err, "reserve set radius")
-				}
-				db.metrics.StorageRadius.Set(float64(radius))
-				db.logger.Info("reserve radius decrease", "radius", radius)
+			t := threshold(db.reserve.Capacity())
+			if count >= t {
+				continue
 			}
+
+			// Decrement the storage radius. The decrement is gated only on
+			// the reserve fill state (count < threshold) and the operator's
+			// minimum-radius floor. There is no sync-rate gate here, which
+			// mirrors the unreserve path that raises radius without
+			// consulting sync activity. A previous SyncRate() == 0 gate
+			// proved structurally unreachable on live networks: peer churn
+			// kept historical sync above zero, and the resetIntervals call
+			// in puller.onChange (triggered by this very radius change)
+			// retriggered historical sync, locking the gate closed (issues
+			// #5396, #5428). When count is well below threshold, jump by
+			// two steps to keep adjustments bounded but recover faster from
+			// large gaps; under uniform CAC bin distribution each step
+			// roughly doubles count-within-radius.
+			steps := uint8(1)
+			if count*4 <= t && radius-1 > db.reserveOptions.minimumRadius {
+				steps = 2
+			}
+			radius -= steps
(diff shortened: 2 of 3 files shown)


acud:
would like to get input from @significance on this

significance:
@acud any specific angle? salud has to a large extent escaped my scrutiny so far, but i can dig in sooner rather than later if needed 😌

sbackend123 (commented):
I would add test for case radius < minimum_radius just to make sure nobody would override it is future

martinconic:
> I would add test for case radius < minimum_radius just to make sure nobody would override it is future

Added, thank you.

sbackend123 (approved):

martinconic:
Closing this since it will be addressed in another PR

Issue #5114, opened 2025-06-11
Add Automated Gas Fee Retry Mechanism with Dynamic Fee Estimation to Bee


gacevicljubisa:
#### Summary

Add an automated gas fee retry mechanism to Bee for Ethereum transactions (e.g., postage batch creation, redistribution for staked nodes) to prevent delays due to low gas fees. Use `eth_feeHistory` RPC to dynamically estimate low, market, and aggressive fees, starting with a market fee and retrying up to 5 times with a 20% gas price increase if unconfirmed after 1 minute. Provide configurable retry parameters and retain header-based gas overrides for developers via the Bee API.

#### Motivation

Transactions in Bee, such as postage batch creation for developers and redistribution for staked node operators, can be delayed or stuck in the mempool due to low gas fees, especially during network congestion. This frustrates users, disrupts app functionality, and prevents node operators from participating in the redistribution game, reducing storage incentives. An automated retry mechanism with dynamic fee estimation ensures reliable transaction confirmation, improves user experience, and reduces the need for manual gas management. Developers benefit from a robust default behavior, while node operators maintain consistent participation. Optional header overrides via the Bee API preserve flexibility for advanced use cases.

#### Implementation

1. Fetch Dynamic Fees:

    - Use `eth_feeHistory` RPC to retrieve base fees and priority fees (10th, 50th, 90th percentiles) for the last 100 blocks.
    - Calculate: Low = base fee + 10th percentile priority fee, Market = base fee + 50th percentile priority fee, Aggressive = base fee + 90th percentile priority fee.
    - Use `eth_estimateGas` to determine the gas limit for each transaction.

2. Retry Logic:

    - On transaction submission, start with the Market fee.
    - If unconfirmed after 1 minute, resubmit with the same nonce but a 20% higher gas price, up to 5 attempts (5-minute timeout).
    - Log each retry (timestamp, gas price, nonce) for debugging.
    - If all attempts fail, return a clear error (e.g., “Transaction failed after 5 attempts due to network congestion. Please try again.”).

3. Configuration:

    - If needed add retry parameters to Bee’s config (e.g., maxRetries, retryDelay, gasIncreasePercent, maxGasPrice).
    - Default settings: 5 retries, 1-minute delay, 20% gas increase, no max gas price.

4. Header Overrides:

    - Retain support for Gas-Price and Gas-Limit headers in the Bee API for developers.
    - Headers override default retry logic only if explicitly set (e.g., with a “disable-retry” flag).

[...]

PR #5464, opened 2026-05-18
feat: transactions with retry

sbackend123:
### Checklist

### Description
Add automatic gas-fee retry for Ethereum transactions in Bee. Transactions that stay unconfirmed are re-broadcast with the same nonce and an escalated priority fee, using dynamic fees from eth_feeHistory. Retry behaviour is configurable, survives node restarts, and is used for redistribution and postage operations.

[Diagram in miro](https://miro.com/app/board/uXjVHSDoqRM=/?share_link_id=82358424075) 

**What changed**
1. Transaction retry (pkg/transaction)
New SendWithRetry on the transaction service: sends EIP-1559 transactions with dynamic fee estimation from eth_feeHistory and automatic escalation across priority tiers. Requires automatic gas pricing: a request carrying an explicit Gas-Price is rejected.

Fees are derived from fresh fee history (default window: last 100 blocks) at percentiles 10 / 50 / 90, mapped to tiers low / market / aggressive. maxFeePerGas = 2 × baseFee + tip.

Sending walks a tier range [start_tier … end_tier]. Each tier gets 2 broadcast attempts (internal constant). If no receipt arrives within the retry delay, the tx is rebroadcast — reusing the same nonce — at the same tier (with fresh fee history) or escalated to the next tier once the per-tier budget is spent.

Persists retry state in the state store and resumes in-flight retries after restart, skipping nonces already confirmed on-chain.

Stops on non-retryable errors.

2. Configuration (cmd/bee, pkg/node)
CLI flags (defaults: start `market`, ceiling `aggressive`, 1 min delay, no price cap):

--transaction-fee-priority — starting fee tier (low / market / aggressive), default market
--transaction-fee-priority-max — escalation ceiling tier, default aggressive
--transaction-retry-delay — wait for a receipt before escalating, default 1m
--transaction-fee-max-tx-price-wei — max maxFeePerGas in wei, 0 = no limit
--fee-history-block-count — fee-history window depth, default 100

Constraint: priority <= priority-max, otherwise the node fails to start. Number of attempts per tier and the +15% bump are internal, not exposed as flags.

3. Call sites
Redistribution (pkg/storageincentives/redistribution): commit / reveal / claim use SendWithRetry via sendAndWait. Config-only — no header overrides.

Postage (pkg/postage/postagecontract): batch create, top-up, dilute, approve use SendWithRetry when retry is not disabled; otherwise fall back to Send + WaitForReceipt.

More detailed description is in [doc](https://docs.google.com/document/d/1D7kqHVKA_gPEUDBMi0sYlQVGt1braenGUZSYw3U6vOI/edit?tab=t.0#heading=h.ui40psetqep)

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5114

### AI Disclosure


pkg/node/node.go
```diff
@@ -128,71 +128,89 @@ type Bee struct {
...
+	TracingEnabled                     bool
+	TracingEndpoint                    string
+	TracingServiceName                 string
+	TrxDebugMode                       bool
+	WarmupTime                         time.Duration
+	WelcomeMessage                     string
+	WhitelistedWithdrawalAddress       []string
+}
+
+func txRetryConfigFromOptions(o *Options) transaction.TransactionsRetryConfig {
```
acud: since we already have the same logic in the `cmd` package (which precedes this package execution on runtime), wouldn't it make sense to build the config just once and pass it to this package as the right type?

pkg/transaction/wrapped/wrapped.go
```diff
@@ -20,37 +21,58 @@ import (
...
 	if blockSyncInterval == 0 {
 		blockSyncInterval = 1
 	}
 
+	if feeHistoryBlockCount == 0 {
+		feeHistoryBlockCount = feeHistoryDefaultBlockCount
+	}
+
+	var rewardPerc []float64
+	if len(feeHistoryRewardPercentiles) >= 3 {
```
acud: i find this a bit confusing. why do we need this? why not just inject the default value and eliminate the choices here?

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,525 @@
...
+
+// escalateGasTip returns tip * (100+increasePct)/100 — a single escalation step.
+func escalateGasTip(tip *big.Int, increasePct int) *big.Int {
+	if tip == nil {
+		return nil
+	}
+	return new(big.Int).Div(new(big.Int).Mul(new(big.Int).Set(tip), big.NewInt(int64(100+increasePct))), big.NewInt(100))
+}
+
+// suggestGasFeeGasTipCapWithHistory returns maxFeePerGas (gasFeeCap) and maxPriorityFeePerGas (gasTipCap)
```
acud: slight headache from this method: long variable names and this comment that tries to explain it all makes it really difficult to read for the amount of lines that it has. i'd rather convert this comment to inline comments explaining the branching, and reduce the variable/method names in the PR in general (adding more words to the method name doesn't necessarily add value when reading it)

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,525 @@
...
+	}
+	_ = t.store.Delete(retryKey)
+	for _, h := range state.AllTxHashes {
+		_ = t.store.Delete(pendingTransactionKey(h))
+	}
+	if state.LastTxHash != (common.Hash{}) {
+		_ = t.store.Delete(pendingTransactionKey(state.LastTxHash))
+	}
+}
+func (t *transactionService) retry(ctx context.Context, txRetryKey string, request *TxRequest) (common.Hash, *types.Receipt, error) {
```
acud: this is really hard to review. i'm not sure about this. also, i'm not sure i fully understand why statestore persistence is needed. do we really need to persist all of the data every time we submit a transaction? apart from the nonce and the transaction data, there's no guarantee that the extra data persisted around a transaction will be relevant once a node restarts (it can just be again inferred via RPC calls and configuration)

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,525 @@
...
+		prevGasTipCap = fh.MarketTip
+	} else {
+		escalatedGasTip = escalateGasTip(prevGasTipCap, t.txRetryGasIncreasePercent)
+	}
+
+	gasFeeCap = new(big.Int).Mul(header.BaseFee, big.NewInt(2))
+	gasFeeCapWithEscalatedTip := new(big.Int).Add(new(big.Int).Set(gasFeeCap), escalatedGasTip)
+	gasFeeCapWithPreviousTip := new(big.Int).Add(new(big.Int).Set(gasFeeCap), prevGasTipCap)
+
+	t.logger.V(1).Register().Debug("suggest gas fees for retry",
```
acud: why do we need the register call? why can't you just use `t.logger`

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,525 @@
...
+
+	for attempt := txState.NextAttempt; attempt < t.txMaxRetries; attempt++ {
+		if txState.NonceAssigned {
+			nonce = &txState.Nonce
+		}
+
+		signedTx, err := t.broadcastTx(ctx, request, nonce, txState.PreviousTip, attempt)
+		if err != nil {
+			if isErrCritical(err) {
+				t.logger.Error(err,
```
acud: why are you using `t.logger` here and not `loggerV1`, not sure why both usages are needed

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,525 @@
...
+				"attempt", attempt,
+				"tx_hash", txState.LastTxHash,
+				"nonce", txState.Nonce,
+				"wait_error", waitErr,
+				"description", request.Description)
+		}
+	}
+
+	exhaustionErr := fmt.Errorf("transaction failed after %d attempts (nonce=%d, description=%s)", t.txMaxRetries, txState.Nonce, txState.Description)
+	t.logger.Error(exhaustionErr,
```
acud: quite a few cases here of log and return error which is against the style guide. in general a bit too much logging i'd say, not sure if this is needed for merging into trunk but ok for now if it is needed for debugging purposes

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,525 @@
...
+			txState.Value = signedTx.Value()
+		}
+	}
+	if txState.NonceAssigned {
+		return t.store.Put(retryStateKey(txState.Nonce), txState)
+	}
+	return nil
+}
+
+func isErrCritical(err error) bool {
```
acud: `notRetriable`?

acud (commented):
lots and lots of code here, i think some complexity can be reduced and readability can be improved

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,576 @@
...
+
+func retryStateKey(nonce uint64) string {
+	return fmt.Sprintf("%s%020d", retryStatePrefix, nonce)
+}
+
+// SendWithRetry sends an EIP-1559 transaction using fee-history tiers with automatic
+// escalation. Each tier gets attemptsPerTier broadcast rounds with fresh eth_feeHistory
+// data. A +15% mempool bump floor is applied to ensure replacement transactions are accepted.
+func (t *transactionService) SendWithRetry(ctx context.Context, request *TxRequest) (txHash common.Hash, receipt *types.Receipt, err error) {
+	if request.GasPrice != nil {
```
martinconic, line 46: Gas-Price header is broken for postage stamp ops. SendWithRetry rejects any request with an explicit GasPrice (send_tx_with_retry.go:51-56), but the postage call sites (postagecontract/contract.go:168 and :212) still copy GasPrice from the header and route to SendWithRetry by default. So POST /stamps/... with Gas-Price set and no Disable-Retry always fails  and since the error isn't a matched sentinel, it surfaces as an opaque HTTP 500 instead of a 400. This is a regression (the old Send path honored Gas-Price) and it affects bee-js.

martinconic (changes requested):
This PR adds new API headers  like Disable-Retry and Fee-Priority, but they're not documented in the OpenAPI spec. Please add all necessary parameters in the corresponding yaml files

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,576 @@
...
+	attempt := 1 // start attempts counting from 1 for metrics
+	defer func() {
+		if terminateTxErr != nil {
+			t.logger.Error(terminateTxErr,
+				"send with retry: finished with error",
+				"attempt", attempt,
+				"tx_hash", txState.LastTxHash,
+				"nonce", txState.Nonce,
+				"to", request.To.String(),
+				"value", request.Value.String(),
```
martinconic: Both `To` and `Value` and will cause a panic if `nil`, right?

pkg/transaction/send_tx_with_retry.go
sbackend123: Thanks for catching that.

I see three possible approaches:

1. Make the error explicit and add Disable-Retry usage in bee-js when Gas-Price is set. This is a breaking change, but on the other hand we are introducing new functionality.

2. If Gas-Price is set, automatically route to Send without retry.

3. Keep the old behavior as the default, and instead of a Disable-Retry flag, introduce the opposite — e.g. a With-Retry opt-in flag.

Given existing clients, option 3 seems the most logical to me. I don’t like option 2 because it’s implicit behavior, and we may end up having to revisit it in a later release anyway — which would likely bring us back to option 1.

@gacevicljubisa WDYT?

pkg/transaction/send_tx_with_retry.go
sbackend123: We discussed this briefly on the call, but leaving it here for the record:

I think persistence is needed because:

1. We choose the next transaction nonce based in part on the transactions we are currently tracking. In this case, we cannot avoid persisting the transaction altogether. And we cannot persist it only once — with each rebroadcast the hash changes because the fees change, and from a monitoring perspective we would end up with several “stuck” transactions sharing the same nonce.

2. With the default 1 min delay, in the worst case — even with just two adjacent tiers (low → market, market → aggressive) and a price spike — retries can take 3–4 minutes. If the node is shut down during that window, the transaction may already be in the mempool and even mined, but user would have no way to know. As a result, user would send it again and lose fees, and that is exactly the scenario we are trying to avoid in the first place.

pkg/transaction/send_tx_with_retry.go
gacevicljubisa: Gas-Price is a spend ceiling, retry contradicts it, and it should act as a client override. I would go with option 2. I would even consider droping the Disable-Retry, if you set Gas-Price you've already opted out of retry. If Gas-Price is not used, pricing is given to the node and retry is just part of that. But we should document this in Swarm.yaml that Gas-Price disables escalation.

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,586 @@
...
+
+	tip := tierTip(tier, fh)
+
+	// Apply mempool bump floor: replacement tx must pay at least +15% over previous.
+	if previousTip != nil && previousTip.Sign() > 0 {
+		bumpedTip := applyMempoolBump(previousTip)
+		if tip.Cmp(bumpedTip) < 0 {
+			tip = bumpedTip
+		}
+	}
```
gacevicljubisa, line 88: The same floor should apply to the fee cap as well, not only for the tip?

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,586 @@
...
+		start = parsed
+	}
+	return tierRange(start, t.endTier), nil
+}
+
+func (t *transactionService) prepareTransactionForTier(ctx context.Context, request *TxRequest, nonce uint64, tier feeTier, previousTip *big.Int) (*types.Transaction, error) {
+	gasLimit, err := t.estimateGasLimit(ctx, request)
+	if err != nil {
+		return nil, err
+	}
```
gacevicljubisa: Previous attempt for gas limit as well can be reused

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,586 @@
...
+		return
+	}
+	_ = t.store.Delete(retryKey)
+	for _, h := range state.AllTxHashes {
+		_ = t.store.Delete(pendingTransactionKey(h))
+	}
+	if state.LastTxHash != (common.Hash{}) {
+		if !keepLast {
+			_ = t.store.Delete(pendingTransactionKey(state.LastTxHash))
+		}
```
gacevicljubisa: if keepLast, should we call `t.waitForPendingTx(state.LastTxHash)` because the last one may still be mined?

pkg/storageincentives/redistribution/redistribution.go
```diff
@@ -18,7 +19,8 @@ import (
 )
 
 const (
-	loggerName      = "redistributionContract"
+	loggerName = "redistributionContract"
+	// BoostTipPercent is used where the node sends transactions without retry.
 	BoostTipPercent = 50
```
acud, line 23: this is now just used in one place (agent.go) and i wonder if that callsite also needs to be migrated or if it is covered in a subsequent PR

pkg/transaction/wrapped/fee_history.go
```diff
@@ -0,0 +1,51 @@
...
+		if idx >= len(row) {
+			continue
+		}
+		if row[idx] == nil {
+			continue
+		}
+		vals = append(vals, new(big.Int).Set(row[idx]))
+	}
+	if len(vals) == 0 {
+		return big.NewInt(0)
```
acud: not to slow things down about this, but if there's no entries here, does it mean that all subsequent transactions sent using this quantile's tip value will fail because the value they're using is `0`?

pkg/transaction/fee_tier.go
```diff
@@ -0,0 +1,79 @@
...
+	case feeTierAggressive:
+		return "aggressive"
+	default:
+		return "unknown"
+	}
+}
+
+// ParseFeePriority validates a fee priority tier name (for API headers and config).
+func ParseFeePriority(s string) (string, error) {
+	t, err := parseFeeTier(s)
```
acud: what's this function needed for? why can't you just inline the inner unexported function here?

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,586 @@
...
+}
+
+func retryStateKey(nonce uint64) string {
+	return fmt.Sprintf("%s%020d", retryStatePrefix, nonce)
+}
+
+// SendWithRetry sends an EIP-1559 transaction using fee-history tiers with automatic
+// escalation. Each tier gets attemptsPerTier broadcast rounds with fresh eth_feeHistory
+// data. A +15% mempool bump floor is applied to ensure replacement transactions are accepted.
+func (t *transactionService) SendWithRetry(ctx context.Context, request *TxRequest) (txHash common.Hash, receipt *types.Receipt, err error) {
```
acud, line 45: nit: i have a feeling this is a little bit misplaced. also, the fact that another metrics collector has been created here (which also partially overlaps with the metrics collector in the `wrapped` package smells like it should really be there alongside, and you should just reuse the existing metrics collector there.

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,586 @@
...
+
+func (t *transactionService) prepareTransactionForTier(ctx context.Context, request *TxRequest, nonce uint64, tier feeTier, previousTip *big.Int) (*types.Transaction, error) {
+	gasLimit, err := t.estimateGasLimit(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	gasFeeCap, gasTipCap, err := t.suggestGasFeeForTier(ctx, tier, previousTip)
+	if err != nil {
+		return nil, err
```
acud: so if the suggest gas fee failed for whatever reason, then the whole transaction is fail and not even sent? there is a soft-fail there which fails when `fh` is empty

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,586 @@
...
+// When fixedNonce is nil a new nonce is allocated (first attempt);
+// otherwise the supplied nonce is reused (replacement transaction).
+func (t *transactionService) broadcastTx(ctx context.Context, request *TxRequest, fixedNonce *uint64, tier feeTier, previousTip *big.Int) (*types.Transaction, error) {
+	var nonce uint64
+
+	if fixedNonce != nil {
+		nonce = *fixedNonce
+	} else {
+		t.lock.Lock()
+		defer t.lock.Unlock()
```
acud, line 135: danger... this will only unlock once the whole method _runs to completion_.
see example here: https://go.dev/play/p/Cg6j6EpNtKl

pkg/transaction/send_tx_with_retry.go
acud: the mining while node is shut down is a separate use case that needs to be handled separately by the startup logic. what i meant is that if the node is offline, the fee pricing fee goes "out-of-fashion" just because the fee snapshot may be completely different with different values once you start the node next time.

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,586 @@
...
+				"send with retry: finished with error",
+				"attempt", attempt,
+				"tx_hash", txState.LastTxHash,
+				"nonce", txState.Nonce,
+				"to", formatToForLog(request, &txState),
+				"value", formatValueForLog(request, &txState),
+				"description", request.Description)
+		}
+
+		monitorLast := errors.Is(terminateTxErr, ErrTxMaxPriceExceeded) || errors.Is(terminateTxErr, ErrAllAttemptsExhausted)
```
acud: don't understand what is `monitorLast`... it is just a bool, what does it mean

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,586 @@
...
+		t.deleteRetryStateAndPending(txRetryKey, txState, monitorLast)
```
acud: are you sure you're not ending up with stale zombie db records here? for me this bool toggle is a bit tricky since the downstream logic may or may not remove some of the database records and that may lead to statestore growth. i see that there is some unit test coverage for db entry assertion (for deletion) but it's another 1000 lines of machine written code to go through

acud (changes requested):

pkg/transaction/send_tx_with_retry.go
sbackend123: Yes, and I did it consciously, same as it is in [Send](https://github.com/ethersphere/bee/blob/master/pkg/transaction/transaction.go#L176) method. I think it is made so because of possible nonce race

pkg/transaction/send_tx_with_retry.go
sbackend123: This is done so we can keep monitoring the last broadcast attempt.

Suppose we have exhausted all retries or hit the user-configured gas price ceiling. The last broadcast transaction may still be in the mempool, which means it can still be mined.

In that case, we clear its retry state, but keep it registered as pending. Look at [deleteRetryStateAndPending](https://github.com/ethersphere/bee/pull/5464/changes#diff-f987aa8316e6527a6d932d1792084d2fe5d375130e62eae8892852bef336a3c4R201)

pkg/storageincentives/redistribution/redistribution.go
sbackend123: I think BoostTipPercent should stay as-is (the other PR keeps it unchanged too). In agent.go we use it to check whether the user has enough funds with a buffer, and BoostTipPercent is effectively the market-tier fee tip.

pkg/transaction/send_tx_with_retry.go
sbackend123: You're right, pending entry would sit in the store until the node restarted (when waitForAllPendingTx would pick it up again.
Added monitoring on it, as its made by Send()

pkg/api/api.go
```diff
@@ -557,8 +558,9 @@ func (s *Service) gasConfigMiddleware(handlerName string) func(h http.Handler) h
 			logger := s.logger.WithName(handlerName).Build()
 
 			headers := struct {
-				GasPrice *big.Int `map:"Gas-Price"`
-				GasLimit uint64   `map:"Gas-Limit"`
+				GasPrice    *big.Int `map:"Gas-Price"`
+				GasLimit    uint64   `map:"Gas-Limit"`
+				FeePriority string   `map:"Fee-Priority"`
```
acud, line 563: this is not in the openapi spec

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,529 @@
...
+}
+
+func (t *transactionService) tierRangeForRequest(ctx context.Context) ([]feeTier, error) {
+	start := t.startTier
+	if override := sctx.GetFeePriority(ctx); override != "" {
+		parsed, err := ParseFeeTier(override)
+		if err != nil {
+			return nil, fmt.Errorf("fee priority: %w", err)
+		}
+		if parsed > t.endTier {
```
acud, line 114: if there's an override then you're clamping it? not sure this makes much sense no? if there's an override from the API then i'd kinda expect it to win here

pkg/transaction/send_tx_with_retry.go
acud: then at least put a very visible comment that this is intentional...

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,529 @@
...
+
+	if fixedNonce != nil {
+		nonce = *fixedNonce
+	} else {
+		t.lock.Lock()
+		defer t.lock.Unlock()
+
+		n, err := t.nextNonce(ctx)
+		if err != nil {
+			return nil, err
```
acud, line 139: error instrumentation is missing across this file and probably worth to go through the entire changeset for this - please instrument the errors whenever a function can return more than one error, otherwise we cannot know where it came from once it lands in the log

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,529 @@
...
+		GasTipCap: gasTipCap,
+		Data:      request.Data,
+	})
+
+	signedTx, err := t.signer.SignTx(tx, t.chainID)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrSignTransaction, err)
+	}
+
+	t.logger.Info("send with retry: broadcast",
```
acud, line 170: as a general comment: this is fine while the branch is in development but i'd be leaning towards having more concise logging. no need to log so much data. the `Send` method in this package is a good example for this

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,529 @@
...
+	if oldHash != (common.Hash{}) {
+		_ = t.store.Delete(pendingTransactionKey(oldHash))
+		_ = t.store.Delete(storedTransactionKey(oldHash))
+	}
+
+	return nil
+}
+
+// attemptResult is the outcome of a single broadcast+wait cycle.
+type attemptResult struct {
```
acud, line 237: not sure why is this type needed. usually i'd say this is needed when you want to pass multiple values in one go, normally as a result of an operation over a channel. here there are no channels, and go supports multiple return values, which i think would be nicer to use here

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,529 @@
...
+
+			if res.signedTx != nil {
+				previousTip = res.signedTx.GasTipCap()
+			}
+			attempt++
+		}
+	}
+
+	terminateTxErr = ErrAllAttemptsExhausted
+	return rs.lastTxHash, nil, terminateTxErr
```
acud, line 307: why care about the last tx hash when there's an error?

EDIT: because of the defer statement. usually if you have an error there's no need to return the other values, but here it does because of the defer. i would lean away from the defer in this case

pkg/transaction/send_tx_with_retry.go
```diff
@@ -0,0 +1,529 @@
...
+		"retry_delay", t.txRetryDelay,
+		"nonce_assigned", rs.nonceAssigned)
+
+	var (
+		previousTip    *big.Int
+		terminateTxErr error
+		nonce          *uint64
+	)
+	attempt := 1
+	defer func() { t.finishRetry(rs, request, attempt, terminateTxErr) }()
```
acud, line 269: this is not so idiomatic, and mostly just making everything here way more complicated. i would go with explicitly calling this method wherever needed and then explicitly returning, instead of returning and implicitly calling this always. it may save on a few lines but it makes the code less linear to read imo

acud (changes requested):
looking way better, i still have a few comments but i think we're almost there. thanks for you patience

PR #5466, opened 2026-05-19
feat(openapi): document utilization; add utilizedPercentage to postag…

### LINKED ISSUE #5226: Document `utilization` property of postage batches; consider adding `utilizedPercentage`
#### Summary
The `utilization` property of the postage batch refers to the highest written slot among the 65,536 buckets.

Utilization is a discrete value, where one unit corresponds to 256MB of data (theoretically) written.

For example, in a postage batch of depth 17, utilization 1 represents 50% usage, but for depth 18, only 25%.

To calculate the real utilization percentage, this formula must be applied:

```ts
/**
 * Utility function that calculates usage of postage batch based on its utilization, depth and bucket depth.
 *
 * @returns {number} A number between 0 and 1 representing the usage of the postage batch.
 */
export function getStampUsage(utilization: number, depth: number, bucketDepth: number): number {
  return utilization / Math.pow(2, depth - bucketDepth)
}
```

It is safe to assume that end-users and developers do NOT care about this Swarm-specific raw utilization value, as it is meaningless on its own. What they are most likely interested in is the usage percentage.

Currently, developers who are relying on the raw API could very easily (and incorrectly) assume that `utilization` represents usage percentage. Unfortunately, it seems the `utilization` property is not documented anywhere in our docs.

I strongly suggest documenting `utilization` in the OpenAPI specs and in our official docs first.

Additionally, I propose to introduce a new property, `utilizedPercentage`, which returns a number in the range `0..1` and provides a ready-to-use percentage for displaying usage.

martinconic:
### Checklist

### Description
Closes #5226

  Adds a description to the existing PostageBatch.utilization property in
  the OpenAPI spec, clarifying that it is a raw bucket-count value (not a
  usage percentage) and giving the formula for fractional usage.

  Adds a new utilizedPercentage field to the /stamps and
  /stamps/{batchID} responses, exposing utilization / 2^(depth -
  bucketDepth) as a float in [0, 1] so API consumers don't have to derive
  it themselves and can't accidentally treat the raw utilization as a
  percentage.

### AI Disclosure

### DIFF
--- a/pkg/api/postage.go
+++ b/pkg/api/postage.go
@@ -139,17 +139,18 @@ func (s *Service) postageCreateHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 type postageStampResponse struct {
-	BatchID       hexByte        `json:"batchID"`
-	Utilization   uint32         `json:"utilization"`
-	Usable        bool           `json:"usable"`
-	Label         string         `json:"label"`
-	Depth         uint8          `json:"depth"`
-	Amount        *bigint.BigInt `json:"amount"`
-	BucketDepth   uint8          `json:"bucketDepth"`
-	BlockNumber   uint64         `json:"blockNumber"`
-	ImmutableFlag bool           `json:"immutableFlag"`
-	Exists        bool           `json:"exists"`
-	BatchTTL      int64          `json:"batchTTL"`
+	BatchID            hexByte        `json:"batchID"`
+	Utilization        uint32         `json:"utilization"`
+	UtilizedPercentage float64        `json:"utilizedPercentage"`
+	Usable             bool           `json:"usable"`
+	Label              string         `json:"label"`
+	Depth              uint8          `json:"depth"`
+	Amount             *bigint.BigInt `json:"amount"`
+	BucketDepth        uint8          `json:"bucketDepth"`
+	BlockNumber        uint64         `json:"blockNumber"`
+	ImmutableFlag      bool           `json:"immutableFlag"`
+	Exists             bool           `json:"exists"`
+	BatchTTL           int64          `json:"batchTTL"`
 }
 
 type postageStampsResponse struct {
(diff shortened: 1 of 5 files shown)


akrem-chabchoub (approved):

pkg/api/postage.go
```diff
@@ -139,17 +139,18 @@ func (s *Service) postageCreateHandler(w http.ResponseWriter, r *http.Request) {
...
-	Depth         uint8          `json:"depth"`
-	Amount        *bigint.BigInt `json:"amount"`
-	BucketDepth   uint8          `json:"bucketDepth"`
-	BlockNumber   uint64         `json:"blockNumber"`
-	ImmutableFlag bool           `json:"immutableFlag"`
-	Exists        bool           `json:"exists"`
-	BatchTTL      int64          `json:"batchTTL"`
+	BatchID            hexByte        `json:"batchID"`
+	Utilization        uint32         `json:"utilization"`
+	UtilizedPercentage float64        `json:"utilizedPercentage"`
```
gacevicljubisa: The field returns a fraction in [0, 1], maybe apropriate name would be in this case `utilizationRatio` or `utilizationFraction`?

pkg/postage/stampissuer.go
```diff
@@ -222,6 +222,14 @@ func (si *StampIssuer) Utilization() uint32 {
 	return si.data.MaxBucketCount
 }
 
+// UtilizationPercentage returns the batch fullness as a fraction in the
+// range [0, 1], computed as Utilization / 2^(BatchDepth - BucketDepth).
+// A value of 1 means the most-filled bucket is full and any further write
+// to that bucket would overflow the batch.
+func (si *StampIssuer) UtilizationPercentage() float64 {
```
gacevicljubisa: Maybe just a minor defensive option to avoid divide by zero, not sure if it can happen:

```go
if si.data.BatchDepth <= si.data.BucketDepth {
      return 0
  }
```

sbackend123 (approved):

akrem-chabchoub (approved):

pkg/postage/stampissuer.go
martinconic: Nice observation, this should not happen and would be guarded when you buy a batch, but yes , it does not heart to be more defensive.

gacevicljubisa (approved):

PR #5468, opened 2026-05-20
fix(api): allow GET /wallet when swap or chequebook is disabled

### LINKED ISSUE #5233: Cannot get node wallet balance when chequebook is disabled
### Context
`2.6.0-d0aa8b93`

### Summary
Since the `GET /wallet` endpoint also returns `chequebookContractAddress`, this endpoint only works when the chequebook is enabled.

### Expected behavior
`GET /wallet` should work with chequebook disabled. Maybe set `chequebookContractAddress` to an empty string or `null` when it is not available.

### Actual behavior

`GET http://localhost:1633/wallet`

```
{
"code": 403,
"message": "Swap is disabled. This endpoint is unavailable."
}
```

or

```
{
"code": 403,
"message": "Chequebook is disabled. This endpoint is unavailable."
}
```

### Steps to reproduce
Set `swap-enable: false` or `chequebook-enable: false`

martinconic:
### Checklist

### Description
The /wallet endpoint was gated behind both swap and chequebook
  availability, and erc20Service was only initialized when swap was
  enabled. As a result, node operators could not read their xDAI/xBZZ
  balances with swap-enable=false or chequebook-enable=false.

  Initialize the chequebook factory and erc20Service whenever the chain
  backend is enabled, independently of swap/chequebook. The lookup stays
  fail-hard when swap is enabled (preserving existing startup behaviour)
  and is best-effort when swap is off, so nodes on custom chains without a
  known factory address still start. Gate /wallet and /wallet/withdraw on
  chain availability instead of swap/chequebook, and make the handler
  tolerate a nil erc20Service.

  Add an explicit startup guard rejecting swap-enable=true with the chain
  backend disabled, which previously failed deep in the chequebook factory
  call; without it the new flow would leave swap nil and panic on
  /settlements and related endpoints.

  Closes #5233

### AI Disclosure


pkg/node/node.go
```diff
@@ -535,46 +539,56 @@ func NewBee(
 		}
 	}
 
-	if o.SwapEnable {
-		chequebookFactory, err := InitChequebookFactory(logger, chainBackend, chainID, transactionService, o.SwapFactoryAddress)
-		if err != nil {
-			return nil, fmt.Errorf("init chequebook factory: %w", err)
+	if chainEnabled {
+		chequebookFactory, ferr := InitChequebookFactory(logger, chainBackend, chainID, transactionService, o.SwapFactoryAddress)
```
gacevicljubisa, line 543: When swap is disabled, the factory is only being initialized so we can read the BZZ token address. Could we use postagecontract.LookupERC20Address above this instead?
LookupERC20Address resolves the same token from the postage stamp contract, and would let us skip InitChequebookFactory entirely on the swap disabled path?

pkg/node/node.go
```diff
@@ -535,46 +539,56 @@ func NewBee(
 		}
 	}
 
-	if o.SwapEnable {
-		chequebookFactory, err := InitChequebookFactory(logger, chainBackend, chainID, transactionService, o.SwapFactoryAddress)
-		if err != nil {
-			return nil, fmt.Errorf("init chequebook factory: %w", err)
+	if chainEnabled {
```
gacevicljubisa, line 542: Consider extracting this whole chainEnabled block into a private helper, there are to many conditionals, it is hard to follow...

gacevicljubisa (commented):
Nice fix.

We can consider addressing one [TODO](https://github.com/ethersphere/bee/blob/a0c59532494d5813a1be0b148abb02febd430393/pkg/config/chain.go#L14) to add BzzAddress to [ChainConfig](https://github.com/ethersphere/bee/blob/a0c59532494d5813a1be0b148abb02febd430393/pkg/config/chain.go#L16) and drop LookupERC20Address entirely. The BZZ token address is deterministic per chain, and no reason to lok it up at runtime. It is maybe out of scope here, but maybe we can make an additional PR? Or maybe even in this one? Wdyt @acud @janos ?

martinconic:
> Nice fix.
> 
> We can consider addressing one [TODO](https://github.com/ethersphere/bee/blob/a0c59532494d5813a1be0b148abb02febd430393/pkg/config/chain.go#L14) to add BzzAddress to [ChainConfig](https://github.com/ethersphere/bee/blob/a0c59532494d5813a1be0b148abb02febd430393/pkg/config/chain.go#L16) and drop LookupERC20Address entirely. The BZZ token address is deterministic per chain, and no reason to lok it up at runtime. It is maybe out of scope here, but maybe we can make an additional PR? Or maybe even in this one? Wdyt @acud @janos ?

This is a good suggestion and task , I would make this separately and keep this PR light, covering only the scope of the reported issue.

pkg/node/node.go
martinconic: Good catch — postagecontract.LookupERC20Address is in fact already called below in the postage setup (node.go:726) to resolve bzzTokenAddress, so the factory's ERC20Address here is redundant on the swap-disabled path. Reusing it would let us drop the factory init entirely when swap is off. The wrinkle is ordering: erc20Service is built here, while bzzTokenAddress only resolves ~180 lines lower, so switching cleanly means reordering the erc20 init relative to the postage block. I'd rather keep this PR scoped to the /wallet gating fix and do the lookup reuse as a follow-up so the reordering gets its own review. WDYT?

pkg/node/node.go
martinconic: I agree, and we would need some refactoring in general not only for chainEnabled here, chainEnabled and other options are  hard to follow in multiple areas. This should be a different PR and some effort is already ongoing in [#5471](https://github.com/ethersphere/bee/pull/5471)

gacevicljubisa:
> > Nice fix.
> > We can consider addressing one [TODO](https://github.com/ethersphere/bee/blob/a0c59532494d5813a1be0b148abb02febd430393/pkg/config/chain.go#L14) to add BzzAddress to [ChainConfig](https://github.com/ethersphere/bee/blob/a0c59532494d5813a1be0b148abb02febd430393/pkg/config/chain.go#L16) and drop LookupERC20Address entirely. The BZZ token address is deterministic per chain, and no reason to lok it up at runtime. It is maybe out of scope here, but maybe we can make an additional PR? Or maybe even in this one? Wdyt @acud @janos ?
> 
> This is a good suggestion and task , I would make this separately and keep this PR light, covering only the scope of the reported issue.

I do not see anything critical in this PR, and we could make some minimal refactoring along the way, for this related changes, because current can be simplified with small effort.

martinconic:
> > > Nice fix.
> > > We can consider addressing one [TODO](https://github.com/ethersphere/bee/blob/a0c59532494d5813a1be0b148abb02febd430393/pkg/config/chain.go#L14) to add BzzAddress to [ChainConfig](https://github.com/ethersphere/bee/blob/a0c59532494d5813a1be0b148abb02febd430393/pkg/config/chain.go#L16) and drop LookupERC20Address entirely. The BZZ token address is deterministic per chain, and no reason to lok it up at runtime. It is maybe out of scope here, but maybe we can make an additional PR? Or maybe even in this one? Wdyt @acud @janos ?
> > 
> > 
> > This is a good suggestion and task , I would make this separately and keep this PR light, covering only the scope of the reported issue.
> 
> I do not see anything critical in this PR, and we could make some minimal refactoring along the way, for this related changes, because current can be simplified with small effort.

Agreed this is worth doing, but I'd rather keep this PR scoped to the reported issue (#5233) and not risk the clean bug-fix on a change that needs authoritative per-chain BZZ addresses, a new CLI override, a custom-chain fallback. I've opened #5475 to track the `BzzAddress`/`ChainConfig` migration and dropping `LookupERC20Address`.

@gacevicljubisa, I implemented your requested changes here: [#5476](https://github.com/ethersphere/bee/pull/5476)

gacevicljubisa (approved):

pkg/api/wallet.go
```diff
@@ -37,12 +37,15 @@ func (s *Service) walletHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	bzz, err := s.erc20Service.BalanceOf(r.Context(), s.ethereumAddress)
-	if err != nil {
-		logger.Debug("unable to acquire erc20 balance", "error", err)
-		logger.Error(nil, "unable to acquire erc20 balance")
-		jsonhttp.InternalServerError(w, "unable to acquire erc20 balance")
-		return
+	bzz := new(big.Int)
+	if s.erc20Service != nil {
```
akrem-chabchoub, line 41: I think when `s.erc20Service` is `nil` we could return `bzz` to `null` instead of `0` ?
Because in current implementation the user will get `0` in both cases of real `0` balance and `s.erc20Service` is `nil`

akrem-chabchoub (approved):

PR #5470, opened 2026-05-20
refactor: redundancy getter exhaust

nugaon:
Small cleanup of `pkg/replicas/getter.go` prompted by a beekeeper CI failure where the `return nil, nil` at the bottom of `Get` was somehow reached, causing a nil-panic downstream in `joiner.New`.

Changes:

- **`NONE` short-circuit**: `NewGetter` with `redundancy.NONE` now returns `g.Getter.Get` directly. The replica machinery (goroutines, ticker, replicator) is pointless for level 0, and emitting `ErrSwarmageddon` when the chunk is simply not found at level 0 was semantically wrong.
- **`return nil, nil` → `errGetterExhausted`**: the trailing return is structurally unreachable (the replicator exhausts exactly at the last loop iteration, turning it into an indefinite wait on `resultC`/`errc`; the loop always exits via success or Swarmageddon). Renamed to a named sentinel with a matching comment so the intent is clear if it ever surfaces again.
- **Comment above loop**: explains the batch-doubling design (2, 2, 4, 8, 16 for PARANOID) and why the loop cannot fall through to the bottom.
- **Test**: `level=0` all-fail case no longer asserts `ErrSwarmageddon`; the raw error is returned via the short-circuit.

### DIFF
--- a/pkg/replicas/getter.go
+++ b/pkg/replicas/getter.go
@@ -24,7 +24,13 @@ import (
 // then the probability of Swarmageddon is less than 0.000001
 // assuming the error rate of chunk retrievals stays below the level expressed
 // as depth by the publisher.
-var ErrSwarmageddon = errors.New("swarmageddon has begun")
+var (
+	ErrSwarmageddon = errors.New("swarmageddon has begun")
+	// errGetterExhausted is returned when the retry loop exhausts all levels without
+	// receiving a result or enough errors to trigger ErrSwarmageddon.
+	// This path should never be reached under normal operation.
+	errGetterExhausted = errors.New("replicas getter: exhausted all levels without result (unexpected)")
+)
 
 // getter is the private implementation of storage.Getter, an interface for
 // retrieving chunks. This getter embeds the original simple chunk getter and extends it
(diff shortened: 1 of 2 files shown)


gacevicljubisa (approved):

martinconic (approved):

pkg/replicas/getter.go
```diff
@@ -24,7 +24,13 @@ import (
 // then the probability of Swarmageddon is less than 0.000001
 // assuming the error rate of chunk retrievals stays below the level expressed
 // as depth by the publisher.
-var ErrSwarmageddon = errors.New("swarmageddon has begun")
+var (
+	ErrSwarmageddon = errors.New("swarmageddon has begun")
+	// errGetterExhausted is returned when the retry loop exhausts all levels without
+	// receiving a result or enough errors to trigger ErrSwarmageddon.
+	// This path should never be reached under normal operation.
+	errGetterExhausted = errors.New("replicas getter: exhausted all levels without result (unexpected)")
```
acud, line 32: similar comment to the other PR - this also suggests incorrect error handling. not a big fan of this but approving because this package's legibility is limited for me.

acud (approved):

PR #5472, opened 2026-05-21
feat(api): add patch endpoint to update issuer label

### LINKED ISSUE #5459: Allow renaming (relabelling) postage batches
#### Summary

Postage batches have an optional `label` property which can be set at the time of purchase, but it is not on-chain, only saved locally in Bee's database as metadata. The only purpose of `label` is to attach a tag to the postage batch so that users remember what they purchased them for. For example, I have a postage batch specifically for long-duration `static_websites`, a different one for mutable `testing`, and so on.

#### Motivation

There are cases when I want to re-label them, which is currently not possible:

- I forgot to set a label in the first place
- I no longer use it for its original purpose / want to use for a different reason, and I want to update the label to reflect this
- I set the label correctly but purchased with the wrong parameters (e.g. I wanted `mutable_test` but accidentally made it immutable), then I want to throw away the `mutable_test` label and create a new postage batch with the correct parameters
- When the Bee node database is corrupted / nuked, and postage batches are re-synced, they all reset to the "recovered" label

#### Implementation

This should be a fairly easy (🤞) implementation in Bee with a new endpoint: `PATCH /stamps/:batchId` accepting a `{"label":string}` body (or `PUT /stamps/:batchId/label`).

### AI Disclosure

akrem-chabchoub:
### Checklist

### Description
Adds a `PATCH /stamps/{batch_id}` endpoint that lets a node operator update the label of an existing postage batch issuer. The label is local metadata only, it has no on-chain effect and requires no transaction. The change is persisted to the stamp store so it survives node restarts.

### AI Disclosure


martinconic (approved):

pkg/postage/service.go
```diff
@@ -247,6 +248,20 @@ func (ps *service) GetStampIssuer(batchID []byte) (*StampIssuer, func() error, e
...
 }
 
+// UpdateIssuerLabel updates the label of the stamp issuer with the given batch ID and persists the change.
+func (ps *service) UpdateIssuerLabel(batchID []byte, label string) error {
+	ps.mtx.Lock()
+	defer ps.mtx.Unlock()
+
+	for _, st := range ps.issuers {
+		if bytes.Equal(batchID, st.data.BatchID) {
+			st.data.Label = label
```
gacevicljubisa, line 259: There is data race on data.Label when we have concurent GET + PATCH on the same batch.

gacevicljubisa (commented):
Looks ok. 
Maybe you can consider adding TestUpdateIssuerLabel on the postage service level which will cover ErrNotFound, and the in-memory update

gacevicljubisa (approved):

PR #5473, opened 2026-05-21
feat: max tx cost limit for redistribution game

sbackend123:
### Checklist

### Description
Extends transactions with retry with per-call overrides and adds redistribution-game behaviour on top of existing settings.

**Transaction layer**
WithRetryDelay — override retry delay for one SendWithRetry call.
WithIgnoreMaxPrice — bypass maxTxPrice when a predicate allows it (predicate gets proposed max_fee_per_gas in wei).

**Redistribution game**
Retry delay: commit / reveal / claim use min(transaction-retry-delay, 35s) so more attempts fit in a phase without changing the global delay for other txs.

Gas: no GasPrice from context; fees via SendWithRetry. MinEstimatedGasLimit lowered from 500k to 250k.

Only claim may bypass maxTxPrice. In the last 10 blocks of the round, if expected reward > spent for fees, the cap is ignored for that attempt. 

A long retry delay can leave the user unable to resend the transaction before the redistribution game phase ends, which means losing the right to participate in that round. For that reason, the delay is shortened for commit, reveal, and claim transactions.

A claim transaction is also roughly 1.5× more expensive than commit and reveal. With a user-configured max tx cost, funds may already be spent on commit and reveal while claim never goes through because of the limit — leaving the user at a net loss. To avoid that, the node is allowed to ignore the configured limit.

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5142

### AI Disclosure


go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/ethereum/go-ethereum v1.16.9
 	github.com/ethersphere/batch-archive v0.0.7
 	github.com/ethersphere/go-price-oracle-abi v0.6.9
-	github.com/ethersphere/go-storage-incentives-abi v0.9.4
+	github.com/ethersphere/go-storage-incentives-abi v0.9.3-rc4
```
martinconic, line 14: this is intentional and will be updated before the merge?

pkg/transaction/send_tx_with_retry.go
```diff
@@ -272,14 +321,18 @@ func (t *transactionService) retry(ctx context.Context, txRetryKey string, reque
 		t.recordRetryComplete(attempt, terminateTxErr)
```
martinconic: is attempt, or attempt+1 like in line 311 ?

go.mod
sbackend123: Only for testing on lighttestnet, will be removed before merge

pkg/transaction/send_tx_with_retry.go
martinconic: Both `To` and `Value`  and will cause a panic if  `nil`, right?

pkg/transaction/transaction.go
```diff
@@ -437,6 +438,18 @@ func (t *transactionService) prepareTransaction(ctx context.Context, request *Tx
 	if err != nil {
 		return nil, err
 	}
+	if request.GasFeeCap != nil {
```
martinconic: This block (and the TxRequest.GasFeeCap field at line 88) is only reachable via the legacy Send path, but I don't see any caller that sets request.GasFeeCap — the redistribution changes here route through SendWithRetry/prepareTransactionForTier, which doesn't use prepareTransaction. Is this OK ?

sbackend123: redundant code, removed

PR #5476, opened 2026-05-25
refactor: add BzzAddress to ChainConfig and remove LookupERC20Address

### LINKED ISSUE #5475: refactor: add BzzAddress to ChainConfig and remove postagecontract.LookupERC20Address
### Context

  Follow-up from PR #5468 and the existing TODO at `pkg/config/chain.go:14`.

  The BZZ token address is deterministic per chain, yet today it is resolved on-chain twice during node bootstrap:

  - via the chequebook factory `ERC20Address(ctx)` → feeds `erc20Service` (used by `/wallet` and chequebook), and
  - via `postagecontract.LookupERC20Address` at `pkg/node/node.go:726` → feeds the postage stamp and staking contracts.

  ### Proposal

  - Add `BzzAddress common.Address` to `ChainConfig` (`pkg/config/chain.go`), populated for mainnet and testnet.
  - Add a corresponding CLI/config override (e.g. `--bzz-token-address`) so custom/unknown chains can supply it.
  - Replace the `LookupERC20Address` call site with `chainCfg.BzzAddress` and remove the function (and `TestLookupERC20Address`).
  - Build `erc20Service` from the static address, decoupling `/wallet` from the chequebook factory entirely (simplifies the best-effort init added in #5468).

### AI Disclosure

martinconic:
### Checklist

### Description
  Follow-up to #5468 and the review comment there. Resolves the long-standing TODO in `pkg/config/chain.go` to source the BZZ token address from static config instead of looking it up on-chain at runtime.

  The address is deterministic per chain but was resolved on-chain **twice** during node bootstrap:

  - `chequebookFactory.ERC20Address(ctx)` → feeds `erc20Service` (and the `/wallet` endpoint), and
  - `postagecontract.LookupERC20Address` → feeds the postage stamp and staking contracts.

  Changes:

  - Add `BzzAddress` to `ChainConfig`, populated for mainnet and testnet from the `go-storage-incentives-abi` constants (same source already used for the staking/postage addresses).
  - Add a `--bzz-token-address` override for custom chains.
  - Resolve the address once and reuse it for `erc20Service` and the postage/staking contracts; remove `postagecontract.LookupERC20Address` and its test.
  - Add a startup guard rejecting `swap-enable=true` with the chain backend disabled. Removing the factory `ERC20Address` call also removed the implicit fail-fast it provided for that misconfiguration; the guard restores it with a clearer message instead of a later nil panic on swap endpoints.

Fixes [#5457](https://github.com/ethersphere/bee/issues/5475)

### AI Disclosure


cmd/bee/cmd/cmd.go
```diff
@@ -299,6 +299,7 @@ func (c *command) setAllFlags(cmd *cobra.Command) {
 	cmd.Flags().Duration(optionNameBlockchainRpcIdleTimeout, 90*time.Second, "blockchain rpc idle connection timeout")
 	cmd.Flags().Duration(optionNameBlockchainRpcKeepalive, 30*time.Second, "blockchain rpc TCP keepalive interval")
 	cmd.Flags().String(optionNameSwapFactoryAddress, "", "swap factory addresses")
+	cmd.Flags().String(optionNameBzzTokenAddress, "", "bzz token contract address (required on chains without a built-in default)")
```
gacevicljubisa: default is already passed through the config? 
maybe just `bzz token contract address`, to match the description with other contracts specified in the config.

akrem-chabchoub (approved):
Maybe you can just add the `bzz-token-address` to the `docker-compose.yml` file ?

akrem-chabchoub (changes requested):
Maybe you can just add the bzz-token-address to the docker-compose.yml file ?

Also the [beekeeper integration test](https://github.com/ethersphere/bee/actions/runs/26504540408/job/78053485288?pr=5476) is failing, not sure if its related to this pr ?

akrem-chabchoub (approved):

pkg/node/node.go
```diff
@@ -545,56 +546,55 @@ func NewBee(
 		}
 	}
 
-	if chainEnabled {
-		chequebookFactory, ferr := InitChequebookFactory(logger, chainBackend, chainID, transactionService, o.SwapFactoryAddress)
-		if o.SwapEnable && ferr != nil {
-			return nil, fmt.Errorf("init chequebook factory: %w", ferr)
-		}
+	chainCfg, found := config.GetByChainID(chainID)
```
acud: `found` isn't used, seems like a mistake

pkg/node/node.go
```diff
@@ -545,56 +546,55 @@ func NewBee(
...
-			}
-		} else {
-			logger.Warning("unable to init chequebook factory; BZZ balance will be unavailable via /wallet", "error", ferr)
+	bzzTokenAddress := chainCfg.BzzAddress
+	if o.BzzTokenAddress != "" {
+		if !common.IsHexAddress(o.BzzTokenAddress) {
+			return nil, errors.New("malformed bzz token address")
 		}
+		bzzTokenAddress = common.HexToAddress(o.BzzTokenAddress)
+	} else if chainEnabled && bzzTokenAddress == (common.Address{}) {
```
acud: 1. why not to use the compare method which is on the `common.Address` type?

2. why do the input sanitation here if you're checking `chainEnabled` later again too? i would suggest to move the check there

pkg/node/node.go
```diff
@@ -545,56 +546,55 @@ func NewBee(
...
-					return nil, fmt.Errorf("factory fail: %w", err)
-				}
-				logger.Warning("unable to resolve ERC20 token address; BZZ balance will be unavailable via /wallet", "error", err)
-			} else {
-				erc20Service = erc20.New(transactionService, erc20Address)
-			}
-		} else {
-			logger.Warning("unable to init chequebook factory; BZZ balance will be unavailable via /wallet", "error", ferr)
+	bzzTokenAddress := chainCfg.BzzAddress
+	if o.BzzTokenAddress != "" {
```
acud: since this parameter is already passed from the `cmd` package, i would suggest to do the input sanitation there instead of pulling it here (and to not pass a string as the option)

acud (changes requested):

pkg/node/node.go
martinconic: Kept == (common.Address{}) to match the existing convention — it's used consistently across the codebase

pkg/node/node.go
martinconic: `found` is used(line 753), I also renamed this to be maybe more relevant.

acud (approved):
Look OK. Just a couple of comments about naming - the PR title mentiones `BzzAddress` which is an actual type in the codebase and this is very misleading because it has nothing to do with the scope of the PR. Some variables/fields are also named like this in the changeset and I would suggest to rename them to something more like `TokenContractAddress` instead of `BzzAddress`

pkg/node/node.go
```diff
@@ -545,56 +546,53 @@ func NewBee(
...
-					o.SwapInitialDeposit,
-					erc20Service,
-				)
-				if err != nil {
-					return nil, fmt.Errorf("init chequebook service: %w", err)
-				}
-			}
-
-			chequeStore, cashoutService = initChequeStoreCashout(
+		if o.ChequebookEnable && chainEnabled {
```
gacevicljubisa: `chainEnabled` is always true here, because we have a check above where we reutrn error in case `o.SwapEnable && !chainEnabled`. `chainEnabled` condition can be removed.

pkg/node/node.go
```diff
@@ -545,56 +546,53 @@ func NewBee(
...
+
 	if chainEnabled {
-		chequebookFactory, ferr := InitChequebookFactory(logger, chainBackend, chainID, transactionService, o.SwapFactoryAddress)
-		if o.SwapEnable && ferr != nil {
-			return nil, fmt.Errorf("init chequebook factory: %w", ferr)
+		if bzzTokenAddress == (common.Address{}) {
+			return nil, errors.New("no known bzz token address for this network; provide --bzz-token-address")
 		}
+		erc20Service = erc20.New(transactionService, bzzTokenAddress)
+	}
```
gacevicljubisa, line 562: Full bzz token address verification can be moved here, not to have it partially before calling constructor and inside of it. But `o.BzzTokenAddress` need to be changed to `string` type. And it can be done only if `chainEnabled`. No need to error out if disabled. 
```go
	if chainEnabled {
		if o.BzzTokenAddress != "" {
			if !common.IsHexAddress(o.BzzTokenAddress) {
				return nil, errors.New("malformed bzz token address")
			}
			bzzTokenAddress = common.HexToAddress(o.BzzTokenAddress)
		}
		if bzzTokenAddress == (common.Address{}) {
			return nil, errors.New("no known bzz token address for this network; provide --bzz-token-address")
		}
		erc20Service = erc20.New(transactionService, bzzTokenAddress)
	}
```

martinconic: In the previous review round Elad asked specifically for the opposite — sanitize the address in cmd and pass a typed common.Address to NewBee rather than a raw string (done in 599f33bf). That aligns it with how the other contract-address flags are validated and keeps the constructor free of string parsing. I'd prefer no to flip it back unless we settle on one convention — @acud, any objection to consolidating validation in node.go as @gacevicljubisa suggests, or keep it cmd-side?

gacevicljubisa: I have missed that comment. Most of verification is done inside the constructor, and my idea was to have it in one place and all can be inside this one if statement when chain is enabled. In any case no need to return error if chain is disabled.

acud (approved):

gacevicljubisa (approved):

akrem-chabchoub (approved):

pkg/node/node.go
acud: ideally validation for commandline stuff is done in the `cmd` package. `node.go` already contains "the universe and more". i can't see why adding more business logic here is beneficial. the less branching you do in this file on stuff which is still not strongly typed - the better! also from a perspective of separation of concerns: provided cli flags are part of the envelope that exactly sits under the responsibility of the cmd packge (it handles everything that has to do with commandline interface).

cmd/bee/cmd/start.go
```diff
@@ -285,6 +286,14 @@ func buildBeeNode(ctx context.Context, c *command, cmd *cobra.Command, logger lo
...
 	}
 
+	var bzzTokenAddress common.Address
+	if a := c.config.GetString(optionNameBzzTokenAddress); a != "" {
+		if !common.IsHexAddress(a) {
+			return nil, errors.New("malformed bzz token address")
+		}
+		bzzTokenAddress = common.HexToAddress(a)
+	}
+
```
gacevicljubisa, line 296: Maybe to add Info log level for the contract address?

PR #5477, opened 2026-05-26
feat: v2.8.0

gacevicljubisa:
### Checklist

### Description
Bundles three related peer-identity features into the v2.8.0 release. Each BzzAddress is now signed over more fields and verified before it can enter the address book.

**Overlay nonce in signature** — the BzzAddress signature now covers the overlay nonce, and hive gossip validates this signature before storing records instead of accepting them at face value. Protocol version bumped.

**Signed timestamps on address records** — every BzzAddress carries a signed Unix timestamp. Future timestamps beyond clock-skew tolerance are rejected; gossip updates must be strictly newer and outside a minimum update interval, while handshake updates bypass the interval check for legitimate live reconnects. Zero-timestamp records are treated as legacy.

**Chequebook address verification** — every BzzAddress carries the peer's chequebook contract address, covered by the signature. On first ingestion three on-chain checks run: `issuer()` matches the peer's Ethereum address, deployed bytecode hash matches the canonical ERC20SimpleSwap hash, and balance meets a minimum threshold. A uniqueness check enforces a 1-to-1 overlay↔chequebook mapping. The address book acts as the verification gate (owner/bytecode checks cached on entry); balance is re-checked on each handshake. Verification is permissive when the verifier is not wired up.

A statestore migration re-wraps existing legacy address-book entries into the new verified envelope format.

### AI Disclosure


akrem-chabchoub (approved):

martinconic (approved):

acud (approved):

janos (approved):
This is a larger changeset that include also conflicts resolution, which is not so clear to track what has changed. The release v2.8.0 is tagged on a commit that is not on a master, not even on any dedicated branch in this repository, so the git exact state of that release will not be on the master branch.

In order not to introduce any regressions, it is needed to check the conflict resolution between the tagged release, master and the branch on this PR, I have used range diff `git range-diff origin/feat/v2.8.0..v2.8.0 origin/master..origin/feat/v2.8.0` with coordination with Ljubisa.

PR #5486, opened 2026-06-03
feat: mic moc

nugaon:
### Description

Adds WebSocket subscription endpoints for receiving real-time Single Owner Chunk (SOC) payloads filtered by owner address (MIC) or identifier (MOC):

- **MIC (Mined ID Chunk)**: Subscribe to all SOCs from a specific owner Ethereum address via `/mic/subscribe/{owner}`
- **MOC (Mined Owner Chunk)**: Subscribe to all SOCs with a specific identifier regardless of owner via `/moc/subscribe/{id}`

Both listeners integrate with push/pull sync to dispatch matching SOC payloads to WebSocket subscribers.

Includes optimized SOC listener implementation using `RWMutex` with atomic fast-path to minimize lock contention on the hot sync path. The same optimization is applied to the existing GSOC listener for consistency.

8.1.0 → 8.2.0 (new endpoints are additive features per SemVer)

Enables real-time notifications for specific SOC patterns without polling.

### Related Issue (Optional)

[MIC/MOC SWIP](https://github.com/ethersphere/SWIPs/pull/80)

### Checklist

### AI Disclosure

### DIFF
--- a/pkg/mic/mic.go
+++ b/pkg/mic/mic.go
@@ -75,7 +75,16 @@ func (l *listener) Handle(c *soc.SOC) {
 		return // no subscriptions, skip lock
 	}
 
-	h := l.getHandlers(c.OwnerAddress())
+	payload := c.WrappedChunk().Data()[swarm.SpanSize:]
+
+	// The read lock is held for the whole iteration so that a concurrent
+	// Subscribe cleanup (write lock) cannot mutate the handlers slice while it
+	// is being ranged over. Each handler is dereferenced and dispatched to its
+	// own goroutine, so the lock is never held while a handler runs.
+	l.handlersMu.RLock()
+	defer l.handlersMu.RUnlock()
+
+	h := l.handlers[string(c.OwnerAddress())]
 	if h == nil {
 		return // no handler
 	}
@@ -83,16 +92,9 @@ func (l *listener) Handle(c *soc.SOC) {
 
 	for _, hh := range h {
 		go func(hh Handler) {
-			hh(c.WrappedChunk().Data()[swarm.SpanSize:])
+			hh(payload)
 		}(*hh)
 	}
-}
-
-func (l *listener) getHandlers(owner []byte) []*Handler {
-	l.handlersMu.RLock()
-	defer l.handlersMu.RUnlock()
-
-	return l.handlers[string(owner)]
 }
 
 func (l *listener) Close() error {
--- a/pkg/mic/mic.go
+++ b/pkg/mic/mic.go
@@ -1,0 +1,106 @@
+// Copyright 2026 The Swarm Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package mic provides subscriptions to single-owner chunks by their owner
+// ethereum address. A MIC (Mined ID Chunk) subscription is matched against the
+// owner of every incoming single-owner chunk, regardless of its id.
+package mic
+
+import (
+	"encoding/hex"
+	"sync"
+	"sync/atomic"
+
+	"github.com/ethersphere/bee/v2/pkg/log"
+	"github.com/ethersphere/bee/v2/pkg/soc"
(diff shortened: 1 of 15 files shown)


pkg/mic/mic.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+func (l *listener) Close() error {
+	close(l.quit)
+	l.handlersMu.Lock()
+	defer l.handlersMu.Unlock()
+
+	l.handlers = make(map[string][]*Handler) // unset handlers on shutdown
+
```
sbackend123, line 106: possibly minor, but shouldn't we drop `subCount` to 0 here?

sbackend123 (changes requested):
pkg/mic/mic.go and pkg/moc/moc.go are nearly identical — the only real difference is which field of the SOC is used as the subscription key (owner vs id).

I'd suggest extracting a shared listener implementation and keeping mic / moc as thin wrappers that only supply that key function. Both packages already expose a Listener interface; the shared code can implement it once, with a keyFn func(*soc.SOC) string (or []byte) passed in at construction time.

WDYT?

acud (changes requested):
Agree with @sbackend123 - I suggest to try to write this with generics.

zelig:
Note on delivery semantics for subscription consumers: push-sync delivery is at-least-once — honest retries alone can deliver the same chunk repeatedly, and the handler dispatches to `gsocHandler` before the store's duplicate check, so subscription handlers fire per *delivery*, not per *chunk*. (Replaying captured chunk bytes triggers it too, at no mining cost.) Pure pub/sub is fine with this, but services that create per-request state from a subscription (session/watch-entry patterns, e.g. request/response over MOCs) must dedupe by chunk address. Probably worth a line in the mic/moc package docs so consumers don't learn it in production.

pkg/mic/mic.go
zelig: In fact, is it not more go sytle to insert a context on the handler which is then calcelled when the listener is closed?

pkg/mic/mic.go
```diff
@@ -0,0 +1,108 @@
...
+	"sync/atomic"
+
+	"github.com/ethersphere/bee/v2/pkg/log"
+	"github.com/ethersphere/bee/v2/pkg/soc"
+	"github.com/ethersphere/bee/v2/pkg/swarm"
+)
+
+// Handler defines code to be executed upon reception of a MIC message.
+// It is used as a parameter definition.
+type Handler func([]byte)
```
zelig, line 22: I think maybe this function should take context and a soc object. not sure why just the payload is in the argument. In all cases I have in mind, ID, owner, wrapped chunk etc are all relevant.

pkg/moc/moc.go
```diff
@@ -0,0 +1,108 @@
...
+
+	h := l.handlers[string(c.ID())]
+	if h == nil {
+		return // no handler
+	}
+	l.logger.Debug("new incoming MOC message", "soc id", hex.EncodeToString(c.ID()), "wrapped chunk address", c.WrappedChunk().Address())
+
+	for _, hh := range h {
+		go func(hh Handler) {
+			hh(payload)
```
zelig, line 95: the entire soc is the argument.

pkg/moc/moc.go
```diff
@@ -0,0 +1,108 @@
...
+	"sync/atomic"
+
+	"github.com/ethersphere/bee/v2/pkg/log"
+	"github.com/ethersphere/bee/v2/pkg/soc"
+	"github.com/ethersphere/bee/v2/pkg/swarm"
+)
+
+// Handler defines code to be executed upon reception of a MOC message.
+// It is used as a parameter definition.
+type Handler func([]byte)
```
zelig, line 22: in fact if the handler is taking ctx , and a soc

zelig (changes requested):
I think this should be implemented with one generic soc hook , with argument which makes all parts of the soc available, not just the payload, ideally already parsed and validated, and could also tell which part the hook matched on. 
See SWIP-60 also for topic match variants.

PR #5493, opened 2026-06-05
fix(handshake): reuse session-stable signed BzzAddress

gacevicljubisa:
### Checklist

### Description
Every handshake created and signed a new BzzAddress with a fresh timestamp, so each record the node advertised differed from the previous one even when nothing changed.
Receiving nodes therefore treated every record as an update: each reconnect caused an addressbook write on the peer, and re-gossiped records with advancing timestamps caused periodic addressbook writes across the whole network.

- The handshake service now mints the signed BzzAddress once and reuses it while the advertised underlay set and the local chequebook stay unchanged, so reconnects and re-gossip present a byte-identical record that peers skip without writing.
- A new record is minted only when the advertised underlays or the local chequebook actually change, with a monotonically increasing timestamp that stays valid across same-second mints and wall-clock regressions.
- The cache keeps only the most recently minted record: since peers reject handshake records older than the one they hold (`bzz.CheckTimestamp`), only the latest record is ever valid to present, so a single entry is sufficient — and bounded by design, regardless of what observed underlays peers report.
- `putHandshakeAddress` now skips the addressbook write when the stored record is identical and already verified, so a reconnecting peer costs one cached read and zero disk writes; the chequebook registry mapping is still re-registered and unverified records are still upgraded.

### Related Issue (Optional)
Closes https://github.com/ethersphere/bee-private/issues/23

### AI Disclosure

### DIFF
--- a/pkg/p2p/libp2p/internal/handshake/handshake.go
+++ b/pkg/p2p/libp2p/internal/handshake/handshake.go
@@ -102,8 +102,7 @@ type Service struct {
 	mu                    sync.RWMutex
 	hostAddresser         Addresser
 	now                   func() time.Time
-
-	addrCache addressCache // session-stable signed address, keyed by chequebook + underlays
+	addrCache             addressCache // session-stable signed address, keyed by chequebook + underlays
 }
 
 // Info contains the information received from the handshake.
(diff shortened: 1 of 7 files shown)


gacevicljubisa:
https://github.com/ethersphere/bee/actions/runs/27019032415/job/79741634855

janos (approved):

pkg/p2p/libp2p/internal/handshake/handshake.go
```diff
@@ -147,14 +148,15 @@ func New(signer crypto.Signer, advertisableAddresser AdvertisableAddressResolver
 }
 
 // SetChequebookAddress sets the local chequebook address included in
-// subsequent signed BzzAddress payloads. The zero value clears it.
+// subsequent signed BzzAddress payloads; the zero value clears it. The
+// chequebook is part of the signed-address cache key, so the next handshake
+// misses the cache and re-signs with the new chequebook.
 func (s *Service) SetChequebookAddress(addr common.Address) {
 	if (addr == common.Address{}) {
 		s.chequebookAddr.Store(nil)
```
acud, line 156: i wonder why we can't just save `&addr` here... why not save the default value?

acud (approved):

akrem-chabchoub (approved):

PR #5497, opened 2026-06-08
feat(gsoc): fine grained API

nugaon:
### Checklist

### Description

Extends the GSOC (`/gsoc/subscribe/{address}`) WebSocket endpoint so that clients can configure which SOC fields are serialized and sent on every incoming message, and optionally cache the wrapped chunk locally.

**New headers:**
- `Swarm-Soc-Fields` — comma-separated list of fields to include in the binary message. Allowed values: `address`, `recoveredPubKey`, `identifier`, `signature`, `wrappedAddress`, `span`, `payload`. Defaults to `payload` for backward compatibility.
- `Swarm-Cache-Wrapped-Chunk` — when `true`, the wrapped chunk of every incoming GSOC message is stored in the local cache so it can later be resolved through the bytes endpoint (useful when the SOC wraps a root chunk larger than 4 KB).

**API / handler changes:**
- The `gsoc.Handler` signature is changed from `func([]byte)` to `func(*soc.SOC)` so subscribers have access to all SOC properties.
- WebSocket `ReadBufferSize` and `WriteBufferSize` are sized to `swarm.SocMaxChunkSize` and `maxSocFieldsSize` respectively to avoid split writes for the worst-case serialized message.

**OpenAPI:**
- Documented both new headers in `SwarmCommon.yaml` and referenced them in the `/gsoc/subscribe/{address}` operation.
- Bumped the Bee API version from `8.1.0` to `8.2.0`.

Bee API: `8.1.0` → `8.2.0`

Previously the GSOC subscription returned only the raw payload of the wrapped chunk. Consumers who also needed the span, the SOC address, the signature, or the recovered public key had no way to obtain them through the WebSocket. This change makes the subscription fully configurable while keeping the default behavior unchanged.

Caching the wrapped chunk is a separate convenience for users who publish large content through GSOC and want recipients to be able to fetch the full content via the standard bytes endpoint.

### AI Disclosure
- [ x I have reviewed the AI generated code thoroughly.

### DIFF
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 
 info:
-  version: 8.1.0
+  version: 8.2.0
   title: Bee API
   description: "API endpoints for interacting with the Swarm network, supporting file operations, messaging, and node management"
 
@@ -859,9 +859,17 @@ paths:
             $ref: "SwarmCommon.yaml#/components/schemas/SwarmAddress"
           required: true
           description: "Single Owner Chunk address (which may have multiple payloads)"
+        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmSocFieldsParameter"
+        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmCacheWrappedChunkParameter"
       responses:
         "200":
-          description: Establishes a WebSocket subscription for incoming messages on the Single Owner Chunk address
+          description: >
+            Establishes a WebSocket subscription for incoming messages on the
+            Single Owner Chunk address. Each message is the binary serialization
+            of the Single Owner Chunk fields requested through the
+            swarm-soc-fields header (defaults to the wrapped chunk payload).
+        "400":
+          $ref: "SwarmCommon.yaml#/components/responses/400"
         "500":
           $ref: "SwarmCommon.yaml#/components/responses/500"
         default:
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 
 info:
-  version: 8.1.1
+  version: 8.2.0
   title: Bee API
   description: "API endpoints for interacting with the Swarm network, supporting file operations, messaging, and node management"
 
@@ -971,9 +971,17 @@ paths:
             $ref: "SwarmCommon.yaml#/components/schemas/SwarmAddress"
           required: true
           description: "Single Owner Chunk address (which may have multiple payloads)"
+        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmSocFieldsParameter"
+        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmCacheWrappedChunkParameter"
       responses:
         "200":
-          description: Establishes a WebSocket subscription for incoming messages on the Single Owner Chunk address
+          description: >
+            Establishes a WebSocket subscription for incoming messages on the
+            Single Owner Chunk address. Each message is the binary serialization
+            of the Single Owner Chunk fields requested through the
+            swarm-soc-fields header (defaults to the wrapped chunk payload).
+        "400":
+          $ref: "SwarmCommon.yaml#/components/responses/400"
         "500":
           $ref: "SwarmCommon.yaml#/components/responses/500"
         default:
--- a/openapi/SwarmCommon.yaml
+++ b/openapi/SwarmCommon.yaml
@@ -1144,6 +1144,29 @@ components:
       required: false
       description: Associate upload with an existing Tag UID
 
+    SwarmSocFieldsParameter:
+      in: header
+      name: swarm-soc-fields
+      schema:
+        type: string
+      required: false
+      description: >
+        Comma separated list of Single Owner Chunk fields to be serialized and
+        channeled on every incoming GSOC message, in the given order. Allowed
+        values are: address, recoveredPubKey, identifier, signature,
+        wrappedAddress, span, payload. When omitted it defaults to "payload".
+
+    SwarmCacheWrappedChunkParameter:
+      in: header
+      name: swarm-cache-wrapped-chunk
+      schema:
+        type: boolean
+      required: false
+      description: >
+        Indicates whether the wrapped chunk of every incoming GSOC message should
+        be cached locally so that it can be resolved through the bytes endpoint
+        (useful when the single owner chunk wraps a root chunk larger than 4KB).
+
     SwarmPinParameter:
       in: header
       name: swarm-pin
--- a/openapi/SwarmCommon.yaml
+++ b/openapi/SwarmCommon.yaml
@@ -1155,6 +1155,8 @@ components:
         channeled on every incoming GSOC message, in the given order. Allowed
         values are: address, recoveredPubKey, identifier, signature,
         wrappedAddress, span, payload. When omitted it defaults to "payload".
+        In order to have random access on the response bytes define payload
+        as the last field in the list since it has variable length.
 
     SwarmCacheWrappedChunkParameter:
       in: header
--- a/openapi/SwarmCommon.yaml
+++ b/openapi/SwarmCommon.yaml
@@ -1149,6 +1149,7 @@ components:
       name: swarm-soc-fields
       schema:
         type: string
+        default: "payload"
       required: false
       description: >
         Comma separated list of Single Owner Chunk fields to be serialized and
--- a/pkg/api/gsoc.go
+++ b/pkg/api/gsoc.go
@@ -183,7 +183,7 @@ func (s *Service) gsocListeningWs(conn *websocket.Conn, socAddress swarm.Address
 
 		b, err := socFieldsBytes(c, fields)
 		if err != nil {
-			s.logger.Debug("gsoc ws: serialize soc fields failed", "error", err)
+			s.logger.Warning("gsoc ws: serialize soc fields failed", "error", err)
 			return
 		}
 
--- a/pkg/api/gsoc.go
+++ b/pkg/api/gsoc.go
@@ -5,14 +5,107 @@
 package api
 
 import (
+	"bytes"
+	"context"
+	"fmt"
 	"net/http"
+	"slices"
+	"strings"
 	"time"
 
 	"github.com/ethersphere/bee/v2/pkg/jsonhttp"
+	"github.com/ethersphere/bee/v2/pkg/soc"
 	"github.com/ethersphere/bee/v2/pkg/swarm"
 	"github.com/gorilla/mux"
 	"github.com/gorilla/websocket"
 )
+
+// SOC field identifiers that can be requested through the SwarmSocFieldsHeader
+// to be serialized and channeled on every incoming GSOC chunk.
+const (
+	socFieldAddress         = "address"
+	socFieldRecoveredPubKey = "recoveredpubkey"
+	socFieldIdentifier      = "identifier"
+	socFieldSignature       = "signature"
+	socFieldWrappedAddress  = "wrappedaddress"
+	socFieldSpan            = "span"
+	socFieldPayload         = "payload"
+)
+
+var validSocFields = []string{
+	socFieldAddress,
+	socFieldRecoveredPubKey,
+	socFieldIdentifier,
+	socFieldSignature,
+	socFieldWrappedAddress,
+	socFieldSpan,
+	socFieldPayload,
+}
+
+// maxSocFieldsSize is the maximum size of a serialized SOC fields message when
+// every field is requested: the whole single owner chunk (identifier +
+// signature + span + payload, i.e. SocMaxChunkSize) plus the derived metadata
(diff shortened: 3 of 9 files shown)


pkg/api/gsoc.go
```diff
@@ -41,25 +156,39 @@ func (s *Service) gsocWsHandler(w http.ResponseWriter, r *http.Request) {
...
+
+		b, err := socFieldsBytes(c, fields)
+		if err != nil {
+			s.logger.Warning("gsoc ws: serialize soc fields failed", "error", err)
+			return
+		}
+
 		select {
-		case dataC <- m:
+		case dataC <- b:
```
sbackend123, line 202: Not sure how realistic this scenario is for us, but it looks like with many subscribers and a poor connection we could see unbounded growth of blocked goroutines.

sbackend123 (commented):
General comment: payload is the only variable-length field, but nothing stops clients from putting it before fixed fields (e.g. payload,identifier). That’s parseable only if you know all fixed sizes and derive payload_len = frame_len − fixed_total — which isn’t documented in OpenAPI. A typical left-to-right parser will misread or fail on those orderings. Worth either requiring payload to be last, or documenting field sizes and the parsing rule explicitly.

nugaon:
great catch @sbackend123 !

I wouldn't put any check for the serialization order. The payload always can be recovered and define its length from the response bytes, because all other properties have fixed length -> if `payload` is defined in the middle, just remove fixed properties length from the start and from the end, the remaining is `payload`.
Nevertheless, I'm going to add it to the openapi docs it has varlen and for random access response bytes, one should define it to the end of the props array.

sbackend123 (approved):

openapi/SwarmCommon.yaml
```diff
@@ -1144,6 +1144,31 @@ components:
       required: false
       description: Associate upload with an existing Tag UID
 
+    SwarmSocFieldsParameter:
+      in: header
+      name: swarm-soc-fields
+      schema:
+        type: string
```
martinconic, line 1151: We should have a default right? - default: "payload"

pkg/api/gsoc.go
```diff
@@ -41,29 +156,50 @@ func (s *Service) gsocWsHandler(w http.ResponseWriter, r *http.Request) {
...
 		case <-gone:
 			return
 		case <-s.quit:
 			return
+		default:
+			s.logger.Warning("gsoc ws: slow consumer, closing connection")
+			_ = conn.WriteControl(websocket.CloseMessage,
```
martinconic: In `gsocListeningWs`, the subscriber callback registered via `s.gsoc.Subscribe` is invoked asynchronously in a goroutine spawned by gsoc.Handle.

When the subscriber buffer (`dataC`, capacity 2) fills up due to a slow client, the callback enters the default: branch:

```go

default:
    s.logger.Warning("gsoc ws: slow consumer, closing connection")
    _ = conn.WriteControl(websocket.CloseMessage,
        websocket.FormatCloseMessage(websocket.ClosePolicyViolation, "slow consumer"),
        time.Now().Add(writeDeadline))
    _ = conn.Close()
    return
```

Simultaneously, the main `gsocListeningWs` goroutine is executing its event loop:

```go

case b := <-dataC:
    err = conn.SetWriteDeadline(time.Now().Add(writeDeadline))
    ...
    err = conn.WriteMessage(websocket.BinaryMessage, b)
case <-ticker.C:
    ...
    err = conn.WriteMessage(websocket.PingMessage, nil)
```

gorilla/websocket.Conn is NOT thread-safe for concurrent writers. Calling conn.WriteControl and conn.Close from the callback goroutine while the main loop calls conn.WriteMessage or conn.SetWriteDeadline causes a data race on the connection's writer buffers and internal state.

Do not call conn.WriteControl or conn.Close directly inside the callback goroutine. Instead, signal the main loop to close the connection, or close a signal channel so that all WebSocket frame writes are strictly executed by the single writer goroutine in `gsocListeningWs`

pkg/api/gsoc.go
```diff
@@ -5,15 +5,108 @@
...
+// public key and wrapped chunk address).
+const maxSocFieldsSize = swarm.SocMaxChunkSize +
+	swarm.HashSize + // soc address
+	soc.OwnerPubKeySize + // recovered public key
+	swarm.HashSize // wrapped chunk address
+
+// parseSocFields parses the SwarmSocFieldsHeader value into a list of SOC field
+// identifiers. When the header is empty it defaults to the payload field only,
+// which preserves backward compatibility.
+func parseSocFields(header string) ([]string, error) {
```
martinconic, line 61: Vulnerability:

`parseSocFields` splits the `Swarm-Soc-Fields `header by commas and validates each string against `validSocFields`. However, it never de-duplicates fields or caps the number of fields.

If a client sends an HTTP header with repeated field names:
```Swarm-Soc-Fields: payload,payload,payload,payload... (10,000 repetitions)```

- `parseSocFields` returns a slice containing 10,000 "payload" items.
- For every incoming SOC chunk, `socFieldsBytes` iterates over all 10,000 items, appending the ~4KB chunk payload into a single bytes.Buffer.
- Each chunk produces a ~40 MB buffer per subscriber. With 10 incoming chunks, this single connection allocates 400 MB of RAM.

An unauthenticated client can easily cause an Out-Of-Memory (OOM) panic on the Bee node.

As a fix maybe in `parseSocFields`, de-duplicate field names and enforce a maximum slice length equal to `len(validSocFields)`

File: pkg/api/gsoc.go [hunk 5]
```diff
@@ -41,29 +156,50 @@ func (s *Service) gsocWsHandler(w http.ResponseWriter, r *http.Request) {
...
 
 	var (
-		dataC  = make(chan []byte)
-		gone   = make(chan struct{})
-		ticker = time.NewTicker(s.WsPingPeriod)
-		err    error
+		dataC       = make(chan []byte, 2) // small buffer to decouple producer/consumer
+		gone        = make(chan struct{})
+		ticker      = time.NewTicker(s.WsPingPeriod)
+		ctx, cancel = context.WithCancel(context.Background()) // for storing cached chunks
```
martinconic: Would this be the case?
`ctx, cancel = context.WithCancel(context.Background())` is created inside `gsocListeningWs` and canceled when the WebSocket connection terminates (defer cancel()).

When `cacheWrappedChunk` is enabled, `s.storer.Cache().Put(ctx, c.WrappedChunk())` uses this connection-bound ctx. If the WebSocket connection closes while a chunk is being processed, cancel() is called and storer.Cache().Put fails with context.Canceled.

Caching a chunk in local storage is a side-effect for the node that should not be aborted simply because the subscriber's WebSocket closed.

pkg/api/gsoc.go
```diff
@@ -41,29 +156,50 @@ func (s *Service) gsocWsHandler(w http.ResponseWriter, r *http.Request) {
...
+		err         error
 	)
 	defer func() {
+		cancel()
```
martinconic: same as above

pkg/api/gsoc_test.go
```diff
@@ -134,7 +138,121 @@ func TestGsocPong(t *testing.T) {
 	}
 }
 
+// TestGsocWebsocketWrappedChunkData verifies that the Swarm-Soc-Fields header
```
martinconic, line 145: The unit tests added in gsoc_test.go only test happy paths. For example the following test cases are missing:

- Invalid Headers: Passing invalid field names in Swarm-Soc-Fields and asserting HTTP 400 Bad Request.
- Duplicate Fields: Sending duplicate fields and verifying de-duplication/rejection.
- Slow Consumer: Verifying clean connection shutdown when a consumer falls behind.
- Ordering Verification: Verifying that sequential Handle calls preserve message ordering.

openapi/Swarm.yaml
```diff
@@ -859,9 +859,17 @@ paths:
             $ref: "SwarmCommon.yaml#/components/schemas/SwarmAddress"
           required: true
           description: "Single Owner Chunk address (which may have multiple payloads)"
+        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmSocFieldsParameter"
+        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmCacheWrappedChunkParameter"
       responses:
         "200":
-          description: Establishes a WebSocket subscription for incoming messages on the Single Owner Chunk address
+          description: >
```
martinconic, line 978: Deficiencies
Undocumented Fixed Byte Sizes: The OpenAPI spec description lists available fields but fails to specify the exact binary byte size for each fixed field:

address: 32 bytes
recoveredPubKey: 33 bytes (compressed secp256k1 public key)
identifier: 32 bytes
signature: 65 bytes (secp256k1 signature + recovery id)
wrappedAddress: 32 bytes
span: 8 bytes (uint64)
payload: variable (0 to 4096 bytes)
Without explicit field lengths, API consumers cannot parse multi-field binary streams without inspecting Bee source code.

Variable-Length Field Order Warning: If payload is placed before fixed-length fields (e.g. Swarm-Soc-Fields: payload,span), parsing span requires scanning from the end of the frame. The spec should clearly state that payload MUST be specified as the last field when requesting multiple fields.

martinconic (changes requested):

openapi/Swarm.yaml
nugaon: I don't think so, it is not necessary to be specified as the last field. see the PR comment above your review.

pkg/api/gsoc.go
```diff
@@ -41,29 +164,51 @@ func (s *Service) gsocWsHandler(w http.ResponseWriter, r *http.Request) {
...
-func (s *Service) gsocListeningWs(conn *websocket.Conn, socAddress swarm.Address) {
+func (s *Service) gsocListeningWs(conn *websocket.Conn, socAddress swarm.Address, fields []string, cacheWrappedChunk bool) {
 	defer s.wsWg.Done()
 
 	var (
-		dataC  = make(chan []byte)
-		gone   = make(chan struct{})
-		ticker = time.NewTicker(s.WsPingPeriod)
-		err    error
+		dataC    = make(chan []byte, 2) // small buffer to decouple producer/consumer
```
martinconic, line 174: dataC has a capacity of only 2 items.

If a burst of 3 SOC chunks arrives in rapid succession:
Chunk 1 fills slot 1 of dataC.
Chunk 2 fills slot 2 of dataC.
Chunk 3 arrives while the writer goroutine is executing conn.WriteMessage(...) for Chunk 1 over the network socket.

Because dataC is full for even a few microseconds, Chunk 3 immediately drops into the default: branch, marks the connection as a slow consumer, and tears down the WebSocket connection.

Under realistic network conditions, healthy clients will experience random unexpected WebSocket disconnects during normal chunk bursts.

I would increase the channel buffer to 32

martinconic:
Thank you for the changes, look good, only one thing found above.

pkg/api/gsoc.go
acud: [why](https://github.com/ethersphere/bee/blob/master/CODINGSTYLE.md#channel-size-is-one-or-none)?

i don't think that the solution here is to increase the buffer size (you can still fall into that case with a `32` sized buffer.) also, usually the http library itself might employ buffering on the websocket - the fact that the write returned does not necessarily mean the socket flushed it down the wire. 

so we are essentially building more buffers over an existing buffer. the other problem here is the actual usage of a channel. 

if we remove the buffering, we essentially also throw out the window the message sequencing - once multiple writers compete over writing to a channel the sequence of arrival of the writers is lost since the runtime would randomize the selection of who-writes-next. 

if message ordering is important (not sure about that) then the call should probably write into a guarded FIFO stack that should be flushed onto the socket, if the ordering does not matter (it rarely does) then removing the buffer alongside the `default` case would probably make for a design with less moving parts - allowing the http library to fail loudly on a missed write deadline and propagating that error - killing the subscription along the way.

martinconic: I agree @acud , if we remove the buffered channel I think all looks good from my side.

martinconic (approved):

akrem-chabchoub:
Thanks @nugaon
After this PR is merged I noticed some flaky tests.
`gsoc.Handle` spawns a goroutine per message, so message order isn't actually guaranteed, worth checking since it could be the cause 👀
https://github.com/ethersphere/bee/actions/runs/33607360631/job/100174304103?pr=5591
https://github.com/ethersphere/bee/actions/runs/33604392092/job/100164967870?pr=5590
https://github.com/ethersphere/bee/actions/runs/33607360631/job/100178668965?pr=5591
https://github.com/ethersphere/bee/actions/runs/33632319071/job/100282470247
https://github.com/ethersphere/bee/actions/runs/33632319071/job/100254552443
https://github.com/ethersphere/bee/actions/runs/33632319071/job/100292965376

nugaon:
Good catch, thanks! Fixed in #5593 — `gsoc.Handle` now calls each subscriber handler synchronously in registration order instead of spawning a goroutine per handler, so message delivery order is preserved. Updated `TestRegister` accordingly since it relied on the async fan-out via an unbuffered channel.

PR #5499, opened 2026-06-10
fix: prevent second batch store reset wiping postage snapshot on --resync

### LINKED ISSUE #5495: Shutdown during `--resync` with postage snapshot: `get: get batch <batchID>: storage: not found`
### Context
2.8.0-41d6efc6

**`config.yaml`**

```
api-addr: localhost:1633
blockchain-rpc-endpoint: ***
cors-allowed-origins: ['*']
data-dir: ***
chequebook-enable: true
swap-enable: true
full-node: false
mainnet: true
password: ***
resolver-options: ***
```

Command:

`./bee start --config=config.yaml --resync=true`

### Summary

I needed to nuke and resync my node due to chequebook corruption (I jumped between 2.7 and 2.8 with this node earlier).

When using the postage snapshot, my Bee node prints the error `get: get batch 5b915b7bbeb9908ffc0fc4c68aadaf4cea63c6e951299ad67f058b1e7ae8cfe9: storage: not found` during syncing and shuts down soon afterwards.

```
"time"="2026-06-05 17:53:31.872312" "level"="warning" "logger"="node/batchservice" "msg"="batch service: resync requested, resetting batch store"
"time"="2026-06-05 17:53:31.877762" "level"="warning" "logger"="node/batchservice" "msg"="batch service: batch store has been reset. your node will now resync chain data. this might take a while..."
"time"="2026-06-05 17:53:31.877797" "level"="info" "logger"="node" "msg"="loading batch snapshot"
"time"="2026-06-05 17:53:32.965755" "level"="info" "logger"="node" "msg"="batch snapshot loaded successfully" "log_count"=132388 "max_block_height"=46358865
"time"="2026-06-05 17:54:10.092186" "level"="info" "logger"="node" "msg"="waiting to sync postage contract data, this may take a while... more info available in Debug loglevel"
"time"="2026-06-05 17:54:10.145840" "level"="info" "logger"="node" "msg"="started postage contract data sync in the background..."
"time"="2026-06-05 17:54:10.145876" "level"="warning" "logger"="node/batchservice" "msg"="batch service: resync requested, resetting batch store"
"time"="2026-06-05 17:54:10.146066" "level"="info" "logger"="node" "msg"="starting in light mode"
"time"="2026-06-05 17:54:10.151388" "level"="warning" "logger"="node/batchservice" "msg"="batch service: batch store has been reset. your node will now resync chain data. this might take a while..."
"time"="2026-06-05 17:54:10.335473" "level"="info" "logger"="node/multiresolver" "msg"="connected" "tld"="" "endpoint"="https://ethereum-rpc.publicnode.com"
"time"="2026-06-05 17:54:11.702994" "level"="error" "logger"="node/listener" "msg"="failed syncing event listener; shutting down node error" "error"="get: get batch 5b915b7bbeb9908ffc0fc4c68aadaf4cea63c6e951299ad67f058b1e7ae8cfe9: storage: not found"
"time"="2026-06-05 17:54:11.703012" "level"="error" "logger"="node" "msg"="unable to sync batches" "error"="get: get batch 5b915b7bbeb9908ffc0fc4c68aadaf4cea63c6e951299ad67f058b1e7ae8cfe9: storage: not found"
"time"="2026-06-05 17:54:20.932223" "level"="info" "logger"="node" "msg"="node warmup check initiated. monitoring activity rate to determine readiness." "startTime"="2026-06-05 17:54:20.932221 +0200 CEST m=+49.888515168"
"time"="2026-06-05 17:54:20.932333" "level"="info" "logger"="node" "msg"="node warmup complete. system is considered stable and ready." "stabilizationTime"="2026-06-05 17:54:20.932221 +0200 CEST m=+49.888515168" "totalMonitoredEvents"=1 "warmupDurationSeconds"=49.691165209
"time"="2026-06-05 17:54:20.932427" "level"="info" "logger"="node" "msg"="node warmup stabilization complete, updating API status"
"time"="2026-06-05 17:54:20.932616" "level"="info" "logger"="node/pusher" "msg"="got a chunk being smuggled"
"time"="2026-06-05 17:54:43.256320" "level"="info" "logger"="node" "msg"="shutting down..."
```

When I disable the postage snapshot with `skip-postage-snapshot`, the Bee node can sync properly and works as expected.

### Expected behavior
No errors when using the postage snapshot.

### Actual behavior
Node shutdown due to possible postage snapshot issue.

My theory is that the snapshot missed a topup event, and considered the postage batch expired sooner than it should have, and then couldn't apply an additional topup or dilute event to the same postage batch since it was no longer found in the local database. However, this is just a naive guess.

### Steps to reproduce
I can reproduce this locally by running `db nuke` and then starting with `resync`. 

### Possible solution
Extra checks on the postage batch snapshot whether it is complete.

### AI Disclosure

martinconic:
### Checklist

### Description

Fixes #5495: a node started with `--resync` while the postage snapshot is active shuts down during sync with:

```
get: get batch <id>: storage: not found
```

#### Root cause

The batch store is reset **twice** on startup in this configuration:

1. `snapshotBatchSvc.Start()` resets the store (`resync`) and rebuilds it from the snapshot, then sets `postageSyncStart = maxBlockHeight`.
2. The live `batchSvc` was constructed with the same `o.Resync` flag, so its `Start()` reset the store **again**, discarding the snapshot that was just loaded.

Live sync then started from the snapshot block height against an **empty** store, so the first top-up/dilute event referencing a batch created before the snapshot height failed with `storage: not found`, and the node shut down.

This matches the report exactly: two `"resync requested, resetting batch store"` log lines, and the fact that `skip-postage-snapshot` works around it (only one reset, and sync replays from contract genesis where every create precedes its top-up).

Note: this is **independent of #5343/#5482**. The flow originates in #5094 and is still present on master; the reporter's build (`2.8.0-41d6efc6`) predates the #5343 revert, but reverting #5343 does not address this.

#### Fix

The two reset sites are not redundant — one rebuilds from the snapshot, the other is the fallback rebuild for the no-snapshot / snapshot-failed / non-mainnet cases — they were just both armed at once.

Construct the live batch service **after** the snapshot path has run and pass resync only when the snapshot did not rebuild the store:

```go
batchSvc, err = batchservice.New(..., o.Resync && !snapshotLoaded)
```

The reset logic stays in its single existing place (`batchservice.Start()`), exactly one service is armed to reset, and there is no runtime toggle. The store is still reset exactly once in every other case (snapshot skipped, failed, not applicable, or a non-resync start).

### Testing

- `TestResyncControlsReset` in `batchservice` asserts both directions of the contract `node.go` relies on:
  - `resync=true` ⇒ store reset once
  - `resync=false` ⇒ store **not** reset (the post-snapshot live-service case)
- `go test -race ./pkg/postage/batchservice/...` passes
- `go build ./pkg/node/...` passes; `go vet`, `gofumpt`, and `golangci-lint` clean

None.

### Related Issue

Closes #5495

### AI Disclosure

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -45,6 +45,8 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/postage/batchstore"
 	"github.com/ethersphere/bee/v2/pkg/postage/listener"
 	"github.com/ethersphere/bee/v2/pkg/postage/postagecontract"
+	"github.com/ethersphere/bee/v2/pkg/postage/snapshot"
+	"github.com/ethersphere/bee/v2/pkg/postage/snapshot/archive"
 	"github.com/ethersphere/bee/v2/pkg/pricer"
 	"github.com/ethersphere/bee/v2/pkg/pricing"
 	"github.com/ethersphere/bee/v2/pkg/pss"
@@ -892,21 +894,17 @@ func NewBee(
 	// store is reset only when --resync is requested. Either way the store is
 	// reset in a single place (batchservice.New), which avoids a second reset
 	// wiping the snapshot that was just loaded (see #5495).
-	var snapshot *batchservice.Snapshot
+	var batchSnapshot *batchservice.Snapshot
 	if !o.SkipPostageSnapshot && !batchStoreExists && (networkID == mainnetNetworkID) && beeNodeMode != api.UltraLightMode {
-		chainBackend := NewSnapshotLogFilterer(logger, archiveSnapshotGetter{})
-
-		snapshotEventListener := listener.New(b.syncingStopped, logger, chainBackend, postageStampContractAddress, postageStampContractABI, o.BlockTime, postageSyncingStallingTimeout, postageSyncingBackoffTimeout)
-
-		snapshot = &batchservice.Snapshot{
-			Listener:   snapshotEventListener,
-			Backend:    chainBackend,
-			StartBlock: postageSyncStart,
+		batchSnapshot, err = snapshot.New(ctx, logger, archive.Getter{}, b.syncingStopped, postageStampContractAddress, postageStampContractABI, o.BlockTime, postageSyncingStallingTimeout, postageSyncingBackoffTimeout, postageSyncStart)
+		if err != nil {
+			// A corrupt snapshot is not fatal: rebuild from the chain instead.
+			logger.Error(err, "postage snapshot unavailable, syncing from chain instead")
 		}
 	}
 
 	var snapshotLoaded bool
-	batchSvc, snapshotLoaded, err = batchservice.New(ctx, stateStore, batchStore, logger, eventListener, overlayEthAddress.Bytes(), post, sha3.New256, snapshot, o.Resync)
+	batchSvc, snapshotLoaded, err = batchservice.New(ctx, stateStore, batchStore, logger, eventListener, overlayEthAddress.Bytes(), post, sha3.New256, batchSnapshot, o.Resync)
 	if err != nil {
 		return nil, fmt.Errorf("init batch service: %w", err)
 	}
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -887,37 +887,33 @@ func NewBee(
 		}
 	)
 
-	snapshotLoaded := false
+	// When the postage snapshot applies, hand it to the batch service so it
+	// rebuilds the store from the snapshot during construction; otherwise the
+	// store is reset only when --resync is requested. Either way the store is
+	// reset in a single place (batchservice.New), which avoids a second reset
+	// wiping the snapshot that was just loaded (see #5495).
+	var snapshot *batchservice.Snapshot
 	if !o.SkipPostageSnapshot && !batchStoreExists && (networkID == mainnetNetworkID) && beeNodeMode != api.UltraLightMode {
 		chainBackend := NewSnapshotLogFilterer(logger, archiveSnapshotGetter{})
 
 		snapshotEventListener := listener.New(b.syncingStopped, logger, chainBackend, postageStampContractAddress, postageStampContractABI, o.BlockTime, postageSyncingStallingTimeout, postageSyncingBackoffTimeout)
 
-		snapshotBatchSvc, err := batchservice.New(stateStore, batchStore, logger, snapshotEventListener, overlayEthAddress.Bytes(), post, sha3.New256, o.Resync)
-		if err != nil {
-			logger.Error(err, "failed to initialize batch service from snapshot, continuing outside snapshot block...")
-		} else {
-			err = snapshotBatchSvc.Start(ctx, postageSyncStart)
-			syncStatus.Store(true)
-			if err != nil {
-				syncErr.Store(err)
-				logger.Error(err, "failed to start batch service from snapshot, continuing outside snapshot block...")
-			} else {
-				postageSyncStart = chainBackend.maxBlockHeight
-				snapshotLoaded = true
-			}
+		snapshot = &batchservice.Snapshot{
+			Listener:   snapshotEventListener,
+			Backend:    chainBackend,
+			StartBlock: postageSyncStart,
 		}
-		if errClose := snapshotEventListener.Close(); errClose != nil {
-			logger.Error(errClose, "failed to close event listener (snapshot) failure")
-		}
-
 	}
 
-	// Don't resync the live batch service when the snapshot already rebuilt the
-	// store, otherwise it would be discarded and sync would fail (see #5495).
-	batchSvc, err = batchservice.New(stateStore, batchStore, logger, eventListener, overlayEthAddress.Bytes(), post, sha3.New256, o.Resync && !snapshotLoaded)
+	var snapshotLoaded bool
+	batchSvc, snapshotLoaded, err = batchservice.New(ctx, stateStore, batchStore, logger, eventListener, overlayEthAddress.Bytes(), post, sha3.New256, snapshot, o.Resync)
 	if err != nil {
 		return nil, fmt.Errorf("init batch service: %w", err)
+	}
+	if snapshotLoaded {
+		// The snapshot rebuilt the store up to its block height, so the node can
+		// already serve postage requests while the remaining gap syncs live.
+		syncStatus.Store(true)
 	}
 
 	if batchSvc != nil && chainEnabled {
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -787,11 +787,6 @@ func NewBee(
 	eventListener = listener.New(b.syncingStopped, logger, chainBackend, postageStampContractAddress, postageStampContractABI, o.BlockTime, postageSyncingStallingTimeout, postageSyncingBackoffTimeout)
 	b.listenerCloser = eventListener
 
-	batchSvc, err = batchservice.New(stateStore, batchStore, logger, eventListener, overlayEthAddress.Bytes(), post, sha3.New256, o.Resync)
-	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/postage/batchservice/batchservice.go
```diff
@@ -241,6 +241,13 @@ func (svc *batchService) TransactionEnd() error {
 
 var ErrInterruped = errors.New("postage sync interrupted")
 
+// SkipResync disables the resync reset for this service. It is called after a
+// postage snapshot has already reset and rebuilt the batch store, so the live
+// sync continues from the snapshot instead of wiping it (see issue #5495).
+func (svc *batchService) SkipResync() {
+	svc.resync = false
```
acud: to me the fact that there's two places where the batch service could get reset already smells a bit. with this change we have still two places where a reset happens + third path to disable one of them potentially. so the question is asked - why have two places that do this in the first place?

martinconic: The two reset sites aren't pure duplication. They serve different roles:
  - snapshot service → resets + rebuilds from the snapshot
  - live service → resets + rebuilds from chain genesis — this is the fallback path, needed when there's no snapshot
So you can't simply delete one site without merging the two services (they use different listeners — snapshot filterer vs live chain backend).

I changed the implementation to not modify at all `batchservice.go`

pkg/node/node.go
```diff
@@ -892,6 +887,7 @@ func NewBee(
 		}
 	)
 
+	snapshotLoaded := false
```
gacevicljubisa: Suggestion: Instead of having this guard, we could expose `batchservice.Reset(stateStore, batchStore)` and just call it before start in case if --resync is set to true

gacevicljubisa: Additional solution could be to modify the Start interface:

```go
Start(ctx context.Context, listener Listener, startBlock uint64) error
```
where we would pass snapshotEventListener and eventListener respectfully, and then when resync is done, we could set it to false. 

But this requires interface changes.

gacevicljubisa: @acud wdyt?

acud: I am not sure exposing `Reset` is needed. From my perspective consider this:
1. if you want to reset (the flag was set) - you start the batch service the snapshot. If you don't want to reset - snapshot it nil.
2. inside the constructor, it detects - if the snapshot is not nil - reset, replay events, then continue startup normally, returning the instance.

If it is really needed to be able to reset _without_ a snapshot, i.e. force a reset that reads directly from the blockchain, pass both a flag and the snapshot (nullable). Not sure I see the need for external callers to be able to reset the state of that component in that way.

pkg/postage/batchservice/batchservice.go
```diff
@@ -65,37 +100,113 @@ func New(
...
+		}
+	}
+
+	return bs, bs.snapshotLoaded, nil
+}
+
+// reset wipes the batch store so it can be rebuilt from scratch. It is called at
+// most once per startup, from New (snapshot rebuild) or Start (live rebuild).
+func (svc *batchService) reset(dirty bool) error {
+	if dirty {
```
acud: if the reset is the same reset for both cases then i don't see why you need to add this into the signature here. why don't you just log the line at the specific point you've detected that there was a dirty shutdown? putting this here also means that in case something went wrong in between the detection and the logging, you actually lose visibility that a reset was supposed to happen

acud (approved):

akrem-chabchoub (approved):

pkg/postage/batchservice/batchservice.go
```diff
@@ -65,37 +98,113 @@ func New(
...
+	if snapshot != nil {
+		var resetReason string
+		switch {
+		case dirty:
+			resetReason = "batch service: dirty shutdown detected, resetting batch store"
+		case resync:
+			resetReason = "batch service: resync requested, resetting batch store"
+		}
+		if resetReason != "" {
+			logger.Warning(resetReason)
```
acud: i dont still understand why you're not logging that a reset was either requested (in the if block that starts `L108`), and why don't you check the `if dirty` (and log) directly after loading it from the dirty key...? in both cases it makes sense to log it there. i won't insist but this was already raised before, not sure why conditional logging is needed

pkg/postage/batchservice/batchservice.go
```diff
@@ -65,37 +98,113 @@ func New(
...
+	// is not fatal; Start then rebuilds from the chain instead.
+	if snapshot != nil {
+		var resetReason string
+		switch {
+		case dirty:
+			resetReason = "batch service: dirty shutdown detected, resetting batch store"
+		case resync:
+			resetReason = "batch service: resync requested, resetting batch store"
+		}
+		if resetReason != "" {
```
acud: so.. i find it a bit strange that a string value governs whether a reset will happen. it would make sense to do this sort of a check in the `if` directly: `if (dirty || resync) && snapshot != nil`

pkg/postage/batchservice/batchservice.go
```diff
@@ -248,27 +357,32 @@ func (svc *batchService) Start(ctx context.Context, startBlock uint64) (err erro
...
+	var resetReason string
+	switch {
+	case dirty:
+		resetReason = "batch service: dirty shutdown detected, resetting batch store"
+	case svc.resync && !svc.snapshotLoaded:
+		resetReason = "batch service: resync requested, resetting batch store"
+	}
+	if resetReason != "" {
+		svc.logger.Warning(resetReason)
+		if err := svc.reset(); err != nil {
```
acud: if the reset was already known to be needed at time of ctor, why do you need to reset it again here?

acud (changes requested):

pkg/postage/snapshot/archive/archive.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2025 The Swarm Authors. All rights reserved.
```
sbackend123: 2026

pkg/postage/snapshot/factory.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2025 The Swarm Authors. All rights reserved.
```
sbackend123: 2026

pkg/node/node.go
```diff
@@ -892,28 +889,29 @@ func NewBee(
...
+	// reset in a single place (batchservice.New), which avoids a second reset
+	// wiping the snapshot that was just loaded (see #5495).
+	var batchSnapshot *batchservice.Snapshot
 	if !o.SkipPostageSnapshot && !batchStoreExists && (networkID == mainnetNetworkID) && beeNodeMode != api.UltraLightMode {
-		chainBackend := NewSnapshotLogFilterer(logger, archiveSnapshotGetter{})
-
-		snapshotEventListener := listener.New(b.syncingStopped, logger, chainBackend, postageStampContractAddress, postageStampContractABI, o.BlockTime, postageSyncingStallingTimeout, postageSyncingBackoffTimeout)
-
-		snapshotBatchSvc, err := batchservice.New(stateStore, batchStore, logger, snapshotEventListener, overlayEthAddress.Bytes(), post, sha3.New256, o.Resync)
+		batchSnapshot, err = snapshot.New(ctx, logger, archive.Getter{}, b.syncingStopped, postageStampContractAddress, postageStampContractABI, o.BlockTime, postageSyncingStallingTimeout, postageSyncingBackoffTimeout, postageSyncStart)
```
sbackend123, line 899: Minor, but I think passing empty `archive.Getter{}` looks strange, it makes me wondering why it is needed. So I would suggest, in archive.go use `var Default Getter` and then pass here `archive.Default`

sbackend123 (approved):

acud (approved):
lgtm

PR #5500, opened 2026-06-11
chore: improve hive and pullsync message handling

akrem-chabchoub:
### Checklist

### AI Disclosure

### DIFF
--- a/pkg/hive/hive.go
+++ b/pkg/hive/hive.go
@@ -324,6 +324,11 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
 	peersToAdd := make([]swarm.Address, 0, len(peers.Peers))
 
 	for _, p := range peers.Peers {
+		if p == nil {
+			s.logger.Debug("nil peer entry in Peers message, skipping")
+			continue
+		}
+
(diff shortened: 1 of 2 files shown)


martinconic:
I'm not convinced these checks guard against anything reachable. All three slices come out of protobuf unmarshaling, and the decoder allocates a fresh struct for every repeated element — even an empty or garbage wire entry yields a non-nil pointer with zero-value fields, never a literal nil.

gacevicljubisa (approved):

acud (approved):

acud:
> I'm not convinced these checks guard against anything reachable. All three slices come out of protobuf unmarshaling, and the decoder allocates a fresh struct for every repeated element — even an empty or garbage wire entry yields a non-nil pointer with zero-value fields, never a literal nil.

While I haven't done the manual tests to see whether the claim is correct or not, I think it is not a bad thing to do input validation. If the something in the serialization changes the code would just break. Also, don't confuse what happens when you use a payload that is generated by protobuf (the usual serialize code-path), and a maliciously crafted payload the circumvents the protobuf null field guards. You could probably still trick protobuf to decode a null value. Just in the same way you could create a JSON by hand that has a `field: null`, simply because JSON supports it, and circumvent the golang JSON serializer `json:"...,omitempty"` directive.

Issue #5485, opened 2026-06-01
Desync between redistribution game phase tracked by node and on-chain (need to investigate)


sbackend123:
Context
An on-chain round is split into commit → reveal → claim. When round N enters claim, the agent publishes the internal sample phase and runs handleSample. At the start of handleSample, IsPlaying() is called — an eth_call to the contract’s isParticipatingInUpcomingRound(owner, committedDepth) — to check whether the node will fall in the neighbourhood for the next round (N+1). Building the sample can take a long time and may overlap with commit of round N+1; IsPlaying is not called again during that work (only once at the beginning of sample).

Problem
The bee_storageincentives_is_playing_errors metric keeps increasing steadily.
Logs on light testnet show phase failed, phase=sample, with errors like
IsPlaying: ... execution reverted: 0xe2586bcc. As I understand, selector decodes to WrongPhase().
In the contract, isParticipatingInUpcomingRound reverts with WrongPhase() only when the call lands in the on-chain reveal phase (currentPhaseReveal()). It should not revert for phase reasons during claim or commit.
If IsPlaying fails, no sample is created → the node does not participate in that game round.

Hypothesis
I suspect a desync between the phase the node believes it is in and the actual on-chain phase. Because of these frequent errors, the node rarely passes the gate before sampling and participates in the game only rarely

sbackend123:
**More context**
The redistribution game runs in cycles of 152 blocks. Each round is divided into three phases:

commit — blocks 0–37
reveal — blocks 38–75
claim — blocks 76–151

**Where the block number comes from**
The RPC call for the block number is cached. Every n blocks (10 by default), BlockNumber fetches the real block from the network via HeaderByNumber. Between synchronizations, the node estimates the current block by adding to the last known number however many blocks have elapsed since the last sync, based on the network’s block-time.

Contract checks (IsPlaying, IsWinner) run against the real latest block via CallContract. Commit, Reveal, and Claim are sent as separate transactions, but they also execute against the current chain state.

That gives two sources of truth, and the node can run a few blocks ahead of the chain in block number.

**What happens on claim entry**
When phaseCheck sees claim, the claim handler runs handleClaim (checking for a win in the current round) and, in parallel, Publish(sample) → handleSample (preparing for the next round).

If no sample is created, the node will not participate in the next round.

**Problem 1:** the node cannot participate in the round (WrongPhase)
[...]

PR #5504, opened 2026-06-17
feat: block time average

sbackend123:
### Checklist

### Description
Added average block time for cache RPC call BlockNumber(). Average block time is computed as (newHeader.Time − prevAnchor.timestamp) / (newHeader.Number − prevAnchor.number) on each cache refresh, using consensus timestamps from the polled headers. If there is no previous anchor (prev.number == 0), or the block number or timestamp did not strictly increase, the configured default block time is used instead.

Observed intervals above maxAverageBlockTime (30 seconds) are capped, so a one-off long gap (chain stall) cannot inflate extrapolation for an extended period. It prevents case when chain got stuck and then cache used "stuck" block average for next N blocks (what can be hours).

On-chain block time is not constant. When the node extrapolates block numbers using a fixed configured block time, the estimate drifts further from the real chain between RPC refreshes. This drift can be significant in redistribution game scenarios, where phase timing depends on accurate block numbers.

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5485

### AI Disclosure

### DIFF
--- a/pkg/transaction/wrapped/wrapped.go
+++ b/pkg/transaction/wrapped/wrapped.go
@@ -18,7 +18,7 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/transaction/wrapped/cache"
 )
 
-const maxAverageBlockTime = 5 * time.Minute
+const maxAverageBlockTime = 30 * time.Second
 
 var _ transaction.Backend = (*wrappedBackend)(nil)
 
--- a/pkg/transaction/wrapped/wrapped.go
+++ b/pkg/transaction/wrapped/wrapped.go
@@ -18,6 +18,8 @@ import (
 	"github.com/ethersphere/bee/v2/pkg/transaction/wrapped/cache"
 )
 
+const maxAverageBlockTime = 5 * time.Minute
+
 var _ transaction.Backend = (*wrappedBackend)(nil)
 
 type blockNumberAnchor struct {
@@ -158,7 +160,12 @@ func (b *wrappedBackend) computeAverageBlockTime(prev blockNumberAnchor, newNumb
 	elapsed := newTime.Sub(prev.timestamp)
 	blocks := newNumber - prev.number
 
-	return elapsed / time.Duration(blocks)
+	averageBlockTime := elapsed / time.Duration(blocks)
+	if averageBlockTime > maxAverageBlockTime {
+		return maxAverageBlockTime
+	}
+
+	return averageBlockTime
 }
 
 func (b *wrappedBackend) HeaderByNumber(ctx context.Context, number *big.Int) (*types.Header, error) {
--- a/pkg/transaction/wrapped/wrapped.go
+++ b/pkg/transaction/wrapped/wrapped.go
@@ -103,7 +103,7 @@ func (b *wrappedBackend) BlockNumber(ctx context.Context) (uint64, error) {
 
 		newNumber := header.Number.Uint64()
 		newTime := time.Unix(int64(header.Time), 0).UTC()
-		averageBlockTime := b.computeAverageBlockTime(prev, newNumber, newTime, time.Now().UTC())
+		averageBlockTime := b.computeAverageBlockTime(prev, newNumber, newTime)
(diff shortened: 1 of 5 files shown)


pkg/transaction/wrapped/wrapped.go
```diff
@@ -126,10 +136,38 @@ func (b *wrappedBackend) estimatedBlockNumberWithElapsed(anchor blockNumberAncho
...
+func (b *wrappedBackend) computeAverageBlockTime(prev blockNumberAnchor, newNumber uint64, newTime time.Time) time.Duration {
+	if prev.number == 0 || newNumber <= prev.number {
+		return b.blockTime
+	}
+	if !newTime.After(prev.timestamp) {
+		return b.blockTime
+	}
+
+	elapsed := newTime.Sub(prev.timestamp)
+	blocks := newNumber - prev.number
```
acud, line 158: check that `blocks != 0`; if chain is stalling, you will get `prev == newNumber`, resulting a division by zero in `L163`

sbackend123: Nope, because L153
```
	if prev.number == 0 || newNumber <= prev.number {
		return b.blockTime
	}
```	

so if the are equal, we will drop to default block time

pkg/transaction/wrapped/wrapped.go
```diff
@@ -126,10 +136,38 @@ func (b *wrappedBackend) estimatedBlockNumberWithElapsed(anchor blockNumberAncho
...
 
+// computeAverageBlockTime returns the observed block time between prev and the
+// freshly loaded header. Falls back to the configured block time only when prev
+// is unset. When the chain block number did not advance, wall-clock elapsed
+// since prev is treated as the time for one block.
+func (b *wrappedBackend) computeAverageBlockTime(prev blockNumberAnchor, newNumber uint64, newTime time.Time) time.Duration {
+	if prev.number == 0 || newNumber <= prev.number {
+		return b.blockTime
+	}
+	if !newTime.After(prev.timestamp) {
```
acud: why not have this check also on the first `if`?

acud (approved):

gacevicljubisa (approved):
LGTM. Nice refinement and improvement on the anchor/extrapolation approach we discussed in [#5394](https://github.com/ethersphere/bee/pull/5394#issuecomment-4067744700)

pkg/transaction/wrapped/wrapped.go
```diff
@@ -126,10 +136,38 @@ func (b *wrappedBackend) estimatedBlockNumberWithElapsed(anchor blockNumberAncho
...
+	if blockTime == 0 {
+		blockTime = b.blockTime
+	}
+
+	elapsedBlocks := uint64(now.Sub(anchor.timestamp) / blockTime)
 	return anchor.number + elapsedBlocks, elapsedBlocks
 }
 
+// computeAverageBlockTime returns the observed block time between prev and the
+// freshly loaded header. Falls back to the configured block time only when prev
```
martinconic: I think comment is outdated because  `it also falls back when newNumber <= prev.number and when !newTime.After(prev.timestamp)`

pkg/transaction/wrapped/wrapped.go
```diff
@@ -126,10 +136,38 @@ func (b *wrappedBackend) estimatedBlockNumberWithElapsed(anchor blockNumberAncho
...
+	}
+	if !newTime.After(prev.timestamp) {
+		return b.blockTime
+	}
+
+	elapsed := newTime.Sub(prev.timestamp)
+	blocks := newNumber - prev.number
+
+	averageBlockTime := elapsed / time.Duration(blocks)
+	if averageBlockTime > maxAverageBlockTime {
```
martinconic, line 161: Should we also have a lower bound as `minAverageBlockTime`, for maybe some misbehaving RPCs returning a large block jump with a near-equal timestamp ?

sbackend123: I do not think it is actually possible...

PR #5508, opened 2026-06-19
test(kademlia): deflake TestAddressBookQuickPrune and TestOutofDepthPrune

martinconic:
### Checklist

### Description
Deflakes two timing-sensitive tests in `pkg/topology/kademlia` that intermittently fail in CI on unrelated branches:

  - `TestAddressBookQuickPrune` — e.g. `kademlia_test.go: timed out waiting for counter ... got 3 want >= 4`
  - `TestOutofDepthPrune` — e.g. `kademlia_test.go: bin 5, got 14, want more than 16`

  Both failures are test-harness timing bugs: the tests gated async connect/prune progress on fixed `time.Sleep` barriers and on the assumption that each `Trigger()` yields exactly one dial, which breaks under load (a trigger landing inside a peer's `TimeToRetry` window is skipped; a 500ms barrier is not always enough on a loaded runner).

**Product change required by this approach.** synctest demands that every goroutine started inside the bubble terminates before the test ends. Kademlia's metrics collector opens an in-memory leveldb in `New` (`shed.NewDB`), but the `*shed.DB` handle was passed to the collector and then dropped — `Close` never closed it. This is a real (if low-severity) resource/goroutine leak that was masked in tests by the `goleak` ignores in `main_test.go`. We now retain the store handle on the `Kad` struct and close it in `Close`, after the collector is finalized. With the store closed, the leveldb compaction/session goroutines exit.

### AI Disclosure

### DIFF
--- a/pkg/topology/kademlia/kademlia.go
+++ b/pkg/topology/kademlia/kademlia.go
@@ -191,6 +191,7 @@ type Kad struct {
 	logger            log.Logger // logger
 	bootnode          bool       // indicates whether the node is working in bootnode mode
 	collector         *im.Collector
+	metricsDB         *shed.DB      // backing store for the metrics collector; closed on shutdown
 	quit              chan struct{} // quit channel
 	halt              chan struct{} // halt channel
 	done              chan struct{} // signal that `manage` has quit
@@ -246,6 +247,7 @@ func New(
 		logger:            logger.WithName(loggerName).Register(),
 		bootnode:          opt.BootnodeMode,
 		collector:         imc,
+		metricsDB:         sdb,
 		quit:              make(chan struct{}),
 		halt:              make(chan struct{}),
 		done:              make(chan struct{}),
@@ -1594,6 +1596,12 @@ func (k *Kad) Close() error {
 		k.logger.Debug("unable to finalize open sessions", "error", err)
 	}
 	k.logger.Debug("metrics collector finalized", "elapsed", time.Since(start))
+
+	if k.metricsDB != nil {
+		if dbErr := k.metricsDB.Close(); dbErr != nil {
+			k.logger.Debug("unable to close metrics store", "error", dbErr)
+		}
+	}
 
 	return err
 }
--- a/pkg/topology/kademlia/kademlia_test.go
+++ b/pkg/topology/kademlia/kademlia_test.go
@@ -919,13 +919,22 @@ func TestAddressBookQuickPrune(t *testing.T) {
 	// bin 1 which is below storageRadius 2, so connectNeighbours skips it).
 	kad.AddPeers(nonConnPeer.Overlay)
 
-	for range kademlia.MaxConnAttempts {
-		time.Sleep(10 * time.Millisecond)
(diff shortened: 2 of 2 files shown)


pkg/topology/kademlia/kademlia_test.go
```diff
@@ -919,13 +919,22 @@ func TestAddressBookQuickPrune(t *testing.T) {
...
-		time.Sleep(10 * time.Millisecond)
+	// keep driving the manage loop until the peer accumulates maxConnAttempts
+	// failed dials. A single Trigger can be wasted if it lands while the peer
+	// is still within its (TimeToRetry) retry window, so we retrigger until the
+	// counter reaches the threshold instead of firing a fixed number of times.
+	// Once pruned at maxConnAttempts, the peer is removed, so the counter cannot
+	// overshoot.
+	err = spinlock.Wait(spinLockWaitTime, func() bool {
 		kad.Trigger()
+		time.Sleep(10 * time.Millisecond)
```
sbackend123: can we use sync.Test somehow here? 
I think, with time.Sleep() test still might be flaky on other CI

martinconic: Good call — done. Both tests are now rewritten on `testing/synctest` instead of `spinlock.Wait`/`time.Sleep`

pkg/topology/kademlia/kademlia.go
```diff
@@ -191,6 +191,7 @@ type Kad struct {
 	logger            log.Logger // logger
 	bootnode          bool       // indicates whether the node is working in bootnode mode
 	collector         *im.Collector
+	metricsDB         *shed.DB      // backing store for the metrics collector; closed on shutdown
```
acud, line 194: i don't understand why this is needed. it does not appear anywhere else in the diff. if this is added because a test needs it then it probably lives in the wrong place. if this is some leftover from another place then probably should be removed.

acud (changes requested):

pkg/topology/kademlia/kademlia.go
martinconic: It's wired up in two other hunks of this same file: it's assigned in New() (metricsDB: sdb, where sdb is the *shed.DB returned by shed.NewDB for the metrics collector), and closed in Close() right after the collector is finalized.

  The reason it's retained at all: the collector's backing leveldb is opened in New() but its handle was previously dropped on the floor — Close() never closed it, so the leveldb compaction/session goroutines outlived the node. It's a real (if low-severity) resource/goroutine leak; today it's just masked in tests by the goleak ignores in main_test.go.

  So it's not a leftover and not test-only state — it's the minimal wiring needed to actually close a store we already own. The synctest rewrite is what surfaced it (the bubble fails if any goroutine started inside it doesn't exit), but the fix is correct independent of the tests. Happy to pull it into a standalone fix(kademlia): PR if you'd rather keep this one test-only — just let me know which you prefer.

acud (approved):

akrem-chabchoub (approved):

Issue #5491, opened 2026-06-04
add address book pruning


acud:
in the recent changes done in the addressbook, a wrapping data structure was added to indicate whether a bzz address was previously verified.

this type should be expanded to add a `last seen` timestamp. this timestamp should be updated every time we see the overlay over hive or in kademlia.

once an address is not seen for over a month - we can remove it from the address book (check on startup). this probably needs a migration that would set all existing addresses to `seen now()`

gacevicljubisa:
Ass addition, maybe it would be more optimal, that before updating the LastSeen, we use Get to check the time diff is more then 24h, or something like this, before we update.

acud:
> Ass addition, maybe it would be more optimal, that before updating the LastSeen, we use Get to check the time diff is more then 24h, or something like this, before we update.

Why? what is the benefit?

gacevicljubisa:
> > Ass addition, maybe it would be more optimal, that before updating the LastSeen, we use Get to check the time diff is more then 24h, or something like this, before we update.
> 
> Why? what is the benefit?

If we have to often updates of LastSeen, in case if we want to update when peer is detected in the hive, which is preaty agressive, this wlll create big amount of IO. We could GET to check the time diff is more then 24h, where db could use its cache, and write only if more then 24 hours.

janos:
> > > Ass addition, maybe it would be more optimal, that before updating the LastSeen, we use Get to check the time diff is more then 24h, or something like this, before we update.
> > 
> > 
> > Why? what is the benefit?
> 
> If we have to often updates of LastSeen, in case if we want to update when peer is detected in the hive, which is preaty agressive, this wlll create big amount of IO. We could GET to check the time diff is more then 24h, where db could use its cache, and write only if more then 24 hours.

I think that the IO concern is a valid one. The UpdateLastSeen method https://github.com/ethersphere/bee/pull/5513/changes#diff-b817056e32073e46ecb398adec8fcde59e8a72d424d90ea9e213039dbeb7b3a6R110, already calls both Get and Put on state storer. The check that @gacevicljubisa suggested can avoid Put calls with a just a few lines there.

[...]

PR #5513, opened 2026-06-22
feat(addressbook): prune stale entries by last-seen time

martinconic:
### Checklist

### Description
Implements address book pruning (#5491).

The address book recently gained a wrapping verifiedAddress{Address, Verified} struct (v2.8.0, #5477). This PR extends it with a last-seen timestamp and uses it to drop peers we have not seen for over a month, preventing the address book from accumulating stale, unreachable peers indefinitely.

Changes

- pkg/addressbook
  - Add LastSeen (unix seconds) to verifiedAddress.
  - Put stamps LastSeen with the current time on every write (initial insertion and any record update).
  - New Seen(overlays ...swarm.Address): read-modify-write that bumps only the timestamp. Overlays that are absent are skipped, and the write is throttled: an overlay already seen within the last 24h is left untouched, so per-sighting calls don't amplify disk IO (on mainnet, sightings would otherwise mean tens of thousands of redundant writes a day).
  - Put and Seen are serialized by a mutex, so a concurrent Put is never rolled back by Seen's stale read.
  - Entries last seen more than 30 days ago are pruned when the store is opened, inside the constructor, before the store is shared with any writer. Entries with LastSeen == 0 are kept for a later run. Best-effort: a pruning failure never blocks node startup.
  - Inject a clock for testability.
- pkg/hive: hearing about a peer we already know over gossip is a sighting in its own right, so Seen is called for every known peer before timestamp validation, whether or not the re-presented record is newer than the one we hold. Peers mint their bzz.Address once and re-gossip it unchanged for their whole uptime, so this is the path that keeps gossip-only peers fresh; the existing Put only fires when the record is genuinely newer. Failures are debug-logged and non-fatal.
- pkg/topology/kademlia: connected peers are "constantly seen" without generating gossip, so the manage loop marks every currently connected peer as seen every 15 minutes. Failures are warning-logged and non-fatal.
- pkg/statestore/storeadapter: migration step 10 stamps last_seen = now() onto existing entries that lack it, decoding into the current entry type (whose last_seen is omitempty) so all other fields round-trip unchanged. Without this, the first prune after upgrade would wipe the whole address book.

### AI Disclosure


gacevicljubisa:
Here, prune here runs once, at startup only, and it can lead that if we have some stable connections over bigger period of time, where we haven't updated LastSeen, it will remove those on next restart. But those are our most valuable connections. 

Maybe we could use `manage()` method in kademlia, where we could trigger prune addressbook every 24h (those older then 30 days), but excluding all connected peers, and prune should not start before warmup is done (when node is started). Also, we could stamp last contact at the disconnect, so the prune clock starts from when the peer was actually lost. 

One gap is left, what do to with the peer we only hear over hive gossip, but we didn't connect to?

pkg/statestore/storeadapter/migration.go
```diff
@@ -118,3 +121,65 @@ func rewriteAddressbookEnvelope(s storage.Store) migration.StepFn {
...
+			})
+			return false, nil
+		}); err != nil {
+			return fmt.Errorf("iterate addressbook entries: %w", err)
+		}
+
+		now := time.Now().Unix()
+
+		for _, e := range batch {
+			var fields map[string]json.RawMessage
```
acud: this is not needed. you can shadow the type which is currently on `master` as a local type scoped to the function, and use it directly to unmarshal and conversion onto the new type. when i think of it, you could even have the latest version (just annotate the last seen field with the json omitempty instruction), and you could unmarshal the existing old serialization format into it, just set the timestamp and write it back into the db.

pkg/topology/kademlia/kademlia.go
```diff
@@ -455,6 +455,10 @@ func (k *Kad) connectionAttemptsHandler(ctx context.Context, wg *sync.WaitGroup,
 
 		k.connectedPeers.Add(peer.addr)
 
+		if err := k.addressBook.UpdateLastSeen(peer.addr); err != nil {
+			k.logger.Debug("could not update last seen for peer", "peer_address", peer.addr, "error", err)
```
acud: maybe warn?

pkg/topology/kademlia/kademlia.go
```diff
@@ -1214,6 +1218,9 @@ func (k *Kad) onConnected(ctx context.Context, addr swarm.Address) error {
 	k.knownPeers.Add(addr)
 	k.connectedPeers.Add(addr)
 	k.waitNext.Remove(addr)
+	if err := k.addressBook.UpdateLastSeen(addr); err != nil {
+		k.logger.Debug("could not update last seen for peer", "peer_address", addr, "error", err)
```
acud: warn?

acud:
> Here, prune here runs once, at startup only, and it can lead that if we have some stable connections over bigger period of time, where we haven't updated LastSeen, it will remove those on next restart. But those are our most valuable connections.
> 
> Maybe we could use `manage()` method in kademlia, where we could trigger prune addressbook every 24h (those older then 30 days), but excluding all connected peers, and prune should not start before warmup is done (when node is started). Also, we could stamp last contact at the disconnect, so the prune clock starts from when the peer was actually lost.

This is what we've converged to with @janos as a naive starting point to ship.

> One gap is left, what do to with the peer we only hear over hive gossip, but we didn't connect to?

For now let's leave them in. The first line of defense is to get rid of garbage which is in the statestore. A peer which is seen over hive and is not connected to is still a potential peer you'd like to connect to. I agree that over time yes, probably you don't wanna keep those. I suggest to ship this first and see it works, then evolve into more well defined scenarios.

martinconic:
Added  two fixes:
 1. Hive never refreshes LastSeen. Peers mint their bzz.Address once and re-gossip it unchanged, so CheckTimestamp rejects it as not-newer and Put never fires again — LastSeen freezes at first-seen, and gossip-only peers get pruned at 30 days.  Fix: call UpdateLastSeen on ErrTimestampStale/ErrTimestampTooSoon (both imply a known peer). Other rejections don't refresh. Safe against reviving dead peers, since hive only gossips currently-connected peers.
 2. UpdateLastSeen is a lost update, and fix 1 makes it reachable. Get→mutate→Put with no lock; hive runs a goroutine per gossip message, so it now races its own Put for the same overlay. Worst case: stale write-back desyncs the addressbook from the chequebook registry (swap-on only) until re-mint or restart. Fix: mutex on the store, taken by Put and UpdateLastSeen. Regression test fails without it.

pkg/addressbook/addressbook.go
```diff
@@ -17,34 +19,56 @@ import (
 
 const keyPrefix = "addressbook_entry_"
 
+// lastSeenUpdateInterval is the resolution at which UpdateLastSeen persists.
+// Hive reports the same peer on every gossip round, while pruning operates on
+// a scale of weeks, so a write is skipped when the stored value is already
+// this fresh.
+const lastSeenUpdateInterval = 24 * time.Hour
```
acud: don't understand whether this is truly needed. if i see a peer overlay at `time.Now()`, i set the time last seen to `now()`. why do we need to add all of this buffering and logic complications? unless there's really an issue with IO resulting from so many writes (i suspect not) then this is a non-issue.

pkg/addressbook/addressbook.go
```diff
@@ -17,34 +19,56 @@ import (
...
+	// was seen over hive or in kademlia. Used to prune stale entries.
+	LastSeen int64 `json:"last_seen"`
 }
 
 // Interface is the AddressBook interface.
 type Interface interface {
 	GetPutter
 	Remover
+	// UpdateLastSeen marks the overlay as seen at the current time.
+	UpdateLastSeen(overlay swarm.Address) error
```
acud: `Seen(peer)`?

pkg/addressbook/addressbook.go
```diff
@@ -63,12 +87,18 @@ type Remover interface {
 
 type store struct {
 	store storage.StateStorer
+	now   func() time.Time
+
+	// mu serializes the read-modify-write in UpdateLastSeen against Put, so a
+	// concurrent Put is not rolled back by a stale copy of the entry.
+	mu sync.Mutex
```
acud, line 101: i don't see why this is needed. this is not business critical information and i think it is fine with this hypothetical use case happening every now and then, if ever.

pkg/addressbook/addressbook.go
```diff
@@ -86,17 +116,80 @@ func (s *store) Get(overlay swarm.Address) (*bzz.Address, bool, error) {
...
+
 func (s *store) Remove(overlay swarm.Address) error {
 	return s.store.Delete(keyPrefix + overlay.String())
 }
 
+// Prune removes all entries whose overlay has not been seen since before.
+// Entries without a recorded last-seen time (LastSeen == 0) are kept, leaving
+// pruning to a later run once they have been observed.
+func (s *store) Prune(before time.Time) error {
+	cutoff := before.Unix()
```
acud, line 198: you added a lock but you don't use it here. i suggest to remove the lock because the edge case it tries to handle is _very_ hypothetical and inconsequential

pkg/hive/hive.go
```diff
@@ -370,6 +370,17 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
...
 
 		if err := bzz.CheckTimestamp(bzzAddress.Timestamp, existing, bzz.TimestampSourceGossip, s.now()); err != nil {
+			// A peer re-presenting a record we already hold has nothing new to
+			// store, but it is still a sighting: peers mint their bzz.Address
+			// once and gossip it unchanged for their whole uptime. Refresh
+			// last-seen so peers we keep hearing about, but never dial, do not
+			// look stale to the pruner. Both errors imply a known peer, since
+			// an unknown one short-circuits inside CheckTimestamp.
+			if errors.Is(err, bzz.ErrTimestampStale) || errors.Is(err, bzz.ErrTimestampTooSoon) {
+				if err := s.addressBook.UpdateLastSeen(overlayAddr); err != nil {
```
acud: i suggest to update the timestamp outside this error check block, because you anyway update the seen time even on a rejected timestamp here, then why not do it anyway before the check timestamp?

pkg/hive/lastseen_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	if puts, updates := spy.counts(); puts != 2 || updates != 0 {
+		t.Fatalf("newer record: puts=%d updates=%d, want 2/0", puts, updates)
+	}
+}
+
+// TestGossipDoesNotRefreshLastSeenOnInvalidRecord makes sure the refresh is
+// scoped to records that merely are not newer. A record rejected for any other
+// reason is not evidence that the peer is alive.
+func TestGossipDoesNotRefreshLastSeenOnInvalidRecord(t *testing.T) {
```
acud: it would be good to avoid WritingWholeSentencesInTestCaseNames. that's what we have comments for :)

pkg/node/node.go
```diff
@@ -382,6 +383,12 @@ func NewBee(
 
 	addressbook := addressbook.New(stateStore)
 
+	// Prune addressbook entries whose overlays have not been seen recently, so
+	// the address book does not accumulate stale peers indefinitely.
+	if err := addressbook.Prune(time.Now().Add(-addressbookPruneAfter)); err != nil {
```
acud: since you prune only on startup - there's no need to expose the method over an interface - just make sure that addressbook ctor runs the pruning internally.

pkg/topology/kademlia/kademlia.go
```diff
@@ -455,6 +455,10 @@ func (k *Kad) connectionAttemptsHandler(ctx context.Context, wg *sync.WaitGroup,
 
 		k.connectedPeers.Add(peer.addr)
 
+		if err := k.addressBook.UpdateLastSeen(peer.addr); err != nil {
```
acud: the changes in kademlia only handle updating on connection event, however this is only partially true. since peers we are connected to are "constantly seen". i suggest to have one update location inside the kademlia's manage loop that fires every now and then, and just iterate over the whole list of connected peers and mark them as seen. in that same breath, it might be useful to make `Seen` to be variadic, like `Seen(...peer)`

acud (changes requested):

pkg/hive/hive.go
```diff
@@ -369,6 +369,18 @@ func (s *Service) checkAndAddPeers(ctx context.Context, peers pb.Peers) {
 			continue
 		}
 
+		// Hearing about a peer we already know is a sighting in its own right,
```
sbackend123, line 372: it seems Seen() now fires even when the incoming record is later rejected for being in the future (ErrTimestampInFuture) or invalid (ErrTimestampInvalid). These aren't meaningful sightings — a bogus record from a third party shouldn't refresh a legitimate peer's last-seen. The old placement after CheckTimestamp, gated on ErrTimestampStale || ErrTimestampTooSoon, was more precise.

Was moving it before CheckTimestamp intentional?

pkg/addressbook/addressbook.go
```diff
@@ -86,17 +114,81 @@ func (s *store) Get(overlay swarm.Address) (*bzz.Address, bool, error) {
...
 func (s *store) Remove(overlay swarm.Address) error {
 	return s.store.Delete(keyPrefix + overlay.String())
 }
 
+// Prune removes all entries from storer whose overlay has not been seen since
+// before. Entries without a recorded last-seen time (LastSeen == 0) are kept,
+// leaving them to a later run once they have been observed. It is a one-shot
+// maintenance step meant to run at startup, before the address book is shared
+// with its writers, so it takes no lock.
+func Prune(storer storage.StateStorer, before time.Time) error {
```
acud: i still argue this method does not need to be exported - you can just call it from within the constructor when you initially build the store.

acud (approved):
minus the comment about exporting `Prune` (which should anyway be a method, not a pure function), this looks good from my side

pkg/hive/hive.go
acud: I think it was part of the addressing of my previous comments. The address still passes _some_ validation until we reach this point. And it could be that the new address was minted and sent not out of malicious behavior. I am fine whatever you two decide about this - have it before or after does not matter from my side, originally I think it was _inside_ the error check after the `CheckTimestamp` and I wanted to avoid a situation where the errors dictate whether the peer gets marked seen yes/no. Before or after is better from my perspective, I don't really mind which one.

acud: (left side [here](https://github.com/ethersphere/bee/pull/5513/changes/699b4d7a920347a847cf85b4a447df694146c20a#diff-f2029eb5af2437e778c69dd0f6a62d5c26a1a22b880935af73caf8ba3f3b73c7L379))

gacevicljubisa (approved):
Nice work. 

@acud suggested not have 24h update check, so at least in that case, please remove this information from the PR desc before merge.

gacevicljubisa:
> Nice work.
> 
> @acud suggested not have 24h update check, so at least in that case, please remove this information from the PR desc before merge.

Additionally I would still suggest 24h throtle to have it, becuse we have noticed increased disk writes when we we wrote TImestamp on every handshake... If my [query](https://grafana.proxy.ethswarm.org/explore?schemaVersion=1&panes=%7B%22hct%22:%7B%22datasource%22:%22adx13wbz0zhmob%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22sum%20by%20%28pod%29%20%28increase%28bee_hive_timestamp_rejected_total%7B%20pod%3D~%5C%22bee-1-1-0%5C%22,%20reason%3D~%5C%22stale%7Ctoo_soon%5C%22%7D%5B24h%5D%29%29%22,%22range%22:false,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22adx13wbz0zhmob%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22,%22format%22:%22table%22,%22exemplar%22:false,%22interval%22:%22%22%7D%5D,%22range%22:%7B%22from%22:%221784545200000%22,%22to%22:%221784631600000%22%7D,%22compact%22:false%7D%7D&orgId=1) is good, for this mainnet node, we would have more then 90k of seen events in a day, 99% redudant.

acud:
> > Nice work.
> > @acud suggested not have 24h update check, so at least in that case, please remove this information from the PR desc before merge.
> 
> Additionally I would still suggest 24h throtle to have it, becuse we have noticed increased disk writes when we we wrote TImestamp on every handshake... If my [query](https://grafana.proxy.ethswarm.org/explore?schemaVersion=1&panes=%7B%22hct%22:%7B%22datasource%22:%22adx13wbz0zhmob%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22sum%20by%20%28pod%29%20%28increase%28bee_hive_timestamp_rejected_total%7B%20pod%3D~%5C%22bee-1-1-0%5C%22,%20reason%3D~%5C%22stale%7Ctoo_soon%5C%22%7D%5B24h%5D%29%29%22,%22range%22:false,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22adx13wbz0zhmob%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22,%22format%22:%22table%22,%22exemplar%22:false,%22interval%22:%22%22%7D%5D,%22range%22:%7B%22from%22:%221784545200000%22,%22to%22:%221784631600000%22%7D,%22compact%22:false%7D%7D&orgId=1) is good, for this mainnet node, we would have more then 90k of seen events in a day, 99% redudant.

hmm... ok.
but a sidenote - we cannot ignore a stale timestamp and potentially also a "too soon" timestamp - the issuing client might not know it has issued "too soon" and stale timestamp is misleading - we cannot treat this error as an error in this case - as you may receive an older timestamp via gossip, that's legit behavior and should count as "seen".

gacevicljubisa:
> > > Nice work.
> > > @acud suggested not have 24h update check, so at least in that case, please remove this information from the PR desc before merge.
> > 
> > 
> > Additionally I would still suggest 24h throtle to have it, becuse we have noticed increased disk writes when we we wrote TImestamp on every handshake... If my [query](https://grafana.proxy.ethswarm.org/explore?schemaVersion=1&panes=%7B%22hct%22:%7B%22datasource%22:%22adx13wbz0zhmob%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22sum%20by%20%28pod%29%20%28increase%28bee_hive_timestamp_rejected_total%7B%20pod%3D~%5C%22bee-1-1-0%5C%22,%20reason%3D~%5C%22stale%7Ctoo_soon%5C%22%7D%5B24h%5D%29%29%22,%22range%22:false,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22adx13wbz0zhmob%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22,%22format%22:%22table%22,%22exemplar%22:false,%22interval%22:%22%22%7D%5D,%22range%22:%7B%22from%22:%221784545200000%22,%22to%22:%221784631600000%22%7D,%22compact%22:false%7D%7D&orgId=1) is good, for this mainnet node, we would have more then 90k of seen events in a day, 99% redudant.
> 
> hmm... ok. but a sidenote - we cannot ignore a stale timestamp and potentially also a "too soon" timestamp - the issuing client might not know it has issued "too soon" and stale timestamp is misleading - we cannot treat this error as an error in this case - as you may receive an older timestamp via gossip, that's legit behavior and should count as "seen".

Yes. I totaly agree. This metric is capturing just "stale" and "too soon", and real "seen" will be even more then this.

acud (approved):

Issue #5490, opened 2026-06-04
hive gossip improvements


acud:
Right now there's a few limits regarding how and what we gossip on hive.

- `randomSubset` in `kademlia` - this was set when the amount of peers gossiped was so large that it was consuming way too much bandwidth. We should set this to be a more reasonable higher value (but also have a cap in a way that closer bins can still get gossiped correctly and fully)
- add write coalescing in `hive` - when frequent connections occur, it is very easy that single peer gossip messages get sent. This is wasteful. We should emit a message every second or so when gossiping single peers (if the message is big it can be dispatched immediately). Single peer gossips wait

sbackend123:
I have additional improvement:

Currently, in distant bins (below storageRadius), we pick 2 random connected peers per bin (BroadcastBinSize = 2).

This ignores two aspects:
- A bin may have 5 peers or 20. Sharing with 2 means ~40% coverage in the first case and only 10% in the second.
- “Distant” bins can sit next to the neighborhood boundary or be as far as possible.

I think, it makes more sense to  follow: the closer the bin to the neighborhood, the higher the coverage. 

So I suggest , in short:  we replace flat “BroadcastBinSize per bin” with a total fanout_cap on distant bins only, weighted toward bins closer to storageRadius, with random picks within each bin’s quota; neighborhood gossip is unchanged.

Example: with 12 active (not empty) shallow bins and BroadcastBinSize = 2 , so  today ~12 × 2 = 24 recipients, which will be shared info bout new peer with.

With fanout_cap = 36 (same total as flat 3 per bin):
bin 11 — up to 6
bin 10 — up to 5
…
bin 0 — 1
Total ≤ 36.

We get faster, denser discovery near the neighborhood boundary and in “almost-neighbor” bins with same traffic.
@acud @gacevicljubisa @janos

PR #5520, opened 2026-06-26
feat: gossip improvements

sbackend123:
### Checklist

### Description
Adds write coalescing for hive outbound gossip. Single-peer BroadcastPeers calls are buffered per addressee and flushed as one batched message after ~1s (configurable via GossipCoalesceInterval), or immediately when the buffer reaches maxBatchSize (30). Calls with 2+ peers are sent without coalescing

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5490

### AI Disclosure


pkg/hive/hive.go
```diff
@@ -136,35 +147,82 @@ func (s *Service) Protocol() p2p.ProtocolSpec {
...
 
 		// If broadcasting limit is exceeded, return early
 		if !s.outLimiter.Allow(addressee.ByteString(), maxSize) {
-			return nil
+			return total - len(peers), nil
 		}
 
 		select {
+		case <-ctx.Done():
+			return total - len(peers), ctx.Err()
```
acud: why do you care about the count when context is cancelled? also, why do you care about the count when an error is returned in general (below, shutting down/other send error)

pkg/hive/gossip_buffer.go
```diff
@@ -0,0 +1,121 @@
...
+	coalesceThreshold = 2
+)
+
+// gossipBuffer accumulates single-peer outbound gossip per addressee so it can be
+// flushed as one batched message.
+type gossipBuffer struct {
+	mu       sync.Mutex
+	pending  map[string]*pendingGossip // addressee bytestring -> buffered peers
+	interval time.Duration
+	jitter   time.Duration
```
acud: why is this needed?

pkg/hive/gossip_buffer.go
```diff
@@ -0,0 +1,121 @@
...
+// removed and returned so the caller can flush it immediately.
+func (b *gossipBuffer) add(now time.Time, addressee swarm.Address, peers ...swarm.Address) *pendingGossip {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	key := addressee.ByteString()
+	e, ok := b.pending[key]
+	if !ok {
+		var jitter time.Duration
+		if b.jitter > 0 {
```
acud: not sure why this is needed...

pkg/hive/gossip_buffer.go
```diff
@@ -0,0 +1,121 @@
...
+	pending  map[string]*pendingGossip // addressee bytestring -> buffered peers
+	interval time.Duration
+	jitter   time.Duration
+	maxBatch int
+}
+
+type pendingGossip struct {
+	addressee swarm.Address
+	peers     map[string]swarm.Address // peer bytestring -> address (set semantics)
+	deadline  time.Time
```
acud: i wonder whether there's a benefit of keeping a deadline per peer. usually, write coalescing is simple enough:
- have an interval fire at a constant rate
- if new records arrived by interval fire
 - when new entries arrived, optionally, postpone the sending the sending until the next interval firing (and so also you could extend up to a set upper bound, so that entries don't keep collecting forever but also guarantee that information goes out still relatively quickly)
- send all the pending sends

also: usually, when a peer arrives - we gossip that peer to all peers (full nodes) and gossip to that peer all of our connected peers.
this in turn means that sending is almost always involving all connected peers. which in turn also means that the timestamps on the individual pendingGossip entries would be almost identical (making the field even more so redundant)

gacevicljubisa (commented):
Nice work.

One idea: instead of deciding what to do based on how many peers are passed in, it might be cleaner to add a separate method like `GossipPeer(addressee, peer)` on `discovery.Driver` just for the buffered case. Make `BroadcastPeers` plain => `send now and return an error method`, and use `GossipPeer` as async (fire and forget) in [kademlia.go:1082-1090](https://github.com/ethersphere/bee/blob/ea9a08fdb751348db3c2245e648896cb0d208eea/pkg/topology/kademlia/kademlia.go#L1082-L1090) and flush in `startGossipCoalescer` worker. This allows you to drop `coalesceThreshold`. Also, the "buffer is full" logic, you can also move to the background worker with a `wakeup` channel so buffering never blocks or sends directly.

pkg/hive/gossip_buffer.go
```diff
@@ -0,0 +1,101 @@
...
+
+// stagePeers buffers peers for the addressee. If the buffer reaches maxBatch it is
+// removed and returned so the caller can flush it immediately.
+func (b *gossipBuffer) stagePeers(addressee swarm.Address, peers ...swarm.Address) (flushPeers []swarm.Address, flush bool) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	key := addressee.ByteString()
+	peerSet, ok := b.pending[key]
+	if !ok {
```
acud: since `map`s [leak memory by design](https://medium.com/@caring_smitten_gerbil_914/go-maps-and-hidden-memory-leaks-what-every-developer-should-know-17b322b177eb), it would be better to:
1. check whether the b.pending key exists
2. merge its contents with `peers` if it does _before_ writing the key to the map
3. check the max batch size and if the entry really exceeds max batch - return early without writing to the map
4. finally if we are within the bounds of the max batch - write to map

pkg/hive/hive.go
```diff
@@ -296,9 +350,55 @@ func (s *Service) peersHandler(ctx context.Context, peer p2p.Peer, stream p2p.St
...
+
+func (s *Service) flushGossipBatch(addressee swarm.Address, peers []swarm.Address, reason string) {
+	s.recordCoalesceFlush(reason, addressee, peers)
+
+	ctx, cancel := context.WithTimeout(context.Background(), messageTimeout)
+	err := s.broadcastNow(ctx, addressee, true, peers...)
+	if err != nil {
+		s.logger.Debug("coalesced gossip flush failed", "addressee", addressee, "reason", reason, "batch_size", len(peers), "error", err)
+	}
+	cancel()
```
acud: ideally this should be a defer call just after it gets created.

pkg/hive/hive.go
```diff
@@ -296,9 +350,55 @@ func (s *Service) peersHandler(ctx context.Context, peer p2p.Peer, stream p2p.St
...
 
+func (s *Service) startGossipCoalescer() {
+	s.wg.Go(func() {
+		ticker := time.NewTicker(s.gossipBuf.interval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				for _, batch := range s.gossipBuf.takeAll() {
+					s.flushGossipBatch(batch.addressee, batch.peers, coalesceFlushReasonTimer)
```
acud: nit - this is a blocking call that makes slower peers to block other peers from getting the information. i would tend to turn this into `go s.flushGossipBatch`. iirc the latest go compilers make sure the values get copied correctly such that when the iterator changes `batch` values it doesn't change the underlying value for the goroutines already dispatched with that same variable name. but maybe also putting this into a closure won't hurt too much.

pkg/hive/gossip_buffer.go
```diff
@@ -0,0 +1,101 @@
...
+	"maps"
+	"slices"
+	"sync"
+	"time"
+
+	"github.com/ethersphere/bee/v2/pkg/swarm"
+)
+
+const (
+	defaultGossipCoalesceInterval = time.Second
```
acud: nit: i think this can be higher (like 5 sec)? the same for the coalesce threshold - we want to have bigger messages and less often. the timer fires every 5 seconds anyway.

acud (changes requested):

pkg/hive/gossip_buffer.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	out := make([]gossipBatch, 0, len(b.pending))
+	for key, peerSet := range b.pending {
+		out = append(out, gossipBatch{
+			addressee: swarm.NewAddress([]byte(key)),
+			peers:     slices.Collect(maps.Values(peerSet)),
+		})
+	}
+	b.pending = make(map[string]map[string]swarm.Address)
```
acud: `clear(b.pending)`

sbackend123:
> Nice work.
> 
> One idea: instead of deciding what to do based on how many peers are passed in, it might be cleaner to add a separate method like `GossipPeer(addressee, peer)` on `discovery.Driver` just for the buffered case. Make `BroadcastPeers` plain => `send now and return an error method`, and use `GossipPeer` as async (fire and forget) in [kademlia.go:1082-1090](https://github.com/ethersphere/bee/blob/ea9a08fdb751348db3c2245e648896cb0d208eea/pkg/topology/kademlia/kademlia.go#L1082-L1090) and flush in `startGossipCoalescer` worker. This allows you to drop `coalesceThreshold`. Also, the "buffer is full" logic, you can also move to the background worker with a `wakeup` channel so buffering never blocks or sends directly.

Nice idea, but looks like more changes than we need (?)
@acud wdyt?

acud:
> wdyt

not sure... also having a buffered vs unbuffered kinda defeats the purpose of batching the writes together. not sure if i see the case of non-buffered broadcast as needed. the thresholds should take care of that already - when the group is big enough - it is urgent enough (and that would always be the case for a peer that connects and gets a bunch of peers via gossip). so we can use the implicit behavior in this case instead of expanding the interfaces. my 2 cents

pkg/hive/hive.go
```diff
@@ -137,34 +149,77 @@ func (s *Service) Protocol() p2p.ProtocolSpec {
...
-	maxSize := maxBatchSize
+	if len(peers) == 0 {
+		return nil
+	}
+
 	s.metrics.BroadcastPeers.Inc()
 	s.metrics.BroadcastPeersPeers.Add(float64(len(peers)))
 
+	// Already-batched messages go out immediately; single-peer gossips are coalesced.
+	if len(peers) >= coalesceThreshold {
```
acud, line 170: if the addressee already has some peers which are queued for sending through the buffer - they are silently skipped here. not a big issue, can be handled later too. flagging this nevertheless. 

also, you might want to gossip, but the rate limiter won't allow you to send the whole batch together because you can't get enough tokens from the bucket. this puts things as a best effort. i'm not sure we should handle all those edge cases right away but they are definitely worth documenting at least inline and perhaps as a follow up issue.

acud (approved):
lgtm

pkg/hive/hive.go
```diff
@@ -297,9 +352,59 @@ func (s *Service) peersHandler(ctx context.Context, peer p2p.Peer, stream p2p.St
...
 
+func (s *Service) startGossipCoalescer() {
+	s.wg.Go(func() {
+		ticker := time.NewTicker(s.gossipBuf.interval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				for _, batch := range s.gossipBuf.takeAll() {
+					go func(batch gossipBatch) {
```
aloknerurkar: The goroutines started here are not tracked using the waitgroup. If a shutdown happens in worst case these routines will wait till messageTimeout to exit and keep trying to write on p2p. Currently there is no test to Close mid-flush which is why we don't find this leak in the tests.

We should create these go routines using the waitgroup. Also maybe a test to verify everything closes correctly mid-flush.

martinconic: Untracked goroutine, this will bypass s.wg:
 - go func(batch) is not wrapped in s.wg.Go(...).
- When the node shuts down, Service.Close() closes s.quit and calls s.wg.Wait().
- s.wg.Wait() immediately finishes without waiting for the flushGossipBatch goroutines.
- Inside  flushGossipBatch

```go
ctx, cancel := context.WithTimeout(context.Background(), messageTimeout) // 1 minute timeout!
```
The orphaned goroutines keep attempting to dial peers and open streams on the libp2p transport for up to 60 seconds after the node has shut down.

pkg/hive/hive.go
```diff
@@ -137,34 +149,77 @@ func (s *Service) Protocol() p2p.ProtocolSpec {
...
 	for len(peers) > 0 {
 		if maxSize > len(peers) {
 			maxSize = len(peers)
 		}
 
-		// If broadcasting limit is exceeded, return early
 		if !s.outLimiter.Allow(addressee.ByteString(), maxSize) {
+			if coalesced {
+				s.metrics.GossipCoalesceDropped.Add(float64(len(peers)))
+			}
```
aloknerurkar, line 207: Just checking if its worth to add a debug log here rather than silently dropping.

martinconic:
Is the following scenario a valid one?

The Scenario
At time = 0s: Single peer PeerA is announced -> buffered in gossipBuf for RemoteNode.
At time = 1s: A batch of 5 peers [P1, P2, P3, P4, P5] is broadcast -> since length is 5 or more, it bypasses the buffer and sends immediately via Stream 1.
RemoteNode receives [P1..P5] before receiving PeerA.
At time = 5s: The timer fires and sends PeerA in Stream 2.
```
Time 0s: [Buffer: PeerA]
Time 1s: [Immediate Broadcast: P1, P2, P3, P4, P5] ----> Sent in Stream #1
Time 5s: [Coalescer Flush: PeerA] --------------------> Sent in Stream #2
```

Consequences:
Older gossip arrives after newer gossip.
If P1 was also in the buffer, it gets transmitted twice (Stream 1 + Stream 2).
If RemoteNode disconnects at time = 3s, PeerA is purged on disconnect and never delivered, even though later peers were delivered.

pkg/hive/hive.go
```diff
@@ -137,34 +149,77 @@ func (s *Service) Protocol() p2p.ProtocolSpec {
...
+func (s *Service) broadcastNow(ctx context.Context, addressee swarm.Address, coalesced bool, peers ...swarm.Address) error {
+	maxSize := maxBatchSize
+
 	for len(peers) > 0 {
 		if maxSize > len(peers) {
 			maxSize = len(peers)
 		}
 
-		// If broadcasting limit is exceeded, return early
 		if !s.outLimiter.Allow(addressee.ByteString(), maxSize) {
```
martinconic, line 206: Because `takeAll()` already removed all buffered peers from `b.pending`, if `s.outLimiter.Allow(..., 30)` returns `false` (when available tokens < 30), `broadcastNow` returns `nil` and all 30 peers are permanently lost.

Since `limitRate` is `time.Minute` (1 token/minute refill), if a bucket is depleted, a node will drop all 30-peer batches for up to 30 minutes.

We should re-enqueue the rejected peers back into the buffer so they can be sent on subsequent ticks or when tokens become available

pkg/hive/hive.go
```diff
@@ -137,34 +149,77 @@ func (s *Service) Protocol() p2p.ProtocolSpec {
 
 var ErrShutdownInProgress = errors.New("shutdown in progress")
 
+// BroadcastPeers sends peer gossip to the addressee. Calls with fewer than
+// coalesceThreshold peers are buffered and flushed asynchronously; errors
+// during deferred dispatch are logged but not returned to the caller.
+// Calls with coalesceThreshold or more peers are sent immediately.
 func (s *Service) BroadcastPeers(ctx context.Context, addressee swarm.Address, peers ...swarm.Address) error {
```
martinconic, line 161: If a caller invokes `BroadcastPeers` with a context that is already canceled or has a short deadline, the check is bypassed and the peer is buffered anyway, returning `nil`. We should validate `ctx.Err()` before staging into memory

aloknerurkar (commented):
I think it now makes more sense to have one bgCtx in the service which is cancelled on quit. There seems to be a lot of cases where we read ctx.Done + s.quit. This can be condensed. Maybe we can do it in a separate PR.

gacevicljubisa (commented):
> not sure if i see the case of non-buffered broadcast as needed. the thresholds should take care of that already - when the group is big enough - it is urgent enough

One thing to check: on a small cluster (beekeeper, or any node right after start, since only confirmed-reachable peers count) the introduction list at kademlia.go:1155 has fewer than 5 peers, so it gets buffered and returns nil. Does the `Disconnect("failed broadcasting to peer")` branch below it ever run then, and what happens to the list if the peer drops before the tick?

And the reason why I suggested `GossipPeer` is that it lets the caller say "this one can wait" instead of hive guessing from the count. Announce/AnnounceTo use it for the single-peer gossip, `BroadcastPeers` stays send-now-and-return-the-error.

PR overall looks good to me. Just address the comments regarding the untracked goroutines.

gacevicljubisa:
One more thing at hive.go:199: the batch asks for all its tokens at once, and if the limiter says no the whole batch is dropped, not put back. After an addressee's burst is spent (refill is 1 token/min) nothing gets through until 30 tokens accumulate, ~30 minutes for a full batch, and every batch tried in between is lost.

sbackend123:
> Is the following scenario a valid one?
> 
> The Scenario At time = 0s: Single peer PeerA is announced -> buffered in gossipBuf for RemoteNode. At time = 1s: A batch of 5 peers [P1, P2, P3, P4, P5] is broadcast -> since length is 5 or more, it bypasses the buffer and sends immediately via Stream 1. RemoteNode receives [P1..P5] before receiving PeerA. At time = 5s: The timer fires and sends PeerA in Stream 2.
> 
> ```
> Time 0s: [Buffer: PeerA]
> Time 1s: [Immediate Broadcast: P1, P2, P3, P4, P5] ----> Sent in Stream #1
> Time 5s: [Coalescer Flush: PeerA] --------------------> Sent in Stream #2
> ```
> 
> Consequences: Older gossip arrives after newer gossip. If P1 was also in the buffer, it gets transmitted twice (Stream 1 + Stream 2). If RemoteNode disconnects at time = 3s, PeerA is purged on disconnect and never delivered, even though later peers were delivered.

Yes, this scenario is possible, but in my opinion it would only be a problem in a very sparse or unstable topology.
Under normal conditions, the delivery order of peer gossip does not matter. Even if an announcement is lost on disconnect, the peer will most likely be advertised later by other peers, or discovered through subsequent gossip or a handshake.
The point about deduplication is valid, but I would rather not complicate the current solution. Deduplication already exists at the buffer level; we can observe mainnet behavior and extend the implementation later if needed.

pkg/hive/hive.go
```diff
@@ -174,6 +235,7 @@ func (s *Service) SetAddPeersHandler(h func(addr ...swarm.Address)) {
 
 func (s *Service) Close() error {
 	close(s.quit)
+	s.bgCancel()
```
aloknerurkar, line 240: So idea was to not have both. Right now it seems like we are duplicating the functionality. In some cases we are waiting on s.quit and in some cases we are waiting on bgCtx.Done. This has become more confusing than it was.

If we have bgCtx we can remove s.quit. In broadcastNow function when we are using the caller context we need to merge the cancel functions of both bgCtx and the calling context. That way we obey both cancels. This way we can remove s.quit. But this is not so straightforward, which is why I suggested doing it in a separate PR.

For eg https://pkg.go.dev/context#example-AfterFunc-Merge

pkg/hive/hive.go
sbackend123: Yes, we decided not to cover all corner cases now 
https://github.com/ethersphere/bee/pull/5520#discussion_r3824411816
I can just document it as Elad suggested

aloknerurkar (approved):

gacevicljubisa (approved):

PR #5522, opened 2026-06-30
fix: add bounds checking for data offset in joiner readAtOffset

nugaon:
### Checklist

### Description

Adds bounds checking to the leaf branch of `(*joiner).readAtOffset` in `pkg/file/joiner/joiner.go`. A malformed parent intermediate-chunk span can drive the recursion into the leaf path with `dataOffsetStart` beyond `len(data)`, producing an inverted slice range and a runtime panic. The offsets are now validated and rejected as `ErrMalformedTrie` instead of crashing the node, consistent with the existing malformed-trie handling in the same function.

### Motivation and Context (Optional)

Reported panic, most likely caused by an externally malformed (erasure-coded) chunk whose parent span does not match the actual data:

```
panic: runtime error: slice bounds out of range [16384:4096]

goroutine 6366125 [running]:
github.com/ethersphere/bee/v2/pkg/file/joiner.(*joiner).readAtOffset(...)
    joiner.go:243
```

With a corrupted parent span, `dataOffsetStart = 16384` exceeds `len(data) = 4096`, making `lenDataToCopy` negative and yielding `dataOffsetEnd = 4096`, so `data[16384:4096]` panics. Parity shard spans are GF(2⁸) combinations of data-chunk spans and are not meaningful values, so a corrupted span must be treated as malformed rather than trusted.

### AI Disclosure

### DIFF
--- a/pkg/file/joiner/joiner.go
+++ b/pkg/file/joiner/joiner.go
@@ -240,6 +240,13 @@ func (j *joiner) readAtOffset(
 			dataOffsetEnd = dataOffsetStart + lenDataToCopy
 		}
 
+		if dataOffsetStart < 0 || dataOffsetEnd > int64(len(data)) || dataOffsetStart > dataOffsetEnd {
+			eg.Go(func() error {
+				return ErrMalformedTrie
+			})
+			return
+		}
+
 		bs := data[dataOffsetStart:dataOffsetEnd]
 		n := copy(b[bufferOffset:bufferOffset+int64(len(bs))], bs)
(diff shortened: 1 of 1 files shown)


acud (approved):

janos (commented):
This is a nice bounds check fix, but a regression test would be good to have. Maybe with some fixed chunks data that have wrong offesets.

acud:
@nugaon thanks, we will reimplement with a regression test as we've managed to reproduce this.

PR #5526, opened 2026-07-07
feat(tracing): add telemetry span kind

gacevicljubisa:
### Checklist

### Description

P2P spans are now tagged with client/server span kinds so trace UIs render request/response relationships correctly, and all span attributes are namespaced under swarm.* (e.g. address → swarm.chunk.address, tag_id → swarm.tag.id, success → swarm.operation.success) to align with OpenTelemetry semantic-convention naming. The go.mod indirect dependencies are also consolidated into a single block with no version changes. These are trace-annotation changes only, no behavioral impact.

### AI Disclosure


pkg/pushsync/pushsync.go
```diff
@@ -558,7 +558,7 @@ func (ps *PushSync) push(parentCtx context.Context, resultChan chan<- receiptRes
 
 	defer action.Cleanup()
 
-	spanInner.SetAttributes(attribute.String("peer_address", peer.String()))
+	spanInner.SetAttributes(attribute.String("swarm.peer.address", peer.String()))
```
martinconic, line 561: The server handler tags the remote as `swarm.peer.sender_address` while here the client tags the remote peer as the generic `swarm.peer.address`. Should we have here something like `swarm.peer.target_address` ?

martinconic (approved):

pkg/pushsync/pushsync.go
gacevicljubisa: Good catch on asymmetry. And after some digging, my opinion is that it would be better to drop the role, and use added span kind we just introduced. That way on the client push span, the peer is the destination, and on the server handler span, the peer is the source. This way, spans are role agnostic. What do you think?

akrem-chabchoub (approved):

sbackend123 (approved):

PR #5528, opened 2026-07-07
feat: add safe package for unified goroutine recovery

janos:
### Checklist

### Description
This PR introduces a unified panic recovery approach to protect the Bee node from unexpected crashes in goroutines.

By catching panics inside worker routines, checker loops, and handler callbacks, it is ensures that unexpected panics are safely recovered and converted into either structured log messages or standard Go errors, rather than causing the entire node process to crash.

I have added a new small package in ./pkg/safe for functions that unify the recovery for the bee node.

Other changes are the usage of these functions in various goroutines in the code. Waith groups and errors groups create goroutines internally.

Prevent bee node from crashing on panics.

### AI Disclosure


acud:
if goroutines that are supposed to panic are recovered gracefully and do not cause the application to crash, how do we recover from a state where: 
a. application is an inconsistent state?
b. database is in an inconsistent state? i.e. if a db write failed because the disk is full and we can no longer assume that the node is consistent
c. node setup is done in a wrong way by us (devs) and now the "feedback" (panic) is swallowed into logs

janos:
> if goroutines that are supposed to panic are recovered gracefully and do not cause the application to crash, how do we recover from a state where: a. application is an inconsistent state? b. database is in an inconsistent state? i.e. if a db write failed because the disk is full and we can no longer assume that the node is consistent c. node setup is done in a wrong way by us (devs) and now the "feedback" (panic) is swallowed into logs

That is per goroutine for us to decide. Some goroutines are contained for a specific operation that should not terminate the application. For example, the joiner slice out of bounds panic was related only to one resource to be loaded and had no impact on the operations of the whole application and that panic could have been be recovered. And for the goroutines related to the database should be evaluated if they should be recovered. That is why I have added protection in goroutines that I could find so that they can be evaluated, case per case, as are visible in this PR changeset.

pkg/storer/validate.go
```diff
@@ -151,10 +152,12 @@ func validateWork(logger log.Logger, store storage.Store, readFn func(context.Co
...
-			buf := make([]byte, swarm.SocMaxChunkSize)
-			for item := range iteratateItemsC {
-				validChunk(item, buf[:item.Location.Length])
-			}
+			safe.Run(logger, "reserve-validation-worker", func() {
+				buf := make([]byte, swarm.SocMaxChunkSize)
+				for item := range iteratateItemsC {
+					validChunk(item, buf[:item.Location.Length])
+				}
+			})
```
gacevicljubisa: Would be better if we recover per item? That way one corrupt item will not tear down the worker?

pkg/storer/validate.go
```diff
@@ -325,15 +328,17 @@ func (p *PinIntegrity) Check(ctx context.Context, logger log.Logger, pin string,
...
-						break
+				safe.Run(logger, "pin-integrity-worker", func() {
+					buf := make([]byte, swarm.SocMaxChunkSize)
+					for item := range iteratateItemsC {
+						if ctx.Err() != nil {
+							break
+						}
+						if !validChunk(item, buf[:item.Location.Length]) {
+							invalid.Add(1)
+						}
```
gacevicljubisa: Maybe here as well we could recover per item?

pkg/safe/safe.go
```diff
@@ -0,0 +1,57 @@
...
+		fn()
+	}()
+}
+
+// Run runs the given function synchronously and recovers from panic in it.
+// Panics are logged using the provided logger.
+func Run(logger log.Logger, name string, fn func()) {
+	defer func() {
+		if r := recover(); r != nil {
+			logger.Error(nil, "panic recovered", "name", name, "panic", fmt.Sprintf("%v", r), "stack", string(debug.Stack()))
```
gacevicljubisa: it is a minor thing, but maybe we could check if logger is nil?

pkg/salud/salud.go
```diff
@@ -144,31 +145,33 @@ func (s *service) salud(mode string, durPercentile float64, connsPercentile floa
...
+				if err != nil {
+					s.topology.UpdatePeerHealth(addr, false, dur)
+					return
+				}
+
+				if snapshot.BeeMode != mode {
+					return
+				}
+
+				mtx.Lock()
```
gacevicljubisa, line 165: maybe to defer mtx.Unlock() here if panic happens in IsWithinStorageRadius

gacevicljubisa (commented):
Overall, the PR is well scoped and with clear idea. I would suggest adding some metrics that could track total recovered panics.

pkg/safe/safe.go
acud: and then you completely lose sight of the fact that there was a panic...

janos: > and then you completely lose sight of the fact that there was a panic...

actually, it will panic if the logger is nil, as the recover is already called, but the check for nil logger is a good suggestion

acud: yes so what i meant it... if it panicked and there is no logger, at least you get a panic now (so you know there was even a panic in the first place).
if you check if the logger is nil and don't even call it - then you will: a. never know that a panic was recovered (unless you add metrics), and b. never know that the logger was not set (which is a severe engineering human error, because someone wrote code that did not set the logger where it should have been set). so my claim is that when we do all of these defensive programming we actually lose feedback over the code.
if the choice is between: 1. having a panic that propagates because of a real error, or 2. having a panic that does not show up at all (and be silently recovered, potentially putting the process into an inconsistent state because someone forgot to set a logger up properly then i prefer to have a panic crash the process honestly but that's just me. 
or to conclude: i like panics :)

janos: Actually, you will get the panic that the logger is nil as it was a method call on the nil logger. That does mean that the original panic is lost, but because of that I have added the check on the nil logger as Ljubisa suggested.

We do not need to recover from panics. This PR is just mu suggestion, it depends on your preference. My opinion is that recovering from panics depends on the actual logic in the goroutine. Recent panics in the code show that there are localized problems that could be errors, not crashing the node (joiner bounds and nil bzz address).

The defensive programming is mostly related to the user experience. Where an error can be logged instead of crashing the node, it appears to me like a better user experience.

pkg/postage/listener/listener.go
martinconic, line 366: I the above `listenf()` panics, then safe.Go recovers and the bellow lines are they skipped?

janos: That is a good catch. I have pushed the change to protect the listenf function so that it returns the error even on panic which will be propagated to the channel.

pkg/safe/safe.go
```diff
@@ -0,0 +1,63 @@
...
+// If the recovered value implements error, it is wrapped using %w to preserve error type checking.
+func RunFunc(logger log.Logger, name string, fn func() error) func() error {
+	return func() (err error) {
+		defer func() {
+			if r := recover(); r != nil {
+				if logger != nil {
+					logger.Error(nil, "errgroup goroutine panic recovered", "name", name, "panic", fmt.Sprintf("%v", r), "stack", string(debug.Stack()))
+				}
+				if panicErr, ok := r.(error); ok {
+					err = fmt.Errorf("panic in %s: %w", name, panicErr)
```
martinconic: Should we use %v for panic values? Callers should never be able to mistake a panic for a sentinel error. Same bellow?

pkg/storer/validate.go
```diff
@@ -330,9 +333,11 @@ func (p *PinIntegrity) Check(ctx context.Context, logger log.Logger, pin string,
 					if ctx.Err() != nil {
 						break
 					}
-					if !validChunk(item, buf[:item.Location.Length]) {
-						invalid.Add(1)
-					}
+					safe.Run(logger, "pin-integrity-worker", func() {
+						if !validChunk(item, buf[:item.Location.Length]) {
```
martinconic: buf[:item.Location.Length] is evaluated as the argument, so it panics before validChunk ever runs. Location.Length is a uint16 and buf is SocMaxChunkSize, so a corrupt retrieval index reaches this. The panic is recovered, invalid is never incremented, and the chunk was already counted in total by the iterator.

pkg/safe/safe.go
janos: Makes sense, I have removed the error type check and adjusted the test to ensure this behaviour.

pkg/storer/validate.go
janos: I have pulled out the invalid counter so that it is executed even in case of a panic.

sbackend123:
I might be wrong, but it seems like panic recovery causes silent death of workers such as the blocker sequencer/blacklist worker, puller manage, pusher chunksWorker and so on.

The node can live more or less normally without some workers (e.g. the blocker that blacklists misbehaving peers — not great, but not fatal), but there are more critical ones like puller manage, pusher chunksWorker, SubscribeBin and SubscribePush. It looks to me like in case of a panic they would just silently stop, and the user wouldn't know about it. And to resume their work, they'd have to restart the node.

Do I miss smth?

janos:
> I might be wrong, but it seems like panic recovery causes silent death of workers such as the blocker sequencer/blacklist worker, puller manage, pusher chunksWorker and so on.
> 
> The node can live more or less normally without some workers (e.g. the blocker that blacklists misbehaving peers — not great, but not fatal), but there are more critical ones like puller manage, pusher chunksWorker, SubscribeBin and SubscribePush. It looks to me like in case of a panic they would just silently stop, and the user wouldn't know about it. And to resume their work, they'd have to restart the node.
> 
> Do I miss smth?

You noticed that correctly. I have removed all recoveries from the long running worker goroutines and left only for the short lived ones.

This is the simples thing to do for now. There can be more robust worker recovery mechanisms, with retries on panics, but that may be a scope for hardening "goroutine daemon" workers.

martinconic (approved):

acud (approved):
LGTM just needs a rebase. thanks @janos

acud (approved):
hayde

PR #5531, opened 2026-07-08
fix(pseudosettle): bound the first refreshment of a connection

### LINKED ISSUE #5530: pseudosettle: first-contact allowance anchored to Unix epoch instead of connection time
### Context

Bee `master`, package `pkg/settlement/pseudosettle`. This is a code-level accounting correctness issue found by reading the settlement logic; it is not environment specific.

### Summary

The time-based allowance granted for an incoming refreshment is computed as `(now - lastSettlementTime) * refreshRate`. On first contact with a peer there is no stored settlement record, so `lastSettlementTime` defaults to `0` (the Unix epoch), making the allowance `now * refreshRate`, i.e. the entire Unix epoch scaled by the refresh rate.

### Expected behavior

The time-based allowance should reflect the time elapsed since the peer connected, so the first refreshment forgives only the debt accrued over the elapsed connected time.

### Actual behavior

The first refreshment from a peer can forgive an effectively unbounded debt (bounded only by the peer's current debt), independent of how long the peer has actually been connected, because the allowance window starts at the Unix epoch rather than the connection time.

### Steps to reproduce

1. Connect a peer and let it accrue debt.
2. Have the peer send its first pseudosettle refreshment.
3. Observe that `peerAllowance` returns `min(debt, now*refreshRate)` and the full debt is accepted regardless of connection duration, because `lastTime.Timestamp` defaults to `0` on `storage.ErrNotFound` in `peerAllowance` (`pkg/settlement/pseudosettle/pseudosettle.go`).

### Possible solution

Record the connection time per peer in `init` and anchor the first-contact allowance to it rather than to the epoch, so the initial refreshment only forgives debt accrued over the elapsed connected time. Subsequent settlements are unchanged.

### AI Disclosure
- [x] This issue contains suggestions and text generated by an LLM.
- [x] I have reviewed the AI generated content thoroughly.
- [x] I possess the technical expertise to responsibly review the AI generated content mentioned in this issue.

mfw78:
### Checklist

### Description

`peerAllowance` computes the acceptable incoming payment as `(now - lastSettlementTime) * refreshRate`. With no stored settlement record `lastSettlementTime` defaults to `0`, so the first refreshment from a peer is granted the whole Unix epoch scaled by the refresh rate and forgives its entire outstanding debt in one go. The same holds on reconnection: the stored record survives the disconnect, so a peer returning after an hour is granted an hour of allowance for a connection that is seconds old.

The bound cannot be applied to the recipient alone. `NotifyRefreshmentSent` recomputes the expectation as `min(allegedInterval * refreshRate, attemptedAmount)` and blocklists the peer when the accepted amount falls short, and `allegedInterval` is derived from the payer's own stored record, which defaults to `0` in exactly the same way. Both ends defaulting to the epoch is what makes them agree today, so tightening one side alone would make a patched recipient accept less than every peer expects and be blocklisted on the first refreshment.

This bounds the first refreshment of a connection to a single refresh interval, applied identically on the recipient and the payer. Refreshments after the first use the stored timestamps as before, and reconnection is treated as a first refreshment so allowance no longer accrues while a peer is disconnected.

Anchoring at the connection time was considered and dropped. The quantity being bounded is of the same order as the clock skew the protocol already tolerates, 2 seconds in `ErrTimeOutOfSyncRecent` against a default payment threshold of 3 seconds of refresh rate, so a bound derived from two independently measured clocks would be unreliable, and slack added to absorb the skew would be claimable on every refreshment. Treating an absent or pre-connection timestamp as one interval needs no clock comparison between peers and is identical on both ends by construction.

On rollout: an unpatched payer still expects the full attempted amount on first contact, so it will blocklist a patched recipient for roughly `(debt + paymentThreshold) / refreshRate` seconds on each new peer until the network has upgraded. The payer half is a pure relaxation and can never cause a blocklist, so it can be released on its own first with the recipient bound following later if you would rather not carry that. Happy to split it that way.

One correction to #5530: the allowance is capped by the peer's debt, which the recipient already caps at `disconnectLimit`, so the exposure is one payment threshold plus tolerance per new peer identity rather than an unbounded amount. Still worth closing off given identities are cheap, but the issue overstates the magnitude.

None.

See #5530.

### Related Issue (Optional)

Closes #5530.

N/A

### AI Disclosure

### DIFF
--- a/pkg/settlement/pseudosettle/pseudosettle.go
+++ b/pkg/settlement/pseudosettle/pseudosettle.go
@@ -60,8 +60,9 @@ type Service struct {
 }
 
 type pseudoSettlePeer struct {
-	lock     sync.Mutex // lock to be held during receiving a payment from this peer
-	fullNode bool
+	lock        sync.Mutex // lock to be held during receiving a payment from this peer
+	fullNode    bool
+	connectedAt int64 // unix time the peer connected; anchors the first-contact allowance
 }
 
 type lastPayment struct {
@@ -108,7 +109,7 @@ func (s *Service) init(ctx context.Context, p p2p.Peer) error {
 
 	_, ok := s.peers[p.Address.String()]
 	if !ok {
-		peerData := &pseudoSettlePeer{fullNode: p.FullNode}
+		peerData := &pseudoSettlePeer{fullNode: p.FullNode, connectedAt: s.timeNow().Unix()}
 		s.peers[p.Address.String()] = peerData
 	}
 
@@ -142,14 +143,18 @@ func totalKeyPeer(key []byte, prefix string) (peer swarm.Address, err error) {
 
 // peerAllowance computes the maximum incoming payment value we accept
 // this is the time based allowance or the peers actual debt, whichever is less
-func (s *Service) peerAllowance(peer swarm.Address, fullNode bool) (limit *big.Int, stamp int64, err error) {
+func (s *Service) peerAllowance(peer swarm.Address, fullNode bool, connectedAt int64) (limit *big.Int, stamp int64, err error) {
 	var lastTime lastPayment
 	err = s.store.Get(totalKey(peer, SettlementReceivedPrefix), &lastTime)
 	if err != nil {
 		if !errors.Is(err, storage.ErrNotFound) {
 			return nil, 0, err
 		}
-		lastTime.Timestamp = int64(0)
+		// First contact with this peer: anchor the time-based allowance to the
+		// connection time rather than the zero epoch. Anchoring at zero would
+		// scale the entire Unix epoch by the refresh rate and let the initial
+		// refreshment forgive an effectively unbounded debt.
+		lastTime.Timestamp = connectedAt
 	}
 
 	currentTime := s.timeNow().Unix()
@@ -210,7 +215,7 @@ func (s *Service) handler(ctx context.Context, p p2p.Peer, stream p2p.Stream) (e
 	pseudoSettlePeer.lock.Lock()
 	defer pseudoSettlePeer.lock.Unlock()
 
-	allowance, timestamp, err := s.peerAllowance(p.Address, pseudoSettlePeer.fullNode)
+	allowance, timestamp, err := s.peerAllowance(p.Address, pseudoSettlePeer.fullNode, pseudoSettlePeer.connectedAt)
 	if err != nil {
 		return err
 	}
--- a/pkg/settlement/pseudosettle/pseudosettle.go
+++ b/pkg/settlement/pseudosettle/pseudosettle.go
@@ -60,9 +60,10 @@ type Service struct {
 }
 
 type pseudoSettlePeer struct {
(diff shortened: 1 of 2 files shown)


gacevicljubisa:
What about the payer side, should it match the peerAllowance?

gacevicljubisa:
> What about the payer side, should it match the peerAllowance?

@mfw78

mfw78:
Good catch, it does have to match, and leaving it unmatched would have been worse than the original bug.

`NotifyRefreshmentSent` recomputes the expectation as `min(allegedInterval * refreshRate, attemptedAmount)` and blocklists the peer when the accepted amount falls short. `allegedInterval` is `paymentAck.Timestamp - lastTime.Timestamp`, and on first contact the payer's stored record is absent so its timestamp defaults to `0` in exactly the same way the recipient's does. Both ends defaulting to the epoch is precisely what makes them agree today. Bounding the recipient alone would have made a patched node accept less than every peer expects and get blocklisted on the first refreshment with each new peer.

Pushed a change that applies the bound on both sides. I dropped the connection-time anchor while doing so: the amount being bounded is of the same order as the clock skew the protocol already tolerates, 2 seconds in `ErrTimeOutOfSyncRecent` against a default payment threshold of 3 seconds of refresh rate, so a bound derived from two independently measured clocks would be unreliable, and any slack added to absorb that skew would be claimable on every refreshment rather than once. Both sides now treat an absent or pre-connection timestamp as exactly one refresh interval, which needs no clock comparison between peers and is identical on both ends by construction.

That also covers reconnection, which the first version missed. The stored record survives a disconnect, so a peer returning after an hour was granted an hour of allowance against a connection seconds old. It now gets a single interval, same as first contact.

On the test side, `testCaseAccepted` asserts the payer's alleged interval against the interval the recipient granted on, since that equality is the invariant keeping the two ends from diverging. Reverting either half of the change fails it: without the recipient bound the full 1000 intervals of debt are accepted, without the payer change the payer alleges an interval of 1000000 against the recipient's 1.

Two things worth your call:

Rollout. An unpatched payer still expects the full attempted amount on first contact, so it will blocklist a patched recipient for roughly `(debt + paymentThreshold) / refreshRate` seconds on each new peer until the network has upgraded. Short and self-clearing, but it recurs per peer. The payer half is a pure relaxation and can never cause a blocklist, so it could ship on its own first with the recipient bound following in a later release. Say the word and I will split it.

Magnitude. #5530 overstates this and I would rather correct it than have it read as more than it is. The allowance is capped by the peer's debt, which the recipient already caps at `disconnectLimit`, so the exposure is one payment threshold plus tolerance per new peer identity, not an unbounded amount. Worth closing off given identities are cheap, but it is not the unbounded case the issue describes.

mfw78:
GitHub closed this on its own and refuses to reopen it: the head repository it was raised from is private and detached from the fork network, so the pull request is not recoverable.

Continued in #5542, raised from a public fork with the same branch and commits. @gacevicljubisa your question is answered there.

PR #5544, opened 2026-07-24
feat(pullsync): content-aware chunk checksum for divergent SOC sync (SWIP-101)

martinconic:
### Problem

Two valid single owner chunks can share the same address, batch and stamp while wrapping different content. The pullsync want-check (`ReserveHas(addr, batchID, stampHash)`) is blind to content, so a node holding one variant never requests the other — neighborhoods holding divergent SOCs can never converge, and their reserve samples produce different commitments in the redistribution game.

### Change ([SWIP-101](https://github.com/ethersphere/SWIPs/pull/101))

Offers now carry a 16-byte checksum per chunk instead of batch ID + stamp hash:

- CAC: `keccak256(batchID ‖ stampHash)[:16]`
- SOC: `keccak256(batchID ‖ stampHash ‖ wrappedCacAddress)[:16]`

The receiver wants a chunk iff `!Has(address, sum)`, answered in O(1) by a new existence index (`ChunkSumItem`). Sums are precomputed at write time and stored on `ChunkBinItem`; on delivery the sum is recomputed and mismatches are rejected as unsolicited. Replacing a SOC payload refreshes the sums of co-resident entries under other stamps, keeping the index consistent with deliverable content.

### Breaking change

Pullsync protocol version `1.4.0` → `2.0.0`. Old and new nodes cannot pullsync with each other; needs a coordinated release.

### Migration

`step_08` backfills sums for existing reserve entries (paginated, no stamp loads) and removes dangling or legacy-format leftovers instead of failing startup. `db repair-reserve` rebuilds the sum index alongside the chunk bin items.

### Scope note

This makes divergence *visible and syncable*. Deterministic retention (which chunk a neighborhood converges on) is a follow-up on `feat/pullsync-soc-convergence`. Divergent SOCs under *different* batches occupy different stamp indices and are not reconciled here.

### Testing

Unit tests cover the divergent-SOC sum property, sum-index lockstep with the reserve, sibling-sum refresh on both replacement paths, and the migration against genuine pre-migration records (old 106-byte serialization, dangling entries, orphans, zero stamp hash).


pkg/pullsync/pullsync.go
```diff
@@ -36,7 +36,7 @@ const loggerName = "pullsync"
 
 const (
 	protocolName     = "pullsync"
-	protocolVersion  = "1.4.0"
+	protocolVersion  = "2.0.0"
```
acud, line 39: with the change from @janos we can just backward support the old version through the different semver (but we'd need to duplicate the protocol code for this). let's discuss this.

pkg/storer/internal/reserve/reserve.go
```diff
@@ -449,6 +576,76 @@ func (r *Reserve) removeChunk(
...
+}
+
+// RemoveMalformedChunkBinItems deletes chunkBin entries whose stored value
+// does not match the current serialization, without unmarshaling them. Such
+// entries are orphaned pre-Sum records (a ChunkBinItem without a matching
+// BatchRadiusItem is never rewritten by the Sum backfill migration) and would
+// otherwise fail every full iteration of the namespace: the reserve sampler,
+// pullsync bin subscriptions and the reserve repairer. They have no companion
+// ChunkSumItem to remove. Returns the number of entries deleted.
+func RemoveMalformedChunkBinItems(ctx context.Context, st transaction.Storage) (int, error) {
```
acud, line 965: what is this about? :) i think this would be the first appearance of a self-authored storage wipe mechanism. why do we need such a thing? and why does a migration not cover what we need it to cover?

pkg/storer/internal/reserve/reserve.go
```diff
@@ -290,11 +321,101 @@ func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
...
+// longer holds (peers reject the deliveries as unsolicited) and keep matching
+// offers for content it cannot store.
+//
+// It must be called with no reserve locks held: each sibling entry is updated
+// under its own batch lock, one at a time, mirroring the Put-vs-Eviction lock
+// discipline. The sum is recomputed from the currently committed payload
+// rather than the caller's chunk, so concurrent replacements converge on the
+// content that committed last. Entries whose sum already matches, including
+// the caller's own, are left untouched.
+func (r *Reserve) refreshSiblingSums(ctx context.Context, addr swarm.Address) error {
```
acud, line 500: iff when a chunk changes on the disk we need to bump the binIDs for the indices relating to that chunk

pkg/storer/internal/reserve/reserve.go
```diff
@@ -100,26 +101,50 @@ func New(
 //  3. A new chunk that has the same address belonging to the same stamp index with an already stored chunk will overwrite the existing chunk
 //     if the new chunk has a higher stamp timestamp (regardless of batch type and chunk type, eg CAC & SOC).
 func (r *Reserve) Put(ctx context.Context, chunk swarm.Chunk) error {
+	socReplaced, err := r.putChunk(ctx, chunk)
+	if err != nil {
+		return err
+	}
+	if socReplaced {
```
acud, line 115: not 100% about this either. @sbackend123 last time i heard you were saying you're at the point of have a `putSoc` and `putCac` which i think is generally a good direction (although the review diff might be quite large).

sbackend123: @acud it is in separate PR here https://github.com/ethersphere/bee/pull/5557

pkg/storer/internal/reserve/reserve.go
```diff
@@ -92,217 +94,712 @@ func New(
...
+				"stamp index collision chunk %s lost tie-break: %w",
+				chunk.Address(),
+				storage.ErrDivergentChunkRejected,
+			)
+		}
+	}
+
+	if oldStampIndex.ChunkAddress.Equal(chunk.Address()) {
+		if prev == curr {
+			if bytes.Equal(oldStampIndex.StampHash, stampHash) {
```
acud, line 380: this means this is the exact same chunk, so there's actually nothing to do. but it still might be useful to explicitly handle the case just for the sake of brevity

pkg/storer/internal/reserve/reserve.go
```diff
@@ -92,217 +94,712 @@ func New(
...
+					return false, err
+				}
+				if !wins {
+					return false, fmt.Errorf(
+						"diverging chunk %s lost tie-break: %w",
+						chunk.Address(),
+						storage.ErrDivergentChunkRejected,
+					)
+				}
+			} else if bytes.Compare(oldStampIndex.StampHash, stampHash) < 0 {
```
acud: remove this branch, test case and document the parent `Put` method

pkg/storer/internal/reserve/reserve.go
```diff
@@ -92,217 +94,712 @@ func New(
...
+		if err != nil {
+			return err
+		}
+		shouldInc = !loaded
+		return nil
+	})
+	return
+}
+
+func (r *Reserve) evaluateSOCDivergence(
```
acud, line 269: please document in a nutshell what this method does

pkg/storer/internal/reserve/reserve.go
```diff
@@ -92,217 +94,712 @@ func New(
...
+			}
+		}
 
-				oldBatchRadiusItem := &BatchRadiusItem{
-					Bin:       bin,
-					Address:   oldStampIndex.ChunkAddress,
-					BatchID:   oldStampIndex.BatchID,
-					StampHash: oldStampIndex.StampHash,
+		entrySum := sum
+		if has && !incomingWins {
```
acud, line 239: if this is true, why do we need to do anything instead of returning early here?

pkg/storer/internal/reserve/reserve.go
martinconic: We saw in the meeting that this actually is not the same chunk

pkg/storer/internal/reserve/reserve.go
martinconic: We can’t return early because the incoming chunk brings a new, valid stamp (different batch or index) for this SOC address. Returning early would break arrival-order convergence:

If V1 arrives before V2: both stamps end up in the reserve, with V1 updated to V2's payload via refreshSiblingSums.
If V2 arrives before V1: returning early would drop V1's stamp completely, causing the node to diverge from nodes that received them in reverse order.
So we must still record the stamp in the reserve, but index it with storedSum (the winning payload on disk) so pullsync offers match what we actually serve.

martinconic:
Findings fixed in the last 8 commits
[bee-5544-fixes.md](https://github.com/user-attachments/files/31960061/bee-5544-fixes.md)
[bee-5544-open-findings.md](https://github.com/user-attachments/files/31960071/bee-5544-open-findings.md)

acud:
> bee-5544-fixes.md

`B1` how is this possible? isn't this the most basic case we are catering for here? reading the first sentence `A SOC address already holds a payload under stamp A. A second stamp B arrives with a payload that loses the tie-break (for example an older timestamp).` the only logical continuation is "nothing should happen", no?

`B2` seems hypothetical and i'm not sure whether we should address it

`B3` seems to change significantly the migration performance - we should test whether this makes sense and whether this has a significant impact on the migration time. there are other ways to achieve this same fix i suppose.

`B4` seems hypothetical. is this even possible?

martinconic:
> > bee-5544-fixes.md
> 
> `B1` how is this possible? isn't this the most basic case we are catering for here? reading the first sentence `A SOC address already holds a payload under stamp A. A second stamp B arrives with a payload that loses the tie-break (for example an older timestamp).` the only logical continuation is "nothing should happen", no?
> 
> `B2` seems hypothetical and i'm not sure whether we should address it
> 
> `B3` seems to change significantly the migration performance - we should test whether this makes sense and whether this has a significant impact on the migration time. there are other ways to achieve this same fix i suppose.
> 
> `B4` seems hypothetical. is this even possible?

Hi,  here is the context behind each of those points:

### `B1` (Why "nothing should happen" breaks convergence)
In Swarm, postage stamps sign the **chunk address**, not the payload content. A SOC address can be funded simultaneously by multiple batches.
If a new stamp arrives with a payload that loses the tie-break:
1. **Arrival-order divergence:** 
   - Node 1 receives Stamp A (older) then Stamp B (newer) → retains the newer payload and holds **2 stamps** (A and B).
   - Node 2 receives Stamp B (newer) then Stamp A (older) → if it does "nothing", it only holds **1 stamp** (B only).
   Their reserve sizes, radiuses, and pullsync offers diverge based purely on network arrival order.
2. **Premature eviction:** If Node 2 drops Stamp A, then when Batch B expires, Node 2 deletes the chunk from disk even though Batch A is active and fully paid for.
3. **The `B1` fix:** We keep the winning payload on disk and record Stamp B's entry with `entrySum = ChunkSum(Stamp B, winningPayload)`. Because Stamp B now references the stored payload, we must call `ChunkStore().Put(ctx, stored)` to increment the refcount; otherwise, evicting Batch A would delete the payload from under Stamp B.
### `B2` (Self-healing missing sum index rows)
Under clean operation this is rare, but if a node crashes mid-migration or during `db repair-reserve`, an entry can exist without its `ChunkSumItem`. Previously, `putChunk` treated this as a divergence and rejected identical SOCs with `ErrDivergentChunkRejected` or CACs with "not a single owner chunk", causing pullsync to spin in an infinite re-want loop. `restoreEntrySum` is a lightweight self-healing fallback that repairs the index row when the payload matches.
### `B3` (Migration performance)
Migration performance is **not** impacted for standard nodes:
- The 99.999% happy path (valid chunks being backfilled with sums) **remains 100% batched in 1,000-item transactions** (`st.Run`).
- Single-item transactions are **only** executed when an entry is marked for deletion (`if remove`), which is rare (orphans/corrupted chunks).
- Removals require single-item transactions because batched transactions cannot see pending refcount decrements on shared SOC addresses, which would leak Sharky slots (identical to how `EvictBatchBin` isolates deletions).
### `B4` (Zero-hash stamp cleanup)
The check `bytes.Equal(item.StampHash, swarm.EmptyAddress.Bytes())` was already added to `step_08.go` in July to prune legacy records with unpopulated stamp hashes. However, the existing code called `RemoveChunkWithItem`, which looked up stamps by hash; with an all-zero hash, it returned `nil` and silently skipped deleting the `stampindex` row. This left poisoned slots that caused subsequent writes to fail with `not found`. `B4` simply ensures the cleanup that `step_08` was already doing completes cleanly.

pkg/pullsync/pullsync.go
gacevicljubisa: We could even keep the protocol version `1.4.0`, by adapting the proto message for Chunk:

```proto
message Chunk {
  bytes Address = 1;
  bytes BatchID = 2;
  bytes StampHash = 3;
  bytes Sum = 4;
}
```
where  len(Sum) == 0 can identify the legacy peer, and select the old batchID/stampHash check.

pkg/pullsync/pullsync.go
```diff
@@ -270,14 +273,15 @@ func (s *Syncer) Sync(ctx context.Context, peer swarm.Address, bin uint8, start
 
 	for i := 0; i < len(offer.Chunks); i++ {
 		if offer.Chunks[i] == nil {
-			return 0, 0, fmt.Errorf("nil chunk at index %d in offer from peer %s", i, peer)
+			s.logger.Debug("syncer got nil chunk on offer", "peer_address", peer, "index", i)
+			continue
```
gacevicljubisa, line 277: if no error is returned, maybe metrics that count skipped could be beneficial to check offer. Also, at the end of the method we could still write the error and return top most?

pkg/storer/internal/reserve/reserve.go
```diff
@@ -92,217 +94,806 @@ func New(
...
+//     type-specific chunkstore actions (CAC stores reserve entries; SOC evaluates
+//     divergence against other live stamps).
+//  2. sameAddr=false: the stored entry points at a different address. That
+//     chunk and its reserve metadata are removed and the stamp index is
+//     rewritten. The caller must still call storeReserveEntries and write the
+//     new chunk to the chunkstore.
+func (r *Reserve) resolveStampIndexCollision(
+	ctx context.Context, s transaction.Store,
+	chunk swarm.Chunk, oldStampIndex *stampindex.Item,
+	sum, stampHash []byte, bin uint8,
```
gacevicljubisa, line 416: unused parameter: sum

pkg/storage/storage.go
```diff
@@ -295,6 +302,80 @@ func ChunkType(ch swarm.Chunk) swarm.ChunkType {
...
+func ChunkSumFromParts(batchID, stampHash []byte, ch swarm.Chunk) ([]byte, error) {
+	h := swarm.NewHasher()
+	_, _ = h.Write(batchID)
+	_, _ = h.Write(stampHash)
+
+	if soc.Valid(ch) {
+		s, err := soc.FromChunk(ch)
+		if err != nil {
+			return nil, fmt.Errorf("soc from chunk: %w", err)
+		}
```
gacevicljubisa, line 339: This is already called in the `soc.Valid(ch)`

pkg/storer/internal/reserve/reserve.go
```diff
@@ -92,217 +94,806 @@ func New(
...
+					return false, err
+				}
+				if !wins {
+					return false, fmt.Errorf(
+						"diverging chunk %s lost tie-break: %w",
+						chunk.Address(),
+						storage.ErrDivergentChunkRejected,
+					)
+				}
+			}
```
gacevicljubisa, line 455: Instead of last write wins, would it be better that we have smaller hash wins, which is deterministic for all, something like this:

```go
			} else if bytes.Compare(oldStampIndex.StampHash, stampHash) < 0 {
				return false, fmt.Errorf("stamp collision chunk %s lost stamp-hash tie-break: %w", chunk.Address(), storage.ErrDivergentChunkRejected)
			}
```

PR #5548, opened 2026-07-25
feat(p2p): add versioned handlers function

janos:
### Checklist

### Description
This PR adds VersionedHandler and NewVersionedHandlersFunc to pkg/p2p to streamline protocol backward compatibility across node versions. The suggestion for this feature came from @acud.

When updating protocols while maintaining support for older node releases, checking stream versions manually inside handler bodies can get verbose. NewVersionedHandlersFunc provides a clean, declarative way to associate handlers with minimum semver thresholds.

When using NewVersionedHandlersFunc function it is needed to declare legacy handlers with the versions of the protocol from which that handler is supported and new handlers to define from which version of the protocol it is supported, so the function will select the highest supported handler by the peer's stream. The implementation relies on the recently added Stream.Version() method to get the version of the stream. This function can be used for both handler and "client" functions, as the logic is the same. There is an example added for more clarity on the usage.

### AI Disclosure


martinconic (approved):

pkg/p2p/example_versioned_test.go
```diff
@@ -0,0 +1,148 @@
+// Copyright 2026 The Swarm Authors. All rights reserved.
```
martinconic: Should we change the name of this file to something like `example_versioned_test.go` ? It would be just straight forward to figure what are the tests for.

martinconic (approved):

akrem-chabchoub (approved):

pkg/p2p/p2p.go
```diff
@@ -208,6 +209,40 @@ type HandlerFunc func(context.Context, Peer, Stream) error
...
+
+	return func(ctx context.Context, p Peer, stream Stream) error {
+		v, err := stream.Version()
+		if err != nil {
+			return fmt.Errorf("get stream version: %w", err)
+		}
+
+		for _, h := range sorted {
+			if !v.LessThan(*h.Version) {
+				return h.Handler(ctx, p, stream)
```
acud: Small comment - we might want to have counter metrics on the `(semver,proto-name)` - so we could have in a dashboard metrics around the currently handled protocol version handlers on a node. It would give us good visibility at least for preliminary testing. Might be good to preemptively add that already, wdyt @janos?

acud (approved):
nicely done @janos, thanks for this

gacevicljubisa (approved):

akrem-chabchoub (approved):

pkg/p2p/p2p.go
janos: That would be useful, but we would have to provide the metric counter somehow, preferably not to be in the pacakge space, as we have managed to keep metrics nicely isolated per service. The simplest would be to add the prometheus.Counter as the argument to the NewVersionedHandlersFunc, so that every protocol would inject its own counter. Maybe as an option construct, I will think about it a bit more and figure out something.

janos: In order to have a pluggable metrics for the versioned handlers, I have added options WithMetricCounter and WithOnMatchFunc but I have also moved the versioned handler logic into a dedicated subpackage ./pkg/p2p/versioned to keep the core pkg/p2p API clean. I have also updated the tests and examples to reflect these changes.

acud (approved):
very nice, thanks @janos

PR #5551, opened 2026-07-30
fix(api): bytes head entity headers

### LINKED ISSUE #5545: `HEAD /bytes/<ref>` omits Content-Length and all other entity headers
**Bee version:** 2.8.1-7cf53193 (API 8.1.0).

## Summary

`HEAD /bytes/<ref>` returns a near-empty header set. `GET` on the same
reference returns `Content-Length`, `Content-Type`, `Accept-Ranges`, `Etag` and
`Last-Modified`. RFC 9110 §9.3.2: "The server SHOULD send the same header
fields in response to a HEAD request as it would have sent if the request
method had been GET."

## Reproduction

```
$ curl -s -D- -o /dev/null localhost:1633/bytes/<ref>
HTTP/1.1 200 OK
Accept-Ranges: bytes
Access-Control-Expose-Headers: Content-Disposition
Content-Length: 200000
Content-Type: application/octet-stream
Etag: "<ref>"
Last-Modified: Fri, 24 Jul 2026 21:23:17 GMT

$ curl -s -I localhost:1633/bytes/<ref>
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Accept-Ranges, Content-Encoding
```

`curl` exits 8 (`Weird server reply`) on the HEAD.

## Expected

HEAD returns the same entity headers as GET — at minimum `Content-Length`,
`Content-Type` and `Accept-Ranges`. The size is already known on the GET path
(it is the chunk span), so no new computation is required.

## Why it matters

Size-then-range is the standard way generic HTTP clients consume a large remote
object without downloading it: HEAD for the length, then ranged GETs. Bee's
ranged GET support is correct and works well — the missing HEAD is the only
thing blocking that whole class of client.

Concrete case: `nbdkit`'s stock `curl` plugin refuses to open a Swarm reference,
because it HEADs for the size first. With this fixed, `nbdkit curl
http://localhost:1633/bytes/<ref>` exposes any Swarm reference as a block
device with no Swarm-specific code at all — mount an image, loopback-boot an
ISO, `qemu-img` straight off the network. Today that needs a shim to work
around the missing header.

The same pattern affects `rclone`, `aria2`, HTTP range-readers in data
libraries, and anything built on `libcurl`'s remote-file handling.

gacevicljubisa:
### Checklist

### Description

`HEAD /bytes/<ref>` did not return usable entity headers. Three separate defects:

1. **Invalid `Content-Length`.** The redundancy level is packed into the most significant byte of the root chunk span (`redundancy.EncodeLevel`), so reading the span as `int64` yielded a negative number for any upload with redundancy level >= 1 spanning more than one chunk. Bee sent `Content-Length: -9151314442816647872`; curl rejects that and aborts header parsing (exit 8), which is why the response looked like it had no headers at all.
2. **Encrypted references returned 404.** The handler looked up the full 64-byte reference (address + key) as if it were a chunk address.
3. **`Accept-Ranges` was never sent** on HEAD, for `/bytes` or `/bzz`, even though both serve ranged GETs.

`bytesHeadHandler` now delegates to `downloadHandler` with `headersOnly`, the same path `bzzHeadHandler` already uses. HEAD and GET therefore share one implementation and the length comes from the joiner, which strips the encoded redundancy level and splits an encrypted reference into address and key. `Accept-Ranges` is set on that shared header-only branch, so both endpoints are covered.

Behavior change worth noting: HEAD now reports the same status as GET for references that cannot be served, where it previously always returned 404. Pinned by `TestBytesHeadErrorsMatchGet`.

Documented `Accept-Ranges`, `ETag` and the `500` response on `HEAD /bytes/{address}` and `HEAD /bzz/{address}`.

`info.version` left at 8.1.0. The added response headers are additive, but the HEAD status change above is arguably breaking — happy to bump if you would rather record it.

Size-then-range is how generic HTTP clients read a large remote object: HEAD for the length, then ranged GETs. Bee's ranged GET support already works, so the missing HEAD headers were the only thing blocking that whole class of client (`nbdkit`, `rclone`, `aria2`, anything on libcurl's remote-file handling).

### Related Issue (Optional)

Closes #5545

### AI Disclosure

### DIFF
--- a/pkg/api/bytes_test.go
+++ b/pkg/api/bytes_test.go
@@ -486,6 +486,17 @@ func TestBytesHead(t *testing.T) {
 						jsonhttptest.WithNonEmptyResponseHeader("Last-Modified"),
 					)
 				}
+
+				// The range is resolved against the length decoded from the root
+				// chunk span, so it has to hold for every redundancy level and for
+				// encrypted references, where that span is decrypted first.
+				for _, method := range []string{http.MethodHead, http.MethodGet} {
+					jsonhttptest.Request(t, client, method, resource, http.StatusPartialContent,
+						jsonhttptest.WithRequestHeader(api.RangeHeader, "bytes=0-10"),
+						jsonhttptest.WithExpectedContentLength(11),
+						jsonhttptest.WithExpectedResponseHeader(api.ContentRangeHeader, fmt.Sprintf("bytes 0-10/%d", len(content))),
+					)
+				}
 			})
 		}
 	}
--- a/pkg/api/bytes_test.go
+++ b/pkg/api/bytes_test.go
@@ -441,3 +441,86 @@ func TestBytesRedundancyLevel(t *testing.T) {
 		})
 	}
 }
+
+// TestBytesHead tests that a HEAD request reports the same entity headers as a
+// GET, for every redundancy level and for encrypted references. The redundancy
+// level is encoded into the most significant byte of the root chunk span and has
+// to be stripped before the length is read out of it, and an encrypted reference
+// carries a decryption key that is not part of the root chunk address.
+func TestBytesHead(t *testing.T) {
+	t.Parallel()
+
+	g := mockbytes.New(0, mockbytes.MockTypeStandard).WithModulus(255)
+	content, err := g.SequentialBytes(swarm.ChunkSize * 10)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for level := redundancy.NONE; level <= redundancy.PARANOID; level++ {
+		for _, encrypt := range []bool{false, true} {
+			t.Run(fmt.Sprintf("level %d encrypt %v", level, encrypt), func(t *testing.T) {
+				t.Parallel()
+
+				client, _, _, _ := newTestServer(t, testServerOptions{
+					Storer: mockstorer.New(),
+					Post:   mockpost.New(mockpost.WithAcceptAll()),
+				})
+
+				var resp struct {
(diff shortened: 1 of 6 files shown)


pkg/api/bzz.go
```diff
@@ -794,6 +794,10 @@ func (s *Service) downloadHandler(logger log.Logger, w http.ResponseWriter, r *h
 	w.Header().Add(AccessControlExposeHeaders, ContentDispositionHeader)
 
 	if headersOnly {
```
martinconic, line 801: Early-returning 200 OK here bypasses Go's http.ServeContent, which causes HEAD requests containing a Range header (e.g. HEAD /bytes/<hash> with Range: bytes=0-99) to return 200 OK with the total content length and no Content-Range header.

In contrast, a GET request with Range: bytes=0-99 returns 206 Partial Content with Content-Range: bytes 0-99/<total> and Content-Length: 100.

According to RFC 9110 Section 9.3.2, HEAD requests must return identical entity headers to GET. Conditional HEAD requests (If-None-Match, If-Modified-Since) also fail to return 304 Not Modified for the same reason.

pkg/api/bytes_test.go
martinconic, line 1: Could we add a test case to TestBytesHead for HEAD requests with a Range header (e.g. Range: bytes=0-10)?

martinconic (approved):

acud (approved):
nice

sbackend123 (approved):

PR #5552, opened 2026-07-31
feat(puller): add pullsync rate metric

akrem-chabchoub:
### Checklist

### Description
The bee node shrinks its storage radius only when historical syncing has fully stopped (SyncRate() == 0). Until now, we couldn't see that sync rate anywhere, so when the radius didn't decrease, we couldn't tell why, or how close it was to happening.

With this change, we expose that number as `bee_puller_pullsync_rate`.

Now we can put it on a dashboard and watch the story unfold: the rate is high while we catch up on history → it decays to 0 when syncing finishes → and the moment it hits 0, our radius is allowed to step down. With this metric, we make the previously invisible condition that controls radius decrease visible.

### AI Disclosure

### DIFF
--- a/pkg/puller/metrics.go
+++ b/pkg/puller/metrics.go
@@ -15,9 +15,10 @@ type metrics struct {
 	SyncedCounter         *prometheus.CounterVec // number of synced chunks
 	SyncWorkerErrCounter  prometheus.Counter     // count number of errors
 	MaxUintErrCounter     prometheus.Counter     // how many times we got maxuint as topmost
+	PullsyncRate          prometheus.GaugeFunc   // rate of historical syncing
 }
 
-func newMetrics() metrics {
+func newMetrics(pullsyncRate func() float64) metrics {
 	subsystem := "puller"
 
 	return metrics{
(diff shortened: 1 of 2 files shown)


martinconic (approved):

martinconic:
Please add a description

acud (approved):
lgtm, just that you can still see the rate of historical syncing by looking at the metric of synced chunks, there's a label `historical` that shows you the number of chunks arriving from historical syncing, which is in fact a proxy to this measurement. either way it is fine from my side to expose this metric.

pkg/puller/metrics.go
```diff
@@ -51,6 +52,12 @@ func newMetrics() metrics {
 			Name:      "max_uint_errors",
 			Help:      "Total max uint errors.",
 		}),
+		PullsyncRate: prometheus.NewGaugeFunc(prometheus.GaugeOpts{
```
gacevicljubisa, line 55: Should the name be HistoricalSyncRate, because it doesn't count live syncing?

gacevicljubisa (approved):

janos (approved):

pkg/puller/metrics.go
akrem-chabchoub: I kept the name `pullsync_rate` on purpose though: it's the same value the status endpoint already exposes as `pullsyncRate`, so using the same name makes it easy to correlate the metric with the status API.

Issue #5506, opened 2026-06-19
Add a transport label to libp2p connection metrics (TCP vs WS/WSS vs QUIC)


darkobas2:
### Summary
The libp2p connection metrics don't expose which **transport** a connection uses, so there's no way to observe TCP vs WebSocket (`ws`/`wss`) vs QUIC from metrics. Adding a `transport` label to the connection counters would make transport usage observable.

### Current behaviour
Counters such as `bee_libp2p_created_connection_count`, `bee_libp2p_handled_connection_count`, `bee_libp2p_public_address_connections_total`, and `bee_libp2p_private_address_connections_total` only carry generic labels (pod/instance/etc.) — there is no transport dimension. With WSS multiaddrs in use on some deployments, there is currently no Prometheus-visible way to confirm peers are actually connecting over WSS rather than silently falling back to TCP. Per-peer transport is only visible via the API (node `/addresses` and the peers/topology endpoints), which isn't scrape-able.

### Proposed
Add a `transport` label (values like `tcp`, `ws`, `wss`, `quic-v1`) to the libp2p connection metrics — at minimum the created/handled connection counters and the public/private address connection totals.

### Why it's useful
- Verify WSS-enabled nodes/bootnodes are actually serving and being reached over WSS.
- Catch transport regressions (e.g. WSS quietly failing → everything on TCP) from metrics/alerts alone, without polling the peers API.
- Build "connections by transport" dashboard panels.

Happy to help test a draft implementation.

PR #5565, opened 2026-08-18
feat: add new metrics

sbackend123:
### Checklist

### Description
- Add transport label to libp2p created/handled connection metrics (tcp / ws / wss / quic*).
- Move /wss + QUIC classification into bzz.ClassifyTransport.
- Replace separate public/private counters with public="true|false" on handled_connection_count.
- Pre-init label series so unused transports show as 0, not missing.

**Breaking changes**
- created_connection_count / handled_connection_count: Counter → CounterVec.
- handled_connection_count now has labels {transport, public}.
- Removed public_address_connections_total / private_address_connections_total — use handled_connection_count{public=...} instead.
- Dashboards using count(...) as a denominator will under-count after upgrade; prefer sum by (...).

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5506

### AI Disclosure

### DIFF
--- a/pkg/p2p/libp2p/metrics.go
+++ b/pkg/p2p/libp2p/metrics.go
@@ -5,16 +5,26 @@
 package libp2p
 
 import (
+	"github.com/ethersphere/bee/v2/pkg/bzz"
 	m "github.com/ethersphere/bee/v2/pkg/metrics"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
 	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	connectionTransportLabelName = "transport"
+	connectionTransportHelp      = "The 'transport' label is one of: tcp, ws, wss, quic-v1, quic, unknown."
 )
 
 type metrics struct {
 	// all metrics fields must be exported
 	// to be able to return them by Metrics()
 	// using reflection
-	CreatedConnectionCount     prometheus.Counter
-	HandledConnectionCount     prometheus.Counter
+	CreatedConnectionCount     *prometheus.CounterVec
+	HandledConnectionCount     *prometheus.CounterVec
+	PublicAddressConnections   *prometheus.CounterVec
+	PrivateAddressConnections  *prometheus.CounterVec
 	CreatedStreamCount         prometheus.Counter
 	ClosedStreamCount          prometheus.Counter
 	StreamResetCount           prometheus.Counter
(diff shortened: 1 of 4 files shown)


darkobas2:
Running this on our mainnet dev canary (`dev-bee-storage/bee-1`, build `2.8.2-rc1-66a4ec3c`). Both fixes behave: WSS now lands on its own label instead of leaking into `tcp`, and the zero-init gives flat lines instead of gaps. Thanks for the quick turnaround.

Live from the canary:

```
created_connection_count{transport="tcp"}                 55
handled_connection_count{public="true",transport="tcp"}  116
handled_connection_count{public="true",transport="wss"}    7
handled_connection_count{public="false",transport="tcp"}   2
```

(all other combinations present at 0)

One small suggestion: `quic` and `quic-v1` can never be observed, since bee registers only the TCP and websocket transports (`libp2p.go:358` and `381-383`). Worth keeping the enum cases and the `ClassifyTransport` branches for the future, but dropping the two from `transportLabelValues`, which only drives pre-initialisation. If a QUIC transport is ever added the series will appear on first observation anyway. As it stands that's 6 of the 18 series per node that are structurally always zero.

On cardinality more broadly: this takes a node from 2 series to 18, but a bee node already exposes around 1570, so it works out at roughly a 1% increase. No concern from our side. The `public` label in particular earns its place, as it's the clearest inbound reachability signal we have.

martinconic:
Could we add test cases to TestClassifyTransport and TestTransportTypePriority covering the new transport classifications? Specifically:

* Multiaddresses with the /wss component (ma.P_WSS)
* Multiaddresses with /udp/.../quic-v1 and /udp/.../quic
* Priority tests for TransportQUIC and TransportQUICV1

pkg/p2p/libp2p/metrics.go
```diff
@@ -31,20 +53,41 @@ type metrics struct {
...
 			Name:      "created_connection_count",
-			Help:      "Number of initiated outgoing libp2p connections.",
-		}),
-		HandledConnectionCount: prometheus.NewCounter(prometheus.CounterOpts{
+			Help:      "Number of initiated outgoing libp2p connections. " + transportHelp,
+		},
+		[]string{connectionTransportLabelName},
+	)
+
+	handledConnectionCount := prometheus.NewCounterVec(
```
martinconic, line 70: handled_connection_count includes the public label (true/false), but created_connection_count only has transport.

Is there a reason not to include public on created_connection_count as well? In Connect(), we also have the remote multiaddr via stream.Conn().RemoteMultiaddr(), so having public on both would give symmetric visibility for outbound dials.

martinconic (approved):

akrem-chabchoub (commented):
Looks good, but why this is canceled ?
image

akrem-chabchoub (approved):

PR #5568, opened 2026-08-21
feat: reserve-expanding sync eligibility for redistribution game (SWIP-44)

aloknerurkar:
Checklist

- [x] I have read the coding guide.

Description

Implements SWIP-44: Reserve-Expanding Sync Eligibility.

Currently, Bee excludes a node from participating in the redistribution game whenever any historical pullsync is active (SyncRate() > 0). When a node is expanding its reserve into a lower bin ($d-1$) in preparation for a storage radius decrease, this sync can take hours, causing the node to miss rounds despite maintaining a complete and valid reserve at the current storage depth $d$.

Because the reserve digest $H(R_N(d))$ is computed strictly over PO bins $p \geq d$, chunks arriving in bins $p < d$ are disjoint from the reserve and do not influence the digest. This PR amends the eligibility check to allow participation during sub-depth sync:

1. pkg/puller:
   - Added bin-level active historical sync worker tracking (activeHistSyncs [swarm.MaxBins]int, activeHistSyncsMu sync.RWMutex).
   - Exposed IsBinSyncing(bin uint8) bool and depth-aware IsReserveSynced(depth uint8) bool.
   - Updated mock syncer and added TestIsReserveSynced.
2. pkg/storageincentives:
   - Updated Agent constructor and sampling handler to evaluate reserveSyncedFunc(committedDepth) instead of blanket global sync state.
   - Added unit tests covering game participation during sub-depth sync vs. reserve-depth sync.
3. pkg/node:
   - Wired pullerService.IsReserveSynced(depth) into storageincentives.New.

Open API Spec Version Changes (if applicable)
N/A

Motivation and Context (Optional)
Addresses SWIP-44 (ethersphere/SWIPs#92).

Related Issue (Optional)
- Addresses https://github.com/ethersphere/SWIPs/pull/92

Screenshots (if appropriate):
N/A

AI Disclosure


pkg/storageincentives/staking/contract.go
```diff
@@ -13,10 +13,11 @@ import (
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethersphere/go-sw3-abi/sw3abi"
+
 	"github.com/ethersphere/bee/v2/pkg/sctx"
 	"github.com/ethersphere/bee/v2/pkg/transaction"
 	"github.com/ethersphere/bee/v2/pkg/util/abiutil"
-	"github.com/ethersphere/go-sw3-abi/sw3abi"
```
sbackend123, line 19: looks like unnecessary change

pkg/storageincentives/staking/contract_test.go
```diff
@@ -15,6 +15,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+
```
sbackend123: may be let's remove changes like new empty strings, import order and so on

pkg/storageincentives/main_test.go
```diff
@@ -7,8 +7,9 @@ package storageincentives_test
 import (
 	"testing"
 
-	"github.com/ethersphere/bee/v2/pkg/bmt"
 	"go.uber.org/goleak"
+
```
sbackend123: same comment as above

pkg/storageincentives/agent.go
```diff
@@ -416,8 +417,10 @@ func (a *Agent) handleSample(ctx context.Context, round uint64) (bool, error) {
 	a.metrics.NeighborhoodSelected.Inc()
 	a.logger.Info("neighbourhood chosen", "round", round)
 
-	if !a.state.IsFullySynced() {
-		a.logger.Info("skipping round because node is not fully synced")
+	isSynced := a.reserveSyncedFunc(committedDepth)
+	a.state.SetFullySynced(isSynced)
```
sbackend123: This overwrites IsFullySynced using CommittedDepth, while phase transitions set the same field using StorageRadius. With capacity doubling and an active sync between these depths, the status changes from false to true during sampling even though the physical reserve is still syncing, making /redistributionstate phase-dependent and misleading. Since eligibility already uses the local isSynced value, I would consider not updating IsFullySynced here or introducing a separate committed-reserve status.

sbackend123 (commented):
Nice job!

I would just clean up it a little bit - removing "formatting" changes, so PR looks bigger than it is.

aloknerurkar:
> Nice job!
> 
> I would just clean up it a little bit - removing "formatting" changes, so PR looks bigger than it is.

So this formatter is actually from our coding style doc. We used to follow it but I see now we dont.
https://github.com/ethersphere/bee/blob/master/CODINGSTYLE.md#code-formatting
https://github.com/ethersphere/bee/blob/master/Makefile#L74

I had my editor setup to run these formatters as soon as I open any file. I will revert the changes for now but we should update the coding style and makefile for others in future. I personally prefer this formatting as it breaks down std library, external and internal imports.

sbackend123 (approved):

pkg/node/node.go
```diff
@@ -1318,10 +1319,10 @@ func NewBee(
 
 			redistributionContract := redistribution.New(swarmAddress, overlayEthAddress, logger, transactionService, redistributionContractAddress, abiutil.MustParseABI(chainCfg.RedistributionABI), contractGasLimit)
 
-			isFullySynced := func() bool {
+			isReserveSynced := func(depth uint8) bool {
 				reserveThreshold := reserveCapacity * 5 / 10
 				logger.Debug("Sync status check evaluated", "stabilized", detector.IsStabilized())
-				return localStore.ReserveSize() >= reserveThreshold && pullerService.SyncRate() == 0 && detector.IsStabilized()
+				return localStore.ReserveSize() >= reserveThreshold && pullerService.IsReserveSynced(depth) && detector.IsStabilized()
```
martinconic: When a node uses Capacity Doubling, it stores chunks locally at radius 8, but commits on the blockchain at depth 9. It is responsible for 2 adjacent "sister neighborhoods". The PR only checks if bins >= 9 are synced, completely ignoring bin 8. Is this right?

martinconic: `ReserveSize()` counts all sub-depth chunks, which can trick a new node into thinking it's ready. I think we should check down to `StorageRadius` and use `ReserveSizeWithinRadius()`.

pkg/puller/puller.go
```diff
@@ -409,9 +435,18 @@ func (p *Puller) syncPeerBin(parentCtx context.Context, peer *syncPeer, bin uint
 	}
 
 	if cursor > 0 {
+		p.activeHistSyncsMu.Lock()
+		p.activeHistSyncs[bin]++
```
martinconic, line 439: When bin = 32, Go panics: index out of range [32] with length 32 , right? Should we have dynamic size for `activeHistSyncs` ?

pkg/storageincentives/agent.go
```diff
@@ -221,7 +222,7 @@ func (a *Agent) start(blockTime time.Duration, blocksPerRound, blocksPerPhase ui
 		a.logger.Info("entered new phase", "phase", currentPhase.String(), "round", round, "block", block)
 
 		a.state.SetCurrentEvent(currentPhase, round)
-		a.state.SetFullySynced(a.fullSyncedFunc())
+		a.state.SetFullySynced(a.reserveSyncedFunc(a.store.StorageRadius()))
```
martinconic, line 225: Should we use `committedDepth()` here in case of doubling?

pkg/node/node.go
aloknerurkar: Did you check the [SWIP](https://github.com/ethersphere/SWIPs/pull/92/changes)?
I believe this is intentional and consistent with how the reserve digest itself is calculated.

CommittedDepth() (doublingFactor + StorageRadius()) is what the redistribution game and the sample are actually keyed on — see sample.go's own doc comment[L56-60]. With a physical radius of 8 and a doubling factor of 1, committed depth is 9, and ReserveSample only ever reads chunks at proximity ≥ 9 to the anchor — bin 8 chunks (which straddle both sister neighborhoods) simply aren't part of what gets hashed into the digest.

pkg/puller/puller.go
aloknerurkar: So bin max value is 31. swarm.MaxBins is 32 so 0-31 are the permitted values. But you are right, better safe than sorry.

pkg/storageincentives/agent.go
aloknerurkar: Good question, but I'd keep this as StorageRadius() rather than switching to CommittedDepth() — they answer different questions and this is deliberately the physical one.

This line sets IsFullySynced, which is purely an operator-facing status field (/redistributionstate) — it reports "is my actual local reserve, at my current physical radius, synced," independent of whatever the game happens to be checking eligibility against this round.

CommittedDepth() is doublingFactor + StorageRadius() — it reflects the depth the node has committed to on-chain, which during an active capacity-doubling sync is ahead of the node's real physical radius.

So the two call sites intentionally use different depths for different purposes: this one reports current physical state, the other gates game participation against the committed depth.

acud:
@aloknerurkar please rebase

pkg/puller/puller.go
```diff
@@ -106,6 +107,9 @@ type Puller struct {
 
 	rate *rate.Rate // rate of historical syncing
```
acud, line 108: since you now count the active historical syncing using the number of active goroutines in radius bins, this thing is no longer used if i get it right. there are a couple of changes here - one is the fact we're no longer using these sliding window rate which is probably a good thing, but the problem is that the goroutine approach would be a bit tricky because it would basically flip on/off all the time. maybe a bit more thought needs to do be done here so we can get a smoothed reading that could be a bit more informative (but also that doesn't take half an hour to reach below the reading we need it to get to in order to conclude "not syncing")

PR #5571, opened 2026-08-23
fix(staking): report minimum deposit when commitment would decrease

### LINKED ISSUE #5193: Unclear error message when depositing stake and commitment decreases
### Context
Version: 2.6.0-d0aa8b93

### Summary
It seems that since the [commitment protection](https://github.com/ethersphere/storage-incentives/commit/12d8bda6b0f4496259cc7e98ae3a3ebb36c7129e#diff-bcfc0067dcd0c2736d43b73f5981fd160ada82fb3f479355ffa9b92bf8007839) changes, it is not always possible to stake small values (probably not a problem), and the error message is not helpful (this is the key problem).

### Expected behavior
Good: When I try depositing stake, and the price (price oracle) has already increased such that my new committed stake would be less than before, there should be a clear error message indicating what the new minimum deposit is.

Better: Bee always exposes the minimum deposit value in a new or existing endpoint; 10.0 BZZ first, and then 1 PLUR, or more if the price oracle has increased the price.

### Actual behavior
Precondition:

```
Staked xBZZ: 11.1000000000000002
Withdrawable staked xBZZ: 0.0000000000000000
```

Attempt to stake 1 PLUR:

```
"time"="2025-08-18 17:51:48.114234" "level"="info" "logger"="node/transaction" "msg"="pending transaction confirmed" "sender_address"="0x43E942d82F72ff8185D9fC15024eDE4889c6e2cA" "tx"="0xa03cd216643207972cda69a71a97d6d114ce0c5bfcb4ba11f020a8a6165bddee"
"time"="2025-08-18 17:51:48.229575" "level"="debug" "logger"="node/api/post_stake_deposit" "msg"="deposit failed" "error"="deposit stake: stakedAmount 1: Invalid, intrinsic gas too low: VM execution error."
"time"="2025-08-18 17:51:48.229616" "level"="error" "logger"="node/api/post_stake_deposit" "msg"="deposit failed"
"time"="2025-08-18 17:51:48.229682" "level"="debug" "logger"="node/api" "msg"="api access" "ip"="127.0.0.1" "method"="POST" "host"="localhost:1633" "uri"="/stake/1" "proto"="HTTP/1.1" "status"=500 "size"=39 "duration"="5.395927846s" "user-agent"="axios/0.30.0"
```

The error message is about gas too low?

So I deposit 0.1 BZZ instead of 1 PLUR, which happens successfully:

```
Staked xBZZ: 11.2000000000000002
Withdrawable staked xBZZ: 0.0000000000008196
```

I can now deposit 1 PLUR, which failed previously:

```
Staked xBZZ: 11.2000000000000003
Withdrawable staked xBZZ: 0.0000000000008197
```

A good indicator whether it will succeed of fail is that the withdrawable BZZ is greater than 0.

After waiting a few rounds, the price oracle increased the price:

Image

Due to the price increase, withdrawable BZZ falls back to 0:

```
Staked xBZZ: 11.2000000000000003
Withdrawable staked xBZZ: 0.0000000000000000
```

And right after that, it starts failing with HTTP 500 again whenever I try depositing 1 PLUR:

```
"time"="2025-08-18 18:31:40.161637" "level"="info" "logger"="node/transaction" "msg"="pending transaction confirmed" "sender_address"="0x43E942d82F72ff8185D9fC15024eDE4889c6e2cA" "tx"="0x7aed852b8968f217e87505b868af3562a0a4fbf32d5e8bf15655c6b46b41faf8"
"time"="2025-08-18 18:31:40.279709" "level"="debug" "logger"="node/api/post_stake_deposit" "msg"="deposit failed" "error"="deposit stake: stakedAmount 1: Invalid, intrinsic gas too low: VM execution error."
"time"="2025-08-18 18:31:40.279743" "level"="error" "logger"="node/api/post_stake_deposit" "msg"="deposit failed"
"time"="2025-08-18 18:31:40.279797" "level"="debug" "logger"="node/api" "msg"="api access" "ip"="127.0.0.1" "method"="POST" "host"="localhost:1633" "uri"="/stake/1" "proto"="HTTP/1.1" "status"=500 "size"=39 "duration"="10.451871458s" "user-agent"="axios/0.30.0"
```

And the error message is - confusingly - about gas too low.

I attempted manually setting the gas price and gas limit values with headers, which at least triggered a `transaction reverted` instead of `Invalid, intrinsic gas too low`:

```
"time"="2025-08-18 17:19:51.479107" "level"="debug" "logger"="node/api/post_stake_deposit" "msg"="deposit failed" "error"="deposit stake: stakedAmount 1: transaction reverted: VM execution error."
```

Some examples of such failed transactions:

- https://gnosisscan.io/tx/0x05822ea9c06c246b8b1f6231c8e168581eb8f89deed526facccae1c2a2ff6565
- https://gnosisscan.io/tx/0x7d07311562cb3c2f8d0abf7501316c61769407dda47206902e9ec3eb20f1d273
- https://gnosisscan.io/tx/0x4811a8570572b5118290303db6670114389172b6ac5ba24f5a49d19f84745b38

### Steps to reproduce
1. Stake the initial 10 BZZ
2. Wait for the price oracle to increase the price
3. Attempt to deposit 1 PLUR
4. Observe the error message about gas being too low
5. Deposit 0.1 BZZ instead
6. Attempt to deposit 1 PLUR again
7. Observe that it succeeds
8. Goto 2

### Possible solution
Please see the suggested scenarios in the "Expected behavior" section above.

sbackend123:
### Checklist

### Description

- Fixes [#5193](https://github.com/ethersphere/bee/issues/5193): small stake deposits that would decrease commitment no longer fail with a misleading intrinsic gas too low / HTTP 500.
- Pre-validates deposits against the staking contract rules (initial MIN_STAKE * 2^height floor and non-decreasing commitment using the price oracle) and returns HTTP 400 with the required minimumDeposit.
- Exposes minimumDeposit on GET /stake.
- Resolves the price oracle address from the staking contract (OracleContract) at node startup, so custom/overridden staking contracts use the correct oracle.

- GET /stake → { "stakedAmount": "...", "minimumDeposit": "..." }
- POST /stake/{amount} on insufficient deposit → 400 with { "code": 400, "message": "insufficient stake amount", "minimumDeposit": "..." }
- OpenAPI bumped to 8.2.0

### AI Disclosure

### DIFF
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 
 info:
-  version: 8.1.0
+  version: 8.2.0
   title: Bee API
   description: "API endpoints for interacting with the Swarm network, supporting file operations, messaging, and node management"
 
@@ -2567,7 +2567,7 @@ paths:
   "/stake":
     get:
       summary: Get the staked amount.
-      description: This endpoint fetches the total staked amount from the blockchain.
+      description: This endpoint fetches the total staked amount and the minimum additional deposit from the blockchain. The first deposit is at least 0.1 BZZ times 2^height. Subsequent deposits are at least 1 PLUR, or more if the price oracle has increased since the last deposit.
       tags:
         - Staking
       responses:
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -2558,7 +2558,11 @@ paths:
               schema:
                 $ref: "SwarmCommon.yaml#/components/schemas/StakeTransactionResponse"
         "400":
-          $ref: "SwarmCommon.yaml#/components/responses/400"
+          description: Deposit amount is below the required minimum.
+          content:
+            application/json:
+              schema:
+                $ref: "SwarmCommon.yaml#/components/schemas/StakeDepositErrorResponse"
         "500":
           $ref: "SwarmCommon.yaml#/components/responses/500"
         default:
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -2571,7 +2571,7 @@ paths:
   "/stake":
     get:
       summary: Get the staked amount.
-      description: This endpoint fetches the total staked amount and the minimum additional deposit from the blockchain. The first deposit is at least 0.1 BZZ times 2^height. Subsequent deposits are at least 1 PLUR, or more if the price oracle has increased since the last deposit.
+      description: This endpoint fetches the total staked amount and the minimum additional deposit from the blockchain. The first deposit is at least 10 BZZ times 2^height. Subsequent deposits are at least 1 PLUR, or more if the price oracle has increased since the last deposit.
       tags:
         - Staking
       responses:
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -1224,7 +1224,16 @@ func NewBee(
 		stakingContractAddress = common.HexToAddress(o.StakingContractAddress)
 	}
 
-	stakingContract := staking.New(overlayEthAddress, stakingContractAddress, abiutil.MustParseABI(chainCfg.StakingABI), bzzTokenAddress, transactionService, common.BytesToHash(nonce), contractGasLimit, uint8(o.ReserveCapacityDoubling), chainCfg.IncentivesPriceOracleAddress)
+	stakingContractABI := abiutil.MustParseABI(chainCfg.StakingABI)
+	var stakingPriceOracleAddress common.Address
+	if chainEnabled {
+		stakingPriceOracleAddress, err = staking.GetPriceOracleAddress(ctx, stakingContractAddress, stakingContractABI, transactionService)
+		if err != nil {
+			return nil, fmt.Errorf("get staking price oracle address: %w", err)
+		}
+	}
+
+	stakingContract := staking.New(overlayEthAddress, stakingContractAddress, stakingContractABI, bzzTokenAddress, transactionService, common.BytesToHash(nonce), contractGasLimit, uint8(o.ReserveCapacityDoubling), stakingPriceOracleAddress)
 
 	if chainEnabled {
 
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -1224,16 +1224,7 @@ func NewBee(
 		stakingContractAddress = common.HexToAddress(o.StakingContractAddress)
 	}
 
-	stakingContractABI := abiutil.MustParseABI(chainCfg.StakingABI)
-	var stakingPriceOracleAddress common.Address
-	if chainEnabled {
-		stakingPriceOracleAddress, err = staking.GetPriceOracleAddress(ctx, stakingContractAddress, stakingContractABI, transactionService)
-		if err != nil {
-			return nil, fmt.Errorf("get staking price oracle address: %w", err)
-		}
-	}
-
-	stakingContract := staking.New(overlayEthAddress, stakingContractAddress, stakingContractABI, bzzTokenAddress, transactionService, common.BytesToHash(nonce), contractGasLimit, uint8(o.ReserveCapacityDoubling), stakingPriceOracleAddress)
+	stakingContract := staking.New(overlayEthAddress, stakingContractAddress, abiutil.MustParseABI(chainCfg.StakingABI), bzzTokenAddress, transactionService, common.BytesToHash(nonce), contractGasLimit, uint8(o.ReserveCapacityDoubling))
 
 	if chainEnabled {
 
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -1224,7 +1224,7 @@ func NewBee(
 		stakingContractAddress = common.HexToAddress(o.StakingContractAddress)
 	}
 
-	stakingContract := staking.New(overlayEthAddress, stakingContractAddress, abiutil.MustParseABI(chainCfg.StakingABI), bzzTokenAddress, transactionService, common.BytesToHash(nonce), contractGasLimit, uint8(o.ReserveCapacityDoubling))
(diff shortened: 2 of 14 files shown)


pkg/storageincentives/staking/contract.go
```diff
@@ -154,11 +198,51 @@ func (c *contract) UpdateHeight(ctx context.Context) (common.Hash, bool, error)
...
+	}
+
+	return calculateMinDeposit(potential, committed, price, c.height), nil
+}
+
+// calculateMinDeposit returns the minimum additional deposit in PLUR that manageStake will accept according to contract.
+func calculateMinDeposit(potential, committed *big.Int, price uint32, height uint8) *big.Int {
+	minAdd := big.NewInt(1)
+
+	// The contract applies the minimum stake floor only when creating a stake.
```
aloknerurkar: The staking contract applies the initial minimum based on:
```solidity
_addAmount < MIN_STAKE * 2 ** _height && _stakingSet == 0
```
That is, it checks lastUpdatedBlockNumber == 0, not whether potentialStake == 0.

A possible state is:

1. A stake was previously created, so lastUpdatedBlockNumber != 0.
2. The effective/committed stake later becomes zero.
3. The potential stake is withdrawn, leaving potentialStake == 0.
4. A subsequent deposit is made.

In that state, the contract does not apply the initial MIN_STAKE * 2^height floor, but calculateMinDeposit reports that floor anyway. The API can therefore overstate the required minimum and reject deposits that the contract would accept.

The calculation needs access to the stake’s lastUpdatedBlockNumber or an equivalent “stake has previously been initialized” value. The current getStake helper only returns committed and potential stake.

Suggested fix:
- Return lastUpdatedBlockNumber from getStake.
- Pass an isInitialized/stakingSet boolean into calculateMinDeposit.
- Mirror the contract’s _stakingSet == 0 condition exactly.

A test should cover potentialStake == 0 with lastUpdatedBlockNumber != 0.

pkg/node/node.go
```diff
@@ -1224,7 +1224,16 @@ func NewBee(
 		stakingContractAddress = common.HexToAddress(o.StakingContractAddress)
 	}
 
-	stakingContract := staking.New(overlayEthAddress, stakingContractAddress, abiutil.MustParseABI(chainCfg.StakingABI), bzzTokenAddress, transactionService, common.BytesToHash(nonce), contractGasLimit, uint8(o.ReserveCapacityDoubling))
+	stakingContractABI := abiutil.MustParseABI(chainCfg.StakingABI)
```
aloknerurkar: L1227-L1235
I see a bunch of issues here. This staking contract initialization is leaking unnecessary details. Now pkg/node is responsible for calling a staking-package helper to discover an internal dependency and passing it back to the staking constructor.

If staking contract changes its oracle discovery logic tomorrow, we will have to change pkg/node as well. If chainEnabled is false, we are passing zero address to staking constructor which may end up using this in GetMinDeposit func call.

I see two options:
1. Either the constructor should call this function and return error OR
2. We do lazy oracle resolution. i.e we resolve the oracle contract in GetMinDeposit when it needs the current price.

I would prefer the second option.

Second issue is along with chainEnabled we have storage incentives enabled check as well. We could have nodes that enable chain but not storage incentives. So it is better to do this lazily inside staking contract when GetMinDeposit is called.

pkg/storageincentives/staking/contract.go
```diff
@@ -154,11 +198,51 @@ func (c *contract) UpdateHeight(ctx context.Context) (common.Hash, bool, error)
...
+	if potential.Sign() == 0 {
+		minAdd.Lsh(new(big.Int).Set(MinimumStakeAmount), uint(height))
+	}
+
+	if price != 0 && committed.Sign() > 0 {
+		// User already has committed stake.
+		// Commitment protection: required = committed * price * 2^height
+		required := new(big.Int).SetUint64(uint64(price))
+		required.Lsh(required, uint(height)) // * 2^height
```
aloknerurkar: nit: although this will work as is, I would usually prefer assigning the result into a new variable or the same variable. According to me it is clearer and reduces maintenance risk. This is the style we usually followed in the codebase in general.

For eg. you can look at pkg/accounting/accounting.go:245, :251, and :257

pkg/storageincentives/staking/contract.go
sbackend123: Looks like mix of different scenarios to me, wrote in masttermost for further discussion

pkg/storageincentives/staking/contract.go
```diff
@@ -346,14 +398,91 @@ func (c *contract) getPotentialStake(ctx context.Context) (*big.Int, error) {
...
+	if !ok {
+		return 0, fmt.Errorf("unexpected oracle price type %T", results[0])
+	}
+
+	return price, nil
+}
+
+// getPriceOracleAddress resolves the price oracle from the staking contract on first use.
+func (c *contract) getPriceOracleAddress(ctx context.Context) (common.Address, error) {
+	if (c.priceOracleAddress != common.Address{}) {
```
aloknerurkar, line 457: This should be protected against concurrent access.

sbackend123: We already have stakingAccessHandler, which effectively serializes staking requests, so there is no race in the current call path. We can add a lock here, but with the current implementation it looks redundant to me

aloknerurkar (approved):

openapi/Swarm.yaml
```diff
@@ -2567,7 +2571,7 @@ paths:
   "/stake":
     get:
       summary: Get the staked amount.
-      description: This endpoint fetches the total staked amount from the blockchain.
+      description: This endpoint fetches the total staked amount and the minimum additional deposit from the blockchain. The first deposit is at least 0.1 BZZ times 2^height. Subsequent deposits are at least 1 PLUR, or more if the price oracle has increased since the last deposit.
```
martinconic: is this 0.1 BZZ the right value or should it be 10 BZZ ?

pkg/storageincentives/staking/contract.go
martinconic: I would say @aloknerurkar point of view is still valid and we should protect here as `pkg/storageincentives/staking` is a low-level domain package implementing staking.Contract. Core packages must maintain their own thread-safety invariants and cannot assume they will only ever be called through HTTP middleware (stakingAccessHandler).

martinconic (approved):

PR #5572, opened 2026-08-24
refactor: remove act from chunk and soc api

### LINKED ISSUE #5469: Remove ACT support from SOC and chunk endpoints
ACT (`Swarm-Act` header) is currently accepted on `POST /soc/{owner}/{id}` and `POST /chunks`, but in both cases it is either semantically broken, end-to-end non-functional, or misaligned with the purpose of those endpoints.

---

**`POST /soc/{owner}/{id}` — remove ACT**

1. **Data is never encrypted.** The SOC chunk is stored in plaintext before ACT runs. ACT then wraps only the *reference*, not the payload.
2. **Reference protection is meaningless.** A SOC address is deterministically derived from `owner + id` and publicly computable — encrypting a reference to it provides no protection.
3. **Logical inconsistency.** `putter.Done` is called on the original SOC address, but an ACT-encrypted address (a different address space) is returned to the caller.
4. **No decryption path exists.** `GET /soc/{owner}/{id}` has no ACT decryption middleware at all. There is no way to resolve an ACT-wrapped SOC reference back to its canonical endpoint, making any upload with `Swarm-Act` on this endpoint completely unusable end-to-end.

The owner-based signing model of SOC already provides a natural identity/access primitive; ACT is incompatible with it and should be removed from this endpoint.

---

**`POST /chunks` — remove ACT**

1. **Data is never encrypted.** Raw chunks are content-addressed plaintext. ACT wraps only the content-hash reference, not the data — unlike the BZZ pipeline where the payload is actually encrypted before chunking.
2. **Wrong abstraction layer.** The chunk endpoint exists for atomic, low-level chunk operations — used by tooling that manages chunk-level storage directly. It does not expose file-level semantics like the BZZ or bytes endpoints do. ACT is a file-access-control feature and does not belong at this layer.

---

**Proposal:**
- Remove `Swarm-Act` / `Swarm-Act-History-Address` header support from `POST /soc/{owner}/{id}` and `POST /chunks`.
- ACT should remain only on upload paths where the pipeline performs actual payload encryption: BZZ, bytes, and file/collection endpoints.

---

akrem-chabchoub:
### Checklist

#### Remove ACT support from /chunks and /soc

ACT never worked correctly on these two endpoints, the data isn't encrypted and, for /soc, there's no way to decrypt the reference back. Removed Swarm-Act handling from both, bumped the API version, and updated tests accordingly.
### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5469
### AI Disclosure

### DIFF
--- a/openapi/Swarm.yaml
+++ b/openapi/Swarm.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 
 info:
-  version: 8.1.0
+  version: 9.0.0
   title: Bee API
   description: "API endpoints for interacting with the Swarm network, supporting file operations, messaging, and node management"
 
@@ -345,8 +345,6 @@ paths:
             $ref: "SwarmCommon.yaml#/components/parameters/SwarmPostageBatchId"
           required: false
         - $ref: "SwarmCommon.yaml#/components/parameters/SwarmPostageStamp"
-        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmAct"
-        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmActHistoryAddress"
       requestBody:
         description: Chunk binary data containing at least 8 bytes.
         content:
@@ -362,8 +360,6 @@ paths:
               description: Tag UID from the request `swarm-tag` header if provided.
               schema:
                 $ref: "SwarmCommon.yaml#/components/schemas/Uid"
-            "swarm-act-history-address":
-              $ref: "SwarmCommon.yaml#/components/headers/SwarmActHistoryAddress"
           content:
             application/json:
               schema:
@@ -1013,8 +1009,6 @@ paths:
         - $ref: "SwarmCommon.yaml#/components/parameters/SwarmTagParameter"
         - $ref: "SwarmCommon.yaml#/components/parameters/SwarmPinParameter"
         - $ref: "SwarmCommon.yaml#/components/parameters/SwarmDeferredUpload"
-        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmAct"
-        - $ref: "SwarmCommon.yaml#/components/parameters/SwarmActHistoryAddress"
       requestBody:
         required: true
         description: The SOC binary data, composed of the span (8 bytes) and up to 4KB of payload.
@@ -1035,8 +1029,6 @@ paths:
               description: Tag UID, returned when an upload session is in use (either because `swarm-tag` was supplied, `swarm-deferred-upload` requested deferred mode, or `swarm-pin` was set).
               schema:
                 $ref: "SwarmCommon.yaml#/components/schemas/Uid"
-            "swarm-act-history-address":
-              $ref: "SwarmCommon.yaml#/components/headers/SwarmActHistoryAddress"
         "400":
           $ref: "SwarmCommon.yaml#/components/responses/400"
         "401":
--- a/pkg/api/chunk.go
+++ b/pkg/api/chunk.go
@@ -12,9 +12,7 @@ import (
 	"net/http"
 	"strconv"
 
-	"github.com/ethersphere/bee/v2/pkg/accesscontrol"
 	"github.com/ethersphere/bee/v2/pkg/cac"
-	"github.com/ethersphere/bee/v2/pkg/file/redundancy"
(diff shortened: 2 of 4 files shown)


sbackend123 (approved):

acud (approved):

openapi/Swarm.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 
 info:
-  version: 8.1.0
+  version: 9.0.0
```
martinconic: I think 8.2.0 would be better, I think 9.0.0 is a to drastic jump

martinconic (changes requested):

openapi/Swarm.yaml
akrem-chabchoub: Removing `Swarm-Act` / `Swarm-Act-History-Address` from `/chunks` and `/soc` is a removal of API surface, not a backward-compatible addition — per this repo's existing versioning convention that calls for a major bump, not minor. Precedents: removing auth went 6.0.0 → 7.0.0 (8c61408d), removing dev-mode went 7.4.1 → 8.0.0 (a9d5cb10). 8.2.0 would read as an additive change, which this isn't, so I'd keep the major bump. (Note the diff will also need rebasing — master is already at 8.1.1 after #5574, this branch is based off 8.1.0.)

pkg/api/chunk.go
martinconic, line 223: We removed ACT from POST /chunks, but GET /chunks/{address} and HEAD /chunks/{address} are still wrapped with s.actDecryptionHandler() in pkg/api/router.go, and chunkGetHandler / hasChunkHandler still have the context address resolution:

```go
if v := getAddressFromContext(r.Context()); !v.IsZero() {
    address = v
}
```
Additionally, openapi/Swarm.yaml still lists Swarm-Act-Timestamp, Swarm-Act-Publisher, and Swarm-Act-History-Address on GET /chunks/{address} and HEAD /chunks/{address}.

If ACT is fundamentally unsupported on raw chunks because they are unencrypted and low-level, shouldn't we remove ACT middleware and parameters from GET and HEAD /chunks/{address} as well? Otherwise, we have an asymmetrical API that accepts ACT on download/head but not upload, without any test coverage.

openapi/Swarm.yaml
martinconic, line 1455: Under GET /chunks/{address} and HEAD /chunks/{address}, the ACT parameters are still present:

SwarmActTimestamp
SwarmActPublisher
SwarmActHistoryAddress
If we are cleaning up ACT from the chunks API, these should be removed from the OpenAPI spec as well (along with removing actDecryptionHandler from router.go).

martinconic (changes requested):

openapi/Swarm.yaml
martinconic: You are not removing the entire feature, just the things that are not used.

pkg/api/chunk.go
akrem-chabchoub: @martinconic The ACT headers remain on `GET`/`HEAD /chunks/{address}` **by design**. These requests go through `actDecryptionHandler`, the same middleware used by `GET`/`HEAD /bytes/{address}` and `/bzz/{address}/{path}`.

This middleware is required to resolve ACT wrapped addresses to their actual addresses when the content was encrypted during upload through `POST /bytes` or `POST /bzz`.

Accessing chunks from such encrypted uploads through the low level chunk API is still a valid use case, so the decryption path must remain.

openapi/Swarm.yaml
akrem-chabchoub: @martinconic Same as above:
The ACT headers remain on `GET`/`HEAD /chunks/{address}` **by design**. These requests go through `actDecryptionHandler`, the same middleware used by `GET`/`HEAD /bytes/{address}` and `/bzz/{address}/{path}`.

This middleware is required to resolve ACT wrapped addresses to their actual addresses when the content was encrypted during upload through `POST /bytes` or `POST /bzz`.

Accessing chunks from such encrypted uploads through the low level chunk API is still a valid use case, so the decryption path must remain.

nugaon (approved):
LGTM

gacevicljubisa (approved):

martinconic (approved):

Issue #5563, opened 2026-08-13
Feed lookups have no metrics


darkobas2:
There are no metrics for feeds. `pkg/feeds` has no `metrics.go`, `feedGetHandler` in `pkg/api/feed.go` is uninstrumented, and the `/feeds/{owner}/{topic}` route is mounted without the metric middleware that `/bzz` and `/bytes` use.

Effect: on a production node there is no way to measure real feed lookup volume, failure rate or latency — feed traffic can't be told apart from ordinary chunk retrieval.

A feed lookup histogram would cover it, ideally with labels for lookup type (sequence/epoch) and result (found / not_found / error).

Synthetic beekeeper checks give some coverage today, but they only exercise test traffic and can't tell us anything about real users.

PR #5585, opened 2026-08-30
feat: add feed metrics

sbackend123:
### Checklist

### Description
Add Prometheus metrics for feed lookups at the factory Lookup.At boundary so both GET /feeds and feed-manifest resolution via /bzz are covered.

Align OpenAPI with actual HTTP feed behavior: only sequence is supported.

### Related Issue (Optional)
https://github.com/ethersphere/bee/issues/5563

### AI Disclosure


darkobas2:
Good addition — feed lookups have been something we haven't measured well. A few things stand out, and one of them has real data behind it.

**The buckets don't match what actually happens.** I looked at `beekeeper_check_feed_v2_feed_retrieval_duration_seconds` over the last 30 days (83,865 samples) and got this:

| ≤ | cumulative |
|---|---|
| 1.0s | 0% |
| 2.0s | 5.7% |
| 2.5s | 22.1% |
| 3.0s | 44.6% |
| 5.0s | 98.9% |
| 10.0s | 99.99% |

Eight of the eleven buckets you proposed sit at or below 0.25s, but this data shows nothing there. The real action is between 2–5s — that's where almost everything lives — and you only have `2.5` and `5`. The median is around 3s, which falls between two buckets that are four seconds apart. That means `histogram_quantile` will guess badly right where you need it to be accurate. The 10s ceiling is fine — only 11 samples in 30 days went over it, so pushing higher doesn't help.

One thing to keep in mind: beekeeper measures the full HTTP path from outside and always does a fresh lookup, while `Lookup.At` skips the network client and can use cached chunks. So having a couple of sub-second buckets makes sense — the real gap is 1–5s. Something like `0.1, 0.25, 0.5, 1, 1.5, 2, 2.5, 3, 4, 5, 7.5, 10` would cover both cases.

**`context.DeadlineExceeded` gets lumped into `error`.** You give `Cancelled` its own bucket but not timeouts. When a retrieval is slow, a timeout is operationally different from an actual failure — that's exactly the split you want to see when your graph goes red.

**The OpenAPI change should be its own PR.** You narrowed `FeedType` from `pattern: ^(sequence|epoch)$` to `enum: [sequence]`, which is accurate — `pkg/api/feed.go:75` hardcodes `feeds.Sequence` and `FeedType` only appears in those two `/feeds` parameters. But it's not related to metrics. More importantly, it tightens the published contract: a client that sends `type=epoch` and validates against the spec now gets rejected. A patch bump from 8.1.1 to 8.1.2 is light for a schema narrowing like that.

**No tests.** `lookupResult` is pure and works well with table tests. The code that wraps it has no coverage right now.

**Small things**
- `lookup_started_total` nearly duplicates `lookup_duration_seconds_count`. The only thing it adds is started-minus-completed to show in-flight requests — either explain that in the `Help` text or switch to a gauge.
- `err != nil && errors.Is(err, context.Canceled)` — the first check is unnecessary, `errors.Is(nil, ...)` already returns false.

**I checked these and they're fine:** `ch == nil` is safe here because `sequence.go:194` returns a literal `nil`, no risk of a typed-nil interface trap. Only `node.go` calls `factory.New`, so no duplicate registration. `Metrics()` using reflection and the `metrics.Collector` assertion match the existing `batchStore` pattern. The `{type}` label is useful — `pkg/api/bzz.go:873` pulls the type from a manifest's `swarm-feed-type` metadata, so `epoch` can reach you through `/bzz` even though `/feeds` never sends it.

PR #5586, opened 2026-08-31
fix: postage listener hot retry loop

### LINKED ISSUE #5560: postage listener: backoff is skipped while paging, so a failing eth_getLogs becomes a hot retry loop (~1,650 req/s)
## Observed

Bee 2.7.1, pointed at a Gnosis RPC endpoint that returns an error for `eth_getLogs`, made **123,761 failed requests in 75 seconds** — roughly 1,650 per second — and wrote a 400,000-line log in the process. It did not back off at any point.

```
"level"="warning" "logger"="node/listener" "msg"="could not get blockchain log"
  "error"="method 'eth_getLogs' cannot be served verified right now"
```

…repeated 123,761 times, several per millisecond.

## Mechanism

In `pkg/postage/listener/listener.go`, the loop's only sleep is guarded by `!paged`:

```go
272    var expectedWaitTime time.Duration
273    if lastConfirmedBlock != 0 {
…
277    } else {
278        expectedWaitTime = l.backoffTime
279    }
280
281    if !paged {
282        l.logger.Debug("sleeping until next block batch", "duration", expectedWaitTime)
283        select {
284        case <-time.After(expectedWaitTime):
…
288    }
289    paged = false
```

The `FilterLogs` failure path sets `lastConfirmedBlock = 0` and continues:

```go
events, err := l.ev.FilterLogs(ctx, l.filterQuery(...))
if err != nil {
    …
    l.logger.Warning("could not get blockchain log", "error", err)
    lastConfirmedBlock = 0
    continue
}
```

Setting `lastConfirmedBlock = 0` correctly selects `l.backoffTime` at line 278 — but that wait is never reached. A node that is catching up sets `paged = true` earlier in the same iteration:

```go
if to-from >= pageSize {
    paged = true
    to = from + pageSize - 1
}
```

So the sequence repeats indefinitely: `paged = true` → `FilterLogs` fails → `continue` → `!paged` is false → **sleep skipped** → `paged` set true again. The backoff is computed on every iteration and never applied.

This only bites while paging, which is exactly when a node is furthest behind and has the most work to do — a fresh node syncing from the PostageStamp start block is paging for millions of blocks.

## Impact

- Against a public RPC provider: near-instant rate-limiting or a ban, which then looks like an unrelated failure.
- Against a self-hosted endpoint — a light client, or a node on a phone — it is a denial of service from your own software.
- On metered or mobile connections: sustained request volume with no ceiling.
- Log growth of ~5,300 lines/second at warning level.

## Suggested fix

Reset `paged` in the error paths so the existing backoff is honoured:

```go
if err != nil {
    …
    lastConfirmedBlock = 0
    paged = false          // let the backoff at line 281 apply
    continue
}
```

The same applies to the `BlockNumber` error path just above, which has the identical shape. Worth considering alongside:

- exponential backoff with a cap, rather than a fixed `backoffTime`, since a failing endpoint rarely recovers within one interval
- rate-limiting the warning (log once, then a periodic count) so a persistent backend failure doesn't bury the rest of the log

## Reproduction

Point Bee at any RPC endpoint that errors on `eth_getLogs` but answers `eth_chainId` and `eth_blockNumber` normally, so the node starts and begins paging:

```
bee start --data-dir=<tmp> --password=<pw> \
          --blockchain-rpc-endpoint=<endpoint> \
          --full-node=false --verbosity=debug
```

## Environment

Bee 2.7.1-61fab37b, Linux x86_64, Gnosis (chainID 100, networkID 1), `--full-node=false`. The endpoint was a Myotis light client, which serves every other method Bee needs but does not yet index logs for Gnosis — an ordinary "backend can't answer this one method" case.

aloknerurkar:
### Checklist

### Description
The postage listener enters a hot loop when a backend RPC call (BlockNumber or FilterLogs) fails while it is paging through historical blocks.

paged is only reset to false right before a successful fetch begins, but both error paths continue before reaching that reset. Since the backoff wait is gated on !paged, a failure during paging causes every subsequent retry to skip expectedWaitTime/backoffTime entirely, hammering the RPC endpoint with no delay.

Fix: reset paged = false on both error paths (BlockNumber and FilterLogs), so a failed call during paging still waits the configured backoff before retrying, matching normal (non-paged) error behavior.

Changes

- pkg/postage/listener/listener.go: reset paged = false before continue on both RPC error branches.
- pkg/postage/listener/listener_test.go: added TestListenerBackoffAfterPagedError — a regression test asserting at least backoffTime elapses between failed paged calls and the subsequent retry. Verified it fails without the fix (~54µs gap) and passes with it.

Testing

- go build ./...
- go vet ./pkg/postage/...
- go test -race ./pkg/postage/listener/...

### Related Issue (Optional)
Fixes ethersphere/bee#5560

### AI Disclosure

### DIFF
--- a/pkg/postage/listener/listener.go
+++ b/pkg/postage/listener/listener.go
@@ -302,6 +302,7 @@ func (l *listener) Listen(ctx context.Context, from uint64, updater postage.Even
 				l.metrics.BackendErrors.Inc()
 				l.logger.Warning("could not get block number", "error", err)
 				lastConfirmedBlock = 0
+				paged = false
 				continue
 			}
 
@@ -339,6 +340,7 @@ func (l *listener) Listen(ctx context.Context, from uint64, updater postage.Even
 				l.metrics.BackendErrors.Inc()
 				l.logger.Warning("could not get blockchain log", "error", err)
 				lastConfirmedBlock = 0
+				paged = false
 				continue
 			}
 
--- a/pkg/postage/listener/listener_test.go
+++ b/pkg/postage/listener/listener_test.go
@@ -481,6 +481,89 @@ func TestListenerPageSize(t *testing.T) {
 	})
 }
 
+// TestListenerBackoffAfterPagedError verifies that a failed FilterLogs call
+// while paging still causes the listener to wait for the backoff duration
+// before retrying, instead of hot-looping (regression test for the paged
+// flag not being reset on the error path).
+func TestListenerBackoffAfterPagedError(t *testing.T) {
+	t.Parallel()
+
+	const testBackoff = 300 * time.Millisecond
+
+	f := &backoffFilterer{done: make(chan struct{})}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	l := listener.New(
+		nil,
+		log.Noop,
+		f,
+		postageStampContractAddress,
+		postageStampContractABI,
+		1,
+		stallingTimeout,
+		testBackoff,
+	)
+	testutil.CleanupCloser(t, l)
+
+	go l.Listen(ctx, 0, newEventUpdaterMock())
+
+	select {
+	case <-f.done:
+	case <-time.After(5 * time.Second):
+		t.Fatal("timed out waiting for FilterLogs calls")
+	}
+
+	f.mu.Lock()
+	calls := append([]time.Time(nil), f.calls...)
+	f.mu.Unlock()
+
+	if len(calls) < 3 {
+		t.Fatalf("expected at least 3 FilterLogs calls, got %d", len(calls))
+	}
+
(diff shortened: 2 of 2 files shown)


sbackend123 (approved):

martinconic (approved):

pkg/postage/listener/listener.go
```diff
@@ -302,6 +302,7 @@ func (l *listener) Listen(ctx context.Context, from uint64, updater postage.Even
 				l.metrics.BackendErrors.Inc()
 				l.logger.Warning("could not get block number", "error", err)
 				lastConfirmedBlock = 0
+				paged = false
```
martinconic, line 305: This is redundant as we have it set to false at line 289. Maybe just keep it as a double assurance.

martinconic (approved):

acud (approved):

PR #5591, opened 2026-09-02
fix: flaky test by checking ctx cancellation before racing I/O in protobuf reader/writer

akrem-chabchoub:
### Checklist

### Description
Fixes flaky [`TestIncoming_ContextTimeout`](https://github.com/ethersphere/bee/actions/runs/33604392092/job/100168667421?pr=5590)

`WriteMsgWithContext` and `ReadMsgWithContext` could behave randomly when the context was already canceled.

They started the I/O operation and waited to see whether the I/O finished or the context was canceled. If both happened at the same time, the program could choose either result.

This was especially easy to reproduce with the in-memory `streamtest` stream, where I/O could finish immediately. As a result, `TestIncoming_ContextTimeout` sometimes passed and sometimes failed in CI.

The fix is to check `ctx.Err()` before starting the I/O. If the context is already canceled or expired, the function now immediately returns the context error every time.

### AI Disclosure

### DIFF
--- a/pkg/p2p/libp2p/stream_test.go
+++ b/pkg/p2p/libp2p/stream_test.go
@@ -79,6 +79,7 @@ func TestStreamVersion(t *testing.T) {
 
 			if v == nil {
 				t.Fatal("expected version to be non-nil")
+				return
 			}
 
 			expected := semver.Version{Major: tc.wantMajor, Minor: tc.wantMinor}
--- a/pkg/p2p/protobuf/protobuf.go
+++ b/pkg/p2p/protobuf/protobuf.go
@@ -56,6 +56,10 @@ func newReader(r ggio.Reader) Reader {
 }
 
 func (r Reader) ReadMsgWithContext(ctx context.Context, msg proto.Message) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
 	errChan := make(chan error, 1)
 	go func() {
 		errChan <- r.ReadMsg(msg)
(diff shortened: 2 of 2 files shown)


sbackend123 (approved):

pkg/p2p/libp2p/stream_test.go
```diff
@@ -79,6 +79,7 @@ func TestStreamVersion(t *testing.T) {
 
 			if v == nil {
 				t.Fatal("expected version to be non-nil")
+				return
```
acud, line 82: why do you need to return here? this is unreachable code. from `go doc testing.T`:
```
    A test ends when its Test function returns or calls any of the methods
    T.FailNow, T.Fatal, T.Fatalf, T.SkipNow, T.Skip, or T.Skipf. Those methods,
    as well as the T.Parallel method, must be called only from the goroutine
    running the Test function.
```

acud (approved):

pkg/p2p/libp2p/stream_test.go
akrem-chabchoub: Sorry I merged without seeing your comment.
Linter failed and prevented me from push:
https://github.com/ethersphere/bee/pull/5591/changes/7a40425263ef4387f22e0c6e6c5f69f90a5089e0

image

akrem-chabchoub: I pull master into this branch and run `make lint` and it works without any issues, although nothing in master is referring to some lint changes, this is wired, didn't know why!

I reverted the changes here: https://github.com/ethersphere/bee/pull/5598

PR #5593, opened 2026-09-02
fix(gsoc): client too slow handling

nugaon:
addressing https://github.com/ethersphere/bee/pull/5497#issuecomment-5506663115

### DIFF
--- a/pkg/api/gsoc.go
+++ b/pkg/api/gsoc.go
@@ -171,7 +171,10 @@ func (s *Service) gsocListeningWs(conn *websocket.Conn, socAddress swarm.Address
 	defer s.wsWg.Done()
 
 	var (
-		dataC    = make(chan []byte, 2) // small buffer to decouple producer/consumer
+		// Buffered enough to absorb a legitimate burst of concurrently delivered
+		// GSOC messages (e.g. several chunks pushed to this address at once)
+		// without tripping the slow-consumer detection below.
+		dataC    = make(chan []byte, 16)
 		gone     = make(chan struct{})
 		slow     = make(chan struct{})
 		slowOnce sync.Once
--- a/pkg/api/gsoc_test.go
+++ b/pkg/api/gsoc_test.go
@@ -290,7 +290,7 @@ func TestGsocWebsocketInvalidFieldsHeader(t *testing.T) {
 func TestGsocWebsocketSlowConsumer(t *testing.T) {
 	t.Parallel()
 
-	const messageCount = 10
+	const messageCount = 32 // exceeds dataC's buffer so the overflow is hit deterministically
 
 	var (
 		id         = make([]byte, 32)
--- a/pkg/api/gsoc_test.go
+++ b/pkg/api/gsoc_test.go
@@ -338,7 +338,7 @@ func TestGsocWebsocketSlowConsumer(t *testing.T) {
 	}
 	testutil.CleanupCloser(t, cl)
 
-	// never read from cl, so the dataC buffer (cap 2) fills up almost
+	// never read from cl, so the dataC buffer (cap 16) fills up almost
 	// immediately: the first message blocks the single writer goroutine
 	// (nothing reads the pipe), and the next ones queue up and overflow.
 	for i := range messageCount {
--- a/pkg/gsoc/gsoc.go
+++ b/pkg/gsoc/gsoc.go
@@ -73,9 +73,7 @@ func (l *listener) Handle(c *soc.SOC) {
 	l.logger.Debug("new incoming GSOC message", "GSOC Address", addr, "wrapped chunk address", c.WrappedChunk().Address())
 
 	for _, hh := range h {
-		go func(hh Handler) {
-			hh(c)
-		}(*hh)
+		(*hh)(c)
 	}
 }
 
--- a/pkg/gsoc/gsoc_test.go
+++ b/pkg/gsoc/gsoc_test.go
@@ -6,7 +6,6 @@ package gsoc_test
 
 import (
 	"testing"
-	"time"
 
 	"github.com/ethersphere/bee/v2/pkg/cac"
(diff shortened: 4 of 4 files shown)


akrem-chabchoub:
@nugaon 
Seems the issue still persist ? (CI and beekeeper)
https://github.com/ethersphere/bee/actions/runs/33647834352/job/100307336280?pr=5593
https://github.com/ethersphere/bee/actions/runs/33647834069/job/100306984511?pr=5593

nugaon:
Thanks for flagging — the ordering fix wasn't the full story. The remaining flakiness in the "parallel" ci-gsoc case (10 messages delivered concurrently to one subscriber) was a false-positive "slow consumer" disconnect: the websocket delivery channel only buffered 2 messages, so a legitimate burst tripped the overflow check meant for genuinely stuck clients. Bumped the buffer to 16 in 0be41223 (`TestGsocWebsocketSlowConsumer` adjusted to still hit real overflow deterministically). Let's see if CI is green now.

pkg/gsoc/gsoc.go
```diff
@@ -73,9 +73,7 @@ func (l *listener) Handle(c *soc.SOC) {
 	l.logger.Debug("new incoming GSOC message", "GSOC Address", addr, "wrapped chunk address", c.WrappedChunk().Address())
 
 	for _, hh := range h {
-		go func(hh Handler) {
-			hh(c)
-		}(*hh)
+		(*hh)(c)
```
akrem-chabchoub, line 76: Could calling subscribers synchronously here delay pushsync or pullsync ?

pkg/api/gsoc_test.go
akrem-chabchoub: This comment needs to be updated from cap 2 to cap 16 to match the new dataC capacity.

akrem-chabchoub (changes requested):

pkg/api/gsoc.go
```diff
@@ -171,7 +171,10 @@ func (s *Service) gsocListeningWs(conn *websocket.Conn, socAddress swarm.Address
 	defer s.wsWg.Done()
 
 	var (
-		dataC    = make(chan []byte, 2) // small buffer to decouple producer/consumer
+		// Buffered enough to absorb a legitimate burst of concurrently delivered
+		// GSOC messages (e.g. several chunks pushed to this address at once)
+		// without tripping the slow-consumer detection below.
+		dataC    = make(chan []byte, 16)
```
aloknerurkar, line 177: Not sure why this is needed in this PR. Is there any reasoning behind why 16 is the right value to be used here? If there is truly a burst > 16 we will still trip the slow-consumer detection below. Even the previous value of 2 is questionable.

If this is an actual problem, the correct fix would be something like a time-windowed rate check, or a slightly larger drain-then-reconsider window, or just accepting that a truly overwhelmed consumer should be disconnected and tuning based on expected real-world message rates than using some arbitrary constant. Having this change in this PR feels like the value is made up so the tests pass. It would be better to not have this in the current PR and making a proper fix in a subsequent PR.

pkg/gsoc/gsoc.go
aloknerurkar, line 76: Handle itself can still be invoked concurrently by multiple peer-connection goroutines in pushsync/pullsync for the same GSOC address (different peers forwarding the same or different chunks around the same time) — sequential dispatch only guarantees ordering within one Handle call, not across concurrent calls from different peers. If true ordering matters end-to-end, that's a separate, larger synchronization question this doesn't address.

gacevicljubisa:
@nugaon I pushed `fix/gsoc-dispatch-safety` on top of this branch if you want to take it.

pkg/api/gsoc.go
nugaon: The time-windowed rate check wouldn't be the best fit since the data feed is sporadic.
I don't know what would be the optimal value, wdyt?
we can remove the whole condition as well if you are fine with that.

pkg/gsoc/gsoc.go
nugaon: yeah, I see.
well, the order is not that much of importance, since GSOC does not guarantee any order anyway.

pkg/api/gsoc.go
nugaon: anyway, I think 16 is a fine number, ofc the test scenario a bit unreal but in case of a really popular gsoc topic it can happen 16 messages arrive the same time. ofc, even more popular topic would get maybe 32...

pkg/gsoc/gsoc.go
alokn-mobius: In which case the tests are wrong to expect them in order?

nugaon: There is no order in GSOC.

nugaon: I meant with [this part](https://github.com/ethersphere/SWIPs/pull/41/changes#diff-7116210bdf3a2610eb4e1428061c378913bac8450e132f81a3371f1be4ca18beR152) that even Graffiti Feed couldn't keep canonical order, so is GSOC.

nugaon: okay, order can happen for one client if the updates' lengths are more than pull-sync time.
in a smaller network, maybe push-sync time.

pkg/api/gsoc.go
martinconic: Please check my comment from the previous PR: https://github.com/ethersphere/bee/pull/5497#discussion_r3821495864

acud:
@nugaon we've reverted the original PR that introduces this issue since this is now stopping a 5 people engineering team from working. please reopen the original PR with the relevant fix and we will review this again. thanks for the understanding

nugaon:
@acud, the _correct_ way would have been to _review_ this PR after I addressed all comments and exchanged messages privately to facilitate your _engineer_ work.

I hardly understand this decision... I have also work to do.

PR #5594, opened 2026-09-03
fix: stewardship re-upload dispersed replicas

### LINKED ISSUE #5450: Stewardship Reupload Does Not Re-create Dispersed Replicas for Root Chunk
### Context

- **Bee version**: v2.7.2-rc1
- **Network**: mainnet / testnet
- **Relevant flags**: `--full-node`

### Summary

When calling `PUT /stewardship/{reference}` with a non-zero `Swarm-Redundancy-Level` header, the stewardship service traverses and re-uploads all BMT trie chunks (data shards + RS parity shards) but silently skips re-creating the **dispersed replicas** for the root chunk. Dispersed replicas are SOC-wrapped copies of the root chunk spread across different Kademlia neighborhoods to provide resilience when the root chunk's own neighborhood degrades. After a reupload, these replicas remain absent or expired, leaving the content more vulnerable to retrieval failure than the requested redundancy level implies.

### Expected behavior

A `PUT /stewardship/{reference}` call with `Swarm-Redundancy-Level: N` (where N > 0) should:

1. Re-stamp and re-push all BMT trie chunks (data + RS parity) — already done.
2. Re-create and push the `2^N` dispersed replica SOC chunks for the root chunk to their target neighborhoods, consistent with what `replicas.NewPutter` does at original upload time.

This ensures the full redundancy guarantee for the given level is restored, not just the erasure-coded intermediate structure.

### Actual behavior

Only the BMT trie chunks are re-uploaded. The dispersed replicas are not re-created.

Inspecting the implementation in [`pkg/steward/steward.go`](https://github.com/ethersphere/bee/blob/master/pkg/steward/steward.go):

```go
if err := s.traverser.Traverse(ctx, root, fn, rLevel); err != nil { ... }
return uploaderSession.Done(root)
```

`traversal.Traverse` → `joiner.IterateChunkAddresses` only walks the BMT trie. Dispersed replicas are SOC chunks outside the trie, created by [`replicas.NewPutter`](https://github.com/ethersphere/bee/blob/master/pkg/replicas/putter.go), and are never visited nor re-uploaded.

Calling the API:
```bash
curl -X PUT http://localhost:1633/stewardship/<reference> \
  -H "Swarm-Postage-Batch-Id: <batch-id>" \
  -H "Swarm-Redundancy-Level: 4"
```

Returns `200 OK` with no error, but the dispersed replicas for the root chunk are not re-pushed to the network.

### Steps to reproduce

1. Upload a file with redundancy enabled:
   ```bash
   curl -X POST http://localhost:1633/bzz \
     -H "Swarm-Postage-Batch-Id: <batch-id>" \
     -H "Swarm-Redundancy-Level: 4" \
     -H "Content-Type: application/octet-stream" \
     --data-binary @file.bin
   # note the returned <reference>
   ```
2. Wait for dispersed replica stamps to expire, or simulate their loss.
3. Call reupload:
   ```bash
   curl -X PUT http://localhost:1633/stewardship/<reference> \
     -H "Swarm-Postage-Batch-Id: <new-batch-id>" \
     -H "Swarm-Redundancy-Level: 4"
   ```
4. Verify no dispersed replicas (SOC chunks with well-known owner `swarm.ReplicasOwner`) have been pushed to the network by checking pushsync logs at `level=trace`.

### Possible solution

After the traversal completes successfully, fetch the root chunk from local storage (already cached from traversal) and pass it through `replicas.NewPutter` with the requested `rLevel`:

The root chunk is already in local storage at this point (fetched during traversal), so no extra network retrieval is required.

### AI Disclosure
- [x] This issue contains suggestions and text generated by an LLM.
- [x] I have reviewed the AI generated content thoroughly.
- [x] I possess the technical expertise to responsibly review the AI generated content mentioned in this issue.

aloknerurkar:
### Checklist

### Description
PUT /stewardship/{reference} with a non-zero Swarm-Redundancy-Level re-uploads all BMT trie chunks (data + RS parity) but never re-creates the root chunk's dispersed replicas — the SOC-wrapped copies spread across neighborhoods for retrieval resilience. traversal.Traverse only walks the trie; replicas are created separately by replicas.NewPutter at original upload time and are invisible to traversal.

Fix: after traversal succeeds, if rLevel != redundancy.NONE, fetch the root chunk (already local from traversal), re-stamp it, and push it through replicas.NewPutter(uploaderSession, rLevel) before calling uploaderSession.Done(root) — mirroring what the original upload path does.

### Changes

- pkg/steward/steward.go: added the post-traversal replica re-upload step in Reupload.
- pkg/steward/steward_test.go: updated TestSteward to account for the extra replica traffic on the same push channel (was previously silently absent), and added an assertion that the expected number of dispersed replicas (redundancy.PARANOID.GetReplicaCount()) are actually re-uploaded and retrievable.

### Testing

- go build ./...
- go vet ./pkg/steward/...
- go test -race ./pkg/steward/... -count=5 — deterministic pass
- golangci-lint run ./pkg/steward/...### Open API Spec Version Changes (if applicable)

### Related Issue (Optional)
Fixes ethersphere/bee#5450

### AI Disclosure


pkg/steward/steward.go
```diff
@@ -77,6 +78,22 @@ func (s *steward) Reupload(ctx context.Context, root swarm.Address, stamper post
 		)
 	}
 
+	if rLevel != redundancy.NONE {
+		rootChunk, err := getter.Get(ctx, root)
```
gacevicljubisa: Dispersed replicas are keyed on the 32 byte content address, but root here can be an encrypted reference, which carries the decryption key in its trailing 32 bytes, so it should be trimmed before the lookup.

pkg/steward/steward.go
```diff
@@ -77,6 +78,22 @@ func (s *steward) Reupload(ctx context.Context, root swarm.Address, stamper post
...
+		if err != nil {
+			return errors.Join(fmt.Errorf("get root chunk for dispersed replicas: %w", err), uploaderSession.Cleanup())
+		}
+
+		stamp, err := stamper.Stamp(rootChunk.Address(), rootChunk.Address())
+		if err != nil {
+			return errors.Join(fmt.Errorf("stamping root chunk for dispersed replicas: %w", err), uploaderSession.Cleanup())
+		}
+
+		if err := replicas.NewPutter(uploaderSession, rLevel).Put(ctx, rootChunk.WithStamp(stamp)); err != nil {
```
gacevicljubisa: We should cehck if rootChank is valid content addressed chunk. Also, the replicas need to be stamped, not the root chunk.

gacevicljubisa (changes requested):

PR #5596, opened 2026-09-03
refactor: node mode two regimes

martinconic:
### Checklist

### Description

Follow-up to #5431, addressing the upgrade-path review from @gacevicljubisa. Stacked on `refactor/node-mode-config`.

The legacy branch in `resolveNodeMode` applied some of the new semantics (strict validation, swap forced on) but not all (no defaults restored for light nodes), so existing configs broke in both directions. This PR replaces it with two regimes selected by one switch: whether `node-mode` is set.

**`node-mode` unset (legacy)**: behaves exactly as releases before `node-mode` existed.
- `chequebook-enable` and `storage-incentives-enable` fall back to their former default of `true` when not set. Both remain gated in `NewBee` (chequebook on swap, incentives on full mode), so nothing starts that the previous release would not have started.
- Mode is inferred as before: `full-node: true` gives full, otherwise an RPC endpoint gives light, otherwise ultra-light.
- No new validation. The only check kept is that a full node has an RPC endpoint; previous releases enabled the chain backend for every full node and failed at chain init without one, so this changes the message, not the outcome.
- One deprecation warning names the inferred mode and the removal release (`v2.11.0`, a constant in `start.go`).

**`node-mode` set**: the mode owns the config.
- `full` requires an RPC endpoint and implies `swap-enable`, `chequebook-enable` and `storage-incentives-enable` unless explicitly disabled. An explicit opt-out is honoured and logged as a warning. Bootnodes are exempt from the implied options because `NewBee` never starts swap, push-sync or the incentives agent for them.
- `light` requires an RPC endpoint. Swap and chequebook stay off unless set. `storage-incentives-enable: true` is rejected.
- `ultra-light` rejects `swap-enable: true` and `storage-incentives-enable: true`. An RPC endpoint is ignored.
- `chequebook-enable: true` with `swap-enable: false` is rejected in every mode. Implied values never produce this combination, so it is always an explicit contradiction.

`node-mode: full` is now sufficient on its own. `validateFullMode` and the `--full-node` restore loop are gone.

The `node-mode` flag default is now empty instead of `ultra-light`, so an unset value selects the legacy regime without relying on viper's `IsSet` semantics for pflag defaults, and `--help` no longer shows a default that does not describe what happens.

**Packaging**: `BEE_NODE_MODE` added to the docker-compose passthrough and the env template, `BEE_FULL_NODE` marked deprecated, stale `true` defaults for `BEE_CHEQUEBOOK_ENABLE` and `BEE_STORAGE_INCENTIVES_ENABLE` corrected, bee.yaml variants document the implied options and the deprecated inference.

### Behaviour matrix

| Config | Regime | Result |
|---|---|---|
| RPC + `swap-enable: true`, nothing else | legacy | light, chequebook **true** (was silently lost) |
| `chequebook-enable: true`, swap unset | legacy | starts, ultra-light (was hard exit) |
| `full-node: true` + `bootnode-mode: true` + incentives false | legacy | full, starts (was hard exit) |
| `full-node: true` + RPC only | legacy | full, swap false, chequebook true, incentives true (master verbatim) |
| `full-node: true`, no RPC | legacy | error, as before but earlier |
| `node-mode: full` + RPC only | new | full, swap/chequebook/incentives all true |
| `node-mode: full` + `storage-incentives-enable: false` | new | full, incentives false, warning |
| `node-mode: full` + `swap-enable: false` | new | full, swap and chequebook false, incentives true |
| `node-mode: full` + `bootnode-mode: true` + RPC | new | full, nothing implied |
| `node-mode: light` + RPC | new | light, swap false |
| `node-mode: light` + `chequebook-enable: true`, swap unset | new | error |
| `node-mode: ultra-light` + `swap-enable: true` | new | error |
| `node-mode: full`, no RPC | new | error |

### Test plan

- [x] `TestResolveNodeMode` covers every row above through `viper.New()`.
- [x] `TestResolveNodeModeWithBoundFlags` runs the resolver against the real bound `start` flags, so the empty `node-mode` default selects the legacy regime and the sub-option flag defaults do not count as set.
- [x] `make lint` (0 issues), `go vet`, full `go test ./...` pass.
- [x] Binary smoke tests with `dist/bee start` against an unreachable RPC endpoint (reaching chain init proves resolution passed):
  - legacy, no flags: starts as ultra-light, one deprecation warning naming the mode and `v2.11.0`
  - legacy, `chequebook-enable: true` without swap: starts as ultra-light (was a hard exit)
  - legacy, RPC + `swap-enable: true`: light, chequebook restored, reaches chain init
  - legacy, `full-node` + `bootnode-mode` + `storage-incentives-enable: false`: full, reaches chain init (was a hard exit)
  - legacy, `full-node` without RPC: clear error
  - `node-mode: full` + RPC only: swap, chequebook, incentives implied (debug log), reaches chain init
  - `node-mode: full` + `storage-incentives-enable: false`: opt-out honoured with a warning
  - `node-mode: full` without RPC, `light` + chequebook without swap, `ultra-light` + swap: rejected with clear errors
- [ ] Beekeeper `local.yaml` still uses `full-node`; it keeps working via the legacy regime and migrates to `node-mode` in a follow-up beekeeper PR.

### AI Disclosure


cmd/bee/cmd/start.go
```diff
@@ -373,113 +373,130 @@ func buildBeeNode(ctx context.Context, c *command, cmd *cobra.Command, logger lo
...
-			optionNameChequebookEnable,
-			optionNameStorageIncentivesEnable,
-		} {
-			if !c.config.IsSet(key) {
-				logger.Warning("enabling option implied by legacy --full-node; set it explicitly or use --node-mode=full", "option", key)
-				c.config.Set(key, true)
-			}
-		}
+	modeStr := c.config.GetString(optionNameNodeMode)
+	if modeStr == "" {
```
sbackend123, line 398: Seems, this differs from the documented distinction based on whether node-mode is set. Should we use IsSet to distinguish an omitted value from an explicitly empty one?

packaging/docker/env
```diff
@@ -37,8 +37,8 @@
 # BEE_CACHE_CAPACITY=1000000
 ## enable forwarded content caching (default true)
 # BEE_CACHE_RETRIEVAL=true
-## enable chequebook (default true)
-# BEE_CHEQUEBOOK_ENABLE=true
+## enable chequebook; requires swap; implied by BEE_NODE_MODE=full (default false)
```
sbackend123, line 40: When BEE_NODE_MODE is unset, the legacy resolver restores chequebook-enable to true. Could we clarify that false applies only when BEE_NODE_MODE is explicitly set?

sbackend123 (approved):
LGTM

cmd/bee/cmd/cmd.go
```diff
@@ -357,11 +358,15 @@ func (c *command) setAllFlags(cmd *cobra.Command) {
 	cmd.Flags().String(optionNameSwapFactoryAddress, "", "swap factory addresses")
 	cmd.Flags().String(optionNameBzzTokenAddress, "", "bzz token contract address")
 	cmd.Flags().String(optionNameSwapInitialDeposit, "0", "initial deposit if deploying a new chequebook")
+	cmd.Flags().String(optionNameNodeMode, "", "node operational mode: full, light, or ultra-light (unset: inferred from deprecated full-node and blockchain-rpc-endpoint)")
 	cmd.Flags().Bool(optionNameSwapEnable, false, "enable swap")
-	cmd.Flags().Bool(optionNameChequebookEnable, true, "enable chequebook")
+	cmd.Flags().Bool(optionNameChequebookEnable, false, "enable chequebook (requires swap-enable)")
```
gacevicljubisa: chequebook-enable is changed?

cmd/bee/cmd/cmd.go
```diff
@@ -378,7 +383,7 @@ func (c *command) setAllFlags(cmd *cobra.Command) {
 	cmd.Flags().Bool(optionNamePProfMutex, false, "enable pprof mutex profile")
 	cmd.Flags().StringSlice(optionNameStaticNodes, []string{}, "protect nodes from getting kicked out on bootnode")
 	cmd.Flags().Bool(optionNameAllowPrivateCIDRs, false, "allow to advertise private CIDRs to the public network")
-	cmd.Flags().Bool(optionNameStorageIncentivesEnable, true, "enable storage incentives feature")
+	cmd.Flags().Bool(optionNameStorageIncentivesEnable, false, "enable storage incentives feature (full node only)")
```
gacevicljubisa: storage-incentives flag is also changed?

gacevicljubisa (changes requested):
We should have backwards comaptibile bahviour, and only if new flag is explicitly set, then we may change it.

gacevicljubisa (approved):

